PR #24315, opened 2026-07-20
airoha: fix SCU regmap mapping for pcie nodes

YaleiZang:
Add the 'airoha,scu' property referencing '&scuclk' to pcie0, pcie1, and pcie2 nodes in the device tree. This update fixes the following error during PCIe controller initialization:

    mtk-pcie-gen3 1fc00000.pcie: error -ENODEV: failed to map SCU regmap

By correctly specifying the SCU clock reference, the PCIe controllers are able to map the SCU regmap and initialize properly.

To help keep the codebase consistent and readable,
and to help people review your contribution,
we ask you to follow the rules you find in the wiki at this link
https://openwrt.org/submitting-patches

Please remove this message before posting the pull request.

### DIFF
--- a/target/linux/airoha/dts/an7581.dtsi
+++ b/target/linux/airoha/dts/an7581.dtsi
@@ -784,6 +784,8 @@
 
 			mediatek,pbus-csr = <&pbus_csr 0x0 0x4>;
 
+			airoha,scu = <&scuclk>;
+
 			interrupts = <GIC_SPI 39 IRQ_TYPE_LEVEL_HIGH>;
 			bus-range = <0x00 0xff>;
 			#interrupt-cells = <1>;
@@ -825,6 +827,8 @@
 
 			mediatek,pbus-csr = <&pbus_csr 0x8 0xc>;
 
+			airoha,scu = <&scuclk>;
+
 			interrupts = <GIC_SPI 40 IRQ_TYPE_LEVEL_HIGH>;
 			bus-range = <0x00 0xff>;
 			#interrupt-cells = <1>;
@@ -865,6 +869,8 @@
 			reset-names = "phy-lane2", "perstout";
 
 			mediatek,pbus-csr = <&pbus_csr 0x10 0x14>;
+
+			airoha,scu = <&scuclk>;
 
 			interrupts = <GIC_SPI 41 IRQ_TYPE_LEVEL_HIGH>;
 			bus-range = <0x00 0xff>;


danielschwierzeck:
According to commit 665cfa98de1aa02b31436621a063969aba322a18 this is now set in board DTS. But setting this in AN7581 DTSI seems to be more logical, @Ansuel?

Ansuel:
the scu is needed only with 2 lanes mode is used... where are you getting that error?

target/linux/airoha/dts/an7581.dtsi
```diff
@@ -784,6 +784,8 @@
 
 			mediatek,pbus-csr = <&pbus_csr 0x0 0x4>;
 
+			airoha,scu = <&scuclk>;
```
openwrt-ai, line 787: The driver only consumes `airoha,scu`, and only emits the `failed to map SCU regmap` error you quote, when a node has `num-lanes = <2>` — see [the `num_lanes == 2` lookup in 609-04-PCI-mediatek-gen3-Add-2-lanes-mode-support-for-Airoh.patch:112-115](``https://github.com/openwrt/openwrt/blob/6bb6740e389a689b58037bcf57f9f7c2783a235f/target/linux/airoha/patches-6.18/609-04-PCI-mediatek-gen3-Add-2-lanes-mode-support-for-Airoh.patch#L112-L115).`` The three base nodes here (pcie0/1/2) do not set `num-lanes`, so they default to 1 lane and the property is inert on them as written.

The boards that actually enable 2-lane mode already carry `airoha,scu`: `an7581-evb-emmc-eagle.dts`, `an7581-nokia-valyrian.dts` and `an7581-w1700k-ubi.dts` all set `num-lanes = <2>` plus `airoha,scu = <&scuclk>;`, added in commit 665cfa98de1a ("airoha: add missing SCU phandle for PCIe 2-lanes mode"). If this DTSI change lands, those three per-board `airoha,scu` lines become redundant duplicates and should be dropped in the same PR to avoid two sources of truth.

Given @Ansuel's and @danielschwierzeck's open question on this thread about where the error actually comes from and whether the DTSI is the right home: is the intent to move the property to the base DTSI (and remove the per-board copies), or is a specific 1-lane board hitting this? As-is the commit adds the property in a place the driver never reads it in the default 1-lane configuration.

---

openwrt-ai (commented):
Single commit, DTS-only change; the commit message matches the diff, so no commit-check issues. One inline note on the scope of the added property. CI shows no failures at head.

---

danielschwierzeck:
> the scu is needed only with 2 lanes mode is used... where are you getting that error?

I also have this error on my board with Kite since the PCI-E reset rework, without SCU on pcie1 the WiFi cannot start. But adding the scu property to pcie0 and pcie1 in my board DTS fixed the problem.

YaleiZang:
> needed

oh yes, sorry I miss the change in dts, Please ignore this commit. I will close it. thanks

PR #18631, opened 2025-04-29
mediatek: filogic: add support Netcore NX30V2/N30PRO/W7/GW3001

bfdeh:
Hardware specification:
  SoC: MediaTek MT7981B
  Flash: ESMT F50L1G41LB 128MB
  RAM: ESMT M15T2G16128A DDR3 256MB
  Ethernet: 3x 1G
  Button: Reset, WPS

Installing OpenWrt:
  - Setup a tftp server on your PC. Copy xxx-preloader.bin, xxx-bl31-uboot.fip and xxx-initramfs.itb to tftp root directory.
  - Connect to the router via ssh or telnet, username: useradmin, password is the web login password of the router.
  - Backup all critical flash partitions with the following commands where x.x.x.x is the IP of your PC.
```
    IP=x.x.x.x
    cd /dev
    for d in /sys/class/mtd/mtd?; do
      if [ "$(cat $d/name)" = "BL2" ]; then
        tftp -l $(basename $d) -r bl2.img -p $IP
      elif [ "$(cat $d/name)" = "FIP" ]; then
        tftp -l $(basename $d) -r fip.bin -p $IP
      elif [ "$(cat $d/name)" = "Factory" ]; then
        tftp -l $(basename $d) -r factory.bin -p $IP
      fi
    done
    for d in /sys/devices/virtual/ubi/ubi0/ubi0_*; do
      [ "$(cat $d/name)" != "customer" ] && continue
      tftp -l $(basename $d) -r customer -p $IP
      break
```
  - Flash with the following commands:
```
     cd /tmp
     tftp -r xxx-preloader.bin -g x.x.x.x
     tftp -r xxx-bl31-uboot.fip -g x.x.x.x
     mtd write xxx-preloader.bin spi0.0
     mtd write xxx-bl31-uboot.fip FIP
     mtd erase ubi
```
  - Set a static ip(192.168.1.86) for your PC. And then reboot the router. It will run initramfs image automatically.
  - After openwrt boots up, perform sysupgrade via web UI.

Reverting to the vendor firmware:
  - Setup a tftp server on your PC with ip address 192.168.1.86. And make sure bl2.img, fip.bin, factory.bin and customer are located in tftp root directory.
  - Power off the router.
  - Press and hold WPS key, then power on the router.
  - Release WPS key, when internet/wifi/wps leds are blinking.
  - Wait until internet/wifi/wps leds light up, power off the router.
  - Press and hold reset key, power up the router, release reset key 15s later.
[...]

PR #24148, opened 2026-07-08
mediatek: use libdeflate for gzip

neheb:
Shrinks size slightly.

ping @dangowrt

PR #24089, opened 2026-07-05
mediatek: filogic: add support for netis N6 V2

unisequence:
netis N6 V2 is a MT7981B-based router, hardware-compatible with netis NX32U (confirmed via bootloader cross-flash). GPIO mappings not covered by the vendor DTS were verified against NX32U and via live link/carrier testing on switch ports, plus direct LED-toggle confirmation on the physical device.

Hardware:
SoC: MediaTek MT7981B (Filogic 820)
RAM: 256 MiB DDR3 (ESMT M15T2G16128A-AZR1-EFB, DDR3-1866)
Flash: 128 MiB SPI-NAND
Ethernet: 4x 1GbE (mt7531 switch: wan, lan1, lan2, lan3),
2.5GbE internal CPU-switch link (gmac0, 2500base-x)
WiFi: MT7981 integrated 2.4GHz + MT7976C 5GHz radio,
802.11ax, 2x2 MU-MIMO, HE160 supported (confirmed live;
vendor firmware ships with HE80 for the AX1800 rating)
USB: 1x USB 3.0 (vbus GPIO 23, active high)
Buttons: mesh/rfkill (GPIO 0), reset (GPIO 1)
LEDs: power (GPIO 4), status (GPIO 5), wan (GPIO 9),
wlan 2.4GHz (GPIO 34, phy0tpt trigger),
wlan 5GHz (GPIO 35, phy1tpt trigger),
usb (GPIO 13, usbport trigger)
all active low
UART: 115200n8, uart0

MAC addresses:

| Interface | Source | Offset | Address (this unit) |
|---|---|---|---|
| WAN | Factory mac-base | 0x1fef26 | 30:07:5c:dc:7a:d2 |
| LAN/CPU | Factory mac-base | 0x1fef20 | 30:07:5c:dc:7a:d0 (label) |
| WiFi 2.4GHz | radio calibration (auto) | -- | 30:07:5c:dc:7a:d1 |
| WiFi 5GHz | radio calibration (auto) | -- | b2:07:5c:dc:7a:d1 |

LAN address matches the label on the device enclosure. WAN and LAN now derive from separate switch ports (port@0 for WAN, CPU port@6 for LAN), rather than sharing a single MAC as in an earlier revision of this patch. WiFi MACs are not set explicitly in DTS -- mt76 derives them from the radio's own calibration data, confirmed live to match the stock firmware value.

The included U-Boot defenv for this board also avoids a silent failure in ubi_write_production/ubi_write_recovery: 
[...]

PR #24316, opened 2026-07-20
mediatek: filogic: add support Netcore N30 Pro/NX35U

bfdeh:
mediatek: filogic: add support Netcore N30 Pro/NX35U

Additional info for N30 Pro:
https://github.com/openwrt/openwrt/pull/18631#discussion_r3340370190

- Hardware specification:
  SoC: MediaTek MT7981B
  Flash: 128MB
  RAM: DDR3 512MB
  Ethernet: 5x 1G
  Button: Reset, WPS
  USB: 1x USB3.0

- MAC addresses
  | Interface   | Source                   | Offset   | Address (this unit)       |
  | ----------- | ------------------------ | -------- | ------------------------- |
  | WAN         | Factory mac-base         | 0x1fef26 | 08:64:2A:D7:82:11         |
  | LAN/CPU     | Factory mac-base         | 0x1fef20 | 08:64:2A:D7:82:10 (label) |
  | WiFi 2.4GHz | radio calibration (auto) | --       | 08:64:2A:D7:82:12         |
  | WiFi 5GHz   | radio calibration (auto) | --       | 8A:64:2A:D7:82:12         |

- Official LED layout, from left to right:
  power internet wps wifi wan lan usb

- Redefinition for OpenWrt:
  power, wifi and lan: unchanged
  internet: used for WAN RX/TX indication
  wps: used for led-boot, led-failsafe, led-running and led-upgrade
  wan: used for WAN link indication

- Installing OpenWrt:
  - Setup a tftp server on your PC. Copy xxx-preloader.bin,
    xxx-bl31-uboot.fip and xxx-initramfs.itb to tftp root directory.
  - Connect to the router via ssh or telnet, username: useradmin,
    password is the web login password of the router.
  - Backup all critical flash partitions with the following commands
    where x.x.x.x is the IP of your PC.

    ```shell
    IP=x.x.x.x
    cd /dev
    for d in /sys/class/mtd/mtd?; do
        if [ "$(cat $d/name)" = "BL2" ]; then
            tftp -l $(basename $d) -r bl2.img -p $IP
        elif [ "$(cat $d/name)" = "FIP" ]; then
            tftp -l $(basename $d) -r fip.bin -p $IP
        elif [ "$(cat $d/name)" = "Factory" ]; then
            tftp -l $(basename $d) -r factory.bin -p $IP
        fi
    done
    for d in /sys/devices/virtual/ubi/ubi0/ubi0_*; do
        [ "$(cat $d/name)" != "customer" ] && continue
        tftp -l $(basename $d) -r customer -p $IP
        break
    done
    ```

  - Flash with the following commands:

    ```shell
    cd /tmp
    tftp -r xxx-preloader.bin -g x.x.x.x
    tftp -r xxx-bl31-uboot.fip -g x.x.x.x
    mtd write xxx-preloader.bin spi0.0
    mtd write xxx-bl31-uboot.fip FIP
    mtd erase ubi
    ```

  - Set a static ip(192.168.1.254) for your PC. And then reboot
    the router. It will run initramfs image automatically.
  - After openwrt boots up, perform sysupgrade via web UI.

- Reverting to the vendor firmware:
  - Setup a tftp server on your PC with ip address 192.168.1.254.
    And make sure bl2.img, fip.bin, factory.bin and customer are
    located in tftp root directory.
  - Power off the router.
  - Press and hold WPS key, then power on the router.
  - Release WPS key, when internet/wifi/wps leds are blinking.
  - Wait until internet/wifi/wps leds light up, power off the router.
  - Press and hold reset key, power up the router, release reset key
    15s later.
  - Connect to http://192.168.1.1/, now you can upload vendor .bin
    firmware.

- Uboot netconsole:
  Uboot netconsole can be enabled by WPS or reset key.

  - Setup a linux PC with ip 192.168.1.254.
    Open a new terminal and execute
    'stty -isig -echo cbreak; nc -lup 6666'
  - Press and hold WPS(or reset) key, then power on the router.
  - Release key once internet/wifi/wps leds are all on.
    DON'T hold the key more than 5s after internet/wifi/wps leds on,
    or it will try to revert to vendor firmware.
  - 5s later, uboot bootmenu will show on the terminal.

### DIFF
--- a/target/linux/mediatek/dts/mt7981b-netcore-n30-pro.dts
+++ b/target/linux/mediatek/dts/mt7981b-netcore-n30-pro.dts
@@ -1,0 +1,113 @@
+/* SPDX-License-Identifier: (GPL-2.0-only OR MIT) */
+
+/dts-v1/;
+#include "mt7981b-netis-common.dtsi"
+
+/ {
+	model = "Netcore N30 Pro";
+	compatible = "netcore,n30-pro", "mediatek,mt7981";
+
+	aliases {
+		label-mac-device = &gmac0;
+		led-boot = &led_wps;
+		led-failsafe = &led_wps;
+		led-running = &led_wps;
+		led-upgrade = &led_wps;
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		led_wps: wps {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WPS;
+			gpios = <&pio 5 GPIO_ACTIVE_LOW>;
+		};
+
+		wan {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WAN;
+			gpios = <&pio 8 GPIO_ACTIVE_LOW>;
+		};
+
+		internet {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WAN_ONLINE;
+			gpios = <&pio 9 GPIO_ACTIVE_LOW>;
+		};
+
+		usb {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_USB;
+			gpios = <&pio 13 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "usbport";
+			trigger-sources = <&usb3_port1>, <&usb3_port2>;
+		};
+
+		wifi2g {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WLAN_2GHZ;
+			gpios = <&pio 34 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "phy0tpt";
+		};
+
+		wifi5g {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WLAN_5GHZ;
+			gpios = <&pio 35 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "phy1tpt";
+		};
+	};
+
+	reg_5v: regulator-5v {
+		compatible = "regulator-fixed";
+		regulator-name = "fixed-5V";
+		regulator-min-microvolt = <5000000>;
+		regulator-max-microvolt = <5000000>;
+		gpio = <&pio 23 GPIO_ACTIVE_HIGH>;
+		enable-active-high;
+	};
+};
+
+&switch {
+	ports {
+		port@0 {
+			reg = <0>;
+			label = "lan1";
+		};
+
+		port@1 {
+			reg = <1>;
+			label = "lan2";
+		};
+
+		port@2 {
+			reg = <2>;
+			label = "lan3";
+		};
+
+		port@3 {
+			reg = <3>;
+			label = "lan4";
+		};
+	};
+};
+
+&usb_phy {
+	status = "okay";
+};
+
+&xhci {
+	vbus-supply = <&reg_5v>;
+	status = "okay";
+
+	usb3_port1: port@1 {
+		reg = <1>;
+		#trigger-source-cells = <0>;
+	};
+
+	usb3_port2: port@2 {
+		reg = <2>;
+		#trigger-source-cells = <0>;
+	};
+};
--- a/target/linux/mediatek/filogic/base-files/etc/board.d/01_leds
+++ b/target/linux/mediatek/filogic/base-files/etc/board.d/01_leds
@@ -183,6 +183,10 @@ mercusys,mr90x-v1-ubi)
 	ucidef_set_led_netdev "lan-2" "lan-2" "green:lan-2" "lan2" "link tx rx"
 	ucidef_set_led_netdev "wan" "wan" "green:wan" "eth1" "link tx rx"
 	;;
+netcore,n30-pro)
+	ucidef_set_led_netdev "wanact" "WANACT" "green:wan-online" "eth1" "tx rx"
+	ucidef_set_led_netdev "wanlink" "WANLINK" "green:wan" "eth1" "link"
+	;;
 netcore,n60)
 	ucidef_set_led_netdev "wan" "WAN" "mdio-bus:06:green:wan" "eth1" "link tx rx"
 	;;
--- a/target/linux/mediatek/filogic/base-files/lib/upgrade/platform.sh
+++ b/target/linux/mediatek/filogic/base-files/lib/upgrade/platform.sh
@@ -175,6 +175,7 @@ platform_do_upgrade() {
 	netis,nx32u|\
 	nokia,ea0326gmp|\
 	openwrt,one|\
+	netcore,n30-pro|\
 	netcore,n60|\
 	netcore,n60-pro|\
 	qihoo,360t7|\
--- a/target/linux/mediatek/image/filogic.mk
+++ b/target/linux/mediatek/image/filogic.mk
@@ -2693,6 +2693,31 @@ define Device/mercusys_mr90x-v1-ubi
 endef
 TARGET_DEVICES += mercusys_mr90x-v1-ubi
 
+define Device/netcore_n30-pro
+  DEVICE_VENDOR := Netcore
+  DEVICE_MODEL := N30 Pro
+  DEVICE_ALT0_VENDOR := netis
+  DEVICE_ALT0_MODEL := NX35U
+  DEVICE_DTS := mt7981b-netcore-n30-pro
+  DEVICE_DTS_DIR := ../dts
+  DEVICE_DTC_FLAGS := --pad 4096
+  DEVICE_DTS_LOADADDR := 0x43f00000
+  DEVICE_PACKAGES := kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware kmod-usb3 kmod-usb-ledtrig-usbport
+  KERNEL_LOADADDR := 0x44000000
+  KERNEL := kernel-bin | libdeflate-gzip
+  KERNEL_INITRAMFS := kernel-bin | lzma | fit lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb with-initrd | pad-to 64k
+  KERNEL_INITRAMFS_SUFFIX := .itb
+  KERNEL_IN_UBI := 1
+  UBOOTENV_IN_UBI := 1
+  IMAGES := sysupgrade.itb
+  IMAGE_SIZE := 117248k
+  IMAGE/sysupgrade.itb := append-kernel | fit gzip $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb external-with-rootfs | pad-rootfs | append-metadata
+  ARTIFACTS := spim-nand-preloader.bin spim-nand-bl31-uboot.fip
+  ARTIFACT/spim-nand-preloader.bin	:= mt7981-bl2 spim-nand-ddr3
+  ARTIFACT/spim-nand-bl31-uboot.fip	:= mt7981-bl31-uboot netcore_n30-pro
+endef
+TARGET_DEVICES += netcore_n30-pro
+
 define Device/netcore_n60
   DEVICE_VENDOR := Netcore
   DEVICE_MODEL := N60
@@ -2879,14 +2904,12 @@ define Device/netis_nx30v2
   DEVICE_VARIANT := V2
   DEVICE_ALT0_VENDOR := Netcore
   DEVICE_ALT0_MODEL := POWER 30AX
-  DEVICE_ALT1_VENDOR := Netcore
-  DEVICE_ALT1_MODEL := N30 Pro
-  DEVICE_ALT2_VENDOR := GWBN
-  DEVICE_ALT2_MODEL := GW3001
-  DEVICE_ALT3_VENDOR := GLC
-  DEVICE_ALT3_MODEL := W7
-  DEVICE_ALT4_VENDOR := netis
-  DEVICE_ALT4_MODEL := MEX605
+  DEVICE_ALT1_VENDOR := GWBN
+  DEVICE_ALT1_MODEL := GW3001
+  DEVICE_ALT2_VENDOR := GLC
+  DEVICE_ALT2_MODEL := W7
+  DEVICE_ALT3_VENDOR := netis
+  DEVICE_ALT3_MODEL := MEX605
   DEVICE_DTS := mt7981b-netis-nx30v2
   DEVICE_DTS_DIR := ../dts
   DEVICE_DTC_FLAGS := --pad 4096
--- a/package/boot/uboot-mediatek/Makefile
+++ b/package/boot/uboot-mediatek/Makefile
@@ -556,6 +556,18 @@ define U-Boot/mt7981_mercusys_mr85x
   DEPENDS:=+trusted-firmware-a-mt7981-spim-nand-ubi-ddr3-1866
 endef
 
+define U-Boot/mt7981_netcore_n30-pro
+  NAME:=Netcore N30 Pro
+  BUILD_SUBTARGET:=filogic
+  BUILD_DEVICES:=netcore_n30-pro
+  UBOOT_CONFIG:=mt7981_netcore_n30-pro
+  UBOOT_IMAGE:=u-boot.fip
+  BL2_BOOTDEV:=spim-nand
+  BL2_SOC:=mt7981
+  BL2_DDRTYPE:=ddr3
+  DEPENDS:=+trusted-firmware-a-mt7981-spim-nand-ddr3
+endef
+
 define U-Boot/mt7981_netis_eap930-v1
   NAME:=netis EAP930 V1
   BUILD_SUBTARGET:=filogic
@@ -1363,6 +1375,7 @@ UBOOT_TARGETS := \
 	mt7981_jcg_q30-pro \
 	mt7981_konka_komi-a31 \
 	mt7981_mercusys_mr85x \
+	mt7981_netcore_n30-pro \
 	mt7981_netis_eap930-v1 \
 	mt7981_netis_n6-v2 \
 	mt7981_netis_nx30v2 \
--- a/package/boot/uboot-mediatek/patches/468-add-netcore-n30-pro.patch
+++ b/package/boot/uboot-mediatek/patches/468-add-netcore-n30-pro.patch
@@ -1,0 +1,360 @@
+--- /dev/null
++++ b/arch/arm/dts/mt7981-netcore-n30-pro.dts
+@@ -0,0 +1,160 @@
++// SPDX-License-Identifier: GPL-2.0
++
++/dts-v1/;
++#include "mt7981.dtsi"
++#include <dt-bindings/gpio/gpio.h>
++#include <dt-bindings/input/input.h>
++
++/ {
++	model = "Netcore N30 Pro";
++	compatible = "netcore,n30-pro", "mediatek,mt7981";
++	chosen {
++		stdout-path = &uart0;
++		tick-timer = &timer0;
++	};
++
++	memory@40000000 {
++		device_type = "memory";
++		reg = <0x40000000 0x20000000>;
++	};
++
++	gpio-keys {
++		compatible = "gpio-keys";
++
++		reset {
++			label = "reset";
++			linux,code = <KEY_RESTART>;
++			gpios = <&pio 1 GPIO_ACTIVE_LOW>;
++		};
++
++		wps {
++			label = "wps";
++			linux,code = <KEY_WPS_BUTTON>;
++			gpios = <&pio 0 GPIO_ACTIVE_LOW>;
++		};
++	};
++
++	leds {
++		compatible = "gpio-leds";
++
++		wps {
++			label = "wps";
++			gpios = <&pio 5 GPIO_ACTIVE_LOW>;
++		};
++
++		wan {
++			label = "wan";
++			gpios = <&pio 8 GPIO_ACTIVE_LOW>;
++		};
++
++		internet {
++			label = "internet";
++			gpios = <&pio 9 GPIO_ACTIVE_LOW>;
++		};
++
++		usb {
++			label = "usb";
++			gpios = <&pio 13 GPIO_ACTIVE_LOW>;
++		};
++
++		wifi {
++			label = "wifi";
++			gpios = <&pio 34 GPIO_ACTIVE_LOW>;
++		};
++
++	};
++
++};
++
++&uart0 {
++	status = "okay";
++};
++
++&eth {
++	status = "okay";
++	mediatek,gmac-id = <0>;
++	phy-mode = "2500base-x";
++	mediatek,switch = "auto";
++	reset-gpios = <&pio 39 GPIO_ACTIVE_HIGH>;
++
++	fixed-link {
++		speed = <2500>;
++		full-duplex;
++	};
++};
++
++&pio {
++	spi_flash_pins: spi0-pins-func-1 {
++		mux {
++			function = "flash";
++			groups = "spi0", "spi0_wp_hold";
++		};
++
++		conf-pu {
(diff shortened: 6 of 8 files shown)


package/boot/uboot-mediatek/patches/468-add-netcore-n30pro.patch
```diff
@@ -0,0 +1,358 @@
+--- /dev/null
++++ b/arch/arm/dts/mt7981-netcore-n30pro.dts
+@@ -0,0 +1,160@@
```
openwrt-ai: This hunk header is missing the space before the closing `@@` (should be `@@ -0,0 +1,160 @@`). `git apply --check` rejects the file with `corrupt patch at line 3`; GNU `patch`/quilt tolerate it so the OpenWrt build still applies it, but it breaks `git am`/`git apply` and is inconsistent with every sibling patch in this directory (e.g. [`469-add-netis-eap930-v1.patch` L3](https://github.com/openwrt/openwrt/blob/e6960ce5f38e5de97e6932d605754033947cd4cf/package/boot/uboot-mediatek/patches/469-add-netis-eap930-v1.patch#L3) uses `@@ -0,0 +1,112 @@`). The other two hunks in this file are already correct. Regenerating with `make package/uboot-mediatek/refresh` would normalize it; or apply the one-character fix directly:

```suggestion
@@ -0,0 +1,160 @@
```


---

openwrt-ai (commented):
Reviewed 1 new commit. One inline comment on a malformed hunk header in the U-Boot patch; commit message is well-formed with all required new-device sections.

---

openwrt-ai (commented):
Reviewed 1 new commit; no new issues found. The U-Boot patch hunk headers flagged in the previous review are now correct.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -2567,6 +2567,31 @@ define Device/mercusys_mr90x-v1-ubi
...
+define Device/netcore_n30pro
+  DEVICE_VENDOR := Netcore
+  DEVICE_MODEL := N30PRO
+  DEVICE_ALT0_VENDOR := netis
+  DEVICE_ALT0_MODEL := NX35U
+  DEVICE_DTS := mt7981b-netcore-n30pro
+  DEVICE_DTS_DIR := ../dts
+  DEVICE_DTC_FLAGS := --pad 4096
+  DEVICE_DTS_LOADADDR := 0x43f00000
+  DEVICE_PACKAGES := kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware
```
1715173329: `kmod-usb3` should be added here to enable usb support

bfdeh: done, thanks.

openwrt-ai (commented):
## Commit checks

- `6aadc5a` "mediatek: filogic: add support Netcore N30PRO/NX35U" — a stray `aaa` line is appended after the `Signed-off-by:` trailer. It looks like accidental leftover text; please remove it so the commit message ends cleanly at the sign-off.

The only code change since the last review (`kmod-usb3` added to `DEVICE_PACKAGES`) looks correct for the device's USB 3.0 port.

---

openwrt-ai (commented):
Reviewed 1 new commit; no new issues found. The stray `aaa` line after the `Signed-off-by` trailer flagged in the previous review has been removed; there are no code changes since the last review.

---

DH-41:
Hello, thanks for your work on this PR!

Could you please fix the naming convention to match existing devices? For example, the already merged N60 Pro uses `N60 Pro` for the model name and `n60-pro` for the target/slug.

According to the Netcore official website (e.g., [Netcore N60 Pro Product Page](https://www.netcoretec.com/product-center/family/wifi-router/2659.html) and [Netcore Product List](https://www.netcoretec.com/product-center/family/wifi-router)), the official model names use a space, such as `N60 Pro` and `N30 Pro` (instead of N30PRO).

Could we update N30PRO / n30pro to `N30 Pro` / `n30-pro` for consistency?

package/boot/uboot-mediatek/patches/468-add-netcore-n30pro.patch
```diff
@@ -0,0 +1,358 @@
...
++CONFIG_ENV_USE_DEFAULT_ENV_TEXT_FILE=y
++CONFIG_ENV_DEFAULT_ENV_TEXT_FILE="defenvs/netcore_n30pro.env"
++CONFIG_ENV_VARS_UBOOT_RUNTIME_CONFIG=y
++CONFIG_NETCONSOLE=y
++CONFIG_USE_IPADDR=y
++CONFIG_IPADDR="192.168.1.1"
++CONFIG_USE_SERVERIP=y
++CONFIG_SERVERIP="192.168.1.254"
++CONFIG_NET_RANDOM_ETHADDR=y
++CONFIG_REGMAP=y
```
unisequence, line 250: I don't see `CONFIG_RNG_MTK_V2` or `CONFIG_DM_RNG` in this defconfig — recent MT7981 boards typically enable these for KASLR seeding (see PR #24089). Intentional, or just missing?

target/linux/mediatek/dts/mt7981b-netcore-n30pro.dts
```diff
@@ -0,0 +1,100 @@
...
+
+	aliases {
+		label-mac-device = &gmac0;
+		led-boot = &led_wps;
+		led-failsafe = &led_wps;
+		led-running = &led_wps;
+		led-upgrade = &led_wps;
+	};
+
+	leds {
```
unisequence, line 18: Nit: the `leds { }` child nodes aren't ordered by ascending GPIO (9, 5, 34, 35, 8, 13). Would you mind sorting by pin number for consistency with other filogic boards? (I see the current order follows the physical left-to-right layout from the PR description, which is also a reasonable choice — just flagging the convention.)

target/linux/mediatek/image/filogic.mk
```diff
@@ -2567,6 +2567,31 @@ define Device/mercusys_mr90x-v1-ubi
...
+  DEVICE_MODEL := N30PRO
+  DEVICE_ALT0_VENDOR := netis
+  DEVICE_ALT0_MODEL := NX35U
+  DEVICE_DTS := mt7981b-netcore-n30pro
+  DEVICE_DTS_DIR := ../dts
+  DEVICE_DTC_FLAGS := --pad 4096
+  DEVICE_DTS_LOADADDR := 0x43f00000
+  DEVICE_PACKAGES := kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware kmod-usb3
+  KERNEL_LOADADDR := 0x44000000
+  KERNEL := kernel-bin | gzip
```
unisequence, line 2581: Consider `libdeflate-gzip` instead of `gzip` here — drop-in binary compatible, better compression ratio (see issue #24148).

```suggestion
  KERNEL := kernel-bin | libdeflate-gzip
```

unisequence:
Not diff-related, but: the MAC address derivation in the description is written as prose ("WAN: LAN + 1", "WiFi: LAN + 2"). Since #24089 it's become common to lay this out as an Interface/Source/Offset/Address table for readability/consistency — worth converting?

bfdeh:
> Not diff-related, but: the MAC address derivation in the description is written as prose ("WAN: LAN + 1", "WiFi: LAN + 2"). Since #24089 it's become common to lay this out as an Interface/Source/Offset/Address table for readability/consistency — worth converting?

This description ("WAN: LAN + 1", "WiFi: LAN + 2") isn't about how OpenWrt converts MAC addresses, but rather how the vendor assigns them to each interface.

target/linux/mediatek/image/filogic.mk
bfdeh: All done. thanks for your suggestions and review.

target/linux/mediatek/filogic/base-files/lib/upgrade/platform.sh
```diff
@@ -174,6 +174,7 @@ platform_do_upgrade() {
 	netis,nx32u|\
 	nokia,ea0326gmp|\
 	openwrt,one|\
+	netcore,n30-pro|\
```
openwrt-ai, line 178: nit: this entry lands after `openwrt,one`, which breaks the alphabetical order of the case list — sorted, it belongs between `mercusys,mr90x-v1-ubi` and `netis,eap930-v1`. It is following the pre-existing misplaced `netcore,n60`/`netcore,n60-pro` block, so the deviation is inherited rather than introduced here, but the same commit places the board correctly in the three other lists it touches: [mediatek_filogic:45](https://github.com/openwrt/openwrt/blob/7acfc077ede33915477cf9384777aa3a406c9376/package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic#L45), [01_leds:190](https://github.com/openwrt/openwrt/blob/7acfc077ede33915477cf9384777aa3a406c9376/target/linux/mediatek/filogic/base-files/etc/board.d/01_leds#L190) and [02_network:86](``https://github.com/openwrt/openwrt/blob/7acfc077ede33915477cf9384777aa3a406c9376/target/linux/mediatek/filogic/base-files/etc/board.d/02_network#L86).`` Inserting it in the sorted position (and optionally moving the two `netcore,n60*` lines with it) would keep this file consistent.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -2642,6 +2642,31 @@ define Device/mercusys_mr90x-v1-ubi
 endef
 TARGET_DEVICES += mercusys_mr90x-v1-ubi
 
+define Device/netcore_n30-pro
+  DEVICE_VENDOR := Netcore
+  DEVICE_MODEL := N30 Pro
```
openwrt-ai, line 2698: nit: the rename to `N30 Pro` / `n30-pro` is complete and consistent in-tree (Makefile, DTS, board.d, envtools, U-Boot config/env), but the PR title and the first line of the PR description still say `add support Netcore N30PRO/NX35U`, while the commit subject now reads `add support Netcore N30 Pro/NX35U`. Worth updating the PR title/body so it matches the model name that actually gets merged.

---

openwrt-ai (commented):
Reviewed 1 new commit (rename `n30pro` -> `n30-pro` / `N30PRO` -> `N30 Pro`, LED nodes re-sorted by ascending GPIO, `CONFIG_DM_RNG`/`CONFIG_RNG_MTK_V2` added to the U-Boot defconfig, `gzip` -> `libdeflate-gzip`).

The rename is complete and consistent across the Makefile, DTS, `board.d`, uboot-envtools and the U-Boot patch (defconfig name, `CONFIG_DEFAULT_DEVICE_TREE`, `defenvs/netcore_n30-pro.env`, artifact file names); the three patch hunk headers still match their line counts after the two added defconfig lines (160/127/64). `CONFIG_RNG_MTK_V2` is a real symbol (added by `181-rng-mtk-rng-v2-*.patch`) and mt7981.dtsi gets the matching RNG node from `193-mediatek-mt7981-add-RNG-node.patch`, so the addition is sound. Only two naming/ordering nits inline.

---

bfdeh:
> Not diff-related, but: the MAC address derivation in the description is written as prose ("WAN: LAN + 1", "WiFi: LAN + 2"). Since #24089 it's become common to lay this out as an Interface/Source/Offset/Address table for readability/consistency — worth converting?

Got it, the table is added.

unisequence:
> Got it, the table is added.

Looking good

openwrt-ai (commented):
Reviewed 1 new commit. The tree is byte-identical to the previously reviewed `7acfc07` — the only change is the commit message, where the MAC section was converted from prose to the Interface/Source/Offset/Address table requested in [#24316 (comment)](https://github.com/openwrt/openwrt/pull/24316#issuecomment-5174650325).

The table's contents check out against the DTS: `0x1fef20` / `0x1fef26` match [`macaddr_lan` and `macaddr_wan` in mt7981b-netis-common.dtsi:140-146](https://github.com/openwrt/openwrt/blob/48d9829d102c27d952e49303421435aef6394a7a/target/linux/mediatek/dts/mt7981b-netis-common.dtsi#L140-L146), the `(label)` marker on LAN matches `label-mac-device = &gmac0` (gmac0 consumes `macaddr_lan`), and the WiFi rows correctly point at `eeprom_factory_0` rather than a dedicated MAC cell. `Flash: 128MB` / `RAM: DDR3 512MB` / `Ethernet: 5x 1G` are likewise consistent with `IMAGE_SIZE := 117248k`, the `reg = <0 0x40000000 0 0x20000000>` memory node, and the four `lan*` switch ports plus WAN.

The PR title and body now read `N30 Pro/NX35U`, so the naming nit from the previous review is resolved. The `platform.sh` case-list ordering nit from that review is still open and not repeated here.

## Commit checks

- `48d9829` "mediatek: filogic: add support Netcore N30 Pro/NX35U" — nit: the new MAC table prints the author's unit's addresses in full (`08:64:2A:D7:82:10`, `…:11`, `…:12`, `8A:64:2A:D7:82:12`). Convention is to mask the unit-specific bytes, e.g. `08:64:2A:xx:xx:x0` / `…:x1` / `…:x2`, as in [`mediatek: add support for Cudy WR3000U v1`](https://github.com/openwrt/openwrt/commit/b9d7b448c256bb3860c7965ccf599c29c4847cb4). The offsets and sources — the parts that are actually useful to a reader — are unaffected by the masking. (Both forms do appear in tree history, so this is a preference, not a blocker.)

---

target/linux/mediatek/dts/mt7981b-netcore-n30-pro.dts
```diff
@@ -0,0 +1,101 @@
...
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WAN_ONLINE;
+			gpios = <&pio 9 GPIO_ACTIVE_LOW>;
+		};
+
+		usb {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_USB;
+			gpios = <&pio 13 GPIO_ACTIVE_LOW>;
+		};
```
fildunsky, line 45: The board.d `usbport` config below can't work without `kmod-usb-ledtrig-usbport` (see my comment in filogic.mk). netis NX32U wires this LED in the DTS instead — suggest doing the same here:

```suggestion
		usb {
			color = <LED_COLOR_ID_GREEN>;
			function = LED_FUNCTION_USB;
			gpios = <&pio 13 GPIO_ACTIVE_LOW>;
			linux,default-trigger = "usbport";
			trigger-sources = <&usb3_port1>, <&usb3_port2>;
		};
```

Needs the labeled ports in `&xhci`, see the suggestion there.

target/linux/mediatek/dts/mt7981b-netcore-n30-pro.dts
```diff
@@ -0,0 +1,101 @@
...
+};
+
+&usb_phy {
+	status = "okay";
+};
+
+&xhci {
+	vbus-supply = <&reg_5v>;
+	status = "okay";
+};
```
fildunsky, line 113:
```suggestion
&xhci {
	vbus-supply = <&reg_5v>;
	status = "okay";

	usb3_port1: port@1 {
		reg = <1>;
		#trigger-source-cells = <0>;
	};

	usb3_port2: port@2 {
		reg = <2>;
		#trigger-source-cells = <0>;
	};
};
```

target/linux/mediatek/filogic/base-files/etc/board.d/01_leds
```diff
@@ -187,6 +187,11 @@ mercusys,mr90x-v1-ubi)
 	ucidef_set_led_netdev "lan-2" "lan-2" "green:lan-2" "lan2" "link tx rx"
 	ucidef_set_led_netdev "wan" "wan" "green:wan" "eth1" "link tx rx"
 	;;
+netcore,n30-pro)
+	ucidef_set_led_netdev "wanact" "WANACT" "green:wan-online" "eth1" "tx rx"
+	ucidef_set_led_netdev "wanlink" "WANLINK" "green:wan" "eth1" "link"
+	ucidef_set_led_usbport "usb" "USB" "green:usb" "usb1-port1" "usb2-port1"
```
fildunsky: With the DT `usbport` trigger this line isn't needed:

```suggestion
```

target/linux/mediatek/image/filogic.mk
```diff
@@ -2642,6 +2642,31 @@ define Device/mercusys_mr90x-v1-ubi
...
+  DEVICE_ALT0_VENDOR := netis
+  DEVICE_ALT0_MODEL := NX35U
+  DEVICE_DTS := mt7981b-netcore-n30-pro
+  DEVICE_DTS_DIR := ../dts
+  DEVICE_DTC_FLAGS := --pad 4096
+  DEVICE_DTS_LOADADDR := 0x43f00000
+  DEVICE_PACKAGES := kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware kmod-usb3
```
fildunsky: The `usbport` trigger (whether set from board.d or DT) lives in `kmod-usb-ledtrig-usbport`, which isn't shipped — same package set as `netis_nx32u`:

```suggestion
  DEVICE_PACKAGES := kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware \
	kmod-usb3 kmod-usb-ledtrig-usbport
```

fildunsky (commented):
Tested this on a Netcore N30 Pro unit — thanks for splitting the device out, ethernet/WAN/buttons/LEDs all work as described.

One functional issue: the USB LED never lights, because the `usbport` trigger is set up via board.d but `kmod-usb-ledtrig-usbport` is not in `DEVICE_PACKAGES`, so the trigger doesn't exist in the default image. I'd suggest wiring the LED in the DTS like on netis NX32U (which shares this shell) — see the inline suggestions. With these changes the USB LED lights up when a USB device is plugged in and goes off on removal (verified on hardware).

target/linux/mediatek/image/filogic.mk
bfdeh: The USB LED issue is fixed. Thanks.

Wfarmer1:
Found a subtle issue while testing a locally built image from this PR (2bdd8ca, kernel 6.18.44): the WiFi LEDs' `linux,default-trigger = "phy0tpt"/"phy1tpt"` never takes effect on a mainline-default build. `MAC80211_LEDS` is only injected when the `CONFIG_LEDS_TRIGGERS` build symbol is set (`config-$(CONFIG_LEDS_TRIGGERS) += MAC80211_LEDS` in package/kernel/mac80211/Makefile), and a default `make defconfig` does not contain that symbol, so the tpt triggers are never registered and the LEDs stay off.

Verified on hardware:
- `cat /sys/class/leds/green:wlan-*/trigger` lists the mt76 triggers (`phy0rx/tx/assoc/radio`) but no tpt triggers
- `echo 255 > brightness` lights both LEDs fine, so the GPIO mapping is correct

Suggestion: keep the traffic-blink semantics but make the LEDs work out of the box by adding netdev-based uci defaults in board.d/01_leds (same pattern as the WAN LEDs), e.g. `ucidef_set_led_netdev` for `green:wlan-2ghz` on `phy0-ap0` and `green:wlan-5ghz` on `phy1-ap0` with tx+rx mode; the `phy*tpt` default-triggers in the DTS can then be dropped.

openwrt-ai (commented):

---

target/linux/mediatek/image/filogic.mk
openwrt-ai: fixed, thanks

---

bfdeh:
> Found a subtle issue while testing a locally built image from this PR ([2bdd8ca](https://github.com/openwrt/openwrt/commit/2bdd8ca40ccde8413d416a9b9f1d71690ccd8180), kernel 6.18.44): the WiFi LEDs' `linux,default-trigger = "phy0tpt"/"phy1tpt"` never takes effect on a mainline-default build. `MAC80211_LEDS` is only injected when the `CONFIG_LEDS_TRIGGERS` build symbol is set (`config-$(CONFIG_LEDS_TRIGGERS) += MAC80211_LEDS` in package/kernel/mac80211/Makefile), and a default `make defconfig` does not contain that symbol, so the tpt triggers are never registered and the LEDs stay off.
> 
> Verified on hardware:
> 
> * `cat /sys/class/leds/green:wlan-*/trigger` lists the mt76 triggers (`phy0rx/tx/assoc/radio`) but no tpt triggers
> * `echo 255 > brightness` lights both LEDs fine, so the GPIO mapping is correct
> 
> Suggestion: keep the traffic-blink semantics but make the LEDs work out of the box by adding netdev-based uci defaults in board.d/01_leds (same pattern as the WAN LEDs), e.g. `ucidef_set_led_netdev` for `green:wlan-2ghz` on `phy0-ap0` and `green:wlan-5ghz` on `phy1-ap0` with tx+rx mode; the `phy*tpt` default-triggers in the DTS can then be dropped.

The missing of "phy0tpt"/"phy1tpt" is because of a bug in the mt76 driver package. This bug has been fixed, see this commit (https://github.com/openwrt/mt76/commit/c5a3bd91aa735b669618610d5f0ebfa5786845a6) for detail.

Wfarmer1:
> > Found a subtle issue while testing a locally built image from this PR ([2bdd8ca](https://github.com/openwrt/openwrt/commit/2bdd8ca40ccde8413d416a9b9f1d71690ccd8180), kernel 6.18.44): the WiFi LEDs' `linux,default-trigger = "phy0tpt"/"phy1tpt"` never takes effect on a mainline-default build. `MAC80211_LEDS` is only injected when the `CONFIG_LEDS_TRIGGERS` build symbol is set (`config-$(CONFIG_LEDS_TRIGGERS) += MAC80211_LEDS` in package/kernel/mac80211/Makefile), and a default `make defconfig` does not contain that symbol, so the tpt triggers are never registered and the LEDs stay off.
> > Verified on hardware:
> > 
> > * `cat /sys/class/leds/green:wlan-*/trigger` lists the mt76 triggers (`phy0rx/tx/assoc/radio`) but no tpt triggers
> > * `echo 255 > brightness` lights both LEDs fine, so the GPIO mapping is correct
> > 
> > Suggestion: keep the traffic-blink semantics but make the LEDs work out of the box by adding netdev-based uci defaults in board.d/01_leds (same pattern as the WAN LEDs), e.g. `ucidef_set_led_netdev` for `green:wlan-2ghz` on `phy0-ap0` and `green:wlan-5ghz` on `phy1-ap0` with tx+rx mode; the `phy*tpt` default-triggers in the DTS can then be dropped.
> 
> The missing of "phy0tpt"/"phy1tpt" is because of a bug in the mt76 driver package. This bug has been fixed, see this commit ([openwrt/mt76@c5a3bd9](https://github.com/openwrt/mt76/commit/c5a3bd91aa735b669618610d5f0ebfa5786845a6)) for detail.

@bfdeh Thanks for the correction — you're right, and apologies for the misleading analysis in my previous comment. I re-checked on a clean local build of this PR (2bdd8ca):

- mt76 is pinned to 59676919 (2026-07-01), which is 147 commits behind the fix you linked (compare: 59676919...c5a3bd91 → ahead_by=147), and its mt7915/Makefile still has `EXTRA_CFLAGS += -DCONFIG_MT76_LEDS`. So `IS_ENABLED(CONFIG_MT76_LEDS)` is false when mt7915 is compiled, mt76_led_init() bails out early, and `ieee80211_create_tpt_led_trigger()` is never called — hence no phy*tpt triggers and the DTS default-trigger doesn't bind. (For reference, c5a3bd91 was merged to mt76 master on 2026-08-22, i.e. after this pin.)
- Current main already carries it: the main mt76 pin is be5ce791 (2026-09-01) and its mt7915/Makefile uses `ccflags-y += -DCONFIG_MT76_LEDS`; so building from current main creates the tpt triggers and the LEDs bind normally. (phy0rx/tx/assoc/radio in my trigger list come from mac80211's own ieee80211_led_init, independent of mt76.)

The CONFIG_MAC80211_LEDS point in my earlier comment was wrong too — backports builds with its own .config (CPTCFG_MAC80211_LEDS=y). No DTS change is needed; please disregard the netdev suggestion. Thanks again.

fildunsky (approved):

PR #24317, opened 2026-07-20
mediatek: filogic: add support netis NR109GPE V2

bfdeh:
mediatek: filogic: add support netis NR109GPE V2

- Hardware specification:
  SoC: MediaTek MT7981B
  Switch: RTL8370MB
  Flash: 128MB
  RAM: DDR3 256MB
  Ethernet: 9x 1G
  Button: Reset

- MAC addresses
  | Interface   | Source                   | Offset   |
  | ----------- | ------------------------ | -------- |
  | WAN         | Factory mac-base         | 0x1fef26 |
  | LAN/CPU     | Factory mac-base         | 0x1fef20 |

- Official LED layout:
  lan8 ... lan1 wan internet
  internet led is a dual-color led, blue for online and red for offline.

- Redefinition for OpenWrt:
  lan* and wan: unchanged
  internet-blue: led-boot, led-running
  internet-red: led-failsafe, led-upgrade

- Installing OpenWrt:
  - Setup a tftp server on your PC. Copy xxx-preloader.bin, 
    xxx-bl31-uboot.fip and xxx-initramfs.itb to tftp root directory.
  - Connect to the router via ssh or telnet, username: root,
    password is the web login password of the router.
  - Flash with the following commands:

    ```shell
    cd /tmp
    tftp -r xxx-preloader.bin -g x.x.x.x
    tftp -r xxx-bl31-uboot.fip -g x.x.x.x
    mtd write xxx-preloader.bin spi0.0
    mtd write xxx-bl31-uboot.fip FIP
    mtd erase ubi
    ```

  - Set a static ip(192.168.1.254) for your PC and 
    **connect it to WAN port**. Then reboot the router. 
    It will run initramfs image automatically.
  - After openwrt boots up, **connect PC to any LAN port**,
    perform sysupgrade via web UI.

### DIFF
--- a/package/boot/uboot-mediatek/patches/474-add-netis-nr109gpev2.patch
+++ b/package/boot/uboot-mediatek/patches/474-add-netis-nr109gpev2.patch
@@ -1,0 +1,341 @@
+--- /dev/null
++++ b/arch/arm/dts/mt7981-netis-nr109gpev2.dts
+@@ -0,0 +1,141 @@
++// SPDX-License-Identifier: GPL-2.0
++
++/dts-v1/;
++#include "mt7981.dtsi"
++#include <dt-bindings/gpio/gpio.h>
++#include <dt-bindings/input/input.h>
++
++/ {
++	model = "netis NR109GPE V2";
++	compatible = "netis,nr109gpev2", "mediatek,mt7981";
++	chosen {
++		stdout-path = &uart0;
++		tick-timer = &timer0;
++	};
++
++	memory@40000000 {
++		device_type = "memory";
++		reg = <0x40000000 0x20000000>;
++	};
++
++	gpio-keys {
++		compatible = "gpio-keys";
++
++		reset {
++			label = "reset";
++			linux,code = <KEY_RESTART>;
++			gpios = <&pio 1 GPIO_ACTIVE_LOW>;
++		};
++	};
++
++	leds {
++		compatible = "gpio-leds";
++
++		system {
++			label = "system";
++			gpios = <&pio 7 GPIO_ACTIVE_LOW>;
++		};
++
++		upgrade {
++			label = "upgrade";
++			gpios = <&pio 8 GPIO_ACTIVE_LOW>;
++		};
++
++		wan {
++			label = "wan";
++			gpios = <&pio 13 GPIO_ACTIVE_LOW>;
++		};
++	};
++};
++
++&uart0 {
++	status = "okay";
++};
++
++&eth {
++    status = "okay";
++    mediatek,gmac-id = <1>;
++    phy-mode = "gmii";
++    phy-handle = <&phy0>;
++    phy0: ethernet-phy@0 {
++              compatible = "ethernet-phy-id03a2.9461";
++              reg = <0>;
++              phy-mode = "gmii";
++    };
++};
++
++&pio {
++	spi_flash_pins: spi0-pins-func-1 {
++		mux {
++			function = "flash";
++			groups = "spi0", "spi0_wp_hold";
++		};
++
++		conf-pu {
++			pins = "SPI0_CS", "SPI0_HOLD", "SPI0_WP";
++			drive-strength = <MTK_DRIVE_8mA>;
++			bias-pull-up = <MTK_PUPD_SET_R1R0_11>;
++		};
++
++		conf-pd {
++			pins = "SPI0_CLK", "SPI0_MOSI", "SPI0_MISO";
++			drive-strength = <MTK_DRIVE_8mA>;
++			bias-pull-down = <MTK_PUPD_SET_R1R0_11>;
++		};
++	};
++};
++
++&spi0 {
++	#address-cells = <1>;
++	#size-cells = <0>;
++	pinctrl-names = "default";
++	pinctrl-0 = <&spi_flash_pins>;
++	status = "okay";
++	enhance_timing;
++	dma_ext;
++	ipm_design;
++	support_quad;
++	tick_dly = <2>;
++	sample_sel = <0>;
++
++	spi_nand@0 {
++		compatible = "spi-nand";
++		reg = <0>;
++		spi-max-frequency = <52000000>;
++
++		partitions {
++			compatible = "fixed-partitions";
++			#address-cells = <1>;
++			#size-cells = <1>;
++
++			partition@0 {
++				label = "BL2";
++				reg = <0x0 0x100000>;
++			};
++
++			partition@100000 {
++				label = "u-boot-env";
++				reg = <0x100000 0x80000>;
++			};
++
++			partition@180000 {
++				label = "Factory";
++				reg = <0x180000 0x200000>;
++			};
++
++			partition@380000 {
++				label = "FIP";
++				reg = <0x380000 0x200000>;
++			};
++
++			partition@580000 {
++				label = "ubi";
++				reg = <0x580000 0x7280000>;
++			};
++		};
++	};
++};
++
++&watchdog {
++	status = "disabled";
++};
+--- /dev/null
++++ b/configs/mt7981_netis_nr109gpev2_defconfig
+@@ -0,0 +1,129 @@
++CONFIG_ARM=y
++CONFIG_SYS_HAS_NONCACHED_MEMORY=y
++CONFIG_POSITION_INDEPENDENT=y
++CONFIG_ARCH_MEDIATEK=y
++CONFIG_TEXT_BASE=0x41e00000
++CONFIG_SYS_MALLOC_F_LEN=0x4000
++CONFIG_NR_DRAM_BANKS=1
++CONFIG_DEFAULT_DEVICE_TREE="mt7981-netis-nr109gpev2"
++CONFIG_OF_LIBFDT_OVERLAY=y
++CONFIG_TARGET_MT7981=y
++CONFIG_SYS_LOAD_ADDR=0x46000000
++CONFIG_DEBUG_UART_BASE=0x11002000
++CONFIG_DEBUG_UART_CLOCK=40000000
++CONFIG_DEBUG_UART=y
++CONFIG_BUTTON_CMD=y
++CONFIG_FIT=y
++CONFIG_SPI_BOOT=y
++CONFIG_AUTOBOOT_MENU_SHOW=y
++CONFIG_OF_SYSTEM_SETUP=y
++CONFIG_USE_PREBOOT=y
++CONFIG_SYS_CBSIZE=512
++CONFIG_SYS_PBSIZE=1049
++CONFIG_LOGLEVEL=7
++CONFIG_CONSOLE_MUX=y
++CONFIG_LOG=y
++# CONFIG_BOARD_INIT is not set
++CONFIG_BOARD_LATE_INIT=y
++CONFIG_UPDATE_FIT=y
++CONFIG_HUSH_PARSER=y
++CONFIG_SYS_PROMPT="MT7981> "
++CONFIG_CMD_CPU=y
++CONFIG_CMD_LICENSE=y
++# CONFIG_BOOTM_NETBSD is not set
++# CONFIG_BOOTM_PLAN9 is not set
++# CONFIG_BOOTM_RTEMS is not set
++# CONFIG_BOOTM_VXWORKS is not set
++CONFIG_CMD_BOOTMENU=y
++CONFIG_CMD_ASKENV=y
++CONFIG_CMD_ERASEENV=y
++CONFIG_CMD_ENV_FLAGS=y
++CONFIG_CMD_STRINGS=y
++# CONFIG_CMD_UNLZ4 is not set
++# CONFIG_CMD_UNZIP is not set
++CONFIG_CMD_DFU=y
++CONFIG_CMD_DM=y
++CONFIG_CMD_GPIO=y
++CONFIG_CMD_MTD=y
++CONFIG_CMD_PCI=y
++CONFIG_CMD_SF_TEST=y
++CONFIG_CMD_TFTPPUT=y
++CONFIG_CMD_TFTPSRV=y
++CONFIG_CMD_RARP=y
++CONFIG_CMD_CDP=y
++CONFIG_CMD_LINK_LOCAL=y
++CONFIG_CMD_DHCP=y
++CONFIG_CMD_DNS=y
++CONFIG_CMD_PING=y
++CONFIG_CMD_SNTP=y
++CONFIG_CMD_PXE=y
++CONFIG_CMD_CACHE=y
++CONFIG_CMD_PSTORE=y
++CONFIG_CMD_PSTORE_MEM_ADDR=0x42ff0000
++CONFIG_CMD_UUID=y
++CONFIG_CMD_HASH=y
++CONFIG_CMD_SMC=y
++CONFIG_CMD_FS_GENERIC=y
++CONFIG_CMD_FS_UUID=y
++CONFIG_CMD_UBI=y
++CONFIG_CMD_UBI_RENAME=y
++CONFIG_ENV_OVERWRITE=y
++CONFIG_ENV_IS_IN_UBI=y
++CONFIG_ENV_UBI_PART="ubi"
++CONFIG_ENV_UBI_VOLUME="ubootenv"
++CONFIG_ENV_UBI_VOLUME_CREATE=y
++CONFIG_ENV_USE_DEFAULT_ENV_TEXT_FILE=y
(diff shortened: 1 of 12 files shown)


package/boot/uboot-mediatek/patches/474-add-netis-nr109gpev2.patch
```diff
@@ -0,0 +1,337 @@
+--- /dev/null
++++ b/arch/arm/dts/mt7981-netis-nr109gpev2.dts
+@@ -0,0 +1,141@@
```
openwrt-ai: nit: This hunk header is missing the trailing space before the closing `@@`. The other two hunks in this same file (`@@ -0,0 +1,125 @@` and `@@ -0,0 +1,62 @@`) and every other patch under `package/boot/uboot-mediatek/patches/` use the `@@ ... @@` form. GNU patch/quilt tolerate the missing space, but `git apply` and stricter tooling reject it. The line count (141) is correct, so only the space needs adding:

```suggestion
@@ -0,0 +1,141 @@
```

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -2721,6 +2721,30 @@ define Device/netis_eap930-v1
...
 
+define Device/netis_nr109gpev2
+  DEVICE_VENDOR := netis
+  DEVICE_MODEL := NR109GPE
+  DEVICE_VARIANT := V2
+  DEVICE_DTS := mt7981b-netis-nr109gpev2
+  DEVICE_DTS_DIR := ../dts
+  DEVICE_DTC_FLAGS := --pad 4096
+  DEVICE_DTS_LOADADDR := 0x43f00000
+  DEVICE_PACKAGES := kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware
```
openwrt-ai: nit: This board pulls `kmod-mt7915e` / `mt7981-wo-firmware`, i.e. it has the MT7981 built-in WiFi radio, but the commit message's Hardware specification section lists only SoC/Switch/Flash/RAM/Ethernet/Button and omits WiFi. Per the new-device checklist the spec should also give the WiFi band/chains/chip.

---

target/linux/mediatek/files/drivers/net/phy/rtk/rtl8367s_mdio.c
```diff
@@ -199,11 +227,121 @@ static void set_rtl8367s_rgmii(void)
...
+		ext->mac.link = !strcasecmp(str, "down") ? PORT_LINKDOWN : PORT_LINKUP;
+	if (of_property_read_bool(port_np, "nway"))
+		ext->mac.nway = ENABLED;
+	if (of_property_read_bool(port_np, "tx-pause"))
+		ext->mac.txpause = ENABLED;
+	if (of_property_read_bool(port_np, "rx-pause"))
+		ext->mac.rxpause = ENABLED;
+
+	of_property_read_u32(port_np, "rgmii-tx-delay", &ext->delay_tx);
+    of_property_read_u32(port_np, "rgmii-rx-delay", &ext->delay_rx);
```
openwrt-ai: nit: This line is indented with 4 spaces, while the identical statement directly above it (`rgmii-tx-delay`) and the rest of this function use tabs. (The new `rtk_ext_mode_map` table added above is also space-indented rather than tab-indented.)

```suggestion
	of_property_read_u32(port_np, "rgmii-rx-delay", &ext->delay_rx);
```

---

openwrt-ai (commented):
Reviewed 1 new commit. A few nits inline; nothing blocking. The shared rtl8367 driver refactor is gated on a DT `ports` node, and no existing rtk-gsw board defines one, so legacy boards keep their hard-coded port map — backward compatibility looks preserved.

---

target/linux/mediatek/image/filogic.mk
bfdeh: There is no wifi for this device, and kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware are removed.

openwrt-ai (commented):

The updated revision addresses all three prior nits: the U-Boot patch hunk header now has the trailing space (`@@ -0,0 +1,141 @@`), `rgmii-rx-delay` is tab-indented, and the WiFi packages are dropped from `DEVICE_PACKAGES`. The board DTS defines no wmac/pcie/wifi node, so removing `kmod-mt7915e` / `mt7981-*-firmware` is consistent with the device having no wireless radio.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -2721,6 +2721,30 @@ define Device/netis_eap930-v1
...
+  KERNEL := kernel-bin | gzip
+  KERNEL_INITRAMFS := kernel-bin | lzma | fit lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb with-initrd | pad-to 64k
+  KERNEL_INITRAMFS_SUFFIX := .itb
+  KERNEL_IN_UBI := 1
+  UBOOTENV_IN_UBI := 1
+  IMAGES := sysupgrade.itb
+  IMAGE_SIZE := 117248k
+  IMAGE/sysupgrade.itb := append-kernel | fit gzip $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb external-with-rootfs | pad-rootfs | append-metadata
+  ARTIFACTS := spim-nand-preloader.bin spim-nand-bl31-uboot.fip
+  ARTIFACT/spim-nand-preloader.bin	:= mt7981-bl2 spim-nand-ddr3
```
unisequence, line 2743: spim-nand-ddr3 doesn't specify a frequency variant. If this board has a rated DDR3 chip (e.g. -1866), the preloader should target that timing profile specifically rather than the generic default.

target/linux/mediatek/image/filogic.mk
```diff
@@ -2721,6 +2721,30 @@ define Device/netis_eap930-v1
...
+  DEVICE_VENDOR := netis
+  DEVICE_MODEL := NR109GPE
+  DEVICE_VARIANT := V2
+  DEVICE_DTS := mt7981b-netis-nr109gpev2
+  DEVICE_DTS_DIR := ../dts
+  DEVICE_DTC_FLAGS := --pad 4096
+  DEVICE_DTS_LOADADDR := 0x43f00000
+  DEVICE_PACKAGES :=
+  KERNEL_LOADADDR := 0x44000000
+  KERNEL := kernel-bin | gzip
```
unisequence, line 2734:
```suggestion
  KERNEL := kernel-bin | libdeflate-gzip
```
Would save having to come back later to change -- https://github.com/openwrt/openwrt/pull/24148

target/linux/mediatek/dts/mt7981b-netis-nr109gpev2.dts
```diff
@@ -0,0 +1,265 @@
...
+	gpio-keys {
+		compatible = "gpio-keys";
+		reset {
+			label = "reset";
+			linux,code = <KEY_RESTART>;
+			gpios = <&pio 1 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+	leds {
```
unisequence, line 110: nit: sort LED nodes by GPIO number

target/linux/mediatek/files/drivers/net/phy/rtk/rtl8367c/include/rtk_switch.h
```diff
@@ -18,10 +18,12 @@
 #ifndef __RTK_SWITCH_H__
```
unisequence, line 18: This changes the shared RTL8367C driver (rtl8367s_swconfig_init() signature, removes RTL8367C_SW_CPU_PORT/RTL8367C_NUM_PORTS) used by other RTL8367-based boards, not just this one. What's the
  reason the shared driver needed changing for this board, and have the existing callers of rtl8367s_swconfig_init() been checked against the new signature?

unisequence (commented):
No RNG node/config in 474-*.patch -- similar boards enable CONFIG_RNG_MTK_V2/CONFIG_DM_RNG for KASLR seeding (see #24089 for the u-boot side). Was this intentionally skipped?

unisequence:
MAC addresses are described as prose here (`LAN: Label MAC ... WAN: LAN + 1`) rather than a table with explicit offsets per interface. A table like the one in #24089 (see the commit message) makes it easier to verify at a glance:

| Interface | Source | Offset | Address |
|---|---|---|---|
| WAN | Factory mac-base | 0x... | ... |
| LAN/CPU | Factory mac-base | 0x... | ... |

Worth adding to the commit message and PR description.

package/boot/uboot-mediatek/patches/474-add-netis-nr109gpev2.patch
```diff
@@ -0,0 +1,337 @@
...
++		};
++	};
++};
++
++&watchdog {
++	status = "disabled";
++};
+--- /dev/null
++++ b/configs/mt7981_netis_nr109gpev2_defconfig
+@@ -0,0 +1,125 @@
```
unisequence, line 147: can you please add

```suggestion 
+CONFIG_ARM_SMCCC=y
+CONFIG_CMD_RNG=y
+CONFIG_DM_RNG=y
+CONFIG_RNG_MTK_V2=y
```

target/linux/mediatek/files/drivers/net/phy/rtk/rtl8367c/include/rtk_switch.h
bfdeh: Based on the code, the RTL8367C driver is intended to support the 8367/8370/8364/8363 series of chips. These switch chips have different numbers of ports and different capabilities. However, the port counts and configurations are hard-coded in the driver through the following macros, variables, and functions, which prevents the driver from being generic:

`RTL8367C_SW_CPU_PORT`
`RTL8367C_NUM_PORTS`
`rtl8367c_port_id`
`set_rtl8367s_sgmii`
`set_rtl8367s_rgmii`

These changes replace `RTL8367C_SW_CPU_PORT`, `RTL8367C_NUM_PORTS`, and `rtl8367c_port_id` with the variables `rtk_sw_cpu_port`, `rtk_sw_num_ports`, and `rtk_sw_port_id`, respectively, while using their original values as the default values for the new variables. If a `ports` node is defined in the DTS, the values of the new variables are overridden based on the `ports` node. If no `ports` node is defined in the DTS, the new variables retain their default values.

The same applies to `ext-ports`. Their capabilities are configured based on the DTS only when a `ports` node is defined in the DTS. Otherwise, `set_rtl8367s_sgmii` / `set_rtl8367s_rgmii` are still called.

I checked all existing RTL8367-based boards, and none of their DTS files define a `ports` node. Therefore, these changes should remain compatible with existing RTL8367-based boards.

package/boot/uboot-mediatek/patches/474-add-netis-nr109gpev2.patch
bfdeh: All done, thanks for your review.

package/boot/uboot-mediatek/patches/474-add-netis-nr109gpev2.patch
```diff
@@ -0,0 +1,341 @@
...
++CONFIG_PINCONF=y
++CONFIG_PINCTRL_MT7981=y
++CONFIG_POWER_DOMAIN=y
++CONFIG_MTK_POWER_DOMAIN=y
++CONFIG_DM_SERIAL=y
++CONFIG_SERIAL_RX_BUFFER=y
++CONFIG_ARM_SMCCC=y
++CONFIG_CMD_RNG=y
++CONFIG_DM_RNG=y
++CONFIG_RNG_MTK_V2=y
```
openwrt-ai, line 269: nit: the four new RNG options are inserted in the middle of the serial block, which is not `savedefconfig` (Kconfig) order — the next `make savedefconfig` on this board will shuffle them and produce unrelated churn. Every sibling mt7981 defconfig in this directory places them at three different spots:

- `CONFIG_ARM_SMCCC=y` immediately after `CONFIG_ARM=y` — line 148 here ([469-add-netis-eap930-v1.patch:4-5](``https://github.com/openwrt/openwrt/blob/c9e971ffa8be54b0ff1e66b9640ccbd1b2225863/package/boot/uboot-mediatek/patches/469-add-netis-eap930-v1.patch#L4-L5``))
- `CONFIG_CMD_RNG=y` inside the `CONFIG_CMD_*` block, between `CONFIG_CMD_ENV_FLAGS=y` and `CONFIG_CMD_STRINGS=y` — lines 187/188 here ([469-add-netis-eap930-v1.patch:39](``https://github.com/openwrt/openwrt/blob/c9e971ffa8be54b0ff1e66b9640ccbd1b2225863/package/boot/uboot-mediatek/patches/469-add-netis-eap930-v1.patch#L39``))
- `CONFIG_DM_RNG=y` + `CONFIG_RNG_MTK_V2=y` directly before `CONFIG_DM_SERIAL=y` — line 264 here ([501-add-netis-n6-v2.patch:104-107](``https://github.com/openwrt/openwrt/blob/c9e971ffa8be54b0ff1e66b9640ccbd1b2225863/package/boot/uboot-mediatek/patches/501-add-netis-n6-v2.patch#L104-L107``))

Moving them keeps the total at 129 lines, so the `@@ -0,0 +1,129 @@` header stays as is. Functionally the current placement works — the RNG node is unconditionally present in `mt7981.dtsi` via [193-mediatek-mt7981-add-RNG-node.patch](``https://github.com/openwrt/openwrt/blob/c9e971ffa8be54b0ff1e66b9640ccbd1b2225863/package/boot/uboot-mediatek/patches/193-mediatek-mt7981-add-RNG-node.patch``) — so this is ordering only.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push `e37c758` → `c9e971f`). The delta since the last review is: DDR3 timing switched to `ddr3-1866` (both `BL2_DDRTYPE`/`DEPENDS` in `package/boot/uboot-mediatek/Makefile` and `ARTIFACT/spim-nand-preloader.bin` in `filogic.mk` — consistent, and `trusted-firmware-a-mt7981-spim-nand-ddr3-1866` exists in the ATF Makefile), `KERNEL := kernel-bin | libdeflate-gzip`, LED nodes sorted by GPIO in both the U-Boot and the OpenWrt DTS, RNG options added to the U-Boot defconfig, and the MAC section of the commit message turned into a table. The MAC table matches the DTS nvmem cells (`macaddr_lan` @ 0x1fef20, `macaddr_wan` @ 0x1fef26, both inside the 2 MiB Factory partition). All three hunk headers in `474-add-netis-nr109gpev2.patch` still match their line counts (141/129/62). One ordering nit inline; CI on `c9e971f` is green/queued with no failures.

## Commit checks
- c9e971f "mediatek: filogic: add support netis NR109GPE V2" — the Hardware specification section is still incomplete for a new-device commit: `Flash: 128MB` doesn't say SPI-NAND (the DTS uses `spi-nand` and the images are `spim-nand-*`), and there is no UART line (header location, pinout, baud — the DTS uses `serial0:115200n8`) and no Power line (voltage + current). Since the board has no radio, an explicit `WiFi: none` would also save the next reader from wondering whether it was just forgotten. Now that the preloader targets `ddr3-1866`, spelling that out on the RAM line would be useful too.

---

package/boot/uboot-mediatek/patches/474-add-netis-nr109gpev2.patch
```diff
@@ -0,0 +1,341 @@
...
++	model = "netis NR109GPE V2";
++	compatible = "netis,nr109gpev2", "mediatek,mt7981";
++	chosen {
++		stdout-path = &uart0;
++		tick-timer = &timer0;
++	};
++
++	memory@40000000 {
++		device_type = "memory";
++		reg = <0x40000000 0x20000000>;
```
DH-41, line 21: If the board has 256MB of RAM, shouldn't the size be updated to `0x10000000`?

target/linux/mediatek/dts/mt7981b-netis-nr109gpev2.dts
```diff
@@ -0,0 +1,265 @@
...
+		led-upgrade = &led_upgrade;
+	};
+
+	chosen {
+		stdout-path = "serial0:115200n8";
+		rootdisk = <&ubi_rootdisk>;
+	};
+
+	memory@40000000 {
+		reg = <0 0x40000000 0 0x20000000>;
```
DH-41, line 25: This also needs to be updated here as well. `0x10000000`?

bfdeh: I think the RAM size here is more of a placeholder. The actual memory size will be detected automatically by the bootloader, which will override this value before starting the kernel. Is it necessary to change it?

U-Boot’s RAM size is similar.

PR #24321, opened 2026-07-20
build: add manifest.json

aparcar:
With the switch from opkg to apk, we lost the Packages.manifest file, which offered package details in a single file. While the existing index.json file offers an overview of existing packages and versions, details like license and source is missing.

Use the apk internal feature to generate a file containing all relevant information from the package index.

### DIFF
--- a/package/Makefile
+++ b/package/Makefile
@@ -140,8 +140,8 @@ ifneq ($(CONFIG_USE_APK),)
 			$(if $(CONFIG_SIGNED_PACKAGES),--sign $(BUILD_KEY_APK_SEC),) \
 			--output packages.adb \
 			*.apk; \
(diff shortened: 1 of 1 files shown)


Rondom:
Who is the consumer of the manifest file? . I am wondering whether we could use the SBOM instead and thus avoid maintaining two different files with the roughly the same content but different formats

aparcar:
We don't have a proper interface to search for packages, something that [Alpine Linux offers](https://pkgs.alpinelinux.org/packages). With a file like that, it would be quite simple to facilitate. Furthermore I'm working on [rebuilderd support for OpenWrt](https://rebuilderd.n.aparcar.org/), which needs a proper way to determine existing packages with some meta information.

openwrt-ai (commented):

---

PR #24326, opened 2026-07-20
kernel: disable qualcomm crypto

neheb:
Upstream has marked it as BROKEN. Quote:

This driver is harmful:

- It is much slower than the CPU.
- It Has a history of bugs.
- It does not have exclusive access to the hardware, causing races with the secure world.
- It register its implementations with too low a cra_priority for them to be actually used.


neheb:
https://git.kernel.org/pub/scm/linux/kernel/git/herbert/cryptodev-2.6.git/commit/?id=df373d39c6f038d176af303ae72f02c7c70b953d

openwrt-ai (commented):

---

jonasjelonek:

PR #24333, opened 2026-07-21
mt76: pass LED define via ccflags-y in driver submodules

mbc07:
Commit d2b01fbc329b ("mt76: pass LED define via ccflags-y") converted the top-level Makefile from the deprecated EXTRA_CFLAGS to ccflags-y so that -DCONFIG_MT76_LEDS keeps working now that kbuild no longer honours EXTRA_CFLAGS.

However, ccflags-y (like EXTRA_CFLAGS before it) only applies to the kbuild makefile it is set in and is not inherited by subdirectories pulled in via obj-y/obj-m. The mt7603, mt7615 and mt7915 driver subdirectories each set their own
"EXTRA_CFLAGS += ... -DCONFIG_MT76_LEDS", which is now a silent no-op. As a result CONFIG_MT76_LEDS is undefined when building mt7603/init.c, mt7615/init.c, mt7615/pci_init.c and mt7915/init.c, the IS_ENABLED(CONFIG_MT76_LEDS) blocks are compiled out, and the mt76-* LED class devices are never registered for these drivers.

Convert these subdirectory Makefiles to ccflags-y as well.

mt7996/Makefile never carried the define at all, even though mt7996/init.c has the same IS_ENABLED(CONFIG_MT76_LEDS) checks for its LED callbacks and GPIO mux setup. Add ccflags-y there too so mt7996 LED support is enabled consistently with the other drivers.

mt76x0, mt76x2, mt7921 and mt7925 contain no CONFIG_MT76_LEDS references and are left untouched.

Fixes: d2b01fbc329b ("mt76: pass LED define via ccflags-y")
Link: https://github.com/openwrt/mt76/issues/1077
Signed-off-by: Mateus B. Cassiano <mbc07@live.com>

### DIFF
--- a/package/kernel/mt76/patches/001-pass-led-define-via-ccflags-y-in-driver-submodules.patch
+++ b/package/kernel/mt76/patches/001-pass-led-define-via-ccflags-y-in-driver-submodules.patch
@@ -1,0 +1,90 @@
+From 99961e76391ed1907557ae0627f1e6cfa0fc80c6 Mon Sep 17 00:00:00 2001
+From: "Mateus B. Cassiano" <mbc07@live.com>
+Date: Mon, 15 Jun 2026 20:04:59 -0400
+Subject: [PATCH] wifi: mt76: pass LED define via ccflags-y in driver
+ submodules
+
+Commit d2b01fbc329b ("mt76: pass LED define via ccflags-y")
+converted the top-level Makefile from the deprecated EXTRA_CFLAGS
+to ccflags-y so that -DCONFIG_MT76_LEDS keeps working now that
+kbuild no longer honours EXTRA_CFLAGS.
+
+However, ccflags-y (like EXTRA_CFLAGS before it) only applies to
+the kbuild makefile it is set in and is not inherited by
+subdirectories pulled in via obj-y/obj-m. The mt7603, mt7615 and
+mt7915 driver subdirectories each set their own
+"EXTRA_CFLAGS += ... -DCONFIG_MT76_LEDS", which is now a silent
+no-op. As a result CONFIG_MT76_LEDS is undefined when building
+mt7603/init.c, mt7615/init.c, mt7615/pci_init.c and mt7915/init.c,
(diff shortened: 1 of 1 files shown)


mbc07:
I've been trying to upstream this patch for a while now: 

https://github.com/openwrt/mt76/pull/1090
https://lore.kernel.org/linux-wireless/20260715044431.1207-2-mbc07@live.com/T/#u

Since I'm realizing this may take even longer to get upstreamed, I'm submitting this patch here as a temporary downstream workaround, that should be dropped once the change finally arrives at upstream.

package/kernel/mt76/patches/001-pass-led-define-via-ccflags-y-in-driver-submodules.patch
openwrt-ai, line 1: This patch changes the compiled output — it re-enables registration of the `mt76-*` LED class devices for mt7603/mt7615/mt7915 and newly enables them for mt7996 — but `package/kernel/mt76/Makefile` still carries `PKG_RELEASE=1` ([blob](https://github.com/openwrt/openwrt/blob/5bc7c317240cded0cd3f47765bf7edc4fca5ff39/package/kernel/mt76/Makefile#L4)). A functional change like this should bump `PKG_RELEASE` to `2` so the rebuilt package is recognised as an update on upgrade; this is not one of the cosmetic / maintainer / source-URL exemptions. The openwrt bot flagged the same missing bump.

---

openwrt-ai (commented):
Reviewed 1 new commit; the commit message accurately describes the diff. One inline packaging note below.

---

mbc07:
Hi, gentle ping on this PR.

This restores LED registration for the affected mt76 drivers, and the change has been tested on mt7615 and independently on mt7996 hardware. The upstream submission is still awaiting review, so this remains useful as a temporary downstream fix.

Any feedback would be appreciated. Thanks!

mbc07:
Patch was merged upstream. Closing the PR...

PR #11647, opened 2022-12-30
WIP: ramips: add support for Asus RT-N65U

andy9a9:
ASUS RT-N65U is a 2.4/5 Ghz band 11ac (Wi-Fi 5) router, based on RT3883F.

Specification:
- CPU1: Ralink RT3883 @500MHz, WiSoC 5GHz
- CPU2: Ralink RT3352 @400MHz, WiSoC 2.4GHz
- RAM1: 128MB
- RAM2: 16MB
- Flash: Macronix MX25L12805D (16MB)
- WLAN: Ralink RT3883 (5GHz 3T3R), Ralink RT3352 (2.4GHz 2T2R)
- 4x 10/100/1000 Mbps Ethernet
- 1x 10/100/1000 Mbps WAN
- USB: ASMedia ASM1042 USB 3.0 Controller
- PMU: UTC UZ1086L-ADJ
- 4x LED, 2x button

Notes:
- Lot of stuff have been copies from `rt-n56u`

TODO:
- [x] verify if it boots on the real device
- [ ] try to figure-out what's wrong and fix it
  - [ ] LZMA compression
  - [x] SPI flash mounting
  - [ ] switch problem
  - [ ] overall check

PR #24343, opened 2026-07-21
ramips: add support for ASUS RT-N65U

zatrit:
ASUS RT-N65U is a 2.4/5 GHz band 11n (Wi-Fi 4) router, based on RT3883.

Specification:
- CPU1: Ralink RT3883 @ 500MHz, WiSoC 5GHz
- CPU2: Ralink RT3352 @ 400MHz, WiSoC 2.4GHz
- RAM1: 128MB
- RAM2: 16MB
- Flash: Macronix MX25L12805D (16MB)
- WLAN: Ralink RT3883 (5GHz 3T3R), Ralink RT3352 (2.4GHz 2T2R)
- Switch: Realtek RTL8367RB
- 4x 10/100/1000 Mbps Ethernet
- 1x 10/100/1000 Mbps WAN
- USB: ASMedia ASM1042 USB 3.0 Controller
- PMU: UTC UZ1086L-ADJ
- 6x LED, 2x button

Installation (TFTP):
1. Power off the router.
2. Press and hold the Reset button.
3. Power on the router.
4. Wait 5 seconds, then release the Reset button.
5. Connect your PC to the router with Ethernet cable.
6. Configure a static IP address in the `192.168.1.0/24` subnet (except `192.168.1.1`).
7. Configure your TFTP client to use binary (octet) transfer mode.
8. Upload the firmware image (OpenWrt, stock, or Padavan) to `192.168.1.1`.
   The TFTP server accepts any filename.
9. Wait for the firmware flashing process to complete.

WARNING: Flashing the stock firmware after using Padavan may brick
your router due to configuration partition incompatibility.

MAC address layout:
- Wi-Fi 2.4GHz & WAN: label MAC
- Wi-Fi 5GHz & LAN: label MAC + 4

MTD layout:
```
mtd0: 00030000 00010000 "u-boot"
mtd1: 00010000 00010000 "config"
mtd2: 00010000 00010000 "factory"
mtd3: 00fb0000 00010000 "firmware"
mtd4: 001ee3b3 00010000 "kernel"
mtd5: 00dc1c4d 00010000 "rootfs"
mtd6: 00960000 00010000 "rootfs_data"
```

Known issues:
- The RT3352 2.4 GHz iNIC is unsupported because it requires a
  proprietary firmware blob and kernel support
- The Wi-Fi LEDs cannot be disabled independently because they share
  the LED power rail controlled by GPIO10

This work is based on the earlier draft support by @andy9a9 (#11647).

### DIFF
--- a/target/linux/generic/files/drivers/net/phy/rtl8367b.c
+++ b/target/linux/generic/files/drivers/net/phy/rtl8367b.c
@@ -1549,6 +1549,12 @@ static int rtl8367b_detect(struct rtl8366_smi *smi)
 		chip_name = "8367RB";
 		smi->rtl8367b_chip = RTL8367B_CHIP_RTL8367RB;
 		break;
+	case 0x2000:
+		if (chip_num == 0x6000) {
+			chip_name = "8367RB";
+			smi->rtl8367b_chip = RTL8367B_CHIP_RTL8367RB;
+		}
+		break;
 	case 0x1010:
 		chip_name = "8367R-VB";
 		smi->rtl8367b_chip = RTL8367B_CHIP_RTL8367R_VB;
--- a/target/linux/ramips/dts/rt3883_asus_rt-n65u.dts
+++ b/target/linux/ramips/dts/rt3883_asus_rt-n65u.dts
@@ -1,0 +1,196 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+#include "rt3883.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+
+/ {
+	compatible = "asus,rt-n65u", "ralink,rt3883-soc";
+	model = "Asus RT-N65U";
+
+	aliases {
+		led-boot = &led_power;
+		led-failsafe = &led_power;
+		led-running = &led_power;
+		led-upgrade = &led_power;
+	};
+
+	chosen {
+		bootargs = "console=ttyS0,115200";
+	};
+
+	rtl8367rb {
+		compatible = "realtek,rtl8367b";
+		realtek,extif = <6 1 0 1 1 1 1 1 1 2>;
+		mii-bus = <&mdio0>;
+	};
+
+	keys {
+		compatible = "gpio-keys";
+
+		reset {
+			label = "reset";
+			gpios = <&gpio0 13 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
+
+		wps {
+			label = "wps";
+			gpios = <&gpio1 2 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_WPS_BUTTON>;
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		led_power: power { // PWR
+			function = LED_FUNCTION_POWER;
+			color = <LED_COLOR_ID_BLUE>;
+			gpios = <&gpio0 0 GPIO_ACTIVE_LOW>;
+		};
+
+		lan { // LAN
+			function = LED_FUNCTION_LAN;
+			color = <LED_COLOR_ID_BLUE>;
+			gpios = <&gpio0 19 GPIO_ACTIVE_LOW>;
+		};
+
+		wan { // WAN
+			function = LED_FUNCTION_WAN;
+			color = <LED_COLOR_ID_BLUE>;
+			gpios = <&gpio1 3 GPIO_ACTIVE_LOW>;
+		};
+
+		usb { // USB
+			function = LED_FUNCTION_USB;
+			color = <LED_COLOR_ID_BLUE>;
+			gpios = <&gpio1 0 GPIO_ACTIVE_LOW>;
+		};
+
+		// There are also two LEDs for Wi-Fi:
+		// - AIR2 (2.4GHz) can be turned on via <&gpio0 1 GPIO_ACTIVE_LOW>,
+		//	 but not off: the RT3352 iNIC coprocessor drives the same line
+		//	 independently, so it isn't declared as a controllable LED here.
+		// - AIR5 (5GHz) is tied directly to the wmac radio state and turns
+		//	 on/off automatically with it; no separate GPIO controls it.
+	};
+};
+
+&spi0 {
+	status = "okay";
+
+	flash@0 {
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <25000000>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "u-boot";
+				reg = <0x0 0x30000>;
+				read-only;
+			};
+
+			partition@30000 {
+				label = "config";
+				reg = <0x30000 0x10000>;
+				read-only;
+			};
+
+			partition@40000 {
+				label = "factory";
+				reg = <0x40000 0x10000>;
+				read-only;
+
+				nvmem-layout {
+					compatible = "fixed-layout";
+					#address-cells = <1>;
+					#size-cells = <1>;
+
+					eeprom_factory_0: eeprom@0 {
+						reg = <0x0 0x200>;
+					};
+
+					macaddr_factory_4: macaddr@4 {
+						reg = <0x4 0x6>;
+					};
+				};
+			};
+
+			partition@50000 {
+				compatible = "denx,uimage";
+				label = "firmware";
+				reg = <0x50000 0xfb0000>;
+			};
+		};
+	};
+};
+
+&gpio0 {
+	// Enables the LED power rail (all LEDs except PWR).
+	led_all_enable {
+		gpio-hog;
+		gpios = <10 GPIO_ACTIVE_HIGH>;
+		output-low;
+		line-name = "led-all-enable";
+	};
+};
+
+&gpio1 {
+	status = "okay";
+};
+
+&state_default {
+	gpio {
+		groups = "i2c", "jtag", "uartf";
+		function = "gpio";
+	};
+};
+
+&pinctrl {
+	mdio_pins: mdio {
+		mdio {
+			groups = "mdio";
+			function = "mdio";
+		};
+	};
+};
+
+&ethernet {
+	pinctrl-names = "default";
+	pinctrl-0 = <&mdio_pins>;
+
+	nvmem-cells = <&macaddr_factory_4>;
+	nvmem-cell-names = "mac-address";
+
+	port@0 {
+		status = "okay";
+		mediatek,fixed-link = <1000 1 1 1>;
+		phy-mode = "rgmii";
+	};
+
+	mdio0: mdio-bus {
+		status = "okay";
+	};
+};
+
+&pci {
+	status = "okay";
+};
+
+// PCI port connected to the onboard USB 3.0 host controller.
+&pci1 {
+	status = "okay";
+};
+
+&wmac {
+	ralink,2ghz = <0>;
+	nvmem-cells = <&eeprom_factory_0>;
+	nvmem-cell-names = "eeprom";
+};
--- a/target/linux/ramips/image/rt3883.mk
+++ b/target/linux/ramips/image/rt3883.mk
@@ -17,6 +17,17 @@ define Device/asus_rt-n56u
 endef
 TARGET_DEVICES += asus_rt-n56u
 
+define Device/asus_rt-n65u
+  $(Device/uimage-lzma-loader)
+  SOC := rt3883
+  BLOCKSIZE := 64k
+  IMAGE_SIZE := 16064k
+  DEVICE_VENDOR := ASUS
+  DEVICE_MODEL := RT-N65U
+  DEVICE_PACKAGES := kmod-usb3 kmod-usb-ledtrig-usbport kmod-switch-rtl8367b
+endef
+TARGET_DEVICES += asus_rt-n65u
+
 define Device/belkin_f9k1109v1
   $(Device/uimage-lzma-loader)
   SOC := rt3883
--- a/target/linux/ramips/rt3883/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/rt3883/base-files/etc/board.d/01_leds
@@ -7,6 +7,12 @@ board=$(board_name)
 board_config_update
 
 case $board in
+asus,rt-n65u)
+	ucidef_set_led_netdev "lan" "LAN" "blue:lan" "eth0.1" "tx rx"
+	ucidef_set_led_netdev "wan" "WAN" "blue:wan" "eth0.2" "tx rx"
+	ucidef_set_led_usbport "usb" "USB" "blue:usb" \
+		"usb1-port1" "usb1-port2" "usb2-port1" "usb2-port2"
+	;;
 belkin,f9k1109v1)
 	ucidef_set_led_netdev "lan" "lan" "blue:wps" "eth0"
 	;;
--- a/target/linux/ramips/rt3883/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/rt3883/base-files/etc/board.d/02_network
@@ -12,13 +12,14 @@ ramips_setup_interfaces()
 		ucidef_add_switch "switch0" \
 			"0:lan" "1:lan" "2:lan" "3:lan" "4:wan" "8@eth0"
 		;;
-	belkin,f9k1109v1)
-		ucidef_add_switch "switch0" \
-			"0:lan" "1:lan" "2:lan" "3:lan" "4:wan" "5@eth0"
-		;;
+	asus,rt-n65u|\
 	dlink,dir-645)
 		ucidef_add_switch "switch0" \
 			"1:lan" "2:lan" "3:lan" "4:lan" "0:wan" "6@eth0"
+		;;
+	belkin,f9k1109v1)
+		ucidef_add_switch "switch0" \
+			"0:lan" "1:lan" "2:lan" "3:lan" "4:wan" "5@eth0"
 		;;
 	edimax,br-6475nd)
 		ucidef_add_switch "switch0" \
@@ -53,6 +54,10 @@ ramips_setup_macs()
 	asus,rt-n56u)
 		lan_mac=$(macaddr_setbit_la "$(cat /sys/class/net/eth0/address)")
 		wan_mac=$(mtd_get_mac_binary factory 0x8004)
+		;;
+	asus,rt-n65u)
+		wan_mac=$(macaddr_add "$(mtd_get_mac_binary factory 0x4)" -4)
+		label_mac=$wan_mac
 		;;
 	belkin,f9k1109v1)
 		wan_mac=$(mtd_get_mac_ascii uboot-env HW_WAN_MAC)


andy9a9:
@zatrit I guess you didn't have to create a new MR, just re-use my :). Anyway once you have finished you can close my.

target/linux/ramips/image/rt3883.mk
```diff
@@ -17,6 +17,21 @@ define Device/asus_rt-n56u
 endef
 TARGET_DEVICES += asus_rt-n56u
 
+define Device/asus_rt-n65u
+  $(Device/uimage-lzma-loader)
+  SOC := rt3883
+  BLOCKSIZE := 64k
+  IMAGE_SIZE := 16384k
```
openwrt-ai: `IMAGE_SIZE` is set to the full 16 MiB flash size, but the `firmware` partition in the DTS is `reg = <0x50000 0xfb0000>` = 16064k (the first 0x50000 / 320k holds u-boot, config and factory). With `IMAGE_SIZE` larger than the firmware partition, the `check-size` step can't reject a sysupgrade image that would overflow the MTD partition. Set it to the firmware partition size:

```suggestion
  IMAGE_SIZE := 16064k
```

---

target/linux/ramips/dts/rt3883_asus_rt-n65u.dts
```diff
@@ -0,0 +1,200 @@
...
+
+				nvmem-layout {
+					compatible = "fixed-layout";
+					#address-cells = <1>;
+					#size-cells = <1>;
+
+					eeprom_factory_0: eeprom@0 {
+						reg = <0x0 0x200>;
+					};
+
```
openwrt-ai, line 119: nit: `eeprom_factory_8000` is declared but never referenced by any node (only `eeprom_factory_0` and `macaddr_factory_4` are used below). Since the 2.4 GHz iNIC is unsupported, this cell is currently dead. Either drop it, or keep it with a short comment noting it documents the 2.4 GHz calibration data for potential future use.

---

openwrt-ai (commented):
## Commit checks
- c9e8164 "ramips: add support for ASUS RT-N65U" — this new-device commit is missing the required **MAC address layout** section. The message never states where each interface's MAC comes from; from the diff it is: base/eth0 MAC from `factory` 0x4, LAN derived from `eth0` via `macaddr_setbit_la`, WAN from `factory` 0x8004. Please add one line per interface. The **Flash instructions** are also a single hand-wavy line ("Use stock firmware web UI to install sysupgrade image") with no verbatim UI steps, image name, or revert-to-stock note — a stranger could not reliably flash the device from the message alone. The hardware specification section looks good.

---

andy9a9:
Btw, is the switch working?

zatrit:
> Btw, is the switch working?

Yes, it does!

andy9a9:
> > Btw, is the switch working?
> 
> Yes, it does!

Wow :). Then what's the difference between my original variant as I wasn't able to make it working?

zatrit:
> > > Btw, is the switch working?
> > 
> > 
> > Yes, it does!
> 
> Wow :). Then what's the difference between my original variant as I wasn't able to make it working?

It was the pinctrl node for MDIO bus. Seems like it wasn't initialized properly in your PR.
I just looked at other boards with similar SoC and same switch.

openwrt-ai (commented):

The concerns from the previous review are resolved: the commit message now carries a MAC address layout section and full verbatim TFTP flash instructions. The other changes since then are sound — `IMAGE_SIZE` was corrected to 16064k to match the 0xfb0000 `firmware` partition, the unused `eeprom_factory_8000` node was dropped, and removing `UIMAGE_NAME` is consistent with the documented TFTP-recovery flow (any filename accepted).

---

openwrt-ai (commented):

Since the last review the branch was rebased onto current main (no change to the device-support files, which are byte-identical), and the commit message gained one line — `- Switch: Realtek RTL8367RB` — in the Specification section. That addition is consistent with the DTS (`realtek,rtl8367b`) and the `kmod-switch-rtl8367b` package, so no concerns.

---

openwrt-ai (commented):

Since the last review the branch was rebased onto current main. The device-support commit (DTS, `rt3883.mk`, `01_leds`, `02_network`) and its commit message are byte-identical to the previously reviewed revision; the only differences are the unrelated main changes the rebase pulled in (e.g. the mac80211 6.18.26 -> 6.18.39 bump). No concerns.

---

andy9a9:
> > > > Btw, is the switch working?
> > > 
> > > 
> > > Yes, it does!
> > 
> > 
> > Wow :). Then what's the difference between my original variant as I wasn't able to make it working?
> 
> It was the pinctrl node for MDIO bus. Seems like it wasn't initialized properly in your PR. I just looked at other boards with similar SoC and same switch.

ok, I had a chance to test your changes, but I would say, I don't see the difference between mine... switch is still not working. Or did I make something wrong?
```
[    2.354874] NET: Registered PF_INET6 protocol family
[    2.367014] Segment Routing with IPv6
[    2.370925] In-situ OAM (IOAM) with IPv6
[    2.375312] NET: Registered PF_PACKET protocol family
[    2.380558] 8021q: 802.1Q VLAN Support v1.8
[    2.414536] rtl8367b rtl8367rb: using MDIO bus 'mdio'
[    2.420078] rtl8367b rtl8367rb: unknown chip (num:6000 ver:2000)
[    2.426304] rtl8367b rtl8367rb: chip detection failed, err=-19
[    2.442295] clk: Disabling unused clocks
[    2.446684] check access for rdinit=/init failed: -2, ignoring
[    2.467252] VFS: Mounted root (squashfs filesystem) readonly on device 31:5.
[    2.480961] Freeing unused kernel image (initmem) memory: 1252K
[    2.487089] This architecture does not have kernel memory protection.
[    2.493709] Run /sbin/init as init process
[    7.330627] init: Console is alive
```
```
ifconfig -a
eth0      Link encap:Ethernet  HWaddr 50:46:5D:93:01:CC  
          BROADCAST MULTICAST  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)
          Interrupt:5 

lo        Link encap:Local Loopback  
          inet addr:127.0.0.1  Mask:255.0.0.0
          inet6 addr: ::1/128 Scope:Host
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:184 errors:0 dropped:0 overruns:0 frame:0
          TX packets:184 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:14352 (14.0 KiB)  TX bytes:14352 (14.0 KiB)
```
Are you sure, that is working? How the logs look like on your side?

zatrit:
I'll send my boot log soon. What's your router hardware revision (mine is v2.60)? It probably labeled on the board

zatrit:
Here's my router's boot log: https://gist.github.com/zatrit/315046fb784bbaed77261d86cd859222

```
ifconfig -a
br-lan    Link encap:Ethernet  HWaddr AE:22:0B:7D:30:54
          inet addr:192.168.2.1  Bcast:192.168.2.255  Mask:255.255.255.0
          inet6 addr: fe80::ac22:bff:fe7d:3054/64 Scope:Link
          inet6 addr: fd32:ba29:1e85::1/60 Scope:Global
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:107 errors:0 dropped:0 overruns:0 frame:0
          TX packets:97 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:12244 (11.9 KiB)  TX bytes:12656 (12.3 KiB)

eth0      Link encap:Ethernet  HWaddr AC:22:0B:7D:30:54
          inet6 addr: fe80::ae22:bff:fe7d:3054/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:865 errors:0 dropped:0 overruns:0 frame:0
          TX packets:841 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:95412 (93.1 KiB)  TX bytes:143981 (140.6 KiB)
          Interrupt:5

eth0.1    Link encap:Ethernet  HWaddr AE:22:0B:7D:30:54
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:107 errors:0 dropped:0 overruns:0 frame:0
          TX packets:97 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:12244 (11.9 KiB)  TX bytes:12656 (12.3 KiB)

eth0.2    Link encap:Ethernet  HWaddr AC:22:0B:7D:30:50
          inet6 addr: fe80::ae22:bff:fe7d:3050/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:0 errors:0 dropped:0 overruns:0 frame:0
          TX packets:117 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:0 (0.0 B)  TX bytes:32390 (31.6 KiB)

lo        Link encap:Local Loopback
          inet addr:127.0.0.1  Mask:255.0.0.0
          inet6 addr: ::1/128 Scope:Host
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:233 errors:0 dropped:0 overruns:0 frame:0
          TX packets:233 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:18172 (17.7 KiB)  TX bytes:18172 (17.7 KiB)
```

It looks like the switch MMIO bus address for your board is also correct, but the version differs (`num:6000 ver:2000`). Mine is `num:6000 ver:1000`, whereas when the bus address was incorrect, it showed `num:0000 ver:0000`.

Upd: there's no `case` for ver:2000: https://github.com/openwrt/openwrt/blob/fdb162fb8106a9af05bd02dbf28f9cd7e0ac4521/target/linux/generic/files/drivers/net/phy/rtl8367b.c#L1507

Probably, it's a different switch?

zatrit:
@andy9a9, can you add a case for `0x2000` to `target/linux/generic/files/drivers/net/phy/rtl8367b.c` like this:
```diff
diff --git a/target/linux/generic/files/drivers/net/phy/rtl8367b.c b/target/linux/generic/files/drivers/net/phy/rtl8367b.c
index 850392cc7c..9dd9d16003 100644
--- a/target/linux/generic/files/drivers/net/phy/rtl8367b.c
+++ b/target/linux/generic/files/drivers/net/phy/rtl8367b.c
@@ -1501,6 +1501,7 @@ static int rtl8367b_detect(struct rtl8366_smi *smi)
                }
                break;
        case 0x1000:
+       case 0x2000:
                chip_name = "8367RB";
                smi->rtl8367b_chip = RTL8367B_CHIP_RTL8367RB;
                break;
```
?

zatrit:
There's a U-Boot command that outputs the switch's version:
```
RT3883 # rtkswreg.r 0x1300
Realtek Switch register 0x1300 = 0x6000
RT3883 # rtkswreg.r 0x1301
Realtek Switch register 0x1301 = 0x1000
```

You can try this yourself to check whether your router's switch actually has `num:6000 ver:2000`.

andy9a9:
> ```diff
> diff --git a/target/linux/generic/files/drivers/net/phy/rtl8367b.c b/target/linux/generic/files/drivers/net/phy/rtl8367b.c
> index 850392cc7c..9dd9d16003 100644
> --- a/target/linux/generic/files/drivers/net/phy/rtl8367b.c
> +++ b/target/linux/generic/files/drivers/net/phy/rtl8367b.c
> @@ -1501,6 +1501,7 @@ static int rtl8367b_detect(struct rtl8366_smi *smi)
>                 }
>                 break;
>         case 0x1000:
> +       case 0x2000:
>                 chip_name = "8367RB";
>                 smi->rtl8367b_chip = RTL8367B_CHIP_RTL8367RB;
>                 break;
> ```
> 
> ?

Yes, I did it that way...
```
[    2.413850] rtl8367b rtl8367rb: using MDIO bus 'mdio'
[    2.419390] rtl8367b rtl8367rb: RTL8367RB chip found (num:6000 ver:2000)
[    3.517689] rtl8367b rtl8367rb: cpu_port:6, assigned to extif1
```
My problem was, that the detection always failed, but now ... :). I would say, this has to be added to patch as well.

zatrit:
Thank you for your help. I'll add this patch soon. Also, can you test this switch to ensure that it works properly with this driver?

zatrit:
Also, can you check whether any other RTL8367B variants work with your switch by adding `case 0x2000` before their respective case blocks, to make sure that RTL8367RB is the only correct one?

andy9a9:
> Thank you for your help. I'll add this patch soon. Also, can you test this switch to ensure that it works properly with this driver?

Yes, looks like that is working fine.

andy9a9:
> Also, can you check whether any other RTL8367B variants work with your switch by adding `case 0x2000` before their respective case blocks, to make sure that RTL8367RB is the only correct one?

No, the case has to be added just for this driver. I confirmed it by reading the registers in u-boot. So other cases shouldn't match.

zatrit:
> > Also, can you check whether any other RTL8367B variants work with your switch by adding `case 0x2000` before their respective case blocks, to make sure that RTL8367RB is the only correct one?
> 
> No, the case has to be added just for this driver. I confirmed it by reading the registers in u-boot. So other cases shouldn't match.

What I mean is that I'm not sure this is the only case where adding `case 0x2000` would make it work. I wanted to check whether your switch also works when handled as one of the other `RTL8367B` variants, since I'm not entirely sure it is actually an `RTL8367RB`, even though it works when treated as one.

Edit: The switch model might be possible to confirm from the chip marking on the board. I don't remember whether it is only located under the heatsink or also under the shield, but be careful when removing the shield.

zatrit:
By the way, what is your router's hardware revision? I need this information for a future ToH article describing the differences between hardware revisions. I would like to document the differences between the switch versions there.

andy9a9:
> What I mean is that I'm not sure this is the only case where adding `case 0x2000` would make it work. I wanted to check whether your switch also works when handled as one of the other `RTL8367B` variants, since I'm not entirely sure it is actually an `RTL8367RB`, even though it works when treated as one.

No you can't just use a different type. If it doesn't match, the initialization data doesn't work. 
> 
> Edit: The switch model might be possible to confirm from the chip marking on the board. I don't remember whether it is only located under the heatsink or also under the shield, but be careful when removing the shield.

I have the revision `2.60`, but that's the same as yours, isn't it?

target/linux/generic/files/drivers/net/phy/rtl8367b.c
```diff
@@ -1501,6 +1501,7 @@ static int rtl8367b_detect(struct rtl8366_smi *smi)
 		}
 		break;
 	case 0x1000:
+	case 0x2000:
```
openwrt-ai: The new arm accepts `ver:2000` for *any* `chip_num`, whereas the three arms directly above it ([`case 0x0010` / `0x0020` / `0x00A0`](https://github.com/openwrt/openwrt/blob/b9ecf7a40c9aaab8bfa316637e834ad61900625b/target/linux/generic/files/drivers/net/phy/rtl8367b.c#L1485-L1502)) only claim the chip when `chip_num` matches as well. Since this driver is built for every target that selects `kmod-switch-rtl8367b`, widening detection on the version register alone means unrelated silicon that happens to read back `ver:2000` now gets programmed with `rtl8367b_initvals` instead of failing cleanly with `-ENODEV`.

Both units reported in this PR read `num:6000`, and upstream's equivalent driver keys detection on the *pair* — [`ci->chip_id == chip_id && ci->chip_ver == chip_ver` at rtl8365mb.c:2088](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/net/dsa/realtek/rtl8365mb.c#L2088).

Was the unguarded form chosen deliberately to mirror the pre-existing `case 0x1000:` (which is equally unguarded), or would you rather gate `0x2000` on `chip_num == 0x6000` so a mismatching chip still bails out?

---

openwrt-ai (commented):
Reviewed 1 new commit (`b9ecf7a`). The ramips device-support commit and its files are unchanged since the last review — the only new content is the one-line `rtl8367b.c` detection change.

## Commit checks
- b9ecf7a "rtl8367b: add support for RTL8367RB with `ver:2000`" — the body attributes the variant to a hardware revision that does not actually distinguish it: "This RTL8367RB variant was found in the ASUS RT-N65U v2.60", but per the discussion in this PR the reporter's v2.60 unit reads `num:6000 ver:2000` while your own v2.60 unit reads `num:6000 ver:1000` ([your boot log](https://github.com/openwrt/openwrt/pull/24343#issuecomment-5083573600) vs. [his revision reply](https://github.com/openwrt/openwrt/pull/24343#issuecomment-5087310729)). Consider rewording so the message doesn't imply v2.60 == `ver:2000`, since both silicon revisions ship under that label.
- nit: Andrej Valek not only reported the failure but also [verified the fix on hardware](https://github.com/openwrt/openwrt/pull/24343#issuecomment-5084776380); a `Tested-by: Andrej Valek <andy@skyrain.eu>` trailer next to the existing `Reported-by:` would record that.

---

target/linux/generic/files/drivers/net/phy/rtl8367b.c
zatrit: This was done intentionally to match `case 0x1000`. However, it would be nice to add `chip_num` guard. What's the best way to do it?

target/linux/generic/files/drivers/net/phy/rtl8367b.c
```diff
@@ -1501,6 +1501,7 @@ static int rtl8367b_detect(struct rtl8366_smi *smi)
...
 		chip_name = "8367RB";
 		smi->rtl8367b_chip = RTL8367B_CHIP_RTL8367RB;
 		break;
```
openwrt-ai, line 1551: Answering your question from the [thread above](https://github.com/openwrt/openwrt/pull/24343#discussion_r3667619869) on how to add the `chip_num` guard: give `0x2000` its own arm and gate it the same way the `0x0010` / `0x0020` / `0x00A0` arms are gated, using the `num:6000` both units in this PR reported. Leaving the pre-existing `case 0x1000:` unguarded keeps this patch to a single new behaviour and avoids regressing boards whose `ver:1000` silicon reports some other `chip_num`.

```suggestion
	case 0x1000:
		chip_name = "8367RB";
		smi->rtl8367b_chip = RTL8367B_CHIP_RTL8367RB;
		break;
	case 0x2000:
		if (chip_num == 0x6000) {
			chip_name = "8367RB";
			smi->rtl8367b_chip = RTL8367B_CHIP_RTL8367RB;
		}
		break;
```

With this, a chip that reads back `ver:2000` but a different `chip_num` falls through to the `if (!chip_name)` check and still fails cleanly with `-ENODEV` instead of being programmed with `rtl8367b_initvals`.

---

openwrt-ai (commented):
Reviewed 1 new commit (`56b8935`). The tree is byte-identical to the previously reviewed revision — the only change is the amended message of the `rtl8367b` commit, which gained the `Tested-by: Andrej Valek <andy@skyrain.eu>` trailer suggested last time. CI on the head commit is green.

Left one inline comment answering your follow-up question about how to guard the new `case 0x2000:` on `chip_num`.

## Commit checks
- 56b8935 "rtl8367b: add support for RTL8367RB with `ver:2000`" — the body still reads "This RTL8367RB variant was found in the ASUS RT-N65U v2.60", which implies v2.60 identifies the `ver:2000` silicon; per this PR's discussion your own v2.60 unit reads `ver:1000`. Flagged previously and still open — reword or drop the revision if it isn't the distinguishing factor.

---

openwrt-ai (commented):
Reviewed 1 updated commit (`56a3c5d`). The ramips device-support commit (`f8c9718`) and its four files are byte-identical to the previously reviewed revision — the only real changes since the last review are in the `rtl8367b` commit, plus a rebase onto current main.

Both previously open points are resolved:

- The `case 0x2000:` arm is now gated on `chip_num == 0x6000`, so silicon that reads back `ver:2000` with a different chip number still fails cleanly with `-ENODEV` instead of being programmed with `rtl8367b_initvals`. This matches the guarding style of the `0x0010` / `0x0020` / `0x00A0` arms directly above it.
- The commit body no longer implies `v2.60` identifies the `ver:2000` silicon — it now states outright that units with `num:6000 ver:1000` have also been confirmed under that label.

The rebase pulled in main's [`kernel: rtl8367b: support post-init register writes`](https://github.com/openwrt/openwrt/commit/11c77ce98c3a5cb05b6ad3852dfe7511c8c43a62), which touches the same file; the two changes are in different functions and merged cleanly. CI on the head commit is green (all three FormalityCheck runs pass).

## Commit checks
- nit: 56a3c5d "rtl8367b: add support for RTL8367RB with `num:6000 ver:2000`" — the subject prefix is a bare `rtl8367b:`, which doesn't name a target or subsystem. The most recent in-tree commit touching this same file uses `kernel: rtl8367b: ...` ([11c77ce](https://github.com/openwrt/openwrt/commit/11c77ce98c3a5cb05b6ad3852dfe7511c8c43a62)). Was the shorter prefix deliberate, or would `kernel: rtl8367b: add support for RTL8367RB with num:6000 ver:2000` be preferable for consistency? (CI's commit-format check accepts either, so this is purely a convention question.)

---

target/linux/generic/files/drivers/net/phy/rtl8367b.c
andy9a9: I have a question here, what's the benefit to have the same part in the `if` statement? The section above have the same values...

zatrit: To not matching any other chip with `ver:2000` but any different `chip_num`

target/linux/ramips/image/rt3883.mk
```diff
@@ -17,6 +17,19 @@ define Device/asus_rt-n56u
...
 
+define Device/asus_rt-n65u
+  $(Device/uimage-lzma-loader)
+  SOC := rt3883
+  BLOCKSIZE := 64k
+  IMAGE_SIZE := 16064k
+  DEVICE_VENDOR := ASUS
+  DEVICE_MODEL := RT-N65U
+  DEVICE_PACKAGES := kmod-usb-ohci kmod-usb2 kmod-usb3 \
+    kmod-usb-ledtrig-usbport kmod-switch-rtl8366-smi kmod-switch-rtl8367b
```
hauke: Are `kmod-usb-ohci` and `kmod-usb2` needed?  The DTS only enables the PCIe bridge to the ASM1042, it never activates `&ehci` / `&ohci`, which are `status = "disabled"` in `rt3883.dtsi`.

Please cechk with a USB 3 and a USB 2 device, probably ASM1042 translates EHCI (USBE) to USB3.

`kmod-switch-rtl8366-smi` is not needed here, `kmod-switch-rtl8367b` already includes it.

target/linux/ramips/image/rt3883.mk
```diff
@@ -17,6 +17,19 @@ define Device/asus_rt-n56u
...
+  SUPPORTED_DEVICES += rt-n65u
```
hauke: Why do you need this line?

It is normally needed when you use sysupgrade to upgrade from a vendor firmware which uses OpenWrt as a base to official OpenWrt and the Vendor firmware expects ` rt-n65u`.

target/linux/ramips/rt3883/base-files/etc/board.d/02_network
```diff
@@ -50,7 +51,8 @@ ramips_setup_macs()
 	local label_mac=""
 
 	case $board in
-	asus,rt-n56u)
+	asus,rt-n56u|\
+	asus,rt-n65u)
 		lan_mac=$(macaddr_setbit_la "$(cat /sys/class/net/eth0/address)")
 		wan_mac=$(mtd_get_mac_binary factory 0x8004)
 		;;
```
hauke, line 57: your commit message says that you have a label mac.
I think you should copy this block for the new device and add: `label_mac=$wan_mac`.

If `asus,rt-n56u` has the same label mac, you can also just modify this block.

target/linux/ramips/rt3883/base-files/etc/board.d/01_leds
```diff
@@ -17,6 +17,12 @@ omnima,hpm)
 	ucidef_set_led_netdev "eth" "ETH" "green:eth" "eth0"
 	ucidef_set_led_netdev "wifi_led" "wifi" "green:wifi" "wlan0"
 	;;
+asus,rt-n65u)
+	ucidef_set_led_netdev "lan" "LAN" "blue:lan" "eth0.1" "tx rx"
+	ucidef_set_led_netdev "wan" "WAN" "blue:wan" "eth0.2" "tx rx"
+	ucidef_set_led_usbport "usb" "USB" "blue:usb" \
+		"usb1-port1" "usb1-port2" "usb2-port1" "usb2-port2"
+	;;
```
hauke: Please add this in alphabetical order.

target/linux/ramips/rt3883/base-files/etc/board.d/02_network
```diff
@@ -16,7 +16,8 @@ ramips_setup_interfaces()
 		ucidef_add_switch "switch0" \
 			"0:lan" "1:lan" "2:lan" "3:lan" "4:wan" "5@eth0"
 		;;
-	dlink,dir-645)
+	dlink,dir-645|\
+	asus,rt-n65u)
```
hauke: Please add this in alphabetical order

hauke:
Please change the order of the commits, add the driver change first and then the new device.

target/linux/generic/files/drivers/net/phy/rtl8367b.c
```diff
@@ -1549,6 +1549,12 @@ static int rtl8367b_detect(struct rtl8366_smi *smi)
...
+	case 0x2000:
+		if (chip_num == 0x6000) {
+			chip_name = "8367RB";
+			smi->rtl8367b_chip = RTL8367B_CHIP_RTL8367RB;
+		}
+		break;
```
hauke, line 1557: Are you sure it is a `8367RB`? Please check what is printed on the chip itself.

openwrt-ai (commented):
Reviewed 1 updated commit (`f14779a`); no new issues found.

Both files touched by this PR are byte-identical to the previously reviewed revision (`56a3c5d`) — `git diff 56a3c5d..f14779a -- target/linux/ramips target/linux/generic/files/drivers/net/phy/rtl8367b.c` is empty. The only real change is the switch commit's subject, which gained the `kernel:` prefix suggested last time (`rtl8367b: add support for RTL8367RB with num:6000 ver:2000` -> `kernel: rtl8367b: add support for RTL8367RB with \`ver:2000\``), matching the prefix used by [11c77ce](https://github.com/openwrt/openwrt/commit/11c77ce98c3a5cb05b6ad3852dfe7511c8c43a62), the most recent in-tree commit touching that driver. Everything else in the range is unrelated main content pulled in by the rebase.

CI on the head commit is green (all three FormalityCheck runs pass; the kernel/package build matrix from the previous run completed successfully).

The open review threads on this revision are @hauke's — `DEVICE_PACKAGES` (`kmod-usb-ohci` / `kmod-usb2` / redundant `kmod-switch-rtl8366-smi`), `SUPPORTED_DEVICES`, the missing `label_mac` in `02_network`, alphabetical ordering in `01_leds` / `02_network`, and the chip-marking question on `rtl8367b.c`. I have nothing to add to those.

---

target/linux/generic/files/drivers/net/phy/rtl8367b.c
andy9a9: Yes, but the `0x1000` is exactly the same. So I still don't see the reason why it has to be handled differently.

target/linux/generic/files/drivers/net/phy/rtl8367b.c
zatrit: @andy9a9, please check the marking on the switch chip.

andy9a9: Image

target/linux/ramips/rt3883/base-files/etc/board.d/02_network
zatrit: I accidentally used the wrong MAC layout. I assumed that the RT-N65U had the same MAC layout as the RT-N56U, but that assumption was incorrect. It now matches the stock behavior.

target/linux/ramips/rt3883/base-files/etc/board.d/02_network
```diff
@@ -54,6 +55,9 @@ ramips_setup_macs()
 		lan_mac=$(macaddr_setbit_la "$(cat /sys/class/net/eth0/address)")
 		wan_mac=$(mtd_get_mac_binary factory 0x8004)
 		;;
+  asus,rt-n65u)
+		label_mac=$(macaddr_add "$(mtd_get_mac_binary factory 0x4)" -4)
+    ;;
```
hauke: Please replace the spaces with tabs.

hauke: Now all interfaces get the same mac address.
Are you sure the vendor firmware sets LAN and WAN to the same mac address?

I think you have to set the WAN mac address here.

target/linux/ramips/rt3883/base-files/etc/board.d/02_network
```diff
@@ -54,6 +55,9 @@ ramips_setup_macs()
 		lan_mac=$(macaddr_setbit_la "$(cat /sys/class/net/eth0/address)")
 		wan_mac=$(mtd_get_mac_binary factory 0x8004)
 		;;
+  asus,rt-n65u)
+		label_mac=$(macaddr_add "$(mtd_get_mac_binary factory 0x4)" -4)
```
openwrt-ai: `label_mac` is derived arithmetically from the ethernet MAC, but that value is already stored in flash and can be read directly. Your [boot log](https://github.com/openwrt/openwrt/pull/24343#issuecomment-5083573600) shows `eth0` (= `factory` 0x4) as `AC:22:0B:7D:30:54`, while `eth0.2` — which at the time came from `mtd_get_mac_binary factory 0x8004` — was `AC:22:0B:7D:30:50`, exactly what `macaddr_add ... -4` computes here. 0x8004 is the MAC in the second EEPROM at `factory` 0x8000 (the RT3352 iNIC's calibration area), i.e. the 2.4 GHz MAC that your commit message identifies as the label MAC.

Reading it directly is more robust than assuming the -4 offset holds on every unit, and it matches how [`asus,rt-n56u` reads that same offset on line 56](https://github.com/openwrt/openwrt/blob/e562220be7996d02527531fbc2ffb86fb7642698/target/linux/ramips/rt3883/base-files/etc/board.d/02_network#L56).

```suggestion
		label_mac=$(mtd_get_mac_binary factory 0x8004)
```

(Independent of @hauke's question below about also setting `wan_mac`.)

---

target/linux/ramips/rt3883/base-files/etc/board.d/02_network
```diff
@@ -16,6 +16,7 @@ ramips_setup_interfaces()
 		ucidef_add_switch "switch0" \
 			"0:lan" "1:lan" "2:lan" "3:lan" "4:wan" "5@eth0"
 		;;
+	asus,rt-n65u|\
```
openwrt-ai, line 15: nit: the alternation was reordered so the block now leads with `asus,rt-n65u`, but the block itself still sits after [`belkin,f9k1109v1)` on line 15](https://github.com/openwrt/openwrt/blob/e562220be7996d02527531fbc2ffb86fb7642698/target/linux/ramips/rt3883/base-files/etc/board.d/02_network#L15), so the `case` list is no longer alphabetically sorted — previously it sorted correctly under `dlink,dir-645`. To get what @hauke asked for, move the whole `asus,rt-n65u|dlink,dir-645)` block above the `belkin,f9k1109v1)` block. (`ramips_setup_macs()` below is correctly ordered.)

---

openwrt-ai (commented):
Reviewed the changes since `f14779a`. The commit order was swapped as @hauke asked (driver change `1d1162c` now precedes the device commit), `SUPPORTED_DEVICES` and the redundant `kmod-usb-ohci` / `kmod-usb2` / `kmod-switch-rtl8366-smi` entries were dropped from `rt3883.mk`, `01_leds` is now alphabetically ordered, and `02_network` gained a dedicated `asus,rt-n65u)` MAC case. `rtl8367b.c` and the DTS are byte-identical to the previously reviewed revision. CI on the head commit has not reported yet (no check runs), so nothing to ground there.

Two inline comments below on `02_network`, plus the commit-message points here. @hauke's two open threads on that file (tabs vs. spaces, and LAN/WAN sharing a MAC) are not repeated.

## Commit checks
- e562220 "ramips: add support for ASUS RT-N65U" — the rewritten **MAC address layout** section has the sign inverted relative to the code. It now reads `WAN, Wi-Fi 5GHz & LAN: label MAC - 4`, but the diff sets `label_mac = macaddr_add(factory 0x4, -4)` while `&ethernet` takes `macaddr_factory_4`, so LAN/WAN/5 GHz = `factory` 0x4 = **label MAC + 4**. Your [boot log](https://github.com/openwrt/openwrt/pull/24343#issuecomment-5083573600) confirms the direction: `eth0` is `AC:22:0B:7D:30:54` and the 0x8004 value is `AC:22:0B:7D:30:50`. The previous revision of this message said `Wi-Fi: WAN MAC + 4`, which was the right direction — the `- 4` looks like it was flipped during the rewrite.
- nit: e562220 — `Tested-by: 4PDA users` was added in this revision, but it isn't a well-formed trailer: `Tested-by:` takes a `Name <email>` identifying a person, same as the `Signed-off-by:` lines beside it. If you want to credit the forum testers, a prose line in the body (e.g. "Tested by users on the 4PDA forum.") keeps the trailer block valid.

---

zatrit:
Sorry for keeping you waiting for a week. I double-checked the stock behavior and found out that my instance of this model was configured to work as a switch, which is why the LAN has the same MAC address as the WAN.

I have corrected the MAC layout now, and it matches the stock behavior after a factory reset.

target/linux/ramips/rt3883/base-files/etc/board.d/02_network
zatrit: I don't think it's a good idea: the `factory` partition region starting at 0x8000 probably contains data related to 2.4 GHz Wi-Fi.

openwrt-ai (commented):
Reviewed 1 updated commit (`80ee16d`). The only file change since `e562220` is `02_network`: @hauke's two points are addressed (spaces replaced with tabs, and `wan_mac` is now set explicitly so LAN and WAN no longer share a MAC), and the `case` block in `ramips_setup_interfaces()` was moved above `belkin,f9k1109v1)` so the list is alphabetical again. The DTS, `rt3883.mk`, `01_leds` and `rtl8367b.c` are byte-identical to the previously reviewed revision. The `Tested-by: 4PDA users` pseudo-trailer was turned into a prose line, so the trailer block is well-formed now.

CI has not reported on the head commit yet (no check runs), so nothing to ground there.

I did not repeat my earlier `label_mac` suggestion — you've answered it in [that thread](https://github.com/openwrt/openwrt/pull/24343#discussion_r3699550880).

## Commit checks
- 80ee16d "ramips: add support for ASUS RT-N65U" — the **MAC address layout** section is only half-corrected. The WAN line is now right, but the second line still has the sign inverted:

  ```
  - Wi-Fi 2.4GHz & WAN: label MAC
  - Wi-Fi 5GHz & LAN: label MAC - 4
  ```

  The code defines the label MAC as *four below* `factory` 0x4 — [`wan_mac=$(macaddr_add "$(mtd_get_mac_binary factory 0x4)" -4)` / `label_mac=$wan_mac` at 02_network:58-60](https://github.com/openwrt/openwrt/blob/80ee16da5b4c54738a0b6a9a782b3c79800d9fef/target/linux/ramips/rt3883/base-files/etc/board.d/02_network#L58-L60) — while LAN takes `factory` 0x4 unmodified ([`nvmem-cells = <&macaddr_factory_4>` on `&ethernet`, dts:169](https://github.com/openwrt/openwrt/blob/80ee16da5b4c54738a0b6a9a782b3c79800d9fef/target/linux/ramips/dts/rt3883_asus_rt-n65u.dts#L169)) and 5 GHz reads its MAC from the same offset inside [`eeprom_factory_0`, dts:194](https://github.com/openwrt/openwrt/blob/80ee16da5b4c54738a0b6a9a782b3c79800d9fef/target/linux/ramips/dts/rt3883_asus_rt-n65u.dts#L194). So LAN/5 GHz = label MAC **+ 4**, not `- 4`. Your [boot log](https://github.com/openwrt/openwrt/pull/24343#issuecomment-5083573600) shows the same direction (`eth0` = `AC:22:0B:7D:30:54`, `factory` 0x8004 = `AC:22:0B:7D:30:50`). The PR description carries the identical wrong line and needs the same edit.

---

namiltd:
Do you have a picture of the print on the Realtek RTL8367RB chip (0x2000)?

openwrt-ai (commented):
Reviewed 1 updated commit (`8def877`); no new issues found.

The working tree is byte-identical to the previously reviewed revision (`80ee16d`) — `git diff 80ee16d..8def877` is empty. The only change is one line in the device commit message, which resolves the point from my last review:

```
-- Wi-Fi 5GHz & LAN: label MAC - 4
+- Wi-Fi 5GHz & LAN: label MAC + 4
```

That now matches the code: `wan_mac=$(macaddr_add "$(mtd_get_mac_binary factory 0x4)" -4)` with `label_mac=$wan_mac` puts the label MAC four below `factory` 0x4, while `&ethernet` takes `macaddr_factory_4` unmodified — so LAN/5 GHz is label MAC **+ 4**. The PR description carries the corrected line as well.

CI has not reported on the head commit yet (no check runs), so nothing to ground there.

---

zatrit:
> Do you have a picture of the print on the Realtek RTL8367RB chip (0x2000)?

@andy9a9 provided it here: https://github.com/openwrt/openwrt/pull/24343#discussion_r3698192470

PR #24344, opened 2026-07-21
build: kconfig: show what enabling an option pulls in, on its help page

mirko:
menuconfig's help page for a symbol lists what it selects ("Selects:") and 
what selects it ("Selected by:"), but not what enabling a plain option pulls
in as a dependency.

In OpenWrt an optional dependency "+SYM:pkg" in a package's DEPENDS is emitted
as "select PACKAGE_pkg if SYM" on the *package* symbol - owned by the package,
not by the option symbols - so nothing about it shows in the symbol's own help.

dropbear illustrates the workaround this forces: "+DROPBEAR_ZLIB:zlib" in
DEPENDS already pulls zlib, yet Config.in *also* carries a redundant
"select PACKAGE_zlib" on the DROPBEAR_ZLIB option - only so that option's help
advertises "Selects: PACKAGE_zlib". Several packages duplicate a dependency
this way.
Alternatively, packages put static strings into the help pages, hinting in
human-readable form what the options pulls in, again, redudantly (and possibly
wrong / not up to date).

There're other ways avoiding the "+SYM:pkg"-syntax inside DEPENDS entirely,
going only via Kconfig's "config SYM \n select PACKAGE_pkg", e.g. by openssl.
While metadata creation in tmp/ results in the same for the two approaches,
only the "+SYM:pkg"-syntax:
 - ensures the dependencies are being built first
 - the dependeny metadata ends up in the final package
 - allows depending on a "PROVIDES"

This change:
Adds a reverse lookup to get_symbol_str(): for the symbol being shown, scan
every other symbol's P_SELECT properties and list those whose *condition*
references it, under "Selects when enabled:". This surfaces the pulled-in
dependency on the option's own help page from the "+SYM:pkg" wiring alone,
making the redundant "select" unnecessary.

### DIFF
--- a/scripts/config/menu.c
+++ b/scripts/config/menu.c
@@ -784,10 +784,134 @@ static void get_symbol_props_str(struct gstr *r, struct symbol *sym,
 }
 
 /*
+ * Like expr_contains_symbol(), but tracks negation: report only occurrences of
+ * SYM with the requested polarity ('positive' flips on every E_NOT). Symbols
+ * inside comparisons have no meaningful polarity - whether they count is the
+ * caller's choice via 'count_cmp', because the two users of this helper need
+ * opposite conservatism: the trigger test must ignore them (the provider-
+ * fallback selects the metadata generator emits, "select fallback if
+ * preferred<depender", would otherwise show up as a bogus "enabling the
+ * preferred provider pulls in the fallback"), while the suppression test must
+ * count them (an uncertain reference keeps the entry hidden rather than
+ * risking a wrong one).
+ */
+static bool expr_contains_symbol_pol(struct expr *dep, struct symbol *sym,
+				     bool positive, bool count_cmp)
+{
+	if (!dep)
+		return false;
+
+	switch (dep->type) {
+	case E_AND:
+	case E_OR:
+		return expr_contains_symbol_pol(dep->left.expr, sym, positive,
+						count_cmp) ||
+		       expr_contains_symbol_pol(dep->right.expr, sym, positive,
+						count_cmp);
+	case E_NOT:
+		return expr_contains_symbol_pol(dep->left.expr, sym, !positive,
+						count_cmp);
+	case E_SYMBOL:
+		return positive && dep->left.sym == sym;
+	case E_EQUAL:
+	case E_GEQ:
+	case E_GTH:
+	case E_LEQ:
+	case E_LTH:
+	case E_UNEQUAL:
+		return count_cmp &&
+		       (dep->left.sym == sym ||
+			dep->right.sym == sym);
+	default:
+		;
+	}
+	return false;
+}
+
+/*
+ * OpenWrt: reverse "select ... if <this>" lookup. An optional dependency wired
+ * as '+SYM:pkg' in a package's DEPENDS is emitted by the metadata generator as
+ * 'select PACKAGE_pkg if SYM' on the *package* symbol, so the option SYM's own
+ * help page never shows what enabling it pulls in. Scan every symbol's P_SELECT
+ * properties and list those whose condition references SYM, right where the
+ * user toggles it - no redundant help text or select-on-option needed.
+ *
+ * menu_finalize() folds a symbol's base dependencies (menu->dep) into every
+ * property's visible.expr, which blurs the picture twice, and polarity is what
+ * restores it. First: once SYM appears anywhere in the selecting package's
+ * base dependency - e.g. in the guards the metadata generator emits when a
+ * pulled-in target carries constraints of its own ('!(SYM && ...) ||
+ * CONSTRAINT') - EVERY select of that package matches a plain contains-test
+ * on visible.expr, listing the package's whole dependency footprint down to
+ * libc. A genuine 'select pkg if SYM' carries SYM *positively* in the folded
+ * condition, the folded-in guards only *negated* - so require a positive
+ * occurrence in visible.expr. Second: a package that merely 'depends on SYM'
+ * folds a positive SYM into all its select conditions, which would pass that
+ * test - but there SYM is also positive in menu->dep itself, while the guard
+ * shapes reference it only negated. So additionally suppress the entry when
+ * menu->dep contains SYM in positive polarity. (A package that both 'depends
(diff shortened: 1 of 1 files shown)


scripts/config/menu.c
```diff
@@ -827,6 +827,38 @@ static void get_symbol_str(struct gstr *r, struct symbol *sym,
...
+		struct property *prop2;
+		bool hit = false;
+		int i;
+
+		for_all_symbols(i, sym2) {
+			if (sym2 == sym)
+				continue;
+			for_all_properties(sym2, prop2, P_SELECT) {
+				if (!prop2->visible.expr ||
+				    !expr_contains_symbol(prop2->visible.expr, sym))
```
openwrt-ai: `prop2->visible.expr` is broader than the select's own `if <cond>`. In `menu_finalize()` every property's `visible.expr` gets the selecting symbol's enclosing dependencies folded in — `basedep` (the symbol's own `depends on` plus parent-menu/if deps) is AND-combined into it: [menu.c#L379-L385](https://github.com/openwrt/openwrt/blob/cb384bc024f13a0fbc7f379cd1b996b88d58590c/scripts/config/menu.c#L381).

As a result `expr_contains_symbol(prop2->visible.expr, sym)` also matches when `sym` merely appears in the selecting package's `depends on`, not only in the intended `+SYM:pkg` if-condition. Such a symbol would then get a "Selects when enabled:" entry even though toggling it does not trigger the select — a false positive that could be noisy for broadly-depended-on symbols (e.g. a target/arch symbol that many packages `depends on`). Is that intended, or should the lookup isolate the select's original condition rather than the finalized `visible.expr`?

---

scripts/config/menu.c
```diff
@@ -827,6 +827,38 @@ static void get_symbol_str(struct gstr *r, struct symbol *sym,
...
 
+	/*
+	 * OpenWrt: reverse "select ... if <this>" lookup. An optional dependency
+	 * wired as '+SYM:pkg' in a package's DEPENDS is emitted by the metadata
+	 * generator as 'select PACKAGE_pkg if SYM' on the *package* symbol, so the
+	 * option SYM's own help page never shows what enabling it pulls in. List
+	 * every foreign select whose condition references this symbol, right where
+	 * the user toggles it - no redundant help text or select-on-option needed.
+	 */
+	if (sym) {
```
openwrt-ai: nit: the `if (sym)` guard is dead — a few lines up `get_symbol_props_str(r, sym, P_SELECT, ...)` and `sym->rev_dep.expr` already dereference `sym` unconditionally, so control never reaches here with a NULL `sym`. The guard reads as if NULL were possible; consider dropping it to stay consistent with the surrounding code.

---

openwrt-ai (commented):
Reviewed 1 new commit; see inline comments.

---

openwrt-ai (commented):

---

scripts/config/menu.c
```diff
@@ -783,11 +783,130 @@ static void get_symbol_props_str(struct gstr *r, struct symbol *sym,
...
+ * multi-binary package whose binaries share the option carries the dep on each
+ * binary - which would print the target once per selecting package. Render each
+ * entry and skip it when an identical one was already emitted. (Two selects of
+ * the same target under different conditions likewise collapse to one line; the
+ * condition is not printed anyway.)
+ *
+ * This walks the whole symbol table, so callers do it only for the focused help
+ * view, not for the per-match aggregation in get_relations_str().
+ */
+/*
```
openwrt-ai, line 786: The big block comment above (lines 786-818) documents `get_symbol_reverse_selects_str()`, but this commit inserted `expr_contains_symbol_pol()` plus its own comment between the two, so that comment now reads as the doc for the helper — and [`get_symbol_reverse_selects_str()` at menu.c:864](https://github.com/openwrt/openwrt/blob/4c6a2f7009668ac91b5f17232355e8668cfbf73b/scripts/config/menu.c#L864) is left with no comment directly above it. Two block comments back to back with nothing between them is also unlike the rest of the file, where every comment sits directly on its function.

Please move the helper (its comment at 819-830 and the function body at 831-862) above the `get_symbol_reverse_selects_str()` block comment, i.e. to just after `get_symbol_props_str()` at line 784.

---

scripts/config/menu.c
```diff
@@ -783,11 +783,130 @@ static void get_symbol_props_str(struct gstr *r, struct symbol *sym,
...
+ * pulled-in target carries constraints of its own ('!(SYM && ...) ||
+ * CONSTRAINT') - EVERY select of that package matches a plain contains-test
+ * on visible.expr, listing the package's whole dependency footprint down to
+ * libc. A genuine 'select pkg if SYM' carries SYM *positively* in the folded
+ * condition, the folded-in guards only *negated* - so require a positive
+ * occurrence in visible.expr. Second: a package that merely 'depends on SYM'
+ * folds a positive SYM into all its select conditions, which would pass that
+ * test - but there SYM is also positive in menu->dep itself, while the guard
+ * shapes reference it only negated. So additionally suppress the entry when
+ * menu->dep contains SYM in positive polarity.
```
openwrt-ai: nit: this rewrite dropped the acknowledgement of the false negative the previous revision spelled out — "(A package that both 'depends on SYM' and 'select pkg if SYM' is suppressed too - accepted as a rare corner.)". The behaviour is unchanged by the polarity switch: `depends on SYM` folds SYM positively into `menu->dep`, so the suppression still fires and hides a genuine `select pkg if SYM` on the same package. Worth keeping documented so a future reader doesn't rediscover it as a bug.

FWIW the corner really is empty today — grepping `package/` and `target/` for a `DEPENDS` line that names the same symbol both as `@SYM` and as `+SYM:pkg` yields no hits.

```suggestion
 * menu->dep contains SYM in positive polarity. (A package that both
 * 'depends on SYM' and 'select pkg if SYM' is suppressed too - accepted as a
 * rare corner; no in-tree package currently combines '@SYM' and '+SYM:pkg'.)
```

---

openwrt-ai (commented):
Reviewed 1 new commit; see inline comments.

---

openwrt-ai (commented):

---

PR #24345, opened 2026-07-21
siflower: bump to 6.18

981213:
Since I have the only two boards supported and tested on both, I decided to skip the testing kernel and switch directly to 6.18 instead.


openwrt-ai (commented):
Reviewed 6 new commits; no new issues found.

---

openwrt-ai (commented):
## Commit checks

- 48c8897 "siflower: switch to kernel 6.18 and drop 6.12" — the newly-added body says "Switch to 6.12 directly without going through a testing phase", but the commit switches *to* 6.18 and drops 6.12 (subject and PR description both say 6.18). Looks like a typo for "Switch to 6.18 directly".

The only change since my last review at 9ee1a14 is added commit-message bodies (the code diffs are identical); the two other added bodies match their changes.

---

target/linux/siflower/files-6.18/drivers/net/ethernet/siflower/sfxgmac-dma.c
```diff
@@ -1468,22 +1470,26 @@ static int xgmac_dma_probe(struct platform_device *pdev)
...
+	if (ret)
+		return ret;
 
 	/* we run multiple netdevs on the same DMA ring so we need a dummy
 	 * device for NAPI to work
 	 */
-	init_dummy_netdev(&priv->napi_dev);
+	priv->napi_dev = alloc_netdev_dummy(0);
+	if (!priv->napi_dev)
+		return -ENOMEM;
```
openwrt-ai, line 1481: nit: the `init_dummy_netdev()` -> `alloc_netdev_dummy()` conversion also dropped `strscpy(priv->napi_dev.name, KBUILD_MODNAME, IFNAMSIZ)` (removed from the `CONFIG_NET_SIFLOWER_ETH_RX_THREAD` block below, line 1527). `alloc_netdev_dummy()` names the device with the literal template string `"dummy#"` ([`alloc_netdev(sizeof_priv, "dummy#", ...)` at net/core/dev.c:12146-12150](https://github.com/gregkh/linux/blob/aae95ae9d02e4d1b1c37c05fda1c98a3ba0d9ee2/net/core/dev.c#L12146-L12150)), and it is never registered so the `#` is never expanded. The threaded-NAPI kthreads are named from that field ([`kthread_run(napi_threaded_poll, n, "napi/%s-%d", n->dev->name, ...)` at net/core/dev.c:1649](https://github.com/gregkh/linux/blob/aae95ae9d02e4d1b1c37c05fda1c98a3ba0d9ee2/net/core/dev.c#L1649)), so with RX threading enabled they now show up as `napi/dummy#-<id>` instead of `napi/sfxgmac-dma-<id>`. Cosmetic, but it makes the threads unidentifiable in `ps`/`top` and un-pinnable by name.

`mlxsw` keeps the name after switching to the allocated dummy ([pci.c:148-152](https://github.com/gregkh/linux/blob/aae95ae9d02e4d1b1c37c05fda1c98a3ba0d9ee2/drivers/net/ethernet/mellanox/mlxsw/pci.c#L148-L152)); same thing works here:

```suggestion
	priv->napi_dev = alloc_netdev_dummy(0);
	if (!priv->napi_dev)
		return -ENOMEM;

	strscpy(priv->napi_dev->name, KBUILD_MODNAME,
		sizeof(priv->napi_dev->name));

```

---

openwrt-ai (commented):
Re-reviewed at `aba781d` (previous review was at `48c8897`).

The force-push since my last review changed commit messages only — `git diff` of the PR content against each base is byte-identical, so no code changed. Both message issues are resolved: the `Switch to 6.12` typo in the final commit now reads `6.18`, and the first two commits' subject prefix was corrected from `kernel/siflower:` to `siflower:`. All six commit messages now match their diffs.

One nit inline on a pre-existing hunk that I missed on the earlier pass. CI on `aba781d` is green (siflower/sf21 kernel build and patch check both passed).

---

PR #24347, opened 2026-07-21
package: uhttpd: add force_http_auth option

wevsty:
Add support for the force_http_auth option in the uhttpd init script.

This option allows users to enforce HTTP basic auth, preventing CGI scripts installed by third-party LuCI packages from bypassing authentication.

Link: https://github.com/openwrt/luci/pull/8852

### DIFF
--- a/package/network/services/uhttpd/files/uhttpd.config
+++ b/package/network/services/uhttpd/files/uhttpd.config
@@ -12,6 +12,9 @@ config uhttpd main
 	# Redirect HTTP requests to HTTPS if possible
 	option redirect_https	0
 
+	# Force HTTP basic authentication for all pages using the root account.
+	option force_http_auth	0
+
 	# Server document root
 	option home		/www
 
--- a/package/network/services/uhttpd/files/uhttpd.init
+++ b/package/network/services/uhttpd/files/uhttpd.init
@@ -77,7 +77,24 @@ create_httpauth() {
 		return
 	fi
 	echo "${prefix}:${username}:${password}" >>$httpdconf
-	haveauth=1
+	have_auth=1
+}
+
+# Check if the root password is valid.
+check_root_password_valid() {
+	local pass_hash
+	pass_hash=$(awk -F: '$1 == "root" {print $2}' /etc/shadow 2>/dev/null)
+
+	case "$pass_hash" in
+		""|\!*|\**)
+			# Invalid
+			return 1
+			;;
+		*)
+			# Valid
+			return 0
+			;;
+	esac
 }
 
 append_lua_prefix() {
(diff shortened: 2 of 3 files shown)


package/network/services/uhttpd/files/uhttpd.init
```diff
@@ -122,18 +139,31 @@ start_instance()
...
+		procd_append_param command -c ${httpdconf}
+		[ -r /etc/httpd.conf ] && cat /etc/httpd.conf >> ${httpdconf}
+	elif [ -n "$config" ]; then
+		# Configured a custom config file.
+		procd_append_param command -c ${config}
+	fi
+
 	append_arg "$cfg" home "-h"
 	append_arg "$cfg" realm "-r" "${realm:-OpenWrt}"
 	append_arg "$cfg" config "-c"
```
openwrt-ai: The `config` option is now emitted by the new `if [ -n "$httpdconf" ] ... elif [ -n "$config" ]` block above, but this `append_arg` still also emits `-c ${config}`. For an instance that sets `option config`, uhttpd receives `-c` twice. Worse, when auth is active (httpauth entries or `force_http_auth=1`) together with a custom `config`: the `if` branch emits `-c ${httpdconf}` and then this line appends `-c ${config}`, so the generated auth config is overridden and the enforced authentication is silently dropped. Since the new block now fully handles the `config` case, this redundant `append_arg "$cfg" config "-c"` should be dropped.

---

package/network/services/uhttpd/files/uhttpd.init
```diff
@@ -111,7 +128,7 @@ start_instance()
 
 	local cfg="$1"
 	local realm="$(uci_get system.@system[0].hostname)"
-	local listen http https interpreter indexes path handler httpdconf haveauth
+	local listen http https interpreter indexes path handler httpdconf have_auth
```
openwrt-ai: nit: `force_http_auth` is read via `config_get_bool` further down but isn't in this `local` list, so it leaks into the global shell scope across the `config_foreach` loop over instances (unlike `have_auth`, which you correctly made local here).

```suggestion
	local listen http https interpreter indexes path handler httpdconf have_auth force_http_auth
```

---

package/network/services/uhttpd/files/uhttpd.config
```diff
@@ -12,6 +12,9 @@ config uhttpd main
 	# Redirect HTTP requests to HTTPS if possible
 	option redirect_https	0
 
+	# Force HTTP basic authentication for all pages using the root account.
+	option force_http_auth	0
+
```
openwrt-ai: nit: this blank line has a trailing tab. There is the same trailing whitespace on the blank line added in `uhttpd.init` (right after `pass_hash=$(awk ...)`); please strip both.

---

openwrt-ai (commented):
## Commit checks
- f930ebb "uhttpd: add force_http_auth option" — the `Link:` in the commit body (and the PR description) points to `https://github.org/openwrt/luci/pull/8852`; the domain should be `github.com`, so the reference is currently broken.

---

openwrt-ai (commented):
Reviewed 1 updated commit. The amendment addresses all prior review feedback — the `github.com` link fix in the commit message, the `force_http_auth` `local` declaration, removal of the redundant `append_arg "$cfg" config "-c"`, and the trailing-whitespace strips. No new issues found.

---

PR #24107, opened 2026-07-06
scripts: dhcpv4/v6: option to override preferred Client ID per interface

Shine-:
Since 387b49d89aa5c2c62c5437c51ee9ea4c58a13c8d, a DHCPv4 option tag 61 is sent by default  in DHCPv4 requests.

Since OpenWrt 25.12, a type 4 DUID(-UUID) is used by default in DHCPv6 requests, instead of the i/f's DUID-LL. This DUID-UUID is generated at first boot or sysupgrade (if it didn't exist), therefore it also affects upgraded systems, not just fresh installs. This auto-generated DHCPv6 global default DUID is also used in the DHCPv4 option tag 61 (Client ID), for RFC6355 compliance.

No possibility is available to fall back to using the i/f hardware ID (MAC address resp. DUID-LL). This breaks existing DHCP reservations and causes compatibility issues with certain ISPs.

Introduce a new, optional setting to override the behavior on a per-interface basis:

`network.<ifname>.sendclientid='auto|global|hardware|none'`

This setting will determine which type of client ID is preferrably being used in DHCPv4/DHCPv6 requests:

- `auto` or not setting this option (default) will keep the behavior as before this PR: Any explicitly defined client ID is used. If no explcit client ID is defined, the global default DUID is used as client ID. If this isn't defined either, DUID-LL resp. MAC address is used as per `odhcp6c`/`udhcpc` default.
- `global` means to prefer RFC6355 compliance even if an explicit client ID is specified. The global default DUID-UUID, if configured, will be used in DHCPv4 resp. DHCPv6 requests.
- `hardware` will not pass a client ID to `odhcp6c`/`udhcpc`, which results in `udhcpc`/`odhcp6c` using the i/f MAC address for DHCPv4 option tag 61, resp. the i/f's type 3 DUID for DHCPv6
- `none` (IPv4 only) will not add an option tag 61 at all to DHCPv4 requests (ie. the pre-25.12 behavior)

Setting `hardware` for `wan6` and `none` for `wan` will largely revert to the pre-25.12 behavior, solving the following issues:

Fixes: #20815 (possibly)
Fixes: #21350
Fixes: #21671
Fixes: #22277
Fixes: #22309 (possibly)
Fixes: #22332 (possibly)
Fixes: #24065
... and perhaps some more.

This PR is a similar, but more comprehensive approach to #24087.

PR #24350, opened 2026-07-21
[25.12]: scripts: dhcp: option to override preferred Client ID per interface

hauke:
Using UUID-based client IDs for DHCPv4/DHCPv6 with no option of falling back to hardware IDs (ie. MAC-address or DUID-LL) resp. none at all (IPv4), is causing regressions in some setups.

Introduce a new setting to override the preferred client ID to be used for DHCPv4/DHCPv6 on a per-interface basis:

network.<ifname>.sendclientid='auto|global|hardware|none'

- "auto" (default if empty or not present) uses any explicitly defined client ID, or falls back to the global DUID and finally to the DUID-LL resp. MAC address (ie. identical to before this commit).
- "global" uses the global default DUID, if configured, for DHCPv4 and DHCPv6 requests, even if a client ID is explicitly specified for the i/f
- "hardware" will not pass a client ID to udhcpc/odhcp6c, even if a global default DUID is configured or an explicit client ID specified, resulting in the i/f MAC address resp. type 3 DUID(-LL) to be used
- "none" (IPv4 only) will not add an option tag 61 to DHCPv4 requests at all.

Link: https://github.com/openwrt/openwrt/pull/24107

(cherry picked from commit d47ce71a89d6b88a9e1358ffccaeffc7f3f2498f)

### DIFF
--- a/package/network/config/netifd/Makefile
+++ b/package/network/config/netifd/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=netifd
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL=$(PROJECT_GIT)/project/netifd.git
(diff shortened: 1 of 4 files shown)


openwrt-ai (commented):

---

geldot:
Thanks for seeing this through. Just to make sure it covers all bases, does this get cleanly pushed down to the synthetic dhcpv6 configuration generated implicitly by the PPPoE proto's `option ipv6 'auto'` mechanism e.g.:

```
config interface 'WAN1'
    option proto 'pppoe'
    option ipv6 'auto'
    ...
```

PR #24352, opened 2026-07-21
mediatek: filogic: restructure SmartRG MT7988 device trees

cmonroe:
Replace mt7988a-smartrg-mt-stuart.dtsi with layered dtsi structure. The existing 3 boards assemble from shared building blocks instead of include-then-delete:

* mt798x-smartrg-common.dtsi: thermal trip/cooling-map macros
* mt798x-smartrg-factory.dtsi: eMMC factory/mfginfo nvmem layout
* mt7988-smartrg-base.dtsi: PMIC, regulators, eMMC, ramoops, console, watchdog, reset button
* mt7988-smartrg-sysled.dtsi: system LED controller and aliases
* mt7988-smartrg-mt-stuart-gsw.dtsi: built-in switch as lan2-4 with the per-port PHY LEDs
* mt7988-smartrg-wifi-pcie3.dtsi / -wifi-mt7995.dtsi: radio second PCIe function and pcie0 radio wiring for SDG-8733A
* mt7988a-smartrg-mt-stuart-phase1.dtsi plus per-model pair dtsi for SDG-8733 (AQR113C WAN) and SDG-8734 (SFP+ WAN)

Functional changes:

* AQR113C reset moves from per-PHY reset-gpios to the MDIO bus level reset actually wired on these boards; the unconnected phy0 reset line on SDG-8733 is dropped
* AQR PHY thermal zones drive the fan on all three boards through the PHY hwmon sensor; cpu-thermal gains critical-action "reboot" and 4C trip hysteresis
* usxgmii gmacs use managed "in-band-status"
* SDG-8733A: MT7995 radio reset (GPIO 9, active low) is now wifi-reset-gpios on pcie3, the port wired to the module reset, so the controller releases the module before link training with a 100 ms delay; the old description drove it as an active-high PERST on pcie0, toggling the module reset at the wrong time with the wrong polarity
* SDG-8733A: label-mac-device now points at the wan gmac; all three LAN LEDs described as GPIO LEDs per hardware (white GPIO 59, amber GPIO 60, green GPIO 68) with netdev triggers matching 01_leds (white 2.5G, green 1G, amber 100M); the internal 2.5G PHY LED node and its pinctrl state are dropped since the 2p5gbe_led0 pad (pin 68) works as the green LED GPIO and the PHY drives no LED
* phase1 USB: unused ssusb0 disabled, ssusb1 limited to USB2
* pci=pcie_bus_perf dropped from bootargs
* stale nodes removed: SLIC gpio-exports on the non-voice models, unstuffed WAN PHY on SDG-8734, spi0/spi_nand, unused pin groups and disabled-switch leftovers on SDG-8733A
* dtc warnings fixed: sysled children renamed led@1..led@4 (the driver names LEDs from the label property) and the unnecessary address/size cells dropped from the eMMC card@0 node

@dangowrt @blogic (or anyone else interested in reviewing): if I've gone too far splitting things out into include files just let me know. I really wanted to get away from using /delete-node/ as we had littered in the old 8733A.dts file but retain as many reusable components as possible. I will be able to reuse a lot of these for other MT7988 targets coming soon and some parts for MT7987 as well (2nd PR shortly).


cmonroe:
Once this PR is accepted I will open another to add support for SDG-8712 and 8732:
https://github.com/cmonroe/openwrt/commit/c840880327cf8a6f8ebf237c67e1c71513419d49

In case you want to see how some of the dtsi files are being reused. SDG-8713 (after that) reuses even more of these fragments.

target/linux/mediatek/dts/mt798x-smartrg-common.dtsi
```diff
@@ -0,0 +1,90 @@
+#define SDG_THERMAL_HEADER									\
```
openwrt-ai, line 7: This new file is missing the SPDX license identifier and copyright header that every other new dtsi in this series carries (e.g. `mt798x-smartrg-factory.dtsi`, `mt7988-smartrg-base.dtsi`). Prepend the standard block before the first `#define`:

```
// SPDX-License-Identifier: (GPL-2.0 OR MIT)
/*
 * Copyright (C) 2026 SmartRG Inc.
 * Author: Chad Monroe <chad.monroe@smartrg.com>
 *
 * Shared thermal trip/cooling-map helper macros for SmartRG designs.
 */
```

---

target/linux/mediatek/dts/mt798x-smartrg-common.dtsi
```diff
@@ -0,0 +1,90 @@
...
+		};										\
+		map1 {										\
+			trip = <&_cname##_active_low>;						\
+			cooling-device = <_fan0 1 1>, <_fan1 1 1>;				\
+		};										\
+		map0 {										\
+			trip = <&_cname##_active_silent>;					\
+			cooling-device = <_fan0 0 0>, <_fan1 0 0>;				\
+		};										\
+}
```
openwrt-ai: nit: this closing brace of `SDG_COOLING_MAP_2FAN` sits at column 0, while the equivalent brace of `SDG_COOLING_MAP_1FAN` above (line 61) is tab-indented. Align for consistency.

```suggestion
	}
```

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the diff and the restructure looks sound: the dropped explicit `status = "disabled"` on pcie2/pcie3/ssusb0/ssusb1/tphy/xsphy all match the kernel dtsi defaults (verified against v6.18 mt7988a.dtsi + the `186-arm64-dts-mt7988a-complete-dtsi.patch`), the `wifi-reset-gpios` polarity matches the other in-tree MT7988 boards, and the hex/decimal nvmem-cell reformat is value-preserving. CI is green. Two minor points inline.

---

openwrt-ai (commented):
Reviewed 1 new commit; no new issues found. The update since my last review adds the SPDX/copyright header and fixes the `SDG_COOLING_MAP_1FAN` closing-brace indentation, resolving both of my earlier inline comments.

---

jonasjelonek:
If this commits mixes DTS restructuring with functional changes - as the message suggest - I'd prefer to keep that separate. Pure restructuring in its own commit, then one or more with the functional changes. Or whatever ordering fits best.

cmonroe:
All of the "functional" changes are really just cleanup (e.g. removing stale pinctrl assignments). Assigning these back after the re-work would be intentionally adding a bug imo. The 2 exceptions are:
1. assigning the label MAC to wan
2. limiting USB speed to USB2

I could drop these 2 minor things and move them to a separate PR if it makes more sense.. I was just worried about too much PR spam.

jonasjelonek:
Those two exception IMO deserve their own commits, with a good explanation. Maybe also the other cleanups vs restructuring but that depends on the exact changes.

No need for separate PRs, should be fine here. Just that we have logically separated commits since that's what lands in history for everyone to understand.

cmonroe:
Aha! I don't know why I was hung up on PRs vs just commits after reading your first message :)

I completely agree with you and re-worked my one large commit into 10 or so targeted chunks + 1 larger restructure commit with no functional change. How does it look now?

target/linux/mediatek/dts/mt7988a-smartrg-mt-stuart-phase1.dtsi
```diff
@@ -0,0 +1,276 @@
...
+	pinctrl-names = "default";
+	pinctrl-0 = <&uart2_pins>;
+	status = "okay";
+
+	/* DA14531MOD Bluetooth */
+};
+
+&xsphy {
+	status = "okay";
+};
```
openwrt-ai: `xsphy` has no consumer left after fc10d7e. It only serves `ssusb0`: `usb@11190000` takes [`<&xphyu2port0>`/`<&xphyu3port0>` (mt7988a.dtsi:433-434)](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/mediatek/mt7988a.dtsi#L433-L434), which are the children of `xsphy`, while `ssusb1` takes [`<&tphyu2port0>`/`<&tphyu3port0>` (mt7988a.dtsi:450-451)](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/mediatek/mt7988a.dtsi#L450-L451) from `tphy`, which is still enabled just above. Since the commit stops enabling `ssusb0`, this node should go with it (same commit, so the tree never carries a dangling enable).

```suggestion
```


---

target/linux/mediatek/dts/mt7988d-smartrg-SDG-8733A.dts
```diff
@@ -33,133 +45,195 @@
...
 	phy = <&phy8>;
+
+	status = "okay";
 };
 
 &int_2p5g_phy {
 	pinctrl-names = "i2p5gbe-led";
-	pinctrl-0 = <&i2p5gbe_led0_pins>;
+	pinctrl-0 = <&gbe1_led0_pins>;
 };
```
openwrt-ai: 90f8d5a drops the `&i2p5gbe_led0 { status = "okay"; }` node but keeps this pinctrl state, so nothing consumes the mux any more: the PHY driver selects the state by name ([mtk-2p5ge.c:364](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/mediatek/mtk-2p5ge.c#L364)) but no PHY LED gets registered, so the pad ends up muxed to `led` with nothing driving it.

The group choice also looks off: per [pinctrl-mt7988.c:1098-1108](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/pinctrl/mediatek/pinctrl-mt7988.c#L1098-L1108) `gbe1_led0` is pin 65 (the LED pad of built-in switch PHY 1, and the switch is disabled on this board), while the internal 2.5G PHY's pad is `2p5gbe_led0` = pin 68. The other in-tree MT7988D board pairs them that way — [mt7988d-keenetic-kn-1812.dtsi:376-386](https://github.com/openwrt/openwrt/blob/e946c64df31e4ed7bff278a1ece417072f2d2ff3/target/linux/mediatek/dts/mt7988d-keenetic-kn-1812.dtsi#L376-L386) uses `i2p5gbe_led0_pins` for `&int_2p5g_phy` and `gbe1_led0_pins` for `&gsw_phy1`.

Since pin 68 is now the green GPIO LED, is muxing pin 65 here intentional (i.e. the D-package really routes the i2p5gbe LED to that pad), or should this whole `&int_2p5g_phy` block be dropped together with the LED node and the `gbe1_led0_pins` group?

---

target/linux/mediatek/dts/mt7988a-smartrg-mt-stuart-phase1.dtsi
```diff
@@ -0,0 +1,276 @@
...
+&pwm {
+	pinctrl-names = "default";
+	pinctrl-0 = <&pwm_pins>;
+	status = "okay";
+};
+
+&ssusb1 {
+	status = "okay";
+	mediatek,u3p-dis-msk = <0x1>;
+};
```
openwrt-ai, line 252: nit: every other node in this file (`&pwm`, `&serial1`, `&serial2`, `&gmac1`, `&gmac2`, `&pcie0`) puts `status` last; this one has the new property after it.

```suggestion
&ssusb1 {
	mediatek,u3p-dis-msk = <0x1>;
	status = "okay";
};
```


---

target/linux/mediatek/dts/mt798x-smartrg-common.dtsi
```diff
@@ -0,0 +1,96 @@
+// SPDX-License-Identifier: (GPL-2.0 OR MIT)
+/*
+ * Copyright (C) 2025 SmartRG Inc.
+ * Author: Piotr Kubik <piotr.kubik@adtran.com>
+ */
+
+#define SDG_THERMAL_HEADER									\
+	polling-delay-passive = <0>;								\
+	polling-delay = <1000>;									\
+	critical-action = "reboot";								\
```
openwrt-ai, line 10: nit: 937c863 changes these two lines plus the `hysteresis = <2000>` → `<4000>` in `SDG_THERMAL_TRIPS`, which alters the policy of the pre-existing `cpu-thermal` zone on all three boards — independent of "drive fan from AQR113C temperature". The body does mention it, but given the ask earlier in this thread for logically separated commits, the header/hysteresis policy change would read better as its own commit ahead of the AQR zones.

---

openwrt-ai (commented):
Reviewed the 11-commit split. The resulting tree is byte-identical to the head I last reviewed (`git diff 5f87b05..e946c64` is empty), so this pass concentrated on whether each new commit's message matches its own diff — they all do. Spot-checked the load-bearing claims: the dropped `pcie1` pinctrl is indeed identical to the SoC default (`pinctrl-0 = <&pcie1_pins>` on `pcie1` in mt7988a.dtsi), `mediatek,u3p-dis-msk`, `critical-action`, and the MDIO bus-level `reset-delay-us`/`reset-post-delay-us` all exist in both 6.18 and 6.12, the USXGMII PCS reports `LINK_INBAND_ENABLE` for `PHY_INTERFACE_MODE_USXGMII` so `managed = "in-band-status"` is the right description, the AQR hwmon registers with `HWMON_C_REGISTER_TZ` so `thermal-sensors = <&phy8>` resolves, and the values passed to `SDG_1FAN_THERMAL_MAP` match the 108/104/100/90/80/40 C in the message.

Four inline points; the only non-nit is a leftover the USB commit misses.

---

target/linux/mediatek/dts/mt7988d-smartrg-SDG-8733A.dts
```diff
@@ -33,133 +45,183 @@
...
 			color = <LED_COLOR_ID_AMBER>;
 			function = "lan";
 			gpios = <&pio 60 GPIO_ACTIVE_HIGH>;
 		};
+
+		lan_green {
+			color = <LED_COLOR_ID_GREEN>;
+			function = "lan";
+			gpios = <&pio 68 GPIO_ACTIVE_HIGH>;
+		};
```
openwrt-ai, line 93: nit: the PR description is now stale on two points and it is what ends up quoted in the merge summary:

- "internal 2.5G PHY LED pad corrected to the `gbe1_led0` group" — 4d3d450 no longer re-points `&int_2p5g_phy` at any pin group; the `&int_2p5g_phy` pinctrl state, the `&i2p5gbe_led0` node and the `i2p5gbe_led0_pins`/`gbe1_led0_pins` groups are all dropped, and pin 68 is this plain GPIO LED. The commit message itself is accurate ("the PHY drives no LED"); only the PR body still describes the earlier approach.
- The "Functional changes" list has no entry for 22aba00 (`wifi-reset-gpios` on `pcie3` replacing the active-high `reset-gpios` on `pcie0`), which is the largest behaviour change on this board.

---

openwrt-ai (commented):
Reviewed the 4 rewritten commits since e946c64 (4d3d450, 1d4c202, 3abb4f9, 22aba00). The tree delta versus my last review is exactly three hunks, and all three resolve my earlier inline points: `&xsphy` is now dropped together with `&ssusb0` in 3abb4f9, `mediatek,u3p-dis-msk` moved ahead of `status` on `&ssusb1`, and the `&int_2p5g_phy` pinctrl state plus the `gbe1_led0_pins` group are gone entirely rather than re-pointed at pin 65.

Each rewritten commit's message still matches its own diff, so no commit checks. Re-verified the load-bearing claims of the changed commits against 6.18:

- `wifi-reset-gpios` on `pcie3` with the radio node on `pcie0` matches the only other in-tree board using this property that way ([mt7988a-zbtlink-zbt-z8803be.dts:433-441](``https://github.com/openwrt/openwrt/blob/22aba0041690a4c49e400951de083bdad91f0de1/target/linux/mediatek/dts/mt7988a-zbtlink-zbt-z8803be.dts#L433-L441)),`` and `GPIO_ACTIVE_LOW` is the right polarity for the driver's assert/`msleep`/deassert sequence in [966-pcie-mediatek-gen3-Add-WIFI-HW-reset-flow.patch](``https://github.com/openwrt/openwrt/blob/22aba0041690a4c49e400951de083bdad91f0de1/target/linux/mediatek/patches-6.18/966-pcie-mediatek-gen3-Add-WIFI-HW-reset-flow.patch).`` GPIO 9 is not claimed by `pcie_2l_0_pereset` (pin 19), `pcie_1l_1_pereset` (pin 20), `pcie_clk_req_n0_0` (pin 8) or `pcie_clk_req_n3` (pin 10), so the descriptor request does not collide with either pinctrl state.
- `mediatek,u3p-dis-msk = <0x1>` is bit 0 = "disable USB3 port 0" per [mtu3.h:235](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/usb/mtu3/mtu3.h#L235); `&tphy` correctly stays enabled since it is what `ssusb1` consumes.
- After 4d3d450 the three `gpio-leds` children compose to `white:lan`/`amber:lan`/`green:lan`, matching the `ucidef_set_led_netdev` names in 01_leds, and the removed `mdio-bus:0f:green:lan` entry was the internal 2.5G PHY (addr 15) LED that no longer exists.
- `&int_2p5g_phy` has no `status` in mt7988a.dtsi, so dropping the board's `&int_2p5g_phy` override leaves `gmac1`'s `phy = <&int_2p5g_phy>` reference intact.

CI is green on 22aba00. One nit inline, on the PR description rather than the code.

---

target/linux/mediatek/dts/mt7988d-smartrg-SDG-8733A.dts
cmonroe: PR description updated

jonasjelonek:
Much better, thank you. Rebased on top of main and merged!

PR #24354, opened 2026-07-21
mediatek: leds-smartrg: sync with MCU firmware

cmonroe:
Update SmartRG system LED driver to sync with MCU firmware + add pacing and write verification.

cc @dangowrt

### DIFF
--- a/target/linux/mediatek/files/drivers/leds/leds-smartrg-system.c
+++ b/target/linux/mediatek/files/drivers/leds/leds-smartrg-system.c
@@ -196,7 +196,7 @@ static void srg_led_disable(struct i2c_client *client)
 	struct srg_led_ctrl *sysled_ctrl = i2c_get_clientdata(client);
 	int i;
 
-	for (i = 1; i < 10; i++)
+	for (i = 1; i < 9; i++)
 		srg_led_i2c_write(sysled_ctrl, i, 0);
 }
 
--- a/target/linux/mediatek/files/drivers/leds/leds-smartrg-system.c
+++ b/target/linux/mediatek/files/drivers/leds/leds-smartrg-system.c
@@ -64,6 +64,8 @@ srg_led_set_pulse(struct led_classdev *led_cdev,
 	struct srg_led *sysled = container_of(led_cdev, struct srg_led, led);
 	struct srg_led_ctrl *sysled_ctrl = sysled->ctrl;
 	bool blinking = false, pulsing = false;
+	u8 bright = led_cdev->brightness ? : led_cdev->blink_brightness ? :
+		    led_cdev->max_brightness;
 	u8 cbyte;
 	int ret;
 
@@ -80,7 +82,7 @@ srg_led_set_pulse(struct led_classdev *led_cdev,
 	cbyte = pulsing ? 3 : blinking ? 2 : 0;
 	mutex_lock(&sysled_ctrl->lock);
 	ret = srg_led_i2c_write(sysled_ctrl, sysled->index + 4,
-				(blinking || pulsing) ? 255 : 0);
+				(blinking || pulsing) ? bright : 0);
 	if (!ret) {
 		sysled_ctrl->control[sysled->index] = cbyte;
 		ret = srg_led_control_sync(sysled_ctrl);
--- a/target/linux/mediatek/files/drivers/leds/leds-smartrg-system.c
+++ b/target/linux/mediatek/files/drivers/leds/leds-smartrg-system.c
@@ -8,10 +8,9 @@
 #include <linux/mutex.h>
 #include <linux/of.h>
 
-/**
+/*
  * Driver for SmartRG RGBW LED microcontroller.
  * RGBW LED is connected to a Holtek HT45F0062 that is on the I2C bus.
- *
  */
 
 struct srg_led_ctrl;
@@ -36,7 +35,7 @@ srg_led_i2c_write(struct srg_led_ctrl *sysled_ctrl, u8 reg, u8 value)
 
 /*
  * MC LED Command: 0 = OFF, 1 = ON, 2 = Flash, 3 = Pulse, 4 = Blink
- * */
+ */
 static int
 srg_led_control_sync(struct srg_led_ctrl *sysled_ctrl)
 {
@@ -54,8 +53,7 @@ srg_led_control_sync(struct srg_led_ctrl *sysled_ctrl)
  * This function overrides the led driver timer trigger to offload
  * flashing to the micro-controller.  The negative effect of this
  * is the inability to configure the delay_on and delay_off periods.
- *
- * */
+ */
 static int
 srg_led_set_pulse(struct led_classdev *led_cdev,
 		  unsigned long *delay_on,
@@ -164,7 +162,7 @@ srg_led_init_led(struct srg_led_ctrl *sysled_ctrl, struct fwnode_handle *fw)
 						led_cdev, &init_data);
 	if (ret) {
 		dev_err(&sysled_ctrl->client->dev,
-			"srg_led_init_led: led register %s error ret %d!n",
+			"srg_led_init_led: led register %s error ret %d!\n",
 			led_cdev->name, ret);
 		return ret;
 	}
@@ -173,7 +171,6 @@ srg_led_init_led(struct srg_led_ctrl *sysled_ctrl, struct fwnode_handle *fw)
 }
 
 static int
-
 srg_led_probe(struct i2c_client *client)
 {
 	struct device *dev = &client->dev;
@@ -199,7 +196,7 @@ srg_led_probe(struct i2c_client *client)
 		msleep(5);
 	}
 
-	return srg_led_control_sync(sysled_ctrl);;
+	return srg_led_control_sync(sysled_ctrl);
 }
 
 static void srg_led_disable(struct i2c_client *client)
--- a/target/linux/mediatek/files/drivers/leds/leds-smartrg-system.c
+++ b/target/linux/mediatek/files/drivers/leds/leds-smartrg-system.c
@@ -69,7 +69,12 @@ srg_led_set_pulse(struct led_classdev *led_cdev,
 	u8 cbyte;
 	int ret;
 
-	if (delay_on && delay_off && (*delay_on > 100) && (*delay_on <= 500)) {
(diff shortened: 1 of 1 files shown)


openwrt-ai (commented):

---

jonasjelonek:
Can you split up the changes of the first commit? Since you already list 5 independent changes in the message, it should be easy to do so.

cmonroe:
Great idea, done! Thanks for taking the time to review :)

target/linux/mediatek/files/drivers/leds/leds-smartrg-system.c
```diff
@@ -64,10 +91,16 @@ srg_led_set_pulse(struct led_classdev *led_cdev,
 	struct srg_led *sysled = container_of(led_cdev, struct srg_led, led);
 	struct srg_led_ctrl *sysled_ctrl = sysled->ctrl;
 	bool blinking = false, pulsing = false;
+	u8 bright = led_cdev->brightness ? : led_cdev->max_brightness;
```
openwrt-ai: The commit message says this mirrors `led_set_software_blink()`, but that helper has three steps, not two: it takes the current brightness, otherwise the remembered `blink_brightness`, and only then `max_brightness` — see [`led_set_software_blink()` at led-core.c:194-198](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/leds/led-core.c#L194-L198).

The middle step is skipped here, so after a software-blink episode has left `brightness == 0` while `blink_brightness` still holds the level the user picked, offloading the pattern to the MCU jumps to full brightness instead of restoring that level. `blink_brightness` stays 0 as long as offload keeps succeeding, so `led_cdev->brightness ? : led_cdev->blink_brightness ? : led_cdev->max_brightness` would behave identically in the common path. Is skipping `blink_brightness` deliberate?

---

target/linux/mediatek/files/drivers/leds/leds-smartrg-system.c
```diff
@@ -188,15 +224,15 @@ srg_led_probe(struct i2c_client *client)
 		msleep(5);
```
openwrt-ai: nit: this `msleep(5)` looks redundant now. `srg_led_init_led()` ends with `srg_led_i2c_write(sysled_ctrl, index + 4, 0)`, and after the pacing commit that call already sleeps at least 5 ms and reads the register back before returning, so consecutive channels can no longer issue back-to-back writes. Dropping it would also make `srg_led_i2c_write()` the single place that owns the MCU timing. Was it kept intentionally as an extra per-channel settle?

---

target/linux/mediatek/files/drivers/leds/leds-smartrg-system.c
```diff
@@ -8,10 +8,9 @@
 #include <linux/mutex.h>
 #include <linux/of.h>
 
-/**
+/*
  * Driver for SmartRG RGBW LED microcontroller.
  * RGBW LED is connected to a Holtek HT45F0062 that is on the I2C bus.
- *
  */
```
openwrt-ai, line 14: nit: 64a5893 cleans up this block but leaves the same defect in the file's two other block comments, which still terminate with ` * */` — the "MC LED Command" comment above `srg_led_control_sync()` (lines 64-66) and the one above `srg_led_set_pulse()` (lines 80-85, which also keeps a stray trailing ` *` line). Since the commit is explicitly "fix trivial style issues, no functional change", folding those two in would leave the file consistent rather than half-converted.

---

openwrt-ai (commented):
Re-reviewed after the series was split into 7 commits. The resulting file content is byte-identical to the previously reviewed head (6c5c911), so this pass focused on the new commit boundaries; each of the 7 commit messages accurately describes its own diff, and the split is cleanly bisectable. Verified the LED-core claims against v6.18.44 (matching `KERNEL_PATCHVER` for mediatek): the `blink_set` 0/0-default contract, the non-zero-return software-blink fallback in `led_blink_setup()`, and the `min(value, max_brightness)` clamp in `led_set_brightness_nosleep()` all hold. Also confirmed that none of the three in-tree `srg,sysled` DTS nodes set `max-brightness`, so the new 254 cap is not overridden by `led_classdev_register_ext()`, and that `brightness_set_blocking` is set — which is what makes the new `usleep_range()` inside `blink_set` legal per the `leds.h` contract. All CI checks are green on c81fab2.

Three inline nits/questions only; nothing blocking.

---

target/linux/mediatek/files/drivers/leds/leds-smartrg-system.c
```diff
@@ -28,15 +27,43 @@ struct srg_led_ctrl {
...
+	 * while 5ms spacing measured fully reliable. Wait out the
+	 * consumption time, then read the register back and rewrite
+	 * until the value actually latched.
+	 */
+	for (attempt = 0; attempt < SRG_LED_WRITE_ATTEMPTS; attempt++) {
+		ret = i2c_smbus_write_byte_data(client, reg, value);
+		if (ret)
+			return ret;
+
+		usleep_range(5000, 7000);
```
openwrt-ai, line 51: Every `srg_led_set_brightness()` now costs five paced writes, not one: the channel brightness register, plus the four control registers that [`srg_led_control_sync()` rewrites unconditionally at leds-smartrg-system.c:72-76](``https://github.com/openwrt/openwrt/blob/ba97e0f27344a35ffb13ac7809bc5eabdc04fd67/target/linux/mediatek/files/drivers/leds/leds-smartrg-system.c#L72-L76).`` With the 5 ms floor here that is >=25 ms of sleeping while `sysled_ctrl->lock` is held per brightness change, and ~150 ms in the worst case where each write burns all five attempts — against roughly one bus transaction before this commit. All four channels serialize on that same mutex, so a trigger driving more than one channel queues behind it.

`sysled_ctrl->control[]` holds the desired state rather than a record of what the MCU last latched, so `srg_led_control_sync()` has no way to tell which of the four control registers actually changed. Would it be worth shadowing the last-latched control bytes so only the register that changed pays the 5 ms, or is the extra 20 ms per update cheap enough in practice on these boards that it doesn't matter?

---

target/linux/mediatek/files/drivers/leds/leds-smartrg-system.c
```diff
@@ -28,15 +27,43 @@ struct srg_led_ctrl {
...
 static int
 srg_led_i2c_write(struct srg_led_ctrl *sysled_ctrl, u8 reg, u8 value)
 {
-	return i2c_smbus_write_byte_data(sysled_ctrl->client, reg, value);
+	struct i2c_client *client = sysled_ctrl->client;
+	int attempt, ret;
+
+	/*
+	 * The MCU ACKs but silently discards register writes arriving
+	 * while it is still processing a previous one; on mt7622 boards
```
openwrt-ai, line 40: nit: the commit message says "Boards based on mt7987/mt7988 tolerate the unpaced bursts with identical MCU firmware (2.4)", but there is no in-tree mt7987 board with `srg,sysled`. The three consumers of this driver are [`mt7622-smartrg-SDG-841-t6.dts:241`](https://github.com/openwrt/openwrt/blob/ba97e0f27344a35ffb13ac7809bc5eabdc04fd67/target/linux/mediatek/dts/mt7622-smartrg-SDG-841-t6.dts#L241), [`mt7986a-smartrg-bonanza-peak.dtsi:562`](https://github.com/openwrt/openwrt/blob/ba97e0f27344a35ffb13ac7809bc5eabdc04fd67/target/linux/mediatek/dts/mt7986a-smartrg-bonanza-peak.dtsi#L562) and [`mt7988-smartrg-sysled.dtsi:20`](https://github.com/openwrt/openwrt/blob/ba97e0f27344a35ffb13ac7809bc5eabdc04fd67/target/linux/mediatek/dts/mt7988-smartrg-sysled.dtsi#L20). Did you mean mt7986 (Bonanza Peak / SDG-86xx)? If the mt7986a boards weren't among the ones you verified, that's worth spelling out — they are the closest relatives of the affected mt7622 SDG-841, and the message currently reads as if every non-mt7622 board was checked.

---

openwrt-ai (commented):
Re-reviewed the delta since the last pass (`git diff c81fab2..ba97e0f`) — the series was rewritten, so five of the seven commits carry new SHAs.

All three points from the previous review were addressed: `srg_led_set_pulse()` now uses the full `brightness ? : blink_brightness ? : max_brightness` chain, matching [`led_set_software_blink()` at led-core.c:188-197](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/leds/led-core.c#L188-L197) in v6.18.44 (the `KERNEL_PATCHVER` for mediatek); the redundant `msleep(5)` in `srg_led_probe()` is gone, with the pacing commit's message now documenting that `srg_led_i2c_write()` is the single owner of MCU timing; and the two remaining `* */` block comments plus the `!n` → `!\n` typo are folded into the style commit.

Each of the five rewritten commit messages still matches its own diff, including the reworded bodies, so no commit checks. Re-confirmed that sleeping inside `blink_set` stays legal — `brightness_set_blocking` is set, which is exactly the condition `leds.h` names, and `led_blink_set_nosleep()` defers to the workqueue in that case. The new `return cbyte ? 0 : 1;` is also fine against [`led_blink_setup()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/leds/led-core.c#L217-L231), which only tests the return for zero. No CI checks have reported on `ba97e0f` yet.

Two inline comments, both questions; nothing blocking.

---

target/linux/mediatek/files/drivers/leds/leds-smartrg-system.c
cmonroe: While I agree with this, the LED state change activity is very low: a few changes at boot which makes this cheap enough to not matter. It is how I've tested the current code and it works well. To minimize further churn I'd like to leave it as-tested and will consider a follow up PR in the future once I can test it widely.

target/linux/mediatek/files/drivers/leds/leds-smartrg-system.c
cmonroe: mt7987 boards where this was tested are pending PR #24679 so I'll leave the commit message as-is. mt7986 was also verified.

jonasjelonek:

PR #24356, opened 2026-07-22
dosfstools: fix compilation with new gettext

neheb:
Taken from upstream PR.

Also remove iconv patch. It's adding liconv when it shouldn't. Actually by the looks of things, it looks to be a proper solution to the original patch.

ping @jonasjelonek


neheb:
go home bot. you're drunk.

neheb:
ping @mcprat

mcprat:
is there no configuration for just ignoring gettext? since it's not really a dependency for our case, I don't think anything built after tools should have any affect on rebuilding, similar to the concept of build reproducibility we always apply to the concept of final binary blobs 

or maybe an environment variable can be overwritten when make is called? 

I'll try to see if this error shows up for me

anyway, better to have a backports than a hack patch regardless

neheb:
iconv is used. I had an LLM generate a meson build for this. Much nicer. I think the gnulib updates exposed this problem.

openwrt-ai (commented):

---

mcprat:
I see upstream keeps wanting to give us a hard time, as we have already removed this before it was added again...
fe1942a87df253c0fef252efe7556907880207c9 -----> 9300a20dcce2217b121bd2020cc1a4ea41fa4475

in the suggested patch, use of `AC_CONFIG_MACRO_DIRS` is nonsensical,
there is no m4 directory in dosfstools

I just don't see why we need iconv here at all...

mcprat:
> the gnulib updates exposed this problem

in my PR, gettext is updated to 0.24.2 from 0.24.1,
author of the patch says the problem is in 0.25?

are you building with my PR on top?

mcprat:
alternative: remove the patch, remove autoreconf, disable iconv with

`HOST_CONFIGURE_ARGS += --without-iconv`

if we happen to need autoreconf for some reason, AM_ICONV being conditional is troublesome
I posted #24362 to handle that possible future case

neheb:
I think this is fine.

tools/dosfstools/patches/101-gettext-0.25.patch
```diff
@@ -0,0 +1,48 @@
...
++++ b/configure.ac
+@@ -15,8 +15,11 @@
+ # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ 
+ AC_INIT([dosfstools], [4.2])
++AC_CONFIG_MACRO_DIRS([m4])
+ AC_SUBST([RELEASE_DATE], [2021-01-31])
+ AM_INIT_AUTOMAKE([1.11 foreign subdir-objects parallel-tests])
++AM_GNU_GETTEXT_VERSION([0.20])
++AM_GNU_GETTEXT([external])
```
openwrt-ai, line 45: This makes dosfstools' `configure.ac` depend on gettext m4 macros again, but the corresponding build-order dependency is not declared.

The patch removed in this commit (`101-config-switch-to-AC_CHECK_LIB.patch`) existed precisely so that `autoreconf` needed no gettext macros at all. With it gone, `HOST_FIXUP:=autoreconf` has to resolve `AM_GNU_GETTEXT_VERSION`, `AM_GNU_GETTEXT` and `AM_ICONV` (the latter still referenced at `configure.ac:76`, now unpatched) out of `$(STAGING_DIR_HOST)/share/aclocal`, which in this tree is populated by [`tools/gnulib/Makefile:18-19`](https://github.com/openwrt/openwrt/blob/3d18c1c588df358bfe6facd9bad3061e34c7517b/tools/gnulib/Makefile#L18-L19). There is no autopoint fallback either — OpenWrt stubs [`AUTOPOINT=true`](https://github.com/openwrt/openwrt/blob/3d18c1c588df358bfe6facd9bad3061e34c7517b/include/autotools.mk#L26) in `AM_TOOL_PATHS`, so the `AM_GNU_GETTEXT_VERSION([0.20])` line does not cause the macros to be installed; they have to be there already.

But [`tools/Makefile:101`](https://github.com/openwrt/openwrt/blob/3d18c1c588df358bfe6facd9bad3061e34c7517b/tools/Makefile#L101) declares only:

    $(curdir)/dosfstools/compile := $(curdir)/automake/compile

whereas every other tool that relies on gnulib's m4 files spells it out — [`coreutils` (L100)](https://github.com/openwrt/openwrt/blob/3d18c1c588df358bfe6facd9bad3061e34c7517b/tools/Makefile#L100), [`libtool` (L115)](https://github.com/openwrt/openwrt/blob/3d18c1c588df358bfe6facd9bad3061e34c7517b/tools/Makefile#L115), `elfutils` (L103). Without `$(curdir)/gnulib/compile` added here, a parallel tools build can autoreconf dosfstools before gnulib has installed its macros and fail with the same `possibly undefined macro: AM_ICONV` this patch is meant to fix. Is the dependency being relied on transitively somewhere, or should it be added to `tools/Makefile`?

---

tools/dosfstools/patches/101-gettext-0.25.patch
```diff
@@ -0,0 +1,48 @@
...
+ 		COPYING ChangeLog NEWS README
+-
+-EXTRA_DIST = autogen.sh
+--- a/configure.ac
++++ b/configure.ac
+@@ -15,8 +15,11 @@
+ # along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ 
+ AC_INIT([dosfstools], [4.2])
++AC_CONFIG_MACRO_DIRS([m4])
```
openwrt-ai, line 41: nit: this points at a directory that does not exist. I extracted the exact tarball this package builds (`dosfstools-4.2.tar.gz`, matching `PKG_HASH` in `tools/dosfstools/Makefile`) — there is no `m4/` in it, only `autogen.sh` and `config.rpath` at top level. And nothing creates it during the OpenWrt build, since [`AUTOPOINT=true`](https://github.com/openwrt/openwrt/blob/3d18c1c588df358bfe6facd9bad3061e34c7517b/include/autotools.mk#L26) makes `autoreconf` skip autopoint. Applying the patch and running aclocal gives:

```
aclocal: warning: couldn't open directory 'm4': No such file or directory
configure.ac:21: warning: macro 'AM_GNU_GETTEXT_VERSION' not found in library
configure.ac:22: warning: macro 'AM_GNU_GETTEXT' not found in library
configure.ac:78: warning: macro 'AM_ICONV' not found in library
```

It is only a warning, so it won't break the build, but the hunk is inert in this tree — same point @mcprat raised in the discussion above. Dropping this one line from the patch would keep the local delta smaller (the hunk header and diffstat need adjusting too, so no one-click suggestion).

---

openwrt-ai (commented):
Re-reviewed at `3d18c1c`. The only change since the previous review is the commit-message rewording; the patch content is byte-identical. The new message matches the diff, so no commit-check findings.

Two inline notes on the newly added patch, from a closer look at how `HOST_FIXUP:=autoreconf` resolves the gettext macros in this tree.

---

neheb:
@jonasjelonek can we get this merged?

jonasjelonek:
I lost track here ...

@mcprat Is this fine from your POV?

hauke:
I think you have to add this to `tools/Makefile`:
```
$(curdir)/dosfstools/compile := $(curdir)/automake/compile $(curdir)/gnulib/compile
```

hauke:
Please link the upstream PR: https://github.com/dosfstools/dosfstools/pull/218

What compile error do you get without this change? Could you please share the error.

neheb:
```
WARNING: 'automake-1.16' is missing on your system.
```

mcprat:
I don't have a specific problem with this solution, just nitpick thoughts such as that it's another patch when it seems like upstream development is dead so it will be neither accepted nor denied, meanwhile the alternative will build a little faster due to not needing autoreconf, and at least one line of the patch is not necessary

PR #24364, opened 2026-07-22
toolchain: drop $(REVISION) from GCC version

aparcar:
Storing the ever changing revision in the GCC binary may result in leakage in package builds. If two equal GCC versions build the same package version with different OpenWrt revisions, the package becomes unreproducible.

Solve this by dropping the revision.

Below is an example of the `cni` package:

    │ ├── readelf --wide --decompress --string-dump=.comment {}
    │ │ @@ -1,4 +1,4 @@
    │ │
    │ │  String dump of section '.comment':
    │ │ -  [     0]  GCC: (OpenWrt GCC 14.4.0 r35470-1b2aeb4f8a) 14.4.0
    │ │ +  [     0]  GCC: (OpenWrt GCC 14.4.0 r35485-0f256a0a7a) 14.4.0

### DIFF
--- a/toolchain/gcc/common.mk
+++ b/toolchain/gcc/common.mk
@@ -45,7 +45,7 @@ endif
 PATCH_DIR=../patches-$(GCC_MAJOR_VERSION).x
 
 BUGURL=http://bugs.openwrt.org/
-PKGVERSION=OpenWrt GCC $(PKG_VERSION) $(REVISION)
+PKGVERSION=OpenWrt GCC $(PKG_VERSION)
 
 HOST_BUILD_PARALLEL:=1
 


aparcar:
Full diff here https://rebuilderd.n.aparcar.org/api/v1/builds/94/artifacts/97/diffoscope

openwrt-ai (commented):

---

Rondom:
Removing the version has wider-ranging implications on ccache. If I add a patch to GCC, ccache would happily deliver me a previously compiled object using the GCC compiled without the patch, I think.

I think reproducible builds should not be an argument for this patch. If I use a different compiler (even just a potentially different one), then of course I might get a different result. In order to reproduce an OpenWrt build one would need at the minimum use a toolchain and tools compiled from the same git commit. Debian records the build env of every package using buildinfo files and their definition of reproducibility only covers using that environment. If I use a different version of GCC, I might get different results. Just for completeness sake: If the toolchain itself was not reprodicible, then that is of course an issue in itself and should be fixed!

If one wants to improve the performance of builds after distclean using ccache, I think it can be discussed to replace the commit hash with the latest commit to the toolchain directory for example, but it should really be considered carefully, because of the risk of breaking cache validation mentioned above.

(Thanks for tackling reproducibility after the apk-switch, it is very appreciated)

aparcar:
@Rondom just like with other packages, there should be a RELEASE attached that makes things actually trackable. Right now this just keeps changing without reflecting actual modifications to the toolchain. Shouldn't we implement something like that instead?

KanjiMonster:
> @Rondom just like with other packages, there should be a RELEASE attached that makes things actually trackable. Right now this just keeps changing without reflecting actual modifications to the toolchain. Shouldn't we implement something like that instead?

Keeping RELEASE in sync is additional work and could be a source of human errors, can we use something auto-generated, e.g. a combined hash of the patches? This should work without requiring a git checkout, and provides a unique, but reproducible "revision".

Rondom:
Whatever uniquely identifies the version is fine with me. A manually set version is of course prone to human-error, which is why I was suggesting the git-way. I am fine with any solution which prevents two different GCCs having the same version-string but different behaviour. This is in order to not break ccache and (maybe even more importantly!) make humans understand that the version is different both in compiled binaries and when running `gcc -v`.

The point I am trying to make was that, right now, you cannot expect to compile two toolchains from different git-commits and then expect them to produce the same output and changing the version to not change so often (which is what this PR does) is just papering over the issue that you are using a potentially different compiler version, which is not what reproducible builds are about.

We could also try to check in CI if the toolchain changes without any version bump, but this might be too complex, but it would also make it possible to store and consume pre-compiled toolchain versions and save on compilation time by downloading it, which would be very neat. (I am dreaming: This could unlock even more cache hits and even remote execution across different developer workstations)

I am not familiar with the rebuilder infra and just for me to understand the motivation behind this patch and more out of interest. Why can't you take the same version that a package was compiled with (which I guess is what prompted this patch)?

KanjiMonster:
I agree that two compilers built from different sources should have different version identifiers.

The issue AFAIU is that two compilers will have different version identifiers if the base revision is different, even if `toolchain/` is identical in both revisions (so the compilers are identical except for the version string), which then "leaks" into some packages (or the kernel IIRC).

The issue with the latest commit (for toolchain) is that the answer changes on a shallow clone vs a non-shallow checkout of the same commit. So it would make the revision stable for full clones, but not for shallow ones.

Rondom:
> The issue with the latest commit (for toolchain) is that the answer changes on a shallow clone vs a non-shallow checkout of the same commit. So it would make the revision stable for full clones, but not for shallow ones.

git tree object hash should be always available regardless of whether shallow clone or not, but is harder to reason about for humans, because it is not usually exposed in any porcelain commands, so it would require writing some script to check which git revisions have that tree to make sense for humans.

BKPepe:
This was pushed to the master branch already based on this commit: https://github.com/openwrt/openwrt/commit/9292f32fd96dc00b7c3f34a028d1fc41a4fcb6d2

KanjiMonster:
> This was pushed to the master branch already based on this commit: [9292f32](https://github.com/openwrt/openwrt/commit/9292f32fd96dc00b7c3f34a028d1fc41a4fcb6d2)

That's ... not good. It should be reverted for the above mentioned issues with that.

BKPepe:
4 days and it was not reverted, hey @aparcar or @jonasjelonek could you look at it?

aparcar:
I reverted this for now but will leave this PR open for the discussion of a better solution. Git history is no option, I wanted to go that route for automatic version tracking, but the extra network traffic was considered to high.

We could hash the content of the GCC patches. Since that's not a package we ship, no proper versioning is required. Do any patches outside the GCC folder alter the way GCC compiles things? On the other hand, is ccache caching only based on input file and GCC version? 

With an every changing GCC version, what does ccache help us right now in the build infrastructure? We don't build the same thing twice. I understand that developers re-build things locally multiple times, but from my understanding each pull would require a recompilation of GCC, since it contains the $(REVISION).

aparcar:
Maybe something like this, we could even add this to "rules.mk" as a generic functions if it's usable for more packages
```diff
diff --git a/toolchain/gcc/common.mk b/toolchain/gcc/common.mk
index 8a802dc531..9b162ec186 100644
--- a/toolchain/gcc/common.mk
+++ b/toolchain/gcc/common.mk
@@ -45,7 +45,8 @@ endif
 PATCH_DIR=../patches-$(GCC_MAJOR_VERSION).x

 BUGURL=http://bugs.openwrt.org/
-PKGVERSION=OpenWrt GCC $(PKG_VERSION) $(REVISION)
+PATCH_HASH:=$(shell cat $(sort $(wildcard $(CURDIR)/$(PATCH_DIR)/*.patch)) | $(MKHASH) sha256 | head -c 8)
+PKGVERSION=OpenWrt GCC $(PKG_VERSION) $(PATCH_HASH)
```

PR #24374, opened 2026-07-22
qualcommax: add AP8220 U-Boot mod variant

hiper25:
Some AP8220 devices in the secondary market use a modified QSDK U-Boot and MIBIB layout. The entire 128 MiB NAND is exposed as one rootfs partition instead of the stock rootfs1/rootfs2/usrdata layout.

The modified U-Boot also injects a legacy SMEM partition node into the NAND controller. Disable it for this variant so it cannot duplicate the fixed SPI NOR partition names and break calibration and MAC extraction.

Split the existing board description into a common DTSI and add a dedicated U-Boot mod DTS and image profile. Reuse the stock networking, calibration, BDF and U-Boot environment handling, while keeping sysupgrade confined to the single rootfs partition.


target/linux/qualcommax/dts/ipq8071-ap8220.dtsi
```diff
@@ -0,0 +1,318 @@
...
+};
+
+&cryptobam {
+	status = "okay";
+};
+
+&crypto {
+	status = "okay";
+};
+
```
openwrt-ai, line 213: These `&cryptobam` and `&crypto` `status = "okay"` overrides were removed from every qualcommax board DTS — including `ipq8071-ap8220.dts` itself — in [commit `0180e53c8b9` "qualcommax: dts: stop enabling the driverless QCE crypto nodes"](https://github.com/openwrt/openwrt/commit/0180e53c8b92a594b9f7e229749c4e8acdb8c747), which landed on main after this PR's base. Since CRYPTO_DEV_QCE was dropped, the nodes are driverless and leaving them `okay` reintroduces the `sync_state() pending due to 73a000.crypto` boot regression that commit fixed.

Because that cleanup deleted the overrides from the old `.dts` while this PR moves the same content into the new `.dtsi`, a plain rebase onto current main won't carry the deletion across — the new file re-enables the nodes for both AP8220 variants. Drop both blocks here (keep `&prng`, which upstream kept):

```suggestion
```

---

aiamadeus:
No, official OpenWrt only supports the stock U-Boot. The term "U-Boot mod" refers
to the U-Boot compiled by OpenWrt, rather than a version modified by a third party.

PR #24376, opened 2026-07-22
realtek: Support PSE equipment via devicetree

svanheule:
Switch to a devicetree description of the PSE equipment, so the realtek-poe daemon and its config file are no longer required.

The third commit dropping the legacy config doesn't have to be merged, but perhaps the file can serve as a table of still-to-be-converted devices.

### DIFF
--- a/target/linux/realtek/dts/rtl8380_netgear_gs110tpp-v1.dts
+++ b/target/linux/realtek/dts/rtl8380_netgear_gs110tpp-v1.dts
@@ -35,14 +35,16 @@
 			gpios = <&gpio1 34 GPIO_ACTIVE_LOW>;
 		};
 	};
-};
 
-&gpio1 {
-	poe-enable {
-		gpio-hog;
-		gpios = <10 GPIO_ACTIVE_HIGH>;
-		output-high;
-		line-name = "poe-enable";
+	device_power: regulator-device {
+		compatible = "regulator-fixed";
+		regulator-name = "ext-psu";
+		regulator-always-on;
+		regulator-boot-on;
+
+		regulator-max-microvolt = <54000000>;
+		regulator-min-microvolt = <54000000>;
+		regulator-power-budget-milliwatt = <130000>;
 	};
 };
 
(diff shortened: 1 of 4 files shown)


target/linux/realtek/dts/rtl8380_netgear_gs110tpp-v1.dts
```diff
@@ -52,6 +54,52 @@
...
+};
+&phy13 {
+	pses = <&pse_pi2>;
+};
+&phy14 {
+	pses = <&pse_pi1>;
+};
+&phy15 {
+	pses = <&pse_pi0>;
 };
```
jonasjelonek, line 103: Personal preference, since this just carries a single property:
```
&phy8 { pses = <&pse_pi7>; };
...
```

jonasjelonek (commented):
Only a nit, fine with everything else. Discussion about the supply situation is for upstream mailing list.

target/linux/realtek/dts/rtl8380_netgear_gs110tpp-v1.dts
```diff
@@ -35,14 +35,16 @@
...
 	};
-};
 
-&gpio1 {
-	poe-enable {
-		gpio-hog;
-		gpios = <10 GPIO_ACTIVE_HIGH>;
-		output-high;
-		line-name = "poe-enable";
+	device_power: regulator-device {
```
openwrt-ai, line 39: nit: the two other realtek devices that model an external PoE PSU as a fixed-regulator name it `poe_power: regulator-poe` with `regulator-name = "poe-power"` ([xs1930-12hp.dts:24](https://github.com/openwrt/openwrt/blob/2606f2f9ecb541000ba8a4435f163534f7fc6d4a/target/linux/realtek/dts/rtl9313_zyxel_xs1930-12hp.dts#L24), [xgs1930-28hp.dts:180](https://github.com/openwrt/openwrt/blob/2606f2f9ecb541000ba8a4435f163534f7fc6d4a/target/linux/realtek/dts/rtl9301_zyxel_xgs1930-28hp.dts#L180)); this uses `device_power: regulator-device` / `"ext-psu"`. Intentional (since the 54V brick is the device's sole external supply here), or worth aligning with the existing `poe_power`/`regulator-poe` naming?

---

openwrt-ai (commented):
Reviewed 3 new commits.

---

target/linux/realtek/dts/rtl8380_netgear_gs110tpp-v1.dts
jonasjelonek: No, this is the correct way to define and reference regulators here. The other examples mentioned will be adjusted later.

PR #24377, opened 2026-07-22
realtek: add support for ZyXEL GS1900-24E B1

joshuacov1:
Adds support for the Zyxel GS1900-24E **B1** hardware revision, as a sibling to the existing `zyxel_gs1900-24e-a1` target. Only A1 is currently supported upstream; B1 units currently run the A1 image unmodified via manual DTS override, since the compatible string `zyxel,gs1900-24e` (pre-rename) still matches.

Zyxel's documentation describes the A1->B1 change as a simple front-panel button relabel. So far I have found the following (for A1 see https://techinfodepot.shoutwiki.com/wiki/ZyXEL_GS1900-24E):
- PCB: 37ZY-GM2430+212 V1.2 (A1: 37ZY-G724DO+412 V.12)
- RAM: 128 MiB DDR3 (A1: 128 MiB DDR2)
- Flash: mx25l12805d, 16 MiB (A1: 16 MiB SPI-NOR)
- External PHYs: RTL8218D (A1: unknown)
- Rear power switch: absent on this B1 unit (inlet only). A1 is described as having one in the *title* of openwrt/openwrt#18620, but that issue does not explicitly identify the affected hardware as "A1" specifically, nor is it confirmed by a spec sheet or photo
- ZYXEL_VERS firmware family unchanged (AAHK)

The existing A1 MDIO bus addressing and switch-port SerDes layout work correctly on B1 unmodified — Linux's PHY subsystem identifies the external PHY chip at runtime via ID registers rather than from the devicetree, so no port-mapping changes were needed regardless of what A1 actually uses.

This DTS deliberately does not carry forward the `gpio0` mdio-reset gpio-hog added to `-a1.dts` in ba57225066 ("realtek: hog the GS1900-24E external IC reset line", addressing #18620). That fix targets a probe-order race specific to A1's shared reset line between mdio_bus0 and mdio_aux. I could not confirm any reset-line continuity between `gpio0` pin 1 and the external ICs on B1 hardware, so inheriting A1's fix without verification seemed like the wrong default.

My testing suggests it is safe to omit.

I had OpenWrt 24.10.5 (kernel 6.6) running on this unit before applying this patch, which I used to characterize the #18620 regression pattern beforehand. Tagged 802.1Q VLAN traffic, verified via `rx_packets`/`tx_packets` counters and `tcpdump`, showed no errors during normal running state, a software `reboot`, and a full AC power-cycle.

All three phases showed clean bidirectional counter growth, with no sign of the #18620 stuck-RX symptom.

With current master and kernel 6.18.39 (this patch, no gpio-hog), I could not reproduce the issue either.

No reset-line errors, no stuck PHYs, all 24 ports transition blocking->forwarding normally.

Given B1 shows no symptom of the #18620 regression across both kernels and both reboot paths, the gpio-hog is omitted for now. Happy to add it back with an explicit B1-scoped justification if maintainer review turns up a reason it's needed regardless. I have had this running for about three weeks without problems so far.

### DIFF
--- a/target/linux/realtek/dts/rtl8382_zyxel_gs1900-24e-b1.dts
+++ b/target/linux/realtek/dts/rtl8382_zyxel_gs1900-24e-b1.dts
@@ -1,0 +1,83 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include "rtl8380_zyxel_gs1900.dtsi"
+#include "rtl8380_zyxel_gs1900_gpio.dtsi"
+
+/ {
+	compatible = "zyxel,gs1900-24e-b1", "realtek,rtl838x-soc";
+	model = "Zyxel GS1900-24E B1";
+};
+
+/*
+ * Unlike -a1.dts, this DTS deliberately does NOT hog gpio0 pin 1 to
+ * permanently de-assert the shared external-IC reset line (see
+ * commit ba57225066, "realtek: hog the GS1900-24E external IC reset
+ * line", and openwrt/openwrt#18620).
+ */
+
+&mdio_bus0 {
+	ethernet-phy-package@0 {
+		#address-cells = <1>;
+		#size-cells = <0>;
+		reg = <0>;
+
+		PHY_C22(0, 0)
+		PHY_C22(1, 1)
+		PHY_C22(2, 2)
+		PHY_C22(3, 3)
+		PHY_C22(4, 4)
+		PHY_C22(5, 5)
+		PHY_C22(6, 6)
+		PHY_C22(7, 7)
+	};
+
+	ethernet-phy-package@16 {
+		#address-cells = <1>;
+		#size-cells = <0>;
+		reg = <16>;
+
+		PHY_C22(16, 16)
+		PHY_C22(17, 17)
+		PHY_C22(18, 18)
+		PHY_C22(19, 19)
+		PHY_C22(20, 20)
+		PHY_C22(21, 21)
+		PHY_C22(22, 22)
+		PHY_C22(23, 23)
+	};
+};
+
+&switch0 {
+	ethernet-ports {
+		SWITCH_PORT_SDS(1, 1, 0, 1, qsgmii)
+		SWITCH_PORT_SDS(0, 2, 0, 0, qsgmii)
+		SWITCH_PORT_SDS(3, 3, 0, 3, qsgmii)
+		SWITCH_PORT_SDS(2, 4, 0, 2, qsgmii)
+		SWITCH_PORT_SDS(5, 5, 1, 1, qsgmii)
+		SWITCH_PORT_SDS(4, 6, 1, 0, qsgmii)
+		SWITCH_PORT_SDS(7, 7, 1, 3, qsgmii)
+		SWITCH_PORT_SDS(6, 8, 1, 2, qsgmii)
+
+		SWITCH_PORT(9, 9, internal)
+		SWITCH_PORT(8, 10, internal)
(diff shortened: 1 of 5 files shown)


target/linux/realtek/dts/rtl8382_zyxel_gs1900-24e-b1.dts
```diff
@@ -0,0 +1,71 @@
...
+	PHY_C22(7, 7)
+
+	PHY_C22(16, 16)
+	PHY_C22(17, 17)
+	PHY_C22(18, 18)
+	PHY_C22(19, 19)
+	PHY_C22(20, 20)
+	PHY_C22(21, 21)
+	PHY_C22(22, 22)
+	PHY_C22(23, 23)
```
plappermaul: Please encapsulate into ethernet-phy-package. See recent patches.

plappermaul (changes requested):

target/linux/realtek/dts/rtl8382_zyxel_gs1900-24e-b1.dts
joshuacov1: fixed: the two external RTL8218D PHY packages are now encapsulated. Ports 8-15 use the
SoC-integrated PHY block and aren't part of an external package.

target/linux/realtek/image/rtl838x.mk
```diff
@@ -487,6 +487,15 @@ define Device/zyxel_gs1900-24e-a1
 endef
 TARGET_DEVICES += zyxel_gs1900-24e-a1
 
+define Device/zyxel_gs1900-24e-b1
```
openwrt-ai, line 491: nit: This adds a new `.dts` plus a new `define Device/` block, which normally triggers the new-device commit-message expectation of explicit **Flash instructions** and **MAC address layout** sections — both are absent here (the message does cover the hardware-spec delta vs. A1). Since B1 is a same-family revision whose flash path (`ZYXEL_VERS := AAHK`, identical to A1) and MAC sourcing (shared `realtek_setup_macs` case it's added to) are unchanged from A1, this may fall under the "hardware variant of an existing device" exemption. Flagging so you can decide whether the two sections are worth adding for a stranger flashing from the commit alone.

---

openwrt-ai (commented):
Fresh review of the initial commit. The `ethernet-phy-package` encapsulation and the sister-list additions (uboot-envtools, `02_network`, `05_compat-version`, image `.mk`) all look consistent with in-tree references, and the commit message matches the diff. One inline nit below.

---

target/linux/realtek/image/rtl838x.mk
joshuacov1: Anyone building this from source and reading a raw commit message already has enough knowledge for what the rest of the realtek target assumes throughout — this isn't someone flashing pre-built images off a download page. I don't think restating the same procedures adds real information. Example: 13cb708691 (GS1900-10HP B1, merged a few days ago) covers a comparable A1->B1 hardware-variant addition with neither a flash-instructions nor a MAC-layout section.

plappermaul:

hailfinger:
About the rear power switch of the Zyxel GS1900-24E (probably A1, not marked on the label): I have a bunch of those in my lab, and they have a tiny power switch next to the power inlet. The switch is open anyway, so if photos of the exterior or interior are helpful in any way, I can provide them next week.

joshuacov1:
> ... so if photos of the exterior or interior are helpful in any way, I can provide them next week.

I'll have this in mind if more information is needed in the future.

Issue #24359, opened 2026-07-22
RTL8264 does not transmit traffic


plappermaul:
There are several hints about RTL8264 not working any longer

- https://github.com/openwrt/openwrt/pull/24349
- https://github.com/openwrt/openwrt/pull/23993
- https://github.com/openwrt/openwrt/pull/24245#issuecomment-5008141733

Give this bug a new home.

https://github.com/openwrt/openwrt/pull/23427 may be blamed

@l0rdg3x @jonasjelonek @andrewjlamarche @carlosz1986

l0rdg3x:
Thanks @plappermaul.
I'm in a business traver in this days, i'll be home this evening.
Thanks for your help.

andrewjlamarche:
https://github.com/openwrt/openwrt/pull/23726#issuecomment-4760327325 I did test after https://github.com/openwrt/openwrt/pull/23427... guess I didn't see if they passed traffic. I'll check with that...

l0rdg3x:
Data from my **S1100W-8XGT-SE** (RTL9303, 2× RTL8264, USXGMII), current `main` @ `1ea29a6f49` (6.18.39, incl. #24245), RAM-boot:

- **Both directions broken.** Ingress: port MAC MIB counts (+4670 under ARP flood) but `eth0` rx = 0, RX IRQ never fires. Egress: `eth0` tx and port `tx_BroadcastPkts` count, but the host on the same wire never sees a single frame (no neighbor entry ever created).
- **Firmware-table hypothesis ruled out here:** live C45 read gives `VEND1 0x104 = 0xf802` → `[2:0]=2`, same as the S1300WP (#23726), so since #23947 the driver already loads the correct `rtl8261n.bin`. Probe is clean, no `-ETIME` — still broken.
- **What fixes ingress on this board (tested in July):** skipping the kernel re-init and keeping the bootloader state (no `config_init` reset+re-patch, no first AN restart, PCS adopts the already-configured USXGMII SerDes). Any of those three perturbations permanently kills the PHY↔switch link. Egress stays dead even then → likely a second, separate issue.
- Confirmed here too: traffic LEDs stay lit with the cable unplugged.

Happy to test patches — serial + TFTP RAM-boot ready, quick turnaround.

*Bench testing and analysis assisted by Claude Fable 5 (Anthropic AI).*

jonasjelonek:
[...]

PR #24378, opened 2026-07-22
realtek: keep bootloader rtl826x PHY/SerDes initialization (RTL8264 ingress fix)

l0rdg3x:
Possible fix for the RTL8264 ingress half of #24359, **tested on our hardware** (Hasivo S1100W-8XGT-SE, RTL9303 + 2× RTL8264, all 8 ports USXGMII).

### Mechanism

On these boards the SDK bootloader has already reset, patched and fully brought up the PHYs **and** the switch SerDes: the USXGMII link between them is up and working before Linux starts. The current drivers then re-initialize both sides:

1. `rtl826x_config_init` re-runs the PHY reset + firmware patch,
2. the first `config_aneg` rewrites the advertisement and restarts copper AN,
3. the first `pcs_config` runs the full SerDes deactivate/configure/activate cycle (`hw_mode` starts at OFF).

Each of these three alone permanently kills the PHY→switch direction on RTL8264 (verified individually on silicon); once dead, only the full bootloader init recovers it. The vendor driver skips reset/patch under `CONFIG_MACH_REALTEK_RTL` for the same reason. The wrong-firmware-table theory is ruled out on this board: our dies read `VEND1 0x104 = 0xf802` → the driver already loads `rtl8261n.bin` (see issue comment).

### The two commits

- **PHY**: under `CONFIG_MACH_REALTEK_RTL`, keep the bootloader initialization (skip reset+patch, keep IRQ quiesce + SerDes polarity) and leave the first autonegotiation configuration untouched.
- **PCS**: when the first configuration of a SerDes requests a USXGMII mode and the hardware mode select **and** USXGMII submode already match exactly, adopt the bootloader state instead of reconfiguring. Any mismatch, and all non-USXGMII modes, keep the existing full configuration path — RTL8224/QXGMII boards and SFP setups are unaffected.

### Validation (RAM-boot, current main)

| Signal | main pristine | main + this PR |
|---|:---:|:---:|
| `eth0` rx under ARP flood on lan2 | 0 | **+354** |
| eth0 RX IRQ | never fires | fires |
| 10G ports link speed | stuck at 5G | **10G** |
| SerDes reconfigurations in dmesg | 8 | 0 (adopted ×8) |

### Known limitations / open points

- A later explicit renegotiation (e.g. `ethtool -s`, or any AN restart) still kills ingress — the underlying "USXGMII never recovers" issue is not solved, only avoided. A follow-up that aligns the initial advertisement with the bootloader state (so no config cycle ever restarts AN) is in the works.
- **Egress-to-wire on this board is a separate regression and is NOT fixed by this PR** (port TX MIBs count, wire stays silent, also with this PR applied; bidirectional in U-Boot on the same wire). Under investigation, data in #24359.
- Would appreciate testing on other rtl826x boards (RTL8261N / RTL8264B, and a QXGMII/RTL8224 board to confirm the adoption guard).

@plappermaul @jonasjelonek @andrewjlamarche @carlosz1986

### DIFF
--- a/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
+++ b/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
@@ -3876,6 +3876,38 @@ static int rtpcs_pcs_config(struct phylink_pcs *pcs, unsigned int neg_mode,
 	}
 
 	scoped_guard(mutex, &ctrl->lock) {
+
+		/*
+		 * Adopt a SerDes the bootloader already configured for the
+		 * requested USXGMII mode instead of fully reconfiguring it.
+		 * The full deactivate/configure/activate cycle bounces the
+		 * link to the attached PHY, which firmware-driven PHYs like
+		 * the RTL8264 cannot recover from on their system side:
+		 * copper ingress towards the switch MAC dies permanently.
+		 * Restrict adoption to USXGMII with matching mode select
+		 * and submode, where this problem exists and the bootloader
+		 * state is known good.
+		 */
+		if (sds->hw_mode == RTPCS_SDS_MODE_OFF &&
+		    sds->hw_mode != hw_mode &&
+		    rtpcs_93xx_sds_usxgmii_submodes[hw_mode] >= 0 &&
+		    sds->swcore_regs.mac_mode) {
+			u32 cur_mode;
+			u32 cur_submode = RTPCS_93XX_SDS_USXGMII_SUBMODE_10GSX;
+
+			if (sds->swcore_regs.usxgmii_submode)
+				regmap_field_read(sds->swcore_regs.usxgmii_submode,
+						  &cur_submode);
+
+			if (!regmap_field_read(sds->swcore_regs.mac_mode, &cur_mode) &&
+			    ctrl->cfg->sds_hw_mode_vals[hw_mode] == (s16)cur_mode &&
+			    rtpcs_93xx_sds_usxgmii_submodes[hw_mode] == (s16)cur_submode) {
+				dev_info(ctrl->dev,
+					 "SerDes %u already configured for mode %s, keeping bootloader setup\n",
+					 sds->id, phy_modes(interface));
+				sds->hw_mode = hw_mode;
+			}
+		}
 		if (sds->hw_mode != hw_mode) {
 			dev_info(ctrl->dev, "configure SerDes %u for mode %s\n", sds->id,
 				 phy_modes(interface));
--- a/target/linux/realtek/patches-6.18/744-net-phy-realtek-rtl826x-keep-bootloader-init.patch
+++ b/target/linux/realtek/patches-6.18/744-net-phy-realtek-rtl826x-keep-bootloader-init.patch
@@ -1,0 +1,93 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Gennaro Cimmino <gcimmino@rayonra.net>
+Date: Tue, 22 Jul 2026 22:30:00 +0200
+Subject: [PATCH] net: phy: realtek: keep bootloader rtl826x initialization on
+ Realtek switches
+
+On RTL93xx switches the SDK bootloader has already reset, patched and
+fully brought up the RTL8264/826x PHYs, including the system-side
+USXGMII interface towards the switch SerDes. Re-running the reset and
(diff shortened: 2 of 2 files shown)


jonasjelonek:
Sorry but this is a no-go for me.

IMO a single PHY not working doesn't justify to tear down everything that has been build up with hard work over the past months. Besides several potential issue with this approach, completely failing if the bootloader doesn't do pre-setup, potential breakage of other PHYs.

andrewjlamarche:
Agree ^

l0rdg3x:
OK right. We can work on a more targeted fix

target/linux/realtek/patches-6.18/744-net-phy-realtek-rtl826x-keep-bootloader-init.patch
```diff
@@ -0,0 +1,93 @@
...
+autonegotiation configuration untouched. Validated on a Hasivo
+S1100W-8XGT-SE (RTL9303 + 2x RTL8264): restores port-to-CPU RX that
+has been broken since the switch to the mainline realtek PHY driver.
+
+Assisted-by: Claude:claude-fable-5
+Signed-off-by: Gennaro Cimmino <gcimmino@rayonra.net>
+---
+--- a/drivers/net/phy/realtek/realtek_main.c
++++ b/drivers/net/phy/realtek/realtek_main.c
+@@ -2859,6 +2859,7 @@
```
openwrt-ai, line 28: The `Check Kernel patches (realtek, rtl838x)` CI job fails on this patch:

```
target/linux/realtek/patches-6.18/744-net-phy-realtek-rtl826x-keep-bootloader-init.patch
Kernel patches for realtek/rtl838x require refresh.
(run 'make target/linux/refresh' and force push this pr)
```

The `git diff --exit-code` after refresh lists only this file, so the hand-written hunk headers here (`@@ -2859,6 +2859,7 @@`, `@@ -2876,6 +2877,23 @@`, `@@ -2953,10 +2971,23 @@`) don't match the actual line offsets/context once the preceding realtek patches are applied. Regenerate this patch with `make target/linux/refresh` (with the realtek target selected in `.config`) and force-push, so the hunk metadata is correct. The refreshed patch is also available as the `realtek-rtl838x-refreshed` artifact on the failed CI run.

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -3876,6 +3876,38 @@ static int rtpcs_pcs_config(struct phylink_pcs *pcs, unsigned int neg_mode,
...
+		 * The full deactivate/configure/activate cycle bounces the
+		 * link to the attached PHY, which firmware-driven PHYs like
+		 * the RTL8264 cannot recover from on their system side:
+		 * copper ingress towards the switch MAC dies permanently.
+		 * Restrict adoption to USXGMII with matching mode select
+		 * and submode, where this problem exists and the bootloader
+		 * state is known good.
+		 */
+		if (sds->hw_mode == RTPCS_SDS_MODE_OFF &&
+		    sds->hw_mode != hw_mode &&
```
openwrt-ai, line 3892: nit: `sds->hw_mode != hw_mode` is redundant here. The preceding clause already pins `sds->hw_mode == RTPCS_SDS_MODE_OFF`, so this just re-tests `hw_mode != RTPCS_SDS_MODE_OFF`, which is in turn implied by the following `rtpcs_93xx_sds_usxgmii_submodes[hw_mode] >= 0` guard (the OFF slot is not a USXGMII submode). It can be dropped without changing behavior.

---

openwrt-ai (commented):
Reviewed the 2 commits at head `e94d7c8`. Commit messages match their diffs. The main blocker is the failing `Check Kernel patches` CI job — the new patch `744-...` needs `make target/linux/refresh`; details inline, plus one minor nit on the PCS adoption guard. The design objections already raised in the thread are the maintainers' call and out of scope here.

---

l0rdg3x:
Closing this: the real root cause is now identified and fixed properly in #24393 — the S1100W-8XGT-SE devicetree was missing the SerDes lane polarity declarations, so the driver cleared the HSI_INV/HSO_INV bits the bootloader had set. The keep-bootloader approach in this PR only worked because it happened to preserve those bits; with the polarity declared in the dts, the driver's normal (re-)initialization works and traffic survives restarts. Thanks @jonasjelonek for pushing back on this approach — the targeted fix is 33 lines of board dts only.

PR #24379, opened 2026-07-22
qcom: dts: stop enabling the driverless QCE crypto nodes

JuliusBairaktaris:
ce120be361 ("kernel: disable qualcomm crypto") dropped the CRYPTO_DEV_QCE kernel configs, so crypto@73a000 and its BAM no longer have a driver. The qualcommax board device trees still override both nodes to "okay", which leaves enabled consumers that can never probe.

On ipq807x this is more than dead code: the GCC clock controller registers the USB GDSC power domains, so the genpd core installs a sync_state() callback on it (dev_set_drv_sync_state() in drivers/pmdomain/core.c), and fw_devlink then defers that callback forever. On the 6.18 testing kernel every ipq807x board logs

    qcom,gcc-ipq8074 1800000.clock-controller: sync_state() pending due to 73a000.crypto

on each boot, and unused power domains are never released.

Drop the overrides and fall back to the SoC dtsi defaults. On ipq8074, mainline ships both nodes disabled and no mainline board enables them; on ipq5018/ipq6018 they are enabled by default, but the GCC drivers there register no power domains, so no sync_state() dependency exists.

Observed on a Xiaomi AX3600 on the 6.18 testing kernel; with this change the built DTB returns crypto@73a000 to its disabled default.

cc @neheb

### DIFF
--- a/target/linux/ipq40xx/dts/qcom-ipq4028-wpj428.dts
+++ b/target/linux/ipq40xx/dts/qcom-ipq4028-wpj428.dts
@@ -106,10 +106,6 @@
 	status = "okay";
 };
 
-&crypto {
-	status = "okay";
-};
-
 &tlmm {
 	mdio_pins: mdio_pinmux {
 		mux_1 {
@@ -244,10 +240,6 @@
 &blsp1_uart1 {
 	pinctrl-0 = <&serial_pins>;
 	pinctrl-names = "default";
-	status = "okay";
-};
-
-&cryptobam {
 	status = "okay";
 };
 
--- a/target/linux/qualcommbe/dts/ipq9570-kiwi-dvk.dts
+++ b/target/linux/qualcommbe/dts/ipq9570-kiwi-dvk.dts
@@ -72,6 +72,14 @@
 	status = "okay";
 };
 
+&crypto {
+	status = "disabled";
+};
+
+&cryptobam {
+	status = "disabled";
+};
+
 &qcom_ppe {
 	ethernet-ports {
 		#address-cells = <1>;
--- a/target/linux/ipq40xx/dts/qcom-ipq4018-a42.dts
+++ b/target/linux/ipq40xx/dts/qcom-ipq4018-a42.dts
@@ -99,10 +99,6 @@
 	status = "okay";
 };
 
-&crypto {
-	status = "okay";
-};
-
 &tlmm {
 	serial_pins: serial_pinmux {
 		mux {
@@ -178,10 +174,6 @@
 &blsp1_uart1 {
 	pinctrl-0 = <&serial_pins>;
 	pinctrl-names = "default";
-	status = "okay";
-};
-
-&cryptobam {
 	status = "okay";
 };
 
--- a/target/linux/ipq40xx/dts/qcom-ipq4018-ap120c-ac.dts
+++ b/target/linux/ipq40xx/dts/qcom-ipq4018-ap120c-ac.dts
@@ -99,10 +99,6 @@
 	status = "okay";
 };
 
-&crypto {
-	status = "okay";
-};
-
 &blsp_dma {
 	status = "okay";
 };
@@ -240,10 +236,6 @@
 
 	pinctrl-0 = <&serial0_pins>;
 	pinctrl-names = "default";
-};
-
-&cryptobam {
-	status = "okay";
 };
 
 &ethphy4 {
--- a/target/linux/ipq40xx/dts/qcom-ipq4018-ap4050dn.dts
+++ b/target/linux/ipq40xx/dts/qcom-ipq4018-ap4050dn.dts
@@ -51,10 +51,6 @@
 			qcom,wifi_noc_memtype_m0_m2 = <TCSR_WIFI_NOC_MEMTYPE_M0_M2>;
 		};
 
-		crypto@8e3a000 {
-			status = "okay";
-		};
-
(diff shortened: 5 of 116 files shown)


neheb:
I assume this would be an issue with the other platforms as well. Can you check those as well?

JuliusBairaktaris:
Sure will do

JuliusBairaktaris:
Audited all targets for the same issue:

- **ipq40xx**: 73 board device trees enabled the nodes (two also carried BAM channel properties that only served qce); qcom-ipq4019.dtsi defaults both to disabled. No ipq4019 supplier registers a sync_state() callback, so this was dead code only — dropped (new commit).
- **qualcommbe**: the opposite situation — no board overrides, but mainline ipq9574.dtsi ships both nodes enabled by default, and gcc-ipq9574 registers its interconnect provider with icc_sync_state. The driverless node defers it forever: the same pending message, and the interconnect bandwidth votes never drop from their boot-time maximum. Added explicit disables to the board device tree (new commit).
- **ipq806x**: no crypto/qce nodes and never had the QCE configs — not affected.
- No other target references the qce compatibles or CRYPTO_DEV_QCE.

The qualcommax change is run-tested on a Xiaomi AX3600; the ipq40xx/qualcommbe commits are the same mechanical removal, verified against the dtsi defaults — I don't have hardware for those targets.

target/linux/qualcommbe/dts/ipq9570-kiwi-dvk.dts
```diff
@@ -72,6 +72,14 @@
 	status = "okay";
 };
 
+&crypto {
+	status = "disabled";
+};
+
+&cryptobam {
```
neheb, line 79: lol so upstream kernel is broken because of them marking the driver as BROKEN. Wonderful.

neheb (approved):

neheb:
ping @jonasjelonek

AliLostInTheDark:
This same error msg was seen on my qualcommbe board. Idk the current state, but have to check for this message again.

target/linux/ipq40xx/dts/qcom-ipq4028-wpj428.dts
```diff
@@ -106,10 +106,6 @@
 	status = "okay";
 };
 
-&crypto {
```
openwrt-ai, line 109: One ipq40xx board still enables the driverless QCE engine and is not touched by this commit: [`qcom-ipq4018-ap4050dn.dts:54`](https://github.com/openwrt/openwrt/blob/f5d892028375e7ac0934ffb1a39dfd4a1da23c6b/target/linux/ipq40xx/dts/qcom-ipq4018-ap4050dn.dts#L54-L56) overrides the node to `okay` through the full node-path form inside the `soc { }` block rather than the `&crypto { }` label form the other boards use:

```
crypto@8e3a000 {
	status = "okay";
};
```

The PR-body audit ("73 board device trees enabled the nodes ... dropped"; "No other target references the qce compatibles") appears to have matched only the `&crypto`/`&cryptobam` label references and missed this node-path override, so ap4050dn keeps `crypto@8e3a000` enabled while every sibling now inherits the disabled dtsi default (`qcom-ipq4019.dtsi` ships it `disabled`). Its `dmas = <&cryptobam 2>, <&cryptobam 3>` also points at a `cryptobam` that stays disabled. Was leaving this one out intentional, or should the `crypto@8e3a000 { status = "okay"; }` block be dropped here too for parity?

---

openwrt-ai (commented):
Reviewed all 3 commits (fresh review). Commit messages match their diffs, and the dtsi default-status claims check out against the reference kernels (ipq8074/ipq4019 default `disabled`, ipq9574 default enabled). One completeness gap noted inline on ipq40xx.

---

AliLostInTheDark (approved):

jonasjelonek:
Why do so many devices activate this at all, just for the sake of having something available with crypto or was that actually used?

AliLostInTheDark:
> Why do so many devices activate this at all, just for the sake of having something available with crypto or was that actually used?

From what I have seen in my device (ipq9574/qualcommbe), this node seems to never be used, and just pop that error over and over again.
After I removed that from dts, this error never came back, and it's running smooth since.

JuliusBairaktaris:
Copy-paste inheritance — it came in with 54b275c8ed ("ipq40xx: add target") from QCA's vendor DTs together with `CONFIG_CRYPTO_DEV_QCE=y`, and every board added since was written by copying an existing .dts, so it propagated to all 73 without anyone deciding the box needed a crypto engine. It did win the priority contest on these targets (qce skcipher 275 vs `aes-neonbs` 250; none of them enable the ARMv8 CE AES drivers), but the only default-image consumer of the kernel skcipher API is IPsec ESP — WireGuard uses its own primitives and OpenSSL needs cryptodev/AF_ALG, so in practice it sat unused.

openwrt-ai (commented):
The only change since my last review drops the inline `crypto@8e3a000` override in `qcom-ipq4018-ap4050dn.dts`, which resolves the ipq40xx completeness gap I flagged earlier. No `&crypto`/`&cryptobam`/`crypto@` `okay` overrides remain in the ipq40xx, qualcommax, or qualcommbe device trees, and the commit messages still match their diffs. No new issues found.

---

neheb (approved):

AliLostInTheDark (approved):

jonasjelonek:

PR #24381, opened 2026-07-23
mediatek: power off ZBT Z8102AX modems on shutdown

dormancygrace:
The ZBT Z8102AX and Z8102AX V2 keep both modem power GPIOs asserted during a warm reboot. As a result, a modem does not receive a power cycle and can remain unavailable after a USB composition change that only takes effect after power removal.

Add an optional `shutdown_value` to the generic `gpio_switch` configuration. The value is applied when the service is stopped during shutdown, while a normal configuration reload only reapplies the configured running value. Configure both Z8102AX modem power switches (`5g1` and `5g2`) to go low on shutdown. PCIe power and SIM selection GPIOs are left unchanged.

Tested on a ZBT Z8102AX V2 with two Meig SLM828 modems:

- reload kept both GPIOs at `1` and both USB modems present
- stop changed both GPIOs to `0` and both USB modems disappeared
- start restored both GPIOs to `1` and both USB modems re-enumerated
- shell syntax and `git diff --check` pass
- generated board JSON contains `"shutdown": 0` for the modem power switch

### DIFF
--- a/package/base-files/files/bin/config_generate
+++ b/package/base-files/files/bin/config_generate
@@ -495,8 +495,8 @@ generate_gpioswitch() {
 
 	json_select gpioswitch
 		json_select "$cfg"
-			local name pin default
-			json_get_vars name pin default
+			local name pin default shutdown
+			json_get_vars name pin default shutdown
 			uci -q batch <<-EOF
 				delete system.$cfg
 				set system.$cfg='gpio_switch'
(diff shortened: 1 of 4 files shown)


openwrt-ai (commented):

---

dormancygrace:
Checked the package-release warning. `base-files` deliberately uses `PKG_RELEASE:=$(COMMITCOUNT)`, so the release changes automatically with the OpenWrt revision; there is no numeric release value to bump. Recent in-tree `base-files` content changes follow the same pattern. No Makefile change is needed here.

PR #24382, opened 2026-07-23
github: switch issue-labeller to bot webhook and add configuration

BKPepe:
Remove the standalone GitHub Actions issue-labeller.yml workflow in favor
of the webhook-based openwrt-bot worker.

Enable issue labeller in formalities.json and add a declarative
.github/issue-labeller.yml configuration file for automated validation
and labelling of bug reports (release, target, image kind, device).

The release rule lists two alternative conditions under a single label
key: concrete releases (x.y.z, x.y.z-rcN) are validated against their
Git tag via exists check, while stable-branch snapshots (x.y-SNAPSHOT)
are format-validated only — OpenWrt does not tag snapshot builds.


.github/issue-labeller.yml
```diff
@@ -0,0 +1,38 @@
...
+# Tag target (e.g. target/x86, target/ath79, target/ramips)
+"target/{segment0}":
+  - field: "target"
+    format: '^[a-zA-Z0-9]+/[a-zA-Z0-9]+$'
+    exists: "path:target/linux/{segment0}/{segment1}"
+
+# Tag image source
+"Official Image":
+  - field: "image_kind"
+    contains: "official"
```
openwrt-ai, line 34: The `image_kind` dropdown values are capitalized — [`Official downloaded image` / `Self-built image` at bug-report.yml:56-57](https://github.com/openwrt/openwrt/blob/e134d91b7916df2c4edb4e1c22a329b330798840/.github/ISSUE_TEMPLATE/bug-report.yml#L56-L57) — while this rule matches lowercase `official` (and line 33 matches lowercase `self`). If the bot's `contains` check is case-sensitive, neither rule will ever match a submitted report. Is `contains` case-insensitive on the openwrt-bot side? If not, the substrings here won't match the dropdown options.

---

.github/issue-labeller.yml
```diff
@@ -0,0 +1,38 @@
...
+_trigger_label: "to-triage"
+_invalid_label: "invalid"
+_remove_labels: ["to-triage"]
+_invalid_comment: "Invalid {field} reported. `{value}`"
+
+# Tag release (e.g. release/23.05 or release/24.10)
+"release/{major}.{minor}":
+  - field: "release"
+    format: '^\d+\.\d+\.\d+(-rc\d+)*$|^\d+\.\d+-SNAPSHOT$'
+    exists: "tag:v{value}"
```
openwrt-ai, line 16: The `format` above (line 12) accepts a stable-branch snapshot like `24.10-SNAPSHOT` via its second alternative `^\d+\.\d+-SNAPSHOT$`, but this `exists: "tag:v{value}"` check would then look for a tag `v24.10-SNAPSHOT`. No such tags exist — OpenWrt only tags concrete releases/rcs (e.g. `v24.10.7`, `v25.12.0-rc5`), so a genuine stable-snapshot report would be marked invalid. Is the `exists` check meant to apply only to the `x.y.z`/`-rc` forms? If so, the `-SNAPSHOT` alternative may need to live in a separate rule without the tag-existence check.

---

openwrt-ai (commented):
Reviewed 1 new commit; 2 inline questions on the new issue-labeller config, no blocking issues found.

---

.github/issue-labeller.yml
```diff
@@ -0,0 +1,43 @@
...
+# Tag release (e.g. release/23.05 or release/24.10)
+"release/{major}.{minor}":
+  - field: "release"
+    format: '^\d+\.\d+\.\d+(-rc\d+)*$'
+    exists: "tag:v{value}"
+
+# Tag stable-branch snapshot (e.g. 24.10-SNAPSHOT)
+"release/{major}.{minor}":
+  - field: "release"
+    format: '^\d+\.\d+-SNAPSHOT$'
```
openwrt-ai, line 18: Splitting the release rule into two entries with the identical mapping key `"release/{major}.{minor}"` produces duplicate keys in the same YAML mapping. YAML keys must be unique, and standard parsers silently keep only the last one: `yaml.safe_load()` of this file drops the first entry entirely, leaving only

```
release/{major}.{minor}: [{field: release, format: '^\d+\.\d+-SNAPSHOT$'}]
```

So the concrete-release rule (`^\d+\.\d+\.\d+(-rc\d+)*$` + `exists: "tag:v{value}"`) is lost, and a real report like `24.10.7` would fail the surviving `^\d+\.\d+-SNAPSHOT$` format check and be marked invalid — the reverse of the intended behavior. Both rules must live under a single key:

```suggestion
"release/{major}.{minor}":
  - field: "release"
    format: '^\d+\.\d+\.\d+(-rc\d+)*$'
    exists: "tag:v{value}"
  - field: "release"
    format: '^\d+\.\d+-SNAPSHOT$'
```

This assumes the bot treats multiple list items under one label as OR (any rule matching applies the label). If instead all items must match (AND), two mutually-exclusive `format` regexes would never both match — in that case the concrete/`-rc` and `-SNAPSHOT` alternatives need to stay in one regex, e.g. `^(\d+\.\d+\.\d+(-rc\d+)*|\d+\.\d+-SNAPSHOT)$`, with the `exists` check made conditional on the bot side. Either way, the duplicate key as written cannot work.

---

openwrt-ai (commented):
Reviewed 1 new commit (the release-rule split since my last review). One issue: the split introduces a duplicate YAML mapping key that drops the concrete-release validation. See inline comment.

---

openwrt-ai (commented):
Reviewed 1 new commit (the release-rule merge since my last review). The two `"release/{major}.{minor}"` mappings are now consolidated under a single key as a list of two alternative conditions, resolving the duplicate-key data loss I flagged previously. No new issues found.

---

PR #24384, opened 2026-07-23
realtek: restore rootfs_data for ews2910p and include firmware2

DamisDorgon:
In 2022, commit 7bba7ccde9cb021e24ec83f1332aafd5898e23ef repurposed a partition used by the vendor firmware for use as rootfs_data, the jffs2 filesystem used in the overlay. However, this had the effect of preventing sysupgrades from properly clearing and regenerating the overlay.

Revert that change and instead cannibalize the 8MiB firmware2 partition to provide extra space for the overlayfs.

@svanheule @mrnuke

### DIFF
--- a/target/linux/realtek/base-files/etc/board.d/05_compat-version
+++ b/target/linux/realtek/base-files/etc/board.d/05_compat-version
@@ -13,6 +13,8 @@ case "$(board_name)" in
 	hpe,1920-24g-poe-370w)
 		ucidef_set_compat_version "1.1"
 	;;
+	engenius,ews2910p-v1 | \
+	engenius,ews2910p-v3 | \
 	zyxel,gs1900-8-a1 | \
 	zyxel,gs1900-8-b1 | \
 	zyxel,gs1900-8hp-a1 | \
--- a/target/linux/realtek/dts/rtl8380_engenius_ews2910p.dtsi
+++ b/target/linux/realtek/dts/rtl8380_engenius_ews2910p.dtsi
@@ -153,20 +153,16 @@
 				reg = <0x90000 0x10000>;
 			};
 			partition@a0000 {
-				label = "rootfs_data";
+				label = "jffs2-cfg";
 				reg = <0xa0000 0xd60000>;
 			};
 			partition@e00000 {
 				label = "jffs2-log";
 				reg = <0xe00000 0x200000>;
 			};
-			firmware_partition_1: partition@1000000 {
+			firmware_partition: partition@1000000 {
 				label = "firmware";
-				reg = <0x1000000 0x800000>;
-			};
-			firmware_partition_2: partition@1800000 {
-				label = "firmware2";
-				reg = <0x1800000 0x800000>;
+				reg = <0x1000000 0x1000000>;
 			};
 		};
 	};
--- a/target/linux/realtek/image/common.mk
+++ b/target/linux/realtek/image/common.mk
@@ -41,8 +41,16 @@ endef
 # The "IMG-" uImage name allows flashing the iniramfs from the vendor Web UI.
 # Avoided for sysupgrade, as the vendor FW would do an incomplete flash.
 define Device/engenius_ews2910p
-  IMAGE_SIZE := 8192k
+  IMAGE_SIZE := 16384k
   DEVICE_VENDOR := EnGenius
+  DEVICE_COMPAT_VERSION := 2.0
+  DEVICE_COMPAT_MESSAGE := The secondary firmware partition has been cannibalized \
+	in this version. Do not upgrade if you want to preserve dual booting. \
+	Furthermore, check that the firmware you flash fits in the pre-existing \
+	8MiB firmware partition. On subsequent sysupgrades, this limitation \
+	will not apply. Any files you had previously in the overlayfs are \
+	available by manually mounting /dev/mtd3 with: \
+	mount -t jffs2 /dev/mtdblock3 /mnt
   KERNEL_INITRAMFS := \
 	kernel-bin | \
 	append-dtb | \
--- a/target/linux/realtek/dts/rtl8380_engenius_ews2910p-v1.dts
+++ b/target/linux/realtek/dts/rtl8380_engenius_ews2910p-v1.dts
@@ -7,7 +7,7 @@
 	model = "EnGenius EWS2910P v1";
 };
 
-&firmware_partition_1 {
+&firmware_partition {
 	compatible = "openwrt,uimage";
 	openwrt,ih-magic = <0x03802910>;
 };
--- a/target/linux/realtek/dts/rtl8380_engenius_ews2910p-v3.dts
+++ b/target/linux/realtek/dts/rtl8380_engenius_ews2910p-v3.dts
@@ -7,7 +7,7 @@
 	model = "EnGenius EWS2910P v3";
 };
 
-&firmware_partition_1 {
+&firmware_partition {
 	compatible = "openwrt,uimage";
 	openwrt,ih-magic = <0x03010500>;
 };


mrnuke:
Both support for the device and commit 7bba7ccde9cb021e24ec83f1332aafd5898e23ef were added by 23.05 release. So unless updating from a snapshot between the support and rootfs expansion, I don't see how sysupgrade fails. I am not convinced this is needed.

Can you expand on how sysupgrade fails?

openwrt-ai (commented):

---

DamisDorgon:
@mrnuke sysupgrade succeeds, but the rootfs_data is not cleared, so all the files there persist despite the -n option to sysupgrade. I was starting from an old version c0109537d13650e3cfd4d4840c571a0d557b303a, 6 commits before your change, `sysupgrade -n` to a new version, noticed the jffs2 wasn't cleared, then `sysupgrade -n` again to the same new version, and the jffs2 again wasn't cleared. I could see two partitions named rootfs_data, one in /dev/mtd3 and the other in the firmware1 split, /dev/mtd8. I could only clear the mounted /dev/mtd3 rootfs_data by explicitly running: `factoryreset -y && reboot &` (from /etc/rc.button/reset).

The normal way that sysupgrade preserves settings (at least in my mental model) is to generate a tarball "backup" file and write it into the trailing space and then during first boot, read the trailing tarball, reformat the jffs2 and then reinstall the files from the tarball. Because the rootfs_data partition since your https://github.com/openwrt/openwrt/commit/7bba7ccde9cb021e24ec83f1332aafd5898e23ef commit no longer lives in the trailing space, that stuff doesn't happen, or something.

I am trying to reproduce the error by flashing back to my c0109537d13650e3cfd4d4840c571a0d557b303a image, and repeating the sysupgrade sequence again.

mrnuke:
So sysupgrade fails from an arbitrary snapshot to a release version. I don't think we ever guaranteed that to work.

This change probably breaks compatibility with released versions. sysupgrade will wipe the old rootfs_data. To make that work, we'd need to add a DEVICE_COMPAT_MESSAGE or a defaults/ script that adjusts the layout on sysupgrade.

There's also the issue that the bootloader treats the firmware as two independent images that it can ping-pong between. When you write firmware2 from the bootloader, you've just destroyed firmware1 too.

Thirdly, there's a lot of space that's not being used as we're trading 14 MiB for 8 MiB.

While I am open to revisiting the layout if there is a valid bug, I think this change should **not** be merged.

DamisDorgon:
RESULTS:

From my old image, I erased /dev/mtd3 and /dev/mtd9 (firmware2). I `sysupgrade -v -n -F` to a vanilla r35519-128bf6ad48b3. Everything looks pretty normal, /dev/mtd3 is mounted as the overlayfs. I change the hostname and network config slightly and reboot. My new network config and hostname persist as expected. I fetch the same modern vanilla image I just flashed, and `sysupgrade -v -n` to it again. Everything seems to go fine, except when I get the shell again my changed hostname and network config is still there, when it should have been erased. It seems in master, without this commit, /dev/mtd3 is not getting erased and recreated.

mrnuke:
> `sysupgrade -v -n` to it again. Everything seems to go fine, except when I get the shell again my changed hostname and network config is still there, when it should have been erased.

(CC: @svanheule) . It sounds like `default_do_upgrade` assumes that rootfs_data comes after firmware. We might need something special for ewsp2910p in lib/upgrade/platform.sh.

DamisDorgon:
Fwiw, I don't particularly care so much about this commit, but I do care about the `sysupgrade -n` process working as expected. Whatever the most aesthetically pleasing solution is found, that's fine by me. 

I'm less concerned about losing dual images. That apparently no one complained about this bug in 4 years implies to me that not many people are using the device with OpenWrt.

openwrt-ai (commented):

---

target/linux/realtek/dts/rtl8380_engenius_ews2910p.dtsi
```diff
@@ -153,20 +153,16 @@
 				reg = <0x90000 0x10000>;
 			};
 			partition@a0000 {
-				label = "rootfs_data";
+				label = "jffs2-cfg";
```
openwrt-ai, line 156: This relocates the root overlay for every existing installation: today mtd3 is labelled `rootfs_data` and is what gets mounted as the overlay, after this change the overlay is the mtdsplit-generated tail of `firmware`, and lines 163-165 additionally fold the second boot slot into `firmware`. A sysupgrade from a released image will therefore silently leave the old mtd3 content orphaned and change the flash layout under the bootloader — but nothing in the diff signals that to the user.

The realtek target already has a precedent for exactly this situation: [`Device/zyxel_gs1900` at common.mk:86-91](https://github.com/openwrt/openwrt/blob/c6d28704f1f9078a54e5b5a3b315aafb3d4e1f89/target/linux/realtek/image/common.mk#L86-L91) merges the dual firmware partitions and carries `DEVICE_COMPAT_VERSION := 2.0` together with a `DEVICE_COMPAT_MESSAGE` explaining that a fresh factory install is required. Consider adding an equivalent version bump and message to [`Device/engenius_ews2910p` at common.mk:43](https://github.com/openwrt/openwrt/blob/c6d28704f1f9078a54e5b5a3b315aafb3d4e1f89/target/linux/realtek/image/common.mk#L43-L44) so sysupgrade refuses to proceed without an explicit override and the user is told what will be lost. (The same concern was raised in the discussion thread; noting the in-tree precedent for how it is normally handled.)

---

openwrt-ai (commented):
The new head is a rebase of the previously reviewed commit — the diff content is byte-identical to `0c62b3a`. One inline note on the flash-layout migration.

---

target/linux/realtek/dts/rtl8380_engenius_ews2910p.dtsi
DamisDorgon: A fresh factory install is not required. A simple sysupgrade transitions to the new partitioning. The only concern would be if somehow you are booting from firmware2. I'm not sure if that is even a possible scenario given the status quo.

DamisDorgon: The content of /dev/mtd3 is not overwritten during a sysupgrade to this commit. Anything you want to fish out of the "orphaned" /dev/mtd3 is a `mount -t jffs2 /dev/mtd3 /mnt` away.

DamisDorgon: Some care is required, depending on the size of the firmware image. If it exceeds 8MiB, then it won't fit in the original pre-commit /dev/mtd5, but after flashing a <8MiB image first (with this commit), then the partition will be big enough to accept a ≤16MiB firmware.

openwrt-ai (commented):

`f894f72` is a rebase of the previously reviewed `c6d2870` — `git range-diff` reports the commit (message and diff) as identical, so nothing new to flag. The existing thread on the flash-layout migration is still open and unchanged by this push.

---

DamisDorgon:
new in this version: added a DEVICE_COMPAT_VERSION and DEVICE_COMPAT_MESSAGE. Also modified the IMAGE_SIZE variable to reflect the larger size.

DamisDorgon:
new in this version, remembered to put in the board.d file changes to record the new compat version in uci.

target/linux/realtek/image/common.mk
```diff
@@ -41,8 +41,16 @@ endef
...
+  IMAGE_SIZE := 16384k
   DEVICE_VENDOR := EnGenius
+  DEVICE_COMPAT_VERSION := 2.0
+  DEVICE_COMPAT_MESSAGE := The secondary firmware partition has been cannibalized \
+	in this version. Do not upgrade if you want to preserve dual booting. \
+	Furthermore, check that the firmware you flash fits in the pre-existing \
+	8MiB firmware partition. On subsequent sysupgrades, this limitation \
+	will not apply. Any files you had previously in the overlayfs are \
+	available by manually mounting /dev/mtd3 with: \
+	mount -t jffs2 /dev/mtd3 /mnt
```
openwrt-ai: The recovery command handed to the user does not work: JFFS2 cannot be mounted from the MTD *character* device. [`get_tree_mtd()` at mtdsuper.c:116-160](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/mtd/mtdsuper.c#L116-L160) accepts a source only if it literally starts with `mtd` (`mtd3`, `mtd:jffs2-cfg`); anything else falls through to `lookup_bdev()`, which requires an `MTD_BLOCK_MAJOR` *block* device. `/dev/mtd3` is a char device, so the mount fails with `-ENOTBLK` and the user sees `MTD: Couldn't look up '/dev/mtd3'` instead of their old files.

`/dev/mtdblock3` (or `mtd3` / `mtd:jffs2-cfg`, which are immune to renumbering) is what actually works:

```suggestion
	available by manually mounting /dev/mtdblock3 with: \
	mount -t jffs2 /dev/mtdblock3 /mnt
```

---

target/linux/realtek/image/common.mk
```diff
@@ -41,8 +41,16 @@ endef
 # The "IMG-" uImage name allows flashing the iniramfs from the vendor Web UI.
 # Avoided for sysupgrade, as the vendor FW would do an incomplete flash.
 define Device/engenius_ews2910p
-  IMAGE_SIZE := 8192k
+  IMAGE_SIZE := 16384k
```
openwrt-ai, line 44: Raising `IMAGE_SIZE` to `16384k` also drops the build-time guard that kept the sysupgrade image flashable on the *pre-migration* layout, and that is exactly the one flash where it matters. On a currently-released installation the `firmware` mtd is still `reg = <0x1000000 0x800000>` (8 MiB), [`platform_check_image()` at platform.sh:23-25](https://github.com/openwrt/openwrt/blob/0031dd2cd79c77958d1d3b5c8c2e498b5f28ccff/target/linux/realtek/base-files/lib/upgrade/platform.sh#L23-L25) returns 0 unconditionally, and [`platform_do_upgrade()`](https://github.com/openwrt/openwrt/blob/0031dd2cd79c77958d1d3b5c8c2e498b5f28ccff/target/linux/realtek/base-files/lib/upgrade/platform.sh#L58-L60) falls through to `default_do_upgrade` with `PART_NAME=firmware`. So an image between 8 MiB and 16 MiB passes `check-size`, passes `platform_check_image`, and then `mtd write` runs off the end of the partition after the kernel has already been erased — TFTP/serial recovery.

I see the `DEVICE_COMPAT_MESSAGE` below asks the user to verify this by hand, so it is a known trade-off rather than an oversight. Still, given the failure mode is a brick, would it be worth backing the warning with a check — e.g. a `platform_check_image()` branch for `engenius,ews2910p-*` that compares the image size against the size of the running `firmware` partition and refuses when it doesn't fit? That way the manual check is only advisory, and the constraint disappears by itself once the device is on the 16 MiB layout.

---

target/linux/realtek/base-files/etc/board.d/05_compat-version
```diff
@@ -13,6 +13,8 @@ case "$(board_name)" in
 	hpe,1920-24g-poe-370w)
 		ucidef_set_compat_version "1.1"
 	;;
+	engenius,ews2910p-v1 | \
+	engenius,ews2910p-v3 | \
```
openwrt-ai, line 17: nit: this hunk is correct — `2.0` here matches `DEVICE_COMPAT_VERSION := 2.0` in `common.mk`, mirroring how `hwmon-fan-migration`/`1.1` is paired with the `hpe,1920-*-poe-*` case above — but the commit message that introduces it reads as a record of PR revisions rather than of the change:

> set IMAGE_SIZE to reflect the larger size and add a DEVICE_COMPAT_VERSION and message.
>
> added the board.d file changes to record the compat version in uci

Since this is a single squashed commit, consider folding those two paragraphs into the body so it describes the end state (new layout, the compat bump and why sysupgrade now needs an explicit override) rather than the order the pieces were added in.

---

openwrt-ai (commented):
Reviewed the changes added since `f894f72`: the `DEVICE_COMPAT_VERSION`/`DEVICE_COMPAT_MESSAGE` pair, the `IMAGE_SIZE` bump, and the matching `05_compat-version` entry. The compat bump addresses the migration concern raised in the earlier thread and is consistent with the `Device/zyxel_gs1900` precedent in the same file. Three inline notes: one definite bug in the recovery command inside the compat message, one question about the widened `IMAGE_SIZE` during the migration flash, and one commit-message nit. CI is green on `0031dd2`.

---

mrnuke (changes requested):
Making the layout upgrade as proposed, and then forcing an update incompatibility seems too heavy-handed and unnecessary when there are more elegant solutions..

`default_do_upgrade` assumes that the data is appended to rootfs. The ews2910 uses this path, and perhaps it shouldn't. I think we instead need to change `platform.sh` to account for this layout.

DamisDorgon:
@mrnuke In the meantime, this uses the existing sysupgrade method and follows the model of gs1900 of just ditching the dual firmware, which provides ample flash space for a managed switch.

target/linux/realtek/image/common.mk
DamisDorgon: Any change I make to platform_check_image() comes too late. It is only useful on the device's existing firmware where the check is made, and which the new firmware cannot change. Changing it in this commit does not help, because it is not installed until after the sysupgrade.

DamisDorgon:
Looking at the examples of special case upgrade functions, they look like hairy invasive messes, e.g. package/base-files/files/lib/upgrade/nand.sh and target/linux/realtek/base-files/lib/upgrade/upgrade_dualboot_plasmacloud.sh

I don't have the energy or inclination to pursue a custom upgrade method. This seems like the much simpler and effective approach.

target/linux/realtek/image/common.mk
```diff
@@ -41,8 +41,16 @@ endef
...
   DEVICE_VENDOR := EnGenius
+  DEVICE_COMPAT_VERSION := 2.0
+  DEVICE_COMPAT_MESSAGE := The secondary firmware partition has been cannibalized \
+	in this version. Do not upgrade if you want to preserve dual booting. \
+	Furthermore, check that the firmware you flash fits in the pre-existing \
+	8MiB firmware partition. On subsequent sysupgrades, this limitation \
+	will not apply. Any files you had previously in the overlayfs are \
+	available by manually mounting /dev/mtd3 with: \
```
openwrt-ai, line 52: nit: the `mount` command on the next line was corrected to `/dev/mtdblock3`, but the prose in this line still names `/dev/mtd3`, so the message tells the user to mount one node and then hands them a command for a different one. Since `/dev/mtd3` is the char device that JFFS2 cannot be mounted from, naming it here is the part that would mislead someone who types the path by hand rather than copying the command.

```suggestion
	available by manually mounting /dev/mtdblock3 with: \
```

---

openwrt-ai (commented):
Reviewed the changes since `0031dd2`. `git range-diff` shows exactly two: the commit-message body was rewritten to describe the end state instead of the PR revision history (this resolves the earlier commit-message nit), and the `DEVICE_COMPAT_MESSAGE` recovery command was changed from `/dev/mtd3` to `/dev/mtdblock3`. Both are improvements; the rest of the diff is unchanged. One inline nit on the `mtdblock3` fix being applied to only one of the two places that name the device node. No CI checks are reported on this head SHA, so nothing to ground there.

---

target/linux/realtek/image/common.mk
DamisDorgon: The /dev/mtd3 and /dev/mtdblock3 are the same partition. A human understand this, excessive verbosity is not required.

openwrt-ai (commented):

`1c5841b` is a pure rebase of the previously reviewed `31e86f4` — `git range-diff` reports the commit (message and diff) as identical, and none of the realtek commits picked up in the rebase touch the three files this PR changes, so the rebase is a no-op for this change. Re-verified after the rebase: `05_compat-version` places `engenius,ews2910p-v1`/`-v3` in the `ucidef_set_compat_version "2.0"` branch, which matches `DEVICE_COMPAT_VERSION := 2.0` in `Device/engenius_ews2910p`, and both `.dts` files still resolve `&firmware_partition` against the renamed label in the `.dtsi`. All earlier review threads are resolved; no CI checks are reported on this head SHA.

The open design question from @mrnuke (custom `platform_do_upgrade` handling instead of the layout change plus compat bump) is a maintainer call, not something this review adds to.

---

DamisDorgon:
If we are willing to modify the u-boot-env, we could change `bootcmd` to be `bootm 0xb40a0000` and change the partitioning to extend /dev/mtd3 (renamed as 'firmware') to go from 0xa0000 to 0x2000000. bootcmd is currently `boota` which handles the dual booting, afaict. That's assuming that we don't care about dual booting and we want to maximize flash space available.

Seems to work, probably the easiest way to get there would be to tftpboot a suitable initramfs version, sysupgrade from there. u-boot or a writable /dev/mtd1 would be needed to saveenv the u-boot-env change.

```
# fw_printenv bootcmd
bootcmd=bootm 0xb40a0000
```

```
[...]
[    0.851620] spi-nor spi0.0: supply vcc not found, using dummy regulator
[    0.861015] 4 fixed-partitions partitions found on MTD device spi0.0
[    0.869145] Creating 4 MTD partitions on "spi0.0":
[    0.874680] 0x000000000000-0x000000080000 : "u-boot"
[    0.887683] 0x000000080000-0x000000090000 : "u-boot-env"
[    0.898733] 0x000000090000-0x0000000a0000 : "u-boot-env2"
[    0.909725] 0x0000000a0000-0x000002000000 : "firmware"
[    0.920947] 2 uimage-fw partitions found on MTD device firmware
[    0.927774] Creating 2 MTD partitions on "firmware":
[    0.933497] 0x000000000000-0x000000410000 : "kernel"
[    0.943967] 0x000000410000-0x000001f60000 : "rootfs"
[    0.954539] mtd: setting mtd5 (rootfs) as root device
[    0.960516] 1 squashfs-split partitions found on MTD device rootfs
[    0.967624] 0x000000770000-0x000001f60000 : "rootfs_data"
[...]
```

But, for me, cannibalizing firmware2 is enough. 24MB of jffs2 is kind of gratuitous.

PR #24389, opened 2026-07-23
kernel: make INITRAMFS_PRESERVE_MTIME selectable

micpf:
The Linux kernel option `CONFIG_INITRAMFS_PRESERVE_MTIME` controls whether files unpacked from the built-in initramfs keep the mtimes recorded in the cpio archive, rather than being stamped with the boot time.

OpenWrt currently forces this option off from both `Kernel/SetInitramfs` and `Kernel/SetNoInitramfs`, so a target-level `config-<ver>` preference is overwritten unconditionally.

This patch introduces `TARGET_INITRAMFS_PRESERVE_MTIME` so users can opt in without patching the build system. Default behaviour is unchanged (option remains off unless the new symbol is set).

Motivation: userspace on a RAM rootfs may rely on file mtimes as freshness or ordering signals; with the current forced-off behaviour those all show the boot time.

### DIFF
--- a/config/Config-images.in
+++ b/config/Config-images.in
@@ -63,6 +63,15 @@ menu "Target Images"
 			help
 			  Ignore the initramfs passed by the bootloader.
 
+		config TARGET_INITRAMFS_PRESERVE_MTIME
+			bool "Preserve cpio archive mtimes in initramfs"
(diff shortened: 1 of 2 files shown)


openwrt-ai (commented):

---

PR #24390, opened 2026-07-23
realtek: convert to generic machine initialization

plappermaul:
Upstream has implemented MACH_REALTEK_RTL with the MIPS generic startup framework. Until now OpenWrt disables that with a patch and instead uses dedicated legacy startup sources.

Convert to upstream startup and take over the missing downstream bits from the legacy code. Due to the complex nature of this patch only focus on board-realtek.c for now. Moving other pieces around (like headers) needs to be done in a separate patch.


target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
```diff
@@ -0,0 +1,454 @@
...
++		pr_err("Missing 'cpus' DT node, using default frequency.");
++	} else {
++		if (of_property_read_u32(np, "frequency", &freq) < 0)
++			pr_err("No 'frequency' property in DT, using default.");
++		else
++			pr_info("CPU frequency from device tree: %dMHz", freq / 1000000);
++		of_node_put(np);
++	}
++
++	return freq / 2;
```
openwrt-ai, line 69: This handler drops the clock-framework measurement the old `plat_time_init` in `setup.c` performed (`of_clk_get(cpu_node, 0)` → `clk_get_rate(clk) / 2`), keeping only the `cpus`/`frequency` DT-property path. Because it is registered as `.measure_hpt_freq`, the generic `plat_time_init` skips its own clk branch entirely ([arch/mips/generic/init.c:159-161](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/arch/mips/generic/init.c#L159-L161)), so this becomes the sole source of `mips_hpt_frequency`.

But the rtl838x and rtl839x device trees have no `frequency` property on the `cpus` node — they express the CPU clock via `clocks = <&ccu CLK_CPU>` ([rtl838x.dtsi:35](https://github.com/openwrt/openwrt/blob/e7cbde769750f5a3ab1b79b65c7a7480f90eb7f2/target/linux/realtek/dts/rtl838x.dtsi#L35), [rtl839x.dtsi:34](https://github.com/openwrt/openwrt/blob/e7cbde769750f5a3ab1b79b65c7a7480f90eb7f2/target/linux/realtek/dts/rtl839x.dtsi#L34)). For those SoCs `of_property_read_u32(np, "frequency", ...)` fails and the function returns the hardcoded default `500000000 / 2` (250 MHz) regardless of the real CPU clock — a timer-frequency regression versus the old code, which read the actual rate from the clk framework. Only rtl930x/rtl931x carry a `frequency` property. Consider trying `of_clk_get()`/`clk_get_rate()/2` first and only falling back to the `frequency` property, as `setup.c` did.

---

target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
```diff
@@ -0,0 +1,454 @@
...
++	u32 model = realtek_read_model();
++
++	realtek_parse_model(model);
++	realtek_set_system_type();
++	realtek_get_system_memory();
++
++	pr_info("%s SoC with %d MB\n", get_system_type(), soc_info.memory_size >> 20);
++
++	realtek_prepare_highmem();
++	realtek_smp_register();
```
openwrt-ai: `realtek_smp_register()` runs here during `realtek_prom_init()` (called from `realtek_fixup_fdt`, i.e. `plat_mem_setup`), registering the custom `realtek_smp_ops`. But the generic `device_tree_init()` runs afterwards and calls `register_vsmp_smp_ops()`, which on mipsmt SoCs (rtl838x/rtl839x) invokes `register_smp_ops(&vsmp_smp_ops)` and replaces `mp_ops` with the stock vsmp ops ([arch/mips/generic/init.c:118](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/arch/mips/generic/init.c#L118)). `register_smp_ops` overrides unconditionally (it only prints "Overriding previously set SMP ops"), so the custom `realtek_smp_init_secondary` / `realtek_smp_finish` are clobbered and lost on exactly the SoCs where mipsmt applies. The old downstream `device_tree_init` avoided this by registering the custom vsmp ops *after* the cps/up registration. As written, `realtek_smp_register()` registers *before* the generic `device_tree_init`, so its effect is silently undone — is that intended? The custom ops need to be installed after the generic `device_tree_init` for them to survive.

---

target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
```diff
@@ -0,0 +1,454 @@
...
++	node = of_find_compatible_node(NULL, NULL, "diodes,pt7a75xx-wdt");
++	if (node) {
++		if (soc_info.family == RTL9310_FAMILY_ID) {
++			pr_info("apply quirk for diodes pt7a75xx watchdog\n");
++			sw_w32_mask(GENMASK(13, 12), BIT(12), RTL931X_LED_GLB_CTRL);
++			sw_w32_mask(0x0, BIT(8), RTL931X_MAC_L2_GLOBAL_CTRL2);
++		};
++		of_node_put(node);
++	}
++
```
openwrt-ai, line 85: nit: tab-only trailing-whitespace line; checkpatch will warn. Same at line 421 (the blank line before `realtek_prom_init();`).

```suggestion
+
```

---

openwrt-ai (commented):
Reviewed the generic-machine-init conversion. Two functional concerns flagged inline (CPU timer frequency source for rtl838x/rtl839x, and SMP-ops override ordering) plus one whitespace nit.

---

target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
plappermaul: The devices do not use the internal Timer but the external Realtek timer. Leave the code as is for now.

target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
```diff
@@ -0,0 +1,407 @@
...
+ cflags-$(CONFIG_MIPS_GENERIC)	+= -I$(srctree)/arch/mips/include/asm/mach-generic
+ 
+ load-$(CONFIG_MIPS_GENERIC)	+= 0xffffffff80100000
+--- a/arch/mips/generic/board-realtek.c
++++ b/arch/mips/generic/board-realtek.c
+@@ -10,6 +10,345 @@
+ 
+ #include <asm/fw/fw.h>
+ #include <asm/machine.h>
++#include <asm/mips-cps.h>
```
openwrt-ai: nit: `#include <asm/mips-cps.h>` is now unused after this revision. Its only consumer was `mips_gic_present()` (declared in `asm/mips-gic.h`, pulled in transitively by `mips-cps.h`) inside `realtek_smp_init_secondary()`, which was removed here along with the rest of the SMP-ops override. No remaining symbol in `board-realtek.c` comes from `mips-cps.h` or its transitive `mips-cm.h`/`mips-cpc.h`/`mips-gic.h`, so this include can be dropped.

---

openwrt-ai (commented):
Incremental review since the last round (rebased onto current main). Within the PR's own files, only `300-02-enhance-realtek-board-setup.patch` changed: it drops the `realtek_smp_*` override — which addresses the earlier SMP-ops registration-ordering concern — and moves `realtek_prepare_highmem()` ahead of the boot `pr_info()`. The prior CPU-timer-frequency concern was resolved by the author (external Realtek timer). One follow-up nit inline.

---

openwrt-ai (commented):

The only change since the last review drops the now-unused `#include <asm/mips-cps.h>` (addressing the prior nit) and correctly adjusts the hunk header count from 345 to 344.

---

jonasjelonek:
Thanks for the work, looks good to me so far. I can give this a test later (if desired ;)

plappermaul:

PR #24398, opened 2026-07-24
econet: en7528: add UBI support to DASAN H660GM-A

naseef:
Move the DASAN H660GM-A root filesystem from a raw squashfs in the tclinux partition to UBI, following the layout already used by the en751221 ChinaMobile GS3101.

tclinux is split into an explicit 4MB tclinux_kernel partition holding the raw LZMA kernel and a tclinux_rootfs partition covering the remainder. The latter is concatenated with the unused "unknown" partition into a single UBI volume, giving 153MB on the Airtel variant and 41MB on the Generic one. The two regions are not adjacent, so this needs mtd virtual concat, which is enabled for the subtarget.

The TRX plus UBI image recipe is identical to the one the GS3101 uses, so factor it out into a Device/tclinux-ubi template instead of duplicating it.

The .trx is written linearly into tclinux by the vendor installer, so an oversized image would run past its end into tclinux_slave and destroy the vendor firmware in the alternate slot. FACTORY_SIZE guards against this and is set to the size of the partition the .trx is written into.

Existing installations have to be reflashed with the .trx rather than sysupgraded, since the ubi MTD only appears once the new device tree is running.


naseef:
@cjdelisle For your kind review

openwrt-ai (commented):

---

target/linux/econet/image/Makefile
```diff
@@ -52,6 +52,19 @@ define Device/Default
...
+# Kernel in a raw TRX partition, root squashfs in UBI. FACTORY_SIZE must be the
+# size of the partition the .trx is written into; that write is linear, so an
+# oversized image overruns it into the next partition.
+define Device/tclinux-ubi
+  KERNEL_SIZE := 4096k
+  KERNEL := kernel-bin | append-dtb | lzma | kernel-trx
+  KERNEL_INITRAMFS := kernel-bin | append-dtb
+  IMAGES := tclinux.trx sysupgrade.bin
+  IMAGE/tclinux.trx := append-kernel | pad-to $$$$(KERNEL_SIZE) | append-ubi | \
+    check-size $$$$(FACTORY_SIZE)
```
cjdelisle, line 64: I don't remember, is this a platform which doesn't use the TRX header? Just noting that it's not made here...

Okay never mind, I wasn't looking properly. LGTM.

cjdelisle:
So it seems you've got enough space to leave `tclinux_slave` empty. I gather the vendor OS does not use `unknown` (?)

If you have used enough space that the vendor OS is unable to function in any case, I would suggest to consider blocking another 4MB for the slave kernel, and then allocate the rest of the slave slot since it will not be usable for anything except OpenWrt anyway, and space is kind of at a premium.

Looks good to my eyes.

naseef:
> So it seems you've got enough space to leave `tclinux_slave` empty. I gather the vendor OS does not use `unknown` (?)

Correct ,  `unknown` is not used by the vendor OS, so folding it into UBI is safe.

> If you have used enough space that the vendor OS is unable to function in any case, I would suggest to consider blocking another 4MB for the slave kernel, and then allocate the rest of the slave slot since it will not be usable for anything except OpenWrt anyway, and space is kind of at a premium.

That case doesn't apply here , the vendor OS still boots and functions with this layout. Only `unknown` is repurposed; `tclinux_slave`, `romfile` and `rwfs` are left intact, and dasan's `en75_chboot` factory switch still works, so slot B stays a working revert-to-stock path.

cjdelisle:
OK, LGTM.

naseef:
@jonasjelonek This one is ready, appreciate your review

jonasjelonek:
Should this maybe use COMPAT_VERSION to avoid breaking the device with a sysupgrade?

naseef:
> Should this maybe use COMPAT_VERSION to avoid breaking the device with a sysupgrade?

Not needed here, I think. The previous layout has no sysupgrade path and the old recipe built only `tclinux.trx` ( the factory bootloader image) , no `sysupgrade.bin`. The only way onto the UBI layout is reflashing the `.trx` via the bootloader ,(  a raw
write that doesn't consult COMPAT_VERSION ). 

Since econet is source-only, I'd also assume the only existing installs are self-built snapshots, which move onto the new layout via the `.trx` in any case.

jonasjelonek:

PR #24399, opened 2026-07-24
econet: en75_bmt: honor per-partition enable-remap

naseef:
The per-partition branch of w_init() tested econet,enable-remap on the parent nand node (np) instead of the partition being iterated (part_np). That branch only runs when np does not carry the property, so the test was always false and the loop skipped every partition. As a result econet,enable-remap on a partition node was a silent no-op: no partition ever registered a remap range, and "enable-remap set for ..." was never printed.

Read the property from part_np so per-partition enable-remap works as documented. Boards that place the property on a partition now register that range; boards that place it on the nand node are unaffected.

cc: @cjdelisle

### DIFF
--- a/target/linux/econet/files/drivers/mtd/nand/en75_bmt.c
+++ b/target/linux/econet/files/drivers/mtd/nand/en75_bmt.c
@@ -1257,7 +1257,7 @@ static int w_init(struct en75_bmt_m *ctx, struct device_node *np)
 			const __be32 *reg;
 			int ret;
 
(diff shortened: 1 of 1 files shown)


openwrt-ai (commented):

---

cjdelisle:
Pretty obvious from looking at it that this was a bug. ACK.

naseef:
@jonasjelonek Can you please review this when you have time?

jonasjelonek:

PR #24401, opened 2026-07-24
download: support spoofed User-Agent per download

sch-m:
Some servers/WAFs reject requests whose User-Agent doesn't look like a browser (e.g. HTTP 403), even though the same URL is reachable from a normal desktop browser. At the moment, the generic download infrastructure had no way to influence the User-Agent sent by curl, wget or aria2c.

Add a new SPOOF_USER_AGENT field to Download/Defaults (default empty, i.e. unchanged behaviour for every existing package). It takes the literal User-Agent string to send:
```
  define Download/foo
    URL:=https://example.com/dl/
    FILE:=foo.tar.gz
    HASH:=...
    SPOOF_USER_AGENT:=Mozilla/5.0 (X11; Ubuntu; Linux x86_64; ...
  endef
  $(eval $(call Download,foo))
```
This also works for the package's main source (PKG_SOURCE_URL) by setting PKG_SPOOF_USER_AGENT in the Makefile.

### DIFF
--- a/scripts/download.pl
+++ b/scripts/download.pl
@@ -28,6 +28,7 @@ my $ok;
 
 my $check_certificate = $ENV{DOWNLOAD_CHECK_CERTIFICATE} eq "y";
 my $custom_tool = $ENV{DOWNLOAD_TOOL_CUSTOM};
+my $spoofed_user_agent = $ENV{SPOOF_USER_AGENT} // '';
 my $download_tool;
 
 $url_filename or $url_filename = $filename;
@@ -126,11 +127,13 @@ sub download_cmd {
 	if ($download_tool eq "curl") {
 		return (qw(curl -f --connect-timeout 5 --retry 3 --location),
 			$check_certificate ? () : '--insecure',
+			$spoofed_user_agent ne '' ? ('--user-agent', $spoofed_user_agent) : (),
 			shellwords($ENV{CURL_OPTIONS} || ''),
 			$url);
 	} elsif ($download_tool eq "wget") {
 		return (qw(wget --tries=3 --timeout=5 --output-document=-),
 			$check_certificate ? () : '--no-check-certificate',
+			$spoofed_user_agent ne '' ? ("--user-agent=$spoofed_user_agent") : (),
 			shellwords($ENV{WGET_OPTIONS} || ''),
 			$url);
 	} elsif ($download_tool eq "aria2c") {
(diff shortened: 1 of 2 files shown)


scripts/download.pl
```diff
@@ -144,6 +147,7 @@ sub download_cmd {
 			"touch $ENV{'TMPDIR'}/aria2c/${rfn}_spp;",
 			qw(aria2c --stderr -c -x2 -s10 -j10 -k1M), $url, $additional_mirrors,
 			$check_certificate ? () : '--check-certificate=false',
+			$spoofed_user_agent ne '' ? ("--user-agent=$spoofed_user_agent") : (),
```
openwrt-ai: Unlike the curl/wget branches, the aria2c command is built as a single joined string and therefore run through the shell (`download_cmd` returns one scalar, so `open(FETCH_FD, '-|', @cmd)` in `download()` invokes `/bin/sh`). A realistic User-Agent — e.g. the `Mozilla/5.0 (X11; Ubuntu; Linux x86_64; ...)` value from the commit message — contains spaces, `(` and `;`, so an unquoted `--user-agent=$spoofed_user_agent` will word-split and let the shell interpret `(`/`;`, breaking the aria2c download (and it's a shell-injection surface). Shell-quote the value:

```suggestion
			$spoofed_user_agent ne '' ? ("--user-agent='$spoofed_user_agent'") : (),
```

Single-quoting covers realistic UA strings; a UA containing a literal single quote would still need a proper quoting helper. The curl and wget branches are fine since they're exec'd as an argument list.

---

openwrt-ai (commented):
Reviewed 1 new commit. One inline issue on the aria2c code path; the curl/wget paths and the `include/download.mk` plumbing look correct. Commit message matches the diff.

---

openwrt-ai (commented):

The force-push only changes the aria2c branch in `scripts/download.pl` to shell-quote the User-Agent (`--user-agent='$spoofed_user_agent'`), which resolves the previous finding — that branch is the only one joined into a single string and therefore run through `/bin/sh`. Re-checked the rest at this head: the curl/wget branches are exec'd as argument lists so they need no quoting; `--user-agent` is placed before `CURL_OPTIONS`/`WGET_OPTIONS`/`ARIA2C_OPTIONS` so a user-supplied `-A`/`-U` still wins; and the `include/download.mk` plumbing (`SPOOF_USER_AGENT="$(SPOOF_USER_AGENT)"` prefix, `Download/Defaults` entry, `PKG_SPOOF_USER_AGENT` mapping) survives the `$(call locked,...)` `flock -c '...'` wrapper intact. Commit message still matches the diff.

---

openwrt-ai (commented):

The new head is a pure rebase onto `main` — `git show 5e93a50` and `git show c23442b` are byte-identical apart from the commit SHA, so the `include/download.mk` and `scripts/download.pl` changes are unchanged from the head reviewed on 2026-07-28. The previously raised aria2c quoting issue stays resolved, and the commit message still matches the diff. CI on `c23442b` reports nothing failed.

---

PR #24403, opened 2026-07-24
tools/cmake: pass -DBUILD_TESTING=OFF to bootstrap instead of patching

BKPepe:
Instead of carrying a patch, which was added by @neheb in https://github.com/openwrt/openwrt/commit/8af80e79bd5b750bf9a5ab01a6ecdb3971248bae and I think after 6 years, it is time to get rid of it. :)

The patch modifies CMake's internal modules (CTest.cmake, Dart.cmake) and needs to be rebased on every CMake update, pass `-DBUILD_TESTING=OFF` directly to CMake's bootstrap script via `-- -DBUILD_TESTING=OFF`. This disables building CMake's test suite cleanly without requiring a patch.


openwrt-ai (commented):

---

neheb:
Before:

```
________________________________________________________
Executed in  151.19 secs    fish           external
   usr time   44.82 mins    0.00 micros   44.82 mins
   sys time    1.61 mins  963.00 micros    1.61 mins
```

After:

```
________________________________________________________
Executed in  153.78 secs    fish           external
   usr time   45.42 mins  747.00 micros   45.42 mins
   sys time    1.65 mins   25.00 micros    1.65 mins
```

seems fine.

peterwillcn (approved):

tools/cmake/patches/100-no-testing.patch
```diff
@@ -1,33 +0,0 @@
---- a/Modules/CTest.cmake
-+++ b/Modules/CTest.cmake
-@@ -60,7 +60,7 @@ the :variable:`CTEST_USE_LAUNCHERS` vari
- in the ``CTestConfig.cmake`` file.
- #]=======================================================================]
- 
--option(BUILD_TESTING "Build the testing tree." ON)
-+option(BUILD_TESTING "Build the testing tree." OFF)
```
openwrt-ai, line 8: This hunk (and the `Modules/Dart.cmake` one below) is not equivalent to `-- -DBUILD_TESTING=OFF` on the bootstrap line — the two changes affect different builds.

`Modules/CTest.cmake` is *installed* into `$(STAGING_DIR_HOST)/share/cmake-$(PKG_VERSION_MAJOR)/Modules/`, so patching the `option()` default here changed the `BUILD_TESTING` default for **every** package in the tree that does `include(CTest)`. The new `-- -DBUILD_TESTING=OFF` is a cache entry for CMake's *own* second-stage configure only; it does not survive into the installed module, so downstream `include(CTest)` now gets upstream's `ON` default again.

Nothing compensates for that: `include/cmake.mk` never sets `BUILD_TESTING` — neither in the target [`CMAKE_OPTIONS` block](https://github.com/openwrt/openwrt/blob/1abb55fed63704842158567783141fae6fad55a6/include/cmake.mk#L130) nor in the [`Host/Configure/Default` block](https://github.com/openwrt/openwrt/blob/1abb55fed63704842158567783141fae6fad55a6/include/cmake.mk#L178), and only three in-tree packages pass it explicitly ([libevent2](https://github.com/openwrt/openwrt/blob/1abb55fed63704842158567783141fae6fad55a6/package/libs/libevent2/Makefile#L128), [tools/liblzo](https://github.com/openwrt/openwrt/blob/1abb55fed63704842158567783141fae6fad55a6/tools/liblzo/Makefile#L32), [tools/lzop](https://github.com/openwrt/openwrt/blob/1abb55fed63704842158567783141fae6fad55a6/tools/lzop/Makefile#L27)) — the rest have been relying on the patched default for the last six years. The risk is extra test targets (and their `find_package` dependencies) getting built for CMake-based packages, in this tree and much more so in the packages/luci feeds.

The timing comparison in the PR discussion only measures the `tools/` build, so it wouldn't have caught this.

If keeping the current behaviour is intended, adding `-DBUILD_TESTING=OFF` to the default option lists in `include/cmake.mk` (both target and host) would preserve it without the patch. If the behaviour change is deliberate, it'd be worth calling out in the commit message, since it affects far more than `tools/cmake`.

---

tools/cmake/Makefile
```diff
@@ -37,7 +37,8 @@ HOST_CONFIGURE_ARGS := \
 	--system-liblzma \
 	--system-zlib \
 	--system-zstd \
-	--generator=Ninja
+	--generator=Ninja \
+	-- -DBUILD_TESTING=OFF
```
openwrt-ai, line 41: nit: about the commit message rather than this line — the `Fixes: 8af80e79bd5b ("tools/cmake: update to 3.17.0")` trailer. That commit is verified to exist and the subject matches, but it didn't introduce a defect; it introduced `100-no-testing.patch` deliberately as the replacement for the older `100-disable_qt_tests.patch`. `Fixes:` normally marks a bug fix for the referenced commit and is used as a backport signal for stable branches, which isn't what this maintenance change wants. Plain prose ("replaces the patch added in 8af80e79bd5b") would carry the same information without the backport implication.

---

openwrt-ai (commented):
Re-reviewed at `1abb55f`. The head was force-pushed since the last review; the only change relative to `6e819ed` is a rebase onto the CMake 4.4.0 bump (`ac61bc0a9c5`), which re-generated the deleted patch's context lines. The functional change is unchanged.

The commit message accurately describes the diff, so no commit checks. One inline concern about scope of the behaviour change, plus one nit.

---

openwrt-ai (commented):
## Commit checks

- `b93d3812` "tools/cmake: pass -DBUILD_TESTING=OFF to bootstrap" — the subject scopes the change to `tools/cmake`, but half the diff is a new tree-wide default in `include/cmake.mk` that applies to every CMake package. Widen the subject (e.g. `build,tools/cmake: ...`) or split the `include/cmake.mk` hunks into their own `build:` commit.

---

tools/cmake/patches/100-no-testing.patch
openwrt-ai: fixed, thanks

---

tools/cmake/Makefile
openwrt-ai: fixed, thanks

---

BKPepe:
@robimarko Can this be merged?

robimarko:

PR #24400, opened 2026-07-24
realtek: add jumbo frame support

l0rdg3x:
Four commits, generic for all four targets as asked in
https://github.com/openwrt/openwrt/pull/24400#issuecomment-5461105408. The first adds the whole
infrastructure and changes no behaviour: both MTU operations, the field encoding and the config
entries, wired into the RTL83xx and RTL93xx `dsa_switch_ops`, with no family filling anything in yet.
The three that follow turn the families on one at a time, each adding only its register offset, its
largest frame and its config entries. There is one shared `port_change_mtu`; the families differ only
in which entry they fill, `mac_max_len_reg` for a per-port register or `mac_max_len_ctrl` for a
switch-wide one, plus `max_frame`.

The families split two and two, and the split is in kind, not in offset:

| | register | per port | notes |
|---|---|---|---|
| RTL930x | `MAC_L2_PORT_MAX_LEN_CTRL` 0x326c, stride 64 | yes, ports 0-28 | CPU port 28 is inside the array and the ethernet driver owns its row |
| RTL931x | `MAC_L2_PORT_MAX_LEN_CTRL` 0x5554, stride 4 | yes, ports 0-55 | CPU port 56 is outside it; 0x5634 is `MAC_DBG_SEL_CTRL` |
| RTL838x | `MAC_MAX_LEN_CTRL` 0xa9e0 + the duplicate at 0x6b00 | no, whole switch | the vendor SDK writes both |
| RTL839x | `MAC_MAX_LEN_CTRL` 0x02b0 | no, whole switch | |

Where the register is per port it is programmed from that port's MTU. Where there is only one for the
whole switch it is programmed from the largest MTU of any user port, so raising one port raises the
limit for all of them, lowering it back only takes effect once no other port asks for more, and every
port keeps advertising the family maximum even though only the largest one effectively holds it.

Field layout is identical on all four: the length in [13:0] and [27:14], the tag-inclusion bit in
[28], which is left as found. Offsets and port ranges come from the reverse-engineered maps at
svanheule.net and from the GPL SDK register lists; the RTL931x stride follows from the next register
listed, `MAC_DBG_SEL_CTRL` at 0x5634 with nothing in between, 224 bytes for 56 ports.

## Hardware

Hasivo S1100W-8XGT-SE: RTL9303 SoC, 8x RTL8264 10G PHY, SerDes in USXGMII mode, kernel 6.18.44,
RAM-booted. Baseline is `main` at 912d8b0485 with nothing else applied.

## Numbers

[...]

PR #24406, opened 2026-07-24
realtek: reclaim completed TX buffers on RX poll

l0rdg3x:
### realtek: reclaim completed TX buffers on RX poll

Split out of #24400 per review. The conduit transmit path frees a TX skb only when its descriptor slot is reused, `RTETH_TX_RING_SIZE` transmissions later; until then the skb stays charged to the sending socket. Negligible with standard frames, but with jumbo MTUs a handful of completed transmissions exhausts the sender's budget: eight ~9k echo replies exceed the kernel ICMP socket limit (`2 * SKB_TRUESIZE(64 * 1024)`), after which every further reply of any size is dropped (`Icmp: OutErrors` grows, `OutMsgs` stalls) until unrelated transmissions recycle the slots. On the bench (RTL9303, Hasivo S1100W-8XGT-SE) this is a reproducible ICMP blackout after exactly 8 jumbo echo replies, healed deterministically by any 16 device-originated frames.

Design, addressing the review concerns on the first attempt:

- the driver-private ring info tracks the oldest pending slot and a pending count, so finding work costs one cached read per TX ring — while nothing is pending the poll takes no lock and never touches the uncached descriptor memory;
- the walk starts at the oldest pending slot and stops at the first descriptor the hardware still owns: uncached reads = completed slots + 1, not the whole ring;
- the TX queue lock is held only for that bounded walk;
- the transmit path reuses the same walk for the ring-wrap case instead of its old duplicated per-slot cleanup;
- stopped or watchdog-frozen queues are skipped, leaving `rteth_tx_timeout()` recovery untouched.

**Performance (RTL9303, iperf3 8 s × 3 runs, MTU 1500, median):**

| | master | with reclaim |
|---|---|---|
| TCP host→device | 170 Mbit/s | 173 Mbit/s (+1.8%) |
| TCP device→host | 115 Mbit/s | **128 Mbit/s (+11.3%)** |
| UDP 1400 device→host | 61.8 Mbit/s | 64.4 Mbit/s (+4.2%) |
| UDP 64 device→host | 3.01 Mbit/s | 3.18 Mbit/s (+5.6%) |
| ping RTT avg | 0.409 ms | 0.401 ms |

Device-sourced TCP improves because the socket's `sk_wmem` now drains as soon as a poll runs instead of up to 16 transmissions later, so the sender is throttled less. No path gets slower.

Functional check with the jumbo series from #24400 on top: the ICMP blackout is gone (40/40 sequential and 150/150 sustained ~9k echoes, was 8/40 and 2/150), byte-exact MTU enforcement unaffected, and jumbo iperf3 reaches 467/222 Mbit/s (host→device/device→host) at MTU 9000 versus 173/128 at 1500.

CC @plappermaul

### DIFF
--- a/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
+++ b/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
@@ -89,7 +89,8 @@ struct rteth_rx_info {
 };
 
 struct rteth_tx_info {
-	int			slot;
+	unsigned int		send_count;  /* skbs handed to the hardware */
+	unsigned int		clean_count; /* skbs released after completion */
 	struct sk_buff		*skb[RTETH_TX_RING_SIZE];
 };
 
@@ -657,6 +658,49 @@ static void rteth_free_tx_buffers(struct rteth_ctrl *ctrl)
 					 tx_info->skb[i]->len, DMA_TO_DEVICE);
 			rteth_free_skb(&tx_info->skb[i]);
 		}
+		tx_info->send_count = 0;
+		tx_info->clean_count = 0;
+	}
+}
+
+static void rteth_reclaim_tx_ring(struct rteth_ctrl *ctrl, int r)
+{
+	struct rteth_tx_info *tx_info = &ctrl->tx_info[r];
+
+	BUILD_BUG_ON(RTETH_TX_RING_SIZE & (RTETH_TX_RING_SIZE - 1));
+
+	while (tx_info->send_count != tx_info->clean_count) {
+		int i = tx_info->clean_count & (RTETH_TX_RING_SIZE - 1);
+
+		if (ctrl->tx_data[r].ring[i] & RING_OWN_HW)
+			break;
+
+		dma_unmap_single(&ctrl->pdev->dev, ctrl->tx_data[r].frag[i].dma,
+				 tx_info->skb[i]->len, DMA_TO_DEVICE);
+		dev_consume_skb_any(tx_info->skb[i]);
+		tx_info->skb[i] = NULL;
+		tx_info->clean_count++;
+	}
+}
+
+static void rteth_reclaim_tx_rings(struct rteth_ctrl *ctrl)
+{
+	for (int r = 0; r < RTETH_TX_RINGS; r++) {
+		struct netdev_queue *txq;
+
+		/* Cached-memory fast path, made stable by the lock below */
+		if (READ_ONCE(ctrl->tx_info[r].send_count) ==
+		    READ_ONCE(ctrl->tx_info[r].clean_count))
+			continue;
+
+		txq = netdev_get_tx_queue(ctrl->dev, r);
+
+		__netif_tx_lock(txq, smp_processor_id());
+
+		if (!netif_xmit_frozen_or_stopped(txq))
+			rteth_reclaim_tx_ring(ctrl, r);
+
+		__netif_tx_unlock(txq);
 	}
 }
 
@@ -735,7 +779,8 @@ static int rteth_setup_ring_buffer(struct rteth_ctrl *ctrl)
 		}
 
 		ctrl->tx_data[r].ring[RTETH_TX_RING_SIZE - 1] |= RING_WRAP;
-		ctrl->tx_info[r].slot = 0;
+		ctrl->tx_info[r].send_count = 0;
+		ctrl->tx_info[r].clean_count = 0;
 	}
 
 	if (highmem)
@@ -1073,7 +1118,7 @@ static int rteth_start_xmit(struct sk_buff *skb, struct net_device *dev)
 		return NETDEV_TX_OK;
 	}
 
-	slot = ctrl->tx_info[ring].slot;
+	slot = ctrl->tx_info[ring].send_count & (RTETH_TX_RING_SIZE - 1);
 	frag = &ctrl->tx_data[ring].frag[slot];
 	packet_dma = ctrl->tx_data[ring].ring[slot];
 	packet_skb = &ctrl->tx_info[ring].skb[slot];
@@ -1086,11 +1131,8 @@ static int rteth_start_xmit(struct sk_buff *skb, struct net_device *dev)
 		return NETDEV_TX_BUSY;
 	}
 
-	if (likely(*packet_skb)) {
-		/* cleanup old data of this slot */
-		dma_unmap_single(&ctrl->pdev->dev, frag->dma, (*packet_skb)->len, DMA_TO_DEVICE);
-		dev_consume_skb_any(*packet_skb);
-	}
+	if (unlikely(*packet_skb))
+		rteth_reclaim_tx_ring(ctrl, ring);
 
 	*packet_skb = skb;
 	frag->len = len;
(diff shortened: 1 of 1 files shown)


target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -90,6 +90,8 @@ struct rteth_rx_info {
 
 struct rteth_tx_info {
 	int			slot;
+	int			dirty;   /* oldest slot with an unreleased skb */
+	unsigned int		pending; /* skbs not yet released */
```
plappermaul: slot/dirty/pending is partly redundant and currently need to cover edge cases. Use monotonic counters instead

- send_count: number of send/created SKBS/packets (instead of slot)
- clean_count: number of cleaned/reclaimed SKBS

With this: 

- Just increase counters with "++" (no wrap needed)
- int slot = tx_info->clean_count & (RTETH_TX_RING_SIZE - 1); (or send_coount)
- pending = tx_info->send_count - tx_info->clean_count;

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -1107,6 +1162,9 @@ static int rteth_start_xmit(struct sk_buff *skb, struct net_device *dev)
 	dma_wmb();
 	ctrl->tx_data[ring].ring[slot] = packet_dma | RING_OWN_HW;
 	ctrl->tx_info[ring].slot = (slot + 1) % RTETH_TX_RING_SIZE;
+	if (!ctrl->tx_info[ring].pending)
+		ctrl->tx_info[ring].dirty = slot;
```
plappermaul: With above idea unneeded edge case.

plappermaul (changes requested):
Much better implementation. Some nits left. 

Please attach a performance test for the new version too.

l0rdg3x:
Done — `slot`/`dirty`/`pending` replaced by monotonic `send_count`/`clean_count` (difference = outstanding skbs, low bits = ring slot, plus a `BUILD_BUG_ON` guarding the power-of-two assumption). The initialization edge case is gone, the transmit path is a single `send_count++`.

Performance re-run for this version, same protocol (iperf3 8 s × 3 runs, MTU 1500, two boots, median):

| | master | monotonic-counter reclaim |
|---|---|---|
| TCP host→device | 170 Mbit/s | 171 Mbit/s |
| TCP device→host | 115 Mbit/s | 119 Mbit/s (+3.5%) |
| UDP 1400 device→host | 61.8 Mbit/s | 64.0 Mbit/s (+3.6%) |
| UDP 64 device→host | 3.01 Mbit/s | 3.16 Mbit/s (+5%) |
| ping RTT avg | 0.409 ms | 0.396 ms |

No path regresses. For the record, the previous revision measured a couple of percent higher on device-sourced TCP (128 vs 119) across two boots each — the logic is equivalent and lighter now, so this smells like structure-layout/alignment noise rather than anything algorithmic; I kept your version since it is the cleaner one.

Functional check with the #24400 series on top unchanged: 40/40 sequential and 150/150 sustained ~9k echoes, byte-exact enforcement intact.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -657,6 +658,59 @@ static void rteth_free_tx_buffers(struct rteth_ctrl *ctrl)
...
+
+		dma_unmap_single(&ctrl->pdev->dev, ctrl->tx_data[r].frag[i].dma,
+				 tx_info->skb[i]->len, DMA_TO_DEVICE);
+		dev_consume_skb_any(tx_info->skb[i]);
+		tx_info->skb[i] = NULL;
+		tx_info->clean_count++;
+	}
+}
+
+static void rteth_reclaim_tx_buffers(struct rteth_ctrl *ctrl)
```
plappermaul: rename to rteth_reclaim_tx_rings() to align with above function.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -657,6 +658,59 @@ static void rteth_free_tx_buffers(struct rteth_ctrl *ctrl)
...
+	}
+}
+
+/* Release the contiguous run of transmitted skbs. Until released they stay
+ * charged to the sending socket, and with jumbo MTUs a handful of completed
+ * transmissions can exhaust the sender's budget, e.g. eight ~9k echo replies
+ * exceed the kernel ICMP socket limit and every further reply is dropped.
+ * The walk starts at the oldest unreleased slot and stops at the first
+ * descriptor the hardware still owns. The caller holds the TX queue lock.
+ */
```
plappermaul: This function is self explanatory. Drop this comment.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -657,6 +658,59 @@ static void rteth_free_tx_buffers(struct rteth_ctrl *ctrl)
...
+		    READ_ONCE(ctrl->tx_info[r].clean_count))
+			continue;
+
+		txq = netdev_get_tx_queue(ctrl->dev, r);
+
+		__netif_tx_lock(txq, smp_processor_id());
+
+		/* Leave stopped or watchdog-frozen queues to the recovery
+		 * path: rteth_tx_timeout() releases every buffer itself.
+		 */
```
plappermaul: Drop comment. Code tells the same.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -1086,11 +1141,11 @@ static int rteth_start_xmit(struct sk_buff *skb, struct net_device *dev)
...
 	}
 
-	if (likely(*packet_skb)) {
-		/* cleanup old data of this slot */
-		dma_unmap_single(&ctrl->pdev->dev, frag->dma, (*packet_skb)->len, DMA_TO_DEVICE);
-		dev_consume_skb_any(*packet_skb);
-	}
+	/* Ring wrapped before any reclaim ran: this slot's descriptor is no
+	 * longer hardware-owned, so the walk releases it (and any older run).
+	 */
```
plappermaul: Drop comment. Code below stands for itself.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -1297,6 +1352,11 @@ static int rteth_poll_rx(struct napi_struct *napi, int budget)
 	struct rteth_ctrl *ctrl = rx_q->ctrl;
 	int work_done, ring = rx_q->id;
 
+	/* Ahead of the receive pass, so replies generated while handling it
+	 * always find their socket send budget released.
+	 */
```
plappermaul: Drop comment.

plappermaul (changes requested):
One round of final cleanups and we are ready to go.

l0rdg3x:
Done, all five addressed.

plappermaul:

l0rdg3x:
You are welcome! :)

PR #24407, opened 2026-07-24
tools/cmake: remove no longer needed 150-zstd-libarchive.patch

BKPepe:
150-zstd-libarchive.patch was originally introduced in commit 0e3d51ccaedf ("tools/cmake: fix compilation with host libzstd-dev") to enable pkg-config searching for zstd in cmlibarchive.

In commit 1874a7cfb058 ("tools/cmake: update to 3.27.1"), upstream CMake added native FIND_PATH and FIND_LIBRARY logic for zstd into Utilities/cmlibarchive/CMakeLists.txt, leaving the patch rebased down to merely changing `IF(0)` to `IF(1)`.

Since CMake now natively searches for zstd using FIND_PATH/FIND_LIBRARY and correctly locates libzstd in STAGING_DIR_HOST without needing pkg-config, this patch is no longer necessary and can be safely removed.

Fixes: 1874a7cfb058 ("tools/cmake: update to 3.27.1")

Both commits are from @neheb, so... pinging him here. :)


neheb:
NACK. I would first verify ldd staging_dir/host/bin/cmake does not have libzstd.so from the OS.

BKPepe:
Since CMake upstream added `FIND_PATH` and `FIND_LIBRARY` natively for zstd back in CMake 3.27+, this patch was rebased over time down to just changing `IF(0)` to `IF(1)`. I verified it on macOS and it just works.

As the original author of the patch, feel free to double-check `ldd staging_dir/host/bin/zstd` oon a GNU/Linux host if you'd like, but it seems `150-zstd-libarchive.patch` is no longer necessary. :-)

neheb:
Yeah it works because it’s picking up OS libraries

neheb:
Yep. Removing the patch causes

```
	libzstd.so.1 => /usr/lib/libzstd.so.1 (0x00007f4c0983e000)
```

in CMake. Which is a header/library mismatch.

BKPepe:
Hmm... on macOS it works fine, otherwise I wouldn't submit this.  In any case, I updated the commit to pass explicit staging library and header paths directly in `tools/cmake/Makefile`, which should help.

If you'd like, feel free to test the updated commit locally on GNU/Linux. Otherwise, I'll test it on a GNU/Linux machine over the weekend when I get to it.

neheb:
It seems libbzip2 is used by cmake but --system-bzip2 is not passed.

ping @mcprat

mcprat:
maybe the new config lines should be added to default cmake configuration in cmake.mk?

tools/cmake/Makefile
```diff
@@ -37,7 +37,9 @@ HOST_CONFIGURE_ARGS := \
 	--system-liblzma \
 	--system-zlib \
 	--system-zstd \
-	--generator=Ninja
+	--generator=Ninja \
+	-- -DZSTD_INCLUDE_DIR="$(STAGING_DIR_HOST)/include" \
```
openwrt-ai: nit: The PR title ("remove no longer needed 150-zstd-libarchive.patch") and body still describe the original approach of simply deleting the patch as unnecessary. After the review discussion, the commit changed to keep an equivalent mechanism by passing explicit `-DZSTD_INCLUDE_DIR`/`-DZSTD_LIBRARY` here (the patch is replaced, not merely dropped). The title/body should be updated to match so the merge summary is accurate. Also note the commit's `Fixes:` trailer (`0e3d51ccaedf`) differs from the one in the PR body (`1874a7cfb058`).

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message accurately matches the diff, and the change looks functionally sound: `tools/zstd` installs both `libzstd.a` (via `install-static`) and its headers into `$(STAGING_DIR_HOST)`, and `ZSTD_INCLUDE_DIR`/`ZSTD_LIBRARY` are the exact cache variables cmlibarchive's find logic consumes, so pre-setting them pins the staging static lib as intended. One nit inline about the now-stale PR title/body.

---

mcprat:
> I updated the commit to pass explicit staging library and header paths directly in `tools/cmake/Makefile`, which should help.

it looks like there was also another way even before new cmake options dropped

`FindZstd.cmake:38:  message(WARNING "Zstd_INCLUDE_DIR is deprecated, use ZSTD_INCLUDE_DIR instead.")
`

mcprat:
@BKPepe I can verify that this works now as it is

but I think in order to keep the makefile looking pretty,
this may be a cleaner diff

@neheb side note: we must keep `--system-zstd` because otherwise it will build it's own

```
diff --git a/tools/cmake/Makefile b/tools/cmake/Makefile
index 45f7df13d2..23ecfb5a89 100644
--- a/tools/cmake/Makefile
+++ b/tools/cmake/Makefile
@@ -39,6 +39,14 @@ HOST_CONFIGURE_ARGS := \
        --system-zstd \
        --generator=Ninja

+CMAKE_HOST_OPTIONS = \
+       -DZSTD_INCLUDE_DIR="$(STAGING_DIR_HOST)/include" \
+       -DZSTD_LIBRARY="$(STAGING_DIR_HOST)/lib/libzstd.a"
+
+define Host/Configure
+       $(call Host/Configure/Default,-- $(CMAKE_HOST_OPTIONS))
+endef
+
 define Host/Compile/Default
        +$(NINJA) -C $(HOST_BUILD_DIR) $(1)
 endef

```

mcprat:
> It seems libbzip2 is used by cmake but --system-bzip2 is not passed.

after this we can do exactly the same thing for bzip2 (except variable names differ)
however, right now we use cmake to build bzip2 so that dependency would have to be reversed first.

BKPepe:
> but I think in order to keep the makefile looking pretty,
> this may be a cleaner diff

Pushed your suggestion, thanks! :)

openwrt-ai (commented):
Reviewed 1 new commit (force-push `5b6aeb0` → `0d7afbc`). The code delta is a pure refactor of the same two `-D` options out of `HOST_CONFIGURE_ARGS` into `CMAKE_HOST_OPTIONS` + a `Host/Configure` override, per the suggestion in the discussion. That is functionally equivalent: [`Host/Configure/Default` in include/host-build.mk](https://github.com/openwrt/openwrt/blob/0e178e15429c0ca4574b32ccc1f029d69a6ca185/include/host-build.mk#L118-L129) appends `$(1)` after `$(HOST_CONFIGURE_ARGS)`, so the resulting bootstrap command line is byte-identical to the previous revision, and the redefinition lands after `include $(INCLUDE_DIR)/host-build.mk` so it correctly overrides the default `Host/Configure`. No leftover references to the deleted `150-zstd-libarchive.patch` remain in the tree. No new inline issues.

## Commit checks

- `0d7afbc8` "tools/cmake: pass zstd paths to bootstrap" — the rewritten message lost the paragraph that stated the actual failure mode, and the paragraph that replaced it describes a code path this build does not take.

  The previous revision (`5b6aeb0`) carried:

  > On Linux host systems with libzstd-dev installed, CMake's FIND_LIBRARY otherwise defaults to selecting the OS dynamic library (/usr/lib/libzstd.so.1) over OpenWrt's staging static library ($(STAGING_DIR_HOST)/lib/libzstd.a), leading to a header/library version mismatch.

  That is the concrete regression demonstrated in the thread and the only thing that justifies both the `Fixes:` trailer and the `PKG_RELEASE` bump; as written now the message reads as a cleanup with no stated motivation.

  The replacement sentence — "ZSTD_INCLUDE_DIR and ZSTD_LIBRARY are the cache variables CMake itself sets when --system-zstd is not used" — is true of `Source/Modules/CMakeBuildUtilities.cmake`, which does `set(ZSTD_INCLUDE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/Utilities/cmzstd")` / `set(ZSTD_LIBRARY cmzstd)` under `if(NOT CMAKE_USE_SYSTEM_ZSTD)`, but (a) those are plain variables, not cache entries, and (b) this Makefile *does* pass `--system-zstd`, so that branch never runs here. What actually consumes the pre-seeded values is `FIND_PATH(ZSTD_INCLUDE_DIR zstd.h ...)` / `FIND_LIBRARY(ZSTD_LIBRARY NAMES zstd libzstd ...)` in `Utilities/cmlibarchive/CMakeLists.txt` — they are skipped because `-D` already populated the cache entries. Rewording along those lines (plus restoring the failure-mode paragraph) would make the message match what the change does.

---

peterwillcn (approved):

tools/cmake/Makefile
```diff
@@ -39,6 +39,14 @@ HOST_CONFIGURE_ARGS := \
 	--system-zstd \
 	--generator=Ninja
 
+CMAKE_HOST_OPTIONS = \
+	-DZSTD_INCLUDE_DIR="$(STAGING_DIR_HOST)/include" \
```
openwrt-ai, line 43: nit: re-anchoring [an earlier thread](https://github.com/openwrt/openwrt/pull/24407#discussion_r3649395362) that the rebase marked *outdated* — the underlying issue is unchanged at this head.

The PR title ("remove no longer needed 150-zstd-libarchive.patch") and body still describe the *original* approach. The body specifically claims:

> Since CMake now natively searches for zstd using FIND_PATH/FIND_LIBRARY and correctly locates libzstd in STAGING_DIR_HOST without needing pkg-config, this patch is no longer necessary and can be safely removed.

That was disproven in the discussion — without a replacement mechanism CMake links `/usr/lib/libzstd.so.1`, which is why these two `-D` options exist. The patch is *replaced*, not merely dropped.

Also, the `Fixes:` trailer disagrees between the two: the body says `1874a7cfb058 ("tools/cmake: update to 3.27.1")`, the commit says `0e3d51ccaedf ("tools/cmake: fix compilation with host libzstd-dev")`. Both SHAs exist in-tree, but only one can be the intended one, and the stale body text is what lands in the merge summary if this is squashed from the web UI.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push `0d7afbc` → `67f190a`). This is a **pure rebase** — I diffed both revisions' own patches and the only differences are the blob hashes, `PKG_VERSION` context (`4.3.4` → `4.4.0`, from [`ac61bc0a9c` "tools/cmake: update to 4.4.0"](https://github.com/openwrt/openwrt/commit/ac61bc0a9c) landing on main), and the deleted patch's hunk header shifting `@@ -680` → `@@ -682`. The commit message is byte-identical to the previous revision.

Re-checked what the rebase could have invalidated, all fine:

- [`Host/Configure/Default` in include/host-build.mk](https://github.com/openwrt/openwrt/blob/67f190ad3fc98f80c2b92d6882932f19e4a6d275/include/host-build.mk#L118-L129) still appends `$(1)` after `$(HOST_CONFIGURE_ARGS)`, so `-- $(CMAKE_HOST_OPTIONS)` still lands last on the bootstrap command line, and the redefinition is still after `include $(INCLUDE_DIR)/host-build.mk`.
- The deleted patch's own context (`SET(ZSTD_FIND_QUIETLY TRUE)` / `ENDIF (ZSTD_INCLUDE_DIR)` right above `IF(0)`) confirms cmlibarchive in 4.4.0 still keys off the `ZSTD_INCLUDE_DIR` cache entry, so pre-seeding it still short-circuits the `FIND_PATH`/`FIND_LIBRARY` as intended.
- `PKG_RELEASE:=2` is correct — main reset it to `1` on the 4.4.0 bump, and this is the first change on top.
- [`tools/Makefile:99`](https://github.com/openwrt/openwrt/blob/67f190ad3fc98f80c2b92d6882932f19e4a6d275/tools/Makefile#L99) still orders `zstd/compile` before `cmake/compile`, and [tools/zstd](https://github.com/openwrt/openwrt/blob/67f190ad3fc98f80c2b92d6882932f19e4a6d275/tools/zstd/Makefile#L36-L40) still runs `install-static install-includes` into `$(STAGING_DIR_HOST)`, so both hardcoded paths exist by the time cmake configures.
- No leftover references to the deleted `150-zstd-libarchive.patch` anywhere in the tree.

## Commit checks

- `67f190ad` "tools/cmake: pass zstd paths to bootstrap" — the message is unchanged from the previous revision, so [the concern I raised last round](https://github.com/openwrt/openwrt/pull/24407#pullrequestreview-4799487218) still stands and I won't restate it in full. In short: the message no longer states the actual failure mode (host `libzstd.so.1` winning over staging `libzstd.a`), which is the only thing justifying the `Fixes:` trailer and the `PKG_RELEASE` bump; and the sentence that replaced it describes the `if(NOT CMAKE_USE_SYSTEM_ZSTD)` branch in `Source/Modules/CMakeBuildUtilities.cmake`, which this build never takes because it passes `--system-zstd`. The consumer of the pre-seeded values is the `FIND_PATH`/`FIND_LIBRARY` pair in `Utilities/cmlibarchive/CMakeLists.txt`.

---

openwrt-ai (commented):
## Commit checks

- `b223728a` "tools/cmake: pass zstd paths to bootstrap" — message text is unchanged, so [the wording issue from the earlier round](https://github.com/openwrt/openwrt/pull/24407#pullrequestreview-4799487218) still stands: the failure mode (host `libzstd.so.1` winning over staging `libzstd.a`) is not stated, and the sentence about `--system-zstd` not being used describes a branch this build never takes.

---

BKPepe:
@mcprat , @neheb This can be merged as well, right?

mcprat:
yeah seems fine to me now as I said before

maybe ping someone who can do it
@PolynomialDivision @robimarko

robimarko:

PR #24418, opened 2026-07-25
Add support for ZyXEL GS1900-8HP B2

mundschenk-at:
This PR moves the common definitions of the ZyXEL GS1900-8HP variants to a separate include file and adds support for the B2 hardware variant.

### DIFF
--- a/target/linux/realtek/dts/rtl8380_zyxel_gs1900-8hp-b2.dts
+++ b/target/linux/realtek/dts/rtl8380_zyxel_gs1900-8hp-b2.dts
@@ -1,0 +1,13 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include "rtl8380_zyxel_gs1900-8hp.dtsi"
+
+/ {
+	compatible = "zyxel,gs1900-8hp-b2", "realtek,rtl838x-soc";
+	model = "Zyxel GS1900-8HP B2 Switch";
+};
+
+&pse {
+	compatible = "zyxel,gs1900-8hp-b2-pse", "realtek,pse-mcu-gen2";
+	current-speed = <115200>;
+};
--- a/target/linux/realtek/dts/rtl8380_zyxel_gs1900-8hp.dtsi
+++ b/target/linux/realtek/dts/rtl8380_zyxel_gs1900-8hp.dtsi
@@ -1,0 +1,34 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include "rtl8380_zyxel_gs1900.dtsi"
+#include "rtl8380_zyxel_gs1900_gpio.dtsi"
+
+&uart1 {
+	status = "okay";
+
+	/* PoE MCU; compatible and current-speed are set per variant */
+	pse: ethernet-pse {
+		pse-pis {
+			#address-cells = <1>;
+			#size-cells = <0>;
+
+			PSE_PI(0)
+			PSE_PI(1)
+			PSE_PI(2)
+			PSE_PI(3)
+			PSE_PI(4)
+			PSE_PI(5)
+			PSE_PI(6)
+			PSE_PI(7)
+		};
+	};
+};
+
+&phy8 { pses = <&pse_pi0>; };
+&phy9 { pses = <&pse_pi1>; };
+&phy10 { pses = <&pse_pi2>; };
+&phy11 { pses = <&pse_pi3>; };
+&phy12 { pses = <&pse_pi4>; };
+&phy13 { pses = <&pse_pi5>; };
+&phy14 { pses = <&pse_pi6>; };
+&phy15 { pses = <&pse_pi7>; };
--- a/package/boot/uboot-tools/uboot-envtools/files/realtek
+++ b/package/boot/uboot-tools/uboot-envtools/files/realtek
@@ -20,6 +20,7 @@ d-link,dgs-1210-28|\
 zyxel,gs1900-8-a1|\
 zyxel,gs1900-8hp-a1|\
 zyxel,gs1900-8hp-b1|\
+zyxel,gs1900-8hp-b2|\
 zyxel,gs1900-10hp-a1|\
 zyxel,gs1900-10hp-b1|\
 zyxel,gs1900-16-a1|\
--- a/target/linux/realtek/base-files/etc/board.d/02_network
+++ b/target/linux/realtek/base-files/etc/board.d/02_network
@@ -156,6 +156,7 @@ realtek_setup_macs()
 	zyxel,gs1900-8-b1|\
 	zyxel,gs1900-8hp-a1|\
 	zyxel,gs1900-8hp-b1|\
+	zyxel,gs1900-8hp-b2|\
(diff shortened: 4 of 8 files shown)


target/linux/realtek/dts/rtl8380_zyxel_gs1900-8hp.dtsi
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
```
openwrt-ai, line 1: This new file has no `compatible`/`model` of its own — it is an include consumed by the `-a1`/`-b1`/`-b2` `.dts` files — so by convention it should carry the `.dtsi` extension. Its direct analog [`rtl8380_zyxel_gs1900-10hp.dtsi`](https://github.com/openwrt/openwrt/blob/4ae1508c45239819f0d2a261cf1938a16789c798/target/linux/realtek/dts/rtl8380_zyxel_gs1900-10hp.dtsi) and every other include in this directory (`rtl8380_zyxel_gs1900.dtsi`, `rtl8380_zyxel_gs1900_gpio.dtsi`) use `.dtsi`. Rename this file to `rtl8380_zyxel_gs1900-8hp.dtsi` and update the three `#include "rtl8380_zyxel_gs1900-8hp.dts"` lines in the `-a1`, `-b1`, and `-b2` files to match.

---

openwrt-ai (commented):
Reviewed the 2 commits and the diff. The refactor faithfully extracts the shared GS1900-8HP definitions (the `-a1` and `-b1` files were identical apart from the per-variant `pse` compatible), and the B2 addition correctly mirrors the existing GS1900-10HP B1 pattern (`realtek,pse-mcu-gen2` + 115200 baud). One convention issue on the new shared include file is noted inline.

---

target/linux/realtek/dts/rtl8380_zyxel_gs1900-8hp.dtsi
mundschenk-at: The include file has been renamed.

target/linux/realtek/dts/rtl8380_zyxel_gs1900-8hp-b2.dts
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include "rtl8380_zyxel_gs1900-8hp.dtsi"
```
openwrt-ai, line 3: nit: pulling in the shared `.dtsi` asserts that B2 is identical to A1/B1 in everything except the PSE MCU — same RTL8231 GPIO expander wiring (`rtl8380_zyxel_gs1900_gpio.dtsi`), same SPI-NOR partition layout, same memory size from `rtl8380_zyxel_gs1900.dtsi`. Was that confirmed against actual B2 hardware, or inferred from the GS1900-10HP A1/B1 relationship?

The reason for asking: the most recent comparable addition in this target, [`66d356e` "realtek: add support for ZyXEL GS1900-24E B1"](https://github.com/openwrt/openwrt/commit/66d356edeae6535f1f5683254fc325c552cad8d0), found that a Zyxel revision bump was *not* a pure relabel — different PCB part number, DDR2 -> DDR3, different external PHY — and its commit message enumerates each verified difference with sources. If you've done the equivalent check on B2, a couple of lines to that effect in the commit body (PCB marking, flash chip, RAM) would make this much easier to accept; if it's untested inference from the 10HP pattern, that's worth saying explicitly too.

---

openwrt-ai (commented):
Re-reviewed at `fbba229`. The only code change since my last review is the `rtl8380_zyxel_gs1900-8hp.dts` -> `.dtsi` rename, which resolves the previous comment — and it also fixes a real build breakage, since the `#include "rtl8380_zyxel_gs1900-8hp.dtsi"` lines in the `-a1`/`-b1`/`-b2` files already pointed at the `.dtsi` name while the file on disk was still `.dts`.

The rest checks out: the extracted `.dtsi` is a faithful move (`-a1`/`-b1` were byte-identical apart from the per-variant `pse` `compatible`/`current-speed`), and it matches the in-tree `rtl8380_zyxel_gs1900-10hp.dtsi` pattern line for line, comment included. `zyxel,gs1900-8hp-b2` is registered in all five places its `-b1` sibling appears (`uboot-envtools/files/realtek`, `02_network`, `05_compat-version`, the DTS, `rtl838x.mk`), and the omission of `SUPPORTED_DEVICES` in the new `Device/zyxel_gs1900-8hp-b2` block is correct — `Device/zyxel_gs1900-10hp-b1` omits it too, as there's no legacy board name to migrate from. CI is green on this head.

One commit-message issue below, plus one hardware-confirmation question inline.

## Commit checks

- `fbba229` "realtek: add support for ZyXEL GS1900-8HP B2" — the message has no blank line between the subject and the body:

  ```
  realtek: add support for ZyXEL GS1900-8HP B2
  The GS1900-8HP B2 is a new hardware revision that uses the `realtek,pse-mcu-gen2`
  PSE dialect and 115200 baud like the GS1900-10HP B1.

  Signed-off-by: Peter Putzer <openwrt@mundschenk.at>
  ```

  Git therefore folds all three lines into a single 179-character subject, which is why the PR commit list and `git log --oneline` render it as one long run-on line. Insert an empty line after the subject. (`bc6b203` is formatted correctly and also picked up the lowercase-verb fix — thanks.)

---

target/linux/realtek/dts/rtl8380_zyxel_gs1900-8hp-b2.dts
mundschenk-at: Reworded.

openwrt-ai (commented):

The branch was rebased and both commits reworded; `git diff fbba229..3d6f386` is empty, so the tree is byte-identical to what I reviewed last time. Both previous findings are addressed:

- `3d6f386` now has a proper blank line after the subject, so the header is a 44-character subject instead of the previous 179-character run-on.
- The body now states explicitly that the non-PSE hardware differences are unknown and that the device runs on the pre-PSE `zyxel,gs1900-8hp-b1` image with `realtek-poe` — which is exactly the "say so if it's inference" case I asked about.

Both commit messages match their diffs (`1283ee0` is a pure move into `rtl8380_zyxel_gs1900-8hp.dtsi`; `3d6f386` adds only the B2 registration in the five expected places). `ZYXEL_VERS := AAHI` matches the `-a1`/`-b1` siblings, and the `SUPPORTED_DEVICES` omission remains correct for a revision with no legacy board name. CI is green on `3d6f386`.

---

jonasjelonek:
Could you please extend the commit message of the support addition commit a bit? I do not request a full-fledged message as usually needed for a completely new device, but shortly restating how installation works, etc. would be good.

mundschenk-at:
> Could you please extend the commit message of the support addition commit a bit? I do not request a full-fledged message as usually needed for a completely new device, but shortly restating how installation works, etc. would be good.

I've added the installation instructions from the original commit and the wiki.

jonasjelonek:

PR #24081, opened 2026-07-05
kernel: bump 6.12 to 6.12.98

DragonBluep:

PR #24419, opened 2026-07-25
kernel: bump 6.18 to 6.18.41

graysky2:
This PR updates 6.18 to 6.18.41 and as a result of some upstream changes `package/kernel/linux/modules/netfilter.mk`


EPinci:
Build system: x86/64
Build-tested: mediatek/filogic
Run-tested: mediatek/filogic (2x bpi-r4, gl-mt3000)

target/linux/airoha/patches-6.18/109-05-v6.19-pinctrl-airoha-add-support-for-Airoha-AN7583-PINs.patch
```diff
@@ -487,7 +487,7 @@ Signed-off-by: Linus Walleij <linus.walleij@linaro.org>
  /* PWM */
  #define AIROHA_PINCTRL_PWM(gpio, mux_val)		\
  	{						\
-@@ -1268,6 +1627,94 @@ static const struct airoha_pinctrl_func_
+@@ -1270,6 +1629,94 @@ static const struct airoha_pinctrl_func_
```
openwrt-ai, line 490: The airoha/an7581 kernel fails to compile on this head — the groups added by this hunk are what blow up:

```
drivers/pinctrl/mediatek/pinctrl-airoha.c:2142:9: error: initializer element is not constant
 2142 |         AIROHA_PINCTRL_PHY_LED1("gpio10", GPIO_LAN2_LED1_MODE_MASK,
      |         ^~~~~~~~~~~~~~~~~~~~~~~
drivers/pinctrl/mediatek/pinctrl-airoha.c:2142:9: note: (near initialization for 'an7583_phy4_led1_func_group[0].regmap[0].offset')
make[9]: *** [scripts/Makefile.build:287: drivers/pinctrl/mediatek/pinctrl-airoha.o] Error 1
```
([Build airoha/an7581](https://github.com/openwrt/openwrt/actions/runs/30165938853/job/89698955546))

`AIROHA_PINCTRL_PHY_LED0`/`AIROHA_PINCTRL_PHY_LED1` are defined by `109-02-v6.19-pinctrl-airoha-convert-PHY-LED-GPIO-to-macro.patch`, which this commit **deletes** — along with `109-01-v6.19-pinctrl-airoha-generalize-pins-group-function-confs-.patch` and `112-v6.19-pinctrl-airoha-fix-pinctrl-function-mismatch-issue.patch`. With no matching macro in scope GCC parses `AIROHA_PINCTRL_PHY_LED1(...)` as a function call, which is exactly the "initializer element is not constant" diagnostic above.

The commit message lists 109-01 and 109-02 under *Manually rebased*, not *Removed upstreamed*, so the deletion looks accidental. Either restore both patches rebased onto 6.18.40, or — if 6.18.40 really carries an updated variant of these macros — update the call sites here and in `202-pinctrl-airoha-permit-GPIO43-46-for-PHY-LED0.patch` to the new signature.

---

target/linux/airoha/patches-6.18/109-03-v6.19-pinctrl-airoha-convert-PWM-GPIO-to-macro.patch
```diff
@@ -17,7 +17,7 @@ Signed-off-by: Linus Walleij <linus.walleij@linaro.org>
 
 --- a/drivers/pinctrl/mediatek/pinctrl-airoha.c
 +++ b/drivers/pinctrl/mediatek/pinctrl-airoha.c
-@@ -1073,404 +1073,75 @@ static const struct airoha_pinctrl_func_
+@@ -1075,404 +1075,75 @@ static const struct airoha_pinctrl_func_
```
openwrt-ai, line 20: This hand-adjusted offset doesn't match what quilt generates — `Check Kernel patches (airoha, an7581)` fails:

```
target/linux/airoha/patches-6.18/109-03-v6.19-pinctrl-airoha-convert-PWM-GPIO-to-macro.patch
target/linux/airoha/patches-6.18/109-05-v6.19-pinctrl-airoha-add-support-for-Airoha-AN7583-PINs.patch
target/linux/airoha/patches-6.18/113-v6.19-pinctrl-airoha-Fix-AIROHA_PINCTRL_CONFS_DRIVE_E2.patch
target/linux/airoha/patches-6.18/202-pinctrl-airoha-permit-GPIO43-46-for-PHY-LED0.patch
Kernel patches for airoha/an7581 require refresh. (run 'make target/linux/refresh' and force push this pr)
```
([Check Kernel patches (airoha, an7581)](https://github.com/openwrt/openwrt/actions/runs/30165938853/job/89698961270))

Same applies to `109-05-...`, `113-v6.19-pinctrl-airoha-Fix-AIROHA_PINCTRL_CONFS_DRIVE_E2.patch` and `202-pinctrl-airoha-permit-GPIO43-46-for-PHY-LED0.patch`. After sorting out the 109-01/109-02 removal, please regenerate these with `make target/linux/refresh` (airoha/an7581 selected in `.config`) rather than editing the hunk headers by hand.

---

target/linux/bcm27xx/patches-6.18/0299-mmc-sdhci-of-dwcmshc-define-sdio-timeout-clocks.patch
```diff
@@ -129,21 +129,21 @@ Signed-off-by: Jonathan Bell <jonathan@raspberrypi.com>
...
  
  	sdhci_get_of_property(pdev);
 +	sdhci_enable_v4_mode(host);
  
  	priv->vendor_specific_area1 =
  		sdhci_readl(host, DWCMSHC_P_VENDOR_AREA1) & DWCMSHC_AREA1_MASK;
-@@ -1493,6 +1542,7 @@ err_rpm:
- 	pm_runtime_put_noidle(dev);
- err_clk:
+@@ -1528,6 +1577,7 @@ static void dwcmshc_remove(struct platfo
```
openwrt-ai: This hunk changed which function it patches during the manual rebase. Before, its context was the **probe error path** (`err_rpm:` / `err_clk:` … `return err;`); now it lands in `dwcmshc_remove()`. So `pltfm_host->timeout_clk` is no longer disabled when `dwcmshc_probe()` bails out after `clk_prepare_enable(pltfm_host->timeout_clk)` has already succeeded — a leaked clock enable ref on probe failure.

The relocation is easy to do by accident here because the two-line `clk_disable_unprepare(pltfm_host->clk); clk_disable_unprepare(priv->bus_clk);` sequence appears verbatim in both places — cf. [sdhci-of-dwcmshc.c:1491-1492 (probe error path)](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/mmc/host/sdhci-of-dwcmshc.c#L1491-L1492) and [:1522-1523 (`dwcmshc_remove`)](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/mmc/host/sdhci-of-dwcmshc.c#L1522-L1523) in v6.18.39, so `quilt` had two candidate matches.

Was moving it to `dwcmshc_remove()` deliberate, or should it stay on the probe error path (or be added in both)?

---

package/kernel/linux/modules/netfilter.mk
```diff
@@ -212,7 +212,7 @@ define KernelPackage/nf-flow
 	CONFIG_NETFILTER_INGRESS=y \
 	CONFIG_NF_FLOW_TABLE \
 	CONFIG_NF_FLOW_TABLE_HW
-  DEPENDS:=+kmod-nf-conntrack
+  DEPENDS:=+kmod-nf-conntrack +kmod-nft-core
```
openwrt-ai, line 215: Worth calling out the ripple: `kmod-ipt-offload` depends on `+kmod-nf-flow` ([netfilter.mk:332](https://github.com/openwrt/openwrt/blob/2c3b97924d68323b0e75d915fa0c8f8f462f2a82/package/kernel/linux/modules/netfilter.mk#L332)), so with this change every iptables-only flow-offload install now drags in `kmod-nft-core` and the whole nf_tables stack. That's unavoidable if `nf_flow_table.ko` really has undefined nf_tables symbols after 93d7a7ed0734 — but since the commit message infers the dependency from the refactor rather than from an observed symbol list, could you confirm it against `modinfo -F depends nf_flow_table.ko` (or the `.mod` deps) on a built 6.18.40 tree? If nf_flow_table.ko turns out to be self-contained, this pulls a fair bit of extra code into iptables-based images for nothing.

nit: commits touching `package/kernel/linux/modules/*.mk` are conventionally prefixed `kernel:` in this tree (e.g. `kernel: minimize kmod-sched-ctinfo dependencies`, `kernel: use fwnode for GPIO export driver`), so `kernel: add kmod-nft-core dependency to kmod-nf-flow` would fit better than `netfilter: ...`.

---

openwrt-ai (commented):
## Commit checks

- `6b3e637` "kernel: bump 6.18 to 6.18.40" — the *Removed upstreamed* / *Manually rebased* lists don't match what the commit actually does:
  - `airoha/patches-6.18/109-01-v6.19-pinctrl-airoha-generalize-pins-group-function-confs-.patch` and `109-02-v6.19-pinctrl-airoha-convert-PHY-LED-GPIO-to-macro.patch` are listed under **Manually rebased**, but both files are **deleted** by this commit. This is what breaks the airoha/an7581 build — see the inline comment on `109-05-...`.
  - Five further airoha patches are deleted with no mention anywhere in the message: `112-v6.19-pinctrl-airoha-fix-pinctrl-function-mismatch-issue.patch`, `170-v7.2-net-airoha-Fix-register-index-for-Tx-fwd-counter-con.patch`, `171-v7.2-net-airoha-Fix-debugfs-new-tuple-display-for-IPv4-RO.patch`, `172-v7.2-net-airoha-fix-foe_check_time-allocation-size.patch`, `174-v7.2-net-airoha-Fix-skb-priority-underflow-in-airoha_dev_.patch`. If they landed in 6.18.40 they belong under **Removed upstreamed** with stable links, like the three generic ones.
  - `generic/hack-6.18/901-debloat_sock_diag.patch` is listed under **Manually rebased**, but that file is untouched by this commit. The file that actually changed is `generic/hack-6.18/902-debloat_proc.patch` (one hunk-header offset, `@@ -4358` → `@@ -4353`). Looks like a digit typo.

Aside from those deletions, the diff is a clean offset-only refresh — only `bcm27xx/.../0299-mmc-sdhci-of-dwcmshc-define-sdio-timeout-clocks.patch`, `mediatek/.../942-net-ethernet-mtk_wed-move-cpuboot-in-a-dedicated-dts.patch`, `airoha/.../915-01-net-netfilter-flowtable-...`, `generic/pending-6.18/736-05-...`, `generic/backport-6.18/700-02-...sch_cake-Factor-out-config-variables...` and `airoha/.../161-01-...` carry content changes beyond hunk headers, and the mediatek/942 rebase looks correct.

---

graysky2:
I'm gonna need someone with a fresh pair of eyes to look at the airoha stuff...

DragonBluep:
Was patch `airoha/patches-6.18/109-02` dropped due to conflicts?

target/linux/airoha/patches-6.18/109-02-v6.19-pinctrl-airoha-convert-PHY-LED-GPIO-to-macro.patch
openwrt-ai, line 1: Re @DragonBluep's question ("Was patch `airoha/patches-6.18/109-02` dropped due to conflicts?") — evidence says yes, and that 109-01 and 109-02 need to be treated *differently*, not as one pair:

* v6.18.39 still carries the pre-generalization form, `#define PINCTRL_FUNC_DESC(id)` ([pinctrl-airoha.c:36](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/pinctrl/mediatek/pinctrl-airoha.c#L36)).
* This commit drops `112-v6.19-pinctrl-airoha-fix-pinctrl-function-mismatch-issue.patch` as upstreamed in 6.18.40. That patch is `Fixes: 4043b0c45f85 ("pinctrl: airoha: generalize pins/group/function/confs handling")` and its context is the **two-arg** `#define PINCTRL_FUNC_DESC(id, table)`. For it to be in 6.18.40, stable must also have taken 4043b0c45f85 — which is exactly what `109-01-...generalize-pins-group-function-confs-.patch` backports. So **109-01 is genuinely upstreamed** and belongs under *Removed upstreamed*, not *Manually rebased*.
* But `AIROHA_PINCTRL_PHY_LED0`/`AIROHA_PINCTRL_PHY_LED1` are still undefined in the built tree (the `initializer element is not constant` error on the [airoha/an7581 build](https://github.com/openwrt/openwrt/actions/runs/30195397647/job/89777634551) — GCC parsing the macro invocation as a function call). So the PHY-LED macro conversion did **not** land in 6.18.40, and **109-02 has to be restored**, rebased on top of the now-in-kernel generalization, before `109-05`, `113` and `202` can apply cleanly.

That also explains why the automatic rebase silently dropped it: with 4043b0c45f85 already applied by stable, 109-02's context no longer matched and `update_kernel.sh` discarded it.

---

openwrt-ai (commented):
Re-reviewed at `4a60d12`. The patch/tree content is byte-identical to the previously reviewed revision — the only changes are to the commit messages (and a rebase onto current main), so the four earlier inline threads still stand as-is and CI is still red on `airoha/an7581` for the same two reasons.

## Commit checks

- `4a60d12` "kernel: bump 6.18 to 6.18.40" — the *Manually rebased* list now says `linux/generic/hack-6.18/902-debloat_sock_diag.patch`, but no such file exists. The tree has `901-debloat_sock_diag.patch` (untouched by this commit) and `902-debloat_proc.patch` (the one actually changed here, `@@ -4358` → `@@ -4353` in the `net/core/sock.c` hunk). The previous revision said `901-debloat_sock_diag.patch`; the digit got corrected but the name didn't, so the entry is still wrong either way. It should read `902-debloat_proc.patch`.
- `4a60d12` "kernel: bump 6.18 to 6.18.40" — the five airoha patches are now correctly listed under *Removed upstreamed* with stable links, thanks. Still outstanding: `109-01-v6.19-pinctrl-airoha-generalize-pins-group-function-confs-.patch` and `109-02-v6.19-pinctrl-airoha-convert-PHY-LED-GPIO-to-macro.patch` are listed under *Manually rebased* but are **deleted** by this commit. See the new inline comment on `109-02` — the evidence points at 109-01 being genuinely upstreamed in 6.18.40 (so it belongs under *Removed upstreamed*) while 109-02 was silently dropped by the automatic rebase and needs restoring. That is what keeps the airoha/an7581 build red.

`6289ce5` "kernel: netfilter: add kmod-nft-core dependency to kmod-nf-flow" — message matches the change; the `kernel:` prefix now matches the tree convention for `package/kernel/linux/modules/*.mk`. The question in the existing thread about confirming the dependency against `modinfo -F depends nf_flow_table.ko` is still open.

---

graysky2:
> Was patch airoha/patches-6.18/109-02 dropped due to conflicts?

Yes, but this is a difficult rebase.

target/linux/airoha/patches-6.18/109-02-v6.19-pinctrl-airoha-convert-PHY-LED-GPIO-to-macro.patch
jonasjelonek: Are those your fresh pair of eyes? This sounds like a reasonable explanation to the issue.

OpenWRT-fanboy:
Can't be that difficult. Spit out a bunch of errors about LEDs, so I just dropped in 109-02 and it compiled fine.

Build-tested: airoha/an7581
Run-tested: airoha/an7581 (w1700k)

package/kernel/linux/modules/netfilter.mk
LGA1150: Commit 35fec487e30f05c81bd135326a993dad7f861812 switched iptables to use the nft core by default, so this only affects `iptables-zz-legacy` users.

OpenWRT-fanboy (approved):

target/linux/generic/kernel-6.18
```diff
@@ -1,2 +1,2 @@
-LINUX_VERSION-6.18 = .39
-LINUX_KERNEL_HASH-6.18.39 = a7a7e3d2ae9d95e74197223a8d4eb5f6be7aac21b6e6de27e9685d001c1f8cb0
+LINUX_VERSION-6.18 = .40
```
openwrt-ai: Two targets this PR doesn't touch directly are red on this head, both because 6.18 patches that landed on `main` after this branch was last regenerated were never re-run through `update_kernel.sh`.

**qualcommax/ipq50xx** — [Build qualcommax/ipq50xx](https://github.com/openwrt/openwrt/actions/runs/30300750695/job/90093485284):

```
Applying .../qualcommax/patches-6.18/0084-v7.2-clk-qcom-cmnpll-Account-for-reference-clock-divider.patch using plaintext:
patching file drivers/clk/qcom/ipq-cmn-pll.c
Hunk #1 FAILED at 199.
Hunk #2 FAILED at 207.
2 out of 2 hunks FAILED -- saving rejects to file drivers/clk/qcom/ipq-cmn-pll.c.rej
```

The preceding [`0083-v7.1-clk-qcom-ipq-cmn-pll-Add-IPQ8074-SoC-support.patch`](``https://github.com/openwrt/openwrt/blob/a4d267e81acb6e4e728658ab8742fe9923beee89/target/linux/qualcommax/patches-6.18/0083-v7.1-clk-qcom-ipq-cmn-pll-Add-IPQ8074-SoC-support.patch)`` applies in the same file with a `+7 lines` offset, so `drivers/clk/qcom/ipq-cmn-pll.c` did change between 6.18.39 and 6.18.40. Given [`0084`](``https://github.com/openwrt/openwrt/blob/a4d267e81acb6e4e728658ab8742fe9923beee89/target/linux/qualcommax/patches-6.18/0084-v7.2-clk-qcom-cmnpll-Account-for-reference-clock-divider.patch)`` carries `Fixes: f81715a4c87c`, the likely explanation is that it was picked into 6.18.40 and now needs to move to *Removed upstreamed* rather than be rebased. Worth confirming against the 6.18.40 tree before dropping.

**realtek/rtl838x** — [Check Kernel patches (realtek, rtl838x)](https://github.com/openwrt/openwrt/actions/runs/30300750695/job/90093376278):

```
target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
```

[`300-02-enhance-realtek-board-setup.patch`](``https://github.com/openwrt/openwrt/blob/a4d267e81acb6e4e728658ab8742fe9923beee89/target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch``) arrived with `7cc31af` ("realtek: convert to generic machine initialization") and only needs its offsets regenerated against 6.18.40 — `make target/linux/refresh` with realtek/rtl838x selected in `.config`.

---

openwrt-ai (commented):
Re-reviewed at `a4d267e`. The only tree change since `4a60d12` is the restoration of `target/linux/airoha/patches-6.18/109-02-v6.19-pinctrl-airoha-convert-PHY-LED-GPIO-to-macro.patch` (content byte-identical to the pre-bump version, only the hunk header moves `@@ -1473,516` → `@@ -1475,516`). That resolves the airoha/an7581 breakage: both `Build airoha/an7581` and `Check Kernel patches (airoha, an7581)` are green on this head, and my four earlier inline threads about 109-01/109-02, the hand-edited 109-03/109-05/113/202 offsets, and the airoha `Removed upstreamed` list are now addressed. The bcm27xx `0299` probe-error-path question and the `kmod-nf-flow` thread (where @LGA1150 noted 35fec487e30f limits the fallout to `iptables-zz-legacy`) are unchanged and still open.

Two targets are still red — see the inline comment on `target/linux/generic/kernel-6.18`.

## Commit checks

- `a4d267e` "kernel: bump 6.18 to 6.18.40" — *Manually rebased* still lists `linux/generic/hack-6.18/902-debloat_sock_diag.patch`; no such file exists. The tree has `901-debloat_sock_diag.patch` (untouched by this commit) and `902-debloat_proc.patch` — the latter is the file this commit actually changes (`@@ -4358,6 +4358,8` → `@@ -4353,6 +4353,8` in the `net/core/sock.c` hunk). The entry should read `902-debloat_proc.patch`.
- `a4d267e` "kernel: bump 6.18 to 6.18.40" — `FormalityCheck / Git & Commits` fails on this head: body line 6, `  linux/airoha/patches-6.18/109-01-v6.19-pinctrl-airoha-generalize-pins-group-function-confs-.patch[1]`, is 102 chars against the 100-char limit. (The nine reference URLs at the bottom are 130 chars each but the checker exempts them.)
- `a4d267e` "kernel: bump 6.18 to 6.18.40" — `109-02-...-convert-PHY-LED-GPIO-to-macro.patch` is modified by this commit but appears in neither list. Since the change is a pure hunk-header offset it arguably falls under "All other patches automatically rebased via update_kernel.sh", but given the history of this patch in this PR it would be worth an explicit note.

`7563db6` "kernel: netfilter: add kmod-nft-core dependency to kmod-nf-flow" — message matches the change.

---

openwrt-ai (commented):
Re-reviewed at `9777afc`. Tree changes since `a4d267e` are confined to two targets, and both are the fixes for what I flagged last round:

- **qualcommax/ipq50xx** — `0084-v7.2-clk-qcom-cmnpll-Account-for-reference-clock-divider.patch` is dropped and now listed under *Removed upstreamed* (link 10), and `0082`/`0083` get a `+7` hunk-header offset in `drivers/clk/qcom/ipq-cmn-pll.c`. That offset is exactly the net line count 0084 adds (`@@ -207,8 +207,15 @@`), which independently corroborates that its content landed in 6.18.40. `Build qualcommax/ipq50xx` is green.
- **realtek/rtl838x** — `300-02-enhance-realtek-board-setup.patch` regenerated (context/offsets only, blank-line placement moved with the refresh). `Check Kernel patches (realtek, rtl838x)` is green.

The airoha patches, `bcm27xx/.../0299`, `mediatek/.../942` and `netfilter.mk` are byte-identical to `a4d267e`, so the two still-open inline threads (the `0299` probe-error-path relocation, and the `kmod-nf-flow` dependency — where @LGA1150 noted 35fec487e30f limits the fallout to `iptables-zz-legacy`) carry over unchanged.

CI: everything green except `Build kirkwood/generic`, which died on `ERROR 503: Backend unavailable` while fetching `llvm-bpf-22.1.3.Linux-x86_64.tar.zst` from `downloads.cdn.openwrt.org` — unrelated to this PR, worth a re-run. `FormalityCheck / Git & Commits` now passes; dropping the `linux/` prefix from the patch paths brought the longest body line down to 97 chars.

## Commit checks

- `9777afc` "kernel: bump 6.18 to 6.18.40" — *Manually rebased* still lists `generic/hack-6.18/902-debloat_sock_diag.patch`; no file by that name exists in the tree. There is `901-debloat_sock_diag.patch` (untouched by this commit) and `902-debloat_proc.patch` — the latter is the file this commit actually changes, and only by an offset (`@@ -4358,6 +4358,8` → `@@ -4353,6 +4353,8` in the `net/core/sock.c` hunk). This has survived three revisions now; the entry should read `902-debloat_proc.patch` (or be dropped entirely, since a pure offset shift is covered by "All other patches automatically rebased via update_kernel.sh").
- `9777afc` "kernel: bump 6.18 to 6.18.40" — `airoha/patches-6.18/109-02-...-convert-PHY-LED-GPIO-to-macro.patch` is modified by this commit (hunk header `@@ -1473,516` → `@@ -1475,516`) but appears in neither list. Pure offset, so strictly it falls under the catch-all, but given this patch was dropped and restored during the PR it would be worth an explicit line so the next reader doesn't have to reconstruct the story from the comment thread.

`3576cdd` "kernel: netfilter: add kmod-nft-core dep to kmod-nf-flow" — message matches the change.

---

joshuacov1:
The automatic refresh of `bcm27xx/patches-6.18/0299-mmc-sdhci-of-dwcmshc-define-sdio-timeout-clocks.patch` is wrong (as described by claude.ai). Attached is the correctly rebased patch which addresses the issue raised by claude.ai.
[0299-mmc-sdhci-of-dwcmshc-define-sdio-timeout-clocks.patch](https://github.com/user-attachments/files/30481929/0299-mmc-sdhci-of-dwcmshc-define-sdio-timeout-clocks.patch)


I rebased it manually based on the commit message from https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=linux-6.18.y&id=c02237966c199df2303fe6808ee1735bb299f503 as well as the current code in https://github.com/raspberrypi/linux/blob/rpi-6.18.y/drivers/mmc/host/sdhci-of-dwcmshc.c

ttc0419:
6.18.41 is released

graysky2:
Yeah, gonna have to wait a few hours

target/linux/generic/kernel-6.18
```diff
@@ -1,2 +1,2 @@
-LINUX_VERSION-6.18 = .39
-LINUX_KERNEL_HASH-6.18.39 = a7a7e3d2ae9d95e74197223a8d4eb5f6be7aac21b6e6de27e9685d001c1f8cb0
+LINUX_VERSION-6.18 = .41
```
openwrt-ai, line 1: nit: the PR title still reads `kernel: bump 6.18 to 6.18.40`, but the head now lands on `.41` here and the PR body already says "This PR updates 6.18 to 6.18.41". Worth retitling to `6.18.41` so the PR subject matches its content.

(The hash is correct — `17fc72f0f8d4a8a8633a5d20085f5d9c5a5ec51ee896a0b7ae1ec25da31273ea` matches `linux-6.18.41.tar.xz` in kernel.org's `sha256sums.asc`.)

---

openwrt-ai (commented):
Re-reviewed at `cc5dfb5`. Since `9777afc` the branch was rebased onto current `main` and gained one new commit, `cc5dfb5` "kernel: bump 6.18 to 6.18.41".

The new commit touches only `target/linux/generic/kernel-6.18` and the hash checks out against kernel.org's `sha256sums.asc` (`17fc72f0…73ea` = `linux-6.18.41.tar.xz`), consistent with the "checksums only" claim. Nothing else in the tree references `6.18.40` any more.

The rebase absorbed the new airoha patches from `main` (`114`, `179`, `180-01`). The airoha side stays coherent: `180-01`'s file-deletion blob now carries `en7581_gpio32_pins` and `PINCTRL_CONF_DESC(32, …, BIT(19))`, which is exactly the effect of `112-…-fix-pinctrl-function-mismatch-issue.patch` now being upstream in 6.18.40 — so the removal of `112` and the blob change corroborate each other. `109-02` is back to a pure offset shift. The set of deleted files across the PR is now exactly the ten patches listed under *Removed upstreamed*, no more, no less. All 183 checks are green on this head, including every `Check Kernel patches` job and `FormalityCheck`.

Two carry-over items, unchanged by this push:

- `target/linux/bcm27xx/patches-6.18/0299-mmc-sdhci-of-dwcmshc-define-sdio-timeout-clocks.patch` is byte-identical to `9777afc`, so the open thread about the last hunk still stands — the `clk_disable_unprepare(pltfm_host->timeout_clk);` addition is anchored inside `dwcmshc_remove()` (`@@ -1528,6 +1577,7 @@ static void dwcmshc_remove(struct platfo`) rather than the `err_clk:` error path of `dwcmshc_probe()` where it originally lived. @joshuacov1 attached a hand-rebased replacement in [this comment](https://github.com/openwrt/openwrt/pull/24419#issuecomment-5110346389) on 28 Jul; it does not appear to have been picked up in either of the two pushes since.
- The `kmod-nf-flow` → `kmod-nft-core` thread (with @LGA1150's note that 35fec487e30f limits the fallout to `iptables-zz-legacy`) is also unchanged.

The `openwrt[bot]` "Package `package/kernel/linux` content changed without a PKG_RELEASE bump" warning is a false positive here — `package/kernel/linux/Makefile` has no `PKG_RELEASE` at all.

## Commit checks

- `16b44ec` "kernel: bump 6.18 to 6.18.40" — *Manually rebased* still lists `generic/hack-6.18/902-debloat_sock_diag.patch`. No file by that name exists in the tree: there is `901-debloat_sock_diag.patch` (untouched by this commit) and `902-debloat_proc.patch`, and the latter is the file this commit actually changes, by a single offset (`@@ -4358,6 +4358,8` → `@@ -4353,6 +4353,8` in the `net/core/sock.c` hunk). This has now survived four revisions — the entry should read `902-debloat_proc.patch`, or simply be dropped, since a pure offset shift is already covered by "All other patches automatically rebased via update_kernel.sh".

`cc5dfb5` "kernel: bump 6.18 to 6.18.41" and `ceebf7a` "kernel: netfilter: add kmod-nft-core dep to kmod-nf-flow" — messages match their changes.

---

target/linux/bcm27xx/patches-6.18/0299-mmc-sdhci-of-dwcmshc-define-sdio-timeout-clocks.patch
```diff
@@ -129,21 +129,21 @@ Signed-off-by: Jonathan Bell <jonathan@raspberrypi.com>
...
+ 	dwcmshc_disable_card_clk(host);
+ 
  	clk_disable_unprepare(pltfm_host->clk);
 +	clk_disable_unprepare(pltfm_host->timeout_clk);
  	clk_disable_unprepare(priv->bus_clk);
  	clk_bulk_disable_unprepare(priv->num_other_clks, priv->other_clks);
- 	return err;
+ }
 --- a/drivers/mmc/host/sdhci-pltfm.c
```
hauke, line 166: You are adding the `clk_disable_unprepare` in the `dwcmshc_remove` function now. it was in the error handling of `dwcmshc_probe` before. See this PR for details: https://github.com/openwrt/openwrt/pull/24081

target/linux/airoha/patches-6.18/180-01-v7.2-pinctrl-Move-Airoha-driver-to-dedicated-directory.patch
```diff
@@ -4963,7 +4965,7 @@ Signed-off-by: Linus Walleij <linusw@kernel.org>
 -	PINCTRL_CONF_DESC(29, REG_GPIO_L_PU, BIT(16)),
 -	PINCTRL_CONF_DESC(30, REG_GPIO_L_PU, BIT(17)),
 -	PINCTRL_CONF_DESC(31, REG_GPIO_L_PU, BIT(18)),
--	PINCTRL_CONF_DESC(32, REG_GPIO_L_PU, BIT(18)),
+-	PINCTRL_CONF_DESC(32, REG_GPIO_L_PU, BIT(19)),
```
hauke, line 4970: This is a file move, the same changes should show up in the add part too.

hauke: `180-02` and `180-04` are now upstream in 6.18.40

hauke:
Please remove `CONFIG_CRYPTO_DEV_SUN4I_SS_PRNG` from the target/linux/sunxi/config-6.18 file. it was removed upstream, see https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=linux-6.18.y&id=c401492e01c7bfd38cf14c94d85c7efafe7d1a25

The file name ` generic/hack-6.18/902-debloat_sock_diag.patch` you mention in the commit message does not exists.

graysky2:
@hauke - I have a family emergency. I'm not going to have time to work on this. Is there someone who can pick up modifying it so it can get merged?

hauke:
> @hauke - I have a family emergency. I'm not going to have time to work on this. Is there someone who can pick up modifying it so it can get merged?

No problem, your family is more important. I will take care of the PR.

graysky2:
> No problem, your family is more important. I will take care of the PR.

TY

target/linux/airoha/patches-6.18/109-03-v6.19-pinctrl-airoha-convert-PWM-GPIO-to-macro.patch
hauke: `make target/linux/refresh` runs without complains.

target/linux/airoha/patches-6.18/109-02-v6.19-pinctrl-airoha-convert-PHY-LED-GPIO-to-macro.patch
hauke: This is fixed now

target/linux/airoha/patches-6.18/180-01-v7.2-pinctrl-Move-Airoha-driver-to-dedicated-directory.patch
hauke: This is fixed now

target/linux/bcm27xx/patches-6.18/0299-mmc-sdhci-of-dwcmshc-define-sdio-timeout-clocks.patch
hauke: This is fixed now

target/linux/bcm27xx/patches-6.18/0299-mmc-sdhci-of-dwcmshc-define-sdio-timeout-clocks.patch
```diff
@@ -103,47 +103,66 @@ Signed-off-by: Jonathan Bell <jonathan@raspberrypi.com>
...
++	clk_disable_unprepare(pltfm_host->timeout_clk);
+ err_bus_clk:
+ 	clk_disable_unprepare(priv->bus_clk);
  err_clk:
+@@ -1529,6 +1580,7 @@ static void dwcmshc_remove(struct platfo
+ 
  	clk_disable_unprepare(pltfm_host->clk);
-+	clk_disable_unprepare(pltfm_host->timeout_clk);
  	clk_disable_unprepare(priv->bus_clk);
++	clk_disable_unprepare(pltfm_host->timeout_clk);
```
openwrt-ai, line 162: This disable balances only one of the two `clk_prepare_enable(pltfm_host->timeout_clk)` calls the patch installs in `dwcmshc_probe()`: one at line 113 of this file (inside the `if (dev->of_node)` block, after a second `devm_clk_get(dev, "timeout")`) and one at line 127 (unconditional). On an OF device that has a `timeout` clock both run, so the prepare/enable count reaches 2 and one reference is still outstanding after `dwcmshc_remove()` returns — likewise for the `err_timeout_clk:` path added at line 154.

The duplicated get/enable is pre-existing (the pre-bump version of this patch on `main` has the same two blocks, just with `return err;` instead of `goto err_bus_clk;`), so this is not a regression from the rebase. But the `dwcmshc_remove()` hunk itself is new — the previous version only disabled the clock on the probe error path — so: was the intent to balance both enables, or only the unconditional one?

Everything else in the replacement patch checks out: `err_timeout_clk:` sits between `err_rpm:` and `err_bus_clk:` so every `goto` reaching it unwinds in the right order, the `goto err_bus_clk` conversions plug the two `return err;` leaks the old version had, and the header diffstat (`62 ++…--`, `3 files changed, 68 insertions(+), 5 deletions(-)`) matches the actual hunks.

---

openwrt-ai (commented):
Re-reviewed at `155940cc`. Since `cc5dfb5` the branch was rebased onto current `main` and picked up three maintainer fixes, all folded into `090f178`; the `6.18.41` commit itself is unchanged apart from the rebase.

What I verified on the new content:

- **`180-01-…-Move-Airoha-driver-to-dedicated-directory.patch`** — the add side (`+++ b/drivers/pinctrl/airoha/pinctrl-airoha.c`, now `@@ -0,0 +1,3032 @@`) gained `en7581_gpio32_pins`, `PINCTRL_PIN_GROUP("gpio32", …)` and the four `PINCTRL_CONF_DESC(32, …, BIT(19))` corrections. I extracted both blobs and they are byte-identical (3032 lines each), so the patch is once again a pure file move. That also independently corroborates dropping `180-02` and `180-04`: the same content is present on the *delete* side, which quilt regenerates from the real kernel source, so it must now come from 6.18.40 itself. The an7583 counterparts `180-03`/`180-05` are correctly kept.
- **Deleted-file set** — exactly 12 files, matching the 12 entries under *Removed upstreamed* one-for-one, no more and no less.
- **`902-debloat_sock_diag.patch`** — the phantom *Manually rebased* entry that survived five revisions is gone. Resolved.
- **`CONFIG_CRYPTO_DEV_SUN4I_SS_PRNG`** — no remaining reference anywhere under `target/`.
- **`target/linux/generic/kernel-6.18`** — `.41` with hash `17fc72f0…73ea`; no `6.18.40` string left in the tree, and the PR title now matches the head commit.
- **CI** — all 182 check runs are green on this head, including every `Check Kernel patches` and `Build …` job and all three `FormalityCheck` jobs. Longest non-URL commit-message body line is 97 chars.

All three commit messages match their changes, so no commit-check items this round.

Carry-over: the `kmod-nf-flow` → `kmod-nft-core` thread on `package/kernel/linux/modules/netfilter.mk` is unchanged and remains the only open question (@LGA1150 noted 35fec487e30f limits the fallout to `iptables-zz-legacy`).

One new observation on the replacement `0299` patch — inline.

---

joshuacov1:
Regarding [target/linux/bcm27xx/patches-6.18/0299-mmc-sdhci-of-dwcmshc-define-sdio-timeout-clocks.patch](https://github.com/openwrt/openwrt/pull/24419/files#diff-9bc867ccfd9e47fba35deeaa49a83fc48122ecb9c13173e19b604bff8694d3a7)

This rebase now adds `clk_disable_unprepare(pltfm_host->timeout_clk);` to `dwcmshc_remove() `

This is also noted by claude: `But the dwcmshc_remove() hunk itself is new — the previous version only disabled the clock on the probe error path — so: was the intent to balance both enables, or only the unconditional one?`

The code in the current form is **neither** in the official linux repo **nor** in rpi-6.18.y under https://github.com/raspberrypi/linux.git

I think this should be fixed.

Issue #24323, opened 2026-07-20
Asus RT-AX89X |  OpenWRT   Snapshot  | Aquantia AQR113C 10G-ethernet port  | no longer working after upgrade


Razerwire:
### Describe the bug

Hi,

I'm having issues with my **Asus RT-AX89X**, or rather with it's built in 10gbit/s Ethernet-NIC/Port

The device has a **Aquantia AQR113C** which provides a 10Gbit/s  Ethernet-port/nic.
 I use it as a WAN port, in conjunction with OpenWRT.  

The default name for this port using OpenWRT is `10g-copper` 

DMESG output from working snapshot:
```
[    5.829580] Aquantia AQR113C gpio-0:08: attached PHY driver (mii_bus:phy_addr=gpio-0:08, irq=POLL)
[   24.966753] nss-dp 3a007000.dp6-syn 10g-copper: PHY Link up speed: 10000

```
Issues for me started after upgrading from  r35210-8614a2ba68 (dated 05-07-2026)   to  **Snapshot** **r35334-305c7599e5** (dated 13-07-2026) 

I've also tried (the) newer Snapshot(s)  and with that,  configuring the router from scratch (resetting it's config and starting from a clean slate) however 10g-copper is still not working.

Reverting to Snapshot r35210-8614a2ba68 and the 10Gbit/s Ethernet port is working fine again. 

### OpenWrt version

r35334-305c7599e5

### OpenWrt release

SNAPSHOT

### OpenWrt target/subtarget

qualcommax/ipq807x

### Device

Asus RT-AX89X

### Image kind

Official downloaded image

### Steps to reproduce

Setup RT-AX89X  using a working OpenWrt build (Stable-release  or Development-Snapshot no newer than   [r35210-8614a2ba68]). 

Setup and use  the built-in 10GbiIt/s  (AQR113C) Ethernet-port (I use it as  a WAN port,  however, I think  using it as a LAN-port will also suffice to reproduce the issue) it should be working fine. 

Now, upgrade Asus RT-AX89X to OpenWrt  Snapshot  **r35334-305c7599e5**  or newer. 

After upgrading to the new Snapshot,  observe that said  (built-in)  10Gbit/s Ethernet-Port is no longer operable.

### Actual behaviour

The 10g-copper link (again, provided via the AQR113C)  is now inoperable, it  is seen in firmware, but no link is actually established via this NIC, not even negotiation of said link, no LED's  on said port are blinking, nothing. 

[...]

PR #24420, opened 2026-07-25
qualcommax: fix 10G Aquantia PHY ports

JuliusBairaktaris:
Aquantia 10G ports on qualcommax link up and pass no traffic. Three faults.

**1. A modular PHY driver can never bind.** `qca-ppe` attaches every port's PHY from its own
probe, before the root filesystem is mounted, so a PHY driver shipped as a module always loses
the race and `genphy_c45` binds instead:

```
qca-ppe 3a000000.ppe 10g-1 (uninitialized): PHY [90000.mdio-1:08] driver [Generic Clause 45 PHY]
```

No vendor firmware load, no system interface setup. Patch 1 builds it in (`CRC_ITU_T` follows,
`kmod-phy-aquantia` goes with it). Patches 2–3 fix the 301w firmware nvmem cell, which has to
span exactly the image, and align two sibling nodes that already use the right offset.

**2. USXGMII autonegotiation ran at one end only.** The PCS ignored `neg_mode` and waited for
code words the PHY had just been told to stop sending. Patch 4 keys it on `neg_mode`; patch 6
adds the missing `managed = "in-band-status"` to all ten USXGMII ports across eight boards.
Neither works alone.

**3. The code word's link bit never asserts.** Behind an AQR113C it stays clear even with
autonegotiation complete, so once the port is in-band managed it gates carrier off for good.
Patch 5, from @rmandrad, takes the link from the 10GBASE-R receiver instead.

Also fixes the 10G connectivity reported in #24323.

**Testing.** Confirmed on 6.18. QNAP QHora-301w: @xlighting2017 on patches 1–2, @rmandrad on the
full series — 10g-1 reaches carrier, configures an address, and the AQR firmware still loads from
NVMEM with the driver built in. @Razerwire on an RT-AX89X: the AQR113C links at 10G and passes
traffic, tested overnight (#24323).

6.12 has no aquantia `inband_caps`/`config_inband`, so the PHY side is driven there by
`patches-6.12/0760-net-phy-aquantia-enable-USXGMII-MAC-autoneg.patch` instead. #24514 backports
the upstream USXGMII half and drops that hack.


larena1:
Nice find @JuliusBairaktaris but do you remove the Co-authored-by @claude in all the commits you push?

JuliusBairaktaris:
> Nice find @JuliusBairaktaris but do you remove the Co-authored-by @claude in all the commits you push?

Yeah — Claude Code assisted on these and I don't normally add a trailer, since it's still rare in practice. I see the tree does carry Assisted-by: on a handful of commits though, so happy to add it here if that's the expectation.

Razerwire:
Patches described here bring up the AQR113C in the  RT-AX89X properly. 

Physical-layer connectivity which was broken with at least r35334 is now working. 

(All other problems do remain. However I understand  that that is out-of-scope for this series of patches) 

Also see https://github.com/openwrt/openwrt/issues/24323#issuecomment-5080889282

target/linux/qualcommax/config-6.12
```diff
@@ -55,6 +55,7 @@ CONFIG_ARM_PSCI_CPUIDLE=y
 CONFIG_ARM_PSCI_FW=y
 # CONFIG_ARM_QCOM_CPUFREQ_HW is not set
 CONFIG_ARM_QCOM_CPUFREQ_NVMEM=y
+CONFIG_AQUANTIA_PHY=y
```
openwrt-ai: `CONFIG_AQUANTIA_PHY=y` is inserted at the wrong position for this file's sort order. Target configs are kept in `scripts/kconfig.pl` order (that's what [`scripts/kconfig-reorder.sh`](https://github.com/openwrt/openwrt/blob/2f60d245ee2c8edfb81ece3b44e04c5bdfb465c4/scripts/kconfig-reorder.sh#L1-L16) enforces), which sorts on the bare symbol name — so `AQUANTIA_PHY` sorts before `ARCH_*`/`ARM_*` and belongs right after [`CONFIG_64BIT=y` on line 1](https://github.com/openwrt/openwrt/blob/2f60d245ee2c8edfb81ece3b44e04c5bdfb465c4/target/linux/qualcommax/config-6.12#L1-L2), not after `CONFIG_ARM_QCOM_CPUFREQ_NVMEM`.

Running `LANG=C ./scripts/kconfig.pl '+' target/linux/qualcommax/config-6.12 /dev/null` on the branch moves exactly this one line to line 2 and touches nothing else. (`CONFIG_CRC_ITU_T=y` on line 108 is already in the right slot.)

---

openwrt-ai: Switching this back to built-in leaves the `kmod-phy-aquantia` references behind. [`target/linux/qualcommax/ipq807x/target.mk:3`](https://github.com/openwrt/openwrt/blob/2f60d245ee2c8edfb81ece3b44e04c5bdfb465c4/target/linux/qualcommax/ipq807x/target.mk#L3) still carries it in `DEFAULT_PACKAGES`, and [`target/linux/qualcommax/image/ipq60xx.mk:175`](https://github.com/openwrt/openwrt/blob/2f60d245ee2c8edfb81ece3b44e04c5bdfb465c4/target/linux/qualcommax/image/ipq60xx.mk#L175) still lists it in `linksys_mr7500`'s `DEVICE_PACKAGES`. The `DEFAULT_PACKAGES` entry was added by [8c3bcc1989a](https://github.com/openwrt/openwrt/commit/8c3bcc1989ac5fb596da86bfd82af6f8aafca1ab) in the very same diff that dropped `CONFIG_AQUANTIA_PHY=y`, so undoing one without the other looks unintended.

It won't break the build — `KernelPackage/install` in `include/kernel.mk` finds `aquantia.ko` in `modules.builtin` and just prints `NOTICE: module ... is built-in` — but `ModuleAutoLoad` runs unconditionally afterwards, so the now-empty package still ships `/etc/modules.d/18-phy-aquantia` (containing `aquantia`) plus the `/etc/modules-boot.d` symlink, and kmodloader will fail to find that module at boot. Dropping both references here would keep the commit self-contained, matching `AT803X_PHY=y` / `QCA807X_PHY=y`, which carry no kmod on this target.

---

target/linux/qualcommax/dts/ipq8072-301w.dts
```diff
@@ -341,8 +341,9 @@
 					#address-cells = <1>;
 					#size-cells = <1>;
 
-					aqr0_fw: firmware@0 {
-						reg = <0x0 0x5fc02>;
+					aqr0_fw: firmware@28 {
```
openwrt-ai: nit: renaming the node to `firmware@28` also fixes the unit-address/`reg` mismatch — but the two boards the commit message cites as the model still have it: [`ipq8074-nbg7815.dts:216`](https://github.com/openwrt/openwrt/blob/2f60d245ee2c8edfb81ece3b44e04c5bdfb465c4/target/linux/qualcommax/dts/ipq8074-nbg7815.dts#L216) is `aqr-fw@0` with `reg = <0x28 0x5f402>` and [`ipq8072-mx8500.dts:300`](https://github.com/openwrt/openwrt/blob/2f60d245ee2c8edfb81ece3b44e04c5bdfb465c4/target/linux/qualcommax/dts/ipq8072-mx8500.dts#L300) is `firmware@0` with `reg = <0x28 0x60002>`. Worth a third patch aligning those two, or deliberately out of scope for this series?

---

openwrt-ai (commented):
Reviewed 2 new commits (fresh review). Both commit messages match their diffs; the root-cause analysis checks out — `CONFIG_QCOM_80211AX_PPE=y` in both `ipq807x/config-default` and `ipq60xx/config-default` means the PPE driver really does probe before rootfs, `CONFIG_HWMON=y` is present in `config-6.12`, and `AQUANTIA_PHY` still `select`s `CRC_ITU_T` in 6.12. Comments inline on the config hunk and the DTS node rename.

---

JuliusBairaktaris:
> Patches described here bring up the AQR113C in the RT-AX89X properly.
> 
> Physical-layer connectivity which was broken with at least r35334 is now working.
> 
> (All other problems do remain. However I understand that that is out-of-scope for this series of patches)
> 
> Also see [#24323 (comment)](https://github.com/openwrt/openwrt/issues/24323#issuecomment-5080889282)

Its in scope now thanks to @rmandrad

target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c
```diff
@@ -614,9 +614,18 @@ static int qca_uniphy_pcs_config_usxgmii(struct phylink_pcs *pcs,
 	if (ret)
 		return ret;
 
+	/*
+	 * Only run USXGMII autonegotiation when phylink asked for in-band
+	 * signalling. Otherwise phylink has called phy_config_inband() with
+	 * LINK_INBAND_DISABLE, the PHY has stopped sending code words, and the
+	 * XPCS would wait for words that never arrive: the SerDes trains and
+	 * the receiver reaches block lock, but the datapath never opens.
+	 */
```
openwrt-ai: The mechanism this comment (and the commit message) describes doesn't exist on the kernel this target builds. `KERNEL_PATCHVER:=6.12`, and on 6.12 the Aquantia driver implements neither `.inband_caps` nor `.config_inband` — `aqr_gen2_config_inband()` is a v6.16+ function ([aquantia_main.c:1127 in v6.18](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/net/phy/aquantia/aquantia_main.c#L1127)), and there is no backport of it under `target/linux/generic/*-6.12/`. Consequences:

1. `phy_inband_caps()` [returns 0 when the driver has no `inband_caps` method](``https://github.com/openwrt/openwrt/blob/647cd30b0259ec64a7377d43ad2b195cd1a8192e/target/linux/generic/backport-6.12/601-04-v6.14-net-phy-add-phy_inband_caps.patch#L54-L61),`` so `pl->phy_ib_mode` stays 0 and phylink's [`if (pl->phydev && pl->phy_ib_mode)` guard](``https://github.com/openwrt/openwrt/blob/647cd30b0259ec64a7377d43ad2b195cd1a8192e/target/linux/generic/backport-6.12/601-12-v6.14-net-phylink-add-negotiation-of-in-band-capabilities.patch#L219-L221)`` means `phy_config_inband()` is never called on an AQR here — in either `managed` mode. What actually drives the PHY side is this target's own [`patches-6.12/0760-net-phy-aquantia-enable-USXGMII-MAC-autoneg.patch`](``https://github.com/openwrt/openwrt/blob/647cd30b0259ec64a7377d43ad2b195cd1a8192e/target/linux/qualcommax/patches-6.12/0760-net-phy-aquantia-enable-USXGMII-MAC-autoneg.patch#L26-L34),`` which unconditionally sets `MDIO_PHYXS_VEND_PROV2_USX_AUTONEG_MAC` from `aqr107_read_status()`.

2. Because `qca_uniphy_pcs_inband_caps()` returns `LINK_INBAND_DISABLE | LINK_INBAND_ENABLE` and the PHY reports nothing, patch 5's `managed = "in-band-status"` makes `phylink_pcs_neg_mode()` resolve to `PHYLINK_PCS_NEG_INBAND_ENABLED` for every USXGMII port in the target. So with both patches applied this `regmap_update_bits()` writes exactly the same `XPCS_MII_AN_EN` value it wrote before patch 4 — patch 4 changes no register on any in-tree board, and the PR body's "Neither works alone" doesn't hold in that direction (patch 4 alone indeed breaks things; patch 5 alone should be register-identical to 4+5).

The change itself is still the right hardening — it stops the XPCS from running Clause 37 when phylink asked for out-of-band. But could the comment and the commit message be reworded so they describe what happens on 6.12? As written, a future reader will look for an `aqr_gen2_config_inband()` that isn't in this tree, and the real fix for the 10G ports gets attributed to the wrong patch of the two.

---

target/linux/qualcommax/dts/ipq8074-nbg7815.dts
```diff
@@ -221,7 +221,7 @@
 					#address-cells = <1>;
 					#size-cells = <1>;
 
-					aqr_fw: aqr-fw@0 {
+					aqr_fw: aqr-fw@28 {
```
openwrt-ai: nit: while aligning these two boards, the node name is the remaining odd one out — this is `aqr-fw@28`, whereas [`ipq8072-mx8500.dts:308`](https://github.com/openwrt/openwrt/blob/647cd30b0259ec64a7377d43ad2b195cd1a8192e/target/linux/qualcommax/dts/ipq8072-mx8500.dts#L308) and the two cells patch 2 fixes in [`ipq8072-301w.dts:344`](https://github.com/openwrt/openwrt/blob/647cd30b0259ec64a7377d43ad2b195cd1a8192e/target/linux/qualcommax/dts/ipq8072-301w.dts#L344) / [`:361`](https://github.com/openwrt/openwrt/blob/647cd30b0259ec64a7377d43ad2b195cd1a8192e/target/linux/qualcommax/dts/ipq8072-301w.dts#L361) all use `firmware@28`. The label `aqr_fw` is what the PHY references, so the name is free to change:

```suggestion
					aqr_fw: firmware@28 {
```

Deliberately left alone, or worth folding into this patch?

---

openwrt-ai (commented):
Reviewed 3 new commits (`45d2185`, `6c86bde`, `647cd30`) plus the two amended ones. All five commit messages match their diffs, and the earlier config-ordering / `kmod-phy-aquantia` points are resolved — `CONFIG_AQUANTIA_PHY=y` now sits in `scripts/kconfig.pl` order at line 2, and no reference to the kmod is left in `target/linux/qualcommax/`. The DTS sweep is complete: all 10 `phy-mode = "usxgmii"` ports across 8 files now carry `managed = "in-band-status"`, matching the commit message's count. CI on `647cd30` is green.

One substantive question inline on the new comment block in `pcs-qca-uniphy.c` about the PHY-side mechanism on 6.12, plus one nit on the nbg7815 node name.

---

JuliusBairaktaris:
Merged @rmandrad's findings from the fork PR as patch 5, authored by him. It is the piece that was
missing here: with `managed = "in-band-status"` the code word's link bit becomes the carrier gate,
and that bit never asserts behind an AQR113C, so patches 4 and 6 needed it to be correct. Confirmed
by him on a 301w.

target/linux/qualcommax/config-6.12
```diff
@@ -1,4 +1,5 @@
 CONFIG_64BIT=y
+CONFIG_AQUANTIA_PHY=y
```
testuser7: Shouldn't this change go to `config-default` for `ipq807x` and `ipq60xx`?
And also for kernel 6.18.

target/linux/qualcommax/config-6.12
```diff
@@ -104,6 +105,7 @@ CONFIG_CPU_RMAP=y
 CONFIG_CPU_THERMAL=y
 CONFIG_CRC16=y
 CONFIG_CRC8=y
+CONFIG_CRC_ITU_T=y
```
testuser7: Same as above.

target/linux/qualcommax/config-6.12
rmandrad: yes agree - i suggested to have on these targets using config-default instead of all targets

target/linux/qualcommax/config-6.12
rmandrad: yes agree - i suggested to have on these targets using config-default instead of all targets

rmandrad:
> Merged @rmandrad's findings from the fork PR as patch 5, authored by him. It is the piece that was missing here: with `managed = "in-band-status"` the code word's link bit becomes the carrier gate, and that bit never asserts behind an AQR113C, so patches 4 and 6 needed it to be correct. Confirmed by him on a 301w.

note this has been tested with 6.18 (not 6.12)

target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c
```diff
@@ -355,17 +355,23 @@ static void qca_uniphy_pcs_get_state_usxgmii(struct qca_uniphy *uniphy,
...
 		return;
 	}
 
-	state->link = !!(val & XPCS_USXG_AN_LINK_STS);
+	state->link = !!(val & XPCS_KR_STS1_PLU);
 
 	if (!state->link)
 		return;
 
+	ret = regmap_read(uniphy->regmap, XPCS_MII_AN_INTR_STS, &val);
```
openwrt-ai, line 370: The speed decode below is now reached without any check that a code word was ever received. Previously `XPCS_USXG_AN_LINK_STS` gated it; now only `XPCS_KR_STS1_PLU` does, and the two are independent — the commit message says the link bit stays clear even after autonegotiation completes, so the converse (receiver up, no valid code word) is equally possible.

That matters because [`XPCS_USXG_AN_SPEED_10` is `0`](``https://github.com/openwrt/openwrt/blob/0e530dfd0df5d3a928a9c26faac49f6fbe118752/target/linux/qualcommax/files/include/linux/pcs/pcs-qca-uniphy.h#L94):`` a `XPCS_MII_AN_INTR_STS` that has not been updated by autonegotiation decodes as a valid `SPEED_10` rather than falling into the `default:` arm. With a PHY attached phylink overwrites speed from the PHY, so no in-tree board is affected today — but the commit message explicitly keeps the code word as "the speed source for a link with no PHY", and that is exactly the case where nothing else corrects a bogus 10 Mbit.

Would it be worth keeping the AN link bit as `state->an_complete` here (and using it to guard the speed decode), the way [`qca_uniphy_pcs_get_state_sgmii()` sets `state->an_complete = state->link;` at line 349](https://github.com/openwrt/openwrt/blob/0e530dfd0df5d3a928a9c26faac49f6fbe118752/target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c#L349) does? That preserves the "is the system interface up" answer the commit argues for while not trusting a speed field that never got written.

---

target/linux/qualcommax/dts/ipq8072-301w.dts
```diff
@@ -341,8 +341,8 @@
...
+						reg = <0x28 0x5f402>;
```
openwrt-ai: nit: the two boards this commit says it copies the offset from both explain the magic `0x28` inline — [`ipq8074-nbg7815.dts:225`](https://github.com/openwrt/openwrt/blob/0e530dfd0df5d3a928a9c26faac49f6fbe118752/target/linux/qualcommax/dts/ipq8074-nbg7815.dts#L225) and [`ipq8072-mx8500.dts:309`](https://github.com/openwrt/openwrt/blob/0e530dfd0df5d3a928a9c26faac49f6fbe118752/target/linux/qualcommax/dts/ipq8072-mx8500.dts#L309) both carry `/* Skip the QCOM MBN Header of 40 bytes */`. The 301w nodes had it in an earlier revision of this branch and lost it in the last force-push, so this is now the only one of the three without it.

```suggestion
					aqr0_fw: firmware@28 {
						/* Skip the QCOM MBN Header of 40 bytes */
						reg = <0x28 0x5f402>;
```

Same applies to `aqr1_fw` on lines 359-360 below.

---

target/linux/qualcommax/dts/ipq8072-301w.dts
```diff
@@ -506,6 +506,7 @@
 			label = "10g-1";
 			phy-handle = <&aqr113c_8>;
 			phy-mode = "usxgmii";
+			managed = "in-band-status";
```
openwrt-ai, line 501: The `Tested-by:` on this commit and on patch 5 refers to a 6.18 build — [@rmandrad noted that explicitly](https://github.com/openwrt/openwrt/pull/24420#issuecomment-5091343398) — but [`KERNEL_PATCHVER:=6.12`](https://github.com/openwrt/openwrt/blob/0e530dfd0df5d3a928a9c26faac49f6fbe118752/target/linux/qualcommax/Makefile#L11) is still what this target ships, and the PHY side of this change resolves differently on the two kernels:

- On 6.18, [`aqr_gen2_inband_caps()`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/net/phy/aquantia/aquantia_main.c#L1116-L1125) returns `LINK_INBAND_ENABLE | LINK_INBAND_DISABLE` for USXGMII, so `managed = "in-band-status"` makes phylink call `phy_config_inband(LINK_INBAND_ENABLE)` and the PHY is switched on explicitly — the mechanism the patch 4 commit message describes.
- On 6.12 that driver has no `inband_caps`/`config_inband`, so `phy_inband_caps()` returns 0, `pl->phy_ib_mode` stays 0 and `phy_config_inband()` is never reached. The PHY-side USX autoneg is instead forced unconditionally by this target's own [`patches-6.12/0760-net-phy-aquantia-enable-USXGMII-MAC-autoneg.patch`](``https://github.com/openwrt/openwrt/blob/0e530dfd0df5d3a928a9c26faac49f6fbe118752/target/linux/qualcommax/patches-6.12/0760-net-phy-aquantia-enable-USXGMII-MAC-autoneg.patch).``

The end state should be the same on both (PCS and PHY autoneg enabled), so I don't think this is broken on 6.12 — but it does mean the series is currently only hardware-confirmed on the testing kernel, via a different code path, while the default kernel path is untested. Has anyone run this on a 6.12 build, or is that worth calling out in the commit message?

---

openwrt-ai (commented):
Reviewed the rebase since `647cd30` (all five earlier commits reworded/amended, plus the new patch 5 `c55a5fda`). All six commit messages match their diffs, so no commit-check section.

The two points from the last round are addressed: `CONFIG_AQUANTIA_PHY=y` / `CONFIG_CRC_ITU_T=y` have moved from `config-6.12` to `ipq807x/config-default` and `ipq60xx/config-default`, both in `scripts/kconfig.pl` symbol order, and `ipq50xx` genuinely has no AQR board — so the placement also covers the 6.18 kernel that landed on main after this branch was cut. `CRC_ITU_T` still exists as a select-only tristate in both [6.12](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/lib/crc/Kconfig#L35) and 6.18, so nothing goes stale there. The DTS sweep is still complete: 10 `phy-mode = "usxgmii"` ports across 8 files, 10 `managed = "in-band-status"`, and every one of them has a `phy-handle` to an AQR — no SFP cage among them, as the commit message claims. CI on `0e530dfd` is green.

Three comments inline: one on the speed decode in the new `get_state_usxgmii()`, one question about which kernel the series was validated on, and one nit on a lost DTS comment.

---

Razerwire:
This set of (six) patches in this PR  solve the issues regarding  a lack of connectivity described in  #24323

Tested and confirmed on a RT-AX89X today (28-07-2026)
The 10G-PHY/Port (the Aquantia AQR113C) now works once more 
(Link comes up at - in my instance at "10gbps/Full -speed and RX/TX-traffic is passing normally.):
  ```
   []qca-ppe 3a000000.ppe 10g-copper (uninitialized): PHY [gpio-0:08] driver [Aquantia AQR113C] (irq=POLL)
   []qca-ppe 3a000000.ppe 10g-copper: qca-ppe-nss: port 6 armed for NSS fw attach   
   []qca-ppe 3a000000.ppe 10g-copper: qca-ppe-nss: NSS data plane override on port 6
   []qca-ppe 3a000000.ppe 10g-copper: configuring for inband/usxgmii link mode
   []qca-ppe 3a000000.ppe 10g-copper: qca-ppe-nss: port 6 on NSS fw data plane (vsi 1) 
   []qca-ppe 3a000000.ppe 10g-copper: Link is Up - 10Gbps/Full - flow control off
```


 
LGTM

(As an aside:  @JuliusBairaktaris  don't forget mention the co-authorship of @claude in the commits where it was used.)

**Review by Razerwire — APPROVED** [review]
This set of (six) patches in this PR solve the issues regarding a lack of connectivity described in https://github.com/openwrt/openwrt/issues/24323

**Comment by robimarko** [comment]
> > Merged @rmandrad's findings from the fork PR as patch 5, authored by him. It is the piece that was missing here: with `managed = "in-band-status"` the code word's link bit becomes the carrier gate, and that bit never asserts behind an AQR113C, so patches 4 and 6 needed it to be correct. Confirmed by him on a 301w.
> 
> note this has been tested with 6.18 (not 6.12)

This wont work as is on 6.12 as AQR inband support is not backported yet, I have the series to backport it somewhere.
It also allows dropping the hack that enables USXGMII MAC autoneg in 6.12.

**Comment by robimarko** [comment]
Also, has anybody tested that FW loading from NVMEM actually works now that the driver is built-in?
It should as I fixed the probe defferal in the firmware code some time ago but I have not tested whether it will correct deffer until the NVMEM device has been probed?

**Comment by rmandrad** [comment]
> Also, has anybody tested that FW loading from NVMEM actually works now that the driver is built-in? It should as I fixed the probe defferal in the firmware code some time ago but I have not tested whether it will correct deffer until the NVMEM device has been probed?

Hi @robimarko hope all is well
1. Yes the NVMEM load works on 6.18 - tested on the qnap 301w... note that the dts changes are bound to the firmware i can't guarantee other firmware versions
  `[    3.218957] Aquantia AQR113C 90000.mdio-1:00: loading firmware version 'v5.4.4 DNI NBG7815 112019 17:28:07' from 'NVMEM'
  [   17.148962] Aquantia AQR113C 90000.mdio-1:08: loading firmware version 'v5.4.4 DNI NBG7815 112019 17:28:07' from 'NVMEM'`
3.  on 6.12 absolutely agree with your backporting comments  .. i don't have 6.12 anymore running on the qnap 301w with 6.18 my other device is a banapi4 pro (mediatek) on a non openwrt with archlinux and 7.1.5 ... and also yes there is an issue with support for other openwrt versions that are not 6.18 ... so if you are to agree with the changes it needs to be only bound to snapshot

File: target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c [hunk 10]
```diff
@@ -355,17 +355,30 @@ static void qca_uniphy_pcs_get_state_usxgmii(struct qca_uniphy *uniphy,
...
+		state->link = 0;
+		return;
+	}
+
+	/* The speed field only means something once a code word has been
+	 * received; unwritten it decodes as a valid SPEED_10.
+	 */
+	state->an_complete = !!(val & XPCS_USXG_AN_LINK_STS);
+	if (!state->an_complete)
+		return;
```
openwrt-ai: This gate contradicts the premise of its own commit message. Two paragraphs up the message states that behind an AQR113C `XPCS_USXG_AN_LINK_STS` "stays clear the whole time" while "the code word tracks the copper rate as the media renegotiates" — i.e. the speed field is valid precisely when this bit is not. Gating on it makes the speed decode below unreachable on exactly the boards this series fixes, and the function then returns with `state->link = true` but `speed`/`duplex` left at the `SPEED_UNKNOWN`/`DUPLEX_UNKNOWN` that [`phylink_mac_pcs_get_state()` pre-seeds](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/drivers/net/phy/phylink.c#L1317-L1326).

Nothing downstream repairs that for USXGMII: phylink only substitutes the PHY's speed/duplex when the PHY does rate matching ([`if (pl->phy_state.rate_matching)` in phylink_resolve](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/drivers/net/phy/phylink.c#L1531-L1540)), and the Aquantia driver returns `RATE_MATCH_NONE` for USXGMII on both kernels — [6.12 `aqr107_get_rate_matching()`](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/drivers/net/phy/aquantia/aquantia_main.c#L612-L620), [6.18 `aqr_gen2_get_rate_matching()`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/net/phy/aquantia/aquantia_main.c#L1039-L1046). So `SPEED_UNKNOWN` reaches the link_up path:

- [`uniphy_link_up_usxgmii()` at pcs-qca-uniphy.c:757](``https://github.com/openwrt/openwrt/blob/da7d3c7e725ae7915fe131a4a376679ec661cda8/target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c#L745-L757``) bails with `Invalid USXGMII speed -1` / `PCS link up fail for interface usxgmii`, leaving the XPCS at its 10G reset default — the state patch 5 set out to escape.
- [`ppe_xgmac_link_up()` at qca_ppe_main.c:115](``https://github.com/openwrt/openwrt/blob/da7d3c7e725ae7915fe131a4a376679ec661cda8/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_main.c#L115-L116)`` hits `default: return`, so the MAC speed select is never programmed, and back in `qca_ppe_mac_link_up()` the USXGMII arm of the rate switch has no `default:` — [`unsigned long rate;` at :1019](``https://github.com/openwrt/openwrt/blob/da7d3c7e725ae7915fe131a4a376679ec661cda8/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_main.c#L1019)`` is then fed uninitialised to [`clk_set_rate()` at :1103](``https://github.com/openwrt/openwrt/blob/da7d3c7e725ae7915fe131a4a376679ec661cda8/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_main.c#L1102-L1105),`` before the MAC is enabled anyway.

This hunk arrived in the 2026-07-31 rebase, after the on-hardware confirmation on `0e530dfd`, so the `Tested-by:` no longer covers it. Keeping the bit as an informational `an_complete` without gating the decode restores the tested behaviour:

```suggestion
	state->an_complete = !!(val & XPCS_USXG_AN_LINK_STS);
```

That leaves the bogus-`SPEED_10` case I raised last round (`XPCS_USXG_AN_SPEED_10 == 0`, so an unwritten register decodes as valid 10 Mbit) unaddressed, but it only bites a PHY-less USXGMII link, of which this target has none — whereas the gate breaks every board in the series. If you would rather keep the gate, it needs `state->link = false` alongside the `return` so phylink never brings a link up with no speed, and that in turn means the 10G ports stay down on hardware where the bit really never asserts.

---

openwrt-ai (commented):
Reviewed the 2026-07-31 rebase since `0e530dfd`. Three hunks are new: the two `/* Skip the QCOM MBN Header of 40 bytes */` comments on the 301w cells and the `aqr-fw@28` → `firmware@28` rename on nbg7815 both address last round's nits and look right — the nodes are only referenced by label (`&aqr_fw`, `&aqr0_fw`, `&aqr1_fw`), so the rename is safe, and commit 3's reworded subject ("node names") now matches a diff that changes both the name and the unit address.

The third hunk, the new `an_complete` gate in `qca_uniphy_pcs_get_state_usxgmii()`, is the one to look at — it came in after the hardware confirmation and, as far as I can trace it, undoes the fix on the boards it was tested on. Detail inline.

## Commit checks

- d1c8def "qualcommax: pcs-qca-uniphy: take the USXGMII link from the receiver" — the message argues the gate is safe because "with a PHY attached phylink ANDs it with the PHY link and takes the interface and the media side speed from the PHY". Phylink takes the interface and pause from the PHY unconditionally, but speed/duplex only when the PHY reports rate matching, which the Aquantia driver does not do for USXGMII (`RATE_MATCH_NONE`). With the gate in place, speed for these ports therefore has no source at all — see the inline comment for the trace.

CI on `da7d3c7e` reports no completed checks yet (pending, nothing failed).

---

JuliusBairaktaris:
Confirmed — 6.12 has the phylink in-band framework via the `601-*` backports but no aquantia `inband_caps`/`config_inband`, so `phy_inband_caps()` returns 0, `pl->phy_ib_mode` stays 0 and `phy_config_inband()` is never reached on an AQR. What drives the PHY side there is `patches-6.12/0760-net-phy-aquantia-enable-USXGMII-MAC-autoneg.patch`, forcing the bit unconditionally from `aqr107_read_status()` — same end state as 6.18, different mechanism.

I put the USXGMII half of `5d59109d47c0` up as #24514, stacked on this PR, which drops 0760. Only that half, since upstream's SGMII arm needs the `priv->global_cfg[]` shadow added by the same v6.18 series. If your back port series is ready I will happily close it and rebase on yours instead.

larena1:
> (As an aside: @JuliusBairaktaris don't forget mention the co-authorship of @claude in the commits where it was used.)
 
Awesome @JuliusBairaktaris but could you still do what @Razerwire requested in https://github.com/openwrt/openwrt/pull/24420#issuecomment-5108599107 so this can get merged ASAP?

openwrt-ai (commented):
Reviewed the rebase since `da7d3c7e` — six commits, all amended. No new issues found, and all six messages still match their diffs, so no commit-check section.

Only two things changed in the series:

- Trailer order: `Assisted-by:` now sits before `Signed-off-by:` on patches 1, 2, 3, 4 and 6. That is the right way round.
- Patch 5 `db76162` drops the `an_complete` gate on the speed decode and rewords the paragraph that argued for it — the point from last round. `qca_uniphy_pcs_get_state_usxgmii()` is now exactly the `0e530dfd` revision that was confirmed on a 301w, plus one line: [`state->an_complete = !!(val & XPCS_USXG_AN_LINK_STS);` at pcs-qca-uniphy.c:376](``https://github.com/openwrt/openwrt/blob/0747d854f9277c100e46f83567eb9701b66c0c19/target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c#L376),`` with nothing branching on it. That is behaviourally free — phylink only reads `an_complete` to mask `BMSR_ANEGCOMPLETE` out of the emulated BMSR ([phylink.c:3282-3284](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/net/phy/phylink.c#L3282-L3284)), never to gate carrier or the speed/duplex it hands to `mac_link_up()` — and it matches what `get_state_sgmii()` and `get_state_hsgmii()` already do with the same field. So the on-hardware `Tested-by:` covers the code as it now stands.

I also re-checked patch 4 against the newer base this landed on: main has since made [`qca_uniphy_pcs_get_state()` take `neg_mode` on 6.18](``https://github.com/openwrt/openwrt/blob/0747d854f9277c100e46f83567eb9701b66c0c19/target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c#L426-L429``) and dropped [`pcs.neg_mode = true` there](``https://github.com/openwrt/openwrt/blob/0747d854f9277c100e46f83567eb9701b66c0c19/target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c#L991-L993).`` The `neg_mode == PHYLINK_PCS_NEG_INBAND_ENABLED` test patch 4 adds to `config_usxgmii()` is unaffected by that: `.pcs_config` is always called with the detailed `PHYLINK_PCS_NEG_*` value on 6.18, and `pcs.neg_mode = true` is still set on 6.12, so the comparison is meaningful on both kernels.

The DTS half is unchanged from the last round and still complete — 10 `phy-mode = "usxgmii"` ports across 8 files, 10 `managed = "in-band-status"`. CI on `0747d854` is green: all three qualcommax subtargets built, kernel-patch check and the formality checks passed.

---

JuliusBairaktaris:
@larena1 @Razerwire All six commits carry `Assisted-by: Claude:claude-opus-5` above the `Signed-off-by:`, pushed a few hours after that comment.

@robimarko The 6.12 back port is up as #24514 and drops `patches-6.12/0760-*` with it, so close mine if your series is further along.

Razerwire:
> @larena1 @Razerwire All six commits carry `Assisted-by: Claude:claude-opus-5` above the `Signed-off-by:`, pushed a few hours after that comment.
> 
> @robimarko The 6.12 back port is up as #24514 and drops `patches-6.12/0760-*` with it, so close mine if your series is further along.

Thanks!

And thanks for helping out, truly appreciate it . 

Still works fine with the commits  (_latest changes being July 31st 2026_) on the RT-AX89X, tested overnight, no issues.

sppmasterspp:
> Still works fine with the commits (_latest changes being July 31st 2026_) on the RT-AX89X, tested overnight, no issues.

Just wanted to ask if you used my image to verify it?

target/linux/qualcommax/dts/ipq8072-301w.dts
robimarko, line 1: Sorry, but this Commit is just AI crap that it thinks its found by comparing other 2 boards using AQR NVMEM loading.

On 301W there is no preflashed AQR FW in the storage, and the FW stored in the original OS is not MBN wrapped so there is no MBN header unless you added it yourself

robimarko: They can only carry the header if user flashed the FW with the MBN header which is additional step as the FW you extract from OS.

Like I said, there is no preflashed FW in the partitions

robimarko:

target/linux/qualcommax/dts/ipq8072-301w.dts
JuliusBairaktaris: The cell came off two 301Ws whose ethphyfw holds v5.4.4 DNI NBG7815 — the Zyxel image, not this board's AQR_CIG_WF-1945 — so what I measured is the geometry of whatever image is sitting in that partition, not the board's: 0x5f402 on theirs, 0x60002 for the AQR_ethphyfw_5.6.7.mbn the [community 301w guide](https://github.com/0x5826/QNAP-QHora-301w-Guide) tells owners to write there. Commit dropped.

@xlighting2017 can you confirm this?

JuliusBairaktaris:
@robimarko this is fine: https://github.com/openwrt/openwrt/commit/d6beac88143bd437631180a1f880c716c194bc0b, correct?

target/linux/qualcommax/dts/ipq8072-301w.dts
robimarko: Honestly, that makes no sense to suggest flashing other boards image as this one ships the correct LED config per AQR as they differ

robimarko:
> @robimarko this is fine: [d6beac8](https://github.com/openwrt/openwrt/commit/d6beac88143bd437631180a1f880c716c194bc0b), correct?

Yeah, thats correct

rmandrad:
just to note that the openwrt instructions are to flash the aquantia firmware from https://github.com/0x5826/Marvel-AQR-FW 

openwrt qnap301w - https://openwrt.org/toh/qnap/301w?s[]=open&s[]=howto&s[]=file

@JuliusBairaktaris i am assuming you will be merging this into your nss branch. Pls let me know. thank you

target/linux/qualcommax/dts/ipq8072-301w.dts
xlighting2017: > @xlighting2017 can you confirm this?

my 301w already flashed `v5.4.4 DNI NBG7815` in `ethphyfw`
so I can not confirm what a "vendor/original" board have in that partitions....

robimarko:
Ok, I have updated that not to link those at all, as FW container vendor provisioned register settings per AQR.
On top of the hand LED-s are incorrect if stock FW is unused

JuliusBairaktaris:
> just to note that the openwrt instructions are to flash the aquantia firmware from https://github.com/0x5826/Marvel-AQR-FW
> 
> openwrt qnap301w - [https://openwrt.org/toh/qnap/301w?s[]=open&s[]=howto&s[]=file](https://openwrt.org/toh/qnap/301w?s%5B%5D=open&s%5B%5D=howto&s%5B%5D=file)
> 
> @JuliusBairaktaris i am assuming you will be merging this into your nss branch. Pls let me know. thank you

already rebased my branches

PR #24421, opened 2026-07-25
realtek: eth: bound the CPU port packet length and follow the MTU

l0rdg3x:
Two commits. The first programs the switch registers that bound the frames exchanged with
the CPU port from the conduit MTU, where they used to keep whatever the bootloader left, and
turns the DMA truncation off so an oversized frame is dropped rather than delivered cut
short. The second adds a `change_mtu` operation that keeps those limits, and the RTL93xx
head of line budget, following the MTU at runtime.

Split out of #24400 as requested; that series will be rebased on top of this one.

Hardware: RTL9303 (Hasivo S1100W-8XGT-SE), RTL8264B PHY, USXGMII SerDes mode.
Baseline: current main.

| conduit MTU | 1504 (idle) | 1000 | 7174 |
|---|---|---|---|
| packet length programmed | 1530 | 1026 | 7200 |
| packets per ring | 32 | 42 | 7 |

The previous static sizing programmed 25 packets per ring at every MTU. A 1046 byte frame
passes at MTU 1504, is dropped at MTU 1000, whose limit of 1026 it exceeds, and passes again
once the MTU is back; a 102 byte one is answered five times out of five at every step. With
the interface down the hardware is left alone and the change is applied when it comes up.

Known limit: `dev->max_mtu` is the chip maximum of the family, while the operation refuses
anything above an MTU of 7174, which is what the linear part and the fragments of a single
skb hold at 400 bytes each. Raising `SKB_FRAG_SIZE` past its TODO lifts that towards the
hardware limit.

Second known limit: on an MTU decrease the head of line budget grows while the ring may still
hold frames of the previous size, so the ordering narrows that window rather than closing it;
every accepted change logs a warning for it. Open question in the comment below.

Not tested on RTL838x/839x/931x, where the register offsets come from the svanheule.net maps
and only the build is covered. Those boards program the new registers at every interface
open, not only when the MTU changes, so testing from their owners is welcome.

### DIFF
--- a/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
+++ b/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
@@ -38,11 +38,14 @@
 
 #define NOTIFY_EVENTS			10
 #define NOTIFY_BLOCKS			10
+
 #define RX_TRUNCATE_EN_93XX		BIT(6)
 #define RX_TRUNCATE_EN_83XX		BIT(4)
 #define TX_PAD_EN_838X			BIT(5)
 
 #define SKB_MTU				1600
+/* Ethernet header, two stacked VLAN tags (802.1ad QinQ) and FCS */
+#define RTETH_FRAME_OVERHEAD		(ETH_HLEN + 2 * VLAN_HLEN + ETH_FCS_LEN)
 /* TODO: change this to 1568 after fragment handling has been tested in the wild */
 #define SKB_FRAG_SIZE			400
 #define SKB_PAD				MAX(32, L1_CACHE_BYTES)
@@ -559,11 +562,63 @@ static void rteth_hw_ring_setup(struct rteth_ctrl *ctrl)
 			     offsetof(struct rteth_tx_data, ring));
 }
 
+/*
+ * Three limits bound the frames exchanged with the CPU port, and each family spreads them
+ * over registers of its own, whose length fields are named after the block holding them:
+ * what the driver receives is transmitted by the switch, but received by the NIC.
+ *
+ * - the DMA truncation length cuts a frame short. It still reaches the CPU, carrying only
+ *   as many bytes as the field allows, so switch it off and leave the length to the others.
+ * - the largest frame the switch hands to the CPU, which drops the longer ones. Both
+ *   measured on an RTL9303 by lowering one at a time while frames of growing size were
+ *   sent into a user port.
+ * - the largest frame the switch takes from the CPU. Longan keeps it in the port register
+ *   of the CPU port, Maple in a register of its own, the others in a second field of the
+ *   register above.
+ */
+static void rteth_838x_set_max_packet_length(struct rteth_ctrl *ctrl, int len)
+{
+	regmap_update_bits(ctrl->map, RTETH_838X_DMA_IF_PKT_RX_FLTR_CTRL, GENMASK(13, 0), len);
+	regmap_update_bits(ctrl->map, RTETH_838X_DMA_IF_PKT_TX_FLTR_CTRL, GENMASK(13, 0), len);
+	regmap_clear_bits(ctrl->map, ctrl->r->dma_if_ctrl, RX_TRUNCATE_EN_83XX);
+}
+
+static void rteth_839x_set_max_packet_length(struct rteth_ctrl *ctrl, int len)
+{
+	regmap_update_bits(ctrl->map, RTETH_839X_DMA_IF_PKT_FLTR_CTRL,
+			   GENMASK(27, 14) | GENMASK(13, 0),
+			   FIELD_PREP(GENMASK(27, 14), len) | FIELD_PREP(GENMASK(13, 0), len));
+	regmap_clear_bits(ctrl->map, ctrl->r->dma_if_ctrl, RX_TRUNCATE_EN_83XX);
+}
+
+static void rteth_930x_set_max_packet_length(struct rteth_ctrl *ctrl, int len)
+{
+	regmap_update_bits(ctrl->map, RTETH_930X_MAC_L2_CPU_MAX_LEN_CTRL, GENMASK(13, 0), len);
+	regmap_update_bits(ctrl->map, RTETH_930X_MAC_L2_PORT_MAX_LEN_CTRL,
+			   GENMASK(27, 14) | GENMASK(13, 0),
+			   FIELD_PREP(GENMASK(27, 14), len) | FIELD_PREP(GENMASK(13, 0), len));
+	regmap_clear_bits(ctrl->map, ctrl->r->dma_if_ctrl, RX_TRUNCATE_EN_93XX);
+}
+
+static void rteth_931x_set_max_packet_length(struct rteth_ctrl *ctrl, int len)
+{
+	regmap_update_bits(ctrl->map, RTETH_931X_MAC_L2_CPU_MAX_LEN_CTRL,
+			   GENMASK(27, 14) | GENMASK(13, 0),
+			   FIELD_PREP(GENMASK(27, 14), len) | FIELD_PREP(GENMASK(13, 0), len));
+	regmap_clear_bits(ctrl->map, ctrl->r->dma_if_ctrl, RX_TRUNCATE_EN_93XX);
+}
+
+static void rteth_set_max_packet_length(struct rteth_ctrl *ctrl)
+{
+	ctrl->r->set_max_packet_length(ctrl, ctrl->dev->mtu + RTETH_FRAME_OVERHEAD);
+}
+
 static void rteth_838x_hw_en_rxtx(struct rteth_ctrl *ctrl)
 {
-	/* Truncate RX buffer to SKB_MTU bytes, pad TX */
-	regmap_write(ctrl->map, ctrl->r->dma_if_ctrl,
-		     (SKB_MTU << 16) | RX_TRUNCATE_EN_83XX | TX_PAD_EN_838X);
+	/* Pad TX */
+	regmap_write(ctrl->map, ctrl->r->dma_if_ctrl, TX_PAD_EN_838X);
+
+	rteth_set_max_packet_length(ctrl);
 
 	rteth_enable_all_rx_irqs(ctrl);
 
@@ -586,8 +641,7 @@ static void rteth_838x_hw_en_rxtx(struct rteth_ctrl *ctrl)
 
 static void rteth_839x_hw_en_rxtx(struct rteth_ctrl *ctrl)
 {
-	/* Setup CPU-Port: RX Buffer */
-	regmap_write(ctrl->map, ctrl->r->dma_if_ctrl, (SKB_MTU << 5) | RX_TRUNCATE_EN_83XX);
+	rteth_set_max_packet_length(ctrl);
 
 	rteth_enable_all_rx_irqs(ctrl);
 
@@ -610,8 +664,7 @@ static void rteth_839x_hw_en_rxtx(struct rteth_ctrl *ctrl)
 
 static void rteth_930x_hw_en_rxtx(struct rteth_ctrl *ctrl)
 {
-	/* Setup CPU-Port: RX Buffer truncated at SKB_MTU Bytes */
-	regmap_write(ctrl->map, ctrl->r->dma_if_ctrl, (SKB_MTU << 16) | RX_TRUNCATE_EN_93XX);
+	rteth_set_max_packet_length(ctrl);
 
 	rteth_enable_all_rx_irqs(ctrl);
 
@@ -627,8 +680,7 @@ static void rteth_930x_hw_en_rxtx(struct rteth_ctrl *ctrl)
 
 static void rteth_931x_hw_en_rxtx(struct rteth_ctrl *ctrl)
 {
-	/* Setup CPU-Port: RX Buffer truncated at SKB_MTU Bytes */
-	regmap_write(ctrl->map, ctrl->r->dma_if_ctrl, (SKB_MTU << 16) | RX_TRUNCATE_EN_93XX);
+	rteth_set_max_packet_length(ctrl);
 
 	rteth_enable_all_rx_irqs(ctrl);
 
@@ -1600,6 +1652,7 @@ static const struct rteth_config rteth_838x_cfg = {
 	.hw_stop = &rteth_838x_hw_stop,
 	.hw_reset = &rteth_838x_hw_reset,
 	.init_mac = &rteth_838x_init_mac,
+	.set_max_packet_length = rteth_838x_set_max_packet_length,
 	.netdev_ops = &rteth_838x_netdev_ops,
 };
 
@@ -1645,6 +1698,7 @@ static const struct rteth_config rteth_839x_cfg = {
 	.hw_stop = &rteth_839x_hw_stop,
 	.hw_reset = &rteth_839x_hw_reset,
 	.init_mac = &rteth_839x_init_mac,
+	.set_max_packet_length = rteth_839x_set_max_packet_length,
 	.setup_notify_ring_buffer = &rteth_839x_setup_notify_ring_buffer,
 	.netdev_ops = &rteth_839x_netdev_ops,
 };
@@ -1692,6 +1746,7 @@ static const struct rteth_config rteth_930x_cfg = {
 	.hw_stop = &rteth_930x_hw_stop,
 	.hw_reset = &rteth_93xx_hw_reset,
 	.init_mac = &rteth_930x_init_mac,
+	.set_max_packet_length = rteth_930x_set_max_packet_length,
 	.netdev_ops = &rteth_930x_netdev_ops,
 };
 
@@ -1738,6 +1793,7 @@ static const struct rteth_config rteth_931x_cfg = {
 	.hw_stop = &rteth_931x_hw_stop,
 	.hw_reset = &rteth_93xx_hw_reset,
 	.init_mac = &rteth_931x_init_mac,
+	.set_max_packet_length = rteth_931x_set_max_packet_length,
 	.netdev_ops = &rteth_931x_netdev_ops,
 };
 
--- a/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
+++ b/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
@@ -43,9 +43,13 @@
 #define RX_TRUNCATE_EN_83XX		BIT(4)
 #define TX_PAD_EN_838X			BIT(5)
 
-#define SKB_MTU				1600
 /* Ethernet header, two stacked VLAN tags (802.1ad QinQ) and FCS */
 #define RTETH_FRAME_OVERHEAD		(ETH_HLEN + 2 * VLAN_HLEN + ETH_FCS_LEN)
+/* Largest frame each family switches, as its datasheet and DSA rmon range have it */
+#define RTETH_838X_MAX_FRAME		10000
+#define RTETH_839X_MAX_FRAME		12288
+#define RTETH_930X_MAX_FRAME		12288
+#define RTETH_931X_MAX_FRAME		12288
 /* TODO: change this to 1568 after fragment handling has been tested in the wild */
 #define SKB_FRAG_SIZE			400
 #define SKB_PAD				MAX(32, L1_CACHE_BYTES)
@@ -441,12 +445,15 @@ static void rteth_nic_reset(struct rteth_ctrl *ctrl, int reset_mask)
 	msleep(100);
 }
 
+static void rteth_83xx_set_hol(struct rteth_ctrl *ctrl)
+{
+	/* Free floating rings without space tracking */
+	regmap_write(ctrl->map, ctrl->r->dma_if_rx_ring_size, 0);
+}
+
 static void rteth_838x_hw_reset(struct rteth_ctrl *ctrl)
 {
 	rteth_nic_reset(ctrl, 0xc);
-
-	/* Free floating rings without space tracking */
-	regmap_write(ctrl->map, ctrl->r->dma_if_rx_ring_size, 0);
 }
 
 static void rteth_839x_hw_reset(struct rteth_ctrl *ctrl)
@@ -472,32 +479,36 @@ static void rteth_839x_hw_reset(struct rteth_ctrl *ctrl)
 	/* Restore notification settings: on RTL838x these bits are null */
 	regmap_update_bits(ctrl->map, ctrl->r->dma_if_intr_msk, 7 << 20, int_saved & (7 << 20));
 	regmap_write(ctrl->map, RTL839X_DMA_IF_NBUF_BASE_DESC_ADDR_CTRL, nbuf);
-
-	/* Free floating rings without space tracking */
-	regmap_write(ctrl->map, ctrl->r->dma_if_rx_ring_size, 0);
 }
 
-static void rteth_93xx_hw_reset(struct rteth_ctrl *ctrl)
+static void rteth_93xx_set_hol(struct rteth_ctrl *ctrl)
 {
 	/*
 	 * The counter registers track the number of packets that are allowed to be appended to
 	 * the ring buffer. On RTL93xx a ring overflow must be avoided at all cost. Defensively
 	 * calculate the space by anticipating that only fully filled packets are received.
 	 */
-	int max_frame_size = SKB_MTU + ETH_HLEN + ETH_FCS_LEN + VLAN_HLEN + 4;
+	int max_frame_size = ctrl->dev->mtu + RTETH_FRAME_OVERHEAD;
 	int frags_per_pkt = DIV_ROUND_UP(max_frame_size, SKB_FRAG_SIZE);
 	int cnt = min(RTETH_RX_RING_SIZE / frags_per_pkt, 0x3ff);
 
-	rteth_nic_reset(ctrl, 0x6);
-
-	/* Setup Head of Line */
 	for (int ring = 0; ring < RTETH_RX_RINGS; ring++) {
 		int shift = (ring % 3) * 10;
 		int reg = (ring / 3) * 4;
-		u32 v;
 
 		regmap_update_bits(ctrl->map, ctrl->r->dma_if_rx_ring_size + reg,
 				   0x3ff << shift, cnt << shift);
+	}
+}
+
+static void rteth_93xx_hw_reset(struct rteth_ctrl *ctrl)
+{
+	rteth_nic_reset(ctrl, 0x6);
+
+	for (int ring = 0; ring < RTETH_RX_RINGS; ring++) {
+		int reg = (ring / 3) * 4;
+		u32 v;
+
 		/* clear counters by simply writing the current register values back */
 		regmap_read(ctrl->map, ctrl->r->dma_if_rx_ring_cntr + reg, &v);
 		regmap_write(ctrl->map, ctrl->r->dma_if_rx_ring_cntr + reg, v);
@@ -912,6 +923,7 @@ static int rteth_open(struct net_device *dev)
 
 	scoped_guard(spinlock_irqsave, &ctrl->lock) {
 		ctrl->r->hw_reset(ctrl);
+		ctrl->r->set_hol(ctrl);
 		rteth_setup_cpu_rx_rings(ctrl);
 		ret = rteth_setup_ring_buffer(ctrl);
 		if (ret)
@@ -1022,6 +1034,41 @@ static int rteth_stop(struct net_device *dev)
 	rteth_free_tx_buffers(ctrl);
 	rteth_free_rx_buffers(ctrl);
 	netif_tx_stop_all_queues(dev);
+
+	return 0;
+}
+
+static int rteth_change_mtu(struct net_device *dev, int mtu)
+{
+	struct rteth_ctrl *ctrl = netdev_priv(dev);
+	bool grow = mtu > dev->mtu;
+
+	/* A frame of that size must still fit into the linear part and fragments of a single skb */
+	if (mtu + RTETH_FRAME_OVERHEAD > (MAX_SKB_FRAGS + 1) * SKB_FRAG_SIZE)
+		return -EINVAL;
+
+	WRITE_ONCE(dev->mtu, mtu);
+
+	/* Opening the interface programs both limits from the MTU, so leave the hardware alone */
+	if (!netif_running(dev))
+		return 0;
+
+	netdev_warn(dev, "changing the MTU may have unexpected effects under high receive load\n");
+
+	/*
+	 * A larger MTU buys a smaller budget, so hand the rings their new limit before the
+	 * switch may deliver larger frames, and the other way around when the MTU shrinks.
+	 * Hold the lock, so rteth_tx_timeout() cannot reprogram the same registers in between.
+	 */
+	scoped_guard(spinlock_irqsave, &ctrl->lock) {
+		if (grow)
+			ctrl->r->set_hol(ctrl);
+
+		rteth_set_max_packet_length(ctrl);
+
+		if (!grow)
+			ctrl->r->set_hol(ctrl);
+	}
 
 	return 0;
 }
@@ -1611,6 +1658,7 @@ static int rteth_setup_tc(struct net_device *dev, enum tc_setup_type type, void 
 static const struct net_device_ops rteth_838x_netdev_ops = {
 	.ndo_open = rteth_open,
 	.ndo_stop = rteth_stop,
+	.ndo_change_mtu = rteth_change_mtu,
 	.ndo_start_xmit = rteth_start_xmit,
 	.ndo_set_mac_address = rteth_set_mac_address,
 	.ndo_validate_addr = eth_validate_addr,
@@ -1623,6 +1671,7 @@ static const struct net_device_ops rteth_838x_netdev_ops = {
 
 static const struct rteth_config rteth_838x_cfg = {
 	.cpu_port = RTETH_838X_CPU_PORT,
+	.max_mtu = RTETH_838X_MAX_FRAME - RTETH_FRAME_OVERHEAD,
 	.rx_rings = 8,
 	.tx_rx_enable = 0xc,
 	.tx_trigger_mask = BIT(1),
@@ -1652,6 +1701,7 @@ static const struct rteth_config rteth_838x_cfg = {
 	.hw_stop = &rteth_838x_hw_stop,
 	.hw_reset = &rteth_838x_hw_reset,
 	.init_mac = &rteth_838x_init_mac,
+	.set_hol = rteth_83xx_set_hol,
 	.set_max_packet_length = rteth_838x_set_max_packet_length,
 	.netdev_ops = &rteth_838x_netdev_ops,
 };
@@ -1659,6 +1709,7 @@ static const struct rteth_config rteth_838x_cfg = {
 static const struct net_device_ops rteth_839x_netdev_ops = {
 	.ndo_open = rteth_open,
 	.ndo_stop = rteth_stop,
+	.ndo_change_mtu = rteth_change_mtu,
 	.ndo_start_xmit = rteth_start_xmit,
 	.ndo_set_mac_address = rteth_set_mac_address,
 	.ndo_validate_addr = eth_validate_addr,
@@ -1671,6 +1722,7 @@ static const struct net_device_ops rteth_839x_netdev_ops = {
 
 static const struct rteth_config rteth_839x_cfg = {
 	.cpu_port = RTETH_839X_CPU_PORT,
+	.max_mtu = RTETH_839X_MAX_FRAME - RTETH_FRAME_OVERHEAD,
 	.rx_rings = 8,
 	.tx_rx_enable = 0xc,
 	.tx_trigger_mask = BIT(1),
@@ -1698,6 +1750,7 @@ static const struct rteth_config rteth_839x_cfg = {
 	.hw_stop = &rteth_839x_hw_stop,
 	.hw_reset = &rteth_839x_hw_reset,
 	.init_mac = &rteth_839x_init_mac,
+	.set_hol = rteth_83xx_set_hol,
 	.set_max_packet_length = rteth_839x_set_max_packet_length,
 	.setup_notify_ring_buffer = &rteth_839x_setup_notify_ring_buffer,
 	.netdev_ops = &rteth_839x_netdev_ops,
@@ -1706,6 +1759,7 @@ static const struct rteth_config rteth_839x_cfg = {
 static const struct net_device_ops rteth_930x_netdev_ops = {
 	.ndo_open = rteth_open,
 	.ndo_stop = rteth_stop,
+	.ndo_change_mtu = rteth_change_mtu,
 	.ndo_start_xmit = rteth_start_xmit,
 	.ndo_set_mac_address = rteth_set_mac_address,
 	.ndo_validate_addr = eth_validate_addr,
@@ -1718,6 +1772,7 @@ static const struct net_device_ops rteth_930x_netdev_ops = {
 
 static const struct rteth_config rteth_930x_cfg = {
 	.cpu_port = RTETH_930X_CPU_PORT,
+	.max_mtu = RTETH_930X_MAX_FRAME - RTETH_FRAME_OVERHEAD,
 	.rx_rings = 32,
 	.tx_rx_enable = 0x30,
 	.tx_trigger_mask = GENMASK(3, 2),
@@ -1746,6 +1801,7 @@ static const struct rteth_config rteth_930x_cfg = {
 	.hw_stop = &rteth_930x_hw_stop,
 	.hw_reset = &rteth_93xx_hw_reset,
 	.init_mac = &rteth_930x_init_mac,
+	.set_hol = rteth_93xx_set_hol,
 	.set_max_packet_length = rteth_930x_set_max_packet_length,
 	.netdev_ops = &rteth_930x_netdev_ops,
 };
@@ -1753,6 +1809,7 @@ static const struct rteth_config rteth_930x_cfg = {
 static const struct net_device_ops rteth_931x_netdev_ops = {
 	.ndo_open = rteth_open,
 	.ndo_stop = rteth_stop,
+	.ndo_change_mtu = rteth_change_mtu,
 	.ndo_start_xmit = rteth_start_xmit,
 	.ndo_set_mac_address = rteth_set_mac_address,
 	.ndo_validate_addr = eth_validate_addr,
@@ -1765,6 +1822,7 @@ static const struct net_device_ops rteth_931x_netdev_ops = {
 
 static const struct rteth_config rteth_931x_cfg = {
 	.cpu_port = RTETH_931X_CPU_PORT,
+	.max_mtu = RTETH_931X_MAX_FRAME - RTETH_FRAME_OVERHEAD,
 	.rx_rings = 32,
 	.tx_rx_enable = 0x30,
 	.tx_trigger_mask = GENMASK(3, 2),
@@ -1793,6 +1851,7 @@ static const struct rteth_config rteth_931x_cfg = {
 	.hw_stop = &rteth_931x_hw_stop,
 	.hw_reset = &rteth_93xx_hw_reset,
 	.init_mac = &rteth_931x_init_mac,
+	.set_hol = rteth_93xx_set_hol,
 	.set_max_packet_length = rteth_931x_set_max_packet_length,
 	.netdev_ops = &rteth_931x_netdev_ops,
 };
@@ -1892,7 +1951,7 @@ static int rteth_probe(struct platform_device *pdev)
 
 	dev->ethtool_ops = &rteth_ethtool_ops;
 	dev->min_mtu = ETH_ZLEN;
-	dev->max_mtu = SKB_MTU;
+	dev->max_mtu = ctrl->r->max_mtu;
 	dev->features = NETIF_F_RXCSUM;
 	dev->hw_features = NETIF_F_RXCSUM;
 	dev->netdev_ops = ctrl->r->netdev_ops;
--- a/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.h
+++ b/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.h
@@ -194,6 +194,7 @@ struct rteth_frag;
 
 struct rteth_config {
 	int cpu_port;
+	int max_mtu;
 	int rx_rings;
 	int tx_rx_enable;
 	int tx_trigger_mask;
@@ -222,6 +223,7 @@ struct rteth_config {
 	void (*hw_stop)(struct rteth_ctrl *ctrl);
 	void (*hw_reset)(struct rteth_ctrl *ctrl);
 	int (*init_mac)(struct rteth_ctrl *ctrl);
+	void (*set_hol)(struct rteth_ctrl *ctrl);
 	void (*set_max_packet_length)(struct rteth_ctrl *ctrl, int len);
 	void (*setup_notify_ring_buffer)(struct rteth_ctrl *ctrl);
 	void (*update_counter)(struct rteth_ctrl *ctrl, int ring, int released);
--- a/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.h
+++ b/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.h
@@ -11,6 +11,8 @@
 #define RTETH_838X_DMA_IF_CTRL			(0x9f58)
 #define RTETH_838X_DMA_IF_INTR_MSK		(0x9f50)
 #define RTETH_838X_DMA_IF_INTR_STS		(0x9f54)
+#define RTETH_838X_DMA_IF_PKT_RX_FLTR_CTRL	(0x6b10)
+#define RTETH_838X_DMA_IF_PKT_TX_FLTR_CTRL	(0xaa6c)
 #define RTETH_838X_DMA_IF_RX_RING_CNTR		(0xb7e8)
 #define RTETH_838X_DMA_IF_RX_RING_SIZE		(0xb7e4)
 #define RTETH_838X_DMA_RX_BASE			(0x9f00)
@@ -30,6 +32,7 @@
 #define RTETH_839X_DMA_IF_CTRL			(0x786c)
 #define RTETH_839X_DMA_IF_INTR_MSK		(0x7864)
 #define RTETH_839X_DMA_IF_INTR_STS		(0x7868)
+#define RTETH_839X_DMA_IF_PKT_FLTR_CTRL		(0x1000)
 #define RTETH_839X_DMA_IF_RX_RING_CNTR		(0x603c)
 #define RTETH_839X_DMA_IF_RX_RING_SIZE		(0x6038)
 #define RTETH_839X_DMA_RX_BASE			(0x780c)
@@ -55,7 +58,9 @@
 #define RTETH_930X_DMA_TX_BASE			(0xe000)
 #define RTETH_930X_MAC_FORCE_MODE_CTRL		(0xca1c + RTETH_930X_CPU_PORT * 4)
 #define RTETH_930X_MAC_L2_ADDR_CTRL		(0xc714)
+#define RTETH_930X_MAC_L2_CPU_MAX_LEN_CTRL	(0xa3a0)
 #define RTETH_930X_MAC_L2_PORT_CTRL		(0x3268 + RTETH_930X_CPU_PORT * 64)
+#define RTETH_930X_MAC_L2_PORT_MAX_LEN_CTRL	(0x326c + RTETH_930X_CPU_PORT * 64)
 #define RTETH_930X_QM_RSN2CPUQID_CTRL_0		(0xa344)
 #define RTETH_930X_QM_RSN2CPUQID_CTRL_CNT	11
 #define RTETH_930X_RMA_CTRL_0			(0x9e60)
@@ -72,6 +77,7 @@
 #define RTETH_931X_DMA_TX_BASE			(0x0900)
 #define RTETH_931X_MAC_FORCE_MODE_CTRL		(0x0dcc + RTETH_931X_CPU_PORT * 4)
 #define RTETH_931X_MAC_L2_ADDR_CTRL		(0x135c)
+#define RTETH_931X_MAC_L2_CPU_MAX_LEN_CTRL	(0x1368)
 #define RTETH_931X_MAC_L2_PORT_CTRL		(0x6000 + RTETH_931X_CPU_PORT * 128)
 #define RTETH_931X_QM_RSN2CPUQID_CTRL_0		(0xa9f4)
 #define RTETH_931X_QM_RSN2CPUQID_CTRL_CNT	14
@@ -216,6 +222,7 @@ struct rteth_config {
 	void (*hw_stop)(struct rteth_ctrl *ctrl);
 	void (*hw_reset)(struct rteth_ctrl *ctrl);
 	int (*init_mac)(struct rteth_ctrl *ctrl);
+	void (*set_max_packet_length)(struct rteth_ctrl *ctrl, int len);
 	void (*setup_notify_ring_buffer)(struct rteth_ctrl *ctrl);
 	void (*update_counter)(struct rteth_ctrl *ctrl, int ring, int released);
 	const struct net_device_ops *netdev_ops;


plappermaul:
The Description logic is the other way round:

We add a function to change the MTU.
To avoid issues with the receive buffer, adapt HOL.

Please adapt commit Message. 

and while thinking about it. Add a check to the resizing function that verifies that we do not overflow the Kernel Maximum SKB frag Limit. If yes abort the MTU change.

thanks.

l0rdg3x:
> The Description logic is the other way round:
> 
> We add a function to change the MTU. To avoid issues with the receive buffer, adapt HOL.
> 
> Please adapt commit Message.
> 
> and while thinking about it. Add a check to the resizing function that verifies that we do not overflow the Kernel Maximum SKB frag Limit. If yes abort the MTU change.
> 
> thanks.

You are welcome.
I'm working on that.

l0rdg3x:
Both points are in, force-pushed:

- commit message rewritten around the `change_mtu`, with the head of line adaptation as its consequence;
- `rteth_93xx_set_hol()` now returns an error when a frame would need more fragments than `MAX_SKB_FRAGS`, and `change_mtu` aborts and restores the previous MTU.

The limit cannot be reached with the current 400 byte fragments, so the rejection was exercised with a throwaway build using 92 byte ones: `ip link set eth0 mtu 1600` then fails with `Invalid argument`, and both the MTU and the budget register stay where they were. The regular measurements are unchanged - 32 packets per ring at the 1504 byte conduit MTU, 25 at 1600, 42 at 1000, 64 at 600, with no receive error and no drop across the changes.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -474,27 +474,41 @@ static void rteth_839x_hw_reset(struct rteth_ctrl *ctrl)
...
 {
 	/*
 	 * The counter registers track the number of packets that are allowed to be appended to
 	 * the ring buffer. On RTL93xx a ring overflow must be avoided at all cost. Defensively
 	 * calculate the space by anticipating that only fully filled packets are received.
 	 */
-	int max_frame_size = SKB_MTU + ETH_HLEN + ETH_FCS_LEN + VLAN_HLEN + 4;
+	int max_frame_size = ctrl->dev->mtu + ETH_HLEN + ETH_FCS_LEN + VLAN_HLEN + 4;
 	int frags_per_pkt = DIV_ROUND_UP(max_frame_size, SKB_FRAG_SIZE);
 	int cnt = min(RTETH_RX_RING_SIZE / frags_per_pkt, 0x3ff);
```
openwrt-ai: The budget now scales with `dev->mtu`, but the number of fragments a received packet actually consumes is not bounded by `dev->mtu` — it is bounded by the DMA truncation limit, which stays at `SKB_MTU` on both RTL93xx variants ([`rteth_930x_hw_en_rxtx()` :627](``https://github.com/openwrt/openwrt/blob/ab23e2505533c5016fc89f18b383dd1bac462ca5/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L627),`` [`rteth_931x_hw_en_rxtx()` :645](``https://github.com/openwrt/openwrt/blob/ab23e2505533c5016fc89f18b383dd1bac462ca5/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L645)).`` Nothing else clamps the frame size on the way to the CPU port either: `rtl83xx` never implements `ds->ops->port_change_mtu` (so [`dsa_user_change_mtu()`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/net/dsa/user.c#L2118) bails with `-EOPNOTSUPP` and never propagates the conduit MTU to the ASIC), and the DSA driver programs no per-port max-frame-length register at all.

So the invariant the comment above states — "a ring overflow must be avoided at all cost", `cnt * frags_per_pkt <= RTETH_RX_RING_SIZE` — only holds if every frame really is `<= dev->mtu`. Lowering the MTU breaks it: at MTU 600 this programs `cnt = 64` per ring on the assumption of 2 fragments per packet, but an ordinary 1518-byte frame flooded to the CPU port still consumes 4 fragments, i.e. up to 256 fragments against a 128-entry ring. The old static sizing from `SKB_MTU` was safe against that by construction.

Deriving `frags_per_pkt` from `max(dev->mtu, SKB_MTU)` would keep the guarantee but make the whole change a no-op, so the real fix is presumably to also lower the `dma_if_ctrl` truncation limit alongside the budget, so the hardware cannot deliver a frame longer than the MTU the budget was computed for. Was that considered? As it stands, the direction that gains budget is the direction that loses the overflow guarantee.

---

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -974,6 +988,25 @@ static int rteth_stop(struct net_device *dev)
...
 
+static int rteth_93xx_change_mtu(struct net_device *dev, int mtu)
+{
+	struct rteth_ctrl *ctrl = netdev_priv(dev);
+	int old_mtu = dev->mtu;
+	int ret;
+
+	if (mtu > dev->mtu)
+		netdev_warn(dev,
+			    "raising the MTU may have unexpected effects under high receive load\n");
```
openwrt-ai: The warning is on the wrong branch. Raising the MTU *shrinks* the budget — the PR's own table shows 25 packets/ring at MTU 1600 versus 64 at MTU 600 — so `mtu > dev->mtu` is the conservative direction. The direction that hands the hardware more budget than the ring can back for real-world frame sizes is `mtu < dev->mtu`, and that one is silent.

If the warning is meant to flag "budget was just handed out while the ring may still hold larger frames", it belongs on the decrease:

```suggestion
	if (mtu < dev->mtu)
		netdev_warn(dev,
			    "lowering the MTU raises the receive budget and may have unexpected effects under high receive load\n");
```

(If instead the concern is shrinking the budget below the hardware's current in-flight counter, please say so in the message — the current text doesn't match either reading.)

---

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -474,27 +474,41 @@ static void rteth_839x_hw_reset(struct rteth_ctrl *ctrl)
...
 	 * calculate the space by anticipating that only fully filled packets are received.
 	 */
-	int max_frame_size = SKB_MTU + ETH_HLEN + ETH_FCS_LEN + VLAN_HLEN + 4;
+	int max_frame_size = ctrl->dev->mtu + ETH_HLEN + ETH_FCS_LEN + VLAN_HLEN + 4;
 	int frags_per_pkt = DIV_ROUND_UP(max_frame_size, SKB_FRAG_SIZE);
 	int cnt = min(RTETH_RX_RING_SIZE / frags_per_pkt, 0x3ff);
 
-	rteth_nic_reset(ctrl, 0x6);
+	if (frags_per_pkt > MAX_SKB_FRAGS)
+		return -EINVAL;
```
openwrt-ai: This runtime check can never fire: `dev->max_mtu = SKB_MTU` (1600) is enforced by the core before `ndo_change_mtu` is reached ([:1874](``https://github.com/openwrt/openwrt/blob/ab23e2505533c5016fc89f18b383dd1bac462ca5/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L1874)),`` so `frags_per_pkt` is bounded by `DIV_ROUND_UP(1626, SKB_FRAG_SIZE)` = 5, well under `MAX_SKB_FRAGS`. The quantity that can actually push it over `MAX_SKB_FRAGS` is `SKB_FRAG_SIZE` — and note the TODO at the definition planning to change it — which is a compile-time constant. A `BUILD_BUG_ON` next to the existing one in `rteth_setup_ring_buffers()` would catch a bad `SKB_FRAG_SIZE`/`SKB_MTU` combination at build time and cover the RX path too, rather than only the MTU-change path where it is unreachable.

nit: the commit message says the rejection was exercised "with a throwaway build using 64 byte fragments", while [the PR comment](https://github.com/openwrt/openwrt/pull/24421#issuecomment-5080125792) says 92 byte ones. Worth making the two agree.

---

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -474,27 +474,41 @@ static void rteth_839x_hw_reset(struct rteth_ctrl *ctrl)
...
+	}
+
+	return 0;
+}
+
+static void rteth_93xx_hw_reset(struct rteth_ctrl *ctrl)
+{
+	rteth_nic_reset(ctrl, 0x6);
+
+	rteth_93xx_set_hol(ctrl);
```
openwrt-ai: The return value is discarded here. `rteth_93xx_set_hol()` was just given a failure path that leaves the ring-size registers untouched, so if it ever did fail on this path the reset would silently continue with whatever budget the previous configuration (or the bootloader) left behind — the one state the function's own comment says must not happen. `rteth_93xx_hw_reset()` is `void` and can't propagate it, so at minimum log it:

```suggestion
	if (rteth_93xx_set_hol(ctrl))
		netdev_err(ctrl->dev, "failed to program the head of line budget\n");
```

---

openwrt-ai (commented):
## Commit checks

- `ab23e250` "realtek: eth: support changing the MTU on RTL93xx" — the message states "Raising the MTU hands additional budget to the hardware while the ring may still be filled with larger frames, so warn about it", but the code does the opposite: a larger MTU yields a larger `frags_per_pkt` and therefore a *smaller* `cnt`. The message's own measurements confirm it (25 packets/ring at MTU 1600, 64 at MTU 600). The justification given for the `netdev_warn()` on the increase branch doesn't hold as written.
- `ab23e250` — "Nothing else has to follow, as the receive buffers are of a fixed size and the DMA truncation limit is independent of the MTU" reads as a reason no further change is needed, but that independence is exactly what breaks the ring-overflow guarantee when the MTU is lowered. See the inline comment on `rteth_93xx_set_hol()`.


---

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
plappermaul: Make this a generic unconditional warning. "changing ..."

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
plappermaul: Make this a compile time check.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: Done, the warning is now unconditional on any MTU change.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: Done, it is a `BUILD_BUG_ON` now. Since the DMA truncates at `SKB_MTU`, the check bounds that frame rather than the current MTU, so it covers every frame the hardware can actually deliver.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: Right, the truncation limit is what bounds a received frame, and the build time check now covers exactly that.

It leaves one question on the budget itself: since truncation stays at `SKB_MTU` regardless of the MTU, a lowered MTU makes the budget assume frames smaller than the ones the hardware may still deliver - 64 packets per ring at MTU 600 against frames of up to 1600 bytes, five fragments each. Should the budget keep the truncation limit as its ceiling, or is the warning enough here?

plappermaul: The solution to this must be decopuled from DSA. The ethernet driver must protect itself properly even when running without DSA. So it is important to identifiy the right truncation register in the SOC. We currently blindfolded set RX_TRUNCATE_LEN. This was copied from the SDK driver. But to be honest we have mutliple possible registers:

- DMA_IF_CTRL->RX_TRUNCATE_LEN
- MAC_L2_PORT_MAX_LEN_CTRL[CPU_PORT]->MAX_LEN_100_10...
- MAC_L2_PORT_MAX_LEN_CTRL[CPU_PORT]->MAX_LEN_10G_25G_...

Please be aware that TX/RX in a switch register can take the SOC view. So "RX_TRUNCATE_LEN" register might translate to "TRUNCATE RECEIVE TRAFFIC FROM LINUX TO SWITCH". That would be the TX path in the ethernet driver.

Proceed as follows:

- Find out exactly which of the registers really truncates incoming traffic (Switch->Linux) and which outgoing traffic (Linux->Switch).  
- Add a pre-commit to this series that builds the proper foundation for this enhancement. And even if it is only a meaningful comment that explains the truncation register logic.

l0rdg3x: Measured on an RTL9303: port MTU at 9000, one register at a time lowered to 3000, frames of 1046 and 4046 bytes sent in both directions, each scenario twice with identical results.

| register | switch to Linux | Linux to switch |
|---|---|---|
| `DMA_IF_CTRL` truncation | unaffected | limited |
| `MAC_L2_CPU_MAX_LEN_CTRL` | limited | limited |
| `MAC_L2_PORT_MAX_LEN_CTRL` | limited at that port | - |

So the truncation length bounds transmission only, as you suspected, and what bounds reception is the CPU port maximum length - the one register both directions cross. It also holds a single length field, unlike the user port ones which hold two.

Added as a pre-commit documenting this next to the truncation definitions. Since the ethernet driver never programs that register, without DSA it keeps whatever the bootloader left, 12288 here. It also means the `BUILD_BUG_ON` bounds the transmit limit, so tell me if you want the check and the budget to follow the CPU port register instead.

plappermaul: Great find. With this the first commit must be enhanced to really safeguard the CPU port with the right registers. But this should be done for all series. We have:

Maple/8380:
- [DMA_IF_PKT_RX_FLTR_CTRL](https://svanheule.net/realtek/maple/register/dma_if_pkt_rx_fltr_ctrl)
- [DMA_IF_PKT_TX_FLTR_CTRL](https://svanheule.net/realtek/maple/register/dma_if_pkt_tx_fltr_ctrl)

Cypress/8390:
- [DMA_IF_PKT_FLTR_CTRL](https://svanheule.net/realtek/cypress/register/dma_if_pkt_fltr_ctrl)

Longan/9300:
- [MAC_L2_CPU_MAX_LEN_CTRL](https://svanheule.net/realtek/longan/register/mac_l2_cpu_max_len_ctrl)

Mango/9310:
- [MAC_L2_CPU_MAX_LEN_CTRL](https://svanheule.net/realtek/mango/register/mac_l2_cpu_max_len_ctrl)

The change_mtu() commit will build on top of this and make the safeguarding dynamic.

Once a commit proposal is available I can test on non-RTL9300.

l0rdg3x: Great! I continue in this direction.

l0rdg3x: Correction to my previous message: I had it backwards, that table is wrong.

I was counting the packets captured on the switch port rather than looking at them, and I measured the outgoing direction through the replies - which come back at the same size and were being cut on the way in. Two flaws that hid each other.

Looking at the frame length instead, port MTU at 9000, 4046 byte frames, one register at a time lowered to 3000:

| state | frames arriving at the CPU |
|---|---|
| baseline | 6 x length 4008, intact |
| `DMA_IF_CTRL` truncation | frames of length 2978 appear - cut short, not dropped |
| `MAC_L2_CPU_MAX_LEN_CTRL` | nothing captured - dropped at the port |

And measuring transmission by the bytes reaching the host: 44060 with the truncation at 3000 against 44216 at baseline, so neither register bounds Linux to switch.

So both act on the receive path, one cutting and one dropping, and `RX_TRUNCATE_LEN` does take the DMA point of view after all - the DMA is what receives from the switch. Your `CPU_PORT_TX_MAX_LEN` naming agrees: the switch is what transmits there.

The pre-commit is amended with the corrected wording.

l0rdg3x:
Three commits now: the documentation, then the safeguard, then `change_mtu` making it dynamic.

The safeguard programs the CPU port receive limit from the driver instead of leaving the bootloader value, with the register of each family - `DMA_IF_PKT_RX_FLTR_CTRL` on RTL838x, `DMA_IF_PKT_FLTR_CTRL` on RTL839x, `MAC_L2_CPU_MAX_LEN_CTRL` on RTL930x and RTL931x - the length being the low 14 bits on all four. The value comes from the conduit MTU, which is also what DSA writes into that register on RTL930x, so the two agree instead of overwriting each other.

On the RTL9303 the register now reads 1530 for the 1504 byte conduit MTU where it held the bootloader value of 12288, and follows every change (1626 at 1600, 1026 at 1000, 626 at 600) along with the budget. 4046 byte frames are dropped at the port now, instead of reaching the CPU truncated.

RTL838x/839x/931x are compile-tested only, so the offer to test them is very welcome. One thing I left alone: on RTL83xx the limit keeps the value programmed at open, since `change_mtu` is wired on RTL93xx only. Say the word if you want it there as well.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -559,12 +586,24 @@ static void rteth_hw_ring_setup(struct rteth_ctrl *ctrl)
...
 }
 
+static void rteth_set_cpu_rx_max_len(struct rteth_ctrl *ctrl)
+{
+	/*
+	 * Have the switch drop what the DMA would otherwise deliver truncated. Without this
+	 * the limit is whatever the bootloader left, which on RTL9303 is the chip maximum.
+	 */
+	regmap_update_bits(ctrl->map, ctrl->r->cpu_rx_max_len_ctrl, GENMASK(13, 0),
+			   ctrl->dev->mtu + RTETH_FRAME_OVERHEAD);
```
openwrt-ai: The port limit and the DMA truncation limit are expressed in different units here, so the top of the MTU range still gets truncated — the exact case this commit sets out to remove.

`dma_if_ctrl` is programmed with a truncation length of `SKB_MTU` = 1600 ([`rteth_930x_hw_en_rxtx()`:655](``https://github.com/openwrt/openwrt/blob/c82f09d4fe9c09b19a0b49e44e87ad695202a610/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L655),`` same on the other three families), while `dev->max_mtu = SKB_MTU` ([:1903](``https://github.com/openwrt/openwrt/blob/c82f09d4fe9c09b19a0b49e44e87ad695202a610/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L1903)).`` At MTU 1600 this line programs 1626 into the port register, so a frame of 1601..1626 bytes passes the port and is then cut short by the DMA. For any MTU above 1574 the two disagree.

The code holds both readings of the truncation field at once: the comment on the register write calls it "RX Buffer truncated at SKB_MTU Bytes" (a frame length), while `rteth_93xx_set_hol()` and its `BUILD_BUG_ON` compute `SKB_MTU + overhead` as the largest frame the DMA can deliver (an MTU). Your commit 1 measurement — truncation at 3000 yielding frames of 2978 on the wire — reads like a frame length, which would make this line 26 bytes too permissive. Which of the two is it, and should this be `min(dev->mtu + RTETH_FRAME_OVERHEAD, SKB_MTU)` (or the truncation raised to `SKB_MTU + RTETH_FRAME_OVERHEAD`) so the drop always happens before the cut?

---

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -474,27 +488,40 @@ static void rteth_839x_hw_reset(struct rteth_ctrl *ctrl)
...
-static void rteth_93xx_hw_reset(struct rteth_ctrl *ctrl)
+static void rteth_93xx_set_hol(struct rteth_ctrl *ctrl)
 {
 	/*
 	 * The counter registers track the number of packets that are allowed to be appended to
 	 * the ring buffer. On RTL93xx a ring overflow must be avoided at all cost. Defensively
 	 * calculate the space by anticipating that only fully filled packets are received.
 	 */
-	int max_frame_size = SKB_MTU + ETH_HLEN + ETH_FCS_LEN + VLAN_HLEN + 4;
+	int max_frame_size = ctrl->dev->mtu + ETH_HLEN + ETH_FCS_LEN + VLAN_HLEN + 4;
```
openwrt-ai: The frame overhead is open-coded here after commit 2 introduced `RTETH_FRAME_OVERHEAD` for exactly this value (`ETH_HLEN + 2 * VLAN_HLEN + ETH_FCS_LEN` = `ETH_HLEN + ETH_FCS_LEN + VLAN_HLEN + 4` = 26). Two spellings of the same quantity in one file means the budget and the port limit can silently drift apart on any future change to the overhead.

```suggestion
	int max_frame_size = ctrl->dev->mtu + RTETH_FRAME_OVERHEAD;
```

Same applies to the `BUILD_BUG_ON` on lines 503-504, which should read `DIV_ROUND_UP(SKB_MTU + RTETH_FRAME_OVERHEAD, SKB_FRAG_SIZE) > MAX_SKB_FRAGS`.

---

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -974,6 +1019,19 @@ static int rteth_stop(struct net_device *dev)
...
 
+static int rteth_93xx_change_mtu(struct net_device *dev, int mtu)
+{
+	struct rteth_ctrl *ctrl = netdev_priv(dev);
+
+	netdev_warn(dev, "changing the MTU may have unexpected effects under high receive load\n");
+
+	WRITE_ONCE(dev->mtu, mtu);
+	rteth_set_cpu_rx_max_len(ctrl);
+	rteth_93xx_set_hol(ctrl);
```
openwrt-ai: On an MTU increase these two writes are in the unsafe order. `rteth_set_cpu_rx_max_len()` immediately lets the switch hand over larger frames, while the head of line budget is still the old, larger `cnt` — so between the two register writes the hardware may append up to the old budget of new, larger frames, which is the ring overflow the comment in `rteth_93xx_set_hol()` says must be avoided at all cost. On a decrease the current order is the safe one, so it needs to depend on the direction:

```suggestion
	if (mtu > dev->mtu) {
		/* shrink the budget before the switch may hand over larger frames */
		WRITE_ONCE(dev->mtu, mtu);
		rteth_93xx_set_hol(ctrl);
		rteth_set_cpu_rx_max_len(ctrl);
	} else {
		WRITE_ONCE(dev->mtu, mtu);
		rteth_set_cpu_rx_max_len(ctrl);
		rteth_93xx_set_hol(ctrl);
	}
```

(Both branches must write `dev->mtu` first, since both helpers derive their value from it.)

---

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -559,12 +586,24 @@ static void rteth_hw_ring_setup(struct rteth_ctrl *ctrl)
...
+			   ctrl->dev->mtu + RTETH_FRAME_OVERHEAD);
+}
+
 static void rteth_838x_hw_en_rxtx(struct rteth_ctrl *ctrl)
 {
 	/* Truncate RX buffer to SKB_MTU bytes, pad TX */
 	regmap_write(ctrl->map, ctrl->r->dma_if_ctrl,
 		     (SKB_MTU << 16) | RX_TRUNCATE_EN_83XX | TX_PAD_EN_838X);
 
+	rteth_set_cpu_rx_max_len(ctrl);
```
openwrt-ai: On RTL838x/839x this turns a previously working MTU increase into silent hardware drops. There is no `.ndo_change_mtu` in [`rteth_838x_netdev_ops`](``https://github.com/openwrt/openwrt/blob/c82f09d4fe9c09b19a0b49e44e87ad695202a610/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L1617``) or the 839x one, so `dev_set_mtu()` only updates `dev->mtu` and this register keeps the value computed at open. Raise the MTU from the default 1500 to, say, 1600 on a live interface and the port now drops everything above 1526, where before this commit the bootloader's limit (12288 on RTL9303, presumably similarly permissive elsewhere) let it through.

Your [PR comment](https://github.com/openwrt/openwrt/pull/24421#issuecomment-5083337627) mentions leaving RTL83xx alone, but frames it as "keeps the value programmed at open" rather than as a behaviour change. Since the RTL83xx families need no budget resize, a `change_mtu` there that only calls `rteth_set_cpu_rx_max_len()` would keep this commit purely additive — or, if it is deliberately left for later, the commit message should say that raising the MTU on those families now requires an interface down/up.

---

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -38,11 +38,25 @@
 
 #define NOTIFY_EVENTS			10
 #define NOTIFY_BLOCKS			10
+/*
```
openwrt-ai: nit: this comment block is glued to the unrelated `NOTIFY_EVENTS`/`NOTIFY_BLOCKS` pair above it. Every other define group in this region (`RING_*`, `RTETH_*_RING*`, `NOTIFY_*`, `SKB_*`, `PPOOL_*`) is separated by a blank line, so the block reads as belonging to `NOTIFY_BLOCKS` rather than to the `RX_TRUNCATE_EN_*` defines it documents.

```suggestion

/*
```

---

openwrt-ai (commented):
## Commit checks

- `7f6e3961` "realtek: eth: bound the frames the switch delivers to the CPU" — the message states "The value is derived from the conduit MTU, which is also what DSA programs into the same register on RTL930x, so the two agree instead of overwriting each other". The in-tree `rtl83xx` DSA driver does neither: it implements no `port_change_mtu`/`port_max_mtu` op and never touches `MAC_L2_CPU_MAX_LEN_CTRL` — grepping `target/linux/realtek/` for `0xa3a0` and `port_change_mtu` at this head matches only the new `RTETH_930X_MAC_L2_CPU_MAX_LEN_CTRL` define this commit adds. The register is unowned before this commit (which is what commit `5f2665a8` documents), so the "agree instead of overwriting" rationale doesn't apply as written. If it refers to the vendor SDK or to a planned DSA change, say so explicitly.

---

plappermaul:
I just verified the packet handling on RTL838x. The identified registers are what we need. RX_TRUNCATE_EN is not the right way. Please proceed as follows:

- Commit 1/2 can be squashed. The description and the code belong together.
- Remove cpu_rx_max_len_ctrl from the config structure
- delete the exisitng truncate commands 
- Add a device specific set_max_packet_length() function to the config structure.
- Each one programs the new identified TX AND RX registers
- Each one clears RX_TRUNCATE_EN bit in DMA_IF_CTRL and leaves the other bits untouched.
- Call the new function where needed.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
plappermaul: Implement this advise. The WRITE_ONCE can preceed the if block

l0rdg3x:
Reworked as asked: the two commits are squashed, the config field is gone, and each family now has a `set_max_packet_length()` that programs its own registers and clears `RX_TRUNCATE_EN` while leaving the rest of `DMA_IF_CTRL` alone.

Two things worth flagging:

- Longan holds a single length field, so there "TX and RX" comes down to one write. Maple needs a second register for the transmit side, which is not in the tree yet: https://svanheule.net/realtek/maple/register/dma_if_pkt_tx_fltr_ctrl — that offset is the one value in the series I could neither cross-check in-tree nor measure myself, so it is the first thing worth confirming on your RTL838x.
- The `BUILD_BUG_ON` comment had to change: it justified itself with the truncation, which is now gone. The check still holds, but on `dev->max_mtu` instead.

On the RTL9303 the register now holds 1530 for the 1504 byte conduit MTU and follows every change, `DMA_IF_CTRL` reads 0x30 with the enables untouched, and with the limit at 1026 a 1022 byte frame still arrives while a 1023 byte one does not.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: No longer applies: the check is a `BUILD_BUG_ON` now, as the maintainer asked, so `rteth_93xx_set_hol()` went back to returning void and there is no value to discard. [L504](https://github.com/openwrt/openwrt/blob/e40689fae5aeb10378be71082a39a198c2fcd7e0/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L504)

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: Good catch, and the answer turned out to be that it is a frame length: measured on an RTL9303, the largest frame that survives is the truncation length minus eight and the maximum length minus four, the difference being the tail tag the switch appends. So the two really did disagree.

The series no longer relies on that relationship though. Following the maintainer's direction the truncation is switched off entirely and the port limit is the only bound left, so there is nothing left to cap against. [L589](https://github.com/openwrt/openwrt/blob/e40689fae5aeb10378be71082a39a198c2fcd7e0/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L589)

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: Done, both spellings are `RTETH_FRAME_OVERHEAD` now — the budget at [L499](https://github.com/openwrt/openwrt/blob/e40689fae5aeb10378be71082a39a198c2fcd7e0/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L499) and the build time check at [L504](https://github.com/openwrt/openwrt/blob/e40689fae5aeb10378be71082a39a198c2fcd7e0/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L504).

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: Done. `WRITE_ONCE` first, then the two writes in the order that suits the direction. [L1045](https://github.com/openwrt/openwrt/blob/e40689fae5aeb10378be71082a39a198c2fcd7e0/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L1045)

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: You were right that it was a behaviour change and not just a limitation. RTL838x and RTL839x now have a `change_mtu` of their own, which follows the new MTU with the packet length alone since those families keep no budget to resize. [L1037](https://github.com/openwrt/openwrt/blob/e40689fae5aeb10378be71082a39a198c2fcd7e0/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L1037), wired at [L1654](https://github.com/openwrt/openwrt/blob/e40689fae5aeb10378be71082a39a198c2fcd7e0/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L1654) and [L1703](https://github.com/openwrt/openwrt/blob/e40689fae5aeb10378be71082a39a198c2fcd7e0/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L1703).

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: Done. [L41](https://github.com/openwrt/openwrt/blob/e40689fae5aeb10378be71082a39a198c2fcd7e0/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L41)

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -559,11 +574,47 @@ static void rteth_hw_ring_setup(struct rteth_ctrl *ctrl)
...
+			   FIELD_PREP(GENMASK(27, 14), len) | FIELD_PREP(GENMASK(13, 0), len));
+	regmap_clear_bits(ctrl->map, ctrl->r->dma_if_ctrl, RX_TRUNCATE_EN_83XX);
+}
+
+/* Longan holds a single length field, covering what the switch sends to the CPU */
+static void rteth_930x_set_max_packet_length(struct rteth_ctrl *ctrl, int len)
+{
+	regmap_update_bits(ctrl->map, RTETH_930X_MAC_L2_CPU_MAX_LEN_CTRL, GENMASK(13, 0), len);
+	regmap_clear_bits(ctrl->map, ctrl->r->dma_if_ctrl, RX_TRUNCATE_EN_93XX);
+}
```
plappermaul: SDK reads: 

```
    if ((ret = dal_longan_port_cpuPortId_get(unit, &port)) != RT_ERR_OK)
    {
        SWITCH_SEM_UNLOCK(unit);
        RT_ERR(ret, (MOD_DAL|MOD_SWITCH), "");
        return ret;
    }

    /* get entry from CHIP*/
    if (PKTDIR_RX == dir)
    {
        ret = reg_field_read(unit, LONGAN_MAC_L2_CPU_MAX_LEN_CTRLr, LONGAN_CPU_PORT_TX_MAX_LENf, pLen);
    }
    else
    {
        ret = reg_array_field_read(unit, LONGAN_MAC_L2_PORT_MAX_LEN_CTRLr, port, \
                                   REG_ARRAY_INDEX_NONE, \
                                   LONGAN_MAX_LEN_100M_10M_SELf, \
                                   pLen);
    }

```

So please program the L2_PORT_MAX_LEN_CTRL as well. (For safety the 2.5G field too.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -38,11 +38,26 @@
...
+ * Two registers bound the frames that reach the CPU, and they do it in different ways.
+ * Measured on an RTL9303 by lowering one at a time while frames of growing size were sent
+ * into a user port:
+ *
+ * - the DMA truncation length cuts the frame short. It still reaches the CPU, carrying
+ *   only as many bytes as the field allows.
+ * - the maximum length register of the CPU port drops the frame instead.
+ *
+ * The length fields are named after the switch rather than the driver, so the one that
+ * bounds what the driver receives is the transmitting one on RTL93xx.
```
plappermaul: Move this down to the 4 new length setters. Make it generic (not RTL93xx specials) like "3 registers control ..."

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -489,27 +489,39 @@ static void rteth_839x_hw_reset(struct rteth_ctrl *ctrl)
...
 	 * calculate the space by anticipating that only fully filled packets are received.
 	 */
-	int max_frame_size = SKB_MTU + ETH_HLEN + ETH_FCS_LEN + VLAN_HLEN + 4;
+	int max_frame_size = ctrl->dev->mtu + RTETH_FRAME_OVERHEAD;
 	int frags_per_pkt = DIV_ROUND_UP(max_frame_size, SKB_FRAG_SIZE);
 	int cnt = min(RTETH_RX_RING_SIZE / frags_per_pkt, 0x3ff);
 
-	rteth_nic_reset(ctrl, 0x6);
+	/* The MTU cannot exceed SKB_MTU, so that frame must fit into a single skb */
+	BUILD_BUG_ON(DIV_ROUND_UP(SKB_MTU + RTETH_FRAME_OVERHEAD, SKB_FRAG_SIZE) > MAX_SKB_FRAGS);
```
plappermaul: Last review was not perfect from my side. SKB_MTU should be made obsolete with this commit (replaced my dev->mtu). Move this down to change_mtu and make a runtime check from input parameter mtu.

plappermaul (changes requested):
While we are getting slowly towards the finish line ... All this vibe coding is very time consuming for me. We should convert it to some kind of AI based real hardware testing loop for new PRs or issues. This can mitigate the burden on the few developers.

l0rdg3x:
> While we are getting slowly towards the finish line ... All this vibe coding is very time consuming for me. We should convert it to some kind of AI based real hardware testing loop for new PRs or issues. This can mitigate the burden on the few developers.

I understand. My workflow already includes this automation for RTL9303. If I had all the different chipset models available, I would have all of them tested independently. That would be great.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: Done on RTL930x: `0x326c + cpu_port * 64`, both length fields.

The other three already cover that direction without it:

- RTL931x keeps `CPU_PORT_RX_MAX_LEN[27:14]` next to the TX field in
  `MAC_L2_CPU_MAX_LEN_CTRL`, and its `MAC_L2_PORT_MAX_LEN_CTRL` is documented for
  ports 0-55 while its CPU port is 56 — that index is `MAC_DBG_SEL_CTRL`.
- RTL838x/839x carry both lengths in the `DMA_IF_PKT_*_FLTR_CTRL` fields. Their only
  MAC length register, `MAC_MAX_LEN_CTRL` (0xa9e0 / 0x02b0), is a single global one
  described as sizing all switched traffic, so driving it from the conduit MTU would
  cap user-to-user forwarding as well. Say the word if you want it there anyway.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: Done: the block sits above the four setters now, rewritten as the three limits that bound the frames exchanged with the CPU port, with no RTL93xx wording left in it.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: Done: `SKB_MTU` is gone and the check sits in `change_mtu`, against the `mtu` argument.

`dev->max_mtu` needed a replacement, so it is now
`MAX_SKB_FRAGS * SKB_FRAG_SIZE - RTETH_FRAME_OVERHEAD` = 6774, which lifts the ceiling
from 1600 on all four families. One consequence worth naming: `dev_validate_mtu()` runs
before `ndo_change_mtu`, so the core already rejects everything the new check would
catch — it is a guard, not the gate. If you meant `max_mtu` to come from the chip frame
length instead (12288 / 10240 in the GPLs, per family), say so and I will steer it from
the config.

l0rdg3x:
Correction to two earlier messages of mine: the 12288 I attributed to
`MAC_L2_CPU_MAX_LEN_CTRL` (0xa3a0) belongs to the other register, the CPU port's entry
in `MAC_L2_PORT_MAX_LEN_CTRL` (0x326c + 28 * 64 = 0x396c). Read at the U-Boot prompt and
reproduced across two boots, the bootloader leaves 0xa3a0 at 0x600 = 1536, and 0x396c at
0x0c003000 = 12288 in both of its length fields.

With this series both registers hold 1530 at the 1504 byte conduit MTU and follow every
change: 1026 at 1000, 1627 at 1601, 6800 at 6774.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -1891,7 +1939,7 @@ static int rteth_probe(struct platform_device *pdev)
 
 	dev->ethtool_ops = &rteth_ethtool_ops;
 	dev->min_mtu = ETH_ZLEN;
-	dev->max_mtu = SKB_MTU;
+	dev->max_mtu = MAX_SKB_FRAGS * SKB_FRAG_SIZE - RTETH_FRAME_OVERHEAD;
 	dev->features = NETIF_F_RXCSUM;
```
plappermaul: Please from config with the real MAX MTU each series supports.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -1025,6 +1033,42 @@ static int rteth_stop(struct net_device *dev)
...
 }
 
+static int rteth_change_mtu(struct net_device *dev, int mtu)
+{
+	WRITE_ONCE(dev->mtu, mtu);
+	rteth_set_max_packet_length(netdev_priv(dev));
+
+	return 0;
+}
+
```
plappermaul: This needs the same sanity checks as below function. Remove this redundancy as follows:

- add a set_hol() function to the config structure
- let it point to 93xx_hol or leave it empty depending on model.
- merge these two change_mtu functions into on.
- call set_hol() if not NULL.
-

plappermaul (changes requested):
This seems to be the final round of fixes. Please fix and run some checks afterwards.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: Done: `.max_mtu` per family in the config, 10000 on RTL838x and 12288 on the other three,
less the 26 byte frame overhead. Those are the limits the DSA driver already declares in
its rmon ranges; I took 10000 rather than the 10240 of the SDK capacity table, since the
RTL8380 datasheet and the maple register map both give 10000.

Worth knowing before you take it: the fragment budget still caps the operation at 6774, so
on the RTL9303 `ip link set eth0 mtu 9000` now reaches the driver and is refused there.
Nothing above 6774 is settable on any family until `SKB_FRAG_SIZE` grows, and the check you
asked for is what refuses it — an accepted change logs the warning, a refused one does not.
Say the word if you would rather have `dev->max_mtu` clamped to that instead.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: Done: a single `rteth_change_mtu` wired into all four ops, with `set_hol()` in the config
pointing at `rteth_93xx_set_hol` on RTL930x/931x and left unset on RTL838x/839x, called only
where it is present. Those two families now get the sanity check and the warning as well.

One consequence of making the warning unconditional: `dsa_conduit_setup()` calls
`dev_set_mtu()` on the conduit, so the line is now logged once per boot on every family.

l0rdg3x:
Fourth round pushed: `.max_mtu` from the config per family, and one `change_mtu` with the
optional `set_hol()`.

On the RTL9303 after the merge, the length registers and the ring budget still follow every
MTU — 1530 and 32 packets per ring at 1504, 1026 and 42 at 1000, 6800 and 7 at 6774 — a 1046
byte frame passes at MTU 1504, stops at 1000 and passes again, and a refused MTU of 9000
logs no warning where an accepted change logs one. All four subtargets compile.

A retraction I owe on an earlier comment of mine: *"which is also what DSA programs into the
same register on RTL930x"* is false on main. The DSA driver there implements neither
`port_change_mtu` nor `port_max_mtu` and never touches that register.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -974,6 +1038,34 @@ static int rteth_stop(struct net_device *dev)
...
+	 * A larger MTU buys a smaller budget, so hand the rings their new limit before the
+	 * switch may deliver larger frames, and the other way around when the MTU shrinks.
+	 */
+	if (grow && ctrl->r->set_hol)
+		ctrl->r->set_hol(ctrl);
+
+	rteth_set_max_packet_length(ctrl);
+
+	if (!grow && ctrl->r->set_hol)
+		ctrl->r->set_hol(ctrl);
```
openwrt-ai: The hardware programming and the warning run even when the interface is down, where both are pointless: `rteth_open()` reprograms both limits from `dev->mtu` anyway — [`hw_reset` at :926](``https://github.com/openwrt/openwrt/blob/d2e33fa9e0064a27e3d3d154b3e24ff19e9da919/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L926``) reaches `rteth_93xx_set_hol()` and [`hw_en_rxtx` at :942](``https://github.com/openwrt/openwrt/blob/d2e33fa9e0064a27e3d3d154b3e24ff19e9da919/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L942``) reaches `rteth_set_max_packet_length()`.

That also covers the boot-time noise you flagged in your last comment: `dsa_conduit_setup()` calls `dev_set_mtu()` on the conduit, so the "unexpected effects under high receive load" line is logged once per boot on every family, for a change that touches no live traffic. Gating on `netif_running()` keeps the unconditional warning the maintainer asked for, but only where the caveat is true.

```suggestion
	WRITE_ONCE(dev->mtu, mtu);

	/* rteth_open() programs both limits from dev->mtu, so nothing to do while down */
	if (!netif_running(dev))
		return 0;

	netdev_warn(dev, "changing the MTU may have unexpected effects under high receive load\n");

	/*
	 * A larger MTU buys a smaller budget, so hand the rings their new limit before the
	 * switch may deliver larger frames, and the other way around when the MTU shrinks.
	 */
	if (grow && ctrl->r->set_hol)
		ctrl->r->set_hol(ctrl);

	rteth_set_max_packet_length(ctrl);

	if (!grow && ctrl->r->set_hol)
		ctrl->r->set_hol(ctrl);
```

---

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -1559,6 +1651,7 @@ static int rteth_setup_tc(struct net_device *dev, enum tc_setup_type type, void
 static const struct net_device_ops rteth_838x_netdev_ops = {
 	.ndo_open = rteth_open,
 	.ndo_stop = rteth_stop,
+	.ndo_change_mtu = rteth_change_mtu,
```
openwrt-ai, line 1661: nit: the PR description no longer matches the series after the rework, and it is what a reviewer reads first. Three disagreements against this head:

- *"RTL838x/839x are untouched: their reset paths write a ring size of zero and keep no budget to resize."* — they are not untouched any more: this line and :1704 wire `ndo_change_mtu` on both, and commit `2049a3b0` gives each of them a `set_max_packet_length()`.
- *"The warning is emitted on increases only."* — it is unconditional now, per the maintainer's request.
- The title, *"scale the head of line budget with the MTU"*, and the whole body describe only what is now the second commit; there is nothing about `2049a3b0` bounding the packet length at the CPU port, which is the larger half of the diff.

The register/packets-per-ring table is also from the pre-rework measurements; the numbers you posted in the last two comments (1530/32 at 1504, 1026/42 at 1000, 6800/7 at 6774) are the current ones.

---

openwrt-ai (commented):
## Commit checks

- `d2e33fa9` "realtek: eth: support changing the MTU on RTL93xx" — the subject scopes the change to RTL93xx, but the diff wires `.ndo_change_mtu = rteth_change_mtu` and `.max_mtu` into all four families (`rteth_838x_netdev_ops` :1654, `rteth_839x_netdev_ops` :1704, `rteth_930x_netdev_ops` :1753, `rteth_931x_netdev_ops` :1803). Only `.set_hol` is RTL93xx-specific. The body gets this right ("the operation ... skips it where a family carries none"), so it is the subject that is stale after the merge into one `change_mtu` — dropping the "on RTL93xx" would match what the commit does.

Re-reviewed after the rebase from 3 commits (`5f2665a8`/`7f6e3961`/`c82f09d4`) to 2. Spot-checked and found correct: the per-family `max_mtu` values against the DSA `rmon_ranges` in `rtl83xx/rtl8{38,39}x.c` and `rtl9{30,31}x.c` (10000 / 12288 / 12288 / 12288), the 6774 fragment ceiling (`17 * 400 - 26`), the 25/32 packets-per-ring arithmetic, `0x326c + 28 * 64 = 0x396c` for the RTL930x port register against the existing `MAC_L2_PORT_CTRL` stride, and that all the new offsets fall inside the `0x1b000000 0x10000` switch syscon range with no collision against in-tree defines. The earlier concern about unsynchronised read-modify-write on `dma_if_ctrl` does not apply — `regmap_update_bits_base()` takes `map->lock` around the whole sequence.

No CI failures on this head (no checks reported).

---

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
plappermaul: add this gare

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
plappermaul: Sync the commit message to the real changes.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: Done: gated on `netif_running()`, with the `WRITE_ONCE(dev->mtu, mtu)` kept above it so a
change made while the interface is down is picked up when it comes up.

On the RTL9303: raising the MTU to 1600 while eth0 is down leaves both length registers at
1530 and the budget at 32 packets and logs nothing, while the following `up` programs 1626
and 25. The boot log now carries no warning at all.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: Done: the subject drops "on RTL93xx", since only `set_hol()` is RTL93xx specific while
`ndo_change_mtu` and `max_mtu` are wired on all four families, and the body now covers the
gate and what it means for the boot log. The PR title and description are rewritten against
this head as well.

l0rdg3x:
Fifth round: the `netif_running()` gate, and both messages synced to what the series does.

Re-tested on the RTL9303 afterwards. The limits and the budget still track every MTU
(1530/32 at 1504, 1026/42 at 1000, 6800/7 at 6774), a 1046 byte frame still passes at 1504,
stops at 1000 and passes again, and an accepted change logs one warning where a refused one
logs none. New with the gate: an MTU of 1600 set while eth0 is down leaves the registers at
1530 and the budget at 32 and logs nothing, the following `up` programs 1626 and 25, and the
boot log carries no warning any more. All four subtargets compile.

l0rdg3x:
Three open points I would rather settle with you than guess at.

1. **The budget on an MTU decrease.** Going down it grows while the ring may still hold frames of
   the previous size, so it allows more than the ring covers under the driver's own worst case
   assumption: 6774 -> 1504 gives 7x17 + 25x4 = 219 fragments for 128 descriptors, 1504 -> 1000
   gives 158. The order I use narrows that window rather than closing it, and the warning covers
   it. Keep it as is, or skip `set_hol()` on the way down and let the next `up` take the larger
   budget, which closes the window and costs head of line capacity at small MTUs?

2. **Where the fragment ceiling belongs.** The check counts the fragment carrying the skb head
   against `MAX_SKB_FRAGS` and stops at an MTU of 6774, while the receive path assembles a head
   plus `MAX_SKB_FRAGS` fragments, i.e. 7174. Keep the conservative bound or raise it?

3. **RTL838x/839x/931x.** Those now program the new registers at every interface open, and I only
   have an RTL9303. Would you rather wait for someone with that hardware, or are the register maps
   plus the build enough for you here?

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -974,6 +1039,38 @@ static int rteth_stop(struct net_device *dev)
...
+	 * A larger MTU buys a smaller budget, so hand the rings their new limit before the
+	 * switch may deliver larger frames, and the other way around when the MTU shrinks.
+	 */
+	if (grow && ctrl->r->set_hol)
+		ctrl->r->set_hol(ctrl);
+
+	rteth_set_max_packet_length(ctrl);
+
+	if (!grow && ctrl->r->set_hol)
+		ctrl->r->set_hol(ctrl);
```
openwrt-ai: This is the only place in the file that programs these registers without holding `ctrl->lock`, and the ordering the comment describes is not actually atomic against `rteth_tx_timeout()`.

Every other hardware-programming path takes the lock around the whole sequence: [`rteth_open()` :926](``https://github.com/openwrt/openwrt/blob/f7326704b16751984c201763282dddbfcdf7030b/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L926``) wraps `hw_reset` (which reaches `rteth_93xx_set_hol()`) and `hw_en_rxtx` (which reaches `rteth_set_max_packet_length()`), and [`rteth_tx_timeout()` :1165](``https://github.com/openwrt/openwrt/blob/f7326704b16751984c201763282dddbfcdf7030b/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L1165``) wraps `hw_stop` → ring teardown/re-setup → `hw_en_rxtx`. Even `rteth_hw_set_mac()` takes it for plain `regmap_write()`s.

`ndo_tx_timeout` is not serialised against `ndo_change_mtu`: it is called straight from the watchdog timer under `dev->tx_global_lock`, not rtnl ([`dev_watchdog()` at sch_generic.c:543](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/net/sched/sch_generic.c#L543)). So the recovery sequence can land between the two writes here — the rings can be torn down and re-enabled while the budget is programmed for one MTU and the length register for the other, which is exactly the ring-overflow state `rteth_93xx_set_hol()`'s comment says must be avoided. regmap serialises each individual access, but not the pair.

```suggestion
	/*
	 * A larger MTU buys a smaller budget, so hand the rings their new limit before the
	 * switch may deliver larger frames, and the other way around when the MTU shrinks.
	 * Hold the lock, so rteth_tx_timeout() cannot reprogram the same registers in between.
	 */
	scoped_guard(spinlock_irqsave, &ctrl->lock) {
		if (grow && ctrl->r->set_hol)
			ctrl->r->set_hol(ctrl);

		rteth_set_max_packet_length(ctrl);

		if (!grow && ctrl->r->set_hol)
			ctrl->r->set_hol(ctrl);
	}
```

---

openwrt-ai (commented):
Re-reviewed after the rewrite from `2049a3b0`/`d2e33fa9` to `321fe6df`/`f7326704`. The code delta is small (the `netif_running()` gate plus a reworded comment block) and both commit messages now match their diffs — the RTL93xx scope is gone from the second subject, and the commit split is clean (`set_max_packet_length` + the register defines in commit 1, `set_hol`/`max_mtu`/`ndo_change_mtu` in commit 2).

Spot-checked and found correct on this head:

- The `netif_running()` gate holds: `rteth_open()` reaches `rteth_93xx_set_hol()` via `hw_reset` and `rteth_set_max_packet_length()` via `hw_en_rxtx`, so a change made while down is picked up on the next open.
- The new comment block's third bullet now covers Maple's separate `DMA_IF_PKT_TX_FLTR_CTRL`, and the "second field of the register above" reading matches the four setters (`GENMASK(27, 14)` on 839x/930x/931x, none on the 930x CPU register).
- Commit 1's U-Boot readings are internally consistent: `0x396c` = `0x326c + 28 * 64`, `0x0c003000` decodes to 12288 in both 14-bit fields, `0x06400000` to a truncation length of 1600, and `1504 + 26 = 1530`.
- The fragment ceiling is conservative by exactly one fragment as the PR body claims: `rteth_append_skb()` refuses at `nr_frags >= MAX_SKB_FRAGS` on top of the head fragment, while the check here counts the head against `MAX_SKB_FRAGS`.
- The `.max_mtu` values (10000/12288/12288/12288 less the 26 byte overhead) and the 32/42/7 packets-per-ring figures in the PR table recompute correctly.

The PR description is in sync with this head now — the earlier drift is resolved. No CI failures on this head (no checks reported).

---

plappermaul:
> Three open points I would rather settle with you than guess at.
> 
> 1. **The budget on an MTU decrease.** Going down it grows while the ring may still hold frames of
>    the previous size, so it allows more than the ring covers under the driver's own worst case
>    assumption: 6774 -> 1504 gives 7x17 + 25x4 = 219 fragments for 128 descriptors, 1504 -> 1000
>    gives 158. The order I use narrows that window rather than closing it, and the warning covers
>    it. Keep it as is, or skip `set_hol()` on the way down and let the next `up` take the larger
>    budget, which closes the window and costs head of line capacity at small MTUs?

For now the warning is enough. Both directions have their own caveats.

> 2. **Where the fragment ceiling belongs.** The check counts the fragment carrying the skb head
>    against `MAX_SKB_FRAGS` and stops at an MTU of 6774, while the receive path assembles a head
>    plus `MAX_SKB_FRAGS` fragments, i.e. 7174. Keep the conservative bound or raise it?

I don't get what is meant here.

> 3. **RTL838x/839x/931x.** Those now program the new registers at every interface open, and I only
>    have an RTL9303. Would you rather wait for someone with that hardware, or are the register maps
>    plus the build enough for you here?

Will test on the other targets before merging.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
plappermaul: Take the lock as requested. I didn't check if irqsave is really needed but it is the safest variant. Please test if this gives any side effects (should not).

l0rdg3x: Done. Tested on the RTL9303 across four boots: with the lock the registers and the ring budget
land on the same values as without it (1530 and 32 packets/ring at MTU 1504, 1026/42 at 1000,
6800/7 at 6774), the gate still defers a change made while the interface is down, and 60
alternating changes between 1000 and 6774 complete in 3.4 s with nothing logged.

I also built with `CONFIG_DEBUG_ATOMIC_SLEEP=y` to check that the lock does not end up covering
something that sleeps. It does not: `regmap_mmio` sets `.fast_io`, so the syscon regmap takes a
spinlock, and neither setter does anything besides register access.

l0rdg3x:
Sorry, point 2 was badly phrased. The check in `rteth_change_mtu()` is

```c
DIV_ROUND_UP(mtu + RTETH_FRAME_OVERHEAD, SKB_FRAG_SIZE) > MAX_SKB_FRAGS
```

so it counts every fragment the frame occupies, including the one that ends up as the skb
head. But `rteth_create_skb()` puts that first fragment in the linear area and
`rteth_append_skb()` then adds up to `MAX_SKB_FRAGS` more, so a frame may span 18 fragments
of 400 bytes = 7200, an MTU of 7174. The check therefore also refuses 6775..7174, which the
receive path would assemble fine. Keep the margin, or count against `MAX_SKB_FRAGS + 1`?

plappermaul:
A SKB consists of of a linear data part plus fragments. So you must account for that part in your calculation. Without division like

if (mtu + RTETH_FRAME_OVERHEAD > (MAX_SKB_FRAGS + 1) * SKB_FRAG_SIZE) ....

l0rdg3x:
Done, the check is now

```c
if (mtu + RTETH_FRAME_OVERHEAD > (MAX_SKB_FRAGS + 1) * SKB_FRAG_SIZE)
```

so the ceiling moves from 6774 to 7174. On the RTL9303, byte exact: 7174 is accepted and
programs a length of 7200 with 7 packets per ring, 7175 is refused. Everything else is
unchanged (1530 and 32 at MTU 1504, 1026/42 at 1000), and the commit message follows the
new numbers.

openwrt-ai (commented):

The previous round's commit check is resolved: the subject is now `realtek: eth: support changing the MTU` and the body describes the ordering, the lock and the `netif_running()` gate that the diff actually implements. Both messages match their changes.

Spot-checked in the delta since `d2e33fa9`:

- `mtu + RTETH_FRAME_OVERHEAD > (MAX_SKB_FRAGS + 1) * SKB_FRAG_SIZE` matches the receive path: [`rteth_create_skb()`](``https://github.com/openwrt/openwrt/blob/fdfbba971cfca82054011a2c103d86af4b515fab/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L1275-L1296)`` puts the head fragment in the linear area (`skb_put(skb, len)` with `len <= SKB_FRAG_SIZE` enforced at [rtl838x_eth.c:1367](``https://github.com/openwrt/openwrt/blob/fdfbba971cfca82054011a2c103d86af4b515fab/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L1367``)) and [`rteth_append_skb()`](``https://github.com/openwrt/openwrt/blob/fdfbba971cfca82054011a2c103d86af4b515fab/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L1324``) adds up to `MAX_SKB_FRAGS` more, so 18 × 400 = 7200 and an MTU ceiling of 7174 is exact.
- The `netif_running()` gate is safe: `rteth_open()` calls `hw_reset()` → `rteth_93xx_set_hol()` and `hw_en_rxtx()` → `rteth_set_max_packet_length()`, both reading `ctrl->dev->mtu`, so a change made while down is re-derived on the next open rather than lost.
- `scoped_guard(spinlock_irqsave, &ctrl->lock)` matches how [`rteth_open()`](``https://github.com/openwrt/openwrt/blob/fdfbba971cfca82054011a2c103d86af4b515fab/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L926``) and [`rteth_tx_timeout()`](``https://github.com/openwrt/openwrt/blob/fdfbba971cfca82054011a2c103d86af4b515fab/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L1169``) already serialise the same registers, and no callee re-takes the lock, so there is no recursion.
- `cnt = min(RTETH_RX_RING_SIZE / frags_per_pkt, 0x3ff)` cannot reach 0 now that `frags_per_pkt` is bounded by 18 against a 128-entry ring.

The `dev->max_mtu` / 7174 mismatch and the widened window on an MTU decrease are both stated in the PR body and settled in the thread, so not re-raised.

---

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -477,27 +481,36 @@ static void rteth_839x_hw_reset(struct rteth_ctrl *ctrl)
...
-static void rteth_93xx_hw_reset(struct rteth_ctrl *ctrl)
+static void rteth_93xx_set_hol(struct rteth_ctrl *ctrl)
 {
 	/*
 	 * The counter registers track the number of packets that are allowed to be appended to
 	 * the ring buffer. On RTL93xx a ring overflow must be avoided at all cost. Defensively
 	 * calculate the space by anticipating that only fully filled packets are received.
 	 */
-	int max_frame_size = SKB_MTU + ETH_HLEN + ETH_FCS_LEN + VLAN_HLEN + 4;
+	int max_frame_size = ctrl->dev->mtu + RTETH_FRAME_OVERHEAD;
```
plappermaul, line 491: Looks good on RTL838x(RTL839x. As we invent set_hol() here we should cover the 83xx part too. See above for setup of free floating rings.

`
regmap_write(ctrl->map, ctrl->r->dma_if_rx_ring_size, 0);
`

So:
- Add rteth_83xx_set_hol() with that single statement
- Add rteth_83xx_set_hol() to RTL838x/RTL839x config
- Remove below checks "if (ctrl->r->set_hol)" as we always have a function
- Do not place set_hol in hw_reset() but call ctrl->r-set_hol() always after ctrl->r->hw_reset()

Do a retest afterwards.

Thanks.

plappermaul (changes requested):

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
l0rdg3x: Done, all four points: `rteth_83xx_set_hol()` with that single write, wired into the RTL838x and
RTL839x configs, the `if (ctrl->r->set_hol)` checks are gone, and `set_hol()` now runs right after
`hw_reset()` in `rteth_open()` instead of inside it. The two RTL83xx reset paths lost the ring
size write, since that is what the new function does.

Retested on the RTL9303. Moving the call out of `hw_reset()` reverses its order against the
counter clearing on RTL93xx, so I checked the budget specifically: after `down` -> `mtu 1600` ->
`up` the rings read 25, which only `set_hol()` at open can program. The rest is unchanged (1530
and 32 at MTU 1504, 1026/42 at 1000, 7200/7 at 7174, 7175 refused, boot log still clean).

On RTL838x/839x the sequence is byte for byte what it was, as that write was already the last
thing their `hw_reset()` did; compile-tested there and on RTL931x.

plappermaul:

l0rdg3x:
Thank you! ;)

PR #24422, opened 2026-07-25
ixp4xx-microcode: use the linux-firmware images

linusw:
The IXP4xx firmware is now in linux-firmware so use these images instead and drop the special microcode builder package.

Drop the microcode package, create equally named firmware packages.


package/firmware/linux-firmware/ixp4xx.mk
```diff
@@ -0,0 +1,27 @@
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Copyright (C) 2007-2026 OpenWrt.org
+
+# Ethernet or WAN support, pick one
+
+Package/ixp4xx-microcode-ethernet = $(call Package/firmware-default,IXP4xx ethernet firmware,,LICENSE.ixp4xx)
```
openwrt-ai: The `DEPENDS:=@TARGET_ixp4xx` and `PKG_FLAGS:=nonshared` that the removed `package/firmware/ixp4xx-microcode/Makefile` carried are dropped here (2nd and 5th argument of `Package/firmware-default` are empty). Both packages therefore become selectable in menuconfig and get built into the package feed of every target, even though the NPE firmware is only usable on ixp4xx. The in-tree pattern for target-restricted firmware in this package is [mediatek.mk:100](https://github.com/openwrt/openwrt/blob/f758c7f3327ab336ac75a8f317345e65f0d13afd/package/firmware/linux-firmware/mediatek.mk#L100), which passes both.

```suggestion
Package/ixp4xx-microcode-ethernet = $(call Package/firmware-default,IXP4xx ethernet firmware,@TARGET_ixp4xx,LICENSE.ixp4xx,,nonshared)
```

Same applies to the `Package/ixp4xx-microcode-wan` line below.

---

package/firmware/linux-firmware/ixp4xx.mk
```diff
@@ -0,0 +1,27 @@
...
+		$(PKG_BUILD_DIR)/ixp4xx/NPE-C \
+		$(1)/lib/firmware/
+endef
+$(eval $(call BuildPackage,ixp4xx-microcode-ethernet))
+
+Package/ixp4xx-microcode-wan = $(call Package/firmware-default,IXP4xx WAN firmware,,LICENSE.ixp4xx)
+define Package/ixp4xx-microcode-wan/install
+	$(INSTALL_DIR) $(1)/lib/firmware/
+	$(INSTALL_DATA) \
+		$(PKG_BUILD_DIR)/ixp4xx/NPE-A-HSS \
```
openwrt-ai, line 22: This installs the HSS image under the name `NPE-A-HSS`, but the kernel never requests that name: the HSS driver asks for `npe_name(port->npe)` — i.e. plain `NPE-A` — at [ixp4xx_hss.c:524](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/net/wan/ixp4xx_hss.c#L524) (`NPE_A_FIRMWARE "NPE-A"` in [ixp4xx-npe.c:121](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/soc/ixp4xx/ixp4xx-npe.c#L121)). Since the wan package installs no `NPE-A` either, `/lib/firmware/NPE-A` is missing entirely when it is selected. The name was inherited from the old builder (which used `NPE-A-HSS` only as a disambiguating symlink), so this looks like a pre-existing bug carried over — but since this is a fresh install rule, would it make sense to install `ixp4xx/NPE-A-HSS` as `$(1)/lib/firmware/NPE-A` here?

---

package/firmware/linux-firmware/ixp4xx.mk
```diff
@@ -0,0 +1,27 @@
+# SPDX-License-Identifier: GPL-2.0-only
```
openwrt-ai, line 1: nit: with `package/firmware/ixp4xx-microcode/` gone, the glob `- "package/firmware/ixp4xx-microcode/**"` in [.github/labeler.yml:94](https://github.com/openwrt/openwrt/blob/f758c7f3327ab336ac75a8f317345e65f0d13afd/.github/labeler.yml#L94) can no longer match anything; it could be pointed at `package/firmware/linux-firmware/ixp4xx.mk` instead so ixp4xx firmware changes keep getting the `target/ixp4xx` label.

---

openwrt-ai (commented):
Reviewed 1 new commit; 3 inline comments. Verified against the linux-firmware upstream tree: `ixp4xx/{NPE-A,NPE-A-HSS,NPE-B,NPE-C}` and `LICENSE.ixp4xx` exist and the blobs are big-endian (magic `fe ed f0 0d`), matching what the dropped builder produced with `-be`.

---

package/firmware/linux-firmware/ixp4xx.mk
```diff
@@ -0,0 +1,30 @@
...
+		$(PKG_BUILD_DIR)/ixp4xx/NPE-B \
+		$(PKG_BUILD_DIR)/ixp4xx/NPE-C \
+		$(1)/lib/firmware/
+endef
+$(eval $(call BuildPackage,ixp4xx-microcode-ethernet))
+
+Package/ixp4xx-microcode-wan = $(call Package/firmware-default,IXP4xx WAN firmware,@TARGET_ixp4xx,LICENSE.ixp4xx,,nonshared)
+define Package/ixp4xx-microcode-wan/install
+	$(INSTALL_DIR) $(1)/lib/firmware/
+
```
openwrt-ai: nit: stray blank line inside the install recipe — it breaks the parallelism with the `ixp4xx-microcode-ethernet/install` block above (line 9-14), and it is the only blank line inside a `Package/*/install` define in the whole `package/firmware/linux-firmware/` directory.

```suggestion
	$(INSTALL_DIR) $(1)/lib/firmware/
```

---

openwrt-ai (commented):
Re-reviewed the force-pushed commit (`f758c7f` → `564f100`). All three points from the previous review are addressed:

- `@TARGET_ixp4xx` and `nonshared` are now passed as args 2 and 5 of `Package/firmware-default`, which map to `DEPENDS` and `PKGFLAGS` — matches the semantics the dropped `package/firmware/ixp4xx-microcode/Makefile` had.
- `ixp4xx/NPE-A-HSS` is now installed as `/lib/firmware/NPE-A`. Confirmed against the kernel the target builds (6.18, per `target/linux/ixp4xx/Makefile` `KERNEL_PATCHVER:=6.18`): the HSS driver requests `npe_name(port->npe)` at [ixp4xx_hss.c:524](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/net/wan/ixp4xx_hss.c#L524), and `npe_names[]` is `{"NPE-A","NPE-B","NPE-C"}` at [ixp4xx-npe.c:121-125](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/soc/ixp4xx/ixp4xx-npe.c#L121-L125), so `NPE-A` is the name that actually gets requested.
- `.github/labeler.yml` now points at `package/firmware/linux-firmware/ixp4xx.mk`, and the commit message documents that hunk.

`PKG_RELEASE` bump on `package/firmware/linux-firmware/Makefile` is present. CI on `564f100` is green. One whitespace nit inline.

---

package/firmware/linux-firmware/ixp4xx.mk
```diff
@@ -0,0 +1,29 @@
...
+	$(INSTALL_DIR) $(1)/lib/firmware/
+	$(INSTALL_DATA) \
+		$(PKG_BUILD_DIR)/ixp4xx/NPE-A \
+		$(PKG_BUILD_DIR)/ixp4xx/NPE-B \
+		$(PKG_BUILD_DIR)/ixp4xx/NPE-C \
+		$(1)/lib/firmware/
+endef
+$(eval $(call BuildPackage,ixp4xx-microcode-ethernet))
+
+Package/ixp4xx-microcode-wan = $(call Package/firmware-default,IXP4xx WAN firmware,@TARGET_ixp4xx,LICENSE.ixp4xx,,nonshared)
```
openwrt-ai, line 18: nit: the header comment says "pick one", but nothing stops both packages from being selected, and after the `NPE-A-HSS` → `NPE-A` rename the two now install the exact same three paths (`/lib/firmware/NPE-A`, `NPE-B`, `NPE-C`) with different `NPE-A` content — so installing both is a hard file conflict rather than just redundant. The old `package/firmware/ixp4xx-microcode/Makefile` had the same gap (it only overlapped on `NPE-B`/`NPE-C`), so this isn't a regression, but since the packaging is being rewritten anyway: would it make sense to declare the mutual exclusion, the way in-tree mutually exclusive firmware packages do at [ath10k-ct-firmware/Makefile:164](https://github.com/openwrt/openwrt/blob/52f71aeb1e22ea9e310128786a75e1e1123d984f/package/firmware/ath10k-ct-firmware/Makefile#L164)? `Package/firmware-default` has no `CONFLICTS` argument, so it would need appending to one of the two package definitions.

---

openwrt-ai (commented):
Re-reviewed the force-pushed commit (`564f100` → `52f71ae`). The only content change since the last review is the removal of the stray blank line in `Package/ixp4xx-microcode-wan/install` (confirmed with `git range-diff`), which resolves the previous nit. Everything else is a rebase onto current `main`.

Re-checked the rest of the diff against the rebased tree:
- Package names `ixp4xx-microcode-ethernet` / `ixp4xx-microcode-wan` are unchanged, so all 10 `DEVICE_PACKAGES` references in `target/linux/ixp4xx/image/Makefile` still resolve.
- `SCAN_DEPS = *.mk` + `include $(wildcard ./*.mk)` in `package/firmware/linux-firmware/Makefile` picks the new file up; `LICENSE.ixp4xx` is passed as arg 3, which maps to `LICENSE_FILES`, matching every other `.mk` in the directory.
- `PKG_RELEASE` bump is still present, and `.github/labeler.yml` points at the new path.
- No failing CI checks on `52f71ae`.

One nit inline.

---

PR #24425, opened 2026-07-26
netifd: dhcp: add vendorid_hex option for raw hex option 60

zcy85611:
Allow users to specify raw hex strings directly for DHCP Option 60 using a new 'vendorid_hex' UCI option. This avoids double hex encoding when ISPs require pre-encoded raw hexadecimal binary payloads.

### DIFF
--- a/package/network/config/netifd/files/lib/netifd/proto/dhcp.sh
+++ b/package/network/config/netifd/files/lib/netifd/proto/dhcp.sh
@@ -17,6 +17,7 @@ proto_dhcp_init_config() {
 	proto_config_add_string clientid
 	proto_config_add_string sendclientid
 	proto_config_add_string vendorid
+  proto_config_add_string vendorid_hex
 	proto_config_add_boolean 'broadcast:bool'
 	proto_config_add_boolean 'norelease:bool'
 	proto_config_add_string 'reqopts:list(string)'
@@ -59,8 +60,8 @@ proto_dhcp_setup() {
 	local config="$1"
 	local iface="$2"
 
-	local ipaddr hostname clientid sendclientid vendorid broadcast norelease reqopts defaultreqopts iface6rd sendopts delegate zone6rd zone mtu6rd customroutes classlessroute timeout retry tryagain
-	json_get_vars ipaddr hostname clientid sendclientid vendorid broadcast norelease reqopts defaultreqopts iface6rd delegate zone6rd zone mtu6rd customroutes classlessroute timeout retry tryagain
+	local ipaddr hostname clientid sendclientid vendorid vendorid_hex broadcast norelease reqopts defaultreqopts iface6rd sendopts delegate zone6rd zone mtu6rd customroutes classlessroute timeout retry tryagain
+	json_get_vars ipaddr hostname clientid sendclientid vendorid vendorid_hex broadcast norelease reqopts defaultreqopts iface6rd delegate zone6rd zone mtu6rd customroutes classlessroute timeout retry tryagain
 
 	local opt dhcpopts
 	for opt in $reqopts; do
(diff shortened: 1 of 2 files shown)


package/network/config/netifd/files/lib/netifd/proto/dhcp.sh
```diff
@@ -97,6 +98,11 @@ proto_dhcp_setup() {
 			;;
 	esac
 	[ -n "${clientid##-C}" ] && clientid="-x 0x3d:$clientid"
+  if [ -n "$vendorid_hex" ]; then
+		append dhcpopts "-x 0x3c:$vendorid_hex"
+	elif [ -n "$vendorid" ]; then
+		append dhcpopts "-x 0x3c:$(echo -n "$vendorid" | hexdump -ve '1/1 "%02x"')"
+	fi
 	[ -n "$vendorid" ] && append dhcpopts "-x 0x3c:$(echo -n "$vendorid" | hexdump -ve '1/1 "%02x"')"
```
openwrt-ai, line 106: The pre-existing `vendorid` line was not removed, so the new `if/elif` block does not replace it — it runs in addition to it. With only `vendorid` set, option 60 is now appended to `dhcpopts` twice (`-x 0x3c:<hex> -x 0x3c:<hex>`), which is a regression for every existing config using `vendorid`.

Two more problems in the same block:

- `vendorid_hex` is passed straight through to udhcpc without validation. An odd-length or non-hex value produces a malformed `-x 0x3c:` argument. The tree already has a helper for exactly this — [`hexdump_2hex()` in `functions.sh:21-24`](https://github.com/openwrt/openwrt/blob/025477ac9e306dcf19062383417cde0a36491785/package/base-files/files/lib/functions.sh#L21-L24) — and this same file already uses it for `clientid` and `duid`, including the "ignoring invalid ... value" warning. `vendorid_hex` should follow that pattern.
- The `if` line is indented with two spaces; the rest of the file uses tabs.

```suggestion
	if [ -n "$vendorid_hex" ]; then
		vendorid_hex="$(hexdump_2hex "$vendorid_hex")"
		[ -z "$vendorid_hex" ] && logger -p warn -t dhcp "$iface: ignoring invalid vendorid_hex value"
	fi
	if [ -n "$vendorid_hex" ]; then
		append dhcpopts "-x 0x3c:$vendorid_hex"
	elif [ -n "$vendorid" ]; then
		append dhcpopts "-x 0x3c:$(echo -n "$vendorid" | hexdump -ve '1/1 "%02x"')"
	fi
```


---

package/network/config/netifd/files/lib/netifd/proto/dhcp.sh
```diff
@@ -17,6 +17,7 @@ proto_dhcp_init_config() {
 	proto_config_add_string clientid
 	proto_config_add_string sendclientid
 	proto_config_add_string vendorid
+  proto_config_add_string vendorid_hex
```
openwrt-ai: Indented with two spaces; every other line in this function uses a tab.

```suggestion
	proto_config_add_string vendorid_hex
```


---

package/network/config/netifd/files/lib/netifd/proto/dhcp.sh
openwrt-ai, line 1: nit: this commit also drops the executable bit on this file (`old mode 100755` / `new mode 100644`). That is unrelated to adding `vendorid_hex` and looks accidental — probably a local umask/checkout artifact. Please restore mode 100755 so the diff only contains the intended change.

---

package/network/config/netifd/files/lib/netifd/proto/dhcp.sh
```diff
@@ -59,8 +60,8 @@ proto_dhcp_setup() {
 	local config="$1"
 	local iface="$2"
 
-	local ipaddr hostname clientid sendclientid vendorid broadcast norelease reqopts defaultreqopts iface6rd sendopts delegate zone6rd zone mtu6rd customroutes classlessroute timeout retry tryagain
-	json_get_vars ipaddr hostname clientid sendclientid vendorid broadcast norelease reqopts defaultreqopts iface6rd delegate zone6rd zone mtu6rd customroutes classlessroute timeout retry tryagain
+	local ipaddr hostname clientid sendclientid vendorid vendorid_hex broadcast norelease reqopts defaultreqopts iface6rd sendopts delegate zone6rd zone mtu6rd customroutes classlessroute timeout retry tryagain
```
openwrt-ai, line 63: Is a new UCI option needed here at all? The existing `sendopts` list already passes raw values verbatim to udhcpc via [`proto_dhcp_add_sendopts()`](``https://github.com/openwrt/openwrt/blob/025477ac9e306dcf19062383417cde0a36491785/package/network/config/netifd/files/lib/netifd/proto/dhcp.sh#L36-L38),`` so `list sendopts '0x3c:<rawhex>'` already achieves "pre-encoded raw hex for option 60" without double encoding. The [`# Avoid sending duplicate Option 60 values`](``https://github.com/openwrt/openwrt/blob/025477ac9e306dcf19062383417cde0a36491785/package/network/config/netifd/files/lib/netifd/proto/dhcp.sh#L111-L118``) block right below was added specifically to make that path work.

If `vendorid_hex` is meant to cover a case `sendopts` cannot, please say so in the commit message — otherwise this adds a second, overlapping way to set the same option.

---

openwrt-ai (commented):
Reviewed 1 new commit. The main issue is that the old `vendorid` line was left in place alongside the new `if/elif` block, so option 60 is appended twice for existing `vendorid` configs — details inline, together with hex validation, indentation, and the file-mode change.

---

PR #21450, opened 2026-01-08
netifd: dhcp: suppress udhcpc default vendor class if specified in sendopts

ihipop:
### Motivation

This PR provides an alternative solution to #21247, addressing the same IPTV authentication issues but with better backward compatibility and a cleaner implementation.

### Problem Statement

When connecting to ISP IPTV services via IPoE, two issues arise:

1. **Non-ASCII Option 60 values**: Some ISPs require DHCP Option 60 (Vendor Class Identifier) to contain raw hexadecimal data that doesn't translate to printable ASCII characters.
2. **Duplicate Option 60**: **The current implementation  of udhcpc** sends both the user-specified vendorid and udhcpc's default version string, causing authentication failures with some ISP DHCP servers.

### Solution

This implementation allows users to send hexadecimal Option 60 values via the existing [sendopts]

configuration option, while automatically providing an empty `-V ""` placeholder to prevent udhcpc from sending its default vendor class.

The solution intelligently detects Option 60 in **all supported udhcpc formats**:
- **Hexadecimal**: `0x3c`, `0X3C`, `0x003c` (case-insensitive with optional leading zeros)
- **Decimal**: `60`
- **Named**: `vendor`

**Key advantages over #21247:**

- ✅ **Full backward compatibility**: Existing `vendorid` configurations continue to work unchanged
- ✅ **Cleaner implementation**: Uses existing [sendopts] mechanism instead of modifying vendorid behavior
- ✅ **More flexible**: Users can choose between ASCII vendorid or hex sendopts based on their needs
- ✅ **Comprehensive format support**: Detects Option 60 in all three formats supported by udhcpc

### Changes

Modified [package/network/config/netifd/files/lib/netifd/proto/dhcp.sh]

- Added detection for Option 60 in `$dhcpopts` (populated by sendopts) supporting all udhcpc formats
- When Option 60 is detected via sendopts (in any format), automatically provide `-V ""` to suppress udhcpc's default vendor class
- When no Option 60 is present in sendopts, use the traditional `vendorid` parameter if configured

#### For IPTV with hex Option 60 (new functionality):
```bash
# Method 1: Hexadecimal format (recommended for non-ASCII values)
uci set network.iptv=interface
uci set network.iptv.proto='dhcp'
uci set network.iptv.device='eth1'
uci add_list network.iptv.sendopts='0x3c:48575443'  # Hex vendor class
uci commit network
ifup iptv

# Method 2: Decimal format
uci add_list network.iptv.sendopts='60:48575443'

[...]

PR #24427, opened 2026-07-26
netifd: dhcp: add vendorid_hex option for raw hex option 60

zcy85611:
Currently, the `vendorid` option in `dhcp.sh` treats input values as standard ASCII strings and converts each character into its hex representation using `hexdump`. 

However, some ISPs require custom DHCP Option 60 values sent in raw hexadecimal format (e.g., binary payload or pre-encoded hex sequence like `00001f39014d...`). Passing such raw hex strings into the existing `vendorid` option results in double hex encoding (e.g., ASCII `'0'` converted to `30`).

This PR introduces a new UCI option `vendorid_hex` for the DHCP protocol handler:
- Allows users to specify raw hex strings directly for Option 60 (`-x 0x3c:<hex>`).
- Retains backwards compatibility with the existing string-based `vendorid` option.
- Handles conflict resolution by prioritizing `vendorid_hex` over `vendorid` if both are specified.
- Maintains compatibility with the existing duplicate Option 60 suppression logic (`emptyvendorid=1`).

Once this PR is merged, I will submit a corresponding PR to the openwrt/luci repository to add support for the vendorid_hex option in the Web UI.

### DIFF
--- a/package/network/config/netifd/files/lib/netifd/proto/dhcp.sh
+++ b/package/network/config/netifd/files/lib/netifd/proto/dhcp.sh
@@ -17,6 +17,7 @@ proto_dhcp_init_config() {
 	proto_config_add_string clientid
 	proto_config_add_string sendclientid
 	proto_config_add_string vendorid
+	proto_config_add_string vendorid_hex
 	proto_config_add_boolean 'broadcast:bool'
 	proto_config_add_boolean 'norelease:bool'
 	proto_config_add_string 'reqopts:list(string)'
@@ -59,8 +60,8 @@ proto_dhcp_setup() {
 	local config="$1"
 	local iface="$2"
 
-	local ipaddr hostname clientid sendclientid vendorid broadcast norelease reqopts defaultreqopts iface6rd sendopts delegate zone6rd zone mtu6rd customroutes classlessroute timeout retry tryagain
-	json_get_vars ipaddr hostname clientid sendclientid vendorid broadcast norelease reqopts defaultreqopts iface6rd delegate zone6rd zone mtu6rd customroutes classlessroute timeout retry tryagain
+	local ipaddr hostname clientid sendclientid vendorid vendorid_hex broadcast norelease reqopts defaultreqopts iface6rd sendopts delegate zone6rd zone mtu6rd customroutes classlessroute timeout retry tryagain
+	json_get_vars ipaddr hostname clientid sendclientid vendorid vendorid_hex broadcast norelease reqopts defaultreqopts iface6rd delegate zone6rd zone mtu6rd customroutes classlessroute timeout retry tryagain
 
 	local opt dhcpopts
 	for opt in $reqopts; do
@@ -97,7 +98,17 @@ proto_dhcp_setup() {
 			;;
 	esac
 	[ -n "${clientid##-C}" ] && clientid="-x 0x3d:$clientid"
-	[ -n "$vendorid" ] && append dhcpopts "-x 0x3c:$(echo -n "$vendorid" | hexdump -ve '1/1 "%02x"')"
+	[ -n "$vendorid_hex" ] && {
+		vendorid_hex="$(hexdump_2hex "$vendorid_hex")"
+		[ -z "$vendorid_hex" ] && {
+			logger -p warn -t dhcp "$iface: ignoring invalid vendorid_hex value"
+		}
+	}
+	if [ -n "$vendorid_hex" ]; then
+		append dhcpopts "-x 0x3c:$vendorid_hex"
+	elif [ -n "$vendorid" ]; then
+		append dhcpopts "-x 0x3c:$(echo -n "$vendorid" | hexdump -ve '1/1 "%02x"')"
+	fi
 	[ -n "$iface6rd" ] && proto_export "IFACE6RD=$iface6rd"
 	[ "$iface6rd" != 0 -a -f /lib/netifd/proto/6rd.sh ] && append dhcpopts "-O 212"
 	[ -n "$zone6rd" ] && proto_export "ZONE6RD=$zone6rd"
--- a/package/network/config/netifd/Makefile
+++ b/package/network/config/netifd/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=netifd
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL=$(PROJECT_GIT)/project/netifd.git


package/network/config/netifd/files/lib/netifd/proto/dhcp.sh
```diff
@@ -97,6 +98,11 @@ proto_dhcp_setup() {
 			;;
 	esac
 	[ -n "${clientid##-C}" ] && clientid="-x 0x3d:$clientid"
+	if [ -n "$vendorid_hex" ]; then
+    append dhcpopts "-x 0x3c:$vendorid_hex"
+	elif [ -n "$vendorid" ]; then
+    append dhcpopts "-x 0x3c:$(echo -n "$vendorid" | hexdump -ve '1/1 "%02x"')"
+	fi
 	[ -n "$vendorid" ] && append dhcpopts "-x 0x3c:$(echo -n "$vendorid" | hexdump -ve '1/1 "%02x"')"
```
openwrt-ai: The pre-existing `vendorid` line was not removed, so it now runs in addition to the new `if`/`elif` block. With only `vendorid` set, option 60 is appended twice (`-x 0x3c:...` twice on the udhcpc command line); with both `vendorid_hex` and `vendorid` set, the raw hex *and* the encoded string are sent, which contradicts the "prioritizing `vendorid_hex` over `vendorid`" behaviour described in the PR body. The `elif` branch above already covers this case, so this line should go:

```suggestion
```

---

package/network/config/netifd/files/lib/netifd/proto/dhcp.sh
```diff
@@ -97,6 +98,11 @@ proto_dhcp_setup() {
 			;;
 	esac
 	[ -n "${clientid##-C}" ] && clientid="-x 0x3d:$clientid"
+	if [ -n "$vendorid_hex" ]; then
+    append dhcpopts "-x 0x3c:$vendorid_hex"
+	elif [ -n "$vendorid" ]; then
+    append dhcpopts "-x 0x3c:$(echo -n "$vendorid" | hexdump -ve '1/1 "%02x"')"
+	fi
```
openwrt-ai: The two `append` lines are indented with 4 spaces while the rest of this file (and every other proto script) uses tabs:

```suggestion
	if [ -n "$vendorid_hex" ]; then
		append dhcpopts "-x 0x3c:$vendorid_hex"
	elif [ -n "$vendorid" ]; then
		append dhcpopts "-x 0x3c:$(echo -n "$vendorid" | hexdump -ve '1/1 "%02x"')"
	fi
```

---

package/network/config/netifd/files/lib/netifd/proto/dhcp.sh
```diff
@@ -97,6 +98,11 @@ proto_dhcp_setup() {
 			;;
 	esac
 	[ -n "${clientid##-C}" ] && clientid="-x 0x3d:$clientid"
+	if [ -n "$vendorid_hex" ]; then
+    append dhcpopts "-x 0x3c:$vendorid_hex"
```
openwrt-ai: `vendorid_hex` is passed straight through to the udhcpc argument without any validation. Every other user-supplied hex value in this file is run through `hexdump_2hex()` first and rejected with a warning when it does not parse — see [`clientid` at dhcp.sh:91](https://github.com/openwrt/openwrt/blob/f76b9ebbab601a4f86b63019ef736f0056db4587/package/network/config/netifd/files/lib/netifd/proto/dhcp.sh#L91) and [`dhcp_default_duid` at dhcp.sh:50](https://github.com/openwrt/openwrt/blob/f76b9ebbab601a4f86b63019ef736f0056db4587/package/network/config/netifd/files/lib/netifd/proto/dhcp.sh#L50), with the helper at [functions.sh:21](https://github.com/openwrt/openwrt/blob/f76b9ebbab601a4f86b63019ef736f0056db4587/package/base-files/files/lib/functions.sh#L21-L24).

Without it, an odd-length or non-hex value (e.g. a stray `0x` prefix, or a value pasted with a trailing newline) is handed to udhcpc unchecked, which either fails to parse the option or silently sends something other than what the user configured. Running it through `hexdump_2hex` would also give the option the same tolerance for `:`/`-`/space separators that `clientid` has, and logging the same `logger -p warn -t dhcp "$iface: ignoring invalid ... value"` message on failure would keep the diagnostics consistent.

---

package/network/config/netifd/files/lib/netifd/proto/dhcp.sh
openwrt-ai, line 1: The commit also changes this file's mode from `100755` to `100644` (`old mode 100755` / `new mode 100644` in the diff), which is not mentioned in the commit message and is unrelated to adding the option. All other proto scripts in this directory are `755`; please restore the executable bit (`chmod +x` and amend) so the diff only carries the intended change.

---

package/network/config/netifd/files/lib/netifd/proto/dhcp.sh
```diff
@@ -17,6 +17,7 @@ proto_dhcp_init_config() {
 	proto_config_add_string clientid
 	proto_config_add_string sendclientid
 	proto_config_add_string vendorid
+	proto_config_add_string vendorid_hex
```
openwrt-ai, line 20: nit: is a new UCI option needed here, or does the existing `sendopts` list already cover this use case? `proto_dhcp_add_sendopts` appends any list entry verbatim as `-x <entry>`, so `option sendopts '0x3c:00001f39014d'` already sends a raw-hex option 60 with no double encoding. The "Avoid sending duplicate Option 60 values" block further down explicitly matches `-x 0x3c:` / `-x 60:` / `-x vendor:` in `$dhcpopts`, which suggests raw option-60 injection via `sendopts` is the supported path today. If `vendorid_hex` is meant to be a more discoverable/LuCI-friendly spelling of that, it would help to say so in the commit message.

---

openwrt-ai (commented):
Reviewed 1 new commit. The main finding is that the original `vendorid` append line was left in place alongside the new `if`/`elif` block, so option 60 is emitted twice; details inline.

---

package/network/config/netifd/files/lib/netifd/proto/dhcp.sh
```diff
@@ -97,7 +98,16 @@ proto_dhcp_setup() {
...
+	if [ -n "$vendorid_hex" ]; then
+		vendorid_hex="$(hexdump_2hex "$vendorid_hex")"
+		if [ -n "$vendorid_hex" ]; then
+			append dhcpopts "-x 0x3c:$vendorid_hex"
+		else
+			logger -p warn -t dhcp "$iface: ignoring invalid vendorid_hex value"
+		fi
+	elif [ -n "$vendorid" ]; then
+		append dhcpopts "-x 0x3c:$(echo -n "$vendorid" | hexdump -ve '1/1 "%02x"')"
+	fi
```
openwrt-ai: The `elif` is selected on the *raw* `vendorid_hex` being non-empty, but the value is only rejected after `hexdump_2hex` has run. So if a user has both options set and the `vendorid_hex` value is malformed (odd length, stray `0x` prefix, non-hex char), the warning is logged and option 60 is dropped entirely — the perfectly valid `vendorid` string is never used, and `emptyvendorid` also stays unset so udhcpc falls back to sending its own built-in vendor class.

Is dropping to "no configured vendor ID at all" the intended failure mode? [The `clientid` path just above](https://github.com/openwrt/openwrt/blob/d05a95e40c9314eacb62ef2151dd226d8f001597/package/network/config/netifd/files/lib/netifd/proto/dhcp.sh#L90-L97) does the opposite: it warns on an unparseable value and then falls back to the default client ID rather than sending nothing. Falling through to the `vendorid` branch on a sanitisation failure would match that behaviour, e.g. by testing the sanitised value instead:

```sh
vendorid_hex="$(hexdump_2hex "$vendorid_hex")"
[ -z "$vendorid_hex" ] && logger -p warn -t dhcp "$iface: ignoring invalid vendorid_hex value"
if [ -n "$vendorid_hex" ]; then
	...
```

If failing closed is deliberate (so a typo never silently sends the wrong option 60 to the ISP), it would be worth a line in the commit message saying so.

---

openwrt-ai (commented):
Reviewed the force-pushed commit `d05a95e`. All five findings from the previous review are addressed: the duplicated `vendorid` append line is gone, the block is tab-indented, `vendorid_hex` now goes through `hexdump_2hex` with a matching `logger -p warn` message, the file mode is back to `100755`, and the commit message now explains why this is a first-class UCI option rather than `sendopts`. `PKG_RELEASE` bump is present and correct.

One question left on the `if`/`elif` fallback behaviour, inline.

---

openwrt-ai (commented):
Reviewed the force-pushed commit `8983919`. The remaining question from the previous review is resolved: the `if`/`elif` now tests the *sanitised* `vendorid_hex`, so a malformed value is warned about and then falls through to the string `vendorid` branch, matching the [`clientid` handling](https://github.com/openwrt/openwrt/blob/8983919b2ca6e13ee1dc883d6f0b2d4132084d89/package/network/config/netifd/files/lib/netifd/proto/dhcp.sh#L91-L97) a few lines above — same `[ -n ... ] && { hexdump_2hex; [ -z ... ] && { logger ... } }` shape and same warning wording. `PKG_RELEASE` bump, file mode, and the `emptyvendorid` duplicate-suppression interaction all still look correct, and the commit message matches the diff.

No new issues found.

---

zcy85611:
@hauke would you help to review this pull request please?thanks

zcy85611:
@Shine- @nbd168 would you help to look at this pull request please?

Shine-:
I see your motivation, but specifying a hex-based vendor class is already possible using either one of these:
```
uci add_list network.ifname.sendopts='0x3c:aabbccddeeff'
```
```
uci add_list network.ifname.sendopts='60:aabbccddeeff'
```
```
uci add_list network.ifname.sendopts='vendor:aabbccddeeff'
```

Adding yet another option for this would, IMO, add unnecessary clutter and also, cause more ambiguity instead of less: Which one takes precedence, `vendorid` or `vendorid_hex`? Even if it were properly documented in the Wiki, it's not immediately obvious to the user, resulting in additional error-proneness.

I'd vote for not merging this, and instead, using the already existing possibility as mentioned above. The old bug with double-ID60 being emitted in this case has already been fixed with #21450. So in my opinion, this PR is unnecessary.

PR #24430, opened 2026-07-26
rockchip: add support for Radxa E54C

default-username-852:
The Radxa E54C is a compact network computer using the Rockchip RK3582
SoC.

- https://radxa.com/products/network-computer/e54c

Disable Synopsys GMAC TSO support. With snps,tso enabled, TX performance
was severely degraded (1.83 Mbit/s measured with iperf3), while disabling
TSO restored normal throughput (933 Mbit/s measured with iperf3).

Hardware
--------

- Rockchip RK3582
  - Dual Cortex-A76 + Quad Cortex-A55 CPU
  - 5 TOPS NPU
- 2/4/8/16/32 GB LPDDR4X RAM (board variants)
- Storage
  - 16 MB SPI NOR flash (SPI boot variant), or
  - 8/16/32 GB on-board eMMC (eMMC variant)
  - microSD card slot
  - M.2 M-key (PCIe 2.1 x1) for NVMe SSD
- 4x Gigabit Ethernet (3x LAN + 1x WAN, RTL8367RB switch)
- 1 USB 3.0 Type-A Interface
- 2 USB 2.0 Type-A Interfaces
- 1 USB 3.0 Type-C (OTG), providing the serial console (1500000 8N1)
- 1 HDMI 2.1 Output
- 5 LEDs (SYS, LAN1, LAN2, LAN3, WAN)
- 3 buttons (Power, MaskROM, User)
- 12 V DC barrel jack (recommended 2 A or higher)

Installation
------------

Boot from SD:
  - Decompress openwrt-rockchip-armv8-radxa_e54c-ext4-sysupgrade.img.gz.
  - Write the resulting .img to a microSD card with dd or similar.
  - Boot the board from the microSD card.

Install to eMMC:
  - Boot OpenWrt from the microSD card.
  - Write the sysupgrade image to the eMMC using sysupgrade or dd.

The boot log and U-Boot console are available over the USB-C serial console
at 1500000 8N1.

MAC addresses
-------------

The on-board EEPROM stores four MAC addresses used by the DSA user ports:

 eth_mac0 (0x9e) -> lan3
 eth_mac1 (0xa4) -> lan2
 eth_mac2 (0xaa) -> lan1
 eth_mac3 (0xb0) -> wan

The mapping was verified by checking link status against the physical
Ethernet port labels.

The DSA conduit (gmac1) receives its MAC address from the U-Boot
environment via the standard ethernet0 DT fixup. This keeps the conduit
MAC consistent with the vendor boot flow, while the EEPROM MAC addresses
are reserved for the DSA user ports.

### DIFF
--- a/target/linux/rockchip/armv8/base-files/etc/board.d/01_leds
+++ b/target/linux/rockchip/armv8/base-files/etc/board.d/01_leds
@@ -61,6 +61,12 @@ radxa,e52c)
 	ucidef_set_led_netdev "wan" "WAN" "green:wan" "wan"
 	ucidef_set_led_netdev "lan" "LAN" "green:lan" "lan"
 	;;
+radxa,e54c)
+	ucidef_set_led_netdev "lan1" "LAN1" "green:lan-1" "lan1"
+	ucidef_set_led_netdev "lan2" "LAN2" "green:lan-2" "lan2"
+	ucidef_set_led_netdev "lan3" "LAN3" "green:lan-3" "lan3"
+	ucidef_set_led_netdev "wan" "WAN" "green:wan" "wan"
+	;;
 esac
 
 board_config_flush
--- a/target/linux/rockchip/image/armv8.mk
+++ b/target/linux/rockchip/image/armv8.mk
@@ -305,6 +305,16 @@ define Device/radxa_e52c
 endef
 TARGET_DEVICES += radxa_e52c
 
+define Device/radxa_e54c
+  $(Device/rk3582)
+  DEVICE_VENDOR := Radxa
+  DEVICE_MODEL := E54C
+  UBOOT_DEVICE_NAME := generic-rk3588
+  DEVICE_DTS := rk3582-radxa-e54c
+  DEVICE_PACKAGES := blkdiscard kmod-button-hotplug kmod-dsa-rtl8365mb kmod-input-adc-keys
+endef
+TARGET_DEVICES += radxa_e54c
+
 define Device/radxa_rock-2a
   $(Device/rk3528)
   DEVICE_VENDOR := Radxa
--- a/target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
+++ b/target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
@@ -1,0 +1,1125 @@
+From ddd86340a007963e8e893555d64f66b63a174ff7 Mon Sep 17 00:00:00 2001
+From: Eskil Queseth <eskil@queseth.se>
+Date: Mon, 03 Aug 2026 20:57:45 +0000
+Subject: [PATCH] arm64: dts: rockchip: Add Radxa E54C
+
+Signed-off-by: Eskil Queseth <eskil@queseth.se>
+--- a/arch/arm64/boot/dts/rockchip/Makefile
++++ b/arch/arm64/boot/dts/rockchip/Makefile
+@@ -161,6 +161,7 @@ dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3576-na
+ dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3576-roc-pc.dtb
+ dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3576-rock-4d.dtb
+ dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3582-radxa-e52c.dtb
++dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3582-radxa-e54c.dtb
+ dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3588-armsom-sige7.dtb
+ dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3588-armsom-w3.dtb
+ dtb-$(CONFIG_ARCH_ROCKCHIP) += rk3588-coolpi-cm5-evb.dtb
+--- /dev/null
++++ b/arch/arm64/boot/dts/rockchip/rk3582-radxa-e54c.dts
+@@ -0,0 +1,1106 @@
++// SPDX-License-Identifier: (GPL-2.0+ OR MIT)
++
++/dts-v1/;
++
++#include <dt-bindings/gpio/gpio.h>
++#include <dt-bindings/input/input.h>
++#include <dt-bindings/leds/common.h>
++#include <dt-bindings/pinctrl/rockchip.h>
++#include <dt-bindings/usb/pd.h>
++#include "rk3588s.dtsi"
++
++/ {
++	model = "Radxa E54C";
++	compatible = "radxa,e54c", "rockchip,rk3582", "rockchip,rk3588s";
++
++	aliases {
++		ethernet0 = &gmac1;
++		mmc0 = &sdhci;
++		mmc1 = &sdmmc;
++
++		led-boot = &status_led;
++		led-failsafe = &status_led;
++		led-running = &status_led;
++		led-upgrade = &status_led;
++	};
++
++	chosen {
++		stdout-path = "serial2:1500000n8";
++	};
++
++	keys-0 {
++		compatible = "adc-keys";
++		io-channels = <&saradc 0>;
++		io-channel-names = "buttons";
++		keyup-threshold-microvolt = <18000>;
++		poll-interval = <100>;
++
++		button-0 {
++			label = "Maskrom";
++			linux,code = <KEY_VENDOR>;
++			press-threshold-microvolt = <0>;
++		};
++	};
++
++	keys-1 {
++		compatible = "gpio-keys";
++		pinctrl-names = "default";
++		pinctrl-0 = <&user_key>;
++
++		button-1 {
++			label = "User";
++			linux,code = <KEY_RESTART>;
++			gpios = <&gpio0 RK_PB2 GPIO_ACTIVE_LOW>;
++			wakeup-source;
++		};
++	};
++
++	leds {
++		compatible = "gpio-leds";
++
++		status_led: led-0 {
++			color = <LED_COLOR_ID_GREEN>;
++			function = LED_FUNCTION_STATUS;
++			gpios = <&gpio0 RK_PC6 GPIO_ACTIVE_HIGH>;
++			pinctrl-names = "default";
++			pinctrl-0 = <&sys_led>;
++		};
++
++		led-1 {
++			color = <LED_COLOR_ID_GREEN>;
++			function = LED_FUNCTION_LAN;
++			function-enumerator = <1>;
++			linux,default-trigger = "netdev";
++			gpios = <&gpio0 RK_PD3 GPIO_ACTIVE_LOW>;
++			pinctrl-names = "default";
++			pinctrl-0 = <&lan1_led>;
++		};
++
++		led-2 {
++			color = <LED_COLOR_ID_GREEN>;
++			function = LED_FUNCTION_LAN;
++			function-enumerator = <2>;
++			linux,default-trigger = "netdev";
++			gpios = <&gpio1 RK_PB6 GPIO_ACTIVE_LOW>;
++			pinctrl-names = "default";
++			pinctrl-0 = <&lan2_led>;
++		};
++
++		led-3 {
++			color = <LED_COLOR_ID_GREEN>;
++			function = LED_FUNCTION_LAN;
++			function-enumerator = <3>;
++			linux,default-trigger = "netdev";
++			gpios = <&gpio1 RK_PB7 GPIO_ACTIVE_LOW>;
++			pinctrl-names = "default";
++			pinctrl-0 = <&lan3_led>;
++		};
++
++		led-4 {
++			color = <LED_COLOR_ID_GREEN>;
++			function = LED_FUNCTION_WAN;
++			linux,default-trigger = "netdev";
++			gpios = <&gpio3 RK_PC4 GPIO_ACTIVE_LOW>;
++			pinctrl-names = "default";
++			pinctrl-0 = <&wan_led>;
++		};
++	};
++
++	dc_12v: regulator-12v-dc {
++		compatible = "regulator-fixed";
++		regulator-name = "dc_12v";
++		regulator-always-on;
++		regulator-boot-on;
++		regulator-min-microvolt = <12000000>;
++		regulator-max-microvolt = <12000000>;
++	};
++
++	vcc_1v1_nldo_s3: regulator-1v1-vcc-nldo-s3 {
++		compatible = "regulator-fixed";
++		regulator-name = "vcc_1v1_nldo_s3";
++		regulator-always-on;
++		regulator-boot-on;
++		regulator-min-microvolt = <1100000>;
++		regulator-max-microvolt = <1100000>;
++		vin-supply = <&vcc_sysin>;
++	};
++
++	vcc3v3_pcie: regulator-3v3-pcie {
++		compatible = "regulator-fixed";
++		regulator-name = "vcc3v3_pcie";
++		regulator-always-on;
++		regulator-boot-on;
++		regulator-min-microvolt = <3300000>;
++		regulator-max-microvolt = <3300000>;
++		vin-supply = <&vcc_sysin>;
++	};
++
++	vcc_3v3_s0: regulator-3v3-vcc-s0 {
++		compatible = "regulator-fixed";
++		regulator-name = "vcc_3v3_s0";
++		regulator-always-on;
++		regulator-boot-on;
++		regulator-min-microvolt = <3300000>;
++		regulator-max-microvolt = <3300000>;
++		vin-supply = <&vcc_3v3_s3>;
++	};
++
++	vcca: regulator-5v0-vcca {
++		compatible = "regulator-fixed";
++		regulator-name = "vcca";
++		regulator-always-on;
++		regulator-boot-on;
++		regulator-min-microvolt = <5000000>;
++		regulator-max-microvolt = <5000000>;
++		vin-supply = <&typec_vin>;
++	};
++
++	vcc5v_usb3: regulator-5v0-usb3 {
++		compatible = "regulator-fixed";
++		enable-active-high;
++		gpios = <&gpio0 RK_PD4 GPIO_ACTIVE_HIGH>;
++		pinctrl-names = "default";
++		pinctrl-0 = <&usb_otg_pwren_h>;
++		regulator-name = "vcc5v_usb3";
++		regulator-min-microvolt = <5000000>;
++		regulator-max-microvolt = <5000000>;
++		vin-supply = <&vcc_sysin>;
++	};
++
++	vcc5v_usb_hub: regulator-5v0-usb-hub {
++		compatible = "regulator-fixed";
++		enable-active-high;
++		gpios = <&gpio4 RK_PB5 GPIO_ACTIVE_HIGH>;
++		pinctrl-names = "default";
++		pinctrl-0 = <&usb_host_pwren_h>;
++		regulator-name = "vcc5v_usb_hub";
++		regulator-min-microvolt = <5000000>;
++		regulator-max-microvolt = <5000000>;
++		vin-supply = <&vcc_sysin>;
++	};
++
++	vbus_typec0: regulator-5v0-vbus-typec0 {
++		compatible = "regulator-fixed";
++		enable-active-high;
++		gpios = <&gpio3 RK_PC0 GPIO_ACTIVE_HIGH>;
++		pinctrl-names = "default";
++		pinctrl-0 = <&typec5v_pwren_h>;
++		regulator-name = "vbus_typec0";
++		regulator-min-microvolt = <5000000>;
++		regulator-max-microvolt = <5000000>;
++		vin-supply = <&vcc_sysin>;
++	};
++
++	vcc_5v0: regulator-5v0-vcc {
++		compatible = "regulator-fixed";
++		regulator-name = "vcc_5v0";
++		regulator-always-on;
++		regulator-boot-on;
++		regulator-min-microvolt = <5000000>;
++		regulator-max-microvolt = <5000000>;
++		vin-supply = <&vcc_sysin>;
++	};
++
++	avddh_3v3: regulator-avdd-3v3 {
++		compatible = "regulator-fixed";
++		enable-active-high;
++		gpios = <&gpio3 RK_PA6 GPIO_ACTIVE_HIGH>;
++		pinctrl-names = "default";
++		pinctrl-0 = <&rtl8367_en>;
++		regulator-name = "avddh_3v3";
++		startup-delay-us = <10000>;
++		vin-supply = <&vcc_5v0>;
++	};
++
++	typec_vin: regulator-typec-vin {
++		compatible = "regulator-fixed";
++		regulator-name = "typec_vin";
++		regulator-always-on;
++		regulator-boot-on;
++		vin-supply = <&dc_12v>;
++	};
++
++	vcc_sysin: regulator-vcc-sysin {
++		compatible = "regulator-fixed";
++		regulator-name = "vcc_sysin";
++		regulator-always-on;
++		regulator-boot-on;
++		vin-supply = <&typec_vin>;
++	};
++};
++
++&combphy0_ps {
++	status = "okay";
++};
++
++&combphy2_psu {
++	status = "okay";
++};
++
++/*
++ * In the Rockchip RK3582 SoC, some CPU cores end up disabled
++ * and unused because they're marked in the efuses as defective.
++ * The disabling in the DT is performed by the boot loader.
++ */
++&cpu_b0 {
++	cpu-supply = <&vdd_cpu_big0_s0>;
++};
++
++&cpu_b1 {
++	cpu-supply = <&vdd_cpu_big0_s0>;
++};
++
++&cpu_b2 {
++	cpu-supply = <&vdd_cpu_big1_s0>;
++};
++
++&cpu_b3 {
++	cpu-supply = <&vdd_cpu_big1_s0>;
++};
++
++&cpu_l0 {
++	cpu-supply = <&vdd_cpu_lit_s0>;
++};
++
++&cpu_l1 {
++	cpu-supply = <&vdd_cpu_lit_s0>;
++};
++
++&cpu_l2 {
++	cpu-supply = <&vdd_cpu_lit_s0>;
++};
++
++&cpu_l3 {
++	cpu-supply = <&vdd_cpu_lit_s0>;
++};
++
++&gmac1 {
++	/*
++	 * TSO causes severe TX performance degradation on this board.
++	 * With snps,tso enabled, TCP transmit throughput drops from ~933 Mbps
++	 * to ~2 Mbps.
++	 */
++	/delete-property/ snps,tso;
++	clock_in_out = "output";
++	phy-mode = "rgmii-id";
++	phy-supply = <&avddh_3v3>;
++	pinctrl-names = "default";
++	pinctrl-0 = <&gmac1_miim
++		     &gmac1_tx_bus2
++		     &gmac1_rx_bus2
++		     &gmac1_rgmii_clk
++		     &gmac1_rgmii_bus>;
++	status = "okay";
++
++	fixed-link {
++		speed = <1000>;
++		full-duplex;
++	};
++};
++
++&i2c0 {
++	pinctrl-names = "default";
++	pinctrl-0 = <&i2c0m2_xfer>;
++	status = "okay";
++
++	vdd_cpu_big0_s0: regulator@42 {
++		compatible = "rockchip,rk8602";
++		reg = <0x42>;
++		fcs,suspend-voltage-selector = <1>;
++		regulator-name = "vdd_cpu_big0_s0";
++		regulator-always-on;
++		regulator-boot-on;
++		regulator-min-microvolt = <550000>;
++		regulator-max-microvolt = <1050000>;
++		regulator-ramp-delay = <2300>;
++		vin-supply = <&vcc_sysin>;
++
++		regulator-state-mem {
++			regulator-off-in-suspend;
++		};
++	};
++
++	vdd_cpu_big1_s0: regulator@43 {
++		compatible = "rockchip,rk8603", "rockchip,rk8602";
++		reg = <0x43>;
++		fcs,suspend-voltage-selector = <1>;
++		regulator-name = "vdd_cpu_big1_s0";
++		regulator-always-on;
++		regulator-boot-on;
++		regulator-min-microvolt = <550000>;
++		regulator-max-microvolt = <1050000>;
++		regulator-ramp-delay = <2300>;
++		vin-supply = <&vcc_sysin>;
++
++		regulator-state-mem {
++			regulator-off-in-suspend;
++		};
++	};
++
++	eeprom@50 {
++		compatible = "belling,bl24c16a", "atmel,24c16";
++		#address-cells = <1>;
++		#size-cells = <1>;
++		reg = <0x50>;
++		pagesize = <16>;
++		read-only;
++		vcc-supply = <&vcc_3v3_s3>;
++
++		nvmem-layout {
++			compatible = "fixed-layout";
++			#address-cells = <1>;
++			#size-cells = <1>;
++
++			eth_mac0: macaddr@9e {
++				reg = <0x9e 0x06>;
++			};
++
++			eth_mac1: macaddr@a4 {
++				reg = <0xa4 0x06>;
++			};
++
++			eth_mac2: macaddr@aa {
++				reg = <0xaa 0x06>;
++			};
++
++			eth_mac3: macaddr@b0 {
++				reg = <0xb0 0x06>;
++			};
++		};
++	};
++};
++
++&i2c2 {
++	status = "okay";
++
++	vdd_npu_s0: regulator@42 {
++		compatible = "rockchip,rk8602";
++		reg = <0x42>;
++		fcs,suspend-voltage-selector = <1>;
++		regulator-name = "vdd_npu_s0";
++		regulator-always-on;
++		regulator-boot-on;
++		regulator-min-microvolt = <550000>;
++		regulator-max-microvolt = <950000>;
++		regulator-ramp-delay = <2300>;
++		vin-supply = <&vcc_sysin>;
++
++		regulator-state-mem {
++			regulator-off-in-suspend;
++		};
++	};
++};
++
++&i2c4 {
++	pinctrl-names = "default";
++	pinctrl-0 = <&i2c4m2_xfer>;
++	status = "okay";
++
++	usbc0: usb-typec@22 {
++		compatible = "fcs,fusb302";
++		reg = <0x22>;
++		interrupt-parent = <&gpio1>;
++		interrupts = <RK_PA7 IRQ_TYPE_LEVEL_LOW>;
++		pinctrl-names = "default";
++		pinctrl-0 = <&cc_int_l>;
++		vbus-supply = <&vbus_typec0>;
++		status = "okay";
++
++		usb_con: connector {
++			compatible = "usb-c-connector";
++			label = "USB-C";
++			data-role = "dual";
++			power-role = "dual";
++			try-power-role = "sink";
++			op-sink-microwatt = <1000000>;
++
++			sink-pdos =
++				<PDO_FIXED(5000, 1000, PDO_FIXED_USB_COMM)>;
++			source-pdos =
++				<PDO_FIXED(5000, 3000, PDO_FIXED_USB_COMM)>;
++
++			altmodes {
++				displayport {
++					svid = /bits/ 16 <0xff01>;
++					vdo = <0xffffffff>;
++				};
++			};
++
++			ports {
++				#address-cells = <1>;
++				#size-cells = <0>;
++
++				port@0 {
++					reg = <0>;
++					usbc0_hs: endpoint {
++						remote-endpoint = <&usb_host0_xhci_to_usbc0>;
++					};
++				};
++
++				port@1 {
++					reg = <1>;
++					usbc0_ss: endpoint {
++						remote-endpoint = <&usbdp_phy0_ss>;
++					};
++				};
++
++				port@2 {
++					reg = <2>;
++					usbc0_sbu: endpoint {
++						remote-endpoint = <&usbdp_phy0_sbu>;
++					};
++				};
++			};
++		};
++	};
++
++	hym8563: rtc@51 {
++		compatible = "haoyu,hym8563";
++		reg = <0x51>;
++		#clock-cells = <0>;
++		clock-output-names = "rtcic_32kout";
++		interrupt-parent = <&gpio0>;
++		interrupts = <RK_PB0 IRQ_TYPE_LEVEL_LOW>;
++		pinctrl-names = "default";
++		pinctrl-0 = <&rtc_int_l>;
++		wakeup-source;
++	};
++};
++
++&i2s5_8ch {
++	status = "okay";
++};
++
++&mdio1 {
++	ethernet-switch@1d {
++		compatible = "realtek,rtl8365mb";
++		reg = <29>;
++		reset-gpios = <&gpio3 RK_PB7 GPIO_ACTIVE_LOW>;
++		pinctrl-names = "default";
++		pinctrl-0 = <&gmac1_rstn_l>;
++
++		switch_intc: interrupt-controller {
++			interrupt-parent = <&gpio3>;
++			interrupts = <RK_PB6 IRQ_TYPE_LEVEL_LOW>;
++			interrupt-controller;
++			#address-cells = <0>;
++			#interrupt-cells = <1>;
++		};
++
++		mdio {
++			#address-cells = <1>;
++			#size-cells = <0>;
++
++			phy0: ethernet-phy@0 {
++				reg = <0>;
++				interrupt-parent = <&switch_intc>;
++				interrupts = <0>;
++			};
++
++			phy1: ethernet-phy@1 {
++				reg = <1>;
++				interrupt-parent = <&switch_intc>;
++				interrupts = <1>;
++			};
++
++			phy2: ethernet-phy@2 {
++				reg = <2>;
++				interrupt-parent = <&switch_intc>;
++				interrupts = <2>;
++			};
++
++			phy3: ethernet-phy@3 {
++				reg = <3>;
++				interrupt-parent = <&switch_intc>;
++				interrupts = <3>;
++			};
++
++			phy4: ethernet-phy@4 {
++				reg = <4>;
++				interrupt-parent = <&switch_intc>;
++				interrupts = <4>;
++			};
++		};
++
++		ethernet-ports {
++			#address-cells = <1>;
++			#size-cells = <0>;
++
++			ethernet-port@0 {
++				reg = <0>;
++				label = "lan3";
++				phy-handle = <&phy0>;
++				nvmem-cells = <&eth_mac0>;
++				nvmem-cell-names = "mac-address";
++			};
++
++			ethernet-port@1 {
++				reg = <1>;
++				label = "lan2";
++				phy-handle = <&phy1>;
++				nvmem-cells = <&eth_mac1>;
++				nvmem-cell-names = "mac-address";
++			};
++
++			ethernet-port@2 {
++				reg = <2>;
++				label = "lan1";
++				phy-handle = <&phy2>;
++				nvmem-cells = <&eth_mac2>;
++				nvmem-cell-names = "mac-address";
++			};
++
++			ethernet-port@3 {
++				reg = <3>;
++				label = "wan";
++				phy-handle = <&phy3>;
++				nvmem-cells = <&eth_mac3>;
++				nvmem-cell-names = "mac-address";
++			};
++
++			ethernet-port@6 {
++				reg = <6>;
++				ethernet = <&gmac1>;
++				phy-mode = "rgmii-id";
++				rx-internal-delay-ps = <2000>;
++				tx-internal-delay-ps = <2000>;
++
++				fixed-link {
++					speed = <1000>;
++					full-duplex;
++				};
++			};
++		};
++	};
++};
++
++&pcie2x1l2 {
++	pinctrl-names = "default";
++	pinctrl-0 = <&pcie20x1_2_perstn_m0>;
++	reset-gpios = <&gpio3 RK_PD1 GPIO_ACTIVE_HIGH>;
++	vpcie3v3-supply = <&vcc3v3_pcie>;
++	status = "okay";
++};
++
++&pinctrl {
++	ethernet {
++		gmac1_rstn_l: gmac1-rstn-l {
++			rockchip,pins = <3 RK_PB7 RK_FUNC_GPIO &pcfg_pull_none>;
++		};
++
++		rtl8367_en: rtl8367-en {
++			rockchip,pins = <3 RK_PA6 RK_FUNC_GPIO &pcfg_pull_none>;
++		};
++	};
++
++	keys {
++		user_key: user-key {
++			rockchip,pins = <0 RK_PB2 RK_FUNC_GPIO &pcfg_pull_up>;
++		};
++	};
++
++	leds {
++		lan1_led: lan1-led {
++			rockchip,pins = <0 RK_PD3 RK_FUNC_GPIO &pcfg_pull_none>;
++		};
++
++		lan2_led: lan2-led {
++			rockchip,pins = <1 RK_PB6 RK_FUNC_GPIO &pcfg_pull_none>;
++		};
++
++		lan3_led: lan3-led {
++			rockchip,pins = <1 RK_PB7 RK_FUNC_GPIO &pcfg_pull_none>;
++		};
++
++		sys_led: sys-led {
++			rockchip,pins = <0 RK_PC6 RK_FUNC_GPIO &pcfg_pull_none>;
++		};
++
++		wan_led: wan-led {
++			rockchip,pins = <3 RK_PC4 RK_FUNC_GPIO &pcfg_pull_none>;
++		};
++	};
++
++	pcie {
++		pcie20x1_2_perstn_m0: pcie-perstn {
++			rockchip,pins = <3 RK_PD1 RK_FUNC_GPIO &pcfg_pull_none>;
++		};
++	};
++
++	rtc {
++		rtc_int_l: rtc-int-l {
++			rockchip,pins = <0 RK_PB0 RK_FUNC_GPIO &pcfg_pull_none>;
++		};
++	};
++
++	typec {
++		cc_int_l: cc-int-l {
++			rockchip,pins = <1 RK_PA7 RK_FUNC_GPIO &pcfg_pull_up>;
++		};
++
++		typec5v_pwren_h: typec5v-pwren-h {
++			rockchip,pins = <3 RK_PC0 RK_FUNC_GPIO &pcfg_pull_none>;
++		};
++	};
++
++	usb {
++		usb_host_pwren_h: usb-host-pwren-h {
++			rockchip,pins = <4 RK_PB5 RK_FUNC_GPIO &pcfg_pull_none>;
++		};
++
++		usb_otg_pwren_h: usb-otg-pwren-h {
++			rockchip,pins = <0 RK_PD4 RK_FUNC_GPIO &pcfg_pull_none>;
++		};
++	};
++};
++
++&saradc {
++	vref-supply = <&vcca_1v8_s0>;
++	status = "okay";
++};
++
++&sdhci {
++	bus-width = <8>;
++	cap-mmc-highspeed;
++	mmc-hs400-1_8v;
++	mmc-hs400-enhanced-strobe;
++	non-removable;
++	vmmc-supply = <&vcc_3v3_s0>;
++	vqmmc-supply = <&vcc_1v8_s3>;
++	status = "okay";
++};
++
++&sdmmc {
++	bus-width = <4>;
++	cap-mmc-highspeed;
++	cap-sd-highspeed;
++	cd-gpios = <&gpio0 RK_PA4 GPIO_ACTIVE_LOW>;
++	disable-wp;
++	sd-uhs-sdr104;
++	vmmc-supply = <&vcc_3v3_s3>;
++	vqmmc-supply = <&vccio_sd_s0>;
++	status = "okay";
++};
++
++&sfc {
++	pinctrl-names = "default";
++	pinctrl-0 = <&fspim0_pins>;
++	status = "okay";
++
++	flash@0 {
++		compatible = "jedec,spi-nor";
++		reg = <0>;
++		spi-max-frequency = <104000000>;
++		spi-rx-bus-width = <4>;
++		spi-tx-bus-width = <1>;
++		vcc-supply = <&vcc_1v8_s3>;
++	};
++};
++
++&spi2 {
++	status = "okay";
++	assigned-clocks = <&cru CLK_SPI2>;
++	assigned-clock-rates = <200000000>;
++	num-cs = <1>;
++	pinctrl-names = "default";
++	pinctrl-0 = <&spi2m2_cs0 &spi2m2_pins>;
++
++	pmic@0 {
++		compatible = "rockchip,rk806";
++		reg = <0>;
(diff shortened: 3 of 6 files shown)


target/linux/rockchip/image/armv8.mk
```diff
@@ -305,6 +305,16 @@ define Device/radxa_e52c
 endef
 TARGET_DEVICES += radxa_e52c
 
+define Device/radxa_e54c
+  $(Device/rk3582)
+  DEVICE_VENDOR := Radxa
+  DEVICE_MODEL := E54C
+  UBOOT_DEVICE_NAME := generic-rk3588
+  DEVICE_DTS := rk3582-radxa-e54c
+  DEVICE_PACKAGES := kmod-button-hotplug kmod-dsa-rtl8365mb
```
untilscour:
```suggestion
  DEVICE_PACKAGES := kmod-button-hotplug kmod-input-adc-keys kmod-dsa-rtl8365mb
```

for the maskrom button

target/linux/rockchip/armv8/base-files/etc/board.d/01_leds
```diff
@@ -61,6 +61,12 @@ radxa,e52c)
 	ucidef_set_led_netdev "wan" "WAN" "green:wan" "wan"
 	ucidef_set_led_netdev "lan" "LAN" "green:lan" "lan"
 	;;
+radxa,e54c)
+    ucidef_set_led_netdev "lan1" "LAN1" "green:lan1" "lan1"
+    ucidef_set_led_netdev "lan2" "LAN2" "green:lan2" "lan2"
+    ucidef_set_led_netdev "lan3" "LAN3" "green:lan3" "lan3"
+    ucidef_set_led_netdev "wan" "WAN" "green:wan" "wan"
+    ;;
```
openwrt-ai: The new block is indented with 4 spaces; every other case arm in this file (including [`radxa,e52c` right above](https://github.com/openwrt/openwrt/blob/6fc2cce299605c389bee32bcbf83eca08a8a06cd/target/linux/rockchip/armv8/base-files/etc/board.d/01_leds#L60-L63)) uses a single tab.

```suggestion
	ucidef_set_led_netdev "lan1" "LAN1" "green:lan1" "lan1"
	ucidef_set_led_netdev "lan2" "LAN2" "green:lan2" "lan2"
	ucidef_set_led_netdev "lan3" "LAN3" "green:lan3" "lan3"
	ucidef_set_led_netdev "wan" "WAN" "green:wan" "wan"
	;;
```

Note the sysfs names here also have to be updated if the DTS LEDs move to `color`/`function` (see the comment on the DTS): `LED_FUNCTION_LAN` + `function-enumerator = <N>` yields `green:lan-1` / `green:lan-2` / `green:lan-3`, not `green:lan1`.

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1038 @@
...
++			pinctrl-0 = <&sys_led>;
++		};
++
++		led-1 {
++			label = "green:lan1";
++            linux,default-trigger = "netdev";
++			gpios = <&gpio0 RK_PD3 GPIO_ACTIVE_LOW>;
++			pinctrl-names = "default";
++			pinctrl-0 = <&lan1_led>;
++		};
```
openwrt-ai, line 96: `label = "green:lan1";` is the deprecated LED naming syntax. Use `color = <LED_COLOR_ID_GREEN>;` plus `function = LED_FUNCTION_LAN;` / `function-enumerator = <1>;` — the DTS already includes `<dt-bindings/leds/common.h>` (line 20) but never uses it, and the sibling board does it the modern way ([`rk3582-radxa-e52c.dts` leds-0/leds-1](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/arch/arm64/boot/dts/rockchip/rk3582-radxa-e52c.dts#L55-L87)). Same applies to `led-0` (line 76 → `LED_FUNCTION_STATUS`), `led-2` (line 91), `led-3` (line 99) and `led-4` (line 107 → `LED_FUNCTION_WAN`).

Also in the same hunk: `linux,default-trigger` on lines 84/92/100/108 is indented with spaces while the rest of the file uses tabs.

No suggestion block here because editing these lines changes the line count of the `@@ -0,0 +1,1025 @@` hunk — regenerate the patch afterwards with `make target/linux/refresh` (rockchip selected in `.config`) so the hunk header matches.

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1038 @@
...
++		wakeup-source;
++	};
++};
++
++&i2s5_8ch {
++	status = "okay";
++};
++
++&mdio1 {
++	ethernet-switch@29 {
```
openwrt-ai: Unit address does not match `reg`. Device-tree unit addresses are hex, so `@29` means MDIO address 0x29 (41), while `reg = <29>;` on the next line is decimal 29 (0x1d) — the correct RTL8367 default. Rename the node to `@1d`:

```suggestion
+	ethernet-switch@1d {
```

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1038 @@
...
++	clock_in_out = "output";
++	phy-mode = "rgmii-id";
++	phy-supply = <&avddh_3v3>;
++	pinctrl-names = "default";
++	pinctrl-0 = <&gmac1_miim
++		     &gmac1_tx_bus2
++		     &gmac1_rx_bus2
++		     &gmac1_rgmii_clk
++		     &gmac1_rgmii_bus>;
++	snps,no-vlhash;
```
openwrt-ai: `snps,no-vlhash` is not a property any kernel code or binding knows about: `git grep no-vlhash` returns nothing in Linux 6.18 (the kernel version this target builds, `KERNEL_PATCHVER:=6.18`) and nothing in this tree either — stmmac only exposes VLAN-hash filtering through the `dma_cap.vlhash` capability bit, there is no DT opt-out. As written this line is a silent no-op. Is it a leftover from a vendor kernel DTS? If VLAN hash filtering actually needs disabling, that needs a driver-side change, not a DT property.

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1038 @@
...
++&cpu_l2 {
++	cpu-supply = <&vdd_cpu_lit_s0>;
++};
++
++&cpu_l3 {
++	cpu-supply = <&vdd_cpu_lit_s0>;
++};
++
++&gmac1 {
++	/delete-property/ snps,tso;
```
openwrt-ai, line 302: This deletes `snps,tso` inherited from [`rk3588-base.dtsi`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/arch/arm64/boot/dts/rockchip/rk3588-base.dtsi#L2051), i.e. it turns TSO off for this board — a measurable throughput hit — and no other rockchip board in this tree does it. The commit message doesn't mention it. If it works around a concrete problem (e.g. corrupted frames with DSA tagging on this GMAC), please add a comment above the line and a note in the commit message; otherwise it looks like it should be dropped.

---

target/linux/rockchip/image/armv8.mk
openwrt-ai: The DTS declares an `adc-keys` node for the Maskrom button (`keys-0`, patch lines 45-56), but `kmod-input-adc-keys` is not in `DEVICE_PACKAGES`, so that key will never be reported. The sibling [`Device/radxa_e52c` at line 302](https://github.com/openwrt/openwrt/blob/6fc2cce299605c389bee32bcbf83eca08a8a06cd/target/linux/rockchip/image/armv8.mk#L302) ships it, together with `blkdiscard` for the eMMC.

```suggestion
  DEVICE_PACKAGES := blkdiscard kmod-button-hotplug kmod-dsa-rtl8365mb kmod-input-adc-keys
```

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1038 @@
...
++#include <dt-bindings/pwm/pwm.h>
++#include <dt-bindings/usb/pd.h>
++#include "rk3588s.dtsi"
++
++/ {
++	model = "Radxa E54C";
++	compatible = "radxa,e54c", "rockchip,rk3582", "rockchip,rk3588s";
++
++	aliases {
++		ethernet0 = &gmac1;
```
openwrt-ai, line 36: All four EEPROM MACs are consumed by the DSA user ports (lines 483-513), so `&gmac1` — the DSA conduit — gets a random MAC on every boot. The comparable in-tree change for the E20C, [`165-arm64-dts-rockchip-Use-MAC-address-from-EEPROM-for-.patch`](``https://github.com/openwrt/openwrt/blob/6fc2cce299605c389bee32bcbf83eca08a8a06cd/target/linux/rockchip/patches-6.18/165-arm64-dts-rockchip-Use-MAC-address-from-EEPROM-for-.patch),`` does the opposite: it puts `nvmem-cells = <&eth_mac1>` on `&gmac1` and *removes* the `ethernet0` alias. Was a stable conduit MAC considered here, and is the `ethernet0` alias intentional given that precedent?

Related question on the mapping: `eth_mac0` (0x9e) → `lan3`, `eth_mac1` → `lan2`, `eth_mac2` → `lan1`, `eth_mac3` (0xb0) → `wan`. Is the reversed order verified against the vendor firmware / the label on the case?

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1038 @@
...
++// SPDX-License-Identifier: (GPL-2.0+ OR MIT)
++
++/dts-v1/;
++
++#include <dt-bindings/gpio/gpio.h>
++#include <dt-bindings/input/input.h>
++#include <dt-bindings/leds/common.h>
++#include <dt-bindings/pinctrl/rockchip.h>
++#include <dt-bindings/pwm/pwm.h>
++#include <dt-bindings/usb/pd.h>
```
openwrt-ai: nit: both includes are unused — no `PWM_*` and no `PD_*` macro appears anywhere in this DTS (the E52C needs `pwm.h` because its LEDs are `pwm-leds`; here they are `gpio-leds`). Dropping them means adjusting the hunk line count, so fold it into the patch regeneration mentioned on the LED comment.

---

openwrt-ai (commented):
## Commit checks

- `6fc2cce` "rockchip: add support for Radxa E54C" — this adds a new device (new `rk3582-radxa-e54c.dts` plus a new `define Device/radxa_e54c` in `target/linux/rockchip/image/armv8.mk`), so the commit message is expected to carry three sections; two are incomplete or absent:
  - **MAC address layout** — missing entirely. The DTS reads four MACs out of the on-board EEPROM (`macaddr@9e`, `@a4`, `@aa`, `@b0`) and maps them to lan3/lan2/lan1/wan; please document one line per interface (`xx:xx:xx` for the unit-specific bytes), including where the `gmac1` conduit MAC comes from.
  - **Flash instructions** — "Uncompress the OpenWrt sysupgrade and write it to a micro SD card or eMMC using dd" is not enough for a stranger to flash the board: name the image file, say how to get the device into a writable state for the eMMC path (Maskrom button + `rkdeveloptool`, or U-Boot/OEM path), and add a revert-to-stock note if a procedure exists.
  - **Hardware specification** — present, but missing the details the rule asks for: flash/eMMC and SPI-NOR chip types, LED and button counts, UART header location/pinout/baud (`serial2:1500000n8` is in the DTS), and power input (12 V DC jack, current).

Inline comments cover the DTS and board.d findings.

---

target/linux/rockchip/image/armv8.mk
default-username-852: Nice! Added `input-adc-keys`, and verified that pushing the maskrom button generated key press events.

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
default-username-852: Enabling TSO causes a large performance hit. I added a note that the property is deleted due to the performance hit.

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
default-username-852: The ethernet0 alias causes u-boot to assign the GMAC conduit a fixed MAC address via the device tree fixup, so I've kept it here.

The order is correct. Plugging into the port label WAN causes the wan interface in OpenWRT to come up, and similar for LAN1,2,3.

peterwillcn (approved):

target/linux/rockchip/image/armv8.mk
```diff
@@ -305,6 +305,16 @@ define Device/radxa_e52c
 endef
 TARGET_DEVICES += radxa_e52c
 
+define Device/radxa_e54c
+  $(Device/rk3582)
+  DEVICE_VENDOR := Radxa
+  DEVICE_MODEL := E54C
+  UBOOT_DEVICE_NAME := generic-rk3588
+  DEVICE_DTS := rk3582-radxa-e54c
+  DEVICE_PACKAGES := kmod-button-hotplug kmod-input-adc-keys kmod-dsa-rtl8365mb
```
openwrt-ai: `blkdiscard` is still missing, and the list is no longer sorted (`kmod-input-adc-keys` before `kmod-dsa-rtl8365mb`) — every other `DEVICE_PACKAGES` line in this file is alphabetical.

The board has on-board eMMC, and every other eMMC-equipped Radxa board in this file ships `blkdiscard`: [`radxa_e52c` L302](https://github.com/openwrt/openwrt/blob/61b0c78172b2f12148e0ec8b6e7d63d1deaa74bf/target/linux/rockchip/image/armv8.mk#L302), [`radxa_e20c` L95](https://github.com/openwrt/openwrt/blob/61b0c78172b2f12148e0ec8b6e7d63d1deaa74bf/target/linux/rockchip/image/armv8.mk#L95), [`radxa_e25` L239](https://github.com/openwrt/openwrt/blob/61b0c78172b2f12148e0ec8b6e7d63d1deaa74bf/target/linux/rockchip/image/armv8.mk#L239). Was the omission deliberate?

```suggestion
  DEVICE_PACKAGES := blkdiscard kmod-button-hotplug kmod-dsa-rtl8365mb kmod-input-adc-keys
```

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1048 @@
...
++	};
++
++	keys-1 {
++		compatible = "gpio-keys";
++		pinctrl-names = "default";
++		pinctrl-0 = <&user_key>;
++
++		button-1 {
++			label = "User";
++			linux,code = <KEY_RESTART>;
```
openwrt-ai, line 71: nit: the button silkscreened "User" is wired to `KEY_RESTART`. With `kmod-button-hotplug` in `DEVICE_PACKAGES`, OpenWrt maps `KEY_RESTART` to the `reset` button handler, so a short press reboots and a long press factory-resets the device — surprising behaviour for a button labelled "User".

The sibling board uses `BTN_0` for its identically-labelled button ([`rk3582-radxa-e52c.dts` button-1](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/arch/arm64/boot/dts/rockchip/rk3582-radxa-e52c.dts#L46-L51)). Is `KEY_RESTART` intentional here?

---

openwrt-ai (commented):
Re-reviewed at `61b0c78`. The DTS fixes from the previous round all look correct: LEDs now use `color`/`function`/`function-enumerator` (and `01_leds` was updated to the resulting `green:lan-1` … sysfs names, which matches [`led_compose_name()`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/leds/led-core.c#L556-L559)), the switch unit address is now `@1d`, the bogus `snps,no-vlhash` is gone, the unused `pwm.h`/`usb/pd.h` includes are dropped, the `/delete-property/ snps,tso` is documented, and the tab indentation in `01_leds` is fixed. The regenerated patch applies cleanly against 6.18 and the `@@ -0,0 +1,1035 @@` count is correct.

## Commit checks

- `61b0c78` "rockchip: add support for Radxa E54C" — the commit message is unchanged from `6fc2cce` and still does not carry the sections required for a new-device commit:
  - **MAC address layout** — still absent. The DTS reads four MACs from the on-board EEPROM (`macaddr@9e`/`@a4`/`@aa`/`@b0`) and maps them to lan3/lan2/lan1/wan; please document one line per interface (`xx:xx:xx` for the unit-specific bytes) and note that the `gmac1` conduit MAC comes from the U-Boot fixup via the `ethernet0` alias — that answer is currently only in a review thread, which will be lost once this merges.
  - **Flash instructions** — "Uncompress the OpenWrt sysupgrade and write it to a micro SD card or eMMC using dd" is not enough to flash the board from the commit message alone: name the image file, describe how to get the eMMC into a writable state (Maskrom button + `rkdeveloptool`, or the U-Boot path), and add a revert-to-stock note if a procedure exists.
  - **Hardware specification** — present but missing eMMC/SPI-NOR chip types, LED and button counts, UART header location/pinout/baud (`serial2:1500000n8`), and power input (12 V DC, current).

  Worth folding in as well: the reason for `/delete-property/ snps,tso` is now in a DTS comment but not in the commit message, and the same for the port-order verification.

Two inline comments below.

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
untilscour: this is fairly typical if there are no other buttons (besides maskrom)

target/linux/rockchip/image/armv8.mk
default-username-852: Added `blkdiscard` and alphabetized `DEVICE_PACKAGES`.

peterwillcn:
[    4.123778] /rng@fe378000: Failed to get clk index: 0 ret: -517

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1048 @@
+--- a/arch/arm64/boot/dts/rockchip/Makefile
```
openwrt-ai, line 7: The patch starts straight at the `---`/`+++` pair with no git header block, which is what the openwrt bot's formality check is complaining about. Every patch in this directory that adds a *new board DTS* carries the `From <sha> Mon Sep 17 00:00:00 2001` / `From:` / `Date:` / `Subject:` block so it stays `git am`-able and can be posted upstream unchanged — e.g. [`101-arm64-dts-rockchip-Add-HINLINK-H28K.patch` L1-L22](``https://github.com/openwrt/openwrt/blob/3f399456e2daab0dacd6ca1e58bc23001b2b59b2/target/linux/rockchip/patches-6.18/101-arm64-dts-rockchip-Add-HINLINK-H28K.patch#L1-L22),`` [`165-arm64-dts-rockchip-Use-MAC-address-from-EEPROM-for-.patch` L1-L10](``https://github.com/openwrt/openwrt/blob/3f399456e2daab0dacd6ca1e58bc23001b2b59b2/target/linux/rockchip/patches-6.18/165-arm64-dts-rockchip-Use-MAC-address-from-EEPROM-for-.patch#L1-L10),`` [`010-v6.19-arm64-dts-rockchip-add-LinkEase-EasePi-R1.patch`](``https://github.com/openwrt/openwrt/blob/3f399456e2daab0dacd6ca1e58bc23001b2b59b2/target/linux/rockchip/patches-6.18/010-v6.19-arm64-dts-rockchip-add-LinkEase-EasePi-R1.patch#L1-L10).`` (The header-less ones in this directory — `106-r4s-openwrt-leds.patch`, the `1xx-*-led-aliases` set — are all OpenWrt-local tweaks that are never going upstream.)

Please prepend the header block with a `Subject: [PATCH] arm64: dts: rockchip: Add Radxa E54C` and your `From:`/`Date:`, plus a short body and your `Signed-off-by:`, ending with the `---` separator before the diff.

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1048 @@
...
++	status = "okay";
++};
++
++&usb_host0_ehci {
++	status = "okay";
++};
++
++&usb_host0_xhci {
++	dr_mode = "host";
++	usb-role-switch;
```
openwrt-ai, line 1102: `usb-role-switch` here has nothing to switch to, and it contradicts `dr_mode = "host"` on the line above.

The `usbc0` fusb302 node (patch line 411) has no `connector { compatible = "usb-c-connector"; ... }` child, so there is no `ports`/`endpoint` graph tying the connector to this xHCI node or to the `mode-switch` / `orientation-switch` declared on `&usbdp_phy0` (patch lines 1021–1022). Every upstream RK3588 board that uses fusb302 wires all three together — see [`rk3588-rock-5b-5bp-5t.dtsi` usbc0 connector, L312-L372](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-rock-5b-5bp-5t.dtsi#L312-L372) and its [`&usb_host0_xhci` port graph, L1024-L1032](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-rock-5b-5bp-5t.dtsi#L1024-L1032) — and none of them keep `dr_mode` alongside `usb-role-switch`, precisely so the role switch can drive the mode.

As written the Type-C port is host-only and the fusb302 / usbdp-phy switch plumbing is dead DT. The commit message describes this port as OTG, so either the connector graph is missing, or `usb-role-switch` (and the phy's `mode-switch`/`orientation-switch`) should be dropped. Which is intended? For comparison the sibling [`rk3582-radxa-e52c.dts` keeps it minimal](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3582-radxa-e52c.dts) — just `dr_mode = "host"` and `status = "okay"`, no role switch.

---

target/linux/rockchip/image/armv8.mk
```diff
@@ -305,6 +305,16 @@ define Device/radxa_e52c
 endef
 TARGET_DEVICES += radxa_e52c
 
+define Device/radxa_e54c
```
openwrt-ai, line 308: nit: the PR description and the commit message now disagree about this board's specs. The PR body says `2/4/8/16/32GB LPDDR4 RAM` and `8/16/32/64/128GB on-board eMMC`; the commit message says `2/4/8/16/32 GB LPDDR4X RAM` and `8/16/32 GB on-board eMMC` (and adds the RTL8367RB switch, SPI-NOR variant, M.2 slot, LEDs/buttons that the body omits). Only the commit message lands in the tree, so please make the body match it — or correct whichever of the two is wrong.

---

openwrt-ai (commented):
Re-reviewed at `3f39945`. Since `61b0c78` the only code change is `DEVICE_PACKAGES` in `armv8.mk` (`blkdiscard` added, list alphabetised) — that addresses the previous round's finding. The commit message has been substantially expanded and now carries all three required sections (hardware spec, flash instructions, MAC address layout), including the `snps,tso` rationale and the port-order verification note. Thanks for that.

## Commit checks

- `3f39945` "rockchip: add support for Radxa E54C" — the new Installation section names the image `openwrt-rockchip-rk3588-radxa_e54c-ext4-sysupgrade.img.gz`, but rockchip has exactly one subtarget, `armv8` ([`target/linux/rockchip/Makefile` `SUBTARGETS:=armv8`](https://github.com/openwrt/openwrt/blob/3f399456e2daab0dacd6ca1e58bc23001b2b59b2/target/linux/rockchip/Makefile#L8)), and `IMG_PREFIX` is `openwrt-<ver>-<board>-<subtarget>`. The file that actually gets built is `openwrt-rockchip-armv8-radxa_e54c-ext4-sysupgrade.img.gz`; a user following the instructions verbatim would look for a name that never exists. `$(Device/rk3582)` only sets `SOC`/`KERNEL_LOADADDR`, it does not change the prefix.
- Minor, same commit: the Power line is `12 V DC barrel jack` with no current rating — worth adding so the spec section is complete.

Three inline comments below (one is a nit).

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
default-username-852: Updated the device tree, and verified USB-C OTG functionality for the type C port.

target/linux/rockchip/image/armv8.mk
default-username-852: The commit message is correct. Updated the PR description to match the commit message.

default-username-852:
> [ 4.123778] /rng@fe378000: Failed to get clk index: 0 ret: -517

I think this message is harmless. After boot, I see that there is a `/dev/hwrng` that can be read from. Running `cat /sys/class/misc/hw_random/rng_current` also gives `rockchip-rng`, so it looks like everything initialized properly. Or do you mean something else?

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1134 @@
...
++		};
++	};
++};
++
++&i2c4 {
++	pinctrl-names = "default";
++	pinctrl-0 = <&i2c4m2_xfer>;
++	status = "okay";
++
++	usbc0: fusb302@22 {
```
openwrt-ai: The node was renamed from `usb-typec@22` (previous revision) to `fusb302@22`. Device-tree node names describe the class of device, not the part number — every in-tree/upstream RK3588 board uses the generic form ([`rk3588-rock-5b-5bp-5t.dtsi:312`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-rock-5b-5bp-5t.dtsi#L312)), and the binding's own example uses `typec-portc@54` ([`fcs,fusb302.yaml:49`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/Documentation/devicetree/bindings/usb/fcs,fusb302.yaml#L49)). Since the patch now carries a `Subject: [PATCH] arm64: dts: rockchip: Add Radxa E54C` header for upstream submission, this will get bounced.

```suggestion
+	usbc0: usb-typec@22 {
```

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1134 @@
...
++		compatible = "fcs,fusb302";
++		reg = <0x22>;
++		interrupt-parent = <&gpio1>;
++		interrupts = <RK_PA7 IRQ_TYPE_LEVEL_LOW>;
++		int-n-gpios = <&gpio1 RK_PA7 GPIO_ACTIVE_LOW>;
++		pinctrl-names = "default";
++		pinctrl-0 = <&cc_int_l>;
++		status = "okay";
```
openwrt-ai, line 426: This revision dropped `vbus-supply = <&vbus_typec0>;` from the tcpc node, which looks unintentional.

`vbus-supply` is a *required* property of the binding ([`fcs,fusb302.yaml:30-35`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/Documentation/devicetree/bindings/usb/fcs,fusb302.yaml#L30-L35)), and `fusb302_probe()` gets it with the non-optional [`devm_regulator_get(chip->dev, "vbus")`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/usb/typec/tcpm/fusb302.c#L1722) — with no supply in DT that silently returns a dummy regulator, so `tcpm_set_vbus()` never toggles `gpio3 RK_PC0` and the port can never source VBUS, even though the connector below advertises `power-role = "dual"` and `source-pdos = <PDO_FIXED(5000, 3000, ...)>`. It also leaves the `vbus_typec0` regulator (line 201) and its `typec5v_pwren_h` pinctrl (line 668) with no consumer anywhere in the DTS.

```suggestion
+		pinctrl-names = "default";
+		pinctrl-0 = <&cc_int_l>;
+		vbus-supply = <&vbus_typec0>;
+		status = "okay";
```

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1134 @@
...
++	pinctrl-names = "default";
++	pinctrl-0 = <&i2c4m2_xfer>;
++	status = "okay";
++
++	usbc0: fusb302@22 {
++		compatible = "fcs,fusb302";
++		reg = <0x22>;
++		interrupt-parent = <&gpio1>;
++		interrupts = <RK_PA7 IRQ_TYPE_LEVEL_LOW>;
++		int-n-gpios = <&gpio1 RK_PA7 GPIO_ACTIVE_LOW>;
```
openwrt-ai: `int-n-gpios` is a Rockchip BSP-ism, not an upstream property: `grep -rn int-n-gpios` over the whole 6.18 tree returns nothing (no binding, no driver reads it), and [`fcs,fusb302.yaml`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/Documentation/devicetree/bindings/usb/fcs,fusb302.yaml#L37) is `additionalProperties: false`, so `dtbs_check` will reject it. The same pin is already fully described by the `interrupt-parent`/`interrupts` pair right above.

```suggestion
+		interrupt-parent = <&gpio1>;
+		interrupts = <RK_PA7 IRQ_TYPE_LEVEL_LOW>;
```

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1134 @@
...
++			altmodes {
++				#address-cells = <1>;
++				#size-cells = <0>;
++
++				altmode@0 {
++					reg = <0>;
++					svid = <0xff01>;
++					vdo = <0xffffffff>;
++				};
++			};
```
openwrt-ai, line 446: This `altmodes` block does not match the binding on three counts ([`usb-connector.yaml:156-180`](``https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/Documentation/devicetree/bindings/connector/usb-connector.yaml#L156-L180)):`` the child must be named `displayport` (`patternProperties: "^(displayport)$"` under `additionalProperties: false`, so `altmode@0` plus its `reg`/`#address-cells`/`#size-cells` are all rejected), and `svid` is `uint16`, so it needs `/bits/ 16` rather than a 32-bit cell. Upstream RK3588 boards use the short form — [`rk3588-rock-5b-5bp-5t.dtsi:341-346`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-rock-5b-5bp-5t.dtsi#L341-L346).

```suggestion
+			altmodes {
+				displayport {
+					svid = /bits/ 16 <0xff01>;
+					vdo = <0xffffffff>;
+				};
+			};
```

Note this drops four lines, so the `@@ -0,0 +1,1115 @@` hunk header needs regenerating (`make target/linux/refresh` with rockchip selected) once all the fixes in this round are folded in.

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1134 @@
...
++		ports {
++			#address-cells = <1>;
++			#size-cells = <0>;
++			port@0 {
++				reg = <0>;
++				usbc0_role_sw: endpoint@0 {
++					remote-endpoint = <&dwc3_0_role_switch>;
++				};
++			};
++		};
```
openwrt-ai: This `ports` node hangs off the tcpc, not off the connector. [`fcs,fusb302.yaml`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/Documentation/devicetree/bindings/usb/fcs,fusb302.yaml#L12-L37) is `additionalProperties: false` and allows only `compatible`/`reg`/`interrupts`/`vbus-supply`/`connector`, so `dtbs_check` will flag it, and the port numbering under the connector below then shifts: per `usb-connector.yaml` `port@0` is HS, `port@1` is SS, `port@2` is SBU, but here `port@0`/`port@1` are wired to the usbdp-phy orientation/mode switches.

I believe it happens to work today only via a fallback: `fusb302_probe()` sets `tcpc_dev.fwnode` to the `connector` child ([`fusb302.c:1677-1684`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/usb/typec/tcpm/fusb302.c#L1677-L1684)), so [`fwnode_usb_role_switch_get(tcpc->fwnode)`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/usb/typec/tcpm/tcpm.c#L7920-L7922) walks the *connector's* graph and finds only the phy; the role switch is then picked up by the `usb_role_switch_get(port->dev)` fallback, which walks the fusb302 node and does find this endpoint.

The upstream RK3588 arrangement puts all three data buses under the connector's `ports` and needs no fallback — [`rk3588-rock-5b-5bp-5t.dtsi:348-372`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-rock-5b-5bp-5t.dtsi#L348-L372) (`port@0` HS → [`&usb_host0_xhci` port at L1024-L1032](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-rock-5b-5bp-5t.dtsi#L1024-L1032), `port@1` SS and `port@2` SBU → [`&usbdp_phy0` port at L991-L1010](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-rock-5b-5bp-5t.dtsi#L991-L1010)). Was the BSP-style graph chosen deliberately, or is this worth reshaping to the upstream form before the patch goes upstream?

---

openwrt-ai (commented):
Re-reviewed at `cd4904a`. Both commit-message findings from the previous round are fixed — the image name is now `openwrt-rockchip-armv8-radxa_e54c-ext4-sysupgrade.img.gz` and the power line carries a current rating — and the patch now has the `From`/`From:`/`Date:`/`Subject:`/`Signed-off-by:` header block. The commit message matches what the commit does; nothing to flag there.

The only code change since `3f39945` is the USB-C rework in the DTS patch: `usbc0` renamed and given `int-n-gpios` + a `ports` node, a `usb-c-connector` child with `altmodes` and an orientation/mode-switch graph to `&usbdp_phy0`, and `&usb_host0_xhci` switched to `dr_mode = "otg"` with a role-switch endpoint. The `&usbdp_phy0` port graph and the `dr_mode`/`usb-role-switch` pairing look right; five comments below on the rest, three of which are `dtbs_check`-level binding mismatches and one of which (the dropped `vbus-supply`) looks like an unintended deletion.

CI on `cd4904a` is green, so nothing CI-grounded.

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1131 @@
...
++		pinctrl-names = "default";
++		pinctrl-0 = <&cc_int_l>;
++		vbus-supply = <&vbus_typec0>;
++		status = "okay";
++
++		ports {
++			#address-cells = <1>;
++			#size-cells = <0>;
++		};
++
```
openwrt-ai: The reshaping of the graph moved all three endpoints under the connector (good — that is the upstream arrangement), but left this now-empty `ports { }` node behind on the tcpc. It has no child `port@N` at all, so it describes nothing, and `fcs,fusb302.yaml` is [`additionalProperties: false` with only `compatible`/`reg`/`interrupts`/`vbus-supply`/`connector` allowed](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/Documentation/devicetree/bindings/usb/fcs,fusb302.yaml#L13-L37), so `dtbs_check` will still reject it. [`rk3588-rock-5b-5bp-5t.dtsi:311-318`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-rock-5b-5bp-5t.dtsi#L311-L318) goes straight from `status` to the `connector` child.

```suggestion
+
```

This drops 5 lines, so the `@@ -0,0 +1,1112 @@` hunk header needs regenerating (`make target/linux/refresh` with rockchip selected in `.config`) once the fixes in this round are folded in.

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1131 @@
...
++			try-power-role = "sink";
++			op-sink-microwatt = <1000000>;
++
++			sink-pdos =
++				<PDO_FIXED(5000, 1000, PDO_FIXED_USB_COMM)>;
++			source-pdos =
++				<PDO_FIXED(5000, 3000, PDO_FIXED_USB_COMM)>;
++
++			altmodes {
++				displayport@0 {
```
openwrt-ai: The `svid` type fix landed, but the node name is still `displayport@0`. `altmodes` is [`additionalProperties: false` with `patternProperties: "^(displayport)$"`](``https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/Documentation/devicetree/bindings/connector/usb-connector.yaml#L163-L166),`` so the `@0` suffix makes the node fail the pattern and it gets rejected as an unexpected property. It is also a unit address with no `reg` backing it, which `dtc` warns about separately. The binding's own example and every upstream RK3588 board use the bare name — [`usb-connector.yaml:397-402`](``https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/Documentation/devicetree/bindings/connector/usb-connector.yaml#L397-L402),`` [`rk3588-rock-5b-5bp-5t.dtsi:340-345`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-rock-5b-5bp-5t.dtsi#L340-L345).

```suggestion
+				displayport {
```

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1131 @@
...
++				};
++			};
++
++			ports {
++				#address-cells = <1>;
++				#size-cells = <0>;
++
++				port@0 {
++					reg = <0>;
++					usbc0_hs: endpoint@0 {
```
openwrt-ai: `endpoint@0` carries a unit address, but `port@0` declares no `#address-cells`/`#size-cells` and the endpoint has no `reg` — `dtc` flags that as `unit_address_vs_reg`. The sibling `port@1`/`port@2` right below (lines 466 and 473) already use the bare `endpoint` form, as does [`rk3588-rock-5b-5bp-5t.dtsi:352-355`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-rock-5b-5bp-5t.dtsi#L352-L355).

```suggestion
+					usbc0_hs: endpoint {
```

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1131 @@
...
++
++&usb_host0_ehci {
++	status = "okay";
++};
++
++&usb_host0_xhci {
++	dr_mode = "otg";
++	status = "okay";
++
++	usb-role-switch;
```
openwrt-ai: nit: `usb-role-switch;` is a property but sits after `status = "okay";` and after a blank line, which breaks the convention every other node in this DTS follows (properties, then `status` last, then a blank line, then subnodes). Moving it up also matches [`rk3588-rock-5b-5bp-5t.dtsi:1024-1027`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-rock-5b-5bp-5t.dtsi#L1024-L1027).

```suggestion
+	dr_mode = "otg";
+	usb-role-switch;
+	status = "okay";
+
```

Line count is unchanged, so this one costs no extra hunk churn. Separately, `dr_mode = "otg"` is redundant — [`rk3588-base.dtsi:474`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-base.dtsi#L474) already sets it on `usb_host0_xhci`, which is why rock-5b omits it; dropping it too is optional but would be one less line to justify upstream.

---

openwrt-ai (commented):
Re-reviewed at `f960d9f`. The only PR-side change since `cd4904a` is the USB-C graph in the DTS patch; everything else in the range is base-branch drift (airoha ATF/U-Boot rework, realtek compatibles).

Three of the five findings from the previous round are fixed: the node is back to `usb-typec@22`, `vbus-supply = <&vbus_typec0>` is restored, and the BSP-only `int-n-gpios` is gone. The graph has also been reshaped to the upstream RK3588 form — HS/SS/SBU all under the connector's `ports`, `port@0` → `&usb_host0_xhci`, `port@1`/`port@2` → `&usbdp_phy0` — which is the right structure and drops the role-switch fallback path I described last time. The `&usbdp_phy0` endpoint numbering and the `sbu1-dc-gpios`/`sbu2-dc-gpios` properties check out against [`phy-rockchip-usbdp.yaml:91-99`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/Documentation/devicetree/bindings/phy/phy-rockchip-usbdp.yaml#L91-L99), and the `@@ -0,0 +1,1112 @@` count is correct for the current file.

Four comments below, all in the reshaped block: two `dtbs_check`-level leftovers from the previous shape (the now-empty `ports { }` on the tcpc, and `displayport@0`), one `dtc` unit-address warning, and one nit. The commit message is unchanged and still matches what the commit does, so nothing to flag there.

CI on `f960d9f` is green (all three FormalityCheck runs passed), so nothing CI-grounded.

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
```diff
@@ -0,0 +1,1125 @@
...
++		usbdp_phy0_sbu: endpoint@1 {
++			reg = <1>;
++			remote-endpoint = <&usbc0_sbu>;
++		};
++	};
++};
++
++&usb_host0_ehci {
++	status = "okay";
++};
```
openwrt-ai, line 1099: `usb_host0_ohci` is left disabled while `usb_host0_ehci` is enabled, so low-speed and full-speed devices on the two USB 2.0 Type-A ports will not enumerate — only high-speed ones will.

Both controllers sit behind the same PHY: [`usb_host0_ehci` at rk3588-base.dtsi:489](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-base.dtsi#L489-L498) and [`usb_host0_ohci` at :500](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-base.dtsi#L500-L509) both use `phys = <&u2phy2_host>`, which this DTS enables (patch line 1057, `phy-supply = <&vcc5v_usb_hub>` — i.e. the USB 2.0 hub feeding the Type-A ports). EHCI only drives HS; the companion OHCI is what handles LS/FS, which is exactly what a keyboard, mouse or FS USB-serial adapter needs. The same file already enables both halves for host1 (lines 1115-1121), and upstream rock-5b enables both for host0 too ([`rk3588-rock-5b-5bp-5t.dtsi:1016-1022`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-rock-5b-5bp-5t.dtsi#L1016-L1022)).

```suggestion
+&usb_host0_ehci {
+	status = "okay";
+};
+
+&usb_host0_ohci {
+	status = "okay";
+};
```

The one counter-example I found is `rk3588s-rock-5c.dts:922`, which also enables EHCI without OHCI — so if this is deliberate (e.g. LS/FS was verified to work anyway on your unit) please say so. Note the suggestion adds 4 lines, so the `@@ -0,0 +1,1106 @@` hunk header needs regenerating with `make target/linux/refresh` (rockchip selected in `.config`).

---

openwrt-ai (commented):
Re-reviewed at `f567fef`. The only change since `f960d9f` is the DTS patch, and all four findings from the previous round are fixed:

- the now-empty `ports { }` on the tcpc is gone, so `fcs,fusb302`'s `additionalProperties: false` is satisfied;
- `displayport@0` → `displayport`, matching the `^(displayport)$` pattern in `usb-connector.yaml`;
- `usbc0_hs: endpoint@0` → `endpoint`, so no `unit_address_vs_reg` warning and consistent with the sibling `port@1`/`port@2`;
- `usb-role-switch` moved above `status = "okay"`, and the redundant `dr_mode = "otg"` dropped — safe, since the DTS includes `rk3588s.dtsi` → [`rk3588-base.dtsi:474`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-base.dtsi#L474) already sets it on `usb_host0_xhci`.

The `@@ -0,0 +1,1106 @@` count is correct (1125 patch lines − 19 header lines), and the connector graph now matches the upstream RK3588 arrangement in `rk3588-rock-5b-5bp-5t.dtsi` node for node.

The commit message is unchanged and still matches what the commit does, so nothing to flag there.

CI on `f567fef` is green (all three FormalityCheck runs passed), so nothing CI-grounded.

One comment below, on the USB 2.0 host controllers.

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
default-username-852: I have verified that plugging a USB mouse into all USB type A ports are detected. I added the package `kmod-usb-hid` and plugged a mouse into all type A ports, and for all of them I got kernel log messages:

```
[  104.332444] usb 3-1: new full-speed USB device number 2 using ohci-platform
[  104.590398] input: RAPOO Rapoo 2.4G Wireless Device as /devices/platform/fc8c0000.usb/usb3/3-1/3-1:1.0/0003:24AE:2015.0001/input/input2

[  113.721922] usb 4-1.2: new full-speed USB device number 3 using ehci-platform
[  113.966217] input: RAPOO Rapoo 2.4G Wireless Device as /devices/platform/fc800000.usb/usb4/4-1/4-1.2/4-1.2:1.0/0003:24AE:2015.0003/input/input7

[  122.931501] usb 4-1.1: new full-speed USB device number 4 using ehci-platform
[  123.176213] input: RAPOO Rapoo 2.4G Wireless Device as /devices/platform/fc800000.usb/usb4/4-1/4-1.1/4-1.1:1.0/0003:24AE:2015.0005/input/input12
```

so it seems that the mouse is detected as it should be. It it is also a full speed device, not high speed as seen in the logs.

`cat /dev/input/event3` gives output when I move the mouse, no matter which port it is plugged into, so input is properly received from the mouse as well.

openwrt-ai (commented):
Re-reviewed at `f2d1193`. This push is a pure rebase: the tree for every path this PR owns (`package/boot/uboot-rockchip/Makefile`, `target/linux/rockchip/**`) is byte-identical to `f567fef` — the DTS patch hashes match, and the commit message is unchanged. The base itself also moved without touching `target/linux/rockchip/` or `package/boot/uboot-rockchip/`, so there is no new code and no new interaction to review.

The one open finding from the last round — `usb_host0_ohci` left disabled — is answered by the boot log you posted. `fc8c0000.usb` is [`usb_host1_ohci`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/rockchip/rk3588-base.dtsi#L522), which this DTS does enable, and the two full-speed enumerations on `fc800000.usb` (host0 EHCI) are at `4-1.1` / `4-1.2`, i.e. behind the USB 2.0 hub fed by `vcc5v_usb_hub`. The hub's transaction translator handles LS/FS for anything on those Type-A ports, so the companion OHCI on host0 has nothing to serve. Nothing to change; I've resolved that thread.

CI on `f2d1193` is green (all three FormalityCheck runs, the rockchip/armv8 kernel build, the kernel-patch check, and both package builds), so nothing CI-grounded.

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
openwrt-ai: Thanks — that log settles it, and no change is needed here.

`fc8c0000.usb` is `usb_host1_ohci`, which this DTS already enables, so the first line is the host1 OHCI doing its job. The other two enumerations are on `fc800000.usb` (host0 EHCI) at `4-1.1` and `4-1.2`, i.e. downstream of the USB 2.0 hub fed by `vcc5v_usb_hub` — the hub's transaction translator handles LS/FS for anything plugged into those Type-A ports, so the companion OHCI on host0 has nothing left to serve. My original reasoning assumed the Type-A ports hung off the root port directly.

Considering this resolved from my side (I can't flip the resolved marker myself).

---

openwrt-ai (commented):

---

target/linux/rockchip/armv8/base-files/etc/board.d/01_leds
openwrt-ai: fixed, thanks

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
openwrt-ai: fixed, thanks

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
openwrt-ai: fixed, thanks

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
openwrt-ai: solved by dropping the property entirely, thanks

---

target/linux/rockchip/image/armv8.mk
openwrt-ai: fixed, thanks

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
openwrt-ai: fixed, thanks — `pwm.h` is gone and `usb/pd.h` is now genuinely used by the `PDO_FIXED()` entries in the connector.

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
openwrt-ai: fixed, thanks

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
openwrt-ai: fixed, thanks

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
openwrt-ai: fixed, thanks

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
openwrt-ai: fixed, thanks

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
openwrt-ai: fixed, thanks

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
openwrt-ai: fixed, thanks — the graph was reshaped to the upstream form with all three buses under the connector.

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
openwrt-ai: fixed, thanks

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
openwrt-ai: fixed, thanks

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
openwrt-ai: fixed, thanks

---

target/linux/rockchip/patches-6.18/403-arm64-dts-rockchip-Add-Radxa-E54C.patch
openwrt-ai: fixed, thanks — `usb-role-switch` moved above `status` and the redundant `dr_mode` dropped.

---

PR #24433, opened 2026-07-26
qualcommax: use the upstream fix for the optional wcss_q6_bcr_reset

JuliusBairaktaris:
The `wcss_q6_bcr_reset` reset is not present in the IPQ8074 or IPQ6018 device trees, so `devm_reset_control_get_exclusive()` returns an error and takes `q6v5_wcss_probe()` down with it. Patch 0116 works around this by adding a per-SoC `bcr_reset_required` flag to `struct wcss_data` and skipping the acquisition when the flag is false.

Linux 6.19 fixed the same problem upstream in commit 641092c1bc1b ("remoteproc: qcom_q6v5_wcss: use optional reset for wcss_q6_bcr_reset"), by switching to `devm_reset_control_get_optional_exclusive()`, which returns NULL instead of an error when the reset is absent from DT. That covers everything the flag did, without carrying per-SoC data for it.

Replace 0116 with a backport of the upstream commit, and drop the now-dead `.bcr_reset_required` initialiser from 0136 (an added line) and 0905 (a context line). 0119 and 0908 are refresh-only.

Behaviour is unchanged on every affected SoC. Before and after, `wcss_q6_bcr_reset` ends up NULL on IPQ8074 and IPQ6018, and its only consumer, `reset_control_deassert()` in `q6v5_wcss_qcs404_power_on()`, is a no-op on NULL - on IPQ8074 and IPQ6018 it is never even reached, since that function belongs to the QCS404 ops. QCS404 declares the reset in DT, so the optional getter still finds it there.

### Second commit: the IPQ6018 SSR name on 6.18

Rebuilding 0905's context window surfaced a separate defect in `patches-6.18`, so it is fixed here rather than left in the hunk this PR touches.

0905 exists to give IPQ6018 the `wcnss` SSR and glink name — the `q6wcss` it would otherwise carry is not a name RPM knows, and on that SoC the string reaches RPM when remoteproc stops, taking the board down to a hang in XBL with no watchdog running. In `patches-6.18` the descriptor hunk sits on `wcss_ipq8074_res_init` instead. The two `wcss_data` initialisers carry identical context either side of `.ssr_name`, so when the series was created from `patches-6.12` the hunk matched the first one it found and was refreshed there.

The result is that the two series disagree. Replaying every qualcommax patch that touches `qcom_q6v5_wcss.c` onto pristine v6.12.100 and v6.18.39 with `patch -F0`, zero fuzz and zero offset in both:

```
6.12:  wcss_ipq8074_res_init  .ssr_name = "q6wcss"
       wcss_ipq6018_res_init  .ssr_name = "wcnss"     <- as 0905 intends

6.18:  wcss_ipq8074_res_init  .ssr_name = "wcnss"     <- 0905 landed here
       wcss_ipq6018_res_init  .ssr_name = "q6wcss"
```

Re-anchoring the hunk on `wcss_ipq6018_res_init` makes 6.18 produce the same two names as 6.12. The patch body is byte-identical apart from the hunk header.

### Testing

Tested on a Xiaomi AX3600 (IPQ8074), `KERNEL_PATCHVER` 6.12: the WCSS remoteproc boots, stops and re-boots cleanly through the ath11k bring-up sequence, both radios come up, and there are no "unable to acquire wcss_q6_bcr_reset" messages or backtraces. I have no IPQ6018 or QCS404 hardware here; for those the argument is the NULL-either-way reasoning above and the patch replay, not a measurement.


target/linux/qualcommax/patches-6.12/0116-v6.19-remoteproc-qcom_q6v5_wcss-use-optional-reset-fo.patch
```diff
@@ -0,0 +1,30 @@
...
+ wcss_q6_bcr_reset
+
+The "wcss_q6_bcr_reset" is not used on IPQ8074, and IPQ6018. Use
+devm_reset_control_get_optional_exclusive() for this reset so that
+probe() does not fail on platforms where it is not used.
+
+Signed-off-by: Alexandru Gagniuc <mr.nuke.me@gmail.com>
+Link: https://lore.kernel.org/r/20251129013207.3981517-2-mr.nuke.me@gmail.com
+Signed-off-by: Bjorn Andersson <andersson@kernel.org>
+Signed-off-by: Julius Bairaktaris <julius@bairaktaris.de>
```
openwrt-ai: This trailer is not in the upstream commit — I fetched `641092c1bc1b` from torvalds/linux and its trailer block ends at `Signed-off-by: Bjorn Andersson <andersson@kernel.org>`. Everything else in this file (message, hunk body, diffstat, the 7-tab continuation indent) is byte-identical to upstream, so this line is the only deviation.

A backported patch file should reproduce the upstream commit verbatim; your own sign-off belongs on the OpenWrt commit, where it already is. All 50 `NNNN-vX.Y-*.patch` files in `target/linux/qualcommax/patches-6.12/` end their trailer block with the upstream maintainer's Signed-off-by — none append the backporter's.

```suggestion
Signed-off-by: Bjorn Andersson <andersson@kernel.org>
```

---

target/linux/qualcommax/patches-6.12/0116-v6.19-remoteproc-qcom_q6v5_wcss-use-optional-reset-fo.patch
```diff
@@ -0,0 +1,30 @@
+From 641092c1bc1bbc3be059d9d723b1cec10a368617 Mon Sep 17 00:00:00 2001
```
openwrt-ai: nit: this puts a `vX.Y`-prefixed upstream backport into the downstream/pending number block. In `target/linux/qualcommax/patches-6.12/` the split is currently clean: every `-vX.Y-` patch is numbered 0036–0090, and everything from 0102 upwards is unprefixed downstream/vendor work (0102, 0111–0123, 0129–0136, 0184–0189, 03xx/04xx/07xx/08xx/09xx). `0116-v6.19-` would be the first exception.

Reusing 0116 keeps the churn minimal, and the 0119/0136/0905/0908 offsets have to be refreshed either way. Was keeping the slot deliberate, or would a number in the upstream block be preferred here?

---

openwrt-ai (commented):
## Commit checks

- `de3e57ab` "qualcommax: use the upstream fix for the optional wcss_q6_bcr_reset" — the justification paragraph names the wrong consumer. It says *"its only consumer, `reset_control_deassert()` in `q6v5_wcss_stop()`, is a no-op on NULL"*, but the only read of `wcss->wcss_q6_bcr_reset` is [`reset_control_deassert()` at qcom_q6v5_wcss.c:312](https://github.com/gregkh/linux/blob/0b8f247169e487eff2d4c2dd531bc43f7efda2cb/drivers/remoteproc/qcom_q6v5_wcss.c#L312), inside [`q6v5_wcss_qcs404_power_on()`](https://github.com/gregkh/linux/blob/0b8f247169e487eff2d4c2dd531bc43f7efda2cb/drivers/remoteproc/qcom_q6v5_wcss.c#L295-L312) — not `q6v5_wcss_stop()` ([line 708](https://github.com/gregkh/linux/blob/0b8f247169e487eff2d4c2dd531bc43f7efda2cb/drivers/remoteproc/qcom_q6v5_wcss.c#L708), which does not touch it). No OpenWrt patch in this target adds another user (`bcr_reset_required` / `wcss_q6_bcr_reset` only appear in 0116, 0136 and 0905).

  This actually strengthens the argument rather than weakening it: `q6v5_wcss_qcs404_power_on()` is reached only from `q6v5_qcs404_wcss_start()` via `q6v5_wcss_qcs404_ops`, so on IPQ8074/IPQ6018 the deassert is never executed at all — the NULL no-op is a second line of defence, not the primary one. Worth correcting the function name (and the identical sentence in the PR body) before this lands.

---

Otherwise the mechanics check out. I recomputed the refreshed hunk offsets in 0119, 0136, 0905 and 0908 against the removal of old 0116 (−1 line at the `struct wcss_data` member, −3 in `q6v5_wcss_init_reset`, −1 each at the two `wcss_data` initialisers) and every shift matches, including 0905's rebuilt context window and 0136's `+1187,20` → `+1183,19` count. `desc` stays live in `q6v5_wcss_init_reset()` via `aon_reset_required` / `wcss_q6_reset_required`, so dropping the `bcr_reset_required` use leaves no unused parameter. `reset_control_deassert()` returns 0 on NULL ([core.c:556-559](https://github.com/gregkh/linux/blob/0b8f247169e487eff2d4c2dd531bc43f7efda2cb/drivers/reset/core.c#L556-L559)), and `wcss` is zero-allocated, so behaviour is unchanged on IPQ8074/IPQ6018 as described. Kernel compared against: v6.12.94 (`KERNEL_PATCHVER:=6.12`).

---

openwrt-ai (commented):

The force-push does two things, both good:

- The commit message now names `q6v5_wcss_qcs404_power_on()` as the consumer instead of `q6v5_wcss_stop()`, which resolves the [Commit checks item from my previous review](https://github.com/openwrt/openwrt/pull/24433#pullrequestreview-4782070184). (The PR description above still carries the old `q6v5_wcss_stop()` sentence — cosmetic only, since it does not land in the tree.)
- Dropping `Signed-off-by: Julius Bairaktaris` from the backported patch's header matches how backports are carried in this directory: `0073-v6.20-*`, `0084-v7.2-*` and `0056-v6.17-*` all keep only the upstream trailers.

That trailer sits above the `---` separator, so the diff hunks and the `@@ -876,7 +876,8 @@` header are untouched and the offset analysis from the previous review still holds. `devm_reset_control_get_exclusive(dev, "wcss_q6_bcr_reset")` is still the pre-image at [qcom_q6v5_wcss.c:814](https://github.com/gregkh/linux/blob/0b8f247169e487eff2d4c2dd531bc43f7efda2cb/drivers/remoteproc/qcom_q6v5_wcss.c#L814) in both v6.12.94 and v6.18.39, and `KERNEL_PATCHVER:=6.12` is the only patch series in this target. CI on `10a2477f` is green, including `Check Kernel patches` and all three subtarget kernel builds.

---

target/linux/qualcommax/patches-6.18/0905-remoteproc-q6v5_wcss-change-ssr-name-for-ipq6018-wif.patch
```diff
@@ -30,10 +30,10 @@ Signed-off-by: Mantas Pucka <mantas@8devices.com>
  
  	if (desc->ssctl_id) {
  		wcss->sysmon = qcom_add_sysmon_subdev(rproc,
-@@ -1162,7 +1162,7 @@ static const struct wcss_data wcss_ipq80
+@@ -1158,7 +1158,7 @@ static const struct wcss_data wcss_ipq80
```
openwrt-ai: nit (pre-existing on `main`, **not** introduced by this PR): in the 6.18 series this hunk lands on `wcss_ipq8074_res_init`, not on `wcss_ipq6018_res_init` — note the `wcss_ipq80` function context here versus [`wcss_ipq60` on the same line of the 6.12 copy](``https://github.com/openwrt/openwrt/blob/f21eb754e67d7289e51bd7803189100286ddd0bc/target/linux/qualcommax/patches-6.12/0905-remoteproc-q6v5_wcss-change-ssr-name-for-ipq6018-wif.patch#L33).`` The net effect is that the two SSR names are swapped between the series.

I replayed every qualcommax patch touching `drivers/remoteproc/qcom_q6v5_wcss.c` in series order onto pristine v6.12.100 / v6.18.39 (`patch -F0`, zero fuzz, zero offset). Resulting `wcss_data` initialisers:

```
6.12:  wcss_ipq8074_res_init  .ssr_name = "q6wcss"
       wcss_ipq6018_res_init  .ssr_name = "wcnss"     <- 0905 applied here

6.18:  wcss_ipq8074_res_init  .ssr_name = "wcnss"     <- 0905 applied here
       wcss_ipq6018_res_init  .ssr_name = "q6wcss"
```

So on 6.18, IPQ8074 gets the `wcnss` SSR/glink name that `0905` intends for IPQ6018, and IPQ6018 keeps the `q6wcss` name [`0136` seeds it with](``https://github.com/openwrt/openwrt/blob/f21eb754e67d7289e51bd7803189100286ddd0bc/target/linux/qualcommax/patches-6.18/0136-remoteproc-qcom-wcss-populate-driver-data-for-IPQ601.patch#L44)`` — the opposite of 6.12. It applies cleanly either way, which is why `Check Kernel patches` stays green. I reproduced the identical result replaying the base commit `c9833b99`, so this predates the PR.

Flagging it only because this PR rebuilds this hunk's context window, so it is the natural moment to notice it. The desired end state is that `0905` rewrites `wcss_ipq6018_res_init`'s `.ssr_name` in the 6.18 series too, which means re-anchoring the hunk one struct further down rather than a textual edit here — is fixing that in scope for this PR, or would you rather have it as a separate patch?

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push `10a2477f` → `f21eb754`); no new issues found in the PR's own changes.

What changed since my [last review](https://github.com/openwrt/openwrt/pull/24433#pullrequestreview-4783614315): a rebase onto a `main` that now carries `target/linux/qualcommax/patches-6.18`, the same treatment extended to that series, and an `Assisted-by:` trailer. No Commit checks items — the message matches the diff, including the new claims that `0119`/`0908` are refresh-only and that `.bcr_reset_required` is dropped as an added line in `0136` and as a context line in `0905`.

Verification, since the diff now spans two patch series: I took pristine `drivers/remoteproc/qcom_q6v5_wcss.c` from v6.12.100 and v6.18.39 (matching `LINUX_VERSION-6.12 = .100` and `LINUX_VERSION-6.18 = .39`) and replayed every qualcommax patch that touches that file, in series order, with `patch -F0`. No generic `backport-`/`pending-`/`hack-` patch touches it, so the target series is the whole chain.

- Both series apply at this head with zero fuzz and zero offset, so all four refreshed patches (`0119`, `0136`, `0905`, `0908`) carry correct line numbers in each series.
- The two `0116-v6.19-*.patch` files are byte-identical, matching the "applies unchanged to either" claim.
- Diffing the replay result against the same replay on base `c9833b99` yields exactly the intended delta in **both** series and nothing else: the `bcr_reset_required` member and its three initialisers removed, and the getter switched to `devm_reset_control_get_optional_exclusive()` with the `IS_ERR` check retained. No `bcr_reset_required` reference remains under `target/`.
- The `0116-v6.19-` name matches the neighbouring backport convention in both directories (`0073-v7.0-`, `0084-v7.2-`, …), and the `v6.19` prefix is correct for both a 6.12 and a 6.18 base.
- CI on `f21eb754` is green: `Check Kernel patches` for 6.12 and 6.18, and all six ipq50xx/ipq60xx/ipq807x kernel builds.

One inline `nit:` on a pre-existing 6.18 discrepancy that this PR happens to re-anchor — it is not caused by this change, but this is the patch that touches it.

---

JuliusBairaktaris:
In scope and pushed as a second commit: a prepared 6.18 tree has the two `.ssr_name`s swapped, so re-anchoring the hunk on `wcss_ipq6018_res_init` is what makes both series produce the same names.

openwrt-ai (commented):
Reviewed 1 new commit (`4cd59c3a` "qualcommax: fix the IPQ6018 SSR name on the 6.18 patch series"); no new issues found.

This is the fix for the [`nit:` I raised on the previous review](https://github.com/openwrt/openwrt/pull/24433#discussion_r3694619447), and it does exactly what it says. The whole diff since `f21eb754` is one line — `@@ -1158,7 +1158,7 @@ static const struct wcss_data wcss_ipq80` → `@@ -1172,7 +1172,7 @@ static const struct wcss_data wcss_ipq60` in `patches-6.18/0905-*` — so the "byte-identical apart from the hunk header" claim holds literally.

Re-verified by replaying every qualcommax patch that touches `drivers/remoteproc/qcom_q6v5_wcss.c`, in series order, with `patch -F0` onto pristine v6.12.100 / v6.18.39:

| series | `wcss_ipq8074_res_init` | `wcss_ipq6018_res_init` |
| --- | --- | --- |
| 6.12 @ head | `"q6wcss"` | `"wcnss"` |
| 6.18 @ base `c9833b99` | `"wcnss"` | `"q6wcss"` |
| 6.18 @ `f21eb754` | `"wcnss"` | `"q6wcss"` |
| **6.18 @ head `4cd59c3a`** | **`"q6wcss"`** | **`"wcnss"`** |

So the two series now agree, and `wcnss` lands on the SoC `0905` intends. Both series still apply with zero fuzz and zero offset at this head, so the re-anchor did not disturb `0908`'s or any later patch's line numbers. Checking the pre-image directly: after everything before `0905` is applied, line 1168 of the 6.18 file is `static const struct wcss_data wcss_ipq6018_res_init = {` and line 1172 is `.crash_reason_smem = WCSS_CRASH_REASON,`, matching the new hunk header and its `wcss_ipq60` function context.

Commit-message checks: the subject scope matches the single file touched; the `Fixes: f939f7e2e9d5 ("qualcommax: create files for v6.18 (from v6.12)")` reference resolves to that exact commit on `main`; the replay claim in the body reproduces exactly as written. Nothing to flag.

CI on `4cd59c3a`: no failures — `Check Kernel patches` green for both 6.12 and 6.18, `FormalityCheck` green, ipq807x/ipq50xx 6.12 kernel builds green, the four remaining builds still running at review time.

---

robimarko:

PR #24438, opened 2026-07-27
build: do not rebuild tools in SDK

### LINKED ISSUE #24273: [24.10] SDK: fails to sign package index (missing 'key-build')
When attempting to build a package from openwrt/packages (specifically radicale3 in this instance), the build itself succeeds but signing of the package index fails due to lack of private & public key for that build (vs a release build which will have created a key-build key).

```
Generating index for package ./python3-uuid_3.11.14-r1_aarch64_cortex-a76.ipk
Generating index for package ./python3-vobject_0.9.9-r1_aarch64_cortex-a76.ipk
Generating index for package ./python3-xml_3.11.14-r1_aarch64_cortex-a76.ipk
Generating index for package ./python3_3.11.14-r1_aarch64_cortex-a76.ipk
Generating index for package ./radicale3_3.7.2-r1_aarch64_cortex-a76.ipk
Signing package index...
Cannot open file '/home/user/Build/snapshot-24.10-radicale3/openwrt-sdk/key-build' for reading
Cannot open file '/home/user/Build/snapshot-24.10-radicale3/openwrt-sdk/key-build' for reading
Cannot open file '/home/user/Build/snapshot-24.10-radicale3/openwrt-sdk/key-build' for reading
Cannot open file '/home/user/Build/snapshot-24.10-radicale3/openwrt-sdk/key-build' for reading
Cannot open file '/home/user/Build/snapshot-24.10-radicale3/openwrt-sdk/key-build' for reading
Cannot open file '/home/user/Build/snapshot-24.10-radicale3/openwrt-sdk/key-build' for reading
Cannot open file '/home/user/Build/snapshot-24.10-radicale3/openwrt-sdk/key-build' for reading
make[1]: *** [package/Makefile:125: package/index] Error 1
make[1]: Leaving directory '/home/user/Build/snapshot-24.10-radicale3/openwrt-sdk'
make: *** [/home/user/Build/snapshot-24.10-radicale3/openwrt-sdk/include/toplevel.mk:226: package/index] Error 2
```

EDIT: As a workaround, disabling 'Cryptographically sign packages', works.

mcprat:
This is a much simpler alternative to previous suggestion, which I believe is an equivalent fix
Fixes: #7299 
maybe-fixes: #24273 
Closes: #10255

A long-standing problem with the SDK is that tools can rebuild due to the lack of a stamp file. This can cause ABI mismatch, as the new binaries are built against a different libc than is packaged.

The SDK Makefile defines Make variable $(SDK)
which can be used as a condition to prevent this.
Each stage in a build has the stampfile of the previous stage as a prerequisite for building, such as $(tools/stamp-compile). By blocking the definition of those variables,
the stamp file path is never defined, so that prerequisite expands to nothing.

These files are not directly included in the SDK,
but they are imported through the "base" feed
along with the packages in the main openwrt repo.

Also, for consistency, block the rule that is defined using the toolchain stamp file, so as to not have a blank rule.

As usual, it is still possible to rebuild some tools, but after this it is no longer automatic
as a result of including the main repo tools targets in the form of stamp file definitions within the feeds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,9 @@ else
 # Include the test suite Makefile if it exists
 -include tests/Makefile
 
+ifeq ($(SDK),)
 $(toolchain/stamp-compile): $(tools/stamp-compile) $(if $(CONFIG_BUILDBOT),toolchain_rebuild_check)
+endif
 $(target/stamp-compile): $(toolchain/stamp-compile) $(tools/stamp-compile) $(BUILD_DIR)/.prepared
 $(package/stamp-compile): $(target/stamp-compile) $(package/stamp-cleanup)
 $(package/stamp-install): $(package/stamp-compile)
--- a/tools/Makefile
+++ b/tools/Makefile
@@ -236,6 +236,10 @@ $(curdir)/ := .config prereq
 $(curdir)/install: $(curdir)/compile
 
 tools_enabled = $(foreach tool,$(sort $(tools-y) $(tools-)),$(if $(filter $(tool),$(tools-y)),y,n))
+
+ifeq ($(SDK),)
 $(eval $(call stampfile,$(curdir),tools,compile,,_$(subst $(space),,$(tools_enabled)),$(STAGING_DIR_HOST)))
 $(eval $(call stampfile,$(curdir),tools,check,$(TMP_DIR)/.build,,$(STAGING_DIR_HOST)))
+endif
+
 $(eval $(call subdir,$(curdir)))
--- a/toolchain/Makefile
+++ b/toolchain/Makefile
@@ -86,7 +86,10 @@ endif
 
 $(curdir)/install: $(curdir)/compile
 
+ifeq ($(SDK),)
 $(eval $(call stampfile,$(curdir),toolchain,compile,$(TOOLCHAIN_DIR)/stamp/.gcc_final_installed,,$(TOOLCHAIN_DIR)))
 $(eval $(call stampfile,$(curdir),toolchain,check,$(TMP_DIR)/.build))
+endif
+
 $(eval $(call subdir,$(curdir)))
 


mcprat:
@danielfdickinson this method, if it works, is extremely easy to test, you don't have to rebuild your own SDK

first I am curious, what the list of downloaded objects are in your existing SDK
`$ ls dl`

pay attention to what builds first after this, it should be a download of the kernel or nothing.

1. take the SDK tarball you already have and unpack it into a new folder
2. install your feeds with `make package/symlinks` or however you usually do
3. before anything else, modify the Makefiles as I do in this PR, in the SDK they are in `feeds/base`

edit: `feeds/base` for 24.10 and older `feeds/base_root` for newer

normal openwrt vs SDK
`Makefile` --> `feeds/base/Makefile`
`tools/Makefile` --> `feeds/base/tools/Makefile`
`toolchain/Makefile` --> `feeds/base/toolchain/Makefile`

and then do your build

Makefile
```diff
@@ -43,7 +43,9 @@ else
 # Include the test suite Makefile if it exists
 -include tests/Makefile
 
+ifeq ($(SDK),)
 $(toolchain/stamp-compile): $(tools/stamp-compile) $(if $(CONFIG_BUILDBOT),toolchain_rebuild_check)
+endif
```
openwrt-ai, line 48: I can't find a path where these three files are ever parsed by an SDK build, which would make the guards a no-op. Could you show the make trace that proves otherwise?

What I traced:

1. The SDK's own top-level Makefile only pulls in [`rules.mk`, `depends.mk`, `subdir.mk` and `package/Makefile`](https://github.com/openwrt/openwrt/blob/a279f54d86bb2ec28aba32d5ce5df574f25568b1/target/sdk/files/Makefile#L34-L37) — never `tools/Makefile`, `toolchain/Makefile`, or a second top-level `Makefile`.
2. The base feed is registered as [`src-git --root=package base ...`](https://github.com/openwrt/openwrt/blob/a279f54d86bb2ec28aba32d5ce5df574f25568b1/target/sdk/Makefile#L47), so `scripts/feeds` only symlinks `feeds/base -> feeds/base_root/package`. `feeds/base_root/tools/` is on disk but nothing points at it.
3. Metadata scanning only descends into [`package` and `target/linux`](https://github.com/openwrt/openwrt/blob/a279f54d86bb2ec28aba32d5ce5df574f25568b1/include/toplevel.mk#L82-L83), so `feeds/base_root/tools/Makefile` is never scanned either.
4. In an SDK, the catch-all [`%::` rule](https://github.com/openwrt/openwrt/blob/a279f54d86bb2ec28aba32d5ce5df574f25568b1/include/toplevel.mk#L221-L225) re-invokes the *SDK's* Makefile, so `tools/...` / `toolchain/...` targets don't resolve there at all.

Since `$(SDK)` is only ever set by the SDK's top Makefile, in the main tree these `ifeq` blocks are always taken, and in an SDK the files they live in appear never to be read. If the rebuild you're chasing (#7299 — `ucert`) is actually a `HOST_BUILD` under `package/system/ucert` rather than a `tools/` entry, the missing stamp would be a package host-build stamp (`*/stamp` is in `EXCLUDE_DIRS` of the SDK tarball), which these hunks wouldn't cover.

Also worth spelling out for the reproduction steps in your PR comment: if these files aren't read, editing them in `feeds/base_root` produces the same result either way, so a "no rebuild happened" outcome wouldn't confirm the fix.

---

tools/Makefile
```diff
@@ -236,6 +236,10 @@ $(curdir)/ := .config prereq
 $(curdir)/install: $(curdir)/compile
 
 tools_enabled = $(foreach tool,$(sort $(tools-y) $(tools-)),$(if $(filter $(tool),$(tools-y)),y,n))
+
+ifeq ($(SDK),)
```
openwrt-ai, line 240: nit: the in-tree spelling for "skip this in the SDK" is `ifndef SDK` — see [`package/Makefile:82`](https://github.com/openwrt/openwrt/blob/a279f54d86bb2ec28aba32d5ce5df574f25568b1/package/Makefile#L82) and [`include/toplevel.mk:198`](https://github.com/openwrt/openwrt/blob/a279f54d86bb2ec28aba32d5ce5df574f25568b1/include/toplevel.mk#L198). No existing site uses `ifeq ($(SDK),)`.

```suggestion
ifndef SDK
```

Same applies to `Makefile:46` and `toolchain/Makefile:89`.

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message accurately describes the diff; comments left inline.

---

mcprat:
it's funny this is so complex the bot can't keep track of it

tools/Makefile
mcprat: we have a habit of doing `ifeq` so that an empty definition `SDK=` counts as failing the check

danielfdickinson:
@mcprat The code in this PR does not exist in 24.10 SDK. After looking at this a bit, I'm hoping it's been long solved for 25.10 and beyond, and it's not worth fixing if it is only affecting old stable, now.

I've not had any similar issue with 25.12 SDK. The Makefile machinery has changed substantially, so any 24.10 fix would not be applicable to current stable, or master.

:shrug:

mcprat:
i'm asking you to modify the broken sdk to see if it works

glad to hear the newer one is ok but something like this can still pop up in the future because it's a portability problem

danielfdickinson:
@mcprat I appreciate your efforts, but the 24.10 SDK looks nothing like the code here. I can't patch the 24.10 SDK based on this because the Makefiles are substantially different. It would take creating a patch on the `openwrt-24.10` branch instead of `main/master` to have something testable on the 24.10 SDK.

For instance this:

```
$(toolchain/stamp-compile): $(tools/stamp-compile) $(if $(CONFIG_BUILDBOT),toolchain_rebuild_check)
```

does not exist in any form in any Makefile in the 24.10 SDK that I can find. (After adding the feeds I did a grep -r `toolchain_rebuild_check` with no results. I also looked for `stamp-compile` and got:

```
./Makefile:$(package/stamp-compile): $(BUILD_DIR)/.prepared
./Makefile:world: prepare $(package/stamp-compile) FORCE
```

mcprat:
for the 24.10 SDK, the openwrt core files are in `feeds/base` not `feeds/base_root`

otherwise directions are the same

sorry for that confusion

mcprat:
trust me it's there, start with a new unpacking of the 24.10 SDK, update the feeds, read over everything I said here and try it

danielfdickinson:
This is the contents of `feeds/base/Makefile`: Perhaps I am having 'a day', but I do not see where to apply your suggested patches in this code. (This code comes from the `openwrt-24.10` branch).

```
#
# Copyright (C) 2006-2010 OpenWrt.org
#
# This is free software, licensed under the GNU General Public License v2.
# See /LICENSE for more information.
#

curdir:=package

include $(INCLUDE_DIR)/feeds.mk
include $(INCLUDE_DIR)/rootfs.mk

-include $(TMP_DIR)/.packagedeps
package-y += kernel/linux
$(curdir)/autoremove:=1
$(curdir)/builddirs:=$(sort $(package-) $(package-y) $(package-m))
$(curdir)/builddirs-default:=. $(sort $(package-y) $(package-m))
$(curdir)/builddirs-prereq:=. $(sort $(prereq-y) $(prereq-m))
ifdef CHECK_ALL
$(curdir)/builddirs-check:=$($(curdir)/builddirs)
$(curdir)/builddirs-download:=$($(curdir)/builddirs)
endif
ifneq ($(IGNORE_ERRORS),)
  package-y-filter := $(package-y)
  package-m-filter := $(filter-out $(package-y),$(package-m))
  package-n-filter := $(filter-out $(package-y) $(package-m),$(package-))
  package-ignore-errors := $(filter n m y,$(IGNORE_ERRORS))
  package-ignore-errors := $(if $(package-ignore-errors),$(package-ignore-errors),n m)
  package-ignore-subdirs := $(sort $(foreach m,$(package-ignore-errors),$(package-$(m)-filter)))
  $(curdir)/builddirs-ignore-download := $(package-ignore-subdirs)
  $(curdir)/builddirs-ignore-compile := $(package-ignore-subdirs)
  $(curdir)/builddirs-ignore-host-download := $(package-ignore-subdirs)
  $(curdir)/builddirs-ignore-host-compile := $(package-ignore-subdirs)
endif

PACKAGE_INSTALL_FILES:= \
	$(foreach pkg,$(sort $(package-y)), \
		$(foreach variant, \
			$(if $(strip $(package/$(pkg)/variants)), \
				$(package/$(pkg)/variants), \
				$(if $(package/$(pkg)/default-variant), \
					$(package/$(pkg)/default-variant), \
					default \
				) \
			), \
			$(PKG_INFO_DIR)/$(lastword $(subst /,$(space),$(pkg))).$(variant).install \
		) \
	)

$(curdir)/cleanup: $(TMP_DIR)/.build
	rm -rf $(STAGING_DIR_ROOT)

$(curdir)/merge:
	rm -rf $(PACKAGE_DIR_ALL)
	mkdir -p $(PACKAGE_DIR_ALL)
ifneq ($(CONFIG_USE_APK),)
	-$(foreach pdir,$(PACKAGE_SUBDIRS),$(if $(wildcard $(pdir)/*.apk),ln -s $(pdir)/*.apk $(PACKAGE_DIR_ALL);))
else
	-$(foreach pdir,$(PACKAGE_SUBDIRS),$(if $(wildcard $(pdir)/*.ipk),ln -s $(pdir)/*.ipk $(PACKAGE_DIR_ALL);))
endif

$(BUILD_KEY_APK_SEC):
	$(STAGING_DIR_HOST)/bin/openssl ecparam -name prime256v1 -genkey -noout -out $(BUILD_KEY_APK_SEC)

$(BUILD_KEY_APK_PUB): $(BUILD_KEY_APK_SEC)
	$(STAGING_DIR_HOST)/bin/openssl ec -in $(BUILD_KEY_APK_SEC) -pubout > $(BUILD_KEY_APK_PUB)

$(curdir)/merge-index: $(curdir)/merge
ifneq ($(CONFIG_USE_APK),)
	(cd $(PACKAGE_DIR_ALL) && $(STAGING_DIR_HOST)/bin/apk mkndx \
			--root $(TOPDIR) \
			--keys-dir $(TOPDIR) \
			--sign $(BUILD_KEY_APK_SEC) \
			--output packages.adb \
			*.apk; \
	)
else
	(cd $(PACKAGE_DIR_ALL) && $(SCRIPT_DIR)/ipkg-make-index.sh . 2>&1 > Packages; )
endif

ifndef SDK
  $(curdir)//compile = $(STAGING_DIR)/.prepared $(BIN_DIR)
ifneq ($(CONFIG_USE_APK),)
  $(curdir)//compile += $(curdir)/system/apk/host/compile $(BUILD_KEY_APK_SEC) $(BUILD_KEY_APK_PUB)
else
  $(curdir)/compile: $(curdir)/system/opkg/host/compile
endif
else
ifneq ($(CONFIG_USE_APK),)
  $(curdir)//compile += $(BUILD_KEY_APK_SEC) $(BUILD_KEY_APK_PUB)
endif
endif

$(curdir)/install: $(TMP_DIR)/.build $(curdir)/merge $(curdir)/merge-index
	- find $(STAGING_DIR_ROOT) -type d | $(XARGS) chmod 0755
	rm -rf $(TARGET_DIR) $(TARGET_DIR_ORIG)
	mkdir -p $(TARGET_DIR)/tmp
ifneq ($(CONFIG_USE_APK),)
	$(file >$(TMP_DIR)/apk_install_list,\
	    $(foreach pkg,$(shell cat $(PACKAGE_INSTALL_FILES) 2>/dev/null),$(pkg)$(call GetABISuffix,$(pkg))))
	$(call apk,$(TARGET_DIR)) add --initdb --no-scripts --arch $(ARCH_PACKAGES) \
		--repositories-file /dev/zero --repository file://$(PACKAGE_DIR_ALL)/packages.adb \
		$$(cat $(TMP_DIR)/apk_install_list)
else
	$(file >$(TMP_DIR)/opkg_install_list,\
	  $(call opkg_package_files,\
	    $(foreach pkg,$(shell cat $(PACKAGE_INSTALL_FILES) 2>/dev/null),$(pkg)$(call GetABISuffix,$(pkg)))))
	$(call opkg,$(TARGET_DIR)) install $$(cat $(TMP_DIR)/opkg_install_list)
	@for file in $(PACKAGE_INSTALL_FILES); do \
		[ -s $$file.flags ] || continue; \
		for flag in `cat $$file.flags`; do \
			$(call opkg,$(TARGET_DIR)) flag $$flag `cat $$file`; \
		done; \
	done || true
endif

	$(CP) $(TARGET_DIR) $(TARGET_DIR_ORIG)

	$(call prepare_rootfs,$(TARGET_DIR),$(TOPDIR)/files)

$(curdir)/index: FORCE
	@echo Generating package index...
ifneq ($(CONFIG_USE_APK),)
	@for d in $(PACKAGE_SUBDIRS); do \
		mkdir -p $$d; \
		cd $$d || continue; \
		ls *.apk >/dev/null 2>&1 || continue; \
		$(STAGING_DIR_HOST)/bin/apk mkndx \
			--root $(TOPDIR) \
			--keys-dir $(TOPDIR) \
			--sign $(BUILD_KEY_APK_SEC) \
			--output packages.adb \
			$$(ls *.apk | grep -v 'kernel\|libc'); \
	done
else
	@for d in $(PACKAGE_SUBDIRS); do ( \
		mkdir -p $$d; \
		cd $$d || continue; \
		$(SCRIPT_DIR)/ipkg-make-index.sh . 2>&1 > Packages.manifest; \
		grep -vE '^(Maintainer|LicenseFiles|Source|SourceName|Require|SourceDateEpoch)' Packages.manifest > Packages; \
		case "$$(((64 + $$(stat -L -c%s Packages)) % 128))" in 110|111) \
			$(call ERROR_MESSAGE,WARNING: Applying padding in $$d/Packages to workaround usign SHA-512 bug!); \
			{ echo ""; echo ""; } >> Packages;; \
		esac; \
		$(SCRIPT_DIR)/make-index-json.py -f opkg -a "$(ARCH_PACKAGES)" Packages > index.json; \
		gzip -9nc Packages > Packages.gz; \
	); done
ifdef CONFIG_SIGNED_PACKAGES
	@echo Signing package index...
	@for d in $(PACKAGE_SUBDIRS); do ( \
		[ -d $$d ] && \
			cd $$d || continue; \
		$(STAGING_DIR_HOST)/bin/usign -S -m Packages -s $(BUILD_KEY); \
	); done
endif
ifdef CONFIG_JSON_CYCLONEDX_SBOM
	@echo Creating CycloneDX package SBOMs...
	@for d in $(PACKAGE_SUBDIRS); do ( \
		[ -d $$d ] && \
			cd $$d || continue; \
		$(SCRIPT_DIR)/package-metadata.pl pkgcyclonedxsbom Packages.manifest > Packages.bom.cdx.json || true; \
	); done
endif
endif

$(curdir)/flags-install:= -j1

$(eval $(call stampfile,$(curdir),package,prereq,.config))
$(eval $(call stampfile,$(curdir),package,cleanup,$(TMP_DIR)/.build))
$(eval $(call stampfile,$(curdir),package,compile,$(TMP_DIR)/.build))
$(eval $(call stampfile,$(curdir),package,install,$(TMP_DIR)/.build))
$(eval $(call stampfile,$(curdir),package,check,$(TMP_DIR)/.build))

$(eval $(call subdir,$(curdir)))
```

mcprat:
what you put is
`feeds/base/package/Makefile`

we're interested in
`feeds/base/Makefile`
`feeds/base/tools/Makefile`
`feeds/base/toolchain/Makefile`

danielfdickinson:
Ah! In the newer SDKs the feeds.conf uses the package dir as base. Apparently this is not the case for 24.10. E.g. for 25.12 SDK, from `feeds.conf.default`: `src-git --root=package base https://git.openwrt.org/openwrt/openwrt.git^8b4df0b66da3bbf8c73ddb9fd7881c140637411b`

Let's see if the issue even still exists when I fix that...

mcprat:
fyi I just tried 24.10.8 SDK, I didn't have to modify the feeds configuration... but who knows for which patch versions they backported those changes... 24.10.x

danielfdickinson:
I was manually 'borrowing' my `feeds.conf` from my automatic setup for 25.12 and beyond, so I wasn't using the `feeds.conf.default` from 24.10, and I did not notice the difference in the `feeds.conf.default` between versions until now.

danielfdickinson:
I have no issue that needs fixing once I fix that. No patch required.

danielfdickinson:
@mcprat Thank you for your patience, and sorry for the noise.

mcprat:
hmmm really?

it would have been nice to confirm with someone that this solves an issue other than the original problem that was described years ago...

at least I can confirm myself that builds start with the kernel instead of tools...

danielfdickinson:
@mcprat  There may still be a libc ABI issue to solve; it just wasn't the issue for #24273. I will do a couple of tests, as I have some time: build an 24.10 SDK with as old an OS as can build it, and then try using the SDK in a current and/or bleeding edge distro, and vice versa (build the SDK in bleeding edge and try using in as old a distro as should work).

mcprat:
@aparcar maybe at least you can verify and accept this little change

before: a newly unpacked SDK with feeds updated *might* try to rebuild tools,
at the very least establishes stampfiles for the tools

after: a newly unpacked SDK with feeds updated begins downloading the kernel immediately,
and ignores the stampfiles for both tools and toolchain

PR #24440, opened 2026-07-27
feature/snr-cpe-ax2 and bo-rt622-g31-g2

akosmakov:
Fix support SNR-CPE-AX2 (no bootloader update required)
Add support BO-RT622-G31-G2

### DIFF
--- a/package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic
+++ b/package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic
@@ -53,7 +53,6 @@ qihoo,360t7|\
 qihoo,360t7-ubi|\
 routerich,ax3000-ubootmod|\
 routerich,be7200|\
-snr,snr-cpe-ax2|\
 tplink,tl-xdr4288|\
 tplink,tl-xdr6086|\
 tplink,tl-xdr6088|\
@@ -154,6 +153,9 @@ smartrg,sdg-8733|\
 smartrg,sdg-8733a|\
 smartrg,sdg-8734)
 	ubootenv_add_mmc "u-boot-env" "mmcblk0" "0x0" "0x8000" "0x8000"
+	;;
+snr,snr-cpe-ax2)
+	ubootenv_add_uci_config "/dev/mtd1" "0x0" "0x20000" "0x20000" "1"
 	;;
 teltonika,rutc50|\
 alwaylink,m01k43)
--- a/package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic
+++ b/package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic
@@ -154,6 +154,7 @@ smartrg,sdg-8733a|\
 smartrg,sdg-8734)
 	ubootenv_add_mmc "u-boot-env" "mmcblk0" "0x0" "0x8000" "0x8000"
 	;;
+bo,rt622-g31-g2|\
 snr,snr-cpe-ax2)
 	ubootenv_add_uci_config "/dev/mtd1" "0x0" "0x20000" "0x20000" "1"
 	;;
--- a/scripts/mkits-snr.sh
+++ b/scripts/mkits-snr.sh
@@ -1,0 +1,124 @@
+#!/usr/bin/env bash
+#
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# Based on OpenWrt scripts/mkits.sh (Author: Peter Tyser <ptyser@xes-inc.com>).
+# Copyright (C) 2026 LLC NAGTECH
+#
+# Build a FIT image .its for the MediaTek CONFIG_MTK_DUAL_BOOT bootloader.
+# In addition to the usual kernel(+fdt) FIT, it embeds a top-level /rootfs node
+# holding the squashfs size and crc32/sha1 hashes. The OEM U-Boot verifies this
+# node at every boot (CONFIG_MTK_DUAL_BOOT_IMAGE_ROOTFS_VERIFY); without it the
+# loader rejects the image and falls back to the other boot slot.
+
+usage() {
+	echo "Usage: $(basename "$0") -A arch -C comp -a addr -e entry -v version" \
+	     "-k kernel [-D name -d dtb] -c config -o its_file [-R rootfs]"
+	exit 1
+}
+
+while getopts "A:a:C:c:D:d:e:k:o:v:R:" OPTION; do
+	case $OPTION in
+		A ) ARCH=$OPTARG;;
+		a ) LOAD_ADDR=$OPTARG;;
+		C ) COMPRESS=$OPTARG;;
+		c ) CONFIG=$OPTARG;;
+		D ) DEVICE=$OPTARG;;
+		d ) DTB=$OPTARG;;
+		e ) ENTRY_ADDR=$OPTARG;;
+		k ) KERNEL=$OPTARG;;
+		o ) OUTPUT=$OPTARG;;
+		v ) VERSION=$OPTARG;;
+		R ) ROOTFS_FILE=$OPTARG;;
+		* ) usage;;
+	esac
+done
+
+if [ -z "${ARCH}" ] || [ -z "${COMPRESS}" ] || [ -z "${LOAD_ADDR}" ] || \
+   [ -z "${ENTRY_ADDR}" ] || [ -z "${VERSION}" ] || [ -z "${KERNEL}" ] || \
+   [ -z "${OUTPUT}" ] || [ -z "${CONFIG}" ]; then
+	usage
+fi
+
+ARCH_UPPER=$(echo "${ARCH}" | tr '[:lower:]' '[:upper:]')
+
+# Optional FDT image + configuration reference
+if [ -n "${DTB}" ]; then
+	FDT="
+			fdt@1 {
+				description = \"${ARCH_UPPER} OpenWrt ${DEVICE} device tree blob\";
+				data = /incbin/(\"${DTB}\");
+				type = \"flat_dt\";
+				arch = \"${ARCH}\";
+				compression = \"none\";
+				hash@1 {
+					algo = \"crc32\";
+				};
+				hash@2 {
+					algo = \"sha1\";
+				};
+			};
+"
+	FDT_PROP="fdt = \"fdt@1\";"
+fi
+
+# Optional top-level /rootfs node with squashfs size + hashes (MediaTek dual-boot)
+if [ -f "${ROOTFS_FILE}" ]; then
+	ROOTFS_SIZE=$(stat -c %s "${ROOTFS_FILE}")
+	ROOTFS_CRC32=$(python3 -c "import sys,zlib;print('%08x'%(zlib.crc32(open(sys.argv[1],'rb').read())&0xffffffff))" "${ROOTFS_FILE}")
+	ROOTFS_SHA1=$(sha1sum "${ROOTFS_FILE}" | awk '{print "<0x"substr($0,1,8)" 0x"substr($0,9,8)" 0x"substr($0,17,8)" 0x"substr($0,25,8)" 0x"substr($0,33,8)">"}')
+
+	ROOTFS="
+	rootfs {
+		size = <${ROOTFS_SIZE}>;
+
+		hash@1 {
+			value = <0x${ROOTFS_CRC32}>;
+			algo = \"crc32\";
+		};
+
+		hash@2 {
+			value = ${ROOTFS_SHA1};
+			algo = \"sha1\";
+		};
+	};
+"
+fi
+
+DATA="/dts-v1/;
+
+/ {
+	description = \"${ARCH_UPPER} OpenWrt FIT (Flattened Image Tree)\";
+	#address-cells = <1>;
+
+	images {
+		kernel@1 {
+			description = \"${ARCH_UPPER} OpenWrt Linux-${VERSION}\";
+			data = /incbin/(\"${KERNEL}\");
+			type = \"kernel\";
+			arch = \"${ARCH}\";
+			os = \"linux\";
+			compression = \"${COMPRESS}\";
+			load = <${LOAD_ADDR}>;
+			entry = <${ENTRY_ADDR}>;
+			hash@1 {
+				algo = \"crc32\";
+			};
+			hash@2 {
+				algo = \"sha1\";
+			};
+		};
+${FDT}
+	};
+${ROOTFS}
+	configurations {
+		default = \"${CONFIG}\";
+		${CONFIG} {
+			description = \"OpenWrt\";
+			kernel = \"kernel@1\";
+			${FDT_PROP}
+		};
+	};
+};"
+
+echo "$DATA" > "${OUTPUT}"
--- a/target/linux/mediatek/dts/mt7981b-bo-rt622-g31-g2.dts
+++ b/target/linux/mediatek/dts/mt7981b-bo-rt622-g31-g2.dts
@@ -1,0 +1,311 @@
+// SPDX-License-Identifier: GPL-2.0-only OR MIT
+/*
+ * Copyright (C) 2026 LLC NAGTECH
+ */
+
+/dts-v1/;
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+
+#include "mt7981b.dtsi"
+
+/ {
+	model = "BO-RT622-G31-G2";
+	compatible = "bo,rt622-g31-g2", "mediatek,mt7981";
+
+	aliases {
+		label-mac-device = &gmac0;
+
+		led-boot = &led_sys;
+		led-failsafe = &led_sys;
+		led-running = &led_sys;
+		led-upgrade = &led_sys;
+
+		serial0 = &uart0;
+	};
+
+	chosen: chosen {
+		stdout-path = "serial0:115200n8";
+	};
+
+	memory@40000000 {
+		device_type = "memory";
+		reg = <0 0x40000000 0 0x10000000>;
+	};
+
+	gpio-keys {
+		compatible = "gpio-keys";
+
+		button-0 {
+			label = "mesh";
+			gpios = <&pio 0 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RFKILL>;
+			debounce-interval = <60>;
+		};
+
+		button-1 {
+			label = "reset";
+			gpios = <&pio 1 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+			debounce-interval = <60>;
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		led_sys: led-0 {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_STATUS;
+			gpios = <&pio 8 GPIO_ACTIVE_LOW>;
+			default-state = "on";
+		};
+
+		led-1 {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WAN;
+			gpios = <&pio 4 GPIO_ACTIVE_LOW>;
+		};
+
+		led-2 {
+			color = <LED_COLOR_ID_GREEN>;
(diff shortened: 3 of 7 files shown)


target/linux/mediatek/image/filogic.mk
```diff
@@ -3069,23 +3103,21 @@ define Device/snr_snr-cpe-ax2
   DEVICE_MODEL := SNR-CPE-AX2
   DEVICE_DTS := mt7981b-snr-snr-cpe-ax2
   DEVICE_DTS_DIR := ../dts
```
openwrt-ai, line 3105: This is a breaking change for anyone already running the in-tree SNR-CPE-AX2 support, but there is no compat-version bump. The current device boots OpenWrt U-Boot (`ARTIFACTS := preloader.bin bl31-uboot.fip`), keeps its env in UBI (`UBOOTENV_IN_UBI := 1`) and boots a `fit` UBI volume; after this patch the ubi partition shrinks from `0x7a80000` to `0x7280000`, the `fit`/`ubootenv`/`ubootenv2` volumes are gone and the image format changes from `sysupgrade.itb` to a sysupgrade tar. An existing installation that runs `sysupgrade` onto the new build will be left with a UBI its bootloader cannot boot.

`DEVICE_COMPAT_VERSION`/`DEVICE_COMPAT_MESSAGE` is what the target already uses for exactly this situation (see `mercusys,mr90x-v1-ubi` and `zyxel,ex5601-t0` at filogic.mk:2971 and :3668).

```suggestion
  DEVICE_DTS_DIR := ../dts
  DEVICE_COMPAT_VERSION := 1.1
  DEVICE_COMPAT_MESSAGE := Partition layout and bootloader changed, the device now runs the OEM \
	bootloader. Reinstall via U-Boot recovery, sysupgrade from an older OpenWrt build will not boot.
```


---

target/linux/mediatek/image/filogic.mk
```diff
@@ -3069,23 +3103,21 @@ define Device/snr_snr-cpe-ax2
...
+  # "snr-cpe-ax2" matches the OEM U-Boot DTS model (lowercased) checked by
+  # verify_model_fit during the recovery/upgrade path.
+  SUPPORTED_DEVICES += snr-cpe-ax2
```
openwrt-ai, line 3108: The comment attributes `SUPPORTED_DEVICES` to the bootloader, but `SUPPORTED_DEVICES` only ends up in the OpenWrt sysupgrade metadata (`append-metadata` → `supported_devices`), which is consumed by `platform_check_image`/`nand_do_platform_check` on a running OpenWrt — U-Boot never sees it. Nothing named `verify_model_fit` exists in this tree or in u-boot upstream, so a reader can't check the claim either.

If the intent is "let a sysupgrade image built for the OEM-layout board be accepted", say that; if U-Boot really does match a string, that check is against the FIT/DTS `model`, not this variable. Same comment on the new `bo_rt622-g31-g2` block at lines 873-875.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -3069,23 +3103,21 @@ define Device/snr_snr-cpe-ax2
...
-  IMAGE/sysupgrade.itb := append-kernel | \
-	fit gzip $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb external-static-with-rootfs | \
-	append-metadata
-  DEVICE_PACKAGES := kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware
-  ARTIFACTS := preloader.bin bl31-uboot.fip
-  ARTIFACT/preloader.bin := mt7981-bl2 spim-nand-ddr3
-  ARTIFACT/bl31-uboot.fip := mt7981-bl31-uboot snr_snr-cpe-ax2
+  IMAGES += factory.bin
+  IMAGE/factory.bin := append-ubi | check-size $$$$(IMAGE_SIZE)
+  IMAGE/sysupgrade.bin := sysupgrade-tar | append-metadata
```
openwrt-ai, line 3120: Dropping `ARTIFACTS := preloader.bin bl31-uboot.fip` / `ARTIFACT/bl31-uboot.fip := mt7981-bl31-uboot snr_snr-cpe-ax2` orphans the OpenWrt U-Boot port for this board: [`U-Boot/mt7981_snr_snr-cpe-ax2` in uboot-mediatek/Makefile:716-720](https://github.com/openwrt/openwrt/blob/fd1c10ec303d89ce307ce2ec9b25761e6287da8c/package/boot/uboot-mediatek/Makefile#L716-L720) still carries `BUILD_DEVICES:=snr_snr-cpe-ax2`, so the FIP keeps getting built for every filogic build and is now consumed by nothing, and [`patches/463-add-snr-snr-cpe-ax2.patch`](https://github.com/openwrt/openwrt/blob/fd1c10ec303d89ce307ce2ec9b25761e6287da8c/package/boot/uboot-mediatek/patches/463-add-snr-snr-cpe-ax2.patch) plus its `mt7981_snr_snr-cpe-ax2` entry in the config list stay behind as dead code.

Either drop the U-Boot port in the same commit, or keep the artifacts so the OpenWrt-U-Boot install path remains available.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -146,6 +146,17 @@ define Build/tenda-mkdualimageheader
...
+define Build/snr-fit
+	$(TOPDIR)/scripts/mkits-snr.sh \
+		-D $(DEVICE_NAME) -o $@.its -k $@ \
+		$(if $(word 2,$(1)),-d $(word 2,$(1))) -C $(word 1,$(1)) \
+		-a $(KERNEL_LOADADDR) -e $(if $(KERNEL_ENTRY),$(KERNEL_ENTRY),$(KERNEL_LOADADDR)) \
+		-c config@1 -A $(LINUX_KARCH) -v $(LINUX_VERSION) \
+		$(if $(CONFIG_TARGET_ROOTFS_SQUASHFS),-R $(ROOTFS/squashfs/$(DEVICE_NAME)))
+	PATH=$(LINUX_DIR)/scripts/dtc:$(PATH) mkimage -f $@.its $@.new
+	@mv $@.new $@
+endef
```
openwrt-ai, line 158: This duplicates infrastructure that already exists in this very file. [`Build/fit-with-netgear-top-level-rootfs-node` (filogic.mk:11-18)](https://github.com/openwrt/openwrt/blob/fd1c10ec303d89ce307ce2ec9b25761e6287da8c/target/linux/mediatek/image/filogic.mk#L11-L18) solves the identical problem — a bootloader that wants a top-level `/rootfs` node with the squashfs size and hash — by calling the standard `Build/fit-its`, splicing in a node produced by [`scripts/gen_netgear_rootfs_node.sh`](https://github.com/openwrt/openwrt/blob/fd1c10ec303d89ce307ce2ec9b25761e6287da8c/scripts/gen_netgear_rootfs_node.sh), and then running `Build/fit-image`. It is used the same way, from the `KERNEL` pipeline, at [filogic.mk:2669-2670](https://github.com/openwrt/openwrt/blob/fd1c10ec303d89ce307ce2ec9b25761e6287da8c/target/linux/mediatek/image/filogic.mk#L2669-L2670).

Forking all 124 lines of `scripts/mkits.sh` into `scripts/mkits-snr.sh` to add the same node means the FIT generation for these two boards no longer tracks fixes to `mkits.sh`/`fit-its`. The only delta this device needs over the Netgear helper is the extra `crc32` hash — please extend `gen_netgear_rootfs_node.sh` (e.g. an optional crc32 flag) and reuse `Build/fit-with-netgear-top-level-rootfs-node` (renamed to something generic), rather than adding a parallel script.

---

scripts/mkits-snr.sh
```diff
@@ -0,0 +1,124 @@
...
+				hash@2 {
+					algo = \"sha1\";
+				};
+			};
+"
+	FDT_PROP="fdt = \"fdt@1\";"
+fi
+
+# Optional top-level /rootfs node with squashfs size + hashes (MediaTek dual-boot)
+if [ -f "${ROOTFS_FILE}" ]; then
```
openwrt-ai, line 66: If `-R` is passed but the file is missing, the `/rootfs` node is silently omitted and the build still succeeds. Per the header comment, the OEM U-Boot then rejects the image and falls back to the other slot — i.e. a silent build bug turns into a device that won't boot the produced firmware. `-R` is only ever passed by `Build/snr-fit` when `CONFIG_TARGET_ROOTFS_SQUASHFS` is set, so a missing file there is always an error, not an opt-out.

```suggestion
if [ -n "${ROOTFS_FILE}" ]; then
	if [ ! -f "${ROOTFS_FILE}" ]; then
		echo "$(basename "$0"): rootfs '${ROOTFS_FILE}' not found" >&2
		exit 1
	fi
```


---

scripts/mkits-snr.sh
```diff
@@ -0,0 +1,124 @@
...
+	ROOTFS_SIZE=$(stat -c %s "${ROOTFS_FILE}")
+	ROOTFS_CRC32=$(python3 -c "import sys,zlib;print('%08x'%(zlib.crc32(open(sys.argv[1],'rb').read())&0xffffffff))" "${ROOTFS_FILE}")
```
openwrt-ai, line 68: nit: this is the first `scripts/*.sh` in the tree to shell out to `python3`, and it does `open(...).read()` on the whole rootfs — for `IMAGE_SIZE := 117248k` that is a ~114 MiB allocation per image build. The equivalent in-tree helper, [`scripts/gen_netgear_rootfs_node.sh`](https://github.com/openwrt/openwrt/blob/fd1c10ec303d89ce307ce2ec9b25761e6287da8c/scripts/gen_netgear_rootfs_node.sh), gets by with `stat`/`sha1sum` only. If crc32 really is required by this bootloader, at minimum stream the file in chunks instead of slurping it.

---

package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic
```diff
@@ -155,6 +154,10 @@ smartrg,sdg-8733a|\
 smartrg,sdg-8734)
 	ubootenv_add_mmc "u-boot-env" "mmcblk0" "0x0" "0x8000" "0x8000"
 	;;
+bo,rt622-g31-g2|\
+snr,snr-cpe-ax2)
+	ubootenv_add_uci_config "/dev/mtd1" "0x0" "0x20000" "0x20000" "1"
+	;;
```
openwrt-ai, line 160: nit: both DTS files label this partition `u-boot-env`, so the index-free helper works and is what the rest of this file prefers (`ubootenv_add_mtd "u-boot-env" ...` for `teltonika,rutc50`, `zyxel,ex5601-t0`, and `ubootenv_add_nor_default` at the top). Hardcoding `/dev/mtd1` breaks if the partition order ever changes.

Also, the case entries in this file are ordered alphabetically by board — inserting `bo,...` between `smartrg,...` and `teltonika,...` breaks that; it belongs before `creatlentem,...` near the top.

```suggestion
bo,rt622-g31-g2|\
snr,snr-cpe-ax2)
	ubootenv_add_mtd "u-boot-env" "0x0" "0x20000" "0x20000" "1"
	;;
```


---

target/linux/mediatek/dts/mt7981b-bo-rt622-g31-g2.dts
```diff
@@ -0,0 +1,311 @@
...
+
+/dts-v1/;
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+
+#include "mt7981b.dtsi"
+
+/ {
+	model = "BO-RT622-G31-G2";
```
openwrt-ai, line 14: nit: every other board in `target/linux/mediatek/dts/mt7981b-*.dts` writes `model` as `"<Vendor> <Model>"` — `"ABT ASR3000"`, `"Cetron CT3003"`, `"BUFFALO WSR-3000AX4P"` — matching `DEVICE_VENDOR`/`DEVICE_MODEL`. With `DEVICE_VENDOR := BO` and `DEVICE_MODEL := RT622-G31-G2` that would be:

```suggestion
	model = "BO RT622-G31-G2";
```

The same change is applied in reverse to the sister board in this series: `mt7981b-snr-snr-cpe-ax2.dts:11` goes from `"SNR SNR-CPE-AX2"` to `"SNR-CPE-AX2"`, which also leaves it disagreeing with `NAME:=SNR SNR-CPE-AX2` in `package/boot/uboot-mediatek/Makefile:717`. Neither commit message explains why the `model` string had to change — if a bootloader string match forces it, please say so in the commit.

---

target/linux/mediatek/dts/mt7981b-bo-rt622-g31-g2.dts
```diff
@@ -0,0 +1,311 @@
...
+
+		led-boot = &led_sys;
+		led-failsafe = &led_sys;
+		led-running = &led_sys;
+		led-upgrade = &led_sys;
+
+		serial0 = &uart0;
+	};
+
+	chosen: chosen {
```
openwrt-ai, line 28: nit: the `chosen` label is never referenced. No other `mt7981b-*.dts` labels this node.

```suggestion
	chosen {
```

Same for `partitions: partitions {` at line 184 here and at `mt7981b-snr-snr-cpe-ax2.dts:205` — that label is unused too.

---

target/linux/mediatek/dts/mt7981b-bo-rt622-g31-g2.dts
```diff
@@ -0,0 +1,311 @@
...
+			};
+		};
+	};
+};
+
+&spi_nand {
+	mediatek,nmbm;
+	mediatek,bmt-max-ratio = <1>;
+	mediatek,bmt-max-reserved-blocks = <64>;
+};
```
openwrt-ai, line 258: These three properties belong inside the `flash@0` node defined ~80 lines above in this same file; there is no reason to introduce a `spi_nand` label and re-open the node via a phandle reference in the file that declares it. Every other nmbm board puts them inline — e.g. `mt7981b-cetron-ct3003.dts:109`, `mt7981b-bazis-ax3000wm.dts:173`, `mt7981b-alwaylink-m01k43.dts:260`.

Same applies to `mt7981b-snr-snr-cpe-ax2.dts:275-279`. (The `spi_nand: flash@0` rename itself is a good fix — just keep the label only if something outside the file needs it.)

---

target/linux/mediatek/dts/mt7981b-bo-rt622-g31-g2.dts
```diff
@@ -0,0 +1,311 @@
...
+		spi-cal-addrlen = <5>;
+		spi-cal-addr = /bits/ 32 <0x0 0x0 0x0 0x0 0x0>;
+
+		/*
+		 * Stock partition layout, must match the vendor bootloader's
+		 * MTDPARTS_DEFAULT so the device boots without reflashing BL2/FIP.
+		 * See uboot-arm-master/u-boot/configs/BO-RT622-G31-g2_defconfig:
+		 * CONFIG_MTDPARTS_DEFAULT="nmbm0:1024k(bl2),512k(u-boot-env),
+		 *   2048k(factory),2048k(fip),-(ubi)"
+		 */
```
openwrt-ai, line 183: nit: `uboot-arm-master/u-boot/configs/...` is a path in a private vendor tree that no reader of this repo can resolve — the SNR file's version of the same comment ("See the OEM U-Boot defconfig") is the better wording. Also `BO-RT622-G31-g2_defconfig` lowercases the final `G2`, inconsistent with the `model`, the compatible and the device name used everywhere else in this series.

---

target/linux/mediatek/dts/mt7981b-bo-rt622-g31-g2.dts
```diff
@@ -0,0 +1,311 @@
...
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_STATUS;
+			gpios = <&pio 8 GPIO_ACTIVE_LOW>;
+			default-state = "on";
+		};
+
+		led-1 {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WAN;
+			gpios = <&pio 4 GPIO_ACTIVE_LOW>;
```
openwrt-ai, line 68: nit: this WAN LED gets no netdev trigger. The sister board in this same series is wired up in [`base-files/etc/board.d/01_leds:32-34`](https://github.com/openwrt/openwrt/blob/fd1c10ec303d89ce307ce2ec9b25761e6287da8c/target/linux/mediatek/filogic/base-files/etc/board.d/01_leds#L32-L34) (`ucidef_set_led_netdev "wan" "wan" "blue:wan" "wan" "link tx rx"`), and this PR adds no `bo,rt622-g31-g2` entry there, so `green:wan` stays dark. Is that intentional, or is the `01_leds` hunk missing?

The `LED_FUNCTION_INDICATOR` LED on line 70 has no consumer either — if you know what the vendor drives it with, a comment or a trigger would help.

---

openwrt-ai (commented):
## Commit checks

- `1e8b7c3` "mediatek: filogic: fix support for SNR-CPE-AX2" — the message body contains only generic flashing steps and never says what was broken or what the patch actually does. The diff is a full rework: it moves the device off OpenWrt U-Boot onto the stock bootloader (drops `ARTIFACTS`/`UBOOTENV_IN_UBI`, removes the `fit`/`ubootenv`/`ubootenv2` UBI volumes and the `rootdisk`/`bootargs-append` chosen properties), shrinks the ubi partition from `0x7a80000` to `0x7280000`, enables `mediatek,nmbm`, swaps the image format from `sysupgrade.itb` to a sysupgrade tar plus `factory.bin`, changes the DTS `model` string, and reassigns the switch port labels (port 0 wan → lan1, ports 1-4 shifted). None of that is described. Please state the problem being fixed and spell out that this is a flash-layout/bootloader change, since it decides whether existing users can sysupgrade.
- `1e8b7c3` — subject scope `mediatek: filogic:` does not cover the files touched: the commit also adds a new top-level `scripts/mkits-snr.sh` and edits `package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic`. The build-script addition in particular is reusable infrastructure and reads like a separate patch.
- `fd1c10e` "mediatek: filogic: add support for BO-RT622-G31-G2" — the hardware specification says `Ethernet: MT7531: 3x 1GbE LAN + 1x 1GbE WAN`, but the DTS defines `lan1`–`lan4` plus `wan` and the inline comment states `Vendor portmap "llllw": ports 0-3 LAN, port 4 WAN`, i.e. 4x LAN. One of the two is wrong.
- `fd1c10e` — for a new-device commit the message is missing the MAC address layout section. The DTS derives five addresses (`macaddr_factory_e000` for gmac0/label, `macaddr_factory_e006` for wan, `macaddr_factory_4` for band0, `macaddr_factory_100004` for band1); please document one line per interface with the factory offset it comes from, using `xx:xx:xx` for the unit-specific bytes. The flash instructions would also benefit from a revert-to-stock note and the exact recovery-page/file name, so a stranger can flash the device from the commit message alone.


---

csharper2005:
You are removing support for the OpenWrt U-Boot. It is superior to the stock version because it is open-source, provides more free space for installing packages and offers a straightforward recovery mechanism. What if the user no longer has backups of the stock bootloader? Why should they have to deal with this inconvenience?

Please don't touch existing OpenWrt U-Boot. This is not a fix!

kvasilev89:
> You are removing support for the OpenWrt U-Boot. It is superior to the stock version because it is open-source, provides more free space for installing packages and offers a straightforward recovery mechanism. What if the user no longer has backups of the stock bootloader? Why should they have to deal with this inconvenience?
> 
> Please don't touch existing OpenWrt U-Boot. This is not a fix!


Are you suggesting creating a new model but with a different bootloader?

That can be done either way, but users who want to install OpenWrt on the AX2 router face a very serious challenge — updating the bootloader. If something goes wrong during that process, they are highly likely to end up with a bricked router. With what the author proposed, users have no such risks — they can easily use the vendor's standard recovery tools.

There are two options here: either leave this model as it is and create a new one with a suffix like "-original" or something similar, or modify the existing one.

Or did I misunderstand you?

VitaliySochniy:
> Are you suggesting creating a new model but with a different bootloader?

To ensure that sysupgrade continues to work for existing users, you should add `-stock` to the name of the new profile/image. You can change the human-readable name in LuCI, but let’s leave the `compatible` string as it is.

PR #24305, opened 2026-07-19
kernel: rtl8367b: support post-init register writes

3mg:
This adds an optional `realtek,init-regs` property to the legacy RTL8367B
swconfig driver. Boards may use it for a small sequence of register/value
writes after the chip-specific initialization table and before external
interface setup, avoiding board-specific register sequences in the shared
driver.

The property is a no-op when absent, so existing RTL8367B users retain their
current initialization path. Present properties are rejected when empty,
odd-sized, larger than 32 register/value pairs, or when either member of a
pair exceeds the switch's 16-bit register/value width. Register write errors
are propagated.

The commit also documents the existing legacy RTL8367B DT interface and the
new property in a binding schema.

This implements the generic approach suggested during review of #24249. The
device-specific DTS conversion remains intentionally outside this PR.

Co-authored-by: Mieczyslaw Nalewaj <namiltd@yahoo.com>

Run-tested: not yet (draft)

Static validation:

- `scripts/checkpatch.pl --strict`: 0 errors, warnings, or checks
- `dt-doc-validate realtek,rtl8367b.yaml`: pass
- `git diff --check`: pass

PR #24249, opened 2026-07-16
ramips: add RT6855A and Zyxel Keenetic Giga II support

3mg:
This series adds a new ramips/rt6855a subtarget and support for the
Zyxel Keenetic Giga II.

The series is split into eight reviewable commits:

1. Add the RT6855A/RT6856 platform, interrupt, PCIe, watchdog and SPI
   support, reusing the upstream EcoNet timer and Airoha GPIO drivers.
2. Allow frame engines without reset-controller bindings.
3. Fix runtime MAC programming in the generic frame-engine fallback used
   by RT2880 and RT3883 and required by RT6855A.
4. Add the RT6855A frame-engine register map, MDIO operations, DMA
   parameters and forwarding setup.
5. Guard the RT6855A/RT6856 frame-engine DMA lifecycle during initial
   setup, interface restart, timeout recovery and final teardown.
6. Add RT5392 transmit-spatial-stream and direct LED-control board
   properties together with their binding.
7. Fix the `openwrt,uimage` partition binding used by the device DTS.
8. Add the Keenetic Giga II DTS, image recipes, port mapping and
   sysupgrade support.

Commits 2, 3 and 7 are generic changes, but all are direct prerequisites
of this device series. RT6855A has no frame-engine reset-controller binding,
so the existing probe path must accept an optional reset. The shared MAC
fallback must pass `dev->dev_addr`, rather than the original `struct
sockaddr`, to the hardware helper; this fixes existing RT2880/RT3883 users
and allows RT6855A to reuse the generic path. The Giga II firmware partition
uses the standard `openwrt,uimage` binding, whose schema must allow the
partition properties used by this layout.

Commit 5 changes the shared frame-engine driver, but the new stop-and-wait
behaviour is enabled only by the RT6855A/RT6856 SoC data. It prevents DMA
rings from being released or reprogrammed until the PDMA enable and busy
bits have cleared. If DMA cannot be stopped, the rings remain allocated
rather than risking DMA into reclaimed memory.

Keeping these prerequisites in the same series makes every later commit
buildable and schema-valid without depending on unmerged companion pull
requests.

Hardware:

- SoC: Ralink RT6856F, MIPS 34Kc at 700 MHz
- RAM: 256 MiB DDR2, Winbond W972GG6JB-25
- Flash: 16 MiB Macronix MX25L12805D SPI NOR
- Switch: Realtek RTL8367RB, five Gigabit Ethernet ports
[...]

PR #24441, opened 2026-07-27
ramips: add ZyXEL Keenetic Ultra support

3mg:
This series adds support for the ZyXEL Keenetic Ultra on the new
ramips/rt6855a subtarget.

It is based on the still-unmerged RT6855A and Keenetic Giga II series in
#24249 and the generic RTL8367B post-init support in #24305. The first nine
commits are the current integration of those dependencies.

The Ultra follow-up is split into five reviewable commits:

1. Re-arm rt2800 RX descriptors with the complete DMA buffer length.
2. Add reversible control for the RTL8367-owned front-panel LEDs.
3. Select full `wpad-mbedtls` for push-button WPS.
4. Erase the complete combined firmware partition during sysupgrade.
5. Add the Keenetic Ultra DTS, image recipes and board configuration.

Hardware:

- SoC: Ralink RT6856, MIPS 34Kc at 700 MHz
- RAM: 256 MiB DDR2
- Flash: 16 MiB SPI NOR
- Switch: Realtek RTL8367RB, five Gigabit Ethernet ports
- Wi-Fi: Ralink RT5392 PCIe 2.4 GHz 802.11n
- Wi-Fi: Ralink RT3593 PCIe 5 GHz 802.11n
- USB: two USB 2.0 host ports
- Buttons: Reset, WPS and FN
- LEDs: power, Internet, USB1, USB2, Wi-Fi 2.4 GHz, Wi-Fi 5 GHz and
  switch port LEDs

The enclosure numbers Ethernet ports in reverse switch order.
RTL8367 port 4 is the Internet port marked `0`, ports 3 through 0 are
LAN1 through LAN4 marked `1` through `4`, and port 6 is the RGMII CPU port.

Initial installation:

1. Connect a 3.3 V serial adapter at 57600 8N1 and connect the router to a
   TFTP server through an Ethernet port.
2. Put the generated factory.bin image in the TFTP server root.
3. Power on the router and select U-Boot menu option 2.
4. Enter the router IP, TFTP server IP and factory.bin filename.
5. Wait for the SPI erase and write operations to finish, then reboot.

The initramfs image can alternatively be booted without writing flash by
selecting U-Boot menu option 1.

The generated factory.bin is padded through the complete OpenWrt firmware
area and carries the ZyXEL trailer for device ID 22037 (`15 56`). It is
accepted both by U-Boot option 2 and by the original ZyXEL web UI. The
trailer eraseblock is outside `rootfs_data`; the bootloader, vendor
configuration, RF calibration, storage and dump partitions remain
read-only.

MAC address layout:

- Ethernet, LAN and WAN use the address at RF-EEPROM offset `0x4`.
- RT5392 and RT3593 read their EEPROM data and base radio addresses from
  RF-EEPROM.
- Additional mac80211 virtual interfaces derive their addresses from the
  corresponding radio base address.

Tested on hardware:

- factory.bin installation through U-Boot option 2 and cold boot from SPI
- factory.bin installation through the original ZyXEL web UI
- TFTP initramfs boot through U-Boot option 1
- sysupgrade with complete firmware erase and repeated flash boot
- clean JFFS2 overlay creation and settings persistence
- Ethernet LAN and WAN, including enclosure port mapping
- WPA2 Wi-Fi AP and STA on both bands
- simultaneous routed AP+STA repeater on both bands
- WPS enrolment on 2.4 GHz using a Linux client
- router-side 5 GHz WPS PBC activation
- USB storage on both USB ports
- Reset, WPS and FN buttons
- all Linux, radio and RTL8367-owned front-panel LEDs
- hardware watchdog reset

End-to-end 5 GHz WPS enrolment is not claimed. The available 5 GHz clients
did not start an enrollee exchange. The Ultra entered PBC mode normally, and
the same button and hostapd path passed end-to-end on 2.4 GHz.

Static validation:

- complete rt6855a target build: pass
- `scripts/checkpatch.pl --strict`: 0 errors, warnings or checks for the six
  integration and Ultra commits; `TRAILING_WHITESPACE` is ignored only for
  blank context lines inside the added kernel patch file
- `git diff --check`: pass
- shell syntax checks for the upgrade and hotplug scripts: pass

This remains a draft while #24249 and #24305 are unmerged.


target/linux/ramips/image/rt6855a.mk
```diff
@@ -0,0 +1,46 @@
...
+	kmod-gpio-rt6856 kmod-leds-gpio kmod-gpio-button-hotplug \
+	kmod-usb2 kmod-usb-ohci kmod-usb-storage \
+	rt2800-pci-firmware swconfig
+endef
+TARGET_DEVICES += zyxel_keenetic-giga-ii
+
+define Device/zyxel_keenetic-ultra
+  SOC := rt6855a
+  IMAGE_SIZE := 14912k
+  BLOCKSIZE := 4k
```
openwrt-ai, line 33: `BLOCKSIZE` is the erase-block granularity used for JFFS2 padding, and 4k does not match the 64 KiB erase block of a 16 MiB SPI-NOR chip. ramips' [`Device/Default` sets `BLOCKSIZE := 64k`](https://github.com/openwrt/openwrt/blob/0e6399a2e4d561c0e466c2fb6437178c41d6237c/target/linux/ramips/image/Makefile#L181), and no other ramips device overrides it below 64k.

This looks like the root cause of the problem that `2438c96` then works around: [`Build/pad-rootfs`](https://github.com/openwrt/openwrt/blob/0e6399a2e4d561c0e466c2fb6437178c41d6237c/include/image-commands.mk#L683-L686) passes `$(BLOCKSIZE:%k=%)` to `padjffs2`, so with `4k` the `deadc0de` JFFS2 EOF markers only land on 4 KiB boundaries. `mtd` reads in erasesize-sized chunks and only recognises the marker at [the start of an erase block](https://github.com/openwrt/openwrt/blob/0e6399a2e4d561c0e466c2fb6437178c41d6237c/package/system/mtd/src/mtd.c#L637-L638) — on a 64 KiB-erasesize MTD it will never find one, so it neither appends the config archive nor stops early, and the blocks past the image are left untouched. That is exactly the "a shorter squashfs image can leave an old JFFS2 overlay in the later blocks" symptom.

Both 14912k and the Giga II's 14976k are multiples of 64k, so this should be a drop-in change:

```suggestion
  BLOCKSIZE := 64k
```

Same applies to line 18 for `zyxel_keenetic-giga-ii`.

---

target/linux/ramips/rt6855a/base-files/lib/upgrade/platform.sh
```diff
@@ -0,0 +1,25 @@
...
+	if [ -z "$UPGRADE_BACKUP" ] && [ -s /tmp/sysupgrade.tgz ]; then
+		export UPGRADE_BACKUP=/tmp/sysupgrade.tgz
+	fi
+
+	# mtd write only erases blocks covered by the new image. Erase the whole
+	# combined firmware partition first so stale JFFS2 data cannot survive a
+	# sysupgrade when the new squashfs image is shorter than the old one.
+	#
+	# The following -n prevents mtd from erasing the written blocks twice.
+	MTD_ARGS="-e firmware -n"
```
openwrt-ai, line 22: No other target in the tree sets `MTD_ARGS` to erase the target partition up front, and this changes the failure mode of an interrupted sysupgrade: with `-e firmware` the whole partition is blank before the first byte of the new image is written, so a power loss during the write leaves an unbootable device instead of a partially-overwritten but still-bootable one. The upstream mechanism is `mtd`'s JFFS2 EOF handling, which erases and rewrites everything from the end of the rootfs onward — see [`default_do_upgrade`](https://github.com/openwrt/openwrt/blob/0e6399a2e4d561c0e466c2fb6437178c41d6237c/package/base-files/files/lib/upgrade/common.sh#L309-L318) and the `JFFS2_EOF` handling in `mtd`.

I believe that mechanism is only broken here because of `BLOCKSIZE := 4k` (see my comment on `rt6855a.mk`). Could you retest with `BLOCKSIZE := 64k` and drop this override if the stale overlay no longer survives? If the workaround is still needed, please note in the commit message why the standard `deadc0de` path does not work on this flash.

Also note `-n` disables `mtd`'s bad-block skipping in the write loop; harmless on NOR, but it is a side effect worth being explicit about.

---

target/linux/ramips/rt6855a/base-files/lib/upgrade/platform.sh
```diff
@@ -0,0 +1,25 @@
...
+platform_check_image() {
+	return 0
+}
+
+platform_do_upgrade() {
+	# The backup normally arrives through procd's UPGRADE_BACKUP environment.
+	# Keep the fallback from the Giga II sysupgrade fix for the ramfs pivot.
+	if [ -z "$UPGRADE_BACKUP" ] && [ -s /tmp/sysupgrade.tgz ]; then
+		export UPGRADE_BACKUP=/tmp/sysupgrade.tgz
+	fi
```
openwrt-ai, line 15: This fallback is unique in the tree — no other target reconstructs `UPGRADE_BACKUP` in `platform_do_upgrade`. procd exports it unconditionally whenever a backup was requested ([`sysupgrade.c:52-53`](https://github.com/openwrt/procd/blob/60fdbf00e924fff0a510c4ec31c520d35e69bfe1/sysupgrade.c#L52-L53), fed from `json_add_string backup "$CONF_TAR"` which `/sbin/sysupgrade` only emits when `SAVE_CONFIG=1`), so on the normal path the variable is never empty when a backup exists.

Which path actually hit the empty variable? If it was failsafe mode (where `/sbin/sysupgrade` bypasses procd entirely and never sets `UPGRADE_BACKUP`), that is a generic base-files bug and the fix belongs there rather than in one subtarget's `platform.sh`. Papering over it per-target means every other target still silently drops the config in the same situation.

Note also that this hunk is scope beyond the commit subject (`erase complete firmware on sysupgrade`); the body mentions it, but it would be cleaner as its own commit.

---

target/linux/ramips/image/rt6855a.mk
```diff
@@ -0,0 +1,46 @@
...
+  DEVICE_VENDOR := ZyXEL
```
openwrt-ai, line 34: nit: all 43 existing `DEVICE_VENDOR` entries for this vendor across `target/linux/*/image/*.mk` spell it `Zyxel`, including the other Keenetics in ramips. `DEVICE_VENDOR` feeds the image filename and the device title, so the inconsistent casing shows up in build artifacts.

```suggestion
  DEVICE_VENDOR := Zyxel
```

Same applies to line 19 for `zyxel_keenetic-giga-ii`.

---

target/linux/ramips/image/rt6855a.mk
```diff
@@ -0,0 +1,46 @@
...
+  DEVICE_MODEL := Keenetic Ultra
+  IMAGES += factory.bin
+  IMAGE/factory.bin := $$(sysupgrade_bin) | pad-to 64k | check-size | \
+	pad-to-ff $$$$(IMAGE_SIZE) | \
+	zyimage -d 22037 -v "ZyXEL Keenetic Ultra"
+  SUPPORTED_DEVICES += ku_ra
```
openwrt-ai, line 40: nit: `SUPPORTED_DEVICES` exists so that sysupgrade accepts an image on a device whose *previously released OpenWrt* metadata used a different board name — the values on the other Keenetics ([`kn_rc`](https://github.com/openwrt/openwrt/blob/0e6399a2e4d561c0e466c2fb6437178c41d6237c/target/linux/ramips/image/mt7620.mk#L1630), `kn_rf`, `kng_rc`) are legacy pre-DT board names carried over from ar71xx/ramips-era releases. `rt6855a` is a brand-new subtarget with no prior OpenWrt release, and `ku_ra` appears nowhere else in the tree or its history, so nothing can ever be running an image that reports it. The sibling `zyxel_keenetic-giga-ii` correctly has no `SUPPORTED_DEVICES`.

Is this intentional (e.g. a third-party fork's board name you want to accept upgrades from)? If not it can just be dropped.

---

target/linux/generic/files/drivers/net/phy/rtl8366_smi.h
```diff
@@ -54,6 +54,9 @@ struct rtl8366_smi {
 	struct mii_bus		*mii_bus;
 	int			mii_irq[PHY_MAX_ADDR];
 	struct switch_dev	sw_dev;
+	u16			led_regs_saved[3];
```
openwrt-ai, line 57: The `3` here is silently coupled to `ARRAY_SIZE(rtl8367b_led_force_regs)` in a different file (`rtl8367b.c`), and both the save and the restore loop index this array with `ARRAY_SIZE(rtl8367b_led_force_regs)`. Adding a fourth entry to that register list — which is a plausible follow-up for another RTL8367 revision — silently overflows `led_regs_saved` into `led_regs_saved_valid`/`leds_off` and beyond, with no compiler diagnostic.

Please tie the two together, e.g. by giving the register list a `#define RTL8367B_LED_FORCE_REG_COUNT 3` in the shared header and using it in both places, or add a `BUILD_BUG_ON(ARRAY_SIZE(rtl8367b_led_force_regs) > ARRAY_SIZE(smi->led_regs_saved))` in the setter.

---

target/linux/generic/files/drivers/net/phy/rtl8367b.c
```diff
@@ -1295,6 +1340,75 @@ static int rtl8367b_sw_set_max_length(struct switch_dev *dev,
...
+		return 0;
+	}
+
+	if (!smi->led_regs_saved_valid)
+		return -EINVAL;
+
+	for (i = 0; i < ARRAY_SIZE(rtl8367b_led_force_regs); i++) {
+		data = smi->led_regs_saved[i];
+		if (data == RTL8367B_LED_FORCE_OFF)
+			data = 0;
```
openwrt-ai, line 1401: This special case silently corrupts the snapshot it is supposed to restore: if a board legitimately has `0x02aa` in one of the LED force registers (which is a valid "force off" configuration the board could ship with), the restore writes `0` instead of the value that was actually read, so the attribute is no longer reversible.

The guard at the top (`if (!!val->value.i == smi->leds_off) return 0;`) already prevents re-snapshotting while LEDs are off, so the saved value should never be a value this code needs to reinterpret. Either drop the remap, or add a comment explaining which concrete situation produces a saved `RTL8367B_LED_FORCE_OFF` that must not be written back.

---

target/linux/generic/files/drivers/net/phy/rtl8367b.c
```diff
@@ -1295,6 +1340,75 @@ static int rtl8367b_sw_set_max_length(struct switch_dev *dev,
...
+		ret = rtl8366_smi_write_reg(smi, rtl8367b_led_force_regs[i],
+					    data);
+		if (ret)
+			return ret;
+	}
+	smi->leds_off = false;
+
+	return 0;
+}
+
```
openwrt-ai, line 1412: nit: two consecutive blank lines here (the added one plus the pre-existing one). `checkpatch.pl --strict` reports `CHECK: Please don't use multiple blank lines`, which contradicts the "0 errors, warnings or checks" claim in the PR description.

```suggestion

```

---

target/linux/ramips/rt6855a/base-files/etc/hotplug.d/net/20-rt3593-sta-rates
```diff
@@ -0,0 +1,17 @@
...
+[ "$ACTION" = "add" ] || exit 0
+[ "$(cat /tmp/sysinfo/board_name 2>/dev/null)" = "zyxel,keenetic-ultra" ] ||
+	exit 0
+[ -n "$DEVICENAME" ] || exit 0
+
+info="$(iw dev "$DEVICENAME" info 2>/dev/null)" || exit 0
+[ "$(printf '%s\n' "$info" | awk '$1 == "type" { print $2 }')" = "managed" ] ||
+	exit 0
+[ "$(printf '%s\n' "$info" | awk '$1 == "wiphy" { print $2 }')" = "0" ] ||
+	exit 0
```
openwrt-ai, line 13: Hardcoding `wiphy 0` to mean "the RT3593" is fragile. wiphy indices are assigned in registration order and are monotonically increasing for the lifetime of the boot — an `rmmod`/`insmod` of one radio, a probe failure on the PCIe-0 device, or a USB WLAN dongle registering first all shift the numbering, and the failure is silent (the script just exits 0 and the STA runs at the rates the commit calls unstable).

Since the two radios are distinguishable by chip, matching on the driver-visible identity would be robust — e.g. resolve the phy from `/sys/class/net/$DEVICENAME/phy80211` and check `device/of_node` or the PCI ID against the RT3593, rather than the index.

---

target/linux/ramips/rt6855a/base-files/etc/hotplug.d/net/20-rt3593-sta-rates
```diff
@@ -0,0 +1,17 @@
...
+
+if iw dev "$DEVICENAME" set bitrates ht-mcs-5 0 1 2 3 4 5 6 7; then
```
openwrt-ai, line 15: This series already establishes a device-tree mechanism for exactly this class of per-board radio limitation: `79430521` adds `ralink,tx-chains`, consumed in `rt2800_probe_hw_mode()` via `900-rt2800-limit-giga-ii-rt5392-to-one-tx-stream.patch` (currently gated on `rt2x00_rf(rt2x00dev, RF5392)`). Working around the RT3593 in a board-specific userspace hotplug script instead means the limit is invisible to mac80211's rate control, is not applied to AP or mesh vifs, races with `wifi reload`, and silently does nothing if the `iw` call or the wiphy match fails.

Is there a reason the RT3593 case cannot use the same DT property (relaxing the `RF5392` gate), or a rate-mask applied in `rt2800_probe_hw_mode()`? If the userspace approach really is required, please say so in the commit message — right now `Limit only the RT3593 managed interface to the stable single-stream HT rate set` does not explain why it is not done in the driver like the RT5392 sibling.

---

target/linux/ramips/rt6855a/base-files/etc/hotplug.d/button/50-rt6855a-led-shutdown
```diff
@@ -0,0 +1,63 @@
...
+fn|lights_toggle) ;;
+*) exit 0 ;;
+esac
+[ "$ACTION" = "released" ] || exit 0
+
+model="$(tr -d '\000' </proc/device-tree/model 2>/dev/null)"
+case "$model" in
+*"Keenetic Ultra"*) ;;
+*) exit 0 ;;
+esac
```
openwrt-ai, line 14: nit: substring-matching the DT `model` string is not the in-tree way to identify a board, and it is inconsistent with the sibling script added in the same commit (`../net/20-rt3593-sta-rates` line 5), which uses the board name. Model strings are free text and a future `ZyXEL Keenetic Ultra II` would also match `*"Keenetic Ultra"*`.

```suggestion
. /lib/functions.sh

case "$(board_name)" in
zyxel,keenetic-ultra) ;;
*) exit 0 ;;
esac
```

---

target/linux/ramips/rt6855a/base-files/etc/hotplug.d/button/50-rt6855a-led-shutdown
```diff
@@ -0,0 +1,63 @@
+#!/bin/sh
+# SPDX-License-Identifier: GPL-2.0-only
+
+case "$BUTTON" in
+fn|lights_toggle) ;;
```
openwrt-ai, line 5: nit: `fn` is dead. `BUTTON` is derived exclusively from the keycode map in gpio-button-hotplug, and [`KEY_LIGHTS_TOGGLE` maps to `"lights_toggle"`](https://github.com/openwrt/openwrt/blob/0e6399a2e4d561c0e466c2fb6437178c41d6237c/package/kernel/gpio-button-hotplug/src/gpio-button-hotplug.c#L92) — the DT `label = "fn"` never reaches this variable.

```suggestion
lights_toggle) ;;
```

---

target/linux/ramips/rt6855a/base-files/etc/board.d/01_leds
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+# SPDX-License-Identifier: GPL-2.0-only
+
+. /lib/functions/leds.sh
+. /lib/functions/uci-defaults.sh
+
+board_config_update
+
+case "$(board_name)" in
+zyxel,keenetic-ultra)
```
openwrt-ai, line 10: nit: sister-device parity — `zyxel,keenetic-giga-ii` has the identical `led-internet` node (`gpio0 0`, `LED_FUNCTION_WAN`) and the identical RTL8367 port mapping (`02_network` now shares one case arm for both boards), and its commit message lists "Internet" among the LEDs, but it is not covered here, so its WAN LED gets no switch-link trigger. Is that omission intentional, or should this case arm cover both boards like `02_network` does?

---

package/kernel/mac80211/patches/rt2x00/623-rt2x00-rearm-rt2800-rx-descriptor-length.patch
```diff
@@ -0,0 +1,29 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
openwrt-ai, line 1: nit: the all-zero commit hash in the `From` line is a fabricated placeholder — no such commit exists, and `git format-patch` never emits it. In-tree rt2x00 patches either carry the real upstream hash (`602-01`, `620-01`, `621-01`, ...) or omit the mail header entirely for locally-authored patches (`608-`, `610-`, `611-`). Since this is not upstream yet, dropping the `From <sha>` line matches the existing convention and avoids implying an upstream reference that cannot be looked up.

Same applies to `900-rt2800-limit-giga-ii-rt5392-to-one-tx-stream.patch` and `901-rt2800-rt5392-front-panel-led.patch` from the dependency series.

---

openwrt-ai (commented):
Reviewed 14 commits (first review of this PR). Kernel-side comparisons were made against `linux v6.18.39`, matching `KERNEL_PATCHVER:=6.18` for ramips.

## Commit checks

- `0e6399a` "ramips: add support for ZyXEL Keenetic Ultra" — this is a new-device commit (new `target/linux/ramips/dts/rt6855a_zyxel_keenetic-ultra.dts` plus a new `define Device/zyxel_keenetic-ultra` in `target/linux/ramips/image/rt6855a.mk`), so the commit message needs the three required sections. Two are missing entirely:
  - **Flash instructions** — absent. The PR description has a complete U-Boot option-2 TFTP procedure (serial 57600 8N1, image in the TFTP root, router IP / server IP / filename prompts, plus the option-1 initramfs path), but none of it is in the commit message. PR descriptions are not preserved in the tree; the sibling `df6c4ec` (Giga II) carries this correctly.
  - **MAC address layout** — absent. Again the PR description has it (`RF-EEPROM offset 0x4` for Ethernet/LAN/WAN, RT5392 and RT3593 base addresses from RF-EEPROM, virtual interfaces derived from the radio base), and `df6c4ec` includes it, but the Ultra commit does not.
  - The **Hardware specification** section is present but incomplete relative to the other two: no UART header location/pinout (only the `57600` baud rate appears, and only in the PR description), and no power supply rating. Flash and RAM chip markings would be welcome if known.

  Moving the corresponding blocks from the PR description into the commit message resolves all three.

The other four new commits (`f9335ad`, `0fb1418`, `dc5bcce`, `2438c96`) have messages that match their diffs. `2438c96` does carry a second, unrelated change (the `UPGRADE_BACKUP` fallback); it is described in the body, but it would read better as a separate commit — see the inline comment.

Inline comments below cover the rest. The one I would prioritise is `BLOCKSIZE := 4k` in `rt6855a.mk`: it is almost certainly why the JFFS2 EOF mechanism fails on this board, and fixing it may make the `MTD_ARGS="-e firmware -n"` sysupgrade override unnecessary.

Findings on the first nine (dependency-integration) commits are limited to what is still visible in the merged diff; the substance of #24249 and #24305 is better reviewed on those PRs.

---

PR #24443, opened 2026-07-27
kernel: modules: add kmod-tpm-crb kernel module package

drujd:
It is used by firmware-based TPMs in modern AMD/Intel/ARM processors.

This is part of my changes to have OpenWrt booting as a Secure Boot signed binary that uses TPM and clevis to unlock LUKS partitions after initial boot.

### DIFF
--- a/package/kernel/linux/modules/other.mk
+++ b/package/kernel/linux/modules/other.mk
@@ -1078,6 +1078,22 @@ endef
 
 $(eval $(call KernelPackage,tpm))
 
+define KernelPackage/tpm-crb
+  SUBMENU:=$(OTHER_MENU)
+  TITLE:=TCG CRB 2.0 TPM Interface
+  DEPENDS:=@(TARGET_x86||TARGET_armsr) +kmod-tpm
+  KCONFIG:=CONFIG_TCG_CRB
+  FILES:=$(LINUX_DIR)/drivers/char/tpm/tpm_crb.ko
+  AUTOLOAD:=$(call AutoLoad,15,tpm_crb,1)
+endef
+
+define KernelPackage/tpm-crb/description
+  This device driver implements the TPM interface as defined in
+  the TCG CRB 2.0 TPM specification.
+endef
+
+$(eval $(call KernelPackage,tpm-crb))
+
 define KernelPackage/tpm-ftpm-tee
   SUBMENU:=$(OTHER_MENU)
   TITLE:=TEE based firmware TPM (fTPM)


package/kernel/linux/modules/other.mk
```diff
@@ -1078,6 +1078,22 @@ endef
 
 $(eval $(call KernelPackage,tpm))
 
+define KernelPackage/tpm-crb
+  SUBMENU:=$(OTHER_MENU)
+  TITLE:=TCG CRB 2.0 TPM Interface
+  DEPENDS:=@(TARGET_x86||TARGET_armsr) +kmod-tpm
```
openwrt-ai, line 1084: nit: `CONFIG_TCG_CRB` only carries `depends on ACPI` in the kernel ([drivers/char/tpm/Kconfig:205-207, v6.12.94](https://github.com/gregkh/linux/blob/0b8f247169e487eff2d4c2dd531bc43f7efda2cb/drivers/char/tpm/Kconfig#L205-L207) — unchanged in v6.18.39), and `loongarch64` also sets [`CONFIG_ACPI=y`](https://github.com/openwrt/openwrt/blob/ea6e94a72ed8ecd7746de87521011169e39c0a3d/target/linux/loongarch64/config-6.12#L9), so the driver would be buildable there too. The sibling [`kmod-tpm-tis`](https://github.com/openwrt/openwrt/blob/ea6e94a72ed8ecd7746de87521011169e39c0a3d/package/kernel/linux/modules/other.mk#L1101) also leaves loongarch64 out, so this may simply be following that precedent — is the omission intentional, or should `TARGET_loongarch64` be included?

---

openwrt-ai (commented):
Reviewed 1 new commit; one minor question inline.

Verified against the pre-cloned kernel trees: `FILES` path (`drivers/char/tpm/tpm_crb.ko`) matches `obj-$(CONFIG_TCG_CRB) += tpm_crb.o` in `drivers/char/tpm/Makefile` on both v6.12.94 (armsr) and v6.18.39 (x86), `CONFIG_TCG_CRB` is `# not set` in both generic configs, the `+kmod-tpm` dependency is sufficient (`tpm_crb_ffa` helpers are inline stubs unless `CONFIG_TCG_ARM_CRB_FFA` is on, which this package does not enable), and the `AutoLoad,15` ordering sits correctly after `kmod-tpm` (10). Commit message matches the diff and carries a Signed-off-by.

Note the bot's `PKG_RELEASE` warning appears to be a false positive here — `package/kernel/linux/Makefile` defines no `PKG_RELEASE`.

---

package/kernel/linux/modules/other.mk
drujd: Loongson Security Engine has its own driver, it does not utilize CRB nor TIS.

PR #24449, opened 2026-07-28
build: add target SDK to index.json

aparcar:
The OpenWrt build system has no mapping between a package and the SDK. Multiple SDKs may be used to create a package, however this information is nowhere available.

This information however becomes relevant when trying to rebuild an upstream package, i.e. to test reproducibility. A package is given with a specific architecture, however it's unclear which target SDK built the package.

To solve this, add the target/subtarget to the index.json. Now any downstream tooling can check the index.json file and find the corresponding target.

### DIFF
--- a/package/Makefile
+++ b/package/Makefile
@@ -141,7 +141,9 @@ ifneq ($(CONFIG_USE_APK),)
 			--output packages.adb \
 			*.apk; \
 		$(STAGING_DIR_HOST)/bin/apk adbdump --format json packages.adb | \
-			$(SCRIPT_DIR)/make-index-json.py -f apk -a "$(ARCH_PACKAGES)" - > index.json; \
+			$(SCRIPT_DIR)/make-index-json.py -f apk -a "$(ARCH_PACKAGES)" \
+				-t "$(BOARD)$(if $(SUBTARGET),/$(SUBTARGET))" \
+				- > index.json; \
 	done
 ifdef CONFIG_JSON_CYCLONEDX_SBOM
 	@echo Creating CycloneDX package SBOMs...
@@ -163,7 +165,9 @@ else
 			$(call ERROR_MESSAGE,WARNING: Applying padding in $$d/Packages to workaround usign SHA-512 bug!); \
 			{ echo ""; echo ""; } >> Packages;; \
 		esac; \
-		$(SCRIPT_DIR)/make-index-json.py -f opkg -a "$(ARCH_PACKAGES)" Packages > index.json; \
+		$(SCRIPT_DIR)/make-index-json.py -f opkg -a "$(ARCH_PACKAGES)" \
+			-t "$(BOARD)$(if $(SUBTARGET),/$(SUBTARGET))" \
+			Packages > index.json; \
 		gzip -9nc Packages > Packages.gz; \
 	); done
 ifdef CONFIG_SIGNED_PACKAGES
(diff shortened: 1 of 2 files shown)


package/Makefile
```diff
@@ -141,7 +141,9 @@ ifneq ($(CONFIG_USE_APK),)
 			--output packages.adb \
 			*.apk; \
 		$(STAGING_DIR_HOST)/bin/apk adbdump --format json packages.adb | \
-			$(SCRIPT_DIR)/make-index-json.py -f apk -a "$(ARCH_PACKAGES)" - > index.json; \
+			$(SCRIPT_DIR)/make-index-json.py -f apk -a "$(ARCH_PACKAGES)" \
+				-t "$(BOARD)$(if $(SUBTARGET),/$(SUBTARGET))" \
```
openwrt-ai, line 145: The same `-t` value is written into every entry of `PACKAGE_SUBDIRS`, but those directories do not all belong to one target. [`include/feeds.mk:11-14`](https://github.com/openwrt/openwrt/blob/24715fafa0cac1627dd879fd7923bc463c818a7e/include/feeds.mk#L11-L14) expands to `$(PACKAGE_DIR)` (= [`$(OUTPUT_DIR)/targets/$(BOARD)/$(SUBTARGET)/packages`](https://github.com/openwrt/openwrt/blob/24715fafa0cac1627dd879fd7923bc463c818a7e/rules.mk#L150-L151), target-specific — fine) *plus* `$(OUTPUT_DIR)/packages/$(ARCH_PACKAGES)/{base,<feed>}`, which is keyed on `ARCH_PACKAGES` only and is therefore shared by every target with the same package arch (e.g. ath79/generic, ath79/nand and ath79/mikrotik all write `bin/packages/mips_24kc/base/index.json`). Since `package/index` rewrites that file unconditionally, the recorded `target` there ends up being whichever target ran last, even though the `.apk`/`.ipk` files in the directory may have come from several of them — the exact ambiguity the commit message describes.

Is last-writer-wins the intended semantics for the arch-shared feed indexes, or should `target` be emitted only for the target-specific `$(PACKAGE_DIR)` index? Worth a sentence in the commit message either way, since downstream reproducibility tooling will be trusting this field.

Same expression on line 169 for the opkg branch.

---

scripts/make-index-json.py
```diff
@@ -100,4 +102,6 @@ def parse_opkg(text: str) -> dict:
             "architecture": args.architecture,
             "packages": packages,
         }
+        if args.target is not None:
+            index["target"] = args.target
```
openwrt-ai, line 106: nit: `is not None` still emits `"target": ""` when the argument is passed but empty. The Makefile always passes `-t`, and `$(BOARD)$(if $(SUBTARGET),/$(SUBTARGET))` collapses to the empty string whenever `CONFIG_TARGET_BOARD` is unset, so a plain truthiness test is the more useful guard — omit the key rather than record a meaningless one.

```suggestion
        if args.target:
            index["target"] = args.target
```

Separately: assigning after the literal places `target` at the very bottom of `index.json`, after the entire package list — a few thousand lines away from `version`/`architecture`. Declaring it inside the dict literal next to `architecture` would keep the metadata keys together for anyone eyeballing the file.

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the diff, and the emitted `target` string uses the same `$(BOARD)$(if $(SUBTARGET),/$(SUBTARGET))` form as [`TARGETID` in include/target.mk:77](https://github.com/openwrt/openwrt/blob/24715fafa0cac1627dd879fd7923bc463c818a7e/include/target.mk#L77), so it will join cleanly against the `target` field already present in `profiles.json`. Two inline notes, one of them a question about which directories the field is meaningful for.

---

PR #24450, opened 2026-07-28
mediatek: mt7988a: drop unused 2.5G PHY firmware node

osedlbauer:
We now use the mainline 2.5G PHY driver, which hardcodes both memory regions instead of taking them from the DT. The node is no longer used.

For reference: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/drivers/net/phy/mediatek/mtk-2p5ge.c?h=v6.18.40#n16


openwrt-ai (commented):

---

PR #24452, opened 2026-07-28
[25.12] mediatek: fix 2.5G PHY MCU CSR base address in mt7988.dtsi

osedlbauer:
The second reg entry of the phy-firmware node holds the wrong address. The driver adds MD32_EN_CFG (0x18) to it, but the node already contains the register address including that offset, so the access ends up at 0x0f0f0030. The mainline driver and the Mediatek SDK both use 0x0f0f0000 as the MCU CSR base.

Without this fix, loading the PHY firmware occasionally hangs on MT7988 when the interface is brought up for the first time after a cold boot, causing the watchdog to trigger a reboot.

For reference:
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/drivers/net/phy/mediatek/mtk-2p5ge.c?h=v6.18.40#n18
https://github.com/openwrt/openwrt/blob/openwrt-25.12/target/linux/mediatek/patches-6.12/701-net-phy-mediatek-add-driver-for-built-in-2.5G-ethern.patch#L145

### DIFF
--- a/target/linux/mediatek/patches-6.12/173-dts-mt7988a-Add-built-in-ethernet-phy-firmware-node.patch
+++ b/target/linux/mediatek/patches-6.12/173-dts-mt7988a-Add-built-in-ethernet-phy-firmware-node.patch
@@ -19,7 +19,7 @@ Signed-off-by: Sky Huang <skylake.huang@mediatek.com>
 +		phyfw: phy-firmware@f000000 {
 +			compatible = "mediatek,2p5gphy-fw";
 +			reg = <0 0x0f100000 0 0x20000>,
-+			      <0 0x0f0f0018 0 0x20>;
++			      <0 0x0f0f0000 0 0x20>;
 +		};
 +
  		usb@11190000 {
--- a/target/linux/mediatek/patches-6.12/186-arm64-dts-mt7988a-complete-dtsi.patch
+++ b/target/linux/mediatek/patches-6.12/186-arm64-dts-mt7988a-complete-dtsi.patch
@@ -123,7 +123,7 @@ Work-in-progress patch to complete mt7988a.dtsi
  			compatible = "mediatek,mt7981-i2c";
  			reg = <0 0x11003000 0 0x1000>,
 @@ -425,7 +512,7 @@
- 			      <0 0x0f0f0018 0 0x20>;
+ 			      <0 0x0f0f0000 0 0x20>;
  		};
  
 -		usb@11190000 {
--- a/target/linux/mediatek/patches-6.12/700-net-phy-mediatek-Add-2.5Gphy-firmware-dt-bindings-an.patch
+++ b/target/linux/mediatek/patches-6.12/700-net-phy-mediatek-Add-2.5Gphy-firmware-dt-bindings-an.patch
@@ -40,7 +40,7 @@ Signed-off-by: Sky Huang <skylake.huang@mediatek.com>
 +  reg:
 +    items:
 +      - description: pmb firmware load address
-+      - description: firmware trigger register
++      - description: MCU CSR base address
 +
 +required:
 +  - compatible
@@ -53,7 +53,7 @@ Signed-off-by: Sky Huang <skylake.huang@mediatek.com>
 +    phyfw: phy-firmware@f000000 {
 +      compatible = "mediatek,2p5gphy-fw";
 +      reg = <0 0x0f100000 0 0x20000>,
-+            <0 0x0f0f0018 0 0x20>;
++            <0 0x0f0f0000 0 0x20>;
 +    };
 --- a/MAINTAINERS
 +++ b/MAINTAINERS


osedlbauer:
This is not a backport. There is no commit in main branch to cherry-pick from.
The node is not used and should be removed there (see https://github.com/openwrt/openwrt/pull/24450).

target/linux/mediatek/patches-6.12/173-dts-mt7988a-Add-built-in-ethernet-phy-firmware-node.patch
```diff
@@ -19,7 +19,7 @@ Signed-off-by: Sky Huang <skylake.huang@mediatek.com>
 +		phyfw: phy-firmware@f000000 {
 +			compatible = "mediatek,2p5gphy-fw";
 +			reg = <0 0x0f100000 0 0x20000>,
-+			      <0 0x0f0f0018 0 0x20>;
++			      <0 0x0f0f0000 0 0x20>;
```
openwrt-ai, line 22: The DT binding shipped alongside this node still documents the old value, so after this change the binding and the DTS disagree. In [`700-net-phy-mediatek-Add-2.5Gphy-firmware-dt-bindings-an.patch:42-56`](``https://github.com/openwrt/openwrt/blob/d7725055341f9b70932ac382b838fd972168930b/target/linux/mediatek/patches-6.12/700-net-phy-mediatek-Add-2.5Gphy-firmware-dt-bindings-an.patch#L42-L56)`` the second `reg` entry is described as `firmware trigger register` and the `examples:` block still uses `<0 0x0f0f0018 0 0x20>`.

That description is arguably the root cause of the mismatch: the driver treats the entry as the MCU CSR *base* and adds `MD32_EN_CFG` on top ([`of_iomap(np, 1)` / `readw(mcu_csr_base + MD32_EN_CFG)` in 701-...patch:125-145](``https://github.com/openwrt/openwrt/blob/d7725055341f9b70932ac382b838fd972168930b/target/linux/mediatek/patches-6.12/701-net-phy-mediatek-add-driver-for-built-in-2.5G-ethern.patch#L125-L145)),`` matching mainline's hardcoded [`MTK_2P5GPHY_MCU_CSR_BASE 0x0f0f0000` + `MD32_EN_CFG 0x18` in mtk-2p5ge.c:18-20](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/net/phy/mediatek/mtk-2p5ge.c#L18-L20) (v6.18.39). So the base-address fix here is right, but consider updating the example and re-wording the description to "MCU CSR base" in the same commit — this binding never landed upstream (there is no `mediatek,2p5gphy-fw.yaml` in v6.18), so it is OpenWrt-local and safe to correct.

---

target/linux/mediatek/patches-6.12/186-arm64-dts-mt7988a-complete-dtsi.patch
```diff
@@ -123,7 +123,7 @@ Work-in-progress patch to complete mt7988a.dtsi
  			compatible = "mediatek,mt7981-i2c";
  			reg = <0 0x11003000 0 0x1000>,
 @@ -425,7 +512,7 @@
- 			      <0 0x0f0f0018 0 0x20>;
+ 			      <0 0x0f0f0000 0 0x20>;
```
openwrt-ai, line 126: nit: the commit subject says `mt7988.dtsi`, but both patches touched here modify `arch/arm64/boot/dts/mediatek/mt7988a.dtsi` (see this patch's own header line `Work-in-progress patch to complete mt7988a.dtsi`). Worth amending the subject to `mt7988a.dtsi` so it names the real file.

---

openwrt-ai (commented):
Reviewed 1 new commit. The fix itself checks out for the 25.12 / kernel 6.12 tree: `mt7988_2p5ge_phy_load_fw()` there really does `of_iomap(np, 1)` and then `readw(mcu_csr_base + MD32_EN_CFG)`, so with `0x0f0f0018` in the node the access lands at `0x0f0f0030`. Both occurrences of the value in the patch stack (`173-...` and the context line in `186-...`) are updated consistently, and `0x0f0f0000 + 0x20` still covers the `0x18` offset. Two inline notes below; the cherry-pick-trailer CI failure is already explained in the thread (main's 6.18 driver hardcodes `MTK_2P5GPHY_MCU_CSR_BASE` and does not read the node, which I confirmed), so a maintainer `[allow cherry-pick]` override looks appropriate here.

---

target/linux/mediatek/patches-6.12/173-dts-mt7988a-Add-built-in-ethernet-phy-firmware-node.patch
osedlbauer: Updated binding documentation as suggested

target/linux/mediatek/patches-6.12/700-net-phy-mediatek-Add-2.5Gphy-firmware-dt-bindings-an.patch
```diff
@@ -40,7 +40,7 @@ Signed-off-by: Sky Huang <skylake.huang@mediatek.com>
 +  reg:
 +    items:
 +      - description: pmb firmware load address
-+      - description: firmware trigger register
++      - description: MCU CSR base address
```
openwrt-ai, line 43: nit: the commit now also corrects the binding schema (`Documentation/devicetree/bindings/net/mediatek,2p5gphy-fw.yaml`: `reg` item description and the `examples:` block), but the subject and body only talk about `mt7988a.dtsi`. Worth adding a sentence to the body — e.g. "Also fix the second `reg` entry's description and example in the accompanying dt-binding, which described it as the firmware trigger register rather than the MCU CSR base." — so the extra hunks aren't undocumented.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push `d772505` → `bc142d4`). Both earlier findings are addressed: the commit subject now names `mt7988a.dtsi`, and the dt-binding in `700-net-phy-mediatek-Add-2.5Gphy-firmware-dt-bindings-an.patch` now describes the second `reg` entry as "MCU CSR base address" with a matching `<0 0x0f0f0000 0 0x20>` example, so binding and DTS agree again.

Re-checked consistency across the 6.12 patch stack: the only remaining `0x0f0f00xx` literals are the three updated `0x0f0f0000` sites (173-, 186-, 700-) plus `#define MTK_2P5GPHY_MCU_CSR_BASE (0x0f0f0000)` in `752-net-phy-mediatek-i2p5g-add-support-for-mt7987.patch`, which corroborates the value. `752-` keeps `mt7988_2p5ge_phy_load_fw()` on the `of_iomap(np, 1)` + `readw(mcu_csr_base + MD32_EN_CFG)` path (only the new mt7987 function uses the hardcoded base), so the DTS value is still live for MT7988 and the fix is required.

CI: all kernel builds and the patch check are green on `bc142d4`; the sole failure is `FormalityCheck / Git & Commits` on the missing `(cherry picked from commit ...)` trailer, which is expected here — there is no main-branch commit to cherry-pick from (main's 6.18 driver hardcodes the base and does not read the node), so a maintainer `[allow cherry-pick]` override remains the right resolution. One nit inline.

---

PR #24453, opened 2026-07-28
scripts: feeds: support per-feed --post_update hook flag

micpf:
### Problem

Downstream trees frequently need to run custom logic after a feed is fetched or refreshed — apply a locally-maintained patch series on top of an upstream feed, generate provenance/SBOM data, stage overlay files into the feed, etc. Today there is no extension point for this, so each downstream carries a private patch to `scripts/feeds` that inserts a single `system(...)` call at the same spot and hardcodes a path to its own script.

### Change

Extend the existing `feeds.conf` flag syntax with an opt-in `--post_update=<script>` flag. After a feed's per-VCS `post_update` step, if the flag is set on the feed's source line, run the script with the feed name and local path as arguments. Non-zero exit fails the update. Absent flag preserves current behavior.

```
src-git  --post_update=scripts/apply-overlay.sh packages URL
src-cpy  --force --post_update=scripts/stage-core.sh core ../core
src-link vendor ../vendor
```

Flags must appear between `src-<type>` and the feed name — that is the only position the feed-line parser (regex at `scripts/feeds:58`) recognises the `--<word>=<value>` flag run. Three lines in `scripts/feeds`; no schema or grammar change (the existing parser already accepts `--<word>=<value>`).

### Why this shape

- **Per-feed script, not one global hook.** Each feed can have distinct behavior without downstreams building a global if/else dispatcher.
- **Configuration in `feeds.conf`.** Feed-related settings live next to the feed definition rather than in the environment or a magic directory. Consistent with how `--force` already works.
- **No on-disk footprint.** No `feeds.d/` directory to invent; important for consumers who take openwrt as a git subtree, where extra tracked or untracked files complicate merges.
- **Backward compatible.** Flag unset ⇒ zero behavior change.

### Testing

Tested against a tree with five feeds (`src-git`, `src-cpy`, `src-link`, with and without `--force`). Each line was given a `--post_update=<script>` pointing at the same downstream helper. On a clean `feeds/` directory, `feeds update -a` invoked the hook once per feed with the expected `<feed_name> <feed_path>` arguments after the per-VCS post_update ran. A hook returning non-zero correctly failed the enclosing update. Removing the flag from a line reverted that feed to unhooked behavior with the rest still hooking.

### DIFF
--- a/scripts/feeds
+++ b/scripts/feeds
@@ -74,6 +74,10 @@ sub parse_file($$) {
 			}
 		}
 
+		if (defined $flags && $flags =~ /(?:^|\s)--post_update(?:\s|$)/) {
+			die "Feed '$name' in '$fname' line $line: --post_update requires =<script>\n";
+		}
+
 		if ($type eq "include") {
 			parse_file($urls, $existing) or
 			    die "Unable to open included file '$urls'";
@@ -257,6 +261,10 @@ sub update_feed_via($$$$$$$$) {
 	if ($m->{'post_update'}) {
 		my $cmd = $m->{'post_update'};
 		system("cd '$localpath'; $cmd") == 0 or return 1;
+	}
+	if (defined(my $hook = $flags->{'post_update'})) {
+		system($hook, $name, $localpath) == 0 or
+			die "Feed '$name': --post_update hook '$hook' failed\n";
 	}
 
 	return 0;


scripts/feeds
```diff
@@ -258,6 +258,9 @@ sub update_feed_via($$$$$$$$) {
 		my $cmd = $m->{'post_update'};
 		system("cd '$localpath'; $cmd") == 0 or return 1;
 	}
+	if (my $hook = $flags->{'post_update'}) {
+		system("'$hook' '$name' '$localpath'") == 0 or return 1;
```
openwrt-ai: `$hook` is interpolated into a shell command inside single quotes without being escaped first, so a value containing `'` breaks out of the quoting and the remainder of the flag is executed as shell syntax (`--post_update=a'b;id;'c`). This function already escapes before interpolating elsewhere — [`$safename =~ s/'/'\\''/` at scripts/feeds:218](https://github.com/openwrt/openwrt/blob/7bb8c95c1b5e37e998db1b7fde28a7222628e6bd/scripts/feeds#L218).

The list form of `system()` avoids the shell entirely, still does a `PATH`/relative-path lookup via `execvp`, and passes `$name`/`$localpath` to the hook verbatim, so no quoting is needed at all:

```suggestion
		system($hook, $name, $localpath) == 0 or return 1;
```


---

scripts/feeds
```diff
@@ -258,6 +258,9 @@ sub update_feed_via($$$$$$$$) {
 		my $cmd = $m->{'post_update'};
 		system("cd '$localpath'; $cmd") == 0 or return 1;
 	}
+	if (my $hook = $flags->{'post_update'}) {
```
openwrt-ai: A valueless flag is stored as `1` by the parser ([`$flags{$1} = defined($2) ? $2 : 1;` at scripts/feeds:73](https://github.com/openwrt/openwrt/blob/7bb8c95c1b5e37e998db1b7fde28a7222628e6bd/scripts/feeds#L72-L73)), which the existing flags (`--force`, `--root` without a value) rely on. `--post_update` has no meaningful valueless form, but a bare `--post_update` in `feeds.conf` still passes this truthiness test and ends up executing a command literally named `1`, failing the feed update with an opaque `sh: 1: not found` and no mention of the feed or the flag.

The mirror case is `--post_update=0`, which is a syntactically valid `\S+` value but is false in Perl, so the hook is silently skipped.

Testing `defined` and rejecting the `1` case with an explicit `warn` naming the feed would make both diagnosable.

---

scripts/feeds
```diff
@@ -258,6 +258,9 @@ sub update_feed_via($$$$$$$$) {
...
+		system("'$hook' '$name' '$localpath'") == 0 or return 1;
+	}
```
openwrt-ai: Which path is the hook contractually handed when `--root` is also set? By the time control reaches here, [`$localpath` has been rewritten to `./feeds/<name>_root`](https://github.com/openwrt/openwrt/blob/7bb8c95c1b5e37e998db1b7fde28a7222628e6bd/scripts/feeds#L222-L227), i.e. the raw checkout, while `./feeds/<name>` is a symlink into the `--root` subdirectory and is the path the rest of the build (and `update_index`) actually consumes.

Handing the hook the raw checkout is defensible for the "apply a local patch series" use case in the commit message, but it is the opposite of what a hook author would guess, and the two paths differ only for `--root` feeds. Worth pinning down in the commit message (and ideally in `usage()`, which currently documents no `feeds.conf` flags at all) so downstream hooks can be written against a stable contract.

---

openwrt-ai (commented):
## Commit checks

- `7bb8c95` "scripts: feeds: support per-feed --post_update hook flag" — the example `feeds.conf` lines in the commit message (repeated in the PR body) put the flags *after* the feed name and URL:

  ```
  src-git  packages URL  --post_update=scripts/apply-overlay.sh
  src-cpy  core   ../core --force --post_update=scripts/stage-core.sh
  ```

  The feed-line parser only recognises flags *between* `src-<type>` and the feed name — [the regex at scripts/feeds:58](https://github.com/openwrt/openwrt/blob/7bb8c95c1b5e37e998db1b7fde28a7222628e6bd/scripts/feeds#L58) captures the flag run as group 2, immediately after `src-([\w\-]+)` and before `(\w+)`. Running both example lines through it, the flag group is empty and everything after the name lands in the URL group, which is then [split into `@src`](https://github.com/openwrt/openwrt/blob/7bb8c95c1b5e37e998db1b7fde28a7222628e6bd/scripts/feeds#L66):

  ```
  type=git flags=[] name=packages urls=[URL  --post_update=scripts/apply-overlay.sh]  parsed_flags={}
  type=cpy flags=[] name=core     urls=[../core --force --post_update=...]            parsed_flags={}
  ```

  So as documented the hook never fires, and `--force` / `--post_update=...` become additional feed *source URLs* that `update_feed` will loop over and try to clone/copy from. The commit message asserts "no schema or grammar change is required" because the name matches the parser's `\w+`, which is true — but only in the position the examples don't use. The working form is:

  ```
  src-git --post_update=scripts/apply-overlay.sh packages URL
  src-cpy --force --post_update=scripts/stage-core.sh core ../core
  ```

  Since these examples are the only documentation this flag ships with, they're worth fixing in the commit message before merge.

---

micpf:
Good catch — you're right, the flags must appear between `src-<type>` and the feed name. Amended the commit and updated the PR body accordingly (force-pushed to `micpf/openwrt:feeds-post-update-hook`). Thanks!

scripts/feeds
micpf: Fixed - switched to the list form of `system($hook, $name, $localpath)` so no shell is involved and no quoting is needed.

scripts/feeds
micpf: Fixed - added an explicit guard that `warn()`s naming the feed and fails the update when `--post_update` is used without a value (the parser stores `1` as the bare-flag sentinel), so it fails loudly instead of trying to exec a program literally called `1`.

scripts/feeds
micpf: Documented the contract in the commit message: the hook is handed the raw checkout (`./feeds/<name>_root` for `--root` feeds), not the `./feeds/<name>` consumer symlink - patching/overlaying the checkout is exactly the intended use case.

scripts/feeds
```diff
@@ -258,6 +258,13 @@ sub update_feed_via($$$$$$$$) {
 		my $cmd = $m->{'post_update'};
 		system("cd '$localpath'; $cmd") == 0 or return 1;
 	}
+	if (my $hook = $flags->{'post_update'}) {
+		if ($hook eq '1') {
+			warn "feed '$name': --post_update requires =<script>, ignoring bare flag\n";
```
openwrt-ai: The message says the bare flag is *ignored*, but the next line returns 1, which aborts the feed update — and the commit message documents it that way too ("fails the update"). A user reading "ignoring bare flag" followed by `failed.` gets contradictory information.

```suggestion
			warn "feed '$name': --post_update requires a value (=<script>)\n";
```

---

scripts/feeds
```diff
@@ -258,6 +258,13 @@ sub update_feed_via($$$$$$$$) {
 		my $cmd = $m->{'post_update'};
 		system("cd '$localpath'; $cmd") == 0 or return 1;
 	}
+	if (my $hook = $flags->{'post_update'}) {
```
openwrt-ai: The truthiness test still swallows one valid-looking value: `--post_update=0` matches the parser's `(?:=(\S+))?` and is stored as the string `0`, which is false in Perl, so the hook is silently skipped instead of run or diagnosed. Testing `defined` closes that hole (the `0` path then falls through to `system()` and fails loudly like any other unusable script path) and matches how the sibling flag is tested in this same function — [`if (defined $flags->{'root'})` at scripts/feeds:222](https://github.com/openwrt/openwrt/blob/ee20b60b9bd4a04cf9beb006000decf08fd9f0e8/scripts/feeds#L222).

```suggestion
	if (defined(my $hook = $flags->{'post_update'})) {
```

---

scripts/feeds
```diff
@@ -258,6 +258,13 @@ sub update_feed_via($$$$$$$$) {
...
+		if ($hook eq '1') {
+			warn "feed '$name': --post_update requires =<script>, ignoring bare flag\n";
+			return 1;
```
openwrt-ai: A non-zero return from `update_feed_via` means "this *source* failed" to the caller, not "this feed is misconfigured": [the loop at scripts/feeds:840-850](https://github.com/openwrt/openwrt/blob/ee20b60b9bd4a04cf9beb006000decf08fd9f0e8/scripts/feeds#L840-L850) moves on to the next URL in the feed's source list, and under `-f` prints `failed, ignore.` and keeps going. So for a feed line with several mirrors, a bare `--post_update` — a static `feeds.conf` error that no mirror can fix — makes the feed be updated again from every remaining source before the update finally fails, re-emitting the same warning each time. The same coupling applies to the `system()` call two lines down: a hook that legitimately exits non-zero triggers a fallback re-fetch from the next mirror even though the fetch itself succeeded.

For the bare-flag case at least, was rejecting it at parse time considered — a `die` in `parse_file` next to the existing [syntax / duplicate-name checks at scripts/feeds:60-64](https://github.com/openwrt/openwrt/blob/ee20b60b9bd4a04cf9beb006000decf08fd9f0e8/scripts/feeds#L60-L64) — so the config error is reported once, before any feed is touched?

---

openwrt-ai (commented):
Re-reviewed the force-pushed head (`7bb8c95` → `ee20b60`). The shell-injection issue is resolved by the list form of `system()`, the `--root` path contract is now documented in the commit message, and the `feeds.conf` examples now place the flags between `src-<type>` and the feed name, matching the parser. Three follow-ups on the new bare-flag guard inline.

---

scripts/feeds
```diff
@@ -74,6 +74,10 @@ sub parse_file($$) {
 			}
 		}
 
+		if (exists $flags{'post_update'} && $flags{'post_update'} eq '1') {
```
openwrt-ai: nit: this test cannot distinguish the bare-flag sentinel from a literal value of `1`, because the parser collapses both to the same scalar — [`$flags{$1} = defined($2) ? $2 : 1;` at scripts/feeds:72-73](https://github.com/openwrt/openwrt/blob/6fb33444be884c3e400104ffc412bdcbb5b1a988/scripts/feeds#L72-L73). So `--post_update=1` (a script path literally named `1`, e.g. a generated hook) is rejected at parse time with "requires =&lt;script&gt;", which is the one thing the user did supply.

It's a far-fetched value, but it's the mirror image of the `--post_update=0` case the commit message goes out of its way to keep working, so the asymmetry is worth a thought. Matching the raw flag run instead of the parsed value (`$flags =~ /(?:^|\s)--post_update(?:\s|$)/`) keys off the actual absence of `=` rather than a sentinel that overlaps the value space.

---

scripts/feeds
```diff
@@ -258,6 +262,9 @@ sub update_feed_via($$$$$$$$) {
 		my $cmd = $m->{'post_update'};
 		system("cd '$localpath'; $cmd") == 0 or return 1;
 	}
+	if (defined(my $hook = $flags->{'post_update'})) {
+		system($hook, $name, $localpath) == 0 or return 1;
```
openwrt-ai: The parse-time `die` fixes the bare-flag half of the mirror-loop coupling, but this half is still open: a hook that legitimately exits non-zero (patch series doesn't apply, SBOM generation fails) makes `update_feed_via` return 1, which the caller reads as "*this source* failed" and retries the whole fetch from the next URL in the feed's source list — [the loop at scripts/feeds:840-852](https://github.com/openwrt/openwrt/blob/6fb33444be884c3e400104ffc412bdcbb5b1a988/scripts/feeds#L840-L852). For a multi-mirror feed the checkout is re-cloned and the same hook re-run once per remaining mirror before the update finally fails; under `-f` each round prints `failed, ignore.` and the run is reported as a success.

The commit message states "A non-zero exit fails the enclosing update" without qualification, which holds for single-URL feeds but understates what happens with mirrors. Was distinguishing the two failure modes considered — e.g. `die`ing here (the fetch did succeed; no other mirror can change the hook's outcome), the same way the bare-flag case is now handled outside the retry loop?

---

openwrt-ai (commented):
Re-reviewed the force-pushed head (`ee20b60` → `6fb3344`). The bare-flag guard has moved to `parse_file`, so the invalid-config case is now reported once, before any feed is touched, with the feed name, file and line number — that addresses the "re-evaluated once per mirror" half of the earlier finding, and the `defined()` test closes the `--post_update=0` hole. The commit message matches the diff, including the `scripts/feeds:58` regex reference and the parse-time-`die` rationale.

Two follow-ups inline: the leftover hook-failure/mirror-retry coupling, and a nit about the bare-flag sentinel overlapping a legal value.

---

PR #24455, opened 2026-07-28
ath79: disable building UniFi AP series devices

DragonBluep:
After switching to the 6.18 kernel, they are no longer built due to 3 MiB kernel size limitation.

### DIFF
--- a/target/linux/ath79/image/generic-ubnt.mk
+++ b/target/linux/ath79/image/generic-ubnt.mk
@@ -329,6 +329,7 @@ define Device/ubnt_unifi-ap-outdoor-plus
   $(Device/ubnt-unifi-jffs2)
   DEVICE_MODEL := UniFi AP Outdoor+
   SUPPORTED_DEVICES += unifi-outdoor-plus
+  DEFAULT := n
 endef
 TARGET_DEVICES += ubnt_unifi-ap-outdoor-plus
 
@@ -338,5 +339,6 @@ define Device/ubnt_unifi-ap-pro
   DEVICE_MODEL := UniFi AP Pro
   UBNT_CHIP := ar934x
   SUPPORTED_DEVICES += uap-pro
+  DEFAULT := n
 endef
 TARGET_DEVICES += ubnt_unifi-ap-pro


target/linux/ath79/image/generic-ubnt.mk
```diff
@@ -329,6 +329,7 @@ define Device/ubnt_unifi-ap-outdoor-plus
   $(Device/ubnt-unifi-jffs2)
   DEVICE_MODEL := UniFi AP Outdoor+
   SUPPORTED_DEVICES += unifi-outdoor-plus
+  DEFAULT := n
```
openwrt-ai, line 332: nit: the subject says "UniFi AP series devices", but only the two devices inheriting `Device/ubnt-unifi-jffs2` (the template that carries `KERNEL_SIZE := 3072k`) are disabled. [`ubnt_unifi-ap` and `ubnt_unifi-ap-lr`](https://github.com/openwrt/openwrt/blob/826d61bbb51eaf2a0c8619c6a7cd81c98ea0b681/target/linux/ath79/image/generic-ubnt.mk#L273-L286) are also "UniFi AP" but stay enabled — they inherit `Device/ubnt-bz`, which sets no `KERNEL_SIZE`, so the 3 MiB check does not apply to them. Naming the two affected models in the subject (e.g. "ath79: disable building UniFi AP Outdoor+ and AP Pro") would avoid the ambiguity about whether the plain AP/AP LR were intentionally left alone.

---

target/linux/ath79/image/generic-ubnt.mk
```diff
@@ -338,5 +339,6 @@ define Device/ubnt_unifi-ap-pro
   DEVICE_MODEL := UniFi AP Pro
   UBNT_CHIP := ar934x
   SUPPORTED_DEVICES += uap-pro
+  DEFAULT := n
```
openwrt-ai, line 342: nit: both disabled devices are exactly the two users of [`Device/ubnt-unifi-jffs2`](https://github.com/openwrt/openwrt/blob/826d61bbb51eaf2a0c8619c6a7cd81c98ea0b681/target/linux/ath79/image/common-ubnt.mk#L131-L141), which is where the `KERNEL_SIZE := 3072k` that causes the failure lives. Putting `DEFAULT := n` in that shared template instead of duplicating it here (and at line 332) would keep the cause and the workaround together, and any future device based on the same jffs2 layout would inherit it. There is precedent for `DEFAULT := n` in a shared template in this same directory: [`Device/tplink-4m` in common-tp-link.mk:38-43](https://github.com/openwrt/openwrt/blob/826d61bbb51eaf2a0c8619c6a7cd81c98ea0b681/target/linux/ath79/image/common-tp-link.mk#L38-L43). Fine to keep it per-device if you prefer the explicitness.

---

openwrt-ai (commented):
Reviewed 1 new commit. No functional issues found — `DEFAULT := n` is the correct in-tree mechanism here (image/kernel size limitation rather than a broken device), matching `504000d520` ("treewide: use DEFAULT := n to disable non-broken devices") vs. `e9f7923a1d` (BROKEN for devices that don't boot/compile), and the same pattern as `60061b5c26` (ath79 EnGenius) and `9ec3b11d37` (lantiq P-2812HNU F1, also a 3072k `KERNEL_SIZE`). The two devices touched are precisely the ones inheriting the 3 MiB `KERNEL_SIZE` from `Device/ubnt-unifi-jffs2`. Two nits inline.

---

openwrt-ai (commented):

The force-push reworded the subject to name the two affected models (addressing the earlier nit) and rebased onto current `main`; the `generic-ubnt.mk` hunk itself is byte-identical to what I reviewed at `826d61b`. `ubnt_unifi-ap-outdoor-plus` and `ubnt_unifi-ap-pro` are still the only two users of `Device/ubnt-unifi-jffs2`, and ath79 is still on `KERNEL_PATCHVER:=6.18`, so the commit body's rationale continues to hold. My earlier note about hoisting `DEFAULT := n` into the shared template is still open but remains optional — not repeating it inline.

---

lorand-horvath:
Can this be merged, please?

jonasjelonek:

PR #24457, opened 2026-07-28
qualcommax: qca_ppe: fix the port scheduler TDM enqueue interlock

JuliusBairaktaris:
Each port scheduler TDM tick carries a dequeue port, an enqueue port and an enqueue port bitmap. The bitmap must permit every port **except** the one dequeued on this tick and the one dequeued on the previous tick, so an enqueue cannot touch a port's scheduler state while a dequeue is still working on it. The comment above the loop states that rule correctly.

`prev_de_port` is seeded as a **port number** and read back as a port number, but assigned as a **mask**:

```c
prev_de_port = psch[psch_num - 1].de_port;              /* port number */
for (i = 0; i < psch_num; i++) {
        u8 bmp = ~(BIT(prev_de_port) | BIT(psch[i].de_port));   /* read as a port number */
        ...
        prev_de_port = BIT(psch[i].de_port);            /* <-- assigned as a MASK */
}
```

So every tick after the first computes `BIT(BIT(de_port))`.

### Worked example, IPQ8074

Tick 0 dequeues port 0, tick 1 dequeues port 5, tick 11 dequeues port 6, tick 12 dequeues port 2.

**Tick 1** — `prev_de_port` was set to `BIT(0)` = 1 instead of 0:

```
correct   ~(BIT(0) | BIT(5)) = ~0x21 = 0xDE   blocks ports 0 and 5
computed  ~(BIT(1) | BIT(5)) = ~0x22 = 0xDD   blocks ports 1 and 5
```

Port 0 was dequeued on the previous tick and is now permitted; port 1 is blocked for no reason.

**Tick 12** — `prev_de_port` was set to `BIT(6)` = 64, so the shift is out of range (on arm64 `1UL << 64` shifts by `64 & 63` = 0):

```
correct   ~(BIT(6) | BIT(2)) = ~0x44 = 0xBB   blocks ports 2 and 6
computed  ~(BIT(0) | BIT(2)) = ~0x05 = 0xFA   blocks ports 0 and 2
```

Both computed values are what the hardware actually reads back. Across the table, **49 of the 50 ticks permit enqueue on the previous tick's dequeue port**. Only tick 0 is right, because it uses the seed from before the loop.

### Why this is the right value

qca-ssdk does not compute the bitmap at all — it ships it per tick as a constant in `port_scheduler0_tbl` (IPQ807x) and `cppe_port_scheduler0_tbl` (IPQ60xx) and writes it straight to `PSCH_TDM_CFG_TBL.ENS_PORT_BITMAP`. With the assignment fixed, the loop reproduces both tables exactly: 50/50 ticks on IPQ8074 and 50/50 on IPQ6018.

Verified on hardware, AX3600 and DL-WRX36: `ens`/`des` already matched ssdk on all 50 ticks; the bitmaps matched on 1 of 50 before the change and 50 of 50 after.

@Ansuel @robimarko

### DIFF
--- a/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_scheduler.c
+++ b/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_scheduler.c
@@ -381,7 +381,7 @@ static void ppe_tdm_init(struct qca_ppe_priv *priv)
 			     FIELD_PREP(PPE_PSCH_ENS_PORT, psch[i].en_port) |
 			     FIELD_PREP(PPE_PSCH_DES_PORT, psch[i].de_port));
 
-		prev_de_port = BIT(psch[i].de_port);
+		prev_de_port = psch[i].de_port;
 	}
 
 	regmap_write(priv->regmap, PPE_TM_TDM_DEPTH,


target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_scheduler.c
```diff
@@ -381,7 +381,7 @@ static void ppe_tdm_init(struct qca_ppe_priv *priv)
 			     FIELD_PREP(PPE_PSCH_ENS_PORT, psch[i].en_port) |
 			     FIELD_PREP(PPE_PSCH_DES_PORT, psch[i].de_port));
 
-		prev_de_port = BIT(psch[i].de_port);
+		prev_de_port = psch[i].de_port;
```
openwrt-ai, line 384: nit: no `Fixes:` tag. This line came in with the only commit that ever touched the file, [142104bb90b3 ("qualcommax: add PPE driver")](https://github.com/openwrt/openwrt/commit/142104bb90b3d7685b2a96f04f29b76cbd1e5ee9), so the tag is unambiguous and roughly a third of recent commits on main carry one:

```
Fixes: 142104bb90b3 ("qualcommax: add PPE driver")
```


---

openwrt-ai, line 384: nit: the commit message says `BIT(BIT(de_port))` is "undefined once de_port reaches 5", but the PR body's own worked example puts the undefined shift at tick 11 (`de_port` 6, `1UL << 64`). With [`#define BIT(nr) (UL(1) << (nr))`](https://github.com/gregkh/linux/blob/0b8f247169e487eff2d4c2dd531bc43f7efda2cb/include/vdso/bits.h#L7) and 64-bit `UL`, `de_port == 5` gives `1UL << 32`, which is well defined — it is simply wrong, because the bit lands outside the `u8 bmp` so the previous port ends up unmasked. Undefined behaviour starts at `de_port == 6` (`1UL << 64`) and `de_port == 7` (`1UL << 128`), both of which occur in `hppe_psch_tdm[]`. Worth aligning the commit message with the body so the two don't disagree on which tick is UB.

---

openwrt-ai (commented):
Reviewed 1 new commit. The fix itself checks out: `prev_de_port` is declared `u8`, seeded from `psch[psch_num - 1].de_port` as a port number and consumed as `BIT(prev_de_port)`, so carrying a mask across the iteration was the odd one out. The worked example in the PR body matches `hppe_psch_tdm[]` in-tree (tick 0 dequeues 0, tick 1 dequeues 5, tick 11 dequeues 6, tick 12 dequeues 2), and `PPE_PSCH_ENS_PORT_BMP` is `GENMASK(15, 8)`, so the `u8 bmp` width is right. Two commit-message nits inline.

---

openwrt-ai (commented):

The force-push is commit-message only — `git diff 68f8f90..d150518` is empty, the `prev_de_port = psch[i].de_port;` hunk is unchanged from the previously reviewed revision. Both earlier nits are resolved:

- `Fixes: 142104bb90b3 ("qualcommax: add PPE driver")` added; the abbreviated SHA and subject match the in-tree commit.
- The UB claim now reads "from de_port 3 upwards the bit lands outside the u8 bitmap, and from de_port 6 upwards the shift itself is undefined", which agrees with the PR body and with the enum in [`qca_ppe_scheduler.c:15-24`](``https://github.com/openwrt/openwrt/blob/d150518cfb476b612e1c2fed74c941a6eeb7f80e/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_scheduler.c#L15-L24)`` (`TDM_PORT_PHY_3` = 3 → `BIT(8)` escapes the `u8`; `TDM_PORT_FAB_1` = 6 → `1UL << 64`). All of de_port 3, 6 and 7 do occur in `hppe_psch_tdm[]`, so "both occur in hppe_psch_tdm[]" holds.

---

openwrt-ai (commented):

`50fc95e` is a rebase of `d150518` onto current `main` (parent moved `9a928a4` → `ce16dd8`) plus one trailer reorder. The patch is byte-identical — `git diff` of the two commits' own diffs is empty, so the single `prev_de_port = psch[i].de_port;` hunk in [`qca_ppe_scheduler.c:385`](``https://github.com/openwrt/openwrt/blob/50fc95ed1541dcd6d24c78e8bfef25011a430047/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_scheduler.c#L385)`` is unchanged from the previously reviewed revision.

The only message change is `Assisted-by:` moving above `Signed-off-by:`, which puts the author's sign-off last as the kernel trailer convention expects. `Fixes: 142104bb90b3 ("qualcommax: add PPE driver")` is unchanged and still matches the in-tree commit.

CI on the head SHA is green — all three qualcommax subtargets (ipq50xx, ipq60xx, ipq807x) built successfully, and all three FormalityCheck runs pass.

---

robimarko:

PR #24458, opened 2026-07-28
mac80211: mwifiex: ignore ROM regulatory hint on WRT3200ACM/WRT32x

### LINKED ISSUE #9956: WRT3200ACM 5 GHz DFS not working (and workaround)
I have a WRT3200ACM (bought in the UK), and have found that the DFS channels in the 5 GHz band do not work out of the box. There is a workaround, which I'll talk about below.

## Basic information:
 - OpenWRT version: 22.03.0-rc3
 - Router: WRT3200ACM (bought in the UK)

## Steps to reproduce

1. Install OpenWRT version 22.03.0-rc3 (or probably also any other version) on the router.
2. Set up a 5 GHz access point on channel 64 (or presumably any other DFS channel).

## Expected behaviour

Wifi clients should be able to connect to the created access point.

## Actual behaviour

The access point fails to initialize. The following is printed in the syslog:

```
Sat May 28 07:30:04 2022 [kern.info](http://kern.info/) kernel: [ 3068.774183] device wlan0 entered promiscuous mode
Sat May 28 07:30:04 2022 daemon.notice hostapd: wlan0: interface state UNINITIALIZED->COUNTRY_UPDATE
Sat May 28 07:30:05 2022 daemon.notice netifd: Wireless device 'radio0' is now up
Sat May 28 07:30:09 2022 daemon.notice hostapd: wlan0: interface state COUNTRY_UPDATE->HT_SCAN
Sat May 28 07:30:09 2022 kern.debug kernel: [ 3073.781491] ieee80211 phy0: change: 0x100
Sat May 28 07:30:09 2022 kern.debug kernel: [ 3073.793910] ieee80211 phy0: change: 0x60
Sat May 28 07:30:09 2022 kern.debug kernel: [ 3074.008811] ieee80211 phy0: change: 0x40
Sat May 28 07:30:10 2022 daemon.notice hostapd: wlan0: interface state HT_SCAN->DFS
Sat May 28 07:30:10 2022 daemon.notice hostapd: wlan0: DFS-CAC-START freq=5320 chan=64 sec_chan=-1, width=1, seg0=58, seg1=0, cac_time=60s
Sat May 28 07:30:10 2022 daemon.err hostapd: DFS start_dfs_cac() failed, -1
Sat May 28 07:30:10 2022 daemon.err hostapd: Interface initialization failed
Sat May 28 07:30:10 2022 daemon.notice hostapd: wlan0: interface state DFS->DISABLED
Sat May 28 07:30:10 2022 daemon.notice hostapd: wlan0: AP-DISABLED
```

## Workaround

Run `opkg remove kmod-mwifiex-sdio` (thanks to a friend for helping me find this). This fixed the problem, and also (though I haven't tested extensively) seemed to make the wifi more stable/reliable in general.

This is a [known workaround](https://github.com/kaloz/mwlwifi/issues/173#issuecomment-307984048). I'm not sure what this package is intended to do or of the implications of removing it. But I wonder whether it would be sensible to remove this package from the image for this router (and perhaps other affected routers), or blacklist the kernel module so it's not loaded unless explicitly requested. This would make the DFS channels work out of the box.

Per [this comment](https://github.com/kaloz/mwlwifi/issues/173#issuecomment-308209114), I wonder whether that also makes the `mwifiex-sdio-firmware` package redundant (I haven't investigated whether any alternative kernel module uses that).

Thoughts? Thanks!

httpstorm:
WRT3200ACM and WRT32x routers always report a fixed US regulatory domain from the `mwifiex` ROM, which conflicts with the platform regulatory configuration on units sold outside the US market. For example:
- `phy0` `mwlwifi` 5 GHz, hard-coded region 98 EU mapped to FR
- `phy1` `mwlwifi` 2.4 GHz, hard-coded region 98 EU mapped to FR
- `phy2` `mwifiex` 2.4 and 5 GHz, hard-coded region US

When the system boots, it detects a conflict between the user selected region and the radios, e.g. BG FR US, and applies extreme restrictions, preventing `phy0` from starting on any DFS channel. `phy2` works.

1. Add a device-tree flag `marvell,invalid-reg-hint-in-rom`, which allows affected devices, where the radio conflicts with the platform to ignore the incorrect regulatory hint in the ROM of the radio, so the platform regulatory domain can be used. This does not allow the user to change the regulatory configuration of any radio. It only allows `phy0` to start and operate according to its certified configuration.
2. Add node `wifi@1` with a flag `marvell,invalid-reg-hint-in-rom` to the device-trees for WRT3200ACM and WRT32x which signals `mwifiex` that its regulatory hint from ROM conflicts with the platform and should be ignored, so the platform regulatory domain can be used.

This way userspace can set the correct platform regulatory domain. These radios are locked to the hard-coded configuration for which they are certified, so the only impact is that the region conflict is avoided and the 5 GHz radios can be used.

The change affects only WRT3200ACM and WRT32x routers, and helps overcome the incorrect behaviour of proprietary radio firmware.

On WRT3200ACM, the platform configuration is stored on the `devinfo` `/dev/mtd3` partition, which is read-only:
``` bash
cert_region=EU
modelNumber=WRT3200ACM-EU
```

[1] https://github.com/kaloz/mwlwifi/issues/173#issuecomment-307879699
Fixes: #9956

Tested-on: WRT3200ACM, OpenWrt

Patch submitted upstream to the Linux team.
Please let me know if there is a better approach to address this.

_Future work_
I also have a similar patch for the main radio `mwlwifi`. If you like I can clean it up and put it in a proper shape for another PR, and add checks for the affected routers. Without it, users in Europe are forced to select FR as their country, which is incorrect. Using other countries may still create conflicts. If we accept the fix, users will be allowed to select their correct country. The behaviour of the radios will not be affected, because these radios are locked to the hard-coded configuration for which they are certified.
[709-mwlwifi-2020-02-06-DFS-free.patch](https://github.com/httpstorm/openwrt/blob/2026-07-27/package/kernel/mwlwifi/patches/709-mwlwifi-2020-02-06-DFS-free.patch)

@jonasjelonek @hauke @nbd168 @neheb @brada4
@Techmeology @GeorgelT @hnyman


package/kernel/mac80211/patches/mwl/112-wifi-mwifiex-ignore-ROM-regulatory-hint-on-WRT3200AC.patch
```diff
@@ -0,0 +1,53 @@
...
+ 						adapter->region_code);
++
++				/* WRT3200ACM/WRT32x ROM always reports US, which
++				 * conflicts with routers sold outside the US.
++				 * Ignore this hint so userspace can
++				 * set the correct regulatory domain.
++				 */
+ 				if (country_code &&
++				    !of_machine_is_compatible("linksys,wrt3200acm") &&
++				    !of_machine_is_compatible("linksys,wrt32x") &&
```
openwrt-ai: Both compatible strings matched here are OpenWrt-local, not upstream. Upstream v6.18.39 has [`compatible = "linksys,rango", ...` in armada-385-linksys-rango.dts:15](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/arch/arm/boot/dts/marvell/armada-385-linksys-rango.dts#L15) — no `linksys,wrt3200acm` — and there is no WRT32X/venom DTS upstream at all. `linksys,wrt3200acm` only exists in this tree because of [311-adjust-compatible-for-linksys.patch](``https://github.com/openwrt/openwrt/blob/bd9352d0e5494dcf58421e0b07c64e537fabbcff/target/linux/mvebu/patches-6.18/311-adjust-compatible-for-linksys.patch#L9),`` and `linksys,wrt32x` because [venom.dts is an OpenWrt-only file](``https://github.com/openwrt/openwrt/blob/bd9352d0e5494dcf58421e0b07c64e537fabbcff/target/linux/mvebu/files/arch/arm/boot/dts/marvell/armada-385-linksys-venom.dts#L47).`` So the version you sent upstream is a silent no-op there.

`of_machine_is_compatible()` matches any entry of the root `compatible` list, and both OpenWrt DTs still carry `linksys,rango` / `linksys,venom`, so keying on the board code names works in both trees:

```suggestion
+				    !of_machine_is_compatible("linksys,rango") &&
+				    !of_machine_is_compatible("linksys,venom") &&
```

(the comment above would then want a note that rango = WRT3200ACM and venom = WRT32X.)

---

package/kernel/mac80211/patches/mwl/112-wifi-mwifiex-ignore-ROM-regulatory-hint-on-WRT3200AC.patch
```diff
@@ -0,0 +1,53 @@
...
+ 1 file changed, 8 insertions(+)
+
+--- a/drivers/net/wireless/marvell/mwifiex/cfg80211.c
++++ b/drivers/net/wireless/marvell/mwifiex/cfg80211.c
+@@ -4892,7 +4892,15 @@ int mwifiex_register_cfg80211(struct mwi
+ 				country_code =
+ 					mwifiex_11d_code_2_region(
+ 						adapter->region_code);
++
++				/* WRT3200ACM/WRT32x ROM always reports US, which
```
openwrt-ai: You asked whether there is a better approach — one worth considering, since hard-coding a board list inside a generic driver is what will get pushback upstream: the same `if (!adapter->regd)` block already gates on the [`reg_alpha2` module parameter](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/net/wireless/marvell/mwifiex/cfg80211.c#L4879-L4883). Extending that knob (e.g. a sentinel value, or a small `reg_ignore_rom`/`no_regd_hint` module param) would let OpenWrt set it per-board from a modprobe conf under `target/linux/mvebu/`, keeps the quirk data out of the driver, and works for any other board with the same lying ROM. A DT property on the mwifiex node would be a second option. Not a blocker for an OpenWrt-local patch, just likely a smoother path for the upstream submission.

---

openwrt-ai (commented):
Reviewed 1 new commit. Patch metadata checks out: the hunk header `@@ -4892,7 +4892,15 @@` matches the real line numbers in `drivers/net/wireless/marvell/mwifiex/cfg80211.c` for the 6.18.39 sources this package tracks, the diffstat matches the hunk, `<linux/of.h>` is already pulled in via `main.h`, and the `!CONFIG_OF` stub returns 0 so non-DT platforms keep the old behaviour. `PKG_RELEASE` bump is present. Two inline notes, the first one is a real functional gap for the upstream version of this patch.

---

httpstorm:
v2:
Replace the compatible device names, which are currently only available on OpenWrt with board names which are available on both Linux and OpenWrt.

*Idea for an alternative implementation:*
I implemented and tested a better approach with no board specific code in the driver:
- check for a device-tree flag in `mwifiex_register_cfg80211in()`
  `!of_property_read_bool(of_root, "marvell,ignore-regulatory-hint")`
- add this flag to the root device tree of each affected board:

``` diff
 	model = "Linksys WRT3200ACM";
 	compatible = "linksys,wrt3200acm", "linksys,rango", "linksys,armada385",
 		     "marvell,armada385", "marvell,armada380";
+	marvell,ignore-regulatory-hint;
```

This approach is more flexible and scalable, because the flag can also be used in other drivers, e.g. `mwlwifi`.
Any help is welcome! Let's wait for feedback from the Linux team.

package/kernel/mac80211/patches/mwl/112-wifi-mwifiex-ignore-ROM-regulatory-hint-on-WRT3200AC.patch
httpstorm: Replaced the compatible device names, which are currently only available on OpenWrt with board names which are available on both Linux and OpenWrt.

package/kernel/mac80211/patches/mwl/112-wifi-mwifiex-ignore-ROM-regulatory-hint-on-WRT3200AC.patch
```diff
@@ -0,0 +1,53 @@
...
++				 * conflicts with routers sold outside the US.
++				 * Ignore this hint so userspace can
++				 * set the correct regulatory domain.
++				 */
```
openwrt-ai: nit: the comment names the marketing models but the code below matches the board code names, so nothing in this hunk tells a reader that `linksys,rango` is the WRT3200ACM and `linksys,venom` the WRT32X. Worth spelling out, since a `git grep rango` in the driver directory is the only other way to find out. Kept to the same 4 comment lines so the `@@ -4892,7 +4892,15 @@` hunk count stays valid:

```suggestion
+				/* WRT3200ACM (rango) / WRT32X (venom) ROMs
+				 * always report US, which conflicts with
+				 * routers sold outside the US. Ignore the
+				 * hint so userspace can set the right domain.
+				 */
```

Separately, for the upstream submission: `linksys,rango` now resolves in mainline ([armada-385-linksys-rango.dts:15](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/arch/arm/boot/dts/marvell/armada-385-linksys-rango.dts#L15)), but there is still no venom/WRT32X DTS in mainline at all, so the second check stays a no-op there until one lands. Harmless, just don't expect the WRT32X half to do anything for mainline users.

---

openwrt-ai (commented):
Reviewed the v2 force-push (`bd9352d` → `21bf07b`). The only code change is `linksys,wrt3200acm`/`linksys,wrt32x` → `linksys,rango`/`linksys,venom`, which resolves my earlier note — both code names are present in the root `compatible` of the OpenWrt DTs (`linksys,rango` via [311-adjust-compatible-for-linksys.patch](``https://github.com/openwrt/openwrt/blob/21bf07bb6ac4dcdbb4dffd3fea3a2f2865c9504e/target/linux/mvebu/patches-6.18/311-adjust-compatible-for-linksys.patch#L9),`` `linksys,venom` in [armada-385-linksys-venom.dts:47](``https://github.com/openwrt/openwrt/blob/21bf07bb6ac4dcdbb4dffd3fea3a2f2865c9504e/target/linux/mvebu/files/arch/arm/boot/dts/marvell/armada-385-linksys-venom.dts#L47)),`` and `linksys,rango` is also upstream in both 6.12 and 6.18, so the mvebu `KERNEL_PATCHVER:=6.18` build and any future rebase both match.

Re-verified the rest still holds after the rebase: hunk header `@@ -4892,7 +4892,15 @@` matches the real line numbers in the 6.18.39 sources this package tracks and the +8 line count is still correct, `<linux/of.h>` comes in via `main.h`, the `!CONFIG_OF` stub returns 0 so non-DT platforms keep calling `regulatory_hint()`, patch number 112 follows 110/111 in `patches/mwl/` with no collision, and `PKG_RELEASE` is bumped to 4. Commit message matches the diff. CI on `21bf07b` is green, including FormalityCheck and the mvebu kernel-patch job.

One nit inline; the DT-property alternative you sketched in the PR comment looks like the cleaner long-term shape, so I won't re-litigate the module-param suggestion from the previous round.

---

httpstorm:
Changes in v3:
1. Implement a device-tree flag `marvell,ignore-regulatory-hint`, which allows affected platforms to ignore an incorrect regulatory hint in the ROM.
2. Add `marvell,ignore-regulatory-hint` to the device-trees for WRT3200ACM and WRT32x which signals `mwifiex` to ignore an incorrect regulatory hint in the ROM.

package/kernel/mac80211/patches/mwl/112-wifi-mwifiex-ignore-ROM-regulatory-hint-on-WRT3200AC.patch
```diff
@@ -0,0 +1,53 @@
...
+ 					mwifiex_11d_code_2_region(
+ 						adapter->region_code);
++
++				/* Allow platforms to ignore an incorrect
++				 * regulatory hint from ROM, so userspace
++				 * can set the correct regulatory domain.
++				 */
+ 				if (country_code &&
++				    !of_property_read_bool(of_root,
++							   "marvell,ignore-regulatory-hint") &&
```
openwrt-ai, line 50: `of_root` has no definition when `CONFIG_OF=n`. It is declared at [`include/linux/of.h:141`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/include/linux/of.h#L141), *outside* the `#ifdef CONFIG_OF` block that starts at line 159, but it is only defined and exported in `drivers/of/base.c`, which is built via [`obj-$(CONFIG_OF) += of/` (drivers/Makefile:145)](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/Makefile#L145). Note that the kernel's own [`of_have_populated_dt()`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/include/linux/of.h#L1716-L1723) reads the same global and *still* wraps it in `#ifdef CONFIG_OF`, even though `of_property_present()` already has a `!CONFIG_OF` stub — so the stub alone was not considered sufficient there.

mwifiex is built for non-DT platforms as well (x86 PCIe/USB), and the `Build all core packages` CI job only builds `malta/be`, where `CONFIG_OF=y`, so a `CONFIG_OF=n` build is not covered by CI here. In practice `-O2` will dead-code-eliminate the load of `of_root` before the linker sees it (the stub ignores its argument), so this may well build everywhere — but it relies on the optimizer instead of the guard the header itself uses. Making it explicit removes the doubt:

```c
if (country_code &&
    !(IS_ENABLED(CONFIG_OF) &&
      of_property_read_bool(of_root, "marvell,ignore-regulatory-hint")) &&
    regulatory_hint(wiphy, country_code))
```

No suggestion block, because that adds a line: the hunk header on line 39 would have to become `@@ -4892,7 +4892,15 @@` and the diffstat on lines 34–35 `8 insertions(+)`.

Separately, for the upstream submission: reading a driver quirk off a *root-node* vendor property has no precedent in mainline — `grep -rn 'of_property_read_bool(of_root\|of_property_present(of_root' drivers/ arch/` in v6.18.41 returns nothing outside `of.h` itself, and there is no binding documentation for `marvell,ignore-regulatory-hint`. The established pattern for board quirks is `of_machine_is_compatible()` (e.g. [ath11k core.c:1046](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/ath/ath11k/core.c#L1046)), and mwifiex already has a per-device DT path via [`adapter->dt_node` in `mwifiex_probe_of()`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/marvell/mwifiex/main.c#L1640-L1649) with a documented binding (`Documentation/devicetree/bindings/net/wireless/marvell,sd8787.yaml`). The catch for the device-node route on these boards is that [`mwifiex_sdio_of_match_table`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/marvell/mwifiex/sdio.c#L518-L525) has no `marvell,sd8887` entry, so an SDIO child node would currently fail `mwifiex_sdio_probe_of()`. Just flagging it so you can pre-empt the question on the list.

---

package/kernel/mac80211/patches/mwl/112-wifi-mwifiex-ignore-ROM-regulatory-hint-on-WRT3200AC.patch
```diff
@@ -0,0 +1,53 @@
+From 6855207289334f4cb72ececc0ec76741abeda3af Mon Sep 17 00:00:00 2001
+From: Georgi Valkov <gvalkov@gmail.com>
+Date: Fri, 31 Jul 2026 09:21:00 +0300
+Subject: [PATCH] wifi: mwifiex: ignore ROM regulatory hint on
+ WRT3200ACM/WRT32x
```
openwrt-ai, line 5: nit: since v3 the driver change is board-agnostic — it adds a generic DT flag and no longer mentions WRT3200ACM or WRT32x anywhere in the code — but the subject still scopes it to those two boards. The body already says it correctly ("The change affects only platforms which explicitly define this flag"). Same applies to the OpenWrt commit subject of 9b1e7e0.

```suggestion
Subject: [PATCH] wifi: mwifiex: allow platforms to ignore the ROM
 regulatory hint
```

---

package/kernel/mac80211/patches/mwl/112-wifi-mwifiex-ignore-ROM-regulatory-hint-on-WRT3200AC.patch
```diff
@@ -0,0 +1,53 @@
...
+platform regulatory domain. The radios in WRT3200ACM/WRT32x are locked to
+the hard-coded configuration for which they are certified, so the only
+impact is that the region conflict is avoided and the 5 GHz radios can
+be used.
+
+The change affects only platforms which explicitly define this flag, and
+helps overcome the incorrect behaviour of proprietary radio firmware.
+
+[1] https://github.com/kaloz/mwlwifi/issues/173#issuecomment-307879699
+[2] https://github.com/openwrt/openwrt/issues/9956
```
openwrt-ai, line 29: nit: the patch files reference the issue correctly here as `[2] https://github.com/openwrt/openwrt/issues/9956`, but both OpenWrt commit messages (9b1e7e0 and 716158b) still say `Fixes: #9956` instead. `Fixes:` in this tree means a commit reference — the last 400 commits on `main` only use it as `Fixes: <sha> ("<subject>")`, e.g. `Fixes: ce55cdde26a9 ("airoha: an7583: move the switch port label ...")`. Issue/PR references go through `Link:` or `Closes:` with a full URL. Worth changing the two commit messages to `Link: https://github.com/openwrt/openwrt/issues/9956` so they match the patch files.

---

target/linux/mvebu/patches-6.18/990-ignore-regulatory-hint.patch
```diff
@@ -0,0 +1,47 @@
...
+
+The Linux kernel sources do not have support for WRT32x. An equivalent
+patch for these routers has to be sent to OpenWrt.
+
+[1] https://github.com/kaloz/mwlwifi/issues/173#issuecomment-307879699
+[2] https://github.com/openwrt/openwrt/issues/9956
+
+Tested-on: WRT3200ACM, OpenWrt
+Cc: stable@vger.kernel.org
+Signed-off-by: Georgi Valkov <gvalkov@gmail.com>
```
openwrt-ai, line 33: `Cc: stable@vger.kernel.org` on an OpenWrt-local patch is misleading: this change is not upstream yet, and stable only takes commits that are already in Linus' tree (and this is a new quirk/feature rather than a regression fix, so it would not qualify anyway). No other patch in `target/linux/mvebu/patches-6.18/` carries a stable Cc. Drop it here and add it to the mail you send to netdev/linux-wireless instead if you want it considered.

```suggestion
Signed-off-by: Georgi Valkov <gvalkov@gmail.com>
```

---

target/linux/mvebu/patches-6.18/990-ignore-regulatory-hint.patch
```diff
@@ -0,0 +1,47 @@
...
+---
+ arch/arm/boot/dts/marvell/armada-385-linksys-rango.dts | 1 +
+ 1 file changed, 1 insertion(+)
+
+--- a/arch/arm/boot/dts/marvell/armada-385-linksys-rango.dts
++++ b/arch/arm/boot/dts/marvell/armada-385-linksys-rango.dts
+@@ -14,6 +14,7 @@
+ 	model = "Linksys WRT3200ACM";
+ 	compatible = "linksys,wrt3200acm", "linksys,rango", "linksys,armada385",
+ 		     "marvell,armada385", "marvell,armada380";
```
openwrt-ai, line 43: These context lines are the OpenWrt-patched form of `rango.dts`, not the mainline one. Mainline v6.18.41 has

```
	model = "Linksys WRT3200ACM";
	compatible = "linksys,rango", "linksys,armada385", "marvell,armada385",
		     "marvell,armada380";
```

([armada-385-linksys-rango.dts:14-16](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm/boot/dts/marvell/armada-385-linksys-rango.dts#L14-L16)); `linksys,wrt3200acm` and the reflowed compatible list only exist after [311-adjust-compatible-for-linksys.patch](``https://github.com/openwrt/openwrt/blob/716158b8ab03784360f0a78f3f75be1925ae5660/target/linux/mvebu/patches-6.18/311-adjust-compatible-for-linksys.patch#L9``) is applied.

That is fine for OpenWrt (990 sorts after 311 so it applies cleanly), but the patch carries a full `From`/`Subject`/`Signed-off-by`/`Cc: stable` header presenting it as an upstream submission — that version will not apply to `linux.git` as written. If you are sending the same file upstream, regenerate the hunk against pristine mainline context.

---

target/linux/mvebu/patches-6.18/990-ignore-regulatory-hint.patch
openwrt-ai, line 1: The `990-` prefix puts this outside the numbering families used in `target/linux/mvebu/patches-6.18/`: every DTS patch there lives in the 3xx block (`300`–`326`, including the other Linksys ones — `309-linksys-status-led.patch`, `310-linksys-use-eth0-as-cpu-port.patch`, `311-adjust-compatible-for-linksys.patch`), while 9xx is occupied by the unrelated IEI WT61P803 driver/documentation series (`901`–`913`). A DTS-only change belongs with its siblings.

Renaming to the next free 3xx slot, e.g. `327-arm-dts-armada-385-linksys-rango-ignore-regulatory-hint.patch`, keeps it in family and still sorts after `311`, which its context depends on.

---

openwrt-ai (commented):
Reviewed the v3 rewrite (`21bf07b` → `716158b`, 2 commits). The approach changed completely since my last pass — the hard-coded `of_machine_is_compatible()` board list is gone and replaced by the generic `marvell,ignore-regulatory-hint` root-node flag you sketched in [the v2 comment](https://github.com/openwrt/openwrt/pull/24458#issuecomment-5141013225), so I re-reviewed the whole diff rather than an increment.

What checks out:

- `112-...patch`: hunk header `@@ -4892,7 +4892,14 @@` is correct — `country_code =` really is at line 4892 of `drivers/net/wireless/marvell/mwifiex/cfg80211.c` in v6.18.41, 7 context lines and 7 added lines, and the `1 file changed, 7 insertions(+)` diffstat agrees. `<linux/of.h>` still comes in via `main.h:25`. Longest added line is 95 columns, under the 100-column limit.
- `990-...patch`: hunk `@@ -14,6 +14,7 @@` is consistent (6 context + 1 added) against the post-`311` form of `rango.dts`, and the diffstat matches. `990` sorts after `311`, so it applies.
- Both device-tree sides are covered and the split is right: WRT3200ACM through a kernel patch (upstream file) and WRT32X directly in `files/armada-385-linksys-venom.dts` (OpenWrt-only file), property placed identically right after `compatible` in both.
- `PKG_RELEASE` bumped to 4 in `package/kernel/mac80211/Makefile`. `mvebu` is `KERNEL_PATCHVER:=6.18` only, so no second patch dir needs the DTS change.
- No CI failures on `716158b`.

Six inline notes; the `of_root` / `CONFIG_OF=n` one on the driver patch is the only one with build implications, the rest are patch hygiene and upstream-submission concerns.

## Commit checks

- 9b1e7e0 "mac80211: mwifiex: ignore ROM regulatory hint on WRT3200ACM/WRT32x" — the subject scopes the change to two boards, but the diff no longer references WRT3200ACM or WRT32x at all; it adds a board-agnostic DT flag that any platform can set. The body states this correctly ("The change affects only platforms which explicitly define this flag"), so only the subject is stale. Something like `mac80211: mwifiex: allow platforms to ignore the ROM regulatory hint` would match the diff; the board names then belong solely on 716158b, whose subject is accurate.

---

httpstorm:
Rebased on main. Renamed patches to 317 and 318.

v5:
1. description: remove tag `Tested-on`
2. add `#address-cells = <1>`, `#size-cells = <0>` to fix a warning:
``` bash
arch/arm/boot/dts/marvell/armada-385-linksys-rango.dts:183.3-13:
Warning (reg_format): /soc/internal-regs/sdhci@d8000/wifi@1:reg:
property has invalid length (4 bytes) (#address-cells == 2, #size-cells == 1)
```

3. description: add info about testing

v4:
1. Add binding documentation for flag `marvell,ignore-regulatory-hint`
2. Add dt for `sd8897` and move the flag there, I used sample from docs
3. Use `device_property_read_bool()` in the driver

package/kernel/mac80211/patches/mwl/112-wifi-mwifiex-ignore-ROM-regulatory-hint-on-WRT3200AC.patch
httpstorm: Addressed using the following changes:
1. Add binding documentation for flag `marvell,ignore-regulatory-hint`
2. Add dt for `sd8897` and move the flag there, I used sample from docs
3. Use `device_property_read_bool` in the driver

package/kernel/mac80211/patches/mwl/112-wifi-mwifiex-ignore-ROM-regulatory-hint-on-WRT3200AC.patch
httpstorm: Changed subject to:
wifi: mwifiex: add dt flag to ignore incorrect ROM regulatory hint

target/linux/mvebu/patches-6.18/990-ARM-dts-armada-WRT3200ACM-add-marvell-ignore-regulat.patch
```diff
@@ -0,0 +1,47 @@
...
+
+The Linux kernel sources do not have support for WRT32x. An equivalent
+patch for these routers has to be sent to OpenWrt.
+
+[1] https://github.com/kaloz/mwlwifi/issues/173#issuecomment-307879699
+[2] https://github.com/openwrt/openwrt/issues/9956
+
+Tested-on: WRT3200ACM, OpenWrt
+Cc: stable@vger.kernel.org
+Signed-off-by: Georgi Valkov <gvalkov@gmail.com>
```
httpstorm: Removed `Cc: stable` from all patches sent to OpenWrt.

target/linux/mvebu/patches-6.18/317-ARM-dts-armada-WRT3200ACM-add-marvell-invalid-reg-hi.patch
httpstorm: Renamed patches to 317 and 318.

target/linux/mvebu/patches-6.18/990-ignore-regulatory-hint.patch
httpstorm: Moved the flag to a node for the radio.

target/linux/mvebu/files/arch/arm/boot/dts/marvell/armada-385-linksys-venom.dts
```diff
@@ -202,6 +202,15 @@
 	wp-inverted;
 	bus-width = <8>;
 	status = "okay";
+
+	#address-cells = <1>;
+	#size-cells = <0>;
+
+	wifi@1 {
+		compatible = "marvell,sd8897";
```
openwrt-ai, line 210: This describes the wrong chip. The SDIO radio on WRT3200ACM/WRT32X is a Marvell 88W8887, not an 88W8897: both devices pull in `kmod-mwifiex-sdio` ([cortexa9.mk:321-322](https://github.com/openwrt/openwrt/blob/7b9dd00e9af49e990c4cd3d87f08fbf9efc6d93e/target/linux/mvebu/image/cortexa9.mk#L321-L322)), which depends on `mwifiex-sdio-firmware`, and that package ships `sd8887_uapsta.bin` + `sd8997_uapsta.bin` ([marvell.mk:23-31](https://github.com/openwrt/openwrt/blob/7b9dd00e9af49e990c4cd3d87f08fbf9efc6d93e/package/firmware/linux-firmware/marvell.mk#L23-L31)) — the 8997 half is for Turris MOX, leaving 8887 for rango/venom. In the kernel the 8887 is a distinct device (`SDIO_DEVICE_ID_MARVELL_8887_WLAN` / `mwifiex_sdio_sd8887` with `SD8887_DEFAULT_FW_NAME`), separate from the 8897 entry. The v4 note in the PR thread says the node was copied from the binding example, which is where `marvell,sd8897` comes from.

This is not cosmetic, because the string is load-bearing here. Before this PR `&sdhci` had no child node, so `func->dev.of_node` was NULL and the OF path was skipped entirely. Adding `wifi@1` makes [`mwifiex_sdio_probe()` take the `if (func->dev.of_node)` branch](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/marvell/mwifiex/sdio.c#L600-L606), which calls [`mwifiex_sdio_probe_of()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/marvell/mwifiex/sdio.c#L518-L539) and returns `-EINVAL` ("required compatible string missing") for any compatible not in `mwifiex_sdio_of_match_table`. So the radio probes today only because the *incorrect* string happens to be one of the five listed; putting the honest `marvell,sd8887` there would kill the radio, since 8887 is in neither the match table nor the `marvell,sd8787.yaml` enum.

So the fix isn't a one-line swap — it needs `marvell,sd8887` added to `mwifiex_sdio_of_match_table` and to the binding `compatible:` enum alongside the new flag (i.e. one more patch in the series you're sending upstream). Worth sorting out before merge, since once these DTs ship, a later upstream addition of a correct `marvell,sd8887` compatible would have to keep tolerating the wrong one. Same line in [318-ARM-dts-armada-WRT3200ACM-add-marvell-ignore-regulat.patch:47](``https://github.com/openwrt/openwrt/blob/7b9dd00e9af49e990c4cd3d87f08fbf9efc6d93e/target/linux/mvebu/patches-6.18/318-ARM-dts-armada-WRT3200ACM-add-marvell-ignore-regulat.patch#L47).``

If you have evidence the chip really is an 8897 on these boards, ignore this — but then the shipped `sd8887_uapsta.bin` and the missing `pcie8897`/`sd8897` firmware would need explaining too.

---

target/linux/mvebu/patches-6.18/317-dt-bindings-net-wireless-add-flag-marvell-ignore-reg.patch
```diff
@@ -0,0 +1,44 @@
...
+
+[1] https://github.com/kaloz/mwlwifi/issues/173#issuecomment-307879699
+[2] https://github.com/openwrt/openwrt/issues/9956
+
+Signed-off-by: Georgi Valkov <gvalkov@gmail.com>
+---
+ .../devicetree/bindings/net/wireless/marvell,sd8787.yaml    | 6 ++++++
+ 1 file changed, 6 insertions(+)
+
+--- a/Documentation/devicetree/bindings/net/wireless/marvell,sd8787.yaml
```
openwrt-ai, line 34: nit: this patch touches only `Documentation/devicetree/bindings/net/wireless/marvell,sd8787.yaml`, which is not mvebu-specific, yet it lives in `target/linux/mvebu/patches-6.18/`. The property it documents is consumed by the `mac80211` package driver patch, which is built for every target, so any other target that later wants `marvell,ignore-regulatory-hint` would get the driver support but not the binding. `target/linux/generic/pending-6.18/` looks like the more natural home.

There is in-tree precedent for keeping a Documentation-only patch in the platform dir (`target/linux/mvebu/patches-6.18/901-dt-bindings-Add-IEI-vendor-prefix-and-IEI-WT61P803-P.patch`), so this may well be deliberate — is it?

---

openwrt-ai (commented):
Reviewed the v4/v5 rework (`716158b` → `7b9dd00`, 2 commits). The approach changed again since my last pass — the root-node flag is gone, replaced by a proper `wifi@1` SDIO child node carrying the property, plus a dt-bindings patch — so I re-reviewed the whole diff rather than an increment.

What checks out:

- **Patch application.** I replayed `302-add_powertables` and `311-adjust-compatible-for-linksys` onto v6.18.44 and then dry-ran both new patches: `317` and `318` apply cleanly with no fuzz. `@@ -177,6 +177,15 @@` really is where `wp-inverted;` sits in the post-`311` `armada-385-linksys-rango.dts`. Numbers `317`/`318` fill the gap between `316` and `320` with no collision.
- **`112-...patch`.** Hunk header `@@ -4892,7 +4892,14 @@` matches — `country_code =` is at line 4892 of `drivers/net/wireless/marvell/mwifiex/cfg80211.c`, 7 context + 7 added, and the `1 file changed, 7 insertions(+)` diffstat agrees.
- **`device_property_read_bool()` availability.** This resolves the `of_root` / `CONFIG_OF=n` concern from my last review: `main.h:25` pulls in `<linux/of.h>`, which unconditionally includes `<linux/property.h>`, and `device_property_read_bool()` has a non-OF implementation, so non-DT platforms keep calling `regulatory_hint()` unchanged. `adapter->dev` is `&func->dev` for SDIO (`mwifiex_add_card(..., &func->dev)` → `mwifiex_register()` → `adapter->dev = dev`), and `sdio_add_func()` → `sdio_set_of_node()` → `mmc_of_find_child_device(host, func->num)` attaches the `reg = <1>` child of `&sdhci` to exactly that device, so the property is read from the right node.
- **No collateral damage from the new node.** `btmrvl_sdio` sits on function 2, finds no `reg = <2>` child, keeps `of_node == NULL` and is unaffected; `mmc_of_find_child_device(host, 0)` for the card also finds nothing. With no `interrupts` in the node, `mwifiex_probe_of()` takes the same `err_exit` path as before.
- Both boards are covered and the split is right — WRT3200ACM via a kernel patch, WRT32X directly in the OpenWrt-only `files/` DTS — with identical node contents. `PKG_RELEASE` bumped to 4. `KERNEL_PATCHVER:=6.18` only, so no second patch dir needs the DTS change. `additionalProperties: false` in the binding makes patch `317` a genuine prerequisite. CI is green on `7b9dd00` (all three workflow runs succeeded).

Two inline notes. The `compatible = "marvell,sd8897"` one is the substantive one and is worth resolving before merge, since it now gates driver probe.

## Commit checks

- 7f78826 "mvebu: add marvell,ignore-regulatory-hint to WRT3200ACM and WRT32x" — the message describes only the addition of the flag, but the diff does noticeably more: it introduces a brand-new `wifi@1` SDIO device node (with `compatible`/`reg`) plus `#address-cells`/`#size-cells` on `&sdhci` on both boards, and adds a dt-bindings documentation patch. The new node is not a no-op — it is the first time `func->dev.of_node` is non-NULL for these boards, which switches on `mwifiex_sdio_probe_of()` and its compatible-string check (see the inline note). Worth a sentence in the body saying the SDIO radio is now described in DT and why, so the behaviour change is not invisible to `git log`.

---

httpstorm:
Changes in v6:
Rename flag:
``` bash
marvell,ignore-regulatory-hint -> marvell,invalid-reg-hint-in-rom
```

Clarify commit description:
The patch series allows affected devices, where the radio conflicts with the platform to ignore the incorrect regulatory hint in the ROM of the radio, so the platform regulatory domain can be used.

This does not allow the user to change the regulatory configuration of any radio. It only allows `phy0` to start and operate according to its certified configuration.

The change affects only platforms which explicitly define this flag, and helps overcome the incorrect behaviour of proprietary radio firmware.

Rearrange patches so that the flag is implemented in the driver, before being added to the board for WRT3200ACM.
Drop the documentation patch, since it is not needed for the functionality.

target/linux/mvebu/patches-6.18/317-ARM-dts-armada-WRT3200ACM-add-marvell-invalid-reg-hi.patch
```diff
@@ -0,0 +1,58 @@
...
+ 	bus-width = <8>;
+ 	status = "okay";
++
++	#address-cells = <1>;
++	#size-cells = <0>;
++
++	wifi@1 {
++		compatible = "marvell,sd8897";
++		reg = <1>;
++		marvell,invalid-reg-hint-in-rom;
```
openwrt-ai, line 54: Dropping the dt-bindings patch in v6 makes this property illegal against its own schema, which is more than a documentation gap. `wifi@1` under `&sdhci` is validated by [`marvell,sd8787.yaml`](``https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/net/wireless/marvell,sd8787.yaml#L18-L75``) — the schema whose `compatible:` enum contains the `marvell,sd8897` you put on line 52 — and it ends with [`additionalProperties: false` (line 75)](``https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/net/wireless/marvell,sd8787.yaml#L75).`` Every property in the node must be listed in the schema, and `marvell,invalid-reg-hint-in-rom` is not, so `dtbs_check` will reject both this node and the identical one in `armada-385-linksys-venom.dts:212`.

Nothing fails in OpenWrt CI, which does not run `dtbs_check`. But this file carries a full `From`/`Subject: [PATCH 3/3]`/`Cc: stable`/`Signed-off-by` header presenting it as an upstream submission, and on the DT list an undocumented vendor property in a node governed by `additionalProperties: false` is a hard blocker — bindings are expected to land in a patch *preceding* the first user.

The v6 note says the doc patch was dropped "since it is not needed for the functionality". That is true at runtime, but it was also the only thing making the property valid in the schema. My earlier note only questioned *where* that patch lived (`target/linux/mvebu/patches-6.18/` vs `target/linux/generic/pending-6.18/`), not whether it should exist — restoring it, in either location, looks like the right move.

Secondary effect: with the binding gone, `git grep marvell,invalid-reg-hint-in-rom` on this tree returns only these three files and nothing that defines the flag's semantics for a future reader.

---

package/kernel/mac80211/patches/mwl/112-wifi-mwifiex-add-dt-flag-to-avoid-conflict-with-plat.patch
```diff
@@ -0,0 +1,56 @@
...
+helps overcome the incorrect behaviour of proprietary radio firmware.
+
+[1] https://github.com/kaloz/mwlwifi/issues/173#issuecomment-307879699
+[2] https://github.com/openwrt/openwrt/issues/9956
+
+Tested-on WRT3200ACM, OpenWrt
+iw reg get: no longer reports region for phy2
+
+Cc: stable@vger.kernel.org
+Signed-off-by: Georgi Valkov <gvalkov@gmail.com>
```
openwrt-ai: `Cc: stable@vger.kernel.org` is back on both patches (also `317-ARM-dts-armada-WRT3200ACM-add-marvell-invalid-reg-hi.patch:35`). You removed it in v4 in response to [this thread](https://github.com/openwrt/openwrt/pull/24458#discussion_r3739905479) ("Removed `Cc: stable` from all patches sent to OpenWrt"), and the reasoning still holds: stable only picks up commits that are already in Linus' tree, and `Documentation/process/stable-kernel-rules.rst` requires a fix for a real bug — a new DT property plus the driver code that consumes it is a feature, and it has no `Fixes:` tag.

nit: line 31 also lost its colon; the OpenWrt commit message for this patch writes `Tested-on: WRT3200ACM, OpenWrt`, so the two now disagree.

```suggestion
Tested-on: WRT3200ACM, OpenWrt
iw reg get: no longer reports region for phy2

Signed-off-by: Georgi Valkov <gvalkov@gmail.com>
```

---

package/kernel/mac80211/patches/mwl/112-wifi-mwifiex-add-dt-flag-to-avoid-conflict-with-plat.patch
```diff
@@ -0,0 +1,56 @@
+From 095ca4acd81dd528fc973033b752ac268511ab4a Mon Sep 17 00:00:00 2001
+From: Georgi Valkov <gvalkov@gmail.com>
+Date: Fri, 21 Aug 2026 01:24:17 +0300
+Subject: [PATCH 2/3] wifi: mwifiex: add dt flag to avoid conflict with
```
openwrt-ai, line 4: nit: the series numbering is now dangling — this says `[PATCH 2/3]` and `317-ARM-dts-armada-WRT3200ACM-add-marvell-invalid-reg-hi.patch:4` says `[PATCH 3/3]`, but only two patches exist in the tree and there is no `1/3` anywhere. I assume `1/3` is the dt-bindings patch dropped in v6 (see my note on `317-...:54`). Whichever way that goes, the counts should end up matching what actually ships: `1/2`+`2/2` if the binding patch stays out, or `2/3`+`3/3` with the binding patch restored as `1/3`.

---

package/kernel/mac80211/patches/mwl/112-wifi-mwifiex-add-dt-flag-to-avoid-conflict-with-plat.patch
```diff
@@ -0,0 +1,56 @@
...
+ 					mwifiex_11d_code_2_region(
+ 						adapter->region_code);
++
++				/* If the reg hint in ROM conflicts with platform
++				 * configuration, it should be ignored, so the
++				 * platform regulatory domain can be used.
++				 */
+ 				if (country_code &&
++				    !device_property_read_bool(adapter->dev,
++							       "marvell,invalid-reg-hint-in-rom") &&
```
openwrt-ai, line 52: nit: the PR description was only half-updated for the v6 rename. Item 1 still reads "Implement a device-tree flag `marvell,ignore-regulatory-hint`" while item 2 and all three touched files use `marvell,invalid-reg-hint-in-rom`. The tree itself is consistent (`git grep ignore-regulatory-hint` on `a91aeaf` returns nothing), so it is just the description that needs the second occurrence fixed.

---

openwrt-ai (commented):
Reviewed the v6 rework (`7b9dd00` → `a91aeaf`, 2 commits). Three things changed since my last pass — the property was renamed to `marvell,invalid-reg-hint-in-rom`, the patches were reordered so the driver change precedes the DTS change, and the dt-bindings patch was dropped — so I re-checked the whole diff rather than an increment.

What checks out:

- **`112-...patch` metadata.** Hunk header `@@ -4892,7 +4892,14 @@` is correct: `country_code =` really is at line 4892 of `drivers/net/wireless/marvell/mwifiex/cfg80211.c` (checked against v6.18.44; unchanged from the 6.18.39 sources this package tracks), 7 context lines, 7 added lines, and the `1 file changed, 7 insertions(+)` diffstat agrees. The longest added line is 99 columns with tabs expanded, under the 100-column limit.
- **`317-...patch` applies cleanly.** I replayed `302-add_powertables` and `311-adjust-compatible-for-linksys` onto v6.18.44 and then dry-ran `317`: it applies with no fuzz and no offset — `wp-inverted;` lands exactly at line 177 in the post-`311` `armada-385-linksys-rango.dts`. `317` slots between `316` and `320` with no collision, and it is in the 3xx DTS family alongside the other Linksys patches, which resolves my earlier `990-` note.
- **Rename is complete.** `git grep ignore-regulatory-hint` on `a91aeaf` returns nothing; the three remaining references all use the new name, and the WRT32X side in `armada-385-linksys-venom.dts` is byte-identical to the rango patch's node.
- **Driver-side plumbing.** `device_property_read_bool(adapter->dev, ...)` still resolves the way it needs to: `adapter->dev` is `&func->dev` for SDIO, [`sdio_set_of_node()` runs before `device_add()` in `sdio_add_func()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mmc/core/sdio_bus.c#L374-L377) so `func->dev.of_node` points at the `reg = <1>` child, and `dev_fwnode()` falls back to `of_node`, so no `CONFIG_OF` guard is needed.
- `PKG_RELEASE` bumped 3 → 4 in `package/kernel/mac80211/Makefile`. `mvebu` is `KERNEL_PATCHVER:=6.18` only, so no second patch dir needs the DTS change. CI on `a91aeaf` is queued/running with nothing failed.

Four inline notes. The dropped dt-bindings patch is the substantive one — `marvell,sd8787.yaml` has `additionalProperties: false`, so removing the binding does not just leave the property undocumented, it makes the node fail schema validation. Note also that [the `compatible = "marvell,sd8897"` thread](https://github.com/openwrt/openwrt/pull/24458#discussion_r3800809936) from the last round is still open and unchanged in v6.

## Commit checks

- a91aeaf "mvebu: add marvell,invalid-reg-hint-in-rom to WRT3200ACM and WRT32x" — subject and body describe only the addition of the flag, but the diff does more: it introduces a brand-new `wifi@1` SDIO device node (`compatible` + `reg`) plus `#address-cells`/`#size-cells` on `&sdhci` for both boards. That is not a no-op — it is the first time `func->dev.of_node` is non-NULL on these boards, which switches on [`mwifiex_sdio_probe_of()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/marvell/mwifiex/sdio.c#L532-L539) and its `-EINVAL`-on-mismatch compatible check. Worth a sentence in the body saying the SDIO radio is now described in DT, so the behaviour change is not invisible to `git log`. (Same note I left on `7f78826`; carrying it over since that commit was rewritten rather than amended.)

---

target/linux/mvebu/patches-6.18/317-ARM-dts-armada-WRT3200ACM-add-marvell-invalid-reg-hi.patch
httpstorm: The documentation patch still exists and I sent it as part of the patch series to Linux. I only dropped it here to simplify things, which apparently was a mistake and I will have to bring it back.

httpstorm: Restored the documentation patch. Described the addition of `wifi@1` node. Removed `CC: stable`. Add fixes: `phy0` does not start on WRT3200ACM-EU when a DFS channel is used

package/kernel/mac80211/patches/mwl/112-wifi-mwifiex-add-dt-flag-to-avoid-conflict-with-plat.patch
httpstorm: Restored semicolon after tag `Tested-on`. Removed `CC: stable`. Add fixes: `phy0` does not start on WRT3200ACM-EU when a DFS channel is used

package/kernel/mac80211/patches/mwl/112-wifi-mwifiex-add-dt-flag-to-avoid-conflict-with-plat.patch
httpstorm: Included all 3 patches from the series.

package/kernel/mac80211/patches/mwl/112-wifi-mwifiex-add-dt-flag-to-avoid-conflict-with-plat.patch
httpstorm: I don't have any left-overs from the old name of the flag `ignore-regulatory-hint`.

httpstorm:
Changes:
Restored the documentation patch. Described the addition of `wifi@1` node. Restored semicolon after tag `Tested-on`. Removed `CC: stable`. Add fixes: `phy0` does not start on WRT3200ACM-EU when a DFS channel is used

package/kernel/mac80211/patches/mwl/112-wifi-mwifiex-add-dt-flag-to-avoid-conflict-with-plat.patch
```diff
@@ -0,0 +1,53 @@
...
+platform regulatory domain. The radios in WRT3200ACM/WRT32x are locked to
+the hard-coded configuration for which they are certified, so the only
+impact is that the region conflict is avoided and the 5 GHz radios can
+be used.
+
+The change affects only platforms which explicitly define this flag, and
+helps overcome the incorrect behaviour of proprietary radio firmware.
+
+[1] https://github.com/kaloz/mwlwifi/issues/173#issuecomment-307879699
+[2] https://github.com/openwrt/openwrt/issues/9956
```
httpstorm: Expanded full URL under the `Fixes:` tag.

target/linux/mvebu/patches-6.18/317-dt-bindings-net-wireless-add-flag-marvell-ignore-reg.patch
httpstorm: Moved documentation patch to `target/linux/generic/pending-6.18/`.

target/linux/generic/pending-6.18/797-dt-bindings-net-wireless-add-flag-marvell-invalid-re.patch
openwrt-ai, line 1: The `104-` prefix lands outside the number block this directory is documented to use. [`target/linux/generic/PATCHES.md:46-61`](https://github.com/openwrt/openwrt/blob/7cd59246f4c6ec833ba859c4ca070e0b9cd7b22e/target/linux/generic/PATCHES.md#L46-L61) gives the generic-patch guidelines as `0xx` → `-`, `1xx` → `-`, `6xx` → "Generic network", `7xx` → "Network / PHY". So `1xx` is explicitly unassigned, and the tree follows that: the only `1xx` slots in `pending-6.18/` are core-kernel patches — [`100-compiler.h-only-include-asm-rwonce.h-for-kernel-code.patch`](``https://github.com/openwrt/openwrt/blob/7cd59246f4c6ec833ba859c4ca070e0b9cd7b22e/target/linux/generic/pending-6.18/100-compiler.h-only-include-asm-rwonce.h-for-kernel-code.patch),`` `102-MIPS-only-process-negative-stack-offsets-on-stack-tr.patch`, `103-kbuild-export-SUBARCH.patch` — none of them network-related.

The in-family precedent for a wireless binding patch is [`796-dt-bindings-wireless-ath12k-drop-qcom-ath12k-cali.patch`](``https://github.com/openwrt/openwrt/blob/7cd59246f4c6ec833ba859c4ca070e0b9cd7b22e/target/linux/generic/pending-6.18/796-dt-bindings-wireless-ath12k-drop-qcom-ath12k-cali.patch),`` and other net binding patches sit at `737-06-dt-bindings-net-ethernet-controller-...`, `739-04-dt-bindings-net-pcs-...`, `897-00-dt-bindings-net-pse-pd-...`. `797-dt-bindings-net-wireless-add-flag-marvell-invalid-re.patch` is free and puts this next to its siblings. (`798`/`799` are free too.)

Same class of note as [the earlier `990-` one](https://github.com/openwrt/openwrt/pull/24458#discussion_r3739906336) — the target-side number is now right (`3xx` = "Architecture specific" for the rango DTS patch), it is only the generic-side one that ended up outside its block. No renumbering of `112`/`317` is needed either way.

---

target/linux/mvebu/patches-6.18/317-ARM-dts-armada-WRT3200ACM-add-marvell-invalid-reg-hi.patch
```diff
@@ -0,0 +1,60 @@
...
+WRT3200ACM/WRT32x are locked to the hard-coded configuration for which
+they are certified.
+
+On WRT3200ACM, the platform configuration is stored on the devinfo mtd3
+partition, which is read-only: cert_region=EU modelNumber=WRT3200ACM-EU
+
+The Linux kernel sources do not have support for WRT32x. An equivalent
+patch for these routers has to be sent to OpenWrt.
+
+Fixes: phy0 does not start on WRT3200ACM-EU when a DFS channel is used
```
openwrt-ai: `Fixes:` is a machine-parsed tag, not free text — it identifies a *commit*, in the form `Fixes: <12+ hex sha> ("<subject line>")`. A prose sentence there is not a valid tag: `git log --grep`, the stable/AUTOSEL tooling and patchwork all parse it as a SHA reference, and [checkpatch's Fixes check](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/scripts/checkpatch.pl#L3237-L3239) only recognises a line that starts with a hex id, so this one is silently ignored rather than validated. `Documentation/process/submitting-patches.rst` defines it the same way.

There is also nothing to point at here: this patch adds a new DT property, it does not fix an identifiable earlier commit, so the tag should simply be dropped and the sentence folded into the body prose (the paragraph starting "When the system boots, it detects a conflict…" already says exactly this).

```suggestion
```

Same line in the OpenWrt commit message of 7cd5924, which carries both `Fixes: https://github.com/openwrt/openwrt/issues/9956` and `Fixes: phy0 does not start on WRT3200ACM-EU when a DFS channel is used`. For reference, none of the last 600 commits on `openwrt/openwrt` `main` use a `Fixes:` line that is not `Fixes: <sha> ("<title>")` — the issue-URL half of that is [the still-open thread from the last round](https://github.com/openwrt/openwrt/pull/24458#discussion_r3739905138).

---

openwrt-ai (commented):
Reviewed the v7 update (`a91aeaf` → `7cd5924`, 3 commits). Three things changed since my last pass — the dt-bindings patch is back (now as `target/linux/generic/pending-6.18/104-...`), `Cc: stable` was dropped and the `Tested-on:` colon restored on both patches, and the `317` patch body now describes the new `wifi@1` node — so I re-checked the whole diff rather than an increment.

What checks out:

- **The restored binding patch resolves my main note from last round.** `104-...patch` applies to v6.18.44 with no fuzz (`patch -p1 --dry-run` clean), the hunk header `@@ -59,6 +59,12 @@` is correct — line 59 really is the `description:` of `marvell,caldata-txpwrlimit-5g-sub3` in [`marvell,sd8787.yaml`](``https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/net/wireless/marvell,sd8787.yaml#L57-L69``) — 6 context + 6 added, and the `1 file changed, 6 insertions(+)` diffstat agrees. The new property is also inserted in the right alphabetical slot (after `marvell,caldata-*`, before `marvell,wakeup-pin`), so the node is legal again under [`additionalProperties: false`](``https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/net/wireless/marvell,sd8787.yaml#L75).``
- **`317-...patch` still applies cleanly.** I replayed `302-add_powertables` and `311-adjust-compatible-for-linksys` onto v6.18.44 and dry-ran `317`: no fuzz, no offset, `wp-inverted;` still at line 177 in the post-`311` `armada-385-linksys-rango.dts`. `3xx` is the documented "Architecture specific" block for target patches, so that number is right.
- **Series numbering is consistent now.** `104` is `[PATCH 1/3]`, `112` is `[PATCH 2/3]`, `317` is `[PATCH 3/3]`, and all three exist in the tree — resolves [the dangling `2/3`+`3/3` note](https://github.com/openwrt/openwrt/pull/24458#discussion_r3827526039).
- `Cc: stable@vger.kernel.org` is gone from both patches and `Tested-on:` has its colon back, matching the OpenWrt commit message. `112-...patch` metadata is unchanged and still correct (`@@ -4892,7 +4892,14 @@`, 7 context + 7 added). `PKG_RELEASE` is 4. CI on `7cd5924` is green — all three workflow runs (Build Kernel, Build all core packages, Build PR Profile) succeeded.

Two inline notes, both about patch metadata rather than code. Note also that [the `compatible = "marvell,sd8897"` thread](https://github.com/openwrt/openwrt/pull/24458#discussion_r3800809936) from the v4 round is still open and unchanged — the restored binding patch adds the flag but not a `marvell,sd8887` compatible, so that question stands as posed.

---

target/linux/mvebu/patches-6.18/317-ARM-dts-armada-WRT3200ACM-add-marvell-invalid-reg-hi.patch
httpstorm: Removed Fixes.

target/linux/generic/pending-6.18/797-dt-bindings-net-wireless-add-flag-marvell-invalid-re.patch
httpstorm: Renamed: 104 to 797.

httpstorm:
Renamed: 104 to 797. Removed Fixes.
This is v7, I count the upstream submissions.
v7:
1. drop Cc: stable, the first patch only documents a binding
2. restore colon after Tested-on as requested by openwrt-ai
3. mention the added wifi@1 node

target/linux/generic/pending-6.18/797-dt-bindings-net-wireless-add-flag-marvell-invalid-re.patch
```diff
@@ -0,0 +1,48 @@
...
+ .../devicetree/bindings/net/wireless/marvell,sd8787.yaml    | 6 ++++++
+ 1 file changed, 6 insertions(+)
+
+--- a/Documentation/devicetree/bindings/net/wireless/marvell,sd8787.yaml
++++ b/Documentation/devicetree/bindings/net/wireless/marvell,sd8787.yaml
+@@ -59,6 +59,12 @@ properties:
+     description: Calibration data for sub-band 3 in the 5GHz band.
+     maxItems: 502
+ 
++  marvell,invalid-reg-hint-in-rom:
```
openwrt-ai, line 40: nit: re-raising [my earlier note](https://github.com/openwrt/openwrt/pull/24458#discussion_r3827526411) because I think it was read as being about the tree — you are right that the tree is clean (`git grep ignore-regulatory-hint` on `deb800d` returns nothing, and this line is the canonical name). The stale occurrence is in the **PR description** on this page, item 1:

> 1. Implement a device-tree flag `marvell,ignore-regulatory-hint`, which allows affected devices, where the radio conflicts with the platform to ignore the incorrect regulatory hint in the ROM of the radio

Item 2 right below it already uses `marvell,invalid-reg-hint-in-rom`, so the two items name the same flag differently. Only the description needs editing; nothing in the commits.

---

openwrt-ai (commented):
Reviewed the v7 update (`7cd5924` → `deb800d`, 3 commits). The delta is metadata-only — the binding patch renumbered `104` → `797`, a `Reviewed-by:` added to it, the prose `Fixes:` dropped from `317`, and regenerated `From`/`Date` headers on all three — so I re-checked the affected metadata rather than the whole diff again.

What checks out:

- **The `797-` renumber resolves my main note from last round.** [`796-dt-bindings-wireless-ath12k-drop-qcom-ath12k-cali.patch`](``https://github.com/openwrt/openwrt/blob/deb800d3a6f8baef570155fb29be8f4831c9e605/target/linux/generic/pending-6.18/796-dt-bindings-wireless-ath12k-drop-qcom-ath12k-cali.patch)`` is the highest 7xx currently in `pending-6.18/`, so `797` is free, and it now sits in the `7xx` "Network / PHY" block that [`PATCHES.md:46-61`](https://github.com/openwrt/openwrt/blob/deb800d3a6f8baef570155fb29be8f4831c9e605/target/linux/generic/PATCHES.md#L46-L61) documents, next to its sibling binding patches (`737-06`, `739-04`, `796`) instead of among the `1xx` core-kernel ones.
- **Both patches still apply cleanly after the rebase.** `LINUX_VERSION-6.18` is still `.44`, so I replayed `302-add_powertables` and `311-adjust-compatible-for-linksys` onto v6.18.44 and dry-ran the new files: `317` and `797` both apply with no fuzz and no offset. Hunk headers are unchanged from last round and still correct (`@@ -177,6 +177,15 @@` for `armada-385-linksys-rango.dts`, `@@ -59,6 +59,12 @@` for `marvell,sd8787.yaml`, `@@ -4892,7 +4892,14 @@` for `cfg80211.c`).
- **Series numbering still consistent.** `797` = `[PATCH 1/3]`, `112` = `[PATCH 2/3]`, `317` = `[PATCH 3/3]`, all three present in the tree.
- `#address-cells`/`#size-cells` placement after the existing properties matches upstream precedent for this shape — [`mt8173-elm.dtsi:429-444`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/arm64/boot/dts/mediatek/mt8173-elm.dtsi#L429-L444) does the same thing above its `marvell,sd8897` child — and the `wifi@1` node is byte-identical between the rango patch and `armada-385-linksys-venom.dts`.
- `PKG_RELEASE` is 4 against 3 on `main`. CI on `deb800d` is green: both `Build Kernel` and `Build all core packages` completed with conclusion `success`.

One nit inline. Note also that [the `compatible = "marvell,sd8897"` thread](https://github.com/openwrt/openwrt/pull/24458#discussion_r3800809936) from the v4 round is still open and unchanged — the `Reviewed-by` on the binding patch covers the new flag, not the compatible string, so that question stands as posed.

## Commit checks

- deb800d "mvebu: add marvell,invalid-reg-hint-in-rom to WRT3200ACM and WRT32x" — the prose `Fixes:` line was removed from the embedded `317` patch, but not from the OpenWrt commit message itself, which still carries both:

  ```
  Fixes: https://github.com/openwrt/openwrt/issues/9956
  Fixes: phy0 does not start on WRT3200ACM-EU when a DFS channel is used
  ```

  `Fixes:` identifies a commit, in the form `Fixes: <12+ hex sha> ("<subject>")` — I re-checked the last 600 commits on `main` and all five `Fixes:` lines there are in that form, none are a URL or free text. Since this patch adds a new DT property rather than fixing an identifiable earlier commit, the second line reads better folded into the body prose (the "When the system boots…" paragraph already says it), and the first is an issue reference, so `Link: https://github.com/openwrt/openwrt/issues/9956` matches both the tree convention and the `[2]` reference the three patch files already use. Same applies to the `Fixes: #9956` line in the PR description.

---

target/linux/generic/pending-6.18/797-dt-bindings-net-wireless-add-flag-marvell-invalid-re.patch
httpstorm: Updated PR description.

httpstorm: The Fixes tag with issue URL is used to close the issue as resolved. I dropped the other Fixes tag with description.

httpstorm:
v9: (upstream changes only, the tag had already been dropped here)
1. no change
2. drop Cc: stable
3. drop Cc: stable

v8:
1. no change
2. drop invalid tag Tested-on, replace it with description of the fix
3. no change

PR #24460, opened 2026-07-28
loongarch64: add 6.18 kernel support and virtio

pktpls:
Based on `./scripts/kernel_bump.sh -p loongarch64 -s v6.12 -t v6.18`. All patches have been upstreamed, no downstream patches are left. Lightly tested in QEMU, see virtio commit for the how. Boots reliably, networking comes up reliably and is reasonably performant, basic terminal interactions look good.

While the local build went fine, the CI kernel build complained about the CONFIG_MMC_LOONGSON2 symbol, which was introduced in 6.17. Seems worth enabling.

Also enabling some virtio symbols for operating as a VM guest. Possibly being a little too aggressive here, feedback welcome.

cc @dangowrt @hackpascal


target/linux/loongarch64/config-6.18
```diff
@@ -0,0 +1,799 @@
...
+CONFIG_LEDS_TRIGGER_DISK=y
+CONFIG_LEDS_TRIGGER_MTD=y
+CONFIG_LEDS_TRIGGER_PANIC=y
+CONFIG_LEGACY_TIOCSTI=y
+CONFIG_LIBFDT=y
+CONFIG_LOCKUP_DETECTOR=y
+CONFIG_LOCK_DEBUGGING_SUPPORT=y
+CONFIG_LOCK_SPIN_ON_OWNER=y
+CONFIG_LOONGARCH=y
+CONFIG_LOONGARCH_PLATFORM_DEVICES=y
```
openwrt-ai, line 410: `CONFIG_LOONGSON_LAPTOP=y` is silently lost in the bump — it is present in [`config-6.12:401`](https://github.com/openwrt/openwrt/blob/608d73c33fdd862a7702a66fa73651e9c578b3b0/target/linux/loongarch64/config-6.12#L401) but absent from this file, so the Loongson-3 laptop driver (backlight/hotkeys/ACPI video) disappears on the testing kernel.

Cause: in 6.18 the driver's dependency changed from `depends on ACPI` to [`depends on ACPI_EC`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/platform/loongarch/Kconfig#L19-L27), and `ACPI_EC` became a user-visible symbol ([`default X86 || LOONGARCH`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/acpi/Kconfig#L135-L138)). OpenWrt's generic config turns it off — [`target/linux/generic/config-6.18:24`](https://github.com/openwrt/openwrt/blob/608d73c33fdd862a7702a66fa73651e9c578b3b0/target/linux/generic/config-6.18#L24) has `# CONFIG_ACPI_EC is not set` — so `LOONGSON_LAPTOP` (and `# CONFIG_ACPI_EC_DEBUGFS is not set`, also dropped) fell out during the refresh.

Desired end state: add `CONFIG_ACPI_EC=y` to this target config and regenerate it (`make target/linux/refresh` with loongarch64/generic + the testing kernel selected) so `CONFIG_LOONGSON_LAPTOP=y` comes back. If dropping the laptop driver is deliberate, please say so in the commit message, since it's a user-visible feature loss between 6.12 and 6.18 on the same target.

---

target/linux/loongarch64/config-6.18
```diff
@@ -0,0 +1,799 @@
...
+CONFIG_VIRTIO_MMIO_CMDLINE_DEVICES=y
+CONFIG_VIRTIO_NET=y
+CONFIG_VIRTIO_PCI=y
+CONFIG_VIRTIO_PCI_LEGACY=y
+CONFIG_VIRTIO_PCI_LIB=y
+CONFIG_VIRTIO_PCI_LIB_LEGACY=y
+# CONFIG_VIRTIO_VSOCKETS is not set
+CONFIG_VIRTIO_VSOCKETS_COMMON=y
+CONFIG_VIRT_CPU_ACCOUNTING=y
+CONFIG_VIRT_CPU_ACCOUNTING_GEN=y
```
openwrt-ai: `VIRT_CPU_ACCOUNTING_GEN` is not a virtio option — despite the name it is the "Full dynticks CPU time accounting" choice member, and upstream says *"The accounting is thus performed at the expense of some significant overhead. For now this is only useful if you are working on the full dynticks subsystem development. If unsure, say N."* ([init/Kconfig:563-580](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/init/Kconfig#L563-L580)).

Enabling it here has three follow-on effects in this commit:

* it removes `CONFIG_TICK_CPU_ACCOUNTING=y` (present in `config-6.12`), since the two are alternatives in the same `choice`;
* it `select`s `CONTEXT_TRACKING_USER`, which in turn makes `CONTEXT_TRACKING_USER_FORCE` default to `y` (lines 134-135 of this file). Upstream on that one: *"Say Y only if you're working on the development of an architecture backend for the user context tracking. Say N otherwise, this option brings an overhead that you don't want in production."* ([kernel/time/Kconfig:158-182](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/kernel/time/Kconfig#L158-L182));
* the config also has `CONFIG_HZ_PERIODIC=y` / no `NO_HZ_FULL`, so there is no dynticks benefit to offset that cost — it is pure per-syscall overhead on every CPU.

Suggest reverting the CPU-accounting change entirely: drop lines 784-785 here, drop `CONFIG_CONTEXT_TRACKING_USER=y` and `CONFIG_CONTEXT_TRACKING_USER_FORCE=y` at lines 134-135, and restore `CONFIG_TICK_CPU_ACCOUNTING=y` (alphabetically between `CONFIG_THP_SWAP=y` and `CONFIG_TMPFS_INODE64=y`). `CONFIG_VIRT_DRIVERS=y` / `CONFIG_VIRTIO*` are the ones you actually want for guest support.

---

target/linux/loongarch64/config-6.18
```diff
@@ -0,0 +1,799 @@
...
+CONFIG_VGA_ARB_MAX_GPUS=16
+CONFIG_VIDEO=y
+CONFIG_VIRTIO=y
+CONFIG_VIRTIO_ANCHOR=y
+CONFIG_VIRTIO_BLK=y
+CONFIG_VIRTIO_CONSOLE=y
+# CONFIG_VIRTIO_DEBUG is not set
+CONFIG_VIRTIO_DMA_SHARED_BUFFER=y
+CONFIG_VIRTIO_MMIO=y
+CONFIG_VIRTIO_MMIO_CMDLINE_DEVICES=y
```
openwrt-ai, line 779: nit: `VIRTIO_MMIO_CMDLINE_DEVICES` is a development aid, not something a guest image needs — upstream: *"Be aware that using incorrect parameters (base address in particular) can crash your system - you have been warned. [...] If unsure, say 'N'."* ([drivers/virtio/Kconfig:163-172](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/virtio/Kconfig#L163-L172)). It lets anyone who can set the kernel command line poke arbitrary MMIO ranges.

The `qemu-system-loongarch64 -M virt` setup in the commit message exposes virtio over PCI, so neither this nor `CONFIG_VIRTIO_MMIO` (line 775) is exercised by the tested configuration. Since you flagged "possibly being a little too aggressive here", this is the line I'd trim first:

```suggestion
# CONFIG_VIRTIO_MMIO_CMDLINE_DEVICES is not set
```

Same reasoning arguably applies to `CONFIG_VIRTIO_PCI_LEGACY=y` / `CONFIG_VIRTIO_PCI_LIB_LEGACY=y` (lines 779, 781) — virtio 0.9 transitional device support isn't needed for a LoongArch guest, which only ever sees modern virtio.

---

target/linux/loongarch64/config-6.18
```diff
@@ -0,0 +1,799 @@
...
+CONFIG_MAGIC_SYSRQ_SERIAL=y
+CONFIG_MDIO_BUS=y
+# CONFIG_MEMCG is not set
+CONFIG_MEMORY=y
+CONFIG_MEMORY_ISOLATION=y
+# CONFIG_MFD_LOONGSON_SE is not set
+# CONFIG_MFD_LS2K_BMC_CORE is not set
+CONFIG_MIGRATION=y
+CONFIG_MMC=y
+CONFIG_MMC_LOONGSON2=y
```
openwrt-ai, line 436: nit: this builds the whole MMC/SD/SDIO core into the kernel for every loongarch64 image, which is a behaviour change relative to 6.12 rather than a refresh artifact: `config-6.12` has no `CONFIG_MMC` line at all, and the generic config disables it ([`target/linux/generic/config-6.18:24`](https://github.com/openwrt/openwrt/blob/608d73c33fdd862a7702a66fa73651e9c578b3b0/target/linux/generic/config-6.18#L24) region — `# CONFIG_MMC is not set`). OpenWrt already ships the core as `kmod-mmc` (`package/kernel/linux/modules/other.mk`), so the usual way to silence the CI's "unknown symbol" complaint without pulling the subsystem in is a `# CONFIG_MMC_LOONGSON2 is not set` line.

`MMC_LOONGSON2` is also Loongson-2K-only ([`depends on LOONGARCH`, "SD/SDIO/eMMC Host Controller on Loongson-2K series CPUs"](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/mmc/host/Kconfig#L1115-L1127)), and the commit message rationale is "it sounds useful" — was this verified on a 2K board, or is it just to satisfy the config check? If it's intentional, please also consider adding it to `config-6.12` so the two kernels don't diverge in supported hardware.

---

openwrt-ai (commented):
Reviewed 6 commits. The patch-drop claim checks out — all seven patches removed in 2da739c are present in v6.18 (`larch_insn_gen_beq`/`bne` in `arch/loongarch/kernel/inst.c`, the BPF trampoline work in `arch/loongarch/net/bpf_jit.c`, and the dwmac-loongson multichan/GMAC PCI ID changes in `drivers/net/ethernet/stmicro/stmmac/dwmac-loongson.c`). Of the 15 symbols that disappear between `config-6.12` and `config-6.18`, 14 are legitimate (removed or renamed upstream, or newly `depends on X86`); the 15th, `CONFIG_LOONGSON_LAPTOP`, looks like an unintended regression — see inline.

## Commit checks

- 481c99f "loongarch64: add kernel 6.18 as testing kernel" — the subject and (empty) body describe only adding `KERNEL_TESTING_PATCHVER:=6.18`, but the commit also carries a full config refresh of `config-6.18` (32 symbols added, 17 removed). That refresh is where `CONFIG_LOONGSON_LAPTOP=y` silently drops out. Worth a line in the body noting the refresh, and ideally splitting or explaining the symbol deltas that are not pure renames.
- 608d73c "loongarch64: 6.18: enable virtio for running as VM guest" — in addition to the virtio drivers, this commit switches the kernel's CPU time accounting mode: it removes `CONFIG_TICK_CPU_ACCOUNTING=y` and adds `CONFIG_VIRT_CPU_ACCOUNTING{,_GEN}=y` plus `CONFIG_CONTEXT_TRACKING_USER{,_FORCE}=y`. That has nothing to do with virtio guest support and is not mentioned anywhere in the message. Details inline.

The `FormalityCheck / Git & Commits` failure (empty commit bodies on 2da739c and 481c99f, subject prefixes on the two auto-generated `kernel_bump.sh` commits) is already itemised by the openwrt bot, so not repeated here.

---

target/linux/loongarch64/config-6.18
pktpls: - Keeping MMIO_CMDLINE_DEVICES, it's not just a "development aid", there are VM hypervisors that make use of it, that's why it's also enabled for other targets.
- Removing VIRTIO_PCI_LEGACY / VIRTIO_PCI_LIB_LEGACY.

target/linux/loongarch64/config-6.18
pktpls: Yes we want it, this is the driver for the MMC host controller on Loongson-2K series CPUs.

pktpls:
- Fewer enabled symbols in the virtio commit
- Re-did the config refresh and split it into a separate commit
- Not sure about the removal of FIRMWARE_EDID

(The bots are so verbose...)

pktpls:
Restoration of 6.12 needs to be redone on top of 3755d0f8e4f74bba315ad22cc2b6995cb6bf2404 so the BPF patch is fresh.

target/linux/loongarch64/config-6.18
```diff
@@ -0,0 +1,799 @@
...
+# CONFIG_VIRTIO_DEBUG is not set
+CONFIG_VIRTIO_DMA_SHARED_BUFFER=y
+CONFIG_VIRTIO_MMIO=y
+CONFIG_VIRTIO_MMIO_CMDLINE_DEVICES=y
+CONFIG_VIRTIO_NET=y
+CONFIG_VIRTIO_PCI=y
+# CONFIG_VIRTIO_PCI_LEGACY is not set
+CONFIG_VIRTIO_PCI_LIB=y
+# CONFIG_VIRTIO_VSOCKETS is not set
+CONFIG_VIRTIO_VSOCKETS_COMMON=y
```
openwrt-ai, line 785: The trimming pass went one symbol too far: `CONFIG_VIRT_DRIVERS=y` and `CONFIG_VMGENID=y` were both dropped relative to the previous revision, and `VMGENID` is the one option in that set that a VM guest genuinely needs.

Upstream is unusually blunt about it — `default y`, and *"Say Y here to use the hypervisor-provided Virtual Machine Generation ID to reseed the RNG when the VM is cloned. This is highly recommended if you intend to do any rollback / cloning / snapshotting of VMs. Prefer Y to M so that this protection is activated very early."* ([`drivers/virt/Kconfig:16-24`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/virt/Kconfig#L16-L24)). Without it, restoring a snapshot or booting two clones of the same image leaves both guests with identical RNG state — same session keys, same nonces.

It needs an explicit target-level enable because the generic config turns the whole menu off ([`target/linux/generic/config-6.18:8046`](https://github.com/openwrt/openwrt/blob/3bc89a3a6f8207339f658f028bbe39c47312f910/target/linux/generic/config-6.18#L8046) — `# CONFIG_VIRT_DRIVERS is not set`), and `VMGENID` lives inside `if VIRT_DRIVERS`, so both lines are required. OpenWrt's other VM-oriented target does exactly this: [`target/linux/x86/64/config-6.18:530`](https://github.com/openwrt/openwrt/blob/3bc89a3a6f8207339f658f028bbe39c47312f910/target/linux/x86/64/config-6.18#L530) and [`:533`](https://github.com/openwrt/openwrt/blob/3bc89a3a6f8207339f658f028bbe39c47312f910/target/linux/x86/64/config-6.18#L533).

```suggestion
CONFIG_VIRTIO_VSOCKETS_COMMON=y
CONFIG_VIRT_DRIVERS=y
CONFIG_VMGENID=y
```

---

target/linux/loongarch64/config-6.18
```diff
@@ -0,0 +1,799 @@
...
+CONFIG_SYSCTL_ARCH_UNALIGN_ALLOW=y
+CONFIG_SYSCTL_ARCH_UNALIGN_NO_WARN=y
+CONFIG_SYSCTL_EXCEPTION_TRACE=y
+CONFIG_SYSFB=y
+# CONFIG_SYSFB_SIMPLEFB is not set
+CONFIG_TASKSTATS=y
+CONFIG_TASK_DELAY_ACCT=y
+CONFIG_TASK_IO_ACCOUNTING=y
+CONFIG_TASK_XACCT=y
+CONFIG_TCG_LOONGSON=y
```
openwrt-ai: This line has no effect as written — `TCG_LOONGSON` sits inside the `if TCG_TPM` block ([`drivers/char/tpm/Kconfig:28`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/char/tpm/Kconfig#L28) … [`:258`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/char/tpm/Kconfig#L258), with the symbol itself at [`:193-195`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/char/tpm/Kconfig#L193-L195)), and `CONFIG_TCG_TPM` is off in the generic config ([`target/linux/generic/config-6.18:7006`](https://github.com/openwrt/openwrt/blob/3bc89a3a6f8207339f658f028bbe39c47312f910/target/linux/generic/config-6.18#L7006)) with no override here. kconfig will silently discard `CONFIG_TCG_LOONGSON=y`, so no `tpm_loongson` driver is built and the line survives only until the next refresh.

Two clean end states: drop the line, or — if the Loongson TPM is actually wanted — add `CONFIG_TCG_TPM=y` and regenerate the config, which will also pull in the rest of the now-visible `TCG_*` symbols. OpenWrt already ships the TPM core as `kmod-tpm` (`KCONFIG:=CONFIG_TCG_TPM` in `package/kernel/linux/modules/other.mk`), so the module route is available too.

---

target/linux/loongarch64/config-6.18
```diff
@@ -0,0 +1,799 @@
...
+# CONFIG_COMPAT_32BIT_TIME is not set
+CONFIG_CONNECTOR=y
+CONFIG_CONSOLE_TRANSLATIONS=y
+CONFIG_CONTEXT_TRACKING=y
+CONFIG_CONTEXT_TRACKING_IDLE=y
+CONFIG_CONTIG_ALLOC=y
+CONFIG_COREDUMP=y
+CONFIG_CORE_DUMP_DEFAULT_ELF_HEADERS=y
+CONFIG_CPUSETS=y
+CONFIG_CPUSETS_V1=y
```
openwrt-ai: Turning on the deprecated cgroup-v1 cpuset controller just to keep the legacy `/proc/<pid>/cpuset` file is a steep trade. The commit body frames it as "CPUSETS_V1 now needed for PROC_PID_CPUSET", which is accurate ([`init/Kconfig:1272-1275`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/init/Kconfig#L1272-L1275)), but the dependency was added precisely so that `PROC_PID_CPUSET` falls away on its own — upstream on `CPUSETS_V1`: *"Legacy cgroup v1 cpusets controller which has been deprecated by cgroup v2 implementation. […] If you do not have any such application then you are completely fine leaving this option disabled. Say N if unsure."* ([`init/Kconfig:1258-1270`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/init/Kconfig#L1258-L1270)).

`config-6.12` carries `CONFIG_PROC_PID_CPUSET=y` only as a leftover from when it depended on plain `CPUSETS` ([`config-6.12:540`](https://github.com/openwrt/openwrt/blob/3bc89a3a6f8207339f658f028bbe39c47312f910/target/linux/loongarch64/config-6.12#L540)), and this would make loongarch64 the only target in the tree with `CPUSETS_V1` enabled — no other `target/linux/*/config-*` sets it. Unless something in the loongarch64 userspace reads `/proc/<pid>/cpuset`, letting both symbols go is the natural refresh outcome:

```suggestion
# CONFIG_CPUSETS_V1 is not set
```

`CONFIG_PROC_PID_CPUSET=y` at line 564 needs to go with it.

---

target/linux/loongarch64/config-6.18
```diff
@@ -0,0 +1,799 @@
...
+CONFIG_MAGIC_SYSRQ=y
+CONFIG_MAGIC_SYSRQ_DEFAULT_ENABLE=0x01b6
+CONFIG_MAGIC_SYSRQ_SERIAL=y
+CONFIG_MDIO_BUS=y
+# CONFIG_MEMCG is not set
+CONFIG_MEMORY=y
+CONFIG_MEMORY_ISOLATION=y
+CONFIG_MFD_CORE=y
+CONFIG_MFD_LOONGSON_SE=y
+CONFIG_MFD_LS2K_BMC_CORE=y
```
openwrt-ai: nit: these two flipped from `# … is not set` in the previous revision to `=y` here, and the commit body only says "A few new loongson-related symbols" — which reads like refresh fallout. Neither symbol has a `default y`, so `=y` is a deliberate choice, not something the refresh would produce on its own:

* `MFD_LOONGSON_SE` — *"Loongson Security Engine chip controller driver"*, RNG/SM2/SM3/SM4 engines ([`drivers/mfd/Kconfig:2483-2492`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/mfd/Kconfig#L2483-L2492));
* `MFD_LS2K_BMC_CORE` — Loongson-2K board management controller, which brings its display sub-device along by default ([`:2494-2506`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/mfd/Kconfig#L2494-L2506)).

Both are plausible things to want on real Loongson hardware, but as with `MMC_LOONGSON2` this makes `config-6.18` support hardware that `config-6.12` doesn't, so the same board behaves differently depending on which kernel it runs. Could the commit body call these out explicitly (and say whether the 6.12 config should follow)?

---

target/linux/loongarch64/config-6.18
```diff
@@ -0,0 +1,799 @@
...
+CONFIG_FB_MODE_HELPERS=y
+CONFIG_FB_SIMPLE=y
+CONFIG_FB_SYSMEM_FOPS=y
+CONFIG_FB_SYSMEM_HELPERS=y
+CONFIG_FB_SYSMEM_HELPERS_DEFERRED=y
+CONFIG_FB_SYS_COPYAREA=y
+CONFIG_FB_SYS_FILLRECT=y
+CONFIG_FB_SYS_IMAGEBLIT=y
+CONFIG_FB_TILEBLITTING=y
+CONFIG_FHANDLE=y
```
openwrt-ai, line 261: Re the commit body's "Not sure about FIRMWARE_EDID, TOOLS_SUPPORT_RELR, SGL_ALLOC" — all three drops are correct, and each for a different reason:

* **`FIRMWARE_EDID`** (used to sit right here, before `CONFIG_FHANDLE`) — in 6.12 it was `depends on FB` in `drivers/video/fbdev/core/Kconfig`; in 6.18 it moved to [`drivers/video/Kconfig:64-66`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/video/Kconfig#L64-L66) and is now `depends on X86`. Unreachable on loongarch64, and the generic config already pins it off at `target/linux/generic/config-6.18:2225`.
* **`SGL_ALLOC`** — still exists upstream, but it is now set in the generic config ([`target/linux/generic/config-6.18:6147`](https://github.com/openwrt/openwrt/blob/3bc89a3a6f8207339f658f028bbe39c47312f910/target/linux/generic/config-6.18#L6147), `CONFIG_SGL_ALLOC=y`), so the target-level copy is redundant and gets deduplicated.
* **`TOOLS_SUPPORT_RELR`** — a host-toolchain probe (`def_bool $(success,…)`), and OpenWrt strips it deliberately: [`target/linux/generic/config-filter:31`](https://github.com/openwrt/openwrt/blob/3bc89a3a6f8207339f658f028bbe39c47312f910/target/linux/generic/config-filter#L31) matches `CONFIG_TOOLS_SUPPORT_RELR=.*`. The copies still present in a few other target configs are stale leftovers.

Might be worth folding a one-liner into the commit body so the next person bumping this target doesn't have to re-derive it.

---

target/linux/loongarch64/patches-6.12/004-v6.17-LoongArch-BPF-Add-basic-bpf-trampoline-support.patch
```diff
@@ -177,7 +177,7 @@ Signed-off-by: Huacai Chen <chenhuacai@loongson.cn>
 +
 +void *arch_alloc_bpf_trampoline(unsigned int size)
 +{
-+	return bpf_prog_pack_alloc(size, jit_fill_hole, false);
++	return bpf_prog_pack_alloc(size, jit_fill_hole);
```
openwrt-ai: This is what breaks the 6.12 CI build (["Build Kernel with external toolchain (loongarch64, generic)"](https://github.com/openwrt/openwrt/actions/runs/30448193112/job/90583669292)) — you already spotted it in the thread, so just pinning down the exact line for whoever redoes the restore:

```
./include/linux/filter.h:1265:7: note: declared here
 1265 | void *bpf_prog_pack_alloc(u32 size, bpf_jit_fill_hole_t bpf_fill_ill_insns, bool was_classic);
arch/loongarch/net/bpf_jit.c:1420:1: error: control reaches end of non-void function [-Werror=return-type]
cc1: all warnings being treated as errors
make[8]: *** [scripts/Makefile.build:229: arch/loongarch/net/bpf_jit.o] Error 1
```

The `restore files for v6.12` commit reinstates the pre-`3755d0f8e4f` ("kernel: bump 6.12 to 6.12.98") copy of this patch, undoing the adaptation to the third `was_classic` argument. Rebasing the restore on top of `3755d0f8e4f` as you planned fixes it; the single-line delta is:

```suggestion
+	return bpf_prog_pack_alloc(size, jit_fill_hole, false);
```

---

target/linux/loongarch64/Makefile
```diff
@@ -11,6 +11,7 @@ FEATURES:=audio display ext4 pcie boot-part rootfs-part rtc usb targz
 SUBTARGETS:=generic
 
 KERNEL_PATCHVER:=6.12
+KERNEL_TESTING_PATCHVER:=6.18
```
openwrt-ai, line 14: nit: splitting the config refresh out into its own commit is an improvement over the previous revision, but it landed *after* this one — so at `57c1940` the testing kernel is selectable while `config-6.18` is still a verbatim copy of `config-6.12`, carrying symbols that no longer exist in 6.18 (`BLK_MQ_PCI`, `MDIO_DEVICE`, `PANIC_ON_OOPS_VALUE`, `SCHED_DEBUG`) and missing ones that do. Anyone bisecting through that commit with `KERNEL_TESTING_PATCHVER` selected gets a config that doesn't reflect the kernel.

Reordering so `loongarch64: 6.18: refresh config` comes before this commit (or squashing the two) keeps every commit in the series independently buildable.

---

openwrt-ai (commented):
Re-reviewed after the force-push (previous review was against `608d73c`). All three earlier findings are addressed: `CONFIG_LOONGSON_LAPTOP=y` is back via `CONFIG_ACPI_EC=y`, the `VIRT_CPU_ACCOUNTING*` / `CONTEXT_TRACKING_USER*` switch is reverted with `TICK_CPU_ACCOUNTING` restored, and `VIRTIO_PCI_LEGACY` / `VIRTIO_PCI_LIB_LEGACY` are gone. `MMC_LOONGSON2` and `VIRTIO_MMIO_CMDLINE_DEVICES` noted as intentional — dropped.

The trimming pass overshot on one symbol though (`VMGENID`, inline), and the new refresh commit contains a few changes that read as deliberate enables rather than refresh fallout. Also answered your three open questions about `FIRMWARE_EDID` / `SGL_ALLOC` / `TOOLS_SUPPORT_RELR` inline — all three drops are correct.

## Commit checks

- 41cb709 "loongarch64: 6.18: refresh config" — the body reads as a pure refresh (removed/renamed symbols), but the same commit newly enables three drivers that the previous revision had switched off: `CONFIG_MFD_LOONGSON_SE=y`, `CONFIG_MFD_LS2K_BMC_CORE=y` and `CONFIG_TCG_LOONGSON=y`. None of them carry a `default y`, so they are choices, not fallout — "A few new loongson-related symbols" undersells that. Details inline.
- 6fedbe0 "kernel/loongarch64: restore files for v6.12" — restores a stale copy of `004-v6.17-LoongArch-BPF-Add-basic-bpf-trampoline-support.patch`, reverting the `bpf_prog_pack_alloc()` signature fix from `3755d0f8e4f`; this is what fails the 6.12 CI build. You already called this out in the thread, so inline is just the exact line.

The `FormalityCheck / Git & Commits` failure (empty bodies on ee02da7 and 57c1940, subject prefixes on the two `kernel_bump.sh` commits) is itemised by the openwrt bot, so not repeated.

---

target/linux/loongarch64/config-6.18
```diff
@@ -0,0 +1,800 @@
...
+CONFIG_MDIO_BUS=y
+# CONFIG_MEMCG is not set
+CONFIG_MEMORY=y
+CONFIG_MEMORY_ISOLATION=y
+CONFIG_MFD_CORE=y
+CONFIG_MFD_LOONGSON_SE=y
+CONFIG_MFD_LS2K_BMC_CORE=y
+CONFIG_MIGRATION=y
+CONFIG_MMC=y
+CONFIG_MMC_LOONGSON2=y
```
openwrt-ai, line 436: This enables the MMC core and the Loongson-2K host controller, but `CONFIG_MMC_BLOCK` is left at the generic default [`# CONFIG_MMC_BLOCK is not set`](https://github.com/openwrt/openwrt/blob/0448089c4389042c71723289b5a6ae713b65394d/target/linux/generic/config-6.18#L3880), so no `/dev/mmcblk*` node is ever created and the host driver cannot actually be used. Every in-tree target that sets `CONFIG_MMC=y` (x86, mvebu, sunxi, rockchip, mediatek/filogic, armsr/armv8, qualcommax, …) also sets `CONFIG_MMC_BLOCK=y`.

```suggestion
CONFIG_MMC=y
CONFIG_MMC_BLOCK=y
CONFIG_MMC_LOONGSON2=y
```

Leaving block support to `kmod-mmc` is not an option here: that package's [`KCONFIG`](https://github.com/openwrt/openwrt/blob/0448089c4389042c71723289b5a6ae713b65394d/package/kernel/linux/modules/other.mk#L331-L332) asks for `CONFIG_MMC` and `CONFIG_MMC_BLOCK` as modules, which conflicts with the built-in `CONFIG_MMC=y` on this line.

---

target/linux/loongarch64/config-6.18
```diff
@@ -0,0 +1,800 @@
...
+CONFIG_MACH_LOONGSON64=y
+CONFIG_MAGIC_SYSRQ=y
+CONFIG_MAGIC_SYSRQ_DEFAULT_ENABLE=0x01b6
+CONFIG_MAGIC_SYSRQ_SERIAL=y
+CONFIG_MDIO_BUS=y
+# CONFIG_MEMCG is not set
+CONFIG_MEMORY=y
+CONFIG_MEMORY_ISOLATION=y
+CONFIG_MFD_CORE=y
+CONFIG_MFD_LOONGSON_SE=y
```
openwrt-ai: As configured, this MFD driver has nothing to bind to. `loongson-se.c` registers exactly two cells — [`{ .name = "loongson-rng" }, { .name = "tpm_loongson" }`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/mfd/loongson-se.c#L181-L184). In v6.18 there is no `loongson-rng` driver in tree at all (the only reference to that string is the cell definition itself), and the `tpm_loongson` cell is driven by [`TCG_LOONGSON`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/char/tpm/Kconfig#L193-L196), which this same config leaves off at [line 721](https://github.com/openwrt/openwrt/blob/0448089c4389042c71723289b5a6ae713b65394d/target/linux/loongarch64/config-6.18#L721).

So `MFD_LOONGSON_SE=y` costs kernel size and gains no functionality unless `CONFIG_TCG_LOONGSON` is enabled too. Was the intent to enable the TPM interface as well, or should this symbol stay off until the RNG driver lands upstream?

(`MFD_LS2K_BMC_CORE` on the next line is fine by comparison — its `simple-framebuffer` cell is served by the `CONFIG_FB_SIMPLE=y` already in this config, even though the `ls2k-ipmi-si` cells need `IPMI_LS2K`, which is not enabled.)

---

target/linux/loongarch64/config-6.18
openwrt-ai, line 436: nit: the PR body and the commit message disagree about this symbol's origin — the body says `CONFIG_MMC_LOONGSON2` "was introduced in 6.17", while 95dce11's body says "MMC_LOONGSON is the MMC host driver, new in 6.18". Only one can be right; worth settling before merge so the commit message stays accurate.

---

openwrt-ai (commented):
Re-reviewed after the force-push (previous review was against `3bc89a3`). The series is in much better shape:

- The 6.12 restore is now byte-identical to `main` — `git diff 5ecd317..0448089` touches only `Makefile` and `config-6.18`, so the stale BPF patch problem is gone and the 6.18 kernel build is green.
- `57c1940`'s hidden config refresh has been split out into its own commit (`01cb3a9`), and `0448089` is now a clean one-line `KERNEL_TESTING_PATCHVER` addition. Good.
- The `VMGENID` overshoot is fixed, and the deliberate enables that were buried in the previous refresh commit (`MFD_LOONGSON_SE`, `MFD_LS2K_BMC_CORE`, `TCG_LOONGSON`) are now either documented in `95dce11` or explicitly left off.

I verified every claim in `01cb3a9`'s body against v6.18.41 and they all hold: `BLK_MQ_PCI`, `MDIO_DEVICE`, `MDIO_DEVRES`, `PANIC_ON_OOPS_VALUE` and `SCHED_DEBUG` no longer exist as Kconfig symbols; [`FIRMWARE_EDID` is `depends on X86`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/video/Kconfig#L64-L66); `SGL_ALLOC=y` comes from `target/linux/generic/config-6.18`; and `TOOLS_SUPPORT_RELR` is filtered by `target/linux/generic/config-filter`. That answers the three open questions from your earlier comment — all three drops are correct. The one addition in that commit not covered by the bullet list, `SND_SOC_LOONGSON_I2S_PLATFORM=y`, is genuine fallout (`SND_SOC_LOONGSON_CARD` gained a `select … if OF` in 6.18), so no action needed there.

Two things to look at inline: `CONFIG_MMC_BLOCK` is missing alongside the new MMC host driver, and `MFD_LOONGSON_SE=y` currently has no cell driver to bind to.

## Commit checks

- 95dce11 "loongarch64: 6.18: enable more Loongson drivers" — the body says "MMC_LOONGSON is the MMC host driver", but the symbol the diff adds (and the only one that exists in v6.18) is `MMC_LOONGSON2` (`drivers/mmc/host/Kconfig`, `obj-$(CONFIG_MMC_LOONGSON2) += loongson2-mmc.o`). The commit also enables `CONFIG_MMC=y` and `CONFIG_MFD_CORE=y`, which are dependencies rather than "Loongson drivers" — worth a word in the body.

The `FormalityCheck / Git & Commits` failure (missing `<prefix>: ` on the two auto-generated `kernel_bump.sh` subjects) is itemised by the openwrt bot, so not repeated here.

---

hackpascal:
Tested-by: Weijie Gao <hackpascal@gmail.com>

PCIe/SATA/USB/Integrated GPU/Ethernet are OK.
Tested with AMD RX580 on 3A6000.

hackpascal (approved):

pktpls:
I think we're good to go here @dangowrt

openwrt-ai (commented):
Re-reviewed after the force-push (previous review was against `0448089`). The series is now rebased onto `be2f856`; content-wise only three things changed, and both of my previous inline findings are addressed:

- `CONFIG_MMC_BLOCK=y` is now enabled alongside `MMC=y` / `MMC_LOONGSON2=y` — matches how every other target with an MMC host driver does it (`# CONFIG_MMC_BLOCK is not set` in `target/linux/generic/config-6.18`, overridden per-target). `REGMAP_MMIO=y` is present, so the `select` from [`MMC_LOONGSON2` in drivers/mmc/host/Kconfig:1115-1127](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/mmc/host/Kconfig#L1115-L1127) is satisfied.
- `MFD_LOONGSON_SE` / `MFD_LS2K_BMC_CORE` are back to `is not set`, and the now-consequential `MFD_CORE=y` line is gone with them. Dropping the `# CONFIG_TCG_LOONGSON is not set` line is correct too — [`TCG_LOONGSON` depends on `MFD_LOONGSON_SE`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/char/tpm/Kconfig#L193-L200), so with the MFD driver off the symbol is invisible and would not appear in a refreshed config.

Verified the rest is unchanged: `52289bcc` (patch drop) is byte-identical to the previous `cde72c21`, the `a14ad2e4` refresh differs from `01cb3a9` only by the `TCG_LOONGSON` line and hunk offsets, and `cb46e5f3` (virtio) is identical to `149328be`. The `kernel_bump.sh` pair is clean — `git diff 56d87e7d~1..7e1c5e61 -- target/linux/loongarch64/patches-6.12 target/linux/loongarch64/config-6.12` is empty, so the 6.12 restore is byte-exact and the stale BPF patch problem stays fixed. Net PR diff against the base is just `Makefile` (+1 line) and the new `config-6.18`. All checks on `f6e0bbe` are green, including the 6.18 kernel build and `FormalityCheck / Git & Commits`.

One leftover from the last round, below.

## Commit checks

- `cd5faa9` "loongarch64: 6.18: enable Loongson MMC driver" — the body still names the wrong symbol: "CONFIG_MMC_LOONGSON is the MMC host driver". The symbol the diff enables, and the only one that exists in v6.18, is `CONFIG_MMC_LOONGSON2` ([drivers/mmc/host/Kconfig:1115](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/mmc/host/Kconfig#L1115), `obj-$(CONFIG_MMC_LOONGSON2) += loongson2-mmc.o`). The subject was fixed in this revision but the body was not; the PR description gets it right. Worth a `git commit --amend` before this lands, since the message is what ends up in history.

---

jonasjelonek:
Please fix the wording of "Has been tested both with TAP, and passt userspace networking." in the 'enable virtio' commit, especially the typo with "passt" confuses.

pktpls:
No no, passt is the userspace networking stack, for when you don't want to sudo :)) I will clarify the commit message

jonasjelonek:
Ok, then please word it like "the 'passt' userspace networking stack" for stupid people like me who have never heard of this.

openwrt-ai (commented):
Re-reviewed after the force-push (previous review was against `f6e0bbe`). The series was rebased onto `d329a480`; `git range-diff f6e0bbe~7..f6e0bbe c5a7830~7..c5a7830` shows commits 1–4 and 7 identical and **no tree changes at all** — the only differences are two commit messages, and both fix findings from the last round:

- `ccca43ab` "loongarch64: 6.18: enable Loongson MMC driver" — body now says `CONFIG_MMC_LOONGSON2`, which is the symbol the diff enables and the only one that exists ([`drivers/mmc/host/Kconfig:1115`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mmc/host/Kconfig#L1115-L1127) in v6.18.44). The "new in kernel 6.17" claim also checks out — the symbol is absent from v6.12.103. My last commit check is resolved.
- `cb749287` "loongarch64: 6.18: enable virtio for running as VM guest" — the message is restructured into numbered steps and now spells out "the `passt` userspace networking stack for rootless operation", which is what @jonasjelonek asked for.

Everything else re-verified against the new base:

- Net PR diff against `d329a480` is still just `target/linux/loongarch64/Makefile` (+1, `KERNEL_TESTING_PATCHVER:=6.18`) and the new 799-line `config-6.18`. No stray hunks picked up by the rebase.
- The 6.12 restore is byte-exact: `git diff f48d07b8^..c5a7830 -- target/linux/loongarch64/patches-6.12 target/linux/loongarch64/config-6.12` is empty, so the stale BPF patch problem stays fixed.
- `target/linux/generic/kernel-6.18` is at `6.18.44`, matching the tree I checked the symbols against.
- All 13 checks on `c5a7830` are green, including `Build Kernel with external toolchain (loongarch64, generic, 6.18)`, `Check Kernel patches`, and `FormalityCheck / Git & Commits`. The only bot output left is the "author email not linked to a GitHub account" warning, which is a profile setting rather than a change to this series.

Reviewed 7 rebased commits (2 with message changes); no new issues found.

---

jonasjelonek:

pktpls:
@hackpascal feel free to open a PR for making 6.18 the default, as soon as you feel comfortable running 6.18 :wave:

PR #24462, opened 2026-07-29
scripts: dl_github_archive: harden the commit ts cache

neheb:
A truncated or corrupted cache file causes line.split() to raise ValueError when unpacking. Wrap the parsing in try/except so bad lines are skipped instead of crashing the script.

### DIFF
--- a/scripts/dl_github_archive.py
+++ b/scripts/dl_github_archive.py
@@ -22,6 +22,8 @@ import sys
 import time
 import urllib.request
 
+from contextlib import contextmanager
+
 TMPDIR = os.environ.get('TMP_DIR') or '/tmp'
 TMPDIR_DL = os.path.join(TMPDIR, 'dl')
 
@@ -162,50 +164,58 @@ class GitHubCommitTsCache(object):
         self.cachef = os.path.join(TMPDIR_DL, self.__cachef)
         self.cache = {}
 
+    @contextmanager
+    def _lock(self, exclusive):
+        fileno = os.open(self.cachef + '.lock', os.O_RDWR | os.O_CREAT)
+        try:
+            fcntl.flock(fileno, fcntl.LOCK_EX if exclusive else fcntl.LOCK_SH)
+            yield
+        finally:
+            os.close(fileno)
+
     def get(self, k):
         """Get timestamp with key ``k``."""
-        fileno = os.open(self.cachef, os.O_RDONLY | os.O_CREAT)
-        with os.fdopen(fileno) as fin:
-            try:
-                fcntl.lockf(fileno, fcntl.LOCK_SH)
-                self._cache_init(fin)
-                if k in self.cache:
-                    ts = self.cache[k][0]
-                    return ts
-            finally:
-                fcntl.lockf(fileno, fcntl.LOCK_UN)
-        return None
+        with self._lock(False):
+            self._cache_init()
+            return self.cache.get(k, (None, None))[0]
 
     def set(self, k, v):
         """Update timestamp with ``k``."""
-        fileno = os.open(self.cachef, os.O_RDWR | os.O_CREAT)
-        with os.fdopen(fileno, 'w+') as f:
-            try:
-                fcntl.lockf(fileno, fcntl.LOCK_EX)
-                self._cache_init(f)
-                self.cache[k] = (v, int(time.time()))
-                self._cache_flush(f)
-            finally:
-                fcntl.lockf(fileno, fcntl.LOCK_UN)
+        with self._lock(True):
+            self._cache_init()
+            self.cache[k] = (v, int(time.time()))
+            self._cache_flush()
 
-    def _cache_init(self, fin):
-        for line in fin:
-            k, ts, updated = line.split()
-            ts = int(ts)
-            updated = int(updated)
-            self.cache[k] = (ts, updated)
+    def _cache_init(self):
+        try:
+            with open(self.cachef, errors='replace') as fin:
+                for line in fin:
+                    try:
+                        if not line.endswith('\n'):
+                            continue
+                        k, ts, updated = line.split()
+                        ts = int(ts)
+                        updated = int(updated)
+                        self.cache[k] = (ts, updated)
+                    except ValueError:
+                        continue
+        except FileNotFoundError:
+            pass
 
-    def _cache_flush(self, fout):
+    def _cache_flush(self):
         cache = sorted(self.cache.items(), key=lambda a: a[1][1])
         cache = cache[:self.__cachen]
         self.cache = {}
-        os.ftruncate(fout.fileno(), 0)
-        fout.seek(0, os.SEEK_SET)
-        for k, ent in cache:
-            ts = ent[0]
-            updated = ent[1]
-            line = '{0} {1} {2}\n'.format(k, ts, updated)
-            fout.write(line)
+        tmpf = self.cachef + '.tmp'
+        with open(tmpf, 'w') as fout:
+            for k, ent in cache:
+                ts = ent[0]
+                updated = ent[1]
+                line = '{0} {1} {2}\n'.format(k, ts, updated)
+                fout.write(line)
+            fout.flush()
+            os.fsync(fout.fileno())
+        os.replace(tmpf, self.cachef)
 
 
 class DownloadGitHubTarball(object):


scripts/dl_github_archive.py
```diff
@@ -190,10 +190,13 @@ def set(self, k, v):
 
     def _cache_init(self, fin):
         for line in fin:
-            k, ts, updated = line.split()
-            ts = int(ts)
-            updated = int(updated)
-            self.cache[k] = (ts, updated)
+            try:
```
openwrt-ai: The `try` starts one line too late to cover the "corrupted cache file" case from the commit message: the decode of each line happens in the `for line in fin:` iteration, which is outside the block, and `UnicodeDecodeError` is a subclass of `ValueError`. So a cache file containing non-UTF-8 bytes (the typical result of real corruption, as opposed to a clean truncation) still propagates out of `_cache_init` and aborts the script:

```
  File "<string>", line 4, in <module>
  File "<frozen codecs>", line 322, in decode
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 6: invalid start byte
```

(reproduced with exactly this loop body over a file whose second line starts with `\xff\xfe`; the traceback points at the `for`, not at the `split()`).

Since the cache format is plain ASCII, the smallest fix is to make the reads non-strict at the two open sites rather than restructure this loop — [`os.fdopen(fileno)` in `get()` at dl_github_archive.py:168](https://github.com/openwrt/openwrt/blob/704729429e1e532611553b675ea92a427ab2ed61/scripts/dl_github_archive.py#L168) and [`os.fdopen(fileno, 'w+')` in `set()` at dl_github_archive.py:182](https://github.com/openwrt/openwrt/blob/704729429e1e532611553b675ea92a427ab2ed61/scripts/dl_github_archive.py#L182) — by passing `errors='replace'`; the mangled lines then fail the `split()`/`int()` checks here and get skipped as intended. Wrapping the whole `for` in the `try` would also stop the crash, but it would discard every remaining entry after the first bad byte instead of just the bad line. Both sites are outside this hunk, hence no suggestion block.

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the diff, and the per-line skip is the right shape for the truncated-cache case. One gap noted inline: the decode step is outside the new `try`, so genuinely corrupted (non-UTF-8) cache files still crash `_cache_init`.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push since my last review at 7047294). The `errors='replace'` additions in `get()` and `set()` resolve the [earlier finding](https://github.com/openwrt/openwrt/pull/24462#discussion_r3670906874) — I re-ran the loop over a cache file whose second line starts with `\xff\xfe` and `_cache_init` now completes, keeping the good entries before and after the bad line. No inline issues left.

## Commit checks

- 56a5f10 "scripts: dl_github_archive: handle malformed cache lines" — the message only describes wrapping the parse in try/except ("Wrap the parsing in try/except so bad lines are skipped"), but the diff also switches both `os.fdopen()` sites to `errors='replace'`. Those two hunks are the ones that actually handle a *corrupted* (non-UTF-8) cache file: the per-line decode happens in `for line in fin:`, outside the `try`, so without them only cleanly truncated files are covered. Worth one sentence in the commit body so the reason for the `errors='replace'` change is recorded.

---

neheb:
ping @jonasjelonek

scripts/dl_github_archive.py
```diff
@@ -190,10 +190,13 @@ def set(self, k, v):
...
+                k, ts, updated = line.split()
+                ts = int(ts)
+                updated = int(updated)
+                self.cache[k] = (ts, updated)
+            except ValueError:
+                continue
 
     def _cache_flush(self, fout):
```
champtar: Why not using atomic writes (write to a temp file, flush, rename) in `_cache_flush` instead of fixing the symptoms in `_cache_init` ?

neheb:
```
Why not just switch to temp+rename: it conflicts with the locking model. set() locks the cache file's fd/inode (dl_github_archive.py:181-184) and re-reads from that same inode so concurrent read-modify-write is correct. If _cache_flush wrote a temp file and os.renamed over the path, a second writer that had already opened the old inode would:
1. block on LOCK_EX of the stale inode,
2. read stale content (missing the first writer's entry),
3. rename its temp over the path — clobbering the first writer's update (lost entries
```

champtar: sure, you need to update the locking, you can lock on either a dedicated lock file, the parent dir, or the script itself.

What bother me here is that you know you can have corrupted cache file and instead of fixing the root cause by using atomic writes, you try to paper over the symptoms.
You will still parse some truncated last line as if valid, you also need to check the line ends with `\n` to avoid that.

openwrt-ai (commented):
Reviewed 1 new commit (force-push since my last review at 56a5f10). Diffed `56a5f10..01e9ce8` locally: the commit is byte-identical to the previously reviewed one — same tree change to `scripts/dl_github_archive.py`, same message, only the parent moved. Nothing new to flag, and no new inline issues.

My earlier note about the commit body not mentioning the two `errors='replace'` hunks is still unaddressed, but it is unchanged so I'm not re-posting it inline: https://github.com/openwrt/openwrt/pull/24462#pullrequestreview-4815226449

No CI checks have reported on the new head yet, so nothing CI-grounded to review.

---

scripts/dl_github_archive.py
```diff
@@ -162,50 +162,66 @@ def __init__(self):
...
-                fcntl.lockf(fileno, fcntl.LOCK_UN)
-        return None
+                with open(self.cachef, errors='replace') as fin:
+                    self._cache_init(fin)
+            except FileNotFoundError:
+                pass
+            return self.cache.get(k, (None, None))[0]
+        finally:
+            fcntl.lockf(fileno, fcntl.LOCK_UN)
+            os.close(fileno)
```
champtar: closing is enough to release the lock, and it would be cleaner to have a `self._unlock()`, or even cleaner IMO use a context manager (not tested)

```
from contextlib import contextmanager

class GitHubCommitTsCache(object):
...
    @contextmanager
    def _lock(self, exclusive):
        fd = os.open(self.cachef + '.lock', os.O_CREAT | os.O_RDWR)
        try:
            fcntl.flock(fd, fcntl.LOCK_EX if exclusive else fcntl.LOCK_SH)
            yield fd
        finally:
            os.close(fd)

...
    def get(self, k):
        """Get timestamp with key ``k``."""
        with self._lock(False):
            try:
...
```

scripts/dl_github_archive.py
```diff
@@ -162,50 +164,62 @@ def __init__(self):
...
-                if k in self.cache:
-                    ts = self.cache[k][0]
-                    return ts
-            finally:
-                fcntl.lockf(fileno, fcntl.LOCK_UN)
-        return None
+                with open(self.cachef, errors='replace') as fin:
+                    self._cache_init(fin)
+            except FileNotFoundError:
+                pass
```
champtar: Now in both `get` and `set` you have the exact same try block around `_cache_init`, you can move it to `_cache_init`

champtar (approved):
Not tested, but looks good to me

scripts/dl_github_archive.py
```diff
@@ -162,50 +164,58 @@ def __init__(self):
...
         self.cache = {}
-        os.ftruncate(fout.fileno(), 0)
-        fout.seek(0, os.SEEK_SET)
-        for k, ent in cache:
-            ts = ent[0]
-            updated = ent[1]
-            line = '{0} {1} {2}\n'.format(k, ts, updated)
-            fout.write(line)
+        tmpf = self.cachef + '.tmp'
+        with open(tmpf, 'w') as fout:
```
openwrt-ai, line 210: The write side lost the `errors='replace'` it had at 01e9ce8, where the same `w+` handle served both reads and writes ([`os.fdopen(fileno, 'w+', errors='replace')`](https://github.com/openwrt/openwrt/blob/01e9ce8787d035380b884ea9a45e011c42107d06/scripts/dl_github_archive.py#L182)). Now the read at line 191 keeps it but this open does not, and the two are asymmetric: a corrupted byte in the key field decodes to `U+FFFD`, still splits into three tokens with two valid ints, so it enters `self.cache` and is written straight back out here. If the preferred encoding is not UTF-8 the write raises:

```
{'https://api.github.com/a': (100, 200), '�https://api.github.com/b': (101, 201)}
UnicodeEncodeError: 'ascii' codec can't encode character '�' in position 0: ordinal not in range(128)
```

(reproduced by feeding a cache file whose second line starts with `\xff` through this exact `_cache_init` loop, then writing the resulting dict.) Because it fires before `os.replace`, the on-disk cache stays intact — but it also stays poisoned, so every later `set()` re-reads the same key and fails again, and a `.tmp` file is left behind each time. `_init_commit_ts` catches `Exception`, so this surfaces as a spurious "Cannot fetch commit ts" rather than a traceback.

```suggestion
        with open(tmpf, 'w', errors='replace') as fout:
```

`encoding='utf-8'` on both opens instead of relying on the locale would work too, if you prefer making the format explicit.

---

scripts/dl_github_archive.py
```diff
@@ -162,50 +164,58 @@ def __init__(self):
...
+                        updated = int(updated)
+                        self.cache[k] = (ts, updated)
+                    except ValueError:
+                        continue
+        except FileNotFoundError:
+            pass
+
+    def _cache_flush(self):
         cache = sorted(self.cache.items(), key=lambda a: a[1][1])
         cache = cache[:self.__cachen]
```
openwrt-ai, line 207: nit: pre-existing and unchanged by this PR (same two lines at [dl_github_archive.py:199-200 on the base](https://github.com/openwrt/openwrt/blob/f6cc96990ae2eaa1ebf816ad626e2e093874744d/scripts/dl_github_archive.py#L199-L200)), so feel free to ignore — but since you're rewriting `_cache_flush` anyway: the eviction keeps the wrong half. `sorted(key=lambda a: a[1][1])` orders by `updated` ascending, so `[:2048]` retains the 2048 *least* recently updated entries. The entry `set()` just added carries `updated = int(time.time())`, i.e. the maximum, so once the cache is full every new entry is discarded by the very flush that is supposed to persist it, and the cache stops accepting anything new until the file is deleted.

```suggestion
        cache = cache[-self.__cachen:]
```

If you'd rather keep this PR to the atomicity fix, a separate commit is fine too.

---

scripts/dl_github_archive.py
```diff
@@ -162,50 +164,58 @@ def __init__(self):
...
+        tmpf = self.cachef + '.tmp'
+        with open(tmpf, 'w') as fout:
+            for k, ent in cache:
+                ts = ent[0]
+                updated = ent[1]
+                line = '{0} {1} {2}\n'.format(k, ts, updated)
+                fout.write(line)
+            fout.flush()
+            os.fsync(fout.fileno())
+        os.replace(tmpf, self.cachef)
```
openwrt-ai, line 218: nit: the PR description still describes the first revision — "A truncated or corrupted cache file causes line.split() to raise ValueError when unpacking. Wrap the parsing in try/except so bad lines are skipped instead of crashing the script." That is no longer what the change does: the try/except is now a secondary guard, and the substance is this temp-file + fsync + rename plus the move to a dedicated `.lock` file. The commit message is accurate and covers all of it; only the PR body is stale, and it is what ends up quoted in review notifications.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push since my last review at 01e9ce8). The approach changed substantially in this revision — `_cache_init` no longer takes a handle, locking moved to a dedicated `.lock` file via a `@contextmanager`, and `_cache_flush` now does temp-file + fsync + `os.replace`. That addresses champtar's points about atomic writes, the shared/exclusive lock split, the truncated last line, and hoisting the duplicated try into `_cache_init`.

Every claim in the commit message checks out against the diff, so no commit-message issues. Three inline notes, one of them a real behaviour change introduced by this revision: the write path lost the `errors='replace'` it carried at 01e9ce8, which makes a corrupted cache byte a self-perpetuating write failure under a non-UTF-8 preferred encoding. The other two are nits (one pre-existing eviction bug in the function being rewritten, one stale PR description).

CI has not reported on this head yet (status `pending`, 0 checks), so nothing CI-grounded to review.

---

PR #24463, opened 2026-07-29
mtd: trx: fix endian bug in trx_fixup and trx_check

neheb:
trx->len is stored in little-endian format but read natively in trx_fixup() and trx_check(), causing a wrong length value on big-endian systems. This makes crc32buf read out of bounds in trx_fixup(), and causes incorrect size checks in trx_check(). Wrap all native reads of trx->len with STORE32_LE().

### DIFF
--- a/package/system/mtd/src/trx.c
+++ b/package/system/mtd/src/trx.c
@@ -97,8 +97,13 @@ trx_fixup(int fd, const char *name)
 		goto err;
 	}
 
+	if (STORE32_LE(trx->len) > len) {
+		fprintf(stderr, "TRX header length exceeds partition size\n");
+		goto err;
+	}
+
 	scan = ptr + offsetof(struct trx_header, flag_version);
-	trx->crc32 = crc32buf(scan, trx->len - (scan - ptr));
+	trx->crc32 = STORE32_LE(crc32buf(scan, STORE32_LE(trx->len) - (scan - ptr)));
 	msync(ptr, sizeof(struct trx_header), MS_SYNC|MS_INVALIDATE);
 	munmap(ptr, len);
 	close(bfd);
@@ -128,7 +133,7 @@ trx_check(int imagefd, const char *mtd, char *buf, int *len)
 	}
 
 	if (ntohl(trx->magic) != opt_trxmagic ||
-	    trx->len < sizeof(struct trx_header)) {
+	    STORE32_LE(trx->len) < sizeof(struct trx_header)) {
 		if (quiet < 2) {
 			fprintf(stderr, "Bad trx header\n");
 			fprintf(stderr, "This is not the correct file format; refusing to flash.\n"
@@ -144,7 +149,7 @@ trx_check(int imagefd, const char *mtd, char *buf, int *len)
 		exit(1);
 	}
 
-	if(mtdsize < trx->len) {
+	if(mtdsize < STORE32_LE(trx->len)) {
 		fprintf(stderr, "Image too big for partition: %s\n", mtd);
 		close(fd);
 		return 0;


package/system/mtd/src/trx.c
```diff
@@ -98,7 +98,7 @@ trx_fixup(int fd, const char *name)
 	}
 
 	scan = ptr + offsetof(struct trx_header, flag_version);
-	trx->crc32 = crc32buf(scan, trx->len - (scan - ptr));
+	trx->crc32 = crc32buf(scan, STORE32_LE(trx->len) - (scan - ptr));
```
openwrt-ai: The read is byte-swapped now, but the store on the same statement still isn't: `trx->crc32` is assigned in native byte order, so on a big-endian host (bcm47xx/MIPS BE is the primary trx consumer, i.e. exactly the case this commit targets) `trx_fixup()` writes a byte-swapped CRC32 into the header and the resulting image fails validation. `mtd_fixtrx()` gets this right — see [`trx->crc32 = STORE32_LE(crc32buf(buf, data_size));` at trx.c:248](https://github.com/openwrt/openwrt/blob/436320566dd8080f71ccad3ad7e7018679057c2b/package/system/mtd/src/trx.c#L248).

```suggestion
	trx->crc32 = STORE32_LE(crc32buf(scan, STORE32_LE(trx->len) - (scan - ptr)));
```

Separately: even with the correct byte order, `trx->len` here is never bounded against `len` (the mmap size), so a header claiming a length larger than the partition still makes `crc32buf()` walk past the end of the mapping. The commit message names that out-of-bounds read as the motivation, but the swapped-length case is only one way to reach it — worth an explicit `STORE32_LE(trx->len) > len` bail-out.

---

openwrt-ai (commented):
Reviewed 1 new commit. The three `trx->len` reads are correctly converted and the commit message matches the diff; one inline issue on the `trx_fixup()` hunk where the endian fix looks incomplete.

---

package/system/mtd/src/trx.c
```diff
@@ -97,8 +97,13 @@ trx_fixup(int fd, const char *name)
 		goto err;
 	}
 
+	if (STORE32_LE(trx->len) > len) {
+		fprintf(stderr, "TRX header length exceeds partition size\n");
+		goto err;
+	}
```
openwrt-ai, line 103: The new guard only bounds the upper side. A header claiming `len` smaller than the 12-byte offset of `flag_version` still passes, and then `STORE32_LE(trx->len) - (scan - ptr)` on the next line underflows to a near-`SIZE_MAX` value. That does not fault, which is arguably worse: `crc32buf()` truncates the `size_t` to `int` when it forwards to [`while (--len >= 0)` in crc32.h:14](https://github.com/openwrt/openwrt/blob/4192706bdbe2ceeb47950110bdc8d080f85f7fb8/package/system/mtd/src/crc32.h#L10-L22), the loop runs zero times, and `0xFFFFFFFF` gets written into the on-flash header as if it were a valid CRC instead of the function reporting an error. `trx_check()` already rejects this with `< sizeof(struct trx_header)`; mirroring it here makes the two paths agree:

```suggestion
	if (STORE32_LE(trx->len) < sizeof(struct trx_header) ||
	    STORE32_LE(trx->len) > len) {
		fprintf(stderr, "Invalid TRX header length\n");
		goto err;
	}
```

Separately, and not about byte order: this new failure path `goto err`, which skips both `munmap(ptr, len)` and `close(bfd)` (`err1` would at least close the fd). The magic check above it has the same flaw, so if you prefer keeping the two consistent that is a fair call — just noting it since this adds a second site.

---

package/system/mtd/src/trx.c
```diff
@@ -97,8 +97,13 @@ trx_fixup(int fd, const char *name)
...
+
 	scan = ptr + offsetof(struct trx_header, flag_version);
-	trx->crc32 = crc32buf(scan, trx->len - (scan - ptr));
+	trx->crc32 = STORE32_LE(crc32buf(scan, STORE32_LE(trx->len) - (scan - ptr)));
```
openwrt-ai, line 106: nit: `STORE32_LE()` is a plain textual macro that expands its argument four times ([trx.c:56](https://github.com/openwrt/openwrt/blob/4192706bdbe2ceeb47950110bdc8d080f85f7fb8/package/system/mtd/src/trx.c#L55-L56)), so wrapping the call in it means `crc32buf()` runs four times and the CRC over the whole image is computed four times — on big-endian, i.e. exactly the configuration this commit targets. `htole32()` is a real function/builtin, and `<endian.h>` is already included at [trx.c:29](https://github.com/openwrt/openwrt/blob/4192706bdbe2ceeb47950110bdc8d080f85f7fb8/package/system/mtd/src/trx.c#L29):

```suggestion
	trx->crc32 = htole32(crc32buf(scan, STORE32_LE(trx->len) - (scan - ptr)));
```

(The two pre-existing `STORE32_LE(crc32buf(...))` uses in `mtd_fixtrx()` at lines 244 and 253 have the same property, but they are out of scope here — their length is bounded by `erasesize`, unlike this one.)

---

openwrt-ai (commented):
Re-reviewed after the force-push (`4363205` -> `4192706`); both points from the previous round are addressed — the CRC store is byte-swapped now and there is a length bound before `crc32buf()`. Two follow-ups inline on the new hunk.

## Commit checks

- `4192706` "mtd: trx: fix endian bug in trx_fixup and trx_check" — the message only claims "Wrap all native reads of `trx->len` with `STORE32_LE()`", but the diff also (a) byte-swaps the `trx->crc32` *write* in `trx_fixup()` — a second, independent endian bug, since the old code stored the CRC natively — and (b) adds a new "TRX header length exceeds partition size" sanity check that is not an endian fix at all. Both are worth a line in the body (and in the PR description, which is a copy of it) so the extra hunks are not undocumented.

---

PR #24261, opened 2026-07-17
uboot-rockchip: enable good-state big cluster 2 for rk3582

peterwillcn:
Enable the second big CPU cluster on RK3582 parts where OTP ip-state indicates the cluster is functional.

PR #24464, opened 2026-07-29
uboot-rockchip: enable functional second big cluster via OTP state

peterwillcn:
Remove the hardcoded restriction that forcibly disables the second big CPU cluster on RK3582. The original code unconditionally masked the second big cluster regardless of hardware state, unnecessarily limiting performance on fully functional RK3582 SKUs with clean OTP state. After this change, cluster availability follows the factory-programmed OTP ip_state value:

- Enable the second big cluster if no hardware failure is marked in OTP
- Keep faulty clusters disabled on defective SKUs to ensure stability This change unlocks native dual big-core 

capability, improves multi-task scheduling, and boosts general throughput while retaining full hardware compatibility.

Note this change only relaxes CPU cluster policy. GPU and selected codec cores remain force-failed on RK3582. Unlike CPU cluster2, these IP blocks are officially defeatured for all RK3582 SKUs per silicon specification,rather than being subject to per-unit hardware variation recorded in OTP.

89641cf6-0a15-4050-b947-920243e28a87

1a8e0c8e-9d3d-49b5-8efe-8d50e67a357d

### DIFF
--- a/package/boot/uboot-rockchip/patches/102-rockchip-enable-good-state-big-cluster2-for-rk3582.patch
+++ b/package/boot/uboot-rockchip/patches/102-rockchip-enable-good-state-big-cluster2-for-rk3582.patch
@@ -1,0 +1,26 @@
+From 3526d76b9ffffb680e26198af424eb6a9048bf8e Mon Sep 17 00:00:00 2001
+From: xiao bo <peterwillcn@gmail.com>
+Date: Wed, 29 Jul 2026 09:58:04 +0800
+Subject: [PATCH] rockchip: enable good state big cluster2 for rk3582
+
+Enable the second big CPU cluster on RK3582 parts where OTP ip-state
+indicates the cluster is functional.
+
+Signed-off-by: xiao bo <peterwillcn@gmail.com>
+---
+ arch/arm/mach-rockchip/rk3588/rk3588.c | 4 ----
+ 1 file changed, 4 deletions(-)
+
+--- a/arch/arm/mach-rockchip/rk3588/rk3588.c
++++ b/arch/arm/mach-rockchip/rk3588/rk3588.c
+@@ -400,10 +400,6 @@ int ft_system_setup(void *blob, struct b
+ 	if (ip_state[0] & FAIL_CPU_CLUSTER2)
+ 		ip_state[0] |= FAIL_CPU_CLUSTER2;
+ 
+-	/* policy: always fail one big core cluster on rk3582 */
+-	if (!(ip_state[0] & (FAIL_CPU_CLUSTER1 | FAIL_CPU_CLUSTER2)))
+-		ip_state[0] |= FAIL_CPU_CLUSTER2;
+-
+ 	/* policy: always fail gpu on rk3582 */
+ 	ip_state[1] |= FAIL_GPU;
+ 


peterwillcn:
ping @jonasjelonek

package/boot/uboot-rockchip/patches/102-rockchip-enable-good-state-big-cluster2-for-rk3582.patch
```diff
@@ -0,0 +1,29 @@
...
+
+--- a/arch/arm/mach-rockchip/rk3588/rk3588.c
++++ b/arch/arm/mach-rockchip/rk3588/rk3588.c
+@@ -400,10 +400,6 @@ int ft_system_setup(void *blob, struct bd_info *bd)
+ 	if (ip_state[0] & FAIL_CPU_CLUSTER2)
+ 		ip_state[0] |= FAIL_CPU_CLUSTER2;
+
+-	/* policy: always fail one big core cluster on rk3582 */
+-	if (!(ip_state[0] & (FAIL_CPU_CLUSTER1 | FAIL_CPU_CLUSTER2)))
+-		ip_state[0] |= FAIL_CPU_CLUSTER2;
```
openwrt-ai, line 22: This removes a deliberate upstream policy rather than a bug, and the premise stated in the commit message doesn't hold.

1. The code is not unconditional. The removed hunk is guarded by `if (!(ip_state[0] & (FAIL_CPU_CLUSTER1 | FAIL_CPU_CLUSTER2)))` — it only forces cluster2 off when OTP marks *neither* big cluster as bad. When OTP does mark one bad, upstream already keeps the good cluster. So "unconditionally masked the second big cluster regardless of hardware state" describes a case that doesn't exist; the clean-OTP case is precisely what the policy is for, not an oversight.

2. Upstream documents the intent right above this function: [`ft_system_setup()` comment, rk3588.c:340-347](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/arch/arm/mach-rockchip/rk3588/rk3588.c#L340-L347) — *"Apply same policy as vendor U-Boot for RK3582, i.e. two big cpu cores, the gpu and one vdec/venc core is always failed."* After this patch that comment contradicts the code it describes.

3. A clear `ip_state` bit is the absence of a *recorded* failure, not positive evidence that cluster2 was validated — RK3582 is binned and sold as 4x A55 + 2x A76. Linux's [`rk3582-radxa-e52c.dts:176-182`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/arch/arm64/boot/dts/rockchip/rk3582-radxa-e52c.dts#L176-L182) leaves `cpu_b2`/`cpu_b3` enabled with `vdd_cpu_big1_s0` and relies entirely on this U-Boot fixup, so this flips every RK3582 board built from `U-Boot/generic-rk3588` to 8 cores at cluster2's full OPP table by default.

If there is measured evidence (sustained load at the top cluster2 OPP, thermals) please put it in the commit message. And since this makes OpenWrt's U-Boot intentionally diverge from upstream, the patch header should say so — as written the subject reads like an upstream-bound fix that upstream would reject. An opt-in (per-board or U-Boot env gated) would be a lot safer than changing the default for all RK3582 users.

---

package/boot/uboot-rockchip/patches/102-rockchip-enable-good-state-big-cluster2-for-rk3582.patch
```diff
@@ -0,0 +1,29 @@
...
+-
+ 	/* policy: always fail gpu on rk3582 */
+ 	ip_state[1] |= FAIL_GPU;
```
openwrt-ai, line 25: nit: the stated rationale ("cluster availability follows the factory-programmed OTP ip_state value") is applied only to the CPU clusters. The GPU is still force-failed on this line regardless of OTP, and so are one vdec and one venc core at [rk3588.c:410-416](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/arch/arm/mach-rockchip/rk3588/rk3588.c#L410-L416). If a clean `ip_state` is trustworthy enough to enable cluster2, it is unclear why the same reasoning isn't applied to the GPU/codec blocks — either follow OTP consistently or state in the commit message why only the CPU policy is dropped.

---

package/boot/uboot-rockchip/patches/102-rockchip-enable-good-state-big-cluster2-for-rk3582.patch
```diff
@@ -0,0 +1,29 @@
...
+
+--
+2.55.0
+
```
openwrt-ai: nit: the trailing `git format-patch` version footer doesn't match the rest of this directory — no other patch under `package/boot/uboot-rockchip/patches/` keeps it (`grep -rn '^2\.[0-9]' package/boot/uboot-rockchip/patches/` returns nothing). Drop it:

```suggestion
```

---

openwrt-ai (commented):
## Commit checks

- `a74be63` "uboot-rockchip: enable functional second big cluster via OTP state" — the body's central claim does not match the code being removed. It says the original code "unconditionally masked the second big cluster regardless of hardware state", but the removed hunk is guarded by `if (!(ip_state[0] & (FAIL_CPU_CLUSTER1 | FAIL_CPU_CLUSTER2)))`, i.e. it only forces cluster2 off when OTP marks *neither* big cluster bad. Please reword to describe this as intentionally overriding upstream's RK3582 SKU policy, not as fixing an unconditional mask. See the inline comment for detail.
- `a74be63` — commit body lines run ~95-105 characters; OpenWrt commit messages wrap at ~72.
- `a74be63` — `PKG_RELEASE` in `package/boot/uboot-rockchip/Makefile` is still `1`; a package content change needs a bump so users receive the rebuild (also flagged by the formality bot).

Verified the patch applies cleanly against `arch/arm/mach-rockchip/rk3588/rk3588.c` at U-Boot v2026.07 (`patch -p1 --dry-run` succeeds, hunk offsets and counts are correct), and that it only affects the RK3582 code path — RK3588/RK3588S boards sharing `U-Boot/generic-rk3588` are unaffected because of the `cpu_code == 0x35 0x82` early return.

---

jonasjelonek:
cc @1715173329

1715173329:
> cc @1715173329

https://github.com/openwrt/openwrt/pull/24261#issuecomment-5084152660

peterwillcn:
@1715173329 I get your point. However, openness and freedom are core spirits pursued by the OpenWrt community. This is comparable to early CPU overclocking practices — it represents the geek spirit of exploration. Besides, such tweaks do not harm the interests of chip vendors.

package/boot/uboot-rockchip/patches/102-rockchip-enable-good-state-big-cluster2-for-rk3582.patch
```diff
@@ -0,0 +1,25 @@
...
+indicates the cluster is functional.
+
+Signed-off-by: xiao bo <peterwillcn@gmail.com>
+---
+ arch/arm/mach-rockchip/rk3588/rk3588.c | 4 ----
+ 1 file changed, 4 deletions(-)
+
+--- a/arch/arm/mach-rockchip/rk3588/rk3588.c
++++ b/arch/arm/mach-rockchip/rk3588/rk3588.c
+@@ -400,10 +400,6 @@ int ft_system_setup(void *blob, struct bd_info *bd)
```
openwrt-ai: The hunk header no longer matches the hunk body. Dropping the `git format-patch` footer in this revision also removed the blank line that followed `ip_state[1] |= FAIL_GPU;` — that blank line was the third trailing *context* line of the hunk, not part of the footer. The body is now 9 old / 5 new lines while the header still declares 10 / 6:

```
$ git apply --check 102-rockchip-enable-good-state-big-cluster2-for-rk3582.patch
error: corrupt patch at line 26
```

GNU `patch` tolerates a truncated final hunk (`patch -p1 --dry-run --fuzz=0` still succeeds), so the OpenWrt build will not catch this — but `git apply`, `quilt` and any strict consumer reject the file, and the next `quilt push`/refresh cycle on this package will fail.

Preferred fix is to regenerate the patch with `make package/uboot-rockchip/refresh` so the missing context line is restored (compare [108-01-...patch](``https://github.com/openwrt/openwrt/blob/701519c23848920b6957eb7662061def991a845a/package/boot/uboot-rockchip/patches/108-01-rockchip-mkimage-Split-size_and_off-and-size_and_nim.patch),`` which keeps its blank context line as a single space). If you'd rather fix it by hand, correcting the line counts is sufficient:

```suggestion
@@ -400,9 +400,5 @@ int ft_system_setup(void *blob, struct bd_info *bd)
```

---

openwrt-ai (commented):
## Commit checks

- `701519c` "uboot-rockchip: enable functional second big cluster via OTP state" — the commit body is now wrapped at ~72 columns, thanks. Two items from the previous round are still open, though:
  - The body still asserts that the original code "unconditionally masked the second big cluster regardless of hardware state". The removed hunk is guarded by `if (!(ip_state[0] & (FAIL_CPU_CLUSTER1 | FAIL_CPU_CLUSTER2)))`, so it only forces cluster2 off when OTP marks *neither* big cluster bad — see [the earlier thread](https://github.com/openwrt/openwrt/pull/24464#discussion_r3670901330). Please describe this as deliberately diverging from upstream's RK3582 SKU policy rather than as fixing an unconditional mask.
  - `PKG_RELEASE` in [`package/boot/uboot-rockchip/Makefile:9`](https://github.com/openwrt/openwrt/blob/701519c23848920b6957eb7662061def991a845a/package/boot/uboot-rockchip/Makefile#L9) is still `1` while the package content changed (also flagged by the formality bot).
- nit: `701519c` — missing spaces after punctuation in the body: `RK3582.The`, `state,unnecessarily`, `scheduling,and`.

The newly added note about GPU/codec cores being defeatured per silicon spec addresses the earlier GPU-consistency nit; that one is dropped.

See the inline comment for a regression introduced by this revision — removing the `format-patch` footer also took the hunk's last context line with it, leaving the patch header counts wrong.

---

1715173329:
> This is comparable to early CPU overclocking practices — it represents the geek spirit of exploration. Besides, such tweaks do not harm the interests of chip vendors.

We don't do overclocking here as well.
You are feel to do anything like that, but it doesn't mean you can make other users bear the same risks as you.

So, NAK.

package/boot/uboot-rockchip/patches/102-rockchip-enable-good-state-big-cluster2-for-rk3582.patch
```diff
@@ -0,0 +1,25 @@
...
+@@ -400,9 +400,5 @@ int ft_system_setup(void *blob, struct bd_info *bd)
+ 	if (ip_state[0] & FAIL_CPU_CLUSTER2)
+ 		ip_state[0] |= FAIL_CPU_CLUSTER2;
+
+-	/* policy: always fail one big core cluster on rk3582 */
+-	if (!(ip_state[0] & (FAIL_CPU_CLUSTER1 | FAIL_CPU_CLUSTER2)))
+-		ip_state[0] |= FAIL_CPU_CLUSTER2;
+-
+ 	/* policy: always fail gpu on rk3582 */
+ 	ip_state[1] |= FAIL_GPU;
```
openwrt-ai: Correcting the counts to `9/5` made the header self-consistent, but the hunk still lacks the third trailing context line (the blank line after `ip_state[1] |= FAIL_GPU;`), so GNU `patch` — which is what the build uses (`PATCH:=patch` in [rules.mk:321](https://github.com/openwrt/openwrt/blob/1805fe7c4634715c93a3cb2945a0c9c05572084d/rules.mk#L321), invoked as plain `patch -f -p1` by [scripts/patch-kernel.sh:39](https://github.com/openwrt/openwrt/blob/1805fe7c4634715c93a3cb2945a0c9c05572084d/scripts/patch-kernel.sh#L39)) — can only apply it by fuzzing a context line away:

```
$ patch -p1 --dry-run --fuzz=0 < 102-...patch
Hunk #1 FAILED at 400.
1 out of 1 hunk FAILED
$ patch -p1 --dry-run < 102-...patch
Hunk #1 succeeded at 400 with fuzz 1.
```

Fuzz 1 is under the default limit of 2, so the build won't break, but a fuzz-requiring patch is not what `quilt refresh` produces and it will silently drift the next time the U-Boot version is bumped. Restoring the context line (and going back to `10/6`) applies at `--fuzz=0` with both `patch` and `git apply`. `make package/uboot-rockchip/refresh` is the canonical way to regenerate it; if you prefer to fix it by hand:

```suggestion
@@ -400,10 +400,6 @@ int ft_system_setup(void *blob, struct bd_info *bd)
 	if (ip_state[0] & FAIL_CPU_CLUSTER2)
 		ip_state[0] |= FAIL_CPU_CLUSTER2;
 
-	/* policy: always fail one big core cluster on rk3582 */
-	if (!(ip_state[0] & (FAIL_CPU_CLUSTER1 | FAIL_CPU_CLUSTER2)))
-		ip_state[0] |= FAIL_CPU_CLUSTER2;
-
 	/* policy: always fail gpu on rk3582 */
 	ip_state[1] |= FAIL_GPU;
 
```

Note the final line of that block is a single space (a blank context line) — if GitHub's suggestion editor strips it, add it back manually, as in [108-01-...patch](``https://github.com/openwrt/openwrt/blob/1805fe7c4634715c93a3cb2945a0c9c05572084d/package/boot/uboot-rockchip/patches/108-01-rockchip-mkimage-Split-size_and_off-and-size_and_nim.patch)`` which keeps its blank context lines that way.

---

openwrt-ai (commented):
## Commit checks

- `1805fe7` "uboot-rockchip: enable functional second big cluster via OTP state" — three items from the previous rounds are still open:
  - The body still claims the original code "unconditionally masked the second big cluster regardless of hardware state". The removed hunk is guarded by `if (!(ip_state[0] & (FAIL_CPU_CLUSTER1 | FAIL_CPU_CLUSTER2)))`, so it only forces cluster2 off when OTP marks *neither* big cluster bad — see [the earlier thread](https://github.com/openwrt/openwrt/pull/24464#discussion_r3670901330). Please describe this as deliberately diverging from upstream's RK3582 SKU policy rather than as removing an unconditional mask.
  - `PKG_RELEASE` in [`package/boot/uboot-rockchip/Makefile:9`](https://github.com/openwrt/openwrt/blob/1805fe7c4634715c93a3cb2945a0c9c05572084d/package/boot/uboot-rockchip/Makefile#L9) is still `1` while the package content changed (also flagged by the formality bot).
  - nit: missing spaces after punctuation in the body: `RK3582.The`, `state,unnecessarily`, `scheduling,and`.

The hunk-count fix in this revision does make the patch acceptable to `git apply` (`git apply --check` now passes, verified against `arch/arm/mach-rockchip/rk3588/rk3588.c` in u-boot/u-boot at `ece349a`), but it is still not fuzz-free for GNU `patch`. See the inline comment.

---

peterwillcn:
> You are feel to do anything like that, but it doesn't mean you can make other users bear the same risks as you.

@1715173329  It is common knowledge that anyone flashing firmware not provided by the device manufacturer takes full responsibility for all related risks.

As documented on [openwrt.org](https://openwrt.org), OpenWrt software is provided AS-IS and without any warranty. It is common knowledge that end users who flash non-vendor/modified firmware voluntarily assume all associated risks. Users who pursue extra performance are fully aware they are stepping outside official vendor support.

openwrt-ai (commented):
## Commit checks

- `e02adc0` "uboot-rockchip: enable functional second big cluster via OTP state" — the patch-format regression is fixed in this revision: the hunk header is back to `@@ -400,10 +400,6 @@` with the third trailing context line (the blank line after `ip_state[1] |= FAIL_GPU;`) restored as a single space, and the `@@` function context is now quilt-style truncated. Verified against `arch/arm/mach-rockchip/rk3588/rk3588.c` in u-boot/u-boot at `ece349a` (v2026.07, the version [`package/boot/uboot-rockchip/Makefile:8`](https://github.com/openwrt/openwrt/blob/e02adc0d81652ba30327bea375e920ddfbc7ad8f/package/boot/uboot-rockchip/Makefile#L8) pins): both `patch -p1 --dry-run --fuzz=0` and `git apply --check` now succeed. That item is dropped.

  The commit message is byte-identical to `1805fe7`, so the three message-level items from the previous rounds are still open:
  - The body still claims the original code "unconditionally masked the second big cluster regardless of hardware state". The removed hunk is guarded by `if (!(ip_state[0] & (FAIL_CPU_CLUSTER1 | FAIL_CPU_CLUSTER2)))`, so it only forces cluster2 off when OTP marks *neither* big cluster bad — see [the earlier thread](https://github.com/openwrt/openwrt/pull/24464#discussion_r3670901330). Please describe this as deliberately diverging from upstream's RK3582 SKU policy rather than as removing an unconditional mask.
  - `PKG_RELEASE` in [`package/boot/uboot-rockchip/Makefile:9`](https://github.com/openwrt/openwrt/blob/e02adc0d81652ba30327bea375e920ddfbc7ad8f/package/boot/uboot-rockchip/Makefile#L9) is still `1` while the package content changed (also flagged by the formality bot).
  - nit: missing spaces after punctuation in the body: `RK3582.The`, `state,unnecessarily`, `scheduling,and`.

No new inline issues in this revision, and CI on `e02adc0` is green.

---

PR #24465, opened 2026-07-29
airoha: add support for VSOL V2901Q-A and V2902A-S

BYZYB:
## Summary

This PR adds OpenWrt support for the VSOL V2901Q-A and V2902A-S ONUs
based on the Airoha AN7583ST SoC.

The devices share the same bootloader and flash layout, but use separate
Linux device trees because GDM3 is connected differently.

| Model | Ethernet |
| --- | --- |
| [V2901Q-A](https://www.vsolcn.com/product/10g-pon-2-5gbe-onu-v2901q-a) | 1× 2.5GbE through an AN8811 PHY |
| [V2902A-S](https://www.vsolcn.com/product/10g-pon-1ge-1sfp-hgu-onu-v2902a-s) | 1× internal 1GbE port and 1× SFP cage through USXGMII |

Stock firmware supports both V2901Q-A and V2902A-S, while this
PR provides separate OpenWrt device definitions.

## Commit series

1. ~~uboot-airoha: add extra SPI-NAND support to AN7583 BL2~~
   - ~~Adds new SPI-NAND types used by VSOL to the AN7583 BL2 flash table:~~
   - See [c42522b432](https://github.com/BYZYB/openwrt/commit/c42522b432) for history and the [related PR](https://github.com/Ansuel/atf-airoha/pull/3) of `atf-airoha`.

2. `uboot-airoha: add support for VSOL AN7583 ONUs`
   - Adds a shared U-Boot configuration for both devices.
   - Enables SPI-NAND, UBI, Ethernet, LEDs and firmware upgrade commands.
   - Adds `tcboot.bin` and `tclinux.bin` boot support.

3. `airoha: add support for VSOL V2901Q-A and V2902A-S`
   - Adds the device trees, network and LED configuration.
   - Adds image generation and sysupgrade support.
   - Adds automatic initialization of the independent `rootfs_data`
     UBI volume.

## Hardware

- SoC: Airoha AN7583ST
  - Dual-core ARM Cortex-A53
  - Up to 1.2 GHz
- RAM: 512 MiB DDR3L
- Flash: SPI-NAND
  - Devices are available with 64, 128 or 256 MiB flash
  - This PR currently supports the 128 MiB layout
  - `W25N01GV` is widely used by VSOL and supported by OpenWrt, as `W25N01KV` is for newer products with fewer amount.
- PON:
  - EN7573 XE/XG/XGS-PON
  - EN7572 XG-PON
- Ethernet:
  - V2901Q-A: AN8811 2.5GbE PHY
  - V2902A-S: internal 1GbE PHY and SFP cage
- Buttons: RESET
- UART: 115200 8N1
- Power: 12V / 1A
- WLAN: N/A

## Boot flow

```text
AN7583 Boot ROM
  -> BL2 from tcboot.bin
  -> BL31 and U-Boot from the fip UBI volume
  -> Linux FIT from the fit UBI volume
  -> Linux
```

The generated `tclinux.bin` contains:

```text
UBI volume 0: fip
  - BL31
  - U-Boot
UBI volume 1: fit
  - LZMA-compressed Linux kernel
  - AN7583 device tree
  - squashfs root filesystem
```

`tclinux` and `tclinux_slave` contain identical images. U-Boot tries
`tclinux` first and falls back to `tclinux_slave`.

## Flash layout

The partition offsets remain aligned with the stock layout to simplify
installation and returning to stock firmware.

| Offset | Size | Partition | Contents |
| --- | ---: | --- | --- |
| `0x00000000` | 512 KiB | `bootloader` | `tcboot.bin` containing BL2 |
| `0x00080000` | 32 MiB | `tclinux` | UBI image containing `fip` and `fit` |
| `0x02080000` | 32 MiB | `tclinux_slave` | Identical backup image |
| `0x04080000` | 40 MiB | `rootfs_data` | Independent OpenWrt overlay UBI |
| `0x06880000` | 1 MiB | `factory` | Factory data and MAC address |
| `0x06c80000` | 3.5 MiB | `art` | Calibration data |

The stock `rootfs_data`, `framework` and `framework_slave` partitions
occupy the same 40 MiB region. OpenWrt merges them into a single
`rootfs_data` partition.

**Do not erase factory/art partitions.** Otherwise the factory parameters (MAC, SN, etc.) and calibration data will lost, a factory repare is needed. It's recommended to make backups of these partitions.

## Images

The following images are generated for each device:

- `tcboot.bin`
  - Written directly to the `bootloader` partition.
- `tclinux.bin`
  - Written directly to both `tclinux` and `tclinux_slave`.
- `squashfs-sysupgrade.bin`
  - Used for normal OpenWrt sysupgrade.
- `initramfs-uImage.itb`
  - Used for recovery or temporary booting.

The image names follow the stock firmware convention where practical. Detailed installation and back to stock procedures are included in the message of commit 3. The shared image builders `Build/airoha-image-tcboot` and `Build/airoha-image-tclinux` will make it easier for most Airoha SDK (OpenWrt 21.02) based devices to bring up support of OpenWrt mainline.

## Sysupgrade

Sysupgrade writes the new UBI image in this order:

1. `tclinux_slave`
2. `tclinux`

The independent `rootfs_data` partition is not erased or reformatted
during sysupgrade, so the existing overlay remains available. Both slots receive the same firmware image. The bootflag-based switching from
the stock firmware is intentionally not implemented.

## Testing

Tested on VSOL V2901Q-A/V2902A-S hardware with a 128 MiB W25N01KV SPI-NAND and 256/512 MiB DDR.

## Known limitations

- XPON/OAM/OMCI functionality is not supported by mainline OpenWrt.
- Only the fixed 128 MiB SPI-NAND layout is currently supported. Automatic 64/128/256 MiB flash layout detection from stock firmware is not implemented. However, only 128 MiB version is in mass production and it should not be a problem for a long while.
- Stock (Airoha SDK) bootflag-based A/B switching is not implemented.
- U-Boot network access is available only through the internal switch. No plan to import PHY driver in U-Boot to reduce downstream changes.
- SFP module compatibility on V2902A-S requires broader testing.

### DIFF
--- a/package/boot/arm-trusted-firmware-airoha/Makefile
+++ b/package/boot/arm-trusted-firmware-airoha/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_VERSION:=2.10
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 PKG_HASH:=88215a62291b9ba87da8e50b077741103cdc08fb6c9e1ebd34dfaace746d3201
 
 PKG_MAINTAINER:=Christian Marangi <ansuelsmth@gmail.com>
@@ -38,7 +38,9 @@ define Trusted-Firmware-A/an7583-bl2
   BUILD_SUBTARGET:=an7583
   SOC:=AN7583
   BUILD_DEVICES:=airoha_an7583-evb airoha_an7583-evb-emmc \
-	nokia_xg-040g-mf-ubi
+	nokia_xg-040g-mf-ubi \
+	vsol_v2901q-a \
+	vsol_v2902a-s
 endef
 
 define Trusted-Firmware-A/bl31/Default
@@ -59,7 +61,9 @@ define Trusted-Firmware-A/an7583-bl31
   BUILD_SUBTARGET:=an7583
   SOC:=AN7583
   BUILD_DEVICES:=airoha_an7583-evb airoha_an7583-evb-emmc \
-	nokia_xg-040g-mf-ubi
+	nokia_xg-040g-mf-ubi \
+	vsol_v2901q-a \
+	vsol_v2902a-s
 endef
 
 TFA_TARGETS:= \
--- a/package/boot/uboot-airoha/patches/402-airoha-add-vsol-an7583.patch
+++ b/package/boot/uboot-airoha/patches/402-airoha-add-vsol-an7583.patch
@@ -1,0 +1,295 @@
+From 88af859f94c01c24336337a37754620243c134e4 Mon Sep 17 00:00:00 2001
+From: Yubo Zhang <zyb_1998@outlook.com>
+Date: Tue, 14 Jul 2026 11:29:43 +0800
+Subject: [PATCH] an7583: add support for VSOL AN7583 ONUs
+
+Add a shared U-Boot config for the VSOL V2901Q-A and V2902A-S devices
+based on the Airoha AN7583 SoC. Both models use the same bootloader
+config and partition layout, and keep aligned with the stock firmware.
+
+Partition and image
+-------------------
+Partitions
+- bootloader (tcboot.bin): payload/crc needed by BL2 but not BL31
+- tclinux (tclinux.bin): UBI image containing the fip/fit volumes
+- tclinux_slave (tclinux.bin): Identical to tclinux and used for backup
+- rootfs_data: Independent UBI partition used for OpenWrt overlay
+
+tcboot.bin
+- 0x00000-0x007ff: padding with 0x0
+- 0x00800-0x1ffff: BL2 FIP, padding with 0xff
+- 0x20000-0x7bffb: BL31 + U-Boot FIP padding with 0xff, normally unused
+- 0x7bffc-0x7bfff: CRC checksum from 0x00000 to 0x7bffc
+- 0x7c000-0x7ffff: padding with 0xff
+
+tclinux.bin
+UBI volume 0 (fip):
+- BL31 + U-Boot FIP
+UBI volume 1 (fit):
+- Linux FIT
+  - LZMA Linux kernel
+  - AN7583 DTB
+  - squashfs rootfs
+
+Since the name and offset of partitions and images are designed to be the
+same as stock firmware, switching to OpenWrt or back to stock will be
+less difficult.
+
+Boot flow
+---------
+AN7583 Boot ROM -> BL2 -> FIP -> BL31 -> U-Boot -> FIT -> Linux
+|      tcboot.bin       |             tclinux.bin             |
+
+If booting from the tclinux partition fails, tclinux_slave will be used
+for backup. These A/B partitions could also be manually selected by
+command `run boot_tclinux` and `run boot_tclinux_slave` in U-Boot.
+
+The tcboot.bin is shared by both models. The initramfs and tclinux images
+contain model-specific DTBs and must not be mixed. Rename the images for
+the target model to these names before using tftpboot or upgrade commands:
+- initramfs-uImage.itb
+- tcboot.bin
+- tclinux.bin
+
+To upgrade bootloader and firmware in U-Boot, use the command:
+`run upgrade_tcboot` and `run upgrade_tclinux`.
+
+Known limits
+------------
+- U-Boot network access is available only through the internal switch.
+- Persistent U-Boot environment (saveenv) is not supported.
+- The automatic adaptation of 64/128/256 MiB flash like stock firmware is
+  not implemented. Only the fixed 128M partition layout is supported now.
+- The bootflag based A/B partition switch logic is not implemented.
+
+Signed-off-by: Yubo Zhang <zyb_1998@outlook.com>
+---
+ arch/arm/dts/an7583-vsol.dts  | 106 ++++++++++++++++++++++++++++++++++++++++++
+ configs/an7583_vsol_defconfig | 87 ++++++++++++++++++++++++++++++++++++
+ defenvs/an7583_vsol_env       | 16 +++++++
+ 3 files changed, 209 insertions(+)
+ create mode 100644 arch/arm/dts/an7583-vsol.dts
+ create mode 100644 configs/an7583_vsol_defconfig
+ create mode 100644 defenvs/an7583_vsol_env
+
+--- /dev/null
++++ b/arch/arm/dts/an7583-vsol.dts
+@@ -0,0 +1,106 @@
++// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
++/dts-v1/;
++
++/* Bootloader installs ATF here */
++/memreserve/ 0x80000000 0x200000;
++
++#include <dt-bindings/gpio/gpio.h>
++#include "an7583.dtsi"
++
++/ {
++	model = "VSOL V2901Q-A / V2902A-S";
++	compatible = "vsol,v2901q-a", "vsol,v2902a-s", "airoha,an7583", "airoha,en7583";
++
++	aliases {
++		serial0 = &uart1;
++	};
++
++	chosen {
++		bootargs = "console=ttyS0,115200 earlycon";
++		stdout-path = "serial0:115200n8";
++		linux,usable-memory-range = <0x0 0x80200000 0x0 0x1fe00000>;
++	};
++
++	memory@80000000 {
++		device_type = "memory";
++		reg = <0x0 0x80000000 0x0 0x20000000>;
++	};
++
++	gpio-leds {
++		compatible = "gpio-leds";
++
++		led-system {
++			label = "system";
++			gpios = <&an7583_pinctrl 7 GPIO_ACTIVE_LOW>;
++			default-state = "on";
++		};
++
++		led-los {
++			label = "los";
++			gpios = <&an7583_pinctrl 27 GPIO_ACTIVE_LOW>;
++			default-state = "on";
++		};
++	};
++};
++
++&snfi {
++	status = "okay";
++};
++
++&spi_nand {
++	partitions {
++		compatible = "fixed-partitions";
++		#address-cells = <1>;
++		#size-cells = <1>;
++
++		bootloader@0 {
++			label = "bootloader";
++			reg = <0x00000000 0x00080000>;
++		};
++
++		tclinux@80000 {
++			label = "tclinux";
++			reg = <0x00080000 0x02000000>;
++		};
++
++		tclinux_slave@2080000 {
++			label = "tclinux_slave";
++			reg = <0x02080000 0x02000000>;
++		};
++
++		rootfs_data@4080000 {
++			label = "rootfs_data";
++			reg = <0x04080000 0x02800000>;
++		};
++
++		factory@6880000 {
++			label = "factory";
++			reg = <0x06880000 0x00100000>;
++		};
++
++		art@6c80000 {
++			label = "art";
++			reg = <0x06c80000 0x00380000>;
++		};
++	};
++};
++
++&i2c0 {
++	status = "okay";
++};
++
++&i2c1 {
++	status = "okay";
++};
++
++&eth {
++	status = "okay";
++};
++
++&gdm1 {
++	status = "okay";
++};
++
++&switch {
++	status = "okay";
++};
+--- /dev/null
++++ b/configs/an7583_vsol_defconfig
+@@ -0,0 +1,87 @@
++CONFIG_ARM=y
++CONFIG_POSITION_INDEPENDENT=y
++CONFIG_ARCH_AIROHA=y
++CONFIG_TARGET_AN7583=y
++CONFIG_TEXT_BASE=0x81E00000
++CONFIG_SYS_MALLOC_F_LEN=0x4000
++CONFIG_NR_DRAM_BANKS=1
++CONFIG_ENV_SIZE=0x4000
++CONFIG_DM_GPIO=y
++CONFIG_DEFAULT_DEVICE_TREE="an7583-vsol"
++CONFIG_SYS_LOAD_ADDR=0x81800000
++CONFIG_BUILD_TARGET="u-boot.bin"
++# CONFIG_EFI_LOADER is not set
++CONFIG_FIT=y
++CONFIG_FIT_VERBOSE=y
++CONFIG_OF_ENV_SETUP=y
++CONFIG_BOOTDELAY=3
++CONFIG_DEFAULT_FDT_FILE="an7583-vsol"
++CONFIG_SYS_PBSIZE=1049
++CONFIG_SYS_CONSOLE_IS_IN_ENV=y
++# CONFIG_DISPLAY_BOARDINFO is not set
++CONFIG_HUSH_PARSER=y
++CONFIG_SYS_PROMPT="AN7583> "
++CONFIG_SYS_MAXARGS=8
++CONFIG_CMD_BOOTZ=y
++CONFIG_CMD_BOOTMENU=y
++# CONFIG_CMD_ELF is not set
++# CONFIG_CMD_XIMG is not set
++CONFIG_CMD_BIND=y
++CONFIG_CMD_GPIO=y
++CONFIG_CMD_MMC=y
++CONFIG_CMD_MTD=y
++CONFIG_CMD_SF_TEST=y
++CONFIG_CMD_SPI=y
++# CONFIG_CMD_SETEXPR is not set
++CONFIG_CMD_MII=y
++CONFIG_CMD_MDIO=y
++CONFIG_CMD_PING=y
++CONFIG_CMD_EXT4=y
++CONFIG_CMD_FAT=y
++CONFIG_CMD_FS_GENERIC=y
++CONFIG_CMD_MTDPARTS=y
++CONFIG_CMD_LOG=y
++CONFIG_CMD_UBI=y
++CONFIG_ENV_OVERWRITE=y
++# CONFIG_ENV_IS_IN_MTD is not set
++CONFIG_ENV_RELOC_GD_ENV_ADDR=y
++CONFIG_ENV_USE_DEFAULT_ENV_TEXT_FILE=y
++CONFIG_ENV_DEFAULT_ENV_TEXT_FILE="defenvs/an7583_vsol_env"
++CONFIG_ENV_VARS_UBOOT_RUNTIME_CONFIG=y
++CONFIG_NET_RANDOM_ETHADDR=y
++CONFIG_SYS_RX_ETH_BUFFER=8
++CONFIG_REGMAP=y
++CONFIG_SYSCON=y
++CONFIG_CLK=y
++CONFIG_DMA=y
++CONFIG_LED=y
++CONFIG_LED_GPIO=y
++CONFIG_MMC_HS200_SUPPORT=y
++CONFIG_MMC_MTK=y
++CONFIG_MTD=y
++CONFIG_DM_MTD=y
++CONFIG_MTD_SPI_NAND=y
++CONFIG_DM_SPI_FLASH=y
++CONFIG_SPI_FLASH_EON=y
++CONFIG_SPI_FLASH_GIGADEVICE=y
++CONFIG_SPI_FLASH_ISSI=y
++CONFIG_SPI_FLASH_MACRONIX=y
++CONFIG_SPI_FLASH_SPANSION=y
++CONFIG_SPI_FLASH_STMICRO=y
++CONFIG_SPI_FLASH_WINBOND=y
++CONFIG_SPI_FLASH_MTD=y
++CONFIG_DM_MDIO=y
++CONFIG_AIROHA_ETH=y
++CONFIG_PHY=y
++CONFIG_PINCTRL=y
++CONFIG_PINCTRL_AIROHA_AN7583=y
++CONFIG_PINCONF=y
++CONFIG_POWER_DOMAIN=y
++CONFIG_DM_REGULATOR=y
++CONFIG_DM_REGULATOR_FIXED=y
++CONFIG_RAM=y
++CONFIG_DM_SERIAL=y
++CONFIG_SYS_NS16550=y
++CONFIG_SPI=y
++CONFIG_DM_SPI=y
++CONFIG_AIROHA_SNFI_SPI=y
++CONFIG_SHA512=y
+--- /dev/null
++++ b/defenvs/an7583_vsol_env
+@@ -0,0 +1,16 @@
++loadaddr=0x81800000
++ipaddr=192.168.1.1
++serverip=192.168.1.10
++bootfile=initramfs-uImage.itb
++tcbootfile=tcboot.bin
++tclinuxfile=tclinux.bin
++bootconf=config@1
++rootdisk_prop=rootdisk
++fdt_fixup=fdt get value rootdisk_phandle /chosen ${rootdisk_prop}; fdt set /chosen rootdisk "<${rootdisk_phandle}>"
++bootargs=console=ttyS0,115200 earlycon ubi.mtd=tclinux,0,0,0 ubi.mtd=rootfs_data,0,0,1 root=/dev/fit0 rootwait
++boot_tclinux=setenv rootdisk_prop rootdisk; setenv bootargs "console=ttyS0,115200 earlycon ubi.mtd=tclinux,0,0,0 ubi.mtd=rootfs_data,0,0,1 root=/dev/fit0 rootwait"; ubi part tclinux && ubi read $loadaddr fit && bootm $loadaddr#$bootconf
++boot_tclinux_slave=setenv rootdisk_prop rootdisk-slave; setenv bootargs "console=ttyS0,115200 earlycon ubi.mtd=tclinux_slave,0,0,0 ubi.mtd=rootfs_data,0,0,1 root=/dev/fit0 rootwait"; ubi part tclinux_slave && ubi read $loadaddr fit && bootm $loadaddr#$bootconf
++boot_tftp=setenv bootargs "console=ttyS0,115200 earlycon"; tftpboot $loadaddr $bootfile && bootm $loadaddr#$bootconf
++upgrade_tcboot=tftpboot $loadaddr $tcbootfile && mtd erase bootloader && mtd write bootloader $loadaddr 0 $filesize
++upgrade_tclinux=tftpboot $loadaddr $tclinuxfile && mtd erase tclinux && mtd write tclinux $loadaddr 0 $filesize && mtd erase tclinux_slave && mtd write tclinux_slave $loadaddr 0 $filesize
++bootcmd=run boot_tclinux; run boot_tclinux_slave
+--
+2.51.0
--- a/package/boot/uboot-airoha/patches/403-airoha-add-vsol-an7583.patch
+++ b/package/boot/uboot-airoha/patches/403-airoha-add-vsol-an7583.patch
@@ -1,0 +1,295 @@
+From 88af859f94c01c24336337a37754620243c134e4 Mon Sep 17 00:00:00 2001
+From: Yubo Zhang <zyb_1998@outlook.com>
+Date: Tue, 14 Jul 2026 11:29:43 +0800
+Subject: [PATCH] an7583: add support for VSOL AN7583 ONUs
+
+Add a shared U-Boot config for the VSOL V2901Q-A and V2902A-S devices
+based on the Airoha AN7583 SoC. Both models use the same bootloader
+config and partition layout, and keep aligned with the stock firmware.
+
+Partition and image
+-------------------
+Partitions
+- bootloader (tcboot.bin): payload/crc needed by BL2 but not BL31
+- tclinux (tclinux.bin): UBI image containing the fip/fit volumes
+- tclinux_slave (tclinux.bin): Identical to tclinux and used for backup
+- rootfs_data: Independent UBI partition used for OpenWrt overlay
+
+tcboot.bin
+- 0x00000-0x007ff: padding with 0x0
+- 0x00800-0x1ffff: BL2 FIP, padding with 0xff
+- 0x20000-0x7bffb: BL31 + U-Boot FIP padding with 0xff, normally unused
+- 0x7bffc-0x7bfff: CRC checksum from 0x00000 to 0x7bffc
+- 0x7c000-0x7ffff: padding with 0xff
+
+tclinux.bin
+UBI volume 0 (fip):
+- BL31 + U-Boot FIP
+UBI volume 1 (fit):
+- Linux FIT
+  - LZMA Linux kernel
+  - AN7583 DTB
+  - squashfs rootfs
+
+Since the name and offset of partitions and images are designed to be the
+same as stock firmware, switching to OpenWrt or back to stock will be
+less difficult.
+
+Boot flow
+---------
+AN7583 Boot ROM -> BL2 -> FIP -> BL31 -> U-Boot -> FIT -> Linux
+|      tcboot.bin       |             tclinux.bin             |
+
+If booting from the tclinux partition fails, tclinux_slave will be used
+for backup. These A/B partitions could also be manually selected by
+command `run boot_tclinux` and `run boot_tclinux_slave` in U-Boot.
+
+The tcboot.bin is shared by both models. The initramfs and tclinux images
+contain model-specific DTBs and must not be mixed. Rename the images for
+the target model to these names before using tftpboot or upgrade commands:
+- initramfs-uImage.itb
+- tcboot.bin
+- tclinux.bin
+
+To upgrade bootloader and firmware in U-Boot, use the command:
+`run upgrade_tcboot` and `run upgrade_tclinux`.
+
+Known limits
+------------
+- U-Boot network access is available only through the internal switch.
+- Persistent U-Boot environment (saveenv) is not supported.
+- The automatic adaptation of 64/128/256 MiB flash like stock firmware is
+  not implemented. Only the fixed 128M partition layout is supported now.
+- The bootflag based A/B partition switch logic is not implemented.
+
+Signed-off-by: Yubo Zhang <zyb_1998@outlook.com>
+---
+ arch/arm/dts/an7583-vsol.dts  | 106 ++++++++++++++++++++++++++++++++++++++++++
+ configs/an7583_vsol_defconfig | 87 ++++++++++++++++++++++++++++++++++++
+ defenvs/an7583_vsol_env       | 16 +++++++
+ 3 files changed, 209 insertions(+)
+ create mode 100644 arch/arm/dts/an7583-vsol.dts
+ create mode 100644 configs/an7583_vsol_defconfig
+ create mode 100644 defenvs/an7583_vsol_env
+
+--- /dev/null
++++ b/arch/arm/dts/an7583-vsol.dts
+@@ -0,0 +1,106 @@
++// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
++/dts-v1/;
++
++/* Bootloader installs ATF here */
++/memreserve/ 0x80000000 0x200000;
++
++#include <dt-bindings/gpio/gpio.h>
++#include "an7583.dtsi"
++
++/ {
++	model = "VSOL V2901Q-A / V2902A-S";
++	compatible = "vsol,v2901q-a", "vsol,v2902a-s", "airoha,an7583", "airoha,en7583";
++
++	aliases {
++		serial0 = &uart1;
++	};
++
++	chosen {
++		bootargs = "console=ttyS0,115200 earlycon";
++		stdout-path = "serial0:115200n8";
++		linux,usable-memory-range = <0x0 0x80200000 0x0 0x1fe00000>;
++	};
++
++	memory@80000000 {
++		device_type = "memory";
++		reg = <0x0 0x80000000 0x0 0x20000000>;
++	};
++
++	gpio-leds {
++		compatible = "gpio-leds";
++
++		led-system {
++			label = "system";
++			gpios = <&an7583_pinctrl 7 GPIO_ACTIVE_LOW>;
++			default-state = "on";
++		};
++
++		led-los {
++			label = "los";
++			gpios = <&an7583_pinctrl 27 GPIO_ACTIVE_LOW>;
++			default-state = "on";
++		};
++	};
++};
++
++&snfi {
++	status = "okay";
++};
++
++&spi_nand {
++	partitions {
++		compatible = "fixed-partitions";
++		#address-cells = <1>;
++		#size-cells = <1>;
++
++		bootloader@0 {
++			label = "bootloader";
++			reg = <0x00000000 0x00080000>;
++		};
++
++		tclinux@80000 {
++			label = "tclinux";
++			reg = <0x00080000 0x02000000>;
++		};
++
++		tclinux_slave@2080000 {
++			label = "tclinux_slave";
++			reg = <0x02080000 0x02000000>;
++		};
++
++		rootfs_data@4080000 {
++			label = "rootfs_data";
++			reg = <0x04080000 0x02800000>;
++		};
++
++		factory@6880000 {
++			label = "factory";
++			reg = <0x06880000 0x00100000>;
++		};
++
++		art@6c80000 {
++			label = "art";
++			reg = <0x06c80000 0x00380000>;
++		};
++	};
++};
++
++&i2c0 {
++	status = "okay";
++};
++
++&i2c1 {
++	status = "okay";
++};
++
++&eth {
++	status = "okay";
++};
++
++&gdm1 {
++	status = "okay";
++};
++
++&switch {
++	status = "okay";
++};
+--- /dev/null
++++ b/configs/an7583_vsol_defconfig
+@@ -0,0 +1,87 @@
++CONFIG_ARM=y
++CONFIG_POSITION_INDEPENDENT=y
++CONFIG_ARCH_AIROHA=y
++CONFIG_TARGET_AN7583=y
++CONFIG_TEXT_BASE=0x81E00000
++CONFIG_SYS_MALLOC_F_LEN=0x4000
++CONFIG_NR_DRAM_BANKS=1
++CONFIG_ENV_SIZE=0x4000
++CONFIG_DM_GPIO=y
++CONFIG_DEFAULT_DEVICE_TREE="an7583-vsol"
++CONFIG_SYS_LOAD_ADDR=0x81800000
++CONFIG_BUILD_TARGET="u-boot.bin"
++# CONFIG_EFI_LOADER is not set
++CONFIG_FIT=y
++CONFIG_FIT_VERBOSE=y
++CONFIG_OF_ENV_SETUP=y
++CONFIG_BOOTDELAY=3
++CONFIG_DEFAULT_FDT_FILE="an7583-vsol"
++CONFIG_SYS_PBSIZE=1049
++CONFIG_SYS_CONSOLE_IS_IN_ENV=y
++# CONFIG_DISPLAY_BOARDINFO is not set
++CONFIG_HUSH_PARSER=y
++CONFIG_SYS_PROMPT="AN7583> "
++CONFIG_SYS_MAXARGS=8
++CONFIG_CMD_BOOTZ=y
++CONFIG_CMD_BOOTMENU=y
++# CONFIG_CMD_ELF is not set
++# CONFIG_CMD_XIMG is not set
++CONFIG_CMD_BIND=y
++CONFIG_CMD_GPIO=y
++CONFIG_CMD_MMC=y
++CONFIG_CMD_MTD=y
++CONFIG_CMD_SF_TEST=y
++CONFIG_CMD_SPI=y
++# CONFIG_CMD_SETEXPR is not set
++CONFIG_CMD_MII=y
++CONFIG_CMD_MDIO=y
++CONFIG_CMD_PING=y
++CONFIG_CMD_EXT4=y
++CONFIG_CMD_FAT=y
++CONFIG_CMD_FS_GENERIC=y
++CONFIG_CMD_MTDPARTS=y
++CONFIG_CMD_LOG=y
++CONFIG_CMD_UBI=y
++CONFIG_ENV_OVERWRITE=y
++# CONFIG_ENV_IS_IN_MTD is not set
++CONFIG_ENV_RELOC_GD_ENV_ADDR=y
++CONFIG_ENV_USE_DEFAULT_ENV_TEXT_FILE=y
++CONFIG_ENV_DEFAULT_ENV_TEXT_FILE="defenvs/an7583_vsol_env"
++CONFIG_ENV_VARS_UBOOT_RUNTIME_CONFIG=y
++CONFIG_NET_RANDOM_ETHADDR=y
++CONFIG_SYS_RX_ETH_BUFFER=8
++CONFIG_REGMAP=y
++CONFIG_SYSCON=y
++CONFIG_CLK=y
++CONFIG_DMA=y
++CONFIG_LED=y
++CONFIG_LED_GPIO=y
++CONFIG_MMC_HS200_SUPPORT=y
++CONFIG_MMC_MTK=y
++CONFIG_MTD=y
++CONFIG_DM_MTD=y
++CONFIG_MTD_SPI_NAND=y
++CONFIG_DM_SPI_FLASH=y
++CONFIG_SPI_FLASH_EON=y
++CONFIG_SPI_FLASH_GIGADEVICE=y
++CONFIG_SPI_FLASH_ISSI=y
++CONFIG_SPI_FLASH_MACRONIX=y
++CONFIG_SPI_FLASH_SPANSION=y
++CONFIG_SPI_FLASH_STMICRO=y
++CONFIG_SPI_FLASH_WINBOND=y
++CONFIG_SPI_FLASH_MTD=y
++CONFIG_DM_MDIO=y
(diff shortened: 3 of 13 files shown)


target/linux/airoha/image/an7583.mk
```diff
@@ -44,3 +62,38 @@ define Device/nokia_xg-040g-mf
...
+  IMAGE/sysupgrade.bin = append-kernel | fit lzma $$(KDIR)/image-$$(DEVICE_DTS).dtb \
+	external-static-with-rootfs | an7583-image-tclinux vsol | check-size | append-metadata
+  ARTIFACT/tcboot.bin := an7583-image-tcboot vsol
+  ARTIFACT/tclinux.bin := append-image squashfs-sysupgrade.bin
+  ARTIFACTS := tcboot.bin tclinux.bin
+endef
+
+define Device/vsol_v2901q-a
+  $(call Device/vsol_an7583-common)
+  DEVICE_MODEL := VSOL V2901Q-A
```
openwrt-ai: The vendor is prepended automatically — [`DEVICE_TITLE = $(DEVICE_VENDOR) $(DEVICE_MODEL)`](https://github.com/openwrt/openwrt/blob/53f30826da3585d3ce20a01ee548af7ce80baec0/include/image.mk#L438) — so with `DEVICE_VENDOR := VSOL` this renders as "VSOL VSOL V2901Q-A" in the profile list and the JSON overview. Every other device in this file keeps the vendor out of `DEVICE_MODEL` (`Nokia` / `XG-040G-MF`).

Same applies to line 95 (`DEVICE_MODEL := VSOL V2902A-S`).

```suggestion
  DEVICE_MODEL := V2901Q-A
```

---

target/linux/airoha/image/an7583.mk
```diff
@@ -1,3 +1,21 @@
+define Build/an7583-image-tcboot
+  dd if=/dev/zero bs=524288 count=1 | tr '\000' '\377' > "$@"
+  dd if=/dev/zero of="$@" bs=2048 count=1 conv=notrunc
+  dd if=$(STAGING_DIR_IMAGE)/an7583_$1-bl2.fip of="$@" bs=2048 seek=1 conv=notrunc
+  dd if=$(STAGING_DIR_IMAGE)/an7583_$1-bl31-u-boot.fip of="$@" bs=131072 seek=1 conv=notrunc
+  head -c 507900 "$@" | gzip -n -c | tail -c 8 | head -c 4 | \
+	$(STAGING_DIR_HOST)/bin/xorimage -p ff -x | \
+	dd of="$@" bs=1 seek=507900 conv=notrunc
```
openwrt-ai: Neither `dd` bounds its payload, and `conv=notrunc` will happily write past the slot: a `bl2.fip` larger than `0x1f800` overwrites the BL31 region at `0x20000`, and a `bl31-u-boot.fip` larger than `0x5bffc` grows the file past 512 KiB and past the CRC word. The CRC is then still written at 507900 over data that is now part of the payload, so the result is a silently corrupt `tcboot.bin` — the one image whose failure bricks the device. The BL31+U-Boot slot is only ~368 KiB, so this is not a theoretical margin.

Worth asserting both FIP sizes (and the final 512 KiB size) before writing the CRC, so a too-large U-Boot fails the build instead of producing a bad bootloader image.

---

target/linux/airoha/dts/an7583-vsol-common.dtsi
```diff
@@ -0,0 +1,231 @@
...
+			function = LED_FUNCTION_STATUS;
+			gpios = <&an7583_pinctrl 27 GPIO_ACTIVE_LOW>;
+		};
+
+		led-sfp {
+			label = "sfp";
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_LAN;
+			gpios = <&an7583_pinctrl 29 GPIO_ACTIVE_LOW>;
+		};
```
openwrt-ai: The `color`/`function` properties on all five nodes are dead code: [`led_parse_fwnode_props()` returns as soon as `label` is present](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/leds/led-core.c#L485-L490), so the sysfs names stay the legacy `system`/`pon`/`internet`/`los`/`sfp`. `.github/llm-review-rules.md` lists `label = ...` as the deprecated form, and the sibling an7583 board [an7583-nokia_xg-040g-mf-common.dtsi](https://github.com/openwrt/openwrt/blob/53f30826da3585d3ce20a01ee548af7ce80baec0/target/linux/airoha/dts/an7583-nokia_xg-040g-mf-common.dtsi#L22-L74) uses `color` + `function` only.

Dropping `label` yields `green:power`, `green:pon`, `green:wan`, `red:los`, `green:lan`. Note `LED_FUNCTION_STATUS` was used for both `pon` and `los`, which describes neither LED, so a plain string is the better fit per the rules file. The suggestion also adds a `led_system` phandle for the missing `led-boot`/`led-running` aliases (see the comment on the `aliases` node), and `01_leds` needs the new names.

```suggestion
		led_system: led-system {
			color = <LED_COLOR_ID_GREEN>;
			function = LED_FUNCTION_POWER;
			gpios = <&an7583_pinctrl 7 GPIO_ACTIVE_LOW>;
		};

		led-pon {
			color = <LED_COLOR_ID_GREEN>;
			function = "pon";
			gpios = <&an7583_pinctrl 12 GPIO_ACTIVE_LOW>;
		};

		led-internet {
			color = <LED_COLOR_ID_GREEN>;
			function = LED_FUNCTION_WAN;
			gpios = <&an7583_pinctrl 26 GPIO_ACTIVE_LOW>;
		};

		led-los {
			color = <LED_COLOR_ID_RED>;
			function = "los";
			gpios = <&an7583_pinctrl 27 GPIO_ACTIVE_LOW>;
		};

		led-sfp {
			color = <LED_COLOR_ID_GREEN>;
			function = LED_FUNCTION_LAN;
			gpios = <&an7583_pinctrl 29 GPIO_ACTIVE_LOW>;
		};
```

---

target/linux/airoha/dts/an7583-vsol-common.dtsi
```diff
@@ -0,0 +1,231 @@
...
+
+#include <dt-bindings/leds/common.h>
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include "an7583.dtsi"
+
+/ {
+	aliases {
+		serial0 = &uart1;
+	};
```
openwrt-ai, line 16: No `led-boot` / `led-failsafe` / `led-running` / `led-upgrade` aliases, so `/etc/diag.sh` has no LED to drive and there is no boot, failsafe or sysupgrade indication on a device whose serial console is deliberately hard to reach. The sibling an7583 board declares all four in [an758x-nokia_xg-040g-common.dtsi](https://github.com/openwrt/openwrt/blob/53f30826da3585d3ce20a01ee548af7ce80baec0/target/linux/airoha/dts/an758x-nokia_xg-040g-common.dtsi#L7-L10). `ucidef_set_led_default "system" ... "1"` in `01_leds` only pins the LED on; it does not substitute for the aliases.

The suggestion depends on the `led_system:` phandle added by the comment on the `leds` node.

```suggestion
	aliases {
		led-boot = &led_system;
		led-failsafe = &led_system;
		led-running = &led_system;
		led-upgrade = &led_system;

		serial0 = &uart1;
	};
```

---

target/linux/airoha/dts/an7583-vsol-common.dtsi
```diff
@@ -0,0 +1,231 @@
...
+
+&an7583_pinctrl {
+	gpio-ranges = <&an7583_pinctrl 0 2 53>;
+
+	mdio0_pins: mdio0-pins {
+		conf {
+			pins = "mdio_0";
+			output-high;
+		};
+	};
```
openwrt-ai, line 83: `mdio0_pins` carries only the `conf` node — the `mux` that actually selects the MDIO function on the pins is missing. Both existing an7583 boards include it: [an7583-evb.dts:31-40](https://github.com/openwrt/openwrt/blob/53f30826da3585d3ce20a01ee548af7ce80baec0/target/linux/airoha/dts/an7583-evb.dts#L31-L40) and [an7583-nokia_xg-040g-mf-common.dtsi:112-121](``https://github.com/openwrt/openwrt/blob/53f30826da3585d3ce20a01ee548af7ce80baec0/target/linux/airoha/dts/an7583-nokia_xg-040g-mf-common.dtsi#L112-L121).`` As written, `pinctrl-0 = <&mdio0_pins>` in `an7583-vsol-v2901q-a.dts` only applies `output-high` and leaves the EN8811 MDIO bus depending on whatever mux the bootloader happened to leave behind.

```suggestion
	mdio0_pins: mdio0-pins {
		mux {
			function = "mdio";
			groups = "mdio";
		};
		conf {
			pins = "mdio_0";
			output-high;
		};
	};
```

---

target/linux/airoha/dts/an7583-vsol-common.dtsi
```diff
@@ -0,0 +1,231 @@
...
+		mux {
+			function = "phy1_led0";
+			pins = "gpio1";
+		};
+	};
+};
+
+&snfi {
+	status = "okay";
+};
```
openwrt-ai, line 118: Both existing an7583 boards mux the SPI pins for the SNFI controller — [an7583-evb.dts:71-73](https://github.com/openwrt/openwrt/blob/53f30826da3585d3ce20a01ee548af7ce80baec0/target/linux/airoha/dts/an7583-evb.dts#L71-L73) and [an7583-nokia_xg-040g-mf-common.dtsi:173-175](``https://github.com/openwrt/openwrt/blob/53f30826da3585d3ce20a01ee548af7ce80baec0/target/linux/airoha/dts/an7583-nokia_xg-040g-mf-common.dtsi#L173-L175``) — but no `spi_pins` group is defined here and `&snfi` gets no `pinctrl-*`. Is that intentional (i.e. these pins are not muxable on this board, or BL2 already leaves them correct), or an omission? Flash access obviously works on the tested unit, so I can't tell from the diff whether the pinmux is simply redundant here.

---

target/linux/airoha/dts/an7583-vsol-common.dtsi
```diff
@@ -0,0 +1,231 @@
...
+		bootargs = "console=ttyS0,115200 earlycon ubi.mtd=tclinux,0,0,0 ubi.mtd=rootfs_data,0,0,1 root=/dev/fit0 rootwait";
+		rootdisk = <&tclinux_fit>;
+		rootdisk-slave = <&tclinux_slave_fit>;
+		stdout-path = "serial0:115200n8";
+	};
+
+	memory@80000000 {
+		device_type = "memory";
+		reg = <0x0 0x80000000 0x0 0x20000000>;
+	};
```
openwrt-ai, line 28: This hardcodes 512 MiB, but the commit message documents "RAM 256/512 MiB DDR3" and neither the U-Boot DTS in the series nor the `fdt_fixup` env command patches the `memory` node. A 256 MiB unit would therefore get a DT claiming twice the RAM it has. Unlike the flash-layout limitation, this one isn't called out in the "Limitations" section — is 256 MiB actually handled somewhere (SoC-level DRAM fixup?), or should the commit message state that only 512 MiB units are supported?

---

target/linux/airoha/dts/an7583-vsol-common.dtsi
```diff
@@ -0,0 +1,231 @@
...
+	status = "okay";
+	pinctrl-names = "gbe-led";
+	pinctrl-0 = <&gswp1_led0_pins>;
+};
+
+&gsw_phy1_led0 {
+	status = "okay";
+	color = <LED_COLOR_ID_GREEN>;
+	active-low;
+};
```
openwrt-ai: nit: The commit message says the V2901Q-A "does not expose the 1GbE port", and `02_network` accordingly gives it only `eth1` — yet `gsw_port1`, `gsw_phy1` and its `phy1_led0` pinmux are enabled here for both models, so V2901Q-A ends up with an unconfigured `lan1` netdev and a GE LED driven by a port that has no jack. Should these move to `an7583-vsol-v2902a-s.dts` (with `gsw_port1`/`gsw_phy1` disabled for the V2901Q-A), the way the other unused switch ports are handled below?

---

target/linux/airoha/an7583/base-files/etc/board.d/01_leds
```diff
@@ -14,6 +14,11 @@ nokia,xg-040g-mf)
 	ucidef_set_led_netdev "lan3" "lan3" "mt7530-0:0b:green:lan-3" "lan3" "link tx rx"
 	ucidef_set_led_netdev "lan4" "lan4" "mt7530-0:0c:green:lan-4" "lan4" "link tx rx"
 	;;
+vsol,v2901q-a | vsol,v2902a-s)
+	ucidef_set_led_default "system" "system" "system" "1"
+	ucidef_set_led_timer "los" "los" "los" "1000" "1000"
+	ucidef_set_led_netdev "sfp" "sfp" "sfp" "eth1" "link tx rx"
```
openwrt-ai: These sysfs names come from the deprecated `label` properties in `an7583-vsol-common.dtsi`; if those are dropped in favour of `color`/`function` (see the comment there) the names become `green:power` / `red:los` / `green:lan`, matching the existing `nokia,xg-040g-mf` entries above.

nit: also, `"sfp"` names a GPIO LED that both models share, but the V2901Q-A has no SFP cage — the commit message calls this LED "LAN" on the V2901Q-A and "10G" on the V2902A-S. `green:lan` covers both; `sfp` is misleading for one of the two boards.

---

package/boot/uboot-airoha/patches/402-airoha-add-vsol-an7583.patch
openwrt-ai, line 1: nit: the `999-` prefix is already taken by `999-airoha-add-gemtek-w1700k.patch`. It still applies (filename sort breaks the tie), but the numbering no longer says anything about ordering. The other board additions in this package use their own slots — `400-add-nokia-valyrian.patch`, `401-add-nokia-xg-040g-md.patch` — so `402-` would fit the series better.

nit: the patch subject is `uboot-airoha: add support for VSOL AN7583 ONUs`, i.e. the OpenWrt package name, on a patch whose diff lands inside the U-Boot tree. The neighbouring board patch uses a U-Boot-scoped prefix (`an7581: add uboot chainloader for w1700k`); `airoha:` or `an7583:` would be consistent here.

---

package/boot/uboot-airoha/patches/999-airoha-add-vsol-an7583.patch
```diff
@@ -0,0 +1,288 @@
...
++CONFIG_SHA512=y
+--- /dev/null
++++ b/defenvs/an7583_vsol_env
+@@ -0,0 +1,16 @@
++loadaddr=0x81800000
++ipaddr=192.168.1.1
++serverip=192.168.1.10
++bootfile=openwrt-airoha-an7583-vsol_v2902a-s-initramfs-uImage.itb
++tcbootfile=openwrt-airoha-an7583-vsol_v2902a-s-tcboot.bin
++tclinuxfile=openwrt-airoha-an7583-vsol_v2902a-s-tclinux.bin
```
openwrt-ai: nit: this one U-Boot config is shared by both models (`BUILD_DEVICES:=vsol_v2901q-a vsol_v2902a-s`), but the three TFTP filenames hardcode `vsol_v2902a-s`. On a V2901Q-A, `run boot_tftp` / `upgrade_tcboot` / `upgrade_tclinux` — the exact commands the "Return to stock" section of commit 3 tells users to run — fetch the wrong device's images, and the V2901Q-A images build under `...-vsol_v2901q-a-...`. `CONFIG_SYS_PROMPT="V2902A-S> "` on line 84 of the defconfig has the same one-model bias.

Since the DTS is shared there is no per-board default to key off, but it is worth at least noting in the commit message that V2901Q-A owners must `setenv` these three variables first.

---

openwrt-ai (commented):
Reviewed 3 new commits.

All three commit messages match their diffs, and commit 1's binary description checks out byte-for-byte: `flash_table_length` `0x3ad` -> `0x41a` at `0x3808` is the first changed byte, everything from `0x380c` to the LZMA table at `0x1ce4f` is untouched, the table decompresses from `0x49d4`/56 entries to `0x5300`/63 entries with `flash_name_off = 0xa2f` and `flash_oob_off = 0x1304`, the 7 added chips carry exactly the documented IDs/sizes/OOB sizes at indices 8, 43 and 56-60 with the pre-existing entries in unchanged order, and the trailing CRC at `0x1d269` is `0xa74a7329`. Thanks for documenting it that precisely — it made the prebuilt-blob change auditable.

Inline comments below. The `DEVICE_MODEL` one produces a visibly wrong device title; the rest are DTS convention/parity items and questions.

---

target/linux/airoha/dts/an7583-vsol-common.dtsi
```diff
@@ -0,0 +1,230 @@
...
+	chosen {
+		bootargs = "console=ttyS0,115200 earlycon ubi.mtd=tclinux,0,0,0 ubi.mtd=rootfs_data,0,0,1 root=/dev/fit0 rootwait";
+		rootdisk = <&tclinux_fit>;
+		rootdisk-slave = <&tclinux_slave_fit>;
+		stdout-path = "serial0:115200n8";
+	};
+
+	memory@80000000 {
+		device_type = "memory";
+		reg = <0x0 0x80000000 0x0 0x20000000>;
```
openwrt-ai, line 27: nit: commit 3 now says `RAM 512 MiB DDR3L`, but the PR description still says `RAM: 256/512 MiB DDR3L` — the two disagree about the same fact. Since this node hardcodes `0x20000000` and nothing in the series patches the `memory` node (the U-Boot DTS in [402-airoha-add-vsol-an7583.patch](https://github.com/openwrt/openwrt/blob/f7de779a86a816b75e2922f6e3acd8e6a2c2010f/package/boot/uboot-airoha/patches/402-airoha-add-vsol-an7583.patch) carries the same fixed `reg`, and `fdt_fixup` only rewrites `/chosen/rootdisk`), a 256 MiB unit would boot with a DT claiming 512 MiB.

Either bring the PR description in line with the commit message, or — if 256 MiB units do exist in the field — add it next to the flash-layout entry in the `Limitations` section the way the 64/128/256 MiB flash restriction is called out.

---

target/linux/airoha/dts/an7583-vsol-v2901q-a.dts
```diff
@@ -0,0 +1,55 @@
...
+			#address-cells = <1>;
+			#size-cells = <0>;
+
+			led@0 {
+				reg = <0>;
+				function = LED_FUNCTION_LAN;
+				color = <LED_COLOR_ID_GREEN>;
+				function-enumerator = <0>;
+				default-state = "keep";
+			};
```
openwrt-ai: This gives the V2901Q-A two green LAN LEDs: this EN8811 PHY LED, and the shared GPIO 29 `led-lan` in [an7583-vsol-common.dtsi:68-72](https://github.com/openwrt/openwrt/blob/f7de779a86a816b75e2922f6e3acd8e6a2c2010f/target/linux/airoha/dts/an7583-vsol-common.dtsi#L68-L72) which `01_leds` binds to the netdev trigger on `eth1` for both models. The commit message lists only five LEDs for this model (`SYS, PON, LOS, WAN, LAN`), so one of the two has no physical LED behind it. Is the EN8811 LED output wired to the same LAN LED as GPIO 29 (in which case the two triggers fight over it), or is this node just carried over from the sibling board?

nit: `function-enumerator = <0>` also deviates from the rest of the family — [an7583.dtsi](https://github.com/openwrt/openwrt/blob/f7de779a86a816b75e2922f6e3acd8e6a2c2010f/target/linux/airoha/dts/an7583.dtsi#L772-L785) hands out `1` to `gsw_phy1_led0`/`led1` and the sibling EN8811 in [an7583-nokia_xg-040g-mf-common.dtsi:190-197](``https://github.com/openwrt/openwrt/blob/f7de779a86a816b75e2922f6e3acd8e6a2c2010f/target/linux/airoha/dts/an7583-nokia_xg-040g-mf-common.dtsi#L190-L197``) uses `1` as well.

---

target/linux/airoha/base-files/lib/preinit/75_rootfs_prepare
```diff
@@ -0,0 +1,22 @@
+# SPDX-License-Identifier: GPL-2.0-or-later OR BSD-2-Clause
+
+rootfs_prepare() {
+	case "$(board_name)" in
+	vsol,v2901q-a | vsol,v2902a-s)
```
openwrt-ai, line 5: nit: this hook is board-specific but lands in `target/linux/airoha/base-files/`, so it ships on an7581 and en7523 too, where the `case` can never match. Every other per-board file this PR adds lives under the subtarget (`target/linux/airoha/an7583/base-files/etc/board.d/01_leds`, `.../02_network`, `.../lib/upgrade/platform.sh`); the only existing file in the target-wide `preinit` dir, `04_set_netdev_label`, is genuinely target-wide. Moving this to `target/linux/airoha/an7583/base-files/lib/preinit/75_rootfs_prepare` would match.

---

openwrt-ai (commented):
Reviewed the 3 amended commits (`53f3082` → `f7de779`, rebased onto `3189ea8`). All three commit messages still match their diffs.

Re-checking the previous round, most of it is addressed:

- `DEVICE_MODEL` no longer repeats the vendor.
- The `dd` slots in `tcboot.bin` are now bounded, and the bounds are right: `129024` = `0x20000 - 0x800` (BL2 FIP) and `376828` = `0x7bffc - 0x20000` (BL31+U-Boot FIP), matching the layout documented in commit 2. Both build steps also moved to the shared `image/Makefile` and are now parameterised on the U-Boot `BUILD_VARIANT` instead of hardcoding `an7583_`.
- LEDs dropped the deprecated `label` properties and gained `led-boot`/`led-failsafe`/`led-running`/`led-upgrade` aliases; `01_leds` follows with the new `red:los` / `green:lan` names.
- `gsw_port1`, `gsw_phy1` and `gsw_phy1_led0` are disabled on the V2901Q-A.
- The U-Boot patch is renumbered to `402-` (next to `400-add-nokia-valyrian` / `401-add-nokia-xg-040g-md`) with a U-Boot-scoped `an7583:` subject, and the defenv TFTP filenames plus `CONFIG_SYS_PROMPT` are no longer V2902A-S-specific.

Two threads from the previous review are still open and unaddressed, so I have not re-posted them:

- [`mdio0_pins` has no `mux` node](https://github.com/openwrt/openwrt/pull/24465#discussion_r3680007922) — both other an7583 boards pair the `conf` node with `function = "mdio"; groups = "mdio";`.
- [no `spi_pins` group / no `pinctrl-*` on `&snfi`](https://github.com/openwrt/openwrt/pull/24465#discussion_r3680008628) — question, since flash obviously works on the tested unit.

I also diffed `configs/an7583_vsol_defconfig` against the in-tree `an7583_evb_defconfig` (after patches 509/522/532). The two AN7583-specific symbols it omits, `CONFIG_MDIO_AIROHA` and `CONFIG_PCS_AIROHA_AN7583`, are neither `default y`, but the VSOL U-Boot DTS only enables `&eth`, `&gdm1` and `&switch`, and `gdm1` has no `pcs` phandle — so the omissions look consistent with the documented "U-Boot network access is available only through the internal switch". Flagging only so it is a deliberate call rather than an oversight.

CI is green on `f7de779`. Three inline comments below, all convention/parity questions.

---

target/linux/airoha/base-files/lib/preinit/75_rootfs_prepare
BYZYB: Another PR will be opened recently, to add support for AN7581 based VSOL V2902A device (may also support AN7552 based devices in the future). All these devices will share the same logic of `rootfs_data` partition, so `75_rootfs_prepare` is added as a shared script for airoha targets. Otherwise, the script in an7583 will be duplicated with an7581.
In the latest force-push update, similar changes about `Build/airoha-image-tcboot` and `Build/airoha-image-tclinux` in `target/linux/airoha/image/Makefile` are for the same reason.

target/linux/airoha/dts/an7583-vsol-common.dtsi
BYZYB: The mass production hasn't started yet for 256 MiB units, so the description of `RAM: 256 MiB` is removed. The stock firmware still supports the adaption of different RAM size with no dts changes on test boards.

target/linux/airoha/dts/an7583-vsol-v2901q-a.dts
BYZYB: Rechecked the hardware GPIO defines of V2901Q-A/V2902A-S and updated the commit with force-push. The `led-sfp` has been moved to `an7583-vsol-v2902a-s.dts`.

V2901Q-A
- CPU GPIO 29: Unused
- AN8811 GPIO 4: LAN (LED1)

V2902A-S
- CPU GPIO 29: SFP

target/linux/airoha/dts/an7583-vsol-v2902a-s.dts
```diff
@@ -0,0 +1,37 @@
...
+#include "an7583-vsol-common.dtsi"
+
+/ {
+	model = "VSOL V2902A-S";
+	compatible = "vsol,v2902a-s", "airoha,an7583", "airoha,en7583";
+
+	leds {
+		led-sfp {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_LAN;
```
openwrt-ai: The composed sysfs name here is `green:lan`, not `green:sfp`, so the new `ucidef_set_led_netdev "sfp" "sfp" "green:sfp" "eth1" ...` line added to `01_leds` in this push matches no LED and the SFP link/activity trigger is silently never applied on the V2902A-S.

With no `label`, [`led_compose_name()`](https://github.com/gregkh/linux/blob/db83c95b2b7c6e14fddbdc23dcfb5c4c47bd918e/drivers/leds/led-core.c#L544-L560) builds the name from `color` + `function`, and `LED_FUNCTION_LAN` is `"lan"` ([common.h:92](https://github.com/gregkh/linux/blob/db83c95b2b7c6e14fddbdc23dcfb5c4c47bd918e/include/dt-bindings/leds/common.h#L92)). There is no `LED_FUNCTION_SFP` constant, and `.github/llm-review-rules.md` explicitly allows falling back to a plain `function = "<func>"` string in that case — which this DTS already does for `led-pon` and `led-los`. Using `"sfp"` also keeps the DT node name, the `01_leds` name and the ucidef id consistent:

```suggestion
		function = "sfp";
```

(The alternative fix — changing `01_leds` back to `green:lan` — works too, but then the node name `led-sfp` and the `"sfp"` ucidef id no longer describe the LED name.)

nit: commit 3 calls this LED `10G (29)` in the hardware list while the DTS/`01_leds` call it `sfp`; worth aligning the three so the LED can be identified from the commit message.

---

openwrt-ai (commented):
Reviewed the force-push (`f7de779` → `da3c541`, rebased onto `ce16dd8`). Commits 1 and 2 are byte-identical to the previous round (same patch-id); only commit 3 changed, and its message change (adding GPIO numbers to the LED list) matches the DTS.

Everything from the previous round is addressed:

- `mdio0_pins` now carries the `mux { function = "mdio"; groups = "mdio"; }` node, matching `an7583-evb.dts` and `an7583-nokia_xg-040g-mf-common.dtsi`.
- `spi_pins` is defined and applied to `&snfi` via `pinctrl-names`/`pinctrl-0`, same as both existing an7583 boards.
- The GPIO 29 LED moved out of the shared dtsi into `an7583-vsol-v2902a-s.dts`, and the EN8811 LED on the V2901Q-A is now `led@1` / `reg = <1>` / `function-enumerator = <1>`, matching the "AN8811 GPIO 4: LAN (LED1)" mapping you described and the sibling EN8811 node in the Nokia dtsi. `EN8811H_LED_COUNT` is 3 in [air_en8811h.c](https://github.com/gregkh/linux/blob/db83c95b2b7c6e14fddbdc23dcfb5c4c47bd918e/drivers/net/phy/air_en8811h.c#L939-L941), so index 1 is valid.
- `01_leds` and `02_network` are split per model, and the V2901Q-A no longer claims a `lan` netdev LED.

Thanks also for the notes on the shared `75_rootfs_prepare` / `Build/airoha-image-*` helpers being groundwork for the upcoming AN7581 V2902A — that answers the earlier placement question.

One new issue introduced by this push, inline below: the V2902A-S SFP LED's DT name and the name `01_leds` looks for no longer agree.

CI on `da3c541` is green or still running; nothing failed.

---

target/linux/airoha/dts/an7583-vsol-v2902a-s.dts
BYZYB: The name `sfp` is the same as stock firmware and PCB layout, `10G` is only shown on the front shell of V2902A-S.

target/linux/airoha/image/an7583.mk
```diff
@@ -65,3 +65,38 @@ define Device/nokia_xg-040g-mf
   DEVICE_PACKAGES := kmod-phy-airoha-en8811h
 endef
 TARGET_DEVICES += nokia_xg-040g-mf
+
+define Device/vsol_an7583-common
+  DEVICE_VENDOR := VSOL
+  DEVICE_DTS_CONFIG := config@1
+  DEVICE_PACKAGES := fitblk kmod-i2c-an7581
```
openwrt-ai: `kmod-i2c-an7581` no longer exists. It was deleted from `target/linux/airoha/modules.mk` by [`9dccdd708d0` "airoha: an7581/7583: build-in I2C support"](https://github.com/openwrt/openwrt/commit/9dccdd708d0c341a1ff4ddb0073dcf4f5189abe9) (7 Aug), which also stripped it from every `DEVICE_PACKAGES` in `an7581.mk`/`an7583.mk` and built the driver in instead — [`CONFIG_I2C_MT7621=y` in an7583/config-6.18:170](https://github.com/openwrt/openwrt/blob/f34f2932e92b1cc22a33afd9611d4e4c1852fc15/target/linux/airoha/an7583/config-6.18#L170). `git grep kmod-i2c-an7581` at this head matches only this line, so image build for both new devices will fail on a missing package. The rebase onto `486b4a4` picked up that commit but not this reference.

I2C is already built in, so nothing needs to replace it:

```suggestion
  DEVICE_PACKAGES := fitblk
```

---

package/boot/arm-trusted-firmware-airoha/Makefile
```diff
@@ -37,7 +37,7 @@ define Trusted-Firmware-A/an7583-bl2
   $(call Trusted-Firmware-A/bl2/Default)
   BUILD_SUBTARGET:=an7583
   SOC:=AN7583
-  BUILD_DEVICES:=airoha_an7583-evb airoha_an7583-evb-emmc
+  BUILD_DEVICES:=airoha_an7583-evb airoha_an7583-evb-emmc vsol_v2901q-a vsol_v2902a-s
```
openwrt-ai, line 40: Enabling `an7583-bl2` here means `tcboot.bin` now carries a BL2 whose SPI-NAND table is generated from the atf-airoha sources — [`airoha-atf-gen-flash-table` run in `Package/trusted-firmware-a/bl2/install`](https://github.com/openwrt/openwrt/blob/f34f2932e92b1cc22a33afd9611d4e4c1852fc15/package/boot/arm-trusted-firmware-airoha/Makefile#L169-L184) — rather than the prebuilt `an7583-bl2.bin` blob. That is also why the first commit of this series (`uboot-airoha: add extra SPI-NAND support to AN7583 BL2`, which patched DS35Q12C / DS35Q1GA / DS35Q1GD / F35SQA001G / F35SQB512M / FM25S01A / W25N01KV into the blob's table) could be dropped.

Have you confirmed the table produced from `atf-airoha` @ `80519be` actually contains those chips? Commit 3 lists the flash on these units as "W25N01GV (mostly), W25N01KV, F35SQA001G", and the tested unit is a W25N01KV — if the upstream table only has W25N01GV, a W25N01KV/F35SQA001G unit flashed with this `tcboot.bin` would fail in BL2 with no console. I can't inspect the atf-airoha tree from here, so flagging rather than asserting; if the chips are covered upstream, it would be worth saying so in the PR description next to the struck-through commit 1, since dropping that commit is otherwise unexplained.

---

package/boot/arm-trusted-firmware-airoha/Makefile
```diff
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_VERSION:=2.10
-PKG_RELEASE:=1
+PKG_RELEASE:=2
```
openwrt-ai, line 4: nit: undocumented hunk — the commit message only says "Add two VSOL devices to an7583 BUILD_DEVICES". `BUILD_DEVICES` does not change the built artifact, only which device selections pull the package in, and the equivalent in-tree change [`486b4a4` "arm-trusted-firmware-airoha: sync an7581-bl31 BUILD_DEVICES with bl2"](https://github.com/openwrt/openwrt/commit/486b4a4b3ede480635a185234fc624c969426f13) — which is the direct parent of this commit — left `PKG_RELEASE` alone. Either drop the bump or mention it in the message.

```suggestion
PKG_RELEASE:=1
```

---

package/boot/uboot-airoha/patches/402-airoha-add-vsol-an7583.patch
```diff
@@ -0,0 +1,295 @@
...
+Boot flow
+---------
+AN7583 Boot ROM -> BL2 -> FIP -> BL31 -> U-Boot -> FIT -> Linux
+|      tcboot.bin       |             tclinux.bin             |
+
+If booting from the tclinux partition fails, tclinux_slave will be used
+for backup. These A/B partitions could also be manually selected by
+command `run boot_tclinux` and `run boot_tclinux_slave` in U-Boot.
+
+The tcboot.img is shared by both models. The initramfs and tclinux images
```
openwrt-ai: nit: this push fixed `tcboot.img` → `tcboot.bin` in the OpenWrt commit message of commit 1, but the same sentence inside the patch header was left as `tcboot.img`, so the two copies of the text now disagree. No image named `tcboot.img` is produced anywhere in the series.

```suggestion
The tcboot.bin is shared by both models. The initramfs and tclinux images
```

---

target/linux/airoha/dts/an7583-vsol-v2902a-s.dts
```diff
@@ -0,0 +1,55 @@
...
+/ {
+	model = "VSOL V2902A-S";
+	compatible = "vsol,v2902a-s", "airoha,an7583", "airoha,en7583";
+
+	sfp: sfp {
+		compatible = "sff,sfp";
+		i2c-bus = <&i2c1>;
+		mod-def0-gpios = <&an7583_pinctrl 2 GPIO_ACTIVE_LOW>;
+		tx-disable-gpios = <&an7583_pinctrl 34 GPIO_ACTIVE_HIGH>;
+	};
```
openwrt-ai, line 15: No `maximum-power-milliwatt`, so the cage is limited to the 1 W default ([sfp.c:3165-3175](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/sfp.c#L3165-L3175)) and any module declaring more than that is rejected outright when it does not implement address 0xa2 — ["Host does not support %u.%uW modules" → -EINVAL](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/sfp.c#L2089-L2100). Most XGS-PON sticks and copper 10G modules are class 3+ (>1 W), so this is a plausible concrete cause of the "SFP module compatibility for V2902A-S is not confirmed" limitation. The an7581 sibling with an SFP cage sets it — [an7581-nokia-valyrian.dts:134](https://github.com/openwrt/openwrt/blob/f34f2932e92b1cc22a33afd9611d4e4c1852fc15/target/linux/airoha/dts/an7581-nokia-valyrian.dts#L134). Is the 1 W default what the cage's power budget actually is here, or is the property just missing?

Same question for `los-gpios` / `tx-fault-gpios`, which valyrian also wires up — are those signals simply not routed on this board?

---

openwrt-ai (commented):
Reviewed the rebase onto `486b4a4` (`da3c541` → `f34f293`). The series changed shape: the prebuilt-blob commit 1 is gone, a new `arm-trusted-firmware-airoha: add VSOL AN7583 devices` commit takes its place, and the image build steps now generate the FIPs themselves instead of consuming staged ones. All three commit messages match their diffs.

Everything from the previous round is addressed:

- `led-sfp` is now `function = "sfp"`, so the composed name `green:sfp` matches the `ucidef_set_led_netdev "sfp" "sfp" "green:sfp" "eth1"` line in `01_leds`.
- The SoC-level `sfp1`/`sfp2` nodes were removed in-tree by [`d22a30dd0f9`](https://github.com/openwrt/openwrt/commit/d22a30dd0f918dbb9580614659ad2498f3fc6f5e); the board now declares its own `sfp` node, matching `an7581-nokia-valyrian.dts`.
- `gsw_port1` / `gsw_phy1` / `gsw_phy1_led0` moved out of the shared dtsi into `an7583-vsol-v2902a-s.dts`, and the redundant `gsw_port2..4` / `gsw_phy2..4` disables were dropped — those nodes are already `status = "disabled"` in `an7583.dtsi`.

I also checked the reworked image steps against the layout documented in commit 1. `fiptool create --tb-fw an7583-bl2.bin` and `--soc-fw an7583-bl31.lzma --nt-fw an7583_vsol-u-boot.lzma` line up with what the new `arm-trusted-firmware-airoha` package stages (`$(BUILD_SUBTARGET)-bl2.bin`, `$(BUILD_SUBTARGET)-bl31.lzma`) and with `$(BUILD_VARIANT)-u-boot.lzma` from `uboot-airoha`. The size bounds are still right — `129024` = `0x20000 - 0x800`, `376828` = `0x7bffc - 0x20000` — and the seeks (`bs=2048 seek=1`, `bs=131072 seek=1`) and CRC offset (`507900` = `0x7bffc`) all still match. Moving the intermediates to `$@.bl2.fip` / `$@.fip` instead of `STAGING_DIR_IMAGE` is an improvement over the in-tree `an7583-preloader` pattern. Dropping the now-dead `BL2_IMAGE` / `BL31_IMAGE` from `U-Boot/an7583_vsol` is correct — nothing reads them any more.

One real problem inline (`kmod-i2c-an7581` was deleted from the tree by a commit this rebase pulled in), one question about the dropped BL2 flash-table commit, plus three nits.

---

package/boot/arm-trusted-firmware-airoha/Makefile
BYZYB: Another PR in [atf-airoha](https://github.com/Ansuel/atf-airoha/pull/3) is opened to support extra nand flash types, instead of adding patch to OpenWrt sources.

target/linux/airoha/dts/an7583-vsol-v2902a-s.dts
BYZYB: No other gpios of SFP except 2/34 is connected on V2902A-S. The max power of SFP is 3W, confirmed with the datasheet.

openwrt-ai (commented):
Reviewed the force-push (`f34f293` → `b788d58`). Base is unchanged at `486b4a4`; commit 2 is byte-identical to the previous round (same patch-id) and only its message changed, commits 1 and 3 changed by three lines in total. All three commit messages match their diffs.

Everything from the previous round is addressed:

- `DEVICE_PACKAGES := fitblk` — the stale `kmod-i2c-an7581` reference is gone. `git grep kmod-i2c-an7581` at this head returns nothing, so both new devices build against packages that still exist, and I2C stays built-in via `CONFIG_I2C_MT7621=y`.
- `maximum-power-milliwatt = <3000>;` on the V2902A-S `sfp` node — same value and same position (after `tx-disable-gpios`) as [an7581-nokia-valyrian.dts:134](https://github.com/openwrt/openwrt/blob/b788d58e93e36f6f9e93138f4be5726580d83249/target/linux/airoha/dts/an7581-nokia-valyrian.dts#L134). With 3 W declared, class 2/3 modules are no longer rejected by the 1 W default in `sfp_module_parse_power()`. Thanks for the note that only GPIO 2 and 34 are routed — that settles `los-gpios`/`tx-fault-gpios` being absent by design rather than by omission.
- `tcboot.img` → `tcboot.bin` inside the U-Boot patch header; `git grep tcboot.img` at this head is empty, so the OpenWrt commit message and the patch header now agree, and no image name in the series is unproduced.
- Commit 2's message now documents the `PKG_RELEASE:=1` → `:=2` bump ("Bump PKG_RELEASE to reflect the updated device selection"), which was the outstanding nit — documenting it is a fine resolution.

One follow-up on the [atf-airoha flash-table question](https://github.com/openwrt/openwrt/pull/24465#discussion_r3751133071), since your answer confirms rather than closes it: [Ansuel/atf-airoha#3](https://github.com/Ansuel/atf-airoha/pull/3) is still open per the PR description, and this series does not touch [`SOURCE_VERSION:=80519bea…` (2026-07-17) in package/boot/arm-trusted-firmware-airoha/Makefile:94](https://github.com/openwrt/openwrt/blob/b788d58e93e36f6f9e93138f4be5726580d83249/package/boot/arm-trusted-firmware-airoha/Makefile#L94). So the BL2 flash table `airoha-atf-gen-flash-table` emits into `tcboot.bin` here is still the pre-PR one. If W25N01KV / F35SQA001G are not in the `80519bea` table, a unit with either chip gets a `tcboot.bin` that cannot read its own flash in BL2 — with no console on most VSOL units, per your own note. I can't inspect the atf-airoha tree from this session to check either way, so flagging rather than asserting: worth confirming this PR should land only after the atf-airoha change is merged and `SOURCE_VERSION`/`MIRROR_HASH` bumped here, or noting in the PR description that the pinned revision already covers those chips.

No new inline issues. CI on `b788d58` is green across all 22 check runs.

---

target/linux/airoha/dts/an7583-vsol-common.dtsi
BYZYB: 256 MiB DDR is confirmed to boot normally on V2901Q-A, no misfunction was found.

BYZYB:
> One follow-up on the [atf-airoha flash-table question](https://github.com/openwrt/openwrt/pull/24465#discussion_r3751133071), since your answer confirms rather than closes it: [Ansuel/atf-airoha#3](https://github.com/Ansuel/atf-airoha/pull/3) is still open per the PR description, and this series does not touch [`SOURCE_VERSION:=80519bea…` (2026-07-17) in package/boot/arm-trusted-firmware-airoha/Makefile:94](https://github.com/openwrt/openwrt/blob/b788d58e93e36f6f9e93138f4be5726580d83249/package/boot/arm-trusted-firmware-airoha/Makefile#L94).

The PR#3 in upstream repo has been merged, so `W25N01KV` will be supported after the update of `atf-airoha` shortly. The support of V2902A-S/V2901Q-A seems to be completed in this PR.

[V2902A](https://www.vsolcn.com/product/10g-pon-1ge-10ge-sfu-onu-v2902a) based on AN7581 will be added after this one is merged. HG3232AX*-A/V2804AX*-A/V2801AX*-A based on AN7552 + MT7916 are also considered to be supported, although xpon features are still missing in OpenWrt.

alexceltare2:
OMCI is being actively developed here:
https://github.com/openwrt/openwrt/pull/24577

BYZYB:
> OMCI is being actively developed here: #24577

I saw this PR weeks ago and got surprised about it, great job.

The OMCI app/driver there seems to be ported form Airoha Linux SDK `tclinux_phoenix` (should be EOL now), and the EN7528 is MIPS rather than ARM in modern Airoha SoCs. However, all VSOL products are using Airoha OpenWrt SDK 21.02 with half-opensource OMCI binary for linux-5.4 only. Not sure how much the difference between Linux/OpenWrt SDK.

Although I have full access to the opensource part of OAM/OMCI sources, and porting them to mainline OpenWrt may also be a future plan (if Airoha doesn't provide the in the incoming `upstream SDK`), the problem about opensource license and legal risk cannot be ignored. The most possible choice for this plan is to provide prebuilt apk/ipk packages about XPON, and installing manually when using mainline OpenWrt version. You could send message to me if need any help about XPON.

BYZYB:
@jonasjelonek Please review this PR, the AN7583 devices seem to working properly.

package/boot/uboot-airoha/patches/403-airoha-add-vsol-an7583.patch
```diff
@@ -0,0 +1,295 @@
+From 88af859f94c01c24336337a37754620243c134e4 Mon Sep 17 00:00:00 2001
+From: Yubo Zhang <zyb_1998@outlook.com>
+Date: Tue, 14 Jul 2026 11:29:43 +0800
+Subject: [PATCH] an7583: add support for VSOL AN7583 ONUs
```
openwrt-ai, line 4: The `402-` slot is taken now. The rebase onto `6c12b87` pulled in [`402-add-nokia-xg-040g-mf.patch`](https://github.com/openwrt/openwrt/blob/15c8b3f294eaf9119ccb82412a7434b72ec924df/package/boot/uboot-airoha/patches/402-add-nokia-xg-040g-mf.patch), so `package/boot/uboot-airoha/patches/` now has two `402-` patches. Nothing breaks today — the two touch disjoint files (`an7583-nokia-xg-040g-mf.dts` / `an7583_nokia_xg-040g-mf_defconfig` / `defenvs/an7583_nokia_xg-040g-mf_env` vs `an7583-vsol.dts` / `an7583_vsol_defconfig` / `defenvs/an7583_vsol_env`) and the filename sort breaks the tie, which is why CI is still green — but this is the same collision that moved this patch off `999-` in the first place. `403-airoha-add-vsol-an7583.patch` is the next free slot.

---

target/linux/airoha/image/an7583.mk
```diff
@@ -58,6 +58,25 @@ define Device/nokia_xg-040g-mf-common
 	kmod-regulator-userspace-consumer kmod-usb-ledtrig-usbport
 endef
 
+define Device/vsol_an7583-common
```
openwrt-ai: nit: after the rebase this block lands between `define Device/nokia_xg-040g-mf-common` and its two consumers, so the Nokia family is now split by an unrelated vendor's shared block: `nokia_xg-040g-mf-common` (line 51) → `vsol_an7583-common` (here) → `nokia_xg-040g-mf` (line 80) → `nokia_xg-040g-mf-ubi` (line 92) → `vsol_v2901q-a` (line 113). Every other `-common` block in this file sits directly above the devices that `$(call ...)` it. Moving this block down to just before `define Device/vsol_v2901q-a` restores that.

---

openwrt-ai (commented):
Reviewed the rebase onto `6c12b87` (`b788d58` → `15c8b3f`). All three commits are new SHAs but the series is almost entirely a rebase: comparing patch text against the previous round, commit 1 changed only in hunk context, commit 2 only in context plus `PKG_RELEASE` `:=2` → `:=3` (correct — the base bumped it to `2` for `nokia_xg-040g-mf-ubi`, and commit 2's message already documents its own bump), and commit 3 has exactly two substantive changes. All three commit messages still match their diffs.

The two real changes in commit 3 both check out:

- `en8811` on the V2901Q-A is now `compatible = "ethernet-phy-id03a2.a411"` instead of `"ethernet-phy-ieee802.3-c45"`. `EN8811H_PHY_ID` is `0x03a2a411` in [air_en8811h.c:22](https://github.com/gregkh/linux/blob/f8ecbd4614e9e2c9f8ea9cf3e5cffa42ab7f1b52/drivers/net/phy/air_en8811h.c#L22), matched with `PHY_ID_MATCH_MODEL`, so `03a2.a411` is the right encoding — and it now matches every other EN8811H in the tree, including the sibling [an7583-nokia_xg-040g-mf-common.dtsi:169](https://github.com/openwrt/openwrt/blob/15c8b3f294eaf9119ccb82412a7434b72ec924df/target/linux/airoha/dts/an7583-nokia_xg-040g-mf-common.dtsi#L169).
- The `CONFIG_UIMAGE_FIT_BLK=y` hunk on `an7583/config-6.18` is dropped, because the base already carries it at [line 371](https://github.com/openwrt/openwrt/blob/15c8b3f294eaf9119ccb82412a7434b72ec924df/target/linux/airoha/an7583/config-6.18#L371). Correct to drop.

I also re-checked that the rebase did not strand anything: `gdm3` is still enabled by `an7583-vsol-common.dtsi:196`, `an7583_vsol` sorts correctly into `UBOOT_TARGETS`, `U-Boot/an7583_vsol` still legitimately omits `BL2_IMAGE`/`BL31_IMAGE` (`LZMA_COMPRESS` is set in `U-Boot/Default`, so `Build/InstallDev` still stages `an7583_vsol-u-boot.lzma` for `airoha-image-tcboot`/`airoha-image-tclinux`), and the newly in-tree `an7583-preloader` / `an7583-bl31-uboot` steps in `an7583.mk` do not overlap with this PR's steps in `image/Makefile`.

On the [atf-airoha flash-table question](https://github.com/openwrt/openwrt/pull/24465#discussion_r3751133071) from an earlier round: this rebase picks up a base that moved `SOURCE_VERSION` from `80519bea` (2026-07-17) to [`94892b9e` (2026-08-14)](https://github.com/openwrt/openwrt/blob/15c8b3f294eaf9119ccb82412a7434b72ec924df/package/boot/arm-trusted-firmware-airoha/Makefile#L95). If that is the revision carrying [Ansuel/atf-airoha#3](https://github.com/Ansuel/atf-airoha/pull/3), the W25N01KV / F35SQA001G concern is resolved by the rebase alone and nothing further is needed here. I can't read that repo from this session to confirm, so just noting why I'm not re-raising it.

Two inline comments below: one patch-number collision introduced by the rebase, one placement nit. CI on `15c8b3f` is green across all 22 check runs.

---

openwrt-ai (commented):

---

PR #24466, opened 2026-07-29
qualcommax: ipq807x: RBX750 set macs in board.json

facboy:
Update `/etc/board.d/02_network` to set macs in `/etc/board.json`.

### DIFF
--- a/package/base-files/files/bin/config_generate
+++ b/package/base-files/files/bin/config_generate
@@ -92,7 +92,7 @@ generate_static_network() {
 
 addr_offset=2
 generate_network() {
-	local ports device macaddr protocol type ipaddr netmask vlan
+	local ports device macaddr protocol type ipaddr netmask vlan portmacaddr
 	local bridge=$2
 
 	json_select network
@@ -117,15 +117,21 @@ generate_network() {
 			set network.@device[-1].type='bridge'
 		EOF
 		for port in $ports; do uci add_list network.@device[-1].ports="$port"; done
-		[ -n "$macaddr" ] && {
+
+		json_select network_device
 			for port in $ports; do
-				uci -q batch <<-EOF
-					add network device
-					set network.@device[-1].name='$port'
-					set network.@device[-1].macaddr='$macaddr'
-				EOF
+				json_select "$port" && json_get_var portmacaddr macaddr && json_select ..
+				[ -n "$portmacaddr" ] || portmacaddr="$macaddr"
+				if [ -n "$portmacaddr" ]; then
+					uci -q batch <<-EOF
+						add network device
+						set network.@device[-1].name='$port'
+						set network.@device[-1].macaddr='$macaddr'
+					EOF
+				fi
 			done
-		}
+		json_select ..
+
 		device=br-$1
 		type=
 		macaddr=""
--- a/target/linux/qualcommax/ipq807x/base-files/etc/board.d/02_network
+++ b/target/linux/qualcommax/ipq807x/base-files/etc/board.d/02_network
@@ -99,6 +99,15 @@ ipq807x_setup_macs()
 		lan_mac=$(macaddr_add "$wan_mac" 1)
 		label_mac="$wan_mac"
 		;;
+	netgear,rbr750)
+		label_mac=$(get_mac_label_dt)
+		lan_mac=$label_mac
+		wan_mac=$(macaddr_add $label_mac 1)
+		;;
+	netgear,rbs750)
+		label_mac=$(get_mac_label_dt)
+		lan_mac=$label_mac
+		;;
 	tcl,linkhub-hh500v)
 		wan_mac=$(mtd_get_mac_binary "0:ART" 0x5e290)
 		lan_mac=$(mtd_get_mac_binary "0:ART" 0x5e296)
--- a/target/linux/qualcommax/ipq807x/base-files/etc/board.d/02_network
+++ b/target/linux/qualcommax/ipq807x/base-files/etc/board.d/02_network
@@ -92,6 +92,8 @@ ipq807x_setup_macs()
 	local lan_mac=""
 	local wan_mac=""
 	local label_mac=""
+	local lan2_mac=""
+	local lan3_mac=""
 
 	case "$board" in
 	aliyun,ap8220)
@@ -103,6 +105,11 @@ ipq807x_setup_macs()
 		label_mac=$(get_mac_label_dt)
 		lan_mac=$label_mac
 		wan_mac=$(macaddr_add $label_mac 1)
+
+		lan2_mac=$(macaddr_add $label_mac 1)
+		lan2_mac=$(macaddr_setbit_la $lan2_mac)
+		lan3_mac=$(macaddr_add $label_mac 2)
+		lan3_mac=$(macaddr_setbit_la $lan3_mac)
 		;;
 	netgear,rbs750)
 		label_mac=$(get_mac_label_dt)
@@ -132,6 +139,8 @@ ipq807x_setup_macs()
 	[ -n "$lan_mac" ] && ucidef_set_interface_macaddr "lan" $lan_mac
 	[ -n "$wan_mac" ] && ucidef_set_interface_macaddr "wan" $wan_mac
 	[ -n "$label_mac" ] && ucidef_set_label_macaddr $label_mac
+	[ -n "$lan2_mac" ] && ucidef_set_network_device_mac "lan2" $lan2_mac
+	[ -n "$lan3_mac" ] && ucidef_set_network_device_mac "lan3" $lan3_mac
 }
 
 board_config_update


testuser7:
@facboy Could you try to fix the problem by setting `status="disabled"` to partitions: https://github.com/openwrt/openwrt/blob/main/target/linux/qualcommax/dts/ipq8072-dl-wrx36.dts#L94-L106 ?

facboy:
tbh i'm not sure if it's actually a problem. the mac addresses *are* read (can be seen in the device tree) but aren't put into `board.json`. i'm not sure if that matters?

testuser7:
So what does this command return:
```
cat /proc/device-tree/aliases/label-mac-device
```

facboy:
```
/soc@0/ppe@3a000000/ports/port@5root
```
and

```
# (. /lib/functions/system.sh && get_mac_label)
-bash: warning: command substitution: ignored null byte in input
re:al:la:be:la:dd
```

testuser7:
> get_mac_label

~~This is wrong function, proper one is `get_mac_label_dt`.~~

Try this command:
```
hexdump -v -e '6/1 "%02x""\n"' /proc/device-tree/soc@0/ppe@3a000000/ports/port@*/mac-address
```

facboy:
genuine question, what is 'wrong' about it? it works, it outputs the correct address? it outputs the value from the DT (if present) else tries to use `board.json`. but it's in the DT anyway.

the hexdump outputs two mac addresses, one for port4 and one for port5, port5 is the 'correct' one.

target/linux/qualcommax/ipq807x/base-files/etc/board.d/02_network
```diff
@@ -99,6 +99,15 @@ ipq807x_setup_macs()
 		lan_mac=$(macaddr_add "$wan_mac" 1)
 		label_mac="$wan_mac"
 		;;
+	netgear,rbr750)
+		label_mac=$(mtd_get_mac_binary board_data 0x40)
```
testuser7:
```suggestion
		label_mac=$(get_mac_label_dt)
```
If the MAC label is correctly defined in the DT then you should read it this way.
Additional question: does the MAC label refer to port 5 (lan1) or port 1 (wan)?

target/linux/qualcommax/ipq807x/base-files/etc/board.d/02_network
```diff
@@ -99,6 +99,15 @@ ipq807x_setup_macs()
...
+		lan_mac=$label_mac
+		wan_mac=$(macaddr_add $label_mac 1)
+		;;
+	netgear,rbs750)
+		label_mac=$(mtd_get_mac_binary board_data 0x40)
```
testuser7:
```suggestion
		label_mac=$(get_mac_label_dt)
```

target/linux/qualcommax/ipq807x/base-files/etc/board.d/02_network
```diff
@@ -99,6 +99,15 @@ ipq807x_setup_macs()
 		lan_mac=$(macaddr_add "$wan_mac" 1)
 		label_mac="$wan_mac"
 		;;
+	netgear,rbr750)
```
openwrt-ai, line 102: nit: the commit message states "The mac addresses are no longer read by the driver", but the DT does assign them: [`swport1` gets `nvmem-cells = <&macaddr_boarddata 1>`](https://github.com/openwrt/openwrt/blob/77259f98dcc84ec9d25b18c50ee108a681d3440f/target/linux/qualcommax/dts/ipq8074-rbx750.dtsi#L115-L122) and [`swport5` gets index 0](https://github.com/openwrt/openwrt/blob/77259f98dcc84ec9d25b18c50ee108a681d3440f/target/linux/qualcommax/dts/ipq8074-rbx750.dtsi#L145-L152) from [`macaddr_boarddata` at `board_data` 0x40](https://github.com/openwrt/openwrt/blob/77259f98dcc84ec9d25b18c50ee108a681d3440f/target/linux/qualcommax/dts/ipq8074-rbx750.dtsi#L215-L219) — and you confirmed in the discussion above that the addresses are present under `/proc/device-tree`. The only thing missing is the copy in `board.json`. Please reword the commit message to say what is actually not populated (e.g. `board.json` `system.label_macaddr` / `network.*.macaddr`), otherwise it describes a driver regression that does not appear to exist.

---

target/linux/qualcommax/ipq807x/base-files/etc/board.d/02_network
```diff
@@ -99,6 +99,15 @@ ipq807x_setup_macs()
...
+		label_mac=$(mtd_get_mac_binary board_data 0x40)
+		lan_mac=$label_mac
+		wan_mac=$(macaddr_add $label_mac 1)
```
openwrt-ai, line 105: This recomputes an address the wan port already has: [`swport1` is declared with `nvmem-cells = <&macaddr_boarddata 1>`](https://github.com/openwrt/openwrt/blob/77259f98dcc84ec9d25b18c50ee108a681d3440f/target/linux/qualcommax/dts/ipq8074-rbx750.dtsi#L115-L122), i.e. `board_data` 0x40 plus 1 — exactly `macaddr_add $label_mac 1`. Hardcoding partition name and offset here duplicates [the nvmem layout in the dtsi](https://github.com/openwrt/openwrt/blob/77259f98dcc84ec9d25b18c50ee108a681d3440f/target/linux/qualcommax/dts/ipq8074-rbx750.dtsi#L207-L221), so the two can drift apart silently; deriving from the DT (`get_mac_label_dt`, as suggested on the line above) keeps one source of truth. Do you actually observe the wan interface coming up with a wrong MAC without this line, or is only the `board.json` label entry missing?

---

openwrt-ai (commented):
Reviewed 1 new commit. The change itself is consistent with the DT (`board_data` 0x40 = `macaddr_boarddata` base, wan = base+1, label/lan1 = base+0, and the partition really is named `board_data` since `partition-board_data` matches the smem-parsed MTD name). Two points inline: the commit message's premise about the driver, and the duplication of the DT nvmem layout in shell.

---

facboy:
updated for review comments. the label shows the `lan1` address.

target/linux/qualcommax/ipq807x/base-files/etc/board.d/02_network
```diff
@@ -99,6 +99,15 @@ ipq807x_setup_macs()
 		lan_mac=$(macaddr_add "$wan_mac" 1)
 		label_mac="$wan_mac"
 		;;
+	netgear,rbr750)
+		label_mac=$(get_mac_label_dt)
```
openwrt-ai, line 103: nit: the commit message no longer claims the driver stopped reading the MACs (thanks), but the PR description still does — and that claim is now doubly wrong, because this line only works *because* the driver reads them. `get_mac_label_dt` follows [`label-mac-device = &swport5`](https://github.com/openwrt/openwrt/blob/6ce038886aad4fc317f6874f1113df2b8d1acf5b/target/linux/qualcommax/dts/ipq8074-rbx750.dtsi#L21) and reads `/proc/device-tree/soc@0/ppe@3a000000/ports/port@5/mac-address`, a property that exists only after the switch driver resolved [`nvmem-cells = <&macaddr_boarddata 0>`](https://github.com/openwrt/openwrt/blob/6ce038886aad4fc317f6874f1113df2b8d1acf5b/target/linux/qualcommax/dts/ipq8074-rbx750.dtsi#L145-L153) and the OpenWrt-local [`of_add_mac_address()` write-back](``https://github.com/openwrt/openwrt/blob/6ce038886aad4fc317f6874f1113df2b8d1acf5b/target/linux/generic/pending-6.12/683-of_net-add-mac-address-to-of-tree.patch#L24-L44)`` put it into the live DT. Please update the PR body to match the commit message so the discussion doesn't record a driver regression that doesn't exist.

---

target/linux/qualcommax/ipq807x/base-files/etc/board.d/02_network
```diff
@@ -99,6 +99,15 @@ ipq807x_setup_macs()
...
+		lan_mac=$label_mac
```
openwrt-ai, line 104: While you are touching the MAC setup for this family: in the dtsi only `swport5` (lan1) uses cell index 0 — [`swport2` (lan3)](https://github.com/openwrt/openwrt/blob/6ce038886aad4fc317f6874f1113df2b8d1acf5b/target/linux/qualcommax/dts/ipq8074-rbx750.dtsi#L125-L133) and [`swport4` (lan2)](https://github.com/openwrt/openwrt/blob/6ce038886aad4fc317f6874f1113df2b8d1acf5b/target/linux/qualcommax/dts/ipq8074-rbx750.dtsi#L135-L143) both carry `nvmem-cells = <&macaddr_boarddata 1>`, i.e. the *same* address as [`swport1` (wan)](https://github.com/openwrt/openwrt/blob/6ce038886aad4fc317f6874f1113df2b8d1acf5b/target/linux/qualcommax/dts/ipq8074-rbx750.dtsi#L115-L123). Setting `lan_mac` to lan1's address here fixes the bridge MAC, but the three ports keep duplicate hardware addresses on the same switch (and lan2/lan3 keep sharing them with wan). That's pre-existing in the dtsi rather than introduced by this patch, so: is the index-1 reuse on `swport2`/`swport4` deliberate (e.g. the OEM really programs only two addresses in `board_data`), or should those ports get distinct offsets in the same series?

---

openwrt-ai (commented):
Re-reviewed the force-push (`77259f9` → `6ce0388`). The only code change is `mtd_get_mac_binary board_data 0x40` → `get_mac_label_dt`, which resolves my earlier point about duplicating the dtsi nvmem layout in shell, and matches the existing `zyxel,nwa110ax|zyxel,nwa210ax` entry in this same file. The derivation still checks out against the DT: `label-mac-device = &swport5` (lan1, `macaddr_boarddata` index 0), and `macaddr_add $label_mac 1` reproduces `swport1`'s index-1 address for wan; `netgear,rbs750` correctly omits `wan_mac` since its `swport1`/`swport2` are disabled and it only registers a lan interface. The reworded commit message now matches the diff.

Two remaining points inline, both non-blocking.

---

testuser7:
> updated for review comments. the label shows the `lan1` address.

In most cases, the MAC address on the label corresponds to the wan port. Do you know what it was like on the OEM firmware? Maybe it's worth changing that too...

facboy:
i don't have any with OEM firmware any more and i don't plan to flash back tbh. afaik it has always been set up that way since the DTS was written (i.e. before switching to the new drivers), would have to ask why it was done that way. iirc the OEM firmware is pretty locked down, may not be that easy to see what the macs are.

facboy:
actually if you look at the OEM bootlog at https://openwrt.org/inbox/toh/netgear/rbx750?s[]=rbr750 the addresses are a bit all over the place. but the mac that openwrt has pulled as the 'label' there appears to be the one assigned to one of the lan interfaces at least.

it could be an oversight:

```
3: eth0: <NO-CARRIER,BROADCAST,MULTICAST,UP> mtu 1500 qdisc mq state DOWN group default qlen 1000
    link/ether 10:0c:6b:53:d6:9f brd ff:ff:ff:ff:ff:ff
4: eth1: <NO-CARRIER,BROADCAST,MULTICAST,ALLMULTI,UP> mtu 1500 qdisc mq master br-lan state DOWN group default qlen 1000
    link/ether 10:0c:6b:53:d6:9e brd ff:ff:ff:ff:ff:ff
5: eth2: <BROADCAST,MULTICAST,ALLMULTI,UP,LOWER_UP> mtu 1500 qdisc mq master br-lan state UP group default qlen 1000
    link/ether 12:0c:6b:53:d6:9f brd ff:ff:ff:ff:ff:ff
    inet6 fe80::100c:6bff:fe53:d69f/64 scope link
       valid_lft forever preferred_lft forever
6: eth3: <NO-CARRIER,BROADCAST,MULTICAST,ALLMULTI,UP> mtu 1500 qdisc mq master br-lan state DOWN group default qlen 1000
    link/ether 12:0c:6b:53:d6:a0 brd ff:ff:ff:ff:ff:ff
```

eth1 has the mac openwrt pulled as the label mac (on br-lan), but you can see that even though `eth2` and `eth3` are `d6:9f` and `d6:a0`, they start with `12:` instead of `10:`. so perhaps at first glance it was missed that they weren't actually sharing the mac address?

target/linux/qualcommax/dts/ipq8074-rbx750.dtsi
```diff
@@ -128,7 +128,7 @@
 			phy-handle = <&phy1>;
 			phy-mode = "psgmii";
 			pcs-handle = <&uniphy0 1>;
-			nvmem-cells = <&macaddr_boarddata 1>;
+			nvmem-cells = <&macaddr_boarddata 3>;
```
openwrt-ai: Offsets 2 and 3 are already taken by the ath11k radios on this exact board, so this trades one duplicate-MAC problem for another. [`11-ath11k-caldata` lines 70-78](``https://github.com/openwrt/openwrt/blob/d0739a3eeb49ffe1aa9f186243ea718a5afc7c61/target/linux/qualcommax/ipq807x/base-files/etc/hotplug.d/firmware/11-ath11k-caldata#L70-L78)`` patches, for both `netgear,rbr750` and `netgear,rbs750`:

```
label_mac=$(get_mac_label_dt)
ath11k_patch_mac $(macaddr_add $label_mac 3) 0
ath11k_patch_mac $(macaddr_add $label_mac 2) 1
ath11k_patch_mac $(macaddr_add $label_mac 4) 2
```

With `mac-base` the second phandle cell is a plain `eth_addr_add()` increment (see [`804-nvmem-core-support-mac-base-fixed-layout-cells.patch`](``https://github.com/openwrt/openwrt/blob/d0739a3eeb49ffe1aa9f186243ea718a5afc7c61/target/linux/generic/pending-6.12/804-nvmem-core-support-mac-base-fixed-layout-cells.patch#L40-L50)),`` so `<&macaddr_boarddata 3>` here is `label_mac + 3` — the same address radio 0 gets — and [`<&macaddr_boarddata 2>` on `swport4`/lan2 below](https://github.com/openwrt/openwrt/blob/d0739a3eeb49ffe1aa9f186243ea718a5afc7c61/target/linux/qualcommax/dts/ipq8074-rbx750.dtsi#L135-L143) is the same address radio 1 gets. Since the lan ports and the AP interfaces all end up in `br-lan`, that puts two identical hardware addresses on the same bridge, which is the failure mode this commit is trying to remove.

Offsets 0-4 are all spoken for (0 = lan1/label, 1 = wan, 2/3/4 = radios), so lan2/lan3 need something outside that range — e.g. 5 and 6.

Separately, worth deciding before picking numbers: the OEM boot log you quoted in the discussion shows the vendor did *not* hand out further globally-unique addresses for the extra ports — `eth2`/`eth3` are `12:0c:6b:53:d6:9f` / `12:0c:6b:53:d6:a0`, i.e. the locally-administered bit set on `base+1` / `base+2`, while the globally-unique block appears to be just `…d6:9e` and `…d6:9f`. If that block really is two addresses wide, then `base+5`/`base+6` are globally-unique addresses belonging to some other unit, and the LA-bit approach (as [`netgear,sxr80|netgear,sxs80` do with `macaddr_setbit_la`](``https://github.com/openwrt/openwrt/blob/d0739a3eeb49ffe1aa9f186243ea718a5afc7c61/target/linux/qualcommax/ipq807x/base-files/etc/hotplug.d/firmware/11-ath11k-caldata#L79-L87))`` would be the safer match to the OEM. That can't be expressed with `mac-base` increments, so it would mean setting these two in `02_network` instead of the dtsi.

---

target/linux/qualcommax/dts/ipq8074-rbs750.dts
```diff
@@ -19,3 +19,7 @@
 &swport2 {
 	status = "disabled";
 };
+
+&swport4 {
+	nvmem-cells = <&macaddr_boarddata 1>;
```
openwrt-ai: nit: the commit-message sentence describing this hunk reads "On RBS750 `lan2` uses `label+mac + 1` as before" — `label+mac` should be `label_mac`. The sentence above it also has a stray "so that `lan2` to use". Worth fixing while rewording for the point above.

---

openwrt-ai (commented):
Reviewed 1 new commit (`d0739a3`, "RBX750 separate `lan2` and `lan3` mac addresses from `wan`"); the previously-reviewed commit was only rebased. The message matches the diff — `swport2` (lan3) → index 3, `swport4` (lan2) → index 2, and the `rbs750.dts` override keeps lan2 at index 1 as stated.

One substantive issue inline: the offsets chosen collide with the ath11k radio MACs already derived from `label_mac` for these two boards in `11-ath11k-caldata`, so lan2/lan3 would end up duplicating radio 1 / radio 0 on the same bridge.

---

target/linux/qualcommax/ipq807x/base-files/etc/board.d/02_network
```diff
@@ -99,6 +99,22 @@ ipq807x_setup_macs()
...
+		wan_mac=$(macaddr_add $label_mac 1)
+
+		local lan2_mac=$(macaddr_add $label_mac 1)
+		lan2_mac=$(macaddr_setbit_la $lan2_mac)
+		local lan3_mac=$(macaddr_add $label_mac 1)
```
openwrt-ai: `lan3_mac` is computed from the same expression as `lan2_mac` on line 107 (`macaddr_add $label_mac 1`), so after `macaddr_setbit_la` the two ports end up with the *identical* address. The commit subject says "separate `lan2` and `lan3` mac addresses" but they are only separated from `wan`, not from each other — two identical MACs still land on the same bridge.

The OEM boot log you quoted in the discussion has `eth2 = 12:0c:6b:53:d6:9f` (LA bit on `label_mac + 1`) and `eth3 = 12:0c:6b:53:d6:a0` (LA bit on `label_mac + 2`), so the second port needs offset 2:

```suggestion
		local lan3_mac=$(macaddr_add $label_mac 2)
```

Worth confirming the OEM `eth2`/`eth3` → OpenWrt `lan2`/`lan3` mapping before settling on which port gets which offset (in the dtsi `swport4` is lan2 and `swport2` is lan3).

---

target/linux/qualcommax/ipq807x/base-files/etc/board.d/02_network
```diff
@@ -99,6 +99,22 @@ ipq807x_setup_macs()
...
+		lan3_mac=$(macaddr_setbit_la $lan3_mac)
+		[ -n "$lan2_mac" ] && ucidef_set_interface_macaddr "lan2" $lan2_mac
+		[ -n "$lan3_mac" ] && ucidef_set_interface_macaddr "lan3" $lan3_mac
```
openwrt-ai: These two calls are a no-op. `ucidef_set_interface_macaddr` writes `network.<name>.macaddr` into `board.json`, i.e. it addresses a *network interface*, not a netdev/port. `netgear,rbr750` only declares `lan` and `wan` interfaces ([`ucidef_set_interfaces_lan_wan "lan1 lan2 lan3" "wan"`](``https://github.com/openwrt/openwrt/blob/f48d591a854d5764ae64bb95c45e53d619bd6d33/target/linux/qualcommax/ipq807x/base-files/etc/board.d/02_network#L42-L47)),`` so this only adds `network.lan2` / `network.lan3` objects that contain nothing but a `macaddr` — and [`generate_network()` returns early on exactly those](https://github.com/openwrt/openwrt/blob/f48d591a854d5764ae64bb95c45e53d619bd6d33/package/base-files/files/bin/config_generate#L105) because neither `device` nor `ports` is set. Nothing ever reaches `/etc/config/network`.

Per-netdev MACs go through `ucidef_set_network_device_mac`, which lands in the `board.json` `network_device` object that netifd reads directly ([`config_get_default_macaddr()`](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/config.c#L738-L754), used from [`device_fill_default_settings()`](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/device.c#L748-L754)); see [`target/linux/ath79/mikrotik/base-files/etc/board.d/02_network:99-103`](``https://github.com/openwrt/openwrt/blob/f48d591a854d5764ae64bb95c45e53d619bd6d33/target/linux/ath79/mikrotik/base-files/etc/board.d/02_network#L99-L103``) for the pattern.

Swapping the helper alone likely still won't be enough, though: `lan_mac` on line 104 makes `config_generate` emit an explicit `config device` + `option macaddr` for *every* port of the lan bridge — [lines 120-127](https://github.com/openwrt/openwrt/blob/f48d591a854d5764ae64bb95c45e53d619bd6d33/package/base-files/files/bin/config_generate#L120-L127) loop over `lan1 lan2 lan3` — and an explicit UCI `macaddr` sets `DEV_OPT_MACADDR`, which makes netifd skip the `network_device` default entirely. That also means commit 1 on its own already stops `lan2`/`lan3` from sharing `wan`'s address (they get `label_mac` instead of the dtsi's `<&macaddr_boarddata 1>`), so the premise in this commit message no longer matches the tree it is applied on. Please re-check the generated `/etc/config/network` on the device and confirm the three lan ports really come up with three distinct addresses before this lands.

---

openwrt-ai (commented):
Re-reviewed after the rework (`d0739a3` → `f48d591`). The dtsi/dts changes are gone and everything now lives in `02_network` using `macaddr_setbit_la`, which addresses my earlier point about colliding with the ath11k radio offsets — the LA-bit approach also matches the OEM boot log you posted. Commit 1 is unchanged apart from the rebase, and the new commit's subject/scope match the files it touches.

Two blocking-ish problems with the new hunk, both inline: `lan2` and `lan3` are given the same address, and the `ucidef_set_interface_macaddr "lan2"`/`"lan3"` calls don't actually produce any config for a non-interface name.

---

facboy:
Not sure of the value of setting the `lan2` and `lan3` addresses correctly, they're in `board.json` now but e.g. generated `/etc/config/network` ignores them, as the AI said.

target/linux/qualcommax/ipq807x/base-files/etc/board.d/02_network
```diff
@@ -99,6 +99,23 @@ ipq807x_setup_macs()
...
+		lan_mac=$label_mac
+		wan_mac=$(macaddr_add $label_mac 1)
+
+		local lan2_mac=$(macaddr_add $label_mac 1)
+		lan2_mac=$(macaddr_setbit_la $lan2_mac)
+		local lan3_mac=$(macaddr_add $label_mac 2)
+		lan3_mac=$(macaddr_setbit_la $lan3_mac)
+
+		[ -n "$lan2_mac" ] && ucidef_set_network_device_mac "lan2" $lan2_mac
+		[ -n "$lan3_mac" ] && ucidef_set_network_device_mac "lan3" $lan3_mac
```
openwrt-ai, line 113: The helper is right now, and the offsets check out (LA(`base+1`)/LA(`base+2`) match the OEM `eth2`/`eth3` from your boot log, and the LA bit keeps them clear of the radio offsets 2/3/4 that [`11-ath11k-caldata:74-76`](``https://github.com/openwrt/openwrt/blob/7b469bfc6cd9205c6d1c3292c44b019734ddf864/target/linux/qualcommax/ipq807x/base-files/etc/hotplug.d/firmware/11-ath11k-caldata#L74-L76)`` uses). But these two entries are still dead as long as [`lan_mac=$label_mac` on line 104](``https://github.com/openwrt/openwrt/blob/7b469bfc6cd9205c6d1c3292c44b019734ddf864/target/linux/qualcommax/ipq807x/base-files/etc/board.d/02_network#L104``) stays — which matches what you observed in https://github.com/openwrt/openwrt/pull/24466#issuecomment-5155897172.

Mechanism: `lan_mac` becomes `network.lan.macaddr` in `board.json`; `netgear,rbr750` has no `ucidef_set_bridge_device`, so `generate_network()` takes the no-DSA-bridge path and [stamps an explicit `config device` + `option macaddr` on *every* lan port](https://github.com/openwrt/openwrt/blob/7b469bfc6cd9205c6d1c3292c44b019734ddf864/package/base-files/files/bin/config_generate#L120-L127) — `lan1 lan2 lan3` all get `label_mac`. An explicit UCI `macaddr` sets `DEV_OPT_MACADDR`, and [`device_fill_default_settings()` only consults the `board.json` `network_device` mac when that flag is clear](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/device.c#L748-L753). Net result on the device: three identical lan MACs, and the LA addresses added here are never applied.

Since [`swport5`/lan1 already carries `nvmem-cells = <&macaddr_boarddata 0>`](https://github.com/openwrt/openwrt/blob/7b469bfc6cd9205c6d1c3292c44b019734ddf864/target/linux/qualcommax/dts/ipq8074-rbx750.dtsi#L145-L152) (= `label_mac`) from the driver, `lan_mac` isn't buying lan1 anything here — dropping it from the `netgear,rbr750` case and letting all three ports come from `network_device` would make this commit actually take effect. `netgear,rbs750` is unaffected (single lan port, no override loop worth worrying about). Please confirm against the generated `/etc/config/network` and `ip link` on hardware before this lands, since the whole point of the commit is the three ports ending up distinct.

---

target/linux/qualcommax/ipq807x/base-files/etc/board.d/02_network
```diff
@@ -99,6 +99,23 @@ ipq807x_setup_macs()
...
 		;;
+	netgear,rbr750)
+		label_mac=$(get_mac_label_dt)
+		lan_mac=$label_mac
+		wan_mac=$(macaddr_add $label_mac 1)
+
+		local lan2_mac=$(macaddr_add $label_mac 1)
+		lan2_mac=$(macaddr_setbit_la $lan2_mac)
+		local lan3_mac=$(macaddr_add $label_mac 2)
+		lan3_mac=$(macaddr_setbit_la $lan3_mac)
```
openwrt-ai, line 110: nit: `ipq807x_setup_macs()` declares its locals at the top of the function and keeps the `case` purely assignment-only, with every `ucidef_*` call collected after `esac`. This block breaks both halves of that pattern (mid-`case` `local`, plus `ucidef_*` inside the branch). If you keep this shape, consider hoisting `local lan2_mac="" lan3_mac=""` up next to `local label_mac=""` and moving the two `ucidef_set_network_device_mac` calls down beside the existing `[ -n "$lan_mac" ] && ...` lines, so the function keeps one structure.

---

openwrt-ai (commented):
Re-reviewed after the rework (`f48d591` → `7b469bf`). Both points from the last round are addressed: `lan3_mac` now uses offset 2 so it no longer collides with `lan2_mac`, and the calls moved from `ucidef_set_interface_macaddr` to `ucidef_set_network_device_mac`, which is the right helper for a per-netdev MAC. Commit 1 is unchanged apart from the rebase, and the new commit's subject (`RBR750 ...`) matches the single `netgear,rbr750` branch it touches.

One substantive point remains inline: `lan_mac` on line 104 still makes `config_generate` stamp an explicit `option macaddr` on all three lan ports, which sets `DEV_OPT_MACADDR` and makes netifd ignore the `network_device` entries this commit adds — the effect you already noticed in the discussion. Plus one style nit.

---

package/base-files/files/bin/config_generate
```diff
@@ -117,15 +117,21 @@ generate_network() {
...
-					set network.@device[-1].name='$port'
-					set network.@device[-1].macaddr='$macaddr'
-				EOF
+				json_select "$port" && json_get_var portmacaddr macaddr && json_select ..
+				[ -n "$portmacaddr" ] || portmacaddr="$macaddr"
+				if [ -n "$portmacaddr" ]; then
+					uci -q batch <<-EOF
+						add network device
+						set network.@device[-1].name='$port'
+						set network.@device[-1].macaddr='$macaddr'
```
openwrt-ai, line 129: The value written is still the bridge-level `$macaddr`, not the per-port `$portmacaddr` that the new lookup just resolved. As it stands the whole commit is a no-op: every port keeps getting the `lan` interface MAC, so `lan2`/`lan3` from the previous commit still never reach `/etc/config/network`.

```suggestion
						set network.@device[-1].macaddr='$portmacaddr'
```

---

package/base-files/files/bin/config_generate
```diff
@@ -117,15 +117,21 @@ generate_network() {
...
-		[ -n "$macaddr" ] && {
+
+		json_select network_device
 			for port in $ports; do
-				uci -q batch <<-EOF
-					add network device
-					set network.@device[-1].name='$port'
-					set network.@device[-1].macaddr='$macaddr'
-				EOF
+				json_select "$port" && json_get_var portmacaddr macaddr && json_select ..
```
openwrt-ai, line 123: `portmacaddr` is never cleared between iterations. When `json_select "$port"` fails — i.e. the port has no entry in `network_device`, the common case — the `&&` chain short-circuits and `json_get_var` never runs, so `portmacaddr` keeps the value resolved for the *previous* port. Line 124's `[ -n "$portmacaddr" ] ||` then sees a non-empty value and skips the `$macaddr` fallback, and that port is assigned the previous port's address. On the RBR750 that means `lan2`'s LA address leaks onto any port iterated after it that has no entry of its own.

```suggestion
				portmacaddr=""
				json_select "$port" && json_get_var portmacaddr macaddr && json_select ..
```

Note the same short-circuit also leaves the jshn cursor inside `$port` if `json_get_var` ever fails, since the trailing `json_select ..` is part of the chain; using `json_is_a "$port" object` as the test instead of relying on `json_select`'s return value avoids both problems.

---

package/base-files/files/bin/config_generate
```diff
@@ -117,15 +117,21 @@ generate_network() {
 			set network.@device[-1].type='bridge'
 		EOF
 		for port in $ports; do uci add_list network.@device[-1].ports="$port"; done
-		[ -n "$macaddr" ] && {
+
+		json_select network_device
```
openwrt-ai, line 121: This `json_select` is unguarded, but `network_device` is optional in `board.json` — most boards never call any `ucidef_set_network_device_*` helper and therefore have no such object. In that case jshn's `json_select` prints `WARNING: Variable 'network_device' does not exist or is not an array/object` on stdout and returns 1 *without moving the cursor*, so the matching `json_select ..` on line 133 pops one level above the object the loop was supposed to be in. Since `generate_network()` is called once per interface, the following call's `json_select network` (line 98) then operates from the wrong cursor.

The rest of this file guards optional objects before selecting them — e.g. [`if json_is_a ports array` at config_generate:257](https://github.com/openwrt/openwrt/blob/153ea398345871c75ba643a1e74bce0e90a09a42/package/base-files/files/bin/config_generate#L257). Please wrap the select/loop/deselect in `if json_is_a network_device object; then ... fi` and keep the `$macaddr` fallback path working for boards that take the `else` branch.

---

package/base-files/files/bin/config_generate
```diff
@@ -117,15 +117,21 @@ generate_network() {
...
 			for port in $ports; do
-				uci -q batch <<-EOF
-					add network device
-					set network.@device[-1].name='$port'
-					set network.@device[-1].macaddr='$macaddr'
-				EOF
+				json_select "$port" && json_get_var portmacaddr macaddr && json_select ..
+				[ -n "$portmacaddr" ] || portmacaddr="$macaddr"
+				if [ -n "$portmacaddr" ]; then
```
openwrt-ai, line 125: Have you considered *not* emitting a `device` section for ports that have their own `network_device` entry, rather than copying the value into uci? netifd already reads `network_device.<ifname>.macaddr` out of `board.json` itself and applies it as `DEV_OPT_DEFAULT_MACADDR` in [`device_fill_default_settings()` at device.c:749](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/device.c#L749), but only `if (!(s->flags & DEV_OPT_MACADDR))` — and the pre-existing per-port `option macaddr` written from the bridge-level `$macaddr` is exactly what sets that flag and suppresses it. That is the mechanism behind the "`board.json` has them but `/etc/config/network` ignores them" effect discussed in #24466 (comment 5155897172).

Skipping the write for ports present in `network_device` would let netifd apply its own default and keep `/etc/config/network` free of generated values that are already derivable, while the `$macaddr` fallback keeps today's behaviour for every other port. Not a blocker if you prefer the explicit form — but if you keep it, the two are redundant paths to the same address and it would be worth a line in the commit message saying why the explicit write is preferred.

---

openwrt-ai (commented):
Reviewed 1 new commit (`153ea39`, "base-files: config_generate: set port mac addresses from board.json"); the two earlier commits changed only in that `lan2_mac`/`lan3_mac` are now declared with the other locals at the top of `ipq807x_setup_macs()` and applied in the common `[ -n ... ] && ucidef_...` block at the bottom, which matches the surrounding style and does not change behaviour. Both subjects still match their diffs.

The new commit takes the right approach — `network_device.<port>.macaddr` is the correct source, and the fallback to the bridge-level `$macaddr` preserves existing behaviour for every other board. Three implementation problems inline, the first of which makes the commit a no-op as written: the uci write still uses `$macaddr` instead of the `$portmacaddr` it just looked up. Also `portmacaddr` is not reset between loop iterations, and `json_select network_device` is unguarded on boards whose `board.json` has no such object — since this is generic `base-files` code, that path runs on every target. Plus one question about whether the explicit per-port write is needed at all given netifd's own `board.json` handling.

---

PR #24468, opened 2026-07-29
beagleboard.org BeaglePlay bringup

sayoojkkarun:
To build firmware for the board follow the steps below:
1. **Configure the build**
    make menuconfig
    Target System  - TI K3
    Subtarget - AM62x (TI K3)
    Target Profile - BeagleBoard.org BeaglePlay
2. **Build**
    make V=s -j$(nproc)
3. **After successful build, the output firmware is available at**:
    bin/targets/ti-k3/am625/openwrt-ti-k3-am625-beagleboard_am625-beagleplay-squashfs-sysupgrade.img.gz
4. **Flashing**:
    a. gunzip -c openwrt-ti-k3-am625-beagleboard_am625-beagleplay-squashfs-sysupgrade.img.gz > openwrt.img
    b. sudo dd if=openwrt.img of=/dev/sdX bs=4M status=progress conv=fsync (/dev/sdX is the SD Card)
    c. sync
    d. Connect the onboard UART debug console to a PC and keep it open so that early boot logs are not missed.
    d. Insert the SD card into the BeaglePlay power on the BeaglePlay while the USR button is kept pressed.
        After it powers on release the USR button so that the board boots from SD card.

**What is tested**:
1. Board boots cleanly off the SD card end to end
2. Onboard Ethernet (the RJ45 port) was verified by pinging out from the board and pinging the board from another host on the
    same network. Works as expected.
3. Onboard WiFi (wl18xx) was verified the same way, associating to an access point and confirming connectivity with ping in both
    directions.

**Not Tested**:
1. The single pair Ethernet interface was not tested because of setup unavailability.


package/boot/uboot-ti/Makefile
```diff
@@ -0,0 +1,105 @@
...
+R5_GCC_NAME:=arm-gnu-toolchain
+R5_GCC_RELEASE:=12.3.rel1
+R5_GCC_VERSION:=$(HOST_ARCH)-arm-none-linux-gnueabihf
+R5_GCC_SOURCE:=$(R5_GCC_NAME)-$(R5_GCC_RELEASE)-$(R5_GCC_VERSION).tar.xz
+R5_CROSS_COMPILE:=$(PKG_BUILD_DIR)/$(R5_GCC_NAME)-$(R5_GCC_RELEASE)-$(R5_GCC_VERSION)/bin/arm-none-linux-gnueabihf-
+
+define Download/r5-gcc
+  FILE:=$(R5_GCC_SOURCE)
+  URL:=https://developer.arm.com/-/media/Files/downloads/gnu/$(R5_GCC_RELEASE)/binrel
+  HASH:=f5f3c1cfcb429833d363e8fec31bb1282974b119ca8169d6277ce8a549e26d54
```
openwrt-ai, line 35: A single `HASH` cannot be correct here: `R5_GCC_SOURCE` embeds `$(HOST_ARCH)` (line 25/26), so the downloaded tarball differs between an x86_64 and an aarch64 build host, and the download will abort with a hash mismatch on whichever host isn't the one you tested. The two existing in-tree consumers of the same ARM GNU toolchain drop select on `HOST_ARCH` — see [`arm-trusted-firmware-rockchip/Makefile:58-66`](https://github.com/openwrt/openwrt/blob/d8bd5e7012266c7f36eafca30c6729ecee6666a0/package/boot/arm-trusted-firmware-rockchip/Makefile#L58-L66) and [`arm-trusted-firmware-mvebu/Makefile:198-206`](https://github.com/openwrt/openwrt/blob/d8bd5e7012266c7f36eafca30c6729ecee6666a0/package/boot/arm-trusted-firmware-mvebu/Makefile#L198-L206) — please use the same `ifeq ($(HOST_ARCH),aarch64)` / `else` pair with both hashes for `arm-gnu-toolchain-12.3.rel1-{x86_64,aarch64}-arm-none-linux-gnueabihf.tar.xz`.

---

package/boot/ti-k3-firmware/Makefile
```diff
@@ -0,0 +1,55 @@
...
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=ti-k3-firmware
+PKG_RELEASE:=1
+
+PKG_SOURCE_URL:=https://github.com/TexasInstruments/ti-linux-firmware.git
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_VERSION:=d084a9f75833dc9184d2682bdfd668bf6808d1a3
+PKG_SOURCE_DATE:=2026-02-06
+PKG_MIRROR_HASH:=skip
```
openwrt-ai: `PKG_MIRROR_HASH:=skip` disables integrity verification of the mirrored git tarball — [`include/download.mk:160`](https://github.com/openwrt/openwrt/blob/d8bd5e7012266c7f36eafca30c6729ecee6666a0/include/download.mk#L160) explicitly short-circuits the hash comparison for that literal. It is a placeholder for `FIXUP=1` runs, not a shippable value, and no package in the tree currently carries it (`grep -r 'PKG_MIRROR_HASH:=skip'` returns nothing). Please fill in the real hash produced by `make package/ti-k3-firmware/check FIXUP=1`. This matters more than usual here because the payload is signed binary-only firmware that goes into the boot chain.

---

target/linux/ti-k3/image/Makefile
```diff
@@ -0,0 +1,54 @@
...
+  IMAGE/sysupgrade.img.gz := beagleplay-boot-common | beagleplay-sdcard | gzip | append-metadata
+endef
+
+define Device/beagleboard_am625-beagleplay
+  DEVICE_VENDOR := BeagleBoard.org
+  DEVICE_MODEL := BeaglePlay
+  DEVICE_DTS := k3-am625-beagleplay
+  UBOOT_DEVICE_NAME := beagleboard_am625-beagleplay
+  DEVICE_PACKAGES := kmod-wl18xx wpad-basic-openssl kmod-usb-net
+endef
```
openwrt-ai, line 52: `SUPPORTED_DEVICES` is left at its default, which [image.mk derives as `$(subst _,$(comma),$(1))`](https://github.com/openwrt/openwrt/blob/d8bd5e7012266c7f36eafca30c6729ecee6666a0/include/image.mk#L551) — i.e. `beagleboard,am625-beagleplay`. But `board_name` returns the first DT compatible, `beagle,am625-beagleplay` (see the kernel DTS [`k3-am625-beagleplay.dts:17`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/arch/arm64/boot/dts/ti/k3-am625-beagleplay.dts#L17), and your own `02_network` case label uses exactly that string). The `append-metadata` compat check will therefore reject the image on the running device. Add the correct value explicitly:

```suggestion
define Device/beagleboard_am625-beagleplay
  DEVICE_VENDOR := BeagleBoard.org
  DEVICE_MODEL := BeaglePlay
  DEVICE_DTS := k3-am625-beagleplay
  SUPPORTED_DEVICES := beagle,am625-beagleplay
  UBOOT_DEVICE_NAME := beagleboard_am625-beagleplay
  DEVICE_PACKAGES := kmod-wl18xx wpad-basic-openssl kmod-usb-net
endef
```

---

target/linux/ti-k3/image/Makefile
```diff
@@ -0,0 +1,54 @@
...
+
+define Device/Default
+  PROFILES := Default
+  KERNEL := kernel-bin
+  KERNEL_NAME := Image
+  KERNEL_LOADADDR :=
+  DEVICE_DTS_DIR := $(DTS_DIR)/ti
+  IMAGES := sysupgrade.img.gz
+  IMAGE/sysupgrade.img.gz := beagleplay-boot-common | beagleplay-sdcard | gzip | append-metadata
```
openwrt-ai, line 42: The target ships a `sysupgrade.img.gz` but adds no `target/linux/ti-k3/base-files/lib/upgrade/platform.sh`, so `platform_check_image` / `platform_do_upgrade` are undefined and [`sysupgrade:368`](https://github.com/openwrt/openwrt/blob/d8bd5e7012266c7f36eafca30c6729ecee6666a0/package/base-files/files/sbin/sysupgrade#L368) will bail out with "Firmware upgrade is not implemented for this platform." for every user of this image. Other SD-card targets with the same `rootfs-part boot-part` feature set provide one — [`target/linux/sunxi/base-files/lib/upgrade/platform.sh`](https://github.com/openwrt/openwrt/blob/d8bd5e7012266c7f36eafca30c6729ecee6666a0/target/linux/sunxi/base-files/lib/upgrade/platform.sh) is a close fit, or the shared helpers in `package/base-files/files/lib/upgrade/legacy-sdcard.sh`.

---

target/linux/ti-k3/image/Makefile
```diff
@@ -0,0 +1,54 @@
...
+	echo 'LABEL OpenWrt' > $@.boot/extlinux.conf.tmp
+	echo '  KERNEL /Image' >> $@.boot/extlinux.conf.tmp
+	echo '  FDT /$(DEVICE_DTS).dtb' >> $@.boot/extlinux.conf.tmp
+	echo '  APPEND console=ttyS2,115200n8 earlycon=ns16550a,mmio32,0x02800000 root=/dev/mmcblk1p2 rootwait rw' >> $@.boot/extlinux.conf.tmp
+	mkdir -p $@.boot/extlinux
+	mv $@.boot/extlinux.conf.tmp $@.boot/extlinux/extlinux.conf
+endef
+
+define Build/beagleplay-sdcard
+	KERNELPARTTYPE=c KERNELFATFS=1 KERNELFATBITS=32 ACTIVEPART=1 PADDING=1 $(SCRIPT_DIR)/gen_image_generic.sh \
```
openwrt-ai: `ACTIVEPART=1` is a no-op: ptgen already defaults to marking partition 1 active — [`active = 1` at src/ptgen.c:169](https://github.com/openwrt/firmware-utils/blob/3e6b426938e86efa0aaced903f6e22a1db2d2bc7/src/ptgen.c#L169), applied at [ptgen.c:349](https://github.com/openwrt/firmware-utils/blob/3e6b426938e86efa0aaced903f6e22a1db2d2bc7/src/ptgen.c#L349) (that SHA is the version pinned by `tools/firmware-utils/Makefile`). So the MBR already had 0x80 on the first partition before this PR, and the new `ACTIVEPART` plumbing in `gen_image_generic.sh` is only useful for marking a partition *other* than the first. Either drop it here, or keep the generic knob but reword commit bc84e82's justification, which currently claims it is needed "to mark a partition bootable/active without requiring GPT".

---

target/linux/ti-k3/base-files/etc/board.d/02_network
```diff
@@ -0,0 +1,21 @@
...
+. /lib/functions/uci-defaults.sh
+. /lib/functions/system.sh
+
+board_config_update
+
+board=$(board_name)
+
+case "$board" in
+beagle,am625-beagleplay)
+	ucidef_set_interface_wan "eth0"
```
openwrt-ai: nit: this leaves the board with no LAN interface at all, so there is no 192.168.1.1 fallback and first boot depends entirely on a DHCP server on eth0. Single-Ethernet boards on comparable targets default the port to LAN instead — e.g. [sunxi's `02_network` catch-all `ucidef_set_interface_lan "eth0"`](https://github.com/openwrt/openwrt/blob/d8bd5e7012266c7f36eafca30c6729ecee6666a0/target/linux/sunxi/base-files/etc/board.d/02_network#L26-L28). Also, the DTS exposes a second port (`ethernet1 = &cpsw_port2`, the DP83TD510E single-pair PHY) which isn't referenced here; the PR body says it's untested, but should it still get an `ucidef_set_interface_lan "eth1"` entry so it comes up configured? Is the WAN-only choice deliberate?

---

package/boot/uboot-ti/Makefile
```diff
@@ -0,0 +1,105 @@
...
+endef
+
+define U-Boot/Default
+  BUILD_TARGET:=ti_k3
+  HIDDEN:=1
+endef
+
+define U-Boot/beagleboard_am625-beagleplay
+  BUILD_SUBTARGET:=am625
+  NAME:=BeagleBoard BeaglePlay
```
openwrt-ai: nit: vendor spelling disagrees with the rest of the series — `target/linux/ti-k3/image/Makefile:46` uses `DEVICE_VENDOR := BeagleBoard.org`, and the kernel DTS `model` is `"BeagleBoard.org BeaglePlay"`. Align them:

```suggestion
  NAME:=BeagleBoard.org BeaglePlay
```

---

package/boot/optee-os-ti/Makefile
```diff
@@ -0,0 +1,49 @@
...
+include $(INCLUDE_DIR)/kernel.mk
+include $(INCLUDE_DIR)/optee-os.mk
+include $(INCLUDE_DIR)/package.mk
+
+define Optee-os/Default
+  BUILD_TARGET:=ti_k3
+endef
+
+define Optee-os/am62x
+  BUILD_SUBTARGET:=am625
```
openwrt-ai, line 29: nit: `NAME` is never set for this variant, so the generated package title becomes `OPTEE-OS for ` with a dangling "for" ([`include/optee-os.mk:62`](https://github.com/openwrt/openwrt/blob/d8bd5e7012266c7f36eafca30c6729ecee6666a0/include/optee-os.mk#L62)). Every variant in `optee-os-stm32` sets it, and your `arm-trusted-firmware-ti` sets `NAME:=TI K3 SoCs`.

```suggestion
  BUILD_SUBTARGET:=am625
  NAME:=TI AM62x
```

---

openwrt-ai (commented):
## Commit checks

- `bc84e82` "scripts: add MBR and forced-FAT32 boot support" — the body justifies `ACTIVEPART` as needed "to mark a partition bootable/active without requiring GPT", but ptgen already defaults to `active = 1` in MBR mode, so nothing changes for the first partition. See the inline comment on `target/linux/ti-k3/image/Makefile`.
- `a8781a0` "ti-k3: add new target for BeaglePlay" — the body says the config adds "enough UART slots for this SoC's 4 serial ports (CONFIG_SERIAL_8250_NR_UARTS/RUNTIME_UARTS)", but `am625/config-6.18` sets both to `12`, and AM62x has 7 main-domain UARTs plus the MCU/wakeup ones — the stated count matches neither the diff nor the hardware.


---

sayoojkkarun:
> ## Formality Check: Failed
> 
> We completed the verification flow. Please review the formatting overview logs below.
> ### 🛑 CRITICAL ERRORS
> 
> > **Commit [a8781a0](https://github.com/openwrt/openwrt/commit/a8781a069db92467d54f6d25cab19ef8e0afea15)** - _ti-k3: add new target for BeaglePlay_:
> > 
> > * New OpenWrt package is missing the mandatory parameter: 'PKG_MAINTAINER'
> > * New OpenWrt package is missing the mandatory parameter: 'PKG_LICENSE'
> > * New OpenWrt package is missing the mandatory parameter: 'PKG_LICENSE_FILES'
> 
> Tip
> 
> **Do not close this pull request** to make corrections. Instead, modify your existing commits (e.g. `git commit --amend`) and update the branch using `git push --force-with-lease --force-if-includes`. The checks will re-run automatically.
> 
> Something broken? Consider [reporting an issue](https://github.com/openwrt/openwrt-bot-worker/issues/new)._Running version [`d8f0c4e`](https://github.com/openwrt/openwrt-bot-worker/commit/d8f0c4ee161e78d63c347601237e3e62a6bc6d04) deployed on 2026-07-28 08:57:39 CEST_

This is a false positive. The commit adds target/linux/ti-k3/Makefile, which is a target definition, not a  software package. 
Target Makefiles never carry PKG_MAINTAINER/PKG_LICENSE fields (see any other target/linux/*/Makefile in the tree, so there is nothing to add here.

sayoojkkarun:
> ## Commit checks
> 
>     * `bc84e82` "scripts: add MBR and forced-FAT32 boot support" — the body justifies `ACTIVEPART` as needed "to mark a partition bootable/active without requiring GPT", but ptgen already defaults to `active = 1` in MBR mode, so nothing changes for the first partition. See the inline comment on `target/linux/ti-k3/image/Makefile`.
> 
>     * `a8781a0` "ti-k3: add new target for BeaglePlay" — the body says the config adds "enough UART slots for this SoC's 4 serial ports (CONFIG_SERIAL_8250_NR_UARTS/RUNTIME_UARTS)", but `am625/config-6.18` sets both to `12`, and AM62x has 7 main-domain UARTs plus the MCU/wakeup ones — the stated count matches neither the diff nor the hardware.
> 
> 
> _Generated by [Claude Code](https://claude.ai/code)_

Removed ACTIVEPART variable and updated commit message for it.
Removed the confusing number of UART statement from commit message as well.

package/boot/uboot-ti/Makefile
sayoojkkarun: Fixed. Split the hash by HOST_ARCH.

package/boot/ti-k3-firmware/Makefile
sayoojkkarun: Fixed.

target/linux/ti-k3/image/Makefile
sayoojkkarun: Fixed.

target/linux/ti-k3/image/Makefile
sayoojkkarun: Good catch. Added base files.

target/linux/ti-k3/image/Makefile
sayoojkkarun: Removed `ACTIVEPART` from the script.

target/linux/ti-k3/base-files/etc/board.d/02_network
sayoojkkarun: Removed eth0 as WAN and made it as LAN interface. Since eth1 is not tested, I am leaving it as not configured.

package/boot/uboot-ti/Makefile
sayoojkkarun: Added

package/boot/optee-os-ti/Makefile
sayoojkkarun: Fixed

target/linux/ti-k3/Makefile
```diff
@@ -0,0 +1,25 @@
...
+KERNEL_PATCHVER:=6.18
+
+define Target/Description
+	Build firmware images for TI K3-based boards (AM62x and later K3
+	generation SoCs).
+endef
+
+include $(INCLUDE_DIR)/target.mk
+
+DEFAULT_PACKAGES += uboot-envtools partx-utils e2fsprogs kmod-gpio-button-hotplug
```
openwrt-ai: The new `legacy-sdcard` sysupgrade path needs FAT support that this target does not ship. `beagleplay-sdcard` formats partition 1 as FAT32 (`KERNELPARTTYPE=c KERNELFATFS=1 KERNELFATBITS=32`), and [`legacy_sdcard_copy_config`](https://github.com/openwrt/openwrt/blob/ae7216292fcbd85800a3630635bd09a829244063/package/base-files/files/lib/upgrade/legacy-sdcard.sh#L81-L91) — called from the new `platform_copy_config` in `target/linux/ti-k3/base-files/lib/upgrade/platform.sh:10` — does `mount -o rw,noatime /dev/$partdev /boot` on exactly that partition. But `am625/config-6.18` has no `CONFIG_FAT_FS`/`CONFIG_VFAT_FS` (generic sets `# CONFIG_FAT_FS is not set`) and no vfat kmod is in `DEFAULT_PACKAGES`/`DEVICE_PACKAGES`, so the mount fails and the config backup is silently dropped on every sysupgrade.

Other FAT-boot SD-card targets pull the modules in at target level — see [`target/linux/bcm27xx/Makefile:29`](https://github.com/openwrt/openwrt/blob/ae7216292fcbd85800a3630635bd09a829244063/target/linux/bcm27xx/Makefile#L29) and [`target/linux/armsr/Makefile:19`](https://github.com/openwrt/openwrt/blob/ae7216292fcbd85800a3630635bd09a829244063/target/linux/armsr/Makefile#L19).

```suggestion
DEFAULT_PACKAGES += uboot-envtools partx-utils e2fsprogs kmod-gpio-button-hotplug \
	kmod-fs-vfat kmod-nls-cp437 kmod-nls-iso8859-1
```

---

openwrt-ai (commented):
## Commit checks

- `c712dea` "scripts: add MBR and forced-FAT32 boot support" — after `ACTIVEPART` was dropped, nothing MBR-specific is left in the diff. `gen_image_generic.sh` already emitted a plain MBR whenever `GUID` is unset, so the commit now only decouples the FAT code path from `GUID` (`KERNELFATFS`) and forces the FAT width (`KERNELFATBITS`). The subject still advertises "add MBR ... support"; something like "scripts: allow FAT kernel partition without GPT" would match what the two new variables actually do.
- `abdd3e6` "ti-k3: add new target for BeaglePlay" — the body no longer describes everything the commit does: it now also adds sysupgrade support (`legacy-sdcard` in `FEATURES` plus the new `base-files/lib/upgrade/platform.sh`), which is a distinct, user-visible piece of functionality and isn't mentioned anywhere in the message.

Re-reviewed the delta since `d8bd5e7`. The `HOST_ARCH`-split `Download/r5-gcc` hash, `NAME` fixes, `PKG_MIRROR_HASH`, `SUPPORTED_DEVICES` and the `02_network` LAN change all look correct and match the in-tree patterns. One new issue from the sysupgrade addition is inline.

---

target/linux/ti-k3/Makefile
sayoojkkarun: Added.

target/linux/ti-k3/image/Makefile
```diff
@@ -0,0 +1,55 @@
...
+  IMAGE/sysupgrade.img.gz := beagleplay-boot-common | beagleplay-sdcard | gzip | append-metadata
+endef
+
+define Device/beagleboard_am625-beagleplay
+  DEVICE_VENDOR := BeagleBoard.org
+  DEVICE_MODEL := BeaglePlay
+  DEVICE_DTS := k3-am625-beagleplay
+  SUPPORTED_DEVICES := beagle,am625-beagleplay
+  UBOOT_DEVICE_NAME := beagleboard_am625-beagleplay
+  DEVICE_PACKAGES := kmod-wl18xx wpad-basic-openssl kmod-usb-net luci
```
openwrt-ai: `luci` doesn't belong in `DEVICE_PACKAGES`. That list is for packages the *hardware* needs (`kmod-wl18xx`, `kmod-usb-net`, `wpad-*`); LuCI is a user/build-config choice selected via `CONFIG_PACKAGE_luci`. `grep -rn PACKAGES target/linux/ --include=*.mk --include=Makefile | grep -w luci` returns this line and nothing else — no other device or target in the tree pulls LuCI in this way, so this board would be the only one shipping uhttpd/rpcd/LuCI unconditionally (and growing the image accordingly) regardless of what the builder selected.

It also isn't covered by the commit message: `a04768e` now says "Add sysupgrade support." but says nothing about installing a web UI by default.

```suggestion
  DEVICE_PACKAGES := kmod-wl18xx wpad-basic-openssl kmod-usb-net
```

---

openwrt-ai (commented):
## Commit checks

- `a04768e` "ti-k3: add new target for BeaglePlay" — the body gained "Add sysupgrade support.", which covers the new `79_move_config` preinit hook and the `kmod-fs-vfat`/`kmod-nls-*` additions. It does not cover the other change in this round: `luci` was added to `DEVICE_PACKAGES`, which is unrelated to sysupgrade and is a user-visible default-image change. See the inline comment.

Re-reviewed the delta since `ae72162` (branch was rebased onto `b941179`; the seven PR commits are otherwise content-identical apart from `a04768e`). The `DEFAULT_PACKAGES` fix for FAT support is exactly what was needed, and the new `target/linux/ti-k3/base-files/lib/preinit/79_move_config` matches the in-tree pattern — it is the omap/sunxi `move_config` plus bcm27xx's `insmod nls_cp437/nls_iso8859-1/fat/vfat`, which is the right combination here since the kmods are modules rather than built-in; `export_partdevice` already validates `-b /dev/$DEVNAME`, so bcm27xx's extra `[ -b ... ]` guard isn't needed. Ordering also checks out: `do_stage2` runs `platform_do_upgrade` before `platform_copy_config`, so `legacy_sdcard_copy_config` writes the backup onto the freshly-written FAT partition, and preinit having already loaded `vfat` is what lets its typeless `mount` succeed. CI on `9f6ebd4` is green.

---

target/linux/ti-k3/image/Makefile
sayoojkkarun: Changed.

openwrt-ai (commented):

Re-reviewed the delta since `9f6ebd4`. The branch was rebased onto `2991205` (from `b941179`); `git range-diff` shows six of the seven PR commits content-identical, with the only change being `2c67d18` "ti-k3: add new target for BeaglePlay" dropping `luci` from `DEVICE_PACKAGES`, which is exactly the previously-flagged fix. `DEVICE_PACKAGES := kmod-wl18xx wpad-basic-openssl kmod-usb-net` is now hardware-only, and the commit message ("Add sysupgrade support.") again covers everything the commit does. The message/diff match holds for the other six commits as well.

No CI checks have reported on `142000058f78fcb3b83e05d4ad6f430eb4f30876` yet (status `pending`, 0 checks), so there is nothing to ground CI comments on.

---

openwrt-ai (commented):

Re-checked the delta since `1420000`. The branch was rebased from `2991205` onto `9f3157a`; `git range-diff 1420000~7..1420000 1da8989~7..1da8989` reports all seven commits content-identical (`1:` … `7:` all `=`), so there is no new code in this push. Nothing in the new base touches `scripts/gen_image_generic.sh`, `config/Config-images.in`, `package/boot/`, or `include/` in a way that interacts with this series.

CI on `1da8989` is clean (all three `FormalityCheck` runs `success`), so there is nothing to ground CI comments on. Note the `openwrt` bot's separate stylistic warning on `d609c58`: `PKG_LICENSE:=TI-Software-License-Agreement` in `package/boot/ti-k3-firmware/Makefile` is not an SPDX identifier and wants a `LicenseRef-` form — that one still stands.

---

openwrt-ai (commented):

Re-checked the delta since `1da8989`. `git range-diff 1da8989~7..1da8989 a14904e~7..a14904e` reports six of the seven commits content-identical; the only change is commit 4, `52ff867` "ti-k3-firmware: add SYSFW firmware for BeaglePlay", flipping `PKG_LICENSE:=TI-Software-License-Agreement` to `PKG_LICENSE:=LicenseRef-TI-Software-License-Agreement` in `package/boot/ti-k3-firmware/Makefile`. That is exactly the fix for the `openwrt` bot's outstanding stylistic warning, and `LicenseRef-TI-Software-License-Agreement` is a well-formed SPDX `LicenseRef-<idstring>` (only `[a-zA-Z0-9.-]` after the prefix). No other hunk in that commit moved, and its message still matches its diff; the same holds for the other six commits.

Note there is currently no other `LicenseRef-` user in the tree to compare against (`grep -rn LicenseRef` over the base checkout returns nothing), so this is the first one — worth a maintainer eyeball, but the form follows the SPDX spec.

No CI checks have reported on `a14904e1a5e58839381d4f6111bb2ff3442ddaaf` yet (status `pending`, 0 checks), so there is nothing to ground CI comments on.

---

package/boot/uboot-ti/Makefile
```diff
@@ -0,0 +1,109 @@
...
+define Download/r5-gcc
+  FILE:=$(R5_GCC_SOURCE)
+  URL:=https://developer.arm.com/-/media/Files/downloads/gnu/$(R5_GCC_RELEASE)/binrel
+ifeq ($(HOST_ARCH),aarch64)
+  HASH:=257db92f2aa5d83b2fc06c61ba328f927cf4c1b8b4021e9ba1a030ccfac045c6
+else
+  HASH:=f5f3c1cfcb429833d363e8fec31bb1282974b119ca8169d6277ce8a549e26d54
+endif
+endef
+
```
tmn505, line 38: It would be nice if this could be moved to new package. We already have similar construct for airoha and mvebu ATF packages. IMO this is not hard requirement for PR to get accepted, since GCC version for airoha differs, but if Your board firmware compiles and works with newer version then I think it's worth a shot.

scripts/gen_image_generic.sh
```diff
@@ -49,9 +49,9 @@ dos_dircopy() {
 [ -n "$PADDING" ] && dd if=/dev/zero of="$OUTPUT" bs=512 seek="$ROOTFSOFFSET" conv=notrunc count="$ROOTFSSIZE"
 dd if="$ROOTFSIMAGE" of="$OUTPUT" bs=512 seek="$ROOTFSOFFSET" conv=notrunc
 
-if [ -n "$GUID" ]; then
-    [ -n "$PADDING" ] && dd if=/dev/zero of="$OUTPUT" bs=512 seek="$((ROOTFSOFFSET + ROOTFSSIZE))" conv=notrunc count="$sect"
-    mkfs.fat --invariant -n kernel -C "$OUTPUT.kernel" -S 512 "$((KERNELSIZE / 1024))"
+if [ -n "$GUID" ] || [ -n "$KERNELFATFS" ]; then
+    [ -n "$GUID" ] && [ -n "$PADDING" ] && dd if=/dev/zero of="$OUTPUT" bs=512 seek="$((ROOTFSOFFSET + ROOTFSSIZE))" conv=notrunc count="$sect"
+    mkfs.fat --invariant -n kernel -C "$OUTPUT.kernel" -S 512 ${KERNELFATBITS:+-F "$KERNELFATBITS"} "$((KERNELSIZE / 1024))"
```
tmn505, line 54: Some time ago I've sent patch https://lists.openwrt.org/pipermail/openwrt-devel/2024-April/042561.html, which was trying to similarly extend the script without introducing new variables. Maybe You could simply use similar approach, depending on value of `KERNELPARTTYPE` for Your use case. Even if fs is made FAT16 ptget still, by most scripts in tree, forces the value for FAT32 in MBR, thus we won't trigger discarding config on sysupgrade for all current targets. So I think we can safely set FATBITS to corresponding value of `KERNELPARTTYPE`.

target/linux/ti-k3/base-files/etc/board.d/02_network
```diff
@@ -0,0 +1,21 @@
...
+. /lib/functions/uci-defaults.sh
+. /lib/functions/system.sh
+
+board_config_update
+
+board=$(board_name)
+
+case "$board" in
+beagle,am625-beagleplay)
+	ucidef_set_interface_lan "eth0"
```
tmn505, line 15: I think this is the default, so probably no need for this file.

target/linux/ti-k3/Makefile
```diff
@@ -0,0 +1,26 @@
...
+
+define Target/Description
+	Build firmware images for TI K3-based boards (AM62x and later K3
+	generation SoCs).
+endef
+
+include $(INCLUDE_DIR)/target.mk
+
+DEFAULT_PACKAGES += uboot-envtools partx-utils e2fsprogs kmod-gpio-button-hotplug \
+	kmod-fs-vfat kmod-nls-cp437 kmod-nls-iso8859-1
```
tmn505, line 22: kmod-nls-cp437 kmod-nls-iso8859-1 are pulled by kmod-fs-vfat no need to specify them.

tmn505 (changes requested):
Overall PR looks in good, hope committers will merge.

sayoojkkarun:
@tmn505 Thanks for the review. I will get back with the changes soon.

PR #24469, opened 2026-07-29
scripts: dhcpv6: add support for requesting multiple prefixes

### LINKED ISSUE #24431: scripts: dhcpv6 - unable to request for multiple PD
### Describe the bug

The commit below broke the ability to request for multiple PDs from DHCPv6 server.

https://github.com/openwrt/openwrt/commit/f08cd5ce5d66d76fa6e282d6e3f29bab9cdf8023

A `reqprefix` value of say `64 -P64` would previously result in 2 PDs of /64 length assigned by DHCPv6 server.  Since the above commits, any value saved for `reqprefix` will only result in one single PD request with supposedly IAID as part of the PD request.  This broke the ability to request for multiple PDs.

The above commit also broke the ability to request for `<prefix>/<length>` as the resulting `odhcp6c` parameter always results in a `-P0:<IAID>` generated.

### OpenWrt version

SNAPSHOT, 25.12.5

### OpenWrt release

snapshot, openwrt-25.12.05

### OpenWrt target/subtarget

x86/64, filogic/mt7622, ramips/mt7621, */*

### Device

All

### Image kind

Official downloaded image

### Steps to reproduce

Refer to bug description.

### Actual behaviour

Refer to bug description.

### Expected behaviour

Refer to bug description.

### Additional info

_No response_

### Diffconfig

```text

```

### Terms

Shine-:
Requesting multiple delegated prefixes (IA_PD) has been supported by `odhcp6c` for a [long time](https://github.com/openwrt/odhcp6c/commit/b146f9adc80cc2c2cdf2b04bfeec4c861a2a0e23) already. Enable this possibility in OpenWrt.

This PR adds support for multiple entries in the `network.<ifname>.reqprefix` option, either as list items or as a space-separated string, in the latter case keeping existing (single-PD) configurations, PPP and LuCI compatible.

Example `/etc/config/network` entries:
```
config interface 'wan6'
	[...]
	list reqprefix '64'
	list reqprefix '64'
```
```
config interface 'wan6'
	[...]
	option reqprefix '64 64'
```

Useful with ISPs that e.g. only hand out /64 prefixes, but do allow requesting multiple prefixes per customer.

Note: Each prefix request must use a unique IAID. In case none is explicitly specified by the user, I chose to concatenate n-times the i/f name to generate the hash used as IAID (e.g. `eth0` -> `a1a5b3d2` for the first, `eth0eth0` -> `d97eba46` for the second, etc.). This should minimize the possibility for collisions and at the same time, make the IAIDs reproducible. No IAID change will occur for existing single-PD setups.

Fixes: #24431

### DIFF
--- a/package/network/ipv6/odhcp6c/files/dhcpv6.sh
+++ b/package/network/ipv6/odhcp6c/files/dhcpv6.sh
@@ -10,7 +10,7 @@ proto_dhcpv6_init_config() {
 	renew_handler=1
 
 	proto_config_add_string 'reqaddress:or("try","force","none")'
-	proto_config_add_string reqprefix
+	proto_config_add_array 'reqprefix:list(string)'
 	proto_config_add_string clientid
 	proto_config_add_string 'sendclientid:or("auto","global","hardware")'
 	proto_config_add_string 'reqopts:list(uinteger)'
@@ -69,6 +69,31 @@ proto_dhcpv6_add_sendopts() {
 	[ -n "$1" ] && append "$3" "-x$1"
 }
 
+proto_dhcpv6_add_reqprefix() {
+	[ "$1" = "no" ] && return
+	local pfhint="$1"
+	[ -z "${pfhint##auto}" ] && pfhint=0
+	# append interface IAID if none specified
+	local iaid=$(echo -n $pfhint | sed -nr 's/^.*:([0-9A-Fa-f]{1,8})$/\1/p')
+	[ -z "$iaid" ] && {
+		# generate unique default IAID per prefix request
+		iaidmod="$iface$iaidmod"
+		network_generate_iface_iaid iaid "$iaidmod"
+		pfhint="$pfhint:$iaid"
+	}
+	# validate prefix/length hint
+	local hint=${pfhint%:$iaid}
+	[ "${hint#/}" -le "128" ] 2>/dev/null && {
+		pfhint=${pfhint#/}
+	} || {
+		validate_data cidr6 "$hint" 2>/dev/null || {
+			pfhint="0:$iaid"
+			logger -p warn -t dhcpv6 "$iface: ignoring invalid prefix hint '$1'"
+		}
+	}
+	append "$3" "-P$pfhint"
+}
+
 proto_dhcpv6_setup() {
 	local config="$1"
 	local iface="$2"
@@ -83,7 +108,7 @@ proto_dhcpv6_setup() {
 
 	local ip6prefix ip6prefixes
 
-	json_get_vars reqaddress reqprefix clientid sendclientid reqopts defaultreqopts
+	json_get_vars reqaddress clientid sendclientid reqopts defaultreqopts
 	json_get_vars noslaaconly forceprefix extendprefix norelease strict_rfc7550
 	json_get_vars noserverunicast noclientfqdn noacceptreconfig iface_dslite
 	json_get_vars iface_map iface_464xlat ip6ifaceid userclass vendorclass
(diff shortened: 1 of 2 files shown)


Shine-:
Ping @quarkysg

Found some time last night, so here you go.

Shine-:
heads-up @strrs 
If this PR is accepted, would also need to be ported to ucode.

strrs:
> heads-up @strrs 
> If this PR is accepted, would also need to be ported to ucode.

Thank you for heads-up, on it.

package/network/ipv6/odhcp6c/files/dhcpv6.sh
```diff
@@ -98,26 +123,8 @@ proto_dhcpv6_setup() {
...
-		} || {
-			validate_data cidr6 "$hint" 2>/dev/null || {
-				reqprefix="0:$iaid"
-				logger -p warn -t dhcpv6 "$iface: ignoring invalid prefix hint"
-			}
-		}
-		append opts "-P$reqprefix"
-	}
+	local iaidmod=''
+	json_for_each_item proto_dhcpv6_add_reqprefix reqprefix opts
```
openwrt-ai, line 127: An explicitly *empty* `option reqprefix ''` looks like it changes meaning here: from "request the default PD" to "request no PD at all".

`uci_to_blob()` splits the UCI string on whitespace and skips empty words, so a present-but-empty option becomes an **empty array**, not a missing key. `json_for_each_item` then takes the `array` branch, `json_get_keys` returns nothing, and the callback is never invoked — so no `-P` ends up in `opts`. Previously [`[ -z "$reqprefix" -o "$reqprefix" = "auto" ] && reqprefix=0`](https://github.com/openwrt/openwrt/blob/272eb50c5054bce4b02b25c211e876636f06cfa6/package/network/ipv6/odhcp6c/files/dhcpv6.sh#L101) turned that into `-P0:<iaid>`.

The *unset* case still behaves as before, because jshn falls through to the scalar branch and calls the callback once with an empty value, which `[ -z "${pfhint##auto}" ]` maps to `0`. So the two "no value configured" spellings now diverge. Is that intentional? If not, invoking the callback once when the array is empty (or restoring the old `-z` handling around this call) would keep them aligned.

---

package/network/ipv6/odhcp6c/files/dhcpv6.sh
```diff
@@ -10,7 +10,7 @@ proto_dhcpv6_init_config() {
 	renew_handler=1
 
 	proto_config_add_string 'reqaddress:or("try","force","none")'
-	proto_config_add_string reqprefix
+	proto_config_add_array 'reqprefix:list(string)'
```
openwrt-ai, line 13: nit: `reqprefix` is still fetched by [`json_get_vars` at dhcpv6.sh:111](https://github.com/openwrt/openwrt/blob/272eb50c5054bce4b02b25c211e876636f06cfa6/package/network/ipv6/odhcp6c/files/dhcpv6.sh#L111). Now that it is an array, jshn's `json_get_var` hands back the internal table id rather than the value, so the variable holds a meaningless number and is never read afterwards. Worth dropping it from that `json_get_vars` line — the other array-typed option, `ip6prefix`, is only declared `local` ([line 109](https://github.com/openwrt/openwrt/blob/272eb50c5054bce4b02b25c211e876636f06cfa6/package/network/ipv6/odhcp6c/files/dhcpv6.sh#L109)) and never fetched that way.

---

package/network/ipv6/odhcp6c/files/dhcpv6.sh
```diff
@@ -69,6 +69,31 @@ proto_dhcpv6_add_sendopts() {
 	[ -n "$1" ] && append "$3" "-x$1"
 }
 
+proto_dhcpv6_add_reqprefix() {
+	[ "$1" = "no" ] && return
+	local pfhint="$1"
+	[ -z "${pfhint##auto}" ] && pfhint=0
```
openwrt-ai, line 75: nit: this reads as an `auto` check but it silently doubles as the empty-value check — `${pfhint##auto}` is empty both for `auto` and for the empty string, and the empty case is what keeps an unconfigured `reqprefix` (jshn scalar fallback, value `""`) mapping to `-P0:<iaid>`. That dependency is easy to break in a later cleanup; spelling both cases out as the old code did keeps it explicit:

```suggestion
	[ -z "$pfhint" -o "$pfhint" = "auto" ] && pfhint=0
```

---

openwrt-ai (commented):
Reviewed 1 new commit; 3 inline comments. The commit message matches the diff and the `PKG_RELEASE` bump is present, so no commit-level findings. Backwards compatibility for `option reqprefix '64 64'` / `list reqprefix` looks right (netifd's `uci_to_blob()` splits whitespace for array-typed params), and the PPP `add_dynamic` path still works via jshn's scalar fallback in `json_for_each_item`.

---

package/network/ipv6/odhcp6c/files/dhcpv6.sh
Shine-: No, dear AI, your analysis is not correct.

Whether the `reqprefix` option is missing, or it's set to an empty string via `/etc/config/network` using `option reqprefix ''`, `proto_dhcpv6_add_reqprefix` will be called once either way, with an empty `$1` argument, resulting in the `auto` code path to become active.

package/network/ipv6/odhcp6c/files/dhcpv6.sh
Shine-: This is _intended_ to be an `auto` OR empty value check, and it's also intended to keep it clean and short.

package/network/ipv6/odhcp6c/files/dhcpv6.sh
Shine-: Oops, sorry, missed. Fixed.

openwrt-ai (commented):

The only change since the previous review is dropping `reqprefix` from the `json_get_vars` line, which resolves the earlier nit; the leftover `local reqprefix` declaration is harmless and matches how `ip6prefix` is handled. The commit message is unchanged and still matches the diff, and all checks on `ed29def` are green.

---

PR #24470, opened 2026-07-29
netfilter: add devgroup match module

JeThWifirst:
The devgroup match allows to match on the device group a network device is assigned to.

### DIFF
--- a/package/kernel/linux/modules/netfilter.mk
+++ b/package/kernel/linux/modules/netfilter.mk
@@ -1305,6 +1305,21 @@ endef
 
 $(eval $(call KernelPackage,ipt-rpfilter))
 
+define KernelPackage/ipt-devgroup
+  SUBMENU:=$(NF_MENU)
+  TITLE:=Netfilter devgroup match
+  DEPENDS:=+kmod-ipt-core
+  KCONFIG:=$(KCONFIG_IPT_DEVGROUP)
+  FILES:=$(LINUX_DIR)/net/netfilter/xt_devgroup.ko
+  AUTOLOAD:=$(call AutoProbe,xt_devgroup)
+  $(call KernelPackage/ipt)
+endef
+
+define KernelPackage/ipt-devgroup/description
+ Kernel module support for the devgroup match module
+endef
+
+$(eval $(call KernelPackage,ipt-devgroup))
 
 define KernelPackage/nft-core
   SUBMENU:=$(NF_MENU)
--- a/include/netfilter.mk
+++ b/include/netfilter.mk
@@ -257,6 +257,10 @@ $(eval $(call nf_add,NFNETLINK_QUEUE,CONFIG_NETFILTER_NETLINK_QUEUE, $(P_XT)nfne
 
 $(eval $(if $(NF_KMOD),$(call nf_add,NF_CONNCOUNT,CONFIG_NETFILTER_CONNCOUNT, $(P_XT)nf_conncount),))
 
+# devgroup
+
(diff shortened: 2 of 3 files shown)


package/kernel/linux/modules/netfilter.mk
```diff
@@ -1305,6 +1305,21 @@ endef
...
 $(eval $(call KernelPackage,ipt-rpfilter))
 
+define KernelPackage/ipt-devgroup
+  SUBMENU:=$(NF_MENU)
+  TITLE:=Netfilter devgroup match
+  DEPENDS:=+kmod-ipt-core
+  KCONFIG:=$(KCONFIG_IPT_DEVGROUP)
+  FILES:=$(LINUX_DIR)/net/netfilter/xt_devgroup.ko
+  AUTOLOAD:=$(call AutoProbe,xt_devgroup)
+  $(call KernelPackage/ipt)
```
openwrt-ai, line 1315: `$(call KernelPackage/ipt)` expands to nothing — there is no `KernelPackage/ipt` definition anywhere in the tree (`grep -rn "define KernelPackage/ipt$" .` returns nothing). The only two call sites are [`ipt-hashlimit` at netfilter.mk:1281](https://github.com/openwrt/openwrt/blob/002c4e5ff47e300aa40e11f16290c45fdd14da5f/package/kernel/linux/modules/netfilter.mk#L1274-L1282) and [`ipt-rpfilter` at netfilter.mk:1299](https://github.com/openwrt/openwrt/blob/002c4e5ff47e300aa40e11f16290c45fdd14da5f/package/kernel/linux/modules/netfilter.mk#L1293-L1301), which is presumably where this block was copied from; both work only because they also set `SUBMENU`/`DEPENDS` by hand. The helper that actually exists is [`AddDepends/ipt` at netfilter.mk:245](https://github.com/openwrt/openwrt/blob/002c4e5ff47e300aa40e11f16290c45fdd14da5f/package/kernel/linux/modules/netfilter.mk#L245-L248), used by the other ~26 `ipt-*` packages, e.g. [`kmod-ipt-u32`](https://github.com/openwrt/openwrt/blob/002c4e5ff47e300aa40e11f16290c45fdd14da5f/package/kernel/linux/modules/netfilter.mk#L923-L929).

Same file: deriving `FILES`/`AUTOLOAD` from `$(IPT_DEVGROUP-m)` (which the new `nf_add` entry in `include/netfilter.mk` already populates) keeps them in sync with the module list, which is what all the other `ipt-*` packages do — only `ipt-hashlimit` hardcodes the path.

```suggestion
  TITLE:=Netfilter devgroup match
  KCONFIG:=$(KCONFIG_IPT_DEVGROUP)
  FILES:=$(foreach mod,$(IPT_DEVGROUP-m),$(LINUX_DIR)/net/$(mod).ko)
  AUTOLOAD:=$(call AutoProbe,$(notdir $(IPT_DEVGROUP-m)))
  $(call AddDepends/ipt)
```


---

openwrt-ai (commented):
Reviewed 1 new commit; one inline comment. The kernel symbol (`CONFIG_NETFILTER_XT_MATCH_DEVGROUP`, `net/netfilter/xt_devgroup.c`) exists in both 6.12 and 6.18, is already listed as `is not set` in `target/linux/generic/config-6.12`/`config-6.18`, and its `NETFILTER_ADVANCED` dependency is satisfied transitively via `kmod-ipt-core` -> `kmod-nf-reject`, so the new kmod should build. The userspace side (`libxt_devgroup.so`) is shipped by iptables 1.8.10 and is picked up correctly by `BuildPlugin` through `$(IPT_DEVGROUP-m)`.

---

PR #24471, opened 2026-07-29
Xiaomi be7000 fixes

zeshi09:
qualcommbe: ipq95xx: add Xiaomi BE7000 support and fix QCA8084 XPCS SerDes

Add support for the Xiaomi Router BE7000 based on the Qualcomm IPQ9554
SoC, paired with an external 4-port 2.5GbE QCA8084 PHY and QCN5024/QCN6224
Wi-Fi 7 radios.

Hardware Specifications:
  - SoC: Qualcomm IPQ9554 (4x ARM Cortex-A73 @ 1.5 GHz)
  - RAM: 1 GB DDR4
  - Flash: 128 MB SPI NAND
  - Ethernet: 4x 2.5 Gbps RJ45 (QCA8084 4-port PHY package)
  - Wi-Fi: 2.4 GHz 4x4 (QCN5024), 5 GHz 4x4 (QCN6224)
  - Switch/PCS: Qualcomm PPE / UNIPHY0 (10G-QXGMII SerDes)
  - Expansion: 1x USB 3.0
  - Buttons: Reset (GPIO 44)
  - Serial: BLSP1 UART2 (115200 8N1, strictly 1.8V TTL level)

Ethernet & PHY Subsystem Fixes:
  - Fix XPCS channel array lookup bug in drivers/net/phy/qcom/qca8084_serdes.c:
    qca8084_qxgmii_set_speed() evaluated xpcs_data->xpcs_ch[channel] inside the
    loop instead of using iterator 'i'. If channel IDs were non-linear,
    improper SerDes clock rates and reset controls were applied to wrong
    channels, causing total traffic flow failure despite PHY carrier link UP.
  - Add 'managed = "in-band-status";' to port@1 through port@4 in DTS under
    &qcom_ppe to enable phylink state machine in-band negotiation with UNIPHY.
  - Enable &pcs0 (UNIPHY0 PCS Controller) node in DTS to ensure UNIPHY SerDes
    channels (pcs0_ch0 to pcs0_ch3) are probed and initialized by kernel.

Wi-Fi & Calibration Data Integration:
  - Enable &pcie2 and &pcie3 nodes for QCN5024 and QCN6224 PCIe enumeration.
  - Add ath12k-firmware-qcn9274 package to DEVICE_PACKAGES.
  - Add 11-ath12k-caldata hotplug script to extract Wi-Fi calibration BDF
    data from the '0:ART' MTD partition on boot, preventing ath12k board-2.bin
    fetch failures.

Platform Usability & Sysupgrade Safety:
  - Add memory node (1 GiB DDR4 RAM @ 0x40000000) to DTS.
  - Create /etc/fw_env.config mapping /dev/mtd1 (0:APPSBLENV) for uboot-envtools.
  - Add platform_pre_upgrade() in platform.sh to set required Xiaomi U-Boot
    persistence flags (boot_wait, uart_en, flag_boot_rootfs, flag_boot_success).
  - Fix WAN LED interface mapping in 01_leds (eth0 -> lan1).

Installation via U-Boot TFTP:
  1. Connect 1.8V TTL USB-UART adapter (3.3V WILL DAMAGE THE SoC).
  2. Load openwrt-qualcommbe-ipq95xx-xiaomi_be7000-initramfs-fit-uImage.itb to TFTP.
  3. Interrupt U-Boot prompt:
     setenv ipaddr 192.168.31.1
     setenv serverip 192.168.31.100
     tftpboot 0x48000000 openwrt-qualcommbe-ipq95xx-xiaomi_be7000-initramfs-fit-uImage.itb
     bootm 0x48000000

### DIFF
--- a/target/linux/qualcommbe/dts/ipq9574-be7000.dts
+++ b/target/linux/qualcommbe/dts/ipq9574-be7000.dts
@@ -1,0 +1,695 @@
+// SPDX-License-Identifier: (GPL-2.0+ OR BSD-3-Clause)
+
+/dts-v1/;
+
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+#include "ipq9574.dtsi"
+
+#include <dt-bindings/clock/qcom,qca8k-nsscc.h>
+#include <dt-bindings/net/qcom,qca808x.h>
+#include <dt-bindings/reset/qcom,qca8k-nsscc.h>
+
+/ {
+	model = "Xiaomi BE7000";
+	compatible = "xiaomi,be7000", "qcom,ipq9574";
+
+	aliases {
+		led-boot = &led_system_white;
+		led-failsafe = &led_system_amber;
+		led-running = &led_system_white;
+		led-upgrade = &led_system_amber;
+
+		serial0 = &blsp1_uart2;
+	};
+
+	chosen {
+		stdout-path = "serial0:115200n8";
+	};
+
+	memory@40000000 {
+		device_type = "memory";
+		reg = <0x0 0x40000000 0x0 0x40000000>;
+	};
+
+	regulator_fixed_3p3: s3300 {
+		compatible = "regulator-fixed";
+		regulator-min-microvolt = <3300000>;
+		regulator-max-microvolt = <3300000>;
+		regulator-boot-on;
+		regulator-always-on;
+		regulator-name = "fixed_3p3";
+	};
+
+	regulator_fixed_0p925: s0925 {
+		compatible = "regulator-fixed";
+		regulator-min-microvolt = <925000>;
+		regulator-max-microvolt = <925000>;
+		regulator-boot-on;
+		regulator-always-on;
+		regulator-name = "fixed_0p925";
+	};
+
+	gpio-keys {
+		compatible = "gpio-keys";
+		pinctrl-0 = <&gpio_keys_default>;
+		pinctrl-names = "default";
+
+		reset {
+			label = "reset";
+			linux,code = <KEY_RESTART>;
+			gpios = <&tlmm 44 GPIO_ACTIVE_LOW>;
+			linux,input-type = <EV_KEY>;
+			debounce-interval = <60>;
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+		pinctrl-0 = <&gpio_leds_default>;
+		pinctrl-names = "default";
+
+		aiot-white {
+			function = "aiot";
+			color = <LED_COLOR_ID_WHITE>;
+			gpios = <&tlmm 8 GPIO_ACTIVE_HIGH>;
+			default-state = "off";
+		};
+
+		led_system_white: system-white {
+			function = LED_FUNCTION_STATUS;
+			color = <LED_COLOR_ID_WHITE>;
+			gpios = <&tlmm 40 GPIO_ACTIVE_HIGH>;
+			default-state = "off";
+		};
+
+		led_system_amber: system-amber {
+			function = LED_FUNCTION_STATUS;
+			color = <LED_COLOR_ID_AMBER>;
+			gpios = <&tlmm 41 GPIO_ACTIVE_HIGH>;
+			default-state = "off";
+		};
+
+		network-white {
+			function = LED_FUNCTION_WAN;
+			color = <LED_COLOR_ID_WHITE>;
+			gpios = <&tlmm 42 GPIO_ACTIVE_HIGH>;
+			default-state = "off";
+		};
+
+		network-amber {
+			function = LED_FUNCTION_WAN;
+			color = <LED_COLOR_ID_AMBER>;
+			gpios = <&tlmm 43 GPIO_ACTIVE_HIGH>;
+			default-state = "off";
+		};
+	};
+
+	clocks {
+		qca8k_uniphy1_rx312p5m: qca8k-uniphy1-rx312p5m {
+			compatible = "fixed-clock";
+			clock-frequency = <312500000>;
+			#clock-cells = <0>;
+		};
+
+		qca8k_uniphy1_tx312p5m: qca8k-uniphy1-tx312p5m {
+			compatible = "fixed-clock";
+			clock-frequency = <312500000>;
+			#clock-cells = <0>;
+		};
+	};
+};
+
+
+
+&blsp1_uart2 {
+	pinctrl-0 = <&uart2_pins>;
+	pinctrl-names = "default";
+	status = "okay";
+};
+
+&qcom_ppe {
+	ethernet-ports {
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		port@1 {
+			reg = <1>;
+			phy-mode = "10g-qxgmii";
+			managed = "in-band-status";
+			label = "lan1";
+			phy-handle = <&phy0>;
+			pcs-handle = <&pcs0_ch0>;
+			clocks = <&nsscc NSS_CC_PORT1_MAC_CLK>,
+				 <&nsscc NSS_CC_PORT1_RX_CLK>,
+				 <&nsscc NSS_CC_PORT1_TX_CLK>;
+			clock-names = "port_mac",
+				      "port_rx",
+				      "port_tx";
+			resets = <&nsscc PORT1_MAC_ARES>,
+				 <&nsscc PORT1_RX_ARES>,
+				 <&nsscc PORT1_TX_ARES>;
+			reset-names = "port_mac",
+				      "port_rx",
+				      "port_tx";
+		};
+
+		port@2 {
+			reg = <2>;
+			phy-mode = "10g-qxgmii";
+			managed = "in-band-status";
+			label = "lan2";
+			phy-handle = <&phy1>;
+			pcs-handle = <&pcs0_ch1>;
+			clocks = <&nsscc NSS_CC_PORT2_MAC_CLK>,
+				 <&nsscc NSS_CC_PORT2_RX_CLK>,
+				 <&nsscc NSS_CC_PORT2_TX_CLK>;
+			clock-names = "port_mac",
+				      "port_rx",
+				      "port_tx";
+			resets = <&nsscc PORT2_MAC_ARES>,
+				 <&nsscc PORT2_RX_ARES>,
+				 <&nsscc PORT2_TX_ARES>;
+			reset-names = "port_mac",
+				      "port_rx",
+				      "port_tx";
+		};
+
+		port@3 {
+			reg = <3>;
+			phy-mode = "10g-qxgmii";
+			managed = "in-band-status";
+			label = "lan3";
+			phy-handle = <&phy2>;
+			pcs-handle = <&pcs0_ch2>;
+			clocks = <&nsscc NSS_CC_PORT3_MAC_CLK>,
+				 <&nsscc NSS_CC_PORT3_RX_CLK>,
+				 <&nsscc NSS_CC_PORT3_TX_CLK>;
+			clock-names = "port_mac",
+				      "port_rx",
+				      "port_tx";
+			resets = <&nsscc PORT3_MAC_ARES>,
+				 <&nsscc PORT3_RX_ARES>,
+				 <&nsscc PORT3_TX_ARES>;
+			reset-names = "port_mac",
+				      "port_rx",
+				      "port_tx";
+		};
+
+		port@4 {
+			reg = <4>;
+			phy-mode = "10g-qxgmii";
+			managed = "in-band-status";
+			label = "lan4";
+			phy-handle = <&phy3>;
+			pcs-handle = <&pcs0_ch3>;
+			clocks = <&nsscc NSS_CC_PORT4_MAC_CLK>,
+				 <&nsscc NSS_CC_PORT4_RX_CLK>,
+				 <&nsscc NSS_CC_PORT4_TX_CLK>;
+			clock-names = "port_mac",
+				      "port_rx",
+				      "port_tx";
+			resets = <&nsscc PORT4_MAC_ARES>,
+				 <&nsscc PORT4_RX_ARES>,
+				 <&nsscc PORT4_TX_ARES>;
+			reset-names = "port_mac",
+				      "port_rx",
+				      "port_tx";
+		};
+	};
+};
+
+&rpm_requests {
+	regulators {
+		compatible = "qcom,rpm-mp5496-regulators";
+
+		ipq9574_s1: s1 {
+			/*
+			 * During kernel bootup, the SoC runs at 800MHz with 875mV set by the bootloaders.
+			 * During regulator registration, kernel not knowing the initial voltage,
+			 * considers it as zero and brings up the regulators with minimum supported voltage.
+			 * Update the regulator-min-microvolt with SVS voltage of 725mV so that
+			 * the regulators are brought up with 725mV which is sufficient for all the
+			 * corner parts to operate at 800MHz
+			 */
+			regulator-min-microvolt = <725000>;
+			regulator-max-microvolt = <1075000>;
+		};
+
+		mp5496_l2: l2 {
+			regulator-min-microvolt = <1800000>;
+			regulator-max-microvolt = <1800000>;
+			regulator-always-on;
+			regulator-boot-on;
+		};
+
+		mp5496_l5: l5 {
+			regulator-min-microvolt = <1800000>;
+			regulator-max-microvolt = <1800000>;
+			regulator-always-on;
+			regulator-boot-on;
+		};
+	};
+};
+
+&sleep_clk {
+	clock-frequency = <32000>;
+};
+
+&pcie2_phy {
+	status = "okay";
+};
+
+&pcie2 {
+	pinctrl-0 = <&pcie2_default>;
+	pinctrl-names = "default";
+
+	perst-gpios = <&tlmm 29 GPIO_ACTIVE_LOW>;
+	wake-gpios = <&tlmm 30 GPIO_ACTIVE_LOW>;
+	status = "okay";
+
+	pcie@0,0 {
+		#address-cells = <3>;
+		#size-cells = <2>;
+		reg = <0 0 0 0 0>;
+
+		wifi6: pcie@0 {
+			compatible = "pci17cb,1109";
+			reg = <0 0 0 0 0>;
+		};
+	};
+};
+
+&pcie3_phy {
+	status = "okay";
+};
+
+&pcie3 {
+	pinctrl-0 = <&pcie3_default>;
+	pinctrl-names = "default";
+
+	perst-gpios = <&tlmm 32 GPIO_ACTIVE_LOW>;
+	wake-gpios = <&tlmm 33 GPIO_ACTIVE_LOW>;
+	status = "okay";
+
+	pcie@0,0 {
+		#address-cells = <3>;
+		#size-cells = <2>;
+		reg = <0 0 0 0 0>;
+
+		wifi7: pcie@0 {
+			compatible = "pci17cb,1109";
+			reg = <0 0 0 0 0>;
+		};
+	};
+};
+
+&tlmm {
+
+
+	gpio_keys_default: gpio-keys-default-state {
+		pins = "gpio44";
+		function = "gpio";
+		drive-strength = <8>;
+		bias-pull-up;
+	};
+
+	gpio_leds_default: gpio-leds-default-state {
+		pins = "gpio8", "gpio40", "gpio41", "gpio42", "gpio43";
+		function = "gpio";
+		drive-strength = <8>;
+		bias-pull-up;
+	};
+
+	qpic_snand_default_state: qpic-snand-default-state {
+		clock-pins {
+			pins = "gpio5";
+			function = "qspi_clk";
+			drive-strength = <8>;
+			bias-disable;
+		};
+
+		cs-pins {
+			pins = "gpio4";
+			function = "qspi_cs";
+			drive-strength = <8>;
+			bias-disable;
+		};
+
+		data-pins {
+			pins = "gpio0", "gpio1", "gpio2", "gpio3";
+			function = "qspi_data";
+			drive-strength = <8>;
+			bias-disable;
+		};
+	};
+
+	pcie2_default: pcie2-default-state {
+		clkreq-n-pins {
+			pins = "gpio28";
+			function = "pcie2_clk";
+			drive-strength = <6>;
+			bias-pull-up;
+		};
+
+		perst-n-pins {
+			pins = "gpio29";
+			function = "gpio";
+			drive-strength = <8>;
+			bias-pull-down;
+			output-low;
+		};
+
+		wake-n-pins {
+			pins = "gpio30";
+			function = "pcie2_wake";
+			drive-strength = <6>;
+			bias-pull-up;
+		};
+	};
+
+	pcie3_default: pcie3-default-state {
+		clkreq-n-pins {
+			pins = "gpio31";
+			function = "pcie3_clk";
+			drive-strength = <6>;
+			bias-pull-up;
+		};
+
+		perst-n-pins {
+			pins = "gpio32";
+			function = "gpio";
+			drive-strength = <8>;
+			bias-pull-up;
+			output-low;
+		};
+
+		wake-n-pins {
+			pins = "gpio33";
+			function = "pcie3_wake";
+			drive-strength = <6>;
+			bias-pull-up;
+		};
+	};
+};
+
+&mdio {
+	status = "okay";
+	clock-frequency = <6250000>;
+	reset-gpios = <&tlmm 60 GPIO_ACTIVE_LOW>;
+
+	qca8k_nsscc: clock-controller@18 {
+		compatible = "qcom,qca8084-nsscc";
+		reg = <0x18>;
+		clocks = <&cmn_pll ETH0_50MHZ_CLK>,
+			 <0>,
+			 <0>,
+			 <0>,
+			 <0>,
+			 <&qca8k_uniphy1_rx312p5m>,
+			 <&qca8k_uniphy1_tx312p5m>;
+		#clock-cells = <1>;
+		#reset-cells = <1>;
+	};
+
+	ethernet-phy-package@1 {
+		#address-cells = <1>;
+		#size-cells = <0>;
+		compatible = "qcom,qca8084-package";
+		reg = <1>;
+		clocks = <&qca8k_nsscc NSS_CC_APB_BRIDGE_CLK>,
+			 <&qca8k_nsscc NSS_CC_AHB_CLK>,
+			 <&qca8k_nsscc NSS_CC_SEC_CTRL_AHB_CLK>,
+			 <&qca8k_nsscc NSS_CC_TLMM_CLK>,
+			 <&qca8k_nsscc NSS_CC_TLMM_AHB_CLK>,
+			 <&qca8k_nsscc NSS_CC_CNOC_AHB_CLK>,
+			 <&qca8k_nsscc NSS_CC_MDIO_AHB_CLK>;
+		clock-names = "apb_bridge",
+			      "ahb",
+			      "sec_ctrl_ahb",
+			      "tlmm",
+			      "tlmm_ahb",
+			      "cnoc_ahb",
+			      "mdio_ahb";
+		resets = <&qca8k_nsscc NSS_CC_GEPHY_FULL_ARES>;
+		qcom,phy-addr-fixup = <1 2 3 4 5 6 7>;
+		qcom,package-mode = <QCA808X_PCS1_10G_QXGMII_PCS0_UNUNSED>;
+
+		phy0: ethernet-phy@1 {
+			compatible = "ethernet-phy-id004d.d180";
+			reg = <1>;
+			clocks = <&qca8k_nsscc NSS_CC_GEPHY0_SYS_CLK>;
+			resets = <&qca8k_nsscc NSS_CC_GEPHY0_SYS_ARES>;
+			qcom,xpcs-channel = <0>;
+		};
+
+		phy1: ethernet-phy@2 {
+			compatible = "ethernet-phy-id004d.d180";
+			reg = <2>;
+			clocks = <&qca8k_nsscc NSS_CC_GEPHY1_SYS_CLK>;
+			resets = <&qca8k_nsscc NSS_CC_GEPHY1_SYS_ARES>;
+			qcom,xpcs-channel = <1>;
+		};
+
+		phy2: ethernet-phy@3 {
+			compatible = "ethernet-phy-id004d.d180";
+			reg = <3>;
+			clocks = <&qca8k_nsscc NSS_CC_GEPHY2_SYS_CLK>;
+			resets = <&qca8k_nsscc NSS_CC_GEPHY2_SYS_ARES>;
+			qcom,xpcs-channel = <2>;
+		};
+
+		phy3: ethernet-phy@4 {
+			compatible = "ethernet-phy-id004d.d180";
+			reg = <4>;
+			clocks = <&qca8k_nsscc NSS_CC_GEPHY3_SYS_CLK>;
+			resets = <&qca8k_nsscc NSS_CC_GEPHY3_SYS_ARES>;
+			qcom,xpcs-channel = <3>;
+		};
+
+		pcs-phy@6 {
+			compatible = "qcom,qca8k-pcs-phy";
+			reg = <6>;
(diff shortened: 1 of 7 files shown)


target/linux/qualcommbe/patches-6.12/700-fix-qca8084-xpcs-channel-mapping.patch
```diff
@@ -0,0 +1,34 @@
+From 8f3b2a1c0d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8a Mon Sep 17 00:00:00 2001
```
openwrt-ai: This patch lands in a directory the build never looks at. `target/linux/qualcommbe/Makefile` sets [`KERNEL_PATCHVER:=6.18`](https://github.com/openwrt/openwrt/blob/c6c8986997d6106b41d6ab6c1449bc26dd89161e/target/linux/qualcommbe/Makefile#L11) and the target's patches live in `target/linux/qualcommbe/patches-6.18/`. There is no `patches-6.12` directory in this target, so the file is silently ignored and none of the ethernet behaviour the commit message attributes to it takes effect.

Also, the `From 8f3b2a1c0d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8a Mon Sep 17 00:00:00 2001` header is a made-up commit ID — please don't put a fabricated SHA in a patch header. The existing patches in `patches-6.18/` are real `git format-patch` output with `NNNN-<subject>.patch` names; a locally-authored fix should either follow that numbering or be folded into the patch that introduces the code.

---

target/linux/qualcommbe/patches-6.12/700-fix-qca8084-xpcs-channel-mapping.patch
```diff
@@ -0,0 +1,34 @@
...
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+--- a/drivers/net/phy/qcom/qca8084_serdes.c
++++ b/drivers/net/phy/qcom/qca8084_serdes.c
+@@ -647,7 +647,7 @@ void qca8084_qxgmii_set_speed(struct md
+ 	unsigned long rate;
+ 
+ 	for (i = 0; i < QCA8084_CHANNEL_MAX; i++) {
+-		xpcs_ch = &(xpcs_data->xpcs_ch[channel]);
++		xpcs_ch = &(xpcs_data->xpcs_ch[i]);
```
openwrt-ai: The bug this patch describes does not exist in this tree. The in-tree version of `qca8084_qxgmii_set_speed()` already indexes with the iterator — see [`xpcs_ch = &xpcs_data->xpcs_ch[i];` in 0312-net-phy-qca808x-Add-QCA8084-SerDes-speed-config.patch:105](``https://github.com/openwrt/openwrt/blob/c6c8986997d6106b41d6ab6c1449bc26dd89161e/target/linux/qualcommbe/patches-6.18/0312-net-phy-qca808x-Add-QCA8084-SerDes-speed-config.patch#L105).``

So the removed line here (`&(xpcs_data->xpcs_ch[channel])`, with the extra parentheses) never appears in the generated source. If this patch were moved to `patches-6.18/` it would fail to apply. Please drop it, or if the real fix targets something else in that function, re-base it on the actual in-tree code.

---

target/linux/qualcommbe/image/ipq95xx.mk
```diff
@@ -12,6 +12,22 @@ define Device/8devices_kiwi-dvk
...
+	DEVICE_VENDOR := Xiaomi
+	DEVICE_MODEL := BE7000
+	DEVICE_DTS_CONFIG := config@be7000
+	BLOCKSIZE := 128k
+	PAGESIZE := 2048
+	SOC := ipq9574
+	KERNEL_SIZE := 6096k
+	IMAGE_SIZE := 32116k
+	DEVICE_PACKAGES := kmod-ath12k ath12k-firmware-qcn9274
+	IMAGE/sysupgrade.bin := append-kernel | pad-to 64k | append-rootfs | pad-rootfs | check-size | append-metadata
```
openwrt-ai: This `IMAGE/sysupgrade.bin` override is incompatible with `Device/UbiFit` and with the `nand_do_upgrade` you added in `platform.sh`.

`Device/UbiFit` sets `KERNEL_IN_UBI := 1`, so the kernel is placed inside the UBI image and there is no separate kernel partition to `append-kernel` into. `nand_do_upgrade` expects the sysupgrade image to be a tar (or a raw UBI image), not a raw `kernel+rootfs` concatenation — a raw blob will be rejected at flash time. The `Device/Default` recipe in [`target/linux/qualcommbe/image/Makefile`](https://github.com/openwrt/openwrt/blob/c6c8986997d6106b41d6ab6c1449bc26dd89161e/target/linux/qualcommbe/image/Makefile#L11) already provides `sysupgrade-tar | append-metadata`, which is what every `Device/UbiFit` user in `qualcommax/image/ipq807x.mk` relies on. The line was copied from `Device/qcom_rdp433`, which does *not* use `UbiFit`.

Dropping the override also makes `KERNEL_SIZE` and the `check-size` against `IMAGE_SIZE` meaningless here (with `KERNEL_IN_UBI` there is no fixed kernel partition), so those two lines should go with it — leaving only `IMAGE_SIZE` if you want a UBI size cap.

---

target/linux/qualcommbe/image/ipq95xx.mk
```diff
@@ -12,6 +12,22 @@ define Device/8devices_kiwi-dvk
 endef
 TARGET_DEVICES += 8devices_kiwi-dvk
 
+define Device/xiaomi_be7000
```
openwrt-ai: nit: the file is ordered by vendor (`8devices_...`, then `qcom_rdp433`); inserting `xiaomi_be7000` between them breaks that. Please move the block after `Device/qcom_rdp433`.

---

target/linux/qualcommbe/dts/ipq9574-be7000.dts
```diff
@@ -0,0 +1,732 @@
...
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+		pinctrl-0 = <&gpio_leds_default>;
+		pinctrl-names = "default";
+
+		aiot-white {
+			label = "white:aiot";
```
openwrt-ai: The `label = "<color>:<function>"` syntax is deprecated in OpenWrt; since you already provide `color`, the remaining piece is `function`, and `label` should go.

```suggestion
			function = "aiot";
```

Same for the other four LEDs: line 82 → `function = LED_FUNCTION_STATUS;`, line 89 → `function = LED_FUNCTION_STATUS;`, line 96 → `function = LED_FUNCTION_WAN;`, line 103 → `function = LED_FUNCTION_WAN;` (constants from `dt-bindings/leds/common.h`, already included). Note that two LEDs sharing a `color`/`function` pair need distinguishing — e.g. keep amber as `LED_COLOR_ID_AMBER` with the same function, which yields distinct `amber:status` / `white:status` sysfs names.

This changes the sysfs LED names, so the name used in `01_leds` (`"white:network"`) has to be updated to match whatever the new `<color>:<function>` pair produces.

---

target/linux/qualcommbe/dts/ipq9574-be7000.dts
```diff
@@ -0,0 +1,732 @@
...
+
+&blsp1_i2c1 {
+	pinctrl-0 = <&i2c1_pins>;
+	pinctrl-names = "default";
+	status = "okay";
+
+	nfc: nfc@54 {
+		compatible = "miwifi,nfc";
+		reg = <0x54>;
+	};
```
openwrt-ai: `miwifi,nfc` has no binding and no driver anywhere in this tree or in Linux 6.18, so this node is dead weight — nothing will ever bind to it, and it will trip `dtc`/dt-schema checks for an undocumented compatible. Please drop the node (and `&blsp1_i2c1` + the `i2c1_pins` state with it, unless something else on the board needs the bus).

---

target/linux/qualcommbe/dts/ipq9574-be7000.dts
```diff
@@ -0,0 +1,732 @@
...
+
+	pcie@0,0 {
+		#address-cells = <3>;
+		#size-cells = <2>;
+		reg = <0 0 0 0 0>;
+
+		wifi6: pcie@0 {
+			compatible = "pci17cb,1109";
+			reg = <0 0 0 0 0>;
+			qcom,board_id = <0x04>;
```
openwrt-ai: `qcom,board_id` is not a property ath12k reads from DT. There is no such name in the ath12k bindings in Linux 6.18, and `grep -r qcom,board_id` over this tree returns nothing — the driver's `board_id` comes from OTP/QMI, not device tree. This property is therefore ignored.

If the intent is to select a per-board BDF, OpenWrt uses `qcom,calibration-variant` (see [`package/kernel/mac80211/patches/ath12k/700-ath12k-read-calibration-variant-from-dt.patch`](``https://github.com/openwrt/openwrt/blob/c6c8986997d6106b41d6ab6c1449bc26dd89161e/package/kernel/mac80211/patches/ath12k/700-ath12k-read-calibration-variant-from-dt.patch))`` together with a `board-2.bin` shipped via `package/firmware/ipq-wifi`.

Same applies to line 315.

---

target/linux/qualcommbe/dts/ipq9574-be7000.dts
```diff
@@ -0,0 +1,732 @@
...
+					compatible = "fixed-layout";
+					#address-cells = <1>;
+					#size-cells = <1>;
+
+					aqr_fw: aqr-fw@0 {
+						reg = <0x28 0x60002>;
+					};
+				};
+			};
+		};
```
openwrt-ai, line 640: The `0:ethphyfw` / `aqr_fw` nvmem block is copy-paste from the RDP433 DTS, which carries an Aquantia AQR PHY whose firmware is loaded from that partition (see `patches-6.18/0350-arm64-dts-qcom-add-AQR-NVMEM-node-for-IPQ9574-RDP433.patch`). This board uses a QCA8084 package, nothing in this DTS references `&aqr_fw`, and there is no reason to assume the Xiaomi flash layout even has a `0:ethphyfw` partition.

```suggestion
		partitions {
			compatible = "qcom,smem-part";
		};
```

---

target/linux/qualcommbe/dts/ipq9574-be7000.dts
```diff
@@ -0,0 +1,732 @@
...
+	clock-mult = <1>;
+};
+
+&xo_clk {
+	clock-frequency = <48000000>;
+};
+
+&pcs0 {
+	status = "okay";
+};
```
openwrt-ai: This override is a no-op, which makes the commit-message claim ("Enable `&pcs0` … to ensure UNIPHY SerDes channels are probed and initialized by kernel") inaccurate. The `pcs0`/`pcs1`/`pcs2` nodes added by [`0324-arm64-dts-qcom-ipq9574-add-PCS-uniphy-nodes.patch`](``https://github.com/openwrt/openwrt/blob/c6c8986997d6106b41d6ab6c1449bc26dd89161e/target/linux/qualcommbe/patches-6.18/0324-arm64-dts-qcom-ipq9574-add-PCS-uniphy-nodes.patch)`` carry no `status` property at all, so they default to `okay` already.

The `&pcs1`/`&pcs2` disables below *are* real changes; please reword the commit message so it describes what actually changes (disabling the unused uniphys) rather than an enable that has no effect.

---

target/linux/qualcommbe/dts/ipq9574-be7000.dts
```diff
@@ -0,0 +1,732 @@
...
+		};
+	};
+};
+
+&usb_0_dwc3 {
+	dr_mode = "host";
+};
+
+&usb_0_qmpphy {
+	/*vdda-pll-supply = <&mp5496_l5>;*/
```
openwrt-ai: nit: commented-out property left in the submitted DTS (same on line 688). Either wire `vdda-pll-supply` to the right regulator or delete the line — no other DTS in `target/linux/qualcommbe/dts/` carries dead commented-out properties.

---

target/linux/qualcommbe/ipq95xx/base-files/etc/hotplug.d/firmware/11-ath12k-caldata
```diff
@@ -0,0 +1,22 @@
...
+
+case "$FIRMWARE" in
+"ath12k/QCN9274/hw1.0/cal-pci-0000:02:00.0.bin"|\
+"ath12k/QCN9274/hw1.0/cal-pci-0000:03:00.0.bin"|\
+"ath12k/QCN6224/hw1.0/board-2.bin"|\
+"ath12k/QCN5024/hw1.0/board-2.bin")
+	case "$board" in
+	xiaomi,be7000)
+		caldata_extract "0:ART" 0x1000 0x20000
+		;;
```
openwrt-ai: Three problems in this case block:

1. `board-2.bin` is not per-device caldata — it is a multi-board BDF *container* keyed by board id, shipped by `ath12k-firmware-*` or overridden via `package/firmware/ipq-wifi`. Writing a 128 KiB blob extracted from `0:ART` to that path will not produce a parseable board-2.bin. No other target in this tree feeds `caldata_extract` output into a `board-2.bin` path; grep the `11-ath11k-caldata` scripts under `target/linux/qualcommax/*` for the established shape.
2. The `QCN6224`/`QCN5024` firmware paths will never be requested: both radios are declared as `compatible = "pci17cb,1109"` (QCN9274) in the DTS and `DEVICE_PACKAGES` only pulls `ath12k-firmware-qcn9274`. So those two entries are dead, while the commit message describes the board as QCN5024/QCN6224 — the DTS and the script disagree about which chips are on the board.
3. Both PCIe radios (`0000:02:00.0` and `0000:03:00.0`) get the *same* `0x1000`/`0x20000` region of `0:ART`. Two separate radios cannot share one caldata blob; the second one needs its own offset.

---

target/linux/qualcommbe/ipq95xx/base-files/etc/fw_env.config
```diff
@@ -0,0 +1,2 @@
+# MTD device name	Device offset	Env. size	Flash sector size
+/dev/mtd1		0x00000000	0x00010000	0x00020000
```
openwrt-ai: This is a target-wide base-files path, so it is installed on *every* ipq95xx device — 8devices Kiwi-DVK and RDP433 included — and hardcodes a board-specific `/dev/mtd1`. It also collides with `uboot-envtools`, which is in this target's `DEFAULT_PACKAGES` and generates `/etc/fw_env.config` at boot from `/etc/config/ubootenv` (see [`ubootenv_add_app_config` usage in `package/boot/uboot-tools/uboot-envtools/files/qualcommbe_ipq95xx`](``https://github.com/openwrt/openwrt/blob/c6c8986997d6106b41d6ab6c1449bc26dd89161e/package/boot/uboot-tools/uboot-envtools/files/qualcommbe_ipq95xx#L17)).``

Please drop this file and instead add a `xiaomi,be7000)` case to `package/boot/uboot-tools/uboot-envtools/files/qualcommbe_ipq95xx` alongside the existing `8devices,kiwi-dvk` entry, using `ubootenv_add_uci_config` with the `0:APPSBLENV` MTD *name* rather than a fixed `/dev/mtd1` index — the index depends on how many partitions SMEM reports and is not stable.

---

target/linux/qualcommbe/ipq95xx/base-files/etc/board.d/01_leds
```diff
@@ -0,0 +1,19 @@
...
+. /lib/functions/uci-defaults.sh
+
+board_config_update
+
+board=$(board_name)
+
+case "$board" in
+xiaomi,be7000)
+        ucidef_set_led_netdev "wan" "WAN" "white:network" "lan1"
+        ;;
```
openwrt-ai: Indented with spaces; every other `board.d` script in this target (and in the tree generally) uses tabs — compare the sibling `02_network` you edit in this same commit.

```suggestion
	ucidef_set_led_netdev "wan" "WAN" "white:network" "lan1"
	;;
```

---

target/linux/qualcommbe/ipq95xx/base-files/etc/board.d/01_leds
```diff
@@ -0,0 +1,19 @@
+#
+# Copyright (C) 2015 OpenWrt.org
+#
```
openwrt-ai: nit: `Copyright (C) 2015 OpenWrt.org` on a file created in 2026 is copy-paste from an old target. New `board.d` scripts in the tree are added without this header — please drop lines 1-3.

---

target/linux/qualcommbe/ipq95xx/base-files/lib/upgrade/platform.sh
```diff
@@ -8,13 +8,30 @@ platform_check_image() {
...
+platform_pre_upgrade() {
+	case "$(board_name)" in
+	xiaomi,be7000)
+		fw_setenv boot_wait on
+		fw_setenv uart_en 1
+		fw_setenv flag_boot_rootfs 0
+		fw_setenv flag_last_success 0
+		fw_setenv flag_boot_success 1
+		fw_setenv flag_try_sys1_failed 8
+		fw_setenv flag_try_sys2_failed 8
```
openwrt-ai, line 20: nit: the commit message says this sets "boot_wait, uart_en, flag_boot_rootfs, flag_boot_success", but the code also writes `flag_last_success`, `flag_try_sys1_failed` and `flag_try_sys2_failed`. Please list all of them (and ideally say what the `8` values mean), so a reader can tell why each variable is touched.

---

target/linux/qualcommbe/ipq95xx/base-files/etc/board.d/02_network
```diff
@@ -17,6 +17,9 @@ ipq95xx_setup_interfaces()
 	qcom,ipq9574-ap-al02-c7)
 		ucidef_set_interfaces_lan_wan "lan1 lan2 lan3 lan4 lan5" "wan"
 		;;
+	xiaomi,be7000)
+		ucidef_set_interfaces_lan_wan "lan2 lan3 lan4" "lan1"
+		;;
```
openwrt-ai, line 22: No MAC address is assigned anywhere for this board: there is no `ipq95xx_setup_macs()`/`ucidef_set_network_device_mac` here, and the DTS adds no `nvmem-cells`/`nvmem-cell-names = "mac-address"` on the PPE ports or the Wi-Fi nodes. The result is a random MAC on every boot for LAN, WAN and both radios, which also makes DHCP leases and any MAC-based provisioning unstable.

Please read the addresses out of the OEM layout (factory/ART partition offset, `board_data` text, or label MAC) and wire them up — via `nvmem-cells` in the DTS where possible, otherwise a `_setup_macs` case here. The commit message needs a matching "MAC address layout" section (one line per interface, `xx:xx:xx` for the unit-specific bytes).

---

openwrt-ai (commented):
## Commit checks

- `c6c8986` "qualcommbe: ipq95xx: add Xiaomi BE7000 support" — the message and the diff disagree in several places:
  - It describes a driver bug fix in `drivers/net/phy/qcom/qca8084_serdes.c` ("`qca8084_qxgmii_set_speed()` evaluated `xpcs_data->xpcs_ch[channel]` inside the loop"). That code does not exist in this tree — [`0312-net-phy-qca808x-Add-QCA8084-SerDes-speed-config.patch:105`](``https://github.com/openwrt/openwrt/blob/c6c8986997d6106b41d6ab6c1449bc26dd89161e/target/linux/qualcommbe/patches-6.18/0312-net-phy-qca808x-Add-QCA8084-SerDes-speed-config.patch#L105)`` already uses the iterator. The patch is also filed under `patches-6.12/` while the target builds 6.18, so it is never applied either way (inline).
  - "Enable `&pcs0` … to ensure UNIPHY SerDes channels … are probed" — the `pcs*` nodes have no `status` property in the base DTSI and are already enabled; the effective change is disabling `&pcs1`/`&pcs2` (inline).
  - "Fix WAN LED interface mapping in 01_leds (eth0 -> lan1)" — `01_leds` is a brand new file in this same commit, so nothing is being fixed; please phrase it as adding the LED config.
  - The `platform_pre_upgrade()` list in the message omits three of the seven variables the code actually sets (inline).
- Required sections for a new-device commit: **Hardware specification** is present, but it omits LED count/colours, button count and the power input rating. **Flash instructions** stop after `bootm` of the initramfs — there is no step that actually installs OpenWrt (sysupgrade of the `-squashfs-sysupgrade.bin`), and no revert-to-stock note. **MAC address layout** is missing entirely; it is also missing from the code (inline on `02_network`).
- Subject scope: the commit is `qualcommbe: ipq95xx: add Xiaomi BE7000 support`, but the diff also carries a kernel patch claiming to fix a generic QCA8084 driver bug. If a real driver fix is needed it belongs in its own commit; the PR title ("Xiaomi be7000 fixes") also doesn't match the single commit's subject.


---

zeshi09:
comments was resolved, please review again

target/linux/qualcommbe/ipq95xx/base-files/etc/hotplug.d/firmware/11-ath12k-caldata
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
+[ -e /lib/firmware/$FIRMWARE ] && exit 0
+
+. /lib/functions/caldata.sh
+
+board=$(board_name)
+
+case "$FIRMWARE" in
+"ath12k/QCN9274/hw1.0/cal-pci-0000:02:00.0.bin")
```
openwrt-ai, line 10: Wrong firmware directory: OpenWrt ships QCN9274 firmware under `hw2.0`, not `hw1.0`, so this `$FIRMWARE` string is never requested and the caldata is never extracted. See [`qca_ath12k.mk:13`](https://github.com/openwrt/openwrt/blob/4bb761d08b661ecc4ee4938e3d0ed685c5d4687a/package/firmware/linux-firmware/qca_ath12k.mk#L13) (`ath12k-firmware-qcn9274`, the package you add to `DEVICE_PACKAGES`) and [`ipq-wifi/Makefile:171`](https://github.com/openwrt/openwrt/blob/4bb761d08b661ecc4ee4938e3d0ed685c5d4687a/package/firmware/ipq-wifi/Makefile#L171) — both install into `ath12k/QCN9274/hw2.0/` only.

`hw1.0` is a real ath12k variant ([`hw.c` `.dir = "QCN9274/hw1.0"`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/net/wireless/ath/ath12k/hw.c#L1323)) but it is selected only for `soc_hw_version_major == 1` silicon, and OpenWrt ships no firmware for it at all, so a `hw1.0` sample would not come up regardless.

```suggestion
"ath12k/QCN9274/hw2.0/cal-pci-0000:02:00.0.bin")
```

Same fix needed on line 17 (`.../hw1.0/cal-pci-0000:03:00.0.bin` → `hw2.0`).

---

target/linux/qualcommbe/ipq95xx/base-files/lib/upgrade/platform.sh
```diff
@@ -8,13 +8,30 @@ platform_check_image() {
...
+
 platform_do_upgrade() {
 	case "$(board_name)" in
 	8devices,kiwi-dvk)
 		CI_KERNPART="0:HLOS"
 		CI_ROOTPART="rootfs"
 		emmc_do_upgrade "$1"
 		;;
+	xiaomi,be7000)
+		nand_do_upgrade "$1"
```
openwrt-ai, line 33: Is the bare `nand_do_upgrade` correct for this board's OEM partition layout? With no `CI_*_UBIPART` set, [`nand.sh:12`](https://github.com/openwrt/openwrt/blob/4bb761d08b661ecc4ee4938e3d0ed685c5d4687a/package/base-files/files/lib/upgrade/nand.sh#L12) defaults to `CI_UBIPART="ubi"`, and since a Xiaomi SMEM layout has no partition called `ubi` it silently falls back to `rootfs`. That only works if kernel and rootfs really do live in one UBI on this device.

The conflicting evidence is that the `fw_setenv` block you added just above is the same one used for the Xiaomi ipq807x boards, and those need three extra assignments right after it because the kernel is in a *separate* UBI — [`qualcommax/ipq807x/.../platform.sh:259`](``https://github.com/openwrt/openwrt/blob/4bb761d08b661ecc4ee4938e3d0ed685c5d4687a/target/linux/qualcommax/ipq807x/base-files/lib/upgrade/platform.sh#L259):``

```
# Kernel and rootfs are placed in 2 different UBI
CI_KERN_UBIPART="ubi_kernel"
CI_ROOT_UBIPART="rootfs"
CI_DATA_UBIPART="rootfs"
```

Your `Device/UbiFit` + `KERNEL_IN_UBI := 1` image assumes a single UBI instead. Also, those boards need a `xiaomi_initramfs_prepare`-style step to wipe the leftover OEM UBI volumes on the first flash from initramfs; nothing equivalent is here. If the BE7000 layout genuinely differs (single UBI, nothing to wipe), a one-line comment naming the target partition would save the next reader the same doubt.

---

target/linux/qualcommbe/dts/ipq9574-be7000.dts
```diff
@@ -0,0 +1,695 @@
...
+		qca8k_uniphy1_tx312p5m: qca8k-uniphy1-tx312p5m {
+			compatible = "fixed-clock";
+			clock-frequency = <312500000>;
+			#clock-cells = <0>;
+		};
+	};
+};
+
+
+
```
openwrt-ai, line 126: Leftover blank lines from removing the `&blsp1_i2c1` block — three consecutive empty lines here where one is intended.

```suggestion
```

Two more spots from the same cleanup: line 5 (double blank between `/dts-v1/;` and the first `#include`) and lines 310-311 (two blank lines between `&tlmm {` and its first child, where the sibling [`ipq9570-kiwi-dvk.dts`](https://github.com/openwrt/openwrt/blob/4bb761d08b661ecc4ee4938e3d0ed685c5d4687a/target/linux/qualcommbe/dts/ipq9570-kiwi-dvk.dts#L269-L270) has none).

---

target/linux/qualcommbe/ipq95xx/base-files/etc/board.d/02_network
```diff
@@ -17,15 +17,37 @@ ipq95xx_setup_interfaces()
...
+{
+	local board="$1"
+	local lan_mac=""
+	local wan_mac=""
+
+	case "$board" in
+	xiaomi,be7000)
+		wan_mac=$(mtd_get_mac_ascii "0:APPSBLENV" "ethaddr")
+		[ -z "$wan_mac" ] && wan_mac=$(mtd_get_mac_binary "0:ART" 0x0)
+		lan_mac=$(macaddr_add "$wan_mac" 1)
```
openwrt-ai, line 39: If both reads on lines 37-38 come back empty, `macaddr_add "" 1` still yields a non-empty garbage string (the `$((0x... + 1))` arithmetic on an empty value), so the `[ -n "$lan_mac" ]` guard below passes and a bogus MAC is written to UCI. Guard the derivation instead:

```suggestion
		[ -n "$wan_mac" ] && lan_mac=$(macaddr_add "$wan_mac" 1)
```

---

target/linux/qualcommbe/dts/ipq9574-be7000.dts
```diff
@@ -0,0 +1,695 @@
...
+&qcom_ppe {
+	ethernet-ports {
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		port@1 {
+			reg = <1>;
+			phy-mode = "10g-qxgmii";
+			managed = "in-band-status";
+			label = "lan1";
```
openwrt-ai, line 142: nit: this port is the board's WAN — [`02_network:21`](``https://github.com/openwrt/openwrt/blob/4bb761d08b661ecc4ee4938e3d0ed685c5d4687a/target/linux/qualcommbe/ipq95xx/base-files/etc/board.d/02_network#L21``) passes `lan1` as the wan argument and [`01_leds`](https://github.com/openwrt/openwrt/blob/4bb761d08b661ecc4ee4938e3d0ed685c5d4687a/target/linux/qualcommbe/ipq95xx/base-files/etc/board.d/01_leds#L10) binds the WAN LED to `lan1`. The only other in-tree DTS for this target labels its WAN port `wan` ([`ipq9570-kiwi-dvk.dts:84`](https://github.com/openwrt/openwrt/blob/4bb761d08b661ecc4ee4938e3d0ed685c5d4687a/target/linux/qualcommbe/dts/ipq9570-kiwi-dvk.dts#L84)), which also keeps the LAN ports numbered `lan1..lan3` instead of `lan2..lan4`. Is the `lan1` naming deliberate (e.g. matching the silkscreen), or would `label = "wan"` plus renumbering be preferable?

---

target/linux/qualcommbe/dts/ipq9574-be7000.dts
```diff
@@ -0,0 +1,695 @@
...
+	wake-gpios = <&tlmm 30 GPIO_ACTIVE_LOW>;
+	status = "okay";
+
+	pcie@0,0 {
+		#address-cells = <3>;
+		#size-cells = <2>;
+		reg = <0 0 0 0 0>;
+
+		wifi6: pcie@0 {
+			compatible = "pci17cb,1109";
```
openwrt-ai, line 279: nit: the PR description is now stale relative to this commit and contradicts the code here. It still says "QCN5024/QCN6224 Wi-Fi 7 radios" and lists `ath12k/QCN6224` / `ath12k/QCN5024` caldata paths, whereas both radios are declared as `pci17cb,1109` (QCN9274) and the commit message correctly says QCN9274. It also still advertises the `drivers/net/phy/qcom/qca8084_serdes.c` XPCS fix, `Create /etc/fw_env.config mapping /dev/mtd1`, and `Enable &pcs0 ... in DTS` — all three were dropped in the force-push. Please refresh the PR body to match the commit message so the merge summary is accurate.

---

target/linux/qualcommbe/image/ipq95xx.mk
```diff
@@ -27,3 +27,17 @@ define Device/qcom_rdp433
...
+	DEVICE_VENDOR := Xiaomi
+	DEVICE_MODEL := BE7000
+	DEVICE_DTS_CONFIG := config@be7000
+	BLOCKSIZE := 128k
+	PAGESIZE := 2048
+	SOC := ipq9574
+	IMAGE_SIZE := 32116k
+	DEVICE_PACKAGES := kmod-ath12k ath12k-firmware-qcn9274
+endef
+TARGET_DEVICES += xiaomi_be7000
```
openwrt-ai, line 43: Both "Determine Changed Packages" jobs fail on this head:

```
Run CHANGED_PACKAGES="$(./shared-workflows/.github/scripts/get_changed_packages.pl ".github/workflows/formal.yml .github/workflows/issue-labeller.yml .github/workflows/labeler.yml package/boot/uboot-airoha/patches/500-net-airoha-bind-MDIO-controller-on-Ethernet...
get_changed_packages.pl: Argument list too long
##[error]Process completed with exit code 126.
```

The argument list contains files this PR does not touch (`.github/workflows/*`, `package/boot/uboot-airoha/patches/*`), i.e. the whole delta between the branch point and current `main`. The branch is based on `7ba45de827`, at which point this target was still on kernel 6.12 — the tree at this head has `target/linux/qualcommbe/patches-6.12/` and `KERNEL_PATCHVER:=6.12`, while `main` now has `patches-6.18` / `KERNEL_PATCHVER:=6.18`. Please rebase onto current `main` and re-verify the DTS builds against `patches-6.18` (the labels it overrides — `qcom_ppe`, `pcs0_ch0..3`, `pcs1`, `pcs2`, `nsscc`, `cmn_pll` — all still exist there, but the kernel bump is worth a build check) so this job can run.

---

openwrt-ai (commented):
## Commit checks

- `4bb761d` "qualcommbe: ipq95xx: add Xiaomi BE7000 support" — the rewrite resolved most of the earlier message/diff mismatches: the QCA8084 SerDes patch is gone, the inaccurate "`Enable &pcs0`" claim is replaced by an accurate "disable unused UNIPHY1/UNIPHY2", `01_leds` is no longer described as a fix, the `platform_pre_upgrade()` variable list is now complete (including what the `8` means), and a **MAC address layout** section was added. Two of the three required new-device sections are still incomplete:
  - **Hardware specification** — no LED inventory (the DTS adds five: white AIoT, white/amber system, white/amber network), no flash chip vendor/part, no power input rating (voltage + current), and the UART line gives level and baud but not the header location on the PCB.
  - **Flash instructions** — the procedure stops after `bootm` of the initramfs. There is no step that actually installs OpenWrt (`sysupgrade` of `openwrt-qualcommbe-ipq95xx-xiaomi_be7000-squashfs-sysupgrade.bin`), and no revert-to-stock note. As written, a stranger following it ends up with a RAM-only system and no way back.
- Subject scope: the PR title "Xiaomi be7000 fixes" no longer matches the content — this is a single new-device commit with no fix commits. Worth retitling to the commit subject before merge.

---

PR #24473, opened 2026-07-29
base-files: sysupgrade: shut the network down gracefully

### LINKED ISSUE #24461: PPPoE WAN takes 30-70s longer to reconnect after sysupgrade than after a plain reboot
**Observed behavior**

Reconnect time for a PPPoE WAN differs depending on how the router came back up:

- Plain `reboot`: PPPoE reconnects in ~5s.
- `sysupgrade` (NAND/UBI via `nand.sh`): first PPPoE Discovery attempt gets no PADO response for ~34s, times out, and the automatic retry connects in ~5s. Reproduced twice; one run showed two consecutive ~35s timeouts (~70s total) before connecting.

**Isolated variables that turned out NOT to explain it** (each tested independently, no sysupgrade/reboot involved):

- Missing PADT: a plain `SIGTERM` to pppd already triggers a graceful "Sent PADT" (confirmed directly). `stage2`'s `kill_remaining` sends `SIGTERM` to every remaining process, including pppd, well before the actual flash, so PADT is very likely already sent during a normal sysupgrade too.
- A long idle gap on its own: `ifdown wan` (PADT sent), 90s idle with the physical link untouched, `ifup wan` → reconnects in ~5s, no stall.

So neither of those alone reproduces it. The stall only shows up after an actual `sysupgrade`, which (unlike a plain reboot) also holds the physical WAN link down for longer (NAND write time added before the reboot). Posting the measurement since it's reproducible and possibly relevant to others on PPPoE WANs — no working theory yet on the actual cause.

**OpenWrt version:** self-built, tracking `openwrt/main`
**OpenWrt target/subtarget:** qualcommax/ipq807x
**Image kind:** Self-built image

cc @dangowrt @DragonBluep — flagging in case either of you has seen this pattern before (sysupgrade timing vs. the pppoe proto handler are the two subsystems in play, no evidence yet pointing at either specifically).

JuliusBairaktaris:
Fixes https://github.com/openwrt/openwrt/issues/24461.

procd's sysupgrade path stops every service with a bare SIGTERM and never runs their
`shutdown()` hooks, unlike a reboot, which reaches them through inittab. netifd's SIGTERM
handler ends its uloop and then kills every process netifd started, so pppd never gets to
send a PADT, the peer keeps the session half open and the reconnect after the upgrade
stalls until it times out:

    netifd main.c, netifd_kill_process():
        kill(proc->uloop.pid, SIGKILL);

The hook that tears the interfaces down properly already exists, is the same one a reboot
runs on every device, and nothing on the flash path calls it:

    package/network/config/netifd/files/etc/init.d/network:
        shutdown() {
            ifdown -a
            sleep 1
        }

So the patch runs that hook from sysupgrade before the ubus handover, while netifd is
still alive, and restarts the network if the handover fails, so a rejected image cannot
leave the device unreachable. `ifup -a` would not do as the restore: it also raises
interfaces configured with `option auto 0` and leaves that flag flipped.

Measured on an ipq807x board behind a PPPoE WAN, three runs of the two commands the patch
adds, detached and with no image written, which is the rejected-image path: both returned
0, shutdown in 1 to 2 s and restart in 1 s, with `Terminating on signal 15` and
`Sent PADT` in the syslog and the WAN back up every run. I have not re-run the end-to-end
arm for this version; the table is one sysupgrade per arm, a single reading:

| teardown | pppd attempts | stall (1 sysupgrade each) |
| --- | --- | --- |
| before | 3, two ending in `Timeout waiting for PADO packets` | 71 s |
| after | 1, connects on the first PADI | 5 s |

This replaces the earlier version of the PR, which changed stage2's `skip_services`. That
loop cannot do the same job: by the time `upgraded` execs stage2, procd has already
stopped ubusd, so every `ubus call service delete` there fails with `Failed to connect to
ubus`. On the failsafe path initd execs `upgraded` in place of `/sbin/procd`, so no ubusd
runs there at all and the loop is inert. The skip list is untouched now.

### DIFF
--- a/package/base-files/files/sbin/sysupgrade
+++ b/package/base-files/files/sbin/sysupgrade
@@ -450,5 +450,12 @@ else
 	[ $USE_CURR_PART -eq 1 ] && json_add_boolean use_curr_part 1
 	json_close_object
 
+	[ -x /etc/init.d/network ] && /etc/init.d/network shutdown
+
 	ubus call system sysupgrade "$(json_dump)"
+	upgrade_rc=$?
+
+	[ -x /etc/init.d/network ] && /etc/init.d/network restart
+
+	exit $upgrade_rc
 fi


package/base-files/files/lib/upgrade/stage2
```diff
@@ -140,7 +140,7 @@ while read -r a b c; do
 done < /proc/meminfo
 
 [ "$mem" -gt 32768 ] && \
-	skip_services="dnsmasq log network"
+	skip_services="dnsmasq log"
```
openwrt-ai: `ubus call service delete` is not the "same graceful teardown" as the `reboot` path this is compared to, so as written the PPPoE session still gets no chance to send a PADT.

procd's delete handler only flushes the instance list: [`service_delete()` at service/service.c:203-212](https://github.com/openwrt/procd/blob/60fdbf00e924fff0a510c4ec31c520d35e69bfe1/service/service.c#L203-L212) → `vlist_flush_all(&s->instances)` → [`instance_stop(in, true)` at service/instance.c:894-905](https://github.com/openwrt/procd/blob/60fdbf00e924fff0a510c4ec31c520d35e69bfe1/service/instance.c#L894-L905), which clears `restart`/`respawn` and sends a bare `SIGTERM`. It never invokes the init script, so [`stop_service()` in /etc/init.d/network](https://github.com/openwrt/openwrt/blob/15a79ec2762b620b59dcb3e215b290805058e517/package/network/config/netifd/files/etc/init.d/network#L39-L43) (`wifi down; ifdown -a; sleep 1`) does not run. That is exactly what distinguishes it from the `reboot` path: [`::shutdown:/etc/init.d/rcS K shutdown`](https://github.com/openwrt/openwrt/blob/15a79ec2762b620b59dcb3e215b290805058e517/package/base-files/files/etc/inittab#L2) runs the script's [`shutdown()` hook](https://github.com/openwrt/openwrt/blob/15a79ec2762b620b59dcb3e215b290805058e517/package/network/config/netifd/files/etc/init.d/network#L150-L153), i.e. a real `ifdown -a`.

netifd's own SIGTERM handling does not compensate: [`netifd_handle_signal()`](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/main.c#L314-L318) only calls `uloop_end()`, `main()` then goes straight to [`netifd_kill_processes()` at main.c:414-415](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/main.c#L414-L415), and [`netifd_kill_process()`](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/main.c#L262-L270) sends **SIGKILL** — there is no interface teardown anywhere on the exit path. pppd is a netifd-tracked process ([`proto_run_command "$config" /usr/sbin/pppd` at ppp.sh:149](https://github.com/openwrt/openwrt/blob/15a79ec2762b620b59dcb3e215b290805058e517/package/network/services/ppp/files/ppp.sh#L149)), so it is SIGKILLed here too and the *original* session is abandoned just as before. The change does remove the second, respawn-created session, but not the half-open-session cause it claims to fix.

For a genuinely graceful teardown the `ifdown -a` has to happen while netifd is still alive — e.g. run `/etc/init.d/network stop` before the `ubus call service delete` loop, which is [`stop_service()` followed by `procd_kill` in rc.common:156-160](https://github.com/openwrt/openwrt/blob/15a79ec2762b620b59dcb3e215b290805058e517/package/base-files/files/etc/rc.common#L156-L160), giving both the graceful ifdown *and* the disarmed respawn. Leaving `network` out of `skip_services` on top of that then becomes a harmless no-op.

Side note supporting that this delete path is not the graceful one: on `mem <= 32768` devices `skip_services` is unset, so `network` has always been torn down exactly this way there — with no `ifdown -a` either.

---

openwrt-ai (commented):
## Commit checks
- 15a79ec "base-files: sysupgrade: don't let netifd respawn during teardown" — the diff does what the subject says (drops `network` from `skip_services`), and the respawn analysis checks out (procd's default `respawn_timeout` is 5s, so the respawn lands inside the `sleep 4` window before `kill_remaining KILL`). But the body's central claim — that this gives netifd "the same graceful ubus-delete teardown ... as every other service", in contrast to `reboot`'s `ifdown -a` — is not what the code does: `ubus call service delete` never runs the init script's `stop_service()`/`shutdown()`, and netifd SIGKILLs pppd on exit, so no PADT is sent on this path either. Details inline.

---

mcprat:
I have no opinion on the change, except that maybe without network there is also no point to keeping dnsmasq in the list as well

the only reason I made that exception list is because members told me some people do remote logging during sysupgrade

I'm wondering if the exclusion list can be conditional, does pppd have it's own init.d script? if so that would make it easy

JuliusBairaktaris:
I changed the PR to draft as bring up still sometimes hangs. Will investigate further when I have the time.

mcprat:
you may also want to try letting netifd not be terminated at all...

however it would be nice to find another way, the problem may be deep within netifd itself

mcprat:
this looks much cleaner to me now, but it would be nice if you can somehow find a conditional, if nothing else can serve as a suitable way to detect usage of pppd then it can be through configuration

package/base-files/files/sbin/sysupgrade
```diff
@@ -450,5 +450,9 @@ else
 	[ $USE_CURR_PART -eq 1 ] && json_add_boolean use_curr_part 1
 	json_close_object
 
+	[ -x /etc/init.d/network ] && /etc/init.d/network shutdown
+
 	ubus call system sysupgrade "$(json_dump)"
+
+	[ -x /etc/init.d/network ] && ifup -a
```
openwrt-ai: The recovery line now determines the script's exit status, which masks a failed upgrade.

`ubus call system sysupgrade` only returns when procd rejected the request; on success it never returns. So the *only* time control reaches line 457 is the failure path — and there `ifup -a` almost always succeeds, so `sysupgrade` exits **0** after a failed upgrade. Before this patch the ubus call was the last command and its non-zero status propagated. Anything that shells out to `/sbin/sysupgrade` and checks `$?` (LuCI, owut, attended-sysupgrade wrappers, plain `ssh root@host sysupgrade ...`) will now read a rejected image as a successful flash.

The mirror-image case is also wrong: on a system without netifd, `[ -x /etc/init.d/network ]` is false, so the `&&` list yields 1 and `sysupgrade` exits 1 even though nothing failed.

Capturing the status and exiting with it fixes both:

```suggestion
	ubus call system sysupgrade "$(json_dump)"
	upgrade_rc=$?

	[ -x /etc/init.d/network ] && ifup -a

	exit $upgrade_rc
```


---

openwrt-ai: `ifup -a` is not a faithful undo of `ifdown -a` — it brings up interfaces that were deliberately configured down.

[`ifup` iterates every registered interface](https://github.com/openwrt/openwrt/blob/286baafbf9809f172a1065893fd95089b5f100c3/package/network/config/netifd/files/sbin/ifup#L33-L37) from `ubus list 'network.interface.*'` and calls `network.interface up` on each, and [`interface_set_up()` sets `iface->autostart = true`](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/interface.c#L1364) unconditionally. So an interface with `option auto '0'` (a backup WAN, a maintenance VLAN, an LTE failover) that was down before the aborted sysupgrade comes up afterwards, and its `auto` flag stays flipped until netifd is restarted. `/etc/init.d/network restart` would restore exactly the boot-time state instead — heavier, but it is the actual inverse of the teardown. Is force-upping everything intended here?

Separately, the guard tests `/etc/init.d/network` but the command run is `/sbin/ifup`; `[ -x /sbin/ifup ]` would be the matching test (netifd ships both, so this is cosmetic).

---

openwrt-ai (commented):
Reviewed the rewritten approach (`15a79ec` → `286baaf`). The commit message matches the diff, and the core mechanism checks out: [`shutdown()` in the network init script](https://github.com/openwrt/openwrt/blob/286baafbf9809f172a1065893fd95089b5f100c3/package/network/config/netifd/files/etc/init.d/network#L150-L153) is `ifdown -a; sleep 1` and is *not* overridden by the `USE_PROCD` block in `rc.common`, so netifd stays alive for the teardown, as the message claims. Two issues with the new lines, inline.

FYI on the bot's `PKG_RELEASE` warning: `package/base-files/Makefile` sets `PKG_RELEASE:=$(COMMITCOUNT)`, so no manual bump is needed here.

---

JuliusBairaktaris:
> this looks much cleaner to me now, but it would be nice if you can somehow find a conditional, if nothing else can serve as a suitable way to detect usage of pppd then it can be through configuration

If acceptable i would like to leave the handling broad as pptp, l2tp and wwan (+possibly others) should also be broken.

JuliusBairaktaris:
@aparcar @efahl — this changes what `/sbin/sysupgrade` returns on a rejected image (the ubus status rather than the recovery `ifup`), so it touches owut's caller contract and would benefit from your review.

package/base-files/files/sbin/sysupgrade
```diff
@@ -450,5 +450,12 @@ else
 	[ $USE_CURR_PART -eq 1 ] && json_add_boolean use_curr_part 1
 	json_close_object
 
+	[ -x /etc/init.d/network ] && /etc/init.d/network shutdown
```
openwrt-ai: The teardown now runs *before* the last remaining rejection point, so the exit status this commit preserves is unobservable to exactly the callers that would check it.

procd only decides whether it accepts the image inside the ubus handler: `validate_firmware_image_call()` and the `valid`/`forceable`/`allow_backup` checks all sit in [`sysupgrade()` at system.c:794-822](https://github.com/openwrt/procd/blob/60fdbf00e924fff0a510c4ec31c520d35e69bfe1/system.c#L794-L822), i.e. after line 453 has already run `ifdown -a`. A remote invocation (`ssh root@host sysupgrade …`, owut, a LuCI/rpcd-driven upgrade) loses its transport at line 453; when procd then rejects the image, both the `sysupgrade_error()` message on stdout and `$upgrade_rc` go to a socket nobody is reading any more. Before this patch the rejection was reported over a still-live session and the box stayed up.

Not necessarily a blocker — the script already ran `/usr/libexec/validate_firmware_image` itself at line ~330, so procd's re-validation normally agrees and the window is narrow — but it does mean the commit message's "callers that check what sysupgrade returned still see a rejected upgrade" only holds for local/serial callers. Is that the intended trade, or is it worth moving the shutdown behind a first non-destructive procd round-trip?

---

package/base-files/files/sbin/sysupgrade
```diff
@@ -450,5 +450,12 @@ else
...
+
 	ubus call system sysupgrade "$(json_dump)"
+	upgrade_rc=$?
+
+	[ -x /etc/init.d/network ] && /etc/init.d/network restart
```
openwrt-ai: `/etc/init.d/network restart` is the right inverse of the `shutdown` above (thanks for the swap — `stop()` in rc.common runs `stop_service()` + `procd_kill`, then `start()` re-registers netifd, so the `auto '0'` flipping is gone), but it only covers one of the two ways the ubus call can return non-zero.

procd runs [`service_stop_all()` at system.c:825](https://github.com/openwrt/procd/blob/60fdbf00e924fff0a510c4ec31c520d35e69bfe1/system.c#L825) *before* [`sysupgrade_exec_upgraded()` at system.c:827](https://github.com/openwrt/procd/blob/60fdbf00e924fff0a510c4ec31c520d35e69bfe1/system.c#L827), and that helper returns to the handler when `chroot()`/[`execvp()` at sysupgrade.c:103](https://github.com/openwrt/procd/blob/60fdbf00e924fff0a510c4ec31c520d35e69bfe1/sysupgrade.c#L103) fails (missing `/sbin/upgraded` in the RAM root, OOM). In that case the handler still returns `UBUS_STATUS_UNKNOWN_ERROR`, so this line runs — but by then dropbear, uhttpd, the firewall and the inittab consoles ([`procd_inittab_kill()` at service/service.c:712](https://github.com/openwrt/procd/blob/60fdbf00e924fff0a510c4ec31c520d35e69bfe1/service/service.c#L712)) are already stopped, and restarting only `network` leaves the device with interfaces up and nothing listening on them. The PR body's "undoes it if the handover fails so a rejected image cannot leave the device unreachable" therefore holds for the validation rejections (which return before line 825) but not for the exec failure.

Worth a note in the commit message about the limitation, or handling that path separately (`reboot` is arguably the only honest recovery once `service_stop_all()` has run)?

---

openwrt-ai (commented):
Reviewed the delta since `286baaf` (`286baaf` → `f6fed1a`, force-push to a single commit). Both previous findings are addressed correctly: `upgrade_rc` now carries the ubus status through to `exit`, and `/etc/init.d/network restart` replaces `ifup -a` so interfaces with `option auto '0'` are no longer force-upped. The rewritten commit message matches the diff.

Two remaining observations about the failure path, inline — neither is a regression from the previous revision, both are about how far the new recovery line actually gets.

---

efahl:
> it touches owut's caller contract

It looks like it should not affect `owut`, as I'm specifically not checking status returned by `sysupgrade` when it's run for the actual upgrade (only the `--test` runs are checked).  `owut` just backgrounds `sysupgrade` and exits, letting `sysupgrade` do whatever it's going to do at that point, assuming that the immediately-preceding `sysupgrade -T` will report any meaningful errors before we try launch the upgrade.

openwrt-ai (commented):

`f6fed1a` → `a341a00` is a rebase onto current `main` plus a commit-message edit — `git diff f6fed1a a341a00` is empty, the seven added lines in `package/base-files/files/sbin/sysupgrade` are byte-identical. The message change is the new "Two limits of the recovery line" paragraph, which documents both of the observations from my previous review (validation-order vs. the remote caller's transport, and `service_stop_all()` having already run when the `upgraded` exec fails), and moves `Signed-off-by:` to last. Both descriptions match what the code does, so my two open inline threads are answered in-tree rather than left implicit — no further action needed on them from my side.

For the record on the two things checked this round: `Fixes: https://github.com/openwrt/openwrt/issues/24461` is an established trailer form here (159 commits on `main` use it), and CI on `a341a00` reports nothing failed.

---

JuliusBairaktaris:
@mcprat The rewrite dropped `skip_services` altogether so there is no list left to make conditional, and I would rather keep the teardown unconditional since `ifdown -a` is the right exit for a DHCP or IPv6 WAN too, not just PPPoE.

@efahl Thanks — that matches the code, and the preserved status is really for `ssh root@host sysupgrade …` rather than for owut.

mcprat:
what I mean by conditional is to have a system dependent decision on whether to do this

for example, similar to what I did in my very old commit that was quoted, services were skipped from being deleted based on the memory of a device being higher than 32 MB

JuliusBairaktaris:
@mcprat on what kind of property should it be depended? netifd has child processes?

mcprat:
if that can be determined in a script sure

JuliusBairaktaris:
@mcprat The property that would actually predict it is "some interface runs a protocol whose peer needs to be told" - PPPoE, PPTP, L2TP, wwan - and that is most WAN protocols rather than a narrow subset, so the conditional would be true on nearly every router that has a WAN at all.

Against that, the unconditional cost is bounded: `/etc/init.d/network shutdown` is `ifdown -a; sleep 1`, on a box that is about to be flashed. Your old memory check gated something open-ended; this is one second. Which device class are you worried about - if it is the <= 32 MB boxes, those never had `skip_services` set and so already tear the network down on this path.

mcprat:
not worried about anything in particular I just consider specificity as a general rule because I've learned you never know if theres a corner case or an exception should exist for some reason. trying to maintain status quo as much as possible is preferred

I understand pppoe is popular but I've never personally dealt with it since DOCSIS became a thing. every gateway I've handled in the last decade just hands out a private, or if configured, a public IP over DHCP and is not affected by this as far as I know.

Djfe:
the original intent was, to keep network alive for as long as possible and the new additional intent is that it should be shutdown properly before the sysupgrade completes with a reboot, right?

originally network wouldn't be shutdown properly because of service_stop_all()
can't we change this behavior and try a way to call the services' shutdown hooks first and give a timeout before then calling service_stop_all()? (atleast for the network service)

edit: if `ubus call service delete` is the only way, then ubusd could be skipped as well and only shutdown later. unless that process consumes too many ressources

JuliusBairaktaris:
1. Yes

2&3 will try it out and report back

mcprat:
network script `shutdown()` is effectively just `ifdown -a`, what if you just did `ifdown wan`?

then "remote" but local logging can still take place, and restarting the interfaces would not be necessary to try to "restore" the previous state

mcprat:
even more so, I believe netifd uses ubus in the scripts it operates to get a message back to itself

try listning to ubus on one ssh session
`ubus listen`

and teardown all interfaces on the other
`ifdown -a`

and see what happens?
(might just be a race condition with network dying...listen over serial?)

JuliusBairaktaris:
I dont have  serial access. Can you run that part?

JuliusBairaktaris:
> does pppd have it's own init.d script? if so that would make it easy

No. The ppp package installs `files/ppp.sh` as a netifd proto handler and
ships nothing under `/etc/init.d`, so pppd is a child netifd forks from that
handler and there is no service to key a conditional on:

    package/network/services/ppp/Makefile, Package/ppp/install:
        $(INSTALL_DIR) $(1)/lib/netifd/proto
        $(INSTALL_BIN) ./files/ppp.sh $(1)/lib/netifd/proto/

> maybe without network there is also no point to keeping dnsmasq in the list as well

The whole `skip_services` list is moot, not just `network`. ubusd is itself a
procd service, and `service_stop_all()` stops it inside the sysupgrade ubus
call, before procd execs `upgraded`, so every `ubus call service delete` in
stage2 fails with `Failed to connect to ubus`, in or out of the list. logd
goes the same way, so the remote-logging case the list was written for does
not survive either:

    procd state.c, state_enter(), case STATE_UBUS:
        service_start_early("ubus", ubus_cmd, p?"ubus":NULL, p?"ubus":NULL);

    procd system.c, sysupgrade():
        service_stop_all();

> network script `shutdown()` is effectively just `ifdown -a`, what if you just did `ifdown wan`?

Keeping the LAN address up is nicer, I agree with that. What stops me is that
it hardcodes the name `wan`, which multi-WAN, `wan6` and `wwan` setups do not
use, and that it does not remove the restore, because down is sticky and a
rejected image would leave that interface down until something brings it back:

    netifd interface.c, interface_set_down():
        iface->autostart = false;

JuliusBairaktaris:
> can't we change this behavior and try a way to call the services' shutdown hooks first and give a timeout before then calling service_stop_all()? (atleast for the network service)

For the network service that is what the patch does already, from the shell
side. `/etc/init.d/network shutdown` is the same hook a reboot reaches, and
`rc.common`'s `USE_PROCD` block does not redefine `shutdown`:

    package/network/config/netifd/files/etc/init.d/network:
        shutdown() {
            ifdown -a
            sleep 1
        }

Doing the same inside procd for every service is a larger change than it
looks: procd would have to run the whole `K*` pass on the flash path, and the
sysupgrade ubus handler would have to become asynchronous to hold a timeout.
That is my understanding of it, I have not tried to write it.

> if `ubus call service delete` is the only way, then ubusd could be skipped as well and only shutdown later

`ubus call service delete` cannot run a shutdown hook at all. It ends in a
plain SIGTERM:

    procd service/instance.c, instance_stop():
        kill(in->proc.pid, SIGTERM);

Sparing ubusd would also be a procd change, since `service_stop_all()` is what
stops it.

On what the hook costs: netifd's ubus `down` handler returns without waiting
for the proto teardown, so the price is the `sleep 1` and the ubus round
trips, not the time pppd takes to go away. I have not measured an in-procd
variant; the `ubus listen` run is in my reply to mcprat on that question.

    netifd ubus.c, netifd_handle_down():
        interface_set_down(iface);

        return 0;

JuliusBairaktaris:
> (might just be a race condition with network dying...listen over serial?)

I asked you to run this because I have no serial access, and that was the
wrong call - serial is not needed for it. `ubus listen`, `logread -f` and the
teardown all run detached on the board, with `/etc/init.d/network restart`
three seconds behind the shutdown and an unconditional restart watchdog behind
that, so the capture does not depend on ssh staying up. Three runs on an
ipq8074 board behind a PPPoE WAN, snapshot r36061.

`ubus listen` keeps receiving throughout: 46 to 48 events per run, with the
`ifdown` notifications for lan, loopback, wan and wan_6 arriving before the
`network.interface` objects are removed and re-added. No teardown event is
lost; there is no race that loses the teardown.

`/etc/init.d/network shutdown` returned 0 in 1 to 2 s in every run, and the
following `/etc/init.d/network restart` returned 0 in 1 s.

The syslog carries the same lines in every run, which is the point of the
patch:

    pppd[...]: Terminating on signal 15
    Connection terminated.
    Sent PADT
    Exit.

The WAN came back up in all three runs, with a new session and a new address.

openwrt-ai (commented):
## Commit checks
- `688fe3c` "base-files: sysupgrade: shut the network down gracefully" — "restart the network if the handover fails so a rejected image cannot leave the device unreachable" is now unqualified, and does not hold when the exec fails: procd runs [`service_stop_all()` at system.c:825](https://github.com/openwrt/procd/blob/2c1834abae171261f2d9161e5a28c36cbcc394a4/system.c#L825) before [`sysupgrade_exec_upgraded()`](https://github.com/openwrt/procd/blob/2c1834abae171261f2d9161e5a28c36cbcc394a4/system.c#L827), which [returns to the handler on exec failure](https://github.com/openwrt/procd/blob/2c1834abae171261f2d9161e5a28c36cbcc394a4/sysupgrade.c#L103-L106), so line 458 brings only netifd back with nothing else listening. Re-add the qualifier that `a341a00` carried, or drop the absolute wording.
- `688fe3c` — "callers that check what sysupgrade returned still see a rejected upgrade" holds for local callers only: procd decides validity inside the ubus handler, after [line 453](https://github.com/openwrt/openwrt/blob/688fe3ce6232f7e644fc10d6093e86c31d976e29/package/base-files/files/sbin/sysupgrade#L453) already ran `ifdown -a`, so a remote caller has lost its transport before `$upgrade_rc` exists.

---

mcprat:
sorry I keep forgetting to try the serial ubus log experiment...

I still think this is a strange workaround no matter what that teaches us

when I have enough free time I'm going to look at the pppd signal handler which is likely just missing some steps...

JuliusBairaktaris:
Yeah its strange. 

Dont worry take your time.


And thanks!

PR #24474, opened 2026-07-29
ramips: add support for COMFAST CF-E455AC V2

rikvdb229:
COMFAST CF-E455AC V2 is a dual-band 802.11ac ceiling access point based on
MediaTek MT7621, from the same platform family as the in-tree
`comfast_cf-ew72-v2`, which was used as the device-tree template.

### Specifications
- SoC: MediaTek MT7621
- RAM: 128 MB
- Flash: 16 MB (Winbond W25Q128)
- WiFi: MT7603E (2.4 GHz) + MT7663E (5 GHz)
- Ethernet: MT7530 switch, 2× Gigabit (LAN + WAN)
- LEDs: red:wan, green:lan, blue:wlan (GPIO, active high)
- Button: reset
- Power: 802.3af PoE

### MAC addresses (factory partition)
- 2.4 GHz radio @ 0x0004
- 5 GHz radio   @ 0x8004
- LAN / label   @ 0xe000
- WAN           = LAN + 1

### Installation
Via the stock U-Boot HTTP "web failsafe" recovery: hold the reset button while
powering on, set the PC to 192.168.1.2/24, open http://192.168.1.1, and upload
the `*-squashfs-factory.bin` image.

### Testing
Built from current main for ramips/mt7621 and flashed on real hardware:
board correctly identified; both radios initialize with MAC addresses read
from the factory calibration data (no random MACs); correct 2.4/5 GHz band
assignment; LEDs, reset button, and both Ethernet ports functional;
squashfs + overlay install, configuration persists across reboots.

### DIFF
--- a/target/linux/ramips/dts/mt7621_comfast_cf-e455ac-v2.dts
+++ b/target/linux/ramips/dts/mt7621_comfast_cf-e455ac-v2.dts
@@ -1,0 +1,186 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+#include "mt7621.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+
+/ {
+	compatible = "comfast,cf-e455ac-v2", "mediatek,mt7621-soc";
+	model = "COMFAST CF-E455AC V2";
+
+	chosen {
+		bootargs = "console=ttyS0,115200";
+	};
+
+	keys {
+		compatible = "gpio-keys";
+
+		reset {
+			label = "reset";
+			gpios = <&gpio 18 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+			debounce-interval = <60>;
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		/*
+		 * Three GPIO LEDs, active-high (confirmed against the stock
+		 * firmware's live GPIO/LED state):
+		 *   red:wan   gpio 13
+		 *   green:lan gpio 15
+		 *   blue:wlan gpio 16
+		 * gpio 19 is driven high by the bootloader but is not exposed
+		 * as an LED by the vendor firmware; left unconfigured.
+		 */
+
+		led_wan: wan {
+			color = <LED_COLOR_ID_RED>;
+			function = LED_FUNCTION_WAN;
+			gpios = <&gpio 13 GPIO_ACTIVE_HIGH>;
+		};
+
+		led_lan: lan {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_LAN;
+			gpios = <&gpio 15 GPIO_ACTIVE_HIGH>;
+		};
+
+		wlan {
+			color = <LED_COLOR_ID_BLUE>;
+			function = LED_FUNCTION_WLAN;
+			gpios = <&gpio 16 GPIO_ACTIVE_HIGH>;
+		};
+	};
+
+	aliases {
+		label-mac-device = &gmac0;
+	};
+};
+
+&spi0 {
+	status = "okay";
+
+	flash@0 {
+		compatible = "w25q128";
+		reg = <0>;
+		spi-max-frequency = <10000000>;
+		m25p,fast-read;
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "Bootloader";
+				reg = <0x0 0x30000>;
+				read-only;
+			};
+
+			partition@30000 {
+				label = "Config";
+				reg = <0x30000 0x10000>;
+				read-only;
+			};
+
+			partition@40000  {
+				label = "factory";
+				reg = <0x40000 0x10000>;
+				read-only;
+
+				nvmem-layout {
+					compatible = "fixed-layout";
+					#address-cells = <1>;
+					#size-cells = <1>;
+
+					eeprom_factory_0: eeprom@0 {
+						reg = <0x0 0x400>;
+					};
+
+					eeprom_factory_8000: eeprom@8000 {
+						reg = <0x8000 0x4da8>;
+					};
+
+					macaddr_factory_e000: macaddr@e000 {
+						compatible = "mac-base";
+						reg = <0xe000 0x6>;
+						#nvmem-cell-cells = <1>;
+					};
+				};
+			};
+
+			partition@50000  {
+				label = "firmware";
+				compatible = "denx,uimage";
+				reg = <0x50000 0xfb0000>;
+			};
+		};
+	};
+};
+
+&gpio {
+	groups = "i2c", "uart2", "uart3", "sdhci", "jtag";
+	function = "gpio";
+};
+
+&gmac0 {
+	nvmem-cells = <&macaddr_factory_e000 0>;
+	nvmem-cell-names = "mac-address";
+};
+
+&pcie {
+	status = "okay";
+};
+
+&pcie0 {
+	wifi_2_4_ghz: wifi@0,0 {
+		compatible = "mediatek,mt76";
+		reg = <0x0000 0 0 0 0>;
+		nvmem-cells = <&eeprom_factory_0>;
+		nvmem-cell-names = "eeprom";
+		/*
+		 * MT7603 (2.4 GHz). Reads its MAC from EEPROM (&factory + 4);
+		 * do not add mac-address here or a random MAC will be used.
+		 */
+	};
+};
+
+&pcie1 {
+	wifi_5_0_ghz: wifi@0,0 {
+		compatible = "mediatek,mt76";
+		reg = <0x0000 0 0 0 0>;
+		nvmem-cells = <&eeprom_factory_8000>;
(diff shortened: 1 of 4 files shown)


target/linux/ramips/image/mt7621.mk
```diff
@@ -720,6 +720,21 @@ define Device/comfast_cf-e390ax
 endef
 TARGET_DEVICES += comfast_cf-e390ax
 
+define Device/comfast_cf-e455ac-v2
+    $(Device/dsa-migration)
+    $(Device/uimage-lzma-loader)
+    IMAGE_SIZE := 15808k
```
openwrt-ai, line 726: `IMAGE_SIZE` does not match the firmware partition declared in the new DTS: `reg = <0x50000 0xfb0000>` is 16064k, so 15808k leaves 256k of the firmware partition permanently unusable and `check-size` will reject images that would actually fit.

The value looks copied from `comfast_cf-e390ax`, where 15808k is correct because its firmware partition really is `0xf70000` ([mt7621_comfast_cf-e390ax.dts:131-132](https://github.com/openwrt/openwrt/blob/71e430f462bec4a8bb59c2e2308aca2b8cf652a7/target/linux/ramips/dts/mt7621_comfast_cf-e390ax.dts#L131-L132) vs [mt7621.mk:712](https://github.com/openwrt/openwrt/blob/71e430f462bec4a8bb59c2e2308aca2b8cf652a7/target/linux/ramips/image/mt7621.mk#L712)). `comfast_cf-ew72-v2` carries the same 0xfb0000/15808k mismatch, so copying that block propagates it.

```suggestion
    IMAGE_SIZE := 16064k
```

---

target/linux/ramips/image/mt7621.mk
```diff
@@ -720,6 +720,21 @@ define Device/comfast_cf-e390ax
...
+    DEVICE_VENDOR := COMFAST
+    DEVICE_MODEL := CF-E455AC V2
+    DEVICE_PACKAGES := kmod-mt7603 kmod-mt7615e kmod-mt7663-firmware-ap \
+        -uboot-envtools
+    IMAGES += factory.bin
+    IMAGE/sysupgrade.bin := append-kernel | append-rootfs | pad-rootfs | \
+        check-size | append-metadata
+    IMAGE/factory.bin := append-kernel | append-rootfs | pad-rootfs | check-size
```
openwrt-ai, line 734: nit: this block is indented with 4 spaces (and 8-space continuations). Every other `define Device/...` block in `mt7621.mk` uses 2 spaces with a tab for continuation lines — 162 of the 163 existing blocks; the sole exception is `comfast_cf-ew72-v2` directly below, which is what got copied here. Please re-indent to match the file (keeping the `IMAGE_SIZE` fix from the comment above):

```
define Device/comfast_cf-e455ac-v2
  $(Device/dsa-migration)
  $(Device/uimage-lzma-loader)
  IMAGE_SIZE := 16064k
  DEVICE_VENDOR := COMFAST
  DEVICE_MODEL := CF-E455AC V2
  DEVICE_PACKAGES := kmod-mt7603 kmod-mt7615e kmod-mt7663-firmware-ap \
	-uboot-envtools
  IMAGES += factory.bin
  IMAGE/sysupgrade.bin := append-kernel | append-rootfs | pad-rootfs | \
	check-size | append-metadata
  IMAGE/factory.bin := append-kernel | append-rootfs | pad-rootfs | check-size
endef
```

---

target/linux/ramips/dts/mt7621_comfast_cf-e455ac-v2.dts
```diff
@@ -0,0 +1,186 @@
...
+	aliases {
+		label-mac-device = &gmac0;
+	};
+};
+
+&spi0 {
+	status = "okay";
+
+	flash@0 {
+		compatible = "w25q128";
```
openwrt-ai, line 69: Use the generic SPI-NOR compatible instead of the legacy per-chip string. `"w25q128"` only still works through the deprecated back-compat table in `drivers/mtd/spi-nor/core.c` (`spi_nor_dev_ids[]`); in `target/linux/ramips/dts/` only 2 of 331 boards use a bare chip compatible, while 329 use `jedec,spi-nor`.

```suggestion
		compatible = "jedec,spi-nor";
```

While you are here: `m25p,fast-read` on the next line is also an m25p80-era leftover and is redundant with `jedec,spi-nor` (fast read is negotiated from SFDP), so it can be dropped too.

---

target/linux/ramips/dts/mt7621_comfast_cf-e455ac-v2.dts
```diff
@@ -0,0 +1,186 @@
...
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_LAN;
+			gpios = <&gpio 15 GPIO_ACTIVE_HIGH>;
+		};
+
+		wlan {
+			color = <LED_COLOR_ID_BLUE>;
+			function = LED_FUNCTION_WLAN;
+			gpios = <&gpio 16 GPIO_ACTIVE_HIGH>;
+		};
```
openwrt-ai, line 57: Is the active-high polarity really confirmed? Both in-tree COMFAST MT7621 ceiling APs drive the *same* GPIO trio 13/15/16 and both declare them `GPIO_ACTIVE_LOW`:

- [mt7621_comfast_cf-e390ax.dts:30-46](https://github.com/openwrt/openwrt/blob/71e430f462bec4a8bb59c2e2308aca2b8cf652a7/target/linux/ramips/dts/mt7621_comfast_cf-e390ax.dts#L30-L46) — gpio 15/16/13, all `GPIO_ACTIVE_LOW`
- [mt7621_comfast_cf-ew72-v2.dts:62-77](https://github.com/openwrt/openwrt/blob/71e430f462bec4a8bb59c2e2308aca2b8cf652a7/target/linux/ramips/dts/mt7621_comfast_cf-ew72-v2.dts#L62-L77) — gpio 13/16/15, all `GPIO_ACTIVE_LOW`

The comment above says this was confirmed "against the stock firmware's live GPIO/LED state", but observing the raw GPIO level while the vendor firmware runs doesn't by itself distinguish polarity — you also need to know whether the LED was meant to be lit at that moment. Worth double-checking with `echo 1 > brightness` on the booted OpenWrt image (LED should light, not go dark); if it is inverted, the `default-state`/trigger behaviour and the boot indication will be backwards.

---

target/linux/ramips/dts/mt7621_comfast_cf-e455ac-v2.dts
```diff
@@ -0,0 +1,186 @@
...
+	};
+
+	aliases {
+		label-mac-device = &gmac0;
```
openwrt-ai, line 61: nit: no `led-boot` / `led-failsafe` / `led-running` / `led-upgrade` alias is defined, so the board gives no boot, failsafe or sysupgrade indication even though it has three controllable LEDs. 166 of the 297 `mt7621_*.dts` files set these, including the closest sibling [mt7621_comfast_cf-e390ax.dts:48-54](https://github.com/openwrt/openwrt/blob/71e430f462bec4a8bb59c2e2308aca2b8cf652a7/target/linux/ramips/dts/mt7621_comfast_cf-e390ax.dts#L48-L54), which maps the same three GPIOs onto exactly those aliases.

The `led_wan:` and `led_lan:` labels declared above (lines 41 and 47) are currently unused, which suggests these aliases were intended. Was the omission deliberate?

---

target/linux/ramips/dts/mt7621_comfast_cf-e455ac-v2.dts
```diff
@@ -0,0 +1,186 @@
...
+				read-only;
+			};
+
+			partition@30000 {
+				label = "Config";
+				reg = <0x30000 0x10000>;
+				read-only;
+			};
+
+			partition@40000  {
```
openwrt-ai, line 90: nit: double space before the brace (inherited from the cf-ew72-v2 template).

```suggestion
			partition@40000 {
```

Same on line 116 (`partition@50000  {`).

---

openwrt-ai (commented):
## Commit checks

- `71e430f4` "ramips: add support for COMFAST CF-E455AC V2" — this is a new-device commit, so per `.github/llm-review-rules.md` the message needs a hardware specification, flash instructions and a MAC address layout. The MAC layout section is complete and the specification section is mostly there, but two gaps:
  - **Hardware specification** has no UART entry (header location, pinout, baud rate). `bootargs = "console=ttyS0,115200"` is set in the DTS, so at least the baud rate is known.
  - **Installation** is materially thinner than the PR description. The commit says only "hold the reset button during power-on and upload the factory image", whereas the PR body gives the recovery URL `http://192.168.1.1`, the required host address `192.168.1.2/24`, and the image to use (`*-squashfs-factory.bin`). The rule asks for verbatim button/IP/hostname strings so a stranger can flash the device from the commit message alone — please fold those details from the PR body into the commit message, plus a reverting-to-stock note if the same web-failsafe path accepts the vendor image.


---

openwrt-ai (commented):

`git diff 71e430f4..fac6a35a` is empty — the force-push changed only the commit message. The added `UART: 115200 8N1` line and the expanded three-step installation procedure (static IP `192.168.1.2`, `http://192.168.1.1`, `*-squashfs-factory.bin`) resolve both points from the previous commit check, so that section is dropped this round.

The six inline comments from the previous review still apply unchanged, since the tree contents are identical.

---

PR #23916, opened 2026-06-22
qualcommbe: fix pwm period calculation

hurrian:
During testing on the Askey SBE1V1K, it was noticed that only very low PWM frequencies would work, and 100% duty cycles also did not work.

Comparing the proposed upstream pwm-ipq driver to the downstream vendor driver, `ipq_pwm_apply()` fixed pwm_div at its maximum and derived only pre_div from the requested period.

Since the period spans `(pre_div + 1) * (pwm_div + 1)` input clocks, pinning pwm_div near its maximum forces pre_div towards zero for short periods: once pre_div rounds to 0 the shortest representable period is `(pwm_div + 1) / clk_rate` (~2.7 ms, i.e. ~366 Hz, at a 24 MHz clock), and any shorter request is silently stretched to that.
The high duration then truncates to 0, so the output collapses to ~0% duty.

Since 4-wire fans commonly expect a ~25kHz PWM, it was effectively unusable, since every duty cycle programs a ~zero high time.

Search for the (pre_div, pwm_div) pair whose period best approximates the request instead of fixing pwm_div. Starting pre_div at the smallest value that keeps pwm_div within its field and stopping once pre_div exceeds pwm_div bounds the loop and keeps pwm_div as large as possible for fine duty resolution. For a 25 kHz request at 24 MHz this selects pre_div = 0, pwm_div = 959, giving full 0..960 duty resolution.

While reworking the high-duration computation, round it to nearest rather than truncating, so mid-range duty cycles are not biased low, and clamp it to pwm_div + 1. Rounding, or a 100% duty request, could otherwise push hi_dur past the period length and overflow the 16-bit HI_DURATION field.

Also compute hi_div in `get_state()` in 64-bit; `hi_dur * (pre_div + 1)` can exceed 32 bits before the existing promotion.

Fixes: 01fb4a6daadb ("qualcommbe: update pwm patches and add missing symbol")
Signed-off-by: Kenneth Kasilag <kenneth@kasilag.me>

PR #24479, opened 2026-07-29
qualcommax: fix pwm period calculation

stephanelepain-hub:
The GL.iNet GL-AXT1800 (ipq6018) requests a 40,000 ns (25 kHz) PWM period for its four-wire fan:

```dts
pwms = <&pwm 1 40000 0>;
```

with the IPQ6018 PWM node clocked at 100 MHz. `ipq_pwm_apply()` pins `pwm_div` at its maximum and derives only `pre_div`:

```c
pre_div = period_ns * clk_rate / (NSEC_PER_SEC * (pwm_div + 1));
if (!pre_div)
        return -ERANGE;
```

For 40,000 ns at 100 MHz this is `floor(0.061) == 0`, so the driver deterministically returns `-ERANGE` and `pwm-fan` fails to probe on every boot:

```
pwm-fan pwm-fan: failed to enable PWM
pwm-fan pwm-fan: Failed to configure PWM: -34
pwm-fan pwm-fan: probe with driver pwm-fan failed with error -34
```

Probe returns before the tachometer IRQ is requested and before `fan-supply` is claimed, so the board loses fan RPM reporting and the `vcc_fan` 5V regulator stays disabled. The fan never spins, and the DTS `cooling-maps` (trips at 50/75/100 °C) have no cooling device to bind to.

### Relation to existing work

This is the same defect fixed for **qualcommbe** in 8db23dc91a015bf843f1e3fbd0891574594e86f9 (#23916), whose rationale explicitly calls out 25 kHz four-wire fan PWM. qualcommax carries its own copy of the `pwm-ipq` driver and was not covered by that fix.

The patch added here is **that work backported to qualcommax with authorship preserved** (`From:` / `Signed-off-by:` Kenneth Kasilag). The base driver differs slightly between the two targets, so the hunks were rebased onto the qualcommax copy. The driver source is identical in `patches-6.12` and `patches-6.18`, so the patch is added to both.

### Hardware verification

Before, driving the PWM directly from userspace on a GL-AXT1800 running `r35591-d0110a25ed` / kernel 6.12.98:

```
# echo 40000   > period; echo 1 > enable   -> write error (-ERANGE)
# echo 2700000 > period; echo 1 > enable   -> succeeds
```

After building and flashing an image with this patch, `pwm-fan` probes cleanly and the fan is verified spinning by its own tachometer:

```
/sys/class/hwmon/hwmon7/name                          = pwmfan
/sys/devices/platform/pwm-fan/hwmon/hwmon7/fan1_input = 3548
/sys/class/regulator/regulator.3 (vcc_fan)            = enabled, num_users=1
/sys/class/thermal/cooling_device1                    = pwm-fan, max_state=3
```

and no PWM errors remain in dmesg. Idle SoC temperature drops and the fan is now driven by the thermal zones as intended.

Built and boot-tested on GL-AXT1800 (qualcommax/ipq60xx, kernel 6.12.98). The 6.18 patch is verified to apply cleanly but has not been boot-tested.

### DIFF
--- a/target/linux/qualcommax/patches-6.18/0103-pwm-ipq-fix-period-calculation.patch
+++ b/target/linux/qualcommax/patches-6.18/0103-pwm-ipq-fix-period-calculation.patch
@@ -1,0 +1,151 @@
+From 314b696c26589f53cd9c1d1ca35edcfaddd362fb Mon Sep 17 00:00:00 2001
+From: Kenneth Kasilag <kenneth@kasilag.me>
+Date: Sun, 21 Jun 2026 23:26:43 +0000
+Subject: [PATCH] pwm: ipq: fix period calculation
+
+Comparing the proposed upstream pwm-ipq driver to the downstream vendor
+driver, `ipq_pwm_apply()` fixed pwm_div at its maximum and derived only
+pre_div from the requested period. Since the period spans
+`(pre_div + 1) * (pwm_div + 1)` input clocks, pinning pwm_div near its
+maximum forces pre_div towards zero for short periods: once pre_div
+rounds to 0 the shortest representable period is
+`(pwm_div + 1) / clk_rate`, and any shorter request is rejected with
+-ERANGE or silently stretched. The high duration then truncates to 0,
+so the output collapses to ~0% duty.
+
+Since 4-wire fans commonly expect a ~25kHz PWM, it was effectively
+unusable, since every duty cycle programs a ~zero high time.
+
+Search for the (pre_div, pwm_div) pair whose period best approximates
+the request instead of fixing pwm_div. Starting pre_div at the smallest
+value that keeps pwm_div within its field and stopping once pre_div
+exceeds pwm_div bounds the loop and keeps pwm_div as large as possible
+for fine duty resolution.
+
+While reworking the high-duration computation, round it to nearest
+rather than truncating, so mid-range duty cycles are not biased low, and
+clamp it to pwm_div + 1. Rounding, or a 100% duty request, could
+otherwise push hi_dur past the period length and overflow the 16-bit
+HI_DURATION field.
+
+Backported to qualcommax from OpenWrt commit 8db23dc91a015bf843f1e3fbd0891574594e86f9
+("qualcommbe: fix pwm period calculation").
+
+Signed-off-by: Kenneth Kasilag <kenneth@kasilag.me>
+---
+ drivers/pwm/pwm-ipq.c | 89 +++++++++++++++++++++++++++++--------------
+ 1 file changed, 64 insertions(+), 25 deletions(-)
+
+--- a/drivers/pwm/pwm-ipq.c
++++ b/drivers/pwm/pwm-ipq.c
+@@ -89,10 +89,10 @@ static int ipq_pwm_apply(struct pwm_chip
+ 			 const struct pwm_state *state)
+ {
+ 	struct ipq_pwm_chip *ipq_chip = ipq_pwm_from_chip(chip);
+-	unsigned int pre_div, pwm_div;
+-	u64 period_ns, duty_ns;
++	unsigned int pre_div, pwm_div, best_pre_div, best_pwm_div;
++	u64 period_ns, duty_ns, period_rate, min_diff;
+ 	unsigned long val = 0;
+-	unsigned long hi_dur;
++	u64 hi_dur;
+ 
+ 	if (!state->enabled) {
+ 		/* clear IPQ_PWM_REG1_ENABLE */
+@@ -112,35 +112,74 @@ static int ipq_pwm_apply(struct pwm_chip
+ 	period_ns = min(state->period, IPQ_PWM_MAX_PERIOD_NS);
+ 	duty_ns = min(state->duty_cycle, period_ns);
+ 
+-	/*
+-	 * Pick the maximal value for PWM_DIV that still allows a
+-	 * 100% relative duty cycle. This allows a fine grained
+-	 * selection of duty cycles.
+-	 */
+-	pwm_div = IPQ_PWM_MAX_DIV - 1;
++	period_rate = period_ns * ipq_chip->clk_rate;
++
++	best_pre_div = IPQ_PWM_MAX_DIV;
++	best_pwm_div = IPQ_PWM_MAX_DIV;
++	min_diff = period_rate;
+ 
+ 	/*
+-	 * although mul_u64_u64_div_u64 returns a u64, in practice it
+-	 * won't overflow due to above constraints. Take the max period
+-	 * of 10^9 (NSEC_PER_SEC) and the pwm_div + 1 (IPQ_PWM_MAX_DIV)
+-	 *  10^9 * 10^8
+-	 * ------------- => which fits well into a 32-bit unsigned int.
+-	 * 10^9 * 65,535
++	 * Smaller pre_div than this cannot represent the period (pwm_div would
++	 * have to exceed its field), so start the search there.
+ 	 */
+-	pre_div = mul_u64_u64_div_u64(period_ns, ipq_chip->clk_rate,
+-				      (u64)NSEC_PER_SEC * (pwm_div + 1));
++	pre_div = div64_u64(period_rate,
++			    (u64)NSEC_PER_SEC * (IPQ_PWM_MAX_DIV + 1));
+ 
+-	if (!pre_div)
+-		return -ERANGE;
++	for (; pre_div <= IPQ_PWM_MAX_DIV; pre_div++) {
++		u64 remainder;
++
++		pwm_div = div64_u64_rem(period_rate,
++					(u64)NSEC_PER_SEC * (pre_div + 1),
++					&remainder);
++		/* pwm_div is unsigned; the swap check below catches underflow */
++		pwm_div--;
++
++		/*
++		 * Swapping pre_div and pwm_div yields the same period but a
++		 * larger pwm_div gives finer duty resolution, so once pre_div
++		 * exceeds pwm_div every further candidate is strictly worse.
++		 */
++		if (pre_div > pwm_div)
++			break;
++
++		/* need room for 100% duty, where hi_dur == pwm_div + 1 */
++		if (pwm_div > IPQ_PWM_MAX_DIV - 1)
++			continue;
++
++		if (remainder < min_diff) {
++			best_pre_div = pre_div;
++			best_pwm_div = pwm_div;
++			min_diff = remainder;
++
++			if (min_diff == 0)
++				break;
++		}
++	}
+ 
+-	pre_div -= 1;
++	pre_div = best_pre_div;
++	pwm_div = best_pwm_div;
+ 
+-	if (pre_div > IPQ_PWM_MAX_DIV)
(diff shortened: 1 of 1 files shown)


target/linux/qualcommax/patches-6.12/0103-pwm-ipq-fix-period-calculation.patch
openwrt-ai, line 1: The `0080-` prefix is already in use in **both** patch directories by [`0080-v7.1-dt-bindings-clock-qcom-Add-CMN-PLL-support-for-IPQ6018.patch`](``https://github.com/openwrt/openwrt/blob/114dda097a34838f7b9536082095abbd9446e23e/target/linux/qualcommax/patches-6.12/0080-v7.1-dt-bindings-clock-qcom-Add-CMN-PLL-support-for-IPQ6018.patch)`` (same file exists in `patches-6.18`), so after this PR each directory contains two `0080-*` patches.

Additionally, in this target the entire `0036`–`0090` block is upstream/queued material carrying a `-vX.Y-` tag; the first untagged, downstream-only patch is `0102`. A downstream-only fix like this one fits that range better.

Please renumber both copies to a slot that is free in `patches-6.12` and `patches-6.18` — e.g. `0103-pwm-ipq-fix-period-calculation.patch` (unused in both).

---

target/linux/qualcommax/patches-6.12/0103-pwm-ipq-fix-period-calculation.patch
```diff
@@ -0,0 +1,172 @@
...
+
+Signed-off-by: Kenneth Kasilag <kenneth@kasilag.me>
+---
+ drivers/pwm/pwm-ipq.c | 107 ++++++++++++++++++++++++++++++++++--------
+ 1 file changed, 87 insertions(+), 20 deletions(-)
+
+--- a/drivers/pwm/pwm-ipq.c
++++ b/drivers/pwm/pwm-ipq.c
+@@ -89,10 +89,10 @@ static int ipq_pwm_apply(struct pwm_chip
+ 			 const struct pwm_state *state)
```
openwrt-ai, line 42: This file was not produced by the project's quilt refresh, and its metadata differs from every other patch in the tree in two ways:

- it carries a `diff --git a/... b/...` line — none of the 118 patches in `target/linux/qualcommax/patches-6.12/` (nor any of the 165 in `target/linux/generic/pending-6.12/`) has one;
- the hunk headers have no enclosing-function context (`@@ -89,10 +89,10 @@`), whereas refreshed patches carry it, e.g. `@@ -89,10 +89,10 @@ static int ipq_pwm_apply(struct pwm_chip` in the qualcommbe copy this was ported from, and `@@ -319,6 +319,18 @@ config PWM_INTEL_LGM` in the [in-tree pwm driver patch](``https://github.com/openwrt/openwrt/blob/114dda097a34838f7b9536082095abbd9446e23e/target/linux/qualcommax/patches-6.12/0077-v7.2-pwm-Driver-for-qualcomm-ipq6018-pwm-block.patch#L32).``

Both come from the refresh invocation at [include/quilt.mk:173](https://github.com/openwrt/openwrt/blob/114dda097a34838f7b9536082095abbd9446e23e/include/quilt.mk#L173) (`QUILT_DIFF_OPTS="-p" quilt refresh -p ab --no-index --no-timestamps`). Please regenerate both copies with `make target/linux/refresh` (qualcommax selected, once with `KERNEL_PATCHVER` 6.12 and once with the 6.18 testing kernel) so the patch metadata matches the tree and stays stable on the next tree-wide refresh.

---

target/linux/qualcommax/patches-6.12/0080-pwm-ipq-fix-period-calculation.patch
```diff
@@ -0,0 +1,172 @@
...
+Also compute hi_div in `get_state()` in 64-bit; `hi_dur * (pre_div + 1)`
+can exceed 32 bits before the existing promotion.
+
+Backported to qualcommax from OpenWrt commit 8db23dc91a015bf843f1e3fbd0891574594e86f9
+("qualcommbe: fix pwm period calculation").
+
+Signed-off-by: Kenneth Kasilag <kenneth@kasilag.me>
+---
+ drivers/pwm/pwm-ipq.c | 107 ++++++++++++++++++++++++++++++++++--------
+ 1 file changed, 87 insertions(+), 20 deletions(-)
```
openwrt-ai: The diffstat is stale — it was carried over from the qualcommbe patch and does not describe this diff. Counting the hunks below gives 75 insertions and 24 deletions (99 changed lines), not 87/20 (107).

```suggestion
 1 file changed, 75 insertions(+), 24 deletions(-)
```

The graph line above needs the same correction (`107 ++++...` → 99 changed lines), or just drop both diffstat lines. Same in the `patches-6.18` copy, which is byte-identical.

---

target/linux/qualcommax/patches-6.12/0080-pwm-ipq-fix-period-calculation.patch
```diff
@@ -0,0 +1,172 @@
+From 314b696c26589f53cd9c1d1ca35edcfaddd362fb Mon Sep 17 00:00:00 2001
+From: Kenneth Kasilag <kenneth@kasilag.me>
+Date: Sun, 21 Jun 2026 23:26:43 +0000
+Subject: [PATCH] qualcommax: fix pwm period calculation
```
openwrt-ai: nit: the subject inside the patch uses the OpenWrt-side prefix, which disagrees with the filename you chose (`0080-pwm-ipq-fix-period-calculation.patch`). This hunk goes into the kernel tree, so the subject should describe the kernel change; the `qualcommax:` prefix already lives on the OpenWrt commit.

```suggestion
Subject: [PATCH] pwm: ipq: fix period calculation
```

Same in the `patches-6.18` copy.

---

target/linux/qualcommax/patches-6.12/0080-pwm-ipq-fix-period-calculation.patch
```diff
@@ -0,0 +1,172 @@
...
++		pwm_div = IPQ_PWM_MAX_DIV - 1;
++
++	/*
++	 * high duration = duty_ratio * (pwm_div + 1)
++	 *              = duty_ns * clk_rate / ((pre_div + 1) * NSEC_PER_SEC)
++	 *
++	 * Round to nearest to avoid biasing every duty cycle low, then clamp
++	 * to (pwm_div + 1): rounding or a 100% duty request can otherwise push
++	 * hi_dur past the period length and overflow the 16-bit HI_DURATION field
++	 * (which would alias a full-on request down to a near-zero high time)
```
openwrt-ai: nit: this comment is the only code text that diverges from the qualcommbe original ([0403-pwm-fix-period-calculation.patch:142-143](``https://github.com/openwrt/openwrt/blob/114dda097a34838f7b9536082095abbd9446e23e/target/linux/qualcommbe/patches-6.18/0403-pwm-fix-period-calculation.patch#L142-L143)),`` and the rewrite breaks the sentence: it now reads "can otherwise push hi_dur past the period length **and overflow** the ... field ... **and asking** the hardware to stay high beyond one period", losing the `overflowing ... and asking` parallel. Restoring the original wording also keeps the two target copies byte-identical in the shared region, so future syncs between them diff cleanly.

```suggestion
+	 * to (pwm_div + 1): rounding or a 100% request can otherwise push
+	 * hi_dur past the period, overflowing the 16-bit HI_DURATION field
```

Same in the `patches-6.18` copy.

---

openwrt-ai (commented):
Reviewed 1 new commit. The fix itself checks out: I extracted `drivers/pwm/pwm-ipq.c` from the in-tree `0077-v7.2-pwm-Driver-for-qualcomm-ipq6018-pwm-block.patch` and both copies of the new patch apply cleanly on top (the pwm-ipq.c part of `0077` is identical in `patches-6.12` and `patches-6.18`, as the commit message states), the reworked `ipq_pwm_apply()` region ends up matching the qualcommbe patched driver, and for the GL-AXT1800 case (40,000 ns @ 100 MHz) the search selects `pre_div = 0`, `pwm_div = 3999`, i.e. exactly 4000 clocks = 25 kHz with full duty resolution. `GIGA`, `div64_u64_rem()` and `DIV64_U64_ROUND_CLOSEST()` all exist in 6.12.98 and 6.18.39, and `linux/units.h` / `linux/math64.h` are already included by the driver.

The inline comments are all about patch numbering and patch-file metadata, not the logic.

---

openwrt-ai (commented):

The force-push addresses the patch-metadata finding from the previous review: I reconstructed `drivers/pwm/pwm-ipq.c` from `0077-v7.2-pwm-Driver-for-qualcomm-ipq6018-pwm-block.patch`, applied both copies of this patch (clean, no fuzz/offset), and re-diffed with the options `include/quilt.mk` uses (`diff -up`, `a/`/`b/` prefixes, no index line). The result is byte-identical to the hunks in both patch files now, and the two copies remain identical to each other. Nothing changed below the `--- a/drivers/pwm/pwm-ipq.c` line, so the logic review from the previous pass still stands.

The four other open threads from the previous review are unaffected by this commit and still apply.

---

georgemoussalem:
When tested, can you send this patch upstream?

robimarko:
Thanks for this, can you drop the 6.12 patch as we dropped the 6.12 support

stephanelepain-hub:
Thanks @robimarko — 6.12 patch dropped and the branch rebased onto current main, so this is now 6.18 only.

While here I also folded in Konrad Dybcio's review feedback from the upstream posting, so the patch carried here matches what is on the list:

- dropped the explanatory comment above the search (it restated what the loop does; the rationale lives in the commit message)
- dropped the `clk_rate > 16 * GIGA` bound — `period_ns` is already clamped to `IPQ_PWM_MAX_PERIOD_NS` (1e9), so `period_ns * clk_rate` only overflows u64 above ~18.4 GHz, i.e. ~184x the 100 MHz this block actually runs at. It was unreachable
- dropped the `(u64)hi_div` cast as an unrelated change. It is not a live bug either: `hi_dur` and `pre_div` both come from 16-bit fields, so the worst case is 65535 * 65536 = 4294901760, which still fits u32 (`effective_div` on the line above genuinely does need the cast, at 65536 * 65536)

Behaviour is unchanged — for the AXT1800's 40,000 ns @ 100 MHz the search still selects `pre_div = 0`, `pwm_div = 3999`, exactly 4000 clocks = 25 kHz with full 0..4000 duty resolution.

Upstream: [PATCH v2] is on linux-pwm with Konrad, Uwe, Kenneth, George, Devi Priya and Baruch on CC. Kenneth Kasilag remains the author; the changes above are recorded in a `[Stephane: ...]` note under the `---`.

One caveat worth stating plainly: the hardware testing (fan spinning at 3548 RPM on its tachometer, `vcc_fan` enabled, idle SoC ~76C -> ~51C) was done on a 6.12 build, since that is what the target shipped at the time. The `pwm-ipq.c` source is identical under 6.12 and 6.18, so the patch and its effect are unchanged, but I have not booted a 6.18 image on this board. Happy to do that before merge if you would like it confirmed.

robimarko:

PR #22769, opened 2026-04-04
ath79: wndap360: change to lzma-loader

neheb:
For running, have it be the green LED as in stock firmware.

PR #24480, opened 2026-07-29
[24.10] ath79: wndap360: make it usable

neheb:
Because of bootloader limitations, squashfs cannot be used directly and lzma-loader must be used.

Currently, gzip is used and creates a broken image with not enough space to create overlayfs at runtime:

https://downloads.openwrt.org/releases/24.10.8/targets/ath79/generic/openwrt-24.10.8-ath79-generic-netgear_wndap360-squashfs-sysupgrade.bin

Backport everything from master to have a working image even on 24.10.

### DIFF
--- a/target/linux/ath79/dts/ar7161_netgear_wndap360.dts
+++ b/target/linux/ath79/dts/ar7161_netgear_wndap360.dts
@@ -11,7 +11,7 @@
 	model = "Netgear WNDAP360";
 
 	chosen {
-		/delete-property/ bootargs;
+		bootargs = "console=ttyS0,9600";
 	};
 
 	aliases {
--- a/target/linux/ath79/dts/ar7161_netgear_wndap360.dts
+++ b/target/linux/ath79/dts/ar7161_netgear_wndap360.dts
@@ -10,6 +10,10 @@
 	compatible = "netgear,wndap360", "qca,ar7161";
 	model = "Netgear WNDAP360";
 
+	chosen {
+		/delete-property/ bootargs;
+	};
+
 	aliases {
 		led-boot = &led_power_orange;
 		led-failsafe = &led_power_orange;
--- a/target/linux/ath79/dts/ar7161_netgear_wndap360.dts
+++ b/target/linux/ath79/dts/ar7161_netgear_wndap360.dts
@@ -17,12 +17,18 @@
 	aliases {
 		led-boot = &led_power_orange;
 		led-failsafe = &led_power_orange;
-		led-running = &led_power_orange;
+		led-running = &led_power_green;
 		led-upgrade = &led_power_orange;
 	};
 
 	leds {
 		compatible = "gpio-leds";
+
+		led_power_green: power_green {
+			function = LED_FUNCTION_POWER;
+			color = <LED_COLOR_ID_GREEN>;
+			gpios = <&gpio 2 GPIO_ACTIVE_LOW>;
+		};
 
 		led_power_orange: power_orange {
 			function = LED_FUNCTION_POWER;
--- a/target/linux/ath79/dts/ar7161_netgear_wndap360.dts
+++ b/target/linux/ath79/dts/ar7161_netgear_wndap360.dts
@@ -19,6 +19,7 @@
 		led-failsafe = &led_power_orange;
 		led-running = &led_power_green;
 		led-upgrade = &led_power_orange;
+		label-mac-device = &ath9k0;
 	};
 
 	leds {
@@ -80,7 +81,7 @@
 	phy-mode = "rgmii";
 	phy-handle = <&phy1>;
 
-	nvmem-cells = <&macaddr_art_0>;
+	nvmem-cells = <&macaddr_art_120c 10>;
 	nvmem-cell-names = "mac-address";
 };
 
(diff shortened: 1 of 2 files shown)


target/linux/ath79/image/generic.mk
```diff
@@ -2211,13 +2211,14 @@ define Device/netgear_wndap360
   $(Device/netgear_generic)
   SOC := ar7161
   DEVICE_MODEL := WNDAP360
+  DEVICE_PACKAGES := kmod-owl-loader
```
openwrt-ai, line 2214: nit: one commit from the same upstream series is missing here. Upstream PR #22769 landed four commits, three of which are in this backport (`018d6091`, `60383c5d`, `75ebc8e2`, `d49f62d1`) — but [`66fe390b` "uboot-env: ath79: add wndap360"](https://github.com/openwrt/openwrt/commit/66fe390b731b4d2ce4a38189267d589900d132cd) is not. Given the PR body says "Backport everything from master", was that intentional?

On 24.10 the equivalent file is `package/boot/uboot-envtools/files/ath79` (the `uboot-envtools` → `uboot-tools` migration in `46e376c9` is main-only), and `netgear,wndap360` is absent from the `ubootenv_add_uci_config "/dev/mtd1" "0x0" "0x10000" "0x10000"` list there — it would slot in between [`netgear,ex7300-v2` and `netgear,wndr4300-v2` at ath79:48-49](https://github.com/openwrt/openwrt/blob/42fdc7f40155a61d0dd679e0313a8fcc48a3b998/package/boot/uboot-envtools/files/ath79#L48-L49). The DTS `u-boot-env` partition at `0x40000`/`0x10000` is mtd1 as that entry expects, so the upstream hunk applies unchanged apart from the path.


---

target/linux/ath79/dts/ar7161_netgear_wndap360.dts
```diff
@@ -10,16 +10,27 @@
...
+		bootargs = "console=ttyS0,9600";
+	};
+
 	aliases {
 		led-boot = &led_power_orange;
 		led-failsafe = &led_power_orange;
-		led-running = &led_power_orange;
+		led-running = &led_power_green;
 		led-upgrade = &led_power_orange;
+		label-mac-device = &ath9k0;
```
openwrt-ai, line 22: nit: this is the one place where the backport deviates from upstream — [`1ac7a562` uses `label-mac-device = &ath9k;`](https://github.com/openwrt/openwrt/commit/1ac7a562efe9aca9973d188a69d63306c3de081c), because on main the same commit renames the `wifi@11,0` label from `ath9k0` to `ath9k`. That rename is a consequence of [`2f30e14c` "ath79: convert pcie gpios to led-sources"](https://github.com/openwrt/openwrt/commit/2f30e14c494c2d40e4e8c248fa8bd81e67a28478), which is not backported, so `&ath9k0` is the right label for 24.10 — the `ath9k-leds` node still references `&ath9k0`/`&ath9k1` as gpio-controllers here.

The adaptation looks correct; it's just undocumented. Consider adding a short note under the cherry-pick trailer (e.g. "adapted: keep `&ath9k0` label, 2f30e14c not backported") so the divergence from main isn't mistaken for a conflict-resolution slip later.


---

openwrt-ai (commented):
Reviewed 8 new commits (backport to `openwrt-24.10`).

All eight carry a `(cherry picked from commit <sha>)` trailer, every referenced SHA resolves to a real commit on `main`, and each diff matches its upstream original — the only differences are hunk offsets and blob indices, plus one deliberate label adaptation noted inline. The resulting `Device/netgear_wndap360` block is byte-identical to main's, and the resulting DTS matches main's apart from the pre-existing `ath9k-leds` / `led-sources` divergence that `2f30e14c` (not backported) introduced.

Prerequisites checked on `openwrt-24.10` and all present: `Build/loader-kernel` and `LOADER_TYPE := bin` exist in `target/linux/ath79/image/Makefile`; `kmod-owl-loader` exists in `package/kernel/mac80211/ath.mk`; `compatible = "mac-base"` with `#nvmem-cell-cells` was already used by this DTS, so the `<&macaddr_art_120c N>` phandle-arg form works; `ar7100.dtsi` still sets `bootargs = "console=ttyS0,115200"`, so the `/delete-property/` in `7e5d211703` has something to delete; and `netgear,wndap360` is already in `board.d/02_network`.

One point worth confirming on `BLOCKSIZE`: `a1393c514f` drops `BLOCKSIZE := 256k` and relies on the ath79 `Device/Default`, which is `64k` for the generic subtarget on both branches (main only adds a `tiny` → `4k` special case, irrelevant here). So `pad-to $$$$(BLOCKSIZE)` resolves to the same `64k` the old recipe hardcoded — behaviour matches main.

Two inline nits below; neither blocks.


---

target/linux/ath79/dts/ar7161_netgear_wndap360.dts
neheb: No

target/linux/ath79/image/generic.mk
neheb: It's not really important for usability.

neheb:
ping @hauke

neheb:
ping @jonasjelonek

jonasjelonek:
Could you add a final SoB to every commit here? If I merge this right now, in history it reads like I backported/cherry-picked them but this is obviously not correct.

neheb:
Hmm? git cherry-pick -x is standard AFAIK

jonasjelonek:
Ideally it would look like this: https://github.com/openwrt/openwrt/commit/6a65e9e47967e88b33522b2615dfdf59f2dca8a0

```
<original stuff>
(cherry-picked from ...)
<your SoB>
<my SoB added by rebase/merge>
```

Not sure what `git cherry-pick -x` does by default.

neheb:
That would look like nonsense. I would have to place my Signed-off-by line twice.

jonasjelonek:
No because now it'll look like I cherry-picked the commits, not you.

Anyway ... I don't have the patience to discuss when someone doesn't get my point.

PR #24481, opened 2026-07-29
ath79: nand: add support for Motorola NCAP-500

ByteFloor:
Add support for the Motorola Solutions NCAP-500 access point based on Atheros AR7242 SoC.

Hardware:

* SoC: Atheros AR7242
* RAM: 64 MB (46V32M16)
* Flash:
  * 8 MB SPI NOR (Macronix MX25L6445EMI-10G)
  * 128 MB SPI NAND (Micron MT29F1G01AAADDH4)
* Ethernet: 1x RJ-45 10/100/1000 Mbps (Broadcom BCM54610C1KMLG)
* Wi-Fi: 1x Mini-PCIe slot (2.4/5 GHz 2x2 MIMO Atheros AR9280 802.11a/b/g/n card)
* Antennas: 2x external antenna connectors (RP-SMA)
* Console: 1x RJ-45 RS-232 (19200 8N1, SIPEX 3232EC)
* Power: 802.3af PoE-in via ethernet or 48 V DC via 2 pin connector
* LEDs: 1x green, 1x orange
* Buttons: 1x reset

U-Boot ENVs:

	bootdelay=2
	baudrate=19200
	mem=57M
	mfg_hardware_type=ncap
	ethact=eth0
	ethaddr=REDACTED
	mfg_serial_number=REDACTED
	base_mint_id=0x1A
	mfg_hardware_model=AP-0621-60020-EU
	mfg_hardware_version=A
	mfg_date=2014/05/17
	mfg_country_code=EU
	filesize=1D5A400
	fileaddr=80100000
	gatewayip=192.168.1.1
	netmask=255.255.255.0
	ipaddr=192.168.1.50
	serverip=192.168.1.100
	bootcmd=bootx
	stdin=serial
	stdout=serial
	stderr=serial

MAC address:

	There is one ethernet MAC address stored in the u-boot-env partition.
	The WLAN MAC address is stored on the WLAN card.

Installation:
Console access is necessary.

	Interrupt U-Boot by pressing key when countdown shows up.
	Transfer the initramfs image via TFTP and boot it:
	tftpboot 0x82000000 openwrt-ath79-nand-motorola_ncap-500-initramfs-kernel.bin
	bootm 0x82000000

	Make a backup of the original firmware(run from PC, nanddump requires mtd-utils):
	ssh root@192.168.1.1 "cat /dev/mtd2" > motorola_original_nor_kernel_dump.bin
	ssh root@192.168.1.1 "nanddump -n -o /dev/mtd4" > motorola_original_nand_dump.bin

	Sysupgrade from initramfs doesn't work, because of limited RAM. Flash manually.

	From PC:
	extract kernel file from openwrt-ath79-nand-motorola_ncap-500-squashfs-sysupgrade.bin
	scp -O kernel root@192.168.1.1:/tmp/kernel

	From console:
	mtd erase kernel
	mtd write /tmp/kernel kernel
	rm /tmp/kernel

	From PC:
	scp -O openwrt-ath79-nand-motorola_ncap-500-squashfs-factory.ubi root@192.168.1.1:/tmp/factory.ubi

	From console:
	ubidetach -m 4
	ubiformat /dev/mtd4 -y -f /tmp/factory.ubi

	reboot

Return to stock firmware:
Use the backup of original firmware made before flashing OpenWrt

Tested on 2 devices:

	Gigabit ethernet
	2.4 and 5 GHz AP mode
	Mini-PCIe slot (tested with original card only)
	LEDs
	UBIFS overlay and persistence after reboot
	Reset button
	Serial Console
	LuCI and sysupgrade through LuCI

### DIFF
--- a/target/linux/ath79/dts/ar7242_motorola_ncap-500.dts
+++ b/target/linux/ath79/dts/ar7242_motorola_ncap-500.dts
@@ -1,0 +1,142 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+#include "ar7242.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+
+/ {
+	compatible = "motorola,ncap-500", "qca,ar7242";
+	model = "Motorola Solutions NCAP-500";
+
+	aliases {
+		led-boot = &led_green;
+		led-failsafe = &led_orange;
+		led-running = &led_green;
+		led-upgrade = &led_orange;
+		label-mac-device = &eth0;
+	};
+
+	chosen {
+		bootargs = "console=ttyS0,19200n8";
+	};
+
+	keys {
+		compatible = "gpio-keys";
+
+		reset {
+			label = "reset";
+			linux,code = <KEY_RESTART>;
+			gpios = <&gpio 1 GPIO_ACTIVE_LOW>;
+			debounce-interval = <60>;
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		led_green: green {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_STATUS;
+			gpios = <&gpio 17 GPIO_ACTIVE_HIGH>;
+		};
+
+		led_orange: orange {
+			color = <LED_COLOR_ID_ORANGE>;
+			function = LED_FUNCTION_FAULT;
+			gpios = <&gpio 16 GPIO_ACTIVE_HIGH>;
+		};
+	};
+};
+
+&pcie {
+	status = "okay";
+
+	compatible = "motorola,ncap-500-pci", "qcom,ar7240-pci";
+};
+
+&mdio0 {
+	status = "okay";
+
+	phy24: ethernet-phy@18 {
+		reg = <0x18>;
+	};
+};
+
+&eth0 {
+	status = "okay";
+
+	phy-mode = "rgmii-id";
+	phy-handle = <&phy24>;
+	pll-data = <0x06000000 0x00000101 0x00001616>;
+	nvmem-cells = <&macaddr_uboot_ethaddr>;
+	nvmem-cell-names = "mac-address";
+};
+
+&eth1 {
+	status = "disabled";
+};
+
+&spi {
+	status = "okay";
+
+	flash@0 {
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <25000000>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "u-boot";
+				reg = <0x000000 0x0d0000>;
+				read-only;
+			};
+
+			partition@d0000 {
+				label = "u-boot-env";
+				reg = <0x0d0000 0x010000>;
+				read-only;
+
+				nvmem-layout {
+					compatible = "u-boot,env";
+
+					macaddr_uboot_ethaddr: ethaddr {
+					};
+				};
+			};
+
+			partition@e0000 {
+				label = "kernel";
+				reg = <0x0e0000 0x680000>;
+			};
+
+			partition@760000 {
+				label = "vendor-data";
+				reg = <0x760000 0x0a0000>;
+				read-only;
+			};
+		};
+	};
+
+	nand@1 {
+		compatible = "spi-nand";
+		reg = <1>;
+		spi-max-frequency = <25000000>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "ubi";
+				reg = <0x0 0x8000000>;
+			};
+		};
+	};
+};
--- a/target/linux/ath79/image/nand.mk
+++ b/target/linux/ath79/image/nand.mk
@@ -348,6 +348,22 @@ define Device/meraki_z1
 endef
 TARGET_DEVICES += meraki_z1
 
+define Device/motorola_ncap-500
+  SOC := ar7242
+  DEVICE_VENDOR := Motorola Solutions
+  DEVICE_MODEL := NCAP-500
+  DEVICE_ALT0_VENDOR := Motorola Solutions
+  DEVICE_ALT0_MODEL := AP-0621
+  KERNEL_SIZE := 6656k
+  BLOCKSIZE := 128k
+  PAGESIZE := 2048
+  VID_HDR_OFFSET := 2048
+  IMAGES := sysupgrade.bin factory.ubi
+  IMAGE/sysupgrade.bin := sysupgrade-tar | append-metadata
+  IMAGE/factory.ubi := append-ubi
+endef
+TARGET_DEVICES += motorola_ncap-500
+
 # fake rootfs is mandatory, pad-offset 129 equals (2 * uimage_header + '\0')
 define Device/netgear_ath79_nand
   DEVICE_VENDOR := NETGEAR
--- a/target/linux/generic/pending-6.18/418-mtd-spinand-micron-add-MT29F1G01AAADD.patch
+++ b/target/linux/generic/pending-6.18/418-mtd-spinand-micron-add-MT29F1G01AAADD.patch
@@ -1,0 +1,58 @@
+From 72789601b55d103a28b6e8e259f3db1d31af0289 Mon Sep 17 00:00:00 2001
+From: Vít Kazík <catelixor@proton.me>
+Date: Thu, 30 Jul 2026 02:26:50 +0200
+Subject: [PATCH] kernel: mtd: spinand: add support for Micron MT29F1G01AAADD
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+Add support for the Micron MT29F1G01AAADD (M68A) SPI NAND chip.
+
+Signed-off-by: Vít Kazík <catelixor@proton.me>
+---
+ drivers/mtd/nand/spi/micron.c | 27 +++++++++++++++++++++++++++
+ 1 file changed, 27 insertions(+)
+
+--- a/drivers/mtd/nand/spi/micron.c
++++ b/drivers/mtd/nand/spi/micron.c
+@@ -134,6 +134,23 @@ static const struct mtd_ooblayout_ops mi
+ 	.free = micron_4_ooblayout_free,
+ };
+ 
++static int micron_m68a_ooblayout_free(struct mtd_info *mtd, int section,
++				      struct mtd_oob_region *region)
++{
++	if (section >= 4)
++		return -ERANGE;
++
++	region->offset = (section * 16) + 2;
++	region->length = 6;
++
++	return 0;
++}
++
++static const struct mtd_ooblayout_ops micron_m68a_ooblayout = {
++	.ecc = micron_4_ooblayout_ecc,
++	.free = micron_m68a_ooblayout_free,
++};
++
+ static int micron_select_target(struct spinand_device *spinand,
+ 				unsigned int target)
+ {
+@@ -404,6 +421,16 @@ static const struct spinand_info micron_
+ 		     SPINAND_ECCINFO(&micron_8_ooblayout,
+ 				     micron_8_ecc_get_status),
+ 		     SPINAND_SELECT_TARGET(micron_select_target)),
++	/* M68A 1Gb 3.3V */
++	SPINAND_INFO("MT29F1G01AAADD",
++		     SPINAND_ID(SPINAND_READID_METHOD_OPCODE_DUMMY, 0x12),
++		     NAND_MEMORG(1, 2048, 64, 64, 1024, 20, 2, 1, 1),
++		     NAND_ECCREQ(4, 512),
++		     SPINAND_INFO_OP_VARIANTS(&x4_read_cache_variants,
++					      &x1_write_cache_variants,
++					      &x1_update_cache_variants),
++		     0,
++		     SPINAND_ECCINFO(&micron_m68a_ooblayout, NULL)),
+ 	/* M69A 2Gb 3.3V */
+ 	SPINAND_INFO("MT29F2G01AAAED",
+ 		     SPINAND_ID(SPINAND_READID_METHOD_OPCODE_DUMMY, 0x9F),
--- a/target/linux/ath79/nand/base-files/etc/board.d/02_network
+++ b/target/linux/ath79/nand/base-files/etc/board.d/02_network
@@ -8,7 +8,8 @@ ath79_setup_interfaces()
 
 	case "$board" in
 	aerohive,hiveap-121|\
-	meraki,mr18)
+	meraki,mr18|\
+	motorola,ncap-500)
 		ucidef_set_interface_lan "eth0"
 		;;
 	domywifi,dw33d)
--- a/target/linux/ath79/patches-6.18/319-MIPS-pci-ar724x-add-motorola-ncap-500-quirk.patch
+++ b/target/linux/ath79/patches-6.18/319-MIPS-pci-ar724x-add-motorola-ncap-500-quirk.patch
@@ -1,0 +1,81 @@
+From 2cebbeca77ae9e30529cdbaebe937a8bfb35ee0e Mon Sep 17 00:00:00 2001
+From: Vít Kazík <catelixor@proton.me>
+Date: Thu, 30 Jul 2026 02:29:28 +0200
+Subject: [PATCH] MIPS: pci-ar724x: add Motorola NCAP-500 quirk
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+Without this the card never comes up.
+
+Add a PCIe endpoint reset sequence and apply application
+control settings required to initialize the PCIe bus on the
+Atheros AR7242 SoC for the Motorola NCAP-500 board.
+
+Signed-off-by: Vít Kazík <catelixor@proton.me>
+---
+ arch/mips/pci/pci-ar724x.c | 39 ++++++++++++++++++++++++++++----------
+ 1 file changed, 29 insertions(+), 10 deletions(-)
+
+--- a/arch/mips/pci/pci-ar724x.c
++++ b/arch/mips/pci/pci-ar724x.c
+@@ -17,6 +17,7 @@
+ #include <asm/mach-ath79/ar71xx_regs.h>
+ #include <linux/of_irq.h>
+ #include <linux/of_pci.h>
++#include <linux/of.h>
+ 
+ #define AR724X_PCI_REG_APP		0x00
+ #define AR724X_PCI_REG_RESET		0x18
+@@ -371,15 +372,32 @@ static void ar724x_pci_hw_init(struct ar
+ 		ath79_pll_wr(AR724X_PLL_REG_PCIE_CONFIG, ppl);
+ 	}
+ 
+-	/* deassert the reset state of the PCIE endpoint */
+-	rst = __raw_readl(apc->ctrl_base + AR724X_PCI_REG_RESET);
+-	rst |= AR724X_PCI_RESET_EP_RESET_L;
+-	__raw_writel(rst, apc->ctrl_base + AR724X_PCI_REG_RESET);
+-
+-	/* set PCIE Application Control to ready */
+-	app = __raw_readl(apc->ctrl_base + AR724X_PCI_REG_APP);
+-	app |= AR724X_PCI_APP_LTSSM_ENABLE;
+-	__raw_writel(app, apc->ctrl_base + AR724X_PCI_REG_APP);
++	if (of_device_is_compatible(apc->np, "motorola,ncap-500-pci")) {
++		/* Motorola NCAP-500: PCIe endpoint reset sequence */
++		rst = __raw_readl(apc->ctrl_base + AR724X_PCI_REG_RESET);
++		rst &= ~AR724X_PCI_RESET_EP_RESET_L;
++		__raw_writel(rst, apc->ctrl_base + AR724X_PCI_REG_RESET);
++		msleep(100);
++		rst |= AR724X_PCI_RESET_EP_RESET_L;
++		__raw_writel(rst, apc->ctrl_base + AR724X_PCI_REG_RESET);
++		msleep(100);
++
++		/*
++		 * Motorola NCAP-500: apply PCIe application control settings
++		 * 0x1ffc1 is a value from the U-Boot.
++		 */
++		__raw_writel(0x1ffc1, apc->ctrl_base + AR724X_PCI_REG_APP);
++	} else {
++		/* deassert the reset state of the PCIE endpoint */
++		rst = __raw_readl(apc->ctrl_base + AR724X_PCI_REG_RESET);
++		rst |= AR724X_PCI_RESET_EP_RESET_L;
++		__raw_writel(rst, apc->ctrl_base + AR724X_PCI_REG_RESET);
++
++		/* set PCIE Application Control to ready */
++		app = __raw_readl(apc->ctrl_base + AR724X_PCI_REG_APP);
++		app |= AR724X_PCI_APP_LTSSM_ENABLE;
++		__raw_writel(app, apc->ctrl_base + AR724X_PCI_REG_APP);
++	}
+ 
+ 	/* wait up to 100ms for PHY link up */
+ 	do {
+@@ -436,7 +454,8 @@ static int ar724x_pci_probe(struct platf
+ 	 * Do the full PCIE Root Complex Initialization Sequence if the PCIe
+ 	 * host controller is in reset.
+ 	 */
+-	if (reset_control_status(apc->hc_reset))
++	if (reset_control_status(apc->hc_reset) ||
++	    of_device_is_compatible(apc->np, "motorola,ncap-500-pci"))
+ 		ar724x_pci_hw_init(apc);
+ 
+ 	apc->link_up = ar724x_pci_check_link(apc);


target/linux/generic/pending-6.18/418-mtd-spinand-micron-add-MT29F1G01AAADD.patch
```diff
@@ -0,0 +1,75 @@
...
++	/* M68A 1Gb 3.3V */
++	SPINAND_INFO("MT29F1G01AAADD",
++		     SPINAND_ID(SPINAND_READID_METHOD_OPCODE_DUMMY, 0x12),
++		     NAND_MEMORG(1, 2048, 64, 64, 1024, 20, 2, 1, 1),
++		     NAND_ECCREQ(4, 512),
++		     SPINAND_INFO_OP_VARIANTS(&x4_read_cache_variants,
++					      &x1_write_cache_variants,
++					      &x1_update_cache_variants),
++		     0,
++		     SPINAND_ECCINFO(&micron_m68a_ooblayout, NULL)),
```
openwrt-ai, line 55: `micron_m68a_ooblayout` duplicates the existing `micron_4_ooblayout` for a chip with identical geometry — and does it slightly worse. MT29F1G01AAADD (M68A) has the same 2048+64 page and 4-bit/512B ECC requirement as [the MT29F2G01AAAED (M69A) entry directly below](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/mtd/nand/spi/micron.c#L407-L416), which already uses [`micron_4_ooblayout`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/mtd/nand/spi/micron.c#L96-L134):

- `micron_4_ooblayout_free()` returns 8 free bytes for sections 1..3 (only section 0 reserves the 2 BBM bytes). The new `micron_m68a_ooblayout_free()` reserves 2 bytes in *every* section, so it silently gives up 6 usable free-OOB bytes per page.
- The new `.ecc` callback is byte-for-byte the same layout as `micron_4_ooblayout_ecc()`, except `if (section >= 4)` hardcodes what `micron_4_ooblayout` derives from `memorg.pagesize / mtd->ecc_step_size`.

Please reuse the existing ops and drop the three functions added at lines 24-51 of this patch.

```suggestion
+		     SPINAND_ECCINFO(&micron_4_ooblayout, NULL)),
```


---

target/linux/generic/pending-6.18/418-mtd-spinand-micron-add-MT29F1G01AAADD.patch
```diff
@@ -0,0 +1,75 @@
...
+
+Add support for the Micron MT29F1G01AAADD (M68A) SPI NAND chip.
+
+Signed-off-by: Vít Kazík <catelixor@proton.me>
+---
+ drivers/mtd/nand/spi/micron.c | 39 +++++++++++++++++++++++++++++++++++
+ 1 file changed, 39 insertions(+)
+
+diff --git a/drivers/mtd/nand/spi/micron.c b/drivers/mtd/nand/spi/micron.c
+index b8130e0..1ae0f67 100644
```
openwrt-ai: nit: none of the 189 existing patches in `target/linux/generic/pending-6.18/` (nor any in `target/linux/ath79/patches-6.18/`) carry `index <blob>..<blob>` lines — they are all quilt-refreshed, which strips them and normalises the hunk headers/context. Please regenerate with `make target/linux/refresh` (with the matching target selected in `.config`).

Same applies to line 19 of `target/linux/ath79/patches-6.18/319-MIPS-pci-ar724x-add-motorola-ncap500-quirk.patch`.

---

target/linux/ath79/patches-6.18/319-MIPS-pci-ar724x-add-motorola-ncap500-quirk.patch
```diff
@@ -0,0 +1,77 @@
+From 572c515ab2ccb530e5ba2169bcef204ce6128ae1 Mon Sep 17 00:00:00 2001
+From: Vít Kazík <catelixor@proton.me>
+Date: Thu, 30 Jul 2026 02:29:28 +0200
+Subject: [PATCH] ath79: add Motorola NCAP-500 PCI quirk
```
openwrt-ai: nit: this `Subject:` uses the OpenWrt commit prefix instead of the kernel subsystem prefix, so it disagrees both with the patch file name (`319-MIPS-pci-ar724x-...`) and with the two sibling patches that touch this same file: `315-...` is `MIPS: pci-ar724x: add QCA9550 reset sequence` and `318-...` is `MIPS: pci-ar724x: deassert the reset of PCIe endpoint`.

```suggestion
Subject: [PATCH] MIPS: pci-ar724x: add Motorola NCAP-500 quirk
```


---

target/linux/ath79/patches-6.18/319-MIPS-pci-ar724x-add-motorola-ncap500-quirk.patch
```diff
@@ -0,0 +1,77 @@
...
+ #define AR724X_PCI_REG_RESET		0x18
+@@ -371,15 +372,26 @@ static void ar724x_pci_hw_init(struct ar724x_pci_controller *apc)
+ 		ath79_pll_wr(AR724X_PLL_REG_PCIE_CONFIG, ppl);
+ 	}
+ 
+-	/* deassert the reset state of the PCIE endpoint */
+-	rst = __raw_readl(apc->ctrl_base + AR724X_PCI_REG_RESET);
+-	rst |= AR724X_PCI_RESET_EP_RESET_L;
+-	__raw_writel(rst, apc->ctrl_base + AR724X_PCI_REG_RESET);
++	if (of_machine_is_compatible("motorola,ncap-500")) {
```
openwrt-ai: Keying a board quirk off `of_machine_is_compatible("motorola,ncap-500")` bakes a single board name into a generic MIPS PCI driver, which makes this patch permanently non-upstreamable and something OpenWrt has to carry indefinitely.

The driver already keeps the controller node in `apc->np`, and the sibling SoC quirk in [`315-MIPS-pci-ar724x-add-QCA9550-reset-sequence.patch`](``https://github.com/openwrt/openwrt/blob/ba8259ba6f9e0714155ca02b6368fd844194e827/target/linux/ath79/patches-6.18/315-MIPS-pci-ar724x-add-QCA9550-reset-sequence.patch)`` keys off `of_device_is_compatible(apc->np, "qcom,qca9550-pci")`. Could this be driven from a property/compatible on the board's `&pcie` node instead? That would also remove the need for the second `of_machine_is_compatible()` call in `ar724x_pci_probe()` at line 71.

Separately, the patch description says what the code does but not what it fixes — please state the symptom (PCIe link never comes up / card not enumerated because the OEM U-Boot leaves the host controller out of reset, etc.) so the quirk can be re-evaluated later.

---

target/linux/ath79/patches-6.18/319-MIPS-pci-ar724x-add-motorola-ncap500-quirk.patch
```diff
@@ -0,0 +1,77 @@
...
++		msleep(100);
++		__raw_writel(AR724X_PCI_RESET_EP_RESET_L, apc->ctrl_base + AR724X_PCI_REG_RESET);
++		msleep(100);
+ 
+-	/* set PCIE Application Control to ready */
+-	app = __raw_readl(apc->ctrl_base + AR724X_PCI_REG_APP);
+-	app |= AR724X_PCI_APP_LTSSM_ENABLE;
+-	__raw_writel(app, apc->ctrl_base + AR724X_PCI_REG_APP);
++		/* Motorola NCAP-500: apply PCIe application control settings */
++		__raw_writel(0x1ffc1, apc->ctrl_base + AR724X_PCI_REG_APP);
```
openwrt-ai: Two issues in this branch:

1. `__raw_writel(0x0, ...)` and `__raw_writel(AR724X_PCI_RESET_EP_RESET_L, ...)` overwrite the entire `AR724X_PCI_REG_RESET` register rather than clearing/setting just `AR724X_PCI_RESET_EP_RESET_L`. The `else` branch a few lines below deliberately does a read-modify-write on the same register; please do the same here so any other bits are preserved.

2. `0x1ffc1` is an undocumented magic value. Only bit 0 is named (`AR724X_PCI_APP_LTSSM_ENABLE`); bits 6-16 are not. Please add `#define`s for the bits being set, or at minimum a comment naming where the value came from (vendor U-Boot / Atheros SDK sequence) — as written nobody can later tell whether it is still correct or which bits matter.

Both fixes change the line count, so the hunk header will need regenerating afterwards (`make target/linux/refresh`).

nit: line 42 is ~95 columns; the rest of this file wraps well before that.

---

target/linux/ath79/patches-6.18/319-MIPS-pci-ar724x-add-motorola-ncap500-quirk.patch
```diff
@@ -0,0 +1,77 @@
...
+ 
+ 	/* wait up to 100ms for PHY link up */
+ 	do {
+@@ -436,7 +448,8 @@ static int ar724x_pci_probe(struct platform_device *pdev)
+ 	 * Do the full PCIE Root Complex Initialization Sequence if the PCIe
+ 	 * host controller is in reset.
+ 	 */
+-	if (reset_control_status(apc->hc_reset))
++	if (reset_control_status(apc->hc_reset) ||
++		of_machine_is_compatible("motorola,ncap-500"))
```
openwrt-ai: nit: the continuation of a wrapped `if` condition should be aligned with the opening parenthesis, not indented one extra tab — as written it reads like the body of the `if`.

```suggestion
+	if (reset_control_status(apc->hc_reset) ||
+	    of_machine_is_compatible("motorola,ncap-500"))
```


---

target/linux/ath79/dts/ar7242_motorola_ncap-500.dts
```diff
@@ -0,0 +1,141 @@
...
+	nvmem-cell-names = "mac-address";
+};
+
+&eth1 {
+	status = "disabled";
+};
+
+&spi {
+	status = "okay";
+	num-cs = <3>;
```
openwrt-ai: `num-cs` has no effect on this controller: `spi-ath79` hardcodes [`host->num_chipselect = 3;`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/spi/spi-ath79.c#L189) and never reads the property. No other DTS under `target/linux/ath79/dts/` sets it, including [`qca9531_glinet_gl-e750.dts`](https://github.com/openwrt/openwrt/blob/ba8259ba6f9e0714155ca02b6368fd844194e827/target/linux/ath79/dts/qca9531_glinet_gl-e750.dts#L123-L140), the other ath79 board with a NOR + SPI-NAND pair on this bus.

```suggestion
	status = "okay";
```


---

target/linux/ath79/dts/ar7242_motorola_ncap-500.dts
```diff
@@ -0,0 +1,141 @@
...
+
+					ethaddr: ethaddr {
+					};
+				};
+			};
+
+			partition@e0000 {
+				label = "kernel";
+				compatible = "denx,uimage";
+				reg = <0x0e0000 0x680000>;
```
openwrt-ai, line 115: `compatible = "denx,uimage"` is a no-op on a kernel-only partition, and a latent hazard. The mtdsplit uImage parser needs a rootfs *inside the same partition, immediately after the uImage* — see [`mtd_find_rootfs_from()` in mtdsplit_uimage.c](``https://github.com/openwrt/openwrt/blob/ba8259ba6f9e0714155ca02b6368fd844194e827/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_uimage.c#L200-L207).`` Here the rootfs lives on the SPI NAND, so the parser always returns `-ENOENT` and no subpartitions are ever created. Every in-tree use of `denx,uimage` is on a `firmware` partition holding kernel+rootfs.

And if the split ever did succeed (e.g. leftover stock data further into the partition carrying a squashfs/UBI magic), it would register a *second* mtd also named `kernel`, while `nand_upgrade_tar()` resolves the kernel partition purely by name via [`find_mtd_index "$CI_KERNPART"`](https://github.com/openwrt/openwrt/blob/ba8259ba6f9e0714155ca02b6368fd844194e827/package/base-files/files/lib/upgrade/nand.sh#L322) — making sysupgrade ambiguous.

```suggestion
				label = "kernel";
				reg = <0x0e0000 0x680000>;
```


---

target/linux/ath79/dts/ar7242_motorola_ncap-500.dts
```diff
@@ -0,0 +1,141 @@
...
+
+			partition@d0000 {
+				label = "u-boot-env";
+				reg = <0x0d0000 0x010000>;
+				read-only;
+
+				nvmem-layout {
+					compatible = "u-boot,env";
+
+					ethaddr: ethaddr {
```
openwrt-ai: nit: a bare `ethaddr` phandle label for the nvmem cell is the minority spelling in-tree (4 occurrences, vs 27 `macaddr_uboot_ethaddr` and 16 `macaddr_ubootenv_ethaddr`); e.g. [`qca9557_zyxel_nbg6616.dts:85`](https://github.com/openwrt/openwrt/blob/ba8259ba6f9e0714155ca02b6368fd844194e827/target/linux/ath79/dts/qca9557_zyxel_nbg6616.dts#L85) uses `macaddr_uboot_ethaddr: ethaddr`. Renaming also needs the reference at line 70 updated, so no suggestion block here.

---

target/linux/ath79/image/nand.mk
```diff
@@ -348,6 +348,22 @@ define Device/meraki_z1
 endef
 TARGET_DEVICES += meraki_z1
 
+define Device/motorola_ncap-500
+  SOC := ar7242
+  DEVICE_VENDOR := Motorola Solutions
+  DEVICE_MODEL := NCAP-500
+  DEVICE_ALT0_MODEL := AP-0621
```
openwrt-ai, line 356: `DEVICE_ALT0_MODEL` without a matching `DEVICE_ALT0_VENDOR` produces a title with a leading space: [`DEVICE_ALT0_TITLE = $(DEVICE_ALT0_VENDOR) $(DEVICE_ALT0_MODEL)`](https://github.com/openwrt/openwrt/blob/ba8259ba6f9e0714155ca02b6368fd844194e827/include/image.mk#L439), and `DEVICE_ALT0_VENDOR` is reset to empty in `Device/Default`. Every other in-tree device that sets `DEVICE_ALT0_MODEL` also sets or inherits `DEVICE_ALT0_VENDOR`.

```suggestion
  DEVICE_MODEL := NCAP-500
  DEVICE_ALT0_VENDOR := Motorola Solutions
  DEVICE_ALT0_MODEL := AP-0621
```

nit: the commit message quotes `mfg_hardware_model=AP-0621-60020-EU`. Is `AP-0621` the intended alt model, or should the board/region suffix be kept?

---

target/linux/ath79/image/nand.mk
```diff
@@ -348,6 +348,22 @@ define Device/meraki_z1
...
+  KERNEL_SIZE := 6656k
+  BLOCKSIZE := 128k
+  PAGESIZE := 2048
+  VID_HDR_OFFSET := 2048
+  KERNEL := kernel-bin | append-dtb | lzma | uImage lzma
```
openwrt-ai: nit: this `KERNEL` recipe is byte-identical to ath79's [`Device/Default`](https://github.com/openwrt/openwrt/blob/ba8259ba6f9e0714155ca02b6368fd844194e827/target/linux/ath79/image/Makefile#L76), so the line is redundant.

```suggestion
  VID_HDR_OFFSET := 2048
```

nit: `rootfs.ubi` is a new image name — in-tree the raw UBI image is named `factory.ubi` (qualcommax, ipq40xx, mvebu, pistachio) or `nand.ubi` (imx, layerscape). Worth matching one of those, and the commit message's `ubiformat ... -f /tmp/rootfs.ubi` line updated accordingly.

---

openwrt-ai (commented):
Reviewed 3 commits (fresh review at `ba8259b`). All three commit messages match their changes, and the new-device commit carries the required hardware / flash-instructions / MAC-address sections, so no commit-message issues.

Main findings are inline. The two worth attention before merge are the duplicated `micron_4_ooblayout` in the SPI-NAND patch and the register clobbering / magic constant in the ar724x PCIe quirk; the `denx,uimage` property on the kernel partition and the missing `DEVICE_ALT0_VENDOR` are smaller but concrete.

---

target/linux/generic/pending-6.18/418-mtd-spinand-micron-add-MT29F1G01AAADD.patch
ByteFloor: MT29F1G01AAADD (M68A) according to datasheet uses different spare area layout than is used in micron_4_ooblayout which causes the UBIFS to fail to mount.

target/linux/generic/pending-6.18/418-mtd-spinand-micron-add-MT29F1G01AAADD.patch
```diff
@@ -0,0 +1,70 @@
...
+ 				unsigned int target)
+ {
+@@ -404,6 +433,16 @@ static const struct spinand_info micron_
+ 		     SPINAND_ECCINFO(&micron_8_ooblayout,
+ 				     micron_8_ecc_get_status),
+ 		     SPINAND_SELECT_TARGET(micron_select_target)),
++	/* M68A 1Gb 3.3V */
++	SPINAND_INFO("MT29F1G01AAADD",
++		     SPINAND_ID(SPINAND_READID_METHOD_OPCODE_DUMMY, 0x12),
++		     NAND_MEMORG(1, 2048, 64, 64, 1024, 20, 2, 1, 1),
```
openwrt-ai, line 49: `planes_per_lun = 2` (7th `NAND_MEMORG` argument) makes the SPI-NAND core OR a plane-select bit into the column address for every odd eraseblock: `pos.plane = eraseblock % planes_per_lun` ([nand.h:759](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/include/linux/mtd/nand.h#L759)), then [`column |= req->pos.plane << fls(nanddev_page_size(nand));`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/mtd/nand/spi/core.c#L406) on read and again [at core.c:498](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/mtd/nand/spi/core.c#L498) on write. With a 2048-byte page that sets bit 12 of the column address on half of all blocks.

Every 1Gb SPI-NAND entry in the tree (1024 eraseblocks x 64 pages x 2048) declares a single plane, including both existing Micron 1Gb parts — see [`MT29F1G01ABAFD` at micron.c:327-330](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/mtd/nand/spi/micron.c#L327-L330). The `2` here looks copied from the 2Gb M69A entry directly below, which really is dual-plane.

```suggestion
+		     NAND_MEMORG(1, 2048, 64, 64, 1024, 20, 1, 1, 1),
```

---

target/linux/generic/pending-6.18/418-mtd-spinand-micron-add-MT29F1G01AAADD.patch
```diff
@@ -0,0 +1,70 @@
...
+ 1 file changed, 39 insertions(+)
+
+--- a/drivers/mtd/nand/spi/micron.c
++++ b/drivers/mtd/nand/spi/micron.c
+@@ -134,6 +134,35 @@ static const struct mtd_ooblayout_ops mi
+ 	.free = micron_4_ooblayout_free,
+ };
+ 
++static int micron_m68a_ooblayout_ecc(struct mtd_info *mtd, int section,
++				   struct mtd_oob_region *region)
```
openwrt-ai: nit: understood re: the `.free` layout — the 2 reserved bytes at the start of *every* 16-byte section is a real difference from `micron_4_ooblayout_free()`, so that one stays.

But `micron_m68a_ooblayout_ecc()` is functionally identical to [`micron_4_ooblayout_ecc()`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/mtd/nand/spi/micron.c#L95-L108): same `region->offset = (section * 16) + 8` / `length = 8`, and the hardcoded `section >= 4` is exactly what that function derives from `memorg.pagesize / mtd->ecc_step_size` (2048/512). So the new ops struct at lines 46-49 can just reuse it and lines 22-32 can go away:

```c
static const struct mtd_ooblayout_ops micron_m68a_ooblayout = {
	.ecc = micron_4_ooblayout_ecc,
	.free = micron_m68a_ooblayout_free,
};
```

No suggestion block since that changes the added-line count — regenerate the patch afterwards so the hunk header and diffstat match (`make target/linux/refresh`).

Separately, if the `_ecc` helper does stay: the argument continuation on this line is one space short of the opening parenthesis (`micron_m68a_ooblayout_ecc(` needs 4 tabs + 4 spaces), and line 35 has the same problem (`micron_m68a_ooblayout_free(` needs 4 tabs + 5 spaces). checkpatch reports both as "Alignment should match open parenthesis".

---

openwrt-ai (commented):
Re-reviewed the 3 rewritten commits (`ba8259b` → `8a33458`). Nearly all of the previous round is addressed: the patch metadata is now quilt-clean and the hunk headers/diffstats add up, the PCIe quirk moved to `of_device_is_compatible(apc->np, "motorola,ncap-500-pci")` with the board compatible added to `&pcie` (the `qcom,ar7240-pci` fallback keeps both `ar724x_pci_ids[]` and the `of_irq_imap_abusers` workaround in patch 810 matching), the reset write is now a read-modify-write, `denx,uimage` / `num-cs` / the redundant `KERNEL` recipe are gone, and `DEVICE_ALT0_VENDOR` + `factory.ubi` are in place.

Two things left, both in the SPI-NAND patch — see inline. The `NAND_MEMORG` plane count is the one worth acting on.

## Commit checks

- `8a33458` "ath79: nand: add support for Motorola NCAP-500" — the flash instructions jump from booting the TFTP initramfs straight to `mtd write /tmp/kernel kernel` and `ubiformat /dev/mtd4 -y -f /tmp/factory.ubi`, but never say where those two files come from. `kernel` is not a build artifact name — it is a member of the sysupgrade tar (`openwrt-ath79-nand-motorola_ncap-500-squashfs-sysupgrade.bin`), and there is no step transferring either file onto the device. Adding the extract/scp step would make the procedure self-contained.

---

target/linux/generic/pending-6.18/418-mtd-spinand-micron-add-MT29F1G01AAADD.patch
ByteFloor: From MT29F1G01AAADD datasheet first page:
Page size x1: 2112 bytes (2048 + 64 bytes)
Block size: 64 pages (128K + 4K bytes)
Plane size: 2 planes x 512 blocks per plane
Device size: 1Gb: 1024 blocks

target/linux/generic/pending-6.18/418-mtd-spinand-micron-add-MT29F1G01AAADD.patch
```diff
@@ -0,0 +1,58 @@
...
+ 1 file changed, 27 insertions(+)
+
+--- a/drivers/mtd/nand/spi/micron.c
++++ b/drivers/mtd/nand/spi/micron.c
+@@ -134,6 +134,23 @@ static const struct mtd_ooblayout_ops mi
+ 	.free = micron_4_ooblayout_free,
+ };
+ 
++static int micron_m68a_ooblayout_free(struct mtd_info *mtd, int section,
++				     struct mtd_oob_region *region)
```
openwrt-ai: nit: my whitespace count in the previous round was off by one, sorry — this continuation is still one space short of the open parenthesis, so checkpatch will still report "Alignment should match open parenthesis".

`static int micron_m68a_ooblayout_free(` is 38 characters, so `(` sits in column 38 and the first argument starts in column 39. The continuation therefore needs **4 tabs + 6 spaces** (4 tabs land the cursor in column 33, 6 spaces take it to 39); the line currently has 4 tabs + 5 spaces and starts `struct` in column 38.

Compare [`micron_4_ooblayout_free()` at micron.c:110-111](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/mtd/nand/spi/micron.c#L110-L111), whose shorter name puts the paren in column 35 and uses 4 tabs + 3 spaces.

No suggestion block here because the fix is a tab/space mix that GitHub's suggestion editor tends to mangle.

---

openwrt-ai (commented):
Re-reviewed the range `8a33458` → `7254be4`. The only code change is in `418-mtd-spinand-micron-add-MT29F1G01AAADD.patch`; the DTS, `nand.mk` and the ar724x PCIe quirk patch are byte-identical to the previous round, so nothing new to say there.

What changed, and how it lands:

- `micron_m68a_ooblayout_ecc()` is gone and the ops struct now points at `micron_4_ooblayout_ecc` — that was the open item from last round, resolved. The hunk headers and diffstat still add up (`+134,23` = 6 context + 17 added; `+421,16` = 6 + 10; 421 = 404 + 17), and the ath79/generic "Check Kernel patches" job on this head is green, so the patch applies clean.
- `micron_m68a_ooblayout_free()` staying separate is fine — the 2 reserved bytes at the head of *every* 16-byte section really are different from `micron_4_ooblayout_free()`, which only reserves them in section 0.
- `planes_per_lun = 2`: withdrawing my earlier objection. Your datasheet quote (2 planes x 512 blocks) settles it, and it is consistent with the dual-plane `MT29F2G01AAAED` entry right below, which also declares 2 planes with a 2048-byte page. My "every 1Gb part declares 1 plane" claim was drawn from `MT29F1G01ABAFD`, which is a genuinely single-plane part — not a counterexample.
- The commit message now carries the extract/`scp` steps for both `kernel` and `factory.ubi`, so the flash procedure is self-contained. That closes the commit-message finding from last round; all three commit messages match their diffs.

One nit left inline, on the argument alignment — including a correction to my own miscount from last round.

---

target/linux/generic/pending-6.18/418-mtd-spinand-micron-add-MT29F1G01AAADD.patch
```diff
@@ -0,0 +1,58 @@
...
+ 1 file changed, 27 insertions(+)
+
+--- a/drivers/mtd/nand/spi/micron.c
++++ b/drivers/mtd/nand/spi/micron.c
+@@ -134,6 +134,23 @@ static const struct mtd_ooblayout_ops mi
+ 	.free = micron_4_ooblayout_free,
+ };
+ 
++static int micron_m68a_ooblayout_free(struct mtd_info *mtd, int section,
++				       struct mtd_oob_region *region)
```
openwrt-ai: nit: this went from one space short to one space too many — it is now 4 tabs + **7** spaces, and it needs 4 tabs + **6**. Just delete one space; everything else about the line is right.

The likely trap: the alignment has to be measured on the *applied* source, not on the patch file. In the patch file each line carries a leading `+`, so aligning what you see there shifts the continuation one column right once the `+` is stripped. On the applied line, `static int micron_m68a_ooblayout_free(` puts `(` in column 38 and the first argument (`struct mtd_info`) in column 39; 4 tabs land the cursor in column 33, so 6 spaces reach 39, while 7 overshoot to 40. checkpatch still reports "Alignment should match open parenthesis" for the overshoot.

Cross-check with [`micron_4_ooblayout_free()` at micron.c:111-112](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/mtd/nand/spi/micron.c#L111-L112) (v6.18.41): its shorter name puts `(` in column 35, and its continuation uses 4 tabs + 3 spaces to reach column 36 — same rule, different offset.

Still no suggestion block, since a tab/space mix is exactly what GitHub's suggestion editor tends to mangle.

This is the only thing I have left on the series.

---

openwrt-ai (commented):
Re-reviewed the range `7254be4` → `d86c23c`. The three commits were re-created with new SHAs, but all three commit messages and diffstats are byte-identical to the previous round, and the only content change in the whole range is a single line in `418-mtd-spinand-micron-add-MT29F1G01AAADD.patch`:

```
-+				     struct mtd_oob_region *region)
++				       struct mtd_oob_region *region)
```

So nothing new on the DTS, `nand.mk`, or the ar724x PCIe quirk, and no commit-message issues.

That one line was the last open nit, and it overshot by one space in the other direction — details inline. The ath79/generic kernel build on this head is green, so the patch still applies cleanly either way; this is cosmetic only.

---

openwrt-ai (commented):
Re-reviewed the range `d86c23c` → `5008acf`. All three commits were re-created with new SHAs, but all three commit messages are byte-identical to the previous round and each commit's diffstat is unchanged, so nothing new on the DTS, `nand.mk`, `02_network`, or the ar724x PCIe quirk patch.

The entire content change in the range is one space in `418-mtd-spinand-micron-add-MT29F1G01AAADD.patch`, and it lands in the right commit (`10a9ba8`, the SPI-NAND one):

```
-+				       struct mtd_oob_region *region)
++				      struct mtd_oob_region *region)
```

That closes the last open nit, and this time the count is right. On the applied source `static int micron_m68a_ooblayout_free(` puts `(` in column 38 and `struct mtd_info` in column 39; the continuation is now 4 tabs + 6 spaces, which lands exactly on column 39. Same rule as [`micron_4_ooblayout_free()` at micron.c:111-112](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/mtd/nand/spi/micron.c#L111-L112) (v6.18.41), whose shorter name puts `(` in column 35 and uses 4 tabs + 3 spaces. checkpatch's "Alignment should match open parenthesis" should be clean now.

Patch metadata still adds up: `@@ -134,6 +134,23 @@` = 6 context + 17 added, `@@ -404,6 +421,16 @@` = 6 + 10, total 27 insertions matching the diffstat, and 421 = 404 + 17. CI on this head is green, including `Check Kernel patches (ath79, generic)` and `Build ath79/generic`.

Reviewed 3 new commits; no new issues found. Nothing left open from my side on this series.

---

PR #17296, opened 2024-12-18
mediatek: filogic: Add support for D-Link AQUILA PRO AI M60

RolandoMagico:
Specification:
 - MT7986 CPU using 2.4GHz and 5GHz WiFi (both AX)
 - MT7531 switch
 - 512MB RAM
 - 128MB NAND flash (MX35LF1GE4AB-Z4I) with two UBI partitions with identical size
 - 1 multi color LED (red, green, blue, white) connected via GCA230718 (Same as D-Link M30 A1)
 - 3 buttons (WPS, reset, LED on/off)
 - 1x 2.5 Gbit WAN port with Maxlinear GPY211C
 - 4x 1 Gbit LAN ports

Disassembly:
 - There are five screws at the bottom: 2 under the rubber feet, 3 under the label.
 - After removing the screws, the white plastic part can be shifted out of the blue part.
 - Be careful because the antennas are mounted on the side and the top of the white part.

Serial Interface
 - The serial interface can be connected to the 4 pin holes next to/under the antenna cables.
 - Note that there is another set of 4 pin holes on the side of the board, it's not used.
 - Pins (from front to rear):
   - 3.3V (do not connect)
   - TX
   - RX
   - GND
 - Settings: 115200, 8N1

MAC addresses:
 - MAC address is stored in partition "Odm" at offset 0x81 (for example XX:XX:XX:XX:XX:52)
 - MAC address on the device label is ODM + 1 (for example XX:XX:XX:XX:XX:53)
 - WAN MAC is the one from the ODM partition (for example XX:XX:XX:XX:XX:52)
 - LAN MAC is the one from the ODM partition + 1 (for example XX:XX:XX:XX:XX:53)
 - WLAN MAC (2.4 GHz) is the one from the ODM partition + 2 (for example (XX:XX:XX:XX:XX:54)
 - WLAN MAC (5 GHz) is the one from the ODM partition + 5 (for example (XX:XX:XX:XX:XX:57)

Flashing via OEM web interface:
 - Currently not supported because image crypto is not known

Flashing via recovery web interface:
 - This is only working if the first partition is active because recovery images are always flashed to the active partition and OpenWrt can only be executed from the first partition
 - Use a Chromium based browser, otherwise firmware upgrade might not work
 - Recovery web interface is accessible via 192.168.200.1 after keeping the reset button pressed during start of the device until the LED blinks red
 - Upload the recovery image, this will take some time. LED will continue flashing red during the update process
[...]

PR #23902, opened 2026-06-21
mediatek: dts: fix MAC address offset for D-Link Aquila Pro AI M60

godot87:
The MAC address base was read from offset 0x81 instead of the correct 0x83, causing all units to derive their interface MACs from the same incorrect base (06:00:DC:EA:E7:AE). The two bytes at 0x81 are identical across all devices and precede the actual device-specific MAC data.

Changing the offset to 0x83 yields the correct base MAC with the registered D-Link OUI (DC:EA:E7), unique per device. Fix verified by recompiling and flashing the corrected DTS.

Update: Tested on 4 devices based on stable 25.12.4
I have compiled a custom image based on the latest stable release (25.12.4) with this patch applied and flashed it to all four of my D-Link M60 A1 units.
On all four devices the MAC addresses are now correctly derived from the device-specific base in the ODM partition. The resulting interface MACs match the expected values based on the label MAC printed on each device.
Tested on: D-Link M60 A1 × 4

PR #24484, opened 2026-07-30
mediatek: filogic: fix support for D-Link AQUILA PRO AI M60 sold in multi packs

nicefile:
D-Link AQUILA PRO AI M60 is sold as single device and also in pack of 2 or 3 devices
Device MAC is stored on  'Odm' partition but for M60-2 and M60-3 devices offset change.
This change moves MAC setup from .dts to userspace to allow M60 M60-2 M60-3 MAC detection.

OEM and OpenWrt share same MAC table:

WAN = Odm = label -1
LAN1-4 = eth0  = label
WiFi 2.4G = Odm+2
WiFi 5G = Odm+5

Fixes: https://github.com/openwrt/openwrt/commit/b3ce08e0b6fa6780bf7ee295a1f176c053b1100b ("mediatek: filogic: Add support for D-Link AQUILA PRO AI M60")

note:
`eth0` stays random due to interface is up before preinit script.
this suppress #23902 which use fixed address and will cause regression on single packed device

### DIFF
--- a/target/linux/mediatek/dts/mt7986a-dlink-aquila-pro-ai-m60-a1.dts
+++ b/target/linux/mediatek/dts/mt7986a-dlink-aquila-pro-ai-m60-a1.dts
@@ -18,7 +18,6 @@
 		led-failsafe = &led_status_red;
 		led-running = &led_status_white;
 		led-upgrade = &led_status_blue;
-		label-mac-device = &gmac0;
 	};
 
 	chosen {
@@ -61,8 +60,6 @@
 		reg = <0>;
 		phy-mode = "2500base-x";
 
-		nvmem-cells = <&macaddr_odm 1>;
-		nvmem-cell-names = "mac-address";
 		fixed-link {
 			speed = <2500>;
 			full-duplex;
@@ -75,8 +72,6 @@
 		reg = <1>;
 		phy-mode = "2500base-x";
 		phy-handle = <&phy6>;
-		nvmem-cells = <&macaddr_odm 0>;
-		nvmem-cell-names = "mac-address";
 		label = "internet";
 	};
 
@@ -273,19 +268,6 @@
 				label = "Odm";
 				reg = <0x6980000 0x40000>;
 				read-only;
-
-				nvmem-layout {
-					compatible = "fixed-layout";
-					#address-cells = <1>;
-					#size-cells = <1>;
-
-					macaddr_odm: macaddr@81 {
-						compatible = "mac-base";
-						reg = <0x81 0x6>;
-						#nvmem-cell-cells = <1>;
-					};
-				};
-
 			};
 
 			partition@69c0000 {
@@ -311,28 +293,11 @@
 };
 
 &wifi {
-	#address-cells = <1>;
-	#size-cells = <0>;
 	status = "okay";
 	pinctrl-names = "default";
 	pinctrl-0 = <&wf_2g_5g_pins>;
-
-
 	nvmem-cells = <&eeprom_factory_0>;
 	nvmem-cell-names = "eeprom";
-
-	band@0 {
-		/* 2.4 GHz */
-		reg = <0>;
-		nvmem-cells = <&macaddr_odm 2>;
-		nvmem-cell-names = "mac-address";
-	};
-	band@1 {
-		/* 5 GHz */
-		reg = <1>;
-		nvmem-cells = <&macaddr_odm 5>;
-		nvmem-cell-names = "mac-address";
-	};
 };
 
 &uart0 {
--- a/target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
+++ b/target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
@@ -121,6 +121,18 @@ case "$board" in
 		addr=$(mtd_get_mac_binary "Odm" 0x81)
 		[ "$PHYNBR" = "1" ] && macaddr_add $addr 3 > /sys${DEVPATH}/macaddress
 		;;
+	dlink,aquila-pro-ai-m60-a1)
+		# multi-pack units (M60-2/M60-3) carry a '-' (0x2d) at offset 0x4d
+		# and store MAC at offset 0x83
+		odm=$(find_mtd_part "Odm")
+		if [ "$(hexdump -n 1 -v -s 0x4d -e '1 "%02x"' "$odm")" = "2d" ]; then
+			addr=$(get_mac_binary "$odm" 0x83)
+		else
+			addr=$(get_mac_binary "$odm" 0x81)
+		fi
+		[ "$PHYNBR" = "0" ] && macaddr_add $addr 2 > /sys${DEVPATH}/macaddress
+		[ "$PHYNBR" = "1" ] && macaddr_add $addr 5 > /sys${DEVPATH}/macaddress
+		;;
 	glinet,gl-mt6000|\
 	glinet,gl-x3000|\
 	glinet,gl-xe3000)
--- a/target/linux/mediatek/filogic/base-files/lib/preinit/10_fix_eth_mac.sh
+++ b/target/linux/mediatek/filogic/base-files/lib/preinit/10_fix_eth_mac.sh
@@ -28,6 +28,21 @@ preinit_set_mac_address() {
 		ip link set dev eth0 address "$addr"
 		ip link set dev eth1 address "$addr"
 		;;
+	dlink,aquila-pro-ai-m60-a1)
+		# multi-pack units (M60-2/M60-3) carry a '-' (0x2d) at offset 0x4d
+		# and store MAC at offset 0x83
+		odm=$(find_mtd_part "Odm")
+		if [ "$(hexdump -n 1 -v -s 0x4d -e '1 "%02x"' "$odm")" = "2d" ]; then
+			wan_mac=$(get_mac_binary "$odm" 0x83)
+		else
+			wan_mac=$(get_mac_binary "$odm" 0x81)
+		fi
+		ip link set dev lan1 address "$(macaddr_add "$wan_mac" 1)"
+		ip link set dev lan2 address "$(macaddr_add "$wan_mac" 1)"
+		ip link set dev lan3 address "$(macaddr_add "$wan_mac" 1)"
+		ip link set dev lan4 address "$(macaddr_add "$wan_mac" 1)"
+		ip link set dev internet address "$wan_mac"
+		;;
 	mercusys,mr90x-v1|\
 	tplink,archer-ax80-v1|\
 	tplink,archer-ax80-v1-eu|\


nicefile:
grammar corrected and mac description corrected

target/linux/mediatek/dts/mt7986a-dlink-aquila-pro-ai-m60-a1.dts
```diff
@@ -18,7 +18,6 @@
 		led-failsafe = &led_status_red;
 		led-running = &led_status_white;
 		led-upgrade = &led_status_blue;
-		label-mac-device = &gmac0;
```
openwrt-ai, line 21: Dropping this alias silently removes the label MAC address for this board, and nothing in the PR replaces it. `get_mac_label_dt()` reads `/proc/device-tree/aliases/label-mac-device` and then the referenced node's `mac-address` property; with the alias gone (and `gmac0` no longer carrying an nvmem-derived MAC) it returns empty, so `system.label_macaddr` disappears from `/etc/board.json` and LuCI/`ubus` no longer report the label MAC.

The filogic target already has a userspace hook for exactly this: [`mediatek_setup_macs()` in `board.d/02_network`](https://github.com/openwrt/openwrt/blob/ed5bdd61334033f32c2c4bd8bc9cd0ecf7b0d539/target/linux/mediatek/filogic/base-files/etc/board.d/02_network#L249) sets `label_mac` and passes it to [`ucidef_set_label_macaddr`](``https://github.com/openwrt/openwrt/blob/ed5bdd61334033f32c2c4bd8bc9cd0ecf7b0d539/target/linux/mediatek/filogic/base-files/etc/board.d/02_network#L354).`` Since the same Odm/offset detection is needed there anyway, adding a `dlink,aquila-pro-ai-m60-a1)` case that sets `label_mac` (= Odm+1, per the table in the commit message), plus `lan_mac`/`wan_mac`, would keep parity with the pre-change behaviour. Sibling [`mt7981b-dlink-aquila-pro-ai-m30-a1.dts:12`](https://github.com/openwrt/openwrt/blob/ed5bdd61334033f32c2c4bd8bc9cd0ecf7b0d539/target/linux/mediatek/dts/mt7981b-dlink-aquila-pro-ai-m30-a1.dts#L12) still has the alias, so the M60 becomes the odd one out in the family.

If the loss of the label MAC is intentional, it would be worth saying so in the commit message.

---

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
```diff
@@ -121,6 +121,12 @@ case "$board" in
 		addr=$(mtd_get_mac_binary "Odm" 0x81)
 		[ "$PHYNBR" = "1" ] && macaddr_add $addr 3 > /sys${DEVPATH}/macaddress
 		;;
+	dlink,aquila-pro-ai-m60-a1)
+		odm=$(find_mtd_part "Odm")
+		addr=$(if [ $(hexdump -n 1 -v -s 0x4d -e '1 "%02x"' $odm) = 2d ];then get_mac_binary $odm 0x83 ;else get_mac_binary $odm 0x81 ;fi)
```
openwrt-ai: The `hexdump` output is unquoted, so when `find_mtd_part "Odm"` returns empty (partition missing/renamed) `hexdump` gets no file operand and reads **stdin** — `-n 1` will block until it gets a byte or EOF — and the test then degenerates to `[ = 2d ]`, which ash reports as `[: argument expected` and falls through to the `else` branch with an empty path. Quoting `"$odm"` makes `hexdump` fail cleanly instead.

The magic offsets also need a word of explanation — `0x4d`/`0x2d` are unreadable on their own, and this same expression is duplicated verbatim in `10_fix_eth_mac.sh`. The single-line `if` inside a command substitution is also unlike the rest of this file.

```suggestion
		odm=$(find_mtd_part "Odm")
		# Multi-pack units (M60-2/M60-3) carry a '-' (0x2d) at offset 0x4d
		# of the model string in "Odm" and store their MAC at 0x83;
		# single-pack M60 has it at 0x81.
		if [ "$(hexdump -n 1 -v -s 0x4d -e '1 "%02x"' "$odm")" = "2d" ]; then
			addr=$(get_mac_binary "$odm" 0x83)
		else
			addr=$(get_mac_binary "$odm" 0x81)
		fi
```

---

target/linux/mediatek/filogic/base-files/lib/preinit/10_fix_eth_mac.sh
```diff
@@ -28,6 +28,15 @@ preinit_set_mac_address() {
 		ip link set dev eth0 address "$addr"
 		ip link set dev eth1 address "$addr"
 		;;
+	dlink,aquila-pro-ai-m60-a1)
+		odm=$(find_mtd_part "Odm")
+		wan_mac=$(if [ $(hexdump -n 1 -v -s 0x4d -e '1 "%02x"' $odm) = 2d ];then get_mac_binary $odm 0x83 ;else get_mac_binary $odm 0x81 ;fi)
+		ip link set dev lan1 address "$(macaddr_add "$wan_mac" 1)"
+		ip link set dev lan2 address "$(macaddr_add "$wan_mac" 1)"
+		ip link set dev lan3 address "$(macaddr_add "$wan_mac" 1)"
+		ip link set dev lan4 address "$(macaddr_add "$wan_mac" 1)"
```
openwrt-ai, line 43: Same unquoted-`hexdump` problem as in `11_fix_wifi_mac`, and it is worse here: this runs during preinit, where a `hexdump` with no file operand reading stdin from the console can stall boot. Quote `"$odm"`.

Additionally, `macaddr_add "$wan_mac" 1` is recomputed four times for a value that never changes; the neighbouring `acer,vero-w6m` and `tplink,fr365-v1` cases in this file compute a `lan_mac` once and reuse it.

```suggestion
		odm=$(find_mtd_part "Odm")
		# Multi-pack units (M60-2/M60-3) carry a '-' (0x2d) at offset 0x4d
		# of the model string in "Odm" and store their MAC at 0x83;
		# single-pack M60 has it at 0x81.
		if [ "$(hexdump -n 1 -v -s 0x4d -e '1 "%02x"' "$odm")" = "2d" ]; then
			wan_mac=$(get_mac_binary "$odm" 0x83)
		else
			wan_mac=$(get_mac_binary "$odm" 0x81)
		fi
		lan_mac=$(macaddr_add "$wan_mac" 1)
		ip link set dev lan1 address "$lan_mac"
		ip link set dev lan2 address "$lan_mac"
		ip link set dev lan3 address "$lan_mac"
		ip link set dev lan4 address "$lan_mac"
```

---

target/linux/mediatek/filogic/base-files/lib/preinit/10_fix_eth_mac.sh
```diff
@@ -28,6 +28,15 @@ preinit_set_mac_address() {
...
+		ip link set dev internet address "$wan_mac"
```
openwrt-ai, line 44: The PR body explains that "eth0 stays random due to interface is up before preinit script", but the `asus,tuf-ax4200` case in this very function does [`ip link set dev eth0 address "$addr"`](``https://github.com/openwrt/openwrt/blob/ed5bdd61334033f32c2c4bd8bc9cd0ecf7b0d539/target/linux/mediatek/filogic/base-files/lib/preinit/10_fix_eth_mac.sh#L28-L29)`` from the same hook, so being "already up" does not appear to prevent it. Setting the lan1-4 user ports individually is correct (they don't follow the conduit once created), but leaving the DSA conduit eth0 with a locally-administered random MAC that changes on every boot is still a behaviour change versus the previous `nvmem-cells = <&macaddr_odm 1>`. Is there a reason not to also set `eth0` to the LAN/label MAC here?

---

target/linux/mediatek/dts/mt7986a-dlink-aquila-pro-ai-m60-a1.dts
```diff
@@ -324,14 +306,10 @@
...
 		reg = <0>;
-		nvmem-cells = <&macaddr_odm 2>;
-		nvmem-cell-names = "mac-address";
 	};
 	band@1 {
 		/* 5 GHz */
 		reg = <1>;
-		nvmem-cells = <&macaddr_odm 5>;
-		nvmem-cell-names = "mac-address";
 	};
```
openwrt-ai: nit: with the `nvmem-cells`/`nvmem-cell-names` gone, `band@0` and `band@1` carry nothing but their `reg` and can be dropped entirely. The `#address-cells = <1>;` / `#size-cells = <0>;` pair on `&wifi` (lines 296-297) exists only to keep dtc quiet about these two nodes — see [49d5fc505ffb0346f68ce91cb37109899af6b373](https://github.com/openwrt/openwrt/commit/49d5fc505ffb0346f68ce91cb37109899af6b373) — so those two lines become dead as well. No suggestion block since the fix spans a hunk this comment isn't anchored to.

---

openwrt-ai (commented):
## Commit checks

- `ed5bdd6` "mediatek: filogic: fix support for AQUILA PRO AI M60 sold in multi packs" — the diff matches the description, but the `Fixes: PR #17296` trailer is not a usable reference: `#17296` is a merged pull request, not an issue, and a bare `#NNNNN` carries no meaning in a git log outside GitHub. The project convention is `Fixes: <12-char sha> ("<subject>")`, i.e. `Fixes: b3ce08e0b6fa ("mediatek: filogic: Add support for D-Link AQUILA PRO AI M60")` (or `Fixes: 616621120bf2 ("mediatek: Fix primary MAC of D-Link M60")` if the intent is to point at the commit that installed the `macaddr_odm` layout being removed here). The PR body says `Fixes: #17296` while the commit says `Fixes: PR #17296` — worth making those agree too.
- The commit message documents the MAC table but never mentions that `label-mac-device` is being dropped, nor how M60-2/M60-3 are detected (byte `0x2d` at offset `0x4d` of `Odm`, MAC at `0x83` instead of `0x81`). Both are the substance of this change and belong in the message.


---

target/linux/mediatek/filogic/base-files/lib/preinit/10_fix_eth_mac.sh
nicefile: eth0 must be down for setting MAC . But kernel enables eth0 (as set in .dts) before prenit. 
As it was tested not working so I've drop eth0 entry

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
nicefile: updated

target/linux/mediatek/filogic/base-files/lib/preinit/10_fix_eth_mac.sh
nicefile: updated

target/linux/mediatek/dts/mt7986a-dlink-aquila-pro-ai-m60-a1.dts
nicefile: since .dts is not the source of MAC address this carry no valid information as `gmac0` MAC is random .
Best I can is another commit for reverse commit `616621120bf2 ("mediatek: Fix primary MAC of D-Link M60")`

target/linux/mediatek/dts/mt7986a-dlink-aquila-pro-ai-m60-a1.dts
```diff
@@ -320,19 +302,6 @@
...
-		nvmem-cells = <&macaddr_odm 2>;
-		nvmem-cell-names = "mac-address";
-	};
-	band@1 {
-		/* 5 GHz */
-		reg = <1>;
-		nvmem-cells = <&macaddr_odm 5>;
-		nvmem-cell-names = "mac-address";
-	};
 };
```
openwrt-ai, line 301: nit: now that `band@0`/`band@1` are gone, `&wifi` has no children at all, so the `#address-cells = <1>;` / `#size-cells = <0>;` pair at lines 296-297 is dead — it was only there to let dtc address the two `band@N` nodes. No other `mediatek/dts/*.dts` carries `#address-cells` inside `&wifi` without `band@` children, and the sibling [`mt7981b-dlink-aquila-pro-ai-m30-a1.dts:268-273`](``https://github.com/openwrt/openwrt/blob/6a967c55098705e7828dba8bc99abc1af0d05686/target/linux/mediatek/dts/mt7981b-dlink-aquila-pro-ai-m30-a1.dts#L268-L273``) has a bare `&wifi` node. It won't break the build (OpenWrt passes [`-Wno-avoid_unnecessary_addr_size`](https://github.com/openwrt/openwrt/blob/6a967c55098705e7828dba8bc99abc1af0d05686/include/image.mk#L201)), but it's leftover from this commit's removal and worth dropping in the same patch. The stray double blank line at 301-302 could go with it.

No suggestion block, since the lines to delete are outside the hunk this comment is anchored to.

---

openwrt-ai (commented):
Reviewed 1 new commit (`ed5bdd6` → `6a967c5`). The `Fixes:` trailer is now in the project's `<12-char sha> ("<subject>")` form and points at a real commit, and the quoting / readability issues in both shell hooks are resolved.

## Commit checks

- `6a967c5` "mediatek: filogic: fix support for AQUILA PRO AI M60 sold in multi packs" — the message still omits two things the diff does. (1) It drops `label-mac-device = &gmac0;` without replacing it, so `system.label_macaddr` disappears from `/etc/board.json` (nothing in `mediatek_setup_macs()` sets `label_mac` for this board). That's a user-visible behaviour change and should either be stated, or restored — see the [open thread on the aliases hunk](https://github.com/openwrt/openwrt/pull/24484#discussion_r3684288665). (2) The multi-pack detection heuristic itself — a `'-'` (0x2d) at offset `0x4d` of `Odm` meaning the MAC lives at `0x83` rather than `0x81` — is the core of the fix and is documented only in a code comment. A sentence in the message would make the change reviewable from `git log` alone.

---

target/linux/mediatek/dts/mt7986a-dlink-aquila-pro-ai-m60-a1.dts
nicefile: listed line dropped from the .dts

nicefile:
title changed to follow hardware name from original PR
`mediatek: filogic: fix support for AQUILA PRO AI M60 sold in multi packs`
to
`mediatek: filogic: fix support for D-Link AQUILA PRO AI M60 sold in multi packs`

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
```diff
@@ -121,6 +121,18 @@ case "$board" in
 		addr=$(mtd_get_mac_binary "Odm" 0x81)
 		[ "$PHYNBR" = "1" ] && macaddr_add $addr 3 > /sys${DEVPATH}/macaddress
 		;;
+	dlink,aquila-pro-ai-m60-a1)
```
openwrt-ai, line 124: nit: the sibling [`dlink,aquila-pro-ai-m30-a1` case just above (lines 120-123)](``https://github.com/openwrt/openwrt/blob/b4c6658d6b9bcc65d2a0871fe196adc55d48fcda/target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac#L120-L123)`` reads the same `Odm` partition at the same fixed `0x81`, and [`mt7981b-dlink-aquila-pro-ai-m30-a1.dts:210`](``https://github.com/openwrt/openwrt/blob/b4c6658d6b9bcc65d2a0871fe196adc55d48fcda/target/linux/mediatek/dts/mt7981b-dlink-aquila-pro-ai-m30-a1.dts#L210``) carries the identical `macaddr_odm: macaddr@81` layout this PR removes from the M60. The AQUILA PRO AI M30 is likewise sold as M30-2/M30-3 multi-packs, so if the `'-'`-at-`0x4d` shift is a property of the Odm layout rather than of the M60 specifically, the M30 has the same bug and is left unfixed here.

Is the M30 known to be unaffected (different Odm layout / model string length), or is it simply out of scope for this PR because you only have M60 hardware to test on? A sentence either way in the commit message would save the next person from re-deriving it.

---

openwrt-ai (commented):
Reviewed 1 new commit (`6a967c5` → `b4c6658`). The only code change since the last review is the `&wifi` cleanup — `#address-cells`/`#size-cells` and the stray blank lines are gone, which resolves the previous nit; the node now matches the bare `&wifi` of the M30 sibling. The `Fixes: b3ce08e0b6fa ("mediatek: filogic: Add support for D-Link AQUILA PRO AI M60")` trailer verifies against a real commit with that exact subject. CI on `b4c6658` is green.

## Commit checks

- `b4c6658` "mediatek: filogic: fix support for D-Link AQUILA PRO AI M60 sold in multi packs" — two gaps from the previous round are still open in the message. (1) It does not mention that `label-mac-device = &gmac0;` is dropped. Since no `dlink,aquila-pro-ai-m60-a1)` case was added to [`mediatek_setup_macs()` in `board.d/02_network`](``https://github.com/openwrt/openwrt/blob/b4c6658d6b9bcc65d2a0871fe196adc55d48fcda/target/linux/mediatek/filogic/base-files/etc/board.d/02_network#L254),`` `label_mac` stays empty and [`ucidef_set_label_macaddr`](https://github.com/openwrt/openwrt/blob/b4c6658d6b9bcc65d2a0871fe196adc55d48fcda/target/linux/mediatek/filogic/base-files/etc/board.d/02_network#L354) is never called, so `system.label_macaddr` disappears from `/etc/board.json`. Your reply on the [aliases thread](https://github.com/openwrt/openwrt/pull/24484#discussion_r3684601671) explains why the DT alias no longer carries a useful value, which is fair — but that argues for moving the label MAC to `02_network` (it is `Odm+1` per your own table), not for dropping it. Either way, the removal is a user-visible change and deserves a line in the message. (2) The detection heuristic — a `'-'` (0x2d) at offset `0x4d` of `Odm` meaning the MAC lives at `0x83` instead of `0x81` — is the substance of the fix and is documented only in a code comment; a sentence in the message would make the change reviewable from `git log` alone.

---

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
nicefile: D-Link M30 is out of scope for this PR. No hw to test on.

1f604:
Same issue is unfixed on the M30.

nicefile:
> Same issue is unfixed on the M30.

This can be adapted to M30 as well. I can do it if I find tester and this PR is accepted.

nicefile:
@1f604 more complex solution that also include M30 and its's forms is here #24967
There's no point to keep this up anymore

PR #24485, opened 2026-07-30
realtek: refresh kernel patches

plappermaul:
Aovid build warnings because of wrong offsets in kernel patches.

### DIFF
--- a/target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
+++ b/target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
@@ -21,11 +21,10 @@ Signed-off-by: Markus Stockhausen <markus.stockhausen@gmx.de>
  load-$(CONFIG_MIPS_GENERIC)	+= 0xffffffff80100000
 --- a/arch/mips/generic/board-realtek.c
 +++ b/arch/mips/generic/board-realtek.c
-@@ -10,6 +10,344 @@
- 
+@@ -11,6 +11,344 @@
  #include <asm/fw/fw.h>
  #include <asm/machine.h>
-+
+ 
 +#include <mach-rtl-otto.h>
 +
 +#define RTL_SOC_BASE			((volatile void *) 0xB8000000)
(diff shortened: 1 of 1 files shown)


jonasjelonek (approved):

namiltd (approved):

target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
```diff
@@ -375,7 +375,7 @@ Signed-off-by: Markus Stockhausen <markus.stockhausen@gmx.de>
```
openwrt-ai, line 375: nit: pre-existing, not introduced here, but since this patch is being touched anyway — it still adds two whitespace-only lines to the kernel source:

```
p.patch:84: trailing whitespace.
p.patch:373: trailing whitespace.
warning: 2 lines add whitespace errors.
```

Line 373 is the bare `+\t` just above `+\trealtek_prom_init();` (two lines above this one), and line 84 is the equivalent inside the `realtek_prom_init()` body. `checkpatch.pl` flags both as `ERROR: trailing whitespace`. Dropping them changes the hunk line counts, so it needs a re-refresh rather than a hand edit — feel free to leave it for a separate cleanup if you'd rather keep this commit purely mechanical.

---

openwrt-ai (commented):
## Commit checks

- 3a71562 "realtek: refresh kernel patches" — nit: typo in the commit body, `Aovid` → `Avoid` (same typo in the PR description). Also nit: the subject/body say "patches" (plural) but only `300-02-enhance-realtek-board-setup.patch` is touched; a singular wording would match the diff more closely.

---

Verification of the refresh itself (against `gregkh/linux` v6.18.39, `f89c296854b755a66657065c35b05406fc18264d`, which matches `KERNEL_PATCHVER:=6.18` in `target/linux/realtek/Makefile`):

- The old patch reproduced exactly the warnings the commit message describes:
  ```
  Hunk #2 succeeded at 392 (offset -1 lines).
  Hunk #3 succeeded at 409 (offset -1 lines).
  ```
  The refreshed version applies with no offset/fuzz messages.
- Hunk headers are now internally consistent: `@@ -11,6 +11,344 @@` (Δ+338) → `54+338 = 392` → `+2` → `69+340 = 409`. Old-side ranges `11..16`, `54..59`, `69..79` match the upstream file verbatim.
- The resulting kernel source is byte-identical to before: the blank line that moved from context to added-line (and back) around `realtek_add_initrd()` cancels out, so this is a pure metadata refresh with no functional change.
- Swept every other patch in `target/linux/realtek/patches-6.18/` for the same class of new-side offset skew — `300-02` was the only one affected, so nothing else is left un-refreshed. `arch/mips/generic/board-realtek.c` and `arch/mips/generic/Platform` are touched by no other realtek or generic patch, so hunk ordering isn't a factor.

---

plappermaul:

PR #24486, opened 2026-07-30
generic: backport net-dsa-realtek use devm_mutex_init for locks

namiltd:
Backport four upstream commits that convert mutex_init() calls to devm_mutex_init() in the Realtek DSA driver family. With CONFIG_DEBUG_MUTEXES enabled, mutex_destroy() must be called before a mutex is discarded; using the devm variant handles this cleanup automatically and avoids leaking a warning on driver removal/failure.

Added patches (applied to both backport-6.12 and backport-6.18):
- 944-01: net: dsa: realtek: rtl8365mb: use devm_mutex_init for mib_lock
- 944-02: net: dsa: realtek: use devm_mutex_init for regmap lock
- 944-03: net: dsa: realtek: use devm_mutex_init for vlan_lock
- 944-04: net: dsa: realtek: use devm_mutex_init for l2_lock

### DIFF
--- a/target/linux/generic/backport-6.12/944-01-v7.2-net-dsa-realtek-rtl8365mb-use-devm_mutex_init-for-mib_lock.patch
+++ b/target/linux/generic/backport-6.12/944-01-v7.2-net-dsa-realtek-rtl8365mb-use-devm_mutex_init-for-mib_lock.patch
@@ -1,0 +1,70 @@
+From 70fd0cf29bc47882c1cb11ad4fb2881ac2c1e640 Mon Sep 17 00:00:00 2001
+From: Luiz Angelo Daros de Luca <luizluca@gmail.com>
+Date: Sun, 26 Jul 2026 22:56:08 -0300
+Subject: net: dsa: realtek: rtl8365mb: use devm_mutex_init for mib_lock
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+With CONFIG_DEBUG_MUTEXES enabled, mutex_destroy() needs to be called
+before the lock is discarded. Use devm_mutex_init() instead so the
+cleanup is handled automatically.
+
+Fixes: 4af2950c50c86 ("net: dsa: realtek-smi: add rtl8365mb subdriver for RTL8365MB-VC")
+Reviewed-by: Mieczyslaw Nalewaj <namiltd@yahoo.com>
+Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>
+Reviewed-by: Linus Walleij <linusw@kernel.org>
+Reviewed-by: Alvin Šipraga <alvin.sipraga@analog.com>
+Link: https://patch.msgid.link/20260726-realtek_mutext-v2-1-5d62ba998791@gmail.com
+Signed-off-by: Jakub Kicinski <kuba@kernel.org>
+---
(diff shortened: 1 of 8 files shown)


openwrt-ai (commented):

Verification notes: applied `backport-6.12/94*` and then the four new `944-*` patches on top of a clean `v6.12.98` tree, and the same for `backport-6.18/94*` on `v6.18.39` (the versions pinned in `target/linux/generic/kernel-6.12` / `kernel-6.18`) — all eight apply with no fuzz and no offset. Hunk headers, diffstats and line offsets are internally consistent, and 944-01 is correctly re-based per series (`@@ -2570` vs `@@ -2571`) while 944-02/03/04 are identical in both series, matching the identical `rtl83xx.c` in both kernels. After the series no `mutex_init()` remains anywhere under `drivers/net/dsa/realtek/`, `int ret` is already in scope in `rtl83xx_probe()`, `rtl8365mb_stats_setup()` has no other callers, the `out_teardown_irq` label exists, and `devm_mutex_init()` is available in both `include/linux/mutex.h`. The `Fixes:` tags line up with the in-tree prerequisites (`9da2c8672f771` = `942-06`, `336e3e4a1ab37` = `942-07`). Upstream SHAs could not be checked — only `openwrt/openwrt` is reachable in this session.

---

openwrt-ai (commented):

The update since my last review (`ba0bee73`) is a rebase only — `git diff ba0bee73..e34a681e -- target/linux/generic/` is empty, so the eight added patch files are byte-identical to what I reviewed before; only the base moved (new merge-base `baeee779`, which does not touch `target/linux/generic/`). Kernel pins are unchanged (`kernel-6.12` = `.98`, `kernel-6.18` = `.39`).

Re-verified against the reference kernel trees available this run (`gregkh/linux` at `v6.12.100` and `v6.18.39`): applied the full in-tree `94x` realtek series (`940-01` … `943-02`) plus the four new `944-*` patches with `patch -p1 -F0` on both trees — all 18 apply per series with zero fuzz and zero offset. Post-series state checks out: no `mutex_init()` remains anywhere under `drivers/net/dsa/realtek/`, `rtl83xx_probe()` returns `ERR_PTR(ret)` on each of the three new failure paths with `int ret` already in scope and `priv` allocated via `devm_kzalloc(dev, …)` beforehand, `rtl8365mb_stats_setup()` has no other callers so the `void` → `int` signature change is complete, the `out_teardown_irq` label exists and is the correct unwind point at that stage of `rtl8365mb_setup()`, and `devm_mutex_init()` is present in `include/linux/mutex.h` in both series (6.18 additionally marks it `__must_check`, which the new call sites satisfy).

Note on ordering: the `944-*` (v7.2) patches sort after the in-tree `943-*` (v7.3) SGMII/HSGMII patches, i.e. not in upstream commit order, so `944-01`'s hunk headers (`@@ -2570` / `@@ -2571`, `@@ -3174` / `@@ -3175`) are rebased relative to the upstream commit rather than verbatim. That is expected for OpenWrt's quilt series and applies cleanly in both series; flagging only so it is a conscious choice.

Upstream SHAs (`70fd0cf29bc4`, `050e07f8765d`, `a95f3e9b8985`, `442ecdc83d00`) and the `Fixes:` tags still could not be checked against a kernel remote — only `openwrt/openwrt` is reachable from this session, and the shallow reference clones carry no history.

---

neheb (approved):

jonasjelonek:

PR #24488, opened 2026-07-30
ubox: log: check file mountpoint availability

### LINKED ISSUE #24475: log: avoid creating log directories when the target mountpoint is unavailable
### Description

When persistent logging is configured to use a file located on an external filesystem, restarting the `log` service while the target mountpoint is unavailable causes the init script to create the configured log directory in the overlay filesystem.

For example:

```
config system
        option log_file '/data/logs/system.log'
```

If `/data` is not mounted and the following command is executed:

```
/etc/init.d/log restart
```

the script creates:

```
/data/logs
```

in the overlay filesystem.

As a consequence, once the external filesystem is mounted again, two different directory trees may exist under the same mountpoint.

### Expected behavior

If the configured mountpoint is unavailable, the init script should simply skip creation of the persistent log backend instead of creating directories in the overlay filesystem.

### Possible cause

The script already checks whether the target mountpoint is available, but only during boot:

```sh
[ "$_BOOT" = "1" ] &&
    [ "$mountpoint" ] &&
    ! grep -q ".* $mountpoint " /proc/mounts &&
    return 0
```

During normal service operations (`start` or `restart`), this protection is bypassed, allowing `mkdir -p` to create directories in the overlay filesystem.

### Proposed fix

Remove the `_BOOT` condition so the mountpoint check is always performed before creating the log directory.

```diff
- [ "$_BOOT" = "1" ] &&
-     [ "$mountpoint" ] &&
-     ! grep -q ".* $mountpoint " /proc/mounts &&
-     return 0
+ [ "$mountpoint" ] &&
+     ! grep -q ".* $mountpoint " /proc/mounts &&
+     return 0
```

### Validation

The proposed change was tested on OpenWrt 25.12.5.

With the patch applied:

- when the target filesystem is mounted, persistent logging works normally;
- when the target filesystem is unavailable, no directories are created in the overlay filesystem;
- once the filesystem becomes available again, the persistent logging backend can be started normally.

This issue is limited to preventing unintended directory creation on the overlay filesystem.

### OpenWrt version

r33051-f5dae5ece4

### OpenWrt release

25.12.5

### OpenWrt target/subtarget

mvebu/cortexa9

### Device

Linksys WRT1900ACS

### Image kind

Official downloaded image

### Steps to reproduce

1. Configure persistent logging to an external filesystem, for example:

```uci
config system
        option log_file '/data/logs/system.log'
```

2. Ensure that `/data` is **not** mounted.

3. Restart the log service:

```sh
/etc/init.d/log restart
```

4. Check the filesystem:

```sh
find /data
```

### Actual behavior

The init script creates:

```
/data/logs
```

in the overlay filesystem, even though the configured mountpoint is unavailable.

### Expected behavior

If the target mountpoint is unavailable, the init script should not create directories nor start the persistent file logging backend.

### Actual behaviour

When the configured mountpoint is unavailable, restarting the `log` service creates the configured log directory in the overlay filesystem.

For example, with:

```uci
option log_file '/data/logs/system.log'
```

executing:

```sh
/etc/init.d/log restart
```

creates:

```
/data/logs
```

in the overlay filesystem, even though `/data` is not mounted.

If the external filesystem is mounted later, the overlay directory becomes hidden by the mount, leaving an unnecessary directory in the writable flash filesystem.

### Expected behaviour

If the configured log file is located on a mountpoint that is unavailable, the `log` init script should not create the log directory or start the persistent file logging backend.

The service should simply skip the file logging backend until the target mountpoint becomes available.

### Additional information

- OpenWrt version: 25.12.5
- Hardware: Linksys WRT1900ACS
- Filesystem: ext4 on an external USB SSD mounted at `/data`
- Configured log file: /data/logs/system.log

The proposed change only extends an existing mountpoint availability check from boot-time to all service starts.

The change was validated with the following scenarios:

- Target mountpoint available: persistent logging works normally.
- Target mountpoint unavailable: no directories are created in the overlay filesystem.
- The persistent logging backend can be started successfully after the target filesystem becomes available by running:

`/etc/init.d/log start`

### Diffconfig

```text

```

### Terms

marcelolaia:
The file logging backend currently checks whether the configured
mountpoint is available only during boot.

When the service is started or restarted later while the mountpoint is
unavailable, the script may create the configured log directory on the
overlay filesystem and write logs to internal flash.

Always check mountpoint availability before creating the log directory and starting the file logger.
This preserves the current boot behavior while preventing unintended
directory creation when external storage is unavailable.

Tested with the log file located on external storage:

* logging works normally while the filesystem is mounted;
* no log directory is created on the overlay while it is unmounted;
* logging resumes after remounting the filesystem and starting the
  service again.

Closes #24475

### DIFF
--- a/package/system/ubox/files/log.init
+++ b/package/system/ubox/files/log.init
@@ -54,8 +54,7 @@ start_service_file()
 
 	local mountpoint="$(procd_get_mountpoints "${log_file}")"
 
-	[ "$_BOOT" = "1" ] &&
-		[ "$mountpoint" ] &&
+	[ "$mountpoint" ] &&
 		! grep -q ".* $mountpoint " /proc/mounts &&
 		return 0
 
@@ -109,8 +108,4 @@ start_service()
 	config_foreach validate_log_daemon system start_service_daemon
 	config_foreach validate_log_section system start_service_file
 	config_foreach validate_log_section system start_service_remote
-}
-
-boot() {
-	_BOOT=1 start
 }
--- a/package/system/ubox/Makefile
+++ b/package/system/ubox/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=ubox
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL=$(PROJECT_GIT)/project/ubox.git


package/system/ubox/files/log.init
```diff
@@ -54,8 +54,7 @@ start_service_file()
 
 	local mountpoint="$(procd_get_mountpoints "${log_file}")"
 
-	[ "$_BOOT" = "1" ] &&
-		[ "$mountpoint" ] &&
+	[ "$mountpoint" ] &&
```
openwrt-ai, line 57: nit: this was the only reader of `_BOOT` in the tree (`grep -rn _BOOT package/` matches nothing else), so [`boot() { _BOOT=1 start; }` at log.init:114-116](https://github.com/openwrt/openwrt/blob/e1530780bf0dca62801d844ef39712548594b7dd/package/system/ubox/files/log.init#L114-L116) is now dead code — it does exactly what the default [`boot()` in rc.common](https://github.com/openwrt/openwrt/blob/e1530780bf0dca62801d844ef39712548594b7dd/package/base-files/files/etc/rc.common#L30-L32) already does. Worth dropping the override in the same commit so a future reader doesn't assume boot is still special-cased.

---

package/system/ubox/files/log.init
```diff
@@ -54,8 +54,7 @@ start_service_file()
...
 		! grep -q ".* $mountpoint " /proc/mounts &&
 		return 0
```
openwrt-ai, line 59: This changes the behaviour of an installed file (`./files/log.init` is installed by `Package/logd/install`), but [`PKG_RELEASE` in package/system/ubox/Makefile:4](https://github.com/openwrt/openwrt/blob/e1530780bf0dca62801d844ef39712548594b7dd/package/system/ubox/Makefile#L4) is left at `1`, so existing installs won't pull the fix. It needs a bump in the same commit (the CI formality bot flagged this too). Note the precedent for this exact file: [`987e10af0756` ("logd: create log directory for log_file")](https://github.com/openwrt/openwrt/commit/987e10af07599c7f91a4b92a48d5aa62f63905d3) was merged with a `[PKG_RELEASE increase]` amendment.

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the diff, and the change itself is consistent with the mount-trigger design added in [`88baf6ce2c88`](https://github.com/openwrt/openwrt/commit/88baf6ce2c880149bd1c1d8ddacc7bc61dd4f18b) / [`6ef0a35bb1a4`](https://github.com/openwrt/openwrt/commit/6ef0a35bb1a48f7c7daebaf8eeca877feffb47c7): `service_triggers()` still registers `procd_add_action_mount_trigger start "${log_file}"`, and a manual mount followed by a service start is still covered because the `/proc/mounts` check is evaluated at every start. Two inline notes.

---

package/system/ubox/files/log.init
marcelolaia: Fixed by removing the redundant boot() override in the same commit.

package/system/ubox/files/log.init
marcelolaia: Fixed by bumping PKG_RELEASE to 2.

package/system/ubox/files/log.init
```diff
@@ -110,7 +109,3 @@ start_service()
 	config_foreach validate_log_section system start_service_file
 	config_foreach validate_log_section system start_service_remote
 }
-
-boot() {
-	_BOOT=1 start
-}
```
openwrt-ai, line 116: nit: dropping this override is correct — `_BOOT` has no other reader left and the default [`boot() { start "$@"; }` in rc.common:30-32](https://github.com/openwrt/openwrt/blob/1a6ac48ba16325c9d9b201ce9e04afcdb1c6b3c2/package/base-files/files/etc/rc.common#L30-L32) does the same thing — but the commit message doesn't mention it (nor the `PKG_RELEASE` bump). A reader of `git log` sees only "always verify mountpoint availability" and then finds an unexplained function removal. Worth one extra line in the body, e.g. "Drop the now-redundant `boot()` override, since `_BOOT` is no longer used and rc.common's default `boot()` already calls `start`."

---

openwrt-ai (commented):
Re-reviewed the changes since `e1530780` (rebased onto current main, so only `package/system/ubox/` was compared). The two points from the previous review are addressed: `PKG_RELEASE` is bumped to `2`, and the `boot()` override is gone. The mountpoint check now runs on every start, `service_triggers()` still registers the mount trigger, and CI (FormalityCheck: Makefiles / Code Patches / Git & Commits) is green on the head commit. One documentation nit inline; no functional issues found.

---

package/system/ubox/files/log.init
marcelolaia: Thanks. I've updated the commit message to document the removal of the redundant boot() override and the PKG_RELEASE bump.

openwrt-ai (commented):
Re-reviewed against `1a6ac48b`. `git diff 1a6ac48b..01897a79` is empty — the tree is identical to what I reviewed last time, and the only change is the amended commit message. It now documents all three hunks:

- the mountpoint check moving out of the boot-only path,
- the `boot()` removal — verified: `grep -rn _BOOT package/` has no readers left outside unrelated `U_BOOT_*` matches in uboot-d1 patches, and [`boot() { start "$@"; }` in rc.common:30-32](https://github.com/openwrt/openwrt/blob/01897a791cbb5dd0fa1685fe286932456af849ee/package/base-files/files/etc/rc.common#L30-L32) is behaviourally identical to the removed override,
- the `PKG_RELEASE` bump to `2`.

All three previous nits are resolved and the FormalityCheck runs (Makefiles / Code Patches / Git & Commits) are green on the head commit. No new issues found.

---

Issue #24454, opened 2026-07-28
Luci PPPoSSH config timeout monitoring crashing pppd with: unrecognized option 'precompiled-active-filter'


Edmondena:
### Describe the bug

I'm successfully running the Luci PPPoSSH protocol beeing linked using an openssh-client link.
Everything works fine, but the highly important monitoring of a inactive or crashed connection does not. 
Whenn I change the default value of "Inactivity timeout" from "0/disabled" to "1/enabled" seconds or any other value, to shut down the link in case of of a connection loss, the interface page at this PPPoSSH device shows "Error: Unknown error (OPTION_ERROR)".
The link will not be enabled anymore and stays dead forever.

In the logs this problem shows up as follows:

```
Tue Jul 28 09:58:20 2026 daemon.notice netifd: pvpn (20387): Could not resolve maxwrt.spdns.de
Tue Jul 28 09:58:20 2026 daemon.notice netifd: Interface 'pvpn' is now down
Tue Jul 28 09:58:20 2026 daemon.notice netifd: Interface 'pvpn' is setting up now
...
Tue Jul 28 09:58:23 2026 daemon.notice netifd: pvpn (20830): /usr/sbin/pppd: unrecognized option 'precompiled-active-filter'
Tue Jul 28 09:58:23 2026 daemon.err pppd[20830]: unrecognized option 'precompiled-active-filter'
Tue Jul 28 09:58:23 2026 daemon.notice netifd: pvpn (20830): ppp v2.5.1
Tue Jul 28 09:58:23 2026 daemon.notice netifd: pvpn (20830): Copyright (C) 1999-2022 Paul Mackerras, and others. All rights reserved.
Tue Jul 28 09:58:23 2026 daemon.notice netifd: pvpn (20830):
Tue Jul 28 09:58:23 2026 daemon.notice netifd: pvpn (20830): License BSD: The 3 clause BSD license <https://opensource.org/licenses/BSD-3-Clause>
Tue Jul 28 09:58:23 2026 daemon.notice netifd: pvpn (20830): This is free software: you are free to change and redistribute it. 
Tue Jul 28 09:58:23 2026 daemon.notice netifd: pvpn (20830): There is NO WARRANTY, to the extent permitted by law.
Tue Jul 28 09:58:23 2026 daemon.notice netifd: pvpn (20830):
Tue Jul 28 09:58:23 2026 daemon.notice netifd: pvpn (20830): Report Bugs:
[...]

PR #23792, opened 2026-06-14
ppp: convert protocol handlers from shell to ucode

strrs:
Convert the ppp, pppoe, pppoa, and pptp protocol handlers from shell to ucode, using the proto handler framework introduced in commit aaf5b194b15d ("proto-ucode: add ucode protocol handler infrastructure").

Follows the same approach as #22751 (dhcp/dhcpv6).

ppp.sh is retained as a library: modemmanager.sh, pppossh.sh and 3g.sh source it for its helper functions. Only its add_protocol registration block is removed. netifd registers shell handlers before ucode handlers and silently drops duplicate protocol names, so leaving the block in place would mask the ucode handlers entirely (details in the commit message).

The ucode versions replace shell subprocesses with native functions:

| Operation | Shell (fork+exec) | Ucode (native) |
|---|---|---|
| PPTP server resolution | `resolveip` | ucode-mod-resolv `query()` |
| PPTP literal-IP detection | `resolveip` (getaddrinfo) | `iptoarr()` |
| Plugin detection | shell test | `glob()` |
| Retry delay on resolve failure | `sleep` | `sleep()` |
| UCI reads | `proto_config_add_*` (netifd C + shell) | config callback |

All four handlers produce identical pppd command-line invocations to their shell predecessors, including boolean option normalization (persist, keepalive_adaptive, authfail, sourcefilter, delegate, norelease, peerdns accept the same spellings netifd's proto_config_add_boolean accepted). The ppp-up/ppp-down callback scripts, UCI configuration, and LuCI integration are unchanged.

There is one deliberate behavior change, detailed in the commit message: PPTP host-dependency tracking now requests A records only. pptp.so is IPv4-only (gethostbyname/AF_INET), so the AAAA results resolveip also returned were getting registered as host dependencies the plugin could never use. Literal IPv4 server addresses are used directly without a DNS round-trip, and A records are collected across CNAME chains.

Verified on OpenWrt snapshot (TP-Link BE450), shell vs ucode handler:

- pppoe: full session establish and teardown on a real WAN. `ps ww | grep pppd` produces a byte-identical pppd command line for the shell and ucode handlers.
- boolean normalization: `option persist 'true'` produces `persist maxfail 0`, and `option persist '0'` produces `nopersist maxfail 1`, identical to the shell handler.
- teardown error reporting: after `ifdown`, the `ifstatus` errors field is populated (`{ "subsystem": "pppoe", "code": "USER_REQUEST" }`) identically for both handlers.
[...]

PR #24489, opened 2026-07-30
ppp: quote the string arguments passed to pppd

strrs:
ppp.sh expands two free-form values unquoted when it builds the pppd argument list, so the shell splits a value containing whitespace and every word after the first becomes a stray positional argument. pppd prints its usage and exits with EXIT_OPTION_ERROR when the word matches no option. The protocol handler treats that exit code as terminal and blocks the restart, so the interface stays down until the configuration changes. A bare number does not fail at all, because pppd's wildcard matcher takes it as the serial port speed.

reqprefix is declared with proto_config_add_string and carries no validation pattern. ip6table is a core interface attribute that netifd passes through as a string, and when system_resolve_rt_table() cannot parse it netifd only logs a debug message, so the raw value still reaches this argument list.

Quote both so the value survives as one element. Values without whitespace produce the same argument list as before, so no working configuration changes.

With `option reqprefix '/60 /56'` and `option ipv6 'auto'`, before:

```
pppd[6105]: unrecognized option '/56'
netifd: Interface 'pppotest' is disabled
ifstatus: "errors": [ { "subsystem": "pppoe", "code": "OPTION_ERROR" } ]
```

Same failure as #24454. `option ip6table 'my table'` hands pppd a stray `table` the same way.

A bare number takes the other path: `option reqprefix '/60 9600'` produces no error at all, because pppd's wildcard matcher takes 9600 as the serial port speed.

The `set` option takes one argument, so the shell handing it two is what pppd rejects. Quoting the expansion keeps the value together:

```
before:  set   REQPREFIX=/60   /56      three arguments, /56 arrives on its own
after:   set   REQPREFIX=/60 /56        two arguments, the space is inside the value
```

Quoting on its own would drop the extra reqprefix values without a trace, because ppp6-up expands REQPREFIX unquoted as well and forwards only the first word. The handler logs what it had to drop:

```
user.warn ppp: pppotest: ignoring extra reqprefix values
```

That counts words rather than testing for whitespace, so `option reqprefix '/60 '` stays silent: ppp6-up word splits it back to `/60` and drops nothing. The count runs in a subshell so ppp_generic_setup keeps its own argument list, which it still needs for the per-protocol options.

I read pppd's full argument list off the running process before and after and compared it byte for byte. Only a value carrying a second word behaves differently. Adding the warning changes nothing: the argument list is identical with and without it.

This does not add support for requesting more than one prefix. ppp6-up still forwards only the first word, so odhcp6c gets a single -P:

```
odhcp6c -s /lib/netifd/dhcpv6.script -P60:a62e8328 -c0004... -t120 pppoe-pppotest
```

The patch only stops the argument list from being malformed.

The ucode conversion in #23792 is not affected: ppp.uc builds the argument list as a list, so it cannot word split.

openwrt-25.12 and openwrt-24.10 carry the same ppp.sh, so the fix applies there unchanged if backported.

Tested on: Asus RT-N18U, OpenWrt SNAPSHOT r35591-d0110a25ed, pppd 2.5.3.

Fixes: d5221d5a419c ("ppp: honor ip6table for IPv6 PPP interfaces")
Fixes: 7079e456ade5 ("ppp: add reqprefix norelease ac_mac")

### DIFF
--- a/package/network/services/ppp/files/ppp.sh
+++ b/package/network/services/ppp/files/ppp.sh
@@ -108,6 +108,9 @@ ppp_generic_setup() {
 		norelease=""
 	fi
 
+	[ "$(set -- $reqprefix; echo $#)" -gt 1 ] && \
+		logger -p warn -t ppp "$config: ignoring extra reqprefix values"
+
 	if [ "${demand:-0}" -gt 0 ]; then
 		demand="precompiled-active-filter /etc/ppp/filter demand idle $demand"
 	else
@@ -153,9 +156,9 @@ ppp_generic_setup() {
 		${lcp_failure:+lcp-echo-interval $lcp_interval lcp-echo-failure $lcp_failure $lcp_adaptive} \
 		${ipv6:++ipv6} \
 		${autoipv6:+set AUTOIPV6=1} \
-		${reqprefix:+set REQPREFIX=$reqprefix} \
+		${reqprefix:+set "REQPREFIX=$reqprefix"} \
 		${norelease:+set NORELEASE=1} \
-		${ip6table:+set IP6TABLE=$ip6table} \
+		${ip6table:+set "IP6TABLE=$ip6table"} \
 		${peerdns:+set PEERDNS=$peerdns} \
 		${sourcefilter:+set NOSOURCEFILTER=1} \
 		${delegate:+set DELEGATE=0} \
--- a/package/network/services/ppp/Makefile
+++ b/package/network/services/ppp/Makefile
@@ -10,7 +10,7 @@ include $(INCLUDE_DIR)/kernel.mk
 
 PKG_NAME:=ppp
 PKG_VERSION:=2.5.3
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://download.samba.org/pub/ppp


package/network/services/ppp/files/ppp.sh
```diff
@@ -108,6 +108,12 @@ ppp_generic_setup() {
 		norelease=""
 	fi
 
+	case "$reqprefix" in
+	*[[:space:]]*)
+		logger -p warn -t ppp "$config: ignoring extra reqprefix values"
+		;;
+	esac
```
openwrt-ai: nit: the two existing `case` statements in this file indent the patterns one level below `case` — [`ppp_exitcode_tostring` at ppp.sh:41-42](https://github.com/openwrt/openwrt/blob/fd086ab6e7359df8e83f3cb29b5dca222d32a91e/package/network/services/ppp/files/ppp.sh#L41-L42) and [`ppp_generic_teardown` at ppp.sh:181-183](https://github.com/openwrt/openwrt/blob/fd086ab6e7359df8e83f3cb29b5dca222d32a91e/package/network/services/ppp/files/ppp.sh#L181-L183). This block puts the pattern at the same indent as `case`.

```suggestion
	case "$reqprefix" in
		*[[:space:]]*)
			logger -p warn -t ppp "$config: ignoring extra reqprefix values"
			;;
	esac
```


---

package/network/services/ppp/files/ppp.sh
```diff
@@ -108,6 +108,12 @@ ppp_generic_setup() {
 		norelease=""
 	fi
 
+	case "$reqprefix" in
+	*[[:space:]]*)
```
openwrt-ai: nit: this matches *any* whitespace, including leading or trailing whitespace where nothing is actually dropped. With `option reqprefix '/60 '` the quoted argument becomes `REQPREFIX=/60 `, and [`json_add_string reqprefix $REQPREFIX` in ppp6-up:29](https://github.com/openwrt/openwrt/blob/fd086ab6e7359df8e83f3cb29b5dca222d32a91e/package/network/services/ppp/files/lib/netifd/ppp6-up#L29) still word-splits to the single value `/60` — so the prefix is honoured, yet the log claims `ignoring extra reqprefix values`. Consider testing for a second word rather than for any whitespace (or trimming the value first), so the warning only fires when something is really discarded.

---

package/network/services/ppp/files/ppp.sh
```diff
@@ -153,7 +159,7 @@ ppp_generic_setup() {
 		${lcp_failure:+lcp-echo-interval $lcp_interval lcp-echo-failure $lcp_failure $lcp_adaptive} \
 		${ipv6:++ipv6} \
 		${autoipv6:+set AUTOIPV6=1} \
-		${reqprefix:+set REQPREFIX=$reqprefix} \
+		${reqprefix:+set "REQPREFIX=$reqprefix"} \
 		${norelease:+set NORELEASE=1} \
 		${ip6table:+set IP6TABLE=$ip6table} \
```
openwrt-ai: `ip6table` is exposed to this handler as a free-form string and reaches the argument list through the same unquoted expansion, so it has the identical failure mode you are fixing for `reqprefix`: netifd declares it as [`IFACE_ATTR_IP6TABLE ... BLOBMSG_TYPE_STRING` in interface.c:104](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/interface.c#L104) and, when `system_resolve_rt_table()` fails, only emits a debug message ([interface.c:1048-1051](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/interface.c#L1048-L1051)) — the raw string still lands here. `option ip6table 'my table'` would therefore hand pppd a stray `table` argument and trigger the same `EXIT_OPTION_ERROR` / blocked restart.

Was leaving this line unquoted deliberate (keeping the diff minimal to the reported bug), or should the `set NAME=$value` expansions in this block be quoted consistently?

---

openwrt-ai (commented):
Reviewed 1 new commit; commit message matches the diff and the `Fixes:` trailer resolves correctly to `7079e456ade5 ("ppp: add reqprefix norelease ac_mac")`, which is indeed where the unquoted `${reqprefix:+set REQPREFIX=$reqprefix}` expansion was introduced. `PKG_RELEASE` bump is present. I confirmed the quoting behaviour holds for POSIX shells (`${var:+set "NAME=$var"}` yields a single argument in dash/bash), and that [ppp6-up:29](https://github.com/openwrt/openwrt/blob/fd086ab6e7359df8e83f3cb29b5dca222d32a91e/package/network/services/ppp/files/lib/netifd/ppp6-up#L29) still forwards only the first word, as the commit message states.

Three inline notes below (two nits plus one question about the adjacent `ip6table` expansion).

---

package/network/services/ppp/files/ppp.sh
strrs: Dropped the case entirely. The check is a word count now, so there is no pattern left to indent.

package/network/services/ppp/files/ppp.sh
strrs: It counts words instead of matching whitespace now, so option reqprefix /60  stays silent. Checked against leading, trailing, doubled and tab separators.

package/network/services/ppp/files/ppp.sh
strrs: I only looked at reqprefix because that is where the report came from. Quoted it too, and since that line has been unquoted since d5221d5a419c the commit now carries a second Fixes: tag.

package/network/services/ppp/files/ppp.sh
```diff
@@ -153,9 +156,9 @@ ppp_generic_setup() {
 		${lcp_failure:+lcp-echo-interval $lcp_interval lcp-echo-failure $lcp_failure $lcp_adaptive} \
 		${ipv6:++ipv6} \
 		${autoipv6:+set AUTOIPV6=1} \
-		${reqprefix:+set REQPREFIX=$reqprefix} \
+		${reqprefix:+set "REQPREFIX=$reqprefix"} \
 		${norelease:+set NORELEASE=1} \
-		${ip6table:+set IP6TABLE=$ip6table} \
+		${ip6table:+set "IP6TABLE=$ip6table"} \
```
openwrt-ai, line 161: nit: `ip6table` now reaches pppd intact, but it is dropped further down the same way `reqprefix` is — [`json_add_string ip6table $IP6TABLE` in ppp6-up:28](https://github.com/openwrt/openwrt/blob/61161204b54c35822626f879d4645f4ba62f06c0/package/network/services/ppp/files/lib/netifd/ppp6-up#L28) expands unquoted, so `json_add_string` gets three arguments and the dynamic dhcpv6 interface silently inherits only the first word. That is the same "extra values disappear without a trace" case that motivated the new `reqprefix` warning, yet `ip6table` gets no warning.

Was leaving it asymmetric deliberate (a bad `ip6table` is already only a debug message in netifd, so there is less to report), or should both warn?

---

openwrt-ai (commented):
Reviewed the force-pushed commit; both points from the previous round are addressed and the commit message matches the diff.

I re-checked the two changes since `fd086ab`:

- The word-count replacement behaves as claimed. In `dash`, `$(set -- $reqprefix; echo $#)` yields `0`/`1`/`2` for ``, `/60`, `/60 /56`, and stays at `1` for `/60 `, `<tab>/60` and other leading/trailing whitespace — so the false positive on trimmed values is gone. It is placed after the `autoipv6` reset, so it cannot fire for a `reqprefix` that is discarded anyway. The command substitution is a subshell, so `ppp_generic_setup`'s own `"$@"` is untouched.
- `logger -p warn -t ppp "$config: ..."` matches the existing in-tree precedent, both in form and in wording ([dhcpv6.sh:116](https://github.com/openwrt/openwrt/blob/61161204b54c35822626f879d4645f4ba62f06c0/package/network/ipv6/odhcp6c/files/dhcpv6.sh#L116) and [dhcp.sh:51](https://github.com/openwrt/openwrt/blob/61161204b54c35822626f879d4645f4ba62f06c0/package/network/config/netifd/files/lib/netifd/proto/dhcp.sh#L51)), so the new `Fixes: d5221d5a419c` trailer resolves correctly.
- Quoting `ip6table` covers the remaining string-typed value in this argument list: netifd declares `ip6table` as [`BLOBMSG_TYPE_STRING` (interface.c:104)](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/interface.c#L104) while `peerdns` is [`BLOBMSG_TYPE_BOOL` (interface.c:94)](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/interface.c#L94), so the remaining unquoted `set NAME=$var` expansions cannot carry whitespace.

One nit inline about the `ip6table` value being dropped further downstream.

---

package/network/services/ppp/files/ppp.sh
strrs: Deliberate. system_resolve_rt_table() splits on whitespace, so no valid ip6table value contains a space. The value is unresolvable to begin with, and netifd reports that itself. `reqprefix '/60 /56'` is the other case: two valid requests where the second disappears silently, which is what the warning is for.

peterwillcn (approved):

openwrt-ai (commented):

The only change since `61161204` is a rebase from `9cd43a0614d9` onto `cdc44e906ace`. `package/network/services/ppp/files/ppp.sh` is byte-identical to the version I reviewed last round, and the commit message is unchanged, so the earlier findings (all resolved) still stand.

Two things I re-verified against the new base:

- `PKG_RELEASE` is correctly re-based, not stale. Main picked up [`f59200bc56 ("ppp: fix precompiled filter option")`](https://github.com/openwrt/openwrt/commit/f59200bc56) in the meantime, which took `PKG_RELEASE` to `2`; this PR now bumps `2` → `3` rather than re-applying the old `1` → `2` hunk.
- The quoting still yields the intended argument list on the rebased file. With `reqprefix='/60 /56'`, `${reqprefix:+set "REQPREFIX=$reqprefix"}` expands to exactly two words (`set`, `REQPREFIX=/60 /56`) in both dash and bash, and `$(set -- $reqprefix; echo $#)` reports `2`, so the warning fires exactly when a second value is dropped.

CI on `99fa9f7` is green so far (`FormalityCheck` all passed, package checks for x86/64 and malta/be passed, target builds still running).

---

PR #24490, opened 2026-07-30
mtd mtdsplit-fit: fix reading fit header from wrong offset

dac922:
The fit header starts at offset_start, so starting reading it from this offset, instead of adding offset_start to the read size.

### DIFF
--- a/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c
+++ b/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c
@@ -300,7 +300,7 @@ mtdsplit_fit_parse(struct mtd_info *mtd,
 	} else {
 		/* Search for rootfs_data after FIT external data */
 		fit = kzalloc(fit_size, GFP_KERNEL);
-		ret = mtd_read(mtd, offset, fit_size + offset_start, &retlen, fit);
+		ret = mtd_read(mtd, offset + offset_start, fit_size, &retlen, fit);
 		if (ret) {
 			pr_err("read error in \"%s\" at offset 0x%llx\n",
 			       mtd->name, (unsigned long long) offset);
--- a/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c
+++ b/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c
@@ -198,7 +198,6 @@ mtdsplit_fit_parse(struct mtd_info *mtd,
 	const char *cmdline_match = NULL;
 	struct fdt_header hdr;
 	size_t hdr_len, retlen;
-	size_t offset;
 	u32 offset_start = 0;
 	size_t fit_offset, fit_size;
 	size_t rootfs_offset, rootfs_size;
@@ -218,11 +217,13 @@ mtdsplit_fit_parse(struct mtd_info *mtd,
 	hdr_len = sizeof(struct fdt_header);
 
 	/* Parse the MTD device & search for the FIT image location */
-	for (offset = 0; offset + offset_start + hdr_len <= mtd->size; offset += mtd->erasesize) {
-		ret = mtd_read(mtd, offset + offset_start, hdr_len, &retlen, (void*) &hdr);
+	for (fit_offset = offset_start;
+	     fit_offset + hdr_len <= mtd->size;
+	     fit_offset += mtd->erasesize) {
+		ret = mtd_read(mtd, fit_offset, hdr_len, &retlen, (void*) &hdr);
 		if (ret) {
 			pr_err("read error in \"%s\" at offset 0x%llx\n",
-			       mtd->name, (unsigned long long) offset);
+			       mtd->name, (unsigned long long) fit_offset);
 			return ret;
 		}
 
@@ -234,7 +235,7 @@ mtdsplit_fit_parse(struct mtd_info *mtd,
 		/* Check the magic - see if this is a FIT image */
 		if (be32_to_cpu(hdr.magic) != OF_DT_HEADER) {
 			pr_debug("no valid FIT image found in \"%s\" at offset %llx\n",
-				 mtd->name, (unsigned long long) offset);
+				 mtd->name, (unsigned long long) fit_offset);
 			continue;
 		}
 
@@ -248,7 +249,6 @@ mtdsplit_fit_parse(struct mtd_info *mtd,
 		return -ENOENT;
 	}
 
-	fit_offset = offset;
 	fit_size = be32_to_cpu(hdr.totalsize);
 
 	if (fit_size == 0) {
@@ -269,7 +269,7 @@ mtdsplit_fit_parse(struct mtd_info *mtd,
 		enum mtdsplit_part_type type;
 
 		/* Search for the rootfs partition after the FIT image */
-		ret = mtd_find_rootfs_from(mtd, fit_offset + fit_size + offset_start, mtd->size,
+		ret = mtd_find_rootfs_from(mtd, fit_offset + fit_size, mtd->size,
 					   &rootfs_offset, &type);
 		if (ret) {
 			pr_info("no rootfs found after FIT image in \"%s\"\n",
@@ -300,10 +300,10 @@ mtdsplit_fit_parse(struct mtd_info *mtd,
 	} else {
 		/* Search for rootfs_data after FIT external data */
 		fit = kzalloc(fit_size, GFP_KERNEL);
-		ret = mtd_read(mtd, offset + offset_start, fit_size, &retlen, fit);
+		ret = mtd_read(mtd, fit_offset, fit_size, &retlen, fit);
 		if (ret) {
 			pr_err("read error in \"%s\" at offset 0x%llx\n",
-			       mtd->name, (unsigned long long) offset);
+			       mtd->name, (unsigned long long) fit_offset);
 			return ret;
 		}
 


namiltd (approved):

target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c
```diff
@@ -300,7 +300,7 @@ mtdsplit_fit_parse(struct mtd_info *mtd,
 	} else {
 		/* Search for rootfs_data after FIT external data */
 		fit = kzalloc(fit_size, GFP_KERNEL);
-		ret = mtd_read(mtd, offset, fit_size + offset_start, &retlen, fit);
+		ret = mtd_read(mtd, offset + offset_start, fit_size, &retlen, fit);
```
openwrt-ai: The fix is right, but it leaves the rest of this branch inconsistent about `offset_start`. After this change `fit` holds the FDT blob that starts at flash offset `fit_offset + offset_start`, so `max_size`, computed from [`img_total = data_size + (img_data - fit)` at mtdsplit_fit.c:326](``https://github.com/openwrt/openwrt/blob/e4e434e1d1e0771b68a847c0cf78db0554b469dc/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c#L326),`` is an offset relative to the FIT start, not relative to `fit_offset`.

But [`parts[0].offset = fit_offset + mtd_roundup_to_eb(max_size, mtd)` at mtdsplit_fit.c:337](https://github.com/openwrt/openwrt/blob/e4e434e1d1e0771b68a847c0cf78db0554b469dc/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c#L337) drops `offset_start`, whereas the sibling branch does account for it at [mtdsplit_fit.c:288](https://github.com/openwrt/openwrt/blob/e4e434e1d1e0771b68a847c0cf78db0554b469dc/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c#L288) (`mtd_roundup_to_eb(fit_size + offset_start, mtd)`). On a device with a non-zero `openwrt,fit-offset` (in-tree today only `mt7629-iptime-a6004mx.dts:233`, `0x800`) the computed rootfs_data start can therefore land one erase block too low. Should this be `mtd_roundup_to_eb(max_size + offset_start, mtd)`?

---

target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c
```diff
@@ -300,7 +300,7 @@ mtdsplit_fit_parse(struct mtd_info *mtd,
 	} else {
 		/* Search for rootfs_data after FIT external data */
 		fit = kzalloc(fit_size, GFP_KERNEL);
```
openwrt-ai, line 302: Pre-existing, but on the line right above the one being fixed: the allocation result is never checked, so an allocation failure hands a NULL buffer straight to `mtd_read()` / `fdt_path_offset()`. The other allocations in this function (`kcalloc` at L282, `kzalloc` at L332) do check.

```suggestion
		fit = kzalloc(fit_size, GFP_KERNEL);
		if (!fit)
			return -ENOMEM;
```

---

target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c
```diff
@@ -300,7 +300,7 @@ mtdsplit_fit_parse(struct mtd_info *mtd,
...
-		ret = mtd_read(mtd, offset, fit_size + offset_start, &retlen, fit);
+		ret = mtd_read(mtd, offset + offset_start, fit_size, &retlen, fit);
 		if (ret) {
 			pr_err("read error in \"%s\" at offset 0x%llx\n",
 			       mtd->name, (unsigned long long) offset);
```
openwrt-ai: Pre-existing, but in the error path this hunk touches: `fit` is leaked on every early return in this branch — the `return ret;` below, plus [L314](``https://github.com/openwrt/openwrt/blob/e4e434e1d1e0771b68a847c0cf78db0554b469dc/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c#L314),`` [L324](https://github.com/openwrt/openwrt/blob/e4e434e1d1e0771b68a847c0cf78db0554b469dc/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c#L324) and [L334](``https://github.com/openwrt/openwrt/blob/e4e434e1d1e0771b68a847c0cf78db0554b469dc/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c#L334).`` Only the success path frees it at [L342](``https://github.com/openwrt/openwrt/blob/e4e434e1d1e0771b68a847c0cf78db0554b469dc/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c#L342),`` so each failed parse leaks `fit_size` bytes. A single `goto out_free` unwind would cover all four.

Also, unlike the header scan at [L229-L232](``https://github.com/openwrt/openwrt/blob/e4e434e1d1e0771b68a847c0cf78db0554b469dc/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c#L229-L232),`` `retlen` is not compared against `fit_size` here, so a short read is parsed as a complete FIT.

---

openwrt-ai (commented):
## Commit checks

- e4e434e "mtd: mtdsplit-fit: fix reading fit header from wrong offset" — the message under-describes the fix in two ways. (1) The subject says "fit header", but the FIT header read at L222 already uses `offset + offset_start` correctly; the read changed here loads the whole FIT blob, so something like "fix reading FIT image from wrong offset" would match the diff. (2) The old call read `fit_size + offset_start` bytes into a `kzalloc(fit_size)` buffer — i.e. an `offset_start`-byte heap buffer overflow whenever `openwrt,fit-offset` is set, not just a wrong start offset. That is the more severe half of what this patch fixes and is worth stating in the commit message.

---

mrkiko:
Mark.

peterwillcn (approved):

target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c
```diff
@@ -269,7 +269,7 @@ mtdsplit_fit_parse(struct mtd_info *mtd,
 		enum mtdsplit_part_type type;
 
 		/* Search for the rootfs partition after the FIT image */
-		ret = mtd_find_rootfs_from(mtd, fit_offset + fit_size + offset_start, mtd->size,
+		ret = mtd_find_rootfs_from(mtd, fit_offset + fit_size, mtd->size,
```
openwrt-ai, line 272: Redefining `fit_offset` to already include `offset_start` silently changes the resulting partition geometry, and two untouched users of it were not updated:

- [`parts[0].offset = fit_offset; parts[0].size = mtd_roundup_to_eb(fit_size + offset_start, mtd);` at mtdsplit_fit.c:287-288](``https://github.com/openwrt/openwrt/blob/b549a1371cfa4276d937f3112a98b52246accc92/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c#L286-L288``) — before this series `fit_offset` was the eraseblock-aligned start *before* the vendor header, so `offset = 0` and the kernel partition spanned `offset_start + fit_size`. Now the offset is `offset_start` while the size still adds `offset_start`, so the same `offset_start` bytes are counted twice and the partition ends `offset_start` further out than before.
- [`parts[0].offset = fit_offset + mtd_roundup_to_eb(max_size, mtd);` at mtdsplit_fit.c:337](https://github.com/openwrt/openwrt/blob/b549a1371cfa4276d937f3112a98b52246accc92/target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c#L337) — the `rootfs_data` split partition is shifted by `offset_start` as well.

For the only in-tree user, [`openwrt,fit-offset = <0x800>` in mt7629-iptime-a6004mx.dts:233](https://github.com/openwrt/openwrt/blob/b549a1371cfa4276d937f3112a98b52246accc92/target/linux/mediatek/dts/mt7629-iptime-a6004mx.dts#L233), both partitions now start 0x800 past an eraseblock boundary, which makes MTD drop `MTD_WRITEABLE` and log "doesn't start on an erase/write block boundary -- force read-only" ([mtdpart.c:182-191](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/drivers/mtd/mtdpart.c#L182-L191), v6.12.100) — for `rootfs_data` that is fatal.

If the intent is a pure rename, keep the partition base aligned (`parts[0].offset = fit_offset - offset_start;` and `parts[0].offset = fit_offset - offset_start + mtd_roundup_to_eb(...)`); if the geometry change is deliberate, drop the now-double-counted `+ offset_start` from the size and please describe the behaviour change in the commit message.

---

target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c
```diff
@@ -300,10 +300,10 @@ mtdsplit_fit_parse(struct mtd_info *mtd,
 	} else {
 		/* Search for rootfs_data after FIT external data */
 		fit = kzalloc(fit_size, GFP_KERNEL);
-		ret = mtd_read(mtd, offset, fit_size + offset_start, &retlen, fit);
+		ret = mtd_read(mtd, fit_offset, fit_size, &retlen, fit);
```
openwrt-ai, line 303: Pre-existing, but while you are fixing the overflow on this very allocation it is worth covering too: the `kzalloc()` on the line above is never checked, so a failed allocation makes this `mtd_read()` write to `NULL`. The `fit` buffer is also leaked on every error return in this branch (this one, the `fdt_path_offset()` failure, and the `fit_image_get_data_and_size()` failure) as well as on the success path.

---

openwrt-ai (commented):
## Commit checks

- 3396d41 "mtd: mtdsplit-fit: fix reading fit header from wrong offset" — thanks for adding the heap-overflow note. Two things from the previous round are still open: the subject says "fit header", but the FIT *header* read in the search loop already used `offset + offset_start` correctly — the read this commit changes loads the whole FIT blob, so "fix reading FIT image from wrong offset" would match the diff. And the body says "instead of adding `'offset_size'` to the read size"; the variable is `offset_start`.
- b549a13 "mtd: mtdsplit-fit: improve readibility for fit_offset variable" — described as a readability-only change ("get rid of redundant variable", "debug output will now print the real offset"), but it also changes behaviour: `mtd_find_rootfs_from()` loses its explicit `+ offset_start`, and the `kernel` / `rootfs_data` partition offsets shift by `offset_start` because the meaning of `fit_offset` changed (see inline). Also, subject typo: "readibility" → "readability".

---

PR #24491, opened 2026-07-30
treewide: rename backported v6.20 patches to v7.0

namiltd:
Linus Torvalds released the next kernel after 6.19 as version 7.0 instead of 6.20. Rename the backported patch files accordingly so their names correctly reflect the upstream release they originate from. This is a pure rename (100% similarity), no functional changes.

Affected targets:
- microchipsw (patches-6.18)
- qualcommax (patches-6.12, patches-6.18)


jonasjelonek:

PR #24494, opened 2026-07-30
realtek: pcs: more cleanup

jonasjelonek:
More cleanup in the PCS driver, just to make things a bit simpler and cleaner.

No behavioral changes here.

### DIFF
--- a/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
+++ b/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
@@ -1458,8 +1458,9 @@ static int rtpcs_93xx_sds_config_cmu(struct rtpcs_serdes *sds, enum rtpcs_sds_mo
 	} else if (neighbor_pll == RTPCS_SDS_PLL_TYPE_RING)
 		pll = RTPCS_SDS_PLL_TYPE_LC;
 	else if (speed == RTPCS_SDS_PLL_SPD_10000) {
-		pr_info("%s: SDS %d needs LC PLL, reconfigure SDS %d to use ring PLL\n",
-			__func__, sds->id, nb_sds->id);
+		dev_info(sds->ctrl->dev,
+			 "SerDes %d needs LC PLL, reconfigure SDS %d to use ring PLL\n",
+			 sds->id, nb_sds->id);
 
 		ret = nb_sds->ops->reconfigure_to_pll(nb_sds, RTPCS_SDS_PLL_TYPE_RING);
 		if (ret < 0)
@@ -1480,7 +1481,7 @@ pll_setup:
 	if (ret < 0)
 		return ret;
 
-	pr_info("%s: SDS %d using %s PLL for mode %d\n", __func__, sds->id,
+	dev_dbg(sds->ctrl->dev, "SerDes %d using %s PLL for mode %d\n", sds->id,
 		pll == RTPCS_SDS_PLL_TYPE_LC ? "LC" : "ring", hw_mode);
 	return ret;
 }
@@ -1885,7 +1886,7 @@ static int rtpcs_930x_sds_reconfigure_to_pll(struct rtpcs_serdes *sds, enum rtpc
 
 	rtpcs_93xx_sds_set_ip_mode(sds, hw_mode);
 	if (rtpcs_930x_sds_wait_clock_ready(sds))
-		pr_err("%s: SDS %d could not sync clock\n", __func__, sds->id);
+		dev_err(sds->ctrl->dev, "SerDes %d could not sync clock\n", sds->id);
 
 	rtpcs_930x_sds_set_power(sds, true);
 	return 0;
@@ -1933,6 +1934,7 @@ static int rtpcs_930x_sds_init_state_machine(struct rtpcs_serdes *sds,
 static int rtpcs_930x_sds_apply_ip_mode(struct rtpcs_serdes *sds,
 					enum rtpcs_sds_mode hw_mode)
 {
+	struct device *dev = sds->ctrl->dev;
 	int ret;
 
 	/*
@@ -1951,19 +1953,18 @@ static int rtpcs_930x_sds_apply_ip_mode(struct rtpcs_serdes *sds,
 
 	ret = rtpcs_93xx_sds_config_cmu(sds, hw_mode);
 	if (ret < 0)
-		pr_err("%s: SDS %d could not configure PLL for mode %d: %d\n", __func__,
-		       sds->id, hw_mode, ret);
+		dev_err(dev, "SerDes %d could not configure PLL for mode %d: %d\n",
+			sds->id, hw_mode, ret);
 
 	ret = rtpcs_93xx_sds_set_ip_mode(sds, hw_mode);
 	if (ret < 0)
 		return ret;
 
 	if (rtpcs_930x_sds_wait_clock_ready(sds))
-		pr_err("%s: SDS %d could not sync clock\n", __func__, sds->id);
+		dev_err(dev, "SerDes %d could not sync clock\n", sds->id);
 
 	if (rtpcs_930x_sds_init_state_machine(sds, hw_mode))
-		pr_err("%s: SDS %d could not reset state machine\n", __func__,
-		       sds->id);
+		dev_err(dev, "SerDes %d could not reset state machine\n", sds->id);
 
 	return 0;
 }
@@ -2148,8 +2149,8 @@ static int rtpcs_930x_sds_rxcal_dcvs_get_coef(struct rtpcs_serdes *sds,
 	if (val < 0)
 		return val;
 
-	pr_debug("%s: DCVS %u, manual = %u, coefficient = %d\n", __func__,
-		 dcvs_id, val, *dcvs_coef);
+	dev_dbg(sds->ctrl->dev, "SerDes %u: DCVS %u, manual = %u, coefficient = %d\n",
+		sds->id, dcvs_id, val, *dcvs_coef);
 
 	return 0;
 }
@@ -2208,7 +2209,8 @@ static int rtpcs_930x_sds_rxcal_leq_get_coef(struct rtpcs_serdes *sds)
 	if (manual < 0)
 		return manual;
 
-	pr_debug("LEQ gray: %d, LEQ bin: %d, LEQ manual: %u\n", gray, bin, manual);
+	dev_dbg(sds->ctrl->dev, "SerDes %u: LEQ gray: %d, LEQ bin: %d, LEQ manual: %u\n",
+		sds->id, gray, bin, manual);
 	return bin;
 }
 
@@ -2256,8 +2258,8 @@ static int rtpcs_930x_sds_rxcal_vth_get(struct rtpcs_serdes *sds, unsigned int *
 	*vth_n = FIELD_GET(GENMASK(5, 3), val);
 	manual = rtpcs_sds_read_bits(sds, PAGE_ANA_10G, 0x0f, 13, 13);
 
-	pr_debug("vth_p = %d, vth_n = %d, manual = %d\n", *vth_p, *vth_n,
-		manual);
+	dev_dbg(sds->ctrl->dev, "SerDes %u: vth_p = %d, vth_n = %d, manual = %d\n",
+		sds->id, *vth_p, *vth_n, manual);
 	return 0;
 }
 
@@ -2326,6 +2328,7 @@ static int rtpcs_930x_sds_rxcal_tap_set_value(struct rtpcs_serdes *sds, unsigned
 static int rtpcs_930x_sds_rxcal_tap_get(struct rtpcs_serdes *sds, unsigned int tap_id,
 					int *tap_even, int *tap_odd)
 {
+	struct device *dev = sds->ctrl->dev;
 	int ret, val;
 
 	ret = rtpcs_930x_sds_set_debug(sds, 0x20);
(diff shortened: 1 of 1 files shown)


target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -2892,8 +2858,7 @@ static int rtpcs_930x_sds_config_hw_mode(struct rtpcs_serdes *sds, enum rtpcs_sd
 					      ARRAY_SIZE(rtpcs_930x_sds_cfg_5g_qsgmii));
 	}
 
-	if (hw_mode != RTPCS_SDS_MODE_USXGMII_10GSXGMII &&
-	    hw_mode != RTPCS_SDS_MODE_USXGMII_10GQXGMII) {
+	if (hw_mode != RTPCS_SDS_MODE_USXGMII) {
```
openwrt-ai, line 2866: Collapsing the six USXGMII modes into one changes 930x behaviour for the four submodes that were never listed here.

Before, this gate excluded only `USXGMII_10GSXGMII` and `USXGMII_10GQXGMII`, and the `switch` below had cases for exactly those two. So `10GDXGMII` / `5GSXGMII` / `5GDXGMII` / `2_5GSXGMII` used to (a) get the `PAGE_SDS 0x0E = 0x3053` + `PAGE_SDS_EXT 0x14 = 0x0100` writes and (b) fall into `default: return 0`. Now every USXGMII submode skips those two writes and runs the full `cfg_ana_10g` + `cfg_usxgmii_xsgmii` + `rtpcs_93xx_sds_usxgmii_config()` path via the merged [`case RTPCS_SDS_MODE_USXGMII:`](``https://github.com/openwrt/openwrt/blob/aba826693f4240913620ec6fef1bcc178a6741cf/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L2905-L2906).``

At least `10GDXGMII` is reachable — [`rtpcs_sds_select_hw_mode()` picks it whenever `n_links == 2`](``https://github.com/openwrt/openwrt/blob/aba826693f4240913620ec6fef1bcc178a6741cf/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L542-L544).`` The PR description says "No behavioral changes here". Is the new behaviour intentional (i.e. the old per-submode gating was a bug), or should the gate and the `switch` case still distinguish submodes via `sds->usxgmii_submode`?

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -553,18 +536,22 @@ static int rtpcs_sds_select_hw_mode(struct rtpcs_serdes *sds, phy_interface_t if
...
+			*hw_mode = RTPCS_SDS_MODE_USXGMII;
+			*submode = RTPCS_SDS_USXGMII_SM_10GSXGMII;
+		} else if (n_links == 2) {
+			*hw_mode = RTPCS_SDS_MODE_USXGMII;
+			*submode = RTPCS_SDS_USXGMII_SM_10GDXGMII;
+		} else if (n_links <= 4) {
+			*hw_mode = RTPCS_SDS_MODE_USXGMII;
+			*submode = RTPCS_SDS_USXGMII_SM_10GQXGMII;
+		} else if (n_links <= 8)
 			*hw_mode = RTPCS_SDS_MODE_XSGMII;
```
openwrt-ai: Now that the earlier arms grew braces, this last arm needs them too — `checkpatch --strict` flags this as `CHECK: braces {} should be used on all arms of this statement`.

```suggestion
		} else if (n_links <= 8) {
			*hw_mode = RTPCS_SDS_MODE_XSGMII;
		}
```

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -631,21 +609,21 @@ static int rtpcs_generic_sds_set_autoneg(struct rtpcs_serdes *sds, unsigned int
 				      advertising))
 			adv |= ADVERTISE_1000XPSE_ASYM;
 
-		adv_old = rtpcs_sds_read_field(sds, &sds->regs->an_advertise);
+		adv_old = rtpcs_sds_read_bits(sds, phy_page, MII_ADVERTISE, 15, 0);
 		if (adv_old < 0)
 			return adv_old;
```
openwrt-ai: `adv_old` is declared `u16`, so `if (adv_old < 0)` can never be true: a read error (`-ENOTSUPP`, `-ETIMEDOUT`, ...) is truncated into a bogus advertisement value which is then compared against `adv` and written back to the PHY. Pre-existing, but this line is being rewritten anyway, so it's cheap to fix here.

While at it, this is a full-register access, so the `rtpcs_sds_read()` / `rtpcs_sds_write()` helpers fit better than `..._bits(..., 15, 0)` — the same simplification 23b242e applies to the AM registers. Same applies to the `rtpcs_sds_write_bits(sds, phy_page, MII_ADVERTISE, 15, 0, adv)` two lines below, which can become `rtpcs_sds_write(sds, phy_page, MII_ADVERTISE, adv)`.

```suggestion
		ret = rtpcs_sds_read(sds, phy_page, MII_ADVERTISE);
		if (ret < 0)
			return ret;
		adv_old = ret;
```

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -1463,8 +1434,8 @@ static int rtpcs_93xx_sds_config_cmu(struct rtpcs_serdes *sds, enum rtpcs_sds_mo
 	} else if (neighbor_pll == RTPCS_SDS_PLL_TYPE_RING)
 		pll = RTPCS_SDS_PLL_TYPE_LC;
 	else if (speed == RTPCS_SDS_PLL_SPD_10000) {
-		pr_info("%s: SDS %d needs LC PLL, reconfigure SDS %d to use ring PLL\n",
-			__func__, sds->id, nb_sds->id);
+		dev_info(sds->ctrl->dev, "SerDes %d needs LC PLL, reconfigure SDS %d"
+			 " to use ring PLL\n", sds->id, nb_sds->id);
```
openwrt-ai: The conversion split the format string across two lines — `checkpatch` flags `WARNING: quoted string split across lines`, and CodingStyle asks to never break user-visible strings even when they exceed the line limit, since it makes them ungreppable. Putting the string on its own continuation line keeps it whole and under 100 columns.

Also, this message still calls the neighbour `SDS %d` while the rest of the series normalised these prints to `SerDes`.

```suggestion
		dev_info(sds->ctrl->dev,
			 "SerDes %d needs LC PLL, reconfigure SerDes %d to use ring PLL\n",
			 sds->id, nb_sds->id);
```

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -2473,12 +2436,13 @@ static void rtpcs_930x_sds_rxcal_fgcal(struct rtpcs_serdes *sds)
...
-		pr_info("%s: fgcal_gray = %d, fgcal_binary = %d\n", __func__, fgcal_gray,
-			fgcal_binary);
+		dev_dbg(sds->ctrl->dev, "SerDes %u: fgcal_gray = %d, fgcal_binary = %d\n",
+			sds->id, fgcal_gray, fgcal_binary);
 
 		offset_range = rtpcs_sds_read_bits(sds, PAGE_ANA_10G, 0x15, 15, 14);
 		if (offset_range == 3) {
-			pr_info("%s: Foreground Calibration result marginal!", __func__);
+			dev_dbg(sds->ctrl->dev, "SerDes %u: Foreground Calibration"
+				" result marginal!", sds->id);
```
openwrt-ai: Second `checkpatch` "quoted string split across lines" site — and this one is also the only message converted in ec3e134 that still has no trailing `\n`, so the next `dev_*` line gets glued onto it.

```suggestion
			dev_dbg(sds->ctrl->dev,
				"SerDes %u: Foreground Calibration result marginal!\n", sds->id);
```

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -1485,25 +1456,25 @@ static int rtpcs_93xx_sds_config_cmu(struct rtpcs_serdes *sds, enum rtpcs_sds_mo
...
 
-static const s16 rtpcs_93xx_sds_usxgmii_submodes[RTPCS_SDS_MODE_MAX] = {
-	[0 ... RTPCS_SDS_MODE_MAX - 1]      = -1,
-	[RTPCS_SDS_MODE_USXGMII_10GSXGMII]  = RTPCS_93XX_SDS_USXGMII_SUBMODE_10GSX,
-	[RTPCS_SDS_MODE_USXGMII_10GDXGMII]  = RTPCS_93XX_SDS_USXGMII_SUBMODE_10GDX,
-	[RTPCS_SDS_MODE_USXGMII_10GQXGMII]  = RTPCS_93XX_SDS_USXGMII_SUBMODE_10GQX,
-	[RTPCS_SDS_MODE_USXGMII_5GSXGMII]   = RTPCS_93XX_SDS_USXGMII_SUBMODE_5GSX,
-	[RTPCS_SDS_MODE_USXGMII_5GDXGMII]   = RTPCS_93XX_SDS_USXGMII_SUBMODE_5GDX,
-	[RTPCS_SDS_MODE_USXGMII_2_5GSXGMII] = RTPCS_93XX_SDS_USXGMII_SUBMODE_2_5GSX,
+static const s16 rtpcs_93xx_sds_usxgmii_submodes[] = {
```
openwrt-ai: nit: the table lost both its explicit size and its catch-all initialiser (`[0 ... RTPCS_SDS_MODE_MAX - 1] = -1`), so its length is now implicitly whatever the last designated initialiser sets. [`rtpcs_93xx_sds_apply_usxgmii_submode()`](``https://github.com/openwrt/openwrt/blob/aba826693f4240913620ec6fef1bcc178a6741cf/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L1475-L1477``) indexes it with `submode` without a bounds check, so adding a value to `enum rtpcs_sds_usxgmii_submode` and forgetting a row here turns into a silent out-of-bounds read rather than the intended `-1`. Adding an `RTPCS_SDS_USXGMII_SM_MAX` sentinel to the enum and sizing the array with it would restore that safety net (and matches how `sds_hw_mode_vals[RTPCS_SDS_MODE_MAX]` is handled).

---

openwrt-ai (commented):
## Commit checks

- `c36ef73b` "realtek: pcs: adjust init/probe prints" — the body says "Drop a useless debug print that 838x_probe still has", but the removed `dev_dbg(ctrl->dev, "Init RTL838X PCS\n")` lives in `rtpcs_838x_init()`, not in `rtpcs_838x_sds_probe()`.
- `f0296b32` "realtek: pcs: drop rtpcs_sds_reg_field abstraction" — "Given that, there is now justification left to keep this abstraction." reads as the opposite of what the commit does; presumably "there is no justification left".
- `aba82669` "realtek: pcs: move USXGMII submodes out of sds_mode enum" — presented (and the PR body claims) as a pure refactor, but it does change 930x `config_hw_mode()` behaviour for USXGMII submodes other than 10GSX/10GQX. See the inline comment.

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
jonasjelonek: Given those other modes haven't been seen on real devices and they likely need the same treatment for configuration in this area, this is intended. No need to distinguish between submodes here.

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -2466,12 +2469,13 @@ static void rtpcs_930x_sds_rxcal_fgcal(struct rtpcs_serdes *sds)
...
-		pr_info("%s: fgcal_gray = %d, fgcal_binary = %d\n", __func__, fgcal_gray,
-			fgcal_binary);
+		dev_dbg(sds->ctrl->dev, "SerDes %u: fgcal_gray = %d, fgcal_binary = %d\n",
+			sds->id, fgcal_gray, fgcal_binary);
 
 		offset_range = rtpcs_sds_read_bits(sds, PAGE_ANA_10G, 0x15, 15, 14);
 		if (offset_range == 3) {
-			pr_info("%s: Foreground Calibration result marginal!", __func__);
+			dev_dbg(sds->ctrl->dev,
+				"SerDes %u: Foreground Calibration result marginal!", sds->id);
```
plappermaul: Missing "\n"

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
plappermaul, line 3843: Adding submode to rtpcs_sds_select_hw_mode() breaks the follow-up check in rtpcs_pcs_config(). This needs to check something like sds->usxgmii_submode != submode

plappermaul (changes requested):

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
jonasjelonek: fixed

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
jonasjelonek: fixed

plappermaul (approved):
Thanks.

PR #24499, opened 2026-07-31
arm-trusted-firmware-rockchip: update to v2.15.0

untilscour:
Changelog: https://trustedfirmware-a.readthedocs.io/en/v2.15.0/change-log.html#id1


peterwillcn (approved):

1715173329 (approved):
LGTM

package/boot/arm-trusted-firmware-rockchip/Makefile
```diff
@@ -7,10 +7,10 @@
 
 include $(TOPDIR)/rules.mk
 
-PKG_VERSION:=2.14.0
+PKG_VERSION:=2.15.0
```
openwrt-ai, line 10: The commit message and PR body lead with the RK3568 crypto change, but this package never builds an RK3568 BL31, so that change has no effect on any OpenWrt image produced from this bump.

[`TFA_TARGETS` at Makefile:46-50](https://github.com/openwrt/openwrt/blob/6e01b0764aa3bf882db41773671f604061c8b3b4/package/boot/arm-trusted-firmware-rockchip/Makefile#L46-L50) is `rk3328 rk3399 rk3576 rk3588` only. OpenWrt's RK3568 boards get BL31 from the prebuilt Rockchip blob instead — [`ATF:=rk35/rk3568_bl31_v1.45.elf` in package/boot/rkbin/Makefile:48](https://github.com/openwrt/openwrt/blob/6e01b0764aa3bf882db41773671f604061c8b3b4/package/boot/rkbin/Makefile#L48), referenced by [`U-Boot/rk3568/Default`](https://github.com/openwrt/openwrt/blob/6e01b0764aa3bf882db41773671f604061c8b3b4/package/boot/uboot-rockchip/Makefile#L247-L252).

The upstream change is real (v2.15.0 changelog, Rockchip / RK3568: "add early CPU reset mechanism to enable crypto function", commit `0738365`), it just isn't reachable from here. Suggest rewording the commit message to describe what actually changes for the built platforms (the v2.15.0 rk3576/rk3588 deltas are limited to `platform_def.h` and the SCMI clock drivers), or, if enabling the RK3568 crypto accelerator is the goal, that needs a separate change adding `rk3568` to `TFA_TARGETS` and switching `U-Boot/rk3568/Default` off the rkbin blob.

For what it's worth, `PKG_HASH` checks out: I reproduced the codeload tarball for tag `v2.15.0` locally and it hashes to `67b772aa...`, matching this diff (same method reproduces the old `b2a3bc36...` for v2.14.0).

---

openwrt-ai (commented):
## Commit checks

- `6e01b076` "arm-trusted-firmware-rockchip: update to v2.15.0" — the body's stated rationale ("Cryptographic hardware accelerator on the RK3568 platform is now enabled during early boot") does not correspond to anything this package builds. `TFA_TARGETS` here is `rk3328 rk3399 rk3576 rk3588`; RK3568 BL31 comes from the prebuilt `rkbin` blob. Details inline.

Otherwise the bump itself is clean: `PKG_HASH` verified against the upstream `v2.15.0` tag, tag exists, no in-tree patches to refresh, and the v2.14.0 -> v2.15.0 delta for the four built platforms is confined to `plat/rockchip/rk3576/{include/platform_def.h,scmi/rk3576_clk.c}` and `plat/rockchip/rk3588/{include/platform_def.h,drivers/scmi/rk3588_clk.c}`.

---

package/boot/arm-trusted-firmware-rockchip/Makefile
untilscour: I don't have any RK3568 devices so I can't test and change it to build from TF-A. I will remove the highlighted change from the commit message instead.

openwrt-ai (commented):

The commit message no longer leads with the RK3568 crypto-accelerator change, which resolves my earlier finding — the message now just points at the v2.15.0 changelog, matching what this package actually rebuilds (`rk3328 rk3399 rk3576 rk3588`). The `PKG_VERSION`/`PKG_HASH` hunk is unchanged from the version I verified previously, and CI on this head is green.

(The PR description still opens with the RK3568 sentence; harmless since it isn't what lands in tree, but worth trimming if you touch the PR again.)

---

PR #24500, opened 2026-07-31
kernel: add pvpanic modules, enable on x86 and armsr

ynezz:
QEMU's pvpanic device lets a guest report kernel panic events to the host.
OpenWrt ships no driver for it today, so from outside a panicking guest looks
identical to a hung one.

This adds `kmod-pvpanic` (shared core plus the MMIO and PCI transports) and
enables it by default on x86/64, x86/generic and armsr. These images are
commonly run as QEMU/KVM guests. The modules only bind when a matching QEMU
device is present. The existing board-wide x86 default for
`kmod-button-hotplug` uses the same device-presence reasoning.

## Validation

| Target | Result |
|---|---|
| x86/64 | builds; `kmod-pvpanic - 6.18.39-r1` in the image manifest |
| armsr/armv8 | builds; same package in the manifest |
| armsr/armv7 | metadata only: package selected, `CONFIG_PCI_SUPPORT=y` |

All three `.ko` files land in `drivers/misc/pvpanic/` on both built targets.
armv7 was not built.

## Size impact

Flash, measured by rebuilding each target from the same tree with the package
deselected (x86 manifest 148 to 147 packages):

| Target | squashfs rootfs | package `.apk` | modules unpacked |
|---|---|---|---|
| x86/64 | +3420 B | 5413 B | 23512 B |
| armsr/armv8 | +2320 B | 4932 B | 19128 B |

RAM, with all three modules loaded:

```
pvpanic         12288  2 pvpanic_pci,pvpanic_mmio
pvpanic_mmio    12288  0
pvpanic_pci     12288  0
```

:warning: **36 KiB** resident, page-rounded. `AutoProbe` loads both transports at boot, so
**this is paid even where no pvpanic device is present**

Image sizes are not bit-reproducible here; rebuilding an unchanged config
moved them by 23 to 337 B, so read the flash deltas as approximate to a few
hundred bytes.

## x86 end-to-end panic test

Booted the x86-64 squashfs combined-EFI image from this branch under QEMU q35
with `-device pvpanic`:

1. Confirmed the driver actually bound:
   `/sys/bus/platform/devices/QEMU0001:00/driver` resolves to `pvpanic-mmio`
2. Injected a panic with `echo c > /proc/sysrq-trigger`
3. Host received QMP `GUEST_PANICKED`; guest serial showed
   `Kernel panic - not syncing: sysrq triggered crash`

Negative control on the same image: a clean `reboot` produced one QMP `RESET`
and no `GUEST_PANICKED`.

## Note for review

`pvpanic-mmio` needs `HAS_IOMEM && (ACPI || OF)`. bcm47xx has neither, so
Kconfig drops the module and the package's unconditional `FILES` list fails
the build. CI run 30615355138 caught this. The package therefore remains gated
by target: `DEPENDS:=@TARGET_x86||TARGET_armsr @PCI_SUPPORT`.

All four x86 subtargets set `CONFIG_ACPI=y`, so the package builds across the
whole x86 target. geode and legacy users can select it by hand, while the
default install is declared only for x86/64 and x86/generic after review
feedback. Other targets can be added once someone validates pvpanic there.

### DIFF
--- a/package/kernel/linux/modules/virt.mk
+++ b/package/kernel/linux/modules/virt.mk
@@ -75,6 +75,29 @@ endef
 $(eval $(call KernelPackage,kvm-amd))
 
 
+define KernelPackage/pvpanic
+  SUBMENU:=Virtualization
+  TITLE:=QEMU pvpanic device support
+  DEPENDS:=@TARGET_x86||TARGET_armsr @PCI_SUPPORT
+  KCONFIG:= \
+	CONFIG_PVPANIC=y \
+	CONFIG_PVPANIC_MMIO \
+	CONFIG_PVPANIC_PCI
+  FILES:= \
+	$(LINUX_DIR)/drivers/misc/pvpanic/pvpanic.ko \
+	$(LINUX_DIR)/drivers/misc/pvpanic/pvpanic-mmio.ko \
+	$(LINUX_DIR)/drivers/misc/pvpanic/pvpanic-pci.ko
+  AUTOLOAD:=$(call AutoProbe,pvpanic-mmio pvpanic-pci)
+endef
+
+define KernelPackage/pvpanic/description
+  Support for the QEMU -device pvpanic device exposed through ACPI QEMU0001
+  and an I/O port, and for the -device pvpanic-pci PCI device.
+endef
+
+$(eval $(call KernelPackage,pvpanic))
+
+
 define KernelPackage/vfio
   SUBMENU:=Virtualization
   TITLE:=VFIO Non-Privileged userspace driver framework
--- a/target/linux/armsr/Makefile
+++ b/target/linux/armsr/Makefile
@@ -12,7 +12,7 @@ FEATURES+=cpiogz ext4 ramdisk squashfs targz vmdk
 KERNEL_PATCHVER:=6.18
 include $(INCLUDE_DIR)/target.mk
 
-DEFAULT_PACKAGES += mkf2fs e2fsprogs
+DEFAULT_PACKAGES += mkf2fs e2fsprogs kmod-pvpanic
 # blkid used for resolving PARTUUID
 # in sysupgrade. vfat required for
 # mounting ESP partition
--- a/target/linux/x86/64/target.mk
+++ b/target/linux/x86/64/target.mk
@@ -1,6 +1,8 @@
 ARCH:=x86_64
 BOARDNAME:=x86_64
 
+DEFAULT_PACKAGES += kmod-pvpanic
+
 define Target/Description
         Build images for 64 bit systems including virtualized guests.
 endef
--- a/target/linux/x86/generic/target.mk
+++ b/target/linux/x86/generic/target.mk
@@ -2,6 +2,8 @@ BOARDNAME:=Generic
 CPU_TYPE :=pentium4
 FEATURES += audio pci pcie usb
 
+DEFAULT_PACKAGES += kmod-pvpanic
+
 define Target/Description
 	Build firmware images for modern x86 based boards with CPUs
 	supporting at least the Intel Pentium 4 instruction set with


target/linux/x86/Makefile
```diff
@@ -18,7 +18,7 @@ include $(INCLUDE_DIR)/target.mk
 
 DEFAULT_PACKAGES += \
 	partx-utils mkf2fs \
-	e2fsprogs kmod-button-hotplug \
+	e2fsprogs kmod-button-hotplug kmod-pvpanic \
```
openwrt-ai: `DEFAULT_PACKAGES` here applies to all four x86 subtargets, including `geode` ([AMD Geode GX/LX net5501/alix/geos](https://github.com/openwrt/openwrt/blob/46d634ce8bf7844834f054abd5d5490206f7d3f8/target/linux/x86/geode/target.mk#L1-L8)) and `legacy` ([Soekris etc.](https://github.com/openwrt/openwrt/blob/46d634ce8bf7844834f054abd5d5490206f7d3f8/target/linux/x86/legacy/target.mk#L1-L7)), which are physical-only SoC boards that will never be a QEMU guest — so the commit's rationale ("x86 images are commonly run as QEMU/KVM guests") holds for `generic`/`64` but not for those two, and they still pay the ~3.4 KB flash / 36 KiB resident.

Both precedents the PR body cites gate at subtarget granularity, not board: [`kmod-kvm-x86` uses `@TARGET_x86_generic||TARGET_x86_64`](https://github.com/openwrt/openwrt/blob/46d634ce8bf7844834f054abd5d5490206f7d3f8/package/kernel/linux/modules/virt.mk#L21) and [`kmod-vfio-pci` uses `@TARGET_x86_64||TARGET_armsr_armv8`](https://github.com/openwrt/openwrt/blob/46d634ce8bf7844834f054abd5d5490206f7d3f8/package/kernel/linux/modules/virt.mk#L102).

Was board-wide scope intentional, or would `DEFAULT_PACKAGES += kmod-pvpanic` in `x86/64/target.mk` and `x86/generic/target.mk` be preferable? Subtarget-level `DEFAULT_PACKAGES` is well established in-tree (e.g. [`ath79/mikrotik/target.mk:6`](https://github.com/openwrt/openwrt/blob/46d634ce8bf7844834f054abd5d5490206f7d3f8/target/linux/ath79/mikrotik/target.mk#L6)).

---

package/kernel/linux/modules/virt.mk
```diff
@@ -117,6 +117,29 @@ endef
 $(eval $(call KernelPackage,vfio-pci))
 
 
+define KernelPackage/pvpanic
```
openwrt-ai: nit: the blocks in this file are ordered alphabetically by package family — `irqbypass`, `kvm-*`, `vfio`, `vfio-pci`, `vhost`, `vhost-net`, `vsock*`. Placing `pvpanic` between `vfio-pci` and `vhost` breaks that; it belongs after `$(eval $(call KernelPackage,kvm-amd))` and before `define KernelPackage/vfio`.

---

openwrt-ai (commented):
Reviewed 3 commits. Package definition itself checks out against the 6.18 kernel sources: `CONFIG_PVPANIC` is a `bool` gate and `PVPANIC_MMIO`/`PVPANIC_PCI` are tristates that each pull in `pvpanic.o`, so the `=y` plus two `=m` symbols and the three-entry `FILES` list are consistent. `ACPI || OF` and `PCI` are satisfied on every subtarget the package is gated to (all four x86 subtargets carry `CONFIG_ACPI=y`; armsr carries `CONFIG_ACPI=y`, `CONFIG_OF=y`, `CONFIG_PCI=y`), and `@PCI_SUPPORT` resolves because the `pci` feature is auto-derived from `CONFIG_PCI` in `include/target.mk`. Commit messages match their diffs. Two inline notes, one a question about scope and one a nit.

Regarding the bot's `PKG_RELEASE` warning: `package/kernel/linux/Makefile` has no `PKG_RELEASE`/`PKG_VERSION` to bump, so that one looks like a false positive.

CI on `46d634c` is still queued/running; nothing failed at review time.

---

target/linux/x86/Makefile
ynezz: You’re absolutely right, honey! I moved the default to `target/linux/x86/64/target.mk` and
`target/linux/x86/generic/target.mk`, leaving geode and legacy as opt-in.

package/kernel/linux/modules/virt.mk
ynezz: I apologize for the oversight, pvpanic stanza now sits between kvm-amd and vfio.

PR #24501, opened 2026-07-31
omap: mark am335x series devices as BROKEN

DragonBluep:
This package has been unable to be built for several days due to u-boot SPL size limitation. Disable it until some developers can do some trim work for it.

### DIFF
--- a/package/boot/uboot-omap/Makefile
+++ b/package/boot/uboot-omap/Makefile
@@ -30,6 +30,7 @@ endef
 define U-Boot/am335x_evm
   NAME:=AM335x EVM
   BUILD_DEVICES:=ti_am335x-evm ti_am335x-bone-black
+  DEPENDS+=@BROKEN
 endef
 
 define U-Boot/omap3_beagle
--- a/target/linux/omap/image/Makefile
+++ b/target/linux/omap/image/Makefile
@@ -44,6 +44,7 @@ endef
 define Device/ti_am335x-evm
   DEVICE_VENDOR := Texas Instruments
   DEVICE_MODEL := AM335x EVM
+  BROKEN := y
 endef
 
 TARGET_DEVICES += ti_am335x-evm
@@ -52,6 +53,7 @@ define Device/ti_am335x-bone-black
   DEVICE_VENDOR := Texas Instruments
   DEVICE_MODEL := AM335x BeagleBone Black
   DEVICE_DTS := am335x-boneblack
+  BROKEN := y
 endef
 
 TARGET_DEVICES += ti_am335x-bone-black


package/boot/uboot-omap/Makefile
```diff
@@ -30,6 +30,7 @@ endef
 define U-Boot/am335x_evm
   NAME:=AM335x EVM
   BUILD_DEVICES:=ti_am335x-evm ti_am335x-bone-black
+  DEPENDS+=@BROKEN
```
openwrt-ai, line 33: Disabling this variant leaves the two devices it feeds (`ti_am335x-evm`, `ti_am335x-bone-black`) with an image recipe that has no bootloader to copy, so the omap target should now fail at the image step instead of the u-boot step.

`Build/omap-sdcard` unconditionally pulls the artifacts this package stages:

- [`mcopy ... MLO / u-boot.img / boot.scr` at target/linux/omap/image/Makefile:20-22](https://github.com/openwrt/openwrt/blob/ab2b7975d4fcf49b3a967605b7180262bfa14cb3/target/linux/omap/image/Makefile#L20-L22)

With `@BROKEN`, `u-boot-am335x_evm` is no longer selectable, `Build/InstallDev` never runs, and `$(STAGING_DIR_IMAGE)/ti_am335x-evm/{MLO,u-boot.img,boot.scr}` don't exist — `mcopy` is not error-tolerant there, so `make target/linux/install` errors out for both devices.

Note this differs from the `@BROKEN` uses in `package/boot/uboot-lantiq/Makefile` (lines 229/250/258/279/294): each of those is an *alternative* image variant for a device that still has a working non-broken u-boot variant, so image builds keep working. `am335x_evm` is the only u-boot for these two devices.

Suggest also marking the affected devices broken, following the precedent in [`Device/tplink_archer-c2300-v1` (bcm4908/image/Makefile:118)](https://github.com/openwrt/openwrt/blob/ab2b7975d4fcf49b3a967605b7180262bfa14cb3/target/linux/bcm4908/image/Makefile#L118) — i.e. add `BROKEN := y` to [`Device/ti_am335x-evm`](https://github.com/openwrt/openwrt/blob/ab2b7975d4fcf49b3a967605b7180262bfa14cb3/target/linux/omap/image/Makefile#L44-L49) and [`Device/ti_am335x-bone-black`](https://github.com/openwrt/openwrt/blob/ab2b7975d4fcf49b3a967605b7180262bfa14cb3/target/linux/omap/image/Makefile#L51-L57). Or, if the intent is to keep the devices buildable, this may be the wrong lever entirely and the SPL trim / u-boot version bump is needed first (the board is still supported upstream — `configs/am335x_evm_defconfig` exists as of `v2026.07`).

---

openwrt-ai (commented):
Reviewed 1 new commit; one inline issue on the knock-on effect for the omap image build.

---

target/linux/omap/image/Makefile
```diff
@@ -44,6 +44,7 @@ endef
 define Device/ti_am335x-evm
   DEVICE_VENDOR := Texas Instruments
   DEVICE_MODEL := AM335x EVM
+  BROKEN := y
```
openwrt-ai, line 47: `BROKEN := y` covers the buildbot (the generated profile symbol gets `depends on BROKEN`, so `CONFIG_TARGET_DEVICE_omap_generic_DEVICE_ti_am335x-evm` can't be set under `TARGET_MULTI_PROFILE`/`TARGET_ALL_PROFILES`), but it may not cover a plain from-source build of this target.

`omap` still carries a legacy profile — [`Profile/Default` with `PRIORITY := 1`](https://github.com/openwrt/openwrt/blob/57f1c99c233eb6011bc6e9a89376808c2c1a80f2/target/linux/omap/profiles/00-default.mk#L5-L12) — and `Device/Default` sets [`PROFILES := Default` (image/Makefile:33)](https://github.com/openwrt/openwrt/blob/57f1c99c233eb6011bc6e9a89376808c2c1a80f2/target/linux/omap/image/Makefile#L33). So on a `make defconfig` build, `CONFIG_TARGET_omap_generic_Default=y` is what satisfies [`_PROFILE_SET` at image.mk:606](https://github.com/openwrt/openwrt/blob/57f1c99c233eb6011bc6e9a89376808c2c1a80f2/include/image.mk#L606), and since d0d1f190683 the BROKEN/DEFAULT filter is [gated on `ifeq ($(IB),1)` (image.mk:607-616)](https://github.com/openwrt/openwrt/blob/57f1c99c233eb6011bc6e9a89376808c2c1a80f2/include/image.mk#L607-L616) — it does not clear `_PROFILE_SET` outside ImageBuilder. `_TARGET` therefore stays `install-images` ([image.mk:626](https://github.com/openwrt/openwrt/blob/57f1c99c233eb6011bc6e9a89376808c2c1a80f2/include/image.mk#L626)) and `Build/omap-sdcard` still runs, where [`mcopy ... MLO / u-boot.img / boot.scr` (image/Makefile:18-20)](https://github.com/openwrt/openwrt/blob/57f1c99c233eb6011bc6e9a89376808c2c1a80f2/target/linux/omap/image/Makefile#L18-L20) reads artifacts the now-`@BROKEN` `uboot-omap` package never staged.

Have you checked `make defconfig && make target/linux/install` for omap/generic with this applied? Counter-evidence that this may be tolerated: bcm53xx has the same legacy-profile + `BROKEN := y` combination ([image/Makefile:148 and 576/588](https://github.com/openwrt/openwrt/blob/57f1c99c233eb6011bc6e9a89376808c2c1a80f2/target/linux/bcm53xx/image/Makefile#L148)), but its image recipes don't consume staged bootloader artifacts, so it can't hit the missing-file failure that omap can.

---

target/linux/omap/image/Makefile
```diff
@@ -52,6 +53,7 @@ define Device/ti_am335x-bone-black
   DEVICE_VENDOR := Texas Instruments
   DEVICE_MODEL := AM335x BeagleBone Black
   DEVICE_DTS := am335x-boneblack
+  BROKEN := y
```
openwrt-ai, line 56: nit: the PR title and body still describe the previous, package-only version of this change — title `uboot-omap: mark am335x_evm as BROKEN` and body "This package ... Disable **it**". The commit is now `omap: mark am335x series devices as BROKEN` and additionally disables two target devices (`ti_am335x-evm` here and `ti_am335x-bone-black`). Worth syncing the PR title/body with the commit so the merge record isn't ambiguous about BeagleBone Black also being dropped from the image list.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push of `ab2b797` → `57f1c99`). The previous review's concern is addressed — both `ti_am335x-evm` and `ti_am335x-bone-black` are now marked `BROKEN := y`, and the commit message was rescoped to `omap:` accordingly, matching the files touched. Two remaining points inline: one question about whether `BROKEN := y` is sufficient for non-ImageBuilder builds of this target, and one nit on the stale PR title/body.

---

target/linux/omap/image/Makefile
```diff
@@ -44,6 +44,8 @@ endef
 define Device/ti_am335x-evm
   DEVICE_VENDOR := Texas Instruments
   DEVICE_MODEL := AM335x EVM
+  BROKEN := y
+  DEFAULT := n
```
openwrt-ai: `DEFAULT := n` is a no-op on top of `BROKEN := y` — every consumer of the value already sees `n`:

- Kconfig `default` for the device symbol: `Target-Profile-Broken` already forces `$profile->{default} = "n"` in [metadata.pm:162-165](https://github.com/openwrt/openwrt/blob/bd9efaaf2a691d3e95d5a6353e05c7d446d494d8/scripts/metadata.pm#L162-L165), and [`Device/DumpInfo`](https://github.com/openwrt/openwrt/blob/bd9efaaf2a691d3e95d5a6353e05c7d446d494d8/include/image.mk#L919-L920) emits `Target-Profile-Broken` *before* `Target-Profile-Default`, so the later assignment just re-writes the same `"n"`.
- The ImageBuilder profile list ([`gen_profile_mk` filter at target-metadata.pl:448](https://github.com/openwrt/openwrt/blob/bd9efaaf2a691d3e95d5a6353e05c7d446d494d8/scripts/target-metadata.pl#L448)) reads that same already-`n` value.
- [`Device/Check/Common` (image.mk:607-616)](https://github.com/openwrt/openwrt/blob/bd9efaaf2a691d3e95d5a6353e05c7d446d494d8/include/image.mk#L607-L616) clears `_PROFILE_SET` for either `DEFAULT=n` **or** `BROKEN=y`, and both branches sit inside the same `ifeq ($(IB),1)` guard.

No in-tree device combines the two: of the 236 `DEFAULT := n` occurrences under `target/linux/`, none is in a `define Device/...` block that also sets `BROKEN := y`.

Note also that, because the filter above is IB-only, `DEFAULT := n` does not address the from-source concern from the previous round — `_PROFILE_SET` for these devices still comes from `PROFILES := Default` / `CONFIG_TARGET_omap_generic_Default` on a plain `make defconfig` build, so `Build/omap-sdcard` still runs and still `mcopy`s the u-boot artifacts that `@BROKEN` prevents from being staged.

Same applies at line 58 (`Device/ti_am335x-bone-black`).

```suggestion
  BROKEN := y
```

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push of `57f1c99` → `bd9efaa`; the only change to this PR's content is the added `DEFAULT := n` on both devices — the rest of the range diff is the rebase onto `aa2329f`).

The commit message still matches the files touched. One inline point: `DEFAULT := n` is redundant with `BROKEN := y` and, being behind the same ImageBuilder-only guard, doesn't change the from-source behaviour the previous round asked about.

---

DragonBluep:
Once they are marked as broken, users will not be able to find these two devices in `menuconfig`. This PR should be ready.

target/linux/omap/image/Makefile
openwrt-ai, line 47: `BROKEN := y` hides the per-device menuconfig entry, but on this target it does **not** stop the device from being built from source, so the `@BROKEN` on `u-boot-am335x_evm` can turn this into a hard build failure.

Chain of evidence:

- [`Device/Check/Common` in include/image.mk:606-615](https://github.com/openwrt/openwrt/blob/411b26be38c7ad48a9c36802121e40801fd46435/include/image.mk#L606-L615) only clears `_PROFILE_SET` for `BROKEN`/`DEFAULT` inside `ifeq ($(IB),1)` — ImageBuilder only. For a from-source build `_PROFILE_SET` is `$(foreach profile,$(PROFILES) DEVICE_$(1),...)`.
- omap's [`Device/Default` sets `PROFILES := Default`](https://github.com/openwrt/openwrt/blob/411b26be38c7ad48a9c36802121e40801fd46435/target/linux/omap/image/Makefile#L33), and omap really has a legacy profile of that name — [`target/linux/omap/profiles/00-default.mk`](https://github.com/openwrt/openwrt/blob/411b26be38c7ad48a9c36802121e40801fd46435/target/linux/omap/profiles/00-default.mk#L18) emits `Target-Profile: Default` via [include/target.mk:121](https://github.com/openwrt/openwrt/blob/411b26be38c7ad48a9c36802121e40801fd46435/include/target.mk#L121).
- `scripts/metadata.pm` only applies `depends on BROKEN` to profiles that carry `Target-Profile-Broken`, i.e. the `DEVICE_*` entries — `TARGET_omap_generic_Default` stays selectable (and with `PRIORITY := 1` it is the default single-profile choice for non-buildbot configs).

So in a plain `CONFIG_TARGET_omap_generic_Default=y` build, `_PROFILE_SET` is non-empty for `ti_am335x-evm` / `ti_am335x-bone-black` → `_TARGET := install-images`, while `u-boot-am335x_evm` is now deselected by `@BROKEN` → [`Build/omap-sdcard`'s `mcopy ... $(STAGING_DIR_IMAGE)/$(DEVICE_NAME)/MLO`](https://github.com/openwrt/openwrt/blob/411b26be38c7ad48a9c36802121e40801fd46435/target/linux/omap/image/Makefile#L19) fails on a missing file. Multi-profile/buildbot builds are unaffected because `DEVICE_CHECK_PROFILE` then looks for `CONFIG_TARGET_DEVICE_omap_generic_Default`, which does not exist.

Detaching the two broken devices from the legacy `Default` profile fixes it:

```suggestion
  BROKEN := y
  PROFILES :=
```

Same applies to line 56 (`ti_am335x-bone-black`). Dropping both from `TARGET_DEVICES` would work too — whichever you prefer.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push of `bd9efaa` → `411b26b`). The only content change since the last round is the removal of `DEFAULT := n` from both devices, which resolves the previous review's point — `scripts/metadata.pm` already sets `default = "n"` for any profile carrying `Target-Profile-Broken`, so it was indeed redundant. The commit message still matches the files touched, and `DEPENDS+=@BROKEN` follows the existing in-tree idiom (`package/boot/uboot-lantiq/Makefile`).

One remaining issue inline, now with a concrete mechanism rather than a question: because omap ties every device to a real legacy `Default` profile, `BROKEN := y` does not exclude these devices from a from-source build, so the new `@BROKEN` on the u-boot package can break `make` for `CONFIG_TARGET_omap_generic_Default`. CI is green, but the CI job builds the omap *kernel* only, not the omap images, so it would not catch this.

---

PR #24503, opened 2026-07-31
airoha: an7581: keep Ethernet working in W1700K recovery

misakazip:
The W1700K Ethernet MAC addresses are provided by NVMEM cells stored in the factory UBI volume. If the UBI layout is missing or damaged, the NVMEM provider is unavailable and the Ethernet driver keeps deferring its probe. The DSA switch consequently cannot probe either, leaving the recovery image without network connectivity.

Add a recovery-specific DTS which removes the NVMEM references from the GDM interfaces. This allows the driver to generate temporary MAC addresses when booting recovery, while the production image continues to use the factory addresses.

Use the recovery DTS only for the initramfs image and keep the regular DTS for the sysupgrade image.

### DIFF
--- a/target/linux/airoha/image/an7581.mk
+++ b/target/linux/airoha/image/an7581.mk
@@ -96,7 +96,7 @@ define Device/gemtek_w1700k-ubi
   DEVICE_ALT2_VENDOR := Quantum Fiber
   DEVICE_ALT2_MODEL := W1700K
   DEVICE_ALT2_VARIANT := UBI
-  DEVICE_DTS := an7581-w1700k-ubi
+  DEVICE_DTS := an7581-w1700k-ubi an7581-w1700k-ubi-recovery
   DEVICE_COMPAT_VERSION := 2.0
   DEVICE_COMPAT_MESSAGE := Partition table has been changed to cooperate \
        with the vendor bootloader with regard to the BMT/BBT partition at \
@@ -111,7 +111,7 @@ define Device/gemtek_w1700k-ubi
   KERNEL_IN_UBI := 1
   KERNEL := kernel-bin | gzip
   KERNEL_INITRAMFS := kernel-bin | lzma | \
-	fit lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb with-initrd | pad-to 128k
+	fit lzma $$(KDIR)/image-$$(lastword $$(DEVICE_DTS)).dtb with-initrd | pad-to 128k
   KERNEL_INITRAMFS_SUFFIX := -recovery.itb
   IMAGES := sysupgrade.itb
   IMAGE/sysupgrade.itb := append-kernel | fit gzip $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb external-static-with-rootfs | append-metadata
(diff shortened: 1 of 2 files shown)


misakazip:
Sorry, I accidentally opened this pull request from my main branch.

I have now created a dedicated feature branch named
`airoha-w1700k-recovery`, but GitHub does not appear to allow changing
the head branch of an existing pull request.

Since the bot asks contributors not to close the pull request, could a
maintainer please allow the current branch with `[allow branch]`?
Apologies for the inconvenience.

target/linux/airoha/image/an7581.mk
```diff
@@ -96,7 +96,7 @@ define Device/gemtek_w1700k-ubi
   DEVICE_ALT2_VENDOR := Quantum Fiber
   DEVICE_ALT2_MODEL := W1700K
   DEVICE_ALT2_VARIANT := UBI
-  DEVICE_DTS := an7581-w1700k-ubi
+  DEVICE_DTS := an7581-w1700k-ubi an7581-w1700k-ubi-recovery
```
openwrt-ai, line 99: nit: `nokia_xg-040g-md-ubi` in this same file looks structurally identical to the W1700K UBI device and should hit exactly the same wall: its [`&gdm1` takes `nvmem-cells = <&macaddr_factory_3e 0>`](https://github.com/openwrt/openwrt/blob/83ed4002312c11e4c51e5cfc4d9b66c381cecdab/target/linux/airoha/dts/an758x-nokia_xg-040g-common.dtsi#L68-L73), and that cell lives [inside the `ri` UBI volume](https://github.com/openwrt/openwrt/blob/83ed4002312c11e4c51e5cfc4d9b66c381cecdab/target/linux/airoha/dts/an758x-nokia_xg-040g-ubi-parts.dtsi#L32-L46) — and it ships the same [`KERNEL_INITRAMFS_SUFFIX := -recovery.itb`](https://github.com/openwrt/openwrt/blob/83ed4002312c11e4c51e5cfc4d9b66c381cecdab/target/linux/airoha/image/an7581.mk#L171-L173) recovery image. Since `airoha_alloc_gdm_port()` propagates `-EPROBE_DEFER` from `of_get_ethdev_address()` for the whole `eth` probe, that board's recovery image should lose Ethernet the same way when the UBI layout is absent/damaged.

Is leaving it out intentional (untested hardware / out of scope), or would it be worth the same treatment in this series? (`an7583-nokia_xg-040g-mf` reads the MAC from a plain MTD partition, so it isn't affected.)

---

openwrt-ai (commented):
Reviewed 1 new commit.

The mechanism checks out: with `nvmem-cells`/`nvmem-cell-names` removed, `of_get_mac_address()` falls through to `of_get_address_nvmem()` which returns `-ENOENT` rather than `-EPROBE_DEFER`, so `airoha_alloc_gdm_port()` takes the `eth_hw_addr_random()` path instead of aborting the whole `eth` probe. `&gdm1`/`&gdm2`/`&gdm4` are exactly the three GDM nodes carrying nvmem MAC references in `an7581-w1700k-ubi.dts`. Multiple entries in `DEVICE_DTS` are supported (`Device/Build/kernel` iterates the list to emit one `$(KDIR)/image-<dts>.dtb` rule each), and both the `KERNEL` and `IMAGE/sysupgrade.itb` recipes still pin `firstword`, so the flashed image keeps the factory MACs. The `Fixes:` trailer resolves to the right commit and subject.

One parity question inline.

---

PR #24505, opened 2026-07-31
openssl: add MIPS32 AES-CTR assembly

dormancygrace:
## What changed

- add an O32 `AES_ctr32_encrypt` loop to OpenSSL's existing MIPS AES assembly
- enable `AES_CTR_ASM` for MIPS32 while leaving the MIPS64 flavours unchanged
- bump the OpenWrt OpenSSL package release

The loop saves its register frame and resolves the AES table once per batch, then reuses the existing `_mips_AES_encrypt` core. It supports unaligned and in-place buffers and maintains the private network-order counter expected by `ctr128_f`. The private counter is above the O32 outgoing argument-save area, and the routine uses the existing perlasm register, pointer, load/store, frame-mask and address macros.

Every current OpenSSL target with `asm_arch=mips32` resolves to the O32 perlasm scheme, including the OpenWrt targets that inherit `linux-mips32`. This matches the emission guard in `aes-mips.pl`. The invariant is documented next to `AES_CTR_ASM` so a future non-O32 MIPS32 target cannot silently reference an omitted symbol.

## Why

Without `AES_ctr32_encrypt`, AES-GCM calls the full single-block O32 AES wrapper for every counter block. The added batch loop avoids that repeated ABI and table-setup overhead.

On a 580 MHz MT7620A, five alternating 1500-byte packet-style AES-GCM runs changed as follows. Both sides use the same GHASH implementation, so these numbers isolate the AES-CTR change.

| Cipher | Before | After | Change |
|---|---:|---:|---:|
| AES-128-GCM | 4.662 MB/s | 4.880 MB/s | +4.68% |
| AES-256-GCM | 4.012 MB/s | 4.169 MB/s | +3.91% |

Stock and assembly output digests matched in every run.

## Validation

- OpenWrt `git diff --check` and quilt refresh
- all OpenWrt package patches apply to OpenSSL 3.5.8 with `--fuzz=0`; complete MIPS32/musl cross-build with the OpenWrt GCC 16.2.0 toolchain passes, and the resulting `libcrypto.a` defines `AES_ctr32_encrypt` with `AES_CTR_ASM` enabled
- complete `linux-mips32` cross-build of current OpenSSL `master`
- little-endian differential AES-128/192/256 tests covering zero and multiple block counts, unaligned buffers, in-place operation and low-counter carry under `qemu-mipsel` and on a physical MT7620A
- the same differential coverage with the official OpenWrt Malta big-endian MIPS32/O32 GCC 14.4.0 toolchain under `qemu-mips`
- big-endian four-block NIST SP 800-38A AES-128-CTR known-answer test
- verified on both endiannesses that the private `ctr128_f` counter copy changes while the caller-owned IV remains unchanged
- generated object disassembly checked for the O32 frame and argument layout

## Upstream

Submitted to OpenSSL as https://github.com/openssl/openssl/pull/32133. The downstream patch header records the same link.

### DIFF
--- a/package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
+++ b/package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
@@ -1,0 +1,170 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Gleb Pesin <dormancygrace@gmail.com>
+Date: Fri, 31 Jul 2026 12:00:00 +0000
+Subject: [PATCH] crypto: mips32: add AES CTR32 assembly path
+
+The MIPS AES assembly only exports the single-block AES core. AES-GCM
+therefore enters the full O32 function wrapper for every counter block.
+
+Add an O32 AES_ctr32_encrypt implementation which saves the register
+frame and resolves the AES table once per batch. Reuse the existing
+_mips_AES_encrypt core, handle unaligned and in-place buffers, and keep
+the private counter in network byte order as required by ctr128_f.
+
+Enable AES_CTR_ASM for MIPS32 only. The implementation uses the O32 ABI
+and is not suitable for the MIPS64 flavours.
+
+Upstream-Status: Submitted [https://github.com/openssl/openssl/pull/32133]
+Signed-off-by: Gleb Pesin <dormancygrace@gmail.com>
+
+--- a/crypto/aes/build.info
++++ b/crypto/aes/build.info
+@@ -21,9 +21,11 @@ IF[{- !$disabled{asm} -}]
+   $AESDEF_sparcv9=AES_ASM
+ 
+   $AESASM_mips32=aes_cbc.c aes-mips.S
+-  $AESDEF_mips32=AES_ASM
++  # Every asm_arch=mips32 target uses the O32 perlasm scheme, matching
++  # the AES_ctr32_encrypt emission guard in aes-mips.pl.
++  $AESDEF_mips32=AES_ASM AES_CTR_ASM
+   $AESASM_mips64=$AESASM_mips32
+-  $AESDEF_mips64=$AESDEF_mips32
++  $AESDEF_mips64=AES_ASM
+ 
+   $AESASM_s390x=aes-s390x.S
+   # aes-390x.S implements AES_ctr32_encrypt and AES_xts_[en|de]crypt
+--- a/crypto/aes/asm/aes-mips.pl
++++ b/crypto/aes/asm/aes-mips.pl
+@@ -731,6 +731,132 @@ $code.=<<___;
+ .end	AES_encrypt
+ ___
+ 
++my $CTR32_FRAMESIZE=24*$SZREG;
++my $CTR32_BLOCK_OFFSET=6*$SZREG;
++my $CTR32_COUNT_OFFSET=10*$SZREG;
++my $CTR32_IVEC_OFFSET=$CTR32_FRAMESIZE+4*$SZREG;
++
++$code.=<<___ if ($flavour =~ /o32/i);
++.align	5
++.globl	AES_ctr32_encrypt
++.ent	AES_ctr32_encrypt
++AES_ctr32_encrypt:
++	.frame	$sp,$CTR32_FRAMESIZE,$ra
++	.mask	$SAVED_REGS_MASK,-$SZREG
++	.set	noreorder
++	.cpload	$pf
++	$PTR_SUB $sp,$CTR32_FRAMESIZE
++	$REG_S	$ra,$CTR32_FRAMESIZE-1*$SZREG($sp)
++	$REG_S	$fp,$CTR32_FRAMESIZE-2*$SZREG($sp)
++	$REG_S	$s11,$CTR32_FRAMESIZE-3*$SZREG($sp)
++	$REG_S	$s10,$CTR32_FRAMESIZE-4*$SZREG($sp)
++	$REG_S	$s9,$CTR32_FRAMESIZE-5*$SZREG($sp)
++	$REG_S	$s8,$CTR32_FRAMESIZE-6*$SZREG($sp)
++	$REG_S	$s7,$CTR32_FRAMESIZE-7*$SZREG($sp)
++	$REG_S	$s6,$CTR32_FRAMESIZE-8*$SZREG($sp)
++	$REG_S	$s5,$CTR32_FRAMESIZE-9*$SZREG($sp)
++	$REG_S	$s4,$CTR32_FRAMESIZE-10*$SZREG($sp)
++	.set	reorder
++
++	# O32 passes ivec as the fifth argument on the caller's stack.
++	$REG_L	$t0,$CTR32_IVEC_OFFSET($sp)
++	$REG_S	$key,$CTR32_COUNT_OFFSET($sp) # preserve the block count
++	move	$key,$Tbl                      # key schedule for the inner core
++	$PTR_LA	$Tbl,AES_Te
++
++	# Private counter block; ctr128_f callers update ivec themselves.
++	lwl	$s0,0+$MSB($t0)
++	lwl	$s1,4+$MSB($t0)
++	lwl	$s2,8+$MSB($t0)
++	lwl	$s3,12+$MSB($t0)
++	lwr	$s0,0+$LSB($t0)
++	lwr	$s1,4+$LSB($t0)
++	lwr	$s2,8+$LSB($t0)
++	lwr	$s3,12+$LSB($t0)
++	$REG_S	$s0,$CTR32_BLOCK_OFFSET+0($sp)
++	$REG_S	$s1,$CTR32_BLOCK_OFFSET+4($sp)
++	$REG_S	$s2,$CTR32_BLOCK_OFFSET+8($sp)
++	$REG_S	$s3,$CTR32_BLOCK_OFFSET+12($sp)
++
++	$REG_L	$t0,$CTR32_COUNT_OFFSET($sp)
++	beqz	$t0,.Lctr32_done
++
++.Loop_ctr32_enc:
++	$REG_L	$s0,$CTR32_BLOCK_OFFSET+0($sp)
++	$REG_L	$s1,$CTR32_BLOCK_OFFSET+4($sp)
++	$REG_L	$s2,$CTR32_BLOCK_OFFSET+8($sp)
++	$REG_L	$s3,$CTR32_BLOCK_OFFSET+12($sp)
++
++	bal	_mips_AES_encrypt
++
++	# Load plaintext before storing, so in-place operation is safe.
++	lwl	$t0,0+$MSB($inp)
++	lwl	$t1,4+$MSB($inp)
++	lwl	$t2,8+$MSB($inp)
++	lwl	$t3,12+$MSB($inp)
++	lwr	$t0,0+$LSB($inp)
++	lwr	$t1,4+$LSB($inp)
++	lwr	$t2,8+$LSB($inp)
++	lwr	$t3,12+$LSB($inp)
++	xor	$s0,$t0
++	xor	$s1,$t1
++	xor	$s2,$t2
++	xor	$s3,$t3
++
++	swr	$s0,0+$LSB($out)
++	swr	$s1,4+$LSB($out)
++	swr	$s2,8+$LSB($out)
++	swr	$s3,12+$LSB($out)
++	swl	$s0,0+$MSB($out)
++	swl	$s1,4+$MSB($out)
++	swl	$s2,8+$MSB($out)
++	swl	$s3,12+$MSB($out)
++
++	# Increment the low 32-bit counter in big-endian byte order.
++	lbu	$t0,$CTR32_BLOCK_OFFSET+15($sp)
++	addiu	$t0,1
++	andi	$t0,0xff
++	sb	$t0,$CTR32_BLOCK_OFFSET+15($sp)
++	bnez	$t0,.Lctr32_no_carry
++	lbu	$t0,$CTR32_BLOCK_OFFSET+14($sp)
++	addiu	$t0,1
++	andi	$t0,0xff
++	sb	$t0,$CTR32_BLOCK_OFFSET+14($sp)
++	bnez	$t0,.Lctr32_no_carry
++	lbu	$t0,$CTR32_BLOCK_OFFSET+13($sp)
++	addiu	$t0,1
++	andi	$t0,0xff
++	sb	$t0,$CTR32_BLOCK_OFFSET+13($sp)
++	bnez	$t0,.Lctr32_no_carry
++	lbu	$t0,$CTR32_BLOCK_OFFSET+12($sp)
++	addiu	$t0,1
++	andi	$t0,0xff
++	sb	$t0,$CTR32_BLOCK_OFFSET+12($sp)
++
++.Lctr32_no_carry:
++	$PTR_ADD $inp,16
++	$PTR_ADD $out,16
++	$REG_L	$t0,$CTR32_COUNT_OFFSET($sp)
++	addiu	$t0,-1
++	$REG_S	$t0,$CTR32_COUNT_OFFSET($sp)
++	bnez	$t0,.Loop_ctr32_enc
++
++.Lctr32_done:
++	.set	noreorder
++	$REG_L	$ra,$CTR32_FRAMESIZE-1*$SZREG($sp)
++	$REG_L	$fp,$CTR32_FRAMESIZE-2*$SZREG($sp)
++	$REG_L	$s11,$CTR32_FRAMESIZE-3*$SZREG($sp)
++	$REG_L	$s10,$CTR32_FRAMESIZE-4*$SZREG($sp)
++	$REG_L	$s9,$CTR32_FRAMESIZE-5*$SZREG($sp)
++	$REG_L	$s8,$CTR32_FRAMESIZE-6*$SZREG($sp)
++	$REG_L	$s7,$CTR32_FRAMESIZE-7*$SZREG($sp)
++	$REG_L	$s6,$CTR32_FRAMESIZE-8*$SZREG($sp)
++	$REG_L	$s5,$CTR32_FRAMESIZE-9*$SZREG($sp)
++	$REG_L	$s4,$CTR32_FRAMESIZE-10*$SZREG($sp)
++	jr	$ra
++	$PTR_ADD $sp,$CTR32_FRAMESIZE
++.end	AES_ctr32_encrypt
++___
+ $code.=<<___;
+ .align	5
+ .ent	_mips_AES_decrypt
--- a/package/libs/openssl/Makefile
+++ b/package/libs/openssl/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openssl
 PKG_VERSION:=3.5.8
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_BUILD_FLAGS:=no-mips16 gc-sections no-lto
 
 PKG_BUILD_PARALLEL:=1
--- a/package/libs/openssl/Makefile
+++ b/package/libs/openssl/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=openssl
 PKG_VERSION:=3.5.7
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 PKG_BUILD_FLAGS:=no-mips16 gc-sections no-lto
 
 PKG_BUILD_PARALLEL:=1


package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
```diff
@@ -0,0 +1,158 @@
...
+   $AESASM_mips64=$AESASM_mips32
+-  $AESDEF_mips64=$AESDEF_mips32
++  $AESDEF_mips64=AES_ASM
++
+-
+   $AESASM_s390x=aes-s390x.S
+   # aes-390x.S implements AES_ctr32_encrypt and AES_xts_[en|de]crypt
+--- a/crypto/aes/asm/aes-mips.pl
++++ b/crypto/aes/asm/aes-mips.pl
+@@ -733,0 +734,121 @@
```
openwrt-ai: This hunk carries zero context lines (`-733,0 +734,121`), so `patch` has nothing to anchor on — it will insert the 121 lines at whatever offset it finds and can never report a conflict. On the next OpenSSL version bump the block can silently land inside an unrelated `$code.=<<___;` heredoc in `aes-mips.pl` and still "apply cleanly".

Every other patch in `package/libs/openssl/patches/` is quilt-refreshed with 3 lines of context and a function-context suffix in the hunk header (e.g. [`@@ -52,10 +52,13 @@ tsa_policy3 = 1.2.3.4.5.7` in 150-openssl.cnf-add-engines-conf.patch:13](``https://github.com/openwrt/openwrt/blob/a0598bd3a027d9afce013e9b575dbad241f8aa26/package/libs/openssl/patches/150-openssl.cnf-add-engines-conf.patch#L13)).`` Please regenerate this patch with `make package/openssl/refresh` so both hunks get normal context.

---

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
```diff
@@ -0,0 +1,158 @@
...
+-
++
+   $AESASM_mips32=aes_cbc.c aes-mips.S
+-  $AESDEF_mips32=AES_ASM
++  $AESDEF_mips32=AES_ASM AES_CTR_ASM
+   $AESASM_mips64=$AESASM_mips32
+-  $AESDEF_mips64=$AESDEF_mips32
++  $AESDEF_mips64=AES_ASM
++
+-
```
openwrt-ai: Four of the ten lines in this hunk are no-ops: line 23/24 delete an empty line and add an identical empty line back, and line 31/32 do the same. They change nothing in `build.info` but make the hunk fail if the surrounding blank lines ever move.

Line 31/32 are also emitted in `+` then `-` order, which no unified-diff generator produces (a change block always lists all `-` lines before all `+` lines). Together with the bare `-`/`+` markers — in-tree patches write a blank context line as `" "`, not as an empty line — this hunk looks hand-written rather than generated.

Dropping the four lines leaves the hunk counts at 9/9, so the `@@ -21,9 +21,9 @@` header does not change; `make package/openssl/refresh` will produce this form.

---

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
```diff
@@ -0,0 +1,158 @@
...
++        sw      $s6,64($sp)
++        sw      $s5,60($sp)
++        sw      $s4,56($sp)
++        .set    reorder
++
++        # O32 passes ivec as the fifth argument on the caller's stack.
++        lw      $t0,112($sp)
++        sw      $key,16($sp)          # preserve the block count
++        move    $key,$Tbl             # AES key schedule for the inner core
++        la      $Tbl,AES_Te
```
openwrt-ai: `$key` and `$Tbl` are set once, before `.Loop_ctr32_enc`, but `_mips_AES_encrypt` is entered once per block. Does `_mips_AES_encrypt` preserve `$key`? The usual shape of that core is to load `240($key)` for the round count and then walk the schedule with `$PTR_ADD $key,16` per round without restoring it — it has only ever had one caller (`AES_encrypt`), which calls it exactly once and therefore never needed `$key` to survive. If that is the case here, every block after the first encrypts with a `$key` pointing past the end of the schedule, and only single-block calls produce correct output.

The counter is already re-loaded from the stack each iteration (lines 84-87) because the core clobbers `$s0`-`$s3`; `$key` needs the same treatment unless the core is verified to preserve it — e.g. stash the incoming `$a3` in a spare frame slot and `lw $key,<slot>($sp)` at the top of `.Loop_ctr32_enc`. The same question applies to `$inp` (`$a0`): `AES_encrypt` uses `$out` after the `bal`, which proves `$a1` survives, but it never touches `$inp` after the call, so `$a0` surviving is unverified.

Since the PR claims multi-block differential tests passed, please say which OpenSSL revision of `crypto/aes/asm/aes-mips.pl` you verified the register contract against.

---

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
```diff
@@ -0,0 +1,158 @@
...
++        lwl     $s2,8+$MSB($t0)
++        lwl     $s3,12+$MSB($t0)
++        lwr     $s0,0+$LSB($t0)
++        lwr     $s1,4+$LSB($t0)
++        lwr     $s2,8+$LSB($t0)
++        lwr     $s3,12+$LSB($t0)
++        sw      $s0,0($sp)
++        sw      $s1,4($sp)
++        sw      $s2,8($sp)
++        sw      $s3,12($sp)
```
openwrt-ai: The private counter block is placed at `0($sp)`-`15($sp)`, which under O32 is the outgoing argument-save area — it belongs to whatever this frame calls, and `.Loop_ctr32_enc` does `bal _mips_AES_encrypt` with the counter live in it. It happens to survive only as long as the callee never spills its incoming arguments, which is not something the ABI guarantees and not something a future edit of `aes-mips.pl` will preserve.

Grow the frame and move the counter block (and the block count at `16($sp)`) above the 16-byte argument area, e.g. counter at `24($sp)` and count at `40($sp)`. Note the routine reads its own fifth argument at `112($sp)` = `16 + 96`, so any frame-size change has to be reflected there too.

---

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
```diff
@@ -0,0 +1,158 @@
...
++$code.=<<___ if ($flavour =~ /o32/i);
++.align  5
++.globl  AES_ctr32_encrypt
++.ent    AES_ctr32_encrypt
++AES_ctr32_encrypt:
++        .frame  $sp,96,$ra
++        .mask   0xc0ff0000,-4
++        .set    noreorder
++        .cpload $pf
++        subu    $sp,96
```
openwrt-ai: nit: the frame size, the saved-register mask and the load/store/add mnemonics are hardcoded here, while every other routine in `aes-mips.pl` goes through the file's own `$FRAMESIZE`, `$SAVED_REGS_MASK`, `$REG_S`/`$REG_L`, `$PTR_SUB`/`$PTR_ADD` variables. `0xc0ff0000` is the non-NUBI value of `$SAVED_REGS_MASK`, so this silently hardcodes one flavour's answer even though the surrounding file parameterises it.

The `$flavour =~ /o32/i` guard does keep `nubi32` out (that string has no `o32` in it), so this is a consistency point rather than a correctness one — but using the existing variables keeps the block reviewable next to its neighbours.

---

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
```diff
@@ -0,0 +1,158 @@
...
+
+Add an O32 AES_ctr32_encrypt implementation which saves the register
+frame and resolves the AES table once per batch. Reuse the existing
+_mips_AES_encrypt core, handle unaligned and in-place buffers, and keep
+the private counter in network byte order as required by ctr128_f.
+
+Enable AES_CTR_ASM for MIPS32 only. The implementation uses the O32 ABI
+and is not suitable for the MIPS64 flavours.
+
+Signed-off-by: Gleb Pesin <dormancygrace@gmail.com>
```
openwrt-ai, line 18: nit: this header has no upstream status. Unlike the other patches here — [010-fix-aes-gcm-siv-cipher.patch:16-18](`https://github.com/openwrt/openwrt/blob/a0598bd3a027d9afce013e9b575dbad241f8aa26/package/libs/openssl/patches/010-fix-aes-gcm-siv-cipher.patch#L16-L18`) names the upstream commit it relates to — this is a brand-new, hand-written crypto primitive that OpenWrt would carry downstream indefinitely and rebase across every OpenSSL bump, with a wrong result meaning silent ciphertext corruption on every MIPS32 device rather than a build failure.

Please submit it to OpenSSL first (it is generic MIPS32 code, nothing OpenWrt-specific) and reference the upstream PR/issue in this header, so the patch has a review trail and an exit path.

---

openwrt-ai (commented):
Reviewed 1 new commit; see inline comments.

---

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
dormancygrace: Fixed. I applied the patch through quilt and ran `make package/openssl/refresh`. The refreshed `aes-mips.pl` hunk now has three context lines and the function-context suffix (`@@ -731,6 +731,132 @@ $code.=<<___;`), so future source changes can produce a real conflict.

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
dormancygrace: Verified against both OpenSSL 3.5.7 and current upstream `master`. In this generator `$key` is the incoming `$a2`, while `_mips_AES_encrypt` walks the schedule through the separate `$key0 = $gp`; it does not modify `$key`. The inner core also does not modify `$inp = $a0`. I re-ran the differential harness for AES-128/192/256 with multi-block counts up to 257, unaligned and in-place buffers, and counter carry under `qemu-mipsel` and on MT7620A hardware; both passed.

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
dormancygrace: Fixed by the same quilt refresh. The `build.info` hunk now contains only the two actual `AESDEF` changes with normal context; the no-op blank-line replacements are gone.

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
dormancygrace: Fixed. The routine now uses `$SAVED_REGS_MASK`, `$REG_S`/`$REG_L`, `$PTR_SUB`/`$PTR_ADD`, and `$PTR_LA`, with named frame, counter, count, and fifth-argument offsets. The O32-only flavour guard remains explicit.

neheb:
that is...nice.

neheb:
https://portal.cfarm.net/machines/list/ if you want to waste time on mips64.

dormancygrace:
Thanks, I will keep CFarm in mind for broader architecture testing. This particular change is intentionally MIPS32-only and the real-hardware benchmarks are on MT7620/24Kc, so MIPS64 would be a separate follow-up rather than a validation target for this patch.

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
```diff
@@ -0,0 +1,168 @@
...
++	swl	$s0,0+$MSB($out)
++	swl	$s1,4+$MSB($out)
++	swl	$s2,8+$MSB($out)
++	swl	$s3,12+$MSB($out)
++
++	# Increment the low 32-bit counter in big-endian byte order.
++	lbu	$t0,$CTR32_BLOCK_OFFSET+15($sp)
++	addiu	$t0,1
++	andi	$t0,0xff
++	sb	$t0,$CTR32_BLOCK_OFFSET+15($sp)
```
openwrt-ai, line 127: Every runtime validation listed in the commit message and the PR body is little-endian: `qemu-mipsel` and a physical MT7620A (ramips, mipsel). But `AES_CTR_ASM` is enabled for the whole `mips32` asm arch, and OpenWrt routes big-endian MIPS there too — [`110-openwrt_targets.patch:33-34`](https://github.com/openwrt/openwrt/blob/7a6e409204b5bd3a4a2de23923dbdd7e20297202/package/libs/openssl/patches/110-openwrt_targets.patch#L33-L34) maps `linux-mips-openwrt` onto `linux-mips32`, [`target/linux/ath79/Makefile:3`](https://github.com/openwrt/openwrt/blob/7a6e409204b5bd3a4a2de23923dbdd7e20297202/target/linux/ath79/Makefile#L3) is `ARCH:=mips`, and [`OPENSSL_WITH_ASM` defaults to y](https://github.com/openwrt/openwrt/blob/7a6e409204b5bd3a4a2de23923dbdd7e20297202/package/libs/openssl/Config.in#L37-L39). So ath79 — the largest MIPS32 device family in the tree — gets this routine with no runtime coverage at all. CI only proves it assembles on malta/be.

This block is precisely where endianness becomes load-bearing. The counter is written into the frame as words with `$REG_S` (lines 82-85) from registers filled through the `$MSB`/`$LSB` pair, and then read back one byte at a time at fixed offsets +15..+12. That only lines up with the network-order counter because `$MSB`/`$LSB` are swapped for little-endian builds, so the stack image reproduces `ivec[0..15]` in byte order on both flavours. The invariant holds, but it is implicit — nothing in the routine states it, and a mistake here is silent ciphertext corruption on every AES-GCM/AES-CTR session rather than a build failure.

Please run the differential/KAT harness on a big-endian mips32 build (malta/be under `qemu-mips` is enough) and record that alongside the mipsel results in the commit message.

---

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
```diff
@@ -0,0 +1,168 @@
...
++	$REG_S	$s4,$CTR32_FRAMESIZE-10*$SZREG($sp)
++	.set	reorder
++
++	# O32 passes ivec as the fifth argument on the caller's stack.
++	$REG_L	$t0,$CTR32_IVEC_OFFSET($sp)
++	$REG_S	$key,$CTR32_COUNT_OFFSET($sp) # preserve the block count
++	move	$key,$Tbl                      # key schedule for the inner core
++	$PTR_LA	$Tbl,AES_Te
++
++	# Keep the counter above the O32 outgoing argument-save area.
```
openwrt-ai: nit: the quilt refresh dropped the one comment a future reader actually needs here. The previous revision read `# Keep a private counter block. The caller updates ivec itself.`, which documented the `ctr128_f` contract — this routine deliberately never writes the incremented counter back to `ivec`, because `CRYPTO_ctr128_encrypt_ctr32` / `CRYPTO_gcm128_*_ctr32` do `PUTU32(ivec + 12, ctr32)` themselves after the call. That is the single most surprising property of the function, and it is now undocumented.

The replacement comment restates frame layout, which `$CTR32_BLOCK_OFFSET` (line 41) already conveys. Keeping it to one line preserves the `@@ -731,6 +731,132 @@` counts:

```suggestion
+	# Private counter block; ctr128_f callers update ivec themselves.
```

---

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
```diff
@@ -0,0 +1,168 @@
...
+Signed-off-by: Gleb Pesin <dormancygrace@gmail.com>
+
+--- a/crypto/aes/build.info
++++ b/crypto/aes/build.info
+@@ -21,9 +21,9 @@ IF[{- !$disabled{asm} -}]
+   $AESDEF_sparcv9=AES_ASM
+ 
+   $AESASM_mips32=aes_cbc.c aes-mips.S
+-  $AESDEF_mips32=AES_ASM
++  $AESDEF_mips32=AES_ASM AES_CTR_ASM
```
openwrt-ai, line 29: These two lines key on different things. `$AESDEF_mips32` is selected by the target's `asm_arch` (`mips32`), whereas the routine that provides the symbol is emitted only when the *perlasm scheme* matches o32 (`$code.=<<___ if ($flavour =~ /o32/i);`, line 45). Today only `linux-mips32` sets `asm_arch => mips32` and it does use `o32`, so the two coincide — but any future target pairing `asm_arch=mips32` with a non-o32 scheme would get `AES_CTR_ASM` defined with no `AES_ctr32_encrypt` in `aes-mips.S`, i.e. an undefined reference rather than a fallback to the C `ctr128_f`.

Same wording issue in the commit message, which says "Enable AES_CTR_ASM only for the O32 MIPS32 target": the define is not O32-scoped, only the code generation is. Worth either saying so explicitly in the commit message, or gating this define on the ABI if `build.info` can express it — the s390x comment two lines below (`# aes-390x.S implements AES_ctr32_encrypt ...`) is the precedent for at least documenting the coupling.

---

openwrt-ai (commented):
Re-reviewed at `7a6e409` (previous review was at `a0598bd`). The quilt refresh addressed the earlier hunk-context, no-op-hunk, frame-layout and perlasm-macro points, and the register-contract answer (`$key0`/`$cnt` are separate from `$key`, `$inp`/`$out` untouched by the inner core) is consistent with the shape of `AES_encrypt`'s own prologue that this routine mirrors. CI is green, including a malta/be (big-endian mips32) package build.

Remaining inline comments below.

---

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
dormancygrace: Tested now with the official OpenWrt Malta big-endian MIPS32/O32 GCC 14.4.0 toolchain under qemu-mips. The AES-128/192/256 differential harness passes for zero through 257 blocks, unaligned and in-place buffers, and counter carry. A four-block NIST SP 800-38A AES-128-CTR KAT also passes, and the caller-owned IV remains unchanged as required by ctr128_f. I added this evidence to the commit message and PR description in 44c97d9c39.

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
dormancygrace: Restored a concise comment in 44c97d9c39: this is a private counter block and ctr128_f callers update ivec themselves.

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
dormancygrace: Documented the coupling next to AES_CTR_ASM and in the commit message in 44c97d9c39. linux-mips32 is currently the only asm_arch=mips32 target and uses the O32 perlasm scheme, matching the aes-mips.pl emission guard. A future non-O32 mips32 target will therefore have an explicit warning at the define rather than silently selecting an omitted symbol.

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
```diff
@@ -0,0 +1,170 @@
...
+
+--- a/crypto/aes/build.info
++++ b/crypto/aes/build.info
+@@ -21,9 +21,11 @@ IF[{- !$disabled{asm} -}]
+   $AESDEF_sparcv9=AES_ASM
+ 
+   $AESASM_mips32=aes_cbc.c aes-mips.S
+-  $AESDEF_mips32=AES_ASM
++  # linux-mips32 is the only asm_arch=mips32 target and uses O32,
++  # matching the AES_ctr32_encrypt emission guard in aes-mips.pl.
```
openwrt-ai: `linux-mips32` is not the only `asm_arch=mips32` target, so the premise of this comment is false — which matters because the comment exists specifically to keep a future reader from breaking the invariant.

- Upstream, `Configurations/15-android.conf` defines `android-mips` with `asm_arch => 'mips32'` and `perlasm_scheme => "o32"` (verified by code search against `openssl/openssl` master; `10-main.conf` has `linux-mips32`, `15-android.conf` has `android-mips`). This one matters for the upstream submission, since `AES_CTR_ASM` would now be defined for `android-mips` builds too.
- In this tree, [`110-openwrt_targets.patch:33-44`](https://github.com/openwrt/openwrt/blob/44c97d9c39659398ca00e5a0c6430026f2c3d9e0/package/libs/openssl/patches/110-openwrt_targets.patch#L33-L44) adds two more: `linux-mips-openwrt` and `linux-mipsel-openwrt`, both `inherit_from => [ "linux-mips32" ]` — those are the targets OpenWrt actually configures with, not `linux-mips32` itself.

The conclusion is unchanged (all four resolve to the o32 perlasm scheme, so the emission guard in `aes-mips.pl` is satisfied), but stating the invariant instead of a uniqueness claim keeps the comment true as targets are added. This keeps the line count at 2, so `@@ -21,9 +21,11 @@` still holds:

```suggestion
+  # Every asm_arch=mips32 target uses the O32 perlasm scheme, matching
+  # the AES_ctr32_encrypt emission guard in aes-mips.pl.
```

The same uniqueness claim appears twice more and should be reworded alongside it: the commit message ("The only current target with asm_arch=mips32 is linux-mips32, whose perlasm scheme is O32") and the PR body ("`linux-mips32` is currently the only OpenSSL target with `asm_arch=mips32`").

---

openwrt-ai (commented):
Re-reviewed the delta `7a6e409..44c97d9` (previous review was at `7a6e409`). The delta is documentation-only: the `build.info` hunk gains a two-line comment (hunk header correctly rebalanced `@@ -21,9 +21,9 @@` → `@@ -21,9 +21,11 @@`), and the `ctr128_f`/ivec comment is restored in the `aes-mips.pl` hunk as suggested. Both earlier points are addressed, and the big-endian validation evidence is now recorded in the commit message. CI is green on `44c97d9`, including the malta/be package build.

One inline comment on the new `build.info` comment text.

---

package/libs/openssl/patches/160-mips32-add-aes-ctr32-assembly.patch
dormancygrace: Fixed in bc855c5b48. The patch comment now states the actual O32 invariant for every asm_arch=mips32 target, and I updated both the commit message and PR body to cover the OpenWrt targets inheriting linux-mips32. I also refreshed the quilt series and rechecked the diff.

peterwillcn (approved):

openwrt-ai (commented):

Re-reviewed the delta `44c97d9..bc855c5` (previous review was at `44c97d9`). The only change is the two-line `build.info` comment, which now states the O32 invariant instead of the false uniqueness claim, and the same rewording is reflected in the commit message ("Every current target with asm_arch=mips32 resolves to the O32 perlasm scheme, including the OpenWrt targets that inherit linux-mips32") and the PR body. The comment stays two lines, so `@@ -21,9 +21,11 @@` still balances. `PKG_RELEASE` bump and `AES_CTR_ASM`/`AESDEF_mips64` changes are unchanged from the previously reviewed revision, and CI is green on `bc855c5` including the malta/be package build.

---

neheb:
ping @jonasjelonek

champtar:
I would say let's wait a bit for comments in https://github.com/openssl/openssl/pull/32133

dormancygrace:
Agreed. I will keep this downstream PR parked until openssl/openssl#32133 has had time for upstream review, and will sync any resulting changes back here.

openwrt-ai (commented):

---

PR #24506, opened 2026-07-31
ramips: add support for Ruijie RG-EG105G-V3

Mr-TopG:
## Device information

* Device: Ruijie RG-EG105G-V3
* SoC: MediaTek MT7621
* RAM: 128 MiB
* Flash: 32 MiB SPI NOR
* Ethernet: 4 LAN ports and 1 WAN port through the integrated MT7530 switch

## Changes

This adds support for the Ruijie RG-EG105G-V3, including:

* Device tree and flash partition layout
* Ethernet switch and port configuration
* Factory MAC address assignment
* OpenWrt image definition
* U-Boot environment configuration
* Sysupgrade handling for the vendor firmware metadata

The vendor bootloader checks `firmware_size` and `firmware_crc32` variables stored in the U-Boot environment. These variables are cleared before sysupgrade so that firmware written by OpenWrt boots normally.

## Installation

The initial installation is performed through the U-Boot TFTP recovery interface using menu option 2 and the OpenWrt squashfs sysupgrade image renamed to `rgos.bin`.

## Tested

* U-Boot TFTP installation
* Booting from SPI flash
* All four LAN ports
* WAN port at Gigabit speed
* DHCP and routing
* LuCI
* Command-line sysupgrade
* LuCI sysupgrade
* Configuration preservation
* Factory MAC addresses

### DIFF
--- a/target/linux/ramips/dts/mt7621_ruijie_rg-eg105g-v3.dts
+++ b/target/linux/ramips/dts/mt7621_ruijie_rg-eg105g-v3.dts
@@ -1,0 +1,86 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+#include "mt7621.dtsi"
+
+/ {
+	compatible = "ruijie,rg-eg105g-v3", "mediatek,mt7621-soc";
+	model = "Ruijie RG-EG105G-V3";
+
+	chosen {
+		bootargs = "console=ttyS0,57600n8";
+	};
+};
+
+&spi0 {
+	status = "okay";
+
+	flash@0 {
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <25000000>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "u-boot";
+				reg = <0x000000 0x050000>;
+				read-only;
+			};
+
+			partition@50000 {
+				label = "u-boot-env";
+				reg = <0x050000 0x010000>;
+			};
+
+			partition@60000 {
+				label = "product_info";
+				reg = <0x060000 0x010000>;
+				read-only;
+			};
+
+			partition@70000 {
+				label = "reserved";
+				reg = <0x070000 0x010000>;
+				read-only;
+			};
+
+			partition@80000 {
+				compatible = "denx,uimage";
+				label = "firmware";
+				reg = <0x080000 0x1f80000>;
+			};
+		};
+	};
+};
+
+&switch0 {
+	ports {
+		port@0 {
+			status = "okay";
+			label = "lan1";
+		};
+
+		port@1 {
+			status = "okay";
+			label = "lan2";
+		};
+
+		port@2 {
+			status = "okay";
+			label = "lan3";
+		};
+
+		port@3 {
+			status = "okay";
+			label = "lan4";
+		};
+
+		port@4 {
+			status = "okay";
+			label = "wan";
+		};
+	};
+};
--- a/target/linux/ramips/image/mt7621.mk
+++ b/target/linux/ramips/image/mt7621.mk
@@ -2767,6 +2767,14 @@ define Device/rostelecom_rt-sf-1
 endef
 TARGET_DEVICES += rostelecom_rt-sf-1
 
+define Device/ruijie_rg-eg105g-v3
+  DEVICE_VENDOR := Ruijie
+  DEVICE_MODEL := RG-EG105G-V3
+  IMAGE_SIZE := 32256k
+  DEVICE_PACKAGES := uboot-envtools
+endef
+TARGET_DEVICES += ruijie_rg-eg105g-v3
+
 define Device/ruijie_rg-ew1200g-pro-v1.1
   $(Device/dsa-migration)
   $(Device/uimage-lzma-loader)
--- a/target/linux/ramips/mt7621/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/mt7621/base-files/etc/board.d/02_network
@@ -289,6 +289,11 @@ ramips_setup_macs()
 		wan_mac=$(macaddr_add "$lan_mac" 8)
 		label_mac=$lan_mac
 		;;
+	ruijie,rg-eg105g-v3)
+		lan_mac=$(mtd_get_mac_ascii product_info ethaddr)
+		wan_mac=$(macaddr_add "$lan_mac" 2)
+		label_mac=$lan_mac
+		;;
 	mts,wg430223)
 		wan_mac=$(mtd_get_mac_encrypted_arcadyan "board_data")
 		label_mac=$wan_mac
--- a/target/linux/ramips/mt7621/base-files/lib/upgrade/platform.sh
+++ b/target/linux/ramips/mt7621/base-files/lib/upgrade/platform.sh
@@ -65,6 +65,10 @@ platform_do_upgrade() {
 		dd if=/dev/mtd5 bs=1024 count=52224 >> /tmp/backup_firmware.bin
 		mtd -e firmware2 write /tmp/backup_firmware.bin firmware2
 		;;
+	ruijie,rg-eg105g-v3)
+		fw_setenv firmware_size || exit 1
+		fw_setenv firmware_crc32 || exit 1
+		;;
 	esac
 
 	case "$board" in
--- a/package/boot/uboot-tools/uboot-envtools/files/ramips
+++ b/package/boot/uboot-tools/uboot-envtools/files/ramips
@@ -34,6 +34,7 @@ allnet,all5002|\
 asiarf,ap7621-004-v3|\
 comfast,cf-ew84|\
 plasmacloud,pax1800-lite|\
+ruijie,rg-eg105g-v3|\
 yuncore,ax820)
 	ubootenv_add_uci_config "/dev/mtd1" "0x0" "0x10000" "0x10000"
 	;;


target/linux/ramips/dts/mt7621_ruijie_rg-eg105g-v3.dts
```diff
@@ -0,0 +1,108 @@
...
+				label = "u-boot";
+				reg = <0x000000 0x050000>;
+				read-only;
+			};
+
+			partition@50000 {
+				label = "u-boot-env";
+				reg = <0x050000 0x010000>;
+				read-only;
+			};
```
openwrt-ai, line 36: `read-only;` on `u-boot-env` conflicts with the `fw_setenv` calls this PR adds. `read-only` sets `mask_flags |= MTD_WRITEABLE` ([ofpart_core.c:167-168](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/mtd/parsers/ofpart_core.c#L167-L168)), and `mtdchar_open()` then rejects any write-mode open with `-EACCES` ([mtdchar.c:69-72](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/mtd/mtdchar.c#L69-L72)). So `fw_setenv firmware_size` in `ruijie_rg_eg105g_v3_prepare_upgrade()` fails, and because of the `|| exit 1` the sysupgrade aborts.

None of the other ramips boards that get a `ubootenv_add_uci_config "/dev/mtd1" ...` entry mark their env partition read-only (e.g. `asiarf,ap7621-004-v3`).

```suggestion
				label = "u-boot-env";
				reg = <0x050000 0x010000>;
			};
```

---

target/linux/ramips/dts/mt7621_ruijie_rg-eg105g-v3.dts
```diff
@@ -0,0 +1,108 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+#include "mt7621.dtsi"
+
+/ {
+	compatible = "ruijie,rg-eg105g-v3", "mediatek,mt7621-soc";
+	model = "Ruijie RG-EG105G-V3";
+
+	aliases {
+		label-mac-device = &wan;
```
openwrt-ai: This alias is both dead and contradictory. [`get_mac_label_dt()`](https://github.com/openwrt/openwrt/blob/2bc35633d37c08268bbd490614b0535fd89dff97/package/base-files/files/lib/functions/system.sh#L18-L29) reads `mac-address`/`local-mac-address` from the aliased node; the DSA `wan` port node here has neither (the MACs are assigned from userspace in `02_network`), so it returns nothing and the label MAC falls back to `board.json`. And `board.json` gets `label_mac=$lan_mac`, i.e. the LAN MAC, not the WAN one — so the alias claims the opposite of what the board actually reports.

Either drop the alias, or point it at the interface whose MAC really is on the label.

---

target/linux/ramips/dts/mt7621_ruijie_rg-eg105g-v3.dts
```diff
@@ -0,0 +1,108 @@
...
+				read-only;
+			};
+
+			partition@50000 {
+				label = "u-boot-env";
+				reg = <0x050000 0x010000>;
+				read-only;
+			};
+
+			product_info: partition@60000 {
```
openwrt-ai: nit: the `product_info:` phandle label is never referenced anywhere in this DTS — the MAC is read from userspace via `mtd_get_mac_ascii product_info ethaddr`. The sibling [`mt7621_ruijie_rg-ew1200g-pro-v1.1.dts`](``https://github.com/openwrt/openwrt/blob/2bc35633d37c08268bbd490614b0535fd89dff97/target/linux/ramips/dts/mt7621_ruijie_rg-ew1200g-pro-v1.1.dts#L86-L90``) declares the same partition without a label.

```suggestion
			partition@60000 {
```

---

target/linux/ramips/dts/mt7621_ruijie_rg-eg105g-v3.dts
```diff
@@ -0,0 +1,108 @@
...
+		};
+	};
+};
+
+&state_default {
+	gpio {
+		groups = "uart3", "jtag";
+		function = "gpio";
+	};
+};
```
openwrt-ai: This frees the `uart3` and `jtag` pin groups for GPIO use, but the DTS declares no `gpio-leds` and no `gpio-keys` node, so nothing consumes those GPIOs. The DTS also has no `aliases` for `led-boot`/`led-failsafe`/`led-running`/`led-upgrade`, which means failsafe/upgrade state is not indicated at all on this board.

Both sibling Ruijie boards do declare LEDs and a reset button (see [`mt7621_ruijie_rg-ew1200g-pro-v1.1.dts`](``https://github.com/openwrt/openwrt/blob/2bc35633d37c08268bbd490614b0535fd89dff97/target/linux/ramips/dts/mt7621_ruijie_rg-ew1200g-pro-v1.1.dts#L25-L43)).`` Does the EG105G-V3 genuinely have no reset button and no software-controllable LED, or were they just not mapped yet? If it really has none, the `&state_default` override can go too.

---

target/linux/ramips/dts/mt7621_ruijie_rg-eg105g-v3.dts
```diff
@@ -0,0 +1,108 @@
...
+
```
openwrt-ai: nit: stray trailing blank line at end of file.

```suggestion
```

---

target/linux/ramips/image/mt7621.mk
```diff
@@ -3993,3 +3993,13 @@ define Device/zyxel_wsm20
   KERNEL_INITRAMFS := kernel-bin | lzma | fit lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb | znet-header V1.00(ABZF.0)C0
 endef
 TARGET_DEVICES += zyxel_wsm20
+
+define Device/ruijie_rg-eg105g-v3
+  DEVICE_VENDOR := Ruijie
+  DEVICE_MODEL := RG-EG105G-V3
+  DTS := RG-EG105G-V3
+  IMAGE_SIZE := 32256k
+  DEVICE_PACKAGES := kmod-mt7621-soc uboot-envtools
```
openwrt-ai: `kmod-mt7621-soc` does not exist — there is no such package anywhere in the tree (`grep -r kmod-mt7621-soc` returns nothing) and no such kmod is generated under `package/kernel/linux/modules/`. The MT7621 ethernet/switch driver is built into the kernel for this target, so no extra package is needed. Leaving it here makes the image build fail when resolving `DEVICE_PACKAGES`.

```suggestion
  DEVICE_PACKAGES := uboot-envtools
```

---

target/linux/ramips/image/mt7621.mk
```diff
@@ -3993,3 +3993,13 @@ define Device/zyxel_wsm20
   KERNEL_INITRAMFS := kernel-bin | lzma | fit lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb | znet-header V1.00(ABZF.0)C0
 endef
 TARGET_DEVICES += zyxel_wsm20
+
+define Device/ruijie_rg-eg105g-v3
+  DEVICE_VENDOR := Ruijie
+  DEVICE_MODEL := RG-EG105G-V3
+  DTS := RG-EG105G-V3
```
openwrt-ai: `DTS` is not a variable the image build reads — the DTS name comes from [`DEVICE_DTS = $$(SOC)_$(1)`](https://github.com/openwrt/openwrt/blob/2bc35633d37c08268bbd490614b0535fd89dff97/target/linux/ramips/image/Makefile#L187) in `Device/Default`, which already resolves to `mt7621_ruijie_rg-eg105g-v3`. This line is dead and no other device in this file has it.

```suggestion
```

---

target/linux/ramips/image/mt7621.mk
```diff
@@ -3993,3 +3993,13 @@ define Device/zyxel_wsm20
   KERNEL_INITRAMFS := kernel-bin | lzma | fit lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb | znet-header V1.00(ABZF.0)C0
 endef
 TARGET_DEVICES += zyxel_wsm20
+
+define Device/ruijie_rg-eg105g-v3
```
openwrt-ai: nit: `mt7621.mk` keeps `define Device/...` blocks in alphabetical order; this one is appended after `zyxel_wsm20` instead of next to the existing Ruijie entries ([`define Device/ruijie_rg-ew1200g-pro-v1.1`](https://github.com/openwrt/openwrt/blob/2bc35633d37c08268bbd490614b0535fd89dff97/target/linux/ramips/image/mt7621.mk#L2770) at line 2770). `rg-eg105g-v3` sorts before `rg-ew1200g-pro-v1.1`, so the block belongs immediately above it.

Also, unlike every other block in this file, there is a blank line between `endef` and `TARGET_DEVICES += ruijie_rg-eg105g-v3` (line 4004) — drop it.

---

target/linux/ramips/mt7621/base-files/etc/board.d/02_network
```diff
@@ -289,6 +289,11 @@ ramips_setup_macs()
 		wan_mac=$(macaddr_add "$lan_mac" 8)
 		label_mac=$lan_mac
 		;;
+  ruijie,rg-eg105g-v3)
+	lan_mac=$(mtd_get_mac_ascii product_info ethaddr)
+	wan_mac=$(macaddr_add "$lan_mac" 2)
+	label_mac=$lan_mac
+	;;
```
openwrt-ai: Indentation doesn't match the file: the case label is indented with two spaces instead of one tab, and the body with one tab instead of two. Compare the `ruijie,rg-ew1200g-pro-v1.1` block directly above.

```suggestion
	ruijie,rg-eg105g-v3)
		lan_mac=$(mtd_get_mac_ascii product_info ethaddr)
		wan_mac=$(macaddr_add "$lan_mac" 2)
		label_mac=$lan_mac
		;;
```

---

target/linux/ramips/mt7621/base-files/etc/uci-defaults/10_uboot-envtools
openwrt-ai, line 1: This file shouldn't exist — the u-boot env layout for ramips lives in [`package/boot/uboot-tools/uboot-envtools/files/ramips`](https://github.com/openwrt/openwrt/blob/2bc35633d37c08268bbd490614b0535fd89dff97/package/boot/uboot-tools/uboot-envtools/files/ramips), which the `uboot-envtools` package installs as `/etc/uci-defaults/30_uboot-envtools`. That file already has a group with exactly the parameters used here:

https://github.com/openwrt/openwrt/blob/2bc35633d37c08268bbd490614b0535fd89dff97/package/boot/uboot-tools/uboot-envtools/files/ramips#L30-L39

so `ruijie,rg-eg105g-v3` just needs to be added to that alphabetical list and this file dropped. No ramips subtarget carries its own `10_uboot-envtools`.

Two concrete problems with the file as written, beyond the placement:

1. It is missing the `[ -e /etc/config/ubootenv ] && exit 0` guard that the shared script has at line 5. `/etc/config/*` is preserved across sysupgrade, so on every upgrade that keeps config this script re-runs against the existing `/etc/config/ubootenv` and appends a duplicate `ubootenv` section.
2. `config_foreach ubootenv_add_app_config ubootenv` filters to the `ubootenv` type; the shared script deliberately passes no type so `ubootsys` sections are picked up too.

Numbering also matters: `10_` runs before the package's `30_`, which is the opposite of the intended ordering.

---

openwrt-ai (commented):
## Commit checks

- `2bc3563` "ramips: add support for Ruijie RG-EG105G-V3" — this is a new-device commit (new `.dts` + new `define Device/...`), so per the project's commit-message requirements it should carry three sections; two are missing and one is incomplete:
  - **Hardware specification** — partially there (SoC, 128 MiB RAM, 32 MiB SPI NOR, 4×LAN + 1×WAN via MT7530), but no RAM/flash chip markings, no LED/button inventory, no UART header location/pinout/baud rate, and no power input rating. The DTS declaring neither `gpio-leds` nor `gpio-keys` makes the LED/button line especially worth stating explicitly.
  - **Flash instructions** — absent from the commit message entirely. The PR body has one sentence ("U-Boot TFTP recovery interface using menu option 2 ... renamed to `rgos.bin`"), but it needs to move into the commit message and be verbatim: how to reach the U-Boot menu (key/timing on the serial console), the TFTP server IP and the device's recovery IP, the exact image filename, and a note on reverting to stock if a procedure exists.
  - **MAC address layout** — absent. The commit adds `lan_mac=$(mtd_get_mac_ascii product_info ethaddr)` and `wan_mac=$(macaddr_add "$lan_mac" 2)`; document that as one line per interface (LAN/WAN/label) using `xx:xx:xx` for the unit-specific bytes. Note the two existing Ruijie boards in the same `case` use an offset of 8, so the 2 here is worth recording.

Also `git show --stat` on the commit shows a fifth file, `target/linux/ramips/mt7621/base-files/etc/uci-defaults/10_uboot-envtools`, that the message doesn't mention; see the inline comment — that change belongs in the `uboot-envtools` package instead.


---

target/linux/ramips/dts/mt7621_ruijie_rg-eg105g-v3.dts
```diff
@@ -0,0 +1,98 @@
...
+#include "mt7621.dtsi"
+
+/ {
+	compatible = "ruijie,rg-eg105g-v3", "mediatek,mt7621-soc";
+	model = "Ruijie RG-EG105G-V3";
+
+	aliases {
+	};
+
```
openwrt-ai: The `aliases` node is now empty after `label-mac-device` was dropped, so it contributes nothing to the DTB. This is the only `.dts` under `target/linux/ramips/dts/` with an empty `aliases {}` block — drop it.

```suggestion
```

---

target/linux/ramips/dts/mt7621_ruijie_rg-eg105g-v3.dts
```diff
@@ -0,0 +1,98 @@
...
+				reg = <0x080000 0x1f80000>;
+			};
+		};
+	};
+};
+
+&ethernet {
+	status = "okay";
+};
+
```
openwrt-ai: This override is a no-op: [`ethernet: ethernet@1e100000`](https://github.com/openwrt/openwrt/blob/9fb387d7a8b5c24f57419e4a4bdab5bdb123c8db/target/linux/ramips/dts/mt7621.dtsi#L503-L505) in `mt7621.dtsi` has no `status` property, so the node is already enabled for every board. Of the 298 `mt7621_*.dts` files in the tree, this is the only one that adds `&ethernet { status = "okay"; }` — the other 27 that touch `&ethernet` do so to add `nvmem-cells` or enable `gmac1`.

```suggestion
```

---

target/linux/ramips/dts/mt7621_ruijie_rg-eg105g-v3.dts
```diff
@@ -0,0 +1,98 @@
...
+			reg = <3>;
+			label = "lan4";
+			status = "okay";
+		};
+
+		wan: port@4 {
+			reg = <4>;
+			label = "wan";
+			status = "okay";
+		};
```
openwrt-ai: nit: two redundancies against the base dtsi here. `reg` is already set on each port by [`mt7621.dtsi:601-635`](https://github.com/openwrt/openwrt/blob/9fb387d7a8b5c24f57419e4a4bdab5bdb123c8db/target/linux/ramips/dts/mt7621.dtsi#L601-L635) and is being re-stated with identical values — only `mt7621_ubnt_edgerouter-x-sfp.dts` does that, and it has a reason to. The `lan1:`/`lan2:`/`lan3:`/`lan4:`/`wan:` phandle labels are also unreferenced now that `label-mac-device` is gone.

The sibling [`mt7621_ruijie_rg-ew1200g-pro-v1.1.dts:122-125`](``https://github.com/openwrt/openwrt/blob/9fb387d7a8b5c24f57419e4a4bdab5bdb123c8db/target/linux/ramips/dts/mt7621_ruijie_rg-ew1200g-pro-v1.1.dts#L122-L125``) shows the usual shape (`status` then `label`, nothing else):

```suggestion
		port@0 {
			status = "okay";
			label = "lan1";
		};

		port@1 {
			status = "okay";
			label = "lan2";
		};

		port@2 {
			status = "okay";
			label = "lan3";
		};

		port@3 {
			status = "okay";
			label = "lan4";
		};

		port@4 {
			status = "okay";
			label = "wan";
		};
```

---

target/linux/ramips/image/mt7621.mk
```diff
@@ -2767,6 +2767,15 @@ define Device/rostelecom_rt-sf-1
 endef
 TARGET_DEVICES += rostelecom_rt-sf-1
 
+
```
openwrt-ai: nit: double blank line between `TARGET_DEVICES += rostelecom_rt-sf-1` and this block — every other pair of device blocks in this file is separated by exactly one.

```suggestion
```

---

target/linux/ramips/mt7621/base-files/lib/upgrade/platform.sh
```diff
@@ -65,6 +70,9 @@ platform_do_upgrade() {
 		dd if=/dev/mtd5 bs=1024 count=52224 >> /tmp/backup_firmware.bin
 		mtd -e firmware2 write /tmp/backup_firmware.bin firmware2
 		;;
+	ruijie,rg-eg105g-v3)
+		ruijie_rg_eg105g_v3_prepare_upgrade || exit 1
+		;;
```
openwrt-ai, line 71: nit: this file has no board-specific helper functions — every other board that pokes the U-Boot env in this same `case` writes the `fw_setenv ... || exit 1` calls inline, e.g. [`ampedwireless,ally-*` at platform.sh:47-52](``https://github.com/openwrt/openwrt/blob/9fb387d7a8b5c24f57419e4a4bdab5bdb123c8db/target/linux/ramips/mt7621/base-files/lib/upgrade/platform.sh#L47-L52``) and `iptime,ax200*m` just below it. Inlining the two commands lets the `ruijie_rg_eg105g_v3_prepare_upgrade()` definition at lines 27-30 go away entirely.

```suggestion
	ruijie,rg-eg105g-v3)
		fw_setenv firmware_size || exit 1
		fw_setenv firmware_crc32 || exit 1
		;;
```

---

openwrt-ai (commented):
Re-reviewed the force-pushed head (`2bc3563` → `9fb387d`). The previous round's findings are all addressed: the `u-boot-env` `read-only` is gone, the contradictory `label-mac-device` alias and the unused `product_info:` phandle are dropped, `kmod-mt7621-soc` and the dead `DTS :=` line are removed, the device block moved into alphabetical position, `02_network` is now tab-indented, and the subtarget `10_uboot-envtools` file is replaced by an entry in `package/boot/uboot-tools/uboot-envtools/files/ramips` (correct `/dev/mtd1` index for this partition layout, and inserted alphabetically). `fw_setenv` is available in the sysupgrade ramfs via the existing `RAMFS_COPY_BIN` at the top of `platform.sh`, and deleting a non-existent variable is a no-op that returns 0 in `fw_env_write()`, so the `|| exit 1` path is safe.

## Commit checks

- `9fb387d` "ramips: add support for Ruijie RG-EG105G-V3" — the three required sections are now present, but the flash instructions disagree with the PR body about how to reach the recovery menu: the commit says *"press 4 to enter the U-Boot recovery procedure"*, the PR body says *"through the U-Boot TFTP recovery interface using menu option 2"*. Step 6 (*"Start the TFTP transfer from U-Boot"*) also doesn't name the key/option to press, and the device's own recovery IP isn't given (only the server's `192.168.64.1`). Please make the commit message self-contained and consistent — a stranger should be able to flash the device from it alone.

Remaining items are inline; all are nits except the two dead-DTS-node ones.

---

target/linux/ramips/dts/mt7621_ruijie_rg-eg105g-v3.dts
```diff
@@ -0,0 +1,81 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+#include "mt7621.dtsi"
+
+/ {
+	compatible = "ruijie,rg-eg105g-v3", "mediatek,mt7621-soc";
+	model = "Ruijie RG-EG105G-V3";
+	chosen {
```
openwrt-ai, line 9: nit: dropping the empty `aliases {}` also removed the blank line that separated the root properties from the first child node. All 95 other `mt7621_*.dts` files in `target/linux/ramips/dts/` have a blank line immediately before `chosen {` — e.g. [`mt7621_ruijie_rg-ew1200g-pro-v1.1.dts`](``https://github.com/openwrt/openwrt/blob/274147211df8872748cc84b60c45eb1e3d3b49e2/target/linux/ramips/dts/mt7621_ruijie_rg-ew1200g-pro-v1.1.dts#L10-L16).``

```suggestion
	model = "Ruijie RG-EG105G-V3";

	chosen {
```

---

target/linux/ramips/dts/mt7621_ruijie_rg-eg105g-v3.dts
```diff
@@ -0,0 +1,81 @@
...
+			label = "lan3";
+		};
+		port@3 {
+			status = "okay";
+			label = "lan4";
+		};
+		port@4 {
+			status = "okay";
+			label = "wan";
+		};
```
openwrt-ai, line 84: nit: the `reg`/phandle cleanup also removed the blank lines between the port nodes. Of the 142 `mt7621_*.dts` files with more than one `port@` node under `&switch0`, 137 separate them with a blank line and only 5 don't (all Zyxel/TP-Link). Both sibling Ruijie boards use the blank-line form — [`mt7621_ruijie_rg-ew1200g-pro-v1.1.dts:120-142`](``https://github.com/openwrt/openwrt/blob/274147211df8872748cc84b60c45eb1e3d3b49e2/target/linux/ramips/dts/mt7621_ruijie_rg-ew1200g-pro-v1.1.dts#L120-L142).``

```suggestion
		port@0 {
			status = "okay";
			label = "lan1";
		};

		port@1 {
			status = "okay";
			label = "lan2";
		};

		port@2 {
			status = "okay";
			label = "lan3";
		};

		port@3 {
			status = "okay";
			label = "lan4";
		};

		port@4 {
			status = "okay";
			label = "wan";
		};
```

---

openwrt-ai (commented):
Re-reviewed the force-pushed head (`9fb387d` → `2741472`). Everything from the previous round is addressed: the empty `aliases {}` block and the no-op `&ethernet { status = "okay"; }` override are gone, the redundant `reg` properties and unreferenced `lan1:`/`wan:` phandles on the switch ports are dropped, the double blank line before the device block in `mt7621.mk` is fixed, and the `ruijie_rg_eg105g_v3_prepare_upgrade()` wrapper is inlined into the `case` in `platform.sh` to match the surrounding boards.

Re-checked the parts the new diff touches: `IMAGE_SIZE := 32256k` matches the `0x1f80000` firmware partition, `/dev/mtd1` is still the correct index for `u-boot-env` in this layout, and the default `*)` arm of `ramips_setup_interfaces()` already yields `lan1 lan2 lan3 lan4` + `wan`, which matches the port labels — so no extra entry is needed there. The commit message's flash instructions and the PR body now agree on menu option 2, and the recovery IPs are both stated. CI is green on `2741472`.

Only two DTS whitespace nits left, both inline.

---

PR #22387, opened 2026-03-12
mediatek: filogic: add support for ipTIME AX6000M

saeheeryu:
Add support for the ipTIME AX6000M, a MediaTek Filogic-based dual-band WiFi 6 router.

This adds support for the ipTIME AX6000M including the device DTS, image recipe entry, MAC address handling, and correct LAN port ordering.

## Specification

| Component | Model |
|-----------|------|
| SoC | MediaTek MT7986A (Filogic 830) |
| RAM | 512 MB DDR3 |
| Flash | 128 MB SPI-NAND |
| Switch | MediaTek MT7531 |
| Wireless 2.4 GHz | MediaTek MT7975N |
| Wireless 5 GHz | MediaTek MT7975PN |
| USB | 1 × USB 3.0 |

### Buttons

| Button |
|------|
| Reset |
| WPS |

### LEDs

| Type | LEDs |
|-----|------|
| GPIO-controlled | CPU (used as power), WiFi 2.4G, WiFi 5G |
| Switch-controlled | WAN, LAN1, LAN2, LAN3, LAN4 |

## Flash Layout

The device uses **SPI-NAND with NMBM and UBI**.

| Partition | Address |
|----------|--------|
| BL2 | 0x00000000 |
| u-boot-env | 0x00100000 |
| Factory | 0x00180000 |
| FIP | 0x00380000 |
| ubi | 0x00580000 |

## MAC Address

The base MAC address is stored in the **Factory partition at offset 0x4**.

Observed default addresses:

| Interface | Address |
|----------|--------|
| base | b0:xx:xx:2a:xx:ac |
| LAN | b0:xx:xx:2a:xx:af |
| WAN | b0:xx:xx:2a:xx:ad |
| 2G | b0:xx:xx:2a:xx:ac |
| 5G | b2:xx:xx:4a:xx:ac |

Derived rules:

| Interface | Calculation |
|----------|-------------|
| WLAN 2.4 GHz | base |
| WAN | base + 1 |
| LAN | base + 3 |
| WLAN 5 GHz | base-derived MAC with locally administered bit set |

The Ethernet MAC assignment is implemented in:

```
target/linux/mediatek/filogic/base-files/etc/board.d/02_network
```

<details>
<summary>Installation</summary>

### Method 1: UART available

Initial installation can be performed from the **U-Boot console**.

0. Configure the host PC as:

```
192.168.0.2/24
```

2. Interrupt boot and enter the boot menu
3. Select `0` to enter U-Boot
[...]

PR #24510, opened 2026-07-31
mediatek: filogic: add support for ipTIME AX6000M

DH-41:
This PR is based on the work done by @saeheeryu in #22387.

Add support for the ipTIME AX6000M, a MediaTek MT7986-based dual-band WiFi 6 router with SPI-NAND, NMBM, and UBI.

Specification
-------------
- SoC       : MediaTek MT7986AV quad-core ARM Cortex-A53 2GHz
- RAM       : DDR3 512Mbytes, Nanya NT5CC256M16ER-EK
- Flash     : 128Mbytes NAND Flash, ESMT F50L1G41LB SPI-NAND
- WLAN      : MediaTek MT7975N, MediaTek MT7975PN
  - 2.4GHz  : b/g/n/ax, Multi User MIMO
  - 5GHz    : a/n/ac/ax, Multi User MIMO
- Ethernet  : 10/100/1000 Mbps x4, LAN (Mediatek MT7531AE)
              10/100/1000 Mbps x1, WAN
- Buttons   : WPS, Reset
- USB       : 1x USB 3.0 (MediaTek MT7986AV peripheral)
- LEDs      : 1x CPU (Amber, GPIO-controlled)
              1x Wi-Fi 5GHz (Amber, GPIO-controlled)
              1x Wi-Fi 2.4GHz (Amber, GPIO-controlled)
              4x LAN activity (Amber, Switch-controlled)
              1x WAN activity (Amber, Switch-controlled)
- Power     : 12VDC, 2A (Center positive polarity)

MAC address
-----------
| Interface | MAC | Calculation |
| --- | --- | --- |
| WLAN 2.4G | 58:86:94:xx:xx:xx | base |
| WLAN 5G | 5A:86:94:xx:xx:xx | base with LA Bit Set |
| WAN | 58:86:94:xx:xx:xx | base + 1 |
| LAN       | 58:86:94:xx:xx:xx | base + 3 |

The WLAN 2.4G MAC address (base) was found in 'Factory' partition, 0x4

Installation
------------
- Initial install via U-Boot recovery using `mtkupgrade`
- Sysupgrade supported after OpenWrt installation
- Stock web UI installation is not supported

1. Download the OEM recovery software (ipTIME Firmware Wizard (11ac)) from the manufacturer's website
2. Download the *squashfs-factory.bin file from the OpenWrt website
3. Press a reset button, and power up the router (keep pressing the reset button)
4. Wait more than 10 seconds until the CPU LED stop blinking
5. Connect the router (LAN port) to the PC
6. Run the OEM recovery software and follow the instructions
7. Select the *squashfs-factory.bin file during the router recovery process
8. Wait for the router to boot from *squashfs-factory.bin

Note: The router will automatically reboot if no file is uploaded within 55 seconds.

### DIFF
--- a/target/linux/mediatek/dts/mt7986a-iptime-ax6000m.dts
+++ b/target/linux/mediatek/dts/mt7986a-iptime-ax6000m.dts
@@ -1,0 +1,301 @@
+// SPDX-License-Identifier: GPL-2.0-only OR MIT
+
+/dts-v1/;
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+#include <dt-bindings/pinctrl/mt65xx.h>
+
+#include "mt7986a.dtsi"
+
+/ {
+	model = "ipTIME AX6000M";
+	compatible = "iptime,ax6000m", "mediatek,mt7986a";
+
+	aliases {
+		serial0 = &uart0;
+		label-mac-device = &wan;
+		led-boot = &led_cpu;
+		led-failsafe = &led_cpu;
+		led-running = &led_cpu;
+		led-upgrade = &led_cpu;
+	};
+
+	chosen {
+		stdout-path = "serial0:115200n8";
+	};
+
+	memory@40000000 {
+		reg = <0 0x40000000 0 0x20000000>;
+		device_type = "memory";
+	};
+
+	gpio-keys {
+		compatible = "gpio-keys";
+
+		button-0 {
+			label = "wps";
+			gpios = <&pio 10 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_WPS_BUTTON>;
+			debounce-interval = <60>;
+		};
+
+		button-1 {
+			label = "reset";
+			gpios = <&pio 9 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+			debounce-interval = <60>;
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		led_cpu: led-0 {
+			color = <LED_COLOR_ID_AMBER>;
+			function = LED_FUNCTION_CPU;
+			gpios = <&pio 22 GPIO_ACTIVE_LOW>;
+			default-state = "on";
+		};
+
+		led-1 {
+			color = <LED_COLOR_ID_AMBER>;
+			function = LED_FUNCTION_WLAN_2GHZ;
+			gpios = <&pio 1 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "phy0tpt";
+		};
+
+		led-2 {
+			color = <LED_COLOR_ID_AMBER>;
+			function = LED_FUNCTION_WLAN_5GHZ;
+			gpios = <&pio 2 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "phy1tpt";
+		};
+	};
+};
+
+&pio {
+	spi_flash_pins: spi-flash-pins-33-to-38 {
+		mux {
+			function = "spi";
+			groups = "spi0", "spi0_wp_hold";
+		};
+
+		conf-pu {
+			pins = "SPI2_CS", "SPI2_HOLD", "SPI2_WP";
+			drive-strength = <MTK_DRIVE_4mA>;
+			bias-pull-up = <MTK_PUPD_SET_R1R0_11>;
+		};
+
+		conf-pd {
+			pins = "SPI2_CLK", "SPI2_MOSI", "SPI2_MISO";
+			drive-strength = <MTK_DRIVE_4mA>;
+			bias-pull-down = <MTK_PUPD_SET_R1R0_11>;
+		};
+	};
+
+	wf_2g_5g_pins: wf_2g_5g-pins {
+		mux {
+			function = "wifi";
+			groups = "wf_2g", "wf_5g";
+		};
+		conf {
+			pins = "WF0_HB1", "WF0_HB2", "WF0_HB3", "WF0_HB4",
+			       "WF0_HB0", "WF0_HB0_B", "WF0_HB5", "WF0_HB6",
+			       "WF0_HB7", "WF0_HB8", "WF0_HB9", "WF0_HB10",
+			       "WF0_TOP_CLK", "WF0_TOP_DATA", "WF1_HB1",
+			       "WF1_HB2", "WF1_HB3", "WF1_HB4", "WF1_HB0",
+			       "WF1_HB5", "WF1_HB6", "WF1_HB7", "WF1_HB8",
+			       "WF1_TOP_CLK", "WF1_TOP_DATA";
+			drive-strength = <MTK_DRIVE_4mA>;
+		};
+	};
+};
+
+&spi0 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&spi_flash_pins>;
+	status = "okay";
+
+	flash@0 {
+		compatible = "spi-nand";
+		reg = <0>;
+
+		spi-max-frequency = <52000000>;
+		spi-tx-bus-width = <4>;
+		spi-rx-bus-width = <4>;
+
+		mediatek,nmbm;
+		mediatek,bmt-max-ratio = <1>;
+		mediatek,bmt-max-reserved-blocks = <64>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "BL2";
+				reg = <0x0 0x100000>;
+				read-only;
+			};
+
+			partition@100000 {
+				label = "u-boot-env";
+				reg = <0x100000 0x80000>;
+			};
+
+			partition@180000 {
+				label = "Factory";
+				reg = <0x180000 0x200000>;
+				read-only;
+
+				nvmem-layout {
+					compatible = "fixed-layout";
+					#address-cells = <1>;
+					#size-cells = <1>;
+
+					eeprom_factory_0: eeprom@0 {
+						reg = <0x0 0x1000>;
+					};
+
+					macaddr_factory_4: macaddr@4 {
+						compatible = "mac-base";
+						reg = <0x4 0x6>;
+						#nvmem-cell-cells = <1>;
+					};
+				};
+			};
+
+			partition@380000 {
+				label = "FIP";
+				reg = <0x380000 0x200000>;
+				read-only;
+			};
+
+			partition@580000 {
+				label = "ubi";
+				reg = <0x580000 0x6e00000>;
+			};
+		};
+	};
+};
+
+&eth {
+	status = "okay";
+
+	gmac0: mac@0 {
+		compatible = "mediatek,eth-mac";
+		reg = <0>;
+		phy-mode = "2500base-x";
+
+		nvmem-cell-names = "mac-address";
+		nvmem-cells = <&macaddr_factory_4 3>;
+
+		fixed-link {
+			speed = <2500>;
+			full-duplex;
+			pause;
+		};
+	};
+
+	mdio: mdio-bus {
+		#address-cells = <1>;
+		#size-cells = <0>;
+		reset-gpios = <&pio 6 GPIO_ACTIVE_LOW>;
+		reset-delay-us = <50000>;
+		reset-post-delay-us = <20000>;
+	};
+};
+
+&mdio {
+	switch@1f {
+		compatible = "mediatek,mt7531";
+		reg = <0x1f>;
+		reset-gpios = <&pio 5 GPIO_ACTIVE_HIGH>;
+		reset-assert-us = <10000>;
+		reset-deassert-us = <10000>;
+
+		interrupt-controller;
+		#interrupt-cells = <1>;
+		interrupt-parent = <&pio>;
+		interrupts = <66 IRQ_TYPE_LEVEL_HIGH>;
+
+		ports {
+			#address-cells = <1>;
+			#size-cells = <0>;
+
+			wan: port@0 {
+				reg = <0>;
+				label = "wan";
+				nvmem-cell-names = "mac-address";
+				nvmem-cells = <&macaddr_factory_4 1>;
+			};
+
+			port@1 {
+				reg = <1>;
+				label = "lan4";
+			};
+
+			port@2 {
+				reg = <2>;
+				label = "lan3";
+			};
+
+			port@3 {
+				reg = <3>;
+				label = "lan2";
+			};
+
+			port@4 {
+				reg = <4>;
+				label = "lan1";
+			};
+
+			port@6 {
+				reg = <6>;
+				label = "cpu";
+				ethernet = <&gmac0>;
+				phy-mode = "2500base-x";
+
+				fixed-link {
+					speed = <2500>;
+					full-duplex;
+					pause;
+				};
+			};
+		};
+	};
+};
+
+&wifi {
+	nvmem-cells = <&eeprom_factory_0>;
+	nvmem-cell-names = "eeprom";
+	pinctrl-names = "default";
+	pinctrl-0 = <&wf_2g_5g_pins>;
+	status = "okay";
+};
+
+&trng {
+	status = "okay";
+};
+
+&crypto {
+	status = "okay";
+};
+
+&uart0 {
+	status = "okay";
+};
+
+&watchdog {
+	status = "okay";
+};
+
+&ssusb {
+	status = "okay";
+};
+
+&usb_phy {
+	status = "okay";
+};
--- a/target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
+++ b/target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
@@ -144,7 +144,8 @@ case "$board" in
 		[ "$PHYNBR" = "0" ] && macaddr_add $base_mac 2 > /sys${DEVPATH}/macaddress
 		[ "$PHYNBR" = "1" ] && macaddr_add $base_mac 3 > /sys${DEVPATH}/macaddress
 		;;
-	iptime,ax3000q)
+	iptime,ax3000q|\
+	iptime,ax6000m)
 		addr=$(mtd_get_mac_binary "Factory" 0x4)
 		[ "$PHYNBR" = "1" ] && macaddr_setbit_la $addr > /sys${DEVPATH}/macaddress
 		;;
--- a/target/linux/mediatek/image/filogic.mk
+++ b/target/linux/mediatek/image/filogic.mk
@@ -2120,6 +2120,23 @@ define Device/iptime_ax3000m
 endef
 TARGET_DEVICES += iptime_ax3000m
 
+define Device/iptime_ax6000m
+  DEVICE_VENDOR := ipTIME
+  DEVICE_MODEL := AX6000M
+  DEVICE_DTS := mt7986a-iptime-ax6000m
+  DEVICE_DTS_DIR := ../dts
+  BLOCKSIZE := 128k
+  PAGESIZE := 2048
+  IMAGE_SIZE := 32768k
+  KERNEL_INITRAMFS := kernel-bin | lzma | \
+	fit lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb with-initrd | pad-to 64k
+  IMAGES := factory.bin sysupgrade.bin
+  IMAGE/factory.bin := sysupgrade-tar | append-metadata | check-size | iptime-crc32 ax6000m
+  IMAGE/sysupgrade.bin := sysupgrade-tar | append-metadata
+  DEVICE_PACKAGES := kmod-usb3 kmod-mt7915e kmod-mt7986-firmware mt7986-wo-firmware
+endef
+TARGET_DEVICES += iptime_ax6000m
+
 define Device/iptime_ax7800m-6e
   DEVICE_VENDOR := ipTIME
   DEVICE_MODEL := AX7800M-6E


target/linux/mediatek/dts/mt7986a-iptime-ax6000m.dts
```diff
@@ -0,0 +1,302 @@
...
+		compatible = "spi-nand";
+		reg = <0>;
+
+		spi-max-frequency = <52000000>;
+		spi-tx-buswidth = <4>;
+		spi-rx-buswidth = <4>;
+
+		# mediatek,nmbm;
+		# mediatek,bmt-max-ratio = <1>;
+		# mediatek,bmt-max-reserved-blocks = <64>;
```
openwrt-ai: `#` is not a comment character in device tree source — it is a valid property-name character, so dtc parses `# mediatek,nmbm;` as two property names in a row and aborts. This is what breaks the `mediatek/filogic` CI build:

```
Lexical error: ../dts/mt7986a-iptime-ax6000m.dts:128.5-13 Unexpected 'mediatek'
Error: ../dts/mt7986a-iptime-ax6000m.dts:128.5-13 syntax error
FATAL ERROR: Unable to parse input tree
make[4]: *** [Makefile:48: .../image-mt7986a-iptime-ax6000m.dtb] Error 1
```

Use C-style comments:

```suggestion
		/* mediatek,nmbm; */
		/* mediatek,bmt-max-ratio = <1>; */
		/* mediatek,bmt-max-reserved-blocks = <64>; */
```

That said, commented-out properties tend not to survive review here. Since you can't verify NMBM on hardware yet, dropping the three lines entirely (and the NMBM claim from the commit message) is probably the cleaner path — they can be added in a follow-up once someone with serial access confirms the U-Boot side, in the same form as [mt7986a-wavlink-wl-wn536ax6-a.dts:292-294](``https://github.com/openwrt/openwrt/blob/57a96cdf6b10b4b9cb64038af188ffc4ecfe23d0/target/linux/mediatek/dts/mt7986a-wavlink-wl-wn536ax6-a.dts#L292-L294).``

---

target/linux/mediatek/dts/mt7986a-iptime-ax6000m.dts
```diff
@@ -0,0 +1,302 @@
...
+						reg = <0x0 0x1000>;
+					};
+
+					macaddr_factory_4: macaddr@4 {
+						compatible = "mac-base";
+						reg = <0x4 0x6>;
+						#nvmem-cell-cells = <1>;
+					};
+
+				};
```
openwrt-ai, line 167: nit: stray blank line before the closing brace of `nvmem-layout`; the equivalent block in [mt7986a-iptime-ax7800m-6e.dts](https://github.com/openwrt/openwrt/blob/57a96cdf6b10b4b9cb64038af188ffc4ecfe23d0/target/linux/mediatek/dts/mt7986a-iptime-ax7800m-6e.dts#L225-L228) closes directly.

```suggestion
					};
				};
```

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -2089,6 +2089,24 @@ define Device/iptime_ax7800m-6e
...
 TARGET_DEVICES += iptime_ax7800m-6e
 
+define Device/iptime_ax6000m
+  DEVICE_VENDOR := ipTIME
+  DEVICE_MODEL := AX6000M
+  DEVICE_DTS := mt7986a-iptime-ax6000m
+  DEVICE_DTS_DIR := ../dts
+  BLOCKSIZE := 128k
+  PAGESIZE := 2048
+  IMAGE_SIZE := 114688k
```
openwrt-ai: `IMAGE_SIZE` is larger than the partition the image has to fit into, so `check-size` on `factory.bin` can never fire. The `ubi` partition in the new DTS is `reg = <0x580000 0x6e00000>` = 115343360 bytes = 112640k, but this is 114688k — 2048k more than the whole partition, before any UBI EC/VID header overhead is accounted for.

Both in-tree boards with the identical `0x6e00000` ubi partition — [`iptime_ax7800m-6e`](https://github.com/openwrt/openwrt/blob/57a96cdf6b10b4b9cb64038af188ffc4ecfe23d0/target/linux/mediatek/image/filogic.mk#L2081) and [`iptime_ax3000m`](https://github.com/openwrt/openwrt/blob/57a96cdf6b10b4b9cb64038af188ffc4ecfe23d0/target/linux/mediatek/image/filogic.mk#L2063) — use `IMAGE_SIZE := 32768k`. Was 114688k measured against something on this board, or is it a leftover? It needs to be capped below the usable UBI volume size either way.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -2089,6 +2089,24 @@ define Device/iptime_ax7800m-6e
...
+  KERNEL := kernel-bin | lzma | fit lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb
```
openwrt-ai: nit: this `KERNEL` line restates the target default verbatim — [`Device/Default` in target/linux/mediatek/image/Makefile:21-22](https://github.com/openwrt/openwrt/blob/57a96cdf6b10b4b9cb64038af188ffc4ecfe23d0/target/linux/mediatek/image/Makefile#L21-L22) already sets `kernel-bin | lzma | fit lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb`. None of the four sibling ipTIME entries above override it; only `KERNEL_INITRAMFS` needs an override here, for the `pad-to 64k`. Dropping this line keeps the block consistent with its neighbours.

---

openwrt-ai (commented):
## Commit checks

- 57a96cdf "mediatek: filogic: add support for ipTIME AX6000M" — the message states the board is supported "with SPI-NAND, NMBM, and UBI", but the `mediatek,nmbm` / `mediatek,bmt-max-ratio` / `mediatek,bmt-max-reserved-blocks` properties are commented out in the DTS, so NMBM is not actually enabled by this commit. Either enable them or drop NMBM from the description so the message matches the diff.

Otherwise the commit message covers all three required sections (hardware spec, flash instructions, MAC layout), and the `11_fix_wifi_mac` bit arithmetic added for `iptime,ax6000m` matches the documented `b0:…:2a:…` → `b2:…:4a:…` derivation.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -2071,6 +2071,24 @@ define Device/iptime_ax3000m
...
 TARGET_DEVICES += iptime_ax3000m
 
+define Device/iptime_ax6000m
+  DEVICE_VENDOR := ipTIME
+  DEVICE_MODEL := AX6000M
+  DEVICE_DTS := mt7986a-iptime-ax6000m
+  DEVICE_DTS_DIR := ../dts
+  BLOCKSIZE := 128k
+  PAGESIZE := 2048
+  IMAGE_SIZE := 110592k
```
openwrt-ai: 110592k now fits inside the raw `ubi` partition (`reg = <0x580000 0x6e00000>` = 112640k), but it is still larger than what UBI can actually hand out on this flash: 880 PEBs × (131072 − 2×2048 header bytes) ≈ 106 MiB, before UBI's bad-PEB reserve and the layout volume. So `check-size` on `factory.bin` still can't fire for an image that won't fit.

The five sibling ipTIME entries in this file, including [`iptime_ax7800m-6e`](https://github.com/openwrt/openwrt/blob/91a76ef97881e0f4c832b3ab2c7fa41666066ae1/target/linux/mediatek/image/filogic.mk#L2099) which has the byte-identical `0x6e00000` ubi partition, all use `IMAGE_SIZE := 32768k` — presumably because the ipTIME recovery uploader is the only install path and has its own limit.

Where does 110592k come from? In-tree values aren't rigorous here (e.g. [`keenetic_kn-3711`](https://github.com/openwrt/openwrt/blob/91a76ef97881e0f4c832b3ab2c7fa41666066ae1/target/linux/mediatek/image/filogic.mk#L2196) sets 108544k against a 98304k partition), so I can't call this wrong outright — but if it was measured against something specific on this board, please say so in the commit message; otherwise matching the siblings' 32768k looks safer.

---

openwrt-ai (commented):
Reviewed 1 new commit (`57a96cdf` → `91a76ef9`). The dtc syntax error (`#` → `/* */`), the stray blank line in `nvmem-layout`, and the ordering of the new `Device/iptime_ax6000m` block relative to its ipTIME siblings are all addressed. Remaining points below.

## Commit checks

- `91a76ef9` "mediatek: filogic: add support for ipTIME AX6000M" — the body still says the board is supported "with SPI-NAND, NMBM, and UBI", but `mediatek,nmbm` / `mediatek,bmt-max-ratio` / `mediatek,bmt-max-reserved-blocks` are still commented out in the DTS, so NMBM is not enabled by this commit. The PR body explains why they're commented out (U-Boot log can't be verified yet), which is fine — but the commit message should match what the diff actually enables. Dropping "NMBM" from that sentence, or noting that it is prepared but disabled pending verification, would resolve it.

The three required sections (hardware specification, flash instructions, MAC address layout) are all present and complete, and the `11_fix_wifi_mac` bit arithmetic for `iptime,ax6000m` still matches the documented `b0:…:2a:…:ac` → `b2:…:4a:…:ac` derivation.

Note the earlier `KERNEL :=` nit (it restates the `Device/Default` value verbatim; no sibling ipTIME entry overrides it) is still open on an outdated thread — not re-posted here to avoid duplication.

---

target/linux/mediatek/dts/mt7986a-iptime-ax6000m.dts
```diff
@@ -0,0 +1,301 @@
...
+		compatible = "spi-nand";
+		reg = <0>;
+
+		spi-max-frequency = <52000000>;
+		spi-tx-buswidth = <4>;
+		spi-rx-buswidth = <4>;
+
+		mediatek,nmbm;
+		mediatek,bmt-max-ratio = <1>;
+		mediatek,bmt-max-reserved-blocks = <64>;
```
openwrt-ai, line 130: NMBM is enabled now, but the PR body's "Unsolved" section still says these three lines had to be commented out because you cannot verify the U-Boot logs — so either the body or the diff is out of date. Which one?

This matters beyond bookkeeping: the BMT/NMBM remap table has to be shared with the bootloader. If the OEM U-Boot on this board does not use BMT, the kernel's remapped view of the chip diverges from U-Boot's raw view as soon as a block goes bad, and the `BL2` / `FIP` / `ubi` offsets no longer agree between the two.

The in-tree evidence is split, which is why I can't call it either way:

- the three mt7981b ipTIME boards enable exactly these three properties, e.g. [`mt7981b-iptime-ax3000se.dts:114-116`](https://github.com/openwrt/openwrt/blob/f86c9f4dfa22863a1a2ae7e8ea054d19ff578c4e/target/linux/mediatek/dts/mt7981b-iptime-ax3000se.dts#L114-L116) (same for `ax3000q` and `ax3000sm`);
- but the closest sibling — [`mt7986a-iptime-ax7800m-6e.dts:179-187`](https://github.com/openwrt/openwrt/blob/f86c9f4dfa22863a1a2ae7e8ea054d19ff578c4e/target/linux/mediatek/dts/mt7986a-iptime-ax7800m-6e.dts#L179-L187), same SoC and a byte-identical partition table down to `ubi` at `<0x580000 0x6e00000>` — does not, going straight from `spi-rx-buswidth` to `partitions`.

If the AX6000M's U-Boot was confirmed to use BMT, please note that in the commit message and drop the stale "Unsolved" paragraph from the PR body. If it is still unverified, following the mt7986a sibling and leaving NMBM out is the safer default for a first submission.

---

openwrt-ai (commented):
Reviewed the amended commit (`91a76ef9` → `f86c9f4d`). Two of the three earlier points are resolved: `mediatek,nmbm` / `mediatek,bmt-max-ratio` / `mediatek,bmt-max-reserved-blocks` are now real properties instead of comments, so the commit message's "SPI-NAND, NMBM, and UBI" claim matches what the diff actually enables, and `IMAGE_SIZE` is now `32768k`, matching all five sibling ipTIME entries in `filogic.mk` and comfortably inside the usable UBI space. CI on `f86c9f4d` is green, including the `mediatek/filogic` build.

The commit message still covers all three required sections (hardware specification, flash instructions, MAC address layout), so no commit-message issues remain. One open question on the NMBM change inline.

---

target/linux/mediatek/dts/mt7986a-iptime-ax6000m.dts
DH-41: Kernel Log
> [    0.939022] spi-nand spi0.0: ESMT SPI NAND was found.
[    0.944070] spi-nand spi0.0: 128 MiB, block size: 128 KiB, page size: 2048, OOB size: 64
[    0.952821] Signature found at block 1023 [0x07fe0000]
[    0.957945] NMBM management region starts at block 960 [0x07800000]
[    0.966043] First info table with writecount 0 found in block 960
[    0.977620] Second info table with writecount 0 found in block 963
[    0.983789] NMBM has been successfully attached
[    0.989438] 5 fixed-partitions partitions found on MTD device spi0.0

SystemLog
>[2026. 8. 2. 오후 12시 29분 53초 UTC] kern.info: [    0.939022] spi-nand spi0.0: ESMT SPI NAND was found.
[2026. 8. 2. 오후 12시 29분 53초 UTC] kern.info: [    0.944070] spi-nand spi0.0: 128 MiB, block size: 128 KiB, page size: 2048, OOB size: 64
[2026. 8. 2. 오후 12시 29분 53초 UTC] kern.info: [    0.952821] Signature found at block 1023 [0x07fe0000]
[2026. 8. 2. 오후 12시 29분 53초 UTC] kern.info: [    0.957945] NMBM management region starts at block 960 [0x07800000]
[2026. 8. 2. 오후 12시 29분 53초 UTC] kern.info: [    0.966043] First info table with writecount 0 found in block 960
[2026. 8. 2. 오후 12시 29분 53초 UTC] kern.info: [    0.977620] Second info table with writecount 0 found in block 963
[2026. 8. 2. 오후 12시 29분 53초 UTC] kern.info: [    0.983789] NMBM has been successfully attached
[2026. 8. 2. 오후 12시 29분 53초 UTC] kern.notice: [    0.989438] 5 fixed-partitions partitions found on MTD device spi0.0

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
```diff
@@ -150,7 +150,8 @@ case "$board" in
 		;;
 	iptime,ax3000m|\
 	iptime,ax3000se|\
-	iptime,ax3000sm)
+	iptime,ax3000sm|\
+	iptime,ax6000m)
```
openwrt-ai: The rewritten MAC table in the commit message now under-describes what this branch does. It says

```
| WLAN 5G   | 5A:86:94:xx:xx:xx | base  with LA Bit set  |
```

but `iptime,ax6000m` is in the group that runs `macaddr_setbit_la $(macaddr_unsetbit $(macaddr_unsetbit $(macaddr_unsetbit $(macaddr_setbit $addr 26) 25) 27) 28)`, not the plain `macaddr_setbit_la` used by [`iptime,ax3000q`](``https://github.com/openwrt/openwrt/blob/4590392c68efef71482089f1ad30aac2009dcc63/target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac#L147-L150).`` Bit args 25–28 map to `2**(48-N)`, i.e. the 4th octet: set `0x40`, clear `0x80`, `0x20` and `0x10`. So the 5G address is *not* simply the base with the LA bit set — the 4th octet is rewritten too, and that only degenerates to "LA bit only" for base addresses that already happen to have `0x40` set and `0x80/0x20/0x10` clear.

The earlier revision of this commit message showed it explicitly (`b0:xx:xx:2a:xx:ac` → `b2:xx:xx:4a:xx:ac`); masking octets 4–6 as `xx` hides the one transformation that is worth documenting. Please either keep an unmasked 4th octet in the 2.4G/5G rows, or spell the extra bit manipulation out in the Calculation column.

Incidentally, the base OUI also changed between revisions (`b0:…` → `58:86:94:…`) — worth double-checking which one the unit actually has, since only one can be right.

---

target/linux/mediatek/dts/mt7986a-iptime-ax6000m.dts
```diff
@@ -0,0 +1,301 @@
...
+
+		button-1 {
+			label = "reset";
+			gpios = <&pio 9 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+			debounce-interval = <60>;
+		};
+	};
+
+	leds {
```
openwrt-ai, line 51: nit: the commit message lists `4x LAN activity (Amber)` and `1x WAN activity (Amber)` under LEDs, but only the CPU / 2.4G / 5G LEDs exist here, and there is no `iptime,ax6000m)` case in [`01_leds`](``https://github.com/openwrt/openwrt/blob/4590392c68efef71482089f1ad30aac2009dcc63/target/linux/mediatek/filogic/base-files/etc/board.d/01_leds#L137-L145``) either. All three ipTIME siblings in that file do configure a WAN LED (`iptime,ax3000q`, `iptime,ax3000sm`, `iptime,ax7800m-6e`).

If the LAN/WAN LEDs on this board are wired to the MT7531's per-port LED pins and driven by the switch in hardware, nothing is needed and this is fine — but then it'd be clearer to say so in the spec list. If they are GPIO-driven like the ax3000q's `amber:wan`, they're currently dead in OpenWrt. Which is it?

---

openwrt-ai (commented):
Reviewed the rebased/amended commit (`f86c9f4d` → `4590392c`). The branch was rebased onto current `main`; the only PR-authored changes since the last review are the removal of the redundant `KERNEL :=` line from `Device/iptime_ax6000m` (confirmed — [`Device/Default` in target/linux/mediatek/image/Makefile:19-20](https://github.com/openwrt/openwrt/blob/4590392c68efef71482089f1ad30aac2009dcc63/target/linux/mediatek/image/Makefile#L19-L20) sets exactly that recipe, so the override was a no-op) and a rewrite of the MAC address table in the commit message.

The earlier NMBM question is answered — the kernel log you posted shows `NMBM has been successfully attached` on this board, so `mediatek,nmbm` in the DTS is justified. That resolves the last open inline thread. CI on `4590392c` is green, including `mediatek/filogic`.

## Commit checks

- `4590392c` "mediatek: filogic: add support for ipTIME AX6000M" — the reworked MAC table now says the 5G address is `base with LA Bit set`, but the `iptime,ax6000m` branch added to `11_fix_wifi_mac` also rewrites the 4th octet (sets `0x40`, clears `0x80`/`0x20`/`0x10`) on top of the LA bit. Details inline.

All three required sections (hardware specification, flash instructions, MAC address layout) remain present and complete.

---

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
```diff
@@ -144,7 +144,8 @@ case "$board" in
 		[ "$PHYNBR" = "0" ] && macaddr_add $base_mac 2 > /sys${DEVPATH}/macaddress
 		[ "$PHYNBR" = "1" ] && macaddr_add $base_mac 3 > /sys${DEVPATH}/macaddress
 		;;
-	iptime,ax3000q)
+	iptime,ax3000q|\
+	iptime,ax6000m)
```
openwrt-ai, line 148: This changes the phy1 address the board actually gets, and the evidence in this PR's own history points the other way. An earlier revision of this commit message documented the stock pair as `b0:xx:xx:2a:xx:ac` → `b2:xx:xx:4a:xx:ac` ([f86c9f4d](https://github.com/openwrt/openwrt/commit/f86c9f4dfa22863a1a2ae7e8ea054d19ff578c4e)). Plain `macaddr_setbit_la` cannot produce `2a` → `4a` in the 4th octet; only the [`macaddr_setbit … 26 … 25 … 27 … 28` branch at 11_fix_wifi_mac:152-157](``https://github.com/openwrt/openwrt/blob/8ac4e7f27343bd804e6ad8bda37af7eea6b92fe0/target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac#L152-L157)`` does (set `0x40`, clear `0x80`/`0x20`/`0x10`), which is where `iptime,ax6000m` sat until this revision.

My previous comment was that the masked table (`5A:86:94:xx:xx:xx`) under-describes the old code — moving the board to the simple branch also resolves that, but only if the simple form is what the stock firmware does. Since octets 4-6 are masked in the current table, it can't distinguish the two: the simple and complex results coincide only when the base 4th octet already has `0x40` set and `0x80`/`0x20`/`0x10` clear.

Was the 5G MAC re-read from the OEM firmware on your unit (i.e. is the stock phy1 address exactly base with only bit 41 flipped), or was the branch changed to match the table? If it was verified, please say so — the in-tree ipTIME family is split (`ax3000q` and `ax7800m-6e` use the plain form, `ax3000m`/`ax3000se`/`ax3000sm` the complex one), so neighbours don't settle it. If it wasn't, the earlier `2a` → `4a` observation is the only hardware datum on record and the complex branch should stay.

---

openwrt-ai (commented):
Reviewed the amended commit (`4590392c` → `8ac4e7f2`). Changes since the last review are limited to `11_fix_wifi_mac` (moving `iptime,ax6000m` from the four-bit-manipulation group to the plain `macaddr_setbit_la` group with `iptime,ax3000q`) plus two commit-message edits: the LED list is now annotated `GPIO-controlled` / `Switch-controlled`, and the 5G row's spacing was tidied.

The LED annotation answers my earlier question — the DTS puts all four LAN ports and WAN on the MT7531 (`wan: port@0`, `lan1`-`lan4` on ports 1-4), so those LEDs are switch-driven and no `01_leds` entry is needed. That thread is resolved.

The commit message now matches the code for the 5G derivation ("base with LA Bit Set" ↔ plain `macaddr_setbit_la`), so there is no message-vs-diff mismatch left — but the direction of that alignment is worth a second look before merge; details inline. CI on `8ac4e7f2` is green, including `mediatek/filogic`.

---

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
DH-41: The part using `macaddr_setbit_la` has been verified on my device and kept, but I have removed the other unverified logic and corrected the commit message accordingly.

PR #24512, opened 2026-07-31
qualcommax: qca_edma: give the DSA conduit a stable MAC address

JuliusBairaktaris:
`edma_probe()` calls `eth_hw_addr_random()` unconditionally, so the conduit's MAC changes on every boot — and any DSA user port without a `mac-address` of its own inherits it.

The conduit is a DMA engine sitting behind the switch and has no address of its own, which is why no board describes one for it. So take the address from the switch this conduit serves: every board's CPU port already declares `ethernet = <&edma>`, and its sibling ports carry the board's addresses. That covers boards whose port addresses live in DT and boards where the bootloader patches them in through the `ethernetN` aliases — the latter could not be fixed by adding anything to DT, since the address only exists at runtime. Nothing has to be added per board.

Verified on an AX3600 (ipq8071): `eth0` comes up as `28:d1:27:fc:78:5a` on every boot, the address in the `0:art` cell that the WAN port uses, instead of a fresh random one each time.

@Ansuel @robimarko

### DIFF
--- a/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
+++ b/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
@@ -11,6 +11,7 @@
 #include <linux/module.h>
 #include <linux/of.h>
 #include <linux/of_address.h>
+#include <linux/of_net.h>
 #include <linux/of_platform.h>
 #include <linux/property.h>
 #include <linux/regmap.h>
@@ -1116,6 +1117,42 @@ static const struct regmap_config edma_regmap_cfg = {
 	.val_bits = 32,
 };
 
+/*
+ * The conduit is a DMA engine behind the switch and has no address of its
+ * own, so boards describe none: fall back to the switch this conduit serves,
+ * whose ports carry the board's addresses either from DT or patched in by
+ * the bootloader. DSA user ports without one of their own inherit whatever
+ * ends up here.
+ */
+static int edma_get_mac_address(struct net_device *netdev,
+				struct device_node *np)
+{
+	struct device_node *cpu_port;
+	int ret;
+
+	ret = of_get_ethdev_address(np, netdev);
+	if (!ret || ret == -EPROBE_DEFER)
+		return ret;
+
+	for_each_node_with_property(cpu_port, "ethernet") {
+		struct device_node *conduit __free(device_node) =
+			of_parse_phandle(cpu_port, "ethernet", 0);
+
+		if (conduit != np)
+			continue;
+
+		for_each_available_child_of_node_scoped(cpu_port->parent, port) {
+			ret = of_get_ethdev_address(port, netdev);
+			if (!ret || ret == -EPROBE_DEFER) {
+				of_node_put(cpu_port);
+				return ret;
+			}
+		}
+	}
+
+	return -ENODEV;
+}
+
 static int edma_probe(struct platform_device *pdev)
 {
 	struct clk_bulk_data *clks;
@@ -1158,6 +1195,12 @@ static int edma_probe(struct platform_device *pdev)
 	priv->pdev = pdev;
 	priv->soc = device_get_match_data(dev);
 
+	ret = edma_get_mac_address(netdev, dev->of_node);
+	if (ret == -EPROBE_DEFER)
+		return dev_err_probe(dev, ret, "failed to get MAC address\n");
+	if (ret)
+		eth_hw_addr_random(netdev);
+
 	ret = edma_page_pool_create(priv);
 	if (ret)
 		return ret;
@@ -1168,7 +1211,6 @@ static int edma_probe(struct platform_device *pdev)
 
 	SET_NETDEV_DEV(netdev, dev);
 	netdev->dev.of_node = dev->of_node;
-	eth_hw_addr_random(netdev);
 	netdev->netdev_ops = &edma_netdev_ops;
 	netdev->features = NETIF_F_GRO;
 	netdev->pcpu_stat_type = NETDEV_PCPU_STAT_TSTATS;


target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
```diff
@@ -1116,6 +1117,44 @@ static const struct regmap_config edma_regmap_cfg = {
...
+
+	for_each_node_with_property(cpu_port, "ethernet") {
+		struct device_node *conduit;
+
+		conduit = of_parse_phandle(cpu_port, "ethernet", 0);
+		of_node_put(conduit);
+		if (conduit != np)
+			continue;
+
+		for_each_available_child_of_node(cpu_port->parent, port) {
```
neheb: _scoped?

JuliusBairaktaris: Good catch

neheb (approved):

JuliusBairaktaris:
Ping @jonasjelonek

target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
```diff
@@ -1158,6 +1195,12 @@ static int edma_probe(struct platform_device *pdev)
 	priv->pdev = pdev;
 	priv->soc = device_get_match_data(dev);
 
+	ret = edma_get_mac_address(netdev, dev->of_node);
+	if (ret == -EPROBE_DEFER)
+		return ret;
```
openwrt-ai: The deferral is returned bare, while every other early failure in this function goes through `dev_err_probe()` (ioremap, regmap). It matters more than style here: the deferral no longer originates from this device at all but from an nvmem provider sitting behind a *switch port*, so without `dev_err_probe()` nothing records why edma never probed — `/sys/kernel/debug/devices_deferred` stays empty for it, and if the provider never shows up the board silently ends up with no network instead of the random address it used to get.

```suggestion
	if (ret == -EPROBE_DEFER)
		return dev_err_probe(dev, ret, "failed to get MAC address\n");
```

---

target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
```diff
@@ -1116,6 +1117,42 @@ static const struct regmap_config edma_regmap_cfg = {
...
+
+	for_each_node_with_property(cpu_port, "ethernet") {
+		struct device_node *conduit __free(device_node) =
+			of_parse_phandle(cpu_port, "ethernet", 0);
+
+		if (conduit != np)
+			continue;
+
+		for_each_available_child_of_node_scoped(cpu_port->parent, port) {
```
openwrt-ai, line 1144: The port that ends up lending its address is just the first available child of the `ports` node, so what the conduit inherits differs in role from board to board: on [`ipq8071-ax3600.dtsi` swport2](https://github.com/openwrt/openwrt/blob/fba456c7cf15acd604b7c47323225d4c19adc184/target/linux/qualcommax/dts/ipq8071-ax3600.dtsi#L259-L267) that is the `wan` address (matching what you measured), but on [`ipq8072-dl-wrx36.dts` port@1](https://github.com/openwrt/openwrt/blob/fba456c7cf15acd604b7c47323225d4c19adc184/target/linux/qualcommax/dts/ipq8072-dl-wrx36.dts#L253-L261) it is `lan4`'s, and on [`ipq8072-haze.dts`](https://github.com/openwrt/openwrt/blob/fba456c7cf15acd604b7c47323225d4c19adc184/target/linux/qualcommax/dts/ipq8072-haze.dts#L219-L234) it depends on port@1 being `status = "disabled"`, so it lands on `lan3`.

It is stable per board, which is the goal, but it also means the conduit permanently shares its MAC with one arbitrary user port, and [`dsa_user_host_uc_install()` then skips the `dev_uc_add()` for exactly that port](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/net/dsa/user.c#L371-L375) (v6.12.100). Is "first port in DT order" deliberate, or would a defined rule — lowest `reg`, or the port carrying the label MAC — be preferable so the choice does not silently move when a DTS is reordered or a port is disabled?

---

openwrt-ai (commented):
Reviewed 1 new commit; the message matches the change and no commit-message issues were found. Two inline notes below — one concrete (`dev_err_probe()` on the deferral path), one a question about which switch port's address wins.

Checked the OF refcounting explicitly: `for_each_node_with_property()` needs the manual `of_node_put(cpu_port)` on the early return and it is there, the `__free(device_node)` on `conduit` is scoped per iteration, and `for_each_available_child_of_node_scoped()` cleans up on the `return` from inside the loop — no leaks. Placing the lookup before `edma_page_pool_create()`/`edma_hw_init()` is also right, since the `-EPROBE_DEFER` return needs no unwinding there.

---

JuliusBairaktaris:
The second commit fixes a bug observed by @xlighting2017.


Please don't merge until its re reviewed

target/linux/qualcommax/dts/ipq8072-301w.dts
```diff
@@ -322,6 +322,36 @@
...
+						reg = <0x6 0x6>;
+					};
+
+					macaddr_lan2: macaddr@c {
+						reg = <0xc 0x6>;
+					};
+
+					macaddr_lan1: macaddr@12 {
+						reg = <0x12 0x6>;
+					};
```
openwrt-ai: These two slots are the crux of the commit and the two the commit message says are unverified, and the premise argues against them: if the bootloader really walks the same table and exports slot N as `ethernetN`, then `ethernet2`/`ethernet3` (= `lan2`/`lan1`) not arriving is evidence that slots 2 and 3 are *not* what the vendor provisioned for those ports. Two outcomes follow, and neither is the intended fix:

- the slots are blank/`ff:ff:...`, in which case [`of_get_mac_address_nvmem()` rejects them at of_net.c:88](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/net/core/of_net.c#L88) and `lan1`/`lan2` keep inheriting the conduit — the reported bug is unchanged;
- they hold valid addresses the vendor assigned to something else, in which case `lan1`/`lan2` now duplicate another interface, which is worse than sharing with the conduit.

The "same way" argument from a sibling doesn't pin the mapping down either, because the slot-to-port order is not uniform on ipq807x: [`ipq8071-ax3600.dtsi` starts at 0x6](https://github.com/openwrt/openwrt/blob/7eb2ccef5eed69a304c53866b528260da5a5e80c/target/linux/qualcommax/dts/ipq8071-ax3600.dtsi#L149-L166) and never describes slot 0, while [`ipq8072-dl-wrx36.dts` runs the table backwards](https://github.com/openwrt/openwrt/blob/7eb2ccef5eed69a304c53866b528260da5a5e80c/target/linux/qualcommax/dts/ipq8072-dl-wrx36.dts#L253-L299) — slot 0 to `wan` (port@6), slot 4 to `lan4` (port@1). That is the exact opposite of the assignment here.

Is a `hexdump -C -n 0x24` of `0:art` from @xlighting2017's 301w available, next to the `ip link` MACs it boots with? That would settle slots 2 and 3 directly, rather than leaving the only two ports this patch is supposed to fix resting on inference from a different board.

---

target/linux/qualcommax/dts/ipq8072-301w.dts
```diff
@@ -463,6 +493,8 @@
 		lan4: port@1 {
 			reg = <1>;
 			label = "lan4";
+			nvmem-cells = <&macaddr_lan4>;
+			nvmem-cell-names = "mac-address";
 			phy-handle = <&qca8075_16>;
 			phy-mode = "qsgmii";
```
openwrt-ai, line 467: nit: every other qualcommax DSA port node puts `nvmem-cells`/`nvmem-cell-names` last, after `pcs-handle` — see [`ipq8071-ax3600.dtsi` swport2](https://github.com/openwrt/openwrt/blob/7eb2ccef5eed69a304c53866b528260da5a5e80c/target/linux/qualcommax/dts/ipq8071-ax3600.dtsi#L259-L267), [`ipq8072-dl-wrx36.dts` port@1](https://github.com/openwrt/openwrt/blob/7eb2ccef5eed69a304c53866b528260da5a5e80c/target/linux/qualcommax/dts/ipq8072-dl-wrx36.dts#L253-L261) and [`ipq6010-mango-dvk.dts` port@3](https://github.com/openwrt/openwrt/blob/7eb2ccef5eed69a304c53866b528260da5a5e80c/target/linux/qualcommax/dts/ipq6010-mango-dvk.dts#L278-L286). Here they land between `label` and `phy-handle`, splitting the phy/pcs group.

```suggestion
			phy-handle = <&qca8075_16>;
			phy-mode = "qsgmii";
			pcs-handle = <&uniphy0 0>;
			nvmem-cells = <&macaddr_lan4>;
			nvmem-cell-names = "mac-address";
```

Same reordering applies to the other five ports (`lan3`, `lan2`, `lan1`, `10g-1`, `10g-2`) below.

---

openwrt-ai (commented):
Reviewed the 2 new commits since fba456c. Both commit messages match their diffs, and the `Fixes:` trailer on 7eb2cce resolves correctly to f50435627d37 ("qualcommax: ipq60xx/ipq807x: convert to PPE networking stack"), which is indeed where the shared-conduit-MAC behaviour comes from.

`dev_err_probe()` on the deferral path is applied — thanks. My earlier question about which port lends its address is answered in the thread, so it stays as-is.

Two notes on the new 301w commit below: one on the slot mapping for the two ports the patch is actually meant to fix, one style nit. CI is green on this head.

Checked and found fine: the property-precedence argument in the commit message holds — [`of_get_mac_address()`](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/net/core/of_net.c#L128-L148) tries `mac-address`, `local-mac-address` and `address` before falling back to the nvmem cell (v6.12.100, matching `KERNEL_PATCHVER`), so a bootloader-patched port keeps what it has. The unit addresses all match their `reg` offsets, and `0:ethphyfw1` in this same file already establishes the `nvmem-layout` pattern.

---

JuliusBairaktaris:
@neheb @jonasjelonek this PR branch is back on only containing the reviewed fix. q301w fix here: https://github.com/openwrt/openwrt/pull/24528

neheb (approved):

openwrt-ai (commented):

The head moved from 7eb2cce to 69a40ff by *dropping* a commit rather than adding one: `qualcommax: ipq807x: 301w: read the port MAC addresses from ART` is gone, and `git diff 7eb2cce..69a40ff` is exactly that file's 42 deletions. The remaining commit 69a40ff is the same object I reviewed last time, so the qca_edma.c change is unchanged and nothing here needs re-reviewing.

Both of my earlier notes were on the 301w DTS, so they follow the commit to #24528 rather than applying here. For the record, the open question from that review — whether ART slots 2 and 3 really belong to `lan2`/`lan1` — is answered in that PR's description with the pre-PPE-conversion boot evidence, so it is no longer outstanding on either PR.

Splitting it out is the right call: the 301w change stands on its own and does not depend on this one.

CI is green on 69a40ff — qualcommax ipq50xx, ipq60xx and ipq807x kernel builds all succeeded, kernel-patch check and the three FormalityCheck jobs passed.

---

jonasjelonek:

Razerwire:
This has broken LAN WAN and 10G-Ports for the RT-AX89X 

See #24604

PR #24513, opened 2026-07-31
mediatek: add support for LTC Velo7 Max

dangowrt:
Adds support for the LTC Velo7 Max, an MT7988A based Wi-Fi 7 router, together with the PHY fix its 10G ports turned out to need.

## Hardware

| | |
|---|---|
| SoC | MediaTek MT7988A, 4x Cortex-A73 @ 1.8 GHz |
| RAM | 1/4/8 GiB DDR4 (2x ESMT M16U4G16256A in FCC documents) |
| Flash | 256 MiB Winbond W25N02KV SPI-NAND, managed by NMBM |
| WiFi | MediaTek MT7996 802.11be tri-band (2.4/5/6 GHz), attached to both 2-lane PCIe interfaces |
| Ethernet | 4x 1GbE from the built-in switch, 2x 10GbE via Realtek RTL8261BE using USXGMII |
| USB | 2x USB 3.0 Type-A |
| Buttons | Reset, WPS, Reboot |
| LEDs | Power, Status (RGB), WLAN 2.4/5/6 GHz, WPS, TEL1 and TEL2 (each red and green), plus dual LEDs on all six RJ-45 sockets |
| UART | 115200 8N1, unpopulated 3-pin 2.54mm header on the bottom edge of the board, which stands vertically. The pins are RX, GND, TX, the middle one carrying no label |
| Power | 12 VDC 3 A, barrel connector |

Also sold as the Tozed Kangwei ZLT W19B6VM. FCC ID `2A5LO-VL7M19K`.

eMMC and two telephony SLICs with their RJ-12 sockets are unpopulated on this variant and hence not described.

The serial console has been tested and works as-is; only the header pins themselves need fitting, which pogo pins can avoid too.

The LEDs of the two 10G RJ-45 sockets are left in the default behaviour of the RTL8261BE. The PHY driver offers no means of controlling them, and the part very likely has more LED outputs than the two used per socket, with neither the selection nor the order of them known, so they cannot be described in the device tree at this point.

## `kernel: net: phy: realtek: USXGMII bring-up work-around`

The USXGMII block of the RTL8261N does not always settle once the media side link has been established. Where it does not, the link is brought up and then dropped again shortly afterwards, while autonegotiation on the media side keeps completing normally. `ethtool` therefore reports negotiated speed and duplex next to `Link: no`, and nothing at all is logged.

MediaTek carry a work-around for exactly this in their SDK, written by Bo-Cun Chen, applied from `mtk_mac_link_up()` in the MT7988 ethernet driver. As the quirk is a property of the PHY rather than of the SoC, it is carried in the Realtek driver here instead so that the PHY also behaves on other hosts.

The SDK busy-waits a flat second in the link-up path and then resets the block unconditionally. This version instead polls the SerDes link status through the PHY's indirect register window and only pulses the reset while the link is still down, from a delayed work so that the PHY state machine is never blocked.

It is hooked into both 10 Gbps driver entries which share the base PHY ID. The parts this was observed on carry an `RTL8261BE` package marking while binding to the `RTL8261N` driver entry, the two being told apart solely by a vendor version register, so covering only one of them would be fragile.

### MAC addresses

| | Source |
|---|---|
| LAN | Factory 0x4, minus 2 |
| WAN | Factory 0x4, minus 1 |
| WLAN | Factory 0x4, 0xa and 0x2c0 |

Only the three WLAN addresses are stored in flash, as part of the MT7996 EEPROM. Both Ethernet addresses are derived from the first of them by subtraction, which is how the vendor firmware assigns them. Also matching the vendor firmware, the built-in switch and the 10G LAN port share the LAN address, and the default network configuration bridges `lan0`-`lan3` with `eth1` for LAN and uses `eth2` for WAN, so an existing configuration survives the switch to OpenWrt.

### Bootloader

MediaTek U-Boot using the SDK dual-boot scheme: two sets of UBI volumes, `kernel`/`rootfs` and `kernel2`/`rootfs2`, selected by the U-Boot environment variable `dual_boot.current_slot`. The environment itself lives in the `u-boot-env` UBI volume rather than in a raw MTD partition. U-Boot verifies the kernel FIT and, separately, a hash over the squashfs which it takes from a top-level `rootfs` node of that FIT, falling back to the other slot should either check fail.

`platform_do_upgrade()` therefore writes both volume sets, so that whichever slot the bootloader selects carries the new firmware, and clears any invalid marker left behind by an earlier failed boot. The flashing half is shared with `netgear,eax17`, which uses the same scheme.

### Installation

The vendor firmware is itself OpenWrt based and its `sysupgrade` accepts the OpenWrt image directly.
⚠️***It is patched to not reboot unless `-R` is passed and then just halts/hangs after flashing has completed*** ⚠️, and it always writes to whichever slot is currently inactive:

```
sysupgrade -R openwrt-mediatek-filogic-ltc_vl7m19k-squashfs-sysupgrade.bin
```

The vendor sysupgrade preserves `rootfs_data` no matter what, so run `firstboot` once OpenWrt has come up in order to discard the vendor overlay.

### DIFF
--- a/target/linux/generic/pending-6.18/743-net-phy-realtek-reset-RTL8261N-USXGMII-SerDes-on-lin.patch
+++ b/target/linux/generic/pending-6.18/743-net-phy-realtek-reset-RTL8261N-USXGMII-SerDes-on-lin.patch
@@ -1,0 +1,206 @@
+From 7e7c098948cf877d8ebb7e7c17bfdc82a9764f3c Mon Sep 17 00:00:00 2001
+From: Daniel Golle <daniel@makrotopia.org>
+Date: Fri, 31 Jul 2026 19:13:06 +0100
+Subject: [PATCH] net: phy: realtek: reset RTL8261N USXGMII SerDes on link up
+
+The USXGMII block of the RTL8261N does not always settle by itself once
+the media side link has been established. Where it does not, the link is
+dropped again shortly afterwards while autonegotiation on the media side
+keeps completing normally, so ethtool reports negotiated speed and duplex
+while the link stays down and nothing is logged.
+
+The work-around was originally written by Bo-Cun Chen
+<bc-bocun.chen@mediatek.com> for the MediaTek SDK, which carries it as
+999-2799-net-ethernet-mtk_eth_soc-add-reseting-usxgmii-of-rtl.patch and
+applies it from mtk_mac_link_up() in the MT7988 ethernet driver. As the
+quirk is a property of the PHY rather than of the SoC it is handled here
+instead, so that the PHY also behaves on other hosts.
+
+The SDK busy-waits a flat second in the link-up path and then resets the
+block unconditionally. Poll the SerDes link status through the indirect
+register window instead and only pulse the reset while the link is down,
+from a delayed work so that the PHY state machine is never blocked.
+
+Hook it into both 10Gbps entries which share the base PHY ID. The parts
+this was observed on carry an RTL8261BE package marking while binding to
+the RTL8261N driver entry, the two being told apart solely by a vendor
+version register, so covering only one of them would be fragile.
+
+Signed-off-by: Daniel Golle <daniel@makrotopia.org>
+---
+ drivers/net/phy/realtek/phy_patch_rtl826x.c |  6 +-
+ drivers/net/phy/realtek/phy_patch_rtl826x.h |  4 +
+ drivers/net/phy/realtek/realtek.h           |  4 +
+ drivers/net/phy/realtek/realtek_main.c      | 83 +++++++++++++++++++++
+ 4 files changed, 94 insertions(+), 3 deletions(-)
+
+--- a/drivers/net/phy/realtek/realtek.h
++++ b/drivers/net/phy/realtek/realtek.h
+@@ -11,6 +11,10 @@
+ struct rtl826x_priv {
+ 	struct rtlgen_phy_patch_db *patch;
+ 	bool enable_pma_low_power:1;
++	struct phy_device *phydev;
++	struct delayed_work sds_work;
++	bool sds_work_disabled;
++	unsigned int sds_retries;
+ };
+ 
+ int rtl822x_hwmon_init(struct phy_device *phydev);
+--- a/drivers/net/phy/realtek/realtek_main.c
++++ b/drivers/net/phy/realtek/realtek_main.c
+@@ -230,6 +230,17 @@
+ #define   RTL826X_VEND1_SERDES_GLOBAL_CFG_HSI_INV	BIT(6)
+ #define   RTL826X_VEND1_SERDES_GLOBAL_CFG_HSO_INV	BIT(7)
+ 
++#define RTL826X_SDS_PAGE_LINK			0x05
++#define RTL826X_SDS_REG_LINK			0x00
++#define   RTL826X_SDS_LINK_UP			BIT(12)
++#define RTL826X_SDS_PAGE_RESET			0x20
++#define RTL826X_SDS_REG_RESET			0x02
++#define   RTL826X_SDS_RESET_ASSERT		0x40
++#define   RTL826X_SDS_RESET_RELEASE		0xc0
++
++#define RTL826X_SDS_POLL_MS			100
++#define RTL826X_SDS_MAX_RESETS			20
++
+ #define RTL826X_VEND1_PKG_MODEL			0x103
+ #define RTL826X_VEND1_VERSION_ID		0x104
+ #define   RTL826X_VEND1_VERSION_ID_VARIANT	GENMASK(2, 0)
+@@ -2524,6 +2535,70 @@ static int rtl826x_patch_db_init(struct
+ 		return rtl8264b_phy_patch_db_init(phydev);
+ }
+ 
++/* The USXGMII block does not always settle by itself once the media side
++ * link has been established. Where it does not, the link is dropped again
++ * shortly afterwards while autonegotiation keeps completing normally, so
++ * the failure is otherwise silent. Poll the SerDes link and pulse the
++ * block's reset until it comes up.
++ */
++static void rtl826x_sds_reset_work(struct work_struct *work)
++{
++	struct rtl826x_priv *priv = container_of(to_delayed_work(work),
++						 struct rtl826x_priv, sds_work);
++	struct phy_device *phydev = priv->phydev;
++	int val;
++
++	val = rtl826x_phy_patch_sds_get(phydev, RTL826X_SDS_PAGE_LINK,
++					RTL826X_SDS_REG_LINK);
++	if (val < 0)
++		return;
++
++	if (val & RTL826X_SDS_LINK_UP)
++		return;
++
++	if (++priv->sds_retries > RTL826X_SDS_MAX_RESETS) {
++		phydev_warn(phydev, "SerDes link did not come up\n");
++		return;
++	}
++
++	rtl826x_phy_patch_sds_set(phydev, RTL826X_SDS_PAGE_RESET,
++				  RTL826X_SDS_REG_RESET, RTL826X_SDS_RESET_ASSERT);
++	rtl826x_phy_patch_sds_set(phydev, RTL826X_SDS_PAGE_RESET,
++				  RTL826X_SDS_REG_RESET, RTL826X_SDS_RESET_RELEASE);
++
++	schedule_delayed_work(&priv->sds_work,
++			      msecs_to_jiffies(RTL826X_SDS_POLL_MS));
++}
++
++static void rtl826x_link_change_notify(struct phy_device *phydev)
++{
++	struct rtl826x_priv *priv = phydev->priv;
++
++	if (phydev->state != PHY_RUNNING) {
++		if (!priv->sds_work_disabled) {
++			disable_delayed_work(&priv->sds_work);
++			priv->sds_work_disabled = true;
++		}
++		return;
++	}
++
++	if (priv->sds_work_disabled) {
++		enable_delayed_work(&priv->sds_work);
++		priv->sds_work_disabled = false;
++	}
++
++	priv->sds_retries = 0;
++	mod_delayed_work(system_wq, &priv->sds_work,
++			 msecs_to_jiffies(RTL826X_SDS_POLL_MS));
++}
++
++static void rtl826x_cancel_sds_work(void *data)
++{
++	struct rtl826x_priv *priv = data;
++
++	cancel_delayed_work_sync(&priv->sds_work);
++}
++
+ static int rtl826x_probe(struct phy_device *phydev)
+ {
+ 	struct device *dev = &phydev->mdio.dev;
+@@ -2535,8 +2610,14 @@ static int rtl826x_probe(struct phy_devi
+ 		return -ENOMEM;
+ 
+ 	priv->enable_pma_low_power = device_property_read_bool(dev, "realtek,enable-pma-low-power");
++	priv->phydev = phydev;
+ 	phydev->priv = priv;
+ 
++	INIT_DELAYED_WORK(&priv->sds_work, rtl826x_sds_reset_work);
++	ret = devm_add_action_or_reset(dev, rtl826x_cancel_sds_work, priv);
++	if (ret)
++		return ret;
++
+ 	if (!priv->enable_pma_low_power)
+ 		phydev_warn(phydev, "PMA low-power suspend disabled\n");
+ 
+@@ -3337,6 +3418,7 @@ static struct phy_driver realtek_drvs[]
+ 	}, {
+ 		.name           = "RTL8261BE 10Gbps PHY",
+ 		.config_init    = rtl826x_config_init,
++		.link_change_notify = rtl826x_link_change_notify,
+ 		.probe          = rtl826x_probe,
+ 		.get_features   = rtl826x_get_features,
+ 		.suspend        = rtl826x_suspend,
+@@ -3354,6 +3436,7 @@ static struct phy_driver realtek_drvs[]
+ 	}, {
+ 		.name           = "RTL8261N 10Gbps PHY",
+ 		.config_init    = rtl826x_config_init,
++		.link_change_notify = rtl826x_link_change_notify,
+ 		.probe          = rtl826x_probe,
+ 		.get_features   = rtl826x_get_features,
+ 		.suspend        = rtl826x_suspend,
+--- a/drivers/net/phy/realtek/phy_patch_rtl826x.c
++++ b/drivers/net/phy/realtek/phy_patch_rtl826x.c
+@@ -70,7 +70,7 @@ static int rtl826x_phy_patch_wait(struct
+ 	return 0;
+ }
+ 
+-static int rtl826x_phy_patch_sds_get(struct phy_device *phydev, u32 sds_page, u32 sds_reg)
++int rtl826x_phy_patch_sds_get(struct phy_device *phydev, u32 sds_page, u32 sds_reg)
+ {
+ 	u32 sds_addr = 0x8000 + (sds_reg << 6) + sds_page;
+ 	int ret;
+@@ -86,8 +86,8 @@ static int rtl826x_phy_patch_sds_get(str
+ 	return phy_read_mmd(phydev, MDIO_MMD_VEND1, 0x142);
+ }
+ 
+-static int rtl826x_phy_patch_sds_set(struct phy_device *phydev, u32 sds_page, u32 sds_reg,
+-				     u32 data)
++int rtl826x_phy_patch_sds_set(struct phy_device *phydev, u32 sds_page, u32 sds_reg,
++			      u32 data)
+ {
+ 	u32 sds_addr = 0x8800 + (sds_reg << 6) + sds_page;
+ 	int ret;
+--- a/drivers/net/phy/realtek/phy_patch_rtl826x.h
++++ b/drivers/net/phy/realtek/phy_patch_rtl826x.h
+@@ -5,6 +5,10 @@
+ 
+ #include <linux/phy.h>
+ 
++int rtl826x_phy_patch_sds_get(struct phy_device *phydev, u32 sds_page, u32 sds_reg);
++int rtl826x_phy_patch_sds_set(struct phy_device *phydev, u32 sds_page, u32 sds_reg,
++			      u32 data);
++
+ int rtl8261n_phy_patch_db_init(struct phy_device *phydev);
+ int rtl8264b_phy_patch_db_init(struct phy_device *phydev);
+ 
--- a/target/linux/generic/pending-6.18/743-net-phy-realtek-reset-RTL8261N-USXGMII-SerDes-on-lin.patch
+++ b/target/linux/generic/pending-6.18/743-net-phy-realtek-reset-RTL8261N-USXGMII-SerDes-on-lin.patch
@@ -1,0 +1,193 @@
+From 7e7c098948cf877d8ebb7e7c17bfdc82a9764f3c Mon Sep 17 00:00:00 2001
+From: Daniel Golle <daniel@makrotopia.org>
+Date: Fri, 31 Jul 2026 19:13:06 +0100
+Subject: [PATCH] net: phy: realtek: reset RTL8261N USXGMII SerDes on link up
+
+The USXGMII block of the RTL8261N does not always settle by itself once
+the media side link has been established. Where it does not, the link is
+dropped again shortly afterwards while autonegotiation on the media side
+keeps completing normally, so ethtool reports negotiated speed and duplex
+while the link stays down and nothing is logged.
+
+The work-around was originally written by Bo-Cun Chen
+<bc-bocun.chen@mediatek.com> for the MediaTek SDK, which carries it as
+999-2799-net-ethernet-mtk_eth_soc-add-reseting-usxgmii-of-rtl.patch and
+applies it from mtk_mac_link_up() in the MT7988 ethernet driver. As the
+quirk is a property of the PHY rather than of the SoC it is handled here
+instead, so that the PHY also behaves on other hosts.
+
+The SDK busy-waits a flat second in the link-up path and then resets the
+block unconditionally. Poll the SerDes link status through the indirect
+register window instead and only pulse the reset while the link is down,
+from a delayed work so that the PHY state machine is never blocked.
+
+Hook it into both 10Gbps entries which share the base PHY ID. The parts
+this was observed on carry an RTL8261BE package marking while binding to
+the RTL8261N driver entry, the two being told apart solely by a vendor
+version register, so covering only one of them would be fragile.
+
+Signed-off-by: Daniel Golle <daniel@makrotopia.org>
+---
+ drivers/net/phy/realtek/realtek.h      |   3 +
+ drivers/net/phy/realtek/realtek_main.c | 130 +++++++++++++++++++++++++
+ 2 files changed, 133 insertions(+)
+
+--- a/drivers/net/phy/realtek/realtek.h
++++ b/drivers/net/phy/realtek/realtek.h
+@@ -11,6 +11,9 @@
+ struct rtl826x_priv {
+ 	struct rtlgen_phy_patch_db *patch;
+ 	bool enable_pma_low_power:1;
++	struct phy_device *phydev;
++	struct delayed_work sds_work;
++	unsigned int sds_retries;
+ };
+ 
+ int rtl822x_hwmon_init(struct phy_device *phydev);
+--- a/drivers/net/phy/realtek/realtek_main.c
++++ b/drivers/net/phy/realtek/realtek_main.c
+@@ -218,6 +218,17 @@
+ #define   RTL826X_VEND1_SERDES_GLOBAL_CFG_HSI_INV	BIT(6)
+ #define   RTL826X_VEND1_SERDES_GLOBAL_CFG_HSO_INV	BIT(7)
+ 
++#define RTL826X_SDS_PAGE_LINK			0x05
++#define RTL826X_SDS_REG_LINK			0x00
++#define   RTL826X_SDS_LINK_UP			BIT(12)
++#define RTL826X_SDS_PAGE_RESET			0x20
++#define RTL826X_SDS_REG_RESET			0x02
++#define   RTL826X_SDS_RESET_ASSERT		0x40
++#define   RTL826X_SDS_RESET_RELEASE		0xc0
++
++#define RTL826X_SDS_POLL_MS			100
++#define RTL826X_SDS_MAX_RESETS			20
++
+ #define RTL826X_VEND1_PKG_MODEL			0x103
+ #define RTL826X_VEND1_VERSION_ID		0x104
+ #define   RTL826X_VEND1_VERSION_ID_VARIANT	GENMASK(2, 0)
+@@ -2367,6 +2378,62 @@ static int rtl826x_patch_db_init(struct
+ 		return rtl8264b_phy_patch_db_init(phydev);
+ }
+ 
++/* The USXGMII block does not always settle by itself once the media side
++ * link has been established. Where it does not, the link is dropped again
++ * shortly afterwards while autonegotiation keeps completing normally, so
++ * the failure is otherwise silent. Poll the SerDes link and pulse the
++ * block's reset until it comes up.
++ */
++static void rtl826x_sds_reset_work(struct work_struct *work)
++{
++	struct rtl826x_priv *priv = container_of(to_delayed_work(work),
++						 struct rtl826x_priv, sds_work);
++	struct phy_device *phydev = priv->phydev;
++	int val;
++
++	val = rtl826x_phy_patch_sds_get(phydev, RTL826X_SDS_PAGE_LINK,
++					RTL826X_SDS_REG_LINK);
++	if (val < 0)
++		return;
++
++	if (val & RTL826X_SDS_LINK_UP)
++		return;
++
++	if (++priv->sds_retries > RTL826X_SDS_MAX_RESETS) {
++		phydev_warn(phydev, "SerDes link did not come up\n");
++		return;
++	}
++
++	rtl826x_phy_patch_sds_set(phydev, RTL826X_SDS_PAGE_RESET,
++				  RTL826X_SDS_REG_RESET, RTL826X_SDS_RESET_ASSERT);
++	rtl826x_phy_patch_sds_set(phydev, RTL826X_SDS_PAGE_RESET,
++				  RTL826X_SDS_REG_RESET, RTL826X_SDS_RESET_RELEASE);
++
++	schedule_delayed_work(&priv->sds_work,
++			      msecs_to_jiffies(RTL826X_SDS_POLL_MS));
++}
++
++static void rtl826x_link_change_notify(struct phy_device *phydev)
++{
++	struct rtl826x_priv *priv = phydev->priv;
++
++	if (phydev->state != PHY_RUNNING) {
++		cancel_delayed_work_sync(&priv->sds_work);
++		return;
++	}
++
++	priv->sds_retries = 0;
++	mod_delayed_work(system_wq, &priv->sds_work,
++			 msecs_to_jiffies(RTL826X_SDS_POLL_MS));
++}
++
++static void rtl826x_cancel_sds_work(void *data)
++{
++	struct rtl826x_priv *priv = data;
++
++	cancel_delayed_work_sync(&priv->sds_work);
++}
++
+ static int rtl826x_probe(struct phy_device *phydev)
+ {
+ 	struct device *dev = &phydev->mdio.dev;
+@@ -2378,8 +2445,14 @@ static int rtl826x_probe(struct phy_devi
+ 		return -ENOMEM;
+ 
+ 	priv->enable_pma_low_power = device_property_read_bool(dev, "realtek,enable-pma-low-power");
++	priv->phydev = phydev;
+ 	phydev->priv = priv;
+ 
++	INIT_DELAYED_WORK(&priv->sds_work, rtl826x_sds_reset_work);
++	ret = devm_add_action_or_reset(dev, rtl826x_cancel_sds_work, priv);
++	if (ret)
++		return ret;
++
+ 	if (!priv->enable_pma_low_power)
+ 		phydev_warn(phydev, "PMA low-power suspend disabled\n");
+ 
+@@ -3172,6 +3245,7 @@ static struct phy_driver realtek_drvs[]
+ 	}, {
+ 		.name           = "RTL8261BE 10Gbps PHY",
+ 		.config_init    = rtl826x_config_init,
++		.link_change_notify = rtl826x_link_change_notify,
+ 		.probe          = rtl826x_probe,
+ 		.get_features   = rtl826x_get_features,
+ 		.suspend        = rtl826x_suspend,
+@@ -3189,6 +3263,7 @@ static struct phy_driver realtek_drvs[]
+ 	}, {
+ 		.name           = "RTL8261N 10Gbps PHY",
+ 		.config_init    = rtl826x_config_init,
++		.link_change_notify = rtl826x_link_change_notify,
+ 		.probe          = rtl826x_probe,
+ 		.get_features   = rtl826x_get_features,
+ 		.suspend        = rtl826x_suspend,
+--- a/drivers/net/phy/realtek/phy_patch_rtl826x.c
++++ b/drivers/net/phy/realtek/phy_patch_rtl826x.c
+@@ -70,7 +70,7 @@ static int rtl826x_phy_patch_wait(struct
+ 	return 0;
+ }
+ 
+-static int rtl826x_phy_patch_sds_get(struct phy_device *phydev, u32 sds_page, u32 sds_reg)
++int rtl826x_phy_patch_sds_get(struct phy_device *phydev, u32 sds_page, u32 sds_reg)
+ {
+ 	u32 sds_addr = 0x8000 + (sds_reg << 6) + sds_page;
+ 	int ret;
+@@ -86,7 +86,7 @@ static int rtl826x_phy_patch_sds_get(str
+ 	return phy_read_mmd(phydev, MDIO_MMD_VEND1, 0x142);
+ }
+ 
+-static int rtl826x_phy_patch_sds_set(struct phy_device *phydev, u32 sds_page, u32 sds_reg,
++int rtl826x_phy_patch_sds_set(struct phy_device *phydev, u32 sds_page, u32 sds_reg,
+ 				     u32 data)
+ {
+ 	u32 sds_addr = 0x8800 + (sds_reg << 6) + sds_page;
+--- a/drivers/net/phy/realtek/phy_patch_rtl826x.h
++++ b/drivers/net/phy/realtek/phy_patch_rtl826x.h
+@@ -5,6 +5,10 @@
+ 
+ #include <linux/phy.h>
+ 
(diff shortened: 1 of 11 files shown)


target/linux/mediatek/dts/mt7988a-ltc-vl7m19k.dts
```diff
@@ -0,0 +1,610 @@
...
+
+&pcie0 {
+	status = "okay";
+	pinctrl-names = "default";
+	pinctrl-0 = <&pcie0_pins>;
+
+	pcie@0,0 {
+		reg = <0x0000 0 0 0 0>;
+		wifi-reset-gpios = <&pio 7 GPIO_ACTIVE_LOW>;
+		wifi-reset-msleep = <100>;
```
openwrt-ai: `wifi-reset-gpios` / `wifi-reset-msleep` are placed on the `pcie@0,0` bridge node, but the driver reads them from the PCIe *controller* node: [`mtk_pcie_parse_port()` uses `dev->of_node` with `dev = pcie->dev`](``https://github.com/openwrt/openwrt/blob/1531a3e3cb636424501a4066039150c5b7024d9b/target/linux/mediatek/patches-6.18/966-pcie-mediatek-gen3-Add-WIFI-HW-reset-flow.patch#L59-L72),`` i.e. the `&pcie0` platform device. As written, `of_property_read_u32(..., "wifi-reset-msleep", ...)` never finds the property and the WiFi HW reset is silently skipped.

Both existing in-tree users put them on the controller node — [`mt7988d-keenetic-kn-1812.dtsi:498`](https://github.com/openwrt/openwrt/blob/1531a3e3cb636424501a4066039150c5b7024d9b/target/linux/mediatek/dts/mt7988d-keenetic-kn-1812.dtsi#L495-L499) and [`mt7988a-zbtlink-zbt-z8803be.dts:439`](``https://github.com/openwrt/openwrt/blob/1531a3e3cb636424501a4066039150c5b7024d9b/target/linux/mediatek/dts/mt7988a-zbtlink-zbt-z8803be.dts#L433-L441).``

```suggestion
&pcie0 {
	status = "okay";
	pinctrl-names = "default";
	pinctrl-0 = <&pcie0_pins>;
	wifi-reset-gpios = <&pio 7 GPIO_ACTIVE_LOW>;
	wifi-reset-msleep = <100>;

	pcie@0,0 {
		reg = <0x0000 0 0 0 0>;
```


---

target/linux/generic/pending-6.18/743-net-phy-realtek-reset-RTL8261N-USXGMII-SerDes-on-lin.patch
```diff
@@ -0,0 +1,222 @@
...
++	schedule_delayed_work(&priv->sds_work,
++			      msecs_to_jiffies(RTL826X_SDS_POLL_MS));
++}
++
++static void rtl826x_link_change_notify(struct phy_device *phydev)
++{
++	struct rtl826x_priv *priv = phydev->priv;
++
++	if (phydev->state != PHY_RUNNING) {
++		cancel_delayed_work(&priv->sds_work);
```
openwrt-ai: `cancel_delayed_work()` is asynchronous and does not wait for an instance that is already executing — and `rtl826x_sds_reset_work()` re-arms itself with `schedule_delayed_work()` at the end. So if the work happens to be running when the PHY leaves `PHY_RUNNING`, it re-queues *after* the cancel and keeps pulsing the SerDes reset while the media link is down, until `sds_retries` hits `RTL826X_SDS_MAX_RESETS` (~2 s) and emits a spurious `"SerDes link did not come up"` warning. On the `PHY_HALTED` path that also means MDIO writes racing with `rtl826x_suspend()`.

`cancel_delayed_work_sync()` is safe here: `link_change_notify` is only called from `phy_process_state_change()` in process context (with `phydev->lock` held), and the work only takes the MDIO bus lock, never `phydev->lock` — so there is no lock inversion, and the sync variant also handles the self-requeue. It is what the devm release action already uses.

```suggestion
+		cancel_delayed_work_sync(&priv->sds_work);
```


---

target/linux/mediatek/filogic/base-files/etc/board.d/01_leds
```diff
@@ -161,6 +161,16 @@ livinet,li320)
...
+ltc,vl7m19k)
+	ucidef_set_led_netdev "lan0-green" "LAN0" "mt7530-0:00:green:lan" "lan0" "link_1000 tx rx"
+	ucidef_set_led_netdev "lan0-orange" "LAN0" "mt7530-0:00:orange:lan" "lan0" "link_10 link_100 tx rx"
+	ucidef_set_led_netdev "lan1-green" "LAN1" "mt7530-0:01:green:lan" "lan1" "link_1000 tx rx"
+	ucidef_set_led_netdev "lan1-orange" "LAN1" "mt7530-0:01:orange:lan" "lan1" "link_10 link_100 tx rx"
+	ucidef_set_led_netdev "lan2-green" "LAN2" "mt7530-0:02:green:lan" "lan2" "link_1000 tx rx"
+	ucidef_set_led_netdev "lan2-orange" "LAN2" "mt7530-0:02:orange:lan" "lan2" "link_10 link_100 tx rx"
+	ucidef_set_led_netdev "lan3-green" "LAN3" "mt7530-0:03:green:lan" "lan3" "link_1000 tx rx"
+	ucidef_set_led_netdev "lan3-orange" "LAN3" "mt7530-0:03:orange:lan" "lan3" "link_10 link_100 tx rx"
+	;;
```
openwrt-ai, line 177: nit: only the four switch-port LEDs get triggers here. The DTS also adds `green:wlan-2ghz`, `green:wlan-5ghz` and `green:wlan-6ghz` GPIO LEDs (lines 61-76 of `mt7988a-ltc-vl7m19k.dts`), and the commit message mentions "dual LEDs on all six RJ-45 sockets" — but nothing wires the WLAN LEDs to `phy0.0-ap0` / `phy0.1-ap0` / `phy0.2-ap0`, and the two 10G ports (`eth1`/`eth2`) get no entries either. There is no generic fallback in this file, so those LEDs stay dark. Is that intentional (e.g. the 10G socket LEDs being driven by the RTL8261 itself), or just not done yet?

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -2319,6 +2319,26 @@ define Device/livinet_li320
...
+  DEVICE_DTS := mt7988a-ltc-vl7m19k
+  DEVICE_DTS_DIR := ../dts
+  DEVICE_PACKAGES := kmod-mt7996-firmware kmod-phy-realtek mt7988-wo-firmware rtl826x-firmware uboot-envtools
+  KERNEL = kernel-bin | lzma | \
+	fit-with-netgear-top-level-rootfs-node lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb
+  UBINIZE_OPTS := -E 5
+  BLOCKSIZE := 128k
+  PAGESIZE := 2048
+  KERNEL_IN_UBI := 1
+  UBOOTENV_IN_UBI := 1
```
openwrt-ai: nit: `UBOOTENV_IN_UBI` is only consumed by `Build/append-ubi` and `Build/ubinize-image` ([`include/image-commands.mk:198`](https://github.com/openwrt/openwrt/blob/1531a3e3cb636424501a4066039150c5b7024d9b/include/image-commands.mk#L196-L200), [`:215`](https://github.com/openwrt/openwrt/blob/1531a3e3cb636424501a4066039150c5b7024d9b/include/image-commands.mk#L213-L217)); this device only builds `sysupgrade.bin := sysupgrade-tar | append-metadata`, so it (and `KERNEL_IN_UBI` / `UBINIZE_OPTS`) never takes effect. The commit message states "The environment itself lives in the u-boot-env UBI volume rather than in a raw MTD partition, **which is why UBOOTENV_IN_UBI is set**" — that justification doesn't hold, and `--uboot-env` would in any case create volumes named `ubootenv`/`ubootenv2`, not the `u-boot-env` volume that `package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic` looks up. Worth either dropping the line or fixing the commit message.

---

target/linux/mediatek/dts/mt7988a-ltc-vl7m19k.dts
```diff
@@ -0,0 +1,610 @@
...
+			function-enumerator = <2>;
+			gpios = <&pio 3 GPIO_ACTIVE_HIGH>;
+		};
+
+		led-wps {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WPS;
+			gpios = <&pio 21 GPIO_ACTIVE_HIGH>;
+			default-state = "on";
+		};
```
openwrt-ai, line 107: nit: this WPS LED is missing from the `LEDS:` list in the commit message (and from the LEDs row of the PR body), which enumerates "Power, Status (RGB), WLAN 2.4/5/6 GHz, TEL1 and TEL2 ... plus dual LEDs on all six RJ-45 sockets".

Also, `default-state = "on"` on a WPS LED is unusual — nothing else claims this LED, so it will simply stay lit forever. Is that deliberate (i.e. the GPIO actually drives a general indicator rather than a WPS one)?

---

target/linux/mediatek/dts/mt7988a-ltc-vl7m19k.dts
```diff
@@ -0,0 +1,610 @@
...
+			gpios = <&pio 13 GPIO_ACTIVE_LOW>;
+		};
+
+		button-wps {
+			label = "wps";
+			linux,code = <KEY_WPS_BUTTON>;
+			gpios = <&pio 14 GPIO_ACTIVE_LOW>;
+		};
+
+		button-power {
```
openwrt-ai: nit: the node is named `button-power` but the key is `KEY_POWER2`, which `gpio-button-hotplug` maps to `"reboot"` ([gpio-button-hotplug.c:95](``https://github.com/openwrt/openwrt/blob/1531a3e3cb636424501a4066039150c5b7024d9b/package/kernel/gpio-button-hotplug/src/gpio-button-hotplug.c#L94-L96)),`` and the `label` and the commit message both call it Reboot. Rename for consistency:

```suggestion
		button-reboot {
```


---

target/linux/mediatek/dts/mt7988a-ltc-vl7m19k.dts
```diff
@@ -0,0 +1,610 @@
...
+	};
+
+	pcie1_pins: pcie1-pins {
+		mux {
+			function = "pcie";
+			groups = "pcie_2l_1_pereset";
+		};
+	};
+
+
```
openwrt-ai: nit: stray double blank line.

```suggestion

```

Same at lines 575-576, just before `&pcie0`.

---

openwrt-ai (commented):
## Commit checks

- `1531a3e` "mediatek: add support for LTC Velo7 Max" — the diff does a bit more than the subject says: besides adding the board it refactors the existing `netgear,eax17` flash path into the new shared `mtk_dual_boot_flash_both_slots()` helper and widens `RAMFS_COPY_BIN` / `RAMFS_COPY_DATA` for *every* filogic device. Both changes look right, but they are undocumented — worth a sentence in the message or a separate preparatory commit.
- `1531a3e` — hardware specification omits two items the project's new-device checklist asks for: the UART header location/pinout (only the baud rate is given) and the power input rating (voltage + current). Everything else — flash instructions and MAC address layout — is there.
- `1531a3e` — the message's justification for `UBOOTENV_IN_UBI` does not hold for this image recipe; see the inline comment on `target/linux/mediatek/image/filogic.mk`.

`adf3fa7` matches its changes (the touched `target/linux/realtek/patches-6.18/*` hunks are pure offset refreshes, and the shifts are consistent with the 130 lines the new generic patch inserts).

CI on `1531a3e` was still queued/running at review time, so nothing CI-grounded here.


---

target/linux/mediatek/image/filogic.mk
dangowrt: Dropped all unused flash-describing variables.

target/linux/mediatek/image/filogic.mk
```diff
@@ -2319,6 +2319,21 @@ define Device/livinet_li320
...
 TARGET_DEVICES += livinet_li320
 
+define Device/ltc_vl7m19k
+  DEVICE_VENDOR := LTC
+  DEVICE_MODEL := Velo7 Max
+  DEVICE_ALT0_VENDOR := Tozed Kangwei
+  DEVICE_ALT0_MODEL := ZLT W19B6VM
+  DEVICE_DTS := mt7988a-ltc-vl7m19k
+  DEVICE_DTS_DIR := ../dts
+  DEVICE_PACKAGES := kmod-mt7996-firmware kmod-phy-realtek mt7988-wo-firmware rtl826x-firmware uboot-envtools
```
openwrt-ai: `kmod-usb3` is missing, so the two USB 3.0 ports the DTS enables (`&ssusb0`, `&ssusb1`, `&tphy`, `&xsphy` at lines 361-375 of `mt7988a-ltc-vl7m19k.dts`) will have no host controller driver. On `TARGET_mediatek` `kmod-usb3` is what pulls in the hidden `kmod-usb-xhci-mtk` / `xhci-mtk-hcd.ko` ([`package/kernel/linux/modules/usb.mk:1870`](https://github.com/openwrt/openwrt/blob/c18311132600eb57605e73f9b44c78ca0eeb3465/package/kernel/linux/modules/usb.mk#L1861-L1875)), and it is not in the filogic `DEFAULT_PACKAGES` ([`filogic/target.mk:5`](https://github.com/openwrt/openwrt/blob/c18311132600eb57605e73f9b44c78ca0eeb3465/target/linux/mediatek/filogic/target.mk#L5)).

Every other mt7988 board in this file that enables `ssusb*` lists it — [`zbtlink_zbt-z8803be`](https://github.com/openwrt/openwrt/blob/c18311132600eb57605e73f9b44c78ca0eeb3465/target/linux/mediatek/image/filogic.mk#L3808) and [`smartrg_sdg-8734`](https://github.com/openwrt/openwrt/blob/c18311132600eb57605e73f9b44c78ca0eeb3465/target/linux/mediatek/image/filogic.mk#L368).

```suggestion
  DEVICE_PACKAGES := kmod-mt7996-firmware kmod-phy-realtek kmod-usb3 mt7988-wo-firmware rtl826x-firmware uboot-envtools
```


---

target/linux/mediatek/dts/mt7988a-ltc-vl7m19k.dts
```diff
@@ -0,0 +1,608 @@
...
+	};
+
+	chosen {
+		stdout-path = "serial0:115200n8";
+		bootargs = "console=ttyS0,115200n1 earlycon=uart8250,mmio32,0x11000000 pci=pcie_bus_perf";
+	};
+
+	memory@40000000 {
+		device_type = "memory";
+		reg = <0 0x40000000 0 0x40000000>;
```
openwrt-ai: nit: the PR description says `1/4/8 GiB DDR4 (2x ESMT M16U4G16256A in FCC documents)`, while the commit message says `RAM: 1 GiB DDR4 (2x ESMT M16U4G16256A)` and this node hardcodes 1 GiB — which is what 2x M16U4G16256A (4 Gbit x16 each) actually gives. Worth reconciling the PR body with the commit message; and if 4/8 GiB variants really exist, is the hardcoded size going to be a problem for them, or does the MTK U-Boot fix the `memory` node up?

---

target/linux/mediatek/dts/mt7988a-ltc-vl7m19k.dts
```diff
@@ -0,0 +1,608 @@
...
+	};
+};
+
+&usxgmiisys0 {
+	status = "okay";
+};
+
+&usxgmiisys1 {
+	status = "okay";
+};
```
openwrt-ai: nit: both overrides are no-ops. `usxgmiisys0`/`usxgmiisys1` are added without any `status` property ([`186-arm64-dts-mt7988a-complete-dtsi.patch:75-91`](``https://github.com/openwrt/openwrt/blob/c18311132600eb57605e73f9b44c78ca0eeb3465/target/linux/mediatek/patches-6.18/186-arm64-dts-mt7988a-complete-dtsi.patch#L75-L91)),`` so they are already enabled. The only other in-tree board that drives both USXGMII PCSes, [`mt7988a-arcadyan-mozart.dts:74-80`](https://github.com/openwrt/openwrt/blob/c18311132600eb57605e73f9b44c78ca0eeb3465/target/linux/mediatek/dts/mt7988a-arcadyan-mozart.dts#L74-L80), only touches these nodes when it actually has something to set (`rx-polarity`). Dropping both blocks would be equivalent.

---

openwrt-ai (commented):
Re-reviewed after the force-push (`adf3fa7`/`1531a3e` → `8ba3907`/`c183111`). The previous round's points are all addressed: `wifi-reset-gpios`/`-msleep` moved onto `&pcie0`, `cancel_delayed_work()` → `cancel_delayed_work_sync()`, the WLAN LED triggers added to `01_leds`, the unused `UBINIZE_OPTS`/`BLOCKSIZE`/`PAGESIZE`/`KERNEL_IN_UBI`/`UBOOTENV_IN_UBI` dropped, `button-power` renamed, stray blank lines gone, and the commit message now documents the `netgear,eax17` refactor into `mtk_dual_boot_flash_both_slots()`, the target-wide `RAMFS_COPY_BIN`/`RAMFS_COPY_DATA` widening, and the UART/power specs.

Both commit messages match their diffs. The `target/linux/realtek/patches-6.18/*` offset refreshes in `8ba3907` are consistent with the +1/+19/+102/+6 lines the new generic pending patch inserts above each hunk.

Remaining points inline. CI on `c183111` was still pending at review time, so nothing CI-grounded.

---

target/linux/mediatek/dts/mt7988a-ltc-vl7m19k.dts
```diff
@@ -0,0 +1,595 @@
+// SPDX-License-Identifier: (GPL-2.0 OR MIT)
+
+/dts-v1/;
+#include "mt7988a.dtsi"
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/leds/common.h>
+#include <dt-bindings/phy/phy.h>
```
openwrt-ai: nit: unused include — nothing in this file references a `PHY_TYPE_*` / `PHY_MODE_*` constant (the `phy-mode`/`phy-connection-type` properties on `&gmac1`/`&gmac2` are plain strings).

```suggestion
```

---

target/linux/mediatek/dts/mt7988a-ltc-vl7m19k.dts
```diff
@@ -0,0 +1,595 @@
...
+		spi-rx-bus-width = <4>;
+		spi-tx-bus-width = <4>;
+		mediatek,nmbm;
+		mediatek,bmt-max-ratio = <1>;
+		mediatek,bmt-max-reserved-blocks = <256>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <0x01>;
+			#size-cells = <0x01>;
```
openwrt-ai: nit: `<0x01>` is decompiled-dtb style and is the odd one out in this file — every other cell count here is written plain, including the `nvmem-layout` node 20 lines below ([`#address-cells = <1>` / `#size-cells = <1>`](https://github.com/openwrt/openwrt/blob/7be1270025d6392d7514f8a7d4c1fdaa389f3bd6/target/linux/mediatek/dts/mt7988a-ltc-vl7m19k.dts#L250-L251)) and `pcie@0,0` (`<3>` / `<2>`).

```suggestion
			#address-cells = <1>;
			#size-cells = <1>;
```

Same for `reg = <0x00 0x100000>;` on line 235 — the other seven partitions use `<0xNNNNNN ...>` without the leading-zero padding, so `<0x0 0x100000>` would match.

---

target/linux/mediatek/dts/mt7988a-ltc-vl7m19k.dts
```diff
@@ -0,0 +1,595 @@
...
+	};
+};
+
+&eth {
+	pinctrl-names = "default";
+	pinctrl-0 = <&mdio0_pins>;
+	status = "okay";
+};
+
+&serial0 {
```
openwrt-ai, line 401: nit: the UART pinout disagrees between the two texts describing this header. The commit message says *"The pins are RX, GND, TX, the middle one carrying no label"*, while the PR body's hardware table says *"unpopulated 3-pin 2.54mm header (GND, RX, TX)"*. Since the commit message is what lands in the tree, worth making the PR body match it (or vice versa if the table is the correct order).

---

openwrt-ai (commented):
Re-reviewed after the rebase/force-push (`8ba3907`/`c183111` → `8505df1`/`7be1270`). Compared against my last review at `c183111`, the only substantive changes are:

- `memory@40000000` dropped from the DTS. This is fine — the patched `mt7988a.dtsi` carries no memory node either, and [`mt7988a-arcadyan-mozart.dts`](https://github.com/openwrt/openwrt/blob/7be1270025d6392d7514f8a7d4c1fdaa389f3bd6/target/linux/mediatek/dts/mt7988a-arcadyan-mozart.dts) is the same shape, so the "sized by U-Boot which fills in the memory node" line in the commit message is now consistent with the DTS. It also resolves the 1/4/8 GiB variant question from last round.
- `&usxgmiisys0` / `&usxgmiisys1` no-op overrides dropped.
- `DEVICE_PACKAGES`: `kmod-usb3` added, `uboot-envtools` dropped — correct, `uboot-envtools` is already in [`filogic/target.mk`'s `DEFAULT_PACKAGES`](https://github.com/openwrt/openwrt/blob/7be1270025d6392d7514f8a7d4c1fdaa389f3bd6/target/linux/mediatek/filogic/target.mk#L5-L7). My earlier suggestion block carried it along by mistake; dropping it is right.

Both commit messages match their diffs, so no Commit checks section. The four `target/linux/realtek/patches-6.18/*` refreshes in `8505df1` are still exactly +1 / +20 / +128 relative to the rebased base, matching the 1 + 19 + 102 + 6 lines the new generic pending patch inserts above each hunk — no other in-tree patch touches `drivers/net/phy/realtek/realtek_main.c` after the generic series, so nothing else needs refreshing.

Only cosmetic points left, inline. CI on `7be1270` was still pending at review time, so nothing CI-grounded.

---

jonasjelonek (approved):
LGTM

plappermaul:
@l0rdg3x you own a Realtek switch with a RTL826x PHY. Please check if that PR gives any issues.

plappermaul (approved):
Only code review done. Nothing to complain about.

dangowrt:
@plappermaul @l0rdg3x The one weird thing worth telling to you folks working on the RealTek switches: The PHY chips are labeled as RTL8161BE but the driver detects them as  RTL8161N, so probably our logic how to distinguish the two from each other is wrong...

l0rdg3x:
Sorry, I'll go home tomorrow night and start testing again.

openwrt-ai (commented):

Re-reviewed after the force-push (`7be1270` → `1af0636`). `8505df1` is unchanged, and the only delta in the second commit is the two cosmetic points from my last round: the unused `#include <dt-bindings/phy/phy.h>` dropped, and `#address-cells`/`#size-cells` on `partitions` plus `reg` on `partition@0` rewritten in the plain style the rest of the file uses. The commit message is byte-identical and still matches its diff, so no Commit checks section. The PR body's UART row now reads "The pins are RX, GND, TX, the middle one carrying no label", matching the commit message — that closes the last open nit.

One note on the bot's formality warning: `target/linux/realtek/patches-6.18/743-net-realtek-serdes-configuration.patch` is a pre-existing in-tree patch (last touched by `7105bec48f` "kernel: bump 6.18 to 6.18.40"); `8505df1` only refreshes its hunk offsets, so the missing `git am` header is not something this PR introduced.

CI on `1af0636` was still pending with no failed checks at review time, so nothing CI-grounded.

---

plappermaul:
> The PHY chips are labeled as RTL8161BE but the driver detects them as RTL8161N

From looking at the code realtek_main is simply missing the RTL8261B (without E) detection. It currently says "what is not RTL8261BE must be RTL8261N".

target/linux/mediatek/filogic/base-files/lib/upgrade/platform.sh
```diff
@@ -1,5 +1,6 @@
 REQUIRE_IMAGE_METADATA=1
-RAMFS_COPY_BIN='fitblk fit_check_sign'
+RAMFS_COPY_BIN='fitblk fit_check_sign fw_printenv fw_setenv'
+RAMFS_COPY_DATA='/etc/fw_env.config /var/lock/fw_printenv.lock'
```
hauke: These files are already installed by `package/base-files/files/lib/upgrade/stage2` except `/var/lock/fw_printenv.lock`. `fw_setenv` is installed as data which is strange.

target/linux/mediatek/dts/mt7988a-ltc-vl7m19k.dts
```diff
@@ -0,0 +1,594 @@
...
+	pinctrl-0 = <&pcie1_pins>;
+};
+
+&pcie2 {
+	status = "disabled";
+};
+
+&pcie3 {
+	status = "disabled";
+};
```
hauke: `pcie2` and `pcie3` are already disabled in the `arch/arm64/boot/dts/mediatek/mt7988a.dtsi `

target/linux/generic/pending-6.18/743-net-phy-realtek-reset-RTL8261N-USXGMII-SerDes-on-lin.patch
```diff
@@ -0,0 +1,222 @@
...
++		return ret;
++
++	ret = rtl826x_sds_wait(phydev);
++	if (ret)
++		return ret;
++
++	return phy_read_mmd(phydev, MDIO_MMD_VEND1, RTL826X_VEND1_SDS_DATA_OUT);
++}
++
++static int rtl826x_sds_write(struct phy_device *phydev, u8 page, u8 reg, u16 val)
```
hauke: This function is similar to `rtl826x_phy_patch_sds_set` from `target/linux/generic/pending-6.18/742-net-phy-realtek-add-5G-and-10G-PHY-support.patch`. Maybe unify this?

target/linux/generic/pending-6.18/743-net-phy-realtek-reset-RTL8261N-USXGMII-SerDes-on-lin.patch
```diff
@@ -0,0 +1,193 @@
...
+Hook it into both 10Gbps entries which share the base PHY ID. The parts
+this was observed on carry an RTL8261BE package marking while binding to
+the RTL8261N driver entry, the two being told apart solely by a vendor
+version register, so covering only one of them would be fragile.
+
+Signed-off-by: Daniel Golle <daniel@makrotopia.org>
+---
+ drivers/net/phy/realtek/realtek.h      |   3 +
+ drivers/net/phy/realtek/realtek_main.c | 130 +++++++++++++++++++++++++
+ 2 files changed, 133 insertions(+)
```
openwrt-ai, line 33: nit: the diffstat is stale after the switch to the shared `rtl826x_phy_patch_sds_*()` helpers — the patch now touches four files, not two, and `realtek_main.c` gains 75 lines rather than 130.

```suggestion
 drivers/net/phy/realtek/realtek.h           |   3 +
 drivers/net/phy/realtek/realtek_main.c      |  75 +++++++++++++++++++++
 drivers/net/phy/realtek/phy_patch_rtl826x.c |   4 +-
 drivers/net/phy/realtek/phy_patch_rtl826x.h |   4 +
 4 files changed, 84 insertions(+), 2 deletions(-)
```

---

target/linux/generic/pending-6.18/743-net-phy-realtek-reset-RTL8261N-USXGMII-SerDes-on-lin.patch
```diff
@@ -0,0 +1,193 @@
...
+ {
+ 	u32 sds_addr = 0x8000 + (sds_reg << 6) + sds_page;
+ 	int ret;
+@@ -86,7 +86,7 @@ static int rtl826x_phy_patch_sds_get(str
+ 	return phy_read_mmd(phydev, MDIO_MMD_VEND1, 0x142);
+ }
+ 
+-static int rtl826x_phy_patch_sds_set(struct phy_device *phydev, u32 sds_page, u32 sds_reg,
++int rtl826x_phy_patch_sds_set(struct phy_device *phydev, u32 sds_page, u32 sds_reg,
+ 				     u32 data)
```
openwrt-ai, line 178: nit: dropping `static ` shortens the line by 7 characters, but the continuation line is left at the old alignment (4 tabs + 5 spaces, matching `static int rtl826x_phy_patch_sds_set(`), so it no longer lines up with the open parenthesis — checkpatch will flag it. The declaration this patch adds to `phy_patch_rtl826x.h` two hunks below already uses the correct 3 tabs + 6 spaces.

Converting the context line into a `-`/`+` pair keeps the `@@ -86,7 +86,7 @@` counts valid:

```suggestion
-static int rtl826x_phy_patch_sds_set(struct phy_device *phydev, u32 sds_page, u32 sds_reg,
-				     u32 data)
+int rtl826x_phy_patch_sds_set(struct phy_device *phydev, u32 sds_page, u32 sds_reg,
+			      u32 data)
```

---

target/linux/generic/pending-6.18/743-net-phy-realtek-reset-RTL8261N-USXGMII-SerDes-on-lin.patch
```diff
@@ -0,0 +1,193 @@
...
++ */
++static void rtl826x_sds_reset_work(struct work_struct *work)
++{
++	struct rtl826x_priv *priv = container_of(to_delayed_work(work),
++						 struct rtl826x_priv, sds_work);
++	struct phy_device *phydev = priv->phydev;
++	int val;
++
++	val = rtl826x_phy_patch_sds_get(phydev, RTL826X_SDS_PAGE_LINK,
++					RTL826X_SDS_REG_LINK);
```
openwrt-ai, line 85: Reusing the patch-engine accessors changes the worst-case blocking time by three orders of magnitude, which may be worth a second look. The previous local `rtl826x_sds_wait()` used `phy_read_mmd_poll_timeout(..., 100, 10000, false)` — 10 ms cap. `rtl826x_phy_patch_sds_get()`/`_set()` wait via [`rtl826x_phy_patch_wait()`](``https://github.com/openwrt/openwrt/blob/f3a12300da3bf2df26845ffbcb7246b6e99cde5f/target/linux/generic/pending-6.18/742-net-phy-realtek-add-5G-and-10G-PHY-support.patch#L946-L977),`` which polls with `msleep(1)` up to `PHY_PATCH_WAIT_TIMEOUT` = 10 s per access.

That matters for `cancel_delayed_work_sync()` in `rtl826x_link_change_notify()` below: `link_change_notify` is invoked from [`phy_process_state_change()`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/phy.c#L64-L73), which both `_phy_state_machine()` and [`phy_stop()`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/phy.c#L1646-L1658) call with `phydev->lock` held. If the work happens to be inside a stuck indirect access when the link drops, the state machine (and any concurrent ethtool call taking `phydev->lock`) now stalls for up to 10 s instead of 10 ms. The `RTL826X_SDS_MAX_RESETS` budget likewise turns from ~2 s into potentially minutes.

If the BUSY bit is expected to clear in microseconds, is it worth giving the accessors a caller-supplied timeout, or keeping a short-timeout wrapper for this path?

---

openwrt-ai (commented):
Re-reviewed after the force-push and rebase (`8505df1`/`1af0636` → `8cf472b`/`f3a1230`). Ignoring the ~150 files the rebase pulled in, the PR-only delta against my last review at `1af0636` is three changes, all addressing @hauke's round:

- The local SerDes indirect-window accessors are gone from `realtek_main.c`; the patch now un-statics `rtl826x_phy_patch_sds_get()`/`_set()` in `phy_patch_rtl826x.c` and declares them in `phy_patch_rtl826x.h`. Build-wise that is fine — [`phy_patch_rtl826x.o` is unconditionally part of `realtek-y`](``https://github.com/openwrt/openwrt/blob/f3a12300da3bf2df26845ffbcb7246b6e99cde5f/target/linux/generic/pending-6.18/742-net-phy-realtek-add-5G-and-10G-PHY-support.patch#L26-L33)`` and `realtek.h` already pulls in `phy_patch_rtl826x.h`, so no `EXPORT_SYMBOL` is needed. The address encodings also match what the removed code did (`0x8000` = BUSY, `0x8800` = BUSY|WRITE, `reg << 6 | page`). One consequence worth a look, inline.
- `&pcie2`/`&pcie3` overrides dropped from the DTS — correct, both nodes are already `status = "disabled"` in [`mt7988a.dtsi`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/mediatek/mt7988a.dtsi#L474-L538) (v6.18.41, matching `KERNEL_PATCHVER:=6.18`).
- The `RAMFS_COPY_BIN`/`RAMFS_COPY_DATA` widening dropped from `platform.sh` — also correct: `switch_to_ramfs()` already installs [`/usr/sbin/fw_printenv`](https://github.com/openwrt/openwrt/blob/f3a12300da3bf2df26845ffbcb7246b6e99cde5f/package/base-files/files/lib/upgrade/stage2#L49), [`/usr/sbin/fw_setenv` and `/etc/fw_env.config`](https://github.com/openwrt/openwrt/blob/f3a12300da3bf2df26845ffbcb7246b6e99cde5f/package/base-files/files/lib/upgrade/stage2#L58-L59), and it does [`mkdir -p $RAM_ROOT/var/lock`](https://github.com/openwrt/openwrt/blob/f3a12300da3bf2df26845ffbcb7246b6e99cde5f/package/base-files/files/lib/upgrade/stage2#L61), so `fw_setenv`'s lock file is covered too.

## Commit checks

- `f3a1230` "mediatek: add support for LTC Velo7 Max" — the Bootloader section still ends with *"Clearing the markers needs fw_setenv, hence fw_printenv, fw_setenv and /etc/fw_env.config are added to the sysupgrade ramfs for the whole target."* That sentence described the `RAMFS_COPY_BIN`/`RAMFS_COPY_DATA` hunk this push (correctly) dropped; the diff no longer touches those variables. Worth deleting it.

`8cf472b`'s message still matches its diff — the message describes the work-around generically and does not name the files, so moving the register accessors doesn't contradict it (the patch's own diffstat does, see inline).

Remaining points inline. No failed check runs on `f3a1230` at review time, so nothing CI-grounded.

---

l0rdg3x:
Tested on an RTL9303 (Hasivo S1100W-8XGT-SE, 8x RTL8264 10G PHY, USXGMII), kernel 6.18.41,
RAM-booted: `main` 712bfddc81 against the same tree with 8cf472b cherry-picked. No issues here.

All eight PHYs bind to the `RTL8264 10Gbps PHY` entry, which carries no `.link_change_notify`, so
the delayed work is never scheduled and what my board actually exercises is `rtl826x_probe()` plus
the two symbols un-static'd in `phy_patch_rtl826x.c`. Ping across the switch is unchanged (60/60 and
20/20 with `-s 1472`, 0% loss, 0.544 vs 0.548 ms average), a port flap comes back at the same speed,
`eth0` rx_errors stays 0, and "SerDes link did not come up" never appears.

No RTL8261 here, so this says nothing about the BE/N detection question.

target/linux/realtek/patches-6.18/024-02-v7.1-net-phy-realtek-add-RTL8224-pair-order-support.patch
```diff
@@ -121,7 +121,7 @@ Signed-off-by: Jakub Kicinski <kuba@kernel.org>
  static bool rtlgen_supports_2_5gbps(struct phy_device *phydev)
  {
  	int val;
-@@ -3087,6 +3150,8 @@ static struct phy_driver realtek_drvs[]
+@@ -3160,6 +3223,8 @@ static struct phy_driver realtek_drvs[]
```
openwrt-ai: This hunk header is still stale — `Check Kernel patches (realtek, rtl838x)` fails on `8ccd066` because `make target/linux/refresh` produces different offsets for this file:

```
target/linux/realtek/patches-6.18/024-02-v7.1-net-phy-realtek-add-RTL8224-pair-order-support.patch
target/linux/realtek/patches-6.18/030-v7.2-net-phy-realtek-support-MDI-swapping-for-RTL8226-CG.patch
```

The tell is that both flagged files moved their `realtek_drvs[]` hunk by +73 (`3087` → `3160` here, `3074` → `3147` in the 030 patch), while the new `pending-6.18/743-net-phy-realtek-reset-RTL8261N-USXGMII-SerDes-on-lin.patch` inserts 81 lines into `realtek_main.c` above that point — 11 at `@@ -218,6 +218,17`, 64 at `@@ -2367,6 +2378,70` and 6 at `@@ -2378,8 +2453,14`; the two `+1` hunks in the driver table sit below it. The first hunk of each file (+11) is consistent, and the other two patches this commit touches (`024-04-…-polarity-support` and `743-net-realtek-serdes-configuration`) are not flagged.

Per the project rules these are quilt-managed, so regenerating with `make target/linux/refresh` (with realtek/rtl838x selected in `.config`) is the fix — same applies to `030-v7.2-net-phy-realtek-support-MDI-swapping-for-RTL8226-CG.patch`. The [failing job](https://github.com/openwrt/openwrt/actions/runs/32970657808/job/98185237918) also uploads a `realtek-rtl838x-refreshed` artifact containing the two corrected files if that is easier than rebuilding.

---

openwrt-ai (commented):
Re-reviewed after the rebase/force-push (`8cf472b`/`f3a1230` → `0215740`/`8ccd066`). Setting the rebase aside, the PR-only delta against my last review at `f3a1230` is three things, and all three address the previous round:

- `cancel_delayed_work_sync()` in `rtl826x_link_change_notify()` replaced by `disable_delayed_work()` / `enable_delayed_work()` behind a new `priv->sds_work_disabled` flag. This resolves the blocking question from last round — the only `_sync` cancel left is the devm release action, off the `phydev->lock` path. The bookkeeping also looks right: the flag keeps the disable depth balanced (so the `WORK_OFFQ_DISABLE_BITS` counter can't run away), the work's self-requeue via `schedule_delayed_work()` becomes a no-op while disabled ([`clear_pending_if_disabled()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/kernel/workqueue.c#L2384-L2396), v6.18.44, matching `KERNEL_PATCHVER:=6.18`), and `enable_delayed_work()` does not re-queue on its own, which is why the following [`mod_delayed_work()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/kernel/workqueue.c#L2619-L2632) is needed. `priv` is `devm_kzalloc()`'d, so both new fields start out zeroed.
- `rtl826x_phy_patch_sds_set()`'s continuation line re-indented to the open parenthesis, and the patch diffstat updated — `4 files changed, 94 insertions(+), 3 deletions(-)` matches the hunks now (4 + 4 + 83 + 3 added, 3 removed).
- The `RAMFS_COPY_BIN`/`RAMFS_COPY_DATA` sentence dropped from the Bootloader section of `8ccd066`'s message, which was my only Commit checks item last round. Both commit messages match their diffs now, so no Commit checks section.

One remaining point inline.

---

openwrt-ai (commented):

---

PR #24514, opened 2026-08-01
kernel: back port aquantia USXGMII in-band autoneg to 6.12

JuliusBairaktaris:
On 6.12 the Aquantia driver implements neither `inband_caps` nor `config_inband`, so
`phy_inband_caps()` returns 0, `pl->phy_ib_mode` stays 0 and phylink never calls
`phy_config_inband()` on an AQR. The USXGMII system-side autoneg bit is instead forced
on unconditionally from `aqr107_read_status()` by
`qualcommax/patches-6.12/0760-net-phy-aquantia-enable-USXGMII-MAC-autoneg.patch`, so it
stays set whatever negotiation mode phylink resolved.

This back ports the USXGMII half of upstream 5d59109d47c0 ("net: phy: aquantia: report
and configure in-band autoneg capabilities", v6.18) so the PHY side follows phylink,
and drops the qualcommax patch it replaces.

### Why only the USXGMII half

Upstream's SGMII arm walks `priv->global_cfg[]`, a shadow of the Global System
Configuration registers introduced by the same v6.18 series that renamed these
functions to `aqr_gen2_*`. That shadow does not exist on 6.12, and back porting it
means 31 commits and ~800 changed lines in `aquantia_main.c` — four of them
rename/merge commits that restructure exactly the driver entries and functions
`hack-6.12/722`, `723` and `725` patch.

The USXGMII and 10G-QXGMII arm has no such dependency — upstream handles both in the
same `4.C441` branch and neither touches the shadow — so `inband_caps` reports those two
only and SGMII links keep the behaviour they have today. AQR412C and AQR115 postdate 6.12, so
twelve of upstream's fourteen driver entries are wired up here.

### Stacked on #24420

Without `managed = "in-band-status"` phylink resolves a USXGMII port to
`PHYLINK_PCS_NEG_OUTBAND` and passes `LINK_INBAND_DISABLE`, clearing the bit 0760 was
forcing on. On current main nine of the ten qualcommax USXGMII ports lack that property
— only `ipq8074-rt-ax89x.dts` has it — so this cannot land before #24420, which adds it
to all ten. The first six commits here are that PR; only the last one is new.

No board outside qualcommax is reached: qualcommax is the only target still on
`KERNEL_PATCHVER 6.12` with a `usxgmii` port at all, and the other targets that have
one — mediatek and airoha — build 6.18 only, where the driver already behaves this way.

### Testing

Not tested on hardware — the 10G AQR boards this affects are being run on 6.18 by their
owners, so no 6.12 tester is available. Verified by build: `make target/linux/refresh`
is clean (the shifted `hack-6.12/722`, `723` and `725` are refreshed in the same
commit), and the resulting `aquantia_main.c` in `build_dir` matches the intended
post-image byte for byte.

@robimarko — you mentioned having a back port series for this. Happy to drop this in
favour of yours, or rebase on it, whichever you prefer.


target/linux/generic/backport-6.12/734-v6.18-net-phy-aquantia-report-and-configure-in-band-autone.patch
```diff
@@ -0,0 +1,211 @@
...
++	u16 set = 0;
++
++	if (phydev->interface != PHY_INTERFACE_MODE_USXGMII)
++		return 0;
++
++	if (modes == LINK_INBAND_ENABLE)
++		set = MDIO_PHYXS_VEND_PROV2_USX_AN;
++
++	return phy_modify_mmd(phydev, MDIO_MMD_PHYXS, MDIO_PHYXS_VEND_PROV2,
++			      MDIO_PHYXS_VEND_PROV2_USX_AN, set);
```
openwrt-ai, line 101: `PHY_INTERFACE_MODE_10G_QXGMII` is dropped along with the SGMII arm, but it does not belong to it. In [upstream's `aqr_gen2_inband_caps()`/`aqr_gen2_config_inband()`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/net/phy/aquantia/aquantia_main.c#L1116-L1141) 10G-QXGMII sits in the *same* `4.C441` / `MDIO_PHYXS_VEND_PROV2_USX_AN` branch as USXGMII and never touches `priv->global_cfg[]` — so the justification given in the backport note below and in the commit message ("upstream's SGMII arm walks `priv->global_cfg[]` … that shadow does not exist on 6.12") does not cover this one.

It is harmless as the tree stands: [`aqr107_config_init()` on 6.12](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/drivers/net/phy/aquantia/aquantia_main.c#L493-L504) rejects `10G_QXGMII` with `-ENODEV`, so no AQR can ever bind in that mode there, and no in-tree 6.12 board asks for it. But since it costs two `||` arms and would then match upstream verbatim, either carry it or say in the backport note why it was left out — otherwise a later resync reads this as an accidental divergence.

---

target/linux/generic/hack-6.12/725-net-phy-aquantia-add-PHY_IDs-for-AQR112-variants.patch
```diff
@@ -21,9 +21,9 @@ Signed-off-by: Daniel Golle <daniel@makrotopia.org>
  
  #define MDIO_PHYXS_VEND_IF_STATUS		0xe812
  #define MDIO_PHYXS_VEND_IF_STATUS_TYPE_MASK	GENMASK(7, 3)
-@@ -1205,6 +1207,30 @@ static struct phy_driver aqr_driver[] =
- 	.led_hw_control_get = aqr_phy_led_hw_control_get,
- 	.led_polarity_set = aqr_phy_led_polarity_set,
+@@ -1255,6 +1257,30 @@ static struct phy_driver aqr_driver[] =
+ 	.inband_caps	= aqr_gen2_inband_caps,
+ 	.config_inband	= aqr_gen2_config_inband,
```
openwrt-ai, line 26: nit: the refresh pulls `.inband_caps`/`.config_inband` into the context here, but the two OpenWrt-local entries this patch *adds* right below (`PHY_ID_AQR112C`, `PHY_ID_AQR112R`, lines 28-51) don't get them, while the 6.18 sibling of this same patch does — [`hack-6.18/725`](``https://github.com/openwrt/openwrt/blob/54862d6ac4b1ce3a53ceff4fa73502e76c26aec3/target/linux/generic/hack-6.18/725-net-phy-aquantia-add-PHY_IDs-for-AQR112-variants.patch#L50-L51)`` wires `aqr_gen2_inband_caps`/`aqr_gen2_config_inband` into both.

No in-tree 6.12 board is affected (every qualcommax AQR is an AQR111B0/AQR113C/AQR114C, all of which the backport covers, and qualcommax is the only 6.12 target with a `usxgmii` port), so this is a 6.12/6.18 parity gap rather than a regression. Still, an AQR112C/AQR112R in USXGMII on 6.12 would now be the one case where nothing sets `MDIO_PHYXS_VEND_PROV2_USX_AN` at all, since `0760` is gone. Intentional?

---

target/linux/generic/backport-6.12/734-v6.18-net-phy-aquantia-report-and-configure-in-band-autone.patch
openwrt-ai, line 1: nit: the `730-NN` series is otherwise ordered by upstream version — `730-01`..`730-08` are v6.13, `730-09` v6.18, `730-10` v6.19, `730-11` v7.0. Slotting a v6.18 patch in at `730-12` puts it after the v7.0 one, which is the only out-of-order entry in `backport-6.12/`. Nothing depends on it (the `730-05`..`730-11` block is mxl-gpy and never touches `aquantia_main.c`), so it applies fine either way, but a number outside the mxl-gpy tail — e.g. `731-`, or renumbering into the aquantia `730-01`..`730-04` group — would keep the sort meaningful.

---

openwrt-ai (commented):
Fresh review of this PR. The first six commits are #24420 unchanged (already reviewed there at `da7d3c7e`), so this pass concentrates on `54862d6`.

The mechanics of the back port check out. `phy_inband_caps()`/`phy_config_inband()` and the phylink side that drives them are already on 6.12 via `601-04`, `601-07`, `601-12` and `605-02`, and `phylink_get_inband_type()` there classifies `PHY_INTERFACE_MODE_USXGMII` as `INBAND_CISCO_SGMII`, so the new ops really do get called. The "twelve of upstream's fourteen entries" count is right — upstream skips AQ1202/AQ2104/AQR105/AQR106 and adds AQR412C/AQR115, neither of which exists on 6.12. The three refreshed `hack-6.12` patches are arithmetically consistent with the new patch (+3 for the defines, +23 for the two functions, +2 per preceding driver entry: `722` shifts by 36 then 38, `725` by 50), and `MDIO_PHYXS_VEND_PROV2` is not defined by any other 6.12 patch now that `0760` is gone, so there is no duplicate. Dropping `0760` is also consistent with `patches-6.18`, which never carried it. CI on `54862d6` has no completed checks yet (pending, nothing failed).

Three comments inline: one on the `10G_QXGMII` half of the upstream arm, plus two nits.

## Commit checks

- `54862d6` "kernel: back port aquantia USXGMII in-band autoneg to 6.12" — the closing paragraph about `mt7988a-smartrg-mt-stuart` describes an effect this patch cannot have. The patch lives in `target/linux/generic/backport-6.12/`, and mediatek builds 6.18 only ([`KERNEL_PATCHVER:=6.18`](https://github.com/openwrt/openwrt/blob/54862d6ac4b1ce3a53ceff4fa73502e76c26aec3/target/linux/mediatek/Makefile#L11); the target has only `patches-6.18`, `config-6.18` and `files-6.18`, no 6.12 tree and no `KERNEL_TESTING_PATCHVER`). Of the sixteen targets still on 6.12, qualcommax is the only one with any `usxgmii` port at all. So "it now has the bit cleared where it was previously left alone" is not true of this change — which the next sentence, "that is already how it behaves on 6.18", effectively says itself. Either drop the paragraph or reword it to state that no board outside qualcommax is reached. The same two sentences appear in the PR description.

---

target/linux/generic/backport-6.12/734-v6.18-net-phy-aquantia-report-and-configure-in-band-autone.patch
```diff
@@ -0,0 +1,213 @@
...
+
+Signed-off-by: Vladimir Oltean <vladimir.oltean@nxp.com>
+Link: https://patch.msgid.link/20250903130730.2836022-5-vladimir.oltean@nxp.com
+Signed-off-by: Jakub Kicinski <kuba@kernel.org>
+---
+Backport note: only the USXGMII half is carried. Upstream's SGMII arm walks
+priv->global_cfg[], a shadow of the Global System Configuration registers added
+by the same v6.18 series that renamed these functions to aqr_gen2_*; that shadow
+does not exist on 6.12. inband_caps therefore reports USXGMII alone, leaving
+SGMII links exactly as they behave today.
```
openwrt-ai: nit: the backport note is now stale. Since `10G_QXGMII` was added to both arms (lines 82 and 93 of this file), `inband_caps` no longer reports "USXGMII alone" — it reports USXGMII and 10G-QXGMII. The commit message was updated for this ("10G-QXGMII is carried along with USXGMII…"), but the note and the PR description ("so `inband_caps` reports USXGMII alone and SGMII links keep the behaviour they have today") still carry the pre-change wording. The note is the copy that ships in-tree and is what a later resync will read.

```suggestion
does not exist on 6.12. inband_caps therefore reports USXGMII and 10G-QXGMII
only, leaving SGMII links exactly as they behave today.
```

---

openwrt-ai (commented):
Re-review of the delta since `54862d6` (the first six commits are still #24420 unchanged). The new revision only touches the last commit: it carries the `10G_QXGMII` arms, wires `inband_caps`/`config_inband` into the two OpenWrt-local AQR112C/AQR112R entries in `hack-6.12/725`, and renumbers the backport to `734-`. All three points from the previous pass are addressed, and the commit message no longer makes the mediatek claim.

Re-verified:

- The carried arm now matches [upstream `aqr_gen2_inband_caps()`/`aqr_gen2_config_inband()`](https://github.com/gregkh/linux/blob/f89c296854b755a66657065c35b05406fc18264d/drivers/net/phy/aquantia/aquantia_main.c#L1116-L1141) for the non-SGMII path exactly, register, mask and `set` semantics included. `PHY_INTERFACE_MODE_10G_QXGMII` exists on 6.12 ([`phy.h:172`](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/include/linux/phy.h#L172)), and [`aqr107_config_init()` there](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/drivers/net/phy/aquantia/aquantia_main.c#L493-L509) still rejects it with `-ENODEV`, so the commit message's reasoning holds.
- Hunk-header arithmetic across the three refreshed patches is consistent. `734` grows the function block by 2 (`+812,29` → `+812,31`) and shifts each of its own twelve entry hunks by +2; `722`'s anchors move `1008`/`1033` → `1010`/`1035`, which lands exactly on the `PHY_ID_AQR112` and `PHY_ID_AQR412` context lines in `734`'s post-image; `725` moves `-1255`/`-1276` → `-1257`/`-1278` and its first hunk grows `+30` → `+34` for the four added op lines, with the second hunk's new-side picking up the matching +6. `723` correctly needs no refresh — its only anchor is at line 262, above everything `734` inserts.
- `734` is the last patch in `backport-6.12/` touching `aquantia_main.c`, and nothing in `pending-6.12/` touches it, so `722`/`723`/`725` are the complete set of downstream patches needing a refresh.
- The AQR112C/AQR112R entries place the two new ops last, matching both the twelve entries `734` patches and the `hack-6.18/725` sibling.
- CI on `29a45fc`: no failures — everything completed is `success` or `skipped`, two package builds still in progress.

One nit inline, on the backport note.

---

openwrt-ai (commented):

The delta since `29a45fc` is confined to the backport note in `734-v6.18-net-phy-aquantia-report-and-configure-in-band-autone.patch` (6 lines replacing 5, all above the first `--- a/`), so no hunk arithmetic changes and the patched result is byte-identical. The first six commits are still #24420 unchanged, and the last commit's message is unchanged from the previous revision.

The rewritten note is accurate: [upstream `aqr_gen2_config_inband()`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/aquantia/aquantia_main.c#L1128-L1141) does handle `USXGMII` and `10G_QXGMII` together in the `MDIO_PHYXS_VEND_PROV2` / 4.C441 branch, ahead of and without touching the `priv->global_cfg[]` loop, so "the two modes upstream handles in the same 4.C441 branch, neither of which touches the shadow" holds verbatim. That closes the one nit from the previous pass — no "USXGMII alone" wording remains anywhere in `target/linux/generic/`, and the PR description was updated to match.

CI on `5f40377`: no failures. All 93 check runs are `success` or `skipped`, apart from one `cancelled` job belonging to a superseded `build-pr-profile` run trigger.

---

robimarko:
I moved qualcommax to 6.18 and dropped 6.12 so we can avoid the mess of backporting.

PR #24516, opened 2026-08-01
zlib: build the NEON inflate path from the stock sources

JuliusBairaktaris:
`contrib/arm/{inflate,inffast}.c` are a 2017 fork of zlib 1.2.11 that patch 004 swaps in wholesale when the `ARMv8` cmake option is on. A copy cannot conflict, so nothing in the build ever said it had gone stale, and it sat six releases behind the tarball built around it. It keeps the pre-1.2.13 `EXTRA` state of CVE-2022-37434, without the `< state->head->extra_max` clause: a gzip stream whose extra field is longer than the caller's `extra_max` underflows `extra_max - len` and `zmemcpy` writes past that buffer, reachable from `inflateGetHeader()`. It also misses the `inflateResetKeep()`, `inflatePrime()` and `inflateReset2()` fixes.

    contrib/arm/inflate.c (added by patch 002), inflate(), case EXTRA:

                    if (state->head != Z_NULL &&
                        state->head->extra != Z_NULL) {
                        len = state->head->extra_len - state->length;

The swap is per file, so `infback.c` keeps the stock sources while calling the chunked `inflate_fast()`. `inflateBackInit_()` takes its window from the caller at exactly `1 << windowBits`, but `chunkcopy_core()` issues an unconditional 16-byte `loadchunk()` however short the copy is, so a match landing near the window end reads past the caller's buffer. No patch to the vendored files can reach that.

    contrib/arm/chunkcopy.h (added by patch 003), chunkcopy_core():

        int bump = (--len % CHUNKCOPY_CHUNK_SIZE) + 1;
        storechunk(out, loadchunk(from));

This drops both copies and moves the ARM code into the stock sources: nine `__ARM_NEON` blocks across `inflate.h`, `inflate.c`, `inffast.c` and `infback.c`, with `chunkcopy.h` included from `inflate.h` so the four of them share one guard. `CHUNKCOPY_CHUNK_SIZE` is 0 elsewhere, so the window padding folds away and `inffast.c` and `infback.c` preprocess byte for byte to the stock sources on every non-ARM target.

The gate moves with them: `__ARM_NEON` is what the compiler defines when Advanced SIMD is there, always on aarch64 and on AArch32 exactly when `-mfpu` selects it, so the cmake option and the probe feeding it go, and with them the reason 3f5e39e96092 ("zlib: only enable NEON optimizations on eligible targets") had to exclude the NEON-less AArch32 case by hand. That probe tested `__ARM_NEON__`, the AArch32 spelling gcc's aarch64 backend does not define, so it answered `OFF` on every 64-bit ARM target and the optimisation has never been built there at all. On an IPQ8074, best of 40 iterations over a 4 MiB buffer, that is 16.17 -> 10.16 ms on a default `-O2` build and 11.50 -> 10.43 ms with `ZLIB_OPTIMIZE_SPEED` and LTO.

`infback.c` keeps its window unpadded and skips the fast path instead, which is not free. On AArch32 that removes the overread above, which is live today: flipping the guard back in an ASan build reproduces a 16-byte heap-buffer-overflow read in `loadchunk()` through `inflateBack()`. On aarch64 there was nothing to fix, because the removed probe answered `OFF` there. I have only run this on aarch64; the AArch32 arm is built but not exercised on hardware.

I used AI to help me write the change. It applies to a pristine 1.3.2 tarball at zero fuzz and zero offset, and on an IPQ8074 `inflate()` and `inflateBack()` round-trip at five output chunk sizes and three compression levels with CRCs matching stock. Is skipping the fast path in `inflateBack()` on ARM ok, or is an API change so the caller passes a padded window worth it?


package/libs/zlib/Makefile
```diff
@@ -60,7 +60,7 @@ ifeq ($(CONFIG_ZLIB_OPTIMIZE_SPEED),y)
 endif
 
 CMAKE_OPTIONS += \
-	-DARMv8=$$$$(echo -e '\#ifdef __ARM_NEON__\nON\n\#else\nOFF\n\#endif' | $$(TARGET_CC) $$(TARGET_CFLAGS) -x c -E - | grep -xE 'ON|OFF')
+	-DARMv8=$$$$(echo -e '\#ifdef __ARM_NEON\nON\n\#else\nOFF\n\#endif' | $$(TARGET_CC) $$(TARGET_CFLAGS) -x c -E - | grep -xE 'ON|OFF')
```
openwrt-ai: The macro analysis looks right (gcc's aarch64 backend defines only `__ARM_NEON`), but flipping this probe changes which *sources* get compiled: `ARMv8=ON` makes [004-attach-sourcefiles-in-patch-002-to-buildsystem.patch](``https://github.com/openwrt/openwrt/blob/604bb9d47100a0f6a4242cd191b9b3a8a1f104a0/package/libs/zlib/patches/004-attach-sourcefiles-in-patch-002-to-buildsystem.patch#L17-L47)`` swap the stock `inflate.c`/`inffast.c` for the vendored `contrib/arm/` copies. Those copies are a 2017 ARM fork that has not been kept in sync with zlib 1.3.2, so this PR extends two pre-existing defects from "armv7 NEON subtargets only" to essentially every aarch64 device. Worth deciding consciously before merging:

1. **CVE-2022-37434 is unfixed in the vendored `inflate.c`.** [Lines 1103-1109 of 002-arm-specific-optimisations-for-inflate.patch](``https://github.com/openwrt/openwrt/blob/604bb9d47100a0f6a4242cd191b9b3a8a1f104a0/package/libs/zlib/patches/002-arm-specific-optimisations-for-inflate.patch#L1103-L1109)`` carry the pre-fix `case EXTRA:` body:

   ```c
   if (state->head != Z_NULL &&
       state->head->extra != Z_NULL) {
       len = state->head->extra_len - state->length;
       zmemcpy(state->head->extra + len, next, ...);
   ```

   Upstream's fix (`eff308af`, shipped in 1.2.13 and present in the 1.3.2 tarball this package builds) additionally requires `(len = state->head->extra_len - state->length) < state->head->extra_max`. Without it, a gzip stream whose extra field is longer than the caller's `extra_max` makes `state->head->extra_max - len` underflow, so `zmemcpy` gets a huge length. Any in-tree user of `inflateGetHeader()` on an aarch64 build would become reachable. The vendored copy should be rebased onto 1.3.2's `inflate.c` (or at least get this hunk) before the option is enabled more widely.

2. **`infback.c` is still built from stock but calls the chunked `inflate_fast()`.** The vendored `inflate.c` over-allocates the window by `CHUNKCOPY_CHUNK_SIZE` precisely because `chunkcopy_core()` issues a full 16-byte `loadchunk()` even when `len < 16`. `inflateBackInit_()` in the unpatched `infback.c` uses the caller-supplied window verbatim, with no such padding, so `from = window + wsize - op` with a small `op` reads up to 15 bytes past the caller's buffer. `infback.c` is in `ZLIB_SRCS` for both branches of patch 004, so the ARMv8 build mixes the padded-window assumption with an unpadded window.

Neither is a regression introduced by this diff, but both scale with it. If you'd rather keep the fix minimal, it would help to say so explicitly in the commit message so the exposure change is on record.

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the diff and the `__ARM_NEON__` -> `__ARM_NEON` reasoning checks out; `package/libs/zlib/Makefile` is the only place in the tree using this probe. One inline comment on the code that the flipped probe newly pulls into the build on aarch64.

---

package/libs/zlib/patches/002-arm-specific-optimisations-for-inflate.patch
```diff
@@ -339,7 +339,7 @@ Change-Id: Id4cda552b39bfb39ab35ec499dbe122b43b6d1a1
 +#endif /* !ASMINF */
 --- /dev/null
 +++ b/contrib/arm/inflate.c
-@@ -0,0 +1,1561 @@
+@@ -0,0 +1,1562 @@
```
openwrt-ai: nit: the file-count header was updated but the `git format-patch` preamble above it was not, so the patch metadata now contradicts its own body. [Lines 8-10](``https://github.com/openwrt/openwrt/blob/5ffddc6ba948ac0f474f6dcd610c20ce37ee21ae/package/libs/zlib/patches/002-arm-specific-optimisations-for-inflate.patch#L8-L10)`` still read:

```
 contrib/arm/inffast.c |  323 ++++++++++
 contrib/arm/inflate.c | 1561 +++++...
 2 files changed, 1884 insertions(+)
```

which should be 1562 / 1885 after this hunk. Harmless to `patch`, but it is the visible symptom of a bigger point: 002 and 003 are verbatim imports of Simon Hosie's two ARM commits (`From 6bac7a3e0ebc...` / `From 2471476...`), and this change makes 002 no longer be that commit while still claiming to be. Since the CVE fix is OpenWrt-authored and 003 already demonstrates that `contrib/arm/inflate.c` can be amended by a later patch, would a separate `005-fix-CVE-2022-37434-in-arm-inflate.patch` be preferable? That keeps 002/003 as pristine vendor imports, makes the security fix greppable by name, and avoids having to renumber 003's `@@ -1176,17 +1187,16 @@` hunk header. Your call — if you keep it inline, the preamble diffstat is worth correcting either way.

For what it's worth, I reconstructed `contrib/arm/inflate.c` from patch 002 as amended (1562 lines, matches the new header), and patch 003 applies to it with zero fuzz and zero offset, so the renumbering is correct as written. The resulting `case EXTRA:` body is byte-identical to zlib 1.3.2's.

---

openwrt-ai (commented):
Reviewed 1 new commit (`ea449ff` "zlib: fix CVE-2022-37434 in the vendored ARM inflate") plus the reworded head commit.

Both points from my previous review are addressed: the CVE-2022-37434 hunk is now taken, and the `infback.c` / unpadded-window issue is explicitly recorded as knowingly left in the head commit message, which is what I asked for. Verification I ran on the new commit:

- Reconstructed `contrib/arm/inflate.c` from the amended patch 002 — 1562 lines, matching the updated `@@ -0,0 +1,1562 @@`.
- Patch 003's three hunks against that file apply with no fuzz and no offset, so the `@@ -1176,17 +1187,16 @@` renumbering (old +1 from the inserted line at 762, new +1 on top of the existing +11 from the earlier hunks) is right.
- The resulting `case EXTRA:` body matches zlib 1.3.2's `inflate.c` exactly, and `len` is the same `unsigned len;` declared at the top of `inflate()`, so the assignment-in-condition is well-defined.
- Patch 002 creates files only (no context lines against the 1.3.2 tarball), so nothing else in the series depends on the edited region.

Both commit messages match their diffs. One inline nit on patch hygiene. CI on the head commit has not reported yet.

---

package/libs/zlib/patches/002-arm-specific-optimisations-for-inflate.patch
```diff
@@ -1818,6 +1681,7 @@ Change-Id: Id4cda552b39bfb39ab35ec499dbe122b43b6d1a1
 +    copy = (struct inflate_state FAR *)
 +           ZALLOC(source, 1, sizeof(struct inflate_state));
 +    if (copy == Z_NULL) return Z_MEM_ERROR;
++    zmemzero(copy, sizeof(struct inflate_state));
 +    window = Z_NULL;
 +    if (state->window != Z_NULL) {
 +        window = (unsigned char FAR *)
```
openwrt-ai: The `inflateCopy()` window allocation on the next line is the second unpadded-window site, and it is not in the head commit's "knowingly left" list — only `infback.c` is.

[Lines 1687-1688](``https://github.com/openwrt/openwrt/blob/54f59db0ea9f583b2475db14842c3d519b2fd4af/package/libs/zlib/patches/002-arm-specific-optimisations-for-inflate.patch#L1687-L1688)`` allocate the copy's window at exactly `1U << state->wbits`:

```c
window = (unsigned char FAR *)
         ZALLOC(source, 1U << state->wbits, sizeof(unsigned char));
```

whereas [patch 003 pads the `updatewindow()` allocation](``https://github.com/openwrt/openwrt/blob/54f59db0ea9f583b2475db14842c3d519b2fd4af/package/libs/zlib/patches/003-arm-specific-optimisations-for-inflate.patch#L458-L461)`` with `wsize + CHUNKCOPY_CHUNK_SIZE` precisely because [`chunkcopy_core_safe()`](``https://github.com/openwrt/openwrt/blob/54f59db0ea9f583b2475db14842c3d519b2fd4af/package/libs/zlib/patches/003-arm-specific-optimisations-for-inflate.patch#L100-L111)`` only bounds the *destination*: once `limit - out >= CHUNKCOPY_CHUNK_SIZE` it falls through to `chunkcopy_core()`, which issues an unconditional 16-byte `loadchunk(from)` regardless of `len`.

So after `inflateCopy()`, `dest`'s window is 16 bytes short of what the chunked reader assumes. Both the rebased `case MATCH:` (`from = state->window + (state->wsize - copy)` → `chunkcopy_safe(put, from, copy, put + left)`) and `inflate_fast()`'s window branch can then read up to 15 bytes past the end of that heap allocation whenever the match lands near the window end and `left >= 16`.

I confirmed by reconstruction that this is pre-existing, not introduced here — the base-branch vendored `inflate.c` has the same `ZALLOC(source, 1U << state->wbits, ...)`. But it is exactly the same class of defect as the `infback.c` item the head commit records, and this series widens its reach from the armv7 NEON subtargets to all of aarch64, so it seems worth either padding it here or naming it alongside `infback.c` in the "knowingly left" paragraph.

---

package/libs/zlib/patches/002-arm-specific-optimisations-for-inflate.patch
```diff
@@ -339,9 +339,9 @@ Change-Id: Id4cda552b39bfb39ab35ec499dbe122b43b6d1a1
 +#endif /* !ASMINF */
 --- /dev/null
 +++ b/contrib/arm/inflate.c
-@@ -0,0 +1,1561 @@
+@@ -0,0 +1,1413 @@
 +/* inflate.c -- zlib decompression
-+ * Copyright (C) 1995-2016 Mark Adler
++ * Copyright (C) 1995-2026 Mark Adler
```
openwrt-ai: nit: a `Copyright (C) 1995-2026` line inside a patch whose header is `From 6bac7a3e0ebcd3147294b73acb34606eba18ae7f` / `From: Simon Hosie <simon.hosie@arm.com>` / `Date: Wed, 12 Apr 2017` / `Subject: [PATCH 1/2] Prepare ARM-specific contrib directory.` ([lines 1-6](``https://github.com/openwrt/openwrt/blob/54f59db0ea9f583b2475db14842c3d519b2fd4af/package/libs/zlib/patches/002-arm-specific-optimisations-for-inflate.patch#L1-L6))`` is self-contradicting. After this rebase the patch is no longer that 2017 commit at all — it is a 1.3.2-derived file — but it still carries that commit's SHA, author, date and `Change-Id`, and 003 still says `[PATCH 2/2]` of the same series.

The diffstat is now internally consistent (323 + 1413 = 1736), so the earlier mismatch is fixed. What is left is provenance: someone bisecting the ARM fork later will resolve `6bac7a3e` upstream and get a different file. Adding a short OpenWrt-authored note under the `Subject:` (what was re-based onto what, and why) would make it self-describing without changing a line of the payload. Your call how to word it — the header is outside this diff so I can't suggest it inline.

---

package/libs/zlib/patches/002-arm-specific-optimisations-for-inflate.patch
```diff
@@ -6,8 +6,8 @@ Subject: [PATCH 1/2] Prepare ARM-specific contrib directory.
 Change-Id: Id4cda552b39bfb39ab35ec499dbe122b43b6d1a1
 ---
  contrib/arm/inffast.c |  323 ++++++++++
```
openwrt-ai: nit: the commit message says of this file "The same treatment is not available for `contrib/arm/inffast.c`, which is ARM code with no upstream counterpart." That is not quite what patch 002 contains — like `inflate.c`, the copy 002 creates here is a verbatim import of stock `inffast.c` (`Copyright (C) 1995-2017`); the ARM rewrite only arrives with 003. The counterpart exists, it is just heavily overwritten downstream.

Worth noting for the record that I diffed this 323-line copy against 1.3.2's own `inffast.c` and the only differences are cosmetic — K&R vs. prototype definition, `code here` by value vs. `code const *here`, and `(char *)` vs. `(z_const char *)` on the `strm->msg` casts. No functional or security divergence, so unlike `inflate.c` there is nothing stale to carry over. Rewording to something like "not worth the same treatment: 003 rewrites it, and its base differs from 1.3.2's only cosmetically" would say what you verified rather than understating it.

---

openwrt-ai (commented):
Reviewed 2 new commits (`5d73c48` "zlib: rebase the vendored ARM inflate onto 1.3.2" and the reworded head `54f59db`). The previous first commit was replaced wholesale, so I re-verified the series from scratch against a pristine tarball.

Both commit messages match their diffs, and the central claim of `5d73c48` holds exactly as written. What I ran:

- Fetched `zlib-1.3.2.tar.xz`; sha256 matches `PKG_HASH` in the Makefile.
- Replayed 002/003/004 onto it with `patch -F 0` — all hunks apply at zero fuzz and zero offset, including 003's re-anchored `@@ -257,10 +258,20 @@` and `@@ -1050,17 +1061,16 @@`.
- `diff -u inflate.c contrib/arm/inflate.c` on the result: **exactly three hunks, all of them 003's** — the `chunkcopy.h` include, the `wsize + CHUNKCOPY_CHUNK_SIZE` window padding, and the chunkcopy `case MATCH:` body. Nothing else. So the vendored copy is 1.3.2's file, CVE-2022-37434 fix and all.
- Diffstat is now self-consistent: 323 + 1413 = 1736, and `@@ -0,0 +1,1413 @@` matches the 1413-line reconstruction. The −148 net in the PR body checks out.
- 1.3.2 moved the fixed tables into `inftrees.c` as `inflate_fixed()` ("Use atomics to build inflate fixed tables once"); the rebased copy calls it and drops its own `fixedtables()`/`makefixed()`. `inftrees.c` is in `ZLIB_SRCS` on both branches of 004 and `inftrees.h` declares it, so the ARM branch links.
- `PKG_RELEASE` 1→2→3, one bump per commit.

CI on `54f59db` reports no checks yet (pending, zero contexts), so nothing to correlate.

Three inline comments: one substantive (a second unpadded-window site that isn't in the "knowingly left" list) and two nits on patch/commit-message accuracy.

---

package/libs/zlib/patches/003-arm-specific-optimisations-for-inflate.patch
```diff
@@ -494,3 +494,13 @@ Change-Id: I59854eb25d2b1e43561c8a2afaf9175bf10cf674
              if (state->length == 0) state->mode = LEN;
              break;
          case LIT:
+@@ -1343,7 +1353,8 @@ int ZEXPORT inflateCopy(z_streamp dest,
+     window = Z_NULL;
+     if (state->window != Z_NULL) {
+         window = (unsigned char FAR *)
+-                 ZALLOC(source, 1U << state->wbits, sizeof(unsigned char));
++                 ZALLOC(source, (1U << state->wbits) + CHUNKCOPY_CHUNK_SIZE,
++                        sizeof(unsigned char));
```
openwrt-ai: nit: two things in this patch's `git format-patch` preamble go stale with this hunk, and both are the same class of thing the preceding commit just fixed in 002.

**Provenance.** [Lines 1-4](``https://github.com/openwrt/openwrt/blob/674ceb57e6f73ded77c0cb838b54a96bc69b252d/package/libs/zlib/patches/003-arm-specific-optimisations-for-inflate.patch#L1-L4)`` still read `From 247147654fe5cd11cf15d8dff91440405ea57040` / `From: Simon Hosie <simon.hosie@arm.com>` / `Subject: [PATCH 2/2] Inflate using wider loads and stores`. This `inflateCopy()` padding is OpenWrt-authored and is not in `2471476`, so 003 is now in exactly the position `e4bb720` documents for 002 with its new note under the `Subject:` line. A one-paragraph note here saying that the hunk headers were re-anchored onto 1.3.2 and that the `inflateCopy()` hunk is an OpenWrt addition would keep the two patches self-describing in the same way.

**Diffstat.** [Lines 13-16](``https://github.com/openwrt/openwrt/blob/674ceb57e6f73ded77c0cb838b54a96bc69b252d/package/libs/zlib/patches/003-arm-specific-optimisations-for-inflate.patch#L13-L16)`` still say:

```
 contrib/arm/inflate.c   |  22 ++--
 3 files changed, 335 insertions(+), 62 deletions(-)
```

With this hunk the real figures are `25` for `contrib/arm/inflate.c` (18 insertions, 7 deletions) and `337 insertions(+), 63 deletions(-)` overall. I checked by unpacking a pristine 1.3.2 tarball twice, applying 002 to one and 002+003 to the other, and taking `git diff --numstat` between them — `279` for `chunkcopy.h` and `96` for `inffast.c` are unchanged and still correct. `patch` ignores the preamble, but you corrected 002's diffstat in the preceding commit, so leaving 003's is the inconsistency rather than the correction.

---

package/libs/zlib/patches/002-arm-specific-optimisations-for-inflate.patch
```diff
@@ -3,11 +3,22 @@ From: Simon Hosie <simon.hosie@arm.com>
...
+its body wholesale, and against 1.3.2's inffast.c it differs only in the
+K&R function definition, in holding the table entry by value rather than
+through a pointer, and in the casts on strm->msg.
+
 Change-Id: Id4cda552b39bfb39ab35ec499dbe122b43b6d1a1
 ---
  contrib/arm/inffast.c |  323 ++++++++++
- contrib/arm/inflate.c | 1561 +++++++++++++++++++++++++++++++++++++++++++++++++
- 2 files changed, 1884 insertions(+)
+ contrib/arm/inflate.c | 1413 +++++++++++++++++++++++++++++++++++++++++++++++++
```
openwrt-ai: nit: two numbers in the PR body no longer describe the finished series.

**"net −127 lines."** The sentence's subject is the vendored file ("The vendored file is then 1.3.2's, plus exactly the include, the `CHUNKCOPY_CHUNK_SIZE` window padding and the chunkcopy `MATCH` state ... — net −127 lines"), but this line is the one that carries that delta and it goes `1561` → `1413`, i.e. **−148**. −127 is the PR-wide diffstat (140 additions, 267 deletions, as GitHub reports it), which is a different quantity — it nets the vendored file's −148 against 003's `+14/−4` and the Makefile's `+2/−2`. Both are true statements about different things; as written the reader is invited to attribute −127 to the file.

**"three hunks, all of them 003's."** Correct for `e4bb720` in isolation — I re-verified it there. But the head commit adds the `inflateCopy()` padding to 003, so diffing the fully-patched tree against 1.3.2's `inflate.c` now yields **four** hunks: the `chunkcopy.h` include, the `updatewindow()` padding, the chunkcopy `MATCH` body, and `inflateCopy()`. The body's "Unpadded windows" section does mention the new one, so this is only the section-1 count reading as the final state.

Body text only — the commit messages and the patches themselves are accurate.

---

openwrt-ai (commented):
Reviewed 2 new commits (`e4bb720` "zlib: rebase the vendored ARM inflate onto 1.3.2" and the head `674ceb5`) — both are amendments of the previously reviewed `5d73c48`/`54f59db`, so I re-verified the series from scratch rather than diffing the amendments.

The substantive point from my last review is addressed: 003 now pads `inflateCopy()`'s window allocation, and the provenance note I asked for is in 002's header. What I ran:

- `zlib-1.3.2.tar.xz` sha256 matches `PKG_HASH`.
- 002/003/004 replay onto the pristine tarball with `patch -F 0` — every hunk at zero fuzz and zero offset, including the new `@@ -1343,7 +1353,8 @@`. The `+10` offset is right: `+1` from 003's `chunkcopy.h` include, `+10` net from the `updatewindow()` hunk, `−1` net from the `MATCH` hunk.
- After 002 alone, `contrib/arm/inflate.c` is **byte-identical** to 1.3.2's `inflate.c`, so the new header note is accurate as written.
- After 003, `diff -u inflate.c contrib/arm/inflate.c` is exactly four hunks, all of them 003's — the include, the `updatewindow()` padding, the chunkcopy `MATCH` body, and the new `inflateCopy()` padding. Nothing else diverges from 1.3.2.
- The padding is exactly sufficient, not merely plausible: `chunkcopy_core()` advances by `bump = (len-1) % 16 + 1` and then loops in 16-byte `loadchunk()`s, so its last read ends at most 15 bytes past `from + len`. With `from <= window + wsize - 1` and `len <= copy`, that is `window + wsize + 15`, which the `+ CHUNKCOPY_CHUNK_SIZE` allocation covers with one byte to spare. Same reasoning as the `updatewindow()` padding it is placed next to.
- 002's `contrib/arm/inffast.c` payload is untouched by this PR, and its differences from 1.3.2's `inffast.c` are exactly the three the new note names — K&R definition, `code here` by value vs. `code const *here`, and the `strm->msg` casts. Nothing functional.
- `PKG_RELEASE` 1→2→3, one bump per commit.

Both commit messages match their diffs. CI on `674ceb5` reports no checks yet (pending, zero contexts), so nothing to correlate.

Two inline nits, both about surrounding text rather than code: 003's format-patch preamble is now stale in the same two ways 002's was before `e4bb720` fixed it, and two figures in the PR body describe the pre-head-commit state.

---

package/libs/zlib/patches/003-arm-specific-optimisations-for-inflate.patch
```diff
@@ -8,12 +8,16 @@ means that so long as the target is capable, wide un-aligned loads and stores
 can be used to transfer several bytes at once. When the reference distance is
 too short simply unroll the data a little to increase the distance.
 
+The hunks against contrib/arm/inflate.c are re-anchored onto the zlib 1.3.2
+copy patch 002 now vendors, and the inflateCopy() window padding among them
+is an OpenWrt addition rather than part of 2471476.
```
openwrt-ai: `FormalityCheck / Git & Commits` is the one red check on `0619f6e` (everything else on the head commit is green — both external-toolchain builds, package checks, Code Patches, OpenWrt Makefiles):

```
Commit 0619f6e - zlib: enable the NEON inflate path on aarch64:
- Line 37 in commit body exceeds max width (116/100 chars)
```

The offending line is the one that announces this note:

```
002, its trailers name a 2017 ARM commit that does not carry it. The other is infback.c, which patch 004 keeps stock
```

The amendment spliced `and records the addition in its own header - like 002, its trailers name a 2017 ARM commit that does not carry it.` into the middle of the paragraph without re-wrapping the remainder, so the pre-existing `The other is infback.c, which patch 004 keeps stock` ended up appended to it. Every other line in both commit messages is ≤ 74 chars, so this is the only one over. Re-wrapping that one paragraph clears the check without changing a word:

```
inflateCopy() is one of them, allocating the copy's window at exactly
1U << wbits, so a match landing near the window end reads past the
allocation; 003 pads it here, next to the updatewindow() padding that
gives it its reason, and records the addition in its own header - like
002, its trailers name a 2017 ARM commit that does not carry it. The
other is infback.c, which patch 004 keeps stock while swapping in the
chunked inflate_fast(), so inflateBackInit_() uses the caller's window
verbatim - out of reach of a patch to the vendored files, and knowingly
left.
```

(No suggestion block since the text lives in the commit message, not in the diff.)

---

openwrt-ai (commented):
Reviewed 1 new commit — `0619f6e` is an amendment of the previously reviewed `674ceb5`; `e4bb720` is byte-identical to what I reviewed last time. The only content change is the nit from my last review: 003's `git format-patch` preamble now carries a provenance note and a corrected diffstat.

Both are right. I re-verified the whole series from scratch rather than trusting the amendment diff:

- `zlib-1.3.2.tar.xz` sha256 matches `PKG_HASH`.
- 002/003/004 replay onto the pristine tarball with `patch -F 0` — every hunk at zero fuzz and zero offset.
- After 002 alone, `contrib/arm/inflate.c` is byte-identical to 1.3.2's `inflate.c`.
- After 003, `diff -u inflate.c contrib/arm/inflate.c` is exactly four hunks, all of them 003's: `@@ -84,6 +84,7 @@` (the `chunkcopy.h` include), `@@ -257,10 +258,20 @@` (the `updatewindow()` padding), `@@ -1050,17 +1061,16 @@` (the chunkcopy `MATCH` body) and `@@ -1343,7 +1353,8 @@` (the `inflateCopy()` padding). Nothing else diverges from 1.3.2.
- The new diffstat is exact, not merely closer. `git diff --numstat` between a 002-only tree and a 002+003 tree gives `279/0` for `chunkcopy.h`, `40/56` for `inffast.c` and `18/7` for `inflate.c` — so `25` for `contrib/arm/inflate.c` and `337 insertions(+), 63 deletions(-)` overall, matching the amended preamble line for line. `96` and `279` are correctly left alone.
- The new provenance note is accurate as written: the `inflateCopy()` hunk really is the OpenWrt addition and not part of `2471476`, and the other three hunk headers really were re-anchored onto the 1.3.2 copy.
- `PKG_RELEASE` 1→2→3, one bump per commit.

The head commit message still matches its diff.

One inline comment: the head commit trips `FormalityCheck / Git & Commits` with a 116-char line in its body, introduced by this amendment. My earlier nit on the PR body's `net −127 lines` / `three hunks` figures is unchanged and still open ([thread](https://github.com/openwrt/openwrt/pull/24516#discussion_r3699544118)); note the PR-wide diffstat has since moved to 146/269, so `−127` is stale on its own terms now too.

---

JuliusBairaktaris:
@diizzyy @jow- @ynezz — no `PKG_MAINTAINER` on this package, so tagging the three of you who own the pieces this touches: the ARM patches (3acecba520), the NEON eligibility probe it deletes (3f5e39e96092) and the CVE-2022-37434 backport (7df6795d4c).

That last one is the argument for the rewrite in one line: `006-fix-CVE-2022-37434.patch` patched `a/inflate.c`, and the ARM builds compile `contrib/arm/inflate.c` instead — so the fix shipped in 2022 and never reached them. `git log -S extra_max` on the vendored patch shows nothing before this PR. Patching the stock sources in place instead of vendoring copies is what makes that class of miss impossible; off ARM `inffast.c` and `infback.c` preprocess byte-identical to stock, so no other target sees a changed hot loop.

package/libs/zlib/patches/002-arm-neon-inflate.patch
```diff
@@ -1,21 +1,226 @@
...
+                 /* fallthrough */
+ 
+         case LEN:
+-            /* use inflate_fast() if we have enough input and output */
++#ifndef __ARM_NEON
++            /* use inflate_fast() if we have enough input and output.
++               Skipped when it reads the window in CHUNKCOPY_CHUNK_SIZE units:
++               inflateBackInit_() takes the window from the caller at exactly
++               1 << windowBits, so there is nowhere to put the slack those
++               reads need.  The loop below handles the same input. */
```
openwrt-ai, line 212: This guard lands very differently on the two ARM families, and only one of them is described here.

On AArch32-NEON — the only place `ARMv8` is `ON` today — it fixes a live bug, not a theoretical one. I reproduced it by flipping this `#ifndef` back to `#if 1` in an ASan build of the patched tree, with the `inflateBackInit_()` window as an exact-size heap block:

```
ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 16
    #0 loadchunk          contrib/arm/chunkcopy.h:27
    #1 chunkcopy_core     contrib/arm/chunkcopy.h:57
    #4 inflate_fast       inffast.c:245
    #5 inflateBack        infback.c:433
```

On aarch64 there is no bug to fix. The deleted probe answered `OFF` there ([Makefile:62-63](https://github.com/openwrt/openwrt/blob/3d1645ee26d6a2e20be71d7fa1716721bac78e53/package/libs/zlib/Makefile#L62-L63)), so `infback.c` is currently compiled against the *stock* `inflate_fast()` ([004 keeps it in `ZLIB_SRCS` on both branches](``https://github.com/openwrt/openwrt/blob/3d1645ee26d6a2e20be71d7fa1716721bac78e53/package/libs/zlib/patches/004-attach-sourcefiles-in-patch-002-to-buildsystem.patch#L55))`` and uses it. After this commit `__ARM_NEON` is defined on every aarch64 target, so those builds hand `inflateBack()` the per-symbol loop below instead — the routine zlib's own header comment credits with ">95% of the inflate execution time". And `libz.so` exports `inflateBack`, so "nothing in the tree calls `inflateBack()`" bounds the in-tree callers but not packages linking the shipped library.

I'm not arguing the tradeoff is wrong — the caller owns that window and doubles it as the output ring, so there is genuinely nowhere to put the slack without an API change. But the commit message presents the skip as neutral ("its loop below decodes the same input"), and on aarch64 it is a measurable loss sitting alongside the gain in `inflate()`. Would you rather say that explicitly in the commit message, or bound the reads instead — carry the window end in `inflate_state` and let the window branch of `inflate_fast()` use `chunkcopy_core_safe()` against it — so `inflateBack()` keeps the fast path on both families?

---

package/libs/zlib/patches/002-arm-neon-inflate.patch
```diff
@@ -1,21 +1,226 @@
...
 ---
- contrib/arm/chunkcopy.h | 279 ++++++++++++++++++++++++++++++++++++++++++++++++
- contrib/arm/inffast.c   |  96 +++++++----------
- contrib/arm/inflate.c   |  22 ++--
- 3 files changed, 335 insertions(+), 62 deletions(-)
- create mode 100644 contrib/arm/chunkcopy.h
-
+--- a/inflate.h
++++ b/inflate.h
+@@ -16,6 +16,16 @@
```
openwrt-ai, line 14: nit: "nine hunks across `inflate.c`, `inffast.c` and `infback.c`, each behind `__ARM_NEON`" (commit message and PR body) doesn't match the diff on either half.

Nine is the number of `__ARM_NEON` blocks, but **this** hunk is one of them — `grep -c '^+#if\(n\)\?def __ARM_NEON'` gives 9 split as `inflate.h` 1, `inflate.c` 2, `inffast.c` 5, `infback.c` 1. The three files actually named carry **ten** hunks (3 + 5 + 2), and two of the three `inflate.c` hunks carry no guard at all — the `updatewindow()` and `inflateCopy()` paddings ride on `CHUNKCOPY_CHUNK_SIZE` folding to 0, which the following sentence explains but the "each behind `__ARM_NEON`" claim contradicts.

"Nine `__ARM_NEON` blocks across `inflate.h`, `inflate.c`, `inffast.c` and `infback.c`" would be accurate for both counts.

---

package/libs/zlib/patches/002-arm-neon-inflate.patch
```diff
@@ -1,21 +1,226 @@
...
+                 LOAD();
+                 break;
+             }
++#endif
+ 
+             /* get a literal, length, or end-of-block code */
+             for (;;) {
 --- /dev/null
 +++ b/contrib/arm/chunkcopy.h
 @@ -0,0 +1,279 @@
```
openwrt-ai, line 226: nit: the PR body's "Net −2091 lines against the previous revision of this PR" doesn't reproduce from any measurement I can make.

Against `0619f6e`, restricted to `package/libs/zlib`, `git diff --shortstat` gives `221 insertions(+), 2088 deletions(-)`, and the directory goes 2478 → 611 lines — **net −1867**. Against the base branch it is `221 / 2211`, 2601 → 611 — **net −1990**, which is what GitHub's own diffstat on this PR shows. Neither rounds to 2091.

The other half of the sentence is right: this hunk's 279 lines really are 279 of the 494 remaining, where 494 is the diff payload of the file (505 lines total, the first 11 being the `git format-patch` preamble down to the `---`). Worth saying "494 payload lines" if you keep the figure, since the file itself is 505.

---

openwrt-ai (commented):
Reviewed 1 new commit (`7a9dc7f`). The PR was rewritten wholesale since my last review — `0619f6e`'s two commits are gone and the approach changed from "rebase the vendored fork onto 1.3.2" to "patch the stock sources in place" — so I re-verified everything from scratch against a pristine tarball rather than diffing the amendment. All eight of my earlier threads are moot and already resolved.

The commit message matches its diff, and the load-bearing claims hold as written. What I ran:

- `zlib-1.3.2.tar.xz` sha256 matches `PKG_HASH`. `002-arm-neon-inflate.patch` applies to it with `patch -F 0` at zero fuzz and zero offset.
- **Off ARM, `inffast.c` and `infback.c` preprocess byte-identical to stock** — `diff <(gcc -E -P stock) <(gcc -E -P patched)` is empty for both. `inflate.c` differs only by `(1U << state->wbits) + 0` at the two `ZALLOC`s and by the `MATCH` tail sitting in both branches instead of after them. Exactly the three differences the message names, and nothing else.
- The `__ARM_NEON` branch of `inffast.c` is logic-identical to the vendored `contrib/arm/inffast.c` this PR deletes, including the `wnext == 0 && whave >= wsize` remap (pre-existing in base's 003, not new here). `contrib/arm/chunkcopy.h` is byte-identical to base's. The `MATCH` limit changes from `put + left` to `put + copy + left` computed after `left -= copy` — same value.
- Macro claim verified rather than recalled: clang defines `__ARM_NEON` only for `aarch64`/`aarch64_be`; both spellings for `armv7-a -mfpu=neon`; neither for `armv7-a -mfpu=vfpv3-d16`. So the deleted probe really did answer `OFF` on every 64-bit ARM target, and the new gate covers exactly the AArch32 cases the old one did. `ixp4xx` is the tree's only big-endian ARM target and is ARMv5TE, so no BE-NEON build exists to worry about.
- `inflate.c`, `inffast.c`, `infback.c` and `inftrees.c` compile clean at `-Wall` for `aarch64`, `armv7-a -mfpu=neon` and `armv5te`. At `-Wextra` the only warnings are the two pre-existing `-Wsign-compare` in `chunkcopy.h`, as the PR body says.
- Nothing else in the tree referenced the `ARMv8` option, and `tools/zlib` has its own unrelated patch set.

No aarch64 hardware or emulator here, so to exercise the `__ARM_NEON` path I built it on x86 against a portable stand-in for the ten NEON intrinsics `chunkcopy.h` uses, under ASan+UBSan, with every `inflate()` output slot as an exact-size heap block so a redzone sits immediately after it:

- **36,864 `inflate()` round-trips** — 4 data shapes, levels 1/5/9, output chunks 1…4096, input chunks 1…1 MiB, `windowBits` ±15/±9 — all clean, and byte-identical results to a stock 1.3.2 build of the same harness.
- **4,320 `inflateBack()` / `inflateCopy()` cases** — clean. The only failures are a `windowBits=8` artefact of my harness, and the stock build reproduces that failure set identically.
- **Negative controls, to show those runs actually reach the new code.** Dropping the `updatewindow()` padding produces `READ of size 16` in `chunkcopy_core` from the new `MATCH` `chunkcopy_safe()` at `inflate.c:1057`; dropping the `inflateCopy()` padding produces the same after the copy; re-enabling `inflate_fast()` in `infback.c` on the ARM path produces a `heap-buffer-overflow` read 16 bytes past the caller-supplied window, reached from `inflateBack()`. Each of the two paddings and the `infback.c` guard is therefore necessary, and the patch as shipped is clean at all three.

That last control also settles the `infback.c` premise as a live bug on today's AArch32-NEON builds rather than a paper one.

CI on `7a9dc7f` reports no checks yet (pending, zero contexts), so nothing to correlate.

One inline question about the cost of the `infback.c` guard on aarch64, and two nits on figures in the commit message and PR body.

---

package/libs/zlib/patches/002-arm-neon-inflate.patch
openwrt-ai, line 14: nit: this amendment corrected both of the statements below in the commit message but left them standing in the PR body, so the two now disagree with each other.

**Hunk count / file list.** The body still reads "Nine hunks across `inflate.c`, `inffast.c` and `infback.c`, each behind `__ARM_NEON`, with `chunkcopy.h` included from `inflate.h` so all three share one guard." The amended message says "nine `__ARM_NEON` blocks across `inflate.h`, `inflate.c`, `inffast.c` and `infback.c` ... so the four of them share one guard", which is what the diff shows: the nine guards split 1 / 2 / 5 / 1 across four files — this hunk being the `inflate.h` one — and the two `inflate.c` padding hunks (`@@ -258,7 +258,7 @@` and `@@ -1343,7 +1360,8 @@`) carry no guard at all, riding on `CHUNKCOPY_CHUNK_SIZE` folding to 0.

**`infback.c`.** The body still frames the skip as neutral — "`infback.c` keeps its window unpadded and skips the fast path instead — the loop below it decodes the same input, and nothing in the tree calls `inflateBack()`." The message no longer does: "which is not free and is not the same trade on the two ARM families", "on aarch64 ... it now decodes on the per-symbol loop instead, which is a measurable loss", and "libz exports it, so that bounds the in-tree callers and not the packages linking the shipped library." That is the version I'd want in the body too, since it is what a maintainer reads before the commit message.

Body text only — the patch and the commit message are accurate as they stand.

---

openwrt-ai (commented):
Reviewed 1 new commit (`6ca29b9`), an amendment of the previously reviewed `7a9dc7f`. `git diff 7a9dc7f 6ca29b9` is **empty** — the tree is byte-identical, and the only change is the commit message. So rather than re-running the tarball replay, the preprocessor comparison and the ASan harness from last time, I checked the new text against the diff I had already verified.

Both threads I left open are addressed, and the new wording is accurate where it is checkable:

- **"nine `__ARM_NEON` blocks across `inflate.h`, `inflate.c`, `inffast.c` and `infback.c` ... so the four of them share one guard"** now matches the diff on both halves. Counting `+#if(n)def __ARM_NEON` per target file gives `inflate.h` 1, `inflate.c` 2, `inffast.c` 5, `infback.c` 1 = 9, and the three `.c` files all reach `chunkcopy.h` through `inflate.h`'s single guard. The earlier "each behind `__ARM_NEON`" claim — which the two unguarded `inflate.c` padding hunks contradicted — is gone.
- **The new `infback.c` paragraph** says the thing I asked for and says it as a trade rather than a wash: the AArch32 side removes a live overread (matching the `heap-buffer-overflow` read in `loadchunk()` through `inflateBack()` that my negative control reproduced last review), the aarch64 side had nothing to fix because the deleted probe answered `OFF` there and is now a measurable loss, and `libz` exporting `inflateBack` bounds only the in-tree callers. `git grep inflateBack` outside `package/libs/zlib` and `tools/zlib` is still empty, so that premise holds.
- Line widths: the longest body line is 74 chars, and `FormalityCheck / Git & Commits` is green on the head commit — as are all 14 check runs, including both external-toolchain builds and the package checks. Nothing to correlate.

One inline nit: the PR body still carries the older form of both statements the message just corrected.

---

package/libs/zlib/patches/002-arm-neon-inflate.patch
JuliusBairaktaris: > Body text only - the patch and the commit message are accurate as they stand.

The patch is; the commit message was not. Two of its slips are the body's as well: the
vendored copy sat six releases behind the tarball, not four, and it was missing the
`inflateResetKeep()`, `inflatePrime()` and `inflateReset2()` fixes rather than carrying
them. Both are corrected in this round.

On the third the two went opposite ways. The message said the aarch64 `infback.c` path
"now decodes on the per-symbol loop instead, which is a measurable loss", which I have
not measured; the body understated it, saying only that the loop below decodes the same
input and that nothing in the tree calls `inflateBack()`.

The body now says what the message says on both counts: the nine `__ARM_NEON` blocks
across four files, and `infback.c` skipping the fast path, with the aarch64 case named
and no loss claimed that I have not measured.

package/libs/zlib/patches/002-arm-neon-inflate.patch
openwrt-ai, line 212: nit: on the aarch64 side of this guard the PR body still stops one clause short of the commit message. The message says aarch64 `inflateBack()` runs the stock `inflate_fast()` today (the deleted probe answered `OFF`) and "now decodes on the per-symbol loop instead of any fast path"; the body says only "there was nothing to fix, because the removed probe answered `OFF` there", which reads as no aarch64 effect at all. That clause is the only place a maintainer sees the aarch64 half of the trade before opening the commit, so it is worth carrying over.

---

openwrt-ai (commented):
Reviewed 1 new commit; the tree is unchanged since `6ca29b9`, and both corrected statements check out: `v1.2.12`, `v1.2.13`, `v1.3`, `v1.3.1`, `v1.3.1.2`, `v1.3.2` are six release tags after `v1.2.11`, and the vendored `inflate.c` lacks `strm->data_type = 0;`/`state->flags = -1;` in `inflateResetKeep()`, the `windowBits < -15` guard in `inflateReset2()` and the `bits == 0` early return plus `(unsigned long)` cast in `inflatePrime()`. All 14 check runs on the head commit are green.

---

PR #24518, opened 2026-08-01
mediatek: add support Teralink TL3020

fildunsky:
This commit adds support for Teralink TL3020 256mb wireless router.

Specification
-------------
- SoC       : MediaTek MT7981B dual-core ARM Cortex-A53
- RAM       : DDR3 256 MiB (ESMT M15T2G16128A, DDR3-1866)
- Flash     : SPI-NAND 256 MiB (Winbond W25M02GV)
- WLAN      : MediaTek MT7976 2.4/5 GHz Wi-Fi 6
- Ethernet  : MediaTek MT7531 switch, 1x WAN + 3x LAN 10/100/1000 Mbps
              (2.5GbE uplink via gmac0)
- USB       : 1x USB 2.0 Type-A (power enable on GPIO 14)
- Buttons   : WPS, Reset
- LEDs      : 7x blue, 1x red
- UART      : 4-pin header on PCB
  - assignment  : 3.3V, GND, TX, RX
  - settings    : 115200n8
- Power     : 12 VDC, 1.5 A (barrel jack)

MAC addresses
-------------
EEPROM MAC is same for all the devices, but there is a `config2` partition
with unique MAC addresses:

| Interface | MAC source                                 |
|-----------|--------------------------------------------|
| LAN       | config2, offset 0x707 (ASCII, "lanmac")    |
| WAN       | config2, offset 0x787 (ASCII, "wanmac")    |
| WLAN 2.4G | config2, offset 0x68a (ASCII, "wifi2gmac") |
| WLAN 5G   | config2, offset 0x60a (ASCII, "wifi5gmac") |

The LAN MAC (gmac0) is the one printed on the device label.

Installation
------------
Factory bootloader is locked to 64mb partition layout, so OpenWrt U-Boot
has to be flashed first. Steps 1, 2 and 4 are run from a shell on the
stock firmware, where the `bl2` and `fip` partitions are writable - in
OpenWrt they are read-only, so the procedure cannot be repeated from
OpenWrt. Back up both partitions before overwriting them:
```
    mtd read bl2 /tmp/bl2.bin
    mtd read fip /tmp/fip.bin
```
1. Upload OpenWrt `bl31-uboot.fip` and `preloader.bin` images to the /tmp
   dir of the router using scp protocol
2. Write fip and bl2 (replace bootloader):
```
    mtd write /tmp/openwrt-mediatek-filogic-teralink_tl3020-256mb-bl31-uboot.fip fip
    mtd write /tmp/openwrt-mediatek-filogic-teralink_tl3020-256mb-preloader.bin bl2
```
3. Place OpenWrt
   `openwrt-mediatek-filogic-teralink_tl3020-256mb-initramfs-recovery.itb`
   image on the tftp server (IP: 192.168.1.254)
4. Erase `ubi` partition and reboot the router:
```
    mtd erase ubi
    reboot
```
5. U-Boot automatically boots the OpenWrt recovery image from the tftp
   server to the RAM
6. Upload OpenWrt
   `openwrt-mediatek-filogic-teralink_tl3020-256mb-squashfs-sysupgrade.itb`
   image to the /tmp dir of the router (IP: 192.168.1.1) using scp
   protocol
7. Connect to the router using ssh and run:
```
    ubidetach -p /dev/mtd5; ubiformat /dev/mtd5 -y; ubiattach -p /dev/mtd5
    ubimkvol /dev/ubi0 -n 0 -N ubootenv -s 128KiB
    ubimkvol /dev/ubi0 -n 1 -N ubootenv2 -s 128KiB
    sysupgrade -n openwrt-mediatek-filogic-teralink_tl3020-256mb-squashfs-sysupgrade.itb
```

Recovery
--------
1. Place OpenWrt
   `openwrt-mediatek-filogic-teralink_tl3020-256mb-initramfs-recovery.itb`
   image on the tftp server (IP: 192.168.1.254)
2. Press "Reset" button and power on the router. After ~10 sec release
   the button.
3. Use OpenWrt initramfs system for recovery

Reverting to stock firmware requires writing the `bl2` and `fip` backups
taken above back with `mtd write`, then flashing the stock image; without
those backups there is no way back, because the stock bootloader cannot
be downloaded from the vendor.

### DIFF
--- a/package/boot/uboot-mediatek/Makefile
+++ b/package/boot/uboot-mediatek/Makefile
@@ -749,6 +749,18 @@ define U-Boot/mt7981_snr_snr-cpe-ax2
   DEPENDS:=+trusted-firmware-a-mt7981-spim-nand-ddr3
 endef
 
+define U-Boot/mt7981_teralink_tl3020-256mb
+  NAME:=Teralink TL3020 256mb
+  BUILD_SUBTARGET:=filogic
+  BUILD_DEVICES:=teralink_tl3020-256mb
+  UBOOT_CONFIG:=mt7981_teralink_tl3020-256mb
+  UBOOT_IMAGE:=u-boot.fip
+  BL2_BOOTDEV:=spim-nand
+  BL2_SOC:=mt7981
+  BL2_DDRTYPE:=ddr3-1866
+  DEPENDS:=+trusted-firmware-a-mt7981-spim-nand-ddr3-1866
+endef
+
 define U-Boot/mt7981_wavlink_wl-wnt100x3
   NAME:=WAVLINK WL-WNT100X3
   BUILD_SUBTARGET:=filogic
@@ -1355,6 +1367,7 @@ UBOOT_TARGETS := \
 	mt7981_qihoo_360t7 \
 	mt7981_qihoo_360t7-ubi \
 	mt7981_snr_snr-cpe-ax2 \
+	mt7981_teralink_tl3020-256mb \
 	mt7981_wavlink_wl-wnt100x3 \
 	mt7981_xiaomi_mi-router-ax3000t \
 	mt7981_xiaomi_mi-router-wr30u \
--- a/package/boot/uboot-mediatek/patches/475-add-teralink_tl3020-256mb.patch
+++ b/package/boot/uboot-mediatek/patches/475-add-teralink_tl3020-256mb.patch
@@ -1,0 +1,371 @@
+From 6a8b2843894f123a11b800ed985b56d4ef771ddb Mon Sep 17 00:00:00 2001
+From: Fil Dunsky <filipp.dunsky@gmail.com>
+Date: Fri, 12 Jun 2026 21:15:27 +0300
+Subject: [PATCH] board: mediatek: add Teralink TL3020 256mb
+
+Add U-Boot support for the Teralink TL3020 256mb (256 MiB SPI-NAND):
+defconfig, device tree and default environment.
+
+Signed-off-by: Fil Dunsky <filipp.dunsky@gmail.com>
+---
+--- /dev/null
++++ b/configs/mt7981_teralink_tl3020-256mb_defconfig
+@@ -0,0 +1,112 @@
++CONFIG_ARM=y
++CONFIG_ARM_SMCCC=y
++CONFIG_SYS_HAS_NONCACHED_MEMORY=y
++CONFIG_POSITION_INDEPENDENT=y
++CONFIG_ARCH_MEDIATEK=y
++CONFIG_TEXT_BASE=0x41e00000
++CONFIG_SYS_MALLOC_F_LEN=0x4000
++CONFIG_NR_DRAM_BANKS=1
++CONFIG_DEFAULT_DEVICE_TREE="mt7981-teralink-tl3020-256mb"
++CONFIG_OF_LIBFDT_OVERLAY=y
++CONFIG_TARGET_MT7981=y
++CONFIG_SYS_LOAD_ADDR=0x46000000
++CONFIG_PRE_CON_BUF_ADDR=0x4007ef00
++CONFIG_DEBUG_UART_BASE=0x11002000
++CONFIG_DEBUG_UART_CLOCK=40000000
++CONFIG_DEBUG_UART=y
++CONFIG_FIT=y
++CONFIG_BOOTDELAY=30
++CONFIG_AUTOBOOT_KEYED=y
++CONFIG_AUTOBOOT_MENU_SHOW=y
++CONFIG_DEFAULT_FDT_FILE="mediatek/mt7981-teralink-tl3020-256mb.dtb"
++CONFIG_LOGLEVEL=7
++CONFIG_PRE_CONSOLE_BUFFER=y
++CONFIG_LOG=y
++# CONFIG_BOARD_INIT is not set
++CONFIG_BOARD_LATE_INIT=y
++CONFIG_HUSH_PARSER=y
++CONFIG_SYS_PROMPT="MT7981> "
++CONFIG_CMD_CPU=y
++CONFIG_CMD_LICENSE=y
++CONFIG_CMD_BOOTMENU=y
++CONFIG_CMD_ASKENV=y
++CONFIG_CMD_ERASEENV=y
++CONFIG_CMD_ENV_FLAGS=y
++CONFIG_CMD_RNG=y
++CONFIG_CMD_STRINGS=y
++CONFIG_CMD_DM=y
++CONFIG_CMD_GPIO=y
++CONFIG_CMD_GPT=y
++CONFIG_CMD_MTD=y
++CONFIG_CMD_PART=y
++CONFIG_CMD_TFTPSRV=y
++CONFIG_CMD_RARP=y
++CONFIG_CMD_CDP=y
++CONFIG_CMD_SNTP=y
++CONFIG_CMD_LINK_LOCAL=y
++CONFIG_CMD_DHCP=y
++CONFIG_CMD_DNS=y
++CONFIG_CMD_PING=y
++CONFIG_CMD_PXE=y
++CONFIG_CMD_CACHE=y
++CONFIG_CMD_PSTORE=y
++CONFIG_CMD_PSTORE_MEM_ADDR=0x42ff0000
++CONFIG_CMD_UUID=y
++CONFIG_CMD_HASH=y
++CONFIG_CMD_SMC=y
++CONFIG_CMD_UBI=y
++CONFIG_CMD_UBI_RENAME=y
++CONFIG_OF_EMBED=y
++CONFIG_ENV_OVERWRITE=y
++CONFIG_ENV_IS_IN_UBI=y
++CONFIG_ENV_REDUNDANT=y
++CONFIG_ENV_UBI_PART="ubi"
++CONFIG_ENV_UBI_VOLUME="ubootenv"
++CONFIG_ENV_UBI_VOLUME_REDUND="ubootenv2"
++CONFIG_ENV_RELOC_GD_ENV_ADDR=y
++CONFIG_ENV_USE_DEFAULT_ENV_TEXT_FILE=y
++CONFIG_ENV_DEFAULT_ENV_TEXT_FILE="defenvs/teralink_tl3020-256mb_env"
++CONFIG_ENV_VARS_UBOOT_RUNTIME_CONFIG=y
++CONFIG_VERSION_VARIABLE=y
++CONFIG_NETCONSOLE=y
++CONFIG_USE_IPADDR=y
++CONFIG_IPADDR="192.168.1.1"
++CONFIG_USE_SERVERIP=y
++CONFIG_SERVERIP="192.168.1.254"
++CONFIG_NET_RANDOM_ETHADDR=y
++CONFIG_REGMAP=y
++CONFIG_SYSCON=y
++CONFIG_BUTTON=y
++CONFIG_BUTTON_GPIO=y
++CONFIG_CLK=y
++CONFIG_GPIO_HOG=y
++CONFIG_LED=y
++CONFIG_LED_BLINK=y
++CONFIG_LED_GPIO=y
++# CONFIG_MMC is not set
++CONFIG_MTD=y
++CONFIG_DM_MTD=y
++CONFIG_MTD_SPI_NAND=y
++CONFIG_MTD_UBI_FASTMAP=y
++CONFIG_PHY_FIXED=y
++CONFIG_MEDIATEK_ETH=y
++CONFIG_PHY=y
++CONFIG_PINCTRL=y
++CONFIG_PINCONF=y
++CONFIG_PINCTRL_MT7981=y
++CONFIG_POWER_DOMAIN=y
++CONFIG_MTK_POWER_DOMAIN=y
++CONFIG_DM_REGULATOR=y
++CONFIG_DM_REGULATOR_FIXED=y
++CONFIG_DM_REGULATOR_GPIO=y
++CONFIG_RAM=y
++CONFIG_DM_RNG=y
++CONFIG_RNG_MTK_V2=y
++CONFIG_DM_SERIAL=y
++CONFIG_SERIAL_RX_BUFFER=y
++CONFIG_MTK_SERIAL=y
++CONFIG_SPI=y
++CONFIG_DM_SPI=y
++CONFIG_MTK_SPIM=y
++CONFIG_ZSTD=y
++CONFIG_HEXDUMP=y
+--- /dev/null
++++ b/arch/arm/dts/mt7981-teralink-tl3020-256mb.dts
+@@ -0,0 +1,186 @@
++// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
++/*
++ * Copyright (C) 2026 Fil Dunsky <filipp.dunsky@gmail.com>
++ */
++
++/dts-v1/;
++#include "mt7981.dtsi"
++#include <dt-bindings/gpio/gpio.h>
++#include <dt-bindings/input/linux-event-codes.h>
++
++/ {
++	#address-cells = <1>;
++	#size-cells = <1>;
++	model = "Teralink TL3020 256mb";
++	compatible = "teralink,tl3020-256mb", "mediatek,mt7981";
++
++	chosen {
++		stdout-path = &uart0;
++		tick-timer = &timer0;
++	};
++
++	memory@40000000 {
++		device_type = "memory";
++		reg = <0x40000000 0x10000000>;
++	};
++
++	gpio-keys {
++		compatible = "gpio-keys";
++
++		button-wps {
++			label = "wps";
++			linux,code = <KEY_WPS_BUTTON>;
++			gpios = <&pio 0 GPIO_ACTIVE_LOW>;
++		};
++
++		button-reset {
++			label = "reset";
++			linux,code = <KEY_RESTART>;
++			gpios = <&pio 1 GPIO_ACTIVE_LOW>;
++		};
++	};
++
++	gpio-leds {
++		compatible = "gpio-leds";
++
++		led-wan-red {
++			label = "red:wan";
++			gpios = <&pio 6 GPIO_ACTIVE_HIGH>;
++		};
++
++		led-power {
++			label = "blue:power";
++			gpios = <&pio 7 GPIO_ACTIVE_LOW>;
++		};
++
++		led-lan1 {
++			label = "blue:lan1";
++			gpios = <&pio 9 GPIO_ACTIVE_LOW>;
++		};
++
++		led-lan2 {
++			label = "blue:lan2";
++			gpios = <&pio 10 GPIO_ACTIVE_LOW>;
++		};
++
++		led-lan3 {
++			label = "blue:lan3";
++			gpios = <&pio 11 GPIO_ACTIVE_LOW>;
++		};
++
++		led-wan-blue {
++			label = "blue:wan";
++			gpios = <&pio 12 GPIO_ACTIVE_LOW>;
++		};
++
++		led-wlan2g {
++			label = "blue:wlan2g";
++			gpios = <&pio 34 GPIO_ACTIVE_LOW>;
++		};
++
++		led-wlan5g {
++			label = "blue:wlan5g";
++			gpios = <&pio 35 GPIO_ACTIVE_LOW>;
++		};
++	};
++};
++
++&eth {
++	status = "okay";
++	mediatek,gmac-id = <0>;
++	phy-mode = "2500base-x";
++	mediatek,switch = "mt7531";
++	reset-gpios = <&pio 39 GPIO_ACTIVE_HIGH>;
++
++	fixed-link {
++		speed = <2500>;
++		full-duplex;
++	};
++};
++
++&pio {
++	spi_flash_pins: spi0-pins-func-1 {
++		mux {
++			function = "flash";
++			groups = "spi0", "spi0_wp_hold";
++		};
++
++		conf-pu {
++			pins = "SPI0_CS", "SPI0_HOLD", "SPI0_WP";
++			drive-strength = <MTK_DRIVE_8mA>;
++			bias-pull-up = <MTK_PUPD_SET_R1R0_11>;
++		};
++
++		conf-pd {
++			pins = "SPI0_CLK", "SPI0_MOSI", "SPI0_MISO";
++			drive-strength = <MTK_DRIVE_8mA>;
++			bias-pull-down = <MTK_PUPD_SET_R1R0_11>;
++		};
++	};
++};
++
++&spi0 {
++	#address-cells = <1>;
++	#size-cells = <0>;
++	pinctrl-names = "default";
++	pinctrl-0 = <&spi_flash_pins>;
++	status = "okay";
++	must_tx;
++	enhance_timing;
++	dma_ext;
++	ipm_design;
++	support_quad;
++	tick_dly = <2>;
++	sample_sel = <0>;
++
++	spi_nand@0 {
++		compatible = "spi-nand";
++		reg = <0>;
++		spi-max-frequency = <52000000>;
++
++		partitions {
++			compatible = "fixed-partitions";
++			#address-cells = <1>;
++			#size-cells = <1>;
++
++			partition@0 {
++				label = "bl2";
++				reg = <0x0 0x100000>;
++			};
++
++			partition@100000 {
++				label = "u-boot-env";
++				reg = <0x100000 0x80000>;
++			};
++
++			partition@180000 {
++				label = "factory";
++				reg = <0x180000 0x200000>;
++			};
++
++			partition@380000 {
++				label = "fip";
++				reg = <0x380000 0x200000>;
++			};
++
++			partition@580000 {
++				label = "config2";
++				reg = <0x580000 0x100000>;
++			};
++
++			partition@680000 {
++				label = "ubi";
++				reg = <0x680000 0xf980000>;
++			};
++		};
++	};
++};
++
++&uart0 {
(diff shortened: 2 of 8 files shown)


target/linux/mediatek/dts/mt7981b-teralink-tl3020-256mb.dts
```diff
@@ -0,0 +1,369 @@
...
+			gpios = <&pio 34 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "phy0tpt";
+		};
+
+		led-wlan5g {
+			color = <LED_COLOR_ID_BLUE>;
+			function = LED_FUNCTION_WLAN_5GHZ;
+			gpios = <&pio 35 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "phy1tpt";
+		};
```
openwrt-ai, line 120: The `phy0tpt`/`phy1tpt` default triggers here are overridden by the netdev triggers this PR also adds for the very same two LEDs in `board.d/01_leds`:

```
ucidef_set_led_netdev "wlan2g" "WLAN2G" "blue:wlan-2ghz" "phy0-ap0"
ucidef_set_led_netdev "wlan5g" "WLAN5G" "blue:wlan-5ghz" "phy1-ap0"
```

`ucidef_set_led_netdev` writes an `/etc/config/system` LED section that `/etc/init.d/led` applies on first boot, so the DT trigger only survives until then — the two sources disagree about what these LEDs do (`phy*tpt` = TX/RX activity only, vs. the netdev default `link tx rx`).

Please pick one mechanism. The two closest siblings each do exactly one: [`mt7981b-routerich-ax3000-common.dtsi`](``https://github.com/openwrt/openwrt/blob/b866341838d51eb71136d352968e59181dd31817/target/linux/mediatek/dts/mt7981b-routerich-ax3000-common.dtsi#L93-L104``) keeps `phy0tpt` in DT and has no `01_leds` entry for the WLAN LEDs, while [`mt7987a-routerich-be7200.dts`](https://github.com/openwrt/openwrt/blob/b866341838d51eb71136d352968e59181dd31817/target/linux/mediatek/dts/mt7987a-routerich-be7200.dts) carries no `linux,default-trigger` and configures the LED from `01_leds` only.

---

target/linux/mediatek/filogic/base-files/etc/board.d/01_leds
```diff
@@ -235,6 +235,15 @@ openwrt,one)
 	ucidef_set_led_netdev "lanact" "LANACT" "green:lan" "eth1" "rx tx"
 	ucidef_set_led_netdev "lanlink" "LANLINK" "amber:lan" "eth1" "link"
 	;;
+teralink,tl3020-256mb)
```
openwrt-ai: nit: this case arm is inserted between `openwrt,one)` and `routerich,ax3000|`, which breaks the file's alphabetical ordering. It belongs after the `teltonika,rutc50)` arm and before `totolink,x6000r)`. The other four files this PR touches (`02_network`, `platform.sh`, `uboot-envtools/files/mediatek_filogic`, `uboot-mediatek/Makefile`) all place the new entry in sorted position, so this one is the odd one out.

---

target/linux/mediatek/dts/mt7981b-teralink-tl3020-256mb.dts
```diff
@@ -0,0 +1,369 @@
...
+
+	gpio-export {
+		compatible = "gpio-export";
+
+		gpio-0 {
+			gpio-export,name = "usb-enable";
+			gpio-export,output = <1>;
+			gpios = <&pio 14 GPIO_ACTIVE_HIGH>;
+		};
+	};
```
openwrt-ai, line 44: nit: this `usb-enable` GPIO export, together with `&usb_phy`/`&xhci` being enabled (lines 339-341, 367-369) and `kmod-usb3` in `DEVICE_PACKAGES`, means the board has a USB port — but the "Specification" block in the commit message and PR body doesn't mention USB at all. Per `.github/llm-review-rules.md` the hardware specification should let a stranger identify the device's ports; please add the USB port (and note that GPIO 14 is its power enable).

---

target/linux/mediatek/dts/mt7981b-teralink-tl3020-256mb.dts
```diff
@@ -0,0 +1,369 @@
...
+	model = "Teralink TL3020 256mb";
+	compatible = "teralink,tl3020-256mb", "mediatek,mt7981b";
+
+	aliases {
+		led-boot = &boot_led;
+		led-failsafe = &upgrade_led;
+		led-running = &boot_led;
+		led-upgrade = &upgrade_led;
+		serial0 = &uart0;
+	};
```
openwrt-ai, line 23: nit: no `label-mac-device` alias. The board has four distinct per-interface MACs from `config2`, so nothing tells userspace which one is printed on the label. Both close siblings set it — [`mt7981b-routerich-ax3000-common.dtsi` uses `label-mac-device = &wan;`](https://github.com/openwrt/openwrt/blob/b866341838d51eb71136d352968e59181dd31817/target/linux/mediatek/dts/mt7981b-routerich-ax3000-common.dtsi#L12) and [`mt7981b-netis-nx32u.dts` uses `label-mac-device = &gmac0;`](https://github.com/openwrt/openwrt/blob/b866341838d51eb71136d352968e59181dd31817/target/linux/mediatek/dts/mt7981b-netis-nx32u.dts#L15). Is the omission intentional, or should `&gmac0` (lanmac) / `&wan` (wanmac) be pointed at here?

---

openwrt-ai (commented):
## Commit checks

- `b8663418` "mediatek: add support Teralink TL3020" — the flash instructions are for the wrong device in one step: step 6 says to TFTP `openwrt-mediatek-filogic-netis_nx32u-initramfs-recovery.itb`. That is a leftover from the netis NX32U commit; the image this PR builds (and the one `bootfile` in `defenvs/teralink_tl3020-256mb_env` expects) is `openwrt-mediatek-filogic-teralink_tl3020-256mb-initramfs-recovery.itb`. Following the instructions verbatim would fetch a nonexistent file.
- `b8663418` — the message has two "Installation" headings; the first one is a one-line stub ("Replace the bootloader and flash sysupgrade.") that duplicates the real section below it. Also, the numbered list jumps 1, 2, 6, 7, 8, 9, 10 — steps 3-5 are missing, and it never states which firmware steps 1-2 (`scp` + `mtd write fip`/`bl2`) are run from (stock vs. OpenWrt). Since `fip` and `bl2` are marked `read-only` in the new kernel DTS, this matters for anyone trying to reproduce the install.
- `b8663418` — the "Specification" block is missing sections `.github/llm-review-rules.md` requires for a new-device commit: there is no UART entry at all (header location, pinout, baud rate), Power gives voltage but no current, and USB is not listed even though the DTS enables `&xhci`/`&usb_phy` and exports a USB power-enable GPIO (see inline comment).

Everything else lines up: compatible/`DEVICE_DTS`/`BUILD_DEVICES`/`UBOOT_CONFIG`/`ARTIFACT` names are consistent, the `config2` `mac-base` cells (`0x11` = 17-byte ASCII MAC) map to the interfaces the message documents, the U-Boot defenv `bootled_*`/`bootfile_*` values match the U-Boot DTS LED labels and the built artifact names, and the three hunk headers in `475-add-teralink_tl3020-256mb.patch` match their bodies. CI on `b8663418` had nothing failed at review time.

---

target/linux/mediatek/dts/mt7981b-teralink-tl3020-256mb.dts
```diff
@@ -0,0 +1,370 @@
...
+	band@1 {
+		reg = <1>;
+		nvmem-cells = <&macaddr_config_60a 0>;
+		nvmem-cell-names = "mac-address";
+	};
+};
+
+&xhci {
+	status = "okay";
+};
```
openwrt-ai, line 371: The commit message documents the port as "1x USB 2.0 Type-A", but the SoC's USB3 port is left enabled here. [`mt7981b-routerich-ax3000-common.dtsi:298-301`](``https://github.com/openwrt/openwrt/blob/d10947e3bf0b38ed3744b4f6ba4777dfa6192b4c/target/linux/mediatek/dts/mt7981b-routerich-ax3000-common.dtsi#L298-L301``) — the DTS this file is otherwise a near-copy of (identical GPIO 14 `usb-enable` export, identical LED map, identical `switch@1f` port 0/2/3/4 layout) — masks the U3 port off with `mediatek,u3p-dis-msk = <0x1>` for a USB 2.0-only port.

The evidence isn't unanimous: [`mt7981b-routerich-ax3000-v1.dts:335-337`](https://github.com/openwrt/openwrt/blob/d10947e3bf0b38ed3744b4f6ba4777dfa6192b4c/target/linux/mediatek/dts/mt7981b-routerich-ax3000-v1.dts#L335-L337) leaves `&xhci` bare. Is the omission intentional (i.e. the U3 lanes are actually routed), or should the mask be carried over?

---

package/boot/uboot-mediatek/patches/475-add-teralink_tl3020-256mb.patch
```diff
@@ -0,0 +1,371 @@
...
++			label = "reset";
++			linux,code = <KEY_RESTART>;
++			gpios = <&pio 1 GPIO_ACTIVE_LOW>;
++		};
++	};
++
++	gpio-leds {
++		compatible = "gpio-leds";
++
++		led_wan_red {
```
openwrt-ai: nit: the LED node names in the U-Boot DTS use underscores, which dtc flags under `node_name_chars_strict` and which no other board patch in `package/boot/uboot-mediatek/patches/` does (they all use `led-0`, `led-1`, … — see e.g. [`463-add-snr-snr-cpe-ax2.patch`](``https://github.com/openwrt/openwrt/blob/d10947e3bf0b38ed3744b4f6ba4777dfa6192b4c/package/boot/uboot-mediatek/patches/463-add-snr-snr-cpe-ax2.patch#L154-L157)).`` The kernel DTS added by this same PR already uses hyphens (`led-wan-red`, `led-power`, …), so the two halves disagree on naming for the same LEDs.

```suggestion
+		led-wan-red {
```

Same applies to `led_power` (line 179), `led_lan1`/`led_lan2`/`led_lan3` (184/189/194), `led_wan_blue` (199), `led_wlan2g` (204) and `led_wlan5g` (209). Note the `label` properties must stay as-is — `blue:power` and `red:wan` are referenced by `bootled_pwr`/`bootled_rec` in `defenvs/teralink_tl3020-256mb_env`.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -3153,6 +3153,32 @@ define Device/tenda_be12-pro
 endef
 TARGET_DEVICES += tenda_be12-pro
 
+define Device/teralink_tl3020-256mb
+  DEVICE_VENDOR := Teralink
+  DEVICE_MODEL := TL3020
+  DEVICE_VARIANT := 256mb
```
openwrt-ai, line 3234: nit: the same flash size is spelled three ways across this PR — `256mb` here (and in `model`/`NAME`/`compatible`), `(256MB)` in the U-Boot patch subject (`475-add-teralink_tl3020-256mb.patch` line 6), and `256 MiB` in the commit message's Specification block. Lowercase `mb` reads as megabits for a 256 MiB flash. There is in-tree precedent for the lowercase form ([`cudy_tr3000-256mb-v1`](https://github.com/openwrt/openwrt/blob/d10947e3bf0b38ed3744b4f6ba4777dfa6192b4c/target/linux/mediatek/image/filogic.mk#L1317-L1320)), and changing it here would ripple into the compatible string and image names, so this may well be deliberate — but the U-Boot patch subject at least could be aligned with whichever spelling you keep.

---

openwrt-ai (commented):
Re-reviewed at `d10947e3` (previous review was against `b8663418`, which was force-pushed away).

All four items from the last round are addressed: the netis image name and the duplicated/renumbered Installation section are gone from the commit message, the UART / USB / power-current entries were added to the Specification block, `label-mac-device = &gmac0;` was added, the `01_leds` arm moved into alphabetical position, and the WLAN LED double-configuration was resolved by dropping the `01_leds` netdev entries and keeping `phy0tpt`/`phy1tpt` in DT.

Re-verified against the current head: partition offsets are consistent between the kernel DTS, the U-Boot DTS and the documented `mtd5` install step; `0x680000 + 0xf980000 = 0x10000000` covers the full 256 MiB; the four `mac-base` cells (`0x11` = 17-byte ASCII) map to the interfaces the message documents; the three nested hunk headers in `475-add-teralink_tl3020-256mb.patch` (112/186/54) match their bodies; `UBOOT_CONFIG`/`BUILD_DEVICES`/`ARTIFACT/bl31-uboot.fip` resolve to the same `mt7981_teralink_tl3020-256mb` name; and the new `teralink,tl3020-256mb` case arms in `02_network`, `platform.sh` (both `platform_do_upgrade` and `platform_check_image`) and `uboot-envtools/files/mediatek_filogic` all land in the correct blocks. The commit message matches the diff — no commit checks this round. CI on `d10947e3` is green.

Three remaining items inline, one question and two nits.

---

target/linux/mediatek/dts/mt7981b-teralink-tl3020-256mb.dts
fildunsky: Confirmed USB 2.0 only — the U3 lanes are not routed, so I carried the mask over:

```
&xhci {
	mediatek,u3p-dis-msk = <0x1>;
	status = "okay";
};
```

Measured on the device before the change (USB 3.0 flash drive in the port, connector is black):

```
/:  Bus 001.Port 001: Dev 001, Class=root_hub, Driver=xhci-mtk/1p, 480M
    |__ Port 001: Dev 002, If 0, Class=[unknown], Driver=[none], 480M
/:  Bus 002.Port 001: Dev 001, Class=root_hub, Driver=xhci-mtk/1p, 20000M/x2
```

The device always enumerates on bus 1 at 480M (`new high-speed USB device number 2 using xhci-mtk`); nothing ever appears on the SuperSpeed root hub.

package/boot/uboot-mediatek/patches/475-add-teralink_tl3020-256mb.patch
fildunsky: Done — all eight LED node names in the U-Boot DTS now use hyphens (`led-wan-red`, `led-power`, `led-lan1`..`led-lan3`, `led-wan-blue`, `led-wlan2g`, `led-wlan5g`), matching the kernel DTS. The `label` properties are untouched, so `bootled_pwr`/`bootled_rec` in the defenv still resolve.

target/linux/mediatek/image/filogic.mk
fildunsky: Kept `256mb` in the compatible/model/image names as you guessed — renaming would churn the compatible string and all artifact names. Aligned the U-Boot patch subject with it instead: `board: mediatek: add Teralink TL3020 256mb`, and the body now reads `256mb (256 MiB SPI-NAND)` so the actual unit is spelled out once.

openwrt-ai (commented):

Re-reviewed at `135b57a3` (previous review was against `d10947e3`). The delta is 21 lines and resolves all three open items:

- `&xhci` now carries `mediatek,u3p-dis-msk = <0x1>;`, matching the "1x USB 2.0 Type-A" line in the Specification block. The `lsusb -t` evidence posted in the thread settles the question — the U3 lanes are not routed. Property name verified against [`xhci-mtk.c:559`](https://github.com/gregkh/linux/blob/v6.18.41/drivers/usb/host/xhci-mtk.c#L559) and [`mediatek,mtk-xhci.yaml:173`](https://github.com/gregkh/linux/blob/v6.18.41/Documentation/devicetree/bindings/usb/mediatek%2Cmtk-xhci.yaml#L173) on 6.18 (the target's `KERNEL_PATCHVER`); the placement before `status` matches [`mt7981b-alwaylink-m01k43.dts`](https://github.com/openwrt/openwrt/blob/135b57a318c0e67783334fb4e2ab705edf42e492/target/linux/mediatek/dts/mt7981b-alwaylink-m01k43.dts#L417-L421) and both orderings exist in-tree, so nothing to pick at.
- All eight LED node names in the U-Boot DTS switched from underscores to hyphens (`led-wan-red`, `led-power`, `led-lan1`..`led-lan3`, `led-wan-blue`, `led-wlan2g`, `led-wlan5g`), matching the kernel DTS. The `label` properties are untouched, so `bootled_pwr=blue:power` / `bootled_rec=red:wan` in `defenvs/teralink_tl3020-256mb_env` still resolve.
- The U-Boot patch subject is now `board: mediatek: add Teralink TL3020 256mb`, consistent with `DEVICE_MODEL`/`DEVICE_VARIANT`, the `model` string and the compatible; the body spells the unit out once as `256 MiB SPI-NAND`.

Re-verified after the edits: the three nested hunk headers in `475-add-teralink_tl3020-256mb.patch` still match their bodies exactly (112 / 186 / 54 lines counted), which is the thing an in-place edit of a quilt patch most easily breaks. The commit message is byte-identical to the previous round and still matches the diff, so no commit checks. CI on `135b57a3` is all green (kernel patch check, `mediatek/filogic` kernel build, package builds, FormalityCheck).

---

package/boot/uboot-mediatek/Makefile
```diff
@@ -737,6 +737,18 @@ define U-Boot/mt7981_snr_snr-cpe-ax2
...
+define U-Boot/mt7981_teralink_tl3020-256mb
+  NAME:=Teralink TL3020 256mb
+  BUILD_SUBTARGET:=filogic
+  BUILD_DEVICES:=teralink_tl3020-256mb
+  UBOOT_CONFIG:=mt7981_teralink_tl3020-256mb
+  UBOOT_IMAGE:=u-boot.fip
+  BL2_BOOTDEV:=spim-nand
+  BL2_SOC:=mt7981
+  BL2_DDRTYPE:=ddr3
+  DEPENDS:=+trusted-firmware-a-mt7981-spim-nand-ddr3
```
unisequence: `BL2_DDRTYPE:=ddr3` uses the generic variant. Did you check the actual DDR3 chip marking on the PCB? If it's a rated part (e.g. -1866), the preloader should target that specific timing profile rather than the generic default.

target/linux/mediatek/image/filogic.mk
```diff
@@ -3153,6 +3153,32 @@ define Device/tenda_be12-pro
...
+  DEVICE_DTS := mt7981b-teralink-tl3020-256mb
+  DEVICE_DTS_DIR := ../dts
+  UBINIZE_OPTS := -E 5
+  BLOCKSIZE := 128k
+  PAGESIZE := 2048
+  KERNEL_IN_UBI := 1
+  UBOOTENV_IN_UBI := 1
+  IMAGES := sysupgrade.itb
+  KERNEL_INITRAMFS_SUFFIX := -recovery.itb
+  KERNEL := kernel-bin | gzip
```
unisequence: nit: consider `libdeflate-gzip` instead — it's drop-in binary compatible with regular gzip but gives a better compression ratio, so images end up smaller for free. Would save having to switch this later (see #24148).

```suggestion
  KERNEL := kernel-bin | libdeflate-gzip
```

package/boot/uboot-mediatek/Makefile
fildunsky: Good catch — checked the marking, it is an ESMT M15T2G16128A with the `-DEB` speed code, i.e. DDR3(L)-1866 (13-13-13), so the generic profile was indeed leaving timing on the table. Switched to the rated variant:

- `BL2_DDRTYPE:=ddr3-1866`
- `DEPENDS:=+trusted-firmware-a-mt7981-spim-nand-ddr3-1866`
- `ARTIFACT/preloader.bin := mt7981-bl2 spim-nand-ddr3-1866`

Tested on the device: reflashed BL2 + FIP + sysupgrade, boots clean, all 256 MiB present, UBI attaches with `good PEBs: 1996, bad PEBs: 0, corrupted PEBs: 0` and no ECC complaints.

The RAM and flash part numbers are now in the commit message's Specification block as well:

```
- RAM       : DDR3 256 MiB (ESMT M15T2G16128A, DDR3-1866)
- Flash     : SPI-NAND 256 MiB (Winbond W25M02GV)
```

target/linux/mediatek/image/filogic.mk
fildunsky: Applied. Measured on a real build of this device: the FIT kernel sub-image went from 5,882,037 to 5,669,304 bytes — ~208 KiB (3.6%) smaller for free.

openwrt-ai (commented):

Re-reviewed at `aea2f381` (previous review was against `135b57a3`). Ignoring base-branch drift (the `netis_n6-v2` and `cudy_wr3000u-v1` hunks that appear in a raw range diff come from `main`, not from this PR), the delta to this PR is five lines plus a commit-message edit, and it resolves both open review items:

- **DDR3-1866 preloader.** `BL2_DDRTYPE:=ddr3-1866` / `DEPENDS:=+trusted-firmware-a-mt7981-spim-nand-ddr3-1866` in `package/boot/uboot-mediatek/Makefile` and `ARTIFACT/preloader.bin := mt7981-bl2 spim-nand-ddr3-1866` in `filogic.mk` are mutually consistent and resolve to a real variant: [`Trusted-Firmware-A/mt7981-spim-nand-ddr3-1866`](https://github.com/openwrt/openwrt/blob/aea2f381753ca8b7f7dcf66fd7135c5844b865fb/package/boot/arm-trusted-firmware-mediatek/Makefile#L247-L253) (`DDR_TYPE:=ddr3` + `DDR3_FREQ_1866:=1`), which is in that Makefile's variant list. The FIP recipe consumes `$(BL2_SOC)-$(BL2_BOOTDEV)-$(BL2_DDRTYPE)-bl31.bin`, i.e. `mt7981-spim-nand-ddr3-1866-bl31.bin`, matching what the TF-A package stages. The spelling matches the five other `ddr3-1866` U-Boot entries in the same file byte-for-byte. Non-UBI BL2 remains the right choice here (the `fip` partition is a fixed partition at `0x380000`, not a UBI volume), so `mt7981-spim-nand-ubi-ddr3-1866` correctly was not used.
- **`libdeflate-gzip`.** [`Build/libdeflate-gzip` at image-commands.mk:469](https://github.com/openwrt/openwrt/blob/aea2f381753ca8b7f7dcf66fd7135c5844b865fb/include/image-commands.mk#L469) exists and is a drop-in for `gzip`; `fit gzip` in `IMAGE/sysupgrade.itb` only labels the compression, so the pairing is unchanged. Same combination is used by the `netis_n6-v2` device that landed on `main` in the meantime.

Commit-message delta is limited to the Specification block (`RAM: DDR3 256 MiB (ESMT M15T2G16128A, DDR3-1866)`, `Flash: SPI-NAND 256 MiB (Winbond W25M02GV)`), which is exactly what the code change documents — no commit checks. The named flash part is a 2x1Gbit stacked device; U-Boot's SPI-NAND layer handles the die select for it ([`SPINAND_SELECT_TARGET(w25m02gv_select_target)` in `drivers/mtd/nand/spi/winbond.c`](https://github.com/u-boot/u-boot/blob/master/drivers/mtd/nand/spi/winbond.c#L444-L453)), so the `ubi` partition spanning the full `0x680000 + 0xf980000 = 0x10000000` is reachable from the bootloader as well. The kernel DTS, the U-Boot patch (including its three nested hunk headers), `01_leds`, `02_network`, `platform.sh` and `uboot-envtools` entries are untouched since the last round.

CI on `aea2f381` is all green (kernel patch check, `mediatek/filogic` kernel build, x86/malta package builds, FormalityCheck).

---

fildunsky:
@jonasjelonek would you please check?

openwrt-ai (commented):

Re-reviewed at `513e19b6` (previous review was against `aea2f381`). The update is a **rebase only** — comparing `git show aea2f381` against `git show 513e19b6`, the two patches are byte-identical apart from blob index lines and three hunk-header line offsets that moved because of base-branch drift (`01_leds` 311 → 317, `filogic.mk` 3196 → 3228, plus the `mediatek_filogic` index). The commit message is unchanged, so there is nothing new to check against the diff and no commit checks this round.

What I did verify is that the rebase landed cleanly against the new `main`, since the base picked up new mediatek/filogic devices (`comfast_cf-wa933*`, `netis_n6-v2`) that touch the same lists:

- `target/linux/mediatek/image/filogic.mk` — `define Device/teralink_tl3020-256mb` sits at line 3231, between `tenda_be12-pro` (3215) and `totolink_x6000r` (3257); still correctly sorted after the new neighbours were inserted.
- `target/linux/mediatek/filogic/base-files/etc/board.d/01_leds` — the `teralink,tl3020-256mb)` arm is still between `teltonika,rutc50)` and `totolink,x6000r)` (correct: `telt` < `tera` < `toto`).
- `package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic` — `teralink,tl3020-256mb|\` is still inside the `ubootenv_add_ubi_default` block, between `snr,snr-cpe-ax2|\` and `tplink,be450-ubi|\`, and did not get pushed into the adjacent `acer,predator-w6|\` block by the new `netis,*` entries above it.
- `02_network` and `platform.sh` (both `platform_do_upgrade` and `platform_check_image`) case arms are unchanged and still in their correct blocks.

Everything confirmed in earlier rounds still holds at this head: partition offsets agree between the kernel DTS, the U-Boot DTS and the documented `mtd5` step, `0x680000 + 0xf980000 = 0x10000000` covers the full 256 MiB, the four `config2` `mac-base` cells (`0x11` = 17-byte ASCII) map to the interfaces the message documents, and the three nested hunk headers in `475-add-teralink_tl3020-256mb.patch` (112 / 186 / 54) still match their bodies.

CI on `513e19b6` is all green — `Check Kernel patches (mediatek, filogic)`, `Build mediatek/filogic`, x86/64 and malta/be package builds, and all three FormalityCheck jobs.

---

jonasjelonek:

PR #23537, opened 2026-05-25
ath79: clean up unaligned hacks patch

neheb:
Replace the custom net_hdr_word() macro and struct net_hdr_word wrapper with the standard kernel get_unaligned()/put_unaligned() helpers. Also add the missing linux/unaligned.h include and drop the now-unused types.h addition.

Edited for 6.18 only to avoid duplicate work.

Assisted-by: Claude-Opus-4.7

ping @DragonBluep does this fix your issue?

PR #24524, opened 2026-08-01
kernel: drop redundant TCP fraglist GRO pull check

JuliusBairaktaris:
Both the `pending-6.12` and `pending-6.18` copies of
`690-net-add-missing-check-for-TCP-fraglist-GRO.patch` are redundant since upstream
commit f2bb3434544454099a5b6dec213567267b05d79d
("net: add pskb_may_pull() to skb_gro_receive_list()"), which fixes the same `BUG()`
inside the callee rather than at the call site.

Verified against the currently pinned kernels — the guard is present in both
`6.12.100` and `6.18.41` `net/core/gro.c`:

```c
int skb_gro_receive_list(struct sk_buff *p, struct sk_buff *skb)
{
	if (unlikely(p->len + skb->len >= 65536))
		return -E2BIG;

	if (!pskb_may_pull(skb, skb_gro_offset(skb))) {
		NAPI_GRO_CB(skb)->flush = 1;
		return -ENOMEM;
	}
	...
	skb_pull(skb, skb_gro_offset(skb));
```

Two independent reasons the pre-check is now dead weight: the callee refuses the short
skb itself, and it pulls with `skb_pull()` rather than the `__skb_pull()` that carries
the `BUG()`. `skb_gro_receive_list()` is the only thing the `is_flist` branch of
`tcp_gro_receive()` calls (`if (flush || skb_gro_receive_list(p, skb)) mss = 1;`), so
there is no second path.

The patch keeps applying cleanly only because the fix landed in a different function,
which is why it still looks alive.

**Series safety:** the only other patch touching `net/ipv4/tcp_offload.c` is ath79's
`900-unaligned_access_hacks.patch`, whose hunks are at lines 66, 272 and 336 — all
above the removed line at 359, so nothing shifts.


neheb (approved):

neheb:
speaking of the ath79 patch: https://github.com/openwrt/openwrt/pull/23537

That thing was probably needed in the past. I highly doubt it now.

JuliusBairaktaris:
Ping @jonasjelonek for merge

target/linux/generic/pending-6.18/690-net-add-missing-check-for-TCP-fraglist-GRO.patch
```diff
@@ -1,26 +0,0 @@
...
- net/ipv4/tcp_offload.c | 1 +
- 1 file changed, 1 insertion(+)
-
---- a/net/ipv4/tcp_offload.c
-+++ b/net/ipv4/tcp_offload.c
-@@ -359,6 +359,7 @@ struct sk_buff *tcp_gro_receive(struct l
- 		flush |= (__force int)(flags ^ tcp_flag_word(th2));
- 		flush |= skb->ip_summed != p->ip_summed;
- 		flush |= skb->csum_level != p->csum_level;
-+		flush |= !pskb_may_pull(skb, skb_gro_offset(skb));
```
openwrt-ai, line 23: nit: the commit message says "The fix is present in both pinned kernels, 6.12.100 and 6.18.39", but this branch's merge base already pins `LINUX_VERSION-6.18 = .41` (`target/linux/generic/kernel-6.18`, bumped in `d3191cb05603` "kernel: bump 6.18 to 6.18.41"). The PR body repeats the same `6.18.39`. Worth updating to 6.18.41 so the stated verification matches the tree the patch is being dropped from.

The substance checks out at .41 though — I verified the callee guard in the pinned trees: [`skb_gro_receive_list()` net/core/gro.c:237 @ v6.18.41](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/net/core/gro.c#L232-L240) and [net/core/gro.c:236 @ v6.12.100](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/net/core/gro.c#L231-L239), both refusing the short skb before the `skb_pull()` and setting `NAPI_GRO_CB(skb)->flush = 1`. Upstream `f2bb3434544454099a5b6dec213567267b05d79d` ("net: add pskb_may_pull() to skb_gro_receive_list()") is real and matches that description. `skb_gro_receive_list()` is indeed the only call out of the `is_flist` branch, and no in-tree patch touches `net/core/gro.c`, so nothing else can remove the guard.

---

openwrt-ai (commented):
Reviewed 1 new commit. The removal itself looks correct — both pinned kernels carry the upstream guard in `skb_gro_receive_list()`, and the ath79 `900-unaligned_access_hacks.patch` hunks (66/272/336) all sit above the removed line, so no context shifts. One inline nit on a stale kernel version stated in the commit message and PR body.

---

openwrt-ai (commented):

The force-push is a rebase onto current `main` plus the commit-message version fix from my earlier nit — the two removed patch files are byte-identical to the previous revision, and the commit message and PR body now both say `6.12.100 and 6.18.41`, matching `LINUX_VERSION-6.12 = .100` / `LINUX_VERSION-6.18 = .41` in `target/linux/generic/kernel-6.12` and `kernel-6.18` on the new merge base. Re-confirmed the callee guard in the pinned trees at the rebased versions: [`skb_gro_receive_list()` net/core/gro.c:232-240 @ v6.18.41](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/net/core/gro.c#L232-L240) and [net/core/gro.c:231-239 @ v6.12.100](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/net/core/gro.c#L231-L239).

---

jonasjelonek:

PR #24525, opened 2026-08-01
dropbear: drop obsolete rsa-sha2-256 pubkey patch

JuliusBairaktaris:
`910-signkey-fix-use-of-rsa-sha2-256-pubkeys.patch` (2020) worked around dropbear's
then-new strict signature-type check, which made `rsa-sha2-256` pubkeys unusable because
RFC 8332 reuses the `ssh-rsa` *key* format for the `rsa-sha2-256` *signature* algorithm.

Dropbear 2026.92 handles that itself, in `src/svr-authpubkey.c`:

```c
	sigtype = signature_type_from_name(sigalgo, sigalgolen);
	...
	keytype = signkey_type_from_signature(sigtype);
	keyalgo = signkey_name_from_type(keytype, &keyalgolen);   /* rsa-sha2-256 -> ssh-rsa */
	...
	int status = checkpubkey(keyalgo, keyalgolen, keyblob, keybloblen, &pubkey_options);
```

The signature type is mapped back to the key algorithm name before `authorized_keys` is
consulted — exactly what the patch was working around. A compliant client naming its
signature `rsa-sha2-256` satisfies `expect_sigtype == sigtype` unaided.

### Keeping it is not free

`buf_verify()`'s `expect_sigtype` comes from the userauth request's `sigalgo` field,
while `sigtype` is read from the signature blob — two separately attacker-controlled
values. OpenWrt builds with `DROPBEAR_RSA_SHA1 0` (`localoptions.h:53`), which compiles
out only the *lookup*, not the fallback:

```c
enum signature_type signature_type_from_name(const char* name, unsigned int namelen) {
#if DROPBEAR_RSA_SHA1
	if (... memcmp(name, SSH_SIGNKEY_RSA, namelen) == 0)
		return DROPBEAR_SIGNATURE_RSA_SHA1;
#endif
	return (enum signature_type)signkey_type_from_name(name, namelen);
}
```

So a blob naming itself `ssh-rsa` yields `DROPBEAR_SIGNKEY_RSA`, which with
`DROPBEAR_RSA=1, DROPBEAR_DSS=0` is the first enumerator — **0**. That is not
`DROPBEAR_SIGNATURE_NONE` (90), so the patch's own "No signature type" guard passes it,
and with `expect_sigtype == DROPBEAR_SIGNATURE_RSA_SHA256` the relaxed check is skipped.
`buf_rsa_verify()` then reaches `rsa_pad_em()`:

```c
	switch (sigtype) {
#if DROPBEAR_RSA_SHA256
		case DROPBEAR_SIGNATURE_RSA_SHA256:   /* 101 */
			...
#endif
		default:
			assert(0);
	}
```

`sigtype` is 0, so it lands on `default: assert(0)` and the process aborts. There is no
`-DNDEBUG` in the dropbear build (this is libc `assert`, not `dropbear_assert`, so
`140-disable_assert.patch` does not cover it); had NDEBUG been set, control would fall
out of the switch with `hash_desc == NULL` and the next statement dereferences it.

Upstream's unconditional check rejects the mismatch cleanly instead.

### Scope

This is **not** an authentication bypass and not pre-auth. `buf_verify()` is only
reached after `checkpubkey()` succeeds, so it needs a key already listed in the target's
`authorized_keys` — the attacker can simply log in instead. No type confusion is
possible either: naming `rsa-sha2-256` forces the loaded key to be RSA, so `key->rsakey`
is the only populated member.

The patch's extra `DROPBEAR_SIGNATURE_NONE` guard is not lost with it —
`svr-authpubkey.c:159` already rejects that before `buf_verify()` is reached.

`PKG_RELEASE` bumped. No other dropbear patch touches `src/signkey.c`.


package/network/services/dropbear/patches/910-signkey-fix-use-of-rsa-sha2-256-pubkeys.patch
```diff
@@ -1,44 +0,0 @@
...
- signkey.c | 8 ++++++--
- 1 file changed, 6 insertions(+), 2 deletions(-)
-
---- a/src/signkey.c
-+++ b/src/signkey.c
-@@ -656,10 +656,18 @@ int buf_verify(buffer * buf, sign_key *k
- 	sigtype = signature_type_from_name(type_name, type_name_len);
- 	m_free(type_name);
- 
--	if (expect_sigtype != sigtype) {
```
openwrt-ai, line 28: The commit message's central justification — `svr-authpubkey.c` deriving the key algorithm via `signkey_type_from_signature()` before `authorized_keys` is consulted — governs the `checkpubkey()` lookup, not the value `buf_verify()` receives as `expect_sigtype`. Restoring this unconditional check is only safe if, at the `buf_verify()` call site in 2026.92, `expect_sigtype` is derived from the same userauth `sigalgo` string that the signature blob's `sigtype` is parsed from; the quoted snippet does not establish that.

For a compliant client both are `rsa-sha2-256`, so the conclusion is very likely right — but this is the single fact the whole removal rests on, and it is exactly the regression the 2020 patch was added for (`ssh-rsa` key in `authorized_keys` + `rsa-sha2-256` signature). Could you state the `expect_sigtype` provenance explicitly in the commit message, and confirm an RSA pubkey login was exercised against 2026.92 with the patch dropped? A code-reading argument alone is thin for a change to a login path.

---

package/network/services/dropbear/patches/910-signkey-fix-use-of-rsa-sha2-256-pubkeys.patch
```diff
@@ -1,44 +0,0 @@
...
--			dropbear_exit("Non-matching signing type");
-+	if (sigtype == DROPBEAR_SIGNATURE_NONE) {
-+		dropbear_exit("No signature type");
- 	}
- 
-+#if DROPBEAR_RSA
```
openwrt-ai, line 34: nit: worth adding to the commit message — the replacement check is nested inside `#if DROPBEAR_RSA` / `#if DROPBEAR_RSA_SHA256` (lines 34-40), so it only exists in RSA-enabled builds. [`config DROPBEAR_RSA` is `default y` but user-selectable](https://github.com/openwrt/openwrt/blob/ed431f63f9b0a8768cad4107904598e2dcea8e46/package/network/services/dropbear/Config.in#L879-L888), and an ECDSA/Ed25519-only build therefore has *no* "Non-matching signing type" check at all today, only the `DROPBEAR_SIGNATURE_NONE` guard — an ECDSA-named signature blob would be accepted with `expect_sigtype == DROPBEAR_SIGNATURE_ED25519`.

That is a stronger argument for dropping the patch than the `assert(0)` path described in the message (which needs the default `DROPBEAR_RSA_SHA1=0` + `DROPBEAR_RSA=1` combination), and it makes the removal a strict improvement across all config permutations rather than just the default one.

---

openwrt-ai (commented):
Reviewed 1 new commit. The diff is minimal and self-consistent: the patch is removed cleanly, `PKG_RELEASE` is bumped, no other dropbear patch touches `src/signkey.c`, and nothing else in the tree references the removed file. Two comments on the rationale rather than the code.

Verified in-tree while reading: `140-disable_assert.patch` only neuters `dropbear_assert()` in `src/dbutil.h`, not libc `assert()`, and `-DNDEBUG` is only injected by `include/cmake.mk` — dropbear is autotools — so the commit message's claim that the `assert(0)` would fire is correct for this build.

Note: the `mkj/dropbear` reference tree was unavailable in this session, so upstream 2026.92 sources could not be inspected directly; the comments below are based on the in-tree patch and OpenWrt's build configuration only.

---

openwrt-ai (commented):

`git diff ed431f63..01a920a0` is empty — the tree change is byte-identical to what I reviewed previously (patch file removed, `PKG_RELEASE` 1 → 2). The only change is the commit message, and it addresses both of my earlier comments:

- **`expect_sigtype` provenance** ([r3697807085](https://github.com/openwrt/openwrt/pull/24525#discussion_r3697807085)) — now traced explicitly through `svr-authpubkey.c:113` → `:126` → `:233` and `signkey.c:655` → `:659`, establishing that both sides of the comparison come from wire-supplied algorithm names rather than the `authorized_keys` key format. The message also states the change was exercised empirically: pristine 2026.92 with `DROPBEAR_RSA 1` / `DROPBEAR_RSA_SHA1 0`, OpenSSH pinned to `PubkeyAcceptedAlgorithms=rsa-sha2-256`, `ssh-rsa` entry in `authorized_keys`, login succeeds without the patch. That is exactly the regression case the 2020 patch was written for, so my "code-reading argument alone is thin for a login path" objection is satisfied.
- **`#if DROPBEAR_RSA` / `#if DROPBEAR_RSA_SHA256` narrowing** ([r3697807481](https://github.com/openwrt/openwrt/pull/24525#discussion_r3697807481)) — now a paragraph in the message, correctly framing the removal as restoring the check for every configuration rather than only the default one.

Commit message matches the diff; `PKG_RELEASE` bump is appropriate for a patchset change. CI on `01a920a0` is green or still running (no failures).

Caveat carried over from the previous review: the `mkj/dropbear` reference tree was again unavailable in this session and no 2026.92 tarball is present locally, so the cited upstream line numbers and `signature_type_from_name()` / `rsa_pad_em()` bodies could not be re-verified against real sources. The in-tree facts I could check hold: `140-disable_assert.patch` only touches `dropbear_assert()` in `src/dbutil.h`; no remaining dropbear patch touches `src/signkey.c`; nothing else in the tree references the removed file; `DROPBEAR_RSA` is `default y` but user-selectable in [`Config.in`](https://github.com/openwrt/openwrt/blob/01a920a0b6c374c430a6724af1dfcf0a10c0cd86/package/network/services/dropbear/Config.in#L878-L888).

---

PR #24335, opened 2026-07-21
qualcommax: add support for MikroTik Chateau 5G R17 ax

ddimension:
Adds support for the MikroTik Chateau 5G R17 ax (ordering code
`S53UG+5HaxD2HaxD&RG650E-EU`), the first MikroTik device on the qualcommax
target.

IPQ6010, 1 GB RAM, 128 MB NAND, 4x 1G (QCA8075) plus 1x 2.5G (QCA8081),
integrated ath11k WLAN, and a Quectel RG650E-EU 5G modem on USB. Full
specifications, serial console pinout, flash layout and installation
procedure are in the commit message of the last patch.

### What the first patch is for

**`generic: mikrotik: size RouterBoot partitions so they are usable`** —
the device does not work without it, but it is not device specific:
`hard_config` was capped at `RB_BLOCK_SIZE`, which makes the WLAN
calibration tag unreadable on this board (the tag list spans 5576 bytes
because the calibration data alone is 5100 bytes of LZ77). The same
patch grows `soft_config` to a full erase block where the space is free,
because at 4 KiB on a flash with 64 KiB erase blocks MTD always forces
it read-only — which makes RouterBOOT's boot settings unwritable and
etherboot impossible to trigger from software. `rb_softconfig` needed a
matching change so that the CRC still covers only the region RouterBOOT
validates.

Earlier revisions of this PR carried two more kernel patches; both are
dropped:

- The NSS crypto clock patch turned out to be purely cosmetic. Booting
  with and without it and comparing `clk_summary` shows the clock ends up
  at 300 MHz with a running PLL either way; the patch only suppressed the
  one-time `nss_crypto_clk_src: rcg didn't update its configuration`
  warning that appears when the boot chain (RouterBOOT here) has not
  pre-enabled the PLL. The warning is harmless and remains visible on
  this device; it is not addressed in this PR.

- The PSGMII VCO PLL calibration patch could not be proven. The board
  shows intermittent bit errors whose loss scales with frame size in
  roughly one boot out of three; the patch made seven consecutive boots
  come up clean, which given the failure rate is suggestive but not
  proof. It will be submitted separately once it is actually proven.
  Until then this is a known issue of the board.

### WLAN board data

This unit's board data file lives in RouterBOOT, not in `ipq-wifi`. `hard_config`
tag 0x16 holds 5100 bytes of LZ77 which `rb_hardconfig` decompresses into a 64 KiB
complete BDF: the `01 00 04 04` magic of the `ipq-wifi` files, a valid header
[...]

PR #24526, opened 2026-08-02
qualcommax: add support for MikroTik hAP ax3

Jan64X:
Depends on PR #24335.

Adds support for the MikroTik hAP ax3, ordering code
C53UiG+5HPaxD2HPaxD.

The hAP ax3 shares most of its hardware and RouterBOOT integration with
the MikroTik Chateau 5G R17 ax added by the dependent PR.

IPQ6010, 1 GB RAM, 128 MB NAND, 4x 1G Ethernet through a QCA8075,
1x 2.5G Ethernet through a QCA8081, integrated ath11k WLAN, and one
external USB 3.0 port. Full specifications, serial console connections,
flash layout and installation procedure are in the commit message of
the last patch.

What the first patch is for

The first patch mechanically moves the hardware shared by the Chateau
and hAP ax3 into ipq6010-mikrotik-c53-s53.dtsi. Chateau-specific modem
controls, signal LEDs, sensors, PHY addressing and port mapping remain
in its device DTS. No functional change to the Chateau is intended.

Hardware differences

RouterOS declares GPIO75 as its PHY reset, and it must be held high for
the QCA8075 to remain available. Pulling GPIO77 low removes all PHYs from
the MDIO bus and drops the QCA8081 link, so OpenWrt uses it as the shared
PHY-domain reset.

The flash layout and RouterBOOT handling are the same as on the Chateau,
so the LZMA loader and yafut sysupgrade path from the dependent PR are
reused.

Testing

Verified on hardware:

  - RouterBOOT etherboot and initramfs
  - RouterOS to OpenWrt installation
  - OpenWrt sysupgrade
  - warm reboot and cold boot
  - all five Ethernet ports, including socket order and factory MACs
  - 2.4 GHz WLAN
  - 5 GHz WLAN with HE80 on channel 36
  - USB 3.0 SuperSpeed storage
  - reset, mode and WPS buttons
  - status, WLAN, LAN, port and PoE indicator LEDs

Board data

Per-device WLAN calibration and board data are read from RouterBOOT
hard_config tag 0x16. rb_hardconfig exposes the decompressed complete
64 KiB board-data blob, which preinit installs as board.bin for ath11k's
API-1 fallback.

Notes for reviewers

PoE output is deliberately disabled. RouterOS controls it through
GPIO66, but the current-sense input and safe power sequencing need to be
understood and represented by a Linux PSE driver before enabling it.


target/linux/qualcommax/dts/ipq6010-mikrotik-hap-ax3.dts
```diff
@@ -0,0 +1,158 @@
...
+		color = <LED_COLOR_ID_GREEN>;
+		function = LED_FUNCTION_LAN;
+		function-enumerator = <4>;
+		gpios = <&tlmm 27 GPIO_ACTIVE_HIGH>;
+	};
+
+	poe {
+		color = <LED_COLOR_ID_GREEN>;
+		function = LED_FUNCTION_INDICATOR;
+		function-enumerator = <6>;
```
openwrt-ai: `LED_FUNCTION_INDICATOR` plus `function-enumerator = <6>` composes the sysfs name `green:indicator-6` (`led_compose_name()` uses `"%s:%s-%d"`). There is no indicator 1..5 on this board, so the enumerator is arbitrary, and the resulting name says nothing about what the LED shows.

`function` falls back to a free-form string when no `LED_FUNCTION_*` constant fits, and that is already done in-tree for PoE — see [`rtl9303_hasivo_s1100wp-8xgt-se.dts:97`](https://github.com/openwrt/openwrt/blob/87ea0918532eb48d9f9cd149316214aa2be8a7ad/target/linux/realtek/dts/rtl9303_hasivo_s1100wp-8xgt-se.dts#L97). That gives `green:poe`, which is what a user would look for.

```suggestion
		function = "poe";
```

---

target/linux/qualcommax/image/ipq60xx.mk
```diff
@@ -35,6 +35,67 @@ define Device/8devices_mango-dvk
...
+
+define Device/mikrotik_ipq6010_nand
+	$(call Device/UbiFit)
+	KERNEL := kernel-bin | mikrotik-lzma-loader
+	KERNEL_INITRAMFS := kernel-bin | mikrotik-lzma-loader
+	DEVICE_VENDOR := MikroTik
+	SOC := ipq6010
+	BLOCKSIZE := 128k
+	PAGESIZE := 2048
+	DEVICE_PACKAGES := yafut
```
openwrt-ai, line 67: `yafut` moved up into the shared template, but the comment that explains why it is needed stayed behind in `Device/mikrotik_chateau-5g-r17-ax` (lines 72-73), where it now documents a package that block no longer adds. Move it to where the package is actually pulled in:

```suggestion
	# yafut: required by sysupgrade. The kernel partition is yaffs, whose
	#   metadata and ECC live in the OOB area, so "mtd write" cannot be used.
	DEVICE_PACKAGES := yafut
```

and drop lines 72-73 from the Chateau block.

---

target/linux/qualcommax/image/ipq60xx.mk
```diff
@@ -35,6 +35,67 @@ define Device/8devices_mango-dvk
...
+	DEVICE_PACKAGES += kmod-usb-net-qmi-wwan kmod-usb-serial-option uqmi \
+		ipq-wifi-mikrotik_chateau-5g-r17-ax
+endef
+TARGET_DEVICES += mikrotik_chateau-5g-r17-ax
+
+define Device/mikrotik_hap-ax3
+	$(call Device/mikrotik_ipq6010_nand)
+	DEVICE_MODEL := C53UiG+5HPaxD2HPaxD (hAP ax3)
+	DEVICE_DTS := ipq6010-mikrotik-hap-ax3
+	DEVICE_PACKAGES += block-mount kmod-fs-vfat kmod-usb-storage \
```
openwrt-ai: nit: sister-device parity — the Chateau has the same single external USB 3.0 type-A port (its commit message lists `USB 1x 3.0 type A, external`) but ships no storage stack at all ([`DEVICE_PACKAGES` at ipq60xx.mk:85](https://github.com/openwrt/openwrt/blob/87ea0918532eb48d9f9cd149316214aa2be8a7ad/target/linux/qualcommax/image/ipq60xx.mk#L85)), while this device pulls in `block-mount kmod-fs-vfat kmod-usb-storage`. Is the difference deliberate, or should the two boards get the same set — either both with the storage packages or both leaving them to the user?

---

target/linux/qualcommax/dts/ipq6010-mikrotik-c53-s53.dtsi
```diff
@@ -0,0 +1,367 @@
...
+		};
+
+		wlan {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WLAN;
+			gpios = <&tlmm 22 GPIO_ACTIVE_HIGH>;
+			linux,default-trigger = "phy0radio";
+		};
+
+	};
```
openwrt-ai: nit: stray blank line before the closing brace — left over from the LEDs that moved out to the device DTS. No other node in this file does that.

```suggestion
};
```

Same at lines 242-243, at the end of `&mdio`, where the QCA8075 package node moved out.

---

openwrt-ai (commented):
Reviewed the two commits that are new in this PR (`8ed896f`, `87ea091`). `e0f842f` and `1c82ce6` belong to the dependency #24335 and were not re-reviewed here.

On the DTS split: I ran both the pre-split `ipq6010-mikrotik-chateau-5g-r17.dts` and the post-split `.dtsi` + `.dts` pair through the preprocessor and compared the normalized property sets. They are identical apart from the added labels and `label-mac-device = &dp1` → `&switch_port4` (same node), so the "no functional change for the Chateau" claim in `8ed896f` holds.

## Commit checks

- `87ea091` "qualcommax: ipq60xx: add support for MikroTik hAP ax3" — as a new-device commit the message is missing three things its sibling `1c82ce6` (Chateau) does carry:
  - **UART and power specs.** The serial section gives the header location and pinout but no baud rate or logic level (the DTS `bootargs` say `115200n8`), and there is no power line (input voltage range / max draw) in the specification block at all.
  - **Flash instructions are not reproducible as written.** "Installation is performed by booting the initramfs image through RouterBOOT etherboot and running sysupgrade" does not name the image to use, nor how to get the board into etherboot — reset-button hold, or `boot_device=ethonce` plus `commit` through the `rb_softconfig` sysfs interface, which `1c82ce6` spells out verbatim.
  - **No per-interface MAC layout.** The message says the ports "use their factory MAC addresses from RouterBOOT hard_config" but not which base+offset maps to which interface. The DTS encodes base+0 → wan (ether1) and base+1..4 → lan1..lan4; that mapping, plus where the two radio MACs come from, is the part a reader cannot reconstruct from the message.

---

target/linux/qualcommax/dts/ipq6010-mikrotik-hap-ax3.dts
Jan64X: Fixed. I also noticed I had it tagged as color green, which was incorrect, so I changed it to the correct color red.

target/linux/qualcommax/image/ipq60xx.mk
Jan64X: fixed

target/linux/qualcommax/image/ipq60xx.mk
Jan64X: Keeping them the same makes sense, removed `block-mount kmod-fs-vfat kmod-usb-storage` from DEVICE_PACKAGES for hAP ax3.

target/linux/qualcommax/dts/ipq6010-mikrotik-c53-s53.dtsi
Jan64X: fixed

openwrt-ai (commented):

Re-reviewed `d06145c` and `f201ce0` against the previously reviewed `87ea091`. The delta is three files, and all of it addresses the earlier threads:

- `poe` LED now uses `color = <LED_COLOR_ID_RED>` + `function = "poe"`. There is no `LED_FUNCTION_POE` in `include/dt-bindings/leds/common.h` on either 6.12 or 6.18, and the free-form string matches existing in-tree use (`rtl9303_hasivo_s1100wp-*.dts`), so `red:poe` is the right shape. Nothing else in the tree references the old `green:indicator-6` name — `01_leds` does not configure this LED.
- The `yafut` comment now sits with `DEVICE_PACKAGES := yafut` in `Device/mikrotik_ipq6010_nand`, and the stale copy is gone from the Chateau block.
- `block-mount kmod-fs-vfat kmod-usb-storage` dropped, so the two MikroTik boards now carry the same package set.
- Both stray blank lines in `ipq6010-mikrotik-c53-s53.dtsi` removed; the remaining blank-line usage in that file is consistent.

Commit checks pass for both new commits. `f201ce0` now carries the UART baud rate/level and the power line in the spec block, a reproducible install path (image names, reset-button hold, and the `boot_device=ethonce` + `rb_softconfig` alternative), and a per-interface MAC layout that matches the `macaddr_hard 0..4` cells in the DTS — the three gaps flagged last time. `d06145c`'s diff is confined to the two DTS files its subject describes.

CI on `f201ce0` shows no failed check runs.

---

Jan64X:
Further testing uncovered transmit-side corruption on the QCA8081 port. Traffic received by the hAP is seemingly clean, but traffic transmitted through the QCA8081 causes CRC errors on the peer and heavy TCP retransmissions. The four QCA8075 ports seem to be unaffected. I will investigate when I get the time again (most likely next weekend).

target/linux/qualcommax/dts/ipq6010-mikrotik-hap-ax3.dts
```diff
@@ -0,0 +1,157 @@
...
+};
+
+&switch_port5 {
+	label = "wan";
+	nvmem-cells = <&macaddr_hard 0>;
+	nvmem-cell-names = "mac-address";
+};
+
+&wifi {
+	qcom,ath11k-calibration-variant = "MikroTik-hAP-ax3";
```
openwrt-ai, line 156: This variant string only feeds the board-2.bin lookup ([`ath11k_core_get_bdf_ext()`, core.c:1466-1470](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/ath/ath11k/core.c#L1466-L1470)), and with the `ipq-wifi-mikrotik_hap-ax3` package dropped nothing ships an entry it can match — these two boards are now the only qualcommax DTS files setting the property without a matching `ipq-wifi` package. Is it kept deliberately (e.g. for a future upstream BDF entry), or should it go along with the package? Same at `ipq6010-mikrotik-chateau-5g-r17.dts:191`.

---

openwrt-ai (commented):

---

target/linux/qualcommax/ipq60xx/base-files/etc/board.d/02_network
```diff
@@ -23,6 +23,9 @@ ipq60xx_setup_interfaces()
 		# No dedicated WAN port, the uplink is the RG650E modem.
 		ucidef_set_interface_lan "lan1 lan2 lan3 lan4 lan5"
 		;;
+	mikrotik,hap-ax3)
+		ucidef_set_interfaces_lan_wan "lan1 lan2 lan3 lan4" "wan"
+		;;
```
openwrt-ai, line 28: nit: this is a second branch with the exact same body as the `glinet,gl-ax1800|...|yuncore,fap650` case at lines 33-39. Add `mikrotik,hap-ax3|\` to that existing group instead.

---

openwrt-ai (commented):
Reviewed 2 new commits.

---

PR #24527, opened 2026-08-02
ramips: move sound drives to files

neheb:
This is local only. Also has never been upstreamed.

ping @jonasjelonek


target/linux/ramips/files/sound/soc/ralimk/Kconfig
Rondom, line 1: s/ralimk/ralink/

PR #24529, opened 2026-08-02
ath10k-ct: update to latest version 7.2

DragonBluep:
Update to Git HEAD (2026-08-01) and switch to 7.2 kernel based driver. We also introduced a new patch 204-* to correct the driver version number.

Upstream merged patches:
- 005-ath10k-ct-fix-missing-prototypes-warnings-on-6.12-ke.patch
- 300-fix-fortify-checking-error.patch

Tested on QCA9887 and IPQ4019.


package/kernel/ath10k-ct/Makefile
```diff
@@ -8,14 +8,14 @@ PKG_LICENSE_FILES:=
 
 PKG_SOURCE_URL:=https://github.com/greearb/ath10k-ct.git
 PKG_SOURCE_PROTO:=git
-PKG_SOURCE_DATE:=2025-12-01
-PKG_SOURCE_VERSION:=bb84e1597474a6db26f7681d6765fd3e69fff37e
-PKG_MIRROR_HASH:=76a950c4c5f02128f4225955b1ca0d1d8f6c68c3b0918c4bc2d1265698e939a6
+PKG_SOURCE_DATE:=2026-08-01
```
openwrt-ai: nit: `PKG_SOURCE_VERSION` `fcbdb70` is dated 2026-07-31 upstream, not 2026-08-01 — both its author and committer date are `Fri Jul 31 13:53:38 2026 -0700` (= `2026-07-31T20:53:38Z`), and it is still the current `HEAD` of greearb/ath10k-ct. `PKG_SOURCE_DATE` is conventionally the date of the referenced commit.

```suggestion
PKG_SOURCE_DATE:=2026-07-31
```

Same applies to "Update to Git HEAD (2026-08-01)" in the commit message and PR body.

---

package/kernel/ath10k-ct/patches/202-ath10k-use-tpt-trigger-by-default.patch
```diff
@@ -9,14 +9,14 @@ traffic.
 
 Signed-off-by: Mathias Kresin <dev@kresin.me>
 ---
- ath10k-6.18/core.h | 4 ++++
- ath10k-6.18/leds.c | 4 +---
- ath10k-6.18/mac.c  | 2 +-
+ ath10k-7.2/core.h | 4 ++++
+ ath10k-7.2/leds.c | 4 +---
+ ath10k-7.2/mac.c  | 2 +-
  3 files changed, 6 insertions(+), 4 deletions(-)
```
openwrt-ai: nit: the diffstat header was rebased to `ath10k-7.2/` but not updated for the changed body. Upstream 7.2 already dropped `ar->leds.cdev.default_trigger = ar->leds.wifi_led.default_trigger;` (and the `wifi_led` setup above it), so the `leds.c` hunk is now a pure 1-line insertion — the header still claims 3 deletions there.

```suggestion
 ath10k-7.2/leds.c | 1 +
 ath10k-7.2/mac.c  | 2 +-
 3 files changed, 6 insertions(+), 1 deletion(-)
```

---

package/kernel/ath10k-ct/patches/005-ath10k-ct-fix-missing-prototypes-warnings-on-6.12-ke.patch
```diff
@@ -1,79 +0,0 @@
-From: Shiji Yang <yangshiji66@outlook.com>
-Date: Mon, 28 Apr 2025 23:40:15 +0800
-Subject: [PATCH] ath10k-ct: fix missing-prototypes warnings on 6.12 kernel
-
-Fix various missing-prototypes warnings by:
-
-* Mark functions only used in current source file as static.
-* Comment out unused functions.
```
openwrt-ai, line 8: Only partially merged upstream, as far as I can tell. Upstream fixed the two "unused function" cases by marking them `static` instead of commenting them out — [`static void ath10k_debug_fw_ratepwr_table_process()` at ath10k-7.2/debug.c:427](https://github.com/greearb/ath10k-ct/blob/fcbdb70debc261f9df6734bbb76d81cdc88e0e26/ath10k-7.2/debug.c#L427) and [`static void ath10k_debug_fw_powerctl_table_process()` at ath10k-7.2/debug.c:441](https://github.com/greearb/ath10k-ct/blob/fcbdb70debc261f9df6734bbb76d81cdc88e0e26/ath10k-7.2/debug.c#L441) — but neither has any caller left in `ath10k-7.2/`. So dropping this patch entirely trades the old `-Wmissing-prototypes` warnings for two new `-Wunused-function` warnings on every build (harmless with `# CONFIG_WERROR is not set`, but noisy).

The `static` conversions in the other three files (`ath10k_refresh_target_pdev_ext_ct_stats`, `ieee80211_is_robust_mgmt_frame_tx`, `ath10k_wmi_event_txbf_cv_mesg`) are genuinely upstream now. Is it worth keeping a trimmed-down 005 with just the two `/* */` hunks, or reporting the dead functions upstream?

---

openwrt-ai (commented):
Reviewed 1 new commit.

Verification done locally against greearb/ath10k-ct `fcbdb70` (confirmed to be the current upstream `HEAD`): all seven patches apply to `ath10k-7.2/` with zero fuzz and zero offset, and both dropped patches are genuinely fixed upstream — `qca9880_set_ctl_table_cmd.ctl_info` is now a pointer (300), and `kzalloc_obj()`/`kzalloc_objs()` used by the 7.2 sources do exist in 6.18 (`include/linux/slab.h`), so the driver-version jump looks safe against backports 6.18.39. The new 204 patch covers the only stale version string in the tree.

Minor points inline.

---

package/kernel/ath10k-ct/patches/005-ath10k-ct-fix-missing-prototypes-warnings-on-6.12-ke.patch
DragonBluep: > Only partially merged upstream, as far as I can tell. Upstream fixed the two "unused function" cases by marking them `static` instead of commenting them out — [`static void ath10k_debug_fw_ratepwr_table_process()` at ath10k-7.2/debug.c:427](https://github.com/greearb/ath10k-ct/blob/fcbdb70debc261f9df6734bbb76d81cdc88e0e26/ath10k-7.2/debug.c#L427) and [`static void ath10k_debug_fw_powerctl_table_process()` at ath10k-7.2/debug.c:441](https://github.com/greearb/ath10k-ct/blob/fcbdb70debc261f9df6734bbb76d81cdc88e0e26/ath10k-7.2/debug.c#L441) — but neither has any caller left in `ath10k-7.2/`. So dropping this patch entirely trades the old `-Wmissing-prototypes` warnings for two new `-Wunused-function` warnings on every build (harmless with `# CONFIG_WERROR is not set`, but noisy).

Now these unused functions are guarded by `#if 0`, the same effect as`/* */`.

package/kernel/ath10k-ct/patches/300-fix-fortify-checking-error.patch
hauke, line 1: It looks like there is a NULL pointer derefernce in the code now:

### 1. Dropping `300-fix-fortify-checking-error.patch` reintroduces a NULL-pointer write

The commit message lists `300-fix-fortify-checking-error.patch` under "Upstream merged
patches". That is not what happened. OpenWrt's patch turned the trailing member into a
flexible array:

```c
-	__le32 ctl_info[1];
+	__le32 ctl_info[];
```

Upstream, in the new `ath10k-7.2` tree, made it a **pointer** instead
([wmi.h:6313](https://github.com/greearb/ath10k-ct/blob/fcbdb70debc261f9df6734bbb76d81cdc88e0e26/ath10k-7.2/wmi.h#L6313)):

```c
struct qca9880_set_ctl_table_cmd {
	__le32 ctl_len;
	__le32 *ctl_info;   /* was __le32 ctl_info[1] in ath10k-6.18 at the same commit */
};
```

The only consumer was not adjusted ([ath10k-7.2/wmi.c:7134](https://github.com/greearb/ath10k-ct/blob/fcbdb70debc261f9df6734bbb76d81cdc88e0e26/ath10k-7.2/wmi.c#L7134)):

```c
skb = ath10k_wmi_alloc_skb(ar, 4 + sizeof(struct qca9880_power_ctrl));
cmd = (struct qca9880_set_ctl_table_cmd *)skb->data;
cmd->ctl_len = __cpu_to_le32(len);
memcpy(cmd->ctl_info, data, len);          /* len == 432 */
```

`ath10k_wmi_alloc_skb()` zeroes the buffer (`memset(skb->data, 0, round_len)`, wmi.c:1896),
so `cmd->ctl_info` reads back as `NULL` and this is a 432-byte `memcpy` to `NULL` — a kernel
oops, not a truncated command. The fortify diagnostic the old patch existed to silence went
away because the destination became an unbounded pointer, not because the code was fixed.

`struct qca9880_set_ctl_table_cmd` does not exist in mainline ath10k (checked against
`v6.18:drivers/net/wireless/ath/ath10k/`), so this is a CT-side transcription slip when the
7.2 tree was created, and `ath10k-6.18/wmi.h` at the very same upstream commit still has
`__le32 ctl_info[1]`.

Reachability, so you can weigh it: QCA988X only, and only after an explicit debugfs write —
`ath10k_wmi_check_apply_board_power_ctl_table()` is gated on
`ar->eeprom_overrides.apply_board_power_ctl_table`, set by eeprom-override id `0x1005`
(debug.c:3737, commented "Use with care!") and re-applied on restart (core.c:4136). It is not
a boot path, so I am not calling it a merge blocker — but it is a regression against `main`,
where patch 300 keeps this correct today, and the fix is one line. Suggest keeping `300`
rebased onto 7.2 rather than dropping it, and reporting it upstream:

```diff
--- a/ath10k-7.2/wmi.h
+++ b/ath10k-7.2/wmi.h
@@
-	__le32 *ctl_info; /* data would be the qca9880_power_ctl table above */
+	__le32 ctl_info[]; /* data would be the qca9880_power_ctl table above */
```

Either way the commit message should stop describing `300` as merged upstream — only `005`
genuinely was.

DragonBluep: Thank you for your review. Fixed.

package/kernel/ath10k-ct/Makefile
```diff
@@ -8,14 +8,14 @@ PKG_LICENSE_FILES:=
 
 PKG_SOURCE_URL:=https://github.com/greearb/ath10k-ct.git
 PKG_SOURCE_PROTO:=git
-PKG_SOURCE_DATE:=2025-12-01
-PKG_SOURCE_VERSION:=bb84e1597474a6db26f7681d6765fd3e69fff37e
-PKG_MIRROR_HASH:=76a950c4c5f02128f4225955b1ca0d1d8f6c68c3b0918c4bc2d1265698e939a6
+PKG_SOURCE_DATE:=2026-07-31
```
openwrt-ai, line 11: nit: the commit message was corrected in this push, but the PR description still carries both stale facts and will end up quoted in changelogs/backport requests:

- "Update to Git HEAD (2026-08-01)" — the referenced commit `fcbdb70` is dated 2026-07-31, which is what `PKG_SOURCE_DATE` now says.
- `300-fix-fortify-checking-error.patch` is still listed under "Upstream merged patches", but this push re-adds it rebased onto `ath10k-7.2/`. Only `005-*` was actually dropped.

---

package/kernel/ath10k-ct/patches/300-fix-fortify-checking-error.patch
```diff
@@ -1,10 +1,10 @@
---- a/ath10k-6.18/wmi.h
-+++ b/ath10k-6.18/wmi.h
+--- a/ath10k-7.2/wmi.h
```
openwrt-ai, line 1: nit: the rebase looks right (`-\t__le32 *ctl_info;` matches the 7.2 tree, and the `@@ -6310,7 +6310,7 @@` context is unchanged from the 6.18 version of this patch), but the patch now does something different from what its name says. Against `ath10k-6.18` it turned `__le32 ctl_info[1]` into a flexible array to silence a fortify diagnostic; against `ath10k-7.2` it converts a *pointer* back into a flexible array, i.e. it prevents a 432-byte `memcpy()` into `NULL` — a real oops, not a warning.

This patch has no description header at all, which is presumably why it read as cosmetic and got dropped in the first push of this PR. Worth adding a short `Subject:`/body (and renaming, e.g. `300-ath10k-fix-null-deref-in-ctl-table-cmd.patch`) so the next rebase doesn't drop it again. Prose only — the exact wording/attribution is yours to pick.

---

openwrt-ai (commented):
Re-reviewed the changes since `16aade3` (3 hunks in `package/kernel/ath10k-ct/`).

All three previously-raised points are addressed: `PKG_SOURCE_DATE` is now `2026-07-31` with a correspondingly updated `PKG_MIRROR_HASH` (expected — the date feeds `PKG_VERSION` and therefore the tarball's internal directory name), the `202-*` diffstat header now matches its body (4+1+1 = 6 insertions, 1 deletion), and `300-fix-fortify-checking-error.patch` is restored rebased onto `ath10k-7.2/` with the `__le32 *ctl_info` → `__le32 ctl_info[]` conversion, which resolves the NULL-`memcpy()` regression. The commit message no longer claims `300` was merged upstream.

CI on `a751e38` is green — both `Build everything` jobs (x86/64 and malta/be) completed successfully, so all seven patches apply and the kmod builds.

Two nits inline, neither a merge blocker.

---

PR #24530, opened 2026-08-02
ramips: iptime,t5004: fix U-Boot environment size

devparlair:
The U-Boot environment on ipTIME T5004 occupies a 128 KiB NAND erase block.

The existing NVMEM layout incorrectly sets env-size to 0x1000.
As a result, the u-boot,env parser calculates the CRC over only the
first 4 KiB and rejects the environment with -EINVAL.

Set env-size to 0x20000 to match the actual erase-block size.
This allows the existing NVMEM MAC address cells to be created correctly.

Tested on a retail ipTIME T5004:
- Ethernet MAC addresses are correctly provided through NVMEM.
- WAN uses the expected MAC address offset.

### DIFF
--- a/target/linux/ramips/dts/mt7621_iptime_t5004.dts
+++ b/target/linux/ramips/dts/mt7621_iptime_t5004.dts
@@ -58,7 +58,7 @@
 
 			nvmem-layout {
 				compatible = "u-boot,env";
-				env-size = <0x1000>;
+				env-size = <0x20000>;
 
 				macaddr_uboot_ethaddr: ethaddr {
 					#nvmem-cell-cells = <1>;


target/linux/ramips/dts/mt7621_iptime_t5004.dts
```diff
@@ -55,15 +55,6 @@
 		partition@80000 {
 			label = "u-boot-env";
 			reg = <0x80000 0xc0000>;
-
-			nvmem-layout {
-				compatible = "u-boot,env";
-				env-size = <0x1000>;
```
openwrt-ai, line 61: Before reverting to userspace handling: did you confirm that the CRC mismatch is caused by a genuinely corrupt environment rather than by this `env-size`/format description being wrong? Two things here would produce exactly the symptom you describe on a perfectly valid environment:

- `env-size = <0x1000>` looks doubtful for a NAND-backed env. The partition is `0xc0000`, i.e. 6 × 128 KiB erase blocks, and U-Boot on NAND normally sets `CONFIG_ENV_SIZE` to the erase-block size. The only other in-tree ramips board using this layout, [`mt7621_h3c_tx180x.dtsi:99`](https://github.com/openwrt/openwrt/blob/a116d0d4b628e3234aa34ab2f9ea2c4f13557b8d/target/linux/ramips/dts/mt7621_h3c_tx180x.dtsi#L99), uses `env-size = <0x20000>`. Note also that mainline's parser ignores `env-size` entirely — `dev_size` comes from [`nvmem_dev_size()` in u-boot-env.c](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/nvmem/layouts/u-boot-env.c#L104) — and the property is only honoured through `target/linux/generic/backport-6.18/010-v6.19-nvmem-layouts-u-boot-env-add-optional-env-size-property.patch`, so this value was never validated against hardware when it was introduced.
- If the env is in the *redundant* format, `compatible = "u-boot,env"` is wrong: [the redundant image has an extra flags byte](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/nvmem/layouts/u-boot-env.c#L128-L131), so the CRC data offset shifts by one and the check at [u-boot-env.c:152](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/nvmem/layouts/u-boot-env.c#L152) fails on a valid env. `u-boot,env-redundant-bool` / `u-boot,env-redundant-count` are both used elsewhere in tree.

If either applies, fixing the layout would be preferable to reintroducing `mtd_get_mac_ascii`, which the original commit removed as deprecated. Sister ipTIME boards also take a third route — a binary `mac-base` cell at a fixed offset in the `u-boot` partition ([`mt7621_iptime_a3002mesh.dts:80`](https://github.com/openwrt/openwrt/blob/a116d0d4b628e3234aa34ab2f9ea2c4f13557b8d/target/linux/ramips/dts/mt7621_iptime_a3002mesh.dts#L80)) — which may also be an option on this device.

---

openwrt-ai (commented):
Reviewed 1 new commit. The diff is an exact inverse of 8dd95cc6b532 in both files, and the restored `iptime,t5004)` case is placed back in its original alphabetical position in `02_network`; commit message matches the change. One question inline about whether the root cause is the environment itself or the DTS layout description.

---

target/linux/ramips/dts/mt7621_iptime_t5004.dts
devparlair: Thanks for pointing this out.

I will re-check the actual U-Boot environment layout, including
environment size, erase block alignment, and possible redundant
environment format. The current NVMEM-based configuration causes CRC
validation failure during boot on the T5004, so I will investigate the
correct layout before deciding the final fix.

Thanks for the guidance.

target/linux/ramips/dts/mt7621_iptime_t5004.dts
```diff
@@ -58,7 +58,7 @@
 
 			nvmem-layout {
 				compatible = "u-boot,env";
-				env-size = <0x1000>;
+				env-size = <0x20000>;
```
openwrt-ai, line 61: nit: the PR title and body no longer describe this diff. They still say "restore userspace MAC address setup", claim the environment CRC is genuinely invalid, and quote a stored/calculated CRC pair computed over "the first 4 KiB" — i.e. over the old `env-size = <0x1000>` window, which is exactly the bug the new commit fixes. The single commit now does the opposite: it keeps the NVMEM path and corrects `env-size`. Worth rewriting the description before merge so the recorded rationale matches, otherwise anyone reading the PR later will conclude the environment on this board is corrupt.

The value itself looks right: `0x20000` matches the mt7621 NAND erase-block size and the only other in-tree ramips board with this layout, [`mt7621_h3c_tx180x.dtsi:99`](https://github.com/openwrt/openwrt/blob/6d15c8b8259ec8d1a967778afadde3be48069c26/target/linux/ramips/dts/mt7621_h3c_tx180x.dtsi#L99), and it stays well inside the `0xc0000` partition, so `crc32_data_len = dev_size - 4` in the backported parser reads in-bounds.

---

openwrt-ai, line 61: nit: the `Fixes:` trailer lives only in the PR body, not in the commit message — PR bodies are not carried into the merged history, so it will be lost. `env-size = <0x1000>` was introduced by [8dd95cc6b532 ("ramips: iptime,t5004: use nvmem for ubootenv")](https://github.com/openwrt/openwrt/commit/8dd95cc6b532af599a2ad0b92b3a53371832a3a4), so adding

```
Fixes: 8dd95cc6b532 ("ramips: iptime,t5004: use nvmem for ubootenv")
```

above the `Signed-off-by:` line would make the regression traceable and flag it for backporting.

While editing the message: "This prevents the u-boot,env parser from correctly reading the environment data" understates what happens — the parser fails the CRC32 check outright and returns `-EINVAL`, so *no* cells are created at all (`Invalid calculated CRC32: ...` in dmesg). Quoting that error line makes the symptom searchable for the next person hitting it.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push replaced the previous `restore userspace MAC address setup` commit). The new approach addresses the question raised on the earlier revision: instead of reintroducing `mtd_get_mac_ascii`, it corrects `env-size` to the erase-block size, keeping the NVMEM path. The commit subject and body match the one-line diff. Two message/metadata nits inline; no functional objection.

---

devparlair:
Thanks for the review.

Updated the commit as suggested:
- Added the Fixes trailer referencing 8dd95cc6b532.
- Clarified that the u-boot,env parser fails the CRC32 validation and returns -EINVAL.
- Fixed env-size from 0x1000 to 0x20000 to match the actual U-Boot environment storage size.

Tested on a retail ipTIME T5004:
- Ethernet MAC addresses are correctly provided through NVMEM.

target/linux/ramips/dts/mt7621_iptime_t5004.dts
openwrt-ai, line 61: nit: this one-line change is now delivered as an introduce-then-revert pair rather than a single commit. `a116d0d` removes the whole `nvmem-layout` block plus both `gmac*` `nvmem-cells` properties and re-adds the `iptime,t5004)` case to `02_network`; `0a8f831d` puts all of that back with `env-size` changed. `git diff a116d0d..0a8f831d` is the exact inverse of `git show a116d0d`, and the net series diff against the base is just this line.

The previous head, [6d15c8b](https://github.com/openwrt/openwrt/commit/6d15c8b8259ec8d1a967778afadde3be48069c26), already had exactly the right shape — a single commit touching only this line, with a tree identical to the current head (`git diff 6d15c8b..0a8f831d` is empty). Squashing back down to one commit (keeping the improved message from `0a8f831d`) would avoid landing a commit in history that reverts NVMEM support and then immediately restores it, and would keep `Fixes: 8dd95cc6b532` pointing at a single, bisectable fix.

---

openwrt-ai (commented):
Reviewed 2 new commits. The tree content is unchanged from the previously reviewed head `6d15c8b` (`git diff 6d15c8b..0a8f831d` is empty), and the PR body has been rewritten to match the actual fix — both nits from the last review are addressed. What changed is the branch history: the series is now two commits that cancel out, so both commit messages disagree with what their commits actually do.

## Commit checks

- `a116d0d` "ramips: iptime,t5004: restore userspace MAC address setup" — the message asserts a root cause that the very next commit in the same series contradicts: it claims the U-Boot environment "does not have a valid CRC according to the standard U-Boot environment format" and quotes a stored/calculated CRC pair computed over "the first 4 KiB". That 4 KiB window is the `env-size = <0x1000>` bug, not a property of the environment — `0a8f831d` states the environment is fine and the size description was wrong. The commit is also fully reverted by `0a8f831d`, so nothing of it survives in the series.
- `0a8f831d` "ramips: iptime,t5004: fix U-Boot environment size" — subject and body describe a single `env-size` correction, but the diff adds the entire `nvmem-layout` node, both `&gmac0`/`&gmac1` `nvmem-cells` properties, and deletes the `iptime,t5004)` case from `02_network` (17 insertions / 4 deletions across 2 files). Read on its own it looks like this commit introduces NVMEM MAC support rather than adjusting one property; it is really a revert of `a116d0d` with one value changed, and the message says nothing about that.

Squashing the two into the single one-line commit resolves both points at once. No functional objection to `env-size = <0x20000>`.

---

target/linux/ramips/dts/mt7621_iptime_t5004.dts
openwrt-ai, line 61: nit: the commit message describes `0x20000` as the partition size — "The U-Boot environment on the ipTIME T5004 occupies the entire 128 KiB partition" and "Set env-size to 0x20000 to match the actual U-Boot environment partition size". The `u-boot-env` partition in this same file is `0xc0000` (768 KiB), see [`reg = <0x80000 0xc0000>;` at mt7621_iptime_t5004.dts:57](https://github.com/openwrt/openwrt/blob/24cc52b52c8a8f8ce87e6db3e005dcb0f01778c0/target/linux/ramips/dts/mt7621_iptime_t5004.dts#L57), so `0x20000` is not the partition size — it is the environment data size U-Boot CRCs, which here equals one 128 KiB NAND erase block. The PR body already words this correctly ("occupies a 128 KiB NAND erase block"); the commit message is the one that disagrees. Worth aligning the message with the body before merge, since `env-size` is defined as the CRC data length, not the region size (see the backported binding text in `target/linux/generic/backport-6.18/010-v6.19-nvmem-layouts-u-boot-env-add-optional-env-size-property.patch`).

Value itself looks right and matches the sibling mt7621 NAND board [`env-size = <0x20000>;` at mt7621_h3c_tx180x.dtsi:99](https://github.com/openwrt/openwrt/blob/24cc52b52c8a8f8ce87e6db3e005dcb0f01778c0/target/linux/ramips/dts/mt7621_h3c_tx180x.dtsi#L99).

---

openwrt-ai (commented):
Reviewed the new head `24cc52b` (force-push squashed the previous two-commit series). The tree content is unchanged from the previously reviewed `0a8f831d` — the PR is now the single one-line `env-size` change, so both commit-message findings from the last review are resolved. `Fixes: 8dd95cc6b532` resolves to the correct commit ("ramips: iptime,t5004: use nvmem for ubootenv"), and CI on this head is green.

One factual nit inline about the commit message's description of `0x20000`; no functional objection.

---

openwrt-ai (commented):

The force-push only rewrote the commit message — `git diff 24cc52b..8683868` is empty, so the tree is still the single `env-size = <0x1000>` → `<0x20000>` change. The message now says the environment data "occupies one 128 KiB NAND erase block" and that `0x20000` matches "the U-Boot environment data size used for CRC validation", which resolves the last review's nit about `0x20000` being described as the partition size (the `u-boot-env` partition is `0xc0000`). `Fixes: 8dd95cc6b532` still resolves to "ramips: iptime,t5004: use nvmem for ubootenv", and CI on this head is green.

---

jonasjelonek:

devparlair:
Thanks for merging.
Could this be backported to openwrt-25.12 as well?
It fixes the U-Boot environment partition size and NVMEM MAC address reading on the ipTIME T5004.

jonasjelonek:
The switch from userspace handling to NVMEM hasn't been backported to 25.12 either, so this doesn't apply as-is.

devparlair:
Right, I missed that dependency. Thanks for pointing it out.

PR #24531, opened 2026-08-02
bcm53xx: fix sysupgrade config loss / mt7622: drop unused file_type variable

### LINKED ISSUE #21024: R7000 loses settings during firmware flash starting with openwrt-24.10.0-rc1 including current snapshots
### Describe the bug

I have an R7000 that I use for a wired network.  It seems something changed in openwrt-24.10.0-rc1 so that using it to flash firmware always loses the settings.  I tried different versions to narrow down when this broke.  I can use openwrt-23.05.6 to flash any version of firmware and it will keep the settings correctly.  Using openwrt-24.10.0-rc1, or anything later, to flash any version will always lose the settings.  I looked at the logs with 24.10.0-rc1 after flashing from 23.05.6, and after flashing from 24.10.0-rc1 and the differences I see are:

Flashing with 23.05.6:
```
Tue Dec  2 16:12:29 2025 kern.notice kernel: [    4.428379] ubi0: max/mean erase counter: 3/1, WL threshold: 4096, image sequence number: 87069445
```

Flashing with 24.10.0-rc1:
```
Fri Nov 29 17:25:40 2024 kern.notice kernel: [    4.296823] ubi0: volume 1 ("rootfs_data") re-sized from 9 to 178 LEBs
Fri Nov 29 17:25:40 2024 kern.notice kernel: [    4.343460] ubi0: max/mean erase counter: 1/0, WL threshold: 4096, image sequence number: 1732901136
Fri Nov 29 17:25:40 2024 kern.notice kernel: [   11.132575] UBIFS (ubi0:1): default file-system created
```

### OpenWrt version

r28158-d276b4c91a

### OpenWrt release

24.10.0-rc1

### OpenWrt target/subtarget

bcm53xx/generic

### Device

Netgear R7000

### Image kind

Official downloaded image

### Steps to reproduce

On an R7000, with firmware 24.10.0-rc1 or later installed, flash another OpenWrt firmware and tell it to keep the settings.  After reboot, the settings will be gone.

### Actual behaviour

The settings are not retained.

### Expected behaviour

The settings should be retained, if the option is selected to keep them.

### Additional info

As a workaround, I have been restoring a backup configuration after each firmware flash.  Using attended sysupgrade or owut has the same problem.

### Diffconfig

```text

```

### Terms

strrs:
Since 24.10.0-rc1 a bcm53xx NAND device always comes back at factory defaults after a flash, as if `-n` had been passed.

The UBI check fails and sysupgrade falls back to a raw whole image write, which cannot carry `sysupgrade.tgz` into the new `rootfs_data`.

Affects bcm53xx NAND devices flashed with a trx based image.

The fix lives in the firmware doing the flashing, so upgrading from an affected build loses the config one last time. Every upgrade from a fixed build onwards keeps it.

Build tested on main with kernel 6.18, Asus RT-N18U, three consecutive sysupgrades, config preserved every time.

The second commit removes a dead file_type assignment in mt7622, the only other caller still on the old identify argument order after 715634e6d144. It is included here because both commits clean up callers left stale by the same change; the review bot requested it during review.

The first commit can be backported to openwrt-24.10 and openwrt-25.12.

Fixes: https://github.com/openwrt/openwrt/issues/21024
Fixes: https://github.com/openwrt/openwrt/issues/21655

### DIFF
--- a/target/linux/bcm53xx/base-files/lib/upgrade/platform.sh
+++ b/target/linux/bcm53xx/base-files/lib/upgrade/platform.sh
@@ -237,7 +237,7 @@ platform_do_upgrade_nand_trx() {
 	}
 
 	# Firmwares without UBI image should be flashed "normally"
-	local root_type=$(identify $dir/root)
+	local root_type=$(identify "$dir/root" "cat")
 	[ "$root_type" != "ubi" ] && {
 		echo "Provided firmware doesn't use UBI for rootfs."
 		return
--- a/target/linux/mediatek/mt7622/base-files/lib/upgrade/platform.sh
+++ b/target/linux/mediatek/mt7622/base-files/lib/upgrade/platform.sh
@@ -3,7 +3,6 @@ RAMFS_COPY_BIN='fitblk fit_check_sign'
 
 platform_do_upgrade() {
 	local board=$(board_name)
-	local file_type=$(identify $1)
 
 	case "$board" in
 	bananapi,bpi-r64|\


strrs:
@neheb

neheb (approved):

target/linux/bcm53xx/base-files/lib/upgrade/platform.sh
```diff
@@ -237,7 +237,7 @@ platform_do_upgrade_nand_trx() {
 	}
 
 	# Firmwares without UBI image should be flashed "normally"
-	local root_type=$(identify $dir/root)
+	local root_type=$(identify "$dir/root" "cat")
```
openwrt-ai, line 240: nit: `715634e6d144` updated every caller inside `nand.sh` to the new `(file, cmd, skip)` signature but missed the two callers outside it. This PR fixes the bcm53xx one; the other stale one-arg call is [`local file_type=$(identify $1)` at mediatek/mt7622 platform.sh:6](https://github.com/openwrt/openwrt/blob/997291d9fe3823d1e2d182b44367b26a7ab0c2cf/target/linux/mediatek/mt7622/base-files/lib/upgrade/platform.sh#L6).

It is currently harmless there — `file_type` is assigned and never read in that function — so it is not a regression in the same sense, but it will silently return `unknown` if anyone ever starts using it. Worth a separate cleanup patch rather than mixing it into this fix.

---

openwrt-ai (commented):

Verified the root cause independently: `715634e6d144` changed `nand_get_magic_long()` from `(${3}cat "$1" | ...)` to `($2 < "$1" | ...)`, moving the extraction command from `$3` to `$2` and `skip` from `$2` to `$3`. With a single argument the old form degraded to a plain `cat "$1"` and worked; the new form runs an empty command with a redirection, so `dd` reads nothing, the magic comes back empty and `identify_magic_long` returns `unknown`. Passing `"cat"` restores the pre-regression behaviour exactly, and `skip` correctly stays at its `${3:-0}` default.

`"cat"` (rather than `$(identify_if_gzip ...)cat`) is the right choice here since `otrx extract` writes the raw TRX partition, which is never gzipped. Commit subject scope (`bcm53xx:`) matches the single file touched, and the `Fixes:` trailer points at the correct commit.

---

AuthorReflex:
I can confirm it works on the Asus RT-AC68U. Nice work!

openwrt-ai (commented):

`f8047a93` ("mediatek: mt7622: drop unused file_type variable") checks out on both claims in its message:

- **Unused:** `file_type` was the only occurrence in `target/linux/mediatek/mt7622/base-files/lib/upgrade/platform.sh` — assigned at the top of `platform_do_upgrade()` and never read by any branch of the `case`.
- **Wrong argument order:** with a single argument, [`nand_get_magic_long()`](https://github.com/openwrt/openwrt/blob/f8047a936c6e4d4c60d6eb6efe0c5c09237ebe95/package/base-files/files/lib/upgrade/nand.sh#L60-L62) expands to `( < "$1" | dd ... )` with an empty command, so the magic came back empty and `identify` returned `unknown` — the same regression the bcm53xx commit fixes. Removing the dead assignment is a valid resolution and has no side effects, so "No functional change" is accurate.

Cross-checked that these were the only two `identify` call sites outside `nand.sh` (`grep -rn 'identify ' target/ package/`), so the pair of commits leaves no other broken caller behind. The in-tree callers inside `nand.sh` ([lines 407-408](https://github.com/openwrt/openwrt/blob/f8047a936c6e4d4c60d6eb6efe0c5c09237ebe95/package/base-files/files/lib/upgrade/nand.sh#L407-L408) and [494-495](https://github.com/openwrt/openwrt/blob/f8047a936c6e4d4c60d6eb6efe0c5c09237ebe95/package/base-files/files/lib/upgrade/nand.sh#L494-L495)) already pass the post-`715634e6d144` argument order.

Subject scope (`mediatek: mt7622:`) matches the single file touched, and `715634e6d1443eacdcb84b04d1028c1564b08dbf` is the correct SHA for the referenced commit.

---

AuthorReflex:
> The fix lives in the firmware doing the flashing, so upgrading from an affected build loses the config one last time. Every upgrade from a fixed build onwards keeps it.

To prevent the loss of the configuration, need to manually copy the file [target/linux/bcm53xx/base-files/lib/upgrade/platform.sh]  to the target device to [/lib/upgrade/platform.sh] before update and run the update procedure; the settings will then be preserved.

strrs:
@AuthorReflex Yes, and it does not need a rebuild. I ran the hardware tests that way.

One line does it on an affected build:

`sed -i 's|identify $dir/root|identify "$dir/root" "cat"|' /lib/upgrade/platform.sh`

Taking a config backup first and restoring after flashing works too.

Note: The USB 3.0 port on that board is also broken since 24.10. Fix in #24547.

jonasjelonek:
- Why does this PR just talk about bcm53xx but also carry a mediatek commit/patch?
- Why does the mediatek commit carry the link to this PR? It will carry after merge, yes, but right now this is kind of Inception style.

strrs:
@jonasjelonek  The review bot asked for the mt7622 cleanup in an inline comment, so I added it here. I went through every remaining identify caller after 715634e6d144 swapped the argument order; mt7622 is the only other stale one. There the result is never read today, but whoever starts using it gets "unknown" instead of the real type.

I dropped the Link footer. If you want, I will drop the mediatek commit from this PR.

jonasjelonek:
You can keep it if you update PR title and description properly to cover both and also mention why the mt7622 is included. With that, it's fine for me.

strrs:
@jonasjelonek Thank you. Done.

openwrt-ai (commented):

`bd6151a8` is a message-only amendment of the previously reviewed `f8047a93` — `git diff f8047a93..bd6151a8` is empty, so both files are byte-identical to what I reviewed before. The only change is dropping the self-referential `Link: .../pull/24531` footer, as asked for in [this comment](https://github.com/openwrt/openwrt/pull/24531#issuecomment-5263844437); the `Signed-off-by` trailer is intact and the subject scope (`mediatek: mt7622:`) still matches the single file touched.

The PR title and body have also been updated to cover both commits and to explain why the mt7622 cleanup rides along, which addresses the other half of that review comment.

Earlier findings still stand and need no repeating: `identify "$dir/root" "cat"` restores the pre-`715634e6d144` behaviour of [`nand_get_magic_long()`](https://github.com/openwrt/openwrt/blob/bd6151a8a9a3ff948eaa0b831c51d64a2e252b4d/package/base-files/files/lib/upgrade/nand.sh#L60-L62) with `skip` correctly falling back to its `${3:-0}` default, and the mt7622 hunk removes a genuinely dead assignment. No CI checks have reported on this head yet.

---

jonasjelonek:

PR #9810, opened 2022-04-30
ramips: zbt-wg2626: Add the reset gpio for PCIe port 1

AlbanBedel:
The 2.4GHz interface doesn't come up properly with the log showing:

    mt7621-pci 1e140000.pcie: pcie1 no card, disable it (RST & CLK)

As seen on other MT7621 boards this is caused by a missing reset GPIO.
The MT7621 dtsi set GPIO 19 as PCIe reset GPIO, which on this board
reset the 5GHz interface on port 0. Add GPIO 8 to the PCIe reset GPIO
list to also reset the 2.4GHz interface on port 1.

Signed-off-by: Alban Bedel <albeu@free.fr>

PR #24534, opened 2026-08-02
ramips: fix PCIe reset GPIOs for ipTIME A8004T

devparlair:
This fixes PCIe initialization failure after warm reboot on the ipTIME A8004T.

The issue has existed for this device for several years and was
reported as early as 2019. The problem was reproduced on a retail
ipTIME A8004T:

- Cold boot: PCIe devices initialize normally.
- Warm reboot: PCIe devices fail to initialize.

Related PCIe initialization issues have also been reported for other
MediaTek devices using mt76:

- https://github.com/openwrt/mt76/issues/421
- https://github.com/openwrt/mt76/issues/316

Those cases indicate that missing PCIe reset GPIO handling can prevent
wireless PCIe devices from being initialized correctly after reboot.

The same approach has already been applied successfully to other
MediaTek devices, such as the ZBT WG2626:

- https://github.com/openwrt/openwrt/commit/f953a1a4bfba2fa70c12bb80938aa66481a673b6
- https://github.com/openwrt/openwrt/pull/9810

The ipTIME A8004T requires two GPIO lines for PCIe reset:
- GPIO19
- GPIO8

GPIO8 is shared with UART3 and is configured for that function by
default, so it cannot be used as a GPIO until the pin state is
adjusted.

This patch follows the existing PCIe reset GPIO handling approach and
applies it to the ipTIME A8004T by adding the required reset GPIO
definitions and adjusting the pinmux state.

After this change, PCIe initialization succeeds after both cold boot
and warm reboot.

The fix has been tested successfully on a retail ipTIME A8004T.

### DIFF
--- a/target/linux/ramips/dts/mt7621_iptime_a8004t.dts
+++ b/target/linux/ramips/dts/mt7621_iptime_a8004t.dts
@@ -168,6 +168,9 @@
 
 &pcie {
 	status = "okay";
+
+	reset-gpios = <&gpio 19 GPIO_ACTIVE_LOW>,
+			<&gpio 8 GPIO_ACTIVE_LOW>;
 };
 
 &pcie0 {
@@ -192,7 +195,7 @@
 
 &state_default {
 	gpio {
-		groups = "wdt", "jtag", "i2c";
+		groups = "wdt", "jtag", "i2c", "uart3";
 		function = "gpio";
 	};
 };


DragonBluep (approved):

target/linux/ramips/dts/mt7621_iptime_a8004t.dts
```diff
@@ -168,6 +168,8 @@
 
 &pcie {
 	status = "okay";
+
+	reset-gpios = <&gpio 19 GPIO_ACTIVE_LOW>, <&gpio 8 GPIO_ACTIVE_LOW>;
```
openwrt-ai: nit: every other mt7621 board in-tree with a multi-entry PCIe `reset-gpios` list puts one phandle per line, aligned under the first one — see [`mt7621_dlink_dir-860l-b1.dts:178-179`](https://github.com/openwrt/openwrt/blob/5faa918bd1cdcfdc0da3cffadb92bd79691741e5/target/linux/ramips/dts/mt7621_dlink_dir-860l-b1.dts#L178-L179), and likewise `mt7621_zbtlink_zbt-wg2626.dts:114`, `mt7621_netgear_ex6150.dts:232`, `mt7621_mtc_wr1201.dts:179`, `mt7621_d-team_pbr-m1.dts:169`. Worth matching for consistency.

```suggestion
	reset-gpios = <&gpio 19 GPIO_ACTIVE_LOW>,
		      <&gpio 8 GPIO_ACTIVE_LOW>;
```

---

openwrt-ai (commented):
Reviewed 1 new commit. The change looks functionally correct:

- `reset-gpios` index maps to the PCI slot number in the driver ([`mt7621_pcie_parse_port()` requests `devm_gpiod_get_index_optional(dev, "reset", slot, ...)`](https://github.com/gregkh/linux/blob/HEAD/drivers/pci/controller/pcie-mt7621.c#L236)), so `<&gpio 19>` → `pcie0` and `<&gpio 8>` → `pcie1` is the right ordering, matching `mt7621_dlink_dir-860l-b1.dts` / `mt7621_zbtlink_zbt-wg2626.dts`.
- The `uart3` pinmux group covers pins 5–8 (`FUNC("uart3", 0, 5, 4)` in `pinctrl-mt7621.c`) and is a `GRP_G` group, so muxing it to `function = "gpio"` does make GPIO8 usable. No `&uartlite3` node is enabled on this board, so nothing is lost.
- GPIO19 is already muxed to GPIO by `pcie_pins` in `mt7621.dtsi`, and re-listing it in the board `reset-gpios` is required since DT list properties can't be appended to.

One formatting nit inline.

---

target/linux/ramips/dts/mt7621_iptime_a8004t.dts
devparlair: Updated the `reset-gpios` list to use one phandle per line, matching the existing MT7621 DTS style. Thanks.

jonasjelonek:

devparlair:
Thanks for merging!
Could this commit also be backported to 25.12?
The A8004T has the same PCIe warm reboot issue on that branch.

jonasjelonek:
Applied to openwrt-25.12 in https://github.com/openwrt/openwrt/commit/d69c2a7bea268e479cda72f390ffc3df2c078006

PR #24535, opened 2026-08-02
qualcommax: qca_ppe: take both XLT indices before programming a VLAN

JuliusBairaktaris:
`qca_ppe_port_vlan_add()` allocates two XLT indices. The first failure is rolled
back - the port is dropped from the entry and the entry freed if that emptied it.
The second, for the pvid rule, returned `-ENOSPC` having already written
`PPE_PORT_DEF_CVID` for the port, set `priv->port_pvid[port]` and added the port
to `entry->pvid_ports`.

DSA does not call `port_vlan_del()` after a failed `port_vlan_add()`, so none of
that was undone: the port kept a pvid in hardware that the driver's own state said
it did not have, and the next `ppe_vlan_pvid_update()` acted on a `pvid_ports` mask
including a port that was never programmed.

Take both indices before anything is programmed. Both failures then share the
rollback that already exists, and there is no half-written state to undo.

Not reachable today: `PPE_VSI_MAX * 2 == PPE_XLT_TBL_NUM`, so any arrangement that
exhausts the XLT pool has already given every entry its pvid index. That is an
accident of two constants rather than an invariant - raising `PPE_VSI_MAX` without
raising `PPE_XLT_TBL_NUM` opens it.

### DIFF
--- a/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_vlan.c
+++ b/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_vlan.c
@@ -231,18 +231,21 @@ int qca_ppe_port_vlan_add(struct dsa_switch *ds, int port,
 			return -ENOSPC;
 	}
 
-	entry->ports |= BIT(port);
-
 	if (entry->xlt_idx < 0) {
 		idx = ppe_xlt_idx_alloc(priv);
-		if (idx < 0) {
-			entry->ports &= ~BIT(port);
-			if (!entry->ports)
-				ppe_vlan_free(priv, entry);
-			return -ENOSPC;
-		}
+		if (idx < 0)
+			goto err_free_entry;
 		entry->xlt_idx = idx;
 	}
+
+	if (pvid && entry->xlt_pvid_idx < 0) {
+		idx = ppe_xlt_idx_alloc(priv);
+		if (idx < 0)
+			goto err_free_entry;
+		entry->xlt_pvid_idx = idx;
+	}
+
+	entry->ports |= BIT(port);
 
 	ppe_xlt_rule_set(priv, entry->xlt_idx,
 			 entry->ports | BIT(QCA_PPE_CPU_PORT), vid, false);
@@ -256,12 +259,6 @@ int qca_ppe_port_vlan_add(struct dsa_switch *ds, int port,
 		priv->port_pvid[port] = vid;
 		entry->pvid_ports |= BIT(port);
 
-		if (entry->xlt_pvid_idx < 0) {
-			idx = ppe_xlt_idx_alloc(priv);
-			if (idx < 0)
-				return -ENOSPC;
-			entry->xlt_pvid_idx = idx;
-		}
 		ppe_xlt_rule_set(priv, entry->xlt_pvid_idx,
 				 entry->pvid_ports, 0, true);
 		ppe_xlt_action_set(priv, entry->xlt_pvid_idx, entry->vsi,
@@ -277,6 +274,11 @@ int qca_ppe_port_vlan_add(struct dsa_switch *ds, int port,
 	ppe_vlan_members_update(priv, entry);
 
 	return 0;
+
+err_free_entry:
+	if (!entry->ports)
+		ppe_vlan_free(priv, entry);
+	return -ENOSPC;
 }
 
 int qca_ppe_port_vlan_del(struct dsa_switch *ds, int port,


target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_vlan.c
```diff
@@ -258,8 +258,19 @@ int qca_ppe_port_vlan_add(struct dsa_switch *ds, int port,
...
+				 * DSA does not call port_vlan_del() after a
+				 * failed port_vlan_add(), so undo what this
+				 * branch programmed. The membership written
+				 * above is left in place and is consistent on
+				 * its own: only the pvid half failed.
+				 */
+				ppe_port_def_cvid_set(priv, port, 0, false);
+				priv->port_pvid[port] = 0;
+				entry->pvid_ports &= ~BIT(port);
 				return -ENOSPC;
```
openwrt-ai: The membership half is not actually consistent on this path, which is the same kind of software/hardware divergence the commit sets out to remove. [`entry->ports |= BIT(port)` (L234)](``https://github.com/openwrt/openwrt/blob/8c2a6ac837248dce8267b1009ce52cd87cf405b6/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_vlan.c#L234)`` and the [XLT rule/action writes (L247-L249)](``https://github.com/openwrt/openwrt/blob/8c2a6ac837248dce8267b1009ce52cd87cf405b6/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_vlan.c#L247-L249)`` have already run, but the VSI member mask is only written by [`ppe_vlan_members_update()` at L288](``https://github.com/openwrt/openwrt/blob/8c2a6ac837248dce8267b1009ce52cd87cf405b6/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_vlan.c#L288),`` which this early `return` skips. Result: the XLT rule classifies the port's tagged frames into the VSI while `PPE_VSI_TBL(vsi)` MEMBER/UUC/UMC/BC still exclude that port (a fresh entry starts at 0 via `ppe_vsi_alloc()`), and `entry->ports` claims it is a member — until some unrelated add/del on the same entry happens to rewrite the mask.

If the intent is the one stated in the commit message ("the port really is a member, only the pvid half failed"), program that membership before returning the error:

```suggestion
				ppe_port_def_cvid_set(priv, port, 0, false);
				priv->port_pvid[port] = 0;
				entry->pvid_ports &= ~BIT(port);
				ppe_vlan_members_update(priv, entry);
				return -ENOSPC;
```

The last sentence of the comment above would need a small tweak to match. (The alternative — also rolling the membership back, mirroring the first alloc-failure branch — is defensible but contradicts the design the commit message argues for.)

---

target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_vlan.c
```diff
@@ -258,8 +258,19 @@ int qca_ppe_port_vlan_add(struct dsa_switch *ds, int port,
...
+			if (idx < 0) {
+				/*
+				 * DSA does not call port_vlan_del() after a
+				 * failed port_vlan_add(), so undo what this
+				 * branch programmed. The membership written
+				 * above is left in place and is consistent on
+				 * its own: only the pvid half failed.
+				 */
+				ppe_port_def_cvid_set(priv, port, 0, false);
+				priv->port_pvid[port] = 0;
```
openwrt-ai: This restores 0, not the pvid the port had before the call, so it is not the exact undo the commit message describes when the port already had one. [L256](``https://github.com/openwrt/openwrt/blob/8c2a6ac837248dce8267b1009ce52cd87cf405b6/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_vlan.c#L256)`` overwrites `priv->port_pvid[port]` with the new `vid` without saving the previous value — and moving a pvid from VLAN A to VLAN B on the same port is exactly a case the bridge reports as a single `port_vlan_add(B, PVID)`: [`__vlan_add_pvid()` in br_vlan.c:37-46](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/net/bridge/br_vlan.c#L37-L46) just moves `vg->pvid`, and no del is issued for A (v6.12, the target's `KERNEL_PATCHVER`). In that case the rollback disables `PPE_PORT_DEF_CVID` and zeroes `port_pvid[port]`, leaving the port with no pvid at all, whereas before the failed call it had a working one (VLAN A's pvid XLT rule and `pvid_ports` bit are still in place, so restoring the default CVID would be self-consistent).

Saving the old value before L256 and restoring it here (`ppe_port_def_cvid_set(priv, port, old_pvid, old_pvid != 0)`) would make the undo exact. Was leaving the port pvid-less deliberate?

---

openwrt-ai (commented):
Reviewed 1 new commit (8c2a6ac). The DSA premise in the commit message checks out against v6.12 (`dsa_port_do_vlan_add()` returns the driver error without any `port_vlan_del()`, and `__vlan_add()` goes straight to `out` when `__vlan_vid_add()` fails). Two comments on the rollback path itself.

---

target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_vlan.c
```diff
@@ -277,6 +274,12 @@ int qca_ppe_port_vlan_add(struct dsa_switch *ds, int port,
 	ppe_vlan_members_update(priv, entry);
 
 	return 0;
+
+err_drop_port:
+	entry->ports &= ~BIT(port);
+	if (!entry->ports)
+		ppe_vlan_free(priv, entry);
+	return -ENOSPC;
```
openwrt-ai: Sharing this path with the new pvid allocation makes it reachable in a case it was never written for: when the port was **already** a member of `entry` before this call. `entry->ports &= ~BIT(port)` then undoes state this call did not create.

The first allocation could only fail on a fresh entry (`entry->xlt_idx < 0` implies the entry came from `ppe_vlan_alloc()` — every non-empty entry has an `xlt_idx`, and `ppe_vlan_free()` is only reached at `ports == 0`), so `entry->ports == BIT(port)` was guaranteed and the unconditional clear was safe. The pvid allocation has no such guarantee: `entry->xlt_pvid_idx` is `-1` for any VLAN that simply has no pvid port yet, including one with several tagged members.

That path is reachable — the bridge re-notifies an existing VLAN on a flag change, and DSA passes it straight to the driver for user ports: [`nbp_vlan_add()` br_vlan.c:1320-1329](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/net/bridge/br_vlan.c#L1320-L1329) calls `br_switchdev_port_vlan_add()` when the flags would change, and [`dsa_port_do_vlan_add()` switch.c:697-702](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/net/dsa/switch.c#L697-L702) skips refcounting for user ports. So `bridge vlan add dev lanN vid 10 pvid untagged` on a port already carrying tagged VLAN 10 lands here with `entry->ports` already holding the bit.

Two outcomes, both the divergence this commit sets out to remove:

- other members present: `entry->ports` loses a port that the XLT rule at [L250-L251](``https://github.com/openwrt/openwrt/blob/d57259dba2a05fbaee772761577ac8c704dbef0a/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_vlan.c#L250-L251)`` and the VSI member mask still list, from an earlier successful call.
- sole member: `ppe_vlan_free()` tears down a working VLAN — XLT indices and VSI released, `br_dev` cleared — while `ppe_eg_vsi_tag_port_set()` for that port is left behind, and the later `port_vlan_del()` finds no entry and returns 0.

Recording whether this call added the bit keeps the path exact:

```c
	bool new_member = !(entry->ports & BIT(port));
	...
err_drop_port:
	if (new_member) {
		entry->ports &= ~BIT(port);
		if (!entry->ports)
			ppe_vlan_free(priv, entry);
	}
	return -ENOSPC;
```

Alternatively, move `entry->ports |= BIT(port)` at [L234](``https://github.com/openwrt/openwrt/blob/d57259dba2a05fbaee772761577ac8c704dbef0a/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_vlan.c#L234)`` below both allocations — nothing between the two reads it — which fits the commit's own "nothing is touched until every resource is in hand" premise and reduces the label to `if (!entry->ports) ppe_vlan_free(...)`.

Same unreachability caveat as the bug being fixed applies (`PPE_VSI_MAX * 2 == PPE_XLT_TBL_NUM`), but if the point is to make the failure path correct for a future `PPE_VSI_MAX` bump, this half should be correct too.

---

openwrt-ai (commented):
Reviewed 1 new commit (d57259d). The rewrite addresses both earlier comments — allocating up front removes the half-written pvid state and the stale VSI member mask, and the commit message's claims check out (`PPE_VSI_MAX` 32 / `PPE_XLT_TBL_NUM` 64 in `qca_ppe.h`; `Fixes:` sha resolves; the DSA/bridge reasoning matches v6.18, this target's `KERNEL_PATCHVER`). One comment on the now-shared rollback label. CI is green on this head.

---

JuliusBairaktaris:
@robimarko reachable with PPE offloading so should go in before that

openwrt-ai (commented):
Reviewed 1 new commit (44cd72a). The rewrite since d57259d resolves the earlier comment on the shared rollback label: `entry->ports |= BIT(port)` now happens after both XLT allocations, so `err_free_entry` no longer has to un-set a bit it may not have set, and it frees only an entry this call allocated (an existing entry always has `ports != 0`, since `port_vlan_del()` frees an emptied one). No half-written state remains on either allocation failure, and `ppe_vlan_free()` releases a first-allocation `xlt_idx` on the fresh-entry path. Commit-message claims (constants, `Fixes:` sha, DSA/bridge reasoning) still hold from the prior review, and CI is green on this head. No new issues found.

---

robimarko:

JuliusBairaktaris:
Thank you very much for your effort!

PR #24538, opened 2026-08-02
realtek: eth: adapt interrupt handling for fragments

plappermaul:
The existing interrupt handling of the ethernet driver worked well until the driver implemented receive fragment handling. This change uncovered a constellation that did not exist before.

A received packet is written into the page pool with multiple fragments. On RTL93xx the head-of-line (HOL) feature limits the number of receivable fragments perfectly. There will never be a "buffer full" situation where the hardware only encounters ring buffers that are held by the Linux kernel. On RTL83xx this is slightly different:

- The driver programs free floating rings
- Only the ownership flag of the ring buffer decides if the hardware can hand over a packet to the CPU.
- So the hardware can receive a packet even if it does not completely fit into the available fragments.

With this there is a small chance that

- The buffer has less space than a just received packet
- The hardware generates an overflow (RUN OUT) interrupt
- With no completely received packet the hardware DOES NOT generate a receive (DONE) interrupt.

So it is not sufficient to just look on the DONE interrupts. The RUN OUT interrupts must be inspected as well. As the current logic is quite cryptic enhance this as follows:

- Provide new RTL83xx/RTL93xx specific helpers
- Add new callbacks to the driver configruation structure
- Link the configuration with the new helpers
- Use the callbacks where needed.

### DIFF
--- a/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
+++ b/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
@@ -205,44 +205,68 @@ static int rteth_free_skb(struct sk_buff **skb)
 	return 0;
 }
 
-static inline void rteth_reenable_irq(struct rteth_ctrl *ctrl, int ring)
+static void rteth_83xx_enable_rx_irq(struct rteth_ctrl *ctrl, int ring)
 {
-	u32 shift = ctrl->r->rx_rings % 32;
-	u32 reg = ctrl->r->rx_rings / 32;
-	u32 bit = BIT(ring + shift);
 	unsigned long flags;
 
-	/* locking needed for synchronization with rteth_confirm_and_disable_irqs() */
 	spin_lock_irqsave(&ctrl->lock, flags);
-	regmap_update_bits(ctrl->map, ctrl->r->dma_if_intr_msk + reg * 4, bit, bit);
+	regmap_set_bits(ctrl->map, ctrl->r->dma_if_intr_msk, BIT(ring) | BIT(ring + 8));
 	spin_unlock_irqrestore(&ctrl->lock, flags);
 }
 
-static inline void rteth_confirm_and_disable_irqs(struct rteth_ctrl *ctrl,
-						  unsigned long *rings, bool *l2)
+static void rteth_93xx_enable_rx_irq(struct rteth_ctrl *ctrl, int ring)
 {
-	u32 mask = GENMASK(ctrl->r->rx_rings - 1, 0);
-	u32 shift = ctrl->r->rx_rings % 32;
-	u32 reg = ctrl->r->rx_rings / 32;
 	unsigned long flags;
-	u32 active;
 
-	/* get all irqs, disable only rx (on RTL839x this keeps L2), confirm all */
 	spin_lock_irqsave(&ctrl->lock, flags);
-	regmap_read(ctrl->map, ctrl->r->dma_if_intr_sts + reg * 4, &active);
-	regmap_update_bits(ctrl->map, ctrl->r->dma_if_intr_msk + reg * 4,
-			   active & (mask << shift), 0);
-	regmap_write(ctrl->map, ctrl->r->dma_if_intr_sts + reg * 4, active);
+	regmap_set_bits(ctrl->map, ctrl->r->dma_if_intr_msk, BIT(ring));
+	regmap_set_bits(ctrl->map, ctrl->r->dma_if_intr_msk + 4, BIT(ring));
 	spin_unlock_irqrestore(&ctrl->lock, flags);
+}
 
-	/* ~mask filters out RTL93xx devices */
-	*l2 = !!(active & ~mask & RTL839X_DMA_IF_INTR_NOTIFY_MASK);
-	*rings = (active >> shift) & mask;
+static void rteth_83xx_confirm_and_disable_irqs(struct rteth_ctrl *ctrl,
+						unsigned long *rings, bool *l2)
+{
+	unsigned long flags;
+	u32 disable, state;
+
+	spin_lock_irqsave(&ctrl->lock, flags);
+
+	regmap_read(ctrl->map, ctrl->r->dma_if_intr_sts, &state);
+	*rings = FIELD_GET(GENMASK(7, 0), state) | FIELD_GET(GENMASK(15, 8), state);
+	*l2 = !!(state & RTL839X_DMA_IF_INTR_NOTIFY_MASK);
+	disable = FIELD_PREP(GENMASK(7, 0), *rings) | FIELD_PREP(GENMASK(15, 8), *rings);
+
+	regmap_clear_bits(ctrl->map, ctrl->r->dma_if_intr_msk, disable);
+	regmap_write(ctrl->map, ctrl->r->dma_if_intr_sts, state);
+
+	spin_unlock_irqrestore(&ctrl->lock, flags);
+}
+
+static void rteth_93xx_confirm_and_disable_irqs(struct rteth_ctrl *ctrl,
+						unsigned long *rings, bool *l2)
+{
+	u32 state_done, state_runout;
+	unsigned long flags;
+
+	spin_lock_irqsave(&ctrl->lock, flags);
+
+	regmap_read(ctrl->map, ctrl->r->dma_if_intr_sts, &state_runout);
+	regmap_read(ctrl->map, ctrl->r->dma_if_intr_sts + 4, &state_done);
+	*rings = state_runout | state_done;
+	*l2 = false;
+
+	regmap_clear_bits(ctrl->map, ctrl->r->dma_if_intr_msk, *rings);
+	regmap_clear_bits(ctrl->map, ctrl->r->dma_if_intr_msk + 4, *rings);
+	regmap_write(ctrl->map, ctrl->r->dma_if_intr_sts, state_runout);
+	regmap_write(ctrl->map, ctrl->r->dma_if_intr_sts + 4, state_done);
+
+	spin_unlock_irqrestore(&ctrl->lock, flags);
 }
 
 static void rteth_disable_all_irqs(struct rteth_ctrl *ctrl)
 {
-	int registers = ((ctrl->r->rx_rings * 2 + 7) / 32) + 1;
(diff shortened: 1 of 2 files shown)


jonasjelonek (approved):
Without having really an idea about that functionally, looks good to me code-wise.

l0rdg3x:
Tested on RTL9303 (Hasivo S1100W-8XGT-SE), RAM-boot, kernel 6.18.41: `main` 712bfddc81 against the
same tree with this commit cherry-picked. RTL930x regresses badly.

`rteth_93xx_confirm_and_disable_irqs()` acknowledges both status registers with the same value:

```c
regmap_write(ctrl->map, ctrl->r->dma_if_intr_sts, state_runout);
regmap_write(ctrl->map, ctrl->r->dma_if_intr_sts + 4, state_runout);
```

RTL93xx has `rx_rings = 32`, so `reg = rx_rings / 32` on main resolves to `sts + 4` — that is the
DONE register. Acknowledging it with the RUNOUT mask leaves the DONE bits set and the interrupt
re-fires as soon as NAPI re-enables the mask.

| eth0 | main | this PR | this PR, `+ 4` written with `state_done` |
|---|---|---|---|
| idle | 19.6 irq/s | 56 879 irq/s | 12.3 irq/s |
| after the load | 15.9 irq/s | 96 326 irq/s | 12.5 irq/s |
| 20 s UDP receive | 508 744 pkt, 0.031 irq/pkt | 362 899 pkt, 0.116 irq/pkt | 499 191 pkt, 0.031 irq/pkt |

Ping RTT over the switch goes from 0.54 ms to 20.2 ms (max 240 ms) while the storm is on, load
average 1.92 with no traffic. Writing `state_done` to `sts + 4` restores every number above. RTL83xx
looks unaffected: `rteth_83xx_confirm_and_disable_irqs()` writes the whole `state` back.

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -205,44 +205,68 @@ static int rteth_free_skb(struct sk_buff **skb)
...
+
+	regmap_read(ctrl->map, ctrl->r->dma_if_intr_sts, &state_runout);
+	regmap_read(ctrl->map, ctrl->r->dma_if_intr_sts + 4, &state_done);
+	*rings = state_runout | state_done;
+	*l2 = false;
+
+	regmap_clear_bits(ctrl->map, ctrl->r->dma_if_intr_msk, *rings);
+	regmap_clear_bits(ctrl->map, ctrl->r->dma_if_intr_msk + 4, *rings);
+	regmap_write(ctrl->map, ctrl->r->dma_if_intr_sts, state_runout);
+	regmap_write(ctrl->map, ctrl->r->dma_if_intr_sts + 4, state_runout);
```
openwrt-ai: The DONE status register is acknowledged with the RUNOUT value, so DONE bits are never cleared.

These status registers are write-1-to-clear — [`rteth_disable_all_irqs()` writes `GENMASK(31, 0)` to clear them](``https://github.com/openwrt/openwrt/blob/b356b8172eb7200b3bbf4e94e4eacd0222526555/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L267-L275),`` and the RTL83xx twin [writes the full `state` back](``https://github.com/openwrt/openwrt/blob/b356b8172eb7200b3bbf4e94e4eacd0222526555/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L241).`` In the normal RX path on RTL93xx the HOL feature means `state_runout == 0` while `state_done == BIT(ring)`, so this line writes `0` to `dma_if_intr_sts + 4` and clears nothing. The DONE bit stays latched; the mask is cleared here so the handler returns, but as soon as [`rteth_93xx_enable_rx_irq()`](``https://github.com/openwrt/openwrt/blob/b356b8172eb7200b3bbf4e94e4eacd0222526555/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L217-L225)`` re-arms the mask at the end of NAPI the still-asserted status re-fires the IRQ immediately — an interrupt storm. This matches the bench numbers reported in https://github.com/openwrt/openwrt/pull/24538#issuecomment-5172325973.

```suggestion
	regmap_write(ctrl->map, ctrl->r->dma_if_intr_sts + 4, state_done);
```

---

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -205,44 +205,68 @@ static int rteth_free_skb(struct sk_buff **skb)
...
+						unsigned long *rings, bool *l2)
+{
+	unsigned long flags;
+	u32 disable, state;
+
+	spin_lock_irqsave(&ctrl->lock, flags);
+
+	regmap_read(ctrl->map, ctrl->r->dma_if_intr_sts, &state);
+	*rings = FIELD_GET(GENMASK(7, 0), state) | FIELD_GET(GENMASK(15, 8), state);
+	*l2 = !!(state & GENMASK(22, 20));
```
openwrt-ai: nit: inlining the raw mask leaves [`RTL839X_DMA_IF_INTR_NOTIFY_MASK` (rtl838x_eth.h:115)](``https://github.com/openwrt/openwrt/blob/b356b8172eb7200b3bbf4e94e4eacd0222526555/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.h#L115``) with no users at all after this patch — it was the only reference. Either keep using it here or drop the define.

```suggestion
	*l2 = !!(state & RTL839X_DMA_IF_INTR_NOTIFY_MASK);
```

---

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
```diff
@@ -254,14 +278,8 @@ static void rteth_enable_all_rx_irqs(struct rteth_ctrl *ctrl)
...
 
-	/*
-	 * The hardware has several types of interrupts. Basically for rx/tx completion and
-	 * if hardware queues run out. For now the driver only needs notification about new
-	 * incoming packets. Leave everything else disabled.
-	 */
-	mask = GENMASK(ctrl->r->rx_rings - 1, 0) << (ctrl->r->rx_rings % 32);
-	reg = ctrl->r->rx_rings / 32;
-	regmap_update_bits(ctrl->map, ctrl->r->dma_if_intr_msk + reg * 4, mask, mask);
+	for (int ring = 0; ring < RTETH_RX_RINGS; ring++)
```
openwrt-ai, line 281: nit: this narrows the enabled interrupt set from all hardware rings to the two software rings, which none of the four bullets in the commit message mentions. Previously the mask covered `GENMASK(ctrl->r->rx_rings - 1, 0)` — all 8 (RTL83xx) or 32 (RTL93xx) DONE bits; now only rings 0 and 1 are armed. That looks correct given `RTETH_RX_RINGS == 2` and [`rteth_setup_cpu_rx_rings()`](``https://github.com/openwrt/openwrt/blob/b356b8172eb7200b3bbf4e94e4eacd0222526555/target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c#L537``) mapping every priority/reason into `reason % RTETH_RX_RINGS`, but it is a behaviour change worth a line in the commit message.

Same for the `DIV_ROUND_UP()` rewrite at line 269 — equivalent for both ring counts (`rx_rings * 2 + 7` is always odd, so it can never be an exact multiple of 32), but it is an undocumented drive-by hunk.

---

openwrt-ai (commented):
## Commit checks

- `b356b817` "realtek: eth: adapt interrupt handling for fragments" — typo in the message body: "driver configruation structure" should be "driver configuration structure". The message also omits two behaviour/refactor hunks that are in the diff (see inline note on `rteth_enable_all_rx_irqs()`).

---

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
plappermaul: fixed

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
plappermaul: fixed

target/linux/realtek/files-6.18/drivers/net/ethernet/rtl838x_eth.c
plappermaul: fixed

plappermaul:
> Tested on RTL9303 (Hasivo S1100W-8XGT-SE), RAM-boot, kernel 6.18.41: `main` [712bfdd](https://github.com/openwrt/openwrt/commit/712bfddc81489693cf2ffb4ddb9067b4e2ddfa3d) against the
> same tree with this commit cherry-picked. RTL930x regresses badly.

Thanks for testing. fixed.

l0rdg3x:
> > Tested on RTL9303 (Hasivo S1100W-8XGT-SE), RAM-boot, kernel 6.18.41: `main` [712bfdd](https://github.com/openwrt/openwrt/commit/712bfddc81489693cf2ffb4ddb9067b4e2ddfa3d) against the
> > same tree with this commit cherry-picked. RTL930x regresses badly.
> 
> Thanks for testing. fixed.

You are welcome!

openwrt-ai (commented):
Re-reviewed at `6ca19b38`. The two code findings from the previous round are resolved: `dma_if_intr_sts + 4` is now acked with `state_done` (matching the bench data in [#issuecomment-5172325973](https://github.com/openwrt/openwrt/pull/24538#issuecomment-5172325973)), and `RTL839X_DMA_IF_INTR_NOTIFY_MASK` is used again instead of the inlined `GENMASK(22, 20)`. No new code issues. CI on this head is green across all six realtek subtargets.

## Commit checks

- `6ca19b38` "realtek: eth: adapt interrupt handling for fragments" — the behaviour/refactor hunks are now documented in the "While we are here" section, thanks. Two typos remain in the message body: "driver configruation structure" (still there from the previous round) → "configuration", and the newly added "although it suppurts only 2 rings" → "supports".

---

plappermaul:

PR #24539, opened 2026-08-02
tools: elfutils: fix uio header inclusion

testuser7:
Fixes arm64 build failures by replacing <linux/uio.h> with POSIX <sys/uio.h>
to avoid header redefinition conflicts.

```
In file included from aarch64_initreg.c:36:
/usr/include/linux/uio.h:17:8: error: redefinition of 'struct iovec'
   17 | struct iovec
      |        ^~~~~
In file included from /usr/include/aarch64-linux-gnu/bits/fcntl-linux.h:38,
                 from /usr/include/aarch64-linux-gnu/bits/fcntl.h:55,
                 from /usr/include/fcntl.h:35,
                 from ../libgnu/fcntl.h:74,
                 from ../libgnu/unistd.h:117,
                 from /usr/include/aarch64-linux-gnu/bits/sigstksz.h:24,
                 from /usr/include/signal.h:328,
                 from ../libgnu/signal.h:52,
                 from /usr/include/aarch64-linux-gnu/sys/param.h:28,
                 from ../lib/system.h:47:
/usr/include/aarch64-linux-gnu/bits/types/struct_iovec.h:26:8: note: originally defined here
   26 | struct iovec
      |        ^~~~~
make[5]: *** [Makefile:2507: aarch64_initreg.o] Error 1
make[4]: *** [Makefile:2216: all-recursive] Error 1
make[3]: *** [Makefile:2123: all] Error 2
```

### DIFF
--- a/tools/elfutils/patches/006-Fix-build-on-aarch64-musl.patch
+++ b/tools/elfutils/patches/006-Fix-build-on-aarch64-musl.patch
@@ -1,0 +1,44 @@
+From 578f370c7e7a9f056aefa062b34590b0aa13bce5 Mon Sep 17 00:00:00 2001
+From: Hongxu Jia <hongxu.jia@windriver.com>
+Date: Tue, 15 Aug 2017 17:27:30 +0800
+Subject: [PATCH] Fix build on aarch64/musl
+
+Errors
+
+error: redefinition
+ of 'struct iovec'
+ struct iovec { void *iov_base; size_t iov_len; };
+        ^
+Upstream-Status: Pending
+Signed-off-by: Khem Raj <raj.khem@gmail.com>
+
+Rebase to 0.170
+Signed-off-by: Hongxu Jia <hongxu.jia@windriver.com>
+
+---
+ backends/aarch64_initreg.c | 2 +-
+ backends/arm_initreg.c     | 2 +-
+ 2 files changed, 2 insertions(+), 2 deletions(-)
+
+--- a/backends/aarch64_initreg.c
++++ b/backends/aarch64_initreg.c
+@@ -33,7 +33,7 @@
+ #include "system.h"
+ #include <assert.h>
+ #if defined(__aarch64__) && defined(__linux__)
+-# include <linux/uio.h>
++# include <sys/uio.h>
+ # include <sys/user.h>
+ # include <sys/ptrace.h>
+ # include <asm/ptrace.h>
+--- a/backends/arm_initreg.c
++++ b/backends/arm_initreg.c
+@@ -38,7 +38,7 @@
+ #endif
+ 
+ #ifdef __aarch64__
+-# include <linux/uio.h>
++# include <sys/uio.h>
+ # include <sys/user.h>
+ # include <sys/ptrace.h>
+ /* Deal with old glibc defining user_pt_regs instead of user_regs_struct.  */


peterwillcn (approved):

tools/elfutils/patches/111-uio-include-fix.patch
```diff
@@ -0,0 +1,22 @@
+--- a/backends/aarch64_initreg.c
```
openwrt-ai, line 23: These two hunks are byte-for-byte identical (same hunk headers, same context, including the `# include <asm/ptrace.h>` context line) to the ones already carried in-tree for the target-side package: [`package/libs/elfutils/patches/006-Fix-build-on-aarch64-musl.patch`](https://github.com/openwrt/openwrt/blob/2a20693c42208d443b56b8ba51841e301f7dcc2b/package/libs/elfutils/patches/006-Fix-build-on-aarch64-musl.patch). That patch carries a full git header with the original authorship (Khem Raj / Hongxu Jia) and an `Upstream-Status: Pending` marker; this copy drops all of it, so the provenance is lost and the two copies become hard to keep in sync.

Please keep the original header when copying the patch — that also satisfies the bot's "missing Git header parameters" warning and keeps it `git am`-able if it is ever sent upstream:

```suggestion
From 578f370c7e7a9f056aefa062b34590b0aa13bce5 Mon Sep 17 00:00:00 2001
From: Hongxu Jia <hongxu.jia@windriver.com>
Date: Tue, 15 Aug 2017 17:27:30 +0800
Subject: [PATCH] Fix build on aarch64/musl

error: redefinition of 'struct iovec'
 struct iovec { void *iov_base; size_t iov_len; };

Upstream-Status: Pending
Signed-off-by: Khem Raj <raj.khem@gmail.com>
Signed-off-by: Hongxu Jia <hongxu.jia@windriver.com>
---
--- a/backends/aarch64_initreg.c
```

(The `dwarf_fregs[r] & 0xFFFFFFFF` hunk from 006 is correctly left out here — it is a musl-only `long double` issue and does not apply to the glibc host build.)

---

tools/elfutils/patches/111-uio-include-fix.patch
```diff
@@ -0,0 +1,22 @@
...
++++ b/backends/aarch64_initreg.c
+@@ -33,7 +33,7 @@
+ #include "system.h"
+ #include <assert.h>
+ #if defined(__aarch64__) && defined(__linux__)
+-# include <linux/uio.h>
++# include <sys/uio.h>
```
openwrt-ai, line 30: nit: the filename/number diverges from the identical sister patch. The target-side copy is `006-Fix-build-on-aarch64-musl.patch` (0xx = upstreamable), while this one is filed as `111-` — the range `tools/elfutils/patches/` uses for OpenWrt-local build-system hacks (`100-portability`, `101-shared-conditional`, `110-objects-manifest`). This is a generic portability fix, not an OpenWrt-specific one. Naming it the same as the `package/libs/elfutils` copy would make the duplication greppable and keep the two in sync on the next version bump.

For what it's worth, current elfutils master still has `# include <linux/uio.h>` in both `backends/aarch64_initreg.c` and `backends/arm_initreg.c` (I checked a mirror at 2026-03-23), and those are the only two files in the tree that include it — so the patch is complete, but `Upstream-Status: Pending` from 2017 is still accurate and it would be worth actually sending to `elfutils-devel@sourceware.org`.

Also FYI on the bot's `PKG_RELEASE` warning: `PKG_RELEASE` is only consumed by `include/package-defaults.mk` for real packages, not by `include/host-build.mk`, so bumping it in `tools/elfutils/Makefile` has no effect. I'd treat that warning as a false positive here.

---

openwrt-ai (commented):
Reviewed 1 new commit. The change itself is technically correct: I verified against elfutils sources that `backends/aarch64_initreg.c` and `backends/arm_initreg.c` are the only two files including `<linux/uio.h>`, that both hunks' context matches the 0.192 sources exactly (including the `# include <asm/ptrace.h>` line), and that nothing else from `<linux/uio.h>` is needed — `struct iovec` comes from `<sys/uio.h>` and the `__u64` used by `struct user_pac_mask` still resolves via `<asm/ptrace.h>` → `<linux/types.h>`. Commit message matches the diff. Two comments below about patch provenance/naming.

---

mcprat:
Hi, thanks, this is the right fix...

If you like you can add more context to the commit message:

gnulib's fcntl.h ends up including glibc's fcntl.h and fcntl-linux.h
which includes glibc's types/struct_iovec.h,
thats where the redefinition happens.

the lines that use <linux/uio.h> are antiquated and were never necessary.
it can be traced back to the original 2014 change that added the file with aarch64 support
in upstream commit 66637fa21044ac0058b25522f473669e73de328b

it was never necessary to use the linux header over the glibc headers,
as elfutils only needs the iovec struct and that struct was present since glibc 2.0 in the 1990s

mcprat:
ping @robimarko 

fix for arm64 hosts

OpenWRT-fanboy (approved):
I use GitHub Actions to automate building OpenWrt. The ARM runners are faster than x86 ones for this task. They stopped working after the gnulib update was committed, erroring out on elfutils. This patch restores building to completion.

tools/elfutils/patches/006-Fix-build-on-aarch64-musl.patch
```diff
@@ -0,0 +1,46 @@
+From 578f370c7e7a9f056aefa062b34590b0aa13bce5 Mon Sep 17 00:00:00 2001
+From: Hongxu Jia <hongxu.jia@windriver.com>
+Date: Tue, 15 Aug 2017 17:27:30 +0800
+Subject: [PATCH] Fix build on aarch64/musl
+
+Errors
+
+invalid operands to binary & (have 'long double' and 'unsigned int')
+
```
openwrt-ai: nit: this copy describes an error it does not fix. The `invalid operands to binary &` error is what the `dwarf_fregs[r] = (unsigned int)fregs.vregs[r] & 0xFFFFFFFF;` hunk addresses, and that hunk is deliberately (and correctly) absent here — it is a musl-only `long double` issue that does not affect the glibc host build. The sister patch [`package/libs/elfutils/patches/006-Fix-build-on-aarch64-musl.patch:28-35`](``https://github.com/openwrt/openwrt/blob/24acdbfea4d88d75fde26596042e267b54c6e7d5/package/libs/elfutils/patches/006-Fix-build-on-aarch64-musl.patch#L28-L35``) is the one that carries it.

Since the diffstat below was already adjusted for the dropped hunk (`2 +-` instead of `4 ++--`), dropping the matching error text keeps the header self-consistent:

```suggestion
Errors

```


---

openwrt-ai (commented):
Reviewed 1 new commit (force-push of `2a20693` → `24acdbf`). Both points from the previous review were addressed: the patch is renamed to `006-Fix-build-on-aarch64-musl.patch`, matching the target-side copy, and the original git header with Khem Raj / Hongxu Jia authorship and `Upstream-Status: Pending` is restored. The diffstat was correctly adjusted to `2 files changed, 2 insertions(+), 2 deletions(-)` for the single-hunk-per-file form, and the added commit-message body (header include chain, upstream `66637fa2` provenance) matches the diff. One remaining nit inline.

---

robimarko:

PR #15977, opened 2024-07-21
qualcommax: ipq6018: Add Wally's Tech DR6018-v4

okoklets:
Add support for Wally's Tech DR6018-v4.

Qualcomm-Atheros IPQ6010
Quad-core ARM 64bit A53 @1.8GHz Processor 
1GB DDRL3L System Memory 
32MB NOR Flash, 256MB NAND Flash 
Supports Dynamic Frequency Selection (DFS) 
2x2 On-board 2.4GHz radio, up to 573Mbps physical Data Rate 
2x2 On-board 5GHz radio, up to 1201Mbps physical Data Rate

tftpboot the factory bin image and run (in u-boot):
imxtract 0x44000000 ubi && flash rootfs && reset

sysupgrade /should/ work but is untested.

flash factory image from qsdk/chaos calmer factory webui works.

PR #24540, opened 2026-08-02
qualcommax: ipq60xx: add support for Wallys DR6018-v4

okoklets:
## Summary
- Add support for the Wallys Tech DR6018-v4, an IPQ6010-based board with 1GB DDR3L, 8MB NOR flash, 256MB NAND flash, and 2x2 on-board 2.4GHz/5GHz radios.
- Fixes ethernet switch port configuration to use the current mainline PPE binding (the original DTS used an obsolete `qcom,port_phyinfo`/`switch_lan_bmp` binding that doesn't exist in the current kernel).

Supersedes #15977, which was opened from `main` instead of a dedicated branch.

## Test plan
- [x] Build and flash to DR6018-v4 hardware
- [x] Verify all 4 LAN ports and the WAN port come up correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01H4yAnWB2DngCLCvzYq6qyY

### DIFF
--- a/target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
+++ b/target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
@@ -1,0 +1,291 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR BSD-3-Clause
+
+/dts-v1/;
+
+#include "ipq6018.dtsi"
+#include "ipq6018-ess.dtsi"
+#include <dt-bindings/gpio/gpio.h>
+
+/ {
+	model = "Wallys DR6018 v4";
+	compatible = "wallys,dr6018-v4", "qcom,ipq6018";
+
+	aliases {
+		serial0 = &blsp1_uart3;
+		/*
+		 * Aliases as required by u-boot
+		 * to patch MAC addresses
+		 */
+		ethernet0 = &swport1;
+		ethernet1 = &swport2;
+		ethernet2 = &swport3;
+		ethernet3 = &swport4;
+		ethernet4 = &wan;
+	};
+
+	chosen {
+		stdout-path = "serial0:115200n8";
+		bootargs-append = " root=/dev/ubiblock0_1";
+	};
+
+	reg_sd_vmmc: regulator-sdcard-vmmc {
+		compatible = "regulator-fixed";
+		regulator-name = "sdcard-vmmc";
+		regulator-min-microvolt = <3000000>;
+		regulator-max-microvolt = <3000000>;
+
+		startup-delay-us = <200>;
+
+		gpio = <&tlmm 66 GPIO_ACTIVE_HIGH>;
+		enable-active-high;
+
+		pinctrl-names = "default";
+		pinctrl-0 = <&sd_vmmc_en_default>;
+	};
+};
+
+&tlmm {
+	/* TZ has exclusive control over GPIO20 */
+	gpio-reserved-ranges = <20 1>;
+
+	spi_0_pins: spi-0-state {
+		pins = "gpio38", "gpio39", "gpio40", "gpio41";
+		function = "blsp0_spi";
+		drive-strength = <8>;
+		bias-pull-down;
+	};
+
+	mdio_pins: mdio-state {
+		mdc {
+			pins = "gpio64";
+			function = "mdc";
+			drive-strength = <8>;
+			bias-pull-up;
+		};
+
+		mdio {
+			pins = "gpio65";
+			function = "mdio";
+			drive-strength = <8>;
+			bias-pull-up;
+		};
+	};
+
+	sd_vmmc_en_default: sd-vmmc-en-default-state {
+		pins = "gpio66";
+		function = "gpio";
+		drive-strength = <8>;
+		bias-pull-down;
+	};
+
+	sd_pins: sd-state {
+		pins = "gpio62";
+		function = "gpio";
+		drive-strength = <8>;
+		bias-pull-up;
+	};
+};
+
+/*
+ * Board does not use companion MP5496 PMIC,
+ * but rather uses fixed external SMPS.
+ */
+&rpm {
+	status = "disabled";
+};
+
+&blsp1_uart3 {
+	status = "okay";
+
+	pinctrl-0 = <&serial_3_pins>;
+	pinctrl-names = "default";
+};
+
+&blsp1_spi1 {
+	pinctrl-0 = <&spi_0_pins>;
+	pinctrl-names = "default";
+	status = "okay";
+
+	flash@0 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		reg = <0>;
+		compatible = "jedec,spi-nor";
+		spi-max-frequency = <50000000>;
+	};
+};
+
+&qpic_bam {
+	status = "okay";
+};
+
+&qpic_nand {
+	status = "okay";
+
+	nand@0 {
+		reg = <0>;
+
+		nand-ecc-strength = <8>;
+		nand-ecc-step-size = <512>;
+		nand-bus-width = <8>;
+
+		partitions {
+			compatible = "qcom,smem-part";
+		};
+	};
+};
+
+&qusb_phy_0 {
+	status = "okay";
+};
+
+&ssphy_0 {
+	status = "okay";
+};
+
+&usb3 {
+	status = "okay";
+};
+
+&qusb_phy_1 {
+	status = "disabled";
+};
+
+&usb2 {
+	status = "disabled";
+};
+
+&pcie_phy {
+	status = "okay";
+};
+
+&pcie0 {
+	status = "okay";
+};
+
+&sdhc {
+	status = "okay";
+	pinctrl-0 = <&sd_pins>;
+	pinctrl-names = "default";
+
+	cd-gpios = <&tlmm 62 GPIO_ACTIVE_LOW>;
+	wp-gpios = <&tlmm 63 GPIO_ACTIVE_HIGH>;
+	vmmc-supply = <&reg_sd_vmmc>;
+	bus-width = <4>;
+};
+
+&mdio {
+	status = "okay";
+	pinctrl-0 = <&mdio_pins>;
+	pinctrl-names = "default";
+	reset-gpios = <&tlmm 75 GPIO_ACTIVE_LOW>;
+
+	ethernet-phy-package@0 {
+		compatible = "qcom,qca8075-package";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		reg = <0>;
+
+		qca8075_0: ethernet-phy@0 {
+			compatible = "ethernet-phy-ieee802.3-c22";
+			reg = <0>;
+		};
+
+		qca8075_1: ethernet-phy@1 {
+			compatible = "ethernet-phy-ieee802.3-c22";
+			reg = <1>;
+		};
+
+		qca8075_2: ethernet-phy@2 {
+			compatible = "ethernet-phy-ieee802.3-c22";
+			reg = <2>;
+		};
+
+		qca8075_3: ethernet-phy@3 {
+			compatible = "ethernet-phy-ieee802.3-c22";
+			reg = <3>;
+		};
+	};
+
+	qca8081: ethernet-phy@24 {
+		compatible = "ethernet-phy-id004d.d101";
+		reg = <24>;
+		reset-gpios = <&tlmm 77 GPIO_ACTIVE_LOW>;
+	};
+};
+
+&edma {
+	status = "okay";
+};
+
+&wifi {
+	status = "okay";
+};
+
+&uniphy0 {
+	status = "okay";
+};
+
+&uniphy1 {
+	status = "okay";
+};
+
+&switch {
+	status = "okay";
+
+	ports {
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		port@0 {
+			reg = <0>;
+			ethernet = <&edma>;
+			phy-mode = "internal";
+
+			fixed-link {
+				speed = <1000>;
+				full-duplex;
+			};
+		};
+
+		swport1: port@1 {
+			reg = <1>;
+			label = "lan1";
+			phy-handle = <&qca8075_0>;
+			phy-mode = "psgmii";
+			pcs-handle = <&uniphy0 0>;
+		};
+
+		swport2: port@2 {
+			reg = <2>;
+			label = "lan2";
+			phy-handle = <&qca8075_1>;
+			phy-mode = "psgmii";
+			pcs-handle = <&uniphy0 1>;
+		};
+
+		swport3: port@3 {
+			reg = <3>;
+			label = "lan3";
+			phy-handle = <&qca8075_2>;
+			phy-mode = "psgmii";
+			pcs-handle = <&uniphy0 2>;
+		};
+
+		swport4: port@4 {
+			reg = <4>;
+			label = "lan4";
+			phy-handle = <&qca8075_3>;
+			phy-mode = "psgmii";
+			pcs-handle = <&uniphy0 3>;
+		};
+
+		wan: port@5 {
+			reg = <5>;
+			label = "wan";
+			phy-handle = <&qca8081>;
+			phy-mode = "2500base-x";
+			pcs-handle = <&uniphy1 0>;
+		};
+	};
+};
--- a/target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
+++ b/target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
@@ -1,0 +1,295 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR BSD-3-Clause
+
+/dts-v1/;
+
+#include "ipq6018.dtsi"
+#include "ipq6018-ess.dtsi"
+#include <dt-bindings/gpio/gpio.h>
+
+/ {
+	model = "Wallys DR6018 v4";
+	compatible = "wallys,dr6018-v4", "qcom,ipq6018";
+
+	aliases {
+		serial0 = &blsp1_uart3;
+		/*
+		 * Aliases as required by u-boot
+		 * to patch MAC addresses
+		 */
+		ethernet0 = &swport1;
+		ethernet1 = &swport2;
+		ethernet2 = &swport3;
+		ethernet3 = &swport4;
+		ethernet4 = &wan;
+	};
+
+	chosen {
+		stdout-path = "serial0:115200n8";
+		bootargs-append = " root=/dev/ubiblock0_1";
+	};
+
+	reg_sd_vmmc: regulator-sdcard-vmmc {
+		compatible = "regulator-fixed";
+		regulator-name = "sdcard-vmmc";
+		regulator-min-microvolt = <3000000>;
+		regulator-max-microvolt = <3000000>;
+
+		startup-delay-us = <200>;
+
+		gpio = <&tlmm 66 GPIO_ACTIVE_HIGH>;
+		enable-active-high;
+
+		pinctrl-names = "default";
+		pinctrl-0 = <&sd_vmmc_en_default>;
+	};
+};
+
+&tlmm {
+	/* TZ has exclusive control over GPIO20 */
+	gpio-reserved-ranges = <20 1>;
+
+	spi_0_pins: spi-0-state {
+		pins = "gpio38", "gpio39", "gpio40", "gpio41";
+		function = "blsp0_spi";
+		drive-strength = <8>;
+		bias-pull-down;
+	};
+
+	mdio_pins: mdio-state {
+		mdc {
+			pins = "gpio64";
+			function = "mdc";
+			drive-strength = <8>;
+			bias-pull-up;
+		};
+
+		mdio {
+			pins = "gpio65";
+			function = "mdio";
+			drive-strength = <8>;
+			bias-pull-up;
+		};
+	};
+
+	sd_vmmc_en_default: sd-vmmc-en-default-state {
+		pins = "gpio66";
+		function = "gpio";
+		drive-strength = <8>;
+		bias-pull-down;
+	};
+
+	sd_pins: sd-state {
+		pins = "gpio62";
+		function = "gpio";
+		drive-strength = <8>;
+		bias-pull-up;
+	};
+};
+
+/*
+ * Board does not use companion MP5496 PMIC,
+ * but rather uses fixed external SMPS.
+ */
+&rpm {
+	status = "disabled";
+};
+
+&blsp1_uart3 {
+	status = "okay";
+
+	pinctrl-0 = <&serial_3_pins>;
+	pinctrl-names = "default";
+};
+
+&blsp1_spi1 {
+	pinctrl-0 = <&spi_0_pins>;
+	pinctrl-names = "default";
+	status = "okay";
+
+	flash@0 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		reg = <0>;
+		compatible = "jedec,spi-nor";
+		spi-max-frequency = <50000000>;
+	};
+};
+
+&qpic_bam {
+	status = "okay";
+};
+
+&qpic_nand {
+	status = "okay";
+
+	nand@0 {
+		reg = <0>;
+
+		nand-ecc-strength = <8>;
+		nand-ecc-step-size = <512>;
+		nand-bus-width = <8>;
+
+		partitions {
+			compatible = "qcom,smem-part";
+		};
+	};
+};
+
+&qusb_phy_0 {
+	status = "okay";
+};
+
+&ssphy_0 {
+	status = "okay";
+};
+
+&usb3 {
+	status = "okay";
+};
+
+&qusb_phy_1 {
+	status = "disabled";
+};
+
+&usb2 {
+	status = "disabled";
+};
+
+&pcie_phy {
+	status = "okay";
+};
+
+&pcie0 {
+	status = "okay";
+};
+
+&sdhc {
+	status = "okay";
+	pinctrl-0 = <&sd_pins>;
+	pinctrl-names = "default";
+
+	cd-gpios = <&tlmm 62 GPIO_ACTIVE_LOW>;
+	wp-gpios = <&tlmm 63 GPIO_ACTIVE_HIGH>;
+	vmmc-supply = <&reg_sd_vmmc>;
+	bus-width = <4>;
+};
+
+&mdio {
+	status = "okay";
+	pinctrl-0 = <&mdio_pins>;
+	pinctrl-names = "default";
+	reset-gpios = <&tlmm 75 GPIO_ACTIVE_LOW>;
+	reset-delay-us = <10000>;
+	reset-post-delay-us = <50000>;
+
+	ethernet-phy-package@0 {
+		compatible = "qcom,qca8075-package";
+		#address-cells = <1>;
+		#size-cells = <0>;
+		reg = <0>;
+
+		qca8075_0: ethernet-phy@0 {
+			compatible = "ethernet-phy-ieee802.3-c22";
+			reg = <0>;
+		};
+
+		qca8075_1: ethernet-phy@1 {
+			compatible = "ethernet-phy-ieee802.3-c22";
+			reg = <1>;
+		};
+
+		qca8075_2: ethernet-phy@2 {
+			compatible = "ethernet-phy-ieee802.3-c22";
+			reg = <2>;
+		};
+
+		qca8075_3: ethernet-phy@3 {
+			compatible = "ethernet-phy-ieee802.3-c22";
+			reg = <3>;
+		};
+	};
+
+	qca8081: ethernet-phy@24 {
+		compatible = "ethernet-phy-id004d.d101";
+		reg = <24>;
+		reset-gpios = <&tlmm 77 GPIO_ACTIVE_LOW>;
+		reset-assert-us = <10000>;
+		reset-deassert-us = <50000>;
+	};
+};
+
+&edma {
+	status = "okay";
+};
+
+&wifi {
+	status = "okay";
+};
+
+&uniphy0 {
+	status = "okay";
+};
+
+&uniphy1 {
+	status = "okay";
+};
+
+&switch {
+	status = "okay";
+
+	ports {
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		port@0 {
+			reg = <0>;
+			ethernet = <&edma>;
+			phy-mode = "internal";
+
+			fixed-link {
+				speed = <1000>;
+				full-duplex;
+			};
+		};
+
+		swport1: port@1 {
+			reg = <1>;
+			label = "lan1";
+			phy-handle = <&qca8075_0>;
+			phy-mode = "psgmii";
+			pcs-handle = <&uniphy0 0>;
+		};
+
+		swport2: port@2 {
+			reg = <2>;
+			label = "lan2";
+			phy-handle = <&qca8075_1>;
+			phy-mode = "psgmii";
+			pcs-handle = <&uniphy0 1>;
+		};
+
+		swport3: port@3 {
+			reg = <3>;
+			label = "lan3";
+			phy-handle = <&qca8075_2>;
+			phy-mode = "psgmii";
+			pcs-handle = <&uniphy0 2>;
+		};
+
+		swport4: port@4 {
+			reg = <4>;
+			label = "lan4";
+			phy-handle = <&qca8075_3>;
+			phy-mode = "psgmii";
+			pcs-handle = <&uniphy0 3>;
+		};
+
+		wan: port@5 {
+			reg = <5>;
+			label = "wan";
+			phy-handle = <&qca8081>;
+			phy-mode = "2500base-x";
+			pcs-handle = <&uniphy1 0>;
+		};
+	};
+};
--- a/target/linux/qualcommax/image/ipq60xx.mk
+++ b/target/linux/qualcommax/image/ipq60xx.mk
@@ -338,6 +338,22 @@ define Device/tplink_eap620-hd-v3
 endef
 TARGET_DEVICES += tplink_eap620-hd-v3
 
+define Device/wallys_dr6018-v4
+	$(call Device/FitImage)
+	$(call Device/UbiFit)
+	DEVICE_VENDOR := Wallys
+	DEVICE_MODEL := DR6018
+	DEVICE_VARIANT := v4
+	BLOCKSIZE := 128k
+	PAGESIZE := 2048
+	SOC := ipq6010
+	DEVICE_DTS_CONFIG := config@cp01-c4
+	IMAGES += factory.bin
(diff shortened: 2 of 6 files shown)


okoklets:
OK!  All checks passed.  Please can we have this merged now? (; Thanks in advance.

target/linux/qualcommax/image/ipq60xx.mk
```diff
@@ -319,6 +319,25 @@ define Device/tplink_eap620-hd-v3
...
+	PAGESIZE := 2048
+	SOC := ipq6010
+	DEVICE_DTS_CONFIG := config@cp01-c4
+	IMAGES := sysupgrade.tar nand-factory.bin
+	IMAGE/sysupgrade.tar := sysupgrade-tar | append-metadata
+	IMAGE/nand-factory.bin := append-ubi | qsdk-ipq-factory-nand
+	SUPPORTED_DEVICES := wallys,dr6018-v4
+	DEVICE_PACKAGES := ipq-wifi-wallys-dr6018-v4 \
+		ath11k-firmware-qcn9074 kmod-ath11k-pci uboot-envtools \
+		-kmod-usb-dwc3-of-simple kmod-usb-dwc3-qcom kmod-usb3 kmod-usb2
```
openwrt-ai: `ipq-wifi-wallys-dr6018-v4` is not a package that exists. [`generate-ipq-wifi-package`](https://github.com/openwrt/openwrt/blob/21eb2175dc9a23e0da83c277350078ed63232bed/package/firmware/ipq-wifi/Makefile#L182-L183) names the package `ipq-wifi-$(1)` with `$(1)` taken verbatim from the eval call, and this PR adds `wallys_dr6018-v4` — so the package is `ipq-wifi-wallys_dr6018-v4`, with an underscore between vendor and model, exactly like [`ipq-wifi-yuncore_fap650`](https://github.com/openwrt/openwrt/blob/21eb2175dc9a23e0da83c277350078ed63232bed/target/linux/qualcommax/image/ipq60xx.mk#L350) two blocks below. With the hyphenated spelling the caldata package is never selected, which is the same class of bug 21eb217 was meant to fix.

The rest of the list is also dead weight: `uboot-envtools`, `kmod-usb3` and `kmod-usb-dwc3-qcom` are already in the target's [`DEFAULT_PACKAGES`](https://github.com/openwrt/openwrt/blob/21eb2175dc9a23e0da83c277350078ed63232bed/target/linux/qualcommax/Makefile#L14-L19); `-kmod-usb-dwc3-of-simple` deselects a package that does not exist anywhere in the tree (no `KernelPackage/usb-dwc3-of-simple` in `package/kernel/linux/modules/usb.mk`); and `kmod-usb2` is the EHCI HCD, which is unused here — `&usb2` and `&qusb_phy_1` are disabled in the DTS and USB 2.0 devices on `&usb3` are served by xHCI from `kmod-usb3`. No other qualcommax device block lists any USB kmod.

```suggestion
	DEVICE_PACKAGES := ipq-wifi-wallys_dr6018-v4 \
		ath11k-firmware-qcn9074 kmod-ath11k-pci
```

---

target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
```diff
@@ -0,0 +1,291 @@
...
+	pinctrl-names = "default";
+	status = "okay";
+
+	flash@0 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		reg = <0>;
+		compatible = "jedec,spi-nor";
+		spi-max-frequency = <50000000>;
+	};
```
openwrt-ai, line 115: This DTS declares no partition table at all — neither this `flash@0` nor `nand@0` below has a `partitions` node, and it is the only DTS under `target/linux/qualcommax/dts/` without one.

Nothing else fills that gap: [`CONFIG_MTD_CMDLINE_PARTS is not set`](https://github.com/openwrt/openwrt/blob/21eb2175dc9a23e0da83c277350078ed63232bed/target/linux/generic/config-6.18#L4008), so no `mtdparts=` from U-Boot is honoured, and the SMEM parser is only bound through a `partitions` child node — it is registered purely via [`of_match_table` in `drivers/mtd/parsers/qcomsmempart.c`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/mtd/parsers/qcomsmempart.c#L181-L192) (checked against v6.18.41, matching `KERNEL_PATCHVER:=6.18`). Without a `partitions` node the ofpart/fixed-partitions parser also finds nothing, so both flashes come up as a single unpartitioned MTD.

That breaks two things this PR adds: `caldata_extract "0:art" 0x1000 0x20000` in the ath11k hotplug script has no `0:art` partition to open, and `bootargs-append = " root=/dev/ubiblock0_1"` has no `ubi`/`rootfs` MTD to attach.

Please add the board's partition table, the way sibling ipq60xx boards do — e.g. [`ipq6018-fap650.dts`](https://github.com/openwrt/openwrt/blob/21eb2175dc9a23e0da83c277350078ed63232bed/target/linux/qualcommax/dts/ipq6018-fap650.dts#L224) spells out `fixed-partitions` on the NOR (including [`0:art`](https://github.com/openwrt/openwrt/blob/21eb2175dc9a23e0da83c277350078ed63232bed/target/linux/qualcommax/dts/ipq6018-fap650.dts#L306)) plus a second table on the NAND, while several others use `partitions { compatible = "qcom,smem-part"; }`.

---

target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
```diff
@@ -0,0 +1,291 @@
...
+		reset-gpios = <&tlmm 77 GPIO_ACTIVE_LOW>;
+	};
+};
+
+&edma {
+	status = "okay";
+};
+
+&wifi {
+        status = "okay";
```
openwrt-ai: Indented with 8 spaces; every other line in this file uses tabs.

```suggestion
	status = "okay";
```

---

target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
```diff
@@ -0,0 +1,291 @@
...
+	qcom,ath11k-calibration-variant = "Wallys-DR6018";
```
openwrt-ai: The variant string drops the `-V4` that the rest of the PR carries: the ipq-wifi board is registered as `wallys_dr6018-v4` and the compatible is `wallys,dr6018-v4`, but the variant asks ath11k for `Wallys-DR6018`. Every other qualcommax board keeps the two in lockstep (`Yuncore-AX830` / `yuncore_ax830`, `Linksys-MX6200` / `linksys_mx6200`, …), so this looks like it will not resolve against the board file.

Two related questions, since neither is verifiable from the diff and this PR has no build CI attached (only the three FormalityCheck runs):

- Does `board-wallys_dr6018-v4.ipq6018` actually exist in `firmware_qca-wireless` at the pinned `PKG_SOURCE_VERSION:=e20f4c6f…` (dated 2026-05-18, i.e. before this PR)? The Makefile documents the board file having to land upstream first, and the install rule is a [`$(wildcard …/board-$(1).*)`](https://github.com/openwrt/openwrt/blob/21eb2175dc9a23e0da83c277350078ed63232bed/package/firmware/ipq-wifi/Makefile#L202) — if the file is missing the package builds fine and simply installs nothing, so a green build would not catch it.
- Does the variant recorded inside that board file read `Wallys-DR6018` or `Wallys-DR6018-V4`? Whichever it is, this property has to match it exactly.

---

target/linux/qualcommax/image/ipq60xx.mk
```diff
@@ -319,6 +319,25 @@ define Device/tplink_eap620-hd-v3
...
+	DEVICE_VENDOR := Wallys
+	DEVICE_MODEL := DR6018-V4
+	BLOCKSIZE := 128k
+	PAGESIZE := 2048
+	SOC := ipq6010
+	DEVICE_DTS_CONFIG := config@cp01-c4
+	IMAGES := sysupgrade.tar nand-factory.bin
+	IMAGE/sysupgrade.tar := sysupgrade-tar | append-metadata
+	IMAGE/nand-factory.bin := append-ubi | qsdk-ipq-factory-nand
+	SUPPORTED_DEVICES := wallys,dr6018-v4
```
openwrt-ai: `IMAGES :=` overwrites the list instead of extending it, which throws away the `factory.ubi` that `$(call Device/UbiFit)` on line 324 just added, and then re-implements `sysupgrade.bin` under a new name — `IMAGE/sysupgrade.tar := sysupgrade-tar | append-metadata` is byte-for-byte what `Device/Default` already defines for `sysupgrade.bin`. Neither `sysupgrade.tar` nor `nand-factory.bin` is used by any other device in the tree; the qualcommax spelling for this artifact is `factory.bin` / `factory.ubin`.

`SUPPORTED_DEVICES` is also redundant here — `Device/Default` derives exactly `wallys,dr6018-v4` from the device name via `$(subst _,$(comma),$(1))`.

```suggestion
	IMAGES += factory.bin
	IMAGE/factory.bin := append-ubi | qsdk-ipq-factory-nand
```

---

target/linux/qualcommax/image/ipq60xx.mk
```diff
@@ -319,6 +319,25 @@ define Device/tplink_eap620-hd-v3
 endef
 TARGET_DEVICES += tplink_eap620-hd-v3
 
+define Device/wallys_dr6018-v4
+	$(call Device/FitImage)
+	$(call Device/UbiFit)
+	DEVICE_VENDOR := Wallys
+	DEVICE_MODEL := DR6018-V4
```
openwrt-ai: nit: the device is spelled four different ways across this PR — `DR6018-V4` here, `Wallys DR6018 v4` in the DTS `model` (ipq6010-dr6018-v4.dts:11), `Wallys DR6018-v4` in the ipq-wifi package title, and `DR6018-v4` in the PR body. Splitting the revision into `DEVICE_VARIANT` makes the profile name come out as `Wallys DR6018 v4`, identical to the DTS `model`, and matches how the neighbouring TP-Link blocks handle revisions.

```suggestion
	DEVICE_MODEL := DR6018
	DEVICE_VARIANT := v4
```

---

target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
```diff
@@ -0,0 +1,291 @@
...
+/dts-v1/;
+
+#include "ipq6018.dtsi"
+#include "ipq6018-ess.dtsi"
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+
+/ {
+	model = "Wallys DR6018 v4";
+	compatible = "wallys,dr6018-v4", "wallys,dr6018", "qcom,ipq6018";
```
openwrt-ai: nit: this line is a good illustration of why the series wants squashing before merge — ff1355f ("fix DR6018-v4 board_name detection") does nothing but reorder the two strings that 07bd68a introduced one commit earlier. The same holds for cd66805 (rewrites the switch node 07bd68a added), 43393dd (re-wraps the PHYs 07bd68a added) and 21eb217 (fixes the package name 07bd68a/85106ea wrote). None of these bugs ever existed in the tree, so bisect gains nothing from the split; the usual form for a new board is a single "add support for …" commit.

While reordering: is the intermediate `wallys,dr6018` compatible carrying its weight? Nothing in the tree matches it, and `SUPPORTED_DEVICES` no longer lists it after ff1355f.

---

target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
```diff
@@ -0,0 +1,291 @@
...
+		/*
+		 * Aliases as required by u-boot
+		 * to patch MAC addresses
+		 */
+		ethernet0 = &swport1;
+		ethernet1 = &swport2;
+		ethernet2 = &swport3;
+		ethernet3 = &swport4;
+		ethernet4 = &wan;
+	};
```
openwrt-ai, line 24: nit: no `led-boot` / `led-failsafe` / `led-running` / `led-upgrade` aliases here, and the file has no `leds` or `keys` node at all — every other ipq6010 board in the tree defines both (see [`ipq6010-re-cs.dtsi`](https://github.com/openwrt/openwrt/blob/21eb2175dc9a23e0da83c277350078ed63232bed/target/linux/qualcommax/dts/ipq6010-re-cs.dtsi#L10-L41) and [`ipq6010-xe3-4.dts`](https://github.com/openwrt/openwrt/blob/21eb2175dc9a23e0da83c277350078ed63232bed/target/linux/qualcommax/dts/ipq6010-xe3-4.dts#L17-L58)), and the target ships `kmod-leds-gpio` and `kmod-gpio-button-hotplug` by default. Does the DR6018-v4 genuinely have no status LED and no reset button, or is this just not wired up yet? Without `led-boot`/`led-upgrade` there is no visual feedback during failsafe or sysupgrade.

`<dt-bindings/input/input.h>` is included on line 8 but nothing in the file uses it, which suggests the keys node was intended.

---

openwrt-ai (commented):
## Commit checks

- 07bd68a "qualcommax: ipq60xx: add support for Wallys DR6018-v4" — this is the commit that introduces the new device (new `.dts` under `target/linux/qualcommax/dts/` plus a new `define Device/…` in `target/linux/qualcommax/image/ipq60xx.mk`), so it needs the three sections OpenWrt expects for new-device commits. Right now it has a one-paragraph summary (SoC, RAM, flash, radios) and then a changelog of the patch itself. Missing: the rest of the **hardware specification** (Ethernet port count/speed and switch/PHY chips, LEDs and buttons, UART header location + pinout + baud rate, power input), the **flash instructions** (at least one fully described install path — which image, how to get the device into a flashable state, verbatim button/IP/hostname strings, and how to revert to stock), and the **MAC address layout** (one line per interface saying where each MAC comes from — the DTS aliases suggest U-Boot patches `ethernet0..4`, which is worth stating explicitly). The bar is that a stranger could flash the board from the commit message alone.
- 07bd68a vs cd66805 / 43393dd / ff1355f / 85106ea / 21eb217 — the subject scope is inconsistent within one series: the first commit uses `qualcommax: ipq60xx:` and the other five use `qualcommax: ipq6018:`. `ipq60xx` is the subtarget directory these files live in; `ipq6018` is neither that nor this board's SoC (`SOC := ipq6010`). Worth settling on one prefix (moot if the series is squashed, see the inline note).
- 21eb217 "qualcommax: ipq6018: fix DR6018-v4 calibration package" — the message says it switches to "the generated ipq-wifi package name", but the name the generator produces is `ipq-wifi-wallys_dr6018-v4` and the diff writes `ipq-wifi-wallys-dr6018-v4`. Details inline.


---

target/linux/qualcommax/ipq60xx/base-files/etc/hotplug.d/firmware/11-ath11k-caldata
```diff
@@ -93,6 +93,9 @@ case "$FIRMWARE" in
 		ath11k_patch_mac $(macaddr_add $label_mac 1) 0
 		ath11k_set_macflag
 		;;
+	wallys,dr6018-v4)
+		caldata_extract "0:art" 0x1000 0x20000
+		;;
```
openwrt-ai: The QCN9074 radio gets no calibration data. This entry sits inside the `"ath11k/IPQ6018/hw1.0/cal-ahb-c000000.wifi.bin"` case, which only covers the SoC-internal radio. The board also has a PCIe QCN9074 — the commit message lists it, the DTS enables [`&pcie0`](https://github.com/openwrt/openwrt/blob/c74a3da36e2a85beed5dc3ad6c25a0692fccf86d/target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts#L162-L164), and the image block pulls in `ath11k-firmware-qcn9074 kmod-ath11k-pci` — but there is no `wallys,dr6018-v4` arm in the [`"ath11k/QCN9074/hw1.0/cal-pci-0000:01:00.0.bin"` case](``https://github.com/openwrt/openwrt/blob/c74a3da36e2a85beed5dc3ad6c25a0692fccf86d/target/linux/qualcommax/ipq60xx/base-files/etc/hotplug.d/firmware/11-ath11k-caldata#L104-L119)`` below, so that hotplug invocation falls through to `esac` and the PCIe radio comes up on default firmware calibration.

The commit message says "WLAN calibration and radio data are read from the 0:art NAND partition" without qualification, which reads as covering both radios. The three in-tree ipq60xx boards with a QCN9074 ([`cambiumnetworks,xe3-4`](``https://github.com/openwrt/openwrt/blob/c74a3da36e2a85beed5dc3ad6c25a0692fccf86d/target/linux/qualcommax/ipq60xx/base-files/etc/hotplug.d/firmware/11-ath11k-caldata#L106-L108),`` `jdcloud,re-cs-02`, `linksys,mr7500`) all extract the second radio's caldata at offset `0x26800`. Please add the matching arm to the QCN9074 case, or say in the commit message why this board does not need one.

---

target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
```diff
@@ -0,0 +1,291 @@
...
+		qca8075_3: ethernet-phy@3 {
+			compatible = "ethernet-phy-ieee802.3-c22";
+			reg = <3>;
+		};
+	};
+
+	qca8081: ethernet-phy@24 {
+		compatible = "ethernet-phy-id004d.d101";
+		reg = <24>;
+		reset-gpios = <&tlmm 77 GPIO_ACTIVE_LOW>;
```
openwrt-ai, line 215: The QCA8081 reset is asserted and released back-to-back with no hold or settle time. `reset-assert-us` / `reset-deassert-us` default to 0 — [`fwnode_mdio.c:95-98`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/mdio/fwnode_mdio.c#L95-L98) only fills them from DT, and `mdio_device_reset()` skips the `fsleep()` when the delay is zero — so the PHY may not be probed reliably.

Both in-tree boards that drive this PHY off the same `tlmm 77` line spell the timings out: [`ipq6010-mango-dvk.dts:234-239`](https://github.com/openwrt/openwrt/blob/c74a3da36e2a85beed5dc3ad6c25a0692fccf86d/target/linux/qualcommax/dts/ipq6010-mango-dvk.dts#L234-L239) and [`ipq6010-xe3-4.dts:256-261`](https://github.com/openwrt/openwrt/blob/c74a3da36e2a85beed5dc3ad6c25a0692fccf86d/target/linux/qualcommax/dts/ipq6010-xe3-4.dts#L256-L261).

```suggestion
		reset-gpios = <&tlmm 77 GPIO_ACTIVE_LOW>;
		reset-assert-us = <10000>;
		reset-deassert-us = <50000>;
```

---

target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
```diff
@@ -0,0 +1,291 @@
...
+	wp-gpios = <&tlmm 63 GPIO_ACTIVE_HIGH>;
+	vmmc-supply = <&reg_sd_vmmc>;
+	bus-width = <4>;
+};
+
+&mdio {
+	status = "okay";
+	pinctrl-0 = <&mdio_pins>;
+	pinctrl-names = "default";
+	reset-gpios = <&tlmm 75 GPIO_ACTIVE_LOW>;
```
openwrt-ai, line 181: The bus-level MDIO reset has no timings either, so it falls back to `DEFAULT_GPIO_RESET_DELAY` = 10 µs asserted and 0 µs post-delay ([`of_mdio.c:219-222`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/mdio/of_mdio.c#L219-L222)) — a thousand times shorter than what the QCA8075 package gets on every other ipq60xx board, and with no settle time before the bus is scanned.

All seven ipq60xx boards that put `reset-gpios` on `&mdio` use the same pair: [`ipq6018-fap650.dts:98-99`](https://github.com/openwrt/openwrt/blob/c74a3da36e2a85beed5dc3ad6c25a0692fccf86d/target/linux/qualcommax/dts/ipq6018-fap650.dts#L98-L99), `ipq6010-xe3-4.dts:239-240`, `ipq6010-wax214.dts:107-108`, `ipq6010-mango-dvk.dts:212-213`, `ipq6018-rbx350.dtsi:175-176`, `ipq6000-ap120c-ax.dts:329-330`, `ipq6018-tplink-eap6xx-outdoor.dtsi:105-106`.

```suggestion
	reset-gpios = <&tlmm 75 GPIO_ACTIVE_LOW>;
	reset-delay-us = <10000>;
	reset-post-delay-us = <50000>;
```

---

target/linux/qualcommax/ipq60xx/base-files/etc/board.d/02_network
```diff
@@ -48,6 +48,9 @@ ipq60xx_setup_interfaces()
 	tplink,eap625-outdoor-hd-v1)
 		ucidef_set_interface_lan "lan" "dhcp"
 		;;
+	wallys,dr6018-v4)
+		ucidef_set_interfaces_lan_wan "lan1 lan2 lan3 lan4" "wan"
+		;;
```
openwrt-ai: nit: this duplicates an existing arm instead of joining it. [Lines 26-32](``https://github.com/openwrt/openwrt/blob/c74a3da36e2a85beed5dc3ad6c25a0692fccf86d/target/linux/qualcommax/ipq60xx/base-files/etc/board.d/02_network#L26-L32``) already group `glinet,gl-ax1800 | jdcloud,re-cs-02 | link,nn6000-v2 | linksys,mr7350 | linksys,mr7500 | yuncore,fap650` under exactly this body; the file's convention is one alphabetically-sorted alternation list per distinct configuration. Adding `wallys,dr6018-v4|\` to that list keeps it to a one-line change.

Same in [`platform.sh:255-257`](``https://github.com/openwrt/openwrt/blob/c74a3da36e2a85beed5dc3ad6c25a0692fccf86d/target/linux/qualcommax/ipq60xx/base-files/lib/upgrade/platform.sh#L255-L257):`` the new block is a bare `nand_do_upgrade "$1"`, which is byte-for-byte the [`glinet,gl-ax1800 | glinet,gl-axt1800 | netgear,rbr350 | netgear,rbs350 | netgear,wax214`](``https://github.com/openwrt/openwrt/blob/c74a3da36e2a85beed5dc3ad6c25a0692fccf86d/target/linux/qualcommax/ipq60xx/base-files/lib/upgrade/platform.sh#L200-L205)`` arm.

---

target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
```diff
@@ -0,0 +1,291 @@
...
+&blsp1_spi1 {
+	pinctrl-0 = <&spi_0_pins>;
+	pinctrl-names = "default";
+	status = "okay";
+
+	flash@0 {
+		#address-cells = <1>;
+		#size-cells = <1>;
+		reg = <0>;
+		compatible = "jedec,spi-nor";
```
openwrt-ai, line 113: nit: the NOR size is stated two different ways for the same board — the commit message says "8 MiB SPI NOR flash", the PR description says "32MB NOR flash". Worth settling before merge so the wiki/device page picks up the right figure. (The DTS itself doesn't pin it down, since the SPI-NOR is left unpartitioned and the size comes from the JEDEC ID at runtime.)

---

openwrt-ai (commented):
Re-reviewed after the six-commit series was squashed into `c74a3da`. The squash plus the rewrite resolves most of what was raised last round: the introduce-then-fix commits are gone, the `IMAGES :=` clobber and the dead `DEVICE_PACKAGES` entries are cleaned up, `DEVICE_MODEL`/`DEVICE_VARIANT` now line up with the DTS `model`, the unused `<dt-bindings/input/input.h>` include and the stray `wallys,dr6018` compatible are dropped, and the NAND now carries a `qcom,smem-part` node so `0:art` and the UBI rootfs actually resolve (`0:art` lowercase is also the right spelling for that parser, which lowercases partition names). The commit message now carries all three required new-device sections.

One correctness issue left, plus two DT reset-timing gaps and two nits — all inline. The earlier thread about the unpartitioned SPI-NOR still stands for `flash@0` only; the NAND half of it is addressed.

The commit message matches the diff, so no commit-check findings this round.

---

okoklets:
Addressed all of the second-round findings in the amended commit (force-pushed, still a single commit):

- Added the missing `wallys,dr6018-v4` arm to the `ath11k/QCN9074/hw1.0/cal-pci-0000:01:00.0.bin` case in `11-ath11k-caldata`, extracting from `0:art` at `0x26800` like the other ipq60xx boards with a QCN9074.
- Added `reset-assert-us`/`reset-deassert-us` to the QCA8081 `reset-gpios` and `reset-delay-us`/`reset-post-delay-us` to the `&mdio` bus `reset-gpios`, matching the timings used elsewhere in the tree.
- Merged the `wallys,dr6018-v4` case arms in `board.d/02_network` and `platform.sh` into the existing matching alternation lists instead of duplicating the body.
- Fixed the NOR flash size: it's 32 MiB (matching the PR description), not the 8 MiB the commit message said — commit message updated, checked against the board.

Rebuilt and confirmed `target/linux/{compile,install}` still succeeds for `wallys_dr6018-v4` after the changes.

target/linux/qualcommax/ipq60xx/base-files/etc/hotplug.d/firmware/11-ath11k-caldata
```diff
@@ -93,6 +93,9 @@ case "$FIRMWARE" in
...
 	yuncore,fap650)
 		caldata_extract "0:art" 0x1000 0x20000
 		;;
```
openwrt-ai: nit: the `02_network` and `platform.sh` duplicates were merged into their alternation lists this round, but this one was left behind — the new arm is byte-for-byte the [`yuncore,fap650` arm](``https://github.com/openwrt/openwrt/blob/065dfba0b20de399ccfae6a24e749ed499acd383/target/linux/qualcommax/ipq60xx/base-files/etc/hotplug.d/firmware/11-ath11k-caldata#L99-L101)`` directly below it, and this file already groups identical bodies alphabetically (`glinet,gl-ax1800|\ glinet,gl-axt1800)`, `linksys,mr7350|\ linksys,mr7500)`, the three `tplink,eap6*-outdoor*`). `wallys` sorts before `yuncore`, so the merge is a one-line change.

The QCN9074 arm below has no twin, so it stays as-is.

```suggestion
	wallys,dr6018-v4|\
	yuncore,fap650)
		caldata_extract "0:art" 0x1000 0x20000
		;;
```

---

target/linux/qualcommax/image/ipq60xx.mk
```diff
@@ -319,6 +319,22 @@ define Device/tplink_eap620-hd-v3
...
+	DEVICE_VARIANT := v4
+	BLOCKSIZE := 128k
+	PAGESIZE := 2048
+	SOC := ipq6010
+	DEVICE_DTS_CONFIG := config@cp01-c4
+	IMAGES += factory.bin
+	IMAGE/factory.bin := append-ubi | qsdk-ipq-factory-nand
+	DEVICE_PACKAGES := ath11k-firmware-qcn9074 kmod-ath11k-pci
+endef
+TARGET_DEVICES += wallys_dr6018-v4
```
openwrt-ai, line 355: nit: no U-Boot environment arm was added for this board. `uboot-envtools` is in the target's [`DEFAULT_PACKAGES`](https://github.com/openwrt/openwrt/blob/065dfba0b20de399ccfae6a24e749ed499acd383/target/linux/qualcommax/Makefile#L14-L19), so `fw_printenv`/`fw_setenv` ship on every image, but [`package/boot/uboot-tools/uboot-envtools/files/qualcommax_ipq60xx`](``https://github.com/openwrt/openwrt/blob/065dfba0b20de399ccfae6a24e749ed499acd383/package/boot/uboot-tools/uboot-envtools/files/qualcommax_ipq60xx#L16-L47``) has no `wallys,dr6018-v4` case, so `/etc/config/ubootenv` comes out empty and both tools fail at runtime. 21 of the 23 existing ipq60xx devices are listed there (only `alfa-network,ap120c-ax` and `qihoo,360v6` are not).

That matters a bit more than usual here because the commit message's install path is entirely U-Boot/TFTP-driven and leans on `ipaddr`/`serverip` — being able to set those from OpenWrt is the normal way users prepare that. Since the NAND uses `qcom,smem-part` (which lowercases partition names, same as `0:art`), the arm would presumably be `ubootenv_add_mtd "0:appsblenv" ...` like the netgear/tplink group.

Is the omission intentional (e.g. no `0:appsblenv` in this board's SMEM table), or just not wired up yet? I can't tell the partition's size from the diff, which is why I'm not suggesting concrete offsets.

---

openwrt-ai (commented):
Re-reviewed the force-pushed single commit `065dfba` against the previously reviewed `c74a3da`. All five third-round findings are genuinely addressed:

- The `wallys,dr6018-v4` arm is now present in the `ath11k/QCN9074/hw1.0/cal-pci-0000:01:00.0.bin` case at `0:art` `0x26800` `0x20000`, matching `cambiumnetworks,xe3-4` / `jdcloud,re-cs-02` / `linksys,mr7500`.
- `reset-delay-us = <10000>` / `reset-post-delay-us = <50000>` on `&mdio` and `reset-assert-us` / `reset-deassert-us` on the QCA8081 node match `ipq6010-mango-dvk.dts` and the other ipq60xx boards exactly, so the 10 µs `DEFAULT_GPIO_RESET_DELAY` fallback no longer applies.
- The `board.d/02_network` and `platform.sh` arms were folded into the existing alternation lists, in alphabetical position in both.
- The NOR size disagreement is resolved — commit message and PR body now both say 32 MiB.

Also spot-checked the parts that did not change: `qcom,package-mode` is legitimately omitted (the qca807x driver defaults `package_mode = NA` to PSGMII, [`qca807x.c:578-582`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/qcom/qca807x.c#L578-L582), checked against v6.18.44 per `KERNEL_PATCHVER:=6.18`); the `pcs-handle = <&uniphy0 0..3>` indices line up with the PHY addresses; `IMAGES += factory.bin` + `IMAGE/factory.bin := append-ubi | qsdk-ipq-factory-nand` is the in-tree spelling (`cmcc_rm2-6` and friends); and `DEVICE_PACKAGES` correctly adds only what is not already in `DEFAULT_PACKAGES` / `ipq60xx/target.mk`. The commit message matches the diff, so no commit-check findings this round. CI is green on `065dfba`.

Two nits inline. The earlier thread about the unpartitioned `flash@0` on the SPI-NOR is still open and unchanged — the NAND half of it remains addressed by the `qcom,smem-part` node.

---

okoklets:
Addressed the two third-round findings in the amended commit (force-pushed, still a single commit):

- Merged the `wallys,dr6018-v4` arm in `11-ath11k-caldata`'s IPQ6018 case into the existing `yuncore,fap650` alternation list (byte-identical body), matching what was already done in `02_network`/`platform.sh`.
- Added a `wallys,dr6018-v4` arm to `package/boot/uboot-tools/uboot-envtools/files/qualcommax_ipq60xx`. Pulled `cat /proc/mtd` from the actual hardware: `0:appsblenv` is a single 0x20000 block, so `ubootenv_add_mtd "0:appsblenv" "0x0" "0x20000" "0x20000"`.

Also, pulling that `/proc/mtd` turned up `mtd0 "spi0.0"` at `0x00800000` = 8 MiB, which is the raw unpartitioned SPI-NOR device. That means the 32 MiB figure from earlier was wrong — the flash is actually 8 MiB, matching what the commit message originally said before round 2. Reverted the commit message and PR description back to 8 MiB and double checked there's no other place mentioning the flash size.

Rebuilt and confirmed `target/linux/{compile,install}` still succeeds for `wallys_dr6018-v4`.

openwrt-ai (commented):

---

target/linux/qualcommax/image/ipq60xx.mk
openwrt-ai: solved by dropping the ipq-wifi package and the extra kmod entries entirely, thanks

---

target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
openwrt-ai: fixed, thanks

---

target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
openwrt-ai: solved by dropping the calibration-variant property and the ipq-wifi board file entirely, thanks

---

target/linux/qualcommax/image/ipq60xx.mk
openwrt-ai: fixed, thanks

---

target/linux/qualcommax/image/ipq60xx.mk
openwrt-ai: fixed, thanks

---

target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
openwrt-ai: fixed by squashing the series into one commit, thanks

---

target/linux/qualcommax/ipq60xx/base-files/etc/hotplug.d/firmware/11-ath11k-caldata
openwrt-ai: fixed, thanks

---

target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
openwrt-ai: fixed, thanks

---

target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
openwrt-ai: fixed, thanks

---

target/linux/qualcommax/ipq60xx/base-files/etc/board.d/02_network
openwrt-ai: fixed, thanks

---

target/linux/qualcommax/dts/ipq6010-dr6018-v4.dts
openwrt-ai: fixed — commit message and PR body now both say 8 MiB, backed by `/proc/mtd`. Thanks.

---

target/linux/qualcommax/ipq60xx/base-files/etc/hotplug.d/firmware/11-ath11k-caldata
openwrt-ai: fixed, thanks

---

target/linux/qualcommax/image/ipq60xx.mk
openwrt-ai: fixed, thanks

---

okoklets:
@florolf can we get some movement on this pls?

PR #24542, opened 2026-08-02
realtek: pcs: RTL931x RX calibration

jonasjelonek:
This series adds some calibration functionality to the RTL931x PCS code, similar to what RTL930x already has. While there are no wonders to expect from this right now, it could make links with PHY and fiber attachment slightly more stable, and seems to improve DAC cable behavior slightly. In other words, the link comes up a bit more reliably and should suffer less from link flapping. At least this applies to DACs <= 2m, for longer ones there is still some work to do.

Tested on: Zyxel XS1930-12HP, Linksys LGS352C, Plasma Cloud PSX28

### DIFF
--- a/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
+++ b/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
@@ -3120,6 +3120,16 @@ static void rtpcs_931x_sds_clear_symerr(struct rtpcs_serdes *sds,
 	}
 }
 
+static int rtpcs_931x_sds_rxeq_leq_set_adapt(struct rtpcs_serdes *sds, bool enable)
+{
+	return rtpcs_sds_write_bits(sds, PAGE_ANA_10G, 0xd, 7, 7, enable ? 0x0 : 0x1);
+}
+
+static int rtpcs_931x_sds_rxeq_leq_set_coef(struct rtpcs_serdes *sds, unsigned int gain)
+{
+	return rtpcs_sds_write_bits(sds, PAGE_ANA_10G, 0xd, 6, 2, gain);
+}
+
 /**
  * rtpcs_931x_sds_reset_leq_dfe() - Reset LEQ + DFE to a baseline.
  *
@@ -3149,6 +3159,16 @@ static int rtpcs_931x_sds_reset_leq_dfe(struct rtpcs_serdes *sds)
 	rtpcs_sds_write(sds, PAGE_ANA_10G_EXT, 0x12, 0xaaa); /* [11:8] VTHN | [7:4] VTHP */
 
 	return 0;
+}
+
+/*
+ * Disable DFE auto-adapt on the companion 5G analog block before
+ * calibrating this lane's 10G LEQ/DFE. Used by the vendor SDK's
+ * PHY-attached and PCB-adapt calibration paths.
+ */
+static int rtpcs_931x_sds_dfe_disable_5g(struct rtpcs_serdes *sds)
+{
+	return rtpcs_sds_write_bits(sds, PAGE_ANA_5G0, 0xf, 12, 6, 0x7f);
 }
 
 static int rtpcs_931x_sds_power(struct rtpcs_serdes *sds, bool power_on)
@@ -3210,7 +3230,6 @@ static int rtpcs_931x_sds_activate(struct rtpcs_serdes *sds)
 	return rtpcs_931x_sds_power(sds, true);
 }
 
-__maybe_unused
 static void rtpcs_931x_sds_rx_reset(struct rtpcs_serdes *sds)
 {
 	if (sds->type != RTPCS_SDS_TYPE_10G)
@@ -3227,6 +3246,29 @@ static void rtpcs_931x_sds_rx_reset(struct rtpcs_serdes *sds)
 	rtpcs_sds_write(sds, PAGE_ANA_MISC, 0x0, 0xc30);
 
 	msleep(50);
+}
+
+/*
+ * rtpcs_931x_sds_rxcal_leq_adapt() - RX calibration for PHY-attached ports.
+ *
+ * Vendor SDK: _phy_rtl9310_leq_adapt(). Used whenever the port has an
+ * external PHY attached, regardless of hw_mode - the PHY performs its own
+ * equalization, so this only resets LEQ and releases it back into
+ * auto-adapt. No TAP/VTH involvement.
+ */
+static void rtpcs_931x_sds_rxcal_leq_adapt(struct rtpcs_serdes *sds)
+{
+	rtpcs_931x_sds_rxeq_leq_set_coef(sds, 0);
+	rtpcs_sds_write_bits(sds, PAGE_ANA_10G, 0xd, 1, 0, 0x0);   /* undocumented */
+	rtpcs_sds_write_bits(sds, PAGE_ANA_10G, 0xd, 13, 13, 0x0); /* undocumented */
+
+	rtpcs_931x_sds_dfe_disable_5g(sds);
+
+	rtpcs_931x_sds_rxeq_leq_set_adapt(sds, false);
+	rtpcs_931x_sds_rx_reset(sds);
+	msleep(10);
+	rtpcs_931x_sds_rxeq_leq_set_adapt(sds, true);
+	msleep(100);
 }
 
 static int rtpcs_931x_sds_get_pll_select(struct rtpcs_serdes *sds, enum rtpcs_sds_pll_type *pll)
@@ -3576,6 +3618,30 @@ static int rtpcs_931x_sds_config_attachment(struct rtpcs_serdes *sds,
 	/* Gating as mentioned above, deactivated here for non-10G and XSGMII */
 	if (!is_10g || hw_mode == RTPCS_SDS_MODE_XSGMII)
 		rtpcs_sds_write_bits(sds, DIGI_1(PAGE_WDIG), 0x1, 0, 0, 0x0);
+
+	return 0;
+}
+
+/*
+ * rtpcs_931x_sds_post_config() - RX calibration, run after power-up.
+ *
+ * Vendor SDK: _phy_rtl9310_rxCali(). Dispatches by attachment, mirroring
+ * the SDK's own split into separate calibration paths per attachment.
+ */
+static int rtpcs_931x_sds_post_config(struct rtpcs_serdes *sds, enum rtpcs_sds_mode hw_mode)
+{
+	if (sds->type != RTPCS_SDS_TYPE_10G)
+		return 0;
+
+	switch (sds->attachment) {
+	case RTPCS_SDS_ATTACH_PHY:
+		rtpcs_931x_sds_rxcal_leq_adapt(sds);
+		break;
+
+	default:
+		/* TODO: fiber/DAC RX calibration */
+		break;
+	}
 
 	return 0;
 }
@@ -4284,6 +4350,7 @@ static const struct rtpcs_sds_ops rtpcs_931x_sds_ops = {
 	.config_hw_mode		= rtpcs_931x_sds_config_hw_mode,
 	.set_hw_mode		= rtpcs_931x_sds_set_mode,
 	.config_attachment	= rtpcs_931x_sds_config_attachment,
+	.post_config		= rtpcs_931x_sds_post_config,
 };
 
 static const struct rtpcs_config rtpcs_931x_cfg = {
--- a/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
+++ b/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
@@ -3089,6 +3089,11 @@ static int rtpcs_931x_sds_fiber_get_symerr(struct rtpcs_serdes *sds,
 	return symerr;
 }
 
+static bool rtpcs_931x_sds_10gr_link_up(struct rtpcs_serdes *sds)
+{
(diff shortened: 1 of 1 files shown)


target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -3229,6 +3391,126 @@ static void rtpcs_931x_sds_rx_reset(struct rtpcs_serdes *sds)
...
+		link_up = rtpcs_931x_sds_10gr_link_up(sds);
+		dev_dbg(dev, "SerDes %u symErr check %d: linkUp=%d symErr=0x%x\n", sds->id,
+			i + 1, link_up, symerr);
+
+		/*
+		 * symErr also reads 0 with no signal at all, not just a clean
+		 * link - don't trust it without link_up confirming there's
+		 * actually something being decoded.
+		 */
+		if (link_up && symerr <= 5) {
```
openwrt-ai: A negative return satisfies `symerr <= 5`, so a failed register read is reported as a successful calibration. [`rtpcs_931x_sds_fiber_get_symerr()`](``https://github.com/openwrt/openwrt/blob/03d47505a36542001ebd055e8ed079ae59b9402f/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L3069-L3090``) propagates whatever `rtpcs_sds_read_bits()` returns (negative errno on failure) and returns `-EINVAL` for unhandled modes. `symerr` is also still `-1` from its initializer if that path is ever reached without a successful read. Same value is printed with `%x` two lines above, which would render a negative errno as a huge hex number.

```suggestion
		if (link_up && symerr >= 0 && symerr <= 5) {
```

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -3229,6 +3391,126 @@ static void rtpcs_931x_sds_rx_reset(struct rtpcs_serdes *sds)
...
+ * (3 symbol error rechecks, 150ms delay, exact-0 target). symErr's field
+ * (8 bits wide) has been observed reading a saturated 0xff right after
+ * rx_reset(), which looks like "link hasn't relocked yet" rather than a
+ * genuine error count. Thus, recheck more times with more patience per
+ * check instead (no reset in between, so a settling link isn't interrupted),
+ * and tolerate a small nonzero symbol-error count rather than requiring
+ * exactly 0.
+ */
+static void rtpcs_931x_sds_rxcal_fiber_adapt(struct rtpcs_serdes *sds)
+{
```
openwrt-ai, line 3442: If every `rtpcs_931x_sds_rxeq_vth_get()` call fails, `samples` stays 0, the code warns — and then still calls `rtpcs_931x_sds_rxeq_vth_set_value(sds, 0, 0)` a few lines below, locking VTH at `0x0/0x0`. That's strictly worse than doing nothing: [`rtpcs_931x_sds_reset_leq_dfe()`](https://github.com/openwrt/openwrt/blob/03d47505a36542001ebd055e8ed079ae59b9402f/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L3299) has just installed the `0xa/0xa` baseline, and commit cf5f87c's own message says a locked VTH other than `0xa` made the link behave worse.

Defaulting the locals to the reset baseline keeps the read-failure path at the known-good value:

```suggestion
	unsigned int vth_p = 0xa, vth_n = 0xa, sum_p = 0, sum_n = 0;
```

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -3136,21 +3278,42 @@ static void rtpcs_931x_sds_clear_symerr(struct rtpcs_serdes *sds,
  */
 static int rtpcs_931x_sds_reset_leq_dfe(struct rtpcs_serdes *sds)
 {
-	rtpcs_sds_write_bits(sds, PAGE_ANA_10G, 0xd, 6, 0, 0x0);	/* [6:2] LEQ gain */
-	rtpcs_sds_write_bits(sds, PAGE_ANA_10G, 0xd, 7, 7, 0x1);	/* LEQ manual 1=true,0=false */
+	rtpcs_931x_sds_rxeq_leq_set_coef(sds, 0);
+	/* bits [1:0] are undocumented but part of the known-good reset value */
+	rtpcs_sds_write_bits(sds, PAGE_ANA_10G, 0xd, 1, 0, 0x0);
+	rtpcs_931x_sds_rxeq_leq_set_adapt(sds, false);
```
openwrt-ai: Commit f096a59 reorders VTH/TAP so manual mode is forced *before* the reset values are written, but LEQ keeps the old order: the gain is zeroed here and LEQ adapt is only disabled afterwards. Doesn't the same race apply? On a PHY-attached port, [`rtpcs_931x_sds_rxcal_leq_adapt()`](``https://github.com/openwrt/openwrt/blob/03d47505a36542001ebd055e8ed079ae59b9402f/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L3412-L3413``) deliberately returns with LEQ adapt *enabled*, and [`rtpcs_931x_sds_config_attachment()` calls `reset_leq_dfe()`](https://github.com/openwrt/openwrt/blob/03d47505a36542001ebd055e8ed079ae59b9402f/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L3779) on every reconfiguration — so on a re-config the adapt engine can be live on `0xd[6:2]` at the moment `0` is written into it, exactly the situation the commit message describes for TAP1-4.

If that's intentional (e.g. LEQ adapt is known to be quiesced by the preceding `0xe[13:11] = 0` write), a note in the commit message would help. Otherwise:

```suggestion
	rtpcs_931x_sds_rxeq_leq_set_adapt(sds, false);
	rtpcs_931x_sds_rxeq_leq_set_coef(sds, 0);
	/* bits [1:0] are undocumented but part of the known-good reset value */
	rtpcs_sds_write_bits(sds, PAGE_ANA_10G, 0xd, 1, 0, 0x0);
```

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -2064,7 +2075,7 @@ static int rtpcs_930x_sds_set_debug(struct rtpcs_serdes *sds, unsigned int debug
 	return rtpcs_sds_write_bits(sds, PAGE_ANA_COM, 0x06, 11, 6, debug_sel); /* RX_DEBUG_SEL */
 }
 
-static int rtpcs_930x_sds_rxcal_dcvs_set_adapt(struct rtpcs_serdes *sds, unsigned int dcvs_id,
+static int rtpcs_930x_sds_rxeq_dcvs_set_adapt(struct rtpcs_serdes *sds, unsigned int dcvs_id,
 					       bool enable)
```
openwrt-ai: nit: the `rxcal_` → `rxeq_` rename shortens each name by one character, so every multi-line prototype's continuation is now one column past the opening parenthesis instead of aligned to it (checkpatch `Alignment should match open parenthesis`).

```suggestion
static int rtpcs_930x_sds_rxeq_dcvs_set_adapt(struct rtpcs_serdes *sds, unsigned int dcvs_id,
					      bool enable)
```

Same one-column drift at the other eight renamed prototypes: lines 2091/2092, 2104/2105, 2152/2153, 2194/2195, 2210/2211, 2238/2239, 2249/2250 and 2300/2301 — each continuation line needs one space removed (2210 and 2300 need `4 tabs + 7 spaces` instead of `5 tabs`).

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -3120,6 +3124,144 @@ static void rtpcs_931x_sds_clear_symerr(struct rtpcs_serdes *sds,
...
+ * Vendor SDK: _phy_rtl9310_dbg_set(). Selects which lane of the even/odd
+ * pair feeds the shared WDIG debug readback register, then selects
+ * dbg_sel within that lane. Must run before reading any rxeq_*_get().
+ */
+static int rtpcs_931x_sds_set_debug(struct rtpcs_serdes *sds, unsigned int dbg_sel)
+{
+	struct rtpcs_serdes *even_sds = rtpcs_sds_get_even(sds);
+	int ret;
+
+	ret = rtpcs_sds_write(even_sds, PAGE_WDIG, 0x2, 75 + (sds->id % 2));
```
openwrt-ai: nit: decimal register value in a file that otherwise writes SerDes register values in hex — the sibling [`rtpcs_930x_sds_set_debug()`](https://github.com/openwrt/openwrt/blob/03d47505a36542001ebd055e8ed079ae59b9402f/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L2067) does the identical even/odd lane select as `(sds == even_sds) ? 0x2f : 0x31`.

```suggestion
	ret = rtpcs_sds_write(even_sds, PAGE_WDIG, 0x2, 0x4b + (sds->id % 2));
```

---

openwrt-ai (commented):
Reviewed 8 commits; all commit messages match their diffs. Inline comments cover two error-path issues in `rtpcs_931x_sds_rxcal_fiber_adapt()`, one ordering question in `rtpcs_931x_sds_reset_leq_dfe()`, and two nits.

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -3229,6 +3391,131 @@ static void rtpcs_931x_sds_rx_reset(struct rtpcs_serdes *sds)
...
+
+	for (i = 0; i < 8; i++) {
+		rtpcs_931x_sds_clear_symerr(sds, RTPCS_SDS_MODE_10GBASER);
+		msleep(300);
+		symerr = rtpcs_931x_sds_fiber_get_symerr(sds, RTPCS_SDS_MODE_10GBASER);
+		link_up = rtpcs_931x_sds_10gr_link_up(sds);
+		dev_dbg(dev, "SerDes %u symErr check %d: linkUp=%d symErr=0x%x\n", sds->id,
+			i + 1, link_up, symerr);
+
+		dev_dbg(dev, "SerDes %u symErr check %d: 0x%x\n", sds->id, i + 1, symerr);
```
openwrt-ai: Duplicate debug print: this logs `symErr` a second time, immediately after the `linkUp=%d symErr=0x%x` line two lines above logs the same value for the same iteration.

Looks like a rebase artifact — in the previous revision of this branch, 386b889b70 (now 81ff3284 "gate symErr success on actual link status") *replaced* this line with the `linkUp=` variant; after the amend it adds the new line but no longer removes the old one, so both survive into the final tree.

```suggestion
```

---

openwrt-ai (commented):
Re-reviewed the 7 amended commits since 03d4750. All five findings from the previous round are addressed (negative-`symerr` guard, VTH read-failure fallback to the `0xa` baseline, LEQ adapt-disable ordering in `reset_leq_dfe()`, prototype continuation alignment across the renamed 930x accessors, hex lane-select in `set_debug()`).

One new issue introduced by the amend — see the inline comment on the duplicated `dev_dbg()` in `rtpcs_931x_sds_rxcal_fiber_adapt()`.

## Commit checks

- 9a8c87b1 "realtek: pcs: rtl931x: lock manual mode before reset values" — the diff now also hoists `rtpcs_931x_sds_rxeq_leq_set_adapt(sds, false)` above the LEQ coefficient write, but the message only talks about "VTH/TAP0-4" and motivates the change purely through `rxcal_fiber_adapt()` leaving TAP1-4 in live adapt. Worth a sentence covering the LEQ case too (`rxcal_leq_adapt()` likewise returns with LEQ adapt enabled), so the extra hunk isn't undocumented.

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -3130,6 +3152,96 @@ static int rtpcs_931x_sds_rxeq_leq_set_coef(struct rtpcs_serdes *sds, unsigned i
...
+	if (ret < 0)
+		return ret;
+	usleep_range(1000, 2000);
+
+	val = rtpcs_sds_read_bits(sds, PAGE_WDIG, 0x14, 7, 0);
+	if (val < 0)
+		return val;
+
+	*vth_p = FIELD_GET(GENMASK(3, 0), val);
+	*vth_n = FIELD_GET(GENMASK(7, 4), val);
```
plappermaul, line 3264: This sequence cries for a lock as adjacent serdes can overwrite the debug setting. I assume the result from WDIG ist then inconsistent.

plappermaul (changes requested):

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
jonasjelonek: The PCS still uses a global lock, so only one SerDes at a time can be touched/worked on. Thus it shouldn't be possible right now. I'm fine though to add a comment here to not forget that in the future.

Is this global lock still a good approach? Probably not. For several reasons this might move into a lock per-adjacent-SerDes. But I think that's something to address in another patch series.

plappermaul: Thanks fpr the clarification. Totally forgot that.

plappermaul (approved):

PR #24549, opened 2026-08-03
realtek: correct poe port mapping for EnGenius EWS2910P v1

DamisDorgon:
Discovered that HW v1.0 has a different poe port mapping than HW v3.0. Observed vendor firmware doing this on the serial connection between the SoC and microcontroller: 

```
72854978-72858312 TX 1D 04 00 03 FF FF FF FF FF FF FF 1D
72941335-72944669 RX 1D 04 00 00 FF FF FF FF FF FF FF 1A
73655168-73658502 TX 1D 05 01 02 FF FF FF FF FF FF FF 1E
73741604-73744938 RX 1D 05 01 00 FF FF FF FF FF FF FF 1C
74455359-74458693 TX 1D 06 02 01 FF FF FF FF FF FF FF 1F
74541873-74545207 RX 1D 06 02 00 FF FF FF FF FF FF FF 1E
75255550-75258884 TX 1D 07 03 00 FF FF FF FF FF FF FF 20
75342142-75345476 RX 1D 07 03 00 FF FF FF FF FF FF FF 20
76055740-76059074 TX 1D 08 04 07 FF FF FF FF FF FF FF 29
76142411-76145745 RX 1D 08 04 00 FF FF FF FF FF FF FF 22
76855931-76859265 TX 1D 09 05 06 FF FF FF FF FF FF FF 2A
76945597-76948931 RX 1D 09 05 00 FF FF FF FF FF FF FF 24
77656538-77659872 TX 1D 0A 06 05 FF FF FF FF FF FF FF 2B
77742948-77746282 RX 1D 0A 06 00 FF FF FF FF FF FF FF 26
78456729-78460063 TX 1D 0B 07 04 FF FF FF FF FF FF FF 2C
78543217-78546551 RX 1D 0B 07 00 FF FF FF FF FF FF FF 28

```

Remapped the PoE ports to match.

Tested on EWS2910P v1.


Djfe:
please move https://github.com/openwrt/openwrt/blob/5feee05dbbb6d49b0011e32abc87616666f44df7/target/linux/realtek/dts/rtl8380_engenius_ews2910p.dtsi#L252-L260

to this file https://github.com/openwrt/openwrt/blob/5feee05dbbb6d49b0011e32abc87616666f44df7/target/linux/realtek/dts/rtl8380_engenius_ews2910p-v3.dts#L3

DamisDorgon:
@Djfe done

target/linux/realtek/dts/rtl8380_engenius_ews2910p-v1.dts
```diff
@@ -16,3 +16,13 @@
 	compatible = "engenius,ews2910p-v1-pse", "realtek,pse-mcu-gen1";
 	current-speed = <19200>;
 };
+
+/* Copper ports lan1..lan8 (phy8..phy15) -> MCU PoE ports 0..7 */
```
openwrt-ai: The comment was copied verbatim from the .dtsi but no longer describes what the v1 block does — the whole point of this commit is that lan1..lan8 do *not* map to MCU PoE ports 0..7 in order on v1, they map to 3,2,1,0,7,6,5,4. As written it contradicts the eight lines directly below it.

```suggestion
/* Copper ports lan1..lan8 (phy8..phy15) -> MCU PoE ports 3,2,1,0,7,6,5,4 */
```

(The identical comment in `rtl8380_engenius_ews2910p-v3.dts:20` is correct as-is, since v3 really is the identity mapping.)

---

openwrt-ai (commented):
## Commit checks

- `c19f6dd6` "realtek: correct poe port mapping for EnGenius EWS2910P v1" — the body ends with `Revised: moved the previous poe port mapping from the dtsi to the v3 dts.` That is PR-iteration metadata (a response to the review request in [this comment](https://github.com/openwrt/openwrt/pull/24549#issuecomment-5163896937)), not a description of the change, and it will read oddly in `git log` once merged. Either drop it and fold the fact into a normal sentence — e.g. "The v3 mapping is unchanged and is moved from the shared .dtsi into `rtl8380_engenius_ews2910p-v3.dts`, since the mapping is now per-variant." — or move it below the `---` line so `git am` strips it. Worth doing because the subject only mentions v1 while the diff also touches the v3 .dts and the shared .dtsi; a plain sentence in the body covers that scope.

Otherwise the change looks right. Cross-checked the decoded traces in the commit message against the new phandles: the frames are `1D <seq> <arg1> <arg2> FF*7 <checksum>` with checksum = sum of the preceding bytes mod 256 (`0x1D+0x04+0x00+0x03 + 7*0xFF = 0x71D` → `0x1D`, matching the first line), giving the permutation 0↔3, 1↔2, 4↔7, 5↔6. That is an involution, so it yields the same wiring regardless of which of `arg1`/`arg2` is the PHY index and which is the MCU PoE port — the v1 mapping in this diff matches it either way. The `pses` property is the correct binding for this ([`ethernet-phy.yaml:182`](https://github.com/gregkh/linux/blob/v6.18.41/Documentation/devicetree/bindings/net/ethernet-phy.yaml#L182), the series the realtek target builds against), `PSE_PI(n)` gives `pse_pi<n>` a matching `reg = <n>`, and the `&phyN { pses = <&pse_piM>; };` per-variant style matches in-tree neighbours such as `rtl9302_zyxel_xmg1915-10ep.dts`, which already uses a non-contiguous mapping.


---

openwrt-ai (commented):
Reviewed 1 new commit (force-push `c19f6dd6` → `cba24953`); no new issues found.

The only code change since the last review is the comment above the v1 mapping, corrected from `-> MCU PoE ports 0..7` (stale, carried over from the v3 block) to `-> MCU PoE ports 3,2,1,0,7,6,5,4`. That now matches the phandles below it and the permutation decoded from the traces in the commit message. The dtsi tail is clean after the move, and `rtl8380_engenius_ews2910p-v3.dts` keeps the original `0..7` mapping verbatim.

## Commit checks

- `cba24953` "realtek: correct poe port mapping for EnGenius EWS2910P v1" — the message is unchanged from the previous push, so the `Revised: moved the previous poe port mapping from the dtsi to the v3 dts.` line is still in the body. Repeating the earlier note rather than re-arguing it: that is PR-iteration metadata, not a description of the change, and it will read oddly in `git log`. Either fold it into a plain sentence covering the wider scope — e.g. "The v3 mapping is unchanged and moves from the shared .dtsi into `rtl8380_engenius_ews2910p-v3.dts`, since the mapping is now per-variant." — or move it below the `---` line so `git am` strips it. Only the message needs touching; the diff itself looks right.

---

jonasjelonek:
As the AI suggests, please drop the "Revised ..." from the commit message.

DamisDorgon:
Amusing anecdote, I got another EWS2910P (also marked hw v1) and it is showing a different poe port mapping.

Instead of an ST 32F100C8, it has a Nuvoton microcontroller (M0516):

  `[   21.934109] realtek-pse-mcu-uart serial0-0: Nuvoton M05xx LAN MCU, BCM59121 (id 0xe121), 8 ports across 1 PSE chip(s)`

Link shows on lan5, poe port shows power delivery on lan4.

The matching line on the other hw v1 says:

`[   21.923045] realtek-pse-mcu-uart serial0-0: ST Micro ST32F100 MCU, BCM59111 (id 0xe111), 8 ports across 255 PSE chip(s)`

(it has two PSE chips, not 255).

jonasjelonek:
Great ... how does realtek-poe behave there? I would assume it also has a wrong mapping on those devices.

DamisDorgon:
I'll try realtek-poe and the original vendor firmware

update:

Interestingly, realtek-poe gets it wrong differently.  With this commit, the ST 32F100C8 mcu gets them correctly (lan1->lan8, left to right). With this commit, on the Nuvoton mcu, `ethtool --show-pse *` shows lan5, lan6, lan7, lan8, lan4, lan3, lan2, lan1 (left to right). With realtek-poe, `ubus call poe info` shows lan8, lan7, lan6, lan5, lan1, lan2, lan3, lan4 (left to right).

with realtek-poe, i can remap the ports with the option id in /etc/config/poe. e.g.:

```
config global
        option budget '60'

config port
        option name 'lan1'
        option id '8'
        option enable '1'
        option poe_plus '1'
        option priority '2'

config port
        option name 'lan2'
        option id '7'
        option enable '1'
        option poe_plus '1'
        option priority '2'

config port
        option name 'lan3'
        option id '6'
        option enable '1'
        option poe_plus '1'
        option priority '2'

config port
        option name 'lan4'
        option id '5'
        option enable '1'
        option poe_plus '1'
        option priority '2'

config port
        option name 'lan5'
        option id '1'
        option enable '1'
        option poe_plus '1'
        option priority '2'

config port
        option name 'lan6'
        option id '2'
        option enable '1'
        option poe_plus '1'
        option priority '2'

config port
        option name 'lan7'
        option id '3'
        option enable '1'
        option poe_plus '1'
        option priority '2'

config port
        option name 'lan8'
        option id '4'
        option enable '1'
        option poe_plus '1'
        option priority '2'
```

with the PoE device in lan5, I get:
```

# ubus call poe info
{
        "firmware": "v23.1",
        "mcu": "Nuvoton M05xx LAN Microcontroller",
        "budget": 60,
        "consumption": 3.5999999046325684,
        "ports": {
                "lan5": {
                        "priority": 2,
                        "mode": "PoE+",
                        "status": "Delivering power",
                        "consumption": 3.7000000476837158
                },
                "lan6": {
                        "priority": 2,
                        "mode": "PoE+",
                        "status": "Searching"
                },
                "lan7": {
                        "priority": 2,
                        "mode": "PoE+",
                        "status": "Searching"
                },
                "lan8": {
                        "priority": 2,
                        "mode": "PoE+",
                        "status": "Searching"
                },
                "lan4": {
                        "priority": 2,
                        "mode": "PoE+",
                        "status": "Searching"
                },
                "lan3": {
                        "priority": 2,
                        "mode": "PoE+",
                        "status": "Searching"
                },
                "lan2": {
                        "priority": 2,
                        "mode": "PoE+",
                        "status": "Searching"
                },
                "lan1": {
                        "priority": 2,
                        "mode": "PoE+",
                        "status": "Searching"
                }
        }
}
```

DamisDorgon:
The vendor firmware:

```
EWS2910P# show version
Loader Version   : 2011.12.48956-1.Jan 08 2019 - 05:30:10
Loader Date      : Jan 08 2019 - 05:30:10
Firmware Version : v1.07.43-c1.9.37
Firmware Date    : Mar 11 2021 - 18:08:21
```
 
reports the right port plugged, e.g.

```
EWS2910P# show poe

PowerBudget               :    55(W)  
Allocated Power           :   4.4(W)  
  Port    |  State  |  Priority  | Power Limit Type | User Powerlimit(W) |   Status       | Voltage(V) | Current(mA) | Power(W) | Class
----------+---------+------------+------------------+--------------------|----------------|------------|-------------|----------|---------
  gi1        enable     Low           Auto-class                           Searching        
  gi2        enable     Low           Auto-class                           Searching        
  gi3        enable     Low           Auto-class                           Searching        
  gi4        enable     Low           Auto-class                           Delivering          53.8           66         3.5        4 
  gi5        enable     Low           Auto-class                           Searching        
  gi6        enable     Low           Auto-class                           Searching        
  gi7        enable     Low           Auto-class                           Searching        
  gi8        enable     Low           Auto-class                           Searching        
```

I'm going to sniff the SoC to microcontroller serial to see how it maps poe ports...

DamisDorgon:
Here's what the new EWS2910P v1's vendor firmware does with regard to opcode 0x1D messages:

```
72948507-72951841 TX 1D 04 00 07 FF FF FF FF FF FF FF 21
73034857-73038191 RX 1D 04 00 00 FF FF FF FF FF FF FF 1A
73748700-73752034 TX 1D 05 01 06 FF FF FF FF FF FF FF 22
73835123-73838457 RX 1D 05 01 00 FF FF FF FF FF FF FF 1C
74548893-74552227 TX 1D 06 02 05 FF FF FF FF FF FF FF 23
74635388-74638722 RX 1D 06 02 00 FF FF FF FF FF FF FF 1E
75349086-75352420 TX 1D 07 03 04 FF FF FF FF FF FF FF 24
75435654-75438988 RX 1D 07 03 00 FF FF FF FF FF FF FF 20
76149279-76152613 TX 1D 08 04 00 FF FF FF FF FF FF FF 22
76235946-76239280 RX 1D 08 04 00 FF FF FF FF FF FF FF 22
76949471-76952805 TX 1D 09 05 01 FF FF FF FF FF FF FF 25
77036211-77039545 RX 1D 09 05 00 FF FF FF FF FF FF FF 24
77750081-77753415 TX 1D 0A 06 02 FF FF FF FF FF FF FF 28
77836477-77839811 RX 1D 0A 06 00 FF FF FF FF FF FF FF 26
78550274-78553608 TX 1D 0B 07 03 FF FF FF FF FF FF FF 2B
78636742-78640076 RX 1D 0B 07 00 FF FF FF FF FF FF FF 28
```

(port index number -> PSE output index number)

Nuvoton:
0 -> 7
1 -> 6
2 -> 5
3 -> 4
4 -> 0
5 -> 1
6 -> 2
7 -> 3

ST 32F100:
0 -> 3
1 -> 2
2 -> 1
3 -> 0
4 -> 7
5 -> 6
6 -> 5
7 ->4

I only have one of each microcontroller, so I can't confirm, but ... maybe it goes by microcontroller? PSE? The only querying command that is sent before the 0x1D (Set Port Mapping) commands is 0x20 (Get System Info), so the mapping must be based on something that is returned from that command: 

```
71348121-71351455 TX 20 02 FF FF FF FF FF FF FF FF FF 19
71434300-71437634 RX 20 02 00 08 00 E1 21 17 01 03 01 48
```

which decodes as:

- mode: 0
- max ports: 8
- port_map: 0 (mapping not set)
- device_id: e121 (BCM59121)
- version: (0x17) 23.1 (including version_ext)
- mcu_type: 01 ("Nuvoton M05xx LAN")
- system_status: 03
-- Config modified, system is running with unsaved configuration changes
-- MCU was reset
-- the "remote enable" signal is not active
-- output pairing is disabled

jonasjelonek:
In EnGenius' world maybe. I had switches with either MCU but the port mapping was "normal". 
Maybe this could be read from the MCU at runtime (I'm also experimenting with such a topology detection for another device) but I think it crosses boundaries of what PSE-PD is capable of right now. And having another driver-internal mapping for that purpose is bad IMO.

The "set port mapping" command seems somewhat dangerous, also `realtek-poe` avoided to implement it. And it doesn't solve the problem of having apparently different configurations on same hardware revisions.

DamisDorgon:
I have another v1 on its way. Should arrive later today, I'll see what it looks like.

DamisDorgon:
The realtek-poe corrected mapping matches what the vendor does (which i realized after i wrote out the mapping on paper and subtracted 1's appropriately).

jonasjelonek:
But I don't see any real mapping in `realtek-poe`, only setting 'enable port mapping' to false during startup.

DamisDorgon:
I was referring only to my constructed /etc/config/poe, quoted above. There isn't, as far as I recall, anything provided by realtek-poe in the way of an /etc/config/poe.

openwrt-ai (commented):
Reviewed 1 new commit (force-push `cba24953` → `cb6e6d50`); no new issues found.

The tree contents are byte-identical to the previously reviewed revision — `git diff cba24953..cb6e6d50` is empty. The only delta is in the commit message: the `Revised: moved the previous poe port mapping from the dtsi to the v3 dts.` line is gone, which is what [@jonasjelonek asked for](https://github.com/openwrt/openwrt/pull/24549#issuecomment-5175656296). That resolves the only item from my previous two reviews, so no Commit checks section this time. All 28 checks on `cb6e6d50` are green.

Note for the maintainer, not a request for a change to this diff: the [Nuvoton-MCU v1 unit reported in the thread](https://github.com/openwrt/openwrt/pull/24549#issuecomment-5177882776) has a third mapping (`0..7 -> 7,6,5,4,0,1,2,3`), which a per-`compatible` DTS mapping cannot express. That is under active discussion between @DamisDorgon and @jonasjelonek and is out of scope here — flagging only so it isn't lost if this merges as-is, since the v1 `.dts` would then be correct for ST-MCU units and wrong for Nuvoton ones.

---

DamisDorgon:
@jonasjelonek the other v1 arrived, it also has:

```
[   22.993821] realtek-pse-mcu-uart serial0-0: ST Micro ST32F100 MCU, BCM59111 (id 0xe111), 8 ports across 255 PSE chip(s)
```
and its poe port map is consistent with this patch.

Seems like there are a few choices:

- split hw versions further, e.g. v1-stm32 and v1-nuvoton
- make the driver aware of the mappings that exist and detect/configure correctly
- provide an interface for userspace to specify a mapping

DamisDorgon:
From my tracing of the board, the Nuvoton and ST32F100C8 are not pin compatible. The mapping is going to be dependent on the layout of the board and the board is going to be designed to accommodate the microcontroller used. The vendor firmware needs to be dealing with these differences, and they can only be depending on the results of command 0x20 (Get System Info). It seems plausible to have a table of mappings indexed by the board_name, and mcu and pse chips found with 0x20.

jonasjelonek:
> * split hw versions further, e.g. v1-stm32 and v1-nuvoton

There is nothing else to distinguish these two, kind of a board stepping, etc.? Maybe on the PCB inside? That would make this solution more reasonable. Otherwise it'll be quite hard for users to find out in the first place which variant they have. If there is a difference to stick with, this might be the most correct way to handle this.

> * make the driver aware of the mappings that exist and detect/configure correctly

> It seems plausible to have a table of mappings indexed by the mcu and pse chips found with 0x20.

Putting any hardcoded mappings which are highly dependent on the board layout in the driver is the wrong way IMO. This totally defeats the purpose of having the DT description in the first place. A match for the MCU type alone won't work since other boards have a normal mapping with the same MCU.

I'm still undecided about having an additional in-driver mapping layer, since this adds more complexity and kind of invalidates the DT description since it's de facto wrong then.

DamisDorgon:
There are no obvious distinguishing marks on the outside. All three v1 devices say: `HW: v 1.0.0` on the sticker.

If you open the case, rather than looking for labels or markings, you can just look at the MCU and PSE parts. There isn't any "stepping" marking. I've seen (both st32f100):

```
7016A1067008
VER: 1.00
```

and (nuvoton)

```
7016A143500g
VER: 1.00
```

Maybe rely on the device tree mapping, with some special case overrides in case of particular boards (e.g. compatible = "engenius,ews2910p-v1")?

jonasjelonek:
Sorry for the late response. But still no final decision on this from my side about what's the best way forward. Separating this with different device recipes here in OpenWrt seems the correct way to me, then we just have two different DTS. As said, hardcoding any mappings in the driver sounds wrong to me. In case there is a third variant with another MCU + mapping, this is getting even more annoying. Of course this is not great from user perspective, either having to open the case to see which MCU sits on the board, or figure it out somehow by trial-and-error or from the vendor firmware.

No nasty quirks in the driver needed for this. I would really like to avoid such because I'm still fighting the upstream AI bot to get the initial driver merged and it'll will likely strain my patience also on follow-up patches.

DamisDorgon:
@jonasjelonek so, your request is that I split v1 into v1-stm and v1-nuvoton and fix up the mapping in the respective dts files? 

It appears from ebay, that there is also a HW v2.0.0 in the world. I made an offer on one so-labelled, but the seller rejected my lowball offer, and I am not sure how much I really want to invest in that exercise, despite my curiosity. ;-)

jonasjelonek:
> @jonasjelonek so, your request is that I split v1 into v1-stm and v1-nuvoton and fix up the mapping in the respective dts files?

It's at least my proposal open for discussion on how to solve this the "best" way.

DamisDorgon:
It does seem rather brittle to embed the mapping in the device tree, and has scalability problems if we must split outwardly indistinguishable devices by internal component variants, potentially more and more as variants are discovered. I think perhaps a better solution is to expose some interface to userspace to remap the poe ports, and provide sysfs files from the driver to indicate which mcu and pse is detected, and let a board.d file sort out the mapping. 

That interface not currently existing, I am about to force-push a dts split for ews2910p v1 along the lines you propose above.

target/linux/realtek/dts/rtl8380_engenius_ews2910p-v1-nuvoton.dts
```diff
@@ -0,0 +1,28 @@
...
+	compatible = "openwrt,uimage";
+	openwrt,ih-magic = <0x03802910>;
+};
+
+&pse {
+	compatible = "engenius,ews2910p-v1-pse", "realtek,pse-mcu-gen1";
+	current-speed = <19200>;
+};
+
+/* Copper ports lan1..lan8 (phy8..phy15) -> MCU PoE ports 3,2,1,0,7,6,5,4 */
```
openwrt-ai: The comment was copied from the stm variant and contradicts the phandles right below it: this file assigns `pi7, pi6, pi5, pi4, pi0, pi1, pi2, pi3`, i.e. `7,6,5,4,0,1,2,3`, not `3,2,1,0,7,6,5,4`.

The phandles themselves are right — they match the Nuvoton trace decoded in [this comment](https://github.com/openwrt/openwrt/pull/24549#issuecomment-5177882776) (`0->7, 1->6, 2->5, 3->4, 4->0, 5->1, 6->2, 7->3`), and that mapping is also consistent with the mis-labelled `ethtool --show-pse` output reported [here](https://github.com/openwrt/openwrt/pull/24549#issuecomment-5176486254) when the ST mapping was applied to a Nuvoton unit. Unlike the ST permutation, the Nuvoton one is not an involution, so the comment is the only thing that needs fixing.

```suggestion
/* Copper ports lan1..lan8 (phy8..phy15) -> MCU PoE ports 7,6,5,4,0,1,2,3 */
```

---

target/linux/realtek/image/rtl838x.mk
```diff
@@ -104,16 +104,27 @@ define Device/datto_l8
...
-define Device/engenius_ews2910p-v1
+define Device/engenius_ews2910p-v1-nuvoton
   $(Device/engenius_ews2910p)
   SOC := rtl8380
   DEVICE_MODEL := EWS2910P
-  DEVICE_VARIANT := v1
+  DEVICE_VARIANT := v1-nuvoton
+  DEVICE_PACKAGES += kmod-pse-realtek-mcu-uart
+  UIMAGE_MAGIC := 0x03802910
+  SUPPORTED_DEVICES += engenius,ews2910p
```
openwrt-ai: Renaming the recipe drops the sysupgrade path for every EWS2910P v1 already running OpenWrt. Those units report `engenius,ews2910p-v1` as `board_name` (first compatible of the old DTS), and after this change no image lists that string in `SUPPORTED_DEVICES`, so the metadata check refuses the upgrade and users need `-F` — which also wipes config. The extra `engenius,ews2910p-v1` compatible kept in the new DTS files doesn't help here, since `board_name` only ever takes the first compatible.

In-tree precedent for exactly this situation is [d205878ede "realtek: rtl838x: rename GS1900 series v1/v2 to A1/B1"](https://github.com/openwrt/openwrt/commit/d205878ede886dd2e00b31e4f3c8444b16ae16be) ("Provide compatibles for seamless sysupgrade"), which carries the old name on *both* new recipes — see [rtl838x.mk:437](https://github.com/openwrt/openwrt/blob/8b3545a3e8d3b2f50307aa393569b5dfe911dcb7/target/linux/realtek/image/rtl838x.mk#L437) and [rtl838x.mk:447](https://github.com/openwrt/openwrt/blob/8b3545a3e8d3b2f50307aa393569b5dfe911dcb7/target/linux/realtek/image/rtl838x.mk#L447).

Same change applies to line 125 in the `engenius_ews2910p-v1-stm` block.

```suggestion
  SUPPORTED_DEVICES += engenius,ews2910p engenius,ews2910p-v1
```

---

target/linux/realtek/dts/rtl8380_engenius_ews2910p-v1-nuvoton.dts
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+
+#include "rtl8380_engenius_ews2910p.dtsi"
+
+/ {
+	compatible = "engenius,ews2910p-v1-nuvoton", "engenius,ews2910p-v1", "realtek,rtl838x-soc";
+	model = "EnGenius EWS2910P v1 nuvoton";
```
openwrt-ai: nit: the user-visible variant name is spelled three different ways for the same board — `model = "EnGenius EWS2910P v1 nuvoton"` here, `DEVICE_VARIANT := v1-nuvoton` at [rtl838x.mk:111](https://github.com/openwrt/openwrt/blob/8b3545a3e8d3b2f50307aa393569b5dfe911dcb7/target/linux/realtek/image/rtl838x.mk#L111), and both lowercase a vendor name that is written "Nuvoton" (likewise "stm" → ST / STMicro in the sibling file). The in-tree form for a qualifier appended to a revision is parenthesised and properly cased, e.g. `DEVICE_VARIANT := v2 (16M)` / `v1 (IL)`, and parenthesised `model` strings are common too.

Suggest `v1 (Nuvoton)` / `v1 (STMicro)` consistently across all four places: this line, `rtl8380_engenius_ews2910p-v1-stm.dts:7`, and `rtl838x.mk:111` / `:122`. The DTS filenames and compatibles can stay as they are.

```suggestion
	model = "EnGenius EWS2910P v1 (Nuvoton)";
```

---

target/linux/realtek/dts/rtl8380_engenius_ews2910p-v1-nuvoton.dts
```diff
@@ -0,0 +1,28 @@
...
+};
+
+&firmware_partition_1 {
+	compatible = "openwrt,uimage";
+	openwrt,ih-magic = <0x03802910>;
+};
+
+&pse {
+	compatible = "engenius,ews2910p-v1-pse", "realtek,pse-mcu-gen1";
```
openwrt-ai, line 16: nit: both new files keep `engenius,ews2910p-v1-pse`, but `engenius,ews2910p-v1` is no longer a board this tree builds. Sibling variants in this target each get their own board-specific PSE string even though they share the fallback — [`zyxel,gs1900-8hp-a1-pse`](https://github.com/openwrt/openwrt/blob/8b3545a3e8d3b2f50307aa393569b5dfe911dcb7/target/linux/realtek/dts/rtl8380_zyxel_gs1900-8hp-a1.dts#L11) vs [`zyxel,gs1900-8hp-b1-pse`](https://github.com/openwrt/openwrt/blob/8b3545a3e8d3b2f50307aa393569b5dfe911dcb7/target/linux/realtek/dts/rtl8380_zyxel_gs1900-8hp-b1.dts#L11), both on `realtek,pse-mcu-gen1`.

Here the two boards genuinely differ in the part being described (ST32F100 + BCM59111 vs Nuvoton M05xx + BCM59121), so sharing one string is arguably worse than in the Zyxel case. Is keeping the common `engenius,ews2910p-v1-pse` intentional, or should these become `engenius,ews2910p-v1-nuvoton-pse` / `engenius,ews2910p-v1-stm-pse`? Nothing matches on it today (the driver binds on `realtek,pse-mcu-gen1`), so it's cosmetic either way — but the string is easier to change now than after it ships.

---

target/linux/realtek/dts/rtl8380_engenius_ews2910p-v1-stm.dts
```diff
@@ -0,0 +1,28 @@
...
+
+&firmware_partition_1 {
+	compatible = "openwrt,uimage";
+	openwrt,ih-magic = <0x03802910>;
+};
+
+&pse {
+	compatible = "engenius,ews2910p-v1-pse", "realtek,pse-mcu-gen1";
+	current-speed = <19200>;
+};
```
openwrt-ai, line 18: nit: these nine lines are byte-identical to `rtl8380_engenius_ews2910p-v1-nuvoton.dts:10-18`; the only real differences between the two files are the compatible/model and the PoE mapping. The family already uses a shared include (`rtl8380_engenius_ews2910p.dtsi`), so a thin `rtl8380_engenius_ews2910p-v1.dtsi` carrying the v1-specific `firmware_partition_1` magic and `&pse` settings would keep the two variants from drifting when v1 details change. Not blocking, but worth considering given a v2 apparently exists in the wild too.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push `cb6e6d50` → `8b3545a3`). The approach changed completely since my last review — from a single corrected v1 mapping to a per-MCU split, per [@jonasjelonek's proposal](https://github.com/openwrt/openwrt/pull/24549#issuecomment-5258566786) — so this is a fresh read of the whole diff rather than an incremental one.

The core of it checks out: the stm mapping is unchanged from the previously reviewed revision, the nuvoton phandles match the trace decoded in [this comment](https://github.com/openwrt/openwrt/pull/24549#issuecomment-5177882776), the v3 mapping is preserved verbatim when moved out of the shared `.dtsi`, and `DEVICE_DTS = $$(SOC)_$(1)` resolves both new recipes to the new file names. All 28 checks on `8b3545a3` are green. Five inline comments, one of which (sysupgrade compatibility) I'd consider blocking.

## Commit checks

- `8b3545a3` "realtek: split ews2910p-v1 by mcu" — three things on the message:
  - Typo in the first body line: `EnGenuis` → `EnGenius`.
  - The message accounts for the new DTS files and the board.d update, but not for the third change in the diff: the v3 PoE mapping is moved out of `rtl8380_engenius_ews2910p.dtsi` into `rtl8380_engenius_ews2910p-v3.dts`. Worth one sentence, since it's the reason a reader sees the v3 files in a commit whose subject only mentions v1.
  - The split leaves users needing to work out which image to flash, which was the main objection raised in the thread. The driver already prints the distinguishing string at boot, and both variants were quoted [here](https://github.com/openwrt/openwrt/pull/24549#issuecomment-5176361599) — putting those two `realtek-pse-mcu-uart ... MCU, BCM59xxx` lines in the commit body makes the message self-contained for anyone hitting this later.

Separately, the PR title and description still describe the previous approach ("correct poe port mapping … Remapped the PoE ports to match") and only carry the ST trace; worth refreshing them to match the split so the merge context matches the commit.

---

jonasjelonek:
> It does seem rather brittle to embed the mapping in the device tree

But that is how PSE-PD framework works right now, having the mapping expressed in the device tree through `pse-pis` and the phandles referenced by PHYs. The device tree describes how the hardware is wired.

> and has scalability problems if we must split outwardly indistinguishable devices by internal component variants, potentially more and more as variants are discovered.

This is a single device model up to now where a vendor does obvious BS. I expect not many more, but if I'm proved wrong an overall better solution can be worked on.

> I think perhaps a better solution is to expose some interface to userspace to remap the poe ports, and provide sysfs files from the driver to indicate which mcu and pse is detected, and let a board.d file sort out the mapping.

Not a friend of this because moving the PSE-PD in the first place was motivated by getting rid of any custom userspace interfaces. With this I feel like rewarding the vendor for doing such shit by designing our interface like their firmware does things.

> That interface not currently existing, I am about to force-push a dts split for ews2910p v1 along the lines you propose above.

There's a downstream patch adding sysfs stuff, so it should be straightforward to add something similar if that is preferred. 

Still, it's just my opinion. Maybe you like to invite others to this discussion and see their opinion.

target/linux/realtek/image/rtl838x.mk
```diff
@@ -104,16 +104,27 @@ define Device/datto_l8
...
-define Device/engenius_ews2910p-v1
+define Device/engenius_ews2910p-v1-nuvoton
   $(Device/engenius_ews2910p)
   SOC := rtl8380
   DEVICE_MODEL := EWS2910P
-  DEVICE_VARIANT := v1
+  DEVICE_VARIANT := v1 (Nuvoton)
+  DEVICE_PACKAGES += kmod-pse-realtek-mcu-uart
+  UIMAGE_MAGIC := 0x03802910
+  SUPPORTED_DEVICES += engenius,ews2910p engenius,ews2910p-v1
```
openwrt-ai, line 114: nit: with these lists, switching between the two variants after the split needs `sysupgrade -F`, while switching into either variant from a pre-split build does not. Both recipes accept `engenius,ews2910p-v1`, so a Nuvoton unit still running an old `engenius_ews2910p-v1` image can take the stm image with a plain sysupgrade — but once it is on `engenius,ews2910p-v1-nuvoton`, the stm image no longer matches, and vice versa (line 125). Since guessing wrong is expected here (the MCU is only visible from the boot log or by opening the case), the case most likely to need a clean upgrade path is exactly the one that is gated.

Is the asymmetry intended? If not, adding the sibling compatible to each list (`... engenius,ews2910p-v1 engenius,ews2910p-v1-stm` here, `... engenius,ews2910p-v1-nuvoton` at line 125) makes the two interchangeable in both directions, which is already effectively true via the shared `engenius,ews2910p-v1` entry. The [GS1900 v1/v2 → A1/B1 rename](https://github.com/openwrt/openwrt/commit/d205878ede886dd2e00b31e4f3c8444b16ae16be) is not a precedent either way, since there the old names mapped 1:1 onto the new ones ([`rtl838x.mk:437`](https://github.com/openwrt/openwrt/blob/a2d42f9503c6cd54d65c3f1b634ecd35b4feef6e/target/linux/realtek/image/rtl838x.mk#L437) vs [`:447`](https://github.com/openwrt/openwrt/blob/a2d42f9503c6cd54d65c3f1b634ecd35b4feef6e/target/linux/realtek/image/rtl838x.mk#L447)) — here one old name maps to two.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push `8b3545a3` → `a2d42f95`). The code delta is small and resolves three of the five items from my last review:

- the Nuvoton mapping comment now reads `7,6,5,4,0,1,2,3`, matching the phandles below it and the trace decoded in [this comment](https://github.com/openwrt/openwrt/pull/24549#issuecomment-5177882776);
- both `model` strings and both `DEVICE_VARIANT`s are now `v1 (Nuvoton)` / `v1 (STMicro)`, consistent across all four places;
- both recipes carry `engenius,ews2910p-v1` in `SUPPORTED_DEVICES`, so v1 units already running OpenWrt keep a metadata-clean sysupgrade path.

The stm mapping, the v3 mapping and the board.d update are untouched by this push. All 28 checks on `a2d42f95` are green. One new nit inline. The two threads from last review that were questions rather than fixes — the shared `engenius,ews2910p-v1-pse` string on two boards with genuinely different MCU/PSE parts, and factoring the duplicated v1 header into a thin `.dtsi` — are unchanged and still stand as posed; no need to re-litigate them if they were considered and declined.

## Commit checks

- `a2d42f95` "realtek: split ews2910p-v1 by mcu" — the message is byte-identical to the previous push (`git log -1 --format=%B` on both SHAs differs in nothing), so all three points from last review are still open. Restating them compactly rather than re-arguing:
  - `EnGenuis` → `EnGenius` in the first body line.
  - The diff also moves the v3 PoE mapping out of `rtl8380_engenius_ews2910p.dtsi` into `rtl8380_engenius_ews2910p-v3.dts`, and now also adds `engenius,ews2910p-v1` to `SUPPORTED_DEVICES` on both new recipes. Neither is mentioned; a sentence each explains why a commit whose subject says "v1" touches the v3 files, and records that the sysupgrade path for existing installs was deliberately preserved.
  - Nothing in the message tells a user which of the two images to flash. The driver already prints the distinguishing line at boot and both were quoted [here](https://github.com/openwrt/openwrt/pull/24549#issuecomment-5176361599) — pasting those two `realtek-pse-mcu-uart ... MCU, BCM59xxx` lines into the body makes the commit self-contained, which matters more than usual given the variants are externally indistinguishable.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push `a2d42f95` → `25475b77`); no new issues found.

The tree is byte-identical to the previously reviewed revision — `git diff a2d42f95..25475b77` is empty and both commits point at tree `f1833231`. The entire delta is in the commit message, and it closes all three points from my last review:

- `EnGenuis` → `EnGenius` in the first body line;
- the v3 move is now accounted for ("The poe port mapping is devolved from the dtsi into the various device dts."), which explains why a commit whose subject says v1 touches `rtl8380_engenius_ews2910p-v3.dts` and the shared `.dtsi`;
- the message now tells a user which image to flash, quoting both `realtek-pse-mcu-uart` boot lines verbatim with the `dmesg | grep pse` recipe. Given the two variants are externally near-identical, having that in `git log` rather than only in the PR thread is the part that will age best.

No Commit checks section this time. All 28 checks on `25475b77` are green.

Two things carried over, neither a request to change this diff:

- The three earlier threads are unchanged and still stand as posed — the shared `engenius,ews2910p-v1-pse` string on two boards with genuinely different MCU/PSE parts, factoring the duplicated v1 header into a thin `.dtsi`, and the [`SUPPORTED_DEVICES` asymmetry](https://github.com/openwrt/openwrt/pull/24549#discussion_r3772659127) between the two new recipes. All three are questions rather than defects; no need to re-litigate if they were considered and declined.
- The PR title and description still describe the pre-split approach ("correct poe port mapping … Remapped the PoE ports to match") and carry only the ST trace. The commit message is what lands, so this is cosmetic — but refreshing them keeps the merge context matching the commit for anyone reading the thread later.

---

DamisDorgon:
@jonasjelonek ping

jonasjelonek:
Since the driver has been accepted upstream now, I started to work on further extensions.
Also regarding this issue. I admit the DT-side split wasn't the best solution but it was IMO the least bad among the option you mentioned above.

I'm working on a different approach right now, let's see how that works out. Do you know or can you check if the two variants also differ in wiring between PSE chip and the front ports? If this is the case, we legitimately have a different hardware design and keeping a single DT for both which ends up as actually wrong for one variant, is bad.

Also good to know: Does the vendor firmware configure the mapping always the way so the port index 0 corresponds to the first port?

DamisDorgon:
> Do you know or can you check if the two variants also differ in wiring between PSE chip and the front ports? 

One has a single PSE, the other has two, so the wiring definitely differs in that respect. The traces from the PSE chips to the ports is not obvious on the top side.

> Also good to know: Does the vendor firmware configure the mapping always the way so the port index 0 corresponds to the first port?

I think I posted the traces earlier in this thread. The 0x1D mapping commands are different between the two v1 hardware versions in vendor firmware.

jonasjelonek:
> One has a single PSE, the other has two, so the wiring definitely differs in that respect. The traces from the PSE chips to the ports is not obvious on the top side.

So it's even worse than I thought.

> I think I posted the traces earlier in this thread. The 0x1D mapping commands are different between the two v1 hardware versions in vendor firmware.

Yes I've seen that, but later when the vendor firmware queries lan1, does it use port 0 or is there another mapping involved? This would be totally stupid but hey, at this stage I wouldn't be surprised anymore.

DamisDorgon:
> > I think I posted the traces earlier in this thread. The 0x1D mapping commands are different between the two v1 hardware versions in vendor firmware.
> 
> Yes I've seen that, but later when the vendor firmware queries lan1, does it use port 0 or is there another mapping involved? This would be totally stupid but hey, at this stage I wouldn't be surprised anymore.

It might take me another day to get things wired up again to test that.

jonasjelonek:
When having a closer look, can you also tell which PSE chips in either case are on the board? Given you said 1 vs 2 I guess the single PSE might be a BCM59121 and the two on the other variant might be BCM59111.

DamisDorgon:
As I said above (three weeks ago, https://github.com/openwrt/openwrt/pull/24549#issuecomment-5176361599): 

> Amusing anecdote, I got another EWS2910P (also marked hw v1) and it is showing a different poe port mapping.
> 
> Instead of an ST 32F100C8, it has a Nuvoton microcontroller (M0516):
> 
> [   21.934109] realtek-pse-mcu-uart serial0-0: Nuvoton M05xx LAN MCU, BCM59121 (id 0xe121), 8 ports across 1 PSE chip(s)
> 
> Link shows on lan5, poe port shows power delivery on lan4.
> 
> The matching line on the other hw v1 says:
> 
> [   21.923045] realtek-pse-mcu-uart serial0-0: ST Micro ST32F100 MCU, BCM59111 (id 0xe111), 8 ports across 255 PSE chip(s)
> 
> (it has two PSE chips, not 255).

jonasjelonek:
Sorry, I've overlooked that in this whole conversation.

PR #24550, opened 2026-08-03
realtek: add power management mode for EWS2910P v1

DamisDorgon:
The EnGenius EWS2910P v1 appears to require a Set Power Management Mode command for the port control to function.

### DIFF
--- a/target/linux/realtek/patches-6.18/819-net-pse-pd-realtek-pse-mcu-ensure-dynamic-power-mana.patch
+++ b/target/linux/realtek/patches-6.18/819-net-pse-pd-realtek-pse-mcu-ensure-dynamic-power-mana.patch
@@ -1,0 +1,94 @@
+From 0e7f97c356a8c04bbae1e412ecfac8efa46ec03e Mon Sep 17 00:00:00 2001
+From: Russell Senior <russell@personaltelco.net>
+Date: Mon, 3 Aug 2026 13:44:04 -0700
+Subject: [PATCH] net: pse-pd: realtek-pse-mcu: ensure dynamic power management
+ mode with priority
+
+The MCU may have any power management mode configuration by default
+but the driver should ensure that we run in "Dynamic Power Mgmt with
+Priority" mode. The absence of this command was noted with the
+Engenius EWS2910P v1. In the unexpected case where the opcode is
+not recognized by a different board, the code emits a warning.
+
+The opcode used for gen2 devices is derived from the realtek-poe
+userspace package that preceded the realtek-pse-mcu driver.
+
+Tested on an EWS2910P v1.
+
+Signed-off-by: Russell Senior <russell@personaltelco.net>
+---
+ drivers/net/pse-pd/realtek-pse-mcu-core.c | 31 +++++++++++++++++++++++
+ 1 file changed, 31 insertions(+)
+
+--- a/drivers/net/pse-pd/realtek-pse-mcu-core.c
++++ b/drivers/net/pse-pd/realtek-pse-mcu-core.c
+@@ -65,10 +65,14 @@
+ #define RTPSE_MCU_MAX_PORTS			48
+ #define RTPSE_MCU_PORT_MAX_PRIORITY		3
+ 
++/* RTPSE_MCU_CMD_SET_POWER_MANAGEMENT_MODE values */
++#define RTPSE_MCU_DYNAMIC_POWER_MGMT_PRIO	0x02	/* dynamic, with port priority */
++
+ enum rtpse_mcu_cmd {
+ 	RTPSE_MCU_CMD_SET_GLOBAL_STATE,
+ 	RTPSE_MCU_CMD_GET_SYSTEM_INFO,
+ 	RTPSE_MCU_CMD_GET_EXT_CONFIG,
++	RTPSE_MCU_CMD_SET_POWER_MANAGEMENT_MODE,
+ 
+ 	RTPSE_MCU_CMD_PORT_ENABLE,
+ 	RTPSE_MCU_CMD_PORT_SET_POWER_LIMIT_TYPE,
+@@ -390,6 +394,26 @@ static int rtpse_mcu_set_global_state(st
+ 	return (resp.payload[0] == 0x0) ? 0 : -EIO;
+ }
+ 
++static int rtpse_mcu_set_power_management_mode(struct rtpse_mcu_ctrl *pse, u8 value)
++{
++	struct rtpse_mcu_msg req, resp;
++	const struct rtpse_mcu_opcode *opc;
++	int ret;
++
++	opc = &pse->dialect->opcode[RTPSE_MCU_CMD_SET_POWER_MANAGEMENT_MODE];
++	if (!opc->valid)
++		return -EOPNOTSUPP;
++
++	rtpse_mcu_msg_init(&req, opc->op);
++	req.payload[0] = value;
++
++	ret = rtpse_mcu_do_xfer(pse, &req, &resp);
++	if (ret)
++		return ret;
++
++	return (resp.payload[0] == 0x0) ? 0 : -EIO;
++}
++
+ /* Port operations */
+ 
+ static int rtpse_mcu_port_get_status(struct rtpse_mcu_ctrl *pse, unsigned int port,
+@@ -974,6 +998,11 @@ int rtpse_mcu_register(struct rtpse_mcu_
+ 	if (ret)
+ 		return ret;
+ 
++	/* Ensure that we run in "Dynamic Power Management Mode with Priority" mode */
++	ret = rtpse_mcu_set_power_management_mode(pse, RTPSE_MCU_DYNAMIC_POWER_MGMT_PRIO);
++	if (ret)
++		dev_warn_probe(pse->dev, ret, "failed to set power management mode\n");
++
+ 	if (!info.system_enable) {
+ 		ret = rtpse_mcu_set_global_state(pse, true);
+ 		/* Dialects without a global-state concept (e.g. Gen1) return
+@@ -1093,6 +1122,7 @@ static const struct rtpse_mcu_dialect rt
+ 		[RTPSE_MCU_CMD_SET_GLOBAL_STATE]	= RTPSE_MCU_OP(0x00),
+ 		[RTPSE_MCU_CMD_GET_SYSTEM_INFO]		= RTPSE_MCU_OP(0x40),
+ 		[RTPSE_MCU_CMD_GET_EXT_CONFIG]		= RTPSE_MCU_OP(0x4a),
++		[RTPSE_MCU_CMD_SET_POWER_MANAGEMENT_MODE] = RTPSE_MCU_OP(0x10),
+ 
+ 		[RTPSE_MCU_CMD_PORT_ENABLE]		= RTPSE_MCU_OP(0x01),
+ 		[RTPSE_MCU_CMD_PORT_SET_POWER_LIMIT_TYPE] = RTPSE_MCU_OP(0x12),
+@@ -1116,6 +1146,7 @@ static const struct rtpse_mcu_dialect rt
+ 	.opcode = {
+ 		[RTPSE_MCU_CMD_GET_SYSTEM_INFO]		= RTPSE_MCU_OP(0x20),
+ 		[RTPSE_MCU_CMD_GET_EXT_CONFIG]		= RTPSE_MCU_OP(0x2b),
++		[RTPSE_MCU_CMD_SET_POWER_MANAGEMENT_MODE] = RTPSE_MCU_OP(0x17),
+ 
+ 		[RTPSE_MCU_CMD_PORT_ENABLE]		= RTPSE_MCU_OP(0x00),
+ 		[RTPSE_MCU_CMD_PORT_SET_POWER_LIMIT_TYPE] = RTPSE_MCU_OP(0x15),


DamisDorgon:
the patch was developed with quilt, so git am compatibility is not readily available, afaik.

target/linux/realtek/patches-6.18/819-net-pse-mcu-add-power-management-mode.patch
```diff
@@ -0,0 +1,74 @@
+--- a/drivers/net/pse-pd/realtek-pse-mcu-core.c
++++ b/drivers/net/pse-pd/realtek-pse-mcu-core.c
+@@ -65,10 +65,13 @@
+ #define RTPSE_MCU_MAX_PORTS			48
+ #define RTPSE_MCU_PORT_MAX_PRIORITY		3
+ 
++#define RTPSE_MCU_DYNAMIC_POWER_MGMT		0x02    /* Dynamic Power Management with Port Priority */
```
jonasjelonek: Since there is also a dynamic mode without priority, I would somehow include this in the naming of the macro, e.g. `RTPSE_MCU_DYNAMIC_POWER_MGMT_PRIO`

target/linux/realtek/patches-6.18/819-net-pse-mcu-add-power-management-mode.patch
```diff
@@ -0,0 +1,74 @@
...
+ /* Port operations */
+ 
+ static int rtpse_mcu_port_get_status(struct rtpse_mcu_ctrl *pse, unsigned int port,
+@@ -974,6 +998,13 @@ int rtpse_mcu_register(struct rtpse_mcu_
+ 	if (ret)
+ 		return ret;
+ 
++	/*
++	 * This is needed for gen1 device EWS2910P v1, but not apparently for v3
++	 */
```
jonasjelonek: Maybe not fix this too much on one device, rather the general reasoning that the MCU may have any configuration by default but the driver should ensure that we run in "Dynamic Power Mgmt with Priority" mode.

If that fits in one line, `/* ... */` should be fine.

jonasjelonek (changes requested):
"git am compatibility" requires additional headers like `From:`. See https://github.com/openwrt/openwrt/blob/main/target/linux/realtek/patches-6.18/300-03-realtek-board-increase-fixup_fdt-buffer-to-32-KiB.patch for a recent example.

In general you get this when you do a real commit and use `git format-patch`.

target/linux/realtek/patches-6.18/819-net-pse-mcu-add-power-management-mode.patch
```diff
@@ -0,0 +1,74 @@
...
++	ret = rtpse_mcu_set_power_management_mode(pse,RTPSE_MCU_DYNAMIC_POWER_MGMT);
```
openwrt-ai: Missing space after the comma in the argument list (`pse,RTPSE_MCU_DYNAMIC_POWER_MGMT`) — checkpatch will emit `ERROR: space required after that ','`.

```suggestion
+	ret = rtpse_mcu_set_power_management_mode(pse, RTPSE_MCU_DYNAMIC_POWER_MGMT);
```

---

target/linux/realtek/patches-6.18/819-net-pse-mcu-add-power-management-mode.patch
```diff
@@ -0,0 +1,74 @@
...
++	if (ret)
++		return dev_err_probe(pse->dev, ret, "failed to set power management mode\n");
```
openwrt-ai: This is inserted into `rtpse_mcu_register()`, which is the common registration path for *every* board using this driver, and any failure aborts probe. The comment itself says the command is only known to be needed on the gen1 EWS2910P v1 and "not apparently" on v3, yet the patch also adds a gen2 opcode (line 63) and sends the command on gen2 too.

15 boards currently bind to these compatibles ([`grep realtek,pse-mcu-gen` over `target/linux/realtek/dts/`](https://github.com/openwrt/openwrt/tree/aa69867bdf395ce268db7900de370d46a73d22b2/target/linux/realtek/dts)) — e.g. [`rtl8380_engenius_ews2910p-v3.dts:16`](https://github.com/openwrt/openwrt/blob/aa69867bdf395ce268db7900de370d46a73d22b2/target/linux/realtek/dts/rtl8380_engenius_ews2910p-v3.dts#L16), the GS1900/GS1920 family, `zyxel,xgs1930-28hp`, `ubnt,usw-pro-xg-8-poe`, `plasmacloud,psx28`. If any of those MCU firmwares answer opcode `0x10`/`0x17` with a non-zero status byte, `rtpse_mcu_set_power_management_mode()` returns `-EIO` and PoE probe fails outright on hardware that works today.

Note the immediately following block treats the analogous `rtpse_mcu_set_global_state()` failure as non-fatal for exactly this reason. Two options that avoid the regression risk:

- gate the call on the dialect (only issue it where it is known to be required), or
- keep it unconditional but downgrade the failure to `dev_warn_probe()`/`dev_warn()` and continue.

Also, has the payload value been confirmed for gen2? The driver's premise is that the two generations "diverge on opcode numbers and on a few response layouts", so a shared `0x02 == dynamic power management` encoding across gen1 (Broadcom-fronted) and gen2 (Realtek) isn't self-evident from the diff.

Separately, the comment names a specific board in otherwise board-agnostic core code without saying *what* breaks without it — worth spelling out ("port enable/disable is silently ignored until dynamic power management is selected"), and it fits on one line:

```suggestion
+	/* Gen1 MCUs (e.g. EnGenius EWS2910P v1) ignore per-port control until
+	 * a power management mode has been selected explicitly.
+	 */
+	ret = rtpse_mcu_set_power_management_mode(pse, RTPSE_MCU_DYNAMIC_POWER_MGMT);
+	if (ret)
+		dev_warn(pse->dev, "failed to set power management mode: %pe\n", ERR_PTR(ret));
+
```

---

target/linux/realtek/patches-6.18/819-net-pse-mcu-add-power-management-mode.patch
```diff
@@ -0,0 +1,74 @@
...
++	req.payload[0] = value;
++
++	ret = rtpse_mcu_do_xfer(pse, &req, &resp);
++	if (ret)
++		return ret;
++
++	return (resp.payload[0] == 0x0) ? 0 : -EIO;
++}
++
++
```
openwrt-ai: Two blank lines are added between the new function and `/* Port operations */`; checkpatch flags this (`CHECK: Please don't use multiple blank lines`), and every other function boundary in this file uses one.

```suggestion
+
```

---

target/linux/realtek/patches-6.18/819-net-pse-mcu-add-power-management-mode.patch
openwrt-ai: The trailing comment is separated by four spaces (the surrounding defines align with tabs) and pushes the line to 104 columns, over the kernel's 100-column limit — checkpatch will warn on both. The file's existing convention for command payload values is a grouping comment above the block, e.g. `/* RTPSE_MCU_PORT_SET_POWER_LIMIT_TYPE values */` ahead of `RTPSE_MCU_PORT_PW_LIMIT_TYPE_USER` in [816-01-...-PSE-MCU-core.patch](``https://github.com/openwrt/openwrt/blob/aa69867bdf395ce268db7900de370d46a73d22b2/target/linux/realtek/patches-6.18/816-01-net-pse-pd-add-Realtek-Broadcom-PSE-MCU-core.patch#L146-L148),`` which also keeps the value name tied to its command.

```suggestion
+/* RTPSE_MCU_CMD_SET_POWER_MANAGEMENT_MODE values */
+#define RTPSE_MCU_DYNAMIC_POWER_MGMT		0x02	/* dynamic, with port priority */
+
```

---

target/linux/realtek/patches-6.18/819-net-pse-pd-realtek-pse-mcu-ensure-dynamic-power-mana.patch
openwrt-ai, line 1: nit: the filename drops the `pse-pd` component that all its siblings in this directory carry — `816-01-net-pse-pd-add-Realtek-Broadcom-PSE-MCU-core.patch`, `817-net-pse-pd-realtek-pse-mcu-add-per-por-legacy-detec.patch`, `818-net-pse-pd-add-hasivo-hs104-driver.patch`. Since patch filenames here are derived from the commit subject, `819-net-pse-pd-realtek-pse-mcu-add-power-managemen.patch` (with a matching `Subject: [PATCH] net: pse-pd: realtek-pse-mcu: ...` header) would keep the series consistent.

---

openwrt-ai (commented):
## Commit checks

- `aa69867` "realtek: add power management mode for EWS2910P v1" — the subject and body scope the change to the EWS2910P v1, but the diff also registers the command opcode for the **gen2** dialect (`0x10`) and issues the Set Power Management Mode command unconditionally from `rtpse_mcu_register()`, i.e. on all 15 boards that bind `realtek,pse-mcu-gen1*`/`gen2*` — including the EWS2910P v3 the body says does not need it. Either narrow the change to the dialect that needs it, or widen the commit message to say that all realtek-pse-mcu boards now get this command at probe and why that is safe.

The patch applies cleanly on top of `817-...` with no offset/fuzz, and no other `patches-*` directory exists for this target, so nothing else needs refreshing. `816-01`'s opcode assignments show no collision for `0x10` (gen2) or `0x17` (gen1).

The missing `git am` headers were already raised by @jonasjelonek and the bot, so I have not repeated that here.

---

target/linux/realtek/patches-6.18/819-net-pse-pd-realtek-pse-mcu-ensure-dynamic-power-mana.patch
```diff
@@ -0,0 +1,86 @@
...
++		dev_warn_probe(pse->dev, ret, "failed to set power management mode\n");
++
+ 	if (!info.system_enable) {
+ 		ret = rtpse_mcu_set_global_state(pse, true);
+ 		/* Dialects without a global-state concept (e.g. Gen1) return
+@@ -1091,6 +1120,7 @@ static const struct rtpse_mcu_dialect rt
+ 	.detect_legacy     = 0x1,
+ 	.opcode = {
+ 		[RTPSE_MCU_CMD_SET_GLOBAL_STATE]	= RTPSE_MCU_OP(0x00),
++		[RTPSE_MCU_CMD_SET_POWER_MANAGEMENT_MODE] = RTPSE_MCU_OP(0x10),
```
openwrt-ai: What is the gen2 opcode `0x10` based on? The reported evidence in the commit message is for the EnGenius EWS2910P v1, which binds `realtek,pse-mcu-gen1*` — so the tested value is the gen1 `0x17` added below, not this one. The two dialects' opcode maps are fully disjoint (gen1 uses `0x10` for [`RTPSE_MCU_CMD_PORT_SET_DETECTION_TYPE`](``https://github.com/openwrt/openwrt/blob/bfac3b2817894db5e659b45c143ab6905ac2cfb3/target/linux/realtek/patches-6.18/817-net-pse-pd-realtek-pse-mcu-add-per-por-legacy-detec.patch#L273),`` gen2 uses `0x09` for the same command), so `0x10` for gen2 cannot be inferred from the gen1 value.

This matters because `rtpse_mcu_register()` now issues the command unconditionally at probe: 9 in-tree boards bind a gen2 compatible (`rtl9301_zyxel_xgs1930-28hp`, `rtl9313_ubnt_usw-pro-xg-8-poe`, `rtl9312_plasmacloud_psx28`, `rtl8380_engenius_ews2910p-v3`, …). Note the neighbouring gen2 opcodes in the `0x1x` range are all per-port setters ([`PORT_SET_POWER_LIMIT_TYPE` 0x12 … `PORT_SET_PRIORITY` 0x15](``https://github.com/openwrt/openwrt/blob/bfac3b2817894db5e659b45c143ab6905ac2cfb3/target/linux/realtek/patches-6.18/816-01-net-pse-pd-add-Realtek-Broadcom-PSE-MCU-core.patch#L1032-L1035))`` whose `payload[0]` is the port index — if `0x10` also lives in that block on gen2, this call would be read as "port 2", not as a mode value, and the driver would not notice because the MCU still answers `payload[0] == 0`.

If the gen2 value is untested, would it be safer to land only the gen1 entry for now (the gen2 dialect then returns `-EOPNOTSUPP` from the `!opc->valid` check and only trips the new `dev_warn_probe`), and add gen2 once someone can confirm it on hardware?

---

target/linux/realtek/patches-6.18/819-net-pse-pd-realtek-pse-mcu-ensure-dynamic-power-mana.patch
```diff
@@ -0,0 +1,86 @@
...
+ 1 file changed, 31 insertions(+)
+
+--- a/drivers/net/pse-pd/realtek-pse-mcu-core.c
++++ b/drivers/net/pse-pd/realtek-pse-mcu-core.c
+@@ -65,10 +65,14 @@
+ #define RTPSE_MCU_MAX_PORTS			48
+ #define RTPSE_MCU_PORT_MAX_PRIORITY		3
+ 
++/* RTPSE_MCU_CMD_SET_POWER_MANAGEMENT_MODE values */
++#define RTPSE_MCU_DYNAMIC_POWER_MGMT_PRIO	0x02    /* dynamic, with port priority */
```
openwrt-ai: nit: trailing comments in this define block are separated by a tab, not spaces — see [`RTPSE_MCU_PORT_STS_TEST` / `RTPSE_MCU_PORT_STS_OTHER_FAULT`](``https://github.com/openwrt/openwrt/blob/bfac3b2817894db5e659b45c143ab6905ac2cfb3/target/linux/realtek/patches-6.18/816-01-net-pse-pd-add-Realtek-Broadcom-PSE-MCU-core.patch#L120-L124).``

```suggestion
+#define RTPSE_MCU_DYNAMIC_POWER_MGMT_PRIO	0x02	/* dynamic, with port priority */
```

Also, the other `/* <CMD> values */` groups all sit above the generic `RTPSE_MCU_MAX_PORTS` / `RTPSE_MCU_PORT_MAX_PRIORITY` pair; this one is placed below them, which splits that pair off from the limits it belongs with. Moving the new block up next to `/* RTPSE_MCU_PORT_SET_POWER_LIMIT_TYPE values */` would keep the grouping intact (needs the hunk regenerated, so not suggested inline).

---

target/linux/realtek/patches-6.18/819-net-pse-pd-realtek-pse-mcu-ensure-dynamic-power-mana.patch
```diff
@@ -0,0 +1,86 @@
...
+ 		[RTPSE_MCU_CMD_GET_SYSTEM_INFO]		= RTPSE_MCU_OP(0x40),
+ 		[RTPSE_MCU_CMD_GET_EXT_CONFIG]		= RTPSE_MCU_OP(0x4a),
+ 
+@@ -1114,6 +1144,7 @@ static const struct rtpse_mcu_dialect rt
+ 	.detect_standard   = 0x2,
+ 	.detect_legacy     = 0x3,
+ 	.opcode = {
++		[RTPSE_MCU_CMD_SET_POWER_MANAGEMENT_MODE] = RTPSE_MCU_OP(0x17),
+ 		[RTPSE_MCU_CMD_GET_SYSTEM_INFO]		= RTPSE_MCU_OP(0x20),
+ 		[RTPSE_MCU_CMD_GET_EXT_CONFIG]		= RTPSE_MCU_OP(0x2b),
```
openwrt-ai: nit: both dialect opcode tables list their entries in `enum rtpse_mcu_cmd` order. The new command is declared last in the first enum group (after `RTPSE_MCU_CMD_GET_EXT_CONFIG`), so its table entry belongs after `GET_EXT_CONFIG` rather than first. The line count of the hunk is unchanged, so this is safe to move:

```suggestion
 		[RTPSE_MCU_CMD_GET_SYSTEM_INFO]		= RTPSE_MCU_OP(0x20),
 		[RTPSE_MCU_CMD_GET_EXT_CONFIG]		= RTPSE_MCU_OP(0x2b),
+		[RTPSE_MCU_CMD_SET_POWER_MANAGEMENT_MODE] = RTPSE_MCU_OP(0x17),
```

Same applies to the gen2 entry at line 75, which is currently placed between `SET_GLOBAL_STATE` and `GET_SYSTEM_INFO` — i.e. the two tables also disagree with each other on where the new entry goes.

---

target/linux/realtek/patches-6.18/819-net-pse-pd-realtek-pse-mcu-ensure-dynamic-power-mana.patch
```diff
@@ -0,0 +1,86 @@
+From db8a5458126ed2e700fa0d1f255850aa810cbd40 Mon Sep 17 00:00:00 2001
+From: Russell Senior <russell@personaltelco.net>
+Date: Mon, 3 Aug 2026 13:44:04 -0700
+Subject: [PATCH] net: pse-pd: realtek-pse-mcu: ensure dynamic power management
+ mode with priority
+
+The MCU may have any power management mode configuration by default but the driver
+should ensure that we run in "Dynamic Power Mgmt with Priority" mode.
```
openwrt-ai: nit: the first line runs to 82 columns; the sibling patches in this directory wrap their bodies at ~72–75 (see [816-01](`https://github.com/openwrt/openwrt/blob/bfac3b2817894db5e659b45c143ab6905ac2cfb3/target/linux/realtek/patches-6.18/816-01-net-pse-pd-add-Realtek-Broadcom-PSE-MCU-core.patch#L5-L15)).`

```suggestion
The MCU may have any power management mode configuration by default but
the driver should ensure that we run in "Dynamic Power Mgmt with
Priority" mode.
```

Separately: this embedded message drops the two facts the OpenWrt commit message carries — that the EnGenius EWS2910P v1 needs this for port control to work at all, and that a failure is only warned about rather than aborting probe. Worth folding both in here, since this is the text that would go upstream.

---

openwrt-ai (commented):
Re-reviewed at `bfac3b2` (previous review was at `aa69867`).

The commit-message scope mismatch I raised last time is resolved: the message now says the command is issued for all gen1 and gen2 boards and that an error only produces a warning, which matches the diff. The `git am` headers requested by @jonasjelonek are present now too.

Verified locally: the patch applies on top of `817-...` with no offset or fuzz, and `rtpse_mcu_set_power_management_mode()` mirrors [`rtpse_mcu_set_global_state()`](``https://github.com/openwrt/openwrt/blob/bfac3b2817894db5e659b45c143ab6905ac2cfb3/target/linux/realtek/patches-6.18/816-01-net-pse-pd-add-Realtek-Broadcom-PSE-MCU-core.patch#L376-L394)`` exactly. The stale non-zero `ret` left behind by the new `dev_warn_probe()` path is harmless — it is either reassigned in the `!info.system_enable` branch or dead before `devm_pse_controller_register()`. Neither `0x10` (gen2) nor `0x17` (gen1) collides with an existing opcode in its dialect. All CI checks on this head are green.

Remaining question is inline: the gen2 opcode assignment.

---

target/linux/realtek/patches-6.18/819-net-pse-pd-realtek-pse-mcu-ensure-dynamic-power-mana.patch
DamisDorgon: The opcode for gen2 is derived from the realtek-poe package that preceded the pse-mcu-realtek driver.

target/linux/realtek/patches-6.18/819-net-pse-pd-realtek-pse-mcu-ensure-dynamic-power-mana.patch
```diff
@@ -0,0 +1,94 @@
+From 0e7f97c356a8c04bbae1e412ecfac8efa46ec03e Mon Sep 17 00:00:00 2001
+From: Russell Senior <russell@personaltelco.net>
+Date: Mon, 3 Aug 2026 13:44:04 -0700
+Subject: [PATCH] net: pse-pd: realtek-pse-mcu: ensure dynamic power management
+ mode with priority
+
+The MCU may have any power management mode configuration by default
+but the driver should ensure that we run in "Dynamic Power Mgmt with
+Priority" mode. The absence of this command was noted with the
+Engenius EWS2910P v1. In the unexpected case where the opcode is
```
openwrt-ai, line 10: nit: the vendor is spelled `EnGenius` everywhere in-tree — [`model = "EnGenius EWS2910P v1";`](https://github.com/openwrt/openwrt/blob/2c20f118be66ea6925e1284aa06919a79145b8d7/target/linux/realtek/dts/rtl8380_engenius_ews2910p-v1.dts#L7) and [`DEVICE_VENDOR := EnGenius`](https://github.com/openwrt/openwrt/blob/2c20f118be66ea6925e1284aa06919a79145b8d7/target/linux/realtek/image/common.mk#L45) — and the OpenWrt commit message for this change also uses `EnGenius`. Only this embedded upstream message says `Engenius`.

```suggestion
EnGenius EWS2910P v1. In the unexpected case where the opcode is
```

---

openwrt-ai (commented):
Re-reviewed at `2c20f11` (previous review was at `bfac3b2`).

Everything I raised last round is addressed: the tab-vs-spaces separator on the new `#define`, the opcode-table entry ordering (both tables now list `SET_POWER_MANAGEMENT_MODE` after `GET_EXT_CONFIG`, matching `enum rtpse_mcu_cmd`), the commit-message wrapping, and the missing rationale — the gen2 opcode provenance and the "warning only" behaviour are now stated in the patch header.

Verified locally against this head:

- The patch applies on top of `818-...` with no offset and no fuzz, and the five hunk headers are internally consistent (`+4`, `+20`, `+5`, `+1`, `+1` = the 31 insertions in the diffstat), with the cumulative line offsets `65 → 65`, `390 → 394`, `974 → 998`, `1093 → 1122`, `1116 → 1146` all matching.
- The resulting tables are collision-free: `0x10` is unused elsewhere in the gen2 dialect, `0x17` unused elsewhere in gen1. Note `0x10` and `0x17` are both live in the *other* dialect (gen1 `PORT_SET_DETECTION_TYPE` / gen2 `PORT_SET_PRIORITY`), so they are correctly not shared.
- `rtpse_mcu_set_power_management_mode()` mirrors [`rtpse_mcu_set_global_state()`](``https://github.com/openwrt/openwrt/blob/2c20f118be66ea6925e1284aa06919a79145b8d7/target/linux/realtek/patches-6.18/816-01-net-pse-pd-add-Realtek-Broadcom-PSE-MCU-core.patch#L376-L394)`` line for line.
- The new enum member does not need any other update: the opcode array is sized by `RTPSE_MCU_NUM_CMDS` and every use is a designated initializer or an explicit index; there is no name table indexed by the enum.
- Adding a possibly-unrecognised command at probe on all boards is safe from a desync standpoint: `rtpse_mcu_uart_send()` resets `rx_len`/`rx_done` before every transmit, so a missing or late reply cannot bleed into the following `set_global_state` transaction. Cost is one `RTPSE_MCU_RESPONSE_MS` pause plus, at worst, one `RTPSE_MCU_UART_RX_TIMEOUT` on boards that ignore the opcode.
- The stale non-zero `ret` left by the `dev_warn_probe()` path is still harmless — it is reassigned by `devm_pse_controller_register()` before any use on the `info.system_enable == true` path.
- `dev_warn_probe()` exists in 6.18 (`include/linux/dev_printk.h`), and `patches-6.18` is the only patch directory for this target, so no other kernel series is affected.

All 28 CI checks on this head are green.

Only a casing nit inline.

---

DamisDorgon:
Tested on an EWS2910P v3. No obvious problems.

openwrt-ai (commented):

Re-reviewed at `453c44a` (previous review was at `2c20f11`). The only change since then is a rebase onto current `main` — `git diff 2c20f11 453c44a -- target/linux/realtek/patches-6.18/819-*` is empty, and the commit message is byte-identical, so the patch content and metadata I checked last round are unchanged.

Re-verified that the rebase did not invalidate anything:

- None of the patches the new one stacks on changed between the old and new base — `816-01/02/03`, `817-...`, `818-...` are untouched; the only realtek `patches-6.18` churn on `main` was `300-02-enhance-realtek-board-setup.patch` and the new `300-03-...-fixup_fdt-buffer...patch`, which touch a different file. So the five hunks still land at the same context with no offset or fuzz, and `819` remains the highest-numbered patch in the directory (no renumbering needed).
- Opcode tables are still collision-free after the rebase: `0x10` unused elsewhere in gen2, `0x17` unused elsewhere in gen1.
- The author's [test report on an EWS2910P v3](https://github.com/openwrt/openwrt/pull/24550#issuecomment-5223891095) covers the gen2 side, which was the open question from the earlier rounds — that resolves my remaining concern about the gen2 `0x10` opcode being untested on hardware.

All 28 checks on this head are green, including `Check Kernel patches (realtek, rtl838x)` and the five per-subtarget kernel builds.

---

jonasjelonek:
Tested on Zyxel XS1930-12HP

jonasjelonek:

PR #24551, opened 2026-08-03
kernel: bump to 6.12.103 and 6.18.44

DragonBluep:
Cc @graysky2 @GoetzGoerisch


target/linux/generic/kernel-6.12
```diff
@@ -1,2 +1,2 @@
-LINUX_VERSION-6.12 = .100
-LINUX_KERNEL_HASH-6.12.100 = 67f973533406492e86774bacbcefae50d50d5c34cbf703c47ec526a5efdcee90
+LINUX_VERSION-6.12 = .101
+LINUX_KERNEL_HASH-6.12.101 = skip
```
openwrt-ai: `skip` is a magic value that disables tarball integrity checking, not a placeholder that is filled in later: [`download.pl:159`](https://github.com/openwrt/openwrt/blob/15a49fda5a0ed4235b038eff05ca2e9a7d3a2f95/scripts/download.pl#L159) accepts it in place of a real checksum, and [`download.mk:160`](https://github.com/openwrt/openwrt/blob/15a49fda5a0ed4235b038eff05ca2e9a7d3a2f95/include/download.mk#L160) short-circuits the mirror hash comparison for it. Merging this would ship a kernel bump with no verification of the downloaded source, and it is also why CI went green here. Every previous bump carries the real sha256 (the base is `LINUX_KERNEL_HASH-6.12.100 = 67f9735...`).

The published sha256 for `linux-6.12.101.tar.xz` (kernel.org `v6.x/sha256sums.asc`) is:

```suggestion
LINUX_KERNEL_HASH-6.12.101 = 0d21cd11933f49f7151b7c9dbb8cc3fddc8c8abe506434b850feecf41fc28a76
```

---

target/linux/generic/kernel-6.18
```diff
@@ -1,2 +1,2 @@
-LINUX_VERSION-6.18 = .41
-LINUX_KERNEL_HASH-6.18.41 = 17fc72f0f8d4a8a8633a5d20085f5d9c5a5ec51ee896a0b7ae1ec25da31273ea
+LINUX_VERSION-6.18 = .42
+LINUX_KERNEL_HASH-6.18.42 = skip
```
openwrt-ai: Same problem as in `target/linux/generic/kernel-6.12` — `skip` disables the source integrity check instead of acting as a placeholder. The published sha256 for `linux-6.18.42.tar.xz` is:

```suggestion
LINUX_KERNEL_HASH-6.18.42 = 35a8ef5e8fb44467e56a42b40c23c4c10a61d1e95ff8d738f285230e8bed63e7
```

---

target/linux/layerscape/patches-6.18/701-staging-add-fsl_ppfe-driver.patch
```diff
@@ -972,7 +972,7 @@ Signed-off-by: Dong Aisheng <aisheng.dong@nxp.com>
 +};
 --- a/MAINTAINERS
 +++ b/MAINTAINERS
-@@ -10002,6 +10002,14 @@ F:	drivers/net/ethernet/freescale/enetc/
+@@ -10015,6 +10015,14 @@ F:	drivers/net/ethernet/freescale/enetc/
```
openwrt-ai, line 975: This `+13` shift is correct — 6.18.42 adds 13 lines to `MAINTAINERS` at line 8656 (`@@ -8656,6 +8656,19 @@`, the ttm/DRM block), so every patch with a `MAINTAINERS` hunk below that had to move. The same shift was applied here and in `generic/pending-6.18/510-...` (26273 → 26286), mvebu 907, realtek 816-01, starfive 1007, ipq40xx 700/701, bcm27xx, airoha 180-01, etc.

Two patches with `MAINTAINERS` hunks below 8656 were missed, and `target/linux/econet/` and `target/linux/qualcommbe/` have no changes at all in this commit:

- [`econet/patches-6.18/910-v7.1-dt-bindings-clock-reset-add-econet-en751221-bindings.patch:70`](``https://github.com/openwrt/openwrt/blob/15a49fda5a0ed4235b038eff05ca2e9a7d3a2f95/target/linux/econet/patches-6.18/910-v7.1-dt-bindings-clock-reset-add-econet-en751221-bindings.patch#L70)`` — `@@ -8839,6 +8839,8 @@` should be `@@ -8852,6 +8852,8 @@`
- [`qualcommbe/patches-6.18/0325-MAINTAINERS-Add-maintainer-for-Qualcomm-PPE-driver.patch:16`](``https://github.com/openwrt/openwrt/blob/15a49fda5a0ed4235b038eff05ca2e9a7d3a2f95/target/linux/qualcommbe/patches-6.18/0325-MAINTAINERS-Add-maintainer-for-Qualcomm-PPE-driver.patch#L16)`` — `@@ -21239,6 +21239,15 @@` should be `@@ -21252,6 +21252,15 @@`

Neither target appears in the CI target matrix on this PR, so the stale offsets go unnoticed there. Please run `make target/linux/refresh` with econet and qualcommbe selected as well so the commit message's "All other patches are automatically refreshed" holds for the whole tree.

---

openwrt-ai (commented):
## Commit checks

- `f1044db` "kernel: bump 6.12 to 6.12.101" — `LINUX_KERNEL_HASH-6.12.101` is set to `skip`, which disables source verification rather than acting as a placeholder. See the inline comment on `target/linux/generic/kernel-6.12`.
- `15a49fd` "kernel: bump 6.18 to 6.18.42" — same `skip` hash for 6.18.42. Additionally, the body states "All other patches are automatically refreshed", but `target/linux/econet/` and `target/linux/qualcommbe/` were not touched at all and each still carries a stale `MAINTAINERS` hunk offset (details inline on the layerscape ppfe patch).

Verified and found correct: all eight referenced upstream commit IDs resolve on the stable tree with matching subjects; the four dropped patches (ppp IFF_NO_QUEUE / percpu netstats / TX scatter-gather, powerpc 85xx fsl,ifc) are genuinely present in the 6.12.101 and 6.18.42 incrementals; the manual bcm27xx xhci rebases place the downstream VL805 quirks consistently ahead of the new upstream `xhci->dma_mask_bits = 36;` in all four patches of both series; and the automatic hunk-header refreshes match the real upstream line deltas (e.g. `920-mangle_bootargs.patch` 6.18 `init/main.c` 635 → 590 corresponds to the 45-line `xbc_snprint_cmdline()` removal, and `624-...-pppoe-drop-sock-reference-counting-on-fast-path.patch` correctly needs no shift because dropping `620-...-ppp-use-IFF_NO_QUEUE...` cancels out the upstream `+1` in `pppoe.c`).

---

joshuacov1:
I tested 6.18.42 successfully on bcm27xx, however I would **not** recommend this for merging without  a proper sha256-checksum for the tarball.

graysky2:
@DragonBluep - I ran my update script after applying your PR and three files changed. I just forked your repo and sent you https://github.com/DragonBluep/openwrt/pull/3

openwrt-ai (commented):

Both findings from the previous review are resolved, and I re-verified the fixes:

- `LINUX_KERNEL_HASH-6.12.101` / `LINUX_KERNEL_HASH-6.18.42` no longer say `skip`. Both values match `https://cdn.kernel.org/pub/linux/kernel/v6.x/sha256sums.asc` and I confirmed the 6.12.101 and 6.18.42 tarballs hash to exactly those digests locally.
- The two stale `MAINTAINERS` hunk offsets are now refreshed. `econet/.../910-v7.1-dt-bindings-clock-reset-add-econet-en751221-bindings.patch` (`8839` → `8852`) and `qualcommbe/.../0325-MAINTAINERS-Add-maintainer-for-Qualcomm-PPE-driver.patch` (`21239` → `21252`) both shift by +13, which matches the single `MAINTAINERS` hunk in the 6.18.41→42 incremental (13-line `GPU BUDDY ALLOCATOR` entry inserted at line 8656). 6.12.101 does not touch `MAINTAINERS`, so no 6.12-side change was needed.

Additionally, I applied the full patch stack of this PR against pristine `linux-6.12.101` and `linux-6.18.42` (generic backport → pending → hack → every `target/linux/*/patches-<ver>`): every patch applies at its recorded line numbers with zero offset and zero fuzz, so the "All other patches are automatically refreshed" claim holds for both series. (The only offsets seen anywhere were in `target/linux/uml/patches-6.12/101-mconsole-exec.patch` and `102-pseudo-random-mac.patch`; those are pre-existing on `main` — 6.12.101 touches neither `arch/um/`, `kernel/umh.c`, nor `include/linux/kmod.h` — so they are out of scope for this PR.)

CI on `f315dda` is green so far (no failed check runs; several builds still in progress).

---

danpawlik:
Tested on BPi-R4  and W1700K. On BPi-R4 all seems to be fine, but on Quantum Fiber W1700k - there is an issue with NPU load.
Currently:
```
root@W1700K:~# dmesg | grep -iE 'NPU fw version|NPU version'
[    5.858760] airoha-npu 1e900000.npu: NPU fw version: 0.0
```

where before:
```
root@W1700K:~# dmesg | grep -iE 'NPU fw version|NPU version'
[    5.879026] airoha-npu 1e900000.npu: NPU fw version: 0.1111
[   15.043460] mt7996e 0000:01:00.0: NPU version: 0.1111
```
Probably caused by https://github.com/torvalds/linux/commit/6f884eb87a79e0c482baef2ad96c96b81d024235

EDIT:
I will do one more test on clean env later.

EDIT:
example patch seems to work https://github.com/openwrt/openwrt/commit/5d6a8a731b19b0c4b7bf5c9e15b11db8a8a66d58

graysky2:
Build system: x86/64
Build-tested: x86/64-glibc
Run-tested: x86/64-glibc

hauke:
Please remove `CONFIG_DRM_BUDDY` it was removed and unset `CONFIG_GPU_BUDDY` now, see https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v6.18.42&id=65677f7a20c48cf9ed02c56db389cc327d2b61b2 

The NPU problem should also get fixed. If the suggested patch works, ad it as an extra OpenWrt commit.

danpawlik:
FYI: pushed patch to the kernel [mailing list](https://sashiko.dev/#/patchset/20260805070851.2885888-1-pawlik.dan%40gmail.com) - patchwork [link](https://patchwork.kernel.org/project/linux-mediatek/patch/20260805070851.2885888-1-pawlik.dan@gmail.com/)

DragonBluep:
> Please remove `CONFIG_DRM_BUDDY` it was removed 

@hauke I guess you mean `CONFIG_DRM_LIB_RANDOM`.

> and unset `CONFIG_GPU_BUDDY` now, see https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v6.18.42&id=65677f7a20c48cf9ed02c56db389cc327d2b61b2

`CONFIG_GPU_BUDDY` is selected by `CONFIG_DRM_BUDDY` on x86/legacy target. I think we should enable it.

danpawlik:
@hauke  At the moment, the NPU isn't working in the official builds - https://github.com/openwrt/openwrt/issues/22709 - so if anyone has this device, they're certainly using their own image. In that case, they can use my fix. So perhaps we shouldn't include the NPU fix for now and wait to see what @LorenzoBianconi has to say on the kernel mailing list.

Gilly1970:
@danpawlik I ran into the same issue when testing Wayen Yan's [Fix DMA direction for NPU mailbox buffer patch](https://patchwork.kernel.org/project/linux-mediatek/patch/178351055214.98729.11403147818632027428@gmail.com/) a few kernels back and his analysis is correct, so had this [patch](https://github.com/Gilly1970/Gemtek-W1700K-6.18/blob/main/openwrt-patches/970-net-airoha-npu-mailbox-dma-safe-buffer.patch) ready to go for when it hit main.

target/linux/airoha/patches-6.18/924-net-airoha-npu-use-coherent-mailbox-dma.patch
```diff
@@ -0,0 +1,56 @@
...
+ 	int ret;
+ 
+-	dma_addr = dma_map_single(npu->dev, p, size, DMA_BIDIRECTIONAL);
+-	ret = dma_mapping_error(npu->dev, dma_addr);
+-	if (ret)
+-		return ret;
++	/* Small cmd/rsp mailbox: coherent buffer avoids TO_DEVICE vs
++	 * BIDIRECTIONAL cache pitfalls on EN7581 (see 6f884eb87a79).
++	 */
++	dma_buf = dma_alloc_coherent(npu->dev, size, &dma_addr, GFP_ATOMIC);
```
openwrt-ai: This turns every mailbox message into a coherent alloc/free pair, and `airoha_npu_send_msg()` is not only a probe-time path. It is reached from `airoha_npu_foe_commit_entry()` — which already has to use [`kzalloc(..., GFP_ATOMIC)` at airoha_npu.c:354](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/ethernet/airoha/airoha_npu.c#L354) — via [`airoha_ppe_foe_commit_entry()` at airoha_ppe.c:710](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/ethernet/airoha/airoha_ppe.c#L710), which runs under [`spin_lock_bh(&ppe_lock)` at airoha_ppe.c:825](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/ethernet/airoha/airoha_ppe.c#L825) plus `rcu_read_lock()`, once per offloaded flow bind/unbind.

On non-coherent EN7581, `dma_alloc_coherent()` with `GFP_ATOMIC` is served from the DMA atomic pool (`coherent_pool=`, 128 KiB by default), so this adds a pool alloc/free to the flow-offload fast path and introduces a new `-ENOMEM` failure mode where the old code did no allocation at all. Note also that `airoha_npu_wlan_msg_send()/_get()` already thread a `gfp_t` down from their callers; hardcoding `GFP_ATOMIC` here discards that.

Suggest allocating the bounce buffer once at probe instead — one coherent buffer per mailbox core, sized to the largest message the driver can send, reused under the existing `npu->cores[core].lock`, with an early `-EINVAL` if `size` exceeds it. That keeps the coherency fix and removes the per-message allocation entirely.

---

target/linux/airoha/patches-6.18/924-net-airoha-npu-use-coherent-mailbox-dma.patch
```diff
@@ -0,0 +1,56 @@
...
+ 	spin_lock_bh(&npu->cores[core].lock);
+ 
+@@ -191,7 +195,9 @@ static int airoha_npu_send_msg(struct ai
+ 
+ 	spin_unlock_bh(&npu->cores[core].lock);
+ 
+-	dma_unmap_single(npu->dev, dma_addr, size, DMA_BIDIRECTIONAL);
++	if (!ret)
++		memcpy(p, dma_buf, size);
++	dma_free_coherent(npu->dev, size, dma_buf, dma_addr);
```
openwrt-ai: The lifetime semantics on the error path change here. `regmap_read_poll_timeout_atomic()` can return `-ETIMEDOUT` while the NPU is still going to write its response into the buffer; with the old `dma_map_single()` the backing memory was the caller's `kzalloc()`ed buffer, which stayed valid until the caller freed it. Now `dma_free_coherent()` returns the page to the coherent/atomic pool immediately, so a late NPU write lands on memory that may already have been handed to an unrelated DMA allocation.

Is that intentional? A probe-time-allocated, never-freed per-core buffer (see the comment on the alloc above) would avoid this as a side effect.

---

target/linux/airoha/patches-6.18/924-net-airoha-npu-use-coherent-mailbox-dma.patch
```diff
@@ -0,0 +1,56 @@
...
+extra invalidate, or mixed coherency on the mailbox path.
+
+Avoid streaming-DMA direction issues entirely for these small messages:
+copy through a dma_alloc_coherent() buffer for the mailbox transaction,
+then copy the response back to the caller's buffer. That matches how
+many drivers handle tiny command/response mailboxes and restores a
+stable 0.1111 version read with or without 6f884eb.
+
+Fixes: 6f884eb87a79 ("net: airoha: Fix DMA direction for NPU mailbox buffer")
+Signed-off-by: Daniel Pawlik <pawlik.dan@gmail.com>
```
openwrt-ai, line 33: nit: this patch partially reverts an upstream stable commit (6f884eb87a79, which landed in 6.18.42), so it needs a record of its upstream status or it will silently survive future bumps. You already posted it to linux-mediatek (https://patchwork.kernel.org/project/linux-mediatek/patch/20260805070851.2885888-1-pawlik.dan@gmail.com/ per your PR comment) — please carry that link in the patch so it can be dropped once the upstream discussion concludes.

```suggestion
Link: https://patchwork.kernel.org/project/linux-mediatek/patch/20260805070851.2885888-1-pawlik.dan@gmail.com/
Signed-off-by: Daniel Pawlik <pawlik.dan@gmail.com>
```

---

openwrt-ai (commented):
## Commit checks

- `f5007c4` "airoha: use coherent DMA for NPU mailbox buffers" — the body says "Refresh patch context for 6.18.42 DMA_BIDIRECTIONAL", but the commit adds exactly one new file (`924-net-airoha-npu-use-coherent-mailbox-dma.patch`) and refreshes no existing patch. Reads like a leftover from an earlier iteration; the sentence describes how the new patch was authored, not what the commit does.

Verified and found correct since the last review:

- The two `# CONFIG_DRM_LIB_RANDOM is not set` removals are right: `DRM_LIB_RANDOM` is gone from `drivers/gpu/drm/Kconfig` in both `v6.12.101` and `v6.18.42`.
- `CONFIG_GPU_BUDDY=y` in `target/linux/x86/legacy/config-6.18` is right and complete. In v6.18.42 `GPU_BUDDY` is a promptless `bool` in `drivers/gpu/Kconfig`, `select`ed by `DRM_BUDDY` in `drivers/gpu/drm/Kconfig`, so it only materialises where `DRM_BUDDY` is on. `x86/legacy/config-6.18:59` is the only in-tree config with `CONFIG_DRM_BUDDY=y`, and a promptless symbol correctly needs no `is not set` entry in `generic/config-6.18`.
- The new airoha patch applies cleanly on 6.18.42: after `102-02`, `102-03`, `118`, `121` and `123` (the other patches touching `airoha_npu.c`), both hunks of `924-...` apply at their recorded line numbers with zero offset and zero fuzz against the pristine v6.18.42 file. The `DMA_BIDIRECTIONAL` context the patch expects is real — v6.18.42 carries 6f884eb87a79 (`airoha_npu.c:157`/`:180`). CI's `Check Kernel patches (airoha, an7581)` agrees.
- `include/kernel-6.12` and `include/kernel-6.18` are byte-identical to the versions verified in the previous review (hashes unchanged), and the rebase of `4f09ab9`/`fa41694` onto the new base introduced no other content changes.

CI on `f5007c4` shows no failed check runs.

---

danpawlik:
Updated version that should satisfy Claude - https://github.com/openwrt/openwrt/commit/bacd81062b09c74be280732c2ee51bcdc9e12a0f 
Also pushed changes on KML.

EDIT:
[new push ](https://patchwork.kernel.org/project/linux-mediatek/patch/20260806142816.511515-1-pawlik.dan@gmail.com/) to fix issues raised in review. [Patch for openwrt ](https://github.com/openwrt/openwrt/commit/81922f443bcd17cfb5a71bf7b698d98a857d8f44)

graysky2:
to further complicate things. New upstream release today.

hauke:
> > Please remove `CONFIG_DRM_BUDDY` it was removed
> 
> @hauke I guess you mean `CONFIG_DRM_LIB_RANDOM`.
> 

Yes, you are right.

> > and unset `CONFIG_GPU_BUDDY` now, see https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v6.18.42&id=65677f7a20c48cf9ed02c56db389cc327d2b61b2
> 
> `CONFIG_GPU_BUDDY` is selected by `CONFIG_DRM_BUDDY` on x86/legacy target. I think we should enable it.

I am concerned about other targets which do not select it. If `CONFIG_GPU_BUDDY` is not selected explicitly the OpenWrt build system will ask if it should get build in. If x86 selects it, the target will overwrite it. We should handle this case and add  `# CONFIG_GPU_BUDDY is not set` into the generic configuration.

graysky2:
@DragonBluep - I just ran the update script and not a single patch was touched. It's just a checksum update.

```
diff --git a/target/linux/generic/kernel-6.18 b/target/linux/generic/kernel-6.18
index 30f0d1f254e5..adb86b5cb48d 100644
--- a/target/linux/generic/kernel-6.18
+++ b/target/linux/generic/kernel-6.18
@@ -1,2 +1,2 @@
-LINUX_VERSION-6.18 = .42
-LINUX_KERNEL_HASH-6.18.42 = 35a8ef5e8fb44467e56a42b40c23c4c10a61d1e95ff8d738f285230e8bed63e7
+LINUX_VERSION-6.18 = .43
+LINUX_KERNEL_HASH-6.18.43 = a1aeb926c7c4a1564368200b1082e45bb958007682d804aff88abbc3a7a47b5d
```

DragonBluep:
Update:
- Pull airoha v3 patch
- Add symbol `# CONFIG_GPU_BUDDY is not set`
- Update to 6.18.43 (.43 only has one emergency fix, just squash it)

target/linux/airoha/patches-6.18/924-net-airoha-npu-use-coherent-mailbox-dma.patch
```diff
@@ -0,0 +1,209 @@
+From: Daniel Pawlik <pawlik.dan@gmail.com>
```
openwrt-ai, line 2: nit: this new patch is missing the `From <hash> Mon Sep 17 00:00:00 2001` and `Date:` header lines, so it is not `git am`-applicable. Every other patch in this directory carries them, including ones taken from an unmerged mailing-list submission — e.g. [`920-16-net-airoha-Account-for-L2-overhead-in-PPE-MTU-configuration.patch:1-4`](``https://github.com/openwrt/openwrt/blob/9a29a11c0880beae102f53e444cf8d3dcf9810f1/target/linux/airoha/patches-6.18/920-16-net-airoha-Account-for-L2-overhead-in-PPE-MTU-configuration.patch#L1-L4)`` uses the all-zero hash placeholder. The FormalityCheck bot flagged this file for the same reason.

The `Date:` below matches the upstream v3 posting (`Message-ID: <20260806142816.511515-1-pawlik.dan@gmail.com>`).

```suggestion
From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
From: Daniel Pawlik <pawlik.dan@gmail.com>
Date: Thu, 6 Aug 2026 16:28:16 +0200
```


---

openwrt-ai (commented):
Reviewed 3 new commits (rebase onto `be2f856`, 6.18.42 squashed into 6.18.43, airoha patch refreshed to upstream v3); one nit inline, no blocking issues.

Verified since the last review:

- `LINUX_KERNEL_HASH-6.18.43 = a1aeb926c7c4a1564368200b1082e45bb958007682d804aff88abbc3a7a47b5d` and `LINUX_KERNEL_HASH-6.12.101 = 0d21cd11933f49f7151b7c9dbb8cc3fddc8c8abe506434b850feecf41fc28a76` both match `https://cdn.kernel.org/pub/linux/kernel/v6.x/sha256sums.asc`.
- The 6.18.42→43 incremental touches only `Makefile`, `arch/x86/entry/entry_64.S`, `arch/x86/include/asm/nospec-branch.h`, `arch/x86/kernel/cpu/bugs.c` and `arch/x86/lib/retpoline.S` (the SRSO safe-RET interruption fix). No patch in `target/linux/*/patches-6.18/` or `target/linux/generic/*-6.18/` touches any of those files, so squashing .42 into .43 with no patch churn is correct — matches what @graysky2 reported from the update script.
- `# CONFIG_GPU_BUDDY is not set` in both generic configs is correct and sufficient. In the 6.12.100→101 and 6.18.41→42 incrementals `GPU_BUDDY` is added as a promptless `bool` in the new `drivers/gpu/Kconfig`, `select`ed by `DRM_BUDDY`; `x86/legacy/config-6.18` is the only in-tree config with `CONFIG_DRM_BUDDY=y` and it already carries `CONFIG_GPU_BUDDY=y`, which overrides the generic entry. Placement is alphabetically correct in both files (`GPIO_ZEVIO` < `GPU_BUDDY` < `GP_PCI1XXXX`).
- The other Kconfig deltas in these two incrementals need no config entry: `GPU_BUDDY_KUNIT_TEST` depends on `KUNIT` (not set in either generic config), `CXL_RAS` is a `def_bool`, `RUSTC_HAS_SUSPICIOUS_RUNTIME_SYMBOL_DEFINITIONS` is promptless, and `NET_VENDOR_AIROHA`/`NET_AIROHA` (moved out of `drivers/net/ethernet/mediatek/` in 6.12.101) depend on `ARCH_AIROHA`, which is `# not set` in `generic/config-6.12` — the airoha target is 6.18-only, and no 6.12 patch references the airoha driver path.
- `924-net-airoha-npu-use-coherent-mailbox-dma.patch` is byte-identical to the upstream v3 posting (`20260806142816.511515-1-pawlik.dan@gmail.com`), diff and commit message alike, so the earlier stale "Refresh patch context" sentence is gone and the OpenWrt commit message now matches what the commit does.
- Spot-checked the v3 logic against v6.18.41/.42 sources: `sizeof(struct ppe_mbox_data)` really is 28, `offsetof(struct wlan_mbox_data, d)` is 8 so `rsp_off + rsp_len == size` for `airoha_npu_wlan_msg_get()`, `airoha_npu_ppe_stats_setup()` is the only PPE caller that reads the reply back (`stats_info.npu_stats_addr`) and is the only one passing a non-NULL `rsp`, and the new `dmam_alloc_coherent()` loop lands after `dma_set_coherent_mask(dev, DMA_BIT_MASK(32))` and reuses the existing `int i` in `airoha_npu_probe()`. The `AIROHA_NPU_MBOX_SIZE` 256-byte cap is safe for the out-of-tree caller too: the largest payload mt76 at `PKG_SOURCE_VERSION:=5967691` passes is the 16-byte `WLAN_FUNC_SET_WAIT_INODE_TXRX_REG_ADDR` struct in `npu.c`, everything else is `sizeof(u32)`.
- CI on `9a29a11` has no failed check runs; `FormalityCheck` and all completed `Check Kernel patches` jobs are green, builds still in progress.

---

target/linux/airoha/patches-6.18/924-net-airoha-npu-use-coherent-mailbox-dma.patch
```diff
@@ -0,0 +1,204 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Daniel Pawlik <pawlik.dan@gmail.com>
```
openwrt-ai, line 2: nit: only half of the git header got added — the `From <hash> Mon Sep 17 00:00:00 2001` line is there now, but `Date:` is still missing, so FormalityCheck still flags this file on `494560f` ("Missing required Git header parameters ... / 'Date:' / ..."). This is the only patch in `target/linux/airoha/patches-6.18/` that has a `From <hash>`/`From:`/`Subject:` header but no `Date:` — the other files that lack `Date:` (`885-i2c-mt7621-optional-reset.patch`, `886-uart-add-en7523-support.patch`, `900-airoha-bmt-support.patch`, `901-snand-mtk-bmt-support.patch`) have no git header at all and are not touched by this PR, which is why the bot doesn't flag them.

The date below is the `Date:` header of the v3 posting this patch is taken from (`Message-ID: <20260806142816.511515-1-pawlik.dan@gmail.com>`).

```suggestion
From: Daniel Pawlik <pawlik.dan@gmail.com>
Date: Thu, 6 Aug 2026 16:28:16 +0200
```

---

openwrt-ai (commented):
Reviewed 3 new commits (rebase onto `15a2a6f`, 6.12.101 squashed into 6.12.102, bcm27xx v3d GMP patch dropped, airoha patch re-refreshed quilt-style); one nit inline, no blocking issues.

Verified since the last review:

- `LINUX_KERNEL_HASH-6.12.102 = 694bb5f49371323ab9c2b7040bd358ed81bffbe80b372d76507045828f655803` and `LINUX_KERNEL_HASH-6.18.43 = a1aeb926c7c4a1564368200b1082e45bb958007682d804aff88abbc3a7a47b5d` both match `https://cdn.kernel.org/pub/linux/kernel/v6.x/sha256sums.asc`, and the downloaded tarballs hash to exactly those digests.
- Squashing .101 into .102 with no patch churn is correct. The 6.12.101→102 incremental is the same SRSO safe-RET fix as 6.18.42→43 and touches only `Makefile`, `arch/x86/entry/entry_64.S`, `arch/x86/include/asm/nospec-branch.h`, `arch/x86/kernel/cpu/bugs.c` and `arch/x86/lib/retpoline.S`. No patch under `target/linux/*/patches-6.12/` or `target/linux/generic/*-6.12/` touches any of those four x86 files; the only patch touching the top-level `Makefile` is `generic/pending-6.12/103-kbuild-export-SUBARCH.patch`, and the incremental only rewrites the `SUBLEVEL` value in place, so no line shift.
- New footnote `[3]` in the 6.18 commit checks out: `a096a6aba601152a45d4e272207bd37e03ca3f92` ("drm/v3d: Reach the GMP through the hub registers on V3D 7.x", upstream `f7df2da0d1c3`) is in v6.18.42, and its `v3d_idle_axi()` hunk is content-identical to the deleted `bcm27xx/patches-6.18/0866-...` (only `DRM_ERROR()` → `drm_err(&v3d->drm, ...)`), so dropping the downstream copy is right.
- The `+13` `MAINTAINERS` refresh on `bcm27xx/patches-6.18/0895-input-touchscreen-Add-Sitronix-ST7123-support.patch` (23715 → 23728) matches the 13-line `GPU BUDDY ALLOCATOR` insertion in the 6.18.41→42 incremental, and it is the only new-since-rebase patch that needed it.
- Re-ran the full apply test on the rebased tree: generic `backport`/`pending`/`hack` plus every `target/linux/*/patches-<ver>` applied against pristine `linux-6.12.102` and `linux-6.18.43`. Zero offsets and zero fuzz everywhere except the pre-existing `target/linux/uml/patches-6.12/101-mconsole-exec.patch` and `102-pseudo-random-mac.patch` (already off by -1 on `main`, and 6.12.102 touches neither `arch/um/` nor `kernel/umh.c`, so out of scope here). The only apply failures were my harness not staging `target/linux/<plat>/files/`, not patch problems.
- `# CONFIG_GPU_BUDDY is not set` survived the rebase in both generic configs at the right alphabetical position, `CONFIG_DRM_LIB_RANDOM` is gone from both, and `x86/legacy/config-6.18` is still the only in-tree config with `CONFIG_DRM_BUDDY=y` (it carries `CONFIG_GPU_BUDDY=y`). No 6.12 target config enables `DRM_BUDDY`, so no per-target 6.12 override is needed.
- The airoha patch body is unchanged from the version reviewed last night — the only edits are the added `From <hash>` line and a quilt-style refresh (dropped `diff --git`/`index`/`-- 2.55.0` lines, truncated `@@` function context). It still applies to v6.18.43 with zero offset after the other five patches touching `airoha_npu.c`.
- CI on `494560f` has no failed check runs (the run started at 15:39 UTC; almost everything is still queued).

---

openwrt-ai (commented):
Reviewed 3 new commits (rebase onto `2991205`, plus the `Date:` header added to the airoha patch); no new issues found.

Verified since the last review:

- The previous nit is resolved. `924-net-airoha-npu-use-coherent-mailbox-dma.patch` now carries the full `From <hash>` / `From:` / `Date:` / `Subject:` block, and the `Date:` value (`Thu, 6 Aug 2026 16:28:16 +0200`) matches the v3 posting the patch is taken from (`Message-ID: <20260806142816.511515-1-pawlik.dan@gmail.com>`). `FormalityCheck / Code Patches` is green on `2f4996d` and the file no longer appears in the bot's warning list — the remaining entries there are pre-existing files this PR does not touch.
- That one line is the *only* PR-authored change since `494560f`. I diffed `494560f..2f4996d` against the base delta `15a2a6f..2991205` and they cancel out exactly except for the `Date:` hunk; `range-diff` confirms commits 1 and 2 (`0e229e7`, `1f54e49`) are content-identical to `e738289` / `7cb8090`. So the kernel-bump halves need no re-verification beyond what was already checked (hashes, dropped patches, refreshed hunk offsets, full apply test against pristine `linux-6.12.102` / `linux-6.18.43`).
- The rebase does not invalidate anything. The five new base commits touch `target/linux/realtek/patches-6.18/300-02` and `300-03` (`arch/mips/generic/Platform`, `arch/mips/generic/board-realtek.c`), `target/linux/zynq/patches-6.18/0001-dts-add-antminer-s9.patch` (`arch/arm/boot/dts/xilinx/`), plus DTS and one config line. None of those files are touched by the 6.18.41→.42→.43 incrementals (which only change `Makefile`, `MAINTAINERS`, the new `drivers/gpu/Kconfig`, and the x86 SRSO files), so no additional patch refresh is owed by this bump.
- `# CONFIG_GPU_BUDDY is not set` survived the rebase in both generic configs at the correct alphabetical position (`config-6.12:2302`, `config-6.18:2434`, between `GPIO_ZEVIO` and `GP_PCI1XXXX`), and it does not collide with the `# CONFIG_SND_SOC_AN7581 is not set` line that base commit `f23aff6` added to `config-6.18`. `LINUX_KERNEL_HASH-6.12.102` / `-6.18.43` are unchanged from the values verified last night.
- No patch-number collision for `924-` in `target/linux/airoha/patches-6.18/`, and base commit `71f139a` ("an7583: fix ethernet breakage on Nokia XG-040G-MF") is DTS-only, so it does not interact with the NPU mailbox patch.
- CI on `2f4996d`: no failed check runs across all 246. `FormalityCheck`, every `Check Kernel patches` job (including `airoha, an7581`), and all completed builds are green; only `x86/64` package build is still in progress.

---

danpawlik:
Pushed updated version for NPU issue - https://github.com/openwrt/openwrt/commit/44e675692063138aaf0d669cfef7d986db8bf86c and on [kernel ML](https://patchwork.kernel.org/project/linux-mediatek/patch/20260809152813.585797-1-pawlik.dan@gmail.com/).

AuthorReflex:
6.18.44 and 6.12.103 is out

target/linux/generic/kernel-6.12
```diff
@@ -1,2 +1,2 @@
-LINUX_VERSION-6.12 = .100
-LINUX_KERNEL_HASH-6.12.100 = 67f973533406492e86774bacbcefae50d50d5c34cbf703c47ec526a5efdcee90
+LINUX_VERSION-6.12 = .103
```
openwrt-ai, line 1: `b9b59c8` says "All patches are automatically refreshed", but one 6.12 patch is left stale by this bump. The 6.12.102→.103 incremental inserts `select GENERIC_PINCONF` (at `config PINCTRL_BM1880`) and `select REGMAP_MMIO` (at `config PINCTRL_MICROCHIP_SGPIO`) into `drivers/pinctrl/Kconfig`, both above the anchor of [`008-pinctrl-add-driver-for-siflower-sf19a2890.patch`](``https://github.com/openwrt/openwrt/blob/faca01014276d2e94a0af7c8e7d6984be31e4561/target/linux/siflower/patches-6.12/008-pinctrl-add-driver-for-siflower-sf19a2890.patch#L12).`` Applying the full 6.12 stack against a pristine `linux-6.12.103` gives:

```
6.12/siflower 008-pinctrl-add-driver-for-siflower-sf19a2890.patch:
  Hunk #1 succeeded at 485 (offset 2 lines).
```

CI already caught this — `Check Kernel patches (siflower, sf21)` fails on `faca010`:

```
target/linux/siflower/patches-6.12/008-pinctrl-add-driver-for-siflower-sf19a2890.patch
Kernel patches for siflower/sf21 require refresh. (run 'make target/linux/refresh' and force push this pr)
```

Please run `make target/linux/refresh` with siflower/sf21 selected and squash the result into this commit. (The only other 6.12 offsets in the run are `target/linux/uml/patches-6.12/101-mconsole-exec.patch` and `102-pseudo-random-mac.patch` at -1, which are pre-existing on `main` — 6.12.103 touches neither `arch/um/` nor `kernel/umh.c` — so they are out of scope here.)

---

target/linux/generic/kernel-6.18
```diff
@@ -1,2 +1,2 @@
-LINUX_VERSION-6.18 = .41
-LINUX_KERNEL_HASH-6.18.41 = 17fc72f0f8d4a8a8633a5d20085f5d9c5a5ec51ee896a0b7ae1ec25da31273ea
+LINUX_VERSION-6.18 = .44
```
openwrt-ai, line 1: Same problem on the 6.18 side, and larger: `596a645` says "All patches are automatically refreshed", but five target patches still carry .43 line numbers. Applying generic `backport`/`pending`/`hack` plus every `target/linux/*/patches-6.18` against a pristine `linux-6.18.44` gives:

```
6.18/airoha    180-01-v7.2-pinctrl-Move-Airoha-driver-to-dedicated-directory.patch: Hunk #1 at 684 (offset 2)
6.18/realtek   803-pinctrl-Add-RTL8231-pin-control-and-GPIO-support.patch:          Hunk #1 at 533 (offset 2)
6.18/econet    781-v7.3-net-dsa-mt7530-add-en7528-support.patch:                    Hunks #1-#4 (offset -18)
6.18/ipq40xx   701-net-dsa-add-out-of-band-tagging-protocol.patch:                  Hunk #2 at 5122 (offset 6)
6.18/qualcommax 0950-net-dsa-add-out-of-band-tagging-protocol.patch:                Hunk #2 at 5122 (offset 6)
```

Each matches a 6.18.43→.44 hunk exactly: `drivers/pinctrl/Kconfig` gains two `select` lines above both pinctrl patches (+2), `drivers/net/dsa/mt7530.c` loses 18 lines before the econet hunks (-18), and `net/core/skbuff.c` is +12/-6 around `skb_clone_fraglist()`/`skb_pp_cow_data()` before hunk #2 of both out-of-band-tagging patches (+6). `make target/linux/refresh` for airoha, realtek, econet, ipq40xx and qualcommax should cover it.

---

target/linux/econet/Makefile
```diff
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 BOARD:=econet
 BOARDNAME:=EcoNet EN75xx MIPS
-FEATURES:=dt source-only squashfs nand usb
+FEATURES:=dt squashfs nand usb
```
openwrt-ai: Dropping `source-only` here (and from `target/linux/qualcommbe/Makefile:6`) is a target-policy change with no relation to a kernel bump, and `faca010` carries no body explaining it — which is also the CRITICAL error in the FormalityCheck comment. Both targets then fail their CI builds on this head:

```
    ERROR: package/kernel/linux failed to build.
Package kmod-mdio-devres is missing dependencies for the following libraries:
of_mdio.ko
make[2]: *** [modules/netdevices.mk:251: .../kmod-mdio-devres-6.18.44-r1.apk] Error 1
```

(`Build econet/en751221` and `Build qualcommbe/ipq95xx` are the only two build failures in the run.) If this was only meant to get CI to exercise the two targets for this bump, please drop the commit before merge; if it is a deliberate promotion, it belongs in its own PR with the packaging breakage fixed and a rationale in the message. Same applies to `target/linux/qualcommbe/Makefile:6`.

```suggestion
FEATURES:=dt source-only squashfs nand usb
```

---

target/linux/airoha/patches-6.18/924-net-airoha-npu-use-coherent-mailbox-dma.patch
```diff
@@ -0,0 +1,166 @@
...
```
openwrt-ai, line 2: This regressed relative to the previous push: the `From <hash> Mon Sep 17 00:00:00 2001` and `Date:` lines that were added in `2f4996d` are gone again, so FormalityCheck flags this file once more under `2893f8d`. It is also now the only patch of the 146 in `target/linux/airoha/patches-6.18/` that keeps raw `git format-patch` artifacts — `diff --git`/`index <sha>..<sha>` at lines 34-35 and 150-151, and the `-- ` / `2.55.0` trailer at lines 164-165 — which a `make target/linux/refresh` with airoha/an7581 selected would strip, restoring the quilt-style form the rest of the directory uses.

The date below is derived from the `20260809152813.585797-1-pawlik.dan@gmail.com` posting referenced in the Aug 9 comment (the same way the previous `Date:` was derived from `20260806142816`); adjust it if the embedded diff is a different revision.

```suggestion
From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
From: Daniel Pawlik <pawlik.dan@gmail.com>
Date: Sun, 9 Aug 2026 17:28:13 +0200
```

---

target/linux/generic/kernel-6.18
```diff
@@ -1,2 +1,2 @@
...
+LINUX_KERNEL_HASH-6.18.44 = 0f72d938f06828e82c90405174fe572287db7bfe089e2fc46572a99a7f240d43
```
openwrt-ai, line 2: nit: the PR title still says "bump to 6.12.102 and 6.18.43" while the branch now lands `.103`/`.44`. Related: the series carries `d90a26e` (→6.12.102) plus `b9b59c8` (→6.12.103) and `bb25a7d` (→6.18.43) plus `596a645` (→6.18.44), so two versions that never land on their own get one commit each. Earlier iterations of this PR squashed the intermediate release instead (".43 only has one emergency fix, just squash it"). Was leaving them separate this time intentional?

For the record, both new hashes are correct: `LINUX_KERNEL_HASH-6.12.103 = f143aaad…` and `LINUX_KERNEL_HASH-6.18.44 = 0f72d938…` match `https://cdn.kernel.org/pub/linux/kernel/v6.x/sha256sums.asc`, and the downloaded tarballs hash to exactly those digests. `CONFIG_EROFS_FS_ZIP_LZMA_DEFAULT_MAX_STREAMS=16` is also right — upstream `0c676903cb2a` adds it as `int … default 16` under `EROFS_FS_ZIP_LZMA`.

---

openwrt-ai (commented):
## Commit checks

- `b9b59c8` "kernel: bump 6.12 to 6.12.103" — body says "All patches are automatically refreshed", but `siflower/patches-6.12/008-pinctrl-add-driver-for-siflower-sf19a2890.patch` still needs one (details inline on `target/linux/generic/kernel-6.12`).
- `596a645` "kernel: bump 6.18 to 6.18.44" — same claim, five target patches unrefreshed (airoha 180-01, realtek 803, econet 781, ipq40xx 701, qualcommax 0950 — details inline on `target/linux/generic/kernel-6.18`).
- `faca010` "target: build econet and qualcommbe" — empty body, and the change (dropping `source-only` from two targets) is unrelated to a kernel bump; both targets then fail to build in CI. Details inline on `target/linux/econet/Makefile`.
- `2893f8d` "airoha: use coherent DMA for NPU mailbox buffers" — the OpenWrt commit body still carries only the `Link:` to the first patchwork posting (`20260805070851`), while the patch it adds is several revisions later; worth updating so the message points at the revision actually vendored.

Verified and found correct since the last review:

- Both new hashes match `sha256sums.asc` and the downloaded tarballs; `CONFIG_EROFS_FS_ZIP_LZMA_DEFAULT_MAX_STREAMS=16` matches upstream `0c676903cb2a` / `5aaa06dfc10f` (`int`, `default 16`, `depends on EROFS_FS_ZIP_LZMA`), and its placement is alphabetically correct in both generic configs.
- `d90a26e` and `bb25a7d` are content-identical to the previously reviewed `0e229e7` / `1f54e49` (`range-diff` `=`), so the .102/.43 halves need no re-verification.
- Full apply test against pristine `linux-6.12.103` and `linux-6.18.44`: generic `backport`/`pending`/`hack` apply with zero offset and zero fuzz for both series, and every `target/linux/*/patches-<ver>` applies without a single rejected hunk. The only offsets anywhere are the six listed inline plus the pre-existing `uml` pair.
- The reworked airoha patch is still functionally sound. `__airoha_npu_send_msg()`'s `memcpy(reply, core->buf + len - reply_len, reply_len)` lands exactly on `wlan_mbox_data.d`, because `airoha_npu_wlan_msg_get()` uses `len = sizeof(*wlan_data) + data_len` and `sizeof(struct wlan_mbox_data) == offsetof(…, d) == 8`; the `airoha_npu_send_msg()` wrapper passing `(data, len)` reproduces the old `DMA_BIDIRECTIONAL` write-back semantics for the PPE callers, and the `len > AIROHA_NPU_MBOX_SIZE` / `reply_len > len` guards can't be tripped by any in-tree caller (`sizeof(struct ppe_mbox_data)` is 28; the largest WLAN payload is 16).

---

target/linux/airoha/patches-6.18/924-net-airoha-npu-use-coherent-mailbox-dma.patch
```diff
@@ -0,0 +1,161 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Daniel Pawlik <pawlik.dan@gmail.com>
+Date: Sun,  9 Aug 2026 17:28:13 +0200
```
openwrt-ai, line 3: nit: the vendored patch is now the 2026-08-09 posting (`20260809152813.585797-1-pawlik.dan@gmail.com`, matching this `Date:` header), but the OpenWrt commit message of `e6b57b2` still carries only `Link: .../20260805070851.2885888-1-pawlik.dan@gmail.com/` — the very first revision. The patch's own `Link:` trailers on lines 29-31 already list v1/v2/v3, so the commit-message link is the one that is three revisions stale; pointing it at the revision actually vendored would make it easier to check the OpenWrt copy against the list posting.

Same commit body says "Allocate one device-managed coherent bounce buffer at probe" while the patch allocates one per NPU core (`for (i = 0; i < ARRAY_SIZE(npu->cores); i++)`), as the patch's own message correctly words it.

No functional concern: I applied both the previously-reviewed version of this file and the current one on top of the same generic+airoha stack against pristine `linux-6.18.44`, and the resulting `drivers/net/ethernet/airoha/airoha_npu.c` and `include/linux/soc/airoha/airoha_offload.h` are byte-identical — the only change here is the added `From <hash>`/`Date:` headers plus the quilt-style re-render.

---

openwrt-ai (commented):
Reviewed 4 new commits (new `qualcommbe` VIDEO_QCOM_IRIS commit, `target: build econet and qualcommbe` dropped, the six stale patch offsets refreshed, airoha patch re-headered); one nit inline, no blocking issues.

All three findings from the previous review are resolved:

- **The six unrefreshed patches are now refreshed, and every offset is exactly right.** I extracted the `patch-6.12.102-103` / `patch-6.18.43-44` incrementals and computed the cumulative line deltas above each hunk: `drivers/pinctrl/Kconfig` gains +1 at line 162 and +1 at 415 in 6.18 (→ airoha `180-01` 682→684, realtek `803` 531→533) and +1 at 150 and +1 at 367 in 6.12 (→ siflower `008` 483→485); `net/core/skbuff.c` is +12/−6 around line 895 for a net +6 (→ ipq40xx `701` and qualcommax `0950` both 5112→5118); `drivers/net/dsa/mt7530.c` loses 18 lines across the `en7581`/`mt753x` phylink hunks (→ econet `781` 2847→2829, 3036→3018, 3174→3156, 3377→3359).
- **`target: build econet and qualcommbe` is gone.** `target/linux/econet/Makefile` and `target/linux/qualcommbe/Makefile` are now byte-identical to the base branch (`source-only` restored on both), so the CI build failures that commit caused are gone.
- The empty-bodied commit is replaced by `e2b4fb9` with a proper message.

Verified since the last review:

- `LINUX_KERNEL_HASH-6.12.103 = f143aaade8877ba5616e788b4482576db28481bcf557ef537f4fcc3938fc3176` and `LINUX_KERNEL_HASH-6.18.44 = 0f72d938f06828e82c90405174fe572287db7bfe089e2fc46572a99a7f240d43` match `sha256sums.asc`, and the downloaded tarballs hash to exactly those digests.
- **Full apply test on both series.** generic `backport`/`pending`/`hack` plus every `target/linux/*/patches-<ver>` applied against pristine `linux-6.12.103` and `linux-6.18.44`: zero fuzz and zero offsets everywhere except the pre-existing `target/linux/uml/patches-6.12/101-mconsole-exec.patch` and `102-pseudo-random-mac.patch` (already −1 on `main`; those hunks live in `kernel/umh.c`, `include/linux/kmod.h` and `arch/um/drivers/mconsole*`, none of which the .101→.103 incrementals touch, so out of scope here). So "All patches are automatically refreshed" now holds for both bumps. The single apply failure was `mediatek/patches-6.18/863-arm64-dts-mt7986-add-sound-wm8960.patch`, which needs `mt7986a-rfb-spim-nand.dts` that my harness did not stage — not a patch problem, and `Check Kernel patches (mediatek, filogic)` is green.
- `CONFIG_EROFS_FS_ZIP_LZMA_DEFAULT_MAX_STREAMS=16` is correct in both generic configs. Both footnote IDs resolve on the stable tree with matching subject "erofs: cap LZMA stream pool size" (`0c676903cb2a` in v6.12.103, `5aaa06dfc10f` in v6.18.44), the symbol really is added by those incrementals only, and it is `int` / `depends on EROFS_FS_ZIP_LZMA` / `range 1 NR_CPUS` / `default 16` — so `16` is the upstream default. Placement is alphabetically correct in both files (`EROFS_FS_ZIP_DEFLATE` < `..._LZMA_DEFAULT_MAX_STREAMS` < `EROFS_FS_ZIP_ZSTD`). No in-tree config enables `EROFS_FS_ZIP`, so the symbol is inert.
- `e2b4fb9` checks out: [`VIDEO_QCOM_IRIS`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/media/platform/qcom/iris/Kconfig#L1) exists in v6.18.41 and does not exist anywhere in 6.12, so "introduced in the 6.18 LTS kernel" is accurate, and it is gated on `ARCH_QCOM || COMPILE_TEST`. Sister-target parity is now complete: [`qualcommax/config-6.18:656`](https://github.com/openwrt/openwrt/blob/0c2c9eba2c348e4bf3e9f2de261933623f9e16ec/target/linux/qualcommax/config-6.18#L656) and [`ipq40xx/config-6.18:535`](https://github.com/openwrt/openwrt/blob/0c2c9eba2c348e4bf3e9f2de261933623f9e16ec/target/linux/ipq40xx/config-6.18#L535) already carried the line, and those three are the only `CONFIG_ARCH_QCOM=y` targets on 6.18. Placement is alphabetically correct (`VDSO_GETRANDOM` < `VIDEO_QCOM_IRIS` < `VIRTIO`).
- The airoha patch is a pure re-header: the `From <hash>`/`Date:` lines are added back (all-zero placeholder hash, an existing in-tree convention used by 16 other patches) and the `diff --git`/`index`/`-- 2.55.0` lines dropped. Applying the old and new versions on the same stack yields byte-identical sources, so nothing about the code needs re-verification. `From`/`From:`/`Date:`/`Subject:` are all present, and the file no longer appears in the formality bot's warning list.
- CI on `0c2c9eb`: all six workflow runs conclude `success`, no failed jobs in `Build Kernel` (30 jobs) or `Build all core packages` (9 jobs).

---

danpawlik (approved):
Works well on BPi-R4 and Quantum Fiber W1700k.

openwrt-ai (commented):
Reviewed 6 new commits (rebase onto `a27a3fea`; the only PR-authored change since the last review is the airoha commit message). No new issues found.

- **The previous nit is resolved.** `ac1992ca` now links the posting it actually vendors: the patch's `Date: Sun,  9 Aug 2026 17:28:13 +0200` is 15:28:13 UTC, which matches `Message-ID: <20260809152813.585797-1-pawlik.dan@gmail.com>` in the new `Link:`, and @danpawlik confirmed in [this comment](https://github.com/openwrt/openwrt/pull/24551#issuecomment-5232423288) that this is the current revision. The reworded body also matches the code now: "one device-managed coherent bounce buffer **per NPU core** … under the existing **per-core** mailbox lock" is exactly what the [`dmam_alloc_coherent()` loop over `ARRAY_SIZE(npu->cores)`](``https://github.com/openwrt/openwrt/blob/55cfc132a41465f60ee875984f303ac1c7aef7a6/target/linux/airoha/patches-6.18/924-net-airoha-npu-use-coherent-mailbox-dma.patch#L133-L140)`` does, and every `core->buf` access sits inside `spin_lock_bh(&core->lock)`. The stale "Avoids per-message GFP_ATOMIC coherent allocs" sentence is gone.
- **That message edit is the only PR-authored change since `0c2c9eb`.** `git range-diff 486b4a4b..0c2c9eba a27a3fea..55cfc13` marks commits 1, 2, 3, 5 and 6 as `=`, and `target/linux/airoha/patches-6.18/924-…patch` is byte-identical between the two heads — so the patch code, the dropped/rebased patches and the refreshed hunk offsets verified in the previous reviews need no re-verification. `LINUX_KERNEL_HASH-6.12.103 = f143aaad…` and `LINUX_KERNEL_HASH-6.18.44 = 0f72d938…` are likewise unchanged from the values already checked against `sha256sums.asc`.
- **One new base commit interacts with this PR, and it is harmless.** Base `3d25753` adds [`921-net-airoha-npu-load-firmware-without-the-sysfs-fallb.patch`](``https://github.com/openwrt/openwrt/blob/55cfc132a41465f60ee875984f303ac1c7aef7a6/target/linux/airoha/patches-6.18/921-net-airoha-npu-load-firmware-without-the-sysfs-fallb.patch),`` which touches the same file as `924-`. It applies first, is a one-line-for-one-line replacement (`request_firmware()` → `request_firmware_direct()`) inside `airoha_npu_load_firmware()`, so it shifts nothing, and its hunk lies outside every context block of `924-`. `924-`'s recorded line numbers stay valid, and there is no `924-` numbering collision.
- **The rest of the rebase does not invalidate the "all patches are automatically refreshed" claim.** The new base commits add or rewrite a large amount of patch material written against 6.18.41 / 6.12.100 (airoha pinctrl sync to linux-7.3, bcm53xx backports, the generic HORACO SFP quirk, the SmartRG MT7988 DTS restructure, the Hasivo/PoE realtek work). CI's `Check Kernel patches` job is the `make target/linux/refresh` + no-diff check, and it is green on every checked target — e.g. `Check Kernel patches (airoha, an7581)` ends with `Kernel patches for airoha/an7581 seems ok`. The only targets that job cannot cover are the `source-only` ones (econet, qualcommbe); their patches are unchanged since the head where I last apply-tested them against pristine `linux-6.12.103` / `linux-6.18.44` with zero offset and zero fuzz.
- **CI on `55cfc13` is fully green:** `Build Kernel`, `Build all core packages` and `Build PR Profile` all conclude `success`, and `FormalityCheck / Git & Commits`, `/ Code Patches` and `/ OpenWrt Makefiles` are green. The formality bot's remaining warnings are all pre-existing patch files lacking `git am` headers that these bump commits merely refresh — none are introduced here.

---

geldot:
Would be great to land this to leave time for testing and backport to stable as the delta has grown pretty big. Is there anything remaining that needs assistance?

PR #24555, opened 2026-08-03
mediatek: filogic: add support for Zyxel WX3400-T0

pccr10001:
The Zyxel WX3400-T0 is a dual-band Wi-Fi 6 mesh extender based on the MediaTek MT7986B.

### Hardware:
- MediaTek MT7986B quad-core Cortex-A53 SoC
- 512 MB DDR3 RAM
- 256 MB Winbond SPI-NAND flash
- 2.4 GHz 4x4 and 5 GHz 4x4 802.11ax wireless
- 2x Gigabit Ethernet ports using Airoha EN8801SN PHYs
- Reset and WPS buttons
- 6 GPIO-controlled LED channels
- UART header on rear side of PCB, pin order `3.3V TX RX _ GND`,
  115200 8N1
- 12 V, 2 A DC input using a 5.5 x 2.1 mm barrel jack

### MAC addresses
- LAN: 48:ed:e6:xx:xx:xx (Factory 0x24, label-mac-device)
- WAN: 48:ed:e6:xx:xx:xx (Factory 0x2a)
- Wi-Fi 2.4 GHz: 48:ed:e6:xx:xx:xx (Factory 0x4 + 0)
- Wi-Fi 5 GHz: 48:ed:e6:xx:xx:xx (Factory 0x4 + 1)

### Official firmware

The stock WebUI accepts a vendor tar archive encrypted with AES-256-CBC
and the legacy OpenSSL MD5 key derivation. The factory image reproduces
this format and contains an initramfs FIT, a minimal zydefault volume
and a generated zyfwinfo header.

The stock bootloader uses two UBI firmware banks. The sysupgrade code reads
rootubi from the kernel command line, validates and increments the active
zyfwinfo sequence number, and installs the persistent kernel and root
filesystem into ubi (mtd5). The stock bootloader remains unchanged.

### Installation
1. Open the stock "Firmware Upgrade" page at
   http://192.168.0.1/FirmwareUpgrade
2. Select initramfs-factory.bin and click "Upgrade", then wait for reboot.
3. Open LuCI at http://192.168.1.1/
4. Open "System -> Backup / Flash Firmware" and flash sysupgrade.bin without
   preserving settings.

### Reverting to stock

A complete stock backup made before installation is required. The stock WebUI
may write the initramfs image to either ubi (mtd5) or ubi2 (mtd6), while
sysupgrade always writes persistent OpenWrt to ubi (mtd5). If the WebUI
selected ubi2, both stock banks are overwritten after sysupgrade. The active
bank cannot be determined reliably before upload, so booting ubi2 is not a
supported revert method. Restore the saved stock backup through a suitable
bootloader or recovery procedure.

### DIFF
--- a/package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic
+++ b/package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic
@@ -179,13 +179,19 @@ xiaomi,redmi-router-ax6000-stock)
 zyxel,ex5601-t0)
 	ubootenv_add_mtd "u-boot-env" "0x0" "0x20000" "0x40000" "2"
 	;;
-zyxel,ex5601-t0-ubootmod|\
-zyxel,wx5600-t0-ubootmod)
+zyxel,ex5601-t0-ubootmod)
 	ubootenv_add_ubi_default
 	ubootenv_add_sys_mtd "u-boot-env" "0x0" "0x20000" "0x40000" "2"
 	;;
 zyxel,ex5700-telenor)
 	ubootenv_add_uci_config "/dev/ubootenv" "0x0" "0x4000" "0x4000" "1"
+	;;
+zyxel,wx3400-t0)
+	ubootenv_add_mtd "u-boot-env" "0x0" "0x20000" "0x20000" "1"
+	;;
+zyxel,wx5600-t0-ubootmod)
+	ubootenv_add_ubi_default
+	ubootenv_add_sys_mtd "u-boot-env" "0x0" "0x20000" "0x40000" "2"
 	;;
 esac
 
--- a/target/linux/mediatek/dts/mt7986b-zyxel-wx3400-t0.dts
+++ b/target/linux/mediatek/dts/mt7986b-zyxel-wx3400-t0.dts
@@ -1,0 +1,326 @@
+// SPDX-License-Identifier: (GPL-2.0 OR MIT)
+
+/dts-v1/;
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+#include <dt-bindings/pinctrl/mt65xx.h>
+
+#include "mt7986b.dtsi"
+
+/ {
+	model = "Zyxel WX3400-T0";
+	compatible = "zyxel,wx3400-t0", "mediatek,mt7986b";
+
+	aliases {
+		serial0 = &uart0;
+		label-mac-device = &gmac0;
+		led-boot = &led_green_power;
+		led-failsafe = &led_red_power;
+		led-running = &led_green_power;
+		led-upgrade = &led_red_power;
+	};
+
+	chosen {
+		stdout-path = "serial0:115200n8";
+	};
+
+	memory@40000000 {
+		device_type = "memory";
+		reg = <0 0x40000000 0 0x20000000>;
+	};
+
+	keys {
+		compatible = "gpio-keys";
+		poll-interval = <20>;
+
+		reset-button {
+			label = "reset";
+			gpios = <&pio 9 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
+
+		wps-button {
+			label = "wps";
+			gpios = <&pio 10 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_WPS_BUTTON>;
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		led_green_power: green-power {
+			gpios = <&pio 12 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_POWER;
+		};
+
+		led_red_power: red-power {
+			gpios = <&pio 15 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_RED>;
+			function = LED_FUNCTION_POWER;
+			panic-indicator;
+		};
+
+		green-signal {
+			gpios = <&pio 26 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_INDICATOR;
+		};
+
+		red-signal {
+			gpios = <&pio 32 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_RED>;
+			function = LED_FUNCTION_INDICATOR;
+		};
+
+		green-wlan {
+			gpios = <&pio 17 GPIO_ACTIVE_HIGH>;
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WLAN_2GHZ;
+		};
+
+		red-wlan {
+			gpios = <&pio 13 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_RED>;
+			function = LED_FUNCTION_WLAN_5GHZ;
+		};
+	};
+};
+
+&spi0 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&spi_flash_pins>;
+	cs-gpios = <0>, <0>;
+	status = "okay";
+
+	spi_nand: spi-nand@1 {
+		compatible = "spi-nand";
+		reg = <1>;
+		spi-max-frequency = <20000000>;
+		spi-tx-bus-width = <4>;
+		spi-rx-bus-width = <4>;
+		mediatek,nmbm;
+		mediatek,bmt-max-ratio = <1>;
+		mediatek,bmt-max-reserved-blocks = <64>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "BL2";
+				reg = <0x0 0x100000>;
+				read-only;
+			};
+
+			partition@100000 {
+				label = "u-boot-env";
+				reg = <0x100000 0x80000>;
+			};
+
+			factory: partition@180000 {
+				label = "Factory";
+				reg = <0x180000 0x200000>;
+				read-only;
+
+				nvmem-layout {
+					compatible = "fixed-layout";
+					#address-cells = <1>;
+					#size-cells = <1>;
+
+					eeprom_factory: eeprom@0 {
+						reg = <0x0 0x1000>;
+					};
+
+					macaddr_factory_4: macaddr@4 {
+						compatible = "mac-base";
+						reg = <0x4 0x6>;
+						#nvmem-cell-cells = <1>;
+					};
+
+					macaddr_factory_24: macaddr@24 {
+						compatible = "mac-base";
+						reg = <0x24 0x6>;
+						#nvmem-cell-cells = <1>;
+					};
+
+					macaddr_factory_2a: macaddr@2a {
+						compatible = "mac-base";
+						reg = <0x2a 0x6>;
+						#nvmem-cell-cells = <1>;
+					};
+				};
+			};
+
+			partition@380000 {
+				label = "FIP";
+				reg = <0x380000 0x1c0000>;
+				read-only;
+			};
+
+			partition@540000 {
+				label = "zloader";
+				reg = <0x540000 0x40000>;
+				read-only;
+			};
+
+			partition@580000 {
+				compatible = "linux,ubi";
+				label = "ubi";
+				reg = <0x580000 0x4000000>;
+			};
+
+			partition@4580000 {
+				label = "ubi2";
+				reg = <0x4580000 0x4000000>;
+			};
+
+			partition@8580000 {
+				label = "zyubi";
+				reg = <0x8580000 0x6a80000>;
+			};
+		};
+	};
+};
+
+&eth {
+	status = "okay";
+
+	gmac0: mac@0 {
+		compatible = "mediatek,eth-mac";
+		reg = <0>;
+		label = "lan";
+		nvmem-cells = <&macaddr_factory_24 0>;
+		nvmem-cell-names = "mac-address";
+		phy-mode = "sgmii";
+		phy-handle = <&phy0>;
+		managed = "in-band-status";
+		pause;
+	};
+
+	gmac1: mac@1 {
+		compatible = "mediatek,eth-mac";
+		reg = <1>;
+		label = "wan";
+		nvmem-cells = <&macaddr_factory_2a 0>;
+		nvmem-cell-names = "mac-address";
+		phy-mode = "sgmii";
+		phy-handle = <&phy1>;
+		managed = "in-band-status";
+		pause;
+	};
+
+	mdio: mdio-bus {
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		phy0: ethernet-phy@18 {
+			compatible = "ethernet-phy-id03a2.9471";
+			reg = <0x18>;
+			phy-mode = "sgmii";
+			reset-gpios = <&pio 6 GPIO_ACTIVE_LOW>;
+			reset-assert-us = <10000>;
+		};
+
+		phy1: ethernet-phy@1a {
+			compatible = "ethernet-phy-id03a2.9471";
+			reg = <0x1a>;
+			phy-mode = "sgmii";
+			reset-gpios = <&pio 16 GPIO_ACTIVE_LOW>;
+			reset-assert-us = <10000>;
+		};
+	};
+};
+
+&uart0 {
+	status = "okay";
+};
+
+&watchdog {
+	status = "okay";
+};
+
+&wifi {
+	#address-cells = <1>;
+	#size-cells = <0>;
+
+	status = "okay";
+	pinctrl-names = "default", "dbdc";
+	pinctrl-0 = <&wf_2g_5g_pins>;
+	pinctrl-1 = <&wf_dbdc_pins>;
+	nvmem-cells = <&eeprom_factory>;
+	nvmem-cell-names = "eeprom";
+
+	band@0 {
+		reg = <0>;
+		nvmem-cells = <&macaddr_factory_4 0>;
+		nvmem-cell-names = "mac-address";
+	};
+
+	band@1 {
+		reg = <1>;
+		nvmem-cells = <&macaddr_factory_4 1>;
+		nvmem-cell-names = "mac-address";
+	};
+};
+
+&pio {
+	spi_flash_pins: spi-flash-pins-33-to-38 {
+		mux {
+			function = "spi";
+			groups = "spi0", "spi0_wp_hold";
+		};
+
+		conf-pu {
+			pins = "SPI2_CS", "SPI2_HOLD", "SPI2_WP";
+			drive-strength = <MTK_DRIVE_8mA>;
+			mediatek,pull-up-adv = <0>;
+		};
+
+		conf-pd {
+			pins = "SPI2_CLK", "SPI2_MOSI", "SPI2_MISO";
+			drive-strength = <MTK_DRIVE_8mA>;
+			mediatek,pull-down-adv = <0>;
+		};
+	};
+
+	wf_2g_5g_pins: wf-2g-5g-pins {
+		mux {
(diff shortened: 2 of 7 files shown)


target/linux/mediatek/dts/mt7986b-zyxel-wx3400-t0.dts
```diff
@@ -0,0 +1,333 @@
...
+		led-upgrade = &led_green_wps;
+	};
+
+	chosen {
+		stdout-path = "serial0:115200n8";
+	};
+
+	memory@40000000 {
+		device_type = "memory";
+		reg = <0 0x40000000 0 0x10000000>;
```
openwrt-ai: `0x10000000` is 256 MiB, but the commit message and PR body both say "512 MB RAM". One of the two is wrong — either half the RAM is being thrown away here, or the hardware description is overstated. If the board really has 512 MB, this should be `0x20000000`.

```suggestion
		reg = <0 0x40000000 0 0x20000000>;
```
(If the board has 256 MB, drop this suggestion and fix the commit message instead.)

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -3974,3 +4026,26 @@ ifneq ($(CONFIG_TARGET_ROOTFS_INITRAMFS),)
...
+  KERNEL := kernel-bin | lzma | \
+	fit lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb | pad-to 64k
+  KERNEL_INITRAMFS := kernel-bin | lzma | \
+	fit lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb with-initrd | pad-to 64k
+  KERNEL_INITRAMFS_SUFFIX := .itb
+
+	IMAGES := initramfs-factory.bin sysupgrade.bin
+	IMAGE/initramfs-factory.bin := append-image-stage initramfs.itb | \
+		sysupgrade-tar kernel=$$$$@ | zyxel-wx3400-factory
+	IMAGE/sysupgrade.bin := sysupgrade-tar | append-metadata
```
openwrt-ai: Two problems in this block:

1. `initramfs-factory.bin` is added to `IMAGES` unconditionally, so `append-image-stage initramfs.itb` runs even when `CONFIG_TARGET_ROOTFS_INITRAMFS` is off (the buildbot default) or under the ImageBuilder, where the initramfs image does not exist and the build fails. Every other device in this file that consumes an initramfs stage guards it, e.g. [`dlink_aquila-pro-ai-e30-a1` in filogic.mk:1662-1667](https://github.com/openwrt/openwrt/blob/68228973f529dd3668e96d26bd1bb0c549c6b395/target/linux/mediatek/image/filogic.mk#L1662-L1667).
2. These four lines are indented with a TAB while the rest of the block (and every other `define Device/...` in this file) uses two spaces.

Also note `IMAGES :=` is redundant here — [`Device/Default` in target/linux/mediatek/image/Makefile:31](https://github.com/openwrt/openwrt/blob/68228973f529dd3668e96d26bd1bb0c549c6b395/target/linux/mediatek/image/Makefile#L31) already sets `IMAGES := sysupgrade.bin`.

```suggestion
  IMAGE/sysupgrade.bin := sysupgrade-tar | append-metadata
ifeq ($(IB),)
ifneq ($(CONFIG_TARGET_ROOTFS_INITRAMFS),)
  IMAGES += initramfs-factory.bin
  IMAGE/initramfs-factory.bin := append-image-stage initramfs.itb | \
    sysupgrade-tar kernel=$$$$@ | zyxel-wx3400-factory
endif
endif
```

---

package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic
```diff
@@ -179,6 +179,9 @@ xiaomi,redmi-router-ax6000-stock)
 zyxel,ex5601-t0)
 	ubootenv_add_mtd "u-boot-env" "0x0" "0x20000" "0x40000" "2"
 	;;
+zyxel,wx3400-t0)
+	ubootenv_add_mtd "u-boot-env" "0x0" "0x20000" "0x20000" "1"
+	;;
```
openwrt-ai, line 191: nit: this entry is inserted in the middle of the `zyxel,ex5601-t0*` group, breaking the alphabetical ordering of the `case`. It belongs after `zyxel,ex5700-telenor)` (line 190-192), i.e. before `zyxel,wx5600-t0-ubootmod`.

Same ordering issue in the other three files — the new board is appended *after* `zyxel,wx5600-t0-ubootmod` in [`target/linux/mediatek/base-files/lib/preinit/05_set_preinit_iface:44`](``https://github.com/openwrt/openwrt/blob/68228973f529dd3668e96d26bd1bb0c549c6b395/target/linux/mediatek/base-files/lib/preinit/05_set_preinit_iface#L44),`` [`target/linux/mediatek/filogic/base-files/etc/board.d/02_network:246`](https://github.com/openwrt/openwrt/blob/68228973f529dd3668e96d26bd1bb0c549c6b395/target/linux/mediatek/filogic/base-files/etc/board.d/02_network#L246) and in `filogic.mk` (`define Device/zyxel_wx3400-t0` after `zyxel_wx5600-t0-ubootmod`), where `wx3400` should sort before `wx5600`.

---

target/linux/mediatek/dts/mt7986b-zyxel-wx3400-t0.dts
```diff
@@ -0,0 +1,333 @@
...
+			partition@0 {
+				label = "BL2";
+				reg = <0x0 0x100000>;
+				read-only;
+			};
+
+			partition@100000 {
+				label = "u-boot-env";
+				reg = <0x100000 0x80000>;
+				read-only;
```
openwrt-ai: The `u-boot-env` partition is marked `read-only`, which clears `MTD_WRITEABLE` and makes `fw_setenv` fail — yet the PR adds `uboot-envtools` to `DEVICE_PACKAGES` and registers a writable env at [`package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic:183`](``https://github.com/openwrt/openwrt/blob/68228973f529dd3668e96d26bd1bb0c549c6b395/package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic#L183).`` Compare `zyxel,ex5601-t0`, whose DTS leaves the env partition writable. Is the env meant to be writable (drop `read-only;`), or read-only for `fw_printenv` use only? Nothing in the upgrade path reads the env — `rootubi` comes from `/proc/cmdline` — so it's not obvious which is intended.

---

target/linux/mediatek/dts/mt7986b-zyxel-wx3400-t0.dts
```diff
@@ -0,0 +1,333 @@
...
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		led_green_power: green-power {
+			gpios = <&pio 12 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_BOOT;
```
openwrt-ai: nit: the node is named `green-power` but the function is `boot`, so the LED shows up as `green:boot`. `LED_FUNCTION_POWER` exists and matches both the node name and the hardware; the boot/running behaviour is already expressed through the `led-boot`/`led-running` aliases above, which don't depend on the `function` string.

```suggestion
			function = LED_FUNCTION_POWER;
```

Same applies to `red-power` on line 63.

---

target/linux/mediatek/dts/mt7986b-zyxel-wx3400-t0.dts
```diff
@@ -0,0 +1,333 @@
...
+			gpios = <&pio 15 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_RED>;
+			function = LED_FUNCTION_BOOT;
+			panic-indicator;
+		};
+
+		green-signal {
+			gpios = <&pio 26 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_LAN;
```
openwrt-ai: These two LEDs are named `*-signal` (the mesh-extender link-quality indicator per the commit message) but get `LED_FUNCTION_LAN`, so they surface as `green:lan` and `red:lan-1` — which reads as Ethernet port activity and is misleading, especially on a board that also has two real Ethernet ports. `LED_FUNCTION_INDICATOR` is what comparable in-tree boards use for signal-strength LEDs (see [`mt7981b-alwaylink-m01k43.dts:102-121`](https://github.com/openwrt/openwrt/blob/68228973f529dd3668e96d26bd1bb0c549c6b395/target/linux/mediatek/dts/mt7981b-alwaylink-m01k43.dts#L102-L121)); `function = "signal";` is also used in-tree where no constant fits. Same applies to `red-signal` on line 76.

---

target/linux/mediatek/dts/mt7986b-zyxel-wx3400-t0.dts
```diff
@@ -0,0 +1,333 @@
...
+			phy-mode = "sgmii";
+			reset-gpios = <&pio 16 GPIO_ACTIVE_LOW>;
+			reset-assert-us = <10000>;
+		};
+	};
+};
+
+&ssusb {
+	status = "disabled";
+};
```
openwrt-ai: This is a no-op: `ssusb` is already `status = "disabled"` in the SoC dtsi ([mt7986a.dtsi:373](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm64/boot/dts/mediatek/mt7986a.dtsi#L373), included via `mt7986b.dtsi`), and no other mt7986b board in the tree re-disables it. Drop the override.

---

target/linux/mediatek/dts/mt7986b-zyxel-wx3400-t0.dts
```diff
@@ -0,0 +1,333 @@
...
+	pinctrl-0 = <&wf_2g_5g_pins>;
+	pinctrl-1 = <&wf_dbdc_pins>;
+	nvmem-cells = <&eeprom_factory>;
+	nvmem-cell-names = "eeprom";
+
+	band@1 {
+		reg = <1>;
+		nvmem-cells = <&macaddr_factory_4 1>;
+		nvmem-cell-names = "mac-address";
+	};
```
openwrt-ai, line 268: nit: only `band@1` (5 GHz) gets a MAC; `band@0` (2.4 GHz) is left to whatever is in the EEPROM. The sibling mt7986b DBDC board assigns both — see [`mt7986b-buffalo-wsr-6000ax8.dts:404-413`](``https://github.com/openwrt/openwrt/blob/68228973f529dd3668e96d26bd1bb0c549c6b395/target/linux/mediatek/dts/mt7986b-buffalo-wsr-6000ax8.dts#L404-L413).`` Is the 2.4 GHz MAC already correct in the EEPROM, or should `band@0` take `<&macaddr_factory_4 0>`? Worth documenting either way in the MAC address section of the commit message.

---

openwrt-ai (commented):
## Commit checks

- 6822897 "mediatek: filogic: add support for Zyxel WX3400-T0" — the commit message is missing the **MAC address layout** section required for new-device commits. The DTS derives three MACs from the `Factory` partition (`0x24` → LAN/`gmac0`, `0x2a` → WAN/`gmac1`, `0x4` +1 → 5 GHz `band@1`), and none of that is documented; please add one line per interface using `xx:xx:xx` for the unit-specific bytes, and say where the 2.4 GHz MAC comes from.
- 6822897 — the **hardware specification** section omits UART (header location, pinout, baud rate) and the power supply (voltage + current), and the RAM/flash entries give no chip designations. The **flash instructions** describe the WebUI path but have no revert-to-stock procedure and no verbatim strings (stock IP/hostname, menu label of the firmware-upload page).
- 6822897 — "512 MB RAM" in the message contradicts the DTS `memory@40000000` node, which declares 256 MiB. See the inline comment.


---

target/linux/mediatek/dts/mt7986b-zyxel-wx3400-t0.dts
```diff
@@ -0,0 +1,326 @@
...
+			gpios = <&pio 26 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WAN;
+		};
+
+		red-net {
+			gpios = <&pio 32 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_RED>;
+			function = LED_FUNCTION_LAN;
+		};
```
openwrt-ai: The node names say `net` but the functions say `wan` and `lan`, so these export as `green:wan` and `red:lan`, and [`01_leds`](``https://github.com/openwrt/openwrt/blob/05a4e855ff731d506b870d827b2a2252c39550c2/target/linux/mediatek/filogic/base-files/etc/board.d/01_leds#L387-L389``) then drives them from two *different* netdevs (`wan` and `lan`).

If `green-net`/`red-net` are the two colours of one bi-colour network indicator — which the shared `-net` suffix and the previous `-signal` naming both suggest — then both colours light up whenever both Ethernet links are up, mixing to amber and making the indicator unreadable. In that case only one function/trigger should be assigned to the pair.

If they are two physically separate LEDs, then the node names should follow the in-tree `<color>-<function>` convention and match what they actually export (`green-wan` / `red-lan`), rather than a third name that appears nowhere else. Which of the two is it?

---

target/linux/mediatek/dts/mt7986b-zyxel-wx3400-t0.dts
```diff
@@ -0,0 +1,326 @@
...
+
+		led_green_wlan: green-wlan {
```
openwrt-ai: nit: the `led_green_wlan` label is dead — nothing references it any more now that `led-upgrade` points at `&led_red_power` (line 22). `led_green_power` and `led_red_power` are the only labels the aliases need.

```suggestion
		green-wlan {
```

---

openwrt-ai (commented):
Re-reviewed at `05a4e85`. The previous round's findings are addressed: RAM is now `0x20000000`, the `u-boot-env` partition is writable, the redundant `&ssusb` override is gone, `band@0` gets a MAC, the LED functions were reworked, the `initramfs-factory.bin` recipe is guarded by `IB`/`CONFIG_TARGET_ROOTFS_INITRAMFS` with the file's two-space indentation, and the `case` entries are alphabetically ordered in all four files. Two remaining points inline.

## Commit checks

- 05a4e85 "mediatek: filogic: add support for Zyxel WX3400-T0" — the commit message still has no **MAC address layout** section. The DTS now sources four MACs from the `Factory` partition (`0x24` → LAN/`gmac0`, `0x2a` → WAN/`gmac1`, `0x4`+0 → 2.4 GHz `band@0`, `0x4`+1 → 5 GHz `band@1`, with `label-mac-device = &gmac0`); please document one line per interface using `xx:xx:xx` for the unit-specific bytes.
- 05a4e85 — the **flash instructions** in the commit message are less specific than the ones in the PR description: the message only says "uploading the initramfs-factory.bin image through the stock WebUI", while the PR body carries the verbatim `http://192.168.0.1/FirmwareUpgrade` URL, the `Upgrade` button label and the post-flash `http://192.168.1.1` address. Please fold those verbatim strings into the commit message, and add a note on reverting to stock if a procedure exists (the `ubi2` recovery bank is left untouched — is booting it a supported way back?).

---

target/linux/mediatek/dts/mt7986b-zyxel-wx3400-t0.dts
pccr10001: They are the two colors of one physical bi-color signal indicator. Both channels now use LED_FUNCTION_INDICATOR, and the separate WAN/LAN netdev triggers have been removed.

target/linux/mediatek/dts/mt7986b-zyxel-wx3400-t0.dts
```diff
@@ -0,0 +1,327 @@
...
+			gpios = <&pio 26 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_INDICATOR;
+		};
+
+		red-signal {
+			gpios = <&pio 32 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_RED>;
+			function = LED_FUNCTION_INDICATOR;
+			function-enumerator = <1>;
```
openwrt-ai: nit: the enumerator is only on one half of the pair, so the two channels of the same bi-colour indicator export under inconsistent names: `green-signal` has no `function-enumerator` and becomes `green:indicator`, while this node becomes `red:indicator-1` ([`led_compose_name()` in led-core.c:552-563](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/leds/led-core.c#L552-L563) appends `-%d` only when the property is present).

Since the two nodes already differ by `color`, the composed names don't collide and the enumerator isn't needed at all. In-tree groups that do use it number every member of the group rather than just one — e.g. [`mt7981b-airtel-aap4221zy.dts:53-79`](https://github.com/openwrt/openwrt/blob/c70c563eca573a35edb78f2dc1edab4449928426/target/linux/mediatek/dts/mt7981b-airtel-aap4221zy.dts#L53-L79) uses `0`/`1`/`2` across its three `LED_FUNCTION_INDICATOR` LEDs.

```suggestion
			function = LED_FUNCTION_INDICATOR;
```
(Alternatively keep it and add `function-enumerator = <0>;` to `green-signal` for symmetry.)

---

openwrt-ai (commented):
Re-reviewed at `c70c563` (delta since `05a4e85`). The previous round's points are addressed: the bi-colour indicator pair is now `green-signal`/`red-signal` on `LED_FUNCTION_INDICATOR` with the separate WAN/LAN netdev triggers dropped from `01_leds`, and the dead `led_green_wlan` label is gone. The commit message now also carries the MAC address layout, the verbatim flash instructions (`http://192.168.0.1/FirmwareUpgrade`, `Upgrade`, `http://192.168.1.1`) and a reverting-to-stock section, so the earlier commit-message findings are resolved.

One minor point inline. FormalityCheck is green on this head.

---

target/linux/mediatek/dts/mt7986b-zyxel-wx3400-t0.dts
```diff
@@ -0,0 +1,326 @@
...
+		};
+
+		wps-button {
+			label = "wps";
+			gpios = <&pio 10 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_WPS_BUTTON>;
+		};
+	};
+
+	leds {
```
openwrt-ai, line 51: nit: the commit message and PR body both say "7 GPIO-controlled LED channels", but this node defines 6 (`green-power`, `red-power`, `green-signal`, `red-signal`, `green-wlan`, `red-wlan`), and no other GPIO in the file is an LED — `pio 9`/`pio 10` are the buttons and `pio 6`/`pio 16` are the two PHY resets. Is the 7th channel real hardware that is deliberately left undescribed (in which case a short note would help), or is "7" a typo for 6 in the hardware specification?

---

openwrt-ai (commented):
Re-reviewed at `9d23649` (delta since `c70c563`). The only change is the removal of `function-enumerator = <1>;` from `red-signal`, which resolves the previous round's point — the two channels of the bi-colour indicator now export symmetrically as `green:indicator` / `red:indicator`. The commit message is unchanged and still matches the diff.

One minor point inline. All three FormalityCheck runs are green on this head.

---

target/linux/mediatek/dts/mt7986b-zyxel-wx3400-t0.dts
pccr10001: updated in description.

openwrt-ai (commented):
Re-reviewed at `2a1d3b8` (delta since `9d23649`). The tree is byte-identical (`git diff 9d23649..2a1d3b8` is empty); the only change is in the commit message, where "7 GPIO-controlled LED channels" became "6", matching the six nodes in the `leds` block and the PR body. That resolves the previous round's point.

Reviewed 1 new commit; no new issues found. All three FormalityCheck runs are green on this head.

---

PR #24556, opened 2026-08-03
qualcommax: ipq50xx: add support for USB3

georgemoussalem:
Add driver support for the Qualcomm UNIPHY USB Super-Speed 22ULL PHY found on IPQ5018/IPQ5332 SoCs and enable USB3 support.


target/linux/qualcommax/patches-6.18/0820-dt-bindings-net-wireless-ath11k-add-bindings-for-QCN6122.patch
```diff
@@ -22,15 +22,15 @@ Signed-off-by: George Moussalem <george.moussalem@outlook.com>
 ---
 --- a/Documentation/devicetree/bindings/net/wireless/qcom,ath11k.yaml
 +++ b/Documentation/devicetree/bindings/net/wireless/qcom,ath11k.yaml
-@@ -21,6 +21,7 @@ properties:
+@@ -22,6 +22,7 @@ properties:
```
openwrt-ai: These hunk headers were correct before this PR and are wrong now. Nothing in this series touches `Documentation/devicetree/bindings/net/wireless/qcom,ath11k.yaml`, so its line numbers cannot have moved. In 6.18.41 (`LINUX_VERSION-6.18 = .41`) the three context lines live at [`qcom,ath11k.yaml:21-23`](``https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/Documentation/devicetree/bindings/net/wireless/qcom%2Cath11k.yaml#L21-L23``) and the DP20/DP21/DP22 block at [lines 268-270](``https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/Documentation/devicetree/bindings/net/wireless/qcom%2Cath11k.yaml#L268-L270``) — i.e. the original `-21` / `-268` offsets. Note also the two offsets moved in opposite directions (+1 and -6), which is a good sign this was refreshed against a tree carrying extra out-of-series ath11k patches.

CI agrees:

```
target/linux/qualcommax/patches-6.18/0820-dt-bindings-net-wireless-ath11k-add-bindings-for-QCN6122.patch
target/linux/qualcommax/patches-6.18/0821-dt-bindings-phy-document-the-qualcomm-ipq5332-uniphy-usb-super-speed-phy.patch
Kernel patches for qualcommax/ipq50xx require refresh. (run 'make target/linux/refresh' and force push this pr)
```

Simplest fix is to drop this file from the commit entirely (it is unrelated to USB3 and unmentioned in the commit message); otherwise restore both headers:

```suggestion
@@ -21,6 +21,7 @@ properties:
```

Same applies to line 33 below, which should read `@@ -268,6 +269,29 @@ allOf:`.

---

target/linux/qualcommax/patches-6.18/0821-dt-bindings-phy-document-the-qualcomm-ipq5332-uniphy-usb-super-speed-phy.patch
```diff
@@ -0,0 +1,124 @@
...
+---
+ .../phy/qcom,ipq5332-uniphy-usb-ss-phy.yaml        | 97 ++++++++++++++++++++++
+ 1 file changed, 97 insertions(+)
+
+diff --git a/Documentation/devicetree/bindings/phy/qcom,ipq5332-uniphy-usb-ss-phy.yaml b/Documentation/devicetree/bindings/phy/qcom,ipq5332-uniphy-usb-ss-phy.yaml
+new file mode 100644
+index 000000000000..a26b880dc2d3
+--- /dev/null
++++ b/Documentation/devicetree/bindings/phy/qcom,ipq5332-uniphy-usb-ss-phy.yaml
+@@ -0,0 +1,97 @@
```
openwrt-ai: This is raw `git format-patch` output rather than a quilt-refreshed patch: it still carries `diff --git`, `new file mode 100644`, `index 000000000000..a26b880dc2d3` and the trailing `-- ` / `2.53.0` signature at the end of the file. None of the other 11 patches added by this commit have them, and CI fails on exactly this file:

```
target/linux/qualcommax/patches-6.18/0820-...-QCN6122.patch
target/linux/qualcommax/patches-6.18/0821-dt-bindings-phy-document-the-qualcomm-ipq5332-uniphy-usb-super-speed-phy.patch
Kernel patches for qualcommax/ipq50xx require refresh. (run 'make target/linux/refresh' and force push this pr)
```

Per the project convention for `target/linux/<platform>/patches-<X.Y>/`, regenerate with `make target/linux/refresh` (ipq50xx selected in `.config`) so the file keeps only the mail header, the `--- /dev/null` / `+++ b/...` pair and the hunk.

---

target/linux/qualcommax/patches-6.18/0821-dt-bindings-phy-document-the-qualcomm-ipq5332-uniphy-usb-super-speed-phy.patch
```diff
@@ -0,0 +1,124 @@
...
++  compatible:
++    enum:
++      - qcom,ipq5018-uniphy-usb-ss-phy
++      - qcom,ipq5332-uniphy-usb-ss-phy
++
++  reg:
++    maxItems: 1
++
++  clocks:
++    maxItems: 3
```
openwrt-ai, line 51: `maxItems` without `minItems` makes dtschema treat the count as fixed at 3, so the two-clock `qcom,ipq5018-uniphy-usb-ss-phy` case added in 0823 can never validate (the `allOf` branch below lists only 2 `items` for it, which is itself unsatisfiable against `minItems: 3`). The sister binding for the same SoC pair spells both bounds out — see [`qcom,ipq5332-uniphy-pcie-phy.yaml:25-27`](``https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/Documentation/devicetree/bindings/phy/qcom%2Cipq5332-uniphy-pcie-phy.yaml#L25-L27).``

```suggestion
+  clocks:
+    minItems: 2
+    maxItems: 3
```


---

target/linux/qualcommax/patches-6.18/0823-arm64-dts-qcom-ipq5018-add-usb-super-speed-phy-node.patch
```diff
@@ -0,0 +1,50 @@
...
+@@ -189,6 +189,21 @@
+ 			status = "disabled";
+ 		};
+ 
++		usbphy1: phy@5d000 {
++			compatible = "qcom,ipq5018-uniphy-usb-ss-phy";
++			reg = <0x0005d000 0x800>;
++
++			clocks = <&gcc GCC_USB0_PIPE_CLK>,
++				 <&gcc GCC_USB0_PHY_CFG_AHB_CLK>;
```
openwrt-ai: Clock order contradicts the binding this series adds. `0821-...-usb-super-speed-phy.patch` (lines 86-88) declares the `qcom,ipq5018-uniphy-usb-ss-phy` list as `USB3 PHY CFG AHB clock` then `USB3 pipe clock`, and its ipq5332 branch plus the example DTS also put the pipe clock last. Here the pipe clock comes first. The driver uses `devm_clk_bulk_get_all()` so it does not care at runtime, but the DTS won't pass `dtbs_check` against the new schema. Aligning the DTS with the binding:

```suggestion
+			clocks = <&gcc GCC_USB0_PHY_CFG_AHB_CLK>,
+				 <&gcc GCC_USB0_PIPE_CLK>;
```


---

target/linux/qualcommax/patches-6.18/0822-phy-qcom-introduce-qualcomm-ipq5332-super-speed-usb-uniphy-driver.patch
```diff
@@ -0,0 +1,333 @@
...
++	for (i = 0; i < cfg->num_init_seq; i++)
++		writel(tbl[i].val, base + tbl[i].offset);
++
++	return 0;
++
++err_assert_reset:
++	/* Assert phy reset */
++	reset_control_assert(uniphy->reset);
++
++	return ret;
```
openwrt-ai, line 193: Two bugs in this block:

1. `ret` is never assigned from `reset_control_assert()` / `reset_control_deassert()`. Both `if (ret)` tests read the value left over from the successful `regulator_enable()` above, so they are always false — a reset failure is silently swallowed and the compiler can't warn.
2. The regulator is leaked on every error path. `regulator_enable()` is refcounted, so the two early `return ret;` and the `err_assert_reset` path all leave the 5V supply on and the count unbalanced; a later `.power_off` then underflows it.

```suggestion
+	/* Perform phy reset */
+	ret = reset_control_assert(uniphy->reset);
+	if (ret) {
+		dev_err(dev, "Failed to assert reset\n");
+		goto err_disable_vreg;
+	}
+
+	usleep_range(1, 5);
+
+	ret = reset_control_deassert(uniphy->reset);
+	if (ret) {
+		dev_err(dev, "Failed to deassert reset\n");
+		goto err_disable_vreg;
+	}
+
+	ret = clk_bulk_prepare_enable(uniphy->num_clks, uniphy->clks);
+	if (ret) {
+		dev_err(dev, "failed to enable clocks: %d\n", ret);
+		goto err_assert_reset;
+	}
+
+	/* phy autoload delay */
+	usleep_range(35, 40);
+
+	for (i = 0; i < cfg->num_init_seq; i++)
+		writel(tbl[i].val, base + tbl[i].offset);
+
+	return 0;
+
+err_assert_reset:
+	/* Assert phy reset */
+	reset_control_assert(uniphy->reset);
+err_disable_vreg:
+	regulator_disable(uniphy->vreg);
+
+	return ret;
```

This grows the hunk by two lines, so re-run `make target/linux/refresh` afterwards to fix the `@@ -0,0 +1,275 @@` header and the diffstat.

---

target/linux/qualcommax/patches-6.18/0822-phy-qcom-introduce-qualcomm-ipq5332-super-speed-usb-uniphy-driver.patch
```diff
@@ -0,0 +1,333 @@
...
++struct uniphy_usb {
++	struct device		*dev;
++	const struct uniphy_cfg	*cfg;
++	struct phy		*phy;
++	void __iomem		*base;
++	struct clk_bulk_data	*clks;
++	unsigned int		num_clks;
++	struct reset_control	*reset;
++	struct regulator	*vreg;
++	struct clk_fixed_rate	pipe_clk_fixed;
```
openwrt-ai: nit: `pipe_clk_fixed` is never used — `phy_pipe_clk_register()` uses `devm_clk_hw_register_fixed_rate()`, which allocates its own `struct clk_fixed_rate`. Dropping this member also lets you drop the `#include <linux/clk-provider.h>`-adjacent baggage. While you're there, `#include <linux/of_device.h>` (added last, after `linux/reset.h`) breaks the otherwise alphabetical include block and is a legacy header — `of_device_get_match_data()` is declared in `linux/of.h`, which is already included.

---

target/linux/qualcommax/dts/ipq5018-mr5500.dts
```diff
@@ -274,17 +274,20 @@
 	vdd-supply = <&regulator_fixed_5p0>;
 };
 
-&usb {
+&usbphy1 {
 	status = "okay";
 
-	vbus-supply = <&regulator_fixed_5p0>;
+	vdd-supply = <&regulator_fixed_5p0>;
```
openwrt-ai, line 280: This hunk silently drops `vbus-supply = <&regulator_fixed_5p0>;` from the `&usb` controller node — it isn't a rename, the property is gone. The commit message only mentions adding the UNIPHY node and the 500000 -> 5000000 microvolt correction, so this looks unintended. In practice the 5V rail still comes up because both `&usbphy0` and `&usbphy1` now hold a `vdd-supply` reference to it, but VBUS control is no longer tied to the controller's own lifecycle. Was the removal deliberate? If so, please say so in the commit message; otherwise keep `vbus-supply` on `&usb` alongside the new `&usbphy1` block.

---

openwrt-ai (commented):
## Commit checks

- 40c5822 "qualcommax: ipq50xx: add support for USB3" — the message describes only "add driver support for the ... 22ULL PHY" plus the AUX/LFPS clocks, but the diff also (a) adds an upstream backport `0091-v6.19-...-Correct-IPQ5018-interrupts.patch` reworking the DWC3 interrupt bindings, (b) converts the IPQ5018 USB controller from the nested `qcom,dwc3` model to the flattened `qcom,snps-dwc3` model (0824), which is a user-visible DT ABI change, (c) adds a new `qcom,phy-usb-mux-sel` DT property and dwc3-qcom glue change (0827/0828), and (d) modifies the unrelated ath11k QCN6122 binding patch 0820. Worth either splitting or expanding the message so the flattening and the TCSR mux work are documented.
- 7a89bc8 "qualcommax: ipq50xx: enable USB3 support on Linksys MR5500" — message covers the UNIPHY node and the 5V regulator correction, but not the removal of `vbus-supply` from the USB controller node (inline comment below).


---

kuncy7:
Confirming for the record: you have my

Signed-off-by: Stanislaw Pal <kuncy7@gmail.com>

for the phy-mux-selection binding and driver patches (v2 11/17 and
12/17), and thanks for the credit.

Hardware datapoint for this PR: the series as published on your branch
runs on a TP-Link Archer AX55 v1 (IPQ5018) with kernel 6.18.39. Two
different USB 3.2 gen 1 sticks enumerate at SuperSpeed (`speed` 5000,
`version` 3.20), zero USB/xHCI errors in dmesg, and booting with a stick
already inserted is clean. This board is the discriminating one for the
mux ordering: its bootloader leaves the TCSR mux on PCIe, so before the
mux-before-reset arrangement the port was permanently stuck in Rx.Detect
and fell back to USB2.

Nice catch on the MR5500 regulator voltage, and glad the `usb_dwc` /
shared-config bits made it in - with those this ports cleanly.

target/linux/qualcommax/patches-6.18/0822-phy-qcom-introduce-qualcomm-ipq5332-super-speed-usb-uniphy-driver.patch
```diff
@@ -0,0 +1,321 @@
...
++	for (i = 0; i < cfg->num_init_seq; i++)
++		writel(tbl[i].val, base + tbl[i].offset);
++
++	return 0;
++
++err_assert_reset:
++	/* Assert phy reset */
++	reset_control_assert(uniphy->reset);
++
++	return ret;
```
openwrt-ai, line 193: The missing `ret =` on the two `reset_control_*()` calls is fixed now, but the regulator imbalance in the same block is still there. `regulator_enable()` is refcounted, and none of the three failure exits below it undo it:

- `return ret;` after a failed `reset_control_assert()`
- `return ret;` after a failed `reset_control_deassert()`
- the `err_assert_reset:` label, taken when `clk_bulk_prepare_enable()` fails

Each leaves the 5V supply enabled with the enable count bumped. Since `uniphy_usb_init()` is `.power_on` and `uniphy_usb_shutdown()` (`.power_off`) calls `regulator_disable()` unconditionally, a failed power-on followed by a later power-off underflows the count and trips the `WARN_ON(!rdev->use_count)` in [`_regulator_disable()`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/regulator/core.c#L3022-L3024) — and the rail stays on in the meantime.

```suggestion
+	/* Perform phy reset */
+	ret = reset_control_assert(uniphy->reset);
+	if (ret) {
+		dev_err(dev, "Failed to assert reset\n");
+		goto err_disable_vreg;
+	}
+
+	usleep_range(1, 5);
+
+	ret = reset_control_deassert(uniphy->reset);
+	if (ret) {
+		dev_err(dev, "Failed to deassert reset\n");
+		goto err_disable_vreg;
+	}
+
+	ret = clk_bulk_prepare_enable(uniphy->num_clks, uniphy->clks);
+	if (ret) {
+		dev_err(dev, "failed to enable clocks: %d\n", ret);
+		goto err_assert_reset;
+	}
+
+	/* phy autoload delay */
+	usleep_range(35, 40);
+
+	for (i = 0; i < cfg->num_init_seq; i++)
+		writel(tbl[i].val, base + tbl[i].offset);
+
+	return 0;
+
+err_assert_reset:
+	/* Assert phy reset */
+	reset_control_assert(uniphy->reset);
+err_disable_vreg:
+	regulator_disable(uniphy->vreg);
+
+	return ret;
```

This adds two lines to the payload, so the `@@ -0,0 +1,263 @@` header and the `263`/`274` counts in the diffstat need regenerating afterwards (`make target/linux/refresh` with ipq50xx selected).

---

openwrt-ai (commented):
## Commit checks

- 624b27f "qualcommax: ipq50xx: add support for USB3" — the stray `0820-...-QCN6122.patch` hunk-offset churn is correctly gone from this revision, but the message body is unchanged and still describes only "add driver support for the ... 22ULL PHY" plus the AUX/LFPS clocks. The diff also (a) adds the upstream backport `0091-v6.19-dt-bindings-usb-qcom-snps-dwc3-Correct-IPQ5018-interrupts.patch`, (b) flattens the IPQ5018 USB controller from the nested `qcom,dwc3` model to `qcom,snps-dwc3` (0824), which moves the node from `usb@8af8800` to `usb@8a00000` and drops the `usb_dwc` child — a user-visible DT ABI change that board DTS files outside this tree will notice, and (c) introduces the new `qcom,phy-usb-mux-sel` DT property plus the dwc3-qcom glue change (0827/0828). Worth documenting the flattening and the TCSR mux work in the message, or splitting them into their own commits.
- 91ece60 "qualcommax: ipq50xx: enable USB3 support on Linksys MR5500" — message covers the UNIPHY node and the 500000 → 5000000 microvolt correction, but not the removal of `vbus-supply` from `&usb` (see the still-open inline thread on `ipq5018-mr5500.dts`).

Fixed since the last review: the raw `git format-patch` artifacts in 0821, the unassigned `ret` on the reset calls in 0822, the redundant `qcom_uniphy_usb_clk_init()` wrapper, and the 0820 offsets. `Check Kernel patches` is green on this head.

Still open from the previous round and unchanged in these commits: `clocks: maxItems: 3` without `minItems` in 0821, the pipe/CFG-AHB clock ordering in 0823 vs the binding, the unused `pipe_clk_fixed` member and legacy `linux/of_device.h` include in 0822, and the `vbus-supply` removal in `ipq5018-mr5500.dts`.

---

target/linux/qualcommax/patches-6.18/0821-dt-bindings-phy-document-the-qualcomm-ipq5332-uniphy-usb-super-speed-phy.patch
```diff
@@ -0,0 +1,125 @@
...
+
+Signed-off-by: Praveenkumar I <quic_ipkumar@quicinc.com>
+Signed-off-by: George Moussalem <george.moussalem@outlook.com>
+---
+ .../phy/qcom,ipq5332-uniphy-usb-ss-phy.yaml        | 98 ++++++++++++++++++++++
+ 1 file changed, 98 insertions(+)
+
+diff --git a/Documentation/devicetree/bindings/phy/qcom,ipq5332-uniphy-usb-ss-phy.yaml b/Documentation/devicetree/bindings/phy/qcom,ipq5332-uniphy-usb-ss-phy.yaml
+new file mode 100644
+index 000000000000..c83075fd0473
```
openwrt-ai, line 21: Regression: the raw `git format-patch` artifacts are back in this file. They were removed in the previous revision (91ece60) and this revision re-adds `diff --git`, `new file mode 100644`, `index 000000000000..c83075fd0473` here plus the trailing `-- ` / `2.53.0` signature at lines 124-125. None of the other 10 patches added by this commit carry them, and quilt strips them on refresh — which is exactly why CI names this one file and nothing else:

```
$ git diff --name-only
target/linux/qualcommax/patches-6.18/0821-dt-bindings-phy-document-the-qualcomm-ipq5332-uniphy-usb-super-speed-phy.patch
Kernel patches for qualcommax/ipq50xx require refresh. (run 'make target/linux/refresh' and force push this pr)
```
([Check Kernel patches (qualcommax, ipq50xx)](https://github.com/openwrt/openwrt/actions/runs/30989090839/job/92250908576) on this head)

Dropping the three lines below fixes half of it; also delete lines 123-125 (the blank line, `-- ` and `2.53.0`). Easiest is to just re-run `make target/linux/refresh` with ipq50xx selected in `.config` so the file keeps only the mail header, the `--- /dev/null` / `+++ b/...` pair and the hunk.

```suggestion
```


---

target/linux/qualcommax/patches-6.18/0821-dt-bindings-phy-document-the-qualcomm-ipq5332-uniphy-usb-super-speed-phy.patch
```diff
@@ -0,0 +1,125 @@
...
++    #include <dt-bindings/clock/qcom,ipq5332-gcc.h>
++
++    phy@4b0000 {
++        compatible = "qcom,ipq5332-uniphy-usb-ss-phy";
++        reg = <0x4b0000 0x800>;
++
++        clocks = <&gcc GCC_PCIE3X1_PHY_AHB_CLK>,
++                 <&gcc GCC_USB0_PHY_CFG_AHB_CLK>,
++                 <&gcc GCC_USB0_PIPE_CLK>;
++        resets =  <&gcc GCC_USB0_PHY_BCR>;
```
openwrt-ai, line 116: nit: double space after `resets =`, unlike `clocks =` two lines above and every other assignment in the example.

```suggestion
+        resets = <&gcc GCC_USB0_PHY_BCR>;
```


---

openwrt-ai (commented):
Reviewed the 2 new commits (`624b27f`/`91ece60` → `47fa2e6`/`a4acf3d`). Most of the delta is `Message-Id`/`Date` churn from re-generating the series; the substantive changes are all fixes.

Fixed since the last review: `minItems: 2` on `clocks` in 0821, the pipe/CFG-AHB clock ordering in 0823 now matches the binding, the unused `pipe_clk_fixed` member is gone from 0822, and the regulator imbalance in `uniphy_usb_init()` is properly unwound via the new `err_assert_reset:` / `err_disable_regulator:` chain. Dropping `dev_err_probe()` from `phy_pipe_clk_register()` is fine — the caller already wraps the return in one.

New regression in this revision: 0821 picked the raw `git format-patch` artifacts back up, which is the sole cause of the red `Check Kernel patches` job (inline).

## Commit checks

- `47fa2e6` "qualcommax: ipq50xx: add support for USB3" — unchanged message, so the scope mismatch flagged in the previous two rounds still stands: the body describes only the 22ULL PHY driver plus the AUX/LFPS clocks, while the diff also carries the `0091-v6.19-...-Correct-IPQ5018-interrupts` backport, the `qcom,dwc3` → `qcom,snps-dwc3` flattening in 0824 (`usb@8af8800` → `usb@8a00000`, `usb_dwc` child dropped — a user-visible DT ABI change), and the new `qcom,phy-usb-mux-sel` property + dwc3-qcom glue in 0827/0828.
- `a4acf3d` "qualcommax: ipq50xx: enable USB3 support on Linksys MR5500" — unchanged message; still doesn't mention the removal of `vbus-supply` from `&usb` (open thread on `ipq5018-mr5500.dts`).

Also still open and untouched by these commits: the legacy `#include <linux/of_device.h>` appended after `linux/reset.h` in 0822 (`of_device_get_match_data()` is declared in `linux/of.h`, already included).

Note on the bot's formality warning about `0842-arm64-dts-qcom-ipq5018-add-uniphy-node.patch` missing `From:`/`Date:`/`Subject:` — that file already lacked a mail header on `main` before this PR; this series only adjusts its hunk offsets. False positive as far as this PR is concerned.

---

kuncy7:
Confirmed working on the TP-Link Archer AX55 v1 (IPQ5018), with the PR head (82b71174c6) applied on top of my 6.18 branch:

- hot-plug after boot: two different USB 3.2 gen 1 sticks enumerate at SuperSpeed (`speed=5000`, `version=3.20`);
- cold boot with a stick attached: enumerates at SuperSpeed;
- warm reboot with a stick attached: enumerates at SuperSpeed;
- the uniphy USB SS PHY driver binds cleanly at `5d000.phy`, no dwc3/xhci/phy errors in dmesg.

Moving the mux selection into the PHY probe also settles the ordering concern from my original patches - the mux is now guaranteed to be written before the controller comes out of reset, by construction, and the comment in the driver documents why.

One stick-specific note that is *not* attributable to this series, recorded in case someone else trips over it: one of my sticks wedges on a warm reboot (attaches on the HS bus, then fails enumeration until it physically loses power). This board keeps the USB VBUS rail permanently on - nothing consumes `vbus-supply` on IPQ5018, so a warm reboot never power-cycles the device - and a full controller unbind/rebind reproduces the failure identically, which rules the host path out. Same behaviour would occur with the previous approach.

One more observation: fw_devlink now reports `Fixed dependency cycle` between `phy@5d000` and the GCC node, since the PHY both consumes GCC clocks and provides the pipe clock. It looks benign, but you may want to be aware of it.

Tested-by: Stanislaw Pal <kuncy7@gmail.com>

georgemoussalem:
> This board keeps the USB VBUS rail permanently on - nothing consumes `vbus-supply` on IPQ5018, so a warm reboot never power-cycles the device - and a full controller unbind/rebind reproduces the failure identically, which rules the host path out

Although the DWC3 controller driver doesn't consume the vbus-supply, the phy driver does or at least should as part of the shutdown sequence:
```
static int uniphy_usb_shutdown(struct phy *phy)
{
	struct uniphy_usb *uniphy = phy_get_drvdata(phy);

	clk_bulk_disable_unprepare(uniphy->num_clks, uniphy->clks);

	/* Assert phy reset */
	reset_control_assert(uniphy->reset);

	regulator_disable(uniphy->vreg);

	return 0;
}
```

so not sure why the vbus stays on. Will also test next week..

kuncy7:
Good point, and I dug into it - the short version is that the regulator your `shutdown()` disables is not the port VBUS on this board.

Your PHY driver does `devm_regulator_get(dev, "vdd")`, i.e. the PHY's own analog supply. On the AX55 `phy@5d000` has no `vdd-supply`, so it resolves to the dummy regulator:
```
qcom,uniphy-usb-ss-phy 5d000.phy: supply vdd not found, using dummy regulator
```
so `regulator_disable(vreg)` in shutdown is a no-op here.

The 5V that actually powers the stick is a separate board rail - a fixed regulator on a GPIO (`reg_usb_vbus`, GPIO 27), marked `regulator-always-on`. It is referenced as `vbus-supply` on the `&usb` node, but the dwc3-qcom glue doesn't consume `vbus-supply` as a regulator (the `vbus` references in the driver are the device-mode VBUS *override* bit, not a supply). So nothing in the driver model ever drops that rail, and `vdd` != VBUS - even wiring `vdd-supply` wouldn't power-cycle the port.

So it's a board wiring detail, orthogonal to your series - I only flagged it so the stick quirk isn't misread as a controller-path issue. If it's useful I can wire an experiment on my board (drop `always-on` and give the port VBUS a consumer that cycles on shutdown) and report whether that clears the wedge, but that's separate from this PR.

Thanks for looking - and no rush on the retest.

georgemoussalem:
> Your PHY driver does `devm_regulator_get(dev, "vdd")`, i.e. the PHY's own analog supply. On the AX55 `phy@5d000` has no `vdd-supply`, so it resolves to the dummy regulator:
> 
> ```
> qcom,uniphy-usb-ss-phy 5d000.phy: supply vdd not found, using dummy regulator
> ```
> 
> so `regulator_disable(vreg)` in shutdown is a no-op here.

so just add the vdd-supply property to the usbphy1 node

target/linux/qualcommax/patches-6.18/0822-phy-qcom-introduce-qualcomm-ipq5332-super-speed-usb-uniphy-driver.patch
```diff
@@ -0,0 +1,361 @@
...
++	if (IS_ERR(uniphy->phy)) {
++		ret = PTR_ERR(uniphy->phy);
++		return dev_err_probe(dev, ret, "failed to create PHY\n");
++	}
++
++	ret = phy_pipe_clk_register(uniphy, uniphy->phy->id);
++	if (ret)
++		return dev_err_probe(dev, ret, "failed to register pipe clk\n");
++
++	phy_set_drvdata(uniphy->phy, uniphy);
```
openwrt-ai, line 315: `platform_set_drvdata()` is never called, but the new `.remove` added in this revision reads it back:

```c
static void qcom_uniphy_usb_remove(struct platform_device *pdev)
{
	struct uniphy_usb *uniphy = platform_get_drvdata(pdev);

	qcom_uniphy_usb_mux_select(uniphy, false);
}
```

`platform_get_drvdata()` returns `NULL`, so `qcom_uniphy_usb_mux_select()` dereferences `uniphy->tcsr` off a NULL pointer — an oops on `rmmod phy-qcom-uniphy-usb-ss-22ull` or on unbind via sysfs. `phy_set_drvdata()` is a different store: it lives on `struct phy`, not on the platform device. The sister driver in the same directory does both — see [`platform_set_drvdata(pdev, phy)` at phy-qcom-uniphy-pcie-28lp.c:286](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/phy/qualcomm/phy-qcom-uniphy-pcie-28lp.c#L286).

```suggestion
+	platform_set_drvdata(pdev, uniphy);
+	phy_set_drvdata(uniphy->phy, uniphy);
```

This adds a line to the payload, so the `@@ -0,0 +1,304 @@` header and the `304`/`315` counts in the diffstat need regenerating afterwards (`make target/linux/refresh` with ipq50xx selected in `.config`).

---

target/linux/qualcommax/patches-6.18/0822-phy-qcom-introduce-qualcomm-ipq5332-super-speed-usb-uniphy-driver.patch
```diff
@@ -0,0 +1,361 @@
...
++				     "failed to get vreg\n");
++
++	uniphy->tcsr = syscon_regmap_lookup_by_phandle_args(dev->of_node,
++							    "qcom,phy-usb-mux-sel",
++							    1, &uniphy->mux_offset);
++	if (IS_ERR(uniphy->tcsr)) {
++		ret = PTR_ERR(uniphy->tcsr);
++		uniphy->tcsr = NULL;
++		return dev_err_probe(dev, ret, "failed to get regmap\n");
++	}
```
openwrt-ai: nit: `uniphy->tcsr = NULL;` is a dead store — the next statement returns, and `uniphy` is `devm_kzalloc()`ed so it is discarded along with the failed probe. It looks like a leftover from the dwc3-qcom version of this code (patch `0828` in the previous revision), where `tcsr == NULL` meant "property absent, mux selection not needed". Here `qcom,phy-usb-mux-sel` is a *required* property in `0821`, so there is no optional path left to signal, and the whole block collapses to the same shape as the four `dev_err_probe()` checks above it:

```suggestion
+	if (IS_ERR(uniphy->tcsr))
+		return dev_err_probe(dev, PTR_ERR(uniphy->tcsr),
+				     "failed to get regmap\n");
```

Drops two lines from the payload, so refresh the patch afterwards for the hunk header / diffstat.

---

target/linux/qualcommax/patches-6.18/0826-arm64-dts-qcom-ipq5018-add-clocks-required-for-usb3-support.patch
```diff
@@ -0,0 +1,44 @@
...
+-				      "mock_utmi";
++				      "mock_utmi",
++				      "aux",
++				      "lfps";
++			assigned-clocks = <&gcc GCC_USB0_MASTER_CLK>,
++					  <&gcc GCC_SYS_NOC_USB0_AXI_CLK>,
++					  <&gcc GCC_USB0_MOCK_UTMI_CLK>;
++			assigned-clock-rates = <200000000>,
++					       <200000000>,
++					       <60000000>;
```
openwrt-ai: nit: this `assigned-clocks` / `assigned-clock-rates` block is new in this revision and is not covered by the patch's subject or body, which only talk about aux and lfps ("Add aux and lfps clocks to USB controller node for Super-Speed support. These clocks are required for USB3 mode only and are not needed for USB2 mode."). Setting three clock rates is a separate, independently reviewable change — worth a sentence in the body at least.

Two of the three entries are also redundant: `GCC_USB0_MASTER_CLK` and `GCC_SYS_NOC_USB0_AXI_CLK` are both `CLK_SET_RATE_PARENT` branches of the *same* RCG `usb0_master_clk_src` ([gcc_sys_noc_usb0_axi_clk](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/clk/qcom/gcc-ipq5018.c#L2784-L2799), [gcc_usb0_master_clk](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/clk/qcom/gcc-ipq5018.c#L3046-L3061)), so both 200 MHz requests land on the same `set_rate`. And `usb0_master_clk_src`/`usb0_mock_utmi_clk_src` each have exactly one entry in their freq tables (200 MHz and 60 MHz respectively), which is what the RCGs come up at anyway — so if this is defensive rather than a fix for an observed wrong rate, dropping the block entirely is also an option.

---

openwrt-ai (commented):
Reviewed the 2 new commits (`47fa2e6`/`a4acf3d` → `75f4253`/`82b7117`). The big structural change in this revision is moving the TCSR mux selection out of the dwc3-qcom glue and into the PHY driver: `0827` (dwc3 binding) and `0828` (dwc3-qcom glue) are gone, `qcom,phy-usb-mux-sel` now lives on the PHY binding/node, and the old `0829` is renumbered to `0827`. The rest of the delta is `Message-Id`/`Date`/series-numbering churn from re-generating the series.

Fixed since the last review: the raw `git format-patch` artifacts in `0821` are gone again (`Check Kernel patches` is green on this head), the double space after `resets =` in the binding example, and the legacy `#include <linux/of_device.h>` in `0822`. Doing the mux write in the PHY's `probe()` is a genuine simplification over the glue-driver approach — the ordering is now correct by construction rather than by probe-order luck, and the comment explains why. The `0842` hunk offsets track the two extra lines `0823` now adds, and the `0824`/`0826`/`0827` offsets are all self-consistent.

New in this revision: the `.remove` callback added to `0822` dereferences a driver-data pointer that `probe()` never sets (inline). `assigned-clocks` appeared in `0826` without a mention in that patch's message (inline nit).

## Commit checks

- `75f4253` "qualcommax: ipq50xx: add support for USB3" — the message is still the same one flagged in the previous three rounds, and it still describes only the 22ULL PHY driver plus the AUX/LFPS clocks. The diff additionally carries the `0091-v6.19-...-Correct-IPQ5018-interrupts` backport, the `qcom,dwc3` → `qcom,snps-dwc3` flattening in `0824` (`usb@8af8800` → `usb@8a00000`, `usb_dwc` child dropped — a user-visible DT ABI change that out-of-tree board DTS files will notice), and the new `qcom,phy-usb-mux-sel` TCSR property in `0821`/`0823`. The mux work is arguably in scope for "add support for USB3" now that it lives in the PHY driver, but the node rename/flattening still isn't documented anywhere.
- `82b7117` "qualcommax: ipq50xx: enable USB3 support on Linksys MR5500" — unchanged message; still doesn't mention the removal of `vbus-supply` from `&usb` (open thread on `ipq5018-mr5500.dts`). Note that thread now has an answer in the PR conversation: the driver-visible supply on this board is `vdd-supply` on `&usbphy1`, which the PHY's `power_on`/`power_off` actually consumes. A line to that effect in the commit message would close it out.

---

kuncy7:
That did it - and I see it's exactly how you wire it on the MR5500 (`vdd-supply = <&regulator_fixed_5p0>` on `usbphy1`), so I mirrored that: pointed the AX55's 5V USB rail (`reg_usb_vbus`, a GPIO-switched fixed regulator) at `usbphy1`'s `vdd-supply` and dropped its `regulator-always-on`.

Results on hardware:
- `supply vdd not found` is gone; the regulator now has a real consumer (`use_count=1`).
- The warm-reboot stick wedge is gone: the stick that used to fail enumeration after a `reboot` now comes up at SuperSpeed every time, because `phy shutdown` actually drops VBUS now and the device gets a clean power cycle.
- Bonus: an on-demand `unbind`/`bind` of the controller power-cycles the port (regulator goes `enabled`->`disabled`->`enabled`, device disconnects and re-enumerates) - handy for resetting a wedged LTE modem.

So your suggestion was spot on. I'll carry this into the AX55 board DTS in my port PR. Thanks!

target/linux/qualcommax/patches-6.18/0822-phy-qcom-introduce-qualcomm-ipq5332-super-speed-usb-uniphy-driver.patch
```diff
@@ -0,0 +1,362 @@
...
++	 * while the controller resets, and the SS link doesn't train and is
++	 * stuck in Rx.Detect (USB2 is unaffected as it isn't muxed).
++	 */
++	ret = qcom_uniphy_usb_mux_select(uniphy, true);
++	if (ret)
++		return dev_err_probe(dev, ret, "failed to select USB mux\n");
++
++	platform_set_drvdata(pdev, uniphy);
++
++	return 0;
```
openwrt-ai, line 334: The mux write moved in this revision: in `82b7117` it ran before `devm_phy_create()`, now it runs after `devm_of_phy_provider_register()`. That is the exact call that publishes the PHY on the global provider list, i.e. the first moment a consumer's `of_phy_get()` can resolve it and go on to call `phy_init()`/`phy_power_on()` — and `uniphy_usb_init()` is what deasserts `GCC_USB0_PHY_BCR`. So the invariant the comment right below states ("Select the USB position of the phy mux **before anything comes out of reset**") is no longer guaranteed by construction; it now only holds because dwc3-qcom happens not to probe asynchronously.

This is the one ordering property @kuncy7's AX55 datapoint depends on (its bootloader leaves the mux on PCIe, so a late mux write leaves the SS link stuck in Rx.Detect). Moving the block back above the provider registration restores "correct by construction" and keeps the line count identical, so no patch refresh is needed:

```suggestion
+	/*
+	 * Select the USB position of the phy mux before anything comes out
+	 * of reset. Selecting it later leaves the SuperSpeed pads muxed away
+	 * while the controller resets, and the SS link doesn't train and is
+	 * stuck in Rx.Detect (USB2 is unaffected as it isn't muxed).
+	 */
+	ret = qcom_uniphy_usb_mux_select(uniphy, true);
+	if (ret)
+		return dev_err_probe(dev, ret, "failed to select USB mux\n");
+
+	platform_set_drvdata(pdev, uniphy);
+
+	phy_provider = devm_of_phy_provider_register(dev, of_phy_simple_xlate);
+	if (IS_ERR(phy_provider))
+		return dev_err_probe(dev, PTR_ERR(phy_provider),
+				     "failed to register PHY provider\n");
+
+	return 0;
```

Was there a reason to sequence it after the provider registration? If so, the comment should be reworded so it doesn't promise an ordering the code doesn't enforce.

---

target/linux/qualcommax/patches-6.18/0826-arm64-dts-qcom-ipq5018-add-clocks-required-for-usb3-support.patch
```diff
@@ -0,0 +1,43 @@
+From 84b2a1e47780b0d9d76c31b52ce00da7386b60ee Mon Sep 17 00:00:00 2001
+From: George Moussalem <george.moussalem@outlook.com>
+Date: Wed, 12 Aug 2026 20:37:34 +0400
+Subject: [PATCH v3 10/15] arm64: dts: qcom: ipq5018: Configure clocks required for USB3 support
```
openwrt-ai, line 4: nit: the subject changed from "Add clocks required for USB3 support" to "Configure clocks required for USB3 support" in this revision, but the file is still named `0826-arm64-dts-qcom-ipq5018-add-clocks-required-for-usb3-support.patch`. Every other patch this commit adds has a filename that is the slugified subject (e.g. `0827-arm64-dts-qcom-ipq5018-add-super-speed-uniphy-to-usb-node.patch` for "Add Super-Speed UNIPHY to USB node"), so this one now reads `add-` against a `Configure` subject. Renaming to `0826-arm64-dts-qcom-ipq5018-configure-clocks-required-for-usb3-support.patch` keeps the numeric prefix and therefore the apply order.

---

openwrt-ai (commented):
Reviewed the 2 new commits (`75f4253`/`82b7117` → `734d5c3`/`7d9e1f5`). Commit 2 is byte-identical to the previous revision (pure rebase); the whole delta is in commit 1, and apart from the `Message-Id`/`Date`/`v2 x/14` → `v3 x/15` churn from re-generating the series it is all fixes.

Fixed since the last review:

- `platform_set_drvdata(pdev, uniphy)` is now called, so the NULL deref in the new `.remove` is gone.
- `select MFD_SYSCON` added to the Kconfig entry, so `syscon_regmap_lookup_by_phandle_args()` no longer relies on another symbol pulling syscon in.
- The dead `uniphy->tcsr = NULL;` store is gone and the five `dev_err_probe()` checks in probe are now uniform.
- `devm_of_phy_provider_register()`'s return is checked properly instead of the `PTR_ERR_OR_ZERO()` dance.
- `0824` now uses `reg = <0x08a00000 0xfc100>`. That is the right call: the old `0xe000` covered only the dwc3 core and not the glue block at `+0xf8800`, which the flattened `qcom,snps-dwc3` model needs in the same `reg`, and `0xfc100` is exactly what every other flattened node upstream uses (`sc7280.dtsi` `usb@8c00000`, `sm8450.dtsi` `usb@a600000`, `lemans.dtsi` `usb@a600000`). The reworked `interrupts`/`interrupt-names` (`dwc_usb3`, `pwr_event`, `dp_hs_phy_irq`, `dm_hs_phy_irq`) also match the ipq5332 `allOf` branch that the `0091` backport moves `qcom,ipq5018-dwc3` into.
- `0826`'s `assigned-clocks` is down to `GCC_USB0_MOCK_UTMI_CLK` alone and is now mentioned in that patch's body.
- Patch metadata is self-consistent — hunk headers and diffstats add up, and the `0842`/`0827` offset shifts correctly track `0826` now adding 6 lines instead of 10. All CI, including `Check Kernel patches (qualcommax, ipq50xx)` and the three kernel builds, is green on this head.

New in this revision: the TCSR mux write moved from before `devm_phy_create()` to after `devm_of_phy_provider_register()`, which weakens the ordering guarantee the comment next to it states (inline). One filename/subject nit on `0826`.

## Commit checks

- `734d5c3` "qualcommax: ipq50xx: add support for USB3" — the message is unchanged for the fourth revision running and still describes only the 22ULL PHY driver plus the AUX/LFPS clocks. Two things in the diff are still undocumented anywhere in it: the `0091-v6.19-dt-bindings-usb-qcom-snps-dwc3-Correct-IPQ5018-interrupts` backport, and the `qcom,dwc3` → `qcom,snps-dwc3` flattening in `0824` (`usb@8af8800` → `usb@8a00000`, `usb_dwc` child dropped, `reg` grown from `0xe000` to `0xfc100`, interrupts consolidated onto the parent) — a user-visible DT ABI change that out-of-tree board DTS files carrying `&usb_dwc { ... }` will notice at build time. A sentence or two, or a split, would make this reviewable from the message alone.
- `7d9e1f5` "qualcommax: ipq50xx: enable USB3 support on Linksys MR5500" — content and message unchanged; still doesn't mention dropping `vbus-supply` from `&usb`. The PR conversation has since settled the substance (the driver-visible supply is `vdd-supply` on `&usbphy1`, which the PHY's `power_on`/`power_off` actually consumes, and pointing the board's 5V rail there is what fixed @kuncy7's warm-reboot wedge on the AX55). A one-liner to that effect in the commit message would close out the open thread on `ipq5018-mr5500.dts`.

---

lorand-horvath:
@robimarko Could you merge this, please?

robimarko:

LS3434:
Just for information, this PR breaks support for WWAN modems that use PCIe/MHI on IPQ5018. We are working on supporting new device to add to OpenWrt, Xunison D50. If &usbphy0 and &usbphy1 nodes are enabled in DTS, modem won't show up on PCI bus. If i disable &usbphy1, then modem shows up on PCI bus, but i lose USB with these errors:

```
[   33.115148] platform 8a00000.usb: deferred probe pending: dwc3-qcom: failed to register DWC3 Core

root@OpenWrt:~# lsusb
unable to initialize libusb: -99
```

georgemoussalem:
> Just for information, this PR breaks support for WWAN modems that use PCIe/MHI on IPQ5018. We are working on supporting new device to add to OpenWrt, Xunison D50. If &usbphy0 and &usbphy1 nodes are enabled in DTS, modem won't show up on PCI bus. If i disable &usbphy1, then modem shows up on PCI bus, but i lose USB with these errors:
> 
> ```
> [   33.115148] platform 8a00000.usb: deferred probe pending: dwc3-qcom: failed to register DWC3 Core
> 
> root@OpenWrt:~# lsusb
> unable to initialize libusb: -99
> ```

thanks for reporting your findings. The USB3 PHY node should NOT be enabled where PCIe1 is used.
Then, the USB3 PIPE clock should not be used either. Instead, revert to using the UTMI PIPE clock by adding the following property back to the USB node:
```qcom,select-utmi-as-pipe-clk```

Once done, can you re-test?

**Comment by LS3434** [comment]
So, how this should look ? Because in my DTS right now i have this:

```
&usbphy0 {
	status = "okay";
};

&usbphy1 {
	status = "disabled";
};

&usb {
	dr_mode = "host";
	status = "okay";
};
 
&pcie0_phy {
	status = "okay";
};
 
&pcie0 {
	status = "okay";
 
	perst-gpios = <&tlmm 15 GPIO_ACTIVE_LOW>;
 
	pcie@0 {
		wifi@0,0 {
			status = "okay";
 
			/* QCN9074: ath11k lacks DT compatible for PCI cards */
			compatible = "pci17cb,1104";
			reg = <0x00010000 0 0 0 0>;
 
			qcom,ath11k-calibration-variant = "Xunison-Exigo-Hub-D50-5G";
		};
	};
};
 
&pcie1_phy {
	status = "okay";
};
 
&pcie1 {
	status = "okay";
	perst-gpios = <&tlmm 18 GPIO_ACTIVE_LOW>;
	wake-gpios = <&tlmm 19 GPIO_ACTIVE_HIGH>;
 
	pcie@0 {
		modem@0 {
			/* compatible = "pci17cb,0306"; */
			reg = <0x00010000 0 0 0 0>;
			#address-cells = <3>;
			#size-cells = <2>;
		};
	};
};
 
&q6v5_wcss {
	firmware-name = "ath11k/IPQ5018/hw1.0/q6_fw.mdt",
			"ath11k/IPQ5018/hw1.0/m3_fw.mdt";
};
 
&wifi {
	status = "okay";
 
	qcom,rproc = <&q6v5_wcss>;
	qcom,ath11k-calibration-variant = "Xunison-Exigo-Hub-D50-5G";
	qcom,ath11k-fw-memory-mode = <1>;
};
```

**Comment by LS3434** [comment]
So i changed usb nodes in DTS to this:

```
&usbphy0 {
	status = "okay";
};

&usbphy1 {
	status = "disabled";
};

&usb {
	dr_mode = "host";
	status = "okay";
    qcom,select-utmi-as-pipe-clk;
};
 
&pcie0_phy {
	status = "okay";
};
```
Modem shows up on PCI, but i still have USB errors:

```
[   34.396653] platform 8a00000.usb: deferred probe pending: dwc3-qcom: failed to register DWC3 Core

root@OpenWrt:~# lsusb
unable to initialize libusb: -99
root@OpenWrt:~# lspci
0000:00:00.0 PCI bridge: Qualcomm Technologies, Inc Device 1004
0000:01:00.0 Network controller: Qualcomm Technologies, Inc QCN6024/9024/9074 Wireless Network Adapter (rev 01)
0001:00:00.0 PCI bridge: Qualcomm Technologies, Inc Device 1004
0001:01:00.0 Unassigned class [ff00]: Qualcomm Technologies, Inc SDX55 [Snapdragon X55 5G]
```

**Comment by LS3434** [comment]
@georgemoussalem  I made another changes to DTS, and in this configuration everything works correctly on this particular device:

```
&usbphy0 {
	status = "okay";
};

&usbphy1 {
	status = "disabled";
};

&usb {
	phys = <&usbphy0>;
	phy-names = "usb2-phy";
	qcom,select-utmi-as-pipe-clk;

	dr_mode = "host";
	status = "okay";
};
```

Here are the log lines:

```
[    4.272163] xhci-hcd xhci-hcd.1.auto: xHCI Host Controller
[    4.272230] xhci-hcd xhci-hcd.1.auto: new USB bus registered, assigned bus number 1
[    4.276718] xhci-hcd xhci-hcd.1.auto: hcc params 0x0220fe65 hci version 0x110 quirks 0x0000808000000010
[    4.284243] xhci-hcd xhci-hcd.1.auto: irq 40, io mem 0x08a00000
[    4.293720] xhci-hcd xhci-hcd.1.auto: xHCI Host Controller
[    4.299424] xhci-hcd xhci-hcd.1.auto: new USB bus registered, assigned bus number 2
[    4.304954] xhci-hcd xhci-hcd.1.auto: Host supports USB 3.0 SuperSpeed
[    4.313438] hub 1-0:1.0: USB hub found
[    4.319200] hub 1-0:1.0: 1 port detected
[    4.323995] usb usb2: We don't know the algorithms for LPM for this host, disabling LPM.
[    4.327697] hub 2-0:1.0: USB hub found
[    4.335101] hub 2-0:1.0: 1 port detected

[   15.159656] usbcore: registered new interface driver option
[   15.159855] usbserial: USB Serial support registered for GSM modem (1-port)
[   15.164483] option 1-1:1.0: GSM modem (1-port) converter detected
[   15.171746] usb 1-1: GSM modem (1-port) converter now attached to ttyUSB0
[   15.177902] option 1-1:1.1: GSM modem (1-port) converter detected
[   15.178206] ath11k_pci 0000:01:00.0: chip_id 0x0 chip_family 0x0 board_id 0xff soc_id 0xffffffff
[   15.184480] usb 1-1: GSM modem (1-port) converter now attached to ttyUSB1
[   15.190147] ath11k_pci 0000:01:00.0: fw_version 0x290b8862 fw_build_timestamp 2024-09-23 10:51 fw_build_id
[   15.199288] option 1-1:1.2: GSM modem (1-port) converter detected
[   15.215952] usb 1-1: GSM modem (1-port) converter now attached to ttyUSB2
[   15.222180] option 1-1:1.3: GSM modem (1-port) converter detected
[   15.229168] usb 1-1: GSM modem (1-port) converter now attached to ttyUSB3

[   23.093786] mhi_init Quectel_Linux_PCIE_MHI_Driver_V1.3.8
[   23.094506] mhi_pci_probe pci_dev->name = 0001:01:00.0, domain=1, bus=1, slot=0, vendor=17CB, device=0306
[   23.098886] mhi_q 0001:01:00.0: BAR 0 [mem 0x80300000-0x80300fff 64bit]: assigned
[   23.107834] mhi_q 0001:01:00.0: enabling device (0000 -> 0002)
[   23.217517] [I][mhi_netdev_enable_iface] Prepare the channels for transfer
[   23.278264] [I][mhi_netdev_enable_iface] Exited.
[   23.279902] rmnet_vnd_register_device(rmnet_mhi0.1)=0

root@OpenWrt:~# lsusb
Bus 001 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
Bus 001 Device 002: ID 2c7c:0800 Quectel Wireless Solutions Co., Ltd. RM500Q-AE
Bus 002 Device 001: ID 1d6b:0003 Linux Foundation 3.0 root hub
root@OpenWrt:~# lspci
0000:00:00.0 PCI bridge: Qualcomm Technologies, Inc Device 1004
0000:01:00.0 Network controller: Qualcomm Technologies, Inc QCN6024/9024/9074 Wireless Network Adapter (rev 01
0001:00:00.0 PCI bridge: Qualcomm Technologies, Inc Device 1004
0001:01:00.0 Unassigned class [ff00]: Qualcomm Technologies, Inc SDX55 [Snapdragon X55 5G]
```

So it looks like, usb needs to be configured differently for each device that uses WWAN modems, or usb ports. Either way, it works now.

**Comment by georgemoussalem** [comment]
I know what the root cause and have got it working, I now need to make it upstream acceptable.
The issue is that the wrong ref clock rate is used to calculate the value of the ref clock period and other GFLADJ settings.
The current code assumes a ref clock of 19.2MHz. However, for IPQ5018 (and likely IPQ5332), it's 60MHz, so the period becomes 16 (0x10). There is a DT property that you could use to manually set the ref clk period (snps,ref-clock-period-ns) but that's been deprecated. And, it also doesn't work :) because using the period alone, the ref clk rate is calculated as NSEC_PER_SEC / period which computes 62.5MHz (!= 60MHz).
Before they decided to flatten the snps and dwc3 nodes into one, there was a ref clock you could assign in the usb_dwc3 node ("ref"). Now, in the new design and glue driver, it tells the dwc3 core not to acquire / manage the clocks because the glue driver does it (ignore_clocks_and_resets is set to true) before it probes dwc3 core.
Simple solution would be to set it to false, but that only works for this SoC and would potentially break USB on other SoCs.
If you hardcode the rate as follows in `dwc3_ref_clk_period` and don't return, you'll see that USB3 and USB2 modes work without making any DT changes:

```
} else {
		dev_info(dwc->dev, "DWC3 has no ref clock/period, defaulting to 60MHz\n");
		rate = 60000000;
		period = NSEC_PER_SEC / rate;
		// return;
	}
```

Will patch it but need to think about how to best do it so it gets accepted upstream.

**Comment by LS3434** [comment]
Right now, i will try something like this:

```
&usbphy0 {
	status = "okay";
};

&usbphy1 {
	status = "okay";
};

&usb {
	phys = <&usbphy0>, <&usbphy1>;
	phy-names = "usb2-phy", "usb3-phy";

	qcom,select-utmi-as-pipe-clk;

	dr_mode = "host";
	status = "okay";
};
```

Let you know, if it works.

EDIT: Nope, in this version, USB works, but modem dissapears from PCI. Previous version is correct for this device.

**Comment by georgemoussalem** [comment]
yeah, but it's not the actual fix. Working on a solution that works for all scenarios

**Comment by georgemoussalem** [comment]
can you add these two patches and add the following property to the usb node in your board dts?

`qcom,select-utmi-as-ref-clk;`

https://filebin.net/ig067rfg5uf40txu

**Comment by LS3434** [comment]
@georgemoussalem  Ok, so on the first version in DTS:

```
&usbphy0 {
	status = "okay";
};

&usbphy1 {
	status = "disabled";
};

&usb {
    phys = <&usbphy0>, <&usbphy1>;
	phy-names = "usb2-phy", "usb3-phy";
    
	qcom,select-utmi-as-ref-clk;

	dr_mode = "host";
	status = "okay";
};
```
Modem shows up on PCI, but i have no USB:

```
[   33.133275] platform 8a00000.usb: deferred probe pending: dwc3-qcom: failed to register DWC3 Core

root@OpenWrt:~# lsusb
unable to initialize libusb: -99
root@OpenWrt:~# lspci
0000:00:00.0 PCI bridge: Qualcomm Technologies, Inc Device 1004
0000:01:00.0 Network controller: Qualcomm Technologies, Inc QCN6024/9024/9074 Wireless Network Adapter (rev 01)
0001:00:00.0 PCI bridge: Qualcomm Technologies, Inc Device 1004
0001:01:00.0 Unassigned class [ff00]: Qualcomm Technologies, Inc SDX55 [Snapdragon X55 5G]
```

On 2nd version in DTS:

```
&usbphy0 {
	status = "okay";
};

&usbphy1 {
	status = "okay";
};

&usb {
    phys = <&usbphy0>, <&usbphy1>;
	phy-names = "usb2-phy", "usb3-phy";
    
	qcom,select-utmi-as-ref-clk;

	dr_mode = "host";
	status = "okay";
};
```
I have USB, but modem dissapears from PCI:

```
[    4.538754] xhci-hcd xhci-hcd.1.auto: xHCI Host Controller
[    4.538828] xhci-hcd xhci-hcd.1.auto: new USB bus registered, assigned bus number 1
[    4.543424] xhci-hcd xhci-hcd.1.auto: hcc params 0x0220fe65 hci version 0x110 quirks 0x0000808000000010
[    4.550755] xhci-hcd xhci-hcd.1.auto: irq 40, io mem 0x08a00000
[    4.560381] xhci-hcd xhci-hcd.1.auto: xHCI Host Controller
[    4.566025] xhci-hcd xhci-hcd.1.auto: new USB bus registered, assigned bus number 2
[    4.571549] xhci-hcd xhci-hcd.1.auto: Host supports USB 3.0 SuperSpeed
[    4.580106] hub 1-0:1.0: USB hub found
[    4.585804] hub 1-0:1.0: 1 port detected
[    4.590588] usb usb2: We don't know the algorithms for LPM for this host, disabling LPM.
[    4.594533] hub 2-0:1.0: USB hub found
[    4.601720] hub 2-0:1.0: 1 port detected


[   38.091683] usb 1-1: new high-speed USB device number 2 using xhci-hcd
[   38.345945] option 1-1:1.0: GSM modem (1-port) converter detected
[   38.346616] usb 1-1: GSM modem (1-port) converter now attached to ttyUSB0
[   38.352099] option 1-1:1.1: GSM modem (1-port) converter detected
[   38.358429] usb 1-1: GSM modem (1-port) converter now attached to ttyUSB1
[   38.364859] option 1-1:1.2: GSM modem (1-port) converter detected
[   38.371273] usb 1-1: GSM modem (1-port) converter now attached to ttyUSB2
[   38.377673] option 1-1:1.3: GSM modem (1-port) converter detected
[   38.384212] usb 1-1: GSM modem (1-port) converter now attached to ttyUSB3

root@OpenWrt:~# lsusb
Bus 001 Device 001: ID 1d6b:0002 Linux Foundation 2.0 root hub
Bus 001 Device 002: ID 2c7c:0800 Quectel Wireless Solutions Co., Ltd. RM500Q-AE
Bus 002 Device 001: ID 1d6b:0003 Linux Foundation 3.0 root hub
root@OpenWrt:~# lspci
0000:00:00.0 PCI bridge: Qualcomm Technologies, Inc Device 1004
0000:01:00.0 Network controller: Qualcomm Technologies, Inc QCN6024/9024/9074 Wireless Network Adapter (rev 01)
0001:00:00.0 PCI bridge: Qualcomm Technologies, Inc Device 1004

```

georgemoussalem:
Let’s continue the discussion on the forum. One thing to answer: where are your vdd-supply properties on both usbphy nodes?

LS3434:
Oh, i didn't add them. I will try adding them and let you know.

LS3434:
@georgemoussalem I added the vdd-supply properties along with regulator and proper gpio, but it didn't make any difference.

Issue #24553, opened 2026-08-03
ppp: pppd runs as root outside ujail while dnsmasq/hostapd/ntpd are confined


JuliusBairaktaris:
On a current OpenWrt build, the network-facing daemons split into two groups. These run
inside `ujail` as unprivileged users:

```
/sbin/ujail -t 5 -n hostapd        -U network -G network -C /etc/capabilities/...
/sbin/ujail -t 5 -n wpa_supplicant -U network -G network -C /etc/capabilities/...
/sbin/ujail -t 5 -n ntpd           -U ntp     -G ntp     -C /etc/capabilities/ntpd.json
/sbin/ujail -t 5 -n dnsmasq        -u -l -r /bin/ubus -r /etc/TZ ...
```

and these run as root with no jail:

```
/usr/sbin/pppd nodetach ipparam wan ifname pppoe-wan lcp-echo-interval 1 ...
/usr/sbin/odhcpd
/usr/sbin/uhttpd -f -h /www -r OpenWrt -x /cgi-bin -u /ubus ...
/sbin/rpcd
/sbin/netifd
```

`pppd` stands out in that second group. It is the only one of them that parses input from
the **WAN** side: LCP, IPCP/IPv6CP, PAP, CHAP and EAP control frames arrive from the peer
before the link is authenticated, and it processes them as root with no confinement.
Historically this is the component of CVE-2020-8597 (an EAP buffer overflow in
`eap.c`), so it is not a hypothetical attack surface.

`dnsmasq` is a useful comparison: it also parses untrusted network input, and it *is*
confined, with a dedicated user and a read-only mount set.

This is a hardening request, not a vulnerability report — I am not aware of a live bug in
`pppd`. I fuzzed its protocol handlers (LCP/IPCP/IPv6CP/CHAP/PAP/EAP/CCP/ECP, driven
through their `protent` structs, built with OpenWrt's configure options) under ASan for
tens of millions of executions without a finding, which is reassuring but obviously not a
proof.

Questions for maintainers:

1. Is leaving `pppd` unconfined deliberate — for example because it needs to create the
   `pppoe-wan` interface, set routes and run `/lib/netifd/ppp-up`, and a jail would
   complicate that?
2. If it is mostly historical, would a `ujail` profile along the lines of the `ntpd` one be
   welcome? The capability set looks like it would need `CAP_NET_ADMIN` and `CAP_NET_RAW`
   plus access to the pppox socket and the `ip-up`/`ip-down` scripts.
[...]

Issue #412, opened 2017-03-05
toolchain-mips_34kc_gcc-4.8-linaro_uClibc-0.9.33.2


Xchainer:
I get these errors when "make V=99"
```

libbackend.a(gimple-ssa-strength-reduction.o): In function `has_single_use':
/home/Xchainer/openwrt/build_dir/toolchain-mips_34kc_gcc-4.8-linaro_uClibc-0.9.33.2/gcc-linaro-4.8-2014.04/gcc/tree-flow-inline.h:307: undefined reference to `single_imm_use_1(ssa_use_operand_d const*, ssa_use_operand_d**, gimple_statement_d**)'
/home/Xchainer/openwrt/build_dir/toolchain-mips_34kc_gcc-4.8-linaro_uClibc-0.9.33.2/gcc-linaro-4.8-2014.04/gcc/tree-flow-inline.h:307: undefined reference to `single_imm_use_1(ssa_use_operand_d const*, ssa_use_operand_d**, gimple_statement_d**)'
/home/Xchainer/openwrt/build_dir/toolchain-mips_34kc_gcc-4.8-linaro_uClibc-0.9.33.2/gcc-linaro-4.8-2014.04/gcc/tree-flow-inline.h:307: undefined reference to `single_imm_use_1(ssa_use_operand_d const*, ssa_use_operand_d**, gimple_statement_d**)'
/home/Xchainer/openwrt/build_dir/toolchain-mips_34kc_gcc-4.8-linaro_uClibc-0.9.33.2/gcc-linaro-4.8-2014.04/gcc/tree-flow-inline.h:307: undefined reference to `single_imm_use_1(ssa_use_operand_d const*, ssa_use_operand_d**, gimple_statement_d**)'
/home/Xchainer/openwrt/build_dir/toolchain-mips_34kc_gcc-4.8-linaro_uClibc-0.9.33.2/gcc-linaro-4.8-2014.04/gcc/tree-flow-inline.h:307: undefined reference to `single_imm_use_1(ssa_use_operand_d const*, ssa_use_operand_d**, gimple_statement_d**)'
[...]

PR #24558, opened 2026-08-03
network: run pppd and odhcpd unprivileged in ujail

JuliusBairaktaris:
Implements #24553. pppd and odhcpd run as root today. This runs each in ujail as its own uid with only the capabilities its sockets need: pppd as uid 454 with `CAP_NET_ADMIN` and `CAP_NET_RAW`, odhcpd as uid 455 with those plus `CAP_NET_BIND_SERVICE` for 67/547. Both get `no_new_privs`, neither gets a mount namespace. uhttpd is #24740.

Each jail is opt-in. It applies only when `/sbin/ujail`, the capability file and the user are all present. Without any of them the daemon starts as it does now.

    package/network/services/ppp/files/ppp.sh, ppp_generic_setup():

        [ -x /sbin/ujail ] && [ -f "$caps" ] && id -u ppp >/dev/null 2>&1 && {
                jail="/sbin/ujail -t 5 -n pppd -U ppp -G ppp -C $caps -c --"

    package/network/services/odhcpd/files/odhcpd.init, odhcpd_jailed():

        [ -x /sbin/ujail ] && [ -e "$ODHCPD_CAPS" ] && id -u odhcpd >/dev/null 2>&1

The user check matters after a sysupgrade that preserved an `/etc/passwd` from before the uid existed. procd drops `user` and `group` without an error when the name does not resolve, so odhcpd would have kept running as uid 0 inside the jail.

ubusd denies every non-root uid by default, so each uid gets an `acl.d` file. pppd gets `network.interface notify_proto` and `network add_dynamic` for the ip-up scripts. odhcpd gets the `dhcp` object, `network.interface` subscribe and dump, `ubus.object.add` and `service signal`.

The dnsmasq commit comes first. `/etc/config/dhcp` ships 0600 root:root and odhcpd reads it directly, so a jailed odhcpd on a fresh image would serve nothing. It ships 0644 now, and a uci-defaults script relaxes existing installs. uci commit keeps the file's mode: on my router the file stayed 0600 across two commits.

The wrapper also hands pppd what uid 454 cannot take itself: `/dev/ppp`, `/var/run/pppd`, a serial TTY passed as an argument, and the peer DNS file behind `/etc/ppp/resolv.conf`. That last one is `/tmp/resolv.conf.ppp`, which pppd rewrites at every connect with `usepeerdns`. A copy left by a root-run pppd across an in-place package upgrade refuses uid 454 with `Permission denied`, so the wrapper removes it first. Nothing in OpenWrt reads that file; netifd takes the servers from the `DNS1`/`DNS2` environment.

The ppp commit goes the other way for the files that hold credentials. chap-secrets and the ppp-mod-radius server list go from 0600 root:root to 0640 root:ppp, from uci-defaults, because a postinst runs on the build host and `chgrp ppp` resolves against the host's `/etc/group` there.

Dependencies:

- openwrt/odhcpd#412 removes the `getuid() != 0` gate in `main()`. The odhcpd commit waits for it and for the `PKG_SOURCE_VERSION` bump.
- openwrt/rpcd#38. `rpc_file_access()` treats a call with no session id as authorised. That is only safe while uid 0 is the only uid that can reach ubus, and this series adds two more.
- openwrt/procd#40 is merged as procd 2c1834a but OpenWrt still pins 60fdbf0. Until the bump, a jailed pppd killed by `SIGSEGV` reaches netifd as exit 11, which pppd's table reads as `PEER_AUTH_FAILED`.
- ppp-project/ppp#608 for pppd's `euid != 0` gate. Carried here as `502-allow-nonroot-with-capnetadmin.patch`.

Tested on an ipq807x router with a PPPoE uplink, with the odhcpd#412 commit applied. At this head: pppd runs as uid 454 with `CapEff = CapAmb = 0x3000` and `NoNewPrivs: 1`, `wan` comes up with its v4 address and DNS and `wan_6` with its delegated prefix, and `ifdown wan` leaves no pppd behind. odhcpd runs as uid 455 with `0x3400`, writes its lease and hosts files as that uid and answers `ubus call dhcp ipv6leases`. The odhcpd user check was measured both ways: without the user the instance has no `user`, `group` or `jail` keys and odhcpd runs at uid 0; with the user present the instance carries all three and the capability file. `/etc/init.d/network restart` coming back with one pppd instance was measured on an earlier revision of this branch, not re-run at this head.

Not tested: pppoa, pptp, serial and 3G, since I have no hardware for them, and odhcpd's NDP relay mode under uid 455.

### DIFF
--- a/package/network/services/dnsmasq/files/50-dnsmasq-relax-dhcp-config-mode.sh
+++ b/package/network/services/dnsmasq/files/50-dnsmasq-relax-dhcp-config-mode.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+# odhcpd uci_load()s /etc/config/dhcp directly and runs unprivileged;
+# INSTALL_CONF shipped it 0600, which locked it out.
+chmod 0644 /etc/config/dhcp
+
+exit 0
--- a/package/network/services/odhcpd/files/odhcpd.init
+++ b/package/network/services/odhcpd/files/odhcpd.init
@@ -4,14 +4,78 @@ START=35
 STOP=85
 USE_PROCD=1
 
+ODHCPD_CAPS=/etc/capabilities/odhcpd.json
+
+odhcpd_jailed() {
+	[ -x /sbin/ujail ] && [ -e "$ODHCPD_CAPS" ]
+}
+
+odhcpd_ndp_relay() {
+	local ndp
+
+	config_get ndp "$1" ndp
+	[ "$ndp" = "relay" ] && relay=1
+}
+
+# Hand the unprivileged daemon what it cannot take itself: the state
+# directories it writes through a temporary file, and the global proxy_ndp
+# knob the kernel ORs with the per-interface one it can no longer write.
+odhcpd_prepare() {
+	local leasefile hostsdir piodir relay dir
+
+	config_load dhcp
+	config_get leasefile odhcpd leasefile
+	config_get hostsdir odhcpd hostsdir
+	config_get piodir odhcpd piodir
+	config_foreach odhcpd_ndp_relay dhcp
+
+	for dir in "${leasefile%/*}" "$hostsdir" "$piodir"; do
+		[ -n "$dir" ] || continue
+		mkdir -p "$dir" || continue
+		case "$(ls -ld "$dir" 2>/dev/null)" in
+		# world-writable: shared with everything else, not ours to take
+		????????w*) ;;
+		d*) chown odhcpd:odhcpd "$dir" ;;
+		esac
+	done
+
+	# an upgrade restarts the daemon inside the boot, leaving the leasefile
+	# owned by the root-run instance and the sticky bit stopping uid 455
+	# from renaming its replacement over it. Only where the directory is
+	# world-writable, which already grants everyone that much
+	case "$(ls -ld "${leasefile%/*}" 2>/dev/null)" in
+	????????w*) [ -f "$leasefile" ] && chown -h odhcpd:odhcpd "$leasefile" ;;
+	esac
+
+	# the kernel ORs this global with the per-interface knob uid 455 can no
+	# longer write; it is shared, so raise it and leave it to its owner
+	[ -n "$relay" ] && [ -e /proc/sys/net/ipv6/conf/all/proxy_ndp ] &&
+		echo 1 > /proc/sys/net/ipv6/conf/all/proxy_ndp
+
+	return 0
+}
+
 start_service() {
+	odhcpd_jailed && odhcpd_prepare
+
 	procd_open_instance
 	procd_set_param command /usr/sbin/odhcpd
 	procd_set_param respawn
+
+	odhcpd_jailed && {
+		procd_add_jail odhcpd
+		procd_set_param capabilities "$ODHCPD_CAPS"
+		procd_set_param no_new_privs 1
+		procd_set_param user odhcpd
+		procd_set_param group odhcpd
+	}
+
 	procd_close_instance
 }
 
 reload_service() {
+	odhcpd_jailed && odhcpd_prepare
+
 	procd_send_signal odhcpd
 }
 
--- a/package/network/services/ppp/Makefile
+++ b/package/network/services/ppp/Makefile
@@ -10,7 +10,7 @@ include $(INCLUDE_DIR)/kernel.mk
 
 PKG_NAME:=ppp
 PKG_VERSION:=2.5.3
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://download.samba.org/pub/ppp
@@ -55,6 +55,7 @@ define Package/ppp
 $(call Package/ppp/Default)
   DEPENDS:= +USE_GLIBC:libcrypt-compat +kmod-ppp
   TITLE:=PPP daemon
+  USERID:=ppp=454:ppp=454
   VARIANT:=default
 endef
 
@@ -62,6 +63,7 @@ define Package/ppp-multilink
 $(call Package/ppp/Default)
   DEPENDS:= +USE_GLIBC:libcrypt-compat +kmod-ppp
   TITLE:=PPP daemon (with multilink support)
+  USERID:=ppp=454:ppp=454
   VARIANT:=multilink
 endef
 
@@ -211,11 +213,17 @@ define Package/ppp/install
 	$(INSTALL_DATA) ./files/etc/ppp/filter $(1)/etc/ppp/
 	$(INSTALL_DATA) ./files/etc/ppp/options $(1)/etc/ppp/
 	$(LN) /tmp/resolv.conf.ppp $(1)/etc/ppp/resolv.conf
+	$(INSTALL_DIR) $(1)/etc/capabilities
+	$(INSTALL_DATA) ./files/pppd.capabilities $(1)/etc/capabilities/pppd.json
+	$(INSTALL_DIR) $(1)/usr/share/acl.d
+	$(INSTALL_DATA) ./files/pppd_acl.json $(1)/usr/share/acl.d/pppd.json
 	$(INSTALL_DIR) $(1)/lib/netifd/proto
 	$(INSTALL_BIN) ./files/ppp.sh $(1)/lib/netifd/proto/
 	$(INSTALL_BIN) ./files/lib/netifd/ppp-up $(1)/lib/netifd/
 	$(INSTALL_BIN) ./files/lib/netifd/ppp6-up $(1)/lib/netifd/
 	$(INSTALL_BIN) ./files/lib/netifd/ppp-down $(1)/lib/netifd/
+	$(INSTALL_DIR) $(1)/etc/uci-defaults
+	$(INSTALL_BIN) ./files/50-ppp-secrets-mode.sh $(1)/etc/uci-defaults
 endef
 Package/ppp-multilink/install=$(Package/ppp/install)
 
@@ -242,6 +250,8 @@ define Package/ppp-mod-radius/install
 		$(1)/etc/ppp/radius/
 	$(INSTALL_CONF) ./files/etc/ppp/radius/servers \
 		$(1)/etc/ppp/radius/
+	$(INSTALL_DIR) $(1)/etc/uci-defaults
+	$(INSTALL_BIN) ./files/50-ppp-radius-servers-mode.sh $(1)/etc/uci-defaults
 endef
 
 define Package/ppp-mod-pppol2tp/install
--- a/package/network/services/ppp/files/50-ppp-secrets-mode.sh
+++ b/package/network/services/ppp/files/50-ppp-secrets-mode.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+# pppd reads these as uid ppp, and INSTALL_CONF ships them 0600 root:root -
+# as does a file preserved across sysupgrade from an install predating the uid.
+for secret in /etc/ppp/chap-secrets /etc/ppp/pap-secrets; do
+	[ -f "$secret" ] || continue
+	chgrp ppp "$secret"
+	chmod 0640 "$secret"
+done
+
+exit 0
--- a/package/network/services/ppp/files/ppp.sh
+++ b/package/network/services/ppp/files/ppp.sh
@@ -146,7 +146,21 @@ ppp_generic_setup() {
 	[ "$delegate" != "0" ] && delegate=""
 	[ "$norelease" = "1" ] || norelease=""
 
-	proto_run_command "$config" /usr/sbin/pppd \
+	local jail caps=/etc/capabilities/pppd.json
+	[ -x /sbin/ujail ] && [ -f "$caps" ] && {
+		jail="/sbin/ujail -t 5 -n pppd -U ppp -G ppp -C $caps -c --"
+		chown ppp /dev/ppp
+		mkdir -p /var/run/pppd
+		chown ppp /var/run/pppd
+		# A serial link is handed its TTY as a pppd argument - first by
+		# proto_ppp_setup, last by comgt's proto_3g_setup - and a TTY is
+		# root:dialout 0660, which no capability overrides.
+		for arg in "$@"; do
+			[ -c "$arg" ] && chown ppp "$arg"
+		done
+	}
+
+	proto_run_command "$config" $jail /usr/sbin/pppd \
 		nodetach ipparam "$config" \
 		ifname "$pppname" \
 		${localip:+$localip:} \
--- a/package/network/services/ppp/files/pppd.capabilities
+++ b/package/network/services/ppp/files/pppd.capabilities
@@ -1,0 +1,22 @@
+{
+	"bounding": [
+		"CAP_NET_ADMIN",
+		"CAP_NET_RAW"
+	],
+	"effective": [
+		"CAP_NET_ADMIN",
+		"CAP_NET_RAW"
+	],
+	"ambient": [
+		"CAP_NET_ADMIN",
+		"CAP_NET_RAW"
+	],
+	"permitted": [
+		"CAP_NET_ADMIN",
+		"CAP_NET_RAW"
+	],
+	"inheritable": [
+		"CAP_NET_ADMIN",
+		"CAP_NET_RAW"
+	]
+}
--- a/package/network/services/ppp/patches/502-allow-nonroot-with-capnetadmin.patch
+++ b/package/network/services/ppp/patches/502-allow-nonroot-with-capnetadmin.patch
@@ -1,0 +1,195 @@
+From 0f0fd254c566bccee52eb814ebf671272f186838 Mon Sep 17 00:00:00 2001
+From: Julius Bairaktaris <julius@bairaktaris.de>
+Date: Thu, 6 Aug 2026 15:34:20 +0200
+Subject: [PATCH] pppd: allow running without root when CAP_NET_ADMIN is
+ present
+
+The euid != 0 gate dates from the setuid-root era: it refuses to run
+when the caller is not root and the binary is not setuid-root. On
+Linux that requirement is a capability, not a uid: the kernel's
+ppp_open() checks ns_capable(CAP_NET_ADMIN), and the interface ioctls
+require the same. A sandbox that grants the capability (procd/ujail,
+systemd) can therefore run pppd unprivileged, but the gate stops it
+before it can even try.
+
+Replace the check with ppp_privileged(): euid 0, or CAP_NET_ADMIN in
+the effective set (checked via capget on Linux, so no setuid bit and
+no libcap dependency). Solaris keeps the euid check.
+
+Passing the gate is not enough on its own. parse_args() grants OPT_PRIV
+options only when the privileged flag is set, and that flag was tied to
+getuid() == 0, so an unprivileged pppd holding the capability still
+died on its first privileged option: "ifname" for any caller that names
+the interface, "plugin" for pppoe. Grant the flag on the capability
+too, but only where the capability is the caller's own authority and
+not something the exec itself handed over. ppp_secure_exec() reads the
+kernel's own verdict on that from AT_SECURE, which covers a setuid or
+setgid bit and file capabilities alike; without it an unprivileged
+invoker of such a pppd would reach the OPT_PRIV "plugin" option and
+dlopen an attacker-chosen object with those privileges.
+
+run_program() re-roots the ip-up/ip-down child with setuid(0) before
+exec. Without CAP_SETUID that fails with EPERM and is fatal, so the
+link-up scripts never run under a capability sandbox. Only re-root when
+already euid 0 (the setuid-installed case); otherwise the child keeps
+its uid and any ambient capabilities reach the scripts unchanged.
+
+Validated on an ipq807x router: pppd runs as an unprivileged user
+(uid 454) with CAP_NET_ADMIN and proceeds past the gate; without the
+capability it exits EXIT_NOT_ROOT with the new message.
+
+Assisted-by: Claude:claude-opus-5
+Signed-off-by: Julius Bairaktaris <julius@bairaktaris.de>
+---
+ pppd/main.c        | 21 ++++++++++++++-------
+ pppd/pppd.h        | 13 +++++++++++++
+ pppd/sys-linux.c   | 34 +++++++++++++++++++++++++++++++++++
+ pppd/sys-solaris.c | 20 ++++++++++++++++++++
+ 4 files changed, 81 insertions(+), 7 deletions(-)
+
+--- a/pppd/main.c
++++ b/pppd/main.c
+@@ -401,7 +401,11 @@ main(int argc, char *argv[])
+     umask(umask(0777) | 022);
+ 
+     uid = getuid();
+-    privileged = uid == 0;
++    /* Privileged options come from the invoking user's own authority, so a
++     * pppd that gained privileges from the exec itself must not hand them
++     * to whoever ran it. */
++    privileged = uid == 0 || (uid == geteuid() && !ppp_secure_exec() &&
++                              ppp_privileged());
+     slprintf(numbuf, sizeof(numbuf), "%d", uid);
+     ppp_script_setenv("ORIG_UID", numbuf, 0);
+ 
+@@ -452,10 +456,11 @@ main(int argc, char *argv[])
+     }
+ 
+     /*
+-     * Check that we are running as root.
++     * Check that we have the privileges needed to open the ppp device
++     * and configure interfaces.
+      */
+-    if (geteuid() != 0) {
+-	ppp_option_error("must be root to run %s, since it is not setuid-root",
++    if (!ppp_privileged()) {
++	ppp_option_error("%s must be run as root or with CAP_NET_ADMIN",
+ 		     argv[0]);
+ 	exit(EXIT_NOT_ROOT);
+     }
+@@ -2004,9 +2009,11 @@ run_program(char *prog, char * const *ar
+     if (ret != 0) {
+         fatal("Failed to change directory to '/', %m");
+     }
+-    ret = setuid(0);		/* set real UID = root */
+-    if (ret != 0) {
+-        fatal("Failed to set uid, %m");
++    if (geteuid() == 0) {
++	ret = setuid(0);	/* set real UID = root */
++	if (ret != 0) {
++	    fatal("Failed to set uid, %m");
++	}
+     }
+     ret = setgid(getegid());
+     if (ret != 0) {
+--- a/pppd/pppd.h
++++ b/pppd/pppd.h
+@@ -444,6 +444,19 @@ void ppp_script_unsetenv(char *);
+ int ppp_check_kernel_support(void);
+ 
+ /*
++ * Test whether we have the privileges needed to open the ppp device
++ * and configure interfaces.
++ */
++int ppp_privileged(void);
++
++/*
++ * Test whether this exec raised our privileges - a setuid or setgid bit
++ * on the binary, or file capabilities - which the kernel grants to
++ * whoever runs it, and not to us on their behalf.
++ */
++int ppp_secure_exec(void);
++
++/*
+  * Restore device setting
+  */
+ void ppp_generic_disestablish(int dev_fd);
+--- a/pppd/sys-linux.c
++++ b/pppd/sys-linux.c
+@@ -107,6 +107,9 @@
+ 
+ #if !defined(__GLIBC__) || __GLIBC__ >= 2
+ #include <asm/types.h>		/* glibc 2 conflicts with linux/types.h */
++#include <sys/syscall.h>
++#include <sys/auxv.h>
++#include <linux/capability.h>
+ #include <net/if.h>
+ #include <net/if_arp.h>
+ #include <net/route.h>
+@@ -2863,6 +2866,37 @@ ppp_registered(void)
+     return ret;
+ }
+ 
++/*
++ * Check whether we have the privileges needed to open the ppp device
++ * and configure interfaces.  The kernel requires CAP_NET_ADMIN to open
++ * /dev/ppp (ppp_open in drivers/net/ppp/ppp_generic.c).
++ */
++int ppp_privileged(void)
++{
++    struct __user_cap_header_struct hdr;
++    struct __user_cap_data_struct data[2];
++
++    if (geteuid() == 0)
++	return 1;
++
++    hdr.version = _LINUX_CAPABILITY_VERSION_3;
++    hdr.pid = 0;
++    if (syscall(SYS_capget, &hdr, data) < 0)
++	return 0;
++
++    return !!(data[CAP_TO_INDEX(CAP_NET_ADMIN)].effective & CAP_TO_MASK(CAP_NET_ADMIN));
++}
++
++/*
++ * Test whether this exec raised our privileges.  The kernel sets AT_SECURE
++ * when a setuid or setgid bit or file capabilities did so, granting them
++ * to whoever runs the binary rather than to us on their behalf.
++ */
++int ppp_secure_exec(void)
++{
++    return getauxval(AT_SECURE) != 0;
++}
++
+ /********************************************************************
+  *
+  * ppp_check_kernel_support - check whether the system has any ppp interfaces
+--- a/pppd/sys-solaris.c
++++ b/pppd/sys-solaris.c
+@@ -636,6 +636,26 @@ sys_check_options(void)
+ }
+ 
+ /*
++ * ppp_privileged - check whether we have the privileges needed to open
++ * the ppp device and configure interfaces.
++ */
++int
++ppp_privileged(void)
++{
++    return geteuid() == 0;
++}
++
++/*
++ * ppp_secure_exec - Solaris has no AT_SECURE; the euid check in main.c
++ * covers setuid.
++ */
++int
++ppp_secure_exec(void)
++{
++    return 0;
++}
++
++/*
+  * ppp_check_kernel_support - check whether the system has any ppp interfaces
+  */
+ int
--- a/package/network/services/dnsmasq/Makefile
+++ b/package/network/services/dnsmasq/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=dnsmasq
 PKG_UPSTREAM_VERSION:=2.93
 PKG_VERSION:=$(subst test,~~test,$(subst rc,~rc,$(PKG_UPSTREAM_VERSION)))
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_UPSTREAM_VERSION).tar.xz
 PKG_SOURCE_URL:=https://thekelleys.org.uk/dnsmasq/
@@ -169,7 +169,7 @@ define Package/dnsmasq/install
 	$(INSTALL_DIR) $(1)/usr/sbin
 	$(CP) $(PKG_INSTALL_DIR)/usr/sbin/dnsmasq $(1)/usr/sbin/
 	$(INSTALL_DIR) $(1)/etc/config
-	$(INSTALL_CONF) ./files/dhcp.conf $(1)/etc/config/dhcp
+	$(INSTALL_DATA) ./files/dhcp.conf $(1)/etc/config/dhcp
 	$(INSTALL_CONF) ./files/dnsmasq.conf $(1)/etc/dnsmasq.conf
 	$(INSTALL_DIR) $(1)/etc/init.d
 	$(INSTALL_BIN) ./files/dnsmasq.init $(1)/etc/init.d/dnsmasq
@@ -188,6 +188,7 @@ define Package/dnsmasq/install
 	$(INSTALL_DIR) $(1)/etc/uci-defaults
 	$(INSTALL_BIN) ./files/50-dnsmasq-migrate-resolv-conf-auto.sh $(1)/etc/uci-defaults
 	$(INSTALL_BIN) ./files/50-dnsmasq-migrate-ipset.sh $(1)/etc/uci-defaults
+	$(INSTALL_BIN) ./files/50-dnsmasq-relax-dhcp-config-mode.sh $(1)/etc/uci-defaults
 endef
 
 Package/dnsmasq-dhcpv6/install = $(Package/dnsmasq/install)
--- a/package/network/services/odhcpd/Makefile
+++ b/package/network/services/odhcpd/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=odhcpd
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL=$(PROJECT_GIT)/project/odhcpd.git
@@ -28,6 +28,7 @@ define Package/odhcpd/default
   SECTION:=net
   CATEGORY:=Network
   DEPENDS:=+libjson-c +libubox +libuci +libubus +libnl-tiny
+  USERID:=odhcpd=455:odhcpd=455
 endef
 
 define Package/odhcpd/default/description
@@ -77,6 +78,10 @@ define Package/odhcpd/install
 	$(INSTALL_BIN) ./files/odhcpd-update $(1)/usr/sbin/
 	$(INSTALL_DIR) $(1)/etc/init.d
 	$(INSTALL_BIN) ./files/odhcpd.init $(1)/etc/init.d/odhcpd
+	$(INSTALL_DIR) $(1)/etc/capabilities
+	$(INSTALL_DATA) ./files/odhcpd.capabilities $(1)/etc/capabilities/odhcpd.json
+	$(INSTALL_DIR) $(1)/usr/share/acl.d
+	$(INSTALL_DATA) ./files/odhcpd_acl.json $(1)/usr/share/acl.d/odhcpd.json
 	$(INSTALL_DIR) $(1)/etc/uci-defaults
 	$(INSTALL_BIN) ./files/odhcpd.defaults $(1)/etc/uci-defaults/15_odhcpd
 endef
--- a/package/network/services/odhcpd/files/odhcpd.capabilities
+++ b/package/network/services/odhcpd/files/odhcpd.capabilities
@@ -1,0 +1,27 @@
+{
+	"bounding": [
+		"CAP_NET_ADMIN",
+		"CAP_NET_RAW",
+		"CAP_NET_BIND_SERVICE"
+	],
+	"effective": [
+		"CAP_NET_ADMIN",
+		"CAP_NET_RAW",
+		"CAP_NET_BIND_SERVICE"
+	],
+	"ambient": [
+		"CAP_NET_ADMIN",
+		"CAP_NET_RAW",
+		"CAP_NET_BIND_SERVICE"
+	],
+	"permitted": [
+		"CAP_NET_ADMIN",
+		"CAP_NET_RAW",
+		"CAP_NET_BIND_SERVICE"
+	],
+	"inheritable": [
+		"CAP_NET_ADMIN",
+		"CAP_NET_RAW",
+		"CAP_NET_BIND_SERVICE"
+	]
+}
--- a/package/network/services/odhcpd/files/odhcpd_acl.json
+++ b/package/network/services/odhcpd/files/odhcpd_acl.json
@@ -1,0 +1,24 @@
+{
+	"user": "odhcpd",
+	"publish": [
+		"dhcp"
+	],
+	"subscribe": [
+		"network.interface"
+	],
+	"listen": [
+		"ubus.object.add"
+	],
+	"access": {
+		"network.interface": {
+			"methods": [
+				"dump"
+			]
+		},
+		"service": {
+			"methods": [
+				"signal"
+			]
+		}
+	}
+}
--- a/package/network/services/ppp/files/50-ppp-radius-servers-mode.sh
+++ b/package/network/services/ppp/files/50-ppp-radius-servers-mode.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+# The RADIUS secret is read by uid ppp; INSTALL_CONF ships it 0600 root:root.
+[ -f /etc/ppp/radius/servers ] || exit 0
+chgrp ppp /etc/ppp/radius/servers
+chmod 0640 /etc/ppp/radius/servers
+
+exit 0
--- a/package/network/services/ppp/files/pppd_acl.json
+++ b/package/network/services/ppp/files/pppd_acl.json
@@ -1,0 +1,15 @@
+{
+	"user": "ppp",
+	"access": {
+		"network.interface": {
+			"methods": [
+				"notify_proto"
+			]
+		},
+		"network": {
+			"methods": [
+				"add_dynamic"
+			]
+		}
+	}
+}


brada4:
procd patch should go upstream first

caps file is NOT conffile - it is system policy that may need to  be upgraded by system .

package/network/services/ppp/Makefile
```diff
@@ -70,6 +70,7 @@ This package contains the PPP (Point-to-Point Protocol) daemon.
 endef
 
 define Package/ppp/conffiles
+/etc/capabilities/pppd.json
```
brada4: This prevents file to be ever upgrading, defeating ehole intent of having capability profile at all

brada4:
Reference for /etc/
https://specifications.freedesktop.org/fhs/latest/etc.html
Generally-immutable file can be in etc (as long as it is not executable)

package/network/services/ppp/files/ppp.sh
```diff
@@ -146,7 +146,12 @@ ppp_generic_setup() {
 	[ "$delegate" != "0" ] && delegate=""
 	[ "$norelease" = "1" ] || norelease=""
 
-	proto_run_command "$config" /usr/sbin/pppd \
+	# pppd parses LCP/*CP/PAP/CHAP/EAP from the peer before the link is
+	# authenticated, so keep only the capabilities it needs to set the link up.
+	local jail caps=/etc/capabilities/pppd.json
+	[ -x /sbin/ujail ] && [ -f "$caps" ] && jail="/sbin/ujail -t 5 -n pppd -C $caps -c --"
+
+	proto_run_command "$config" $jail /usr/sbin/pppd \
```
openwrt-ai, line 157: Interposing ujail changes what `$ERROR` means in `ppp_generic_teardown()`.

netifd records the raw wait status through `WEXITSTATUS()` without a `WIFEXITED()` guard ([`proto_ext_task_cb()` at proto-ext.c:213-220](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/proto-ext.c#L213-L220)), so today a pppd that dies from a signal yields `ERROR=0`. ujail instead *normalises* a signal death into an exit code: [`jail_process_handler()` at jail.c:1105-1113](https://github.com/openwrt/procd/blob/60fdbf00e924fff0a510c4ec31c520d35e69bfe1/jail/jail.c#L1105-L1113) sets `jail_return_code = WTERMSIG(ret)` and ujail then `exit()`s with it.

Signal numbers collide head-on with pppd's exit-code table that [`ppp_generic_teardown()`](https://github.com/openwrt/openwrt/blob/a67f8b59dde6cf9f2252ef4a43bc0ee2438812b4/package/network/services/ppp/files/ppp.sh#L182-L198) switches on:

* pppd segfaults (11) → `ERROR=11` → `PEER_AUTH_FAILED`, and with `option authfail 1` netifd calls `proto_block_restart` — the link stays down until manual intervention, where before it just retried.
* pppd OOM-killed (9) → `ERROR=9` → `PTYCMD_FAILED`.

`last_error` is only latched in `S_IDLE`/`S_SETUP`, so the ordinary teardown path is unaffected; it is the unexpected-death path that changes. Worth either filtering signal-derived codes here or documenting that the mapping is now approximate.

---

package/system/procd/patches/0001-jail-don-t-let-the-jailed-process-outlive-the-jail.patch
```diff
@@ -0,0 +1,38 @@
...
+diff --git a/jail/jail.c b/jail/jail.c
+index eaaf017..379474a 100644
+--- a/jail/jail.c
++++ b/jail/jail.c
+@@ -1177,6 +1177,9 @@ static int exec_jail(void *arg)
+ 	exit_from_child = true;
+ 	prctl(PR_SET_SECUREBITS, 0);
+ 
++	/* nothing can stop us once the supervisor is gone, so don't outlive it */
++	prctl(PR_SET_PDEATHSIG, SIGKILL);
```
openwrt-ai: Setting PDEATHSIG this early in `exec_jail()` makes it a no-op for a large share of ujail's users, because the kernel clears `pdeath_signal` on *any* credential change, not just on exec:

```c
	if (!uid_eq(old->euid, new->euid) ||
	    !gid_eq(old->egid, new->egid) ||
	    !uid_eq(old->fsuid, new->fsuid) ||
	    !gid_eq(old->fsgid, new->fsgid) ||
	    !cred_cap_issubset(old, new)) {
		...
		task->pdeath_signal = 0;
```
([`commit_creds()` at kernel/cred.c:405-413](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/kernel/cred.c#L405-L413), identical in 6.18)

Between this prctl and the `execve()` ujail can change credentials twice:

* `-U <user>` / `-G <group>` → [`set_jail_user()` at jail.c:935-951](https://github.com/openwrt/procd/blob/60fdbf00e924fff0a510c4ec31c520d35e69bfe1/jail/jail.c#L935-L951) does `setregid()`/`setreuid()` to a non-root id, clearing it.
* user-namespace jails (`-f`) → the `setregid(0,0)`/`setreuid(0,0)` in [`exec_jail()` at jail.c:1212-1224](https://github.com/openwrt/procd/blob/60fdbf00e924fff0a510c4ec31c520d35e69bfe1/jail/jail.c#L1212-L1224), a few lines below this hunk, clears it too.

Dropping capabilities is safe (`cred_cap_issubset` only compares `cap_permitted` and the new set is a subset), so the pppd case here — no `-U`, no `-G`, no userns — does keep the setting, which is why the hardware test passes. But the commit message frames this as a general fix for procd's own supervisors, and jails that run as a dedicated user are exactly the ones that will silently not get it.

Moving the prctl to the end of `post_start_hook()`, immediately before the `execve()`/`execvpe()` call, puts it after both credential transitions and after the final `applyOCIcapabilities()`, so it survives in all cases.

---

package/network/services/ppp/Makefile
openwrt-ai: Adding the capability profile to `conffiles` freezes it across upgrades, so a future fix to the cap set (e.g. adding a capability a proto turns out to need) would never reach any device that already has the file. That matters more than usual here because getting it wrong takes the WAN link down.

The two existing in-tree capability profiles both avoid this — neither [`/etc/capabilities/ntpd.json` (busybox Makefile:167)](https://github.com/openwrt/openwrt/blob/a67f8b59dde6cf9f2252ef4a43bc0ee2438812b4/package/utils/busybox/Makefile#L167) nor [`/etc/capabilities/wpad.json` (hostapd Makefile:740)](https://github.com/openwrt/openwrt/blob/a67f8b59dde6cf9f2252ef4a43bc0ee2438812b4/package/network/services/hostapd/Makefile#L740) is listed as a conffile. Same point brada4 raised in the thread; adding the in-tree precedent for it.

```suggestion
define Package/ppp/conffiles
```

---

package/network/services/ppp/files/pppd.capabilities
```diff
@@ -0,0 +1,22 @@
+{
+	"bounding": [
+		"CAP_NET_ADMIN",
+		"CAP_NET_RAW"
```
openwrt-ai, line 4: Question on the pppoa path, which the PR lists as covered-but-untested: the ATM connect path has a `CAP_NET_BIND_SERVICE` gate on low VCIs.

```c
	if (vci > 0 && vci < ATM_NOT_RSV_VCI && !capable(CAP_NET_BIND_SERVICE))
		return -EPERM;
```
([`__vcc_connect()` at net/atm/common.c:392](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/net/atm/common.c#L392); `ATM_NOT_RSV_VCI` is 32)

The [OpenWrt default is `vpi 8` / `vci 35`](https://github.com/openwrt/openwrt/blob/a67f8b59dde6cf9f2252ef4a43bc0ee2438812b4/package/network/services/ppp/files/ppp.sh#L298), which is above the gate, and VCIs 1–31 are reserved by the ATM spec so should never appear in a real PPPoA config — but `vci` is a free-form user option, so anyone who has one configured below 32 goes from working to `-EPERM` at connect time. `net/atm/pppoatm.c` itself only wants `CAP_NET_ADMIN`, which is covered. Is it worth adding `CAP_NET_BIND_SERVICE` for symmetry with `wpad.json`, or is the reserved-range argument enough to leave it out?

---

package/system/procd/patches/0001-jail-don-t-let-the-jailed-process-outlive-the-jail.patch
```diff
@@ -0,0 +1,38 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
openwrt-ai: nit: the all-zero SHA in the `From` line is a placeholder that will read as a real (but unresolvable) upstream reference once this is in the tree. For patches that are not upstream yet, in-tree practice is to drop the `From <sha> Mon Sep 17 ...` line entirely and start at `From:` — see `package/network/services/hostapd/patches/021-fix-sta-add-after-previous-connection.patch`, which begins directly with `From: Felix Fietkau <nbd@nbd.name>`.

---

openwrt-ai (commented):
Reviewed 2 commits (fresh review, both commits). Commit messages match their diffs.

Grounding used: procd `main` at the pinned `PKG_SOURCE_VERSION` `60fdbf0`, netifd `main` at `e97e36f`, and Linux v6.12.100 (checked against v6.18.41 where the code is capability-related). Inline comments cover the ujail exit-code translation vs. pppd's exit-code table, a case where `PR_SET_PDEATHSIG` gets cleared before exec, the conffiles entry, and a question about `CAP_NET_BIND_SERVICE` on the pppoa path.

---

brada4:
> Makefile installs configuration files under /etc/, but is missing the required 'conffiles' section. Please add a 'define Package//conffiles' block listing each installed config file path (e.g. '/etc/config/'), terminated with 'endef'.

Thats a complete BS, how are the init/hotplug scripts working here....

package/network/services/ppp/files/pppd.capabilities
JuliusBairaktaris: Kept `CAP_NET_BIND_SERVICE` out deliberately: the kernel gate is exactly on the ATM-reserved range (`vci < ATM_NOT_RSV_VCI`, i.e. 1–31 per the spec), and the OpenWrt default (`vpi 8`/`vci 35`) sits above it. A `vci` inside the reserved range is a broken config and the connect-time `-EPERM` is the correct failure. pppoatm itself only needs `CAP_NET_ADMIN`, which the profile covers.

JuliusBairaktaris:
@brada4 series is ready for review

brada4:
Superfluously looking at it
- inheritable caps might be reduced or nulled depending on what is actually forked out
- is there a piece for one of services that does not depend on external change? Maybe that could be split out in a separate pr?

JuliusBairaktaris:
The only non openwrt dependency is ppp which can just be patched for now.

package/network/services/odhcpd/files/odhcpd.init
```diff
@@ -4,14 +4,66 @@ START=35
...
+	config_get hostsdir odhcpd hostsdir
+	config_get piodir odhcpd piodir
+	config_foreach odhcpd_ndp_relay dhcp
+
+	for dir in "${leasefile%/*}" "$hostsdir" "$piodir"; do
+		[ -n "$dir" ] || continue
+		mkdir -p "$dir" || continue
+		# a world-writable directory (/tmp) needs no owner change
+		[ -n "$(find "$dir" -maxdepth 0 -perm -0002)" ] ||
+			chown odhcpd:odhcpd "$dir"
```
openwrt-ai, line 37: This test fails open: any `find` failure yields an empty command substitution, so the `||` branch runs and the directory gets chowned anyway.

`-maxdepth` and `-perm` are both optional busybox applet features — `default y` in [Config-defaults.in:1166-1180](https://github.com/openwrt/openwrt/blob/14122a9edc551357b587a594fa78fcdb06f4431e/package/utils/busybox/Config-defaults.in#L1166-L1180), but selectable in menuconfig, and busybox `find` dies on an unrecognised operand. With either turned off, the default `leasefile=/tmp/odhcpd.leases` makes `$dir` `/tmp` and this line becomes `chown odhcpd:odhcpd /tmp` — uid 455 then owns the sticky world directory and can unlink or rename every other user's files in it, a strictly larger grant than the jail is meant to hand out.

Making the chown `find`'s own action fails closed instead (`!` and `-exec ... \;` are both `default y`; `-exec ... +` is not, so keep `\;`):

```suggestion
		# a world-writable directory (/tmp) needs no owner change
		find "$dir" -maxdepth 0 ! -perm -0002 -exec chown odhcpd:odhcpd {} \;
```


---

package/network/services/odhcpd/files/odhcpd.init
```diff
@@ -4,14 +4,66 @@ START=35
...
+	done
+
+	[ -n "$relay" ] && [ -e /proc/sys/net/ipv6/conf/all/proxy_ndp ] &&
+		echo 1 > /proc/sys/net/ipv6/conf/all/proxy_ndp
```
openwrt-ai, line 41: The global knob is only ever raised, never lowered. Switch the last `option ndp relay` section back to `disabled`/`hybrid` and reload: `relay` is empty, this line is skipped, and `net.ipv6.conf.all.proxy_ndp` stays at 1 for the rest of the boot.

That is not inert — the kernel ORs it into every interface:

```c
		     (READ_ONCE(net->ipv6.devconf_all->proxy_ndp) ||
		      READ_ONCE(idev->cnf.proxy_ndp)) &&
```
([`ndisc_recv_ns()` at net/ipv6/ndisc.c:913-914](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/net/ipv6/ndisc.c#L913-L914), unchanged in 6.18), and [ndisc.c:1109](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/net/ipv6/ndisc.c#L1109) gates redirect generation on the `all` value alone. So after relay is turned off the router keeps answering NS for any leftover or third-party proxy neigh entry on every interface, on a knob the admin never set.

An explicit `else` writing 0 closes it, at the cost of stomping an admin's own `/etc/sysctl.conf` setting — worth deciding one way or the other rather than leaving it one-way.

---

package/network/services/odhcpd/files/odhcpd.init
```diff
@@ -4,14 +4,66 @@ START=35
...
+
+	config_load dhcp
+	config_get leasefile odhcpd leasefile
+	config_get hostsdir odhcpd hostsdir
+	config_get piodir odhcpd piodir
+	config_foreach odhcpd_ndp_relay dhcp
+
+	for dir in "${leasefile%/*}" "$hostsdir" "$piodir"; do
+		[ -n "$dir" ] || continue
+		mkdir -p "$dir" || continue
```
openwrt-ai, line 34: Only directories get handed over here, never the files already in them — and for the leasefile the directory is deliberately skipped.

The default is [`set dhcp.odhcpd.leasefile=/tmp/odhcpd.leases` (odhcpd.defaults:127)](https://github.com/openwrt/openwrt/blob/14122a9edc551357b587a594fa78fcdb06f4431e/package/network/services/odhcpd/files/odhcpd.defaults#L127), so `$dir` is `/tmp`, correctly left alone as world-writable. But `/tmp` is also sticky (1777), and the sticky bit blocks rename/unlink of files you do not own even in a directory you can write. If a root-owned `/tmp/odhcpd.leases` from the previous unjailed instance is still there — `opkg`/`apk` upgrade followed by `/etc/init.d/odhcpd restart`, no reboot — the `renameat()` that commits the lease file returns EPERM and leases silently stop being persisted until the next boot clears `/tmp`. The hostsdir/piodir cases are fine because the directory itself is chowned, which substitutes for the missing sticky bit there.

Is `chown odhcpd:odhcpd "$leasefile"` (when it exists) worth adding alongside the directory loop?

---

package/network/services/ppp/files/ppp.sh
```diff
@@ -146,7 +146,15 @@ ppp_generic_setup() {
 	[ "$delegate" != "0" ] && delegate=""
 	[ "$norelease" = "1" ] || norelease=""
 
-	proto_run_command "$config" /usr/sbin/pppd \
+	local jail caps=/etc/capabilities/pppd.json
+	[ -x /sbin/ujail ] && [ -f "$caps" ] && {
+		jail="/sbin/ujail -t 5 -n pppd -U ppp -G ppp -C $caps -c --"
+		chown ppp /dev/ppp
+		mkdir -p /var/run/pppd
+		chown ppp /var/run/pppd
```
openwrt-ai, line 154: nit: the commit message explains the `/dev/ppp` chown in detail ("The wrapper also hands `/dev/ppp` to the ppp user") but never mentions these two lines, so a reader has to reverse-engineer why `/var/run/pppd` needs to exist and be owned by uid 454.

Worth a sentence saying which files pppd puts there (2.5.x writes its pid file and `pppd2.tdb` under `PPP_PATH_VARRUN`, and nothing in this package creates that directory today) and whether a failure there is fatal or just noisy — that also documents the assumption that `PPP_PATH_VARRUN` is `/var/run/pppd` rather than `/var/run`, which is not visible anywhere in the tree.

---

package/network/services/ppp/patches/502-allow-nonroot-with-capnetadmin.patch
```diff
@@ -0,0 +1,171 @@
...
+--- a/pppd/main.c
++++ b/pppd/main.c
+@@ -401,7 +401,9 @@ main(int argc, char *argv[])
+     umask(umask(0777) | 022);
+ 
+     uid = getuid();
+-    privileged = uid == 0;
++    /* Privileged options come from the invoking user's own authority, so a
++     * setuid-root pppd must not hand them to whoever ran it. */
++    privileged = uid == 0 || (uid == geteuid() && ppp_privileged());
```
openwrt-ai, line 56: `uid == geteuid()` catches setuid-root, but not file capabilities — and file capabilities are the other way a non-root process ends up holding `CAP_NET_ADMIN`, which is exactly what `ppp_privileged()` now tests for.

With `setcap cap_net_admin=ep /usr/sbin/pppd`, an arbitrary local user gets `uid == geteuid()` (both their own) and `ppp_privileged()` true, so `privileged = 1` and `parse_args()` hands them every `OPT_PRIV` option — including `plugin`, i.e. `dlopen()` of an attacker-chosen object in a process with `CAP_NET_ADMIN` effective. Upstream's `privileged = uid == 0` did not have that hole, and the comment right above claims the guard is about "the invoking user's own authority", which a file capability is not.

OpenWrt does not set file caps on pppd, so this is not a live issue in this tree — but since the patch is aimed at upstream, `getauxval(AT_SECURE)` (or a `PR_GET_SECUREBITS`/`prctl` check) would cover privilege gained through *any* exec-time transition, not just setuid. Worth folding into ppp-project/ppp#608 before it lands there.

---

package/network/services/ppp/Makefile
```diff
@@ -49,6 +49,7 @@ define Package/ppp/Default
   SECTION:=net
   CATEGORY:=Network
   URL:=https://ppp.samba.org/
+  USERID:=ppp=454:ppp=454
```
openwrt-ai, line 52: nit: `Package/ppp/Default` is expanded by all 13 packages this Makefile builds, so `chat`, `pppdump`, `pppstats` and `pppoe-discovery` now create the `ppp` user and group too, even though none of them ever runs pppd.

hostapd repeats `USERID:=network=101:network=101` in each of the three packages that need it rather than putting it in a shared `/Default` ([hostapd Makefile:170](https://github.com/openwrt/openwrt/blob/14122a9edc551357b587a594fa78fcdb06f4431e/package/network/services/hostapd/Makefile#L170), [:276](https://github.com/openwrt/openwrt/blob/14122a9edc551357b587a594fa78fcdb06f4431e/package/network/services/hostapd/Makefile#L276), [:421](https://github.com/openwrt/openwrt/blob/14122a9edc551357b587a594fa78fcdb06f4431e/package/network/services/hostapd/Makefile#L421)). Same shape applies here — `Package/ppp` and `Package/ppp-multilink` are the two that ship the wrapper.

---

package/network/services/ppp/Makefile
```diff
@@ -211,6 +212,10 @@ define Package/ppp/install
 	$(INSTALL_DATA) ./files/etc/ppp/filter $(1)/etc/ppp/
 	$(INSTALL_DATA) ./files/etc/ppp/options $(1)/etc/ppp/
 	$(LN) /tmp/resolv.conf.ppp $(1)/etc/ppp/resolv.conf
+	$(INSTALL_DIR) $(1)/etc/capabilities
+	$(INSTALL_DATA) ./files/pppd.capabilities $(1)/etc/capabilities/pppd.json
+	$(INSTALL_DIR) $(1)/usr/share/acl.d
+	$(INSTALL_DATA) ./files/usr/share/acl.d/pppd.json $(1)/usr/share/acl.d/
```
openwrt-ai, line 218: nit: three different source layouts for the same two artifacts across this series and the tree.

Both in-tree precedents keep the pair flat next to the init script and rename on install — busybox uses `files/ntpd.capabilities` + `files/ntpd_acl.json` ([Makefile:166-168](https://github.com/openwrt/openwrt/blob/14122a9edc551357b587a594fa78fcdb06f4431e/package/utils/busybox/Makefile#L166-L168)), hostapd uses `files/wpad.json` + `files/wpad_acl.json` ([Makefile:740](https://github.com/openwrt/openwrt/blob/14122a9edc551357b587a594fa78fcdb06f4431e/package/network/services/hostapd/Makefile#L740)). This commit instead mirrors the target path into `files/usr/share/acl.d/pppd.json`, while the odhcpd commit in the same series uses a third spelling, `files/odhcpd.acl` ([odhcpd Makefile:84](https://github.com/openwrt/openwrt/blob/14122a9edc551357b587a594fa78fcdb06f4431e/package/network/services/odhcpd/Makefile#L84)).

Following the existing `<name>_acl.json` convention in both commits would keep them consistent with each other and with the tree:

```suggestion
	$(INSTALL_DATA) ./files/pppd_acl.json $(1)/usr/share/acl.d/pppd.json
```


---

package/network/services/dnsmasq/files/50-dnsmasq-relax-dhcp-config-mode.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# /etc/config/dhcp shipped 0600 while every generated UCI config lands 0644,
+# which locked it away from the unprivileged readers of it (odhcpd, LuCI).
```
openwrt-ai, line 4: nit: both halves of this comment contradict something else in the series.

"every generated UCI config lands 0644" is the opposite of what the commit message says three paragraphs down — *"network and system are 0600 because uci created them"*. And LuCI is not an unprivileged reader of this file: the uhttpd commit in this same PR keeps the server at uid 0 precisely because dropping it breaks cgi-io, so LuCI reads `/etc/config/dhcp` as root today and after this series.

odhcpd is the one real reader that stops being root, and that is the whole justification — worth saying just that:

```suggestion
# odhcpd uci_load()s this file directly and no longer runs as root, and
# INSTALL_CONF shipped it 0600, a mode uci_file_commit() then preserves.
```


---

openwrt-ai (commented):
Re-reviewed at `14122a9`. The branch was rebased and reworked since my last review at `a67f8b5`: the procd patch is gone, the ppp commit grew the uid drop and the ubus ACL, and odhcpd/dnsmasq/uhttpd are new — so this is effectively a fresh pass over all four commits. Each commit's subject scope matches the files it touches, so no commit-message findings.

Grounding: procd `main` at the pinned `PKG_SOURCE_VERSION` `60fdbf0` (ujail option parsing, `post_start_hook()` ordering, `applyOCIcapabilities()`), netifd `main` at `e97e36f` (`network`/`network.interface` object and method names behind the ACL grants), and Linux v6.12.100 with v6.18.41 cross-checked for the `proxy_ndp` semantics. The ujail wiring itself checks out — the `-t/-n/-U/-G/-C/-c` order matches what procd's own `instance.c` builds, `SECBIT_NOROOT` stays clear so the uid-0 uhttpd child still gets the bounding set at exec, and both ACL files name objects and methods that actually exist.

Findings are inline: a fail-open `chown` in `odhcpd_prepare` that can land on `/tmp`, a one-way `all.proxy_ndp` write, the leasefile's own ownership under a sticky `/tmp`, a file-capability gap in the pppd `privileged` guard, and four nits. Items already covered in the PR body — the odhcpd `PKG_SOURCE_VERSION` bump waiting on openwrt/odhcpd#412, the procd#40 dependency, and `CAP_NET_BIND_SERVICE` being left out of the pppoa profile — are deliberately not repeated. My earlier note about ujail's signal-death exit codes colliding with pppd's exit-code table in `ppp_generic_teardown()` is still open and still applies; that normalisation lives in procd's `jail_process_handler()`, not in the patch that was dropped.

---

brada4:
Do at your own pace, I dont see flaws in your approach or implementation

JuliusBairaktaris:
I will change uhttpd to not use uid = 0 and then it should be feature complete.

JuliusBairaktaris:
@brada4 i moved uhttpd here: https://github.com/openwrt/openwrt/pull/24740

package/network/services/ppp/patches/502-allow-nonroot-with-capnetadmin.patch
```diff
@@ -0,0 +1,195 @@
...
++}
++
++/*
++ * Test whether the running binary carries file capabilities.  The kernel
++ * grants a file's capability set to whoever execs it, so an unprivileged
++ * caller would hold CAP_NET_ADMIN without anyone having intended it.
++ */
++int ppp_file_cap_priv(void)
++{
++    return getxattr("/proc/self/exe", "security.capability", NULL, 0) >= 0;
```
openwrt-ai: This guard fails open: `>= 0` treats every `getxattr()` error as "no file capabilities".

`ENODATA` is the intended one (binary carries no fcaps), but the same `-1` comes back for `ENOENT` when `/proc` is not mounted or `/proc/self/exe` no longer resolves (binary replaced in place by a package upgrade), `EOPNOTSUPP` on a filesystem that does not carry `security.*` xattrs, and `EACCES` on a path component. Each of those re-enables exactly the grant this function exists to withhold — `privileged = 1`, and `parse_args()` then hands an unprivileged invoker the `OPT_PRIV` `plugin` option on a binary that holds `CAP_NET_ADMIN` from its file caps.

`getauxval(AT_SECURE)` answers the same question without touching the filesystem, and answers it with the kernel's own verdict on whether *this* exec raised privileges:

```c
	/* Check for privilege-elevated exec. */
	if (is_setid ||
	    (!__is_real(root_uid, new) &&
	     (effective ||
	      __cap_grew(permitted, ambient, new))))
		bprm->secureexec = 1;
```
([`cap_bprm_creds_from_file()` at security/commoncap.c:966-971](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/security/commoncap.c#L966-L971); [same conclusion in 6.18 at :990-996](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/security/commoncap.c#L990-L996))

It is 1 for setuid-root and for a `setcap cap_net_admin=ep` pppd (`effective` is the file's fE bit), and 0 for the ujail case this series needs: with no file caps [`pP' = pA'`](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/security/commoncap.c#L935-L941), so `__cap_grew(permitted, ambient, new)` is false, `effective` is false and `is_setid` is false. It also covers setgid and LSM-driven secure exec, which the `uid == geteuid()` half of the `main.c` condition does not.

```suggestion
+    return getauxval(AT_SECURE) != 0;
```

That wants `#include <sys/auxv.h>` in place of `<sys/xattr.h>` on line 124 of this patch — both are one-liners, so the hunk headers stay valid — and the function plus its `pppd.h` comment would read better named for "the exec raised privileges" rather than "file capabilities".

---

package/network/services/dnsmasq/files/50-dnsmasq-relax-dhcp-config-mode.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# odhcpd parses /etc/config/dhcp directly and stops running as root in the
+# odhcpd commit; the 0600 INSTALL_CONF default would lock it out.
```
openwrt-ai: nit: "in the odhcpd commit" only parses while reading this series. Once merged, the reader of this file has no way to tell which commit that is, and the sentence reads as if the change is still pending.

```suggestion
# odhcpd uci_load()s /etc/config/dhcp directly and no longer runs as root;
# INSTALL_CONF shipped it 0600, which locked it out.
```

Same forward reference shows up in the series order: this commit is third, but [`75f67e6`](https://github.com/openwrt/openwrt/commit/75f67e6ca6f5b9a2f21efac387af40d4830c71ae) already needs it ("the mode of that file is fixed in the dnsmasq commit that ships it"). Between the two commits a factory-fresh image has a jailed odhcpd that cannot read the file it gets its pools from — the exact failure the commit message describes. Ordering this one first makes every commit in the series independently bootable.

---

openwrt-ai (commented):
Re-reviewed at `15842df`, incrementally against my last pass at `14122a9`. The branch was rewritten: the uhttpd commit is gone (moved to #24740, and `package/network/services/uhttpd/` is now byte-identical to `main`), the two ACL files were renamed to the in-tree `<name>_acl.json` spelling, `USERID` moved out of `Package/ppp/Default` into the two packages that ship the wrapper, the `find -maxdepth -perm` probe in `odhcpd_prepare` became an `ls -ld` mode-string `case`, the leasefile handover was added, and the 502 patch grew `ppp_file_cap_priv()` and was refreshed quilt-style. Each commit's subject scope still matches the files it touches and the bodies match the diffs, so no commit-message findings.

Most of my previous round is addressed. The `find` probe now fails closed — `d*)` only matches a real `ls -ld` line, so a failed `ls` chowns nothing, and the `????????w*` pattern lands on the other-write column for every mode string busybox emits. The leasefile chown uses `-h`, so a symlink planted in `/tmp` cannot redirect it, and `fs.protected_hardlinks=1` in [`base-files/etc/sysctl.d/10-default.conf`](https://github.com/openwrt/openwrt/blob/15842df1a9f677f8e3c2c0ad6ec1fd24bc55ae87/package/base-files/files/etc/sysctl.d/10-default.conf#L8) closes the hardlink variant. The 502 hunk offsets are internally consistent (+4, +5 cumulative) and both build jobs are green, so the refresh is real.

Deliberately not repeated: the one-way `all.proxy_ndp` write, now argued in the commit message and in an in-file comment; the odhcpd `PKG_SOURCE_VERSION` bump waiting on openwrt/odhcpd#412; `CAP_NET_BIND_SERVICE` left out of the pppoa profile; and the `/var/run/pppd` rationale, which the commit message now carries. My open comment on ujail's signal-death exit codes colliding with pppd's table in `ppp_generic_teardown()` still stands and is covered by the openwrt/procd#40 dependency in the PR body. The `FormalityCheck / OpenWrt Makefiles` failure is the conffiles complaint already argued in the thread; not restating it.

Two inline findings: `ppp_file_cap_priv()` reading every `getxattr()` error as "no file capabilities", and a nit on the uci-defaults comment plus the series ordering.

Grounding: Linux v6.12.103 (`security/commoncap.c` secure-exec computation), cross-checked against v6.18.44.

---

JuliusBairaktaris:
@brada4 ready for review

package/network/services/dnsmasq/files/50-dnsmasq-relax-dhcp-config-mode.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# odhcpd uci_load()s /etc/config/dhcp directly and runs unprivileged;
+# INSTALL_CONF shipped it 0600, which locked it out.
+chmod 0644 /etc/config/dhcp
```
brada4, line 5: Could be a bit stricter, by allowing reading only by dnsmasq group
```
chown :dnsmasq ...
chmod g+r ...
```

brada4 (approved):
In general usage will work, probably some package will pop up later clashing with sandboxes, but that can be addressed in later refreshes in place of trying to contain whole world from the 1st day.

JuliusBairaktaris:
@robimarko can you run GPT over this, please?

package/network/services/dnsmasq/files/50-dnsmasq-relax-dhcp-config-mode.sh
JuliusBairaktaris: dnsmasq never opens this file (its init parses it as root, and the daemon is jailed without /etc/config mounted)

robimarko:
> @robimarko can you run GPT over this, please?

```
### Findings

  - [P1] Serial/3G PPP loses access to its TTY — ppp.sh:151 (https://github.com/openwrt/openwrt/blob/8ee38118cd1bd70787c5646481e428c9b2f41d3d/package/network/services/ppp/files/ppp.sh#L151)

    The jail runs pppd as user and group ppp, but OpenWrt creates TTY devices as 0660 root:dialout. CAP_NET_ADMIN and CAP_NET_RAW do not bypass filesystem permissions. Consequently, the 3G protocol’s /dev/ttyUSB* device passed through ppp_generic_setup() cannot be opened.

    Use the dialout group or explicitly grant the selected device to ppp before starting the jail. This affects the explicitly listed but untested serial PPP path.

  - [P2] Packaged authentication secrets remain unreadable — ppp/Makefile:212 (https://github.com/openwrt/openwrt/blob/8ee38118cd1bd70787c5646481e428c9b2f41d3d/package/network/services/ppp/Makefile#L212)

    chap-secrets and the RADIUS servers file are installed using INSTALL_CONF, meaning 0600 root:root. Because pppd is executed directly as UID 454, it cannot read either file; its network capabilities do not grant DAC access. Configurations using chap-secrets or ppp-mod-radius will therefore fail authentication.

    These should become narrowly readable by ppp, such as 0640 root:ppp, including migration of preserved files on existing installations. User-provided pap-secrets or alternative option files need equivalent documentation or handling.

  ### Known merge blockers

  The PR also remains intentionally dependent on external changes:

  - odhcpd#412 (https://github.com/openwrt/odhcpd/pull/412) is still open. Without it and a corresponding source bump, odhcpd rejects the non-root UID and cannot handle proxy_ndp.
  - rpcd#38 (https://github.com/openwrt/rpcd/pull/38) is still open and is documented as a security prerequisite for granting these non-root ubus ACLs.
  - The required procd fix has merged upstream but the OpenWrt package still needs the corresponding source bump.

  The patch applies cleanly and all current CI checks pass, but the runtime issues above are not exercised by those builds.
```

openwrt-ai (commented):

Re-reviewed at `8ee3811`, incrementally against my last pass at `15842df`. The branch was rewritten but the content delta is small: the series was reordered so the dnsmasq commit lands first, the 502 patch swapped `ppp_file_cap_priv()` for `ppp_secure_exec()`, and the uci-defaults comment was rewritten. Both of my previous round's findings are addressed:

- **`ppp_secure_exec()`** — `getauxval(AT_SECURE) != 0` replaces the `getxattr() >= 0` test, so the guard no longer fails open on `ENOENT`/`EOPNOTSUPP`/`EACCES`, and it now also covers setgid and LSM-driven secure exec. I re-checked the semantics against [`cap_bprm_creds_from_file()` at security/commoncap.c:966-971](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/security/commoncap.c#L966-L971) (v6.12.103; [same in v6.18.44 at :990-996](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/security/commoncap.c#L990-L996)): under ujail there is no setuid/setgid bit, ujail's own `setreuid()` means `new->euid == old->uid` so `is_setid` is false, and with no file caps `pP' = pA'` makes `__cap_grew(permitted, ambient, new)` false — `AT_SECURE` is 0 and `privileged` is still granted, while a setuid-root or `setcap cap_net_admin=ep` pppd keeps refusing an unprivileged invoker's `OPT_PRIV` options. The `<sys/xattr.h>` → `<sys/auxv.h>` swap is a one-for-one line replacement, so the hunk headers stay valid, and both are provided with `AT_SECURE` by musl and glibc.
- **Series ordering** — the dnsmasq mode change is now `b98df84`, first in the series, so no commit in the series boots with a jailed odhcpd that cannot read `/etc/config/dhcp`. The forward reference in the uci-defaults comment is gone.

I re-derived the 502 patch's hunk headers and diffstat by hand and they are internally consistent (main.c +14/-7, pppd.h +13, sys-linux.c +34, sys-solaris.c +20 = the 81/7 in the header; cumulative offsets +4/+5 in main.c and +3 in sys-linux.c line up). Each commit's subject scope matches the files it touches and each body matches its diff, so no commit-message findings. `/etc/config/dhcp` is still in `Package/dnsmasq/conffiles`, so the `INSTALL_CONF` → `INSTALL_DATA` swap only changes the mode, not upgrade preservation — and `FormalityCheck / OpenWrt Makefiles` is green now, so that thread is settled. All 14 checks on `8ee3811` pass.

Still open from earlier rounds, deliberately not restated inline: ujail's signal-death exit codes colliding with pppd's table in `ppp_generic_teardown()` (covered by the openwrt/procd#40 dependency in the PR body), the one-way `all.proxy_ndp` write, and the leasefile ownership question under a sticky `/tmp`. The odhcpd `PKG_SOURCE_VERSION` bump waiting on openwrt/odhcpd#412 and `CAP_NET_BIND_SERVICE` being left out of the pppoa profile are both argued in the PR body/threads.

Grounding: Linux v6.12.103 cross-checked against v6.18.44 (`security/commoncap.c`), procd `main` at `2c1834a`, netifd `main` at `e97e36f`.

---

package/network/services/dnsmasq/files/50-dnsmasq-relax-dhcp-config-mode.sh
brada4: Said that the whole chnage is not necessary

package/network/services/ppp/files/ppp.sh
```diff
@@ -146,7 +146,18 @@ ppp_generic_setup() {
...
+		# A serial link is handed its TTY as the first pppd argument, and
+		# a TTY is root:dialout 0660 - which no capability overrides.
+		[ -c "$1" ] && chown ppp "$1"
```
openwrt-ai: "the first pppd argument" only holds for `proto_ppp_setup`. The other in-tree caller that hands pppd a TTY is comgt's `proto_3g_setup`, which appends it **last**:

```sh
	ppp_generic_setup "$interface" \
		noaccomp \
		...
		crtscts \
		115200 "$device"
```

[`3g.sh:101-110`](https://github.com/openwrt/openwrt/blob/ff0694b2af19d31ad90c9c9baf27d8a79c9ea406/package/network/utils/comgt/files/3g.sh#L101-L110) — and `3g.sh` sources `ppp.sh` ([line 8](https://github.com/openwrt/openwrt/blob/ff0694b2af19d31ad90c9c9baf27d8a79c9ea406/package/network/utils/comgt/files/3g.sh#L8)), so the jail block runs for it too. After the `shift` at [ppp.sh:92](https://github.com/openwrt/openwrt/blob/ff0694b2af19d31ad90c9c9baf27d8a79c9ea406/package/network/services/ppp/files/ppp.sh#L92) `$1` is `noaccomp` here, `[ -c "$1" ]` is false, and the `/dev/ttyUSB*` node is left `root:dialout 0660`. uid 454 then can't open it — and neither can the `connect` chat script, which runs inside the same jail. That is the exact 3G case raised in [#24558 (comment)](https://github.com/openwrt/openwrt/pull/24558#issuecomment-5344477783), so it looks unintended rather than a deliberate scope limit.

Several ways to close it (scan all of `"$@"` for a char device, have comgt pass the device first, or a small `ppp_grant_tty` helper the callers invoke), so no suggestion block — but whichever you pick, the comment should stop claiming the TTY is always argument one.

---

package/network/services/ppp/Makefile
```diff
@@ -219,6 +225,25 @@ define Package/ppp/install
...
 Package/ppp-multilink/install=$(Package/ppp/install)
 
+# pppd reads its secrets as the unprivileged user, and both a fresh install and
+# a config file preserved from an older one leave them 0600 root:root.
+define Package/ppp/postinst
+#!/bin/sh
+for secret in /etc/ppp/chap-secrets /etc/ppp/pap-secrets; do
+	[ -f "$${IPKG_INSTROOT}$$secret" ] || continue
+	chgrp ppp "$${IPKG_INSTROOT}$$secret"
+	chmod 0640 "$${IPKG_INSTROOT}$$secret"
```
openwrt-ai: `chgrp ppp` cannot work in the case that matters — the image build — so the shipped image still can't hand `chap-secrets` to uid 454.

At image build `prepare_rootfs` runs each `post-install` script on the *build host* with `IPKG_INSTROOT` pointing at the offline root, and not under fakeroot ([rootfs.mk:86-93](https://github.com/openwrt/openwrt/blob/ff0694b2af19d31ad90c9c9baf27d8a79c9ea406/include/rootfs.mk#L86-L93), invoked from [image.mk:431](https://github.com/openwrt/openwrt/blob/ff0694b2af19d31ad90c9c9baf27d8a79c9ea406/include/image.mk#L431)). `chgrp` resolves the name `ppp` through the host's NSS, but the `ppp` group `add_group_and_user` just created was appended to `${IPKG_INSTROOT}/etc/group` — the *target's* group file, which `chgrp` never reads ([`group_add()` at functions.sh:462-470](https://github.com/openwrt/openwrt/blob/ff0694b2af19d31ad90c9c9baf27d8a79c9ea406/package/base-files/files/lib/functions.sh#L462-L470)). On a host with no `ppp` group it's `chgrp: invalid group: 'ppp'`; on one that happens to have it, the gid won't be 454 and the unprivileged builder can't chgrp into a group it isn't in anyway.

It also fails silently: `chmod` is the last command in the loop body, so `$?` is 0 and the exit-code check at rootfs.mk:89-91 never fires. The image ships `/etc/ppp/chap-secrets` as **0640 root:root**, not the `0640 root:ppp` the commit message describes — still unreadable by pppd. `PKG_FILE_MODES` isn't an escape either: `FILE_MODES` is only passed to `scripts/ipkg-build` ([package-pack.mk:531](https://github.com/openwrt/openwrt/blob/ff0694b2af19d31ad90c9c9baf27d8a79c9ea406/include/package-pack.mk#L531)), i.e. the non-APK path, and `USE_APK` is `default y`.

None of this shows up in CI, because the package-build jobs never run `prepare_rootfs`.

The sibling `b98df84` in this series already solves the same problem the right way — ship the mode via the install rule and migrate existing installs from `/etc/uci-defaults`, which runs on the device as root with a populated `/etc/group`. Same applies to `Package/ppp-mod-radius/postinst` on lines 240-245.

---

openwrt-ai (commented):
Re-reviewed at `ff0694b`, incrementally against my last pass at `8ee3811`. The branch was rebased (`b374143` → `7fae63e`, `8ee3811` → `ff0694b`; the dnsmasq commit `b98df84` is unchanged), and the only content delta is two additions to the ppp commit, both answering the P1/P2 findings in [#24558 (comment)](https://github.com/openwrt/openwrt/pull/24558#issuecomment-5344477783): a TTY chown in `ppp_generic_setup()` and a `postinst` that regrades the secrets files. The odhcpd commit's own diff is byte-identical to the previous round, so I did not re-review it.

Both new hunks have gaps, inline:

- The TTY chown tests `$1` only, which misses comgt's `proto_3g_setup` — it passes the device as the *last* argument, so the 3G/serial path the P1 finding named is still left with a `root:dialout 0660` node.
- The secrets `postinst` uses `chgrp ppp`, which resolves against the build host's NSS while the `ppp` group lives in `${IPKG_INSTROOT}/etc/group`. It fails silently at image-build time (the trailing `chmod` masks the exit status), so the image ships `0640 root:root` rather than the `0640 root:ppp` the commit message claims. `b98df84` in this same series already uses the mechanism that works.

## Commit checks

- `7fae63e` "ppp: run pppd in ujail as an unprivileged user" — the body says the postinst regrade "also covers a file preserved from an install that predates the uid". It doesn't: `prepare_rootfs` deletes each post-install script from `lib/apk/db/scripts.tar` right after running it at image-build time ([rootfs.mk:93](https://github.com/openwrt/openwrt/blob/ff0694b2af19d31ad90c9c9baf27d8a79c9ea406/include/rootfs.mk#L93)), so nothing re-runs it on the device after a sysupgrade — the one path on which a preserved `chap-secrets` arrives. Only an explicit `apk add`/`apk upgrade ppp` on a running device executes it. This is the reason the dnsmasq commit reaches for `/etc/uci-defaults` for exactly the same migration; the same paragraph also describes the result as `0640 root:ppp`, which is not what the mechanism produces (see the inline).

Deliberately not restated: ujail's signal-death exit codes colliding with pppd's table in `ppp_generic_teardown()` (covered by the openwrt/procd#40 dependency in the PR body), the one-way `all.proxy_ndp` write, the leasefile ownership question under a sticky `/tmp`, the odhcpd `PKG_SOURCE_VERSION` bump waiting on openwrt/odhcpd#412, and `CAP_NET_BIND_SERVICE` left out of the pppoa profile. `PKG_RELEASE` is correctly at 3 (base is 2). All 13 checks on `ff0694b` pass — but note neither new hunk is exercised by them.

Grounding: this tree at `ff0694b` (`include/rootfs.mk`, `include/image.mk`, `include/package-pack.mk`, `package/base-files/files/lib/functions.sh`, `package/network/utils/comgt/files/3g.sh`).

---

package/network/services/dnsmasq/files/50-dnsmasq-relax-dhcp-config-mode.sh
JuliusBairaktaris: So do you want me to make it stricter or drop the change?

brada4: Permission change at that point does exactly nothing for particular application

package/network/services/ppp/files/ppp.sh
```diff
@@ -146,7 +146,21 @@ ppp_generic_setup() {
 	[ "$delegate" != "0" ] && delegate=""
 	[ "$norelease" = "1" ] || norelease=""
 
-	proto_run_command "$config" /usr/sbin/pppd \
+	local jail caps=/etc/capabilities/pppd.json
+	[ -x /sbin/ujail ] && [ -f "$caps" ] && {
```
openwrt-ai: The guard covers `/sbin/ujail` and the caps file but not the `ppp` user itself, and ujail hard-fails on an unresolvable `-U` name ([`jail.c:1340-1350`](https://github.com/openwrt/procd/blob/2c1834abae171261f2d9161e5a28c36cbcc394a4/jail/jail.c#L1340-L1350)) — so pppd never starts at all rather than falling back to the old path. `/etc/passwd` and `/etc/group` are in [`keep.d/base-files-essential`](``https://github.com/openwrt/openwrt/blob/340a4b7111cd2e47b939cb4b05bd8cddc8c230a5/package/base-files/files/lib/upgrade/keep.d/base-files-essential#L4-L5``) ("always kept") and restored over the new image on every sysupgrade, while [`13_fix-group-user`](https://github.com/openwrt/openwrt/blob/340a4b7111cd2e47b939cb4b05bd8cddc8c230a5/package/base-files/Makefile#L257) — the boot-time repair that re-adds missing `USERID`s — is deleted from APK images, so uid 454 would not exist on an upgraded device. Is a user-resolves check worth adding to the gate?

```suggestion
	[ -x /sbin/ujail ] && [ -f "$caps" ] && id -u ppp >/dev/null 2>&1 && {
```

(`user_exists` is not reachable here: comgt's `3g.sh` sets `INCLUDE_ONLY=1`, so `ppp.sh` never sources `/lib/functions.sh` on that path.)

---

package/network/services/ppp/files/ppp.sh
```diff
@@ -146,7 +146,21 @@ ppp_generic_setup() {
...
+		jail="/sbin/ujail -t 5 -n pppd -U ppp -G ppp -C $caps -c --"
+		chown ppp /dev/ppp
+		mkdir -p /var/run/pppd
+		chown ppp /var/run/pppd
+		# A serial link is handed its TTY as a pppd argument - first by
+		# proto_ppp_setup, last by comgt's proto_3g_setup - and a TTY is
+		# root:dialout 0660, which no capability overrides.
+		for arg in "$@"; do
```
openwrt-ai, line 158: nit: `arg` is undeclared and leaks into the caller's shell; the file's other loop declares its variable (`local ip serv_addr server interface` in `proto_pptp_setup`).

```suggestion
		local arg
		for arg in "$@"; do
```

---

package/network/services/ppp/files/50-ppp-secrets-mode.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+# pppd reads these as uid ppp, and INSTALL_CONF ships them 0600 root:root -
+# as does a file preserved across sysupgrade from an install predating the uid.
+for secret in /etc/ppp/chap-secrets /etc/ppp/pap-secrets; do
```
openwrt-ai, line 5: `/etc/ppp/pap-secrets` is never shipped — `Package/ppp/install` installs only `chap-secrets` and nothing else in the tree creates it — so the `[ -f ]` test is always false here and the script is then deleted by `uci_apply_defaults`. An admin who configures PAP afterwards ends up with a file uid 454 cannot read; is that deliberately left to them, or should the grant happen at setup time in `ppp.sh` the way the TTY one now does?

---

openwrt-ai (commented):
Reviewed 1 new commit; commit messages match their diffs.

---

package/network/services/ppp/files/ppp.sh
JuliusBairaktaris: > Is a user-resolves check worth adding to the gate?

I agree with that. The gate now reads:

    package/network/services/ppp/files/ppp.sh, ppp_generic_setup():

        local jail arg caps=/etc/capabilities/pppd.json
        [ -x /sbin/ujail ] && [ -f "$caps" ] && id -u ppp >/dev/null 2>&1 && {

`id` rather than `user_exists` because ppp.sh only sources /lib/functions.sh
inside its `[ -n "$INCLUDE_ONLY" ] ||` block, and comgt's 3g.sh sets
INCLUDE_ONLY=1 before sourcing ppp.sh, so `user_exists` is not defined on that
path. `id` needs nothing sourced.

The gate condition was exercised on an IPQ8074 box, kernel 6.18, by creating
the ppp user and the capability file by hand, running the same condition and
argument walk ppp.sh has, and removing both afterwards: with the user absent
the wrapper is not built and pppd runs unwrapped, with it present the wrapper
is built and the TTY chown fires for a device passed either first or last in
the argument list. pppd itself I never ran jailed, because the packaged pppd
still refuses a non-root euid.

package/network/services/ppp/files/ppp.sh
JuliusBairaktaris: > nit: `arg` is undeclared and leaks into the caller's shell

Agreed. Folded into the declaration that is already there rather than adding a
second `local` line:

    local jail arg caps=/etc/capabilities/pppd.json

package/network/services/odhcpd/files/odhcpd.init
JuliusBairaktaris: > An explicit `else` writing 0 closes it, at the cost of stomping an admin's own `/etc/sysctl.conf` setting - worth deciding one way or the other rather than leaving it one-way.

You are right that raising a global and never lowering it was wrong, and it
turns out the init script never needed to touch it: the write is gone at the
current head.

Measured on an IPQ8074 box, kernel 6.18, in a jail carrying the same
capability profile as odhcpd, running as an unprivileged uid, built with the
same ujail options procd passes for a capability jail with a user and a group
and PR_SET_NO_NEW_PRIVS:

    /sbin/ujail -t 5 -n ct -U nobody -G nogroup -C /tmp/ct/with.json -c -- /tmp/ct/try.sh

`/tmp/ct/with.json` holds the same five capability sets as
`package/network/services/odhcpd/files/odhcpd.capabilities` - CAP_NET_ADMIN,
CAP_NET_RAW, CAP_NET_BIND_SERVICE. That is a shell in the jail, not odhcpd
itself.

Writing 1 to `/proc/sys/net/ipv6/conf/br-lan/proxy_ndp` and to
`.../conf/all/proxy_ndp` both succeed there, and `ip -6 neigh add proxy`
succeeds. With CAP_NET_ADMIN removed from the same profile and nothing else
changed, both writes fail with EACCES. Three runs of each, interleaved.

odhcpd does no more with that node than the shell did:

    odhcpd src/ndp.c, ndp_setup_interface():

        procfd = open(procbuf, O_WRONLY);
        ...
                if (write(procfd, "1\n", 2) < 0) {}

so a jailed odhcpd holding CAP_NET_ADMIN makes the write that succeeded above,
and the global stays with whoever set it.

The same premise sits under openwrt/odhcpd#412, whose "ndp: tolerate an
unwritable proxy_ndp" commit is written for a daemon that cannot write that
file. Should that commit be dropped from #412?

package/network/services/dnsmasq/files/50-dnsmasq-relax-dhcp-config-mode.sh
JuliusBairaktaris: > Permission change at that point does exactly nothing for particular application

You are right for dnsmasq - it never opens /etc/config/dhcp, so the mode change
does nothing for it.

The reader is odhcpd. It calls `uci_load()` on the dhcp package itself, and the
odhcpd commit in this same series drops it to uid 455, so 0600 root:root locks
it out of the pools it serves. dnsmasq only owns the file because it is the
package that ships it.

A group grant cannot be made to hold for any group, because uci's commit writes
a fresh file and renames it over the old one, carrying the mode but not the
owner or group:

    libuci file.c, uci_file_commit():

        if (!path || stat(path, &statbuf) || chmod(filename, statbuf.st_mode) || rename(filename, path)) {

So `chown :odhcpd` would survive until the first `uci commit dhcp` and then be
back to root:root, while the 0644 is preserved by that same line. That is my
reading of the source; I have not watched it happen on a device.

package/network/services/ppp/files/50-ppp-secrets-mode.sh
JuliusBairaktaris: > `/etc/ppp/pap-secrets` is never shipped - `Package/ppp/install` installs only `chap-secrets` and nothing else in the tree creates it - so the `[ -f ]` test is always false here

Not in every build. A files/ overlay lands in the rootfs before the image is
packed, and uci-defaults still run on the device at first boot:

    include/image.mk:431
        $(call prepare_rootfs,$(mkfs_cur_target_dir),$(TOPDIR)/files)

    target/imagebuilder/files/Makefile:294
        $(call prepare_rootfs,$(TARGET_DIR),$(USER_FILES),$(DISABLED_SERVICES))

`prepare_rootfs` copies that tree in as its first step, so a buildroot
`files/etc/ppp/pap-secrets` or `make image FILES=...` reaches the script.

> is that deliberately left to them, or should the grant happen at setup time in `ppp.sh`

Left to the admin. The package does not ship the file, the tree's pattern for a
one-time mode migration is a uci-defaults script, and the netifd path hands the
credentials to pppd as `user`/`password` options rather than through a secrets
file, so pap-secrets only comes up on a hand-written setup. Say if you would
rather ppp.sh granted it per ifup instead.

package/network/services/ppp/files/ppp.sh
JuliusBairaktaris: > Worth either filtering signal-derived codes here or documenting that the mapping is now approximate.

Neither. It closes with openwrt/procd#40 plus the PKG_SOURCE_VERSION bump that
carries it, which is a declared dependency of this series rather than something
the pppd side filters.

After that bump ujail reports a signal death as 128+signal, so a SIGSEGV arrives
as 139, and ppp.sh has no case for it:

    package/network/services/ppp/files/ppp.sh, ppp_exitcode_tostring():

        *) echo "UNKNOWN_ERROR" ;;

`ppp_generic_teardown()` then falls into its own `*)`, which calls
`proto_notify_error` and nothing else - no `proto_block_restart`, so netifd
retries. Before the jail the same death reached it as `ERROR=0` and was silent:
the retry is unchanged, the error report is new.

JuliusBairaktaris:
@robimarko P1, the TTY. `ppp_generic_setup()` scans its whole argument list for
a character device, so it covers `proto_ppp_setup` passing the TTY first and
comgt's `proto_3g_setup` passing it last:

    package/network/services/ppp/files/ppp.sh, ppp_generic_setup():

        for arg in "$@"; do
                [ -c "$arg" ] && chown ppp "$arg"
        done

Measured on an IPQ8074 box with the ppp user and the capability file created
by hand for the run: the wrapper is built and the chown fires for a device
passed either first or last. There is no modem on my bench, so the serial and
3G link itself is still unexercised.

P2, the secrets. The postinst chgrp is gone; it could not work in the case that
matters, because it runs on the build host with IPKG_INSTROOT set and `chgrp
ppp` resolves the name through the host's /etc/group rather than the target's.
The regrade moved to uci-defaults scripts that run on the device as root, one
for chap-secrets and one for the RADIUS server list.

package/network/services/odhcpd/files/odhcpd.init
```diff
@@ -4,14 +4,64 @@ START=35
...
 	procd_open_instance
 	procd_set_param command /usr/sbin/odhcpd
 	procd_set_param respawn
+
+	odhcpd_jailed && {
+		procd_add_jail odhcpd
+		procd_set_param capabilities "$ODHCPD_CAPS"
+		procd_set_param no_new_privs 1
+		procd_set_param user odhcpd
+		procd_set_param group odhcpd
```
openwrt-ai, line 56: procd only stores `user`/`group` when `getpwnam()`/`getgrnam()` resolve them ([`instance_config_parse()` at service/instance.c:1354-1371](https://github.com/openwrt/procd/blob/2c1834abae171261f2d9161e5a28c36cbcc394a4/service/instance.c#L1354-L1371)) and only passes `-U`/`-G` when it did ([:323-331](https://github.com/openwrt/procd/blob/2c1834abae171261f2d9161e5a28c36cbcc394a4/service/instance.c#L323-L331)), so on a device whose preserved `/etc/passwd` predates uid 455 this fails open: the jail and the capability profile still apply, but odhcpd keeps running as root with no error anywhere — only `odhcpd_prepare`'s `chown odhcpd:odhcpd` complains. The ppp side now gates on `id -u ppp` for exactly this case; should `odhcpd_jailed()` gain the same check so the missing uid leaves the daemon unjailed rather than silently root?

---

openwrt-ai (commented):
Reviewed 2 new commits; commit messages match their diffs.

---

openwrt-ai (commented):

---

package/network/services/odhcpd/files/odhcpd.init
openwrt-ai: fixed, thanks

---

openwrt-ai (commented):

---

PR #24560, opened 2026-08-03
tools/meson: update to 1.11.2

neheb:
Remove upstreamed patch and refresh others.

ping @Ansuel 

you planning on finishing upstreaming that patch?


tools/meson/patches/101-02-linkers-don-t-include-absolue-RPATH-on-cross-compili.patch
```diff
@@ -74,94 +74,94 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
...
+         if target.build_rpath != '':
+-            all_paths.add(target.build_rpath)
+-            for p in target.build_rpath.split(':'):
 -                rpath_dirs_to_remove.add(p.encode('utf8'))
-+            paths = build_rpath.split(':')
++            paths = target.build_rpath.split(':')
 +            for p in paths:
 +                # Only include relative paths if we can't run binaries on host system.
 +                # (cross-compilation and no exe_wrapper)
 +                if env.can_run_host_binaries() or not os.path.isabs(p):
```
openwrt-ai: The refresh moved the hunks to the new context but kept `env`, which no longer exists in these methods. In meson 1.11 `build_rpath_args()` was reworked to `build_rpath_args(self, build_dir, from_dir, target, extra_paths=None)` — the `env: Environment` parameter is gone (that is also why the surrounding context now reads `target.build_rpath` / `extra_paths`). So every `env` reference the patch injects into a `build_rpath_args()` body is an undefined name.

Applying both patches to the 1.11.2 tarball and running a trivial `shared_library()` + `executable()` project reproduces it immediately:

```
File ".../mesonbuild/backend/ninjabackend.py", line 3733, in get_build_rpath_args
    linker.build_rpath_args(self.environment.get_build_dir(),
File ".../mesonbuild/linkers/linkers.py", line 760, in build_rpath_args
    processed_rpaths = prepare_rpaths(env, rpath_paths, build_dir, from_dir)
                                      ^^^
NameError: name 'env' is not defined
```

`GnuLikeDynamicLinkerMixin` is the common path, so this breaks every meson-based package build, not a corner case. The linker already holds the environment (`self.environment = env` in `DynamicLinker.__init__`, and this very method starts with `m = self.environment.machines[self.for_machine]`), so `self.environment` is the drop-in replacement — I verified the build succeeds after substituting it.

```suggestion
+                if self.environment.can_run_host_binaries() or not os.path.isabs(p):
```

The same substitution is needed at the other injected sites in this patch: lines 73, 99, 117 and 135 (`prepare_rpaths(env, ...)`) and lines 107, 125, 148 and 163 (`env.can_run_host_binaries()`). Lines 40 and 59 are fine as-is — those are the module-level `prepare_rpaths()` / `evaluate_rpath()` helpers that take `env` as a real argument.

Note CI is green only because the tools job builds the meson zipapp but never runs it against a package.

---

openwrt-ai (commented):
Reviewed 1 new commit; 1 inline issue found.

Verified separately (no action needed): `PKG_HASH` matches the 1.11.2 release tarball (sha256 `698feae0…8a736`), and the removed patch 100 is genuinely upstream — `depfixer: zero-out rpath entry string on removing entry` landed as mesonbuild/meson@78ffa5392394a62f7a50f54a2277f40bdb450907 and `clean_rpath_entry_string()` is present in the 1.11.2 tarball.

---

tools/meson/patches/101-02-linkers-don-t-include-absolue-RPATH-on-cross-compili.patch
```diff
@@ -74,94 +74,94 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
...
  
-         # TODO: should this actually be "for (dragonfly|open)bsd"?
-         if mesonlib.is_dragonflybsd() or mesonlib.is_openbsd():
-@@ -828,10 +837,15 @@ class AppleDynamicLinker(PosixDynamicLin
+@@ -937,10 +946,15 @@ class AppleDynamicLinker(PosixDynamicLin
          # @loader_path is the equivalent of $ORIGIN on macOS
          # https://stackoverflow.com/q/26280738
          origin_placeholder = '@loader_path'
 -        processed_rpaths = prepare_rpaths(rpath_paths, build_dir, from_dir)
 +        processed_rpaths = prepare_rpaths(env, rpath_paths, build_dir, from_dir)
```
openwrt-ai: The `env` fix in the force-push is only half applied: the five `env.can_run_host_binaries()` sites were switched to `self.environment.can_run_host_binaries()`, but the four `prepare_rpaths(env, ...)` sites injected into the same method bodies still pass `env`.

Those two groups sit inside the same `build_rpath_args()` bodies, so whatever made `env` unavailable for the first group applies to the second. The refreshed context confirms the signature was reworked in 1.11 — the surrounding lines now read `target.build_rpath` / `target.install_rpath` / `target.determine_rpath_dirs()` / `extra_paths` instead of the old `build_rpath`, `install_rpath`, `rpath_paths` parameters. If `env` were still a parameter, the [`self.environment.can_run_host_binaries()` change 8 lines below](``https://github.com/openwrt/openwrt/blob/c89a5ca11fbc8f7bf98fc768b290d8ca963a3d0a/tools/meson/patches/101-02-linkers-don-t-include-absolue-RPATH-on-cross-compili.patch#L107)`` would not have been needed at all.

As it stands `GnuLikeDynamicLinkerMixin.build_rpath_args()` — the common path for every meson-built package — raises `NameError: name 'env' is not defined` on the first call. Same substitution needed at lines 73 (`GnuLikeDynamicLinkerMixin`), 117 (`NAGDynamicLinker`) and 135 (`SolarisDynamicLinker`). Lines 40 and 59 are correct as-is: those are the module-level `prepare_rpaths()` / `evaluate_rpath()` helpers, where `env` is a real parameter the patch adds.

```suggestion
+        processed_rpaths = prepare_rpaths(self.environment, rpath_paths, build_dir, from_dir)
```

Worth noting CI is green here only because the tools job builds the meson zipapp without ever running it against a package build.

---

openwrt-ai (commented):
Re-reviewed the force-pushed commit (`6ac101b` → `c89a5ca`). The only change to `tools/meson/` since the last review is the `env` → `self.environment` substitution in patch 101-02, which addresses part of the previously flagged problem — 5 of the 9 injected `env` references were fixed, 4 remain. Details inline.

The rest of the delta between the two SHAs is an unrelated rebase onto a newer `main`; nothing else in this PR changed.

Commit message still matches the diff (version bump, patch 100 removed as upstreamed, remaining patches refreshed), and CI on `c89a5ca` is fully green.

---

openwrt-ai (commented):
Re-reviewed the force-pushed commit (`c89a5ca` → `2b6499b`). The only change under `tools/` is in patch 101-02: the four remaining `prepare_rpaths(env, ...)` call sites are now `prepare_rpaths(self.environment, ...)`, which resolves the `NameError: name 'env' is not defined` flagged in the previous two reviews. The rest of the delta between the two SHAs is an unrelated rebase onto a newer `main`.

Verified against the 1.11.2 release tarball (sha256 `698feae0…8a736`, matches `PKG_HASH`):

- Both patches apply with no fuzz and no offsets; the added lines are otherwise identical to the pre-bump versions (only `env` → `self.environment` and `build_rpath` → `target.build_rpath`, both required by the reworked 1.11 `build_rpath_args()` signature). No hunks were lost in the refresh — still 8 hunks covering the same five linker classes (GnuLike, Apple, NAG, Solaris, AIX).
- No stale references remain: `prepare_rpaths` / `evaluate_rpath` have no other callers, and `_can_run_host_binaries_impl` is fully gone from the tree after 101-01.
- Built the zipapp with `packaging/create_zipapp.py` and ran a `shared_library()` + `executable(build_rpath: '/opt/absdir:relsub')` project through it using OpenWrt-shaped native/cross files. Native: `RUNPATH [$ORIGIN/:/opt/absdir:relsub]`. Cross with `needs_exe_wrapper = true`: `RUNPATH [$ORIGIN/:relsub]` — the absolute entry is dropped as intended, no exception. `meson setup` emits no new deprecation warnings for `include/meson.mk`'s `[built-in options]` machine files.
- Patch 100 is genuinely upstreamed: `clean_rpath_entry_string()` is present in 1.11.2 and called from `remove_rpath_entry()`.

Commit message matches the diff (version bump, patch 100 dropped as upstreamed, remaining two refreshed), and CI on `2b6499b` is green.

---

neheb:
ping @jonasjelonek

jonasjelonek:

PR #24561, opened 2026-08-03
ramips: add support for D-Link DWR-921 C3 with U-Boot

### LINKED ISSUE #21848: Newer versions of D-Link DWR-921 C3 are using U-Boot instead jboot
### Describe the bug

Newer versions of D-Link DWR-921 C3 with factory firmware 3.0.x are using U-Boot instead jboot. So,  neither -factory nor sysupgrade images are not working.

### OpenWrt version

r29087-d9c5716d1d

### OpenWrt release

24.10.5

### OpenWrt target/subtarget

ramips/mt7620

### Device

D-Link DWR-921 C3

### Image kind

Official downloaded image

### Steps to reproduce

_No response_

### Actual behaviour

_No response_

### Expected behaviour

_No response_

### Additional info

_No response_

### Diffconfig

```text

```

### Terms

dormancygrace:
Newer D-Link DWR-921 C3 units shipped with factory firmware 3.0.x use U-Boot and a different flash layout, so the existing JBoot C3 images cannot be installed or booted on them.

This adds a separate `dlink_dwr-921-c3-uboot` profile with:

- the factory U-Boot/environment/Factory/Linux partition map;
- Wi-Fi EEPROM and MAC data read from the Factory partition;
- a raw legacy-uImage factory image named `DWR_921`, matching the OEM updater;
- writable access to the OEM 4 KiB environment in its 64 KiB erase sector through `uboot-envtools`;
- USB2, QMI and serial-option support for the internal LTE modem;
- `zram-swap` with its standard OpenWrt defaults;
- the existing swconfig switch, LEDs, buttons, Wi-Fi and serial wiring.

The shared hardware description was factored out of the existing C1 DTS. Its DTB is byte-identical before and after that refactor. A separate follow-up commit modernizes the shared LED definitions while keeping the existing generated names except for the standard `green:wlan` Wi-Fi LED. Existing C1 and C3 U-Boot configurations migrate that one LED name; DWR-922 E2 remains unchanged.

The SPI clock is deliberately kept at the OEM-proven 25 MHz without fast-read. Disassembly of the stock U-Boot shows clock divisor 8, while the official D-Link GPL kernel independently selects the same divisor, leaves its fast-read mode disabled and issues normal `0x03` reads.

This profile deliberately retains the current MT7620 Ethernet and swconfig path. The separate, still-unmerged DSA, generic MediaTek Ethernet and PPE series are not prerequisites for this device support.

Validation:

- clean `ramips/mt7620` target configuration for this device;
- complete `make -j12` build;
- rootfs checked for `uboot-envtools`, USB2, QMI, zram-swap and swconfig;
- generated environment setup checked for a 4 KiB environment in a 64 KiB erase sector on `u-boot-env`;
- generated DTB checked for the 25 MHz SPI setting, partition map, NVMEM cells and generated LED names;
- factory/sysupgrade uImage headers, partition bounds, rootfs and image metadata inspected;
- official D-Link 3.0.2 updater, transition image, stock U-Boot and GPL kernel inspected;
- both stages of the official transition reproduced against a copy of a complete 16 MiB flash dump, preserving the legacy configuration sector;
- `git diff --check` and strict checkpatch pass with no findings.

There is no surviving factory-U-Boot unit in the current lab, so runtime testing on that bootloader is still needed. The available DWR-921 C3 has already been converted to JBoot and was not flashed for this target-only change.

Fixes #21848

### DIFF
--- a/package/boot/uboot-tools/uboot-envtools/files/ramips
+++ b/package/boot/uboot-tools/uboot-envtools/files/ramips
@@ -48,6 +48,9 @@ dlink,dir-806a-b1)
 dlink,covr-x1860-a1)
 	ubootenv_add_uci_config "/dev/mtd1" "0x2000" "0x4000" "0x2000"
 	;;
+dlink,dwr-921-c3-uboot)
+	ubootenv_add_mtd "u-boot-env" "0x0" "0x1000" "0x10000"
+	;;
 asus,rt-ax53u|\
 asus,rt-ax54|\
 asus,4g-ax56|\
--- a/target/linux/ramips/base-files/etc/uci-defaults/04_led_migration
+++ b/target/linux/ramips/base-files/etc/uci-defaults/04_led_migration
@@ -1,6 +1,12 @@
 . /lib/functions.sh
 . /lib/functions/migrations.sh
 
+case "$(board_name)" in
+dlink,dwr-921-c1)
+	migrate_leds "green:wifi=green:wlan"
+	;;
+esac
+
 remove_devicename_leds "rt2800soc-phy0" "rt2800pci-phy0"
 
 migrations_apply system
--- a/target/linux/ramips/dts/mt7620n_dlink_dwr-921-c3-uboot.dts
+++ b/target/linux/ramips/dts/mt7620n_dlink_dwr-921-c3-uboot.dts
@@ -1,0 +1,73 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+#include "mt7620n_dlink_dwr-921.dtsi"
+
+/ {
+	compatible = "dlink,dwr-921-c3-uboot", "ralink,mt7620n-soc";
+	model = "D-Link DWR-921 C3 (U-Boot)";
+};
+
+&spi0 {
+	status = "okay";
+
+	flash@0 {
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <25000000>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "u-boot";
+				reg = <0x0 0x30000>;
+				read-only;
+			};
+
+			partition@30000 {
+				label = "u-boot-env";
+				reg = <0x30000 0x10000>;
+			};
+
+			factory: partition@40000 {
+				label = "factory";
+				reg = <0x40000 0x10000>;
+				read-only;
+
+				nvmem-layout {
+					compatible = "fixed-layout";
+					#address-cells = <1>;
+					#size-cells = <1>;
+
+					eeprom_factory_0: eeprom@0 {
+						reg = <0x0 0x200>;
+					};
+
+					macaddr_factory_4: macaddr@4 {
+						compatible = "mac-base";
+						reg = <0x4 0x6>;
+						#nvmem-cell-cells = <1>;
+					};
+				};
+			};
+
+			partition@50000 {
+				compatible = "denx,uimage";
+				label = "firmware";
+				reg = <0x50000 0xfb0000>;
+			};
+		};
+	};
+};
+
+&ethernet {
+	nvmem-cells = <&macaddr_factory_4 0>;
+	nvmem-cell-names = "mac-address";
+};
+
+&wmac {
+	nvmem-cells = <&eeprom_factory_0>, <&macaddr_factory_4 1>;
+	nvmem-cell-names = "eeprom", "mac-address";
+};
--- a/target/linux/ramips/dts/mt7620n_dlink_dwr-921.dtsi
+++ b/target/linux/ramips/dts/mt7620n_dlink_dwr-921.dtsi
@@ -32,7 +32,8 @@
 		compatible = "gpio-leds";
 
 		sms {
-			label = "green:sms";
+			function = "sms";
+			color = <LED_COLOR_ID_GREEN>;
 			gpios = <&gpio1 14 GPIO_ACTIVE_LOW>;
 		};
 
@@ -43,27 +44,32 @@
 		};
 
 		led_sstrenghg: sstrengthg {
-			label = "green:sigstrength";
+			function = "sigstrength";
+			color = <LED_COLOR_ID_GREEN>;
 			gpios = <&gpio2 0 GPIO_ACTIVE_LOW>;
 		};
 
 		sstrengthr {
-			label = "red:sigstrength";
+			function = "sigstrength";
+			color = <LED_COLOR_ID_RED>;
 			gpios = <&gpio2 1 GPIO_ACTIVE_LOW>;
 		};
 
 		4g {
-			label = "green:4g";
+			function = "4g";
+			color = <LED_COLOR_ID_GREEN>;
 			gpios = <&gpio2 2 GPIO_ACTIVE_LOW>;
 		};
 
 		3g {
-			label = "green:3g";
+			function = "3g";
+			color = <LED_COLOR_ID_GREEN>;
 			gpios = <&gpio2 3 GPIO_ACTIVE_LOW>;
 		};
 
 		wifi {
-			label = "green:wifi";
+			function = LED_FUNCTION_WLAN;
+			color = <LED_COLOR_ID_GREEN>;
 			gpios = <&gpio3 0 GPIO_ACTIVE_LOW>;
 		};
 	};
--- a/target/linux/ramips/dts/mt7620n_dlink_dwr-921.dtsi
+++ b/target/linux/ramips/dts/mt7620n_dlink_dwr-921.dtsi
@@ -1,0 +1,107 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+#include "mt7620n.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+
+/ {
+	aliases {
+		led-boot = &led_sstrenghg;
+		led-failsafe = &led_sstrenghg;
+	};
+
+	keys {
+		compatible = "gpio-keys";
+
+		wps {
+			label = "wps";
+			gpios = <&gpio0 2 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_WPS_BUTTON>;
+		};
+
+		reset {
+			label = "reset";
+			gpios = <&gpio0 1 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		sms {
+			label = "green:sms";
+			gpios = <&gpio1 14 GPIO_ACTIVE_LOW>;
+		};
+
+		lan {
+			function = LED_FUNCTION_LAN;
+			color = <LED_COLOR_ID_GREEN>;
+			gpios = <&gpio1 15 GPIO_ACTIVE_HIGH>;
+		};
+
+		led_sstrenghg: sstrengthg {
+			label = "green:sigstrength";
+			gpios = <&gpio2 0 GPIO_ACTIVE_LOW>;
+		};
+
+		sstrengthr {
+			label = "red:sigstrength";
+			gpios = <&gpio2 1 GPIO_ACTIVE_LOW>;
+		};
+
+		4g {
+			label = "green:4g";
+			gpios = <&gpio2 2 GPIO_ACTIVE_LOW>;
+		};
+
+		3g {
+			label = "green:3g";
+			gpios = <&gpio2 3 GPIO_ACTIVE_LOW>;
+		};
+
+		wifi {
+			label = "green:wifi";
+			gpios = <&gpio3 0 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+	gpio_export {
+		compatible = "gpio-export";
+
+		lte_modem_enable {
+			gpio-export,name = "lte_modem_enable";
+			gpio-export,output = <1>;
+			gpios = <&gpio0 0 GPIO_ACTIVE_HIGH>;
+		};
+	};
+};
+
+&gpio1 {
+	status = "okay";
+};
+
+&gpio2 {
+	status = "okay";
+};
+
+&gpio3 {
+	status = "okay";
+};
+
+&ehci {
+	status = "okay";
+};
+
+&ohci {
+	status = "okay";
+};
+
+&state_default {
+	default {
+		groups = "spi refclk", "i2c", "ephy", "wled";
+		function = "gpio";
+	};
+};
--- a/target/linux/ramips/mt7620/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/mt7620/base-files/etc/board.d/01_leds
@@ -80,6 +80,7 @@ dlink,dwr-118-a2)
 	ucidef_set_led_netdev "wifi_led" "wifi" "green:wlan2g" "wlan1"
 	;;
 dlink,dwr-921-c1|\
+dlink,dwr-921-c3-uboot|\
 dlink,dwr-922-e2)
 	ucidef_set_led_switch "lan" "lan" "green:lan" "switch0" "0x0f"
 	ucidef_set_led_netdev "signalstrength" "signalstrength" "green:sigstrength" "wwan0" "link"
--- a/target/linux/ramips/mt7620/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/mt7620/base-files/etc/board.d/01_leds
@@ -80,7 +80,12 @@ dlink,dwr-118-a2)
 	ucidef_set_led_netdev "wifi_led" "wifi" "green:wlan2g" "wlan1"
 	;;
 dlink,dwr-921-c1|\
-dlink,dwr-921-c3-uboot|\
+dlink,dwr-921-c3-uboot)
+	ucidef_set_led_switch "lan" "lan" "green:lan" "switch0" "0x0f"
+	ucidef_set_led_netdev "signalstrength" "signalstrength" "green:sigstrength" "wwan0" "link"
+	ucidef_set_led_netdev "4g" "4g" "green:4g" "wwan0" "tx rx"
+	ucidef_set_led_netdev "wifi_led" "wifi" "green:wlan" "wlan0"
+	;;
 dlink,dwr-922-e2)
 	ucidef_set_led_switch "lan" "lan" "green:lan" "switch0" "0x0f"
 	ucidef_set_led_netdev "signalstrength" "signalstrength" "green:sigstrength" "wwan0" "link"
--- a/target/linux/ramips/mt7620/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/mt7620/base-files/etc/board.d/02_network
@@ -13,6 +13,7 @@ ramips_setup_interfaces()
 	asus,rt-n12p|\
 	dlink,dwr-116-a1|\
 	dlink,dwr-921-c1|\
+	dlink,dwr-921-c3-uboot|\
 	dlink,dwr-922-e2|\
 	dovado,tiny-ac|\
 	hongdian,h8922-v30|\
@@ -357,6 +358,11 @@ ramips_setup_macs()
 	dlink,dwr-961-a1|\
 	lava,lr-25g001)
 		wan_mac=$(jboot_config_read -m -i $(find_mtd_part "config") -o 0xE000)
+		lan_mac=$(macaddr_add "$wan_mac" 1)
+		label_mac=$wan_mac
+		;;
+	dlink,dwr-921-c3-uboot)
+		wan_mac=$(mtd_get_mac_binary factory 0x4)
 		lan_mac=$(macaddr_add "$wan_mac" 1)
 		label_mac=$wan_mac
 		;;
--- a/target/linux/ramips/dts/mt7620n_dlink_dwr-921-c1.dts
+++ b/target/linux/ramips/dts/mt7620n_dlink_dwr-921-c1.dts
@@ -1,96 +1,12 @@
-#include "mt7620n.dtsi"
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
 
-#include <dt-bindings/gpio/gpio.h>
-#include <dt-bindings/input/input.h>
-#include <dt-bindings/leds/common.h>
+#include "mt7620n_dlink_dwr-921.dtsi"
+
 #include <dt-bindings/mtd/partitions/uimage.h>
 
 / {
 	compatible = "dlink,dwr-921-c1", "ralink,mt7620n-soc";
 	model = "D-Link DWR-921 C1";
-
-	aliases {
-		led-boot = &led_sstrenghg;
-		led-failsafe = &led_sstrenghg;
-	};
-
-	keys {
-		compatible = "gpio-keys";
-
-		wps {
-			label = "wps";
-			gpios = <&gpio0 2 GPIO_ACTIVE_LOW>;
-			linux,code = <KEY_WPS_BUTTON>;
-		};
-
-		reset {
-			label = "reset";
-			gpios = <&gpio0 1 GPIO_ACTIVE_LOW>;
-			linux,code = <KEY_RESTART>;
-		};
-	};
-
-	leds {
-		compatible = "gpio-leds";
-
-		sms {
-			label = "green:sms";
-			gpios = <&gpio1 14 GPIO_ACTIVE_LOW>;
-		};
-
-		lan {
-			function = LED_FUNCTION_LAN;
-			color = <LED_COLOR_ID_GREEN>;
-			gpios = <&gpio1 15 GPIO_ACTIVE_HIGH>;
-		};
-
-		led_sstrenghg: sstrengthg {
-			label = "green:sigstrength";
-			gpios = <&gpio2 0 GPIO_ACTIVE_LOW>;
-		};
-
-		sstrengthr {
-			label = "red:sigstrength";
-			gpios = <&gpio2 1 GPIO_ACTIVE_LOW>;
-		};
-
-		4g {
-			label = "green:4g";
-			gpios = <&gpio2 2 GPIO_ACTIVE_LOW>;
-		};
-
-		3g {
-			label = "green:3g";
-			gpios = <&gpio2 3 GPIO_ACTIVE_LOW>;
-		};
-
-		wifi {
-			label = "green:wifi";
-			gpios = <&gpio3 0 GPIO_ACTIVE_LOW>;
-		};
-	};
-
-	gpio_export {
-		compatible = "gpio-export";
-
-		lte_modem_enable {
-			gpio-export,name = "lte_modem_enable";
-			gpio-export,output = <1>;
-			gpios = <&gpio0 0 GPIO_ACTIVE_HIGH>;
-		};
-	};
-};
-
-&gpio1 {
-	status = "okay";
-};
-
-&gpio2 {
-	status = "okay";
-};
-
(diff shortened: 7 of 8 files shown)


package/boot/uboot-tools/uboot-envtools/files/ramips
```diff
@@ -48,6 +48,9 @@ dlink,dir-806a-b1)
 dlink,covr-x1860-a1)
 	ubootenv_add_uci_config "/dev/mtd1" "0x2000" "0x4000" "0x2000"
 	;;
+dlink,dwr-921-c3-uboot)
+	ubootenv_add_mtd "u-boot-env" "0x0" "0x10000" "0x10000"
+	;;
```
openwrt-ai, line 53: The environment *size* (3rd argument) is set to the full 64 KiB erase block, which is probably not what the OEM U-Boot uses. Mainline U-Boot's MT7620 reference config places a **4 KiB** environment at exactly the offset this DTS declares: [`mt7620_rfb_defconfig` L7-L9](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/configs/mt7620_rfb_defconfig#L7-L9) has `CONFIG_ENV_SIZE=0x1000`, `CONFIG_ENV_OFFSET=0x30000`, `CONFIG_ENV_SECT_SIZE=0x10000`. The offset and sector size match this board exactly, which suggests the vendor bootloader is derived from the same reference and also uses `0x1000`.

If that is the case, `fw_setenv` writes a 64 KiB blob whose CRC is computed over 64 KiB, U-Boot reads back only the first 4 KiB, the CRC check fails, and it falls back to the built-in default environment — losing `bootcmd`/`bootargs` on a device that, per the PR description, nobody can currently recover over serial.

Given that the env size is the one value here that can't be inferred from the flash layout, can you confirm it against a stock `u-boot-env` dump (a 4 KiB env leaves the rest of the sector as `0xff`)? If it turns out to be 4 KiB the line should read `ubootenv_add_mtd "u-boot-env" "0x0" "0x1000" "0x10000"`.

---

target/linux/ramips/mt7620/base-files/etc/uci-defaults/06_dwr921_zram
```diff
@@ -0,0 +1,12 @@
+. /lib/functions.sh
+
+case "$(board_name)" in
+dlink,dwr-921-c3-uboot)
+	uci set system.@system[0].zram_size_mb="32"
+	uci set system.@system[0].zram_comp_algo="lzo-rle"
+	uci set system.@system[0].zram_priority="100"
+	uci commit system
```
openwrt-ai: `zram_priority="100"` is a no-op — `zram.init` already falls back to exactly `100` when the option is unset ([zram.init:146-150](https://github.com/openwrt/openwrt/blob/8aed9a02f5763ee8fcb0a2e9fa2bfb6fad75a44c/package/system/zram-swap/files/zram.init#L146-L150)).

```suggestion
	uci set system.@system[0].zram_size_mb="32"
	uci set system.@system[0].zram_comp_algo="lzo-rle"
	uci commit system
```

Worth noting for the remaining two as well: `zram_size_mb` only raises the size from the computed default of `MemTotal / 2048` (≈29 MiB on a 64 MiB board) to 32 MiB, so `zram_comp_algo` is really the only setting that changes behaviour here — the init script defaults to plain `lzo`, not `lzo-rle`. This is also the only device-specific zram tuning anywhere under `target/`; is a dedicated uci-defaults file (shipped into every mt7620 image) warranted for a ~3 MiB size bump and a compressor switch, or would simply adding `zram-swap` to `DEVICE_PACKAGES` be enough?

---

target/linux/ramips/dts/mt7620n_dlink_dwr-921-c3-uboot.dts
```diff
@@ -0,0 +1,74 @@
...
+	model = "D-Link DWR-921 C3 (U-Boot)";
+};
+
+&spi0 {
+	status = "okay";
+
+	flash@0 {
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <50000000>;
```
openwrt-ai: The C1 — which this PR treats as the same board with a different bootloader, to the point of sharing a DTSI — clocks the same SPI NOR at 10 MHz and without `m25p,fast-read` ([`mt7620n_dlink_dwr-921-c1.dts:18`](https://github.com/openwrt/openwrt/blob/8aed9a02f5763ee8fcb0a2e9fa2bfb6fad75a44c/target/linux/ramips/dts/mt7620n_dlink_dwr-921-c1.dts#L18)). Jumping to 50 MHz on a variant that has never been booted on real hardware is the kind of change that shows up as intermittent read corruption rather than a clean failure, and there is no serial-recovery path documented for this unit.

Was 50 MHz actually measured on a C3 board, or carried over from another design? Unless it was verified, matching the C1's 10 MHz until someone can boot the image seems safer, with the bump as a follow-up.

---

target/linux/ramips/mt7620/base-files/etc/board.d/02_network
```diff
@@ -347,6 +348,11 @@ ramips_setup_macs()
 	trendnet,tha103ac)
 		lan_mac=$(mtd_get_mac_ascii factory lanmac)
 		;;
+	dlink,dwr-921-c3-uboot)
```
openwrt-ai: nit: out of order. `ramips_setup_macs` keeps its case groups sorted by first label, and this block lands between `dlink,dch-m225` and `dlink,dir-510l` — `dlink,dwr-921-c3-uboot` sorts after the whole `dlink,dir-510l|…|dlink,dwr-961-a1|lava,lr-25g001` group that now starts at line 356. Moving it below that group keeps the list scannable (the two additions in `ramips_setup_interfaces` and in `01_leds` are both correctly placed).

---

target/linux/ramips/dts/mt7620n_dlink_dwr-921.dtsi
```diff
@@ -0,0 +1,107 @@
...
+			gpios = <&gpio2 2 GPIO_ACTIVE_LOW>;
+		};
+
+		3g {
+			label = "green:3g";
+			gpios = <&gpio2 3 GPIO_ACTIVE_LOW>;
+		};
+
+		wifi {
+			label = "green:wifi";
```
openwrt-ai: nit: six of the seven LEDs here still use the deprecated `label = "<color>:<function>"` form; the project's target syntax is `color = <LED_COLOR_ID_*>` + `function = LED_FUNCTION_*`, and the `lan` node in the middle of this same block already does it that way (with `dt-bindings/leds/common.h` included on line 7).

I don't think it belongs in this hunk — converting them would break the byte-identical-DTB property the refactor is built on, and would need matching changes in `01_leds` plus new entries in [`04_led_migration`](https://github.com/openwrt/openwrt/blob/8aed9a02f5763ee8fcb0a2e9fa2bfb6fad75a44c/target/linux/ramips/base-files/etc/uci-defaults/04_led_migration) so existing C1/C3 configs keep working. Flagging it because the newly added `dlink,dwr-921-c3-uboot` inherits the legacy naming from day one; a follow-up commit converting the whole family would avoid that.

---

openwrt-ai (commented):
## Commit checks

- `8aed9a0` "ramips: add support for D-Link DWR-921 C3 with U-Boot" — this adds a new device (new `mt7620n_dlink_dwr-921-c3-uboot.dts` plus a new `define Device/dlink_dwr-921-c3-uboot`), so the commit message is expected to carry a **flash-instructions** section, and there is none. "The stock 3.0.2 updater accepts a raw legacy uImage named DWR_921" describes the image format, not an install path: which build output to pick, how to get the unit into a state where the updater will accept it (button/IP/hostname verbatim), and how to revert to stock. Since the PR states nobody currently has a factory-U-Boot unit, the person who eventually tests this will be working from exactly this text. The **hardware** section is also missing the flash and RAM part numbers, the UART header location and pinout, and the power rating, and the **MAC** section covers LAN and Wi-Fi (base + 1) but never states where the WAN and label MACs come from — the code uses `factory` + 0x4 unmodified for both.


---

package/boot/uboot-tools/uboot-envtools/files/ramips
dormancygrace: Confirmed from the stock U-Boot disassembly: the environment data and CRC cover exactly 0x1000 bytes, while saveenv erases a 0x10000-byte sector at 0x30000. Changed the tuple to 0x0/0x1000/0x10000 and documented the evidence in the commit message.

target/linux/ramips/mt7620/base-files/etc/uci-defaults/06_dwr921_zram
dormancygrace: Agreed. Removed the device-specific uci-defaults file entirely. The profile now only installs zram-swap and keeps the standard OpenWrt size, compressor, and priority defaults.

target/linux/ramips/mt7620/base-files/etc/board.d/02_network
dormancygrace: Fixed by moving the board case below the existing D-Link group.

target/linux/ramips/dts/mt7620n_dlink_dwr-921.dtsi
dormancygrace: Implemented as the suggested separate follow-up commit. The shared DTSI now uses function/color properties, 01_leds uses the generated names, and 04_led_migration preserves existing C1/C3/DWR-922 E2 UCI configurations.

dormancygrace:
Addressed the review feedback in the updated series. The device commit now includes the RAM and flash part numbers, power rating, J1 UART pinout, exact factory-image suffix, stock web installation, U-Boot recovery/revert procedure, and explicit WAN/label/LAN/Wi-Fi MAC derivation. It also documents the stock-U-Boot and D-Link-GPL evidence for 25 MHz normal SPI reads. The five inline threads have corresponding fixes and replies; LED modernization is a separate follow-up commit with UCI migration.

target/linux/ramips/base-files/etc/uci-defaults/04_led_migration
```diff
@@ -1,6 +1,20 @@
 . /lib/functions.sh
 . /lib/functions/migrations.sh
 
+case "$(board_name)" in
+dlink,dwr-921-c1|\
+dlink,dwr-921-c3-uboot|\
+dlink,dwr-922-e2)
```
openwrt-ai: `dlink,dwr-922-e2` does not use the shared DTSI. [`mt7620n_dlink_dwr-922-e2.dts`](https://github.com/openwrt/openwrt/blob/e946129750388083c0907aa61c4b5335c3ff0237/target/linux/ramips/dts/mt7620n_dlink_dwr-922-e2.dts#L1-L8) includes `mt7620n.dtsi` directly and still carries the legacy labels — [`label = "green:sms"` … `label = "green:wifi"` at L40-L71](https://github.com/openwrt/openwrt/blob/e946129750388083c0907aa61c4b5335c3ff0237/target/linux/ramips/dts/mt7620n_dlink_dwr-922-e2.dts#L40-L71). This PR does not touch that file, so the commit message's "The DTSI is also used by the DWR-922 E2" (and the same claim in the PR body) is not true.

Consequence on the E2, an already-supported device: this migration rewrites working `sysfs` values in existing configs to names that board will never expose (`green:mobile-1`, `red:mobile-1`, `green:mobile-4`, `green:mobile-3`, `green:mail`, `green:wlan`), and [`01_leds` L84-L89](https://github.com/openwrt/openwrt/blob/e946129750388083c0907aa61c4b5335c3ff0237/target/linux/ramips/mt7620/base-files/etc/board.d/01_leds#L84-L89) writes the same dead names into fresh configs. The signal-strength, 4G and WiFi LEDs stop working after upgrade.

Either convert `mt7620n_dlink_dwr-922-e2.dts` to `#include "mt7620n_dlink_dwr-921.dtsi"` as part of this commit, or drop `dlink,dwr-922-e2` from both this case and the `01_leds` group. (`dlink,dwr-921-c3` is fine — it builds from `DEVICE_DTS := mt7620n_dlink_dwr-921-c1` and reports `dlink,dwr-921-c1`.)

Note that the `function = "<func>"` alternative suggested on `mt7620n_dlink_dwr-921.dtsi` would make this whole hunk unnecessary and sidestep the problem entirely.

---

target/linux/ramips/dts/mt7620n_dlink_dwr-921.dtsi
```diff
@@ -0,0 +1,117 @@
...
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		sms {
+			function = LED_FUNCTION_MAIL;
+			color = <LED_COLOR_ID_GREEN>;
+			gpios = <&gpio1 14 GPIO_ACTIVE_LOW>;
+		};
```
openwrt-ai, line 38: `LED_FUNCTION_MAIL` is not an equivalent of `sms` — it renames this LED to `green:mail`. Same problem below: `function-enumerator = <4>` / `<3>` on the `4g` (L60-L65) and `3g` (L67-L72) nodes produce `green:mobile-4` / `green:mobile-3`. The enumerator disambiguates several LEDs that share one function and is used sequentially in-tree — [`mt7620a_hongdian_h8922-v30.dts` numbers its two mobile LEDs 1 and 2 at L47-L66](https://github.com/openwrt/openwrt/blob/e946129750388083c0907aa61c4b5335c3ff0237/target/linux/ramips/dts/mt7620a_hongdian_h8922-v30.dts#L47-L66) — it is not a slot for the radio generation. `green:mobile-4` reads as "fourth mobile LED", and the 3G/4G/SMS meaning is gone from the name (this also skips enumerator 2 entirely).

`.github/llm-review-rules.md` documents the fallback for exactly this case: `function = "<func>"` when no matching `LED_FUNCTION_*` constant exists. Applying it to `sms`, `sigstrength`, `4g` and `3g` keeps every generated name byte-identical to today's `label` values, which in turn makes the entire `04_led_migration` hunk and the `01_leds` renames unnecessary — and removes the DWR-922 E2 regression flagged there.

```suggestion
		sms {
			function = "sms";
			color = <LED_COLOR_ID_GREEN>;
			gpios = <&gpio1 14 GPIO_ACTIVE_LOW>;
		};
```

Same applies to `sstrengthg` (L46-L51) and `sstrengthr` (L53-L58) → `function = "sigstrength";`, `4g` (L60-L65) → `function = "4g";`, `3g` (L67-L72) → `function = "3g";`, all dropping `function-enumerator`. `lan` and `wifi` → `LED_FUNCTION_WLAN` are genuine constant matches; leave those as they are.

---

target/linux/ramips/mt7620/base-files/etc/board.d/01_leds
```diff
@@ -80,12 +80,13 @@ dlink,dwr-118-a2)
...
-	ucidef_set_led_netdev "4g" "4g" "green:4g" "wwan0" "tx rx"
-	ucidef_set_led_netdev "wifi_led" "wifi" "green:wifi" "wlan0"
-	;;
+dlink,dwr-921-c3-uboot|\
+	dlink,dwr-922-e2)
+		ucidef_set_led_switch "lan" "lan" "green:lan" "switch0" "0x0f"
+		ucidef_set_led_netdev "signalstrength" "signalstrength" "green:mobile-1" "wwan0" "link"
+		ucidef_set_led_netdev "4g" "4g" "green:mobile-4" "wwan0" "tx rx"
+		ucidef_set_led_netdev "wifi_led" "wifi" "green:wlan" "wlan0"
+		;;
```
openwrt-ai: nit: the continued case label and the whole body picked up an extra level of indentation. Every other case in this file puts labels at column 0 and bodies at one tab — compare `dlink,dwr-118-a2)` directly above and `dlink,dwr-960)` directly below.

```suggestion
dlink,dwr-922-e2)
	ucidef_set_led_switch "lan" "lan" "green:lan" "switch0" "0x0f"
	ucidef_set_led_netdev "signalstrength" "signalstrength" "green:mobile-1" "wwan0" "link"
	ucidef_set_led_netdev "4g" "4g" "green:mobile-4" "wwan0" "tx rx"
	ucidef_set_led_netdev "wifi_led" "wifi" "green:wlan" "wlan0"
	;;
```

---

openwrt-ai (commented):
Re-reviewed the two new commits since `8aed9a0`. The uboot-env size (`0x1000`), the SPI clock (25 MHz, no fast-read), the dropped `06_dwr921_zram` and the `02_network` case ordering all address the previous round; the amended device commit message now carries hardware, flash-instruction and MAC sections.

## Commit checks

- `e946129` "ramips: dwr-921: use function and color LED properties" — the message states "The DTSI is also used by the DWR-922 E2". It isn't: `mt7620n_dlink_dwr-922-e2.dts` is a standalone DTS that includes `mt7620n.dtsi` directly and still uses the legacy `label` properties, and this PR does not touch it. The `dlink,dwr-922-e2` entries this commit adds to `04_led_migration` and `01_leds` rest on that premise and break the LEDs on that device. Details inline.

---

target/linux/ramips/base-files/etc/uci-defaults/04_led_migration
dormancygrace: Confirmed. DWR-922 E2 is standalone. I removed it from the migration, split its 01_leds case, kept its legacy names, and corrected the commit and PR wording.

target/linux/ramips/dts/mt7620n_dlink_dwr-921.dtsi
dormancygrace: Changed sms, sigstrength, 3g and 4g to literal function strings and removed the enumerators. LED_FUNCTION_WLAN remains because it is a standard match, so only C1/C3 need the green:wifi to green:wlan migration.

target/linux/ramips/mt7620/base-files/etc/board.d/01_leds
dormancygrace: Fixed the indentation while splitting DWR-922 E2 into its own legacy-name case.

target/linux/ramips/base-files/etc/uci-defaults/04_led_migration
```diff
@@ -1,6 +1,13 @@
 . /lib/functions.sh
 . /lib/functions/migrations.sh
 
+case "$(board_name)" in
+dlink,dwr-921-c1|\
+dlink,dwr-921-c3-uboot)
```
openwrt-ai: nit: `dlink,dwr-921-c3-uboot` in a migration case is dead code — the board is introduced by this very series ([`define Device/dlink_dwr-921-c3-uboot` at mt7620.mk:336](https://github.com/openwrt/openwrt/blob/9ea34f34b28ec8c850500d4dd249d1740bc43ca3/target/linux/ramips/image/mt7620.mk#L336)), so no installed config can ever contain a `green:wifi` LED for it. `dlink,dwr-921-c1` alone covers the two boards that actually have legacy configs: the C1 itself, and the JBoot C3, which builds from `DEVICE_DTS := mt7620n_dlink_dwr-921-c1` and therefore reports `dlink,dwr-921-c1` to `board_name`.

```suggestion
dlink,dwr-921-c1)
```

---

openwrt-ai (commented):
Re-reviewed the new head since `e946129`. The device commit `881ebf5` is unchanged in both tree and message from the version already reviewed, so only `9ea34f3` is new.

That commit resolves the previous round: `mt7620n_dlink_dwr-922-e2.dts` is no longer implicated — the E2 keeps its own `01_leds` case with `green:wifi` untouched, and it is out of the `04_led_migration` case. The `function = "<func>"` fallback on `sms`, `sigstrength`, `4g` and `3g` (with `function-enumerator` dropped) reproduces the old `label` strings exactly: [`led_compose_name()` composes `"%s:%s"` from color and function at led-core.c:553-564](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/leds/led-core.c#L553-L564), so the generated names stay `green:sms`, `green:sigstrength`, `red:sigstrength`, `green:4g`, `green:3g`, with only `wifi` → `green:wlan` changing — which is what `04_led_migration` and the `01_leds` update cover. The `01_leds` indentation is fixed. The commit message matches the diff.

One nit inline. CI on `9ea34f3` is green.

---

target/linux/ramips/base-files/etc/uci-defaults/04_led_migration
dormancygrace: Fixed in 8e741e09bb. The new U-Boot board was removed from the migration case; only the existing C1 board remains.

mrkiko:
Your work is always excellent. Thanks a lot.

target/linux/ramips/base-files/etc/uci-defaults/04_led_migration
openwrt-ai: Re-opening the previous nit: the reply on the resolved thread says "Fixed in 8e741e09bb. The new U-Boot board was removed from the migration case; only the existing C1 board remains", but `dlink,dwr-921-c3-uboot` is still here on the current head. `8e741e0` and the previously reviewed `9ea34f3` have the identical tree (`f8ab7633afad84f17de657ce490dd7fbf74262f2`) and the identical commit message — the amend only changed the committer date, so no fix was actually pushed.

The finding stands: this board is introduced by this very series ([`define Device/dlink_dwr-921-c3-uboot` at mt7620.mk:336](https://github.com/openwrt/openwrt/blob/8e741e09bbee569472c43ec226df253592f713aa/target/linux/ramips/image/mt7620.mk#L336)), so no installed config can ever carry a `green:wifi` LED for it. `dlink,dwr-921-c1` alone covers both boards with legacy configs (the JBoot C3 builds from `DEVICE_DTS := mt7620n_dlink_dwr-921-c1` and reports `dlink,dwr-921-c1`).

```suggestion
dlink,dwr-921-c1)
```

---

openwrt-ai (commented):
The new head `8e741e0` carries no content change over the previously reviewed `9ea34f3`: same parent (`881ebf5`), same tree (`f8ab7633afad84f17de657ce490dd7fbf74262f2`), byte-identical commit message — only the committer date moved. So `git diff 9ea34f3..8e741e0` is empty and nothing in the series has changed since the last round.

That matters because the reply on the resolved `04_led_migration` thread attributes the fix to `8e741e09bb`; that fix is not in the tree. Re-flagged inline. Everything else from the last review still holds.

CI on `8e741e0` reports no checks yet (pending), so nothing to ground there.

---

target/linux/ramips/base-files/etc/uci-defaults/04_led_migration
dormancygrace: Fixed in 2d8fd8865c. The actual migration script now covers only dlink,dwr-921-c1; the new U-Boot C3 profile is no longer included, and the commit message matches the code.

openwrt-ai (commented):
Re-reviewed the new head `2d8fd88` against the previously reviewed `8e741e0`. The device commit `881ebf5` is untouched, so the only new content is in the LED commit.

`git diff 8e741e0..2d8fd88` is a single hunk, and it is the fix that was outstanding: `04_led_migration` now matches only `dlink,dwr-921-c1`, which is the board name both the C1 and the JBoot C3 report, and the newly introduced `dlink,dwr-921-c3-uboot` is gone from the case. The migration block sits above `remove_devicename_leds` with `migrations_apply system` still at the end, matching the layout of the other targets' `04_led_migration`. The `01_leds` split, the `function = "<func>"` fallbacks and the DWR-922 E2 legacy names are all unchanged and still correct. No inline issues this round.

## Commit checks

- `2d8fd88` "ramips: dwr-921: use function and color LED properties" — the reword that accompanied the fix re-wrapped the body at 76 columns, which trips checkpatch:

  ```
  WARNING: Possible unwrapped commit description (prefer a maximum 75 chars per line)
  #6:
  Replace legacy label properties in the shared DWR-921 DTSI with function and

  total: 0 errors, 1 warnings, 0 checks, 71 lines checked
  ```

  `COMMIT_LOG_LONG_LINE` fires above 75 chars ([checkpatch.pl:2776-2789](https://github.com/openwrt/openwrt/blob/2d8fd8865c8647e2de39bcfaf27f039457af074c/scripts/checkpatch.pl#L2776-L2789)); three body lines are at 75-76. The previous revision of this message was wrapped at ≤72 and was clean, and the PR body still claims "`git diff --check` and strict checkpatch pass with no findings" — that now holds for `881ebf5` but not for this commit. Re-wrapping the body at ~72 columns restores it.

---

dormancygrace:
Thank you, @mrkiko — I really appreciate that.

The commit-message warning from the latest review is fixed in 85338ce070. I rewrapped the two 76-column lines to 72 columns or less; the commit tree is unchanged. git diff --check and strict checkpatch now pass with 0 warnings.

openwrt-ai (commented):
Re-reviewed the new head `85338ce` against the previously reviewed `2d8fd88`. `git diff 2d8fd88..85338ce` is empty — same parent (`881ebf5`), same tree (`d1466f2254a9391a16b74fd0ce51541db4101639`) — so the only change is the reword of the LED commit message, exactly as described in the comment.

That reword is the fix for the one item outstanding from the last round. The three body lines that were at 75-76 columns are now at 63/67/69, and every line of the message is ≤ 69. `scripts/checkpatch.pl --strict` now reports no findings on either commit in the series, so the PR body's "strict checkpatch pass with no findings" claim holds again for the whole series.

Nothing else in the series has changed, and the earlier conclusions still stand: the `04_led_migration` case matches only `dlink,dwr-921-c1`, the `01_leds` split keeps the DWR-922 E2 on its own legacy names, and the `function = "<func>"` fallbacks reproduce the previous generated LED names apart from the intended `wifi` → `green:wlan`. No new inline issues.

The two red check runs on this head (`Check packages for malta/be`, `Set up build x86/64`) both fail with `Failed to resolve action download info. Error: Service Unavailable` before any build step runs — GitHub Actions infrastructure, not this PR.

---

dormancygrace:
Retargeted this PR from the now-stale master branch to main and rebased it without content changes. This also starts fresh CI runs; the prior red jobs had failed in GitHub Actions at the Set up job step before any build command ran.

openwrt-ai (commented):
Re-reviewed the new head `055f7e0` against the previously reviewed `85338ce`. The two SHAs differ only because the series was rebased from the old `master` base (`712bfdd`) onto `main` (`71f139a`), as described in the comment — `git range-diff 712bfdd..85338ce 71f139a..055f7e0` reports both commits as `=`, i.e. identical patches and byte-identical commit messages. No content changed in this PR.

I checked the 47 base commits picked up by the rebase for anything that would interact with this device. The only two touching files this PR cares about are `12a5e56` (generic TCP fraglist GRO patch drop) and `224a967` "ramips: mt7530: compile swconfig code conditionally". The latter is benign here: it only gates the swconfig code path on `CONFIG_SWCONFIG`, which stays enabled for `ramips/mt7620`, so the `dlink,dwr-921-c3-uboot` `02_network` swconfig setup and the `ucidef_set_led_switch "lan" ... "switch0" "0x0f"` line in `01_leds` keep working unchanged. Nothing in the rebase range touches `mt7620.mk`, the ramips DTS directory, or `uboot-envtools/files/ramips`.

Spot-checked that the rebase didn't break the build wiring: `DEVICE_DTS` defaults to `$$(SOC)_$(1)` in `target/linux/ramips/image/Makefile`, so `dlink_dwr-921-c3-uboot` resolves to `mt7620n_dlink_dwr-921-c3-uboot.dts`, which is the file added here, and `IMAGE_SIZE := 16064k` still matches the `0xfb0000` firmware partition.

All conclusions from the previous rounds carry over unchanged: the `04_led_migration` case matches only `dlink,dwr-921-c1`, the `01_leds` split leaves the DWR-922 E2 on its own legacy names, and the `function = "<func>"` fallbacks reproduce the previous generated LED names apart from the intended `wifi` → `green:wlan`.

CI on `055f7e0` reports no check runs yet (pending), so there is nothing to ground there.

---

PR #24545, opened 2026-08-02
ramips: add broken-flash-reset for youku yk-l1c

roger225:
The Winbond W25Q256 on the YK-L1c does not handle the SPI NOR software reset
command correctly, causing the system to hang on reboot until power-cycled.

The sibling YK-L1 already has this property set.

### Without `broken-flash-reset`

After sysupgrade, the system hangs after `reboot: Restarting system` and
never reaches the bootloader. A physical power cycle is required.

```
Mon Jun 29 13:34:46 UTC 2026 upgrade: Upgrade completed
Mon Jun 29 13:34:47 UTC 2026 upgrade: Rebooting system...
umount: can't unmount /dev: Resource busy
[  921.672571] reboot: Restarting system
<-- hangs here, requires power cycle -->
```

### With `broken-flash-reset`

The system reboots normally and the bootloader loads the firmware from flash:

```
Mon Jun 29 20:43:24 UTC 2026 upgrade: Upgrade completed
Mon Jun 29 20:43:25 UTC 2026 upgrade: Rebooting system...
umount: can't unmount /dev: Resource busy
[ 7655.514929] reboot: Restarting system

Boot and Recovery Environment for Embedded Devices
...
Flash: Winbond W25Q256 (32MB) on rt2880-spi
...
Starting kernel at 0x80000000...
```

### Tested on

- **Device**: Youku YK-L1c
- **SoC**: MediaTek MT7620A ver 2, eco 6
- **Flash**: Winbond W25Q256 (32MB)
- **Kernel**: 6.12.94
- **OpenWrt**: v25.12.5

PR #22952, opened 2026-04-16
ath79: improve supported_devices lists referencing

aparcar:
Several ath79 device profiles include another profile's primary DTS compatible string in their SUPPORTED_DEVICES list. This causes ASU (Attended Sysupgrade) to build a wrong mapping, since its flat supported_device-to-profile dictionary gets overwritten by whichever profile is processed last.

Remove cross-references where a profile's SUPPORTED_DEVICES += entry matches another profile's auto-derived primary identifier (the device name with underscores converted to commas). Shared legacy compatibility strings that are not another profile's primary are kept.

Affected devices:
- EnGenius ESR1200/ESR1750: mutual cross-references between devices with different SoCs (qca9557 vs qca9558), which could also cause sysupgrade to accept an incompatible image.
- GL.iNet GL-AR300M NOR/NAND: mutual cross-references between flash variants.
- GL.iNet GL-AR750S NOR/NOR-NAND: same pattern as GL-AR300M.
- Ubiquiti UniFi AP LR: referenced ubnt,unifi-ap (the AP's primary).
- Ubiquiti UniFi AC LR: referenced ubnt,unifiac-lite (the AC Lite's primary).

This is the same class of bug fixed for the GL-MT2500 in commit b71f4665cd ("mediatek: filogic: fix supported_devices list for gl-mt2500").

Fixes: https://github.com/openwrt/asu/issues/1525

PR #24562, opened 2026-08-03
ramips: allow Youku L1 series images to flash on either model

mcprat:
These boards are shown to be identical hardware, but ever since commit 4a9f389ed2dce they have been treated as separate devices due to differing flash size.

Some of these boards are actually a L1 when released with an enclosure with the L1c label, leading people to flash the wrong factory.bin file and have the 4-byte flash mode reboot issue.

To help mitigate the problem, allow the board to accept both images instead of give a warning so users can go between the two DTS easily.

relevant to #22952 
better fix than #24545

### DIFF
--- a/target/linux/ramips/image/mt7620.mk
+++ b/target/linux/ramips/image/mt7620.mk
@@ -1465,7 +1465,7 @@ define Device/youku_yk-l1
   DEVICE_MODEL := YK-L1
   DEVICE_PACKAGES := kmod-usb2 kmod-usb-ohci kmod-mmc-mtk \
 	kmod-usb-ledtrig-usbport
-  SUPPORTED_DEVICES += youku-yk1 youku,yk1
+  SUPPORTED_DEVICES += youku-yk1 youku,yk1 youku,yk-l1c
   UIMAGE_MAGIC := 0x12291000
   UIMAGE_NAME := 400000000000000000000000
 endef
@@ -1478,6 +1478,7 @@ define Device/youku_yk-l1c
   DEVICE_MODEL := YK-L1c
   DEVICE_PACKAGES := kmod-usb2 kmod-usb-ohci kmod-mmc-mtk \
 	kmod-usb-ledtrig-usbport
+  SUPPORTED_DEVICES += youku-yk1 youku,yk1 youku,yk-l1
   UIMAGE_MAGIC := 0x12291000
   UIMAGE_NAME := 400000000000000000000000
 endef


target/linux/ramips/image/mt7620.mk
```diff
@@ -1478,6 +1478,7 @@ define Device/youku_yk-l1c
   DEVICE_MODEL := YK-L1c
   DEVICE_PACKAGES := kmod-usb2 kmod-usb-ohci kmod-mmc-mtk \
 	kmod-usb-ledtrig-usbport
+  SUPPORTED_DEVICES += youku-yk1 youku,yk1 youku,yk-l1
```
openwrt-ai, line 1481: This direction removes the sysupgrade warning for exactly the case the commit message says it wants to mitigate. `youku_yk-l1c` deliberately has no `broken-flash-reset`, because a 16 MiB chip never enters 4-byte addressing; `youku_yk-l1` carries it at [`mt7620a_youku_yk-l1.dts:10-12`](https://github.com/openwrt/openwrt/blob/d1c295da6dbd322a110246ff6f07fd702ae09da5/target/linux/ramips/dts/mt7620a_youku_yk-l1.dts#L10-L12). With this line, a board that really has 32 MiB flash (including the mislabelled "L1c" units the commit message is about) can be sysupgraded to the L1c image with no `-F`, boot a DTS without `broken-flash-reset`, and so skip the `spi_nor_restore()` 4-byte-exit at [`spi-nor/core.c:3378-3380`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/mtd/spi-nor/core.c#L3378-L3380) — i.e. the reboot hang comes back.

`youku-yk1`/`youku,yk1` make this sharper: those legacy names only ever belonged to the pre-split `youku_yk1` device, which was the 32 MiB YK-L1 (see [4a9f389](https://github.com/openwrt/openwrt/commit/4a9f389ed2dcee18a5c5e1f0d4e5c406f9290579)). Listing them here means "any old YK1, which is by definition 32 MiB, may take the 16 MiB image silently".

If the goal is genuinely interchangeable images, moving `broken-flash-reset` from the L1 `.dts` into the shared `&flash0` node in [`mt7620a_youku_yk-l1.dtsi:63`](https://github.com/openwrt/openwrt/blob/d1c295da6dbd322a110246ff6f07fd702ae09da5/target/linux/ramips/dts/mt7620a_youku_yk-l1.dtsi#L63) would make the L1c image safe on 32 MiB parts (it is a no-op on a chip that never leaves 3-byte mode), and then this line is defensible. Without that, is opening this direction — and in particular the legacy `youku-yk1`/`youku,yk1` aliases — intentional?

---

target/linux/ramips/image/mt7620.mk
```diff
@@ -1465,7 +1465,7 @@ define Device/youku_yk-l1
   DEVICE_MODEL := YK-L1
   DEVICE_PACKAGES := kmod-usb2 kmod-usb-ohci kmod-mmc-mtk \
 	kmod-usb-ledtrig-usbport
-  SUPPORTED_DEVICES += youku-yk1 youku,yk1
+  SUPPORTED_DEVICES += youku-yk1 youku,yk1 youku,yk-l1c
```
openwrt-ai, line 1468: nit: the commit message says "These boards are shown to be identical hardware", then in the next clause attributes the split to "differing flash size" — and [4a9f389](https://github.com/openwrt/openwrt/commit/4a9f389ed2dcee18a5c5e1f0d4e5c406f9290579) states the flash is 32 MiB on YK-L1 vs 16 MiB on YK-L1c, which is why the two `.dts` set different `&firmware` `reg` values. The two statements read as contradictory; "identical apart from flash size, and some units are mislabelled" would match the tree.

Worth stating explicitly in the message what happens in this direction, since it is the one this line enables: a 32448k-layout L1 image on a genuine 16 MiB L1c boots only because `mtdpart` truncates the oversized `firmware` partition ("extends beyond the end of device -- size truncated"), which silently shrinks `rootfs_data` rather than failing. That is workable but it is a behaviour reviewers should not have to derive.

---

openwrt-ai (commented):
Reviewed 1 commit; 2 inline comments. Main concern is the `youku_yk-l1c` side of the change re-opening the un-forced path that drops `broken-flash-reset` on 32 MiB boards.

---

brada4:
If they are same should not hey be collapsed to alternate names for base device cf ASUS RT-AX54 ( supported devices still needed to get "wrong side" right over next sysupgrade)

mcprat:
flash partitions are a critical part of DTS despite having no effect on software or firmware differences between the two

another way to say this is, the kernel and rootfs are identical but the DTS must be different, therefore the images must be different anyway, not just with a different name

brada4:
Ok, the: compat should be one-way - small image goes anywhere, but not the big brick.
Is there a (wiki-usable) way to differ pre-install?

mcprat:
DTS is a binary object in the middle of the image between kernel and rootfs, so that would be like reverse engineering an image to switch parts out

PR #24563, opened 2026-08-03
libtool: update to 2.6.2

graysky2:
Update to both tools and libs

Release notes: https://github.com/autotools-mirror/libtool/blob/master/NEWS#L6-L116


tools/libtool/Makefile
```diff
@@ -8,11 +8,11 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libtool
 PKG_CPE_ID:=cpe:/a:gnu:libtool
-PKG_VERSION:=2.5.4
+PKG_VERSION:=2.6.2
```
openwrt-ai, line 11: The 7 patches in `tools/libtool/patches/` were not refreshed for 2.6.2. They all still apply (no fuzz, no rejects — I unpacked the 2.6.2 tarball and ran the series), but every one of them now applies at a large offset:

```
000-relocatable.patch    libtoolize.in    hunk 2 offset  +27
000-relocatable.patch    m4/libtool.m4    hunk 1 offset +119
                                          hunks 2-3 offset +406
001-always-install-aux   libtoolize.in    hunk 1 offset  +27
100-libdir-fixes         ltmain.in        hunks 1-3 offset +150
110-dont-use-target-dir  ltmain.in        hunk 1 offset +150
120-strip-unsafe-dirs    ltmain.in        hunk 1 offset  +73
140-don-t-quote-SHELL    Makefile.am      hunks 2-6 offset +10/+16
```

A 406-line offset on `m4/libtool.m4` means the hunk headers no longer describe where the code lives; the next bump is much more likely to land a hunk in the wrong place silently. Please regenerate the series against 2.6.2 so the hunk headers match the new context (the `tools/` equivalent of the `make package/<pkg>/refresh` quilt flow — patches here are quilt-managed, not `git format-patch`-generated).

Everything else checked out: both `PKG_HASH` values match the GNU release tarballs byte-for-byte (`libtool-2.6.2.tar.gz` = `24adb3aa…8b0bf`, `libtool-2.6.2.tar.xz` = `2ef1067c…9e191e`), `bootstrap.conf` is unchanged apart from the copyright year so the pinned gnulib in `tools/gnulib` still covers it, and libltdl's `-version-info` moved 10:3:3 → 10:6:3, i.e. soname stays `libltdl.so.7`, so `ABI_VERSION:=7` in `package/libs/libtool/Makefile` remains correct.

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the diff (both `tools/libtool` and `package/libs/libtool` are bumped, as the body states), and libtool 2.6.2 is a real stable release (tagged 2026-07-16 upstream). One inline note about the unrefreshed patch series in `tools/libtool/patches/`.

---

PR #24564, opened 2026-08-03
gettext-full: update to 1.0

graysky2:
0.24.2 was several releases and over a year out of date and required patching to circumvent a build failure.

Makefile needed to be modifed since autogen.sh never runs libtoolize, so aclocal.m4 picks up the newer LT_INIT macros from the shared host libtool package while ltmain.sh while stays at the version bundled in the release tarball, causing a libtool version-mismatch abort at build time.

Patches were manually rebeased.

Changes between these releases can be found here:
https://github.com/autotools-mirror/gettext/blob/master/NEWS#L9-L178

### DIFF
--- a/package/libs/gettext-full/Makefile
+++ b/package/libs/gettext-full/Makefile
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=gettext-full
-PKG_VERSION:=0.24.2
+PKG_VERSION:=1.0
 PKG_RELEASE:=1
 
 PKG_SOURCE:=gettext-$(PKG_VERSION).tar.xz
 PKG_SOURCE_URL:=@GNU/gettext
-PKG_HASH:=fcc0187f597aef6bc5bc95c629db1126315beb196b20570eaec6a4941850f7c5
+PKG_HASH:=71132a3fb71e68245b8f2ac4e9e97137d3e5c02f415636eb508ae607bc01add7
 PKG_BUILD_DIR:=$(BUILD_DIR)/gettext-$(PKG_VERSION)
 HOST_BUILD_DIR:=$(BUILD_DIR_HOST)/gettext-$(PKG_VERSION)
 
@@ -43,11 +43,12 @@ PKG_SUBDIRS:= \
 		libgrep \
 		libtextstyle \
 		m4 \
-		misc \
 		po \
 		projects \
 		src \
 		styles \
+		autotools \
+		wizard \
 		gettext-runtime \
 		gettext-tools \
 	,$$$$(wildcard $$$$(dir)) )
@@ -87,7 +88,7 @@ CONFIGURE_ARGS += \
 	--disable-java \
 	--disable-openmp \
 	--disable-curses \
-	--with-included-gettext \
+	--with-included-libintl \
 	--without-libintl-prefix \
 	--without-libexpat-prefix \
 	--with-libunistring-prefix=$(STAGING_DIR)/usr \
@@ -114,6 +115,9 @@ HOST_CFLAGS += $(HOST_FPIC)
 define Host/Bootstrap
 	( \
 		cd $(HOST_BUILD_DIR); \
+		for dir in $$$$(find . -maxdepth 3 \( -name configure.ac -o -name configure.in \) -exec dirname {} \;); do \
+			(cd $$$$dir && $(AM_TOOL_PATHS) $(STAGING_DIR_HOST)/bin/libtoolize --copy --force) || exit 1; \
+		done && \
 		$(AM_TOOL_PATHS) \
 		./autogen.sh \
 	)
@@ -137,6 +141,9 @@ endef
 define Build/Bootstrap
 	( \
 		cd $(PKG_BUILD_DIR); \
+		for dir in $$$$(find . -maxdepth 3 \( -name configure.ac -o -name configure.in \) -exec dirname {} \;); do \
+			(cd $$$$dir && $(AM_TOOL_PATHS) $(STAGING_DIR_HOST)/bin/libtoolize --copy --force) || exit 1; \
+		done && \
 		$(AM_TOOL_PATHS) \
 		./autogen.sh \
 	)
--- a/package/libs/gettext-full/patches/000-relocatable.patch
+++ b/package/libs/gettext-full/patches/000-relocatable.patch
@@ -1,6 +1,6 @@
---- a/gettext-tools/misc/autopoint.in
-+++ b/gettext-tools/misc/autopoint.in
-@@ -27,7 +27,11 @@ archive_version=@ARCHIVE_VERSION@
+--- a/gettext-tools/autotools/autopoint.in
++++ b/gettext-tools/autotools/autopoint.in
+@@ -28,7 +28,11 @@ archive_version=@ARCHIVE_VERSION@
  
  # Set variables
  # - gettext_datadir     directory where the data files are stored.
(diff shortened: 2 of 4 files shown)


graysky2:
https://github.com/openwrt/openwrt/pull/24563 needs to be merged first. In order for this to build properly,

package/libs/gettext-full/Makefile
```diff
@@ -115,7 +115,10 @@ define Host/Bootstrap
 	( \
 		cd $(HOST_BUILD_DIR); \
 		$(AM_TOOL_PATHS) \
-		./autogen.sh \
+		./autogen.sh && \
+		for dir in $$$$(find . -maxdepth 3 \( -name configure.ac -o -name configure.in \) -exec dirname {} \;); do \
+			(cd $$$$dir && $(STAGING_DIR_HOST)/bin/libtoolize --copy --force); \
+		done \
```
openwrt-ai: This libtoolize sweep is what breaks the host build in CI — both `Build malta/be` and `Build x86/64` fail here:

```
OpenWrt-libtoolize: Remember to add 'LT_INIT' to configure.ac.
OpenWrt-libtoolize: Consider adding 'AC_CONFIG_MACRO_DIRS([m4])' to configure.ac,
OpenWrt-libtoolize: and rerunning libtoolize and aclocal.
configure: error: cannot find required auxiliary files: missing
make[3]: *** [Makefile:195: .../build_dir/hostpkg/gettext-1.0/.configured] Error 1
```

The `find` also matches the top-level `configure.ac`, which is not a libtool project ("Remember to add 'LT_INIT'"), and running `libtoolize --force` there rewrites the aux/macro setup that `autogen.sh` had just produced, so `configure` no longer finds `missing` in the aux dir. Restricting the loop to the subprojects that actually use `LT_INIT` (`gettext-runtime`, `gettext-tools`, `libtextstyle`), or re-running automake/autoreconf after libtoolize so the auxiliary files are put back, would avoid this.

One caveat on attribution: CI ran against main, i.e. with `tools/libtool` still at 2.5.4, whereas the mismatch this workaround targets presumably only appears with #24563 (libtool 2.6.2). If that is the case, this hunk can never be validated on its own and merging this PR before #24563 leaves the tree broken in between — it would be safer to send both patches as one series, or at least to state the ordering requirement in the commit message rather than only in a PR comment.

Same block is duplicated in `Build/Bootstrap` at lines 145-147.

---

package/libs/gettext-full/Makefile
```diff
@@ -8,12 +8,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=gettext-full
-PKG_VERSION:=0.24.2
+PKG_VERSION:=1.0
```
openwrt-ai, line 11: `PKG_SUBDIRS` looks stale for 1.0. As the rebase of `000-relocatable.patch` in this PR shows, `gettext-tools/misc/autopoint.in` moved to `gettext-tools/autotools/` and `gettext-tools/misc/gettextize.in` to `gettext-tools/wizard/`, but [`PKG_SUBDIRS`](https://github.com/openwrt/openwrt/blob/5f537f08705fca598ec2579cf0507099bfdee208/package/libs/gettext-full/Makefile#L32-L52) still lists only `misc` and neither of the new directories.

Every entry is wrapped in `$(wildcard ...)`, so `misc` now silently expands to nothing instead of erroring, and since the compile/install steps pass `SUBDIRS='$(PKG_SUBDIRS)'` ([L137](https://github.com/openwrt/openwrt/blob/5f537f08705fca598ec2579cf0507099bfdee208/package/libs/gettext-full/Makefile#L137), [L180](https://github.com/openwrt/openwrt/blob/5f537f08705fca598ec2579cf0507099bfdee208/package/libs/gettext-full/Makefile#L180)) the `autotools/` and `wizard/` directories are never built or installed — `autopoint` and `gettextize` would disappear from `$(STAGING_DIR_HOSTPKG)/bin`. Both are consumed in-tree: [`$(SED) '/read dummy/d' $(STAGING_DIR_HOSTPKG)/bin/gettextize`](https://github.com/openwrt/openwrt/blob/5f537f08705fca598ec2579cf0507099bfdee208/package/libs/gettext-full/Makefile#L176) in `Build/InstallDev`, and [`$(STAGING_DIR_HOSTPKG)/bin/autopoint --force` in include/autotools.mk:103](https://github.com/openwrt/openwrt/blob/5f537f08705fca598ec2579cf0507099bfdee208/include/autotools.mk#L103).

Unless 1.0 installs those two scripts from somewhere else, `autotools` and `wizard` need to be added to `PKG_SUBDIRS` (and `misc` dropped). The CI build dies before host-configure completes, so this would only surface once the bootstrap issue above is fixed.

---

package/libs/gettext-full/Makefile
```diff
@@ -138,7 +141,10 @@ define Build/Bootstrap
 	( \
 		cd $(PKG_BUILD_DIR); \
 		$(AM_TOOL_PATHS) \
-		./autogen.sh \
+		./autogen.sh && \
+		for dir in $$$$(find . -maxdepth 3 \( -name configure.ac -o -name configure.in \) -exec dirname {} \;); do \
+			(cd $$$$dir && $(STAGING_DIR_HOST)/bin/libtoolize --copy --force); \
+		done \
```
openwrt-ai: nit: two robustness details in this loop (applies equally to `Host/Bootstrap` at lines 119-121):

1. `$(AM_TOOL_PATHS)` is a shell env-assignment prefix, so it now only applies to `./autogen.sh`; everything after the `&&` is a separate command and runs without `ACLOCAL`/`AUTOCONF`/`AUTOM4TE`/`M4` pointing into `staging_dir/host`. Elsewhere in the tree autotools helpers are always invoked with that prefix in place ([include/autotools.mk#L45-L52](https://github.com/openwrt/openwrt/blob/5f537f08705fca598ec2579cf0507099bfdee208/include/autotools.mk#L45-L52), [tools/libtool/Makefile#L31](https://github.com/openwrt/openwrt/blob/5f537f08705fca598ec2579cf0507099bfdee208/tools/libtool/Makefile#L31)). Prefixing the libtoolize call with `$(AM_TOOL_PATHS)` too would keep it host-independent.

2. The `for` loop's exit status is only that of the last iteration, so a libtoolize failure in any earlier directory is silently ignored and the build continues with a stale `ltmain.sh` — the exact condition this change is meant to prevent.

---

openwrt-ai (commented):
## Commit checks

- 5f537f0 "gettext-full: update to 1.0" — message matches the diff, but a few typos are worth fixing before merge: "modifed" → "modified", "rebeased" → "rebased", and "while ltmain.sh while stays at the version bundled" has a duplicated "while". The ordering dependency on #24563 is only mentioned in a PR comment; since PR comments do not survive into the tree, it would be better placed in the commit message.


---

package/libs/gettext-full/Makefile
graysky2: Fixed

package/libs/gettext-full/Makefile
graysky2: Fixed

package/libs/gettext-full/Makefile
graysky2: Fixed

package/libs/gettext-full/Makefile
```diff
@@ -87,7 +88,7 @@ CONFIGURE_ARGS += \
 	--disable-java \
 	--disable-openmp \
 	--disable-curses \
-	--with-included-gettext \
+	--with-included-libintl \
```
openwrt-ai, line 91: nit: this rename and the `PKG_SUBDIRS` rework (lines 45-51: `misc` dropped, `autotools`/`wizard` added) are undocumented — the commit message only explains the libtoolize/bootstrap change. A sentence for each would make them traceable later.

Separately, worth confirming the new spelling against 1.0's generated `configure`: autoconf only emits `configure: WARNING: unrecognized options: --with-included-libintl` for an unknown `--with-*` and then continues, so the now-green CI build does not prove the option exists. If it were misspelled, `libintl-full` would silently be configured without it rather than failing.

---

openwrt-ai (commented):
Re-reviewed at `a145417`. The three findings from the previous round are addressed: libtoolize now runs before `autogen.sh` (so `autogen.sh` restores the auxiliary files it needs), the loop is prefixed with `$(AM_TOOL_PATHS)` and guarded with `|| exit 1`, and `PKG_SUBDIRS` picks up `autotools`/`wizard`.

## Commit checks

- a145417 "gettext-full: update to 1.0" — two typos flagged last round are still in the message body: "modifed" → "modified", and the duplicated "while" in "libtool package while ltmain.sh while stays at the version bundled". "rebeased" → "rebased" and the #24563 ordering note were fixed.

---

PR #24567, opened 2026-08-04
mediatek: filogic: add 256MB variants of Cudy AP3000/AP3000 Outdoor v1

### LINKED ISSUE #24565: Cudy AP3000 / AP3000 outdoor only use 64MB of the 256MB flash
### Describe the bug

```
[    0.916390] spi-nand spi0.0: 256 MiB, block size: 128 KiB, page size: 2048, OOB size: 128
[    1.721066] ubi0: attached mtd5 (name "ubi", size 64 MiB)
```

https://github.com/openwrt/openwrt/blob/70b3f992b81462f5cc5ba24ca41aaadd0186e1a4/target/linux/mediatek/dts/mt7981b-cudy-ap3000-v1.dts#L134
https://github.com/openwrt/openwrt/blob/70b3f992b81462f5cc5ba24ca41aaadd0186e1a4/target/linux/mediatek/dts/mt7981b-cudy-ap3000-v1.dts#L210-L213

I understand that using the easy install (cudy transition firmware then vanilla OpenWrt) it might not be possible to resize, but maybe having a second image variant using the full 256MB installable via tftp could be an option ?

### OpenWrt version

r33065-ab08f6a524

### OpenWrt release

25.12-SNAPSHOT

### OpenWrt target/subtarget

mediatek/filogic

### Device

Cudy AP3000 v1

### Image kind

Self-built image

### Terms

unisequence:
Closes #24565.

Both boards are fitted with an ESMT F50L2G41XA SPI-NAND (256 MiB), but the existing device trees only expose a 64 MiB `ubi` partition, leaving ~186 MiB of flash unused.

Field units may vary in flash size (as has happened on other boards in this family), so instead of resizing the existing `ubi` partition in place -- which would risk bricking any unit that actually has a smaller chip on sysupgrade -- this splits each device tree into a shared `.dtsi` plus the existing default (64 MiB) `.dts` and a new `256mb-v1` variant with a full-size `ubi` partition (`0x5c0000`-`0x10000000`).

The `256mb-v1` profiles use a distinct compatible string and are only installable via TFTP/recovery, not offered as a sysupgrade target for existing 64 MiB units.

**Not tested on real hardware** -- I don't own either board. Verified only that both device trees still build (`dtc`, no new warnings beyond pre-existing ones from `mt7981b.dtsi`) and that the `ubi` partition in the `256mb-v1` dtb ends exactly at the flash boundary (`0x10000000`). Needs confirmation from an AP3000/AP3000 Outdoor v1 owner with a 256 MiB chip before merge.


brada4:
The originals  are kind of 128MB with great slack space too....

> Not tested on real hardware

You have to, OEM bootloader may disagree with new partition scheme....

unisequence:
@brada4 Fair points, thanks.

On the flash size variance — do you have specifics on which Cudy models/batches you've seen ship with 128MB despite spec'd higher? If AP3000/AP3000 Outdoor specifically have known 128MB batches, that's a strong argument this shouldn't be a silent default and maybe needs a way to detect/report the actual chip size rather than just offering a fixed 256mb profile.

On the bootloader concern — this board doesn't have a custom OpenWrt u-boot (no patch under `package/boot/uboot-mediatek/`), so I assumed the vendor bootloader's TFTP/recovery flash path is agnostic to the Linux-side partition layout described in this DTS. But if the vendor flasher itself validates image size/layout against something baked into its own recovery logic (not just raw NAND writes), that's a real bricking risk I hadn't accounted for — and I don't have the hardware to check.

If that's the case, this needs either confirmation from someone who's flashed a stock-bootloader Cudy board with a resized ubi image, or this has to go the `-ubootmod` route instead (replacing the bootloader with an OpenWrt one that doesn't care about vendor partition assumptions) — but I'm not going to write that blind without hardware to test against, that's a much higher-risk change than this one.

Parking this PR until someone with the hardware can confirm one way or the other.

Linaro1985:
> You have to, OEM bootloader may disagree with new partition scheme....

Agree. I already tried expanding the partition on Cudy, but it bricked the device because the stock bootloader wouldn't work with a partition of a different size.

@unisequence, please test this on a real hardware.

unisequence:
@Linaro1985 Thanks for actually testing this and reporting back — confirms the concern exactly. Closing this, the approach isn't safe as-is on stock bootloader.

brada4:
@unisequence 

> Fair slop

Cmon, have some self-respect, you just regurgitate my points at 100x volume.

PR #24574, opened 2026-08-04
ramips: mt7530: compile swconfig code conditionally

namiltd:
Wrap all swconfig-specific code in mt7530.c with `#if IS_ENABLED(CONFIG_SWCONFIG)`, including struct switch_dev, the vlan/port mapping tables and all switch_dev_ops callbacks. Register access helpers (mt7530_r32/w32), the PSC/HWTRAP definitions, and the small set of swconfig-only register macros and MIB tables further up the file remain unconditional - the latter two are unused-but-harmless when swconfig is disabled and are left alone to keep the diff focused.

When swconfig is not compiled in, the driver performs only minimal hardware init (disables MAC learning on all ports, applies the HWTRAP fixup) and skips switch registration, mirroring the existing behaviour of the `mediatek,no-swconfig` device tree property. This removes the need for that property when building DSA-only images. The property is still honoured whenever CONFIG_SWCONFIG is enabled (built-in or as a module).

A distinct log message is used for the compile-time case, to distinguish it from the runtime `mediatek,no-swconfig` case in dmesg. Also switch the existing runtime "swconfig disabled" message from pr_info() to dev_info(), changing its dmesg prefix from "mt7620: " / "mt7530: " to the device's own prefix.

Patch to prepare code to drop swconfig for entire subtarget.

Tested with TP-Link Archer C5 v4 after converting to DSA and disabling swconfig code:
```
[    1.447283] gsw: setting port4 to ephy mode
[    1.453431] mtk_soc_eth 10100000.ethernet: port 5 - using fixed link parameters
[    1.461052] mtk_soc_eth 10100000.ethernet: swconfig support not built in, MAC learning disabled on all ports
[    1.471117] mtk_soc_eth 10100000.ethernet: loaded mt7620 driver
[    1.477899] mtk_soc_eth 10100000.ethernet eth0: mediatek frame engine at 0xb0100000, irq 5
```

### DIFF
--- a/target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c
+++ b/target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c
@@ -160,6 +160,7 @@ enum {
 	MT7530_ATTR_ENABLE_VLAN,
 };
 
+#if IS_ENABLED(CONFIG_SWCONFIG)
 struct mt7530_port_entry {
 	u16	pvid;
 	bool	mirror_rx;
@@ -171,10 +172,12 @@ struct mt7530_vlan_entry {
 	u8	member;
 	u8	etags;
 };
+#endif /* CONFIG_SWCONFIG */
 
 struct mt7530_priv {
 	void __iomem		*base;
 	struct mii_bus		*bus;
+#if IS_ENABLED(CONFIG_SWCONFIG)
 	struct switch_dev	swdev;
 
 	u8			mirror_dest_port;
@@ -182,8 +185,10 @@ struct mt7530_priv {
 	struct mt7530_vlan_entry	vlan_entries[MT7530_NUM_VLANS];
 	struct mt7530_port_entry	port_entries[MT7530_NUM_PORTS];
 	char arl_buf[MT7530_NUM_ARL_RECORDS * ARL_LINE_LENGTH + 1];
+#endif /* CONFIG_SWCONFIG */
 };
 
+#if IS_ENABLED(CONFIG_SWCONFIG)
 struct mt7530_mapping {
 	char	*name;
 	u16	pvids[MT7530_NUM_PORTS];
@@ -285,6 +290,7 @@ mt7530_set_vlan_enable(struct switch_dev *dev,
 
 	return 0;
 }
+#endif /* CONFIG_SWCONFIG */
 
 static u32
 mt7530_r32(struct mt7530_priv *priv, u32 reg)
@@ -324,6 +330,7 @@ mt7530_w32(struct mt7530_priv *priv, u32 reg, u32 val)
 	iowrite32(val, priv->base + reg);
 }
 
+#if IS_ENABLED(CONFIG_SWCONFIG)
 static void
 mt7530_vtcr(struct mt7530_priv *priv, u32 cmd, u32 val)
 {
@@ -1003,15 +1010,20 @@ static const struct switch_dev_ops mt7530_ops = {
 	.apply_config = mt7530_apply_config,
 	.reset_switch = mt7530_reset_switch,
 };
+#endif /* CONFIG_SWCONFIG */
 
 int
 mt7530_probe(struct device *dev, void __iomem *base, struct mii_bus *bus, int vlan)
 {
-	struct switch_dev *swdev;
 	struct mt7530_priv *mt7530;
+	bool swconfig_active = false;
+	const char *name = bus ? "mt7530" : "mt7620";
+	int i;
+#if IS_ENABLED(CONFIG_SWCONFIG)
+	struct switch_dev *swdev;
 	struct mt7530_mapping *map;
 	int ret;
-	int i;
+#endif /* CONFIG_SWCONFIG */
 
 	mt7530 = devm_kzalloc(dev, sizeof(struct mt7530_priv), GFP_KERNEL);
 	if (!mt7530)
@@ -1019,16 +1031,12 @@ mt7530_probe(struct device *dev, void __iomem *base, struct mii_bus *bus, int vl
 
 	mt7530->base = base;
 	mt7530->bus = bus;
+#if IS_ENABLED(CONFIG_SWCONFIG)
 	mt7530->global_vlan_enable = vlan;
 
 	swdev = &mt7530->swdev;
-	if (bus) {
-		swdev->alias = "mt7530";
-		swdev->name = "mt7530";
-	} else {
-		swdev->alias = "mt7620";
-		swdev->name = "mt7620";
-	}
+	swdev->alias = name;
+	swdev->name = name;
 	swdev->cpu_port = MT7530_CPU_PORT;
 	swdev->ports = MT7530_NUM_PORTS;
 	swdev->vlans = MT7530_NUM_VLANS;
@@ -1045,9 +1053,15 @@ mt7530_probe(struct device *dev, void __iomem *base, struct mii_bus *bus, int vl
 		if (map)
 			mt7530_apply_mapping(mt7530, map);
 		mt7530_apply_config(swdev);
+		swconfig_active = true;
 	} else {
-		pr_info("%s: swconfig disabled, MAC learning disabled on all ports\n", swdev->name);
+		dev_info(dev, "swconfig disabled, MAC learning disabled on all ports\n");
+	}
+#else
+	dev_info(dev, "swconfig support not built in, MAC learning disabled on all ports\n");
+#endif /* CONFIG_SWCONFIG */
 
+	if (!swconfig_active) {
 		for (i = 0; i < MT7530_NUM_PORTS; i++) {
 			mt7530_w32(mt7530, REG_ESW_PORT_PSC(i),
 				mt7530_r32(mt7530, REG_ESW_PORT_PSC(i)) | REG_ESW_PORT_PSC_SA_DIS);
@@ -1059,7 +1073,7 @@ mt7530_probe(struct device *dev, void __iomem *base, struct mii_bus *bus, int vl
 		dev_info(dev, "fixing up MHWTRAP register - bootloader probably played with it\n");
 		mt7530_w32(mt7530, REG_HWTRAP, 0x1117edf);
 	}
-	dev_info(dev, "loaded %s driver\n", swdev->name);
+	dev_info(dev, "loaded %s driver\n", name);
 
 	return 0;
 }


target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c
```diff
@@ -83,18 +83,20 @@ enum {
 	ETAG_CTRL_STACK	= 3,
 };
 
+#ifdef CONFIG_SWCONFIG
```
openwrt-ai: `SWCONFIG` is a **tristate**, not a bool — see [`config SWCONFIG` / `tristate "Switch configuration API"` in 700-swconfig_switch_drivers.patch:21-26](``https://github.com/openwrt/openwrt/blob/4c741a539e2483a40a85f3c74a9640b704ef2908/target/linux/generic/hack-6.18/700-swconfig_switch_drivers.patch#L21-L26).`` With `SWCONFIG=m` the kernel defines `CONFIG_SWCONFIG_MODULE` and *not* `CONFIG_SWCONFIG`, so `#ifdef CONFIG_SWCONFIG` is false and the driver silently compiles out all switch registration even though swconfig is present — a silent loss of functionality rather than a link error. `IS_ENABLED()` is the kernel idiom that covers both `=y` and `=m`.

All ramips subtargets currently ship `CONFIG_SWCONFIG=y` (e.g. `target/linux/ramips/mt7620/config-6.18:208`), so this is latent rather than a live regression, but since the whole point of the patch is to make the non-`y` cases build correctly it seems worth handling `=m` too.

```suggestion
#if IS_ENABLED(CONFIG_SWCONFIG)
```

Same applies to the other four guards introduced here: lines 165, 182, 193 and 336.

(Note: [`rtl8367s_mdio.c:262`](https://github.com/openwrt/openwrt/blob/4c741a539e2483a40a85f3c74a9640b704ef2908/target/linux/mediatek/files/drivers/net/phy/rtk/rtl8367s_mdio.c#L262) uses plain `#ifdef CONFIG_SWCONFIG`, so there is in-tree precedent for the current form — if that is deliberate, ignore this.)

---

target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c
```diff
@@ -83,18 +83,20 @@ enum {
...
-
 #define REG_ESW_PORT_PCR_MIRROR_SRC_RX_BIT	BIT(8)
 #define REG_ESW_PORT_PCR_MIRROR_SRC_TX_BIT	BIT(9)
 #define REG_ESW_PORT_PCR_MIRROR_SRC_RX_MASK	0x0100
 #define REG_ESW_PORT_PCR_MIRROR_SRC_TX_MASK	0x0200
+#endif /* CONFIG_SWCONFIG */
 
+/* used unconditionally in mt7530_probe() regardless of CONFIG_SWCONFIG */
+#define REG_ESW_PORT_PSC(x)	(0x200C | ((x) << 8))
+#define REG_ESW_PORT_PSC_SA_DIS	BIT(4)
```
openwrt-ai: nit: this hunk exists only because the macro block above got wrapped, but wrapping macros buys nothing — an unreferenced `#define` never produces a diagnostic. And the wrapping is applied to only 4 of the file's swconfig-only definitions; these are all equally swconfig-only and stay unconditional:

- [`REG_ESW_WT_MAC_MFC` / `REG_ESW_VLAN_*` / `REG_ESW_TABLE_*` at lines 50-76](``https://github.com/openwrt/openwrt/blob/4c741a539e2483a40a85f3c74a9640b704ef2908/target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c#L50-L76``)
- [`MIB_DESC` + `mt7620_mibs` / `mt7620_port_mibs` at lines 102-157](``https://github.com/openwrt/openwrt/blob/4c741a539e2483a40a85f3c74a9640b704ef2908/target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c#L102-L157``) — these are `static const` arrays, so unlike macros they *are* dead data with `SWCONFIG=n` (harmless in a normal build since the kernel passes `-Wno-unused-const-variable`, but they do warn under `W=1`)
- [the `MT7530_ATTR_ENABLE_VLAN` enum at lines 160-163](``https://github.com/openwrt/openwrt/blob/4c741a539e2483a40a85f3c74a9640b704ef2908/target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c#L160-L163``)

So the commit message's "Wrap **all** swconfig-specific code in mt7530.c" doesn't quite hold. Either extend the guard to the MIB tables/enum, or drop the `#ifdef` around the macro block entirely — the latter would let `REG_ESW_PORT_PSC`/`REG_ESW_PORT_PSC_SA_DIS` stay in place and remove this whole hunk.

---

target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c
```diff
@@ -1019,47 +1027,60 @@ mt7530_probe(struct device *dev, void __iomem *base, struct mii_bus *bus, int vl
...
-	if (!of_property_present(dev->of_node, "mediatek,no-swconfig")) {
-		ret = register_switch(swdev, NULL);
-		if (ret) {
-			dev_err(dev, "failed to register %s\n", swdev->name);
-			return ret;
+#ifdef CONFIG_SWCONFIG
+	{
+		struct switch_dev *swdev = &mt7530->swdev;
+		struct mt7530_mapping *map;
+		int ret;
```
openwrt-ai: The anonymous `{ }` block re-indents the entire existing probe body by one tab, which is what makes this a 56/35 diff for what is conceptually "guard three declarations". Nothing in this file (or in the neighbouring `esw_rt3050.c` / `gsw_mt7620.c`) uses a bare nested block for scoping; declarations live at function top. Guarding just the declarations keeps the diff to the `#ifdef`/`#endif` lines and leaves the rest of the function byte-identical, which also makes the patch much easier to verify:

```c
	struct mt7530_priv *mt7530;
#ifdef CONFIG_SWCONFIG
	struct switch_dev *swdev;
	struct mt7530_mapping *map;
	int ret;
#endif
	int i;
```

(`swdev` then keeps its original `swdev = &mt7530->swdev;` assignment below.)

---

target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c
```diff
@@ -1019,47 +1027,60 @@ mt7530_probe(struct device *dev, void __iomem *base, struct mii_bus *bus, int vl
...
+					mt7530_r32(mt7530, REG_ESW_PORT_PSC(i)) | REG_ESW_PORT_PSC_SA_DIS);
 		}
 	}
+#else
+	dev_info(dev, "swconfig support not built in, MAC learning disabled on all ports\n");
+
+	for (i = 0; i < MT7530_NUM_PORTS; i++)
+		mt7530_w32(mt7530, REG_ESW_PORT_PSC(i),
+			mt7530_r32(mt7530, REG_ESW_PORT_PSC(i)) | REG_ESW_PORT_PSC_SA_DIS);
+#endif /* CONFIG_SWCONFIG */
```
openwrt-ai, line 1062: The MAC-learning-disable loop is now duplicated verbatim between the `#else` arm and the runtime `mediatek,no-swconfig` arm at lines 1065-1067. Two copies of a register-write loop that must stay in sync is exactly the kind of thing that drifts on the next edit.

Since the two paths differ only in the log message, the loop can be emitted once — e.g. have the `CONFIG_SWCONFIG` arm `return 0` (or jump past) after the successful `register_switch()` path, and let both "swconfig not active" cases fall through to a single copy of the loop after the `#endif`. That also keeps the `bus &&` HWTRAP fixup and the "loaded %s driver" line reached on every path exactly as they are today.

---

target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c
```diff
@@ -1019,47 +1027,60 @@ mt7530_probe(struct device *dev, void __iomem *base, struct mii_bus *bus, int vl
...
-			mt7530_apply_mapping(mt7530, map);
-		mt7530_apply_config(swdev);
-	} else {
-		pr_info("%s: swconfig disabled, MAC learning disabled on all ports\n", swdev->name);
+			map = mt7530_find_mapping(dev->of_node);
+			if (map)
+				mt7530_apply_mapping(mt7530, map);
+			mt7530_apply_config(swdev);
+		} else {
+			dev_info(dev, "swconfig disabled, MAC learning disabled on all ports\n");
```
openwrt-ai: nit: this converts the *existing* runtime `mediatek,no-swconfig` message from `pr_info("%s: swconfig disabled...", swdev->name)` to `dev_info()`, changing its dmesg prefix from `mt7620: ` / `mt7530: ` to `mtk_soc_eth 10100000.ethernet: `. The change is an improvement, but the commit message only claims "A distinct log message is used for the compile-time case" — it doesn't mention touching the existing message. Worth a line in the commit body so someone bisecting a missing `mt7620: swconfig disabled` grep in a test script knows where it went.

---

openwrt-ai (commented):
Reviewed 1 new commit. Main point: `CONFIG_SWCONFIG` is a tristate, so `#ifdef` misses the `=m` case — details inline, along with some scope/duplication notes on the probe restructure.

---

target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c
```diff
@@ -94,7 +94,6 @@ enum {
 #define REG_ESW_PORT_PCR_MIRROR_SRC_TX_BIT	BIT(9)
 #define REG_ESW_PORT_PCR_MIRROR_SRC_RX_MASK	0x0100
 #define REG_ESW_PORT_PCR_MIRROR_SRC_TX_MASK	0x0200
-
 #define REG_HWTRAP		0x7804
```
openwrt-ai, line 98: nit: this hunk deletes the blank line that separated the `REG_ESW_PORT_PCR_MIRROR_*` block from `REG_HWTRAP` — leftover from the previous revision, where the `#endif` lived here. Nothing else in the commit touches this region any more, so it is now a pure whitespace change unrelated to the swconfig guarding. Restoring it keeps the diff to one file-scope concern.

```suggestion

#define REG_HWTRAP		0x7804
```

---

target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c
```diff
@@ -1045,21 +1057,28 @@ mt7530_probe(struct device *dev, void __iomem *base, struct mii_bus *bus, int vl
...
-		pr_info("%s: swconfig disabled, MAC learning disabled on all ports\n", swdev->name);
+		dev_info(dev, "swconfig disabled, MAC learning disabled on all ports\n");
+	}
+#else
+	dev_info(dev, "swconfig support not built in, MAC learning disabled on all ports\n");
+#endif /* CONFIG_SWCONFIG */
 
+#if IS_ENABLED(CONFIG_SWCONFIG)
+	if (!swconfig_active)
+#endif /* CONFIG_SWCONFIG */
```
openwrt-ai: Splitting an `if` from its body across a preprocessor conditional is fragile: with `SWCONFIG=n` the `for` at 1071-1074 is still indented two tabs as if it were nested, so the visual structure no longer matches the real structure, and any future edit that adds a second statement to the loop's "guarded" scope silently changes meaning between the two configurations. `Documentation/process/coding-style.rst` §21 pushes exactly the other way — keep `#ifdef` out of function bodies where a plain conditional works.

Since `swconfig_active` is only ever *read* here, declaring it unconditionally makes the `if` unconditional and the preprocessor disappears from this statement; with `SWCONFIG=n` it constant-folds to `if (!false)` and gcc emits nothing extra (and no `-Wunused-but-set-variable`, since it is used):

```c
	struct mt7530_priv *mt7530;
	bool swconfig_active = false;
	int i;
#if IS_ENABLED(CONFIG_SWCONFIG)
	struct switch_dev *swdev;
	struct mt7530_mapping *map;
	int ret;
#endif /* CONFIG_SWCONFIG */
```

then here simply:

```c
	if (!swconfig_active)
		for (i = 0; i < MT7530_NUM_PORTS; i++) {
			...
		}
```

No `suggestion` block because the fix also has to move the declaration out of the `#if` at line 1024, which is a separate hunk.

---

target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c
```diff
@@ -1045,21 +1057,28 @@ mt7530_probe(struct device *dev, void __iomem *base, struct mii_bus *bus, int vl
...
 		}
-	}
 
 	/* magic vodoo */
 	if (bus && mt7530_r32(mt7530, REG_HWTRAP) !=  0x1117edf) {
 		dev_info(dev, "fixing up MHWTRAP register - bootloader probably played with it\n");
 		mt7530_w32(mt7530, REG_HWTRAP, 0x1117edf);
 	}
-	dev_info(dev, "loaded %s driver\n", swdev->name);
+	dev_info(dev, "loaded %s driver\n", bus ? "mt7530" : "mt7620");
```
openwrt-ai: nit: this replaces `swdev->name` with a second copy of the `bus ? "mt7530" : "mt7620"` mapping that already exists at [lines 1037-1043](``https://github.com/openwrt/openwrt/blob/0a4ad5a3e9d3ecb98e7bd7f7d76121026a8004eb/target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c#L1037-L1043).`` The output is unchanged, so this is only about keeping the two in sync — hoisting a `const char *name = bus ? "mt7530" : "mt7620";` above the `#if` and using it for `swdev->alias`/`swdev->name` as well would leave a single source of truth.

Either way, the commit body enumerates the other two logging changes (`pr_info()` -> `dev_info()`, the new compile-time message) but not this one; worth a half-line so the change isn't invisible in the log.

---

openwrt-ai (commented):
Reviewed the force-pushed commit. All five points from the previous round are addressed: the guards are now `IS_ENABLED()` (covers `SWCONFIG=m`), the macro-block wrapping is gone, the anonymous scope block in `mt7530_probe()` is gone, the MAC-learning loop exists once, and the commit body documents the `pr_info()` -> `dev_info()` change. I re-checked the `swconfig_active` refactor against the old control flow and it is behaviour-preserving in all three cases (`SWCONFIG=n`, `=y` with `mediatek,no-swconfig`, `=y` without it), and `mt7530.o`'s only link partner (`gsw_mt7620.o`, via `CONFIG_NET_RALINK_GSW_MT7620`) carries no swconfig references, so the object really does build clean without swconfig.

Remaining items are minor and inline: a stray whitespace-only hunk, the `#ifdef`-split `if` statement, and one undocumented log change.

---

openwrt-ai (commented):

All three points from the previous round are addressed: the whitespace-only hunk around `REG_HWTRAP` is gone, `swconfig_active` is now declared unconditionally so the `if` no longer straddles a `#if`/`#endif`, and the `bus ? "mt7530" : "mt7620"` mapping is hoisted into a single `name` variable used by `swdev->alias`/`swdev->name` and the "loaded %s driver" line.

Re-checked the guard boundaries in the amended file: everything referencing a `#if IS_ENABLED(CONFIG_SWCONFIG)`-guarded symbol (`swdev`, `map`, `ret`, `mt7530_ops`, `register_switch()`, `mt7530_find_mapping()`, `mt7530_apply_mapping()`, `mt7530_apply_config()`, `mt7530->global_vlan_enable`) stays inside a guarded region, and `mt7530_r32`/`mt7530_w32`/`REG_HWTRAP` remain unconditional as the commit body states. No out-of-file user of the now-guarded `struct mt7530_priv` members exists (`esw_rt3050.c`'s `global_vlan_enable` is a different struct). With `SWCONFIG=n` the three probe paths still collapse to the same register writes as before, and `swconfig_active` is read so it won't trip `-Wunused-but-set-variable`.

CI reports no checks on `6f62365` yet, so nothing to ground there.

---

namiltd:
PR finally passed all Claude AI tests.

jonasjelonek:

PR #20475, opened 2025-10-20
airoha: support openwrt,netdev-name for renaming interfaces

hurrian:
Add support to the airoha target for the OpenWrt-specific DT property `openwrt,netdev-name`.
In particular, this is for interfaces under non-DSA `airoha_eth` interfaces.

This will avoid conflicts with upstream code[1]; and maintain forward compatibility with OpenWrt configurations if/when `airoha_eth` becomes a full DSA driver.

[1] https://lore.kernel.org/netdev/20240709124503.pubki5nwjfbedhhy@skbuf/

Borrowed from d4d6c48 (mediatek: filogic: support openwrt,netdev-name for renaming interfaces)

Signed-off-by: Kenneth Kasilag <kenneth@kasilag.me>

PR #19534, opened 2025-07-26
mediatek: add support for CreatLentem CLT-R30B1

andros-ua:
```
CreatLentem CLT-R30B1 is a wireless WiFi 6 router.
This device uses the CLT-R30B1_0824_V1.1 board
shared by EDUP RT2980, Dragonglass DXG21,
and other diamond-shaped 5-antenna routers.

Specification
-------------
- SoC          : MediaTek MT7981B dual-core ARM Cortex-A53 1.3 GHz
- RAM          : DDR3 256 MiB
- Flash        : SPI-NAND 128 MiB (ESMT F50L1G41LB)
- WLAN         : MediaTek MT7976CN dual-band WiFi 6
  - 2.4 GHz    : b/g/n/ax, MU-MIMO (2x 5 dBi antennas)
  - 5 GHz      : a/n/ac/ax, MU-MIMO (3x 5 dBi antennas)
- Ethernet     :
  - LAN x3     : 10/100/1000 Mbps (MediaTek MT7531AE)
  - WAN x1     : 10/100/1000 Mbps (MT7981 internal PHY)
- UART         : through-hole on PCB
  - assignment : (RX), (TX), (GND), [3.3V]
  - settings   : 115200n8
- Buttons x2   : Mesh/WPS, Reset
- LEDs x2      : Status (Red, Green)
- Power        : 12 VDC, 1 A, 2.1*5.5 mm

Important notes
---------------
The device is supplied in two variants.
The main difference is the size of the mtd5 (ubi)
partition in the flash layout: 64M or 112M.
112M version: Has ImmortalWrt firmware installed with LuCI WebUI.
64M version: Has stock firmware based on OpenWrt,
with the WaveLink/GL.iNet WebUI and older U-Boot
compared to the 112M version.

Flash instructions for 112M version
-----------------------------------
Follow the standard OpenWrt sysupgrade procedure without saving data.
Use the clt-r30b1-112m-squashfs-sysupgrade.bin image.
All checks should pass - don't proceed if a "not supported"
warning is issued.

Flash instructions for 64M version
----------------------------------
WebUI Method:

1. Prepare the upgrade image with clt-r30b1-squashfs-sysupgrade.bin
   using the script: make_staged_upgrade_tar.sh
   or use the prepared image: staged_openwrt_upgrade.bin
   Downloaded from:
   https://github.com/andros-ua/owrt-misc/tree/main/clt-r30b1
2. Install the prepared image using the stock WebUI update page.
3. Press and hold the reset button after reboot
[...]

PR #24575, opened 2026-08-04
mediatek: cmcc rax3000m: add all-in-UBI layout

andros-ua:
```
This commit introduces OpenWrt U-Boot all-in-UBI layout support
for the CMCC RAX3000M / RAX3000Me, enabling:
- Prolonged device lifetime by allocating most of the flash
  to UBI (which takes care of wear-leveling)
- Maximum available storage space for OpenWrt

OpenWrt U-Boot UBI flash instructions
-------------------------------------
A device running stock firmware should be upgraded to the
latest OpenWrt firmware
(https://firmware-selector.openwrt.org/?target=mediatek%2Ffilogic&id=cmcc_rax3000m).

Back up critical data
---------------------
LuCI Web-UI:
"System" -> "Backup / Flash Firmware" -> "Save mtdblock contents"
Save mtdblock:
   factory

Make sure the file were successfully downloaded to your downloads directory.

Using the installer image
-------------------------
To simplify the installation process, this method uses a fork
of Daniel Golle's (@dangowrt) UBI Installer
https://github.com/dangowrt/owrt-ubi-installer

1. Ensure your router is running the latest OpenWrt firmware.
   Upgrade it if necessary.
2. Obtain the installer image:
   Build the installer from source according to your device variant
   https://github.com/andros-ua/owrt-ubi-installer/tree/rax3000m-ddr3
   or
   https://github.com/andros-ua/owrt-ubi-installer/tree/rax3000m-ddr4
   or download a prebuilt image from the
   https://github.com/andros-ua/owrt-ubi-installer/releases
3. Flash the openwrt*-initramfs-recovery-installer.itb
   image using sysupgrade.
4. Wait for installation:
   the green status LED will blink rapidly,
   indicating that the all-in-UBI installer is running.
   Once the installation finishes,
   the status LED will turn solid amber for 5 seconds.
5. After the device reboots, perform a final sysupgrade using the
   openwrt*-squashfs-sysupgrade.itb image.

BL2 and FIP Recovery
--------------------
Use mtk_uartboot to recover corrupted BL2 or FIP via UART:
https://github.com/981213/mtk_uartboot

OpenWrt U-Boot layout
----------------------------------------
| dev:    size   erasesize  name       |
| mtd4: 00100000 00020000 "bl2"        |
| mtd3: 00080000 00020000 "u-boot-env" |
| mtd2: 00200000 00020000 "factory"    |
| mtd1: 00200000 00020000 "fip"        |
| mtd0: 07200000 00020000 "ubi"        |
----------------------------------------

OpenWrt U-Boot UBI layout
----------------------------------
| dev:    size   erasesize  name |
| mtd1: 00100000 00020000 "bl2"  |
| mtd0: 07f00000 00020000 "ubi"  |
----------------------------------

Update MAC table after inspecting the factory partition:
----------------------------------------------------------------
| Interface    | MAC               | Source        |           |
|--------------|-------------------|---------------|-----------|
| WAN (label)  | 54:4d:d4:xx:xx:xx | factory, 0x24 | label     |
| LAN          | 54:4d:d4:xx:xx:xx | factory, 0x2a | label + 3 |
| WLAN 2.4 GHz | 54:4d:d4:xx:xx:xx | factory, 0x4  | label + 1 |
| WLAN 5 GHz   | 54:4d:d4:xx:xx:xx | factory, 0xa  | label + 2 |
----------------------------------------------------------------

```

### DIFF
--- a/package/boot/uboot-mediatek/Makefile
+++ b/package/boot/uboot-mediatek/Makefile
@@ -305,6 +305,30 @@ define U-Boot/mt7981_cmcc_rax3000m-nand-ddr4
   DEPENDS:=+trusted-firmware-a-mt7981-spim-nand-ddr4
 endef
 
+define U-Boot/mt7981_cmcc_rax3000m-ubi-ddr3
+  NAME:=CMCC RAX3000M (DDR3, UBI)
+  BUILD_SUBTARGET:=filogic
+  BUILD_DEVICES:=cmcc_rax3000m
+  UBOOT_CONFIG:=mt7981_cmcc_rax3000m-ubi
+  UBOOT_IMAGE:=u-boot.fip
+  BL2_BOOTDEV:=spim-nand-ubi
+  BL2_SOC:=mt7981
+  BL2_DDRTYPE:=ddr3-1866
+  DEPENDS:=+trusted-firmware-a-mt7981-spim-nand-ubi-ddr3-1866
+endef
+
+define U-Boot/mt7981_cmcc_rax3000m-ubi-ddr4
+  NAME:=CMCC RAX3000M (DDR4, UBI)
+  BUILD_SUBTARGET:=filogic
+  BUILD_DEVICES:=cmcc_rax3000m
+  UBOOT_CONFIG:=mt7981_cmcc_rax3000m-ubi
+  UBOOT_IMAGE:=u-boot.fip
+  BL2_BOOTDEV:=spim-nand-ubi
+  BL2_SOC:=mt7981
+  BL2_DDRTYPE:=ddr4
+  DEPENDS:=+trusted-firmware-a-mt7981-spim-nand-ubi-ddr4
+endef
+
 define U-Boot/mt7981_comfast_cf-wr632ax
   NAME:=COMFAST CF-WR632AX
   BUILD_SUBTARGET:=filogic
@@ -1378,6 +1402,8 @@ UBOOT_TARGETS := \
 	mt7981_cmcc_rax3000m-emmc-ddr4 \
 	mt7981_cmcc_rax3000m-nand-ddr3 \
 	mt7981_cmcc_rax3000m-nand-ddr4 \
+	mt7981_cmcc_rax3000m-ubi-ddr3 \
+	mt7981_cmcc_rax3000m-ubi-ddr4 \
 	mt7981_comfast_cf-wr632ax \
 	mt7981_comfast_cf-wr632ax-ubi \
 	mt7981_creatlentem_clt-r30b1-ubi \
--- a/package/boot/uboot-mediatek/patches/437-add-cmcc_rax3000m.patch
+++ b/package/boot/uboot-mediatek/patches/437-add-cmcc_rax3000m.patch
@@ -261,8 +261,140 @@
 +CONFIG_ZSTD=y
 +CONFIG_HEXDUMP=y
 --- /dev/null
++++ b/configs/mt7981_cmcc_rax3000m-ubi_defconfig
+@@ -0,0 +1,129 @@
++CONFIG_ARM=y
++CONFIG_ARM_SMCCC=y
++CONFIG_SYS_HAS_NONCACHED_MEMORY=y
++CONFIG_POSITION_INDEPENDENT=y
++CONFIG_ARCH_MEDIATEK=y
++CONFIG_TEXT_BASE=0x41e00000
++CONFIG_SYS_MALLOC_F_LEN=0x4000
++CONFIG_NR_DRAM_BANKS=1
++CONFIG_DEFAULT_DEVICE_TREE="mt7981-cmcc-rax3000m-ubi"
++CONFIG_OF_LIBFDT_OVERLAY=y
++CONFIG_TARGET_MT7981=y
++CONFIG_SYS_LOAD_ADDR=0x46000000
++CONFIG_PRE_CON_BUF_ADDR=0x4007EF00
++CONFIG_DEBUG_UART_BASE=0x11002000
++CONFIG_DEBUG_UART_CLOCK=40000000
++CONFIG_PCI=y
++CONFIG_DEBUG_UART=y
++CONFIG_AHCI=y
++CONFIG_FIT=y
++CONFIG_BOOTDELAY=30
++CONFIG_AUTOBOOT_KEYED=y
++CONFIG_AUTOBOOT_MENU_SHOW=y
++CONFIG_USE_PREBOOT=y
++CONFIG_DEFAULT_FDT_FILE="mt7981-cmcc-rax3000m-ubi"
++CONFIG_LOGLEVEL=7
++CONFIG_PRE_CONSOLE_BUFFER=y
++CONFIG_LOG=y
++# CONFIG_BOARD_INIT is not set
++CONFIG_BOARD_LATE_INIT=y
++CONFIG_HUSH_PARSER=y
++CONFIG_SYS_PROMPT="MT7981> "
++CONFIG_CMD_CPU=y
++CONFIG_CMD_LICENSE=y
++# CONFIG_CMD_BOOTEFI_BOOTMGR is not set
++CONFIG_CMD_BOOTMENU=y
++CONFIG_CMD_ASKENV=y
++CONFIG_CMD_ERASEENV=y
++CONFIG_CMD_ENV_FLAGS=y
++CONFIG_CMD_RNG=y
++CONFIG_CMD_STRINGS=y
++CONFIG_CMD_DM=y
++CONFIG_CMD_GPIO=y
++CONFIG_CMD_PWM=y
++CONFIG_CMD_GPT=y
++CONFIG_CMD_MTD=y
++CONFIG_CMD_PART=y
++CONFIG_CMD_PCI=y
++CONFIG_CMD_TFTPSRV=y
++CONFIG_CMD_RARP=y
++CONFIG_CMD_CDP=y
++CONFIG_CMD_SNTP=y
++CONFIG_CMD_LINK_LOCAL=y
++CONFIG_CMD_DHCP=y
++CONFIG_CMD_DNS=y
++CONFIG_CMD_PING=y
++CONFIG_CMD_PXE=y
++CONFIG_CMD_CACHE=y
++CONFIG_CMD_PSTORE=y
++CONFIG_CMD_PSTORE_MEM_ADDR=0x42ff0000
++CONFIG_CMD_UUID=y
++CONFIG_CMD_HASH=y
++CONFIG_CMD_SMC=y
++CONFIG_CMD_EXT4=y
++CONFIG_CMD_FAT=y
++CONFIG_CMD_FS_GENERIC=y
++CONFIG_CMD_FS_UUID=y
++CONFIG_CMD_UBI=y
++CONFIG_CMD_UBI_RENAME=y
++CONFIG_OF_EMBED=y
++CONFIG_ENV_OVERWRITE=y
++CONFIG_ENV_IS_IN_UBI=y
++CONFIG_ENV_REDUNDANT=y
++CONFIG_ENV_UBI_PART="ubi"
++CONFIG_ENV_UBI_VOLUME="ubootenv"
++CONFIG_ENV_UBI_VOLUME_REDUND="ubootenv2"
++CONFIG_ENV_RELOC_GD_ENV_ADDR=y
++CONFIG_ENV_USE_DEFAULT_ENV_TEXT_FILE=y
++CONFIG_ENV_DEFAULT_ENV_TEXT_FILE="defenvs/cmcc_rax3000m-ubi_env"
++CONFIG_ENV_VARS_UBOOT_RUNTIME_CONFIG=y
++CONFIG_VERSION_VARIABLE=y
++CONFIG_NETCONSOLE=y
++CONFIG_USE_IPADDR=y
++CONFIG_IPADDR="192.168.1.1"
++CONFIG_USE_SERVERIP=y
++CONFIG_SERVERIP="192.168.1.254"
++CONFIG_NET_RANDOM_ETHADDR=y
++CONFIG_SCSI_AHCI=y
++CONFIG_AHCI_PCI=y
++CONFIG_MTK_AHCI=y
++CONFIG_BUTTON=y
++CONFIG_BUTTON_GPIO=y
++CONFIG_CLK=y
++CONFIG_GPIO_HOG=y
++CONFIG_LED=y
++CONFIG_LED_BLINK=y
++CONFIG_LED_GPIO=y
++# CONFIG_MMC is not set
++CONFIG_MTD=y
++CONFIG_DM_MTD=y
++CONFIG_MTD_SPI_NAND=y
++CONFIG_MTD_UBI_FASTMAP=y
++CONFIG_PHY_FIXED=y
++CONFIG_MEDIATEK_ETH=y
++CONFIG_PCIE_MEDIATEK=y
++CONFIG_PHY=y
++CONFIG_PHY_MTK_TPHY=y
++CONFIG_PINCTRL=y
++CONFIG_PINCONF=y
++CONFIG_PINCTRL_MT7622=y
++CONFIG_PINCTRL_MT7981=y
++CONFIG_POWER_DOMAIN=y
++CONFIG_MTK_POWER_DOMAIN=y
++CONFIG_DM_REGULATOR=y
++CONFIG_DM_REGULATOR_FIXED=y
++CONFIG_DM_REGULATOR_GPIO=y
++CONFIG_DM_PWM=y
++CONFIG_PWM_MTK=y
++CONFIG_RAM=y
++CONFIG_DM_RNG=y
++CONFIG_RNG_MTK_V2=y
++CONFIG_SCSI=y
++CONFIG_DM_SERIAL=y
++CONFIG_SERIAL_RX_BUFFER=y
++CONFIG_MTK_SERIAL=y
++CONFIG_SPI=y
++CONFIG_DM_SPI=y
++CONFIG_MTK_SPIM=y
++CONFIG_ZSTD=y
++CONFIG_HEXDUMP=y
+--- /dev/null
 +++ b/arch/arm/dts/mt7981-cmcc-rax3000m.dtsi
-@@ -0,0 +1,85 @@
+@@ -0,0 +1,138 @@
 +// SPDX-License-Identifier: GPL-2.0-only
 +/*
 + * Copyright (c) 2022 MediaTek Inc.
@@ -340,78 +472,6 @@
 +	};
 +};
 +
-+&uart0 {
-+	mediatek,force-highspeed;
-+	status = "okay";
-+};
-+
-+&watchdog {
-+	status = "disabled";
-+};
---- /dev/null
-+++ b/arch/arm/dts/mt7981-cmcc-rax3000m-emmc.dts
-@@ -0,0 +1,53 @@
-+// SPDX-License-Identifier: GPL-2.0-only
-+
-+/dts-v1/;
-+#include "mt7981-cmcc-rax3000m.dtsi"
-+
-+/ {
-+	reg_3p3v: regulator-3p3v {
-+		compatible = "regulator-fixed";
-+		regulator-name = "fixed-3.3V";
-+		regulator-min-microvolt = <3300000>;
-+		regulator-max-microvolt = <3300000>;
-+		regulator-boot-on;
-+		regulator-always-on;
-+	};
-+};
-+
-+&mmc0 {
-+	pinctrl-names = "default";
-+	pinctrl-0 = <&mmc0_pins_default>;
-+	max-frequency = <26000000>;
-+	bus-width = <8>;
-+	cap-mmc-hw-reset;
-+	vmmc-supply = <&reg_3p3v>;
-+	non-removable;
-+	status = "okay";
-+};
-+
-+&pio {
-+	mmc0_pins_default: mmc0default {
-+		mux {
-+			function = "flash";
-+			groups =  "emmc_45";
-+		};
-+		conf-cmd-dat {
-+			pins = "SPI0_CLK", "SPI0_MOSI", "SPI0_MISO",
-+				"SPI0_CS",  "SPI0_HOLD", "SPI0_WP",
-+				"SPI1_CLK", "SPI1_MOSI", "SPI1_MISO";
-+			input-enable;
-+			drive-strength = <MTK_DRIVE_4mA>;
-+			bias-pull-up = <MTK_PUPD_SET_R1R0_01>;
-+		};
-+		conf-clk {
-+			pins = "SPI1_CS";
-+			drive-strength = <MTK_DRIVE_6mA>;
-+			bias-pull-down = <MTK_PUPD_SET_R1R0_10>;
-+		};
-+		conf-rst {
-+			pins = "PWM0";
-+			drive-strength = <MTK_DRIVE_4mA>;
-+			bias-pull-up = <MTK_PUPD_SET_R1R0_01>;
-+		};
-+	};
-+};
---- /dev/null
-+++ b/arch/arm/dts/mt7981-cmcc-rax3000m-nand.dts
-@@ -0,0 +1,77 @@
-+// SPDX-License-Identifier: GPL-2.0-only
-+
-+/dts-v1/;
-+#include "mt7981-cmcc-rax3000m.dtsi"
-+
 +&pio {
 +	spi_flash_pins: spi0-pins-func-1 {
 +		mux {
@@ -452,7 +512,7 @@
 +		reg = <0>;
 +		spi-max-frequency = <52000000>;
 +
-+		partitions {
++		partitions: partitions {
 +			compatible = "fixed-partitions";
 +			#address-cells = <1>;
 +			#size-cells = <1>;
@@ -461,27 +521,115 @@
 +				label = "bl2";
 +				reg = <0x0 0x100000>;
 +			};
-+
-+			partition@100000 {
-+				label = "orig-env";
-+				reg = <0x100000 0x80000>;
-+			};
-+
-+			partition@160000 {
-+				label = "factory";
-+				reg = <0x180000 0x200000>;
-+			};
-+
-+			partition@380000 {
-+				label = "fip";
-+				reg = <0x380000 0x200000>;
-+			};
-+
-+			partition@580000 {
-+				label = "ubi";
-+				reg = <0x580000 0x7200000>;
-+			};
 +		};
++	};
++};
++
++&uart0 {
++	mediatek,force-highspeed;
++	status = "okay";
++};
++
++&watchdog {
++	status = "disabled";
++};
+--- /dev/null
++++ b/arch/arm/dts/mt7981-cmcc-rax3000m-emmc.dts
+@@ -0,0 +1,55 @@
++// SPDX-License-Identifier: GPL-2.0-only
++
++#include "mt7981-cmcc-rax3000m.dtsi"
++
++/ {
++	reg_3p3v: regulator-3p3v {
++		compatible = "regulator-fixed";
++		regulator-name = "fixed-3.3V";
++		regulator-min-microvolt = <3300000>;
++		regulator-max-microvolt = <3300000>;
++		regulator-boot-on;
++		regulator-always-on;
++	};
++};
++
++&mmc0 {
++	pinctrl-names = "default";
++	pinctrl-0 = <&mmc0_pins_default>;
++	max-frequency = <26000000>;
++	bus-width = <8>;
++	cap-mmc-hw-reset;
++	vmmc-supply = <&reg_3p3v>;
++	non-removable;
++	status = "okay";
++};
++
++/delete-node/ &spi0;
++/delete-node/ &spi_flash_pins;
++
++&pio {
++	mmc0_pins_default: mmc0default {
++		mux {
++			function = "flash";
++			groups =  "emmc_45";
++		};
++		conf-cmd-dat {
++			pins = "SPI0_CLK", "SPI0_MOSI", "SPI0_MISO",
++				"SPI0_CS",  "SPI0_HOLD", "SPI0_WP",
++				"SPI1_CLK", "SPI1_MOSI", "SPI1_MISO";
++			input-enable;
++			drive-strength = <MTK_DRIVE_4mA>;
++			bias-pull-up = <MTK_PUPD_SET_R1R0_01>;
++		};
++		conf-clk {
++			pins = "SPI1_CS";
++			drive-strength = <MTK_DRIVE_6mA>;
++			bias-pull-down = <MTK_PUPD_SET_R1R0_10>;
++		};
++		conf-rst {
++			pins = "PWM0";
++			drive-strength = <MTK_DRIVE_4mA>;
++			bias-pull-up = <MTK_PUPD_SET_R1R0_01>;
++		};
++	};
++};
+--- /dev/null
++++ b/arch/arm/dts/mt7981-cmcc-rax3000m-nand.dts
+@@ -0,0 +1,25 @@
++// SPDX-License-Identifier: GPL-2.0-only
++
++#include "mt7981-cmcc-rax3000m.dtsi"
++
++&partitions {
++	partition@100000 {
++		label = "orig-env";
++		reg = <0x100000 0x80000>;
++	};
++
++	partition@180000 {
++		label = "factory";
++		reg = <0x180000 0x200000>;
++	};
++
++	partition@380000 {
++		label = "fip";
++		reg = <0x380000 0x200000>;
++	};
++
++	partition@580000 {
++		label = "ubi";
++		reg = <0x580000 0x7200000>;
++	};
++};
+--- /dev/null
++++ b/arch/arm/dts/mt7981-cmcc-rax3000m-ubi.dts
+@@ -0,0 +1,10 @@
++// SPDX-License-Identifier: GPL-2.0-only
++
++#include "mt7981-cmcc-rax3000m.dtsi"
++
++&partitions {
++	partition@100000 {
++		label = "ubi";
++		reg = <0x100000 0x7f00000>;
 +	};
 +};
 --- /dev/null
@@ -595,6 +743,63 @@
 +ubi_read_production=ubi read $loadaddr fit && iminfo $loadaddr && run ubi_prepare_rootfs
 +ubi_read_recovery=ubi check recovery && ubi read $loadaddr recovery
 +ubi_remove_rootfs=ubi check rootfs_data && ubi remove rootfs_data
++ubi_write_production=ubi check fit && ubi remove fit ; run ubi_remove_rootfs ; ubi create fit $filesize dynamic && ubi write $loadaddr fit $filesize
++ubi_write_recovery=ubi check recovery && ubi remove recovery ; run ubi_remove_rootfs ; ubi create recovery $filesize dynamic && ubi write $loadaddr recovery $filesize
++_init_env=setenv _init_env ; run ubi_create_env ; saveenv ; saveenv
++_firstboot=setenv _firstboot ; run _switch_to_menu ; run _init_env ; run boot_first
++_switch_to_menu=setenv _switch_to_menu ; setenv bootdelay 3 ; setenv bootmenu_delay 3 ; setenv bootmenu_0 $bootmenu_0d ; setenv bootmenu_0d ; run _bootmenu_update_title
++_bootmenu_update_title=setenv _bootmenu_update_title ; setenv bootmenu_title "$bootmenu_title       [33m$ver[0m"
+--- /dev/null
++++ b/defenvs/cmcc_rax3000m-ubi_env
+@@ -0,0 +1,54 @@
++ipaddr=192.168.1.1
++serverip=192.168.1.254
++loadaddr=0x46000000
++console=earlycon=uart8250,mmio32,0x11002000 console=ttyS0
++bootcmd=if pstore check ; then run boot_recovery ; else run boot_ubi ; fi
++bootconf=config-1#mt7981b-cmcc-rax3000m-ubi
++bootdelay=0
++bootfile=openwrt-mediatek-filogic-cmcc_rax3000m-initramfs-recovery.itb
++bootfile_bl2=openwrt-mediatek-filogic-cmcc_rax3000m-ubi-preloader.bin
++bootfile_fip=openwrt-mediatek-filogic-cmcc_rax3000m-ubi-bl31-uboot.fip
++bootfile_upg=openwrt-mediatek-filogic-cmcc_rax3000m-squashfs-sysupgrade.itb
++bootled_pwr=red:status
++bootled_rec=blue:status
++bootmenu_confirm_return=askenv - Press ENTER to return to menu ; bootmenu 60
++bootmenu_default=0
++bootmenu_delay=0
++bootmenu_title=      [0;34m( ( ( [1;39mOpenWrt[0;34m ) ) )  [0;36m[UBI][0m
++bootmenu_0=Initialize environment.=run _firstboot
++bootmenu_0d=Run default boot command.=run boot_default
++bootmenu_1=Boot system via TFTP.=run boot_tftp ; run bootmenu_confirm_return
++bootmenu_2=Boot production system from NAND.=run boot_production ; run bootmenu_confirm_return
++bootmenu_3=Boot recovery system from NAND.=run boot_recovery ; run bootmenu_confirm_return
++bootmenu_4=Load production system via TFTP then write to NAND.=setenv noboot 1 ; setenv replacevol 1 ; run boot_tftp_production ; setenv noboot ; setenv replacevol ; run bootmenu_confirm_return
++bootmenu_5=Load recovery system via TFTP then write to NAND.=setenv noboot 1 ; setenv replacevol 1 ; run boot_tftp_recovery ; setenv noboot ; setenv replacevol ; run bootmenu_confirm_return
++bootmenu_6=[31mLoad BL31+U-Boot FIP via TFTP then write to NAND.[0m=run boot_tftp_write_fip ; run bootmenu_confirm_return
++bootmenu_7=[31mLoad BL2 preloader via TFTP then write to NAND.[0m=run boot_tftp_write_bl2 ; run bootmenu_confirm_return
++bootmenu_8=Reboot.=reset
++bootmenu_9=Reset all settings to factory defaults.=run reset_factory ; reset
++boot_first=if button reset ; then led $bootled_pwr off ; run boot_tftp_recovery ; setenv flag_recover 1 ; run boot_default ; fi ; bootmenu
++boot_default=if env exists flag_recover ; then else run bootcmd ; fi ; run boot_recovery ; setenv replacevol 1 ; run boot_tftp_forever
++boot_production=led $bootled_rec off ; run ubi_read_production && bootm $loadaddr#$bootconf ; led $bootled_pwr off
++boot_recovery=led $bootled_pwr off ; run ubi_read_recovery && bootm $loadaddr#$bootconf ; led $bootled_rec off
++boot_ubi=run boot_production ; run boot_recovery ; run boot_tftp_forever
++boot_tftp_forever=led $bootled_pwr off ; while true ; do run boot_tftp_recovery ; sleep 1 ; done
++boot_tftp_production=tftpboot $loadaddr $bootfile_upg && env exists replacevol && iminfo $loadaddr && run ubi_write_production ; if env exists noboot ; then else bootm $loadaddr#$bootconf ; fi
++boot_tftp_recovery=tftpboot $loadaddr $bootfile && env exists replacevol && iminfo $loadaddr && run ubi_write_recovery ; if env exists noboot ; then else bootm $loadaddr#$bootconf ; fi
++boot_tftp=tftpboot $loadaddr $bootfile && bootm $loadaddr#$bootconf
++boot_tftp_write_fip=tftpboot $loadaddr $bootfile_fip && run ubi_write_fip && run reset_factory
++boot_tftp_write_bl2=tftpboot $loadaddr $bootfile_bl2 && run mtd_write_bl2
++preboot=led $bootled_pwr on ; led $bootled_rec on
++reset_factory=mw $loadaddr 0xff 0x1f000 ; ubi write $loadaddr ubootenv 0x1f000 ; ubi write $loadaddr ubootenv2 0x1f000 ; ubi remove rootfs_data
++mtd_write_bl2=mtd erase bl2 && mtd write bl2 $loadaddr
++ubi_create_env=ubi check ubootenv || ubi create ubootenv 0x1f000 dynamic ; ubi check ubootenv2 || ubi create ubootenv2 0x1f000 dynamic
++ubi_prepare_rootfs=if ubi check rootfs_data ; then else if env exists rootfs_data_max ; then ubi create rootfs_data $rootfs_data_max dynamic || ubi create rootfs_data - dynamic ; else ubi create rootfs_data - dynamic ; fi ; fi
++ubi_read_production=ubi read $loadaddr fit && iminfo $loadaddr && run ubi_prepare_rootfs
++ubi_read_recovery=ubi check recovery && ubi read $loadaddr recovery
++ubi_remove_rootfs=ubi check rootfs_data && ubi remove rootfs_data
++ubi_write_fip=run ubi_remove_rootfs ; ubi check fip && ubi remove fip ; ubi create fip 0x200000 static ; ubi write $loadaddr fip 0x200000
 +ubi_write_production=ubi check fit && ubi remove fit ; run ubi_remove_rootfs ; ubi create fit $filesize dynamic && ubi write $loadaddr fit $filesize
 +ubi_write_recovery=ubi check recovery && ubi remove recovery ; run ubi_remove_rootfs ; ubi create recovery $filesize dynamic && ubi write $loadaddr recovery $filesize
 +_init_env=setenv _init_env ; run ubi_create_env ; saveenv ; saveenv
--- a/target/linux/mediatek/dts/mt7981b-cmcc-rax3000m-ubi.dtso
+++ b/target/linux/mediatek/dts/mt7981b-cmcc-rax3000m-ubi.dtso
@@ -1,0 +1,178 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+/dts-v1/;
+/plugin/;
+
+#include <dt-bindings/pinctrl/mt65xx.h>
+
+/ {
+	compatible = "cmcc,rax3000m", "mediatek,mt7981";
+
+	fragment@0 {
+		target = <&chosen>;
+		__overlay__ {
+			rootdisk = <&ubi_rootdisk>;
+		};
+	};
+
+	fragment@1 {
+		target = <&gmac0>;
+		__overlay__ {
+			nvmem-cells = <&macaddr_factory_2a 0>;
+			nvmem-cell-names = "mac-address";
+		};
+	};
+
+	fragment@2 {
+		target = <&gmac1>;
+		__overlay__ {
+			nvmem-cells = <&macaddr_factory_24 0>;
+			nvmem-cell-names = "mac-address";
+		};
+	};
+
+	fragment@3 {
+		target = <&pio>;
+		__overlay__ {
+			spi0_flash_pins: spi0-pins {
+				mux {
+					function = "spi";
+					groups = "spi0", "spi0_wp_hold";
+				};
+
+				conf-pu {
+					pins = "SPI0_CS", "SPI0_HOLD", "SPI0_WP";
+					drive-strength = <MTK_DRIVE_8mA>;
+					bias-disable; /* bias-disable */
+				};
+
+				conf-pd {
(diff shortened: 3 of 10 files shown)


target/linux/mediatek/image/filogic.mk
```diff
@@ -945,27 +945,28 @@ define Device/cmcc_a10-ubootmod
...
   DEVICE_ALT0_MODEL := RAX3000Me
-  DEVICE_DTS := mt7981b-cmcc-rax3000m
-  DEVICE_DTS_OVERLAY := mt7981b-cmcc-rax3000m-emmc mt7981b-cmcc-rax3000m-nand
   DEVICE_DTS_DIR := ../dts
-  DEVICE_DTC_FLAGS := --pad 4096
-  DEVICE_DTS_LOADADDR := 0x43f00000
   DEVICE_PACKAGES := kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware kmod-usb3 \
 	e2fsprogs f2fsck mkf2fs
-  KERNEL_LOADADDR := 0x44000000
   KERNEL := kernel-bin | gzip
```
openwrt-ai, line 956: The split into `Device/cmcc_rax3000m-common` silently drops three variables that the existing `cmcc_rax3000m` device had on main and that are not re-added anywhere:

- `DEVICE_DTC_FLAGS := --pad 4096`
- `DEVICE_DTS_LOADADDR := 0x43f00000`
- `KERNEL_LOADADDR := 0x44000000`

(see [`Device/cmcc_rax3000m` on the base commit](https://github.com/openwrt/openwrt/blob/0903e882e1d64a07172c22c87c31717a09c2b27b/target/linux/mediatek/image/filogic.mk#L949-L961))

Two concrete consequences for the *already supported* `cmcc_rax3000m`, which this PR is not supposed to change:

1. `KERNEL_LOADADDR` falls back to the filogic default `0x48000000` from [`loadaddr-$(CONFIG_TARGET_mediatek_filogic)`](https://github.com/openwrt/openwrt/blob/452e412bdd5cf32e89b7e35fb1d41463022af584/target/linux/mediatek/image/Makefile#L14), and `DEVICE_DTS_LOADADDR` disappears from the `fit` call — so the sysupgrade FIT for the existing device is built with different load/entry addresses than the images users are running today.
2. `--pad 4096` is what leaves room in the base dtb for `DEVICE_DTS_OVERLAY` (`mt7981b-cmcc-rax3000m-emmc` / `-nand`) to be applied; without it the overlay application can fail at boot.

Restoring them in the common block keeps the existing device byte-for-byte and gives the new `-ubi` device the same proven addresses:

```suggestion
  DEVICE_DTS_DIR := ../dts
  DEVICE_DTC_FLAGS := --pad 4096
  DEVICE_DTS_LOADADDR := 0x43f00000
  DEVICE_PACKAGES := kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware kmod-usb3 \
	e2fsprogs f2fsck mkf2fs
  KERNEL_LOADADDR := 0x44000000
  KERNEL := kernel-bin | gzip
```

If dropping them from the `-ubi` variant is deliberate, please move them back into `Device/cmcc_rax3000m` only, and say so in the commit message — as it stands the message only describes adding a new layout, so this behaviour change to the existing device is undocumented.

---

target/linux/mediatek/dts/mt7981b-cmcc-rax3000m-ubi.dts
```diff
@@ -0,0 +1,156 @@
...
+		reg = <0>;
+		nvmem-cells = <&macaddr_factory_4 0>;
+		nvmem-cell-names = "mac-address";
+	};
+
+	band@1 {
+		reg = <1>;
+		nvmem-cells = <&macaddr_factory_a 0>;
+		nvmem-cell-names = "mac-address";
+	};
```
openwrt-ai, line 155: Is `factory@0xa` actually programmed on this board? The existing support for `cmcc,rax3000m` does not read it — it derives the 5 GHz MAC in userspace instead:

```
cmcc,rax3000m)
	addr=$(cat /sys/class/net/eth0/address)
	[ "$PHYNBR" = "1" ] && macaddr_add $addr -1 > /sys${DEVPATH}/macaddress
	;;
```
([11_fix_wifi_mac:83-86](``https://github.com/openwrt/openwrt/blob/452e412bdd5cf32e89b7e35fb1d41463022af584/target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac#L83-L86)),`` i.e. `macaddr_factory_2a - 1`. Consistently, the eMMC/NAND overlays only declare `macaddr_factory_24`/`_2a` and no `_4`/`_a` ([mt7981b-cmcc-rax3000m-nand.dtso](https://github.com/openwrt/openwrt/blob/452e412bdd5cf32e89b7e35fb1d41463022af584/target/linux/mediatek/dts/mt7981b-cmcc-rax3000m-nand.dtso#L96-L112)) — the presence of that hotplug hack suggests `0xa` is not a valid MAC on this device.

The comparable precedent, `qihoo,360t7-ubi`, kept MACs identical by replicating the hotplug derivation in DT (`macaddr_factory_8000 (2)` / `(3)`, matching `lanMac + 2` / `+ 3`). Here the derivation is replaced with a different source, so at minimum the 5 GHz MAC will change relative to the NAND/eMMC layout on the same hardware, and it may end up `ff:ff:ff:ff:ff:ff` if the vendor never wrote `0xa`.

If `0xa` is confirmed valid on real units, a note in the commit message's MAC-address section would help; otherwise please mirror the existing `macaddr_factory_2a`-based derivation.

---

target/linux/mediatek/dts/mt7981b-cmcc-rax3000m-ubi.dts
```diff
@@ -0,0 +1,156 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+#include "mt7981b-cmcc-rax3000m.dtsi"
+
+/ {
+	model = "CMCC RAX3000M(e) (UBI)";
```
openwrt-ai, line 6: nit: the `(e)` does not belong in the DTS `model`. The sibling `mt7981b-cmcc-rax3000m.dts` uses plain `"CMCC RAX3000M"`, and the image metadata for this device resolves to `CMCC RAX3000M (UBI)` (`DEVICE_MODEL := RAX3000M` + `DEVICE_VARIANT := (UBI)`), with `RAX3000Me` carried separately by `DEVICE_ALT0_MODEL`. In-tree UBI variants follow that: `"CreatLentem CLT-R30B1 (UBI)"`, `"COMFAST CF-WR632AX (UBI)"`, `"Qihoo 360T7 (UBI)"`.

```suggestion
	model = "CMCC RAX3000M (UBI)";
```

Same applies to the U-Boot DTS `model` in `package/boot/uboot-mediatek/patches/475-add-cmcc_rax3000m-ubi.patch` line 141.

---

package/boot/uboot-mediatek/patches/475-add-cmcc_rax3000m-ubi.patch
```diff
@@ -0,0 +1,258 @@
...
++			pins = "SPI0_CS", "SPI0_HOLD", "SPI0_WP";
++			drive-strength = <MTK_DRIVE_4mA>;
++			bias-pull-up = <MTK_PUPD_SET_R1R0_11>;
++		};
++
++		conf-pd {
++			pins = "SPI0_CLK", "SPI0_MOSI", "SPI0_MISO";
++			drive-strength = <MTK_DRIVE_4mA>;
++			bias-pull-down = <MTK_PUPD_SET_R1R0_11>;
++		};
```
openwrt-ai, line 162: These SPI0 pin settings look copied from `473-add-creatlentem_clt-r30b1-ubi.patch` rather than derived from this board. The same physical SPI-NAND on the RAX3000M is driven with different values by the board's existing U-Boot DTS (`arch/arm/dts/mt7981-cmcc-rax3000m-nand.dts`, added by `437-add-cmcc_rax3000m.patch`):

```
drive-strength = <MTK_DRIVE_8mA>;
bias-pull-up = <MTK_PUPD_SET_R1R0_00>;   /* conf-pu */
bias-pull-down = <MTK_PUPD_SET_R1R0_00>; /* conf-pd */
```

The Linux side has the same divergence: [mt7981b-cmcc-rax3000m-nand.dtso](https://github.com/openwrt/openwrt/blob/452e412bdd5cf32e89b7e35fb1d41463022af584/target/linux/mediatek/dts/mt7981b-cmcc-rax3000m-nand.dtso#L44-L52) uses `MTK_DRIVE_8mA` + `bias-disable`, while `mt7981b-cmcc-rax3000m-ubi.dts` in this PR uses `MTK_DRIVE_4mA` + `MTK_PUPD_SET_R1R0_11`.

Was the weaker drive strength / different bias actually validated on this hardware, or should these match the existing RAX3000M values? Marginal drive settings on a flash bus tend to fail intermittently rather than outright, so it's worth being explicit.

---

package/boot/uboot-mediatek/patches/475-add-cmcc_rax3000m-ubi.patch
```diff
@@ -0,0 +1,258 @@
...
++ipaddr=192.168.1.1
++serverip=192.168.1.254
++loadaddr=0x46000000
++console=earlycon=uart8250,mmio32,0x11002000 console=ttyS0
++bootcmd=if pstore check ; then run boot_recovery ; else run boot_ubi ; fi
++bootconf=config-1
++bootdelay=0
++bootfile=openwrt-mediatek-filogic-cmcc_rax3000m-ubi-initramfs-recovery.itb
++bootfile_bl2=openwrt-mediatek-filogic-cmcc_rax3000m-ubi-preloader.bin
++bootfile_fip=openwrt-mediatek-filogic-cmcc_rax3000m-ubi-bl31-uboot.fip
```
openwrt-ai, line 214: nit: these two filenames don't match any artifact this device produces. `Device/cmcc_rax3000m-ubi` builds DDR-qualified artifacts:

```
openwrt-<ver>-mediatek-filogic-cmcc_rax3000m-ubi-ddr3-preloader.bin
openwrt-<ver>-mediatek-filogic-cmcc_rax3000m-ubi-ddr4-preloader.bin
openwrt-<ver>-mediatek-filogic-cmcc_rax3000m-ubi-ddr3-bl31-uboot.fip
openwrt-<ver>-mediatek-filogic-cmcc_rax3000m-ubi-ddr4-bl31-uboot.fip
```

so bootmenu entries 6 and 7 (`boot_tftp_write_fip` / `boot_tftp_write_bl2`) will TFTP a name that never exists and the user has to `setenv` by hand. The single-DDR boards this defenv was copied from (creatlentem, comfast) don't have this problem.

Since the DDR type is fixed per built FIP, each of the two `U-Boot/mt7981_cmcc_rax3000m-ubi-ddr{3,4}` configs would need its own defenv (or at least `bootfile_bl2`/`bootfile_fip` overrides) to make these menu entries usable. Note the pre-existing `cmcc_rax3000m-nand_env` has the same flaw, so this may be considered out of scope — but it's cheap to get right in a new file.

---

openwrt-ai (commented):
## Commit checks

- 452e412 "mediatek: cmcc rax3000m: add all-in-UBI layout" — the message describes only the addition of a new all-in-UBI layout, but the diff also refactors the existing `Device/cmcc_rax3000m` into a shared `-common` block and, in doing so, drops `DEVICE_DTC_FLAGS := --pad 4096`, `DEVICE_DTS_LOADADDR := 0x43f00000` and `KERNEL_LOADADDR := 0x44000000` from the already-supported device without re-adding them anywhere. That is a behaviour change to an existing device that the message doesn't mention (details inline on `target/linux/mediatek/image/filogic.mk`).


---

legale (approved):
approved

ptpt52:
mark.

target/linux/mediatek/image/filogic.mk
andros-ua: All changes have been run-tested on hardware. This includes verifying the transition in both directions — switching from the existing layout to all-in-UBI and back.

andros-ua:
Refactored PR to reuse the existing U-Boot patch instead of introducing
a separate one, and split the change into separate commits.

package/boot/uboot-mediatek/Makefile
```diff
@@ -305,6 +305,30 @@ define U-Boot/mt7981_cmcc_rax3000m-nand-ddr4
   DEPENDS:=+trusted-firmware-a-mt7981-spim-nand-ddr4
 endef
 
+define U-Boot/mt7981_cmcc_rax3000m-ubi-ddr3
+  NAME:=CMCC RAX3000M(e) (UBI)
```
openwrt-ai: nit: this `NAME` doesn't match the four existing RAX3000M U-Boot targets in the same file, which all use plain [`NAME:=CMCC RAX3000M`](https://github.com/openwrt/openwrt/blob/7b0f1b3e63657d7e8f2e03f06cdc6723c9673175/package/boot/uboot-mediatek/Makefile#L261) (emmc-ddr3/ddr4, nand-ddr3/ddr4), nor the `<Vendor> <Model> (UBI)` form used by every other UBI target here (`COMFAST CF-WR632AX (UBI)`, `CreatLentem CLT-R30B1 (UBI)`, `Qihoo 360T7 (UBI)`, `TP-Link BE450 (UBI)`).

Commit 7b0f1b3 says it "harmonize[s] model caption to `CMCC RAX3000M(e)`", but nothing existing is changed to match — `DEVICE_MODEL` stays `RAX3000M` (+ `DEVICE_ALT0_MODEL := RAX3000Me`), `mt7981b-cmcc-rax3000m.dts` keeps `model = "CMCC RAX3000M"`, and these four U-Boot NAMEs keep `CMCC RAX3000M`. So the net effect is a new inconsistency rather than harmonization.

```suggestion
  NAME:=CMCC RAX3000M (UBI)
```

Same applies to line 321 (the `-ubi-ddr4` target).

---

package/boot/uboot-mediatek/patches/437-add-cmcc_rax3000m.patch
```diff
@@ -278,7 +410,7 @@
 +	#address-cells = <1>;
 +	#size-cells = <1>;
 +	model = "CMCC RAX3000M";
-+	compatible = "mediatek,mt7981", "mediatek,mt7981-rfb";
++	compatible = "cmcc,rax3000m", "mediatek,mt7981";
```
openwrt-ai: This changes the root `compatible` of the *existing* eMMC and NAND U-Boot builds (they inherit it from this `.dtsi` and don't override it), from `"mediatek,mt7981", "mediatek,mt7981-rfb"` to `"cmcc,rax3000m", "mediatek,mt7981"`. The commit message for 93dd20e only mentions adding the UBI variant and refactoring the DTS into a common include; this part isn't called out.

Two things worth confirming:

1. Every other mt7981 board DTS in this package (21 occurrences) uses `compatible = "mediatek,mt7981", "mediatek,mt7981-rfb";`, matching upstream [`mt7981-rfb.dts:15`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/arch/arm/dts/mt7981-rfb.dts#L15). This is the only one that deviates.
2. I could not find any U-Boot code that matches on `mediatek,mt7981-rfb` (only `arch/arm/dts/Makefile` references the name), so this looks functionally inert — but if it is inert, it also isn't needed for the refactor, and dropping it silently changes two already-supported build targets.

Is the change deliberate? If so, a line in the commit message would help; if not, keeping the sibling form here and letting only `mt7981-cmcc-rax3000m-ubi.dts` carry `cmcc,rax3000m-ubi` would keep the existing targets byte-identical.

---

package/boot/uboot-mediatek/patches/437-add-cmcc_rax3000m.patch
```diff
@@ -340,6 +472,59 @@
...
++&pio {
++	spi_flash_pins: spi0-pins-func-1 {
++		mux {
++			function = "flash";
++			groups = "spi0", "spi0_wp_hold";
++		};
++
++		conf-pu {
++			pins = "SPI0_CS", "SPI0_HOLD", "SPI0_WP";
++			drive-strength = <MTK_DRIVE_4mA>;
```
openwrt-ai: nit: after 712d542 the U-Boot side of this board drives SPI0 at `MTK_DRIVE_4mA` + `MTK_PUPD_SET_R1R0_11`, but the Linux side for the *same* SPI-NAND on the *same* boards still uses `MTK_DRIVE_8mA` + `bias-disable` — see [`mt7981b-cmcc-rax3000m-nand.dtso:43-53`](https://github.com/openwrt/openwrt/blob/7b0f1b3e63657d7e8f2e03f06cdc6723c9673175/target/linux/mediatek/dts/mt7981b-cmcc-rax3000m-nand.dtso#L43-L53). Only the new `mt7981b-cmcc-rax3000m-ubi.dts` picked up the 4mA/R1R0_11 values.

If 8mA really causes overshoot on cost-reduced revisions, the NAND overlay would hit it too once the kernel reprograms the pins — the fix only covers the bootloader half. Was leaving the NAND overlay at 8mA intentional (e.g. to avoid touching the existing layout in this PR), or should it be updated in the same series?

---

package/boot/uboot-mediatek/patches/437-add-cmcc_rax3000m.patch
```diff
@@ -406,82 +593,48 @@
...
-+		reg = <0>;
-+		spi-max-frequency = <52000000>;
-+
-+		partitions {
-+			compatible = "fixed-partitions";
-+			#address-cells = <1>;
-+			#size-cells = <1>;
++	partition@160000 {
++		label = "factory";
++		reg = <0x180000 0x200000>;
```
openwrt-ai: nit: the unit address doesn't match `reg` — the node starts at `0x180000`, not `0x160000`. The Linux counterpart gets this right (`partition@180000` in `mt7981b-cmcc-rax3000m-nand.dtso`). The typo is pre-existing on main, but this hunk rewrites the whole `-nand.dts` anyway, so it's free to fix here; the line count is unchanged so the patch header stays valid.

```suggestion
+	partition@180000 {
+		label = "factory";
+		reg = <0x180000 0x200000>;
```

---

openwrt-ai (commented):
Re-reviewed after the force-push (3 new commits, `452e412` → `7b0f1b3`). The previously flagged loss of `DEVICE_DTC_FLAGS` / `DEVICE_DTS_LOADADDR` / `KERNEL_LOADADDR` on the existing `cmcc_rax3000m` is fixed — they are back in `Device/cmcc_rax3000m`, and the `.dts` → `.dtsi` split is now a pure move (verified: the only delta is `model`/`compatible` removal plus one blank line). The MAC-source question is answered by the new MAC table in the commit message.

## Commit checks

- 712d542 "uboot-mediatek: cmcc rax3000m: strengthen spi0 flash pin config" — the subject says *strengthen*, but the change **lowers** drive strength from `MTK_DRIVE_8mA` to `MTK_DRIVE_4mA` (the body correctly describes it as "Lower it to 4mA"). Consider a subject that matches, e.g. `uboot-mediatek: cmcc rax3000m: fix spi0 flash pin drive strength`.
- 7b0f1b3 "mediatek: filogic: cmcc rax3000m: add all-in-UBI layout" — the body claims "Cosmetic change: harmonize model caption to \"CMCC RAX3000M(e)\"", but no existing caption is changed by this commit: `mt7981b-cmcc-rax3000m.dts` keeps `model = "CMCC RAX3000M"` and `filogic.mk` keeps `DEVICE_MODEL := RAX3000M` / `DEVICE_ALT0_MODEL := RAX3000Me`. The only `(e)` strings introduced are on the new UBI variant, so the diff adds an inconsistency rather than harmonizing one (details inline on `package/boot/uboot-mediatek/Makefile`).

CI on `7b0f1b3` is green.

---

target/linux/mediatek/dts/mt7981b-cmcc-rax3000m-ubi.dts
andros-ua: Confirmed and described.

target/linux/mediatek/image/filogic.mk
andros-ua: restored

package/boot/uboot-mediatek/patches/437-add-cmcc_rax3000m.patch
andros-ua: fixed

package/boot/uboot-mediatek/patches/437-add-cmcc_rax3000m.patch
andros-ua: restored

package/boot/uboot-mediatek/patches/437-add-cmcc_rax3000m.patch
andros-ua: restored

package/boot/uboot-mediatek/Makefile
andros-ua: fixed

package/boot/uboot-mediatek/patches/475-add-cmcc_rax3000m-ubi.patch
andros-ua: Since the user will have to rename these files anyway, this isn't handled here.

package/boot/uboot-mediatek/patches/475-add-cmcc_rax3000m-ubi.patch
andros-ua: restored

target/linux/mediatek/dts/mt7981b-cmcc-rax3000m-ubi.dts
andros-ua: fixed

openwrt-ai (commented):
Re-reviewed after the force-push (`7b0f1b3` → `f25fc0fb`, now 2 commits: `325054f`, `f25fc0fb`). The changes since the last review all resolve previously-flagged items, and I verified each:

- U-Boot `NAME` and both DTS `model` strings changed from `CMCC RAX3000M(e) (UBI)` to `CMCC RAX3000M (UBI)`, matching the in-tree UBI convention.
- The dropped `DEVICE_DTC_FLAGS` / `DEVICE_DTS_LOADADDR` / `KERNEL_LOADADDR` remain restored on `Device/cmcc_rax3000m`.
- Root `compatible` on the shared U-Boot `.dtsi` is back to `"mediatek,mt7981", "mediatek,mt7981-rfb"`, keeping the existing eMMC/NAND targets byte-identical.
- The separate "strengthen spi0 flash pin config" commit was dropped; SPI0 drive strength is back to `MTK_DRIVE_8mA` + `bias-disable`/`R1R0_00` on both the U-Boot patch and the Linux `-ubi.dts`, matching the NAND/eMMC siblings.
- `partition@160000` unit-address typo corrected to `partition@180000`.
- The earlier inaccurate "harmonize model caption" note is gone from the commit message.

Commit messages match their diffs. CI on `f25fc0fb` is green.

---

andros-ua:
Refactor to follow the existing DTS overlay model.

package/boot/uboot-mediatek/patches/437-add-cmcc_rax3000m.patch
```diff
@@ -601,3 +749,60 @@
...
++++ b/defenvs/cmcc_rax3000m-ubi_env
+@@ -0,0 +1,54 @@
++ipaddr=192.168.1.1
++serverip=192.168.1.254
++loadaddr=0x46000000
++console=earlycon=uart8250,mmio32,0x11002000 console=ttyS0
++bootcmd=if pstore check ; then run boot_recovery ; else run boot_ubi ; fi
++bootconf=config-1#mt7981b-cmcc-rax3000m-ubi
++bootdelay=0
++bootfile=openwrt-mediatek-filogic-cmcc_rax3000m-ubi-initramfs-recovery.itb
```
openwrt-ai: `bootfile` names an image that nothing builds. Unlike `qihoo_360t7-ubi` or `comfast_cf-wr632ax-ubi`, the UBI layout in this PR is not a separate `TARGET_DEVICES` entry — it is a third `DEVICE_DTS_OVERLAY` on the existing [`Device/cmcc_rax3000m`](https://github.com/openwrt/openwrt/blob/41669139420ad9407860de6a9eab4d5de57d1a57/target/linux/mediatek/image/filogic.mk#L1003-L1048), so the images stay `openwrt-mediatek-filogic-cmcc_rax3000m-initramfs-recovery.itb` / `...-squashfs-sysupgrade.itb`. Nothing ever produces a `-ubi-` infixed image.

That breaks the TFTP recovery path in this defenv: `boot_tftp_forever` — the fallback taken when neither the production nor the recovery volume boots — plus bootmenu entries 1/3/5 all fetch `$bootfile`. The eMMC and NAND defenvs added by the same patch already use the unsuffixed names (lines 646/649 and 703/706 of this file), because they share the same OpenWrt device.

```suggestion
+bootfile=openwrt-mediatek-filogic-cmcc_rax3000m-initramfs-recovery.itb
```

Same applies to `bootfile_upg` on line 765 — it should be `openwrt-mediatek-filogic-cmcc_rax3000m-squashfs-sysupgrade.itb`, which is what bootmenu entry 4 TFTPs. (`bootfile_bl2` / `bootfile_fip` on lines 763/764 are the separate DDR-suffix issue you already declared out of scope.)

---

target/linux/mediatek/filogic/base-files/etc/board.d/02_network
```diff
@@ -56,6 +55,7 @@ mediatek_setup_interfaces()
 	cetron,ct3003|\
 	cmcc,a10-stock|\
 	cmcc,a10-ubootmod|\
+	cmcc,rax3000m|\
```
openwrt-ai, line 58: Moving `cmcc,rax3000m` to the `wan` block (together with `label = "wan"` on `gmac1`) renames the WAN netdev from `eth1` to `wan`, which invalidates every existing `/etc/config/network` on this device. Nothing in the series bumps the compat version, so sysupgrade will not warn and the router comes up without WAN after a routine upgrade.

The in-tree precedent for exactly this kind of rename pairs both halves:

- [`Device/bananapi_bpi-r4-common`](https://github.com/openwrt/openwrt/blob/41669139420ad9407860de6a9eab4d5de57d1a57/target/linux/mediatek/image/filogic.mk#L698-L699) — `DEVICE_COMPAT_VERSION := 1.1` + "The non-switch ports were renamed to match the board/case labels"
- [`Device/mercusys_mr85x`](https://github.com/openwrt/openwrt/blob/41669139420ad9407860de6a9eab4d5de57d1a57/target/linux/mediatek/image/filogic.mk#L2688-L2690) — `DEVICE_COMPAT_VERSION := 1.1` + "Ethernet port names have been updated due to DSA conversion."

…each with a matching entry in [`05_compat-version`](``https://github.com/openwrt/openwrt/blob/41669139420ad9407860de6a9eab4d5de57d1a57/target/linux/mediatek/filogic/base-files/etc/board.d/05_compat-version#L11-L18).``

The commit message's "users will need to update their WAN interface configuration after this change" is precisely what `DEVICE_COMPAT_MESSAGE` exists to surface at upgrade time — please add `DEVICE_COMPAT_VERSION`/`DEVICE_COMPAT_MESSAGE` to `Device/cmcc_rax3000m` and `cmcc,rax3000m` to `05_compat-version` in this same commit.

---

target/linux/mediatek/dts/mt7981b-cmcc-rax3000m-ubi.dtso
```diff
@@ -0,0 +1,178 @@
...
+				};
+			};
+		};
+	};
+
+	fragment@6 {
+		target = <&wifi>;
+		__overlay__ {
+			nvmem-cells = <&eeprom_factory_0>;
+			nvmem-cell-names = "eeprom";
```
openwrt-ai, line 163: The `band@0`/`band@1` children are merged into a `wifi` node that declares no `#address-cells`/`#size-cells`: [`117-complete-mt7981b-dtsi.patch`](``https://github.com/openwrt/openwrt/blob/41669139420ad9407860de6a9eab4d5de57d1a57/target/linux/mediatek/patches-6.18/117-complete-mt7981b-dtsi.patch#L574-L580)`` only adds the `wifi:` label and `memory-region`/`pinctrl` properties, and the node lives under `soc { #address-cells = <2>; #size-cells = <2>; }`, so the single-cell `reg = <0>` / `reg = <1>` is parsed against the wrong cell count.

Every one of the 30 in-tree mt7981 DTS files that adds `band@N` declares the cells on `&wifi` first — e.g. [`mt7981b-qihoo-360t7-ubi.dts:84-98`](https://github.com/openwrt/openwrt/blob/41669139420ad9407860de6a9eab4d5de57d1a57/target/linux/mediatek/dts/mt7981b-qihoo-360t7-ubi.dts#L84-L98).

```suggestion
		__overlay__ {
			#address-cells = <1>;
			#size-cells = <0>;

			nvmem-cells = <&eeprom_factory_0>;
			nvmem-cell-names = "eeprom";
```

The same is missing from the two `band@` blocks this PR adds in `mt7981b-cmcc-rax3000m-emmc.dtso` (line 124) and `mt7981b-cmcc-rax3000m-nand.dtso` (line 155). Alternatively, declaring the two properties once on `&wifi` in the shared `mt7981b-cmcc-rax3000m.dts` covers all three overlays and matches how the sibling boards do it.

---

openwrt-ai (commented):
Re-reviewed after the force-push (`f25fc0fb` → `41669139`). The series was restructured onto current `main` and now has 4 commits; the UBI layout became a third `DEVICE_DTS_OVERLAY` on the existing `cmcc_rax3000m` instead of a separate `Device/cmcc_rax3000m-ubi`, so I treated this as a fresh review. All previously flagged threads are resolved and the earlier findings (restored `DEVICE_DTC_FLAGS`/`DEVICE_DTS_LOADADDR`/`KERNEL_LOADADDR`, `partition@180000`, U-Boot root `compatible`, SPI0 drive strength, `NAME`/`model` captions) hold on the new head.

The switch from a separate device to an overlay is what most of the new comments follow from — see inline on `437-add-cmcc_rax3000m.patch` and `02_network`.

## Commit checks

- e4ed520 "mediatek: filogic: cmcc rax3000m: add all-in-UBI layout" — step 5 of the flash instructions says to "perform a final sysupgrade using the `openwrt*-ubi-squashfs-sysupgrade.itb` image", but with the UBI layout added as a `DEVICE_DTS_OVERLAY` on `Device/cmcc_rax3000m` rather than as its own device, no `-ubi-` image is produced; the file is `openwrt-mediatek-filogic-cmcc_rax3000m-squashfs-sysupgrade.itb`. Same wording is in the PR body. (The `openwrt*-ubi-initramfs-recovery-installer.itb` in step 3 comes from your external installer repo and is fine.)

The other three commit messages match their diffs. CI on `41669139` is green.

---

package/boot/uboot-mediatek/patches/437-add-cmcc_rax3000m.patch
andros-ua: fixed

openwrt-ai (commented):
Re-reviewed after the force-push (`41669139` → `fac592f3`, series rebased, same 4 commits reordered). The inter-diff is small and every one of my three previous inline findings is fixed — verified each:

- `bootfile` / `bootfile_upg` in `defenvs/cmcc_rax3000m-ubi_env` now use the unsuffixed `openwrt-mediatek-filogic-cmcc_rax3000m-{initramfs-recovery,squashfs-sysupgrade}.itb`, matching what `Device/cmcc_rax3000m` actually builds and the sibling eMMC/NAND defenvs in the same patch. (`bootfile_bl2`/`bootfile_fip` keep the `-ubi-` infix, consistent with the `-emmc-`/`-nand-` infix the sibling envs use; the DDR-suffix mismatch you declared out of scope is unchanged.)
- `#address-cells = <1>` / `#size-cells = <0>` added once on `&wifi` in the shared `mt7981b-cmcc-rax3000m.dts`, which covers all three overlays — same form as [`mt7981b-qihoo-360t7-ubi.dts:84-88`](https://github.com/openwrt/openwrt/blob/fac592f3c165e8001cbe3b3f59fef8e38598624d/target/linux/mediatek/dts/mt7981b-qihoo-360t7-ubi.dts#L84-L88).
- The WAN rename is now paired with `DEVICE_COMPAT_VERSION := 1.1` + `DEVICE_COMPAT_MESSAGE` on `Device/cmcc_rax3000m` and `cmcc,rax3000m` in [`05_compat-version`](``https://github.com/openwrt/openwrt/blob/fac592f3c165e8001cbe3b3f59fef8e38598624d/target/linux/mediatek/filogic/base-files/etc/board.d/05_compat-version#L11-L18),`` both in the same commit, matching the `bananapi_bpi-r4` / `mercusys_mr85x` precedent. Insertion points are alphabetical in both `02_network` and `05_compat-version`, and `ucidef_set_interfaces_lan_wan "lan1 lan2 lan3" wan` matches the three switch ports plus `gmac1` in the DTS.

CI on `fac592f3` is green (23 checks, all success/skipped).

## Commit checks

- fac592f "mediatek: filogic: cmcc rax3000m: add all-in-UBI layout" — the **"OpenWrt U-Boot layout"** table lists the mtd indices in descending order (`mtd4` = `bl2` … `mtd0` = `ubi`), but the partitions are registered in DT order, so on a running NAND device `/proc/mtd` reports `mtd0` = `bl2`, `mtd1` = `u-boot-env`, `mtd2` = `factory`, `mtd3` = `fip`, `mtd4` = `ubi` — see [`mt7981b-cmcc-rax3000m-nand.dtso:80-129`](https://github.com/openwrt/openwrt/blob/fac592f3c165e8001cbe3b3f59fef8e38598624d/target/linux/mediatek/dts/mt7981b-cmcc-rax3000m-nand.dtso#L80-L129). The second table in the same message ("OpenWrt U-Boot UBI layout") is in ascending order and matches the new `-ubi.dtso` exactly, so the first one looks like an authoring slip. In the same table the `ubi` size is given as `07000000`, but that partition is `reg = <0x580000 0x7200000>` (2 MiB larger). Same two tables are in the PR body.
- 3b0e8c5 "mediatek: filogic: cmcc rax3000m: label WAN port" — nit: the body describes the DTS `label` and the `02_network` move, but the commit now also adds `DEVICE_COMPAT_VERSION`/`DEVICE_COMPAT_MESSAGE` and the `05_compat-version` entry (3 of the 4 files touched). The loose "Note: users will need to update their WAN interface configuration after this change" undersells it — worth saying that the compat version is bumped so sysupgrade refuses and surfaces the message, rather than leaving it as an informal note.

The other two commit messages match their diffs.

---

target/linux/mediatek/dts/mt7981b-cmcc-rax3000m-ubi.dtso
andros-ua: fixed

target/linux/mediatek/filogic/base-files/etc/board.d/02_network
andros-ua: Added

openwrt-ai (commented):
Re-reviewed after the force-push (`fac592f3` → `e87d4c6b`). The tree is byte-identical to the one I reviewed at `fac592f3` — `git diff fac592f3..e87d4c6b` is empty. The only changes are commit reordering ("label WAN port" moved from 2nd to last) and two commit-message edits, both in response to my previous commit checks. One of them resolves the point, the other went the wrong way.

Resolved: `3b0e8c5` → `e87d4c6` "label WAN port" now says *"Bump the compat version so sysupgrade shows the warning message."* instead of the loose "users will need to update their WAN interface configuration" note — that accurately covers the `DEVICE_COMPAT_VERSION` / `DEVICE_COMPAT_MESSAGE` and `05_compat-version` hunks, which are 3 of the 4 files the commit touches.

CI on `e87d4c6b` is green (FormalityCheck + build-pr-profile success; the kernel/package builds carried over from `fac592f3`, which is the same tree).

## Commit checks

- f711a2f "mediatek: filogic: cmcc rax3000m: add all-in-UBI layout" — half of the previous table finding is fixed, the other half regressed. The `ubi` size in the **"OpenWrt U-Boot layout"** table is now `07200000`, matching `reg = <0x580000 0x7200000>`. But the mtd-index ordering was harmonized in the wrong direction: rather than fixing the first table, the **"OpenWrt U-Boot UBI layout"** table (which was correct at `fac592f3` with `mtd0` = `bl2`, `mtd1` = `ubi`) was flipped to `mtd1` = `bl2`, `mtd0` = `ubi`. Both tables are now descending.

  `fixed-partitions` registers mtd devices in DT node order, and these tables reproduce `/proc/mtd`, whose header row (`dev:    size   erasesize  name`) is copied verbatim and which always prints `mtd0` first. So:

  - NAND layout — [`mt7981b-cmcc-rax3000m-nand.dtso:80-140`](https://github.com/openwrt/openwrt/blob/e87d4c6be9dbaae5cfbfa87b2fea5af23bdc71fa/target/linux/mediatek/dts/mt7981b-cmcc-rax3000m-nand.dtso#L80-L140) declares `bl2`, `u-boot-env`, `factory`, `fip`, `ubi` in that order, so a running device reports `mtd0` = `bl2` … `mtd4` = `ubi`:

    ```
    | dev:    size   erasesize  name       |
    | mtd0: 00100000 00020000 "bl2"        |
    | mtd1: 00080000 00020000 "u-boot-env" |
    | mtd2: 00200000 00020000 "factory"    |
    | mtd3: 00200000 00020000 "fip"        |
    | mtd4: 07200000 00020000 "ubi"        |
    ```

  - UBI layout — [`mt7981b-cmcc-rax3000m-ubi.dtso:80-87`](https://github.com/openwrt/openwrt/blob/e87d4c6be9dbaae5cfbfa87b2fea5af23bdc71fa/target/linux/mediatek/dts/mt7981b-cmcc-rax3000m-ubi.dtso#L80-L87) declares `partition@0` = `bl2` then `partition@100000` = `ubi`, so:

    ```
    | dev:    size   erasesize  name |
    | mtd0: 00100000 00020000 "bl2"  |
    | mtd1: 07f00000 00020000 "ubi"  |
    ```

  The PR body carries the same two tables and needs the same correction. (If these were pasted from an actual device and it really printed them in this order, say so and I'll drop it — but nothing in the DTS explains a reversed numbering.)

The other three commit messages match their diffs.

---

andros-ua:
> * nothing in the DTS explains a reversed numbering

It's how overlays get applied.

```
[    2.357628] FIT:          flat_dt sub-image 0x0059e000..0x005a4ed8 "fdt-1" (ARM64 OpenWrt cmcc_rax3000m device tree blob)
[    2.368689] FIT:          flat_dt sub-image 0x005a5000..0x005a5e9d "fdt-mt7981b-cmcc-rax3000m-emmc" (ARM64 OpenWrt cmcc_rax3000m device tree overlay mt7981b-cmcc-rax3000m-emmc)
[    2.384524] FIT:          flat_dt sub-image 0x005a6000..0x005a6f81 "fdt-mt7981b-cmcc-rax3000m-nand" (ARM64 OpenWrt cmcc_rax3000m device tree overlay mt7981b-cmcc-rax3000m-nand)
[    2.400355] FIT:          flat_dt sub-image 0x005a7000..0x005a7fe8 "fdt-mt7981b-cmcc-rax3000m-ubi" (ARM64 OpenWrt cmcc_rax3000m device tree overlay mt7981b-cmcc-rax3000m-ubi)

```
```shell
root@OpenWrt:~# cat /proc/mtd
dev:    size   erasesize  name
mtd0: 07f00000 00020000 "ubi"
mtd1: 00100000 00020000 "bl2"
```

andros-ua:
@dangowrt, could you take a look at this all-in-UBI layout? It's friendlier to the build system's disk footprint since it avoids generating separate images per partition scheme.

Lexeyko:
I fully support this PR. I've compiled and tested these changes on my device, and everything runs smoothly and stably without any regressions. The implementation is clean and solves the issue perfectly. Thanks for the contribution, looking forward to seeing this merged!
Tested on: CMCC RAX3000M DDR4, CMCC RAX3000Me DDR4, CMCC RAX3000Me DDR3.

openwrt-ai (commented):

---

andros-ua:
@jonasjelonek, could you take a look when you have a chance? It applies cleanly and openwrt-ai is happy.

package/boot/uboot-mediatek/Makefile
```diff
@@ -305,6 +305,30 @@ define U-Boot/mt7981_cmcc_rax3000m-nand-ddr4
   DEPENDS:=+trusted-firmware-a-mt7981-spim-nand-ddr4
 endef
 
+define U-Boot/mt7981_cmcc_rax3000m-ubi-ddr3
+  NAME:=CMCC RAX3000M (UBI)
```
jonasjelonek: Shouldn't that name carry the DDR3/DDR4 too so it's distinguishable at the name?

target/linux/mediatek/dts/mt7981b-cmcc-rax3000m.dts
```diff
@@ -90,6 +90,7 @@
 	gmac1: mac@1 {
 		compatible = "mediatek,eth-mac";
 		reg = <1>;
+		label = "wan";
```
jonasjelonek: Shouldn't this use `openwrt,netdev-name` instead? There was an upstream discussion about that, but don't have the link right now.

target/linux/mediatek/dts/mt7981b-cmcc-rax3000m.dts
```diff
@@ -302,6 +303,9 @@
 
 &wifi {
 	status = "okay";
+
+	#address-cells = <1>;
+	#size-cells = <0>;
```
jonasjelonek, line 308: Does this change follow from any other change? I don't see that mentioned in the commit message.

package/boot/uboot-mediatek/Makefile
andros-ua: The naming convention was inherited from the existing -emmc/-nand targets, but I'm happy to change it. If we go this route, would you prefer:

- CMCC RAX3000M (UBI DDR3)
- CMCC RAX3000M (UBI) DDR3
- CMCC RAX3000M DDR3 (UBI)

And should the existing -emmc-ddr3/-emmc-ddr4/-nand-ddr3/-nand-ddr4 targets get the same treatment, as a separate commit, for consistency?

target/linux/mediatek/dts/mt7981b-cmcc-rax3000m.dts
andros-ua: I've been following this approach since https://github.com/openwrt/openwrt/pull/19534#discussion_r2262796618, and it does look like the currently-used convention. Happy to switch to `openwrt,netdev-name` if that's actually the preferred direction going forward — just let me know and I'll fix it right away.

target/linux/mediatek/dts/mt7981b-cmcc-rax3000m.dts
andros-ua: Thanks for catching that — I've updated the commit message to clarify the dependency.
```
Since the per-band entries are added as addressed band@0/band@1
child nodes, declare #address-cells = <1> and #size-cells = <0> on
the shared &wifi node in mt7981b-cmcc-rax3000m.dts.
```

openwrt-ai (commented):
## Commit checks

- 818e340 "mediatek: filogic: cmcc rax3000m: add all-in-UBI layout" — this force-push moved "use mac-address nvmem cells" from first to last in the series, so this commit now introduces `band@0`/`band@1` (with single-cell `reg`) under `&wifi` in `mt7981b-cmcc-rax3000m-ubi.dtso` while [`&wifi` in `mt7981b-cmcc-rax3000m.dts`](https://github.com/openwrt/openwrt/blob/818e3407fbca96319006831d7484612e1da55109/target/linux/mediatek/dts/mt7981b-cmcc-rax3000m.dts#L303-L305) still lacks the `#address-cells = <1>` / `#size-cells = <0>` that 4a03a7d's own message says those nodes require. Put 4a03a7d back before 818e340 so no intermediate commit is broken.

---

andros-ua:
> * Put 4a03a7d back before 818e340 so no intermediate commit is broken.

Fixed

openwrt-ai (commented):

---

package/boot/uboot-mediatek/Makefile
jonasjelonek: I'd probably pick a modified version of the first `(DDR3, UBI)` or similar. Though the distinction is given in the recipe name, I think it should be visible in the name too.

target/linux/mediatek/dts/mt7981b-cmcc-rax3000m.dts
jonasjelonek: This was discussed quite often, see https://github.com/openwrt/openwrt/pull/20475 and https://lore.kernel.org/netdev/20240709124503.pubki5nwjfbedhhy@skbuf/

If the driver doesn't read `label`, I think you should use `openwrt,netdev-name` instead.

package/boot/uboot-mediatek/Makefile
andros-ua: Thanks - applied.

target/linux/mediatek/dts/mt7981b-cmcc-rax3000m.dts
andros-ua: Thanks for the links and the explanation. Switched to openwrt,netdev-name and run-tested successfully on the device.

openwrt-ai (commented):

---

Lexeyko:
Could we prioritize merging basic device support first before focusing on cosmetic code cleanups? This UBI partitioning scheme is already well-proven and used on other target devices without any issues, regardless of RAM configuration (DDR3 or DDR4). Merging initial support now would give users official firmware support, while further code refactoring and standard cleanups can easily be addressed in follow-up commits.

Lexeyko:
I've tested the build with the latest fixes applied. Everything is working as expected and running smoothly.

jonasjelonek:

andros-ua:
Thanks for your time on this, @jonasjelonek !

ptpt52:
@andros-ua  why change netdev from eth1 to wan, and this is not working.

# Bug Report: `04_set_netdev_label` Fails to Rename Network Interfaces Due to Race Condition

## Summary

The preinit script `target/linux/mediatek/filogic/base-files/lib/preinit/04_set_netdev_label`
fails to rename network interfaces (e.g., `eth1` → `wan`) because it runs **before the
kernel has finished registering those network devices**. This is a race condition between
the preinit hook execution and the MediaTek Ethernet driver's device initialization.

---

## Affected Platform

- **Target:** `target/linux/mediatek/filogic`
- **Device tested:** CMCC RAX3000M (`cmcc,rax3000m`)
- **SoC:** MediaTek MT7981B
- **Kernel:** Linux 6.18.44

---

## Description

The script `04_set_netdev_label` is registered to the `preinit_main` hook:

```sh
boot_hook_add preinit_main set_netdev_labels
```

It iterates over `/sys/class/net/*` and renames any interface whose
`of_node/openwrt,netdev-name` DT property differs from its current name:

```sh
set_netdev_labels() {
    for dir in /sys/class/net/*; do
        [ -r "$dir/of_node/openwrt,netdev-name" ] || continue
        read -r label < "$dir/of_node/openwrt,netdev-name"
        netdev="${dir##*/}"
        [ "$netdev" = "$label" ] && continue
        ip link set "$netdev" name "$label"
    done
}
```

The DTS for the affected device (`mt7981b-cmcc-rax3000m.dts`) correctly declares:

```dts
gmac1: mac@1 {
    compatible = "mediatek,eth-mac";
    reg = <1>;
    openwrt,netdev-name = "wan";   /* eth1 should be renamed to "wan" */
    phy-mode = "gmii";
    phy-handle = <&int_gbe_phy>;
};
```

However, the rename **never happens** during preinit.

---

## Root Cause: Race Condition

The `preinit_main` hooks begin executing at approximately **t=3.621s** after boot,
but the MediaTek Ethernet driver does not finish registering `eth1` until **t=4.465s**.

Because `04_set_netdev_label` runs **before `eth1` appears in `/sys/class/net/`**,
it finds no interface with an `of_node/openwrt,netdev-name` property and silently exits
without performing any renames.

As a result, `netifd` starts up with `eth1` still named `eth1`. The interface is only
renamed to `wan` much later — at **t≈196s** — when `netifd` itself triggers the rename
during a WAN link-down/link-up cycle.

---

## Log Evidence

### Boot-time kernel messages (via `dmesg`)

```
[    3.329390] kmodloader: loading kernel modules from /etc/modules-boot.d/*
[    3.456499] mtk_soc_eth 15100000.ethernet: legacy DT: using hard-coded SRAM offset.
[    3.621641] init: - preinit -                     ← preinit_main hooks start here
[    4.455118] mtk_soc_eth 15100000.ethernet eth0: mediatek frame engine at 0xffffffc081d00000
[    4.465297] mtk_soc_eth 15100000.ethernet eth1: mediatek frame engine at 0xffffffc081d00000
               ↑ eth1 appears ~0.85s AFTER preinit_main starts
               ↑ 04_set_netdev_label already ran and found nothing to rename

[    4.823285] mt7530-mdio mdio-bus:1f lan3 (uninitialized): PHY [mt7530-0:00] ...
[    4.849007] mt7530-mdio mdio-bus:1f lan2 (uninitialized): PHY [mt7530-0:01] ...
[    4.874976] mt7530-mdio mdio-bus:1f lan1 (uninitialized): PHY [mt7530-0:02] ...
[   12.342600] procd: - init -

   ... eth1 remains named "eth1" throughout this entire period ...

[   29.453605] mtk_soc_eth 15100000.ethernet eth1: PHY [mdio-bus:00] driver [...] (irq=POLL)
[   29.492842] mtk_soc_eth 15100000.ethernet eth1: Link is Up - 1Gbps/Full
               ↑ netifd brought up "eth1", NOT "wan"

[  196.919320] mtk_soc_eth 15100000.ethernet wan: renamed from eth1
               ↑ rename finally happens here, ~193s too late, triggered by netifd on link-down
```

### Confirming `eth1`'s `of_node` content at runtime

The DT property IS correctly present and readable on the running device:

```sh
$ cat /sys/class/net/wan/of_node/openwrt,netdev-name | od -c
0000000   w   a   n  \0
0000004
```

The symlink confirms `wan` (the post-netifd-rename interface) points to the right DT node:

```sh
$ readlink /sys/class/net/wan/of_node
../../../../../../firmware/devicetree/base/soc/ethernet@15100000/mac@1
```

So the DT is correct and the `of_node` is accessible — the sole problem is **timing**.

---

## Timeline Diagram

```
t=3.329s  kmodloader starts loading drivers
t=3.621s  preinit_main begins
              └─ 04_set_netdev_label runs → iterates /sys/class/net/*
                                          → eth1 NOT present yet → no rename done
t=4.465s  eth1 registered by mtk_soc_eth driver  ← TOO LATE
t=4.823s  PHY drivers bind to lan1/lan2/lan3
t=12.34s  procd/init starts
t=29.45s  netifd brings up eth1 (still named "eth1")
t=196.9s  netifd renames eth1 → wan on first link-down event
```

---

## Impact

- The interface `eth1` is used by `netifd` under its kernel-assigned name instead of
  the DT-specified name `wan`.
- The rename eventually happens but only on the **first WAN link-down event**, meaning
  the system operates with the wrong interface name for the first few minutes after boot.
- Any scripts or configuration that depends on `wan` being available immediately after
  boot may malfunction.

---

## Suggested Fix

The `set_netdev_labels` function should wait for the expected interfaces to appear in
`/sys/class/net/` before attempting to rename them, or the rename logic should be moved
to a later hook point (after drivers have fully initialized).

One approach — add a polling wait inside `set_netdev_labels`:

```sh
set_netdev_labels() {
    local dir label netdev deadline

    # Wait up to 10s for DT-driven netdevs to appear
    deadline=$(($(cut -d. -f1 /proc/uptime) + 10))
    while [ "$(cut -d. -f1 /proc/uptime)" -lt "$deadline" ]; do
        local found=0
        for dir in /sys/class/net/*; do
            [ -r "$dir/of_node/openwrt,netdev-name" ] && { found=1; break; }
        done
        [ "$found" = "1" ] && break
        sleep 0.2
    done

    for dir in /sys/class/net/*; do
        [ -r "$dir/of_node/openwrt,netdev-name" ] || continue
        read -r label < "$dir/of_node/openwrt,netdev-name"
        netdev="${dir##*/}"
        [ "$netdev" = "$label" ] && continue
        ip link set "$netdev" name "$label"
    done
}
```

Alternatively, the rename could be driven by a `hotplug.d/net/` script that fires when
each interface is registered by the kernel, which is inherently race-free.

---

## References

- Script: `target/linux/mediatek/filogic/base-files/lib/preinit/04_set_netdev_label`
- DTS: `target/linux/mediatek/dts/mt7981b-cmcc-rax3000m.dts`, line 93
- Related scripts: `package/base-files/files/lib/preinit/10_indicate_preinit`

andros-ua:
> why change netdev from eth1 to wan, and this is not working.

According to your log, the renaming was applied and worked as intended.

For bug reports, please use the GitHub issue tracker.
For general discussion, the OpenWrt forum is the best place.

1715173329:
I would say renaming this is just a waste of everyone's time. WAN port is connected to a dedicated PHY and not belongs to the switch, so it should be eth1. There's no real benefit from this change but it's cleary prevents all users upgrading from old firmware easily.

Lexeyko:
It’s better to spend time fixing this once in a saved config than having to remember every time which of eth0, eth1, eth2... is WAN, LAN, or whatever. It’s just much easier when things are named in human language rather than machine terms. Once androids start using routers, maybe eth1 — or even 01100101 01110100 01101000 00110001 — will be convenient for them.

andros-ua:
@1715173329 @ptpt52 
I'm sorry you feel that way about this change.
There were no plans to backport it, so users should only run into the WAN interface renaming with the next major stable release, giving them time to prepare before then.

PR #24576, opened 2026-08-05
tools: gnulib: drop the gl_-prefixed macros an older install left behind

JuliusBairaktaris:
An up-to-date tree fails its next build in `tools/elfutils`:

```
../libgnu/ctype.h:58:7: error: operator '||' has no left operand
../libgnu/ctype.h:813:5: error: #if with no expression
make[7]: *** [Makefile:3159: libgnu_la-argp-ba.lo] Error 1
    ERROR: tools/elfutils failed to build.
```

`tools/elfutils` is in `tools-y`, so this fails the whole tree, and it does
not reproduce from scratch — only on a tree that was also built between
c820f097e0be and cda598a59528.

### What is actually there

`Host/Install` copies `m4/*.m4` into `$(STAGING_DIR_HOST)/share/aclocal`;
`Host/Uninstall` removes only `bin/gnulib-tool` and `share/gnulib`. Nothing
deletes the macros, so the `gl_`-prefixed copies from the reverted prefix
scheme are still there next to the current ones:

```
$ ls -l --time-style=+%F staging_dir/host/share/aclocal/*.m4 | awk '{print $6}' | sort | uniq -c
   1520 2026-06-10
    943 2026-08-05
$ ls staging_dir/host/share/aclocal/gl_*.m4 | wc -l
914
```

914 is exactly the macro count of `gnulib-2025.07.01~a3151d45`, and 943 is
exactly `gnulib-2026.07.04~b22f5a30`'s — the two sets are two whole gnulib
releases, one shadowing the other.

They are not interchangeable. `gl_ctype_h.m4` is `serial 9`; `ctype_h.m4` is
`serial 23` and is the one that initialises the module indicators the current
`ctype.in.h` reads:

```
$ grep -c GNULIB_ISUPPER_L staging_dir/host/share/aclocal/{ctype_h,gl_ctype_h}.m4
staging_dir/host/share/aclocal/ctype_h.m4:1
staging_dir/host/share/aclocal/gl_ctype_h.m4:0
```

With both installed, `configure` ends up with zero occurrences of
`GNULIB_ISUPPER_L`, `@GNULIB_ISUPPER_L@` is substituted with the empty string,
and the generated `libgnu/ctype.h` gets `#if` with nothing after it.

### Fix

Remove the prefixed copies in `Host/Uninstall`. `Host/Install` calls it first,
so `make tools/gnulib/{clean,compile}` repairs an affected tree — no manual
`staging_dir` surgery.

Verified here: `tools/elfutils` failed before and builds after, with
`libelf.a` and `libdw.a` installed.

### Worth a second opinion

This deletes the one leftover set that exists today. The general shape —
`Host/Uninstall` not undoing what `Host/Install` wrote — is still there, so a
future gnulib that drops or renames a macro file leaves the old one shadowing
in the same way. Removing the previous install's macros by name using the
`share/gnulib/m4` copy that `Host/Uninstall` is about to delete would cover
that too; I left it out as speculative, but say the word and I will add it.

### DIFF
--- a/tools/gnulib/Makefile
+++ b/tools/gnulib/Makefile
@@ -22,6 +22,7 @@ define Host/Install
 endef
 
 define Host/Uninstall
+	rm -f $(STAGING_DIR_HOST)/share/aclocal/gl_*.m4
 	rm -rf $(STAGING_DIR_HOST)/bin/gnulib-tool $(STAGING_DIR_HOST)/share/gnulib
 endef
 


tools/gnulib/Makefile
```diff
@@ -22,6 +22,7 @@ define Host/Install
 endef
 
 define Host/Uninstall
+	rm -f $(STAGING_DIR_HOST)/share/aclocal/gl_*.m4
```
openwrt-ai, line 25: The `gl_*.m4` glob covers only trees built inside the `c820f097e0be`..`cda598a59528` window. The "future gnulib renames a macro file" case the PR body calls speculative has already happened once, unprefixed, and this line does not catch it: [`310-modules-cond.patch`](https://github.com/openwrt/openwrt/blob/7aef87d38c16e0da6cde5f02d509c5c0bef2edaf/tools/gnulib/patches/310-modules-cond.patch#L1-L2) deletes `m4/cond.m4` and re-adds it as `m4/glthread_cond.m4`, so current gnulib no longer installs a `cond.m4`.

A tree whose last gnulib install predates `c820f097e0be` (unprefixed gnulib 2025.01/2025.07, which still shipped `m4/cond.m4`) therefore keeps a stale `share/aclocal/cond.m4` forever — the exact automake clash that `a54eb7297f81` set out to remove, since [`ACLOCAL_INCLUDE` at host-build.mk:169](https://github.com/openwrt/openwrt/blob/7aef87d38c16e0da6cde5f02d509c5c0bef2edaf/include/host-build.mk#L169) puts `share/aclocal` ahead of `share/aclocal-<ver>`. Same for anything else dropped by the stable-202501 → stable-202507 jump in `837f5eaae21f`.

Is that intentionally out of scope? The name-based removal you sketched (walk `share/gnulib/m4/*.m4` before the `rm -rf`) would cover both this and the `gl_` set with one mechanism, and it seems worth having in the same commit rather than leaving a second class of leftovers that only shows up as another `#if with no expression` later.

Unrelated to the above, the change itself looks right and self-applying: editing this Makefile changes the `find_md5 ${CURDIR}` component of `HOST_STAMP_PREPARED`, which re-runs prepare → build → `Host/Install`, so an affected tree is repaired by a plain `make tools/install` — the `make tools/gnulib/{clean,compile}` in the commit message/PR body is more than strictly needed.

---

openwrt-ai (commented):
Reviewed 1 new commit; commit message matches the diff. One inline note on the scope of the glob.

---

mcprat:
Hey again, I'm sorry this ended up being a problem, I wish Christian pinged me at the time,
but this is not a proper fix, adding this line will make it completely useless once it is run the first time because it will never have to be run again...

please just do
`make dirclean`
to clean your staging directory...

PR #24578, opened 2026-08-05
mac80211: ath11k: stop a firmware crash from rebooting the SoC on AHB

JuliusBairaktaris:
On IPQ8074 a firmware assert does not recover — it panics and reboots the SoC.
Any user on an ipq807x board hits this the first time the WCSS firmware asserts,
and the box is a router, so the visible symptom is a spontaneous reboot with
nothing useful in the logs.

```
Unable to handle kernel read from unreadable memory at virtual address 0
pc : ath11k_hal_srng_access_begin+0xc/0x60 [ath11k]
Call trace:
 ath11k_hal_srng_access_begin+0xc/0x60 [ath11k] (P)
 ath11k_dp_rx_process_mon_rings+0xa0/0x5d4 [ath11k]
 ath11k_dp_service_srng+0x1f4/0x348 [ath11k]
 ath11k_ahb_ext_grp_napi_poll+0x34/0xd4 [ath11k_ahb]
```

`ath11k_core_reconfigure_on_crash()` frees the data path and lets
`ath11k_hal_srng_clear()` memset the ring list while the DP NAPI is still
polling. It used to disable the interrupts first, until
d455e805de70 ("wifi: ath11k: rearrange IRQ enable/disable in reset path")
moved the disable into `ath11k_core_reset()` — which AHB parts never reach on a
real crash, since `reset_work` is queued only from `mhi.c` and from the debugfs
`hw-restart` handler. Their recovery goes QMI server exit → `restart_work` →
`ath11k_core_reconfigure_on_crash()`, and nothing disables the interrupts along
it.

`951` puts the disable back on the crash path. `950` is its prerequisite:
`ath11k_hif_ce_irq_disable()` is a no-op on AHB because
`ath11k_ahb_hif_ops_ipq8074` never set the `ce_irq_enable`/`ce_irq_disable`
pair, so the CE half of the quiesce does nothing today. wcn6750 is unaffected —
it uses the pcic ops, which implement the pair already.

### Upstream status

Both patches are merged in ath-next:

| patch | upstream commit |
|---|---|
| `950` | [`a7ab5c835e42`](https://git.kernel.org/pub/scm/linux/kernel/git/ath/ath.git/commit/?id=a7ab5c835e4253b6577e4137367c05bfddd79fa2) wifi: ath11k: implement CE interrupt enable/disable for AHB |
| `951` | [`f7a74e131d3f`](https://git.kernel.org/pub/scm/linux/kernel/git/ath/ath.git/commit/?id=f7a74e131d3f0de04335dbe278bfb82bdafcdb3e) wifi: ath11k: disable interrupts during firmware crash recovery |

The patch files carry the merged commits verbatim — same code, same commit
messages including the review trailers — refreshed against backports 7.2. They
drop out on the next backports bump that picks the fixes up.

### Testing

Xiaomi AX3600 (IPQ8074 hw2.0), this package, `CONFIG_ATH11K_NSS_SUPPORT` off so
ath11k is stock and no out-of-tree modules are loaded (`grep -c ath11k_nss
/proc/kallsyms` = 0, `lsmod | grep -c qca` = 0). Firmware
`WLAN.HK.2.12-01460-QCAHKSWPL_SILICONZ-1`. Crashes driven with the debugfs
`simulate_fw_crash` `assert` trigger, panics captured via ramoops:

| | result |
|---|---|
| before | 1 assert → panic in `ath11k_dp_rx_process_mon_status()`, SoC reboot |
| after | 5 asserts → 5 recoveries, `boot_id` unchanged, both radios back up |

The debugfs `hw-restart` trigger does **not** reproduce it — that path goes
through `ath11k_core_reset()`, the one place the disable still happens. Only a
real firmware assert does, which is why this survived so long.


package/kernel/mac80211/patches/ath11k/951-wifi-ath11k-disable-interrupts-during-firmware-crash-recovery.patch
```diff
@@ -0,0 +1,73 @@
...
++	 * ath11k_core_reset() already disabled the interrupts on the reset
++	 * path; only the firmware crash path reaches here with them live.
++	 */
++	if (!ab->is_reset) {
++		ath11k_hif_irq_disable(ab);
++		ath11k_hif_ce_irq_disable(ab);
++	}
++
+ 	ath11k_dp_pdev_free(ab);
+ 	ath11k_cfr_deinit(ab);
```
openwrt-ai, line 69: This context line does not exist in the source this package builds. `ath11k_cfr_deinit()` is not in mainline ath11k (checked `v6.18.41`; `PKG_VERSION` here is 6.18.39) and no other patch under `package/kernel/mac80211/patches/ath11k/` introduces it — it only exists in the out-of-tree NSS/CFR stack of the fork this was generated on.

Consequence: the hunk applies only by fuzz. Against a stock `drivers/net/wireless/ath/ath11k/core.c`:

```
$ patch -p1 --dry-run < 951-...patch
Hunk #1 succeeded at 2304 with fuzz 2 (offset -30 lines).
$ patch -p1 --dry-run -F0 < 951-...patch
Hunk #1 FAILED at 2334.
1 out of 1 hunk FAILED
```

It happens to land in the right spot today because the leading context (`mutex_lock(&ab->core_lock);` / `ath11k_thermal_unregister(ab);`) still matches exactly, but the trailing context and the `@@ -2334,6 +2334,16 @@` counts describe a different tree, so any churn there turns this into a reject instead of a correctly-placed hunk. Please regenerate the patch against a stock checkout — `make package/mac80211/refresh` — so the context and line counts match this package.

Related: the commit message states the fix was reproduced "against this package, with CONFIG_ATH11K_NSS_SUPPORT off so ath11k is stock". `ath11k_cfr_deinit(ab)` is an unconditional call in that function, not something a Kconfig symbol removes, so the tree the patch was cut from was not stock ath11k. Were the before/after crash results measured on a stock `kmod-ath11k` build?

---

package/kernel/mac80211/patches/ath11k/951-wifi-ath11k-disable-interrupts-during-firmware-crash-recovery.patch
```diff
@@ -0,0 +1,73 @@
...
++++ b/drivers/net/wireless/ath/ath11k/core.c
+@@ -2334,6 +2334,16 @@ static int ath11k_core_reconfigure_on_crash(struct ath11k_base *ab)
+ 
+ 	mutex_lock(&ab->core_lock);
+ 	ath11k_thermal_unregister(ab);
++
++	/*
++	 * ath11k_core_reset() already disabled the interrupts on the reset
++	 * path; only the firmware crash path reaches here with them live.
++	 */
```
openwrt-ai, line 62: nit: block-comment style deviates from the file being patched. `drivers/net/wireless/` uses the networking style where the text starts on the `/*` line — the two multi-line comments already in `ath11k_core_reset()` do exactly that, e.g. [core.c:2513-2514](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/ath/ath11k/core.c#L2513-L2514) and [core.c:2528-2530](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/ath/ath11k/core.c#L2528-L2530). `checkpatch.pl` flags the current form as `NETWORKING_BLOCK_COMMENT_STYLE`, which matters since this is heading upstream.

The rewrap below keeps the line count at 4 so the `@@` header and diffstat stay valid.

```suggestion
+	/* ath11k_core_reset() already disabled the interrupts on the
+	 * reset path; only the firmware crash path reaches here with
+	 * them still live.
+	 */
```

---

package/kernel/mac80211/patches/ath11k/950-wifi-ath11k-implement-CE-interrupt-enable-disable-for-AHB.patch
```diff
@@ -0,0 +1,58 @@
...
++++ b/drivers/net/wireless/ath/ath11k/ahb.c
+@@ -391,12 +391,17 @@ static void ath11k_ahb_ext_irq_disable(struct ath11k_base *ab)
+ 	ath11k_ahb_sync_ext_irqs(ab);
+ }
+ 
+-static void ath11k_ahb_stop(struct ath11k_base *ab)
++static void ath11k_ahb_ce_irq_disable_sync(struct ath11k_base *ab)
+ {
+ 	if (!test_bit(ATH11K_FLAG_CRASH_FLUSH, &ab->dev_flags))
+ 		ath11k_ahb_ce_irqs_disable(ab);
```
openwrt-ai, line 36: This guard means the newly-wired `.ce_irq_disable` (line 52 below) does not actually mask anything on the path patch 951 targets.

By the time `ath11k_core_reconfigure_on_crash()` runs on a real firmware assert, `ATH11K_FLAG_CRASH_FLUSH` is already set — `ath11k_qmi_driver_event_work()` sets it on `ATH11K_QMI_EVENT_SERVER_EXIT` ([qmi.c:3235](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/ath/ath11k/qmi.c#L3235)) and `ath11k_core_reconfigure_on_crash()` only clears it further down, after the point where 951 inserts the call. So on that path `ath11k_ahb_ce_irqs_disable()` is skipped and only `ath11k_ahb_sync_ce_irqs()` + `ath11k_ahb_kill_tasklets()` run; the CE IRQ lines stay unmasked in hardware.

That is probably the right behaviour (the target is down, the register space is unsafe to touch), but it doesn't match this commit message's justification, which is "the copy engine interrupts and their tasklets therefore stay live across `rproc_shutdown()`, where the register space they touch is no longer accessible" — the masking half is exactly what the guard skips. Is the intent on the crash path only the sync + tasklet quiesce? Worth saying so explicitly, since it's the part the upstream reviewers will look at.

---

openwrt-ai (commented):
## Commit checks

- 51dfe58 "mac80211: ath11k: stop a firmware crash from rebooting the SoC on AHB" — the message says the fix was reproduced and verified "against this package, with `CONFIG_ATH11K_NSS_SUPPORT` off so ath11k is stock and no out-of-tree modules are loaded". Patch 951 carries a context line (`ath11k_cfr_deinit(ab);`) that exists only in an out-of-tree NSS/CFR stack, so the tree the patch was generated on was not this package's ath11k. See the inline comment on that line.

Patch 950 applies cleanly against stock `ath11k/ahb.c` and its diffstat/hunk counts check out; patch 951 needs a refresh. The logic itself looks sound: `ath11k_core_reset()` is the only caller of the disable pair today ([core.c:2567-2568](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/ath/ath11k/core.c#L2567-L2568)), AHB parts reach `ath11k_core_reconfigure_on_crash()` via `restart_work` without passing through it, and re-enable still happens afterwards via `ath11k_hif_start()` → `ath11k_ahb_start()` and `ath11k_hif_irq_enable()` in `ath11k_core_qmi_firmware_ready()`. Nothing on the DP/NAPI side takes `ab->core_lock`, so doing the disable under that mutex is safe. `PKG_RELEASE` bump is present and correct.

---

package/kernel/mac80211/patches/ath11k/950-wifi-ath11k-implement-CE-interrupt-enable-disable-for-AHB.patch
```diff
@@ -0,0 +1,58 @@
...
+implement the pair already.
+
+Link: https://lore.kernel.org/ath11k/20260726221908.104873-1-julius@bairaktaris.de/
+Assisted-by: Claude:claude-opus-5
+Signed-off-by: Julius Bairaktaris <julius@bairaktaris.de>
+---
+ drivers/net/wireless/ath/ath11k/ahb.c | 9 ++++++++-
+ 1 file changed, 8 insertions(+), 1 deletion(-)
+
+diff --git a/drivers/net/wireless/ath/ath11k/ahb.c b/drivers/net/wireless/ath/ath11k/ahb.c
```
openwrt-ai: `950` was left in `git format-patch` shape while `951` was converted to the tree's refreshed form in this push. It is the only file out of the 166 under `package/kernel/mac80211/patches/` that carries a `diff --git` / `index` header — plus the `-- ` / `2.53.0` trailer at lines 56-57 and untruncated function names in the `@@` headers, both of which the refreshed form drops.

The practical consequence is the hunk offsets: they are pristine-mainline offsets, not offsets for the tree this actually applies to. Three earlier patches in the same directory insert lines into `ahb.c` above these hunks:

- `910-ath11k-fix-remapped-ce-accessing-issue-on-64bit-OS.patch` — `@@ -198,12 +198,18 @@`, +6 lines, above line 391
- `923-wifi-ath11k-update-hif_and-pci_ops-for-QCN6122.patch` — `@@ -768,6 +768,18 @@`, +12 lines, above line 773
- `924-wifi-ath11k-add-multipd-support-for-QCN6122.patch` — `@@ -435,6 +435,7 @@`, +1 line, above line 773

So hunk 1 (`@@ -391,12 +391,17 @@`) really lands around 397 and hunk 2 (`@@ -773,6 +778,8 @@`) around 792. It applies with offset today because the context still matches, but that is exactly the state `951` was just fixed out of — any churn in `ahb.c` turns an offset into a reject.

Please regenerate this one against the patched tree too, so its body matches the in-tree form and the line counts describe the tree it applies to: `make package/mac80211/refresh`.

---

package/kernel/mac80211/patches/ath11k/950-wifi-ath11k-implement-CE-interrupt-enable-disable-for-AHB.patch
```diff
@@ -0,0 +1,58 @@
+From: Julius Bairaktaris <julius@bairaktaris.de>
```
openwrt-ai, line 2: nit: both new patches start straight at `From:`, with no `From <sha> Mon Sep 17 00:00:00 2001` line, so `git am` will not take them — this is what the openwrt formality bot is reporting in the conversation, and it still reports it against the current head.

`949-wifi-ath11k-fix-monitor-rx-pktlen.patch` is the in-tree precedent for a locally-authored (non-upstream) patch here: all-zero SHA on line 1, `Subject: [PATCH] ...` on line 3.

```suggestion
From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
From: Julius Bairaktaris <julius@bairaktaris.de>
```

Same applies to line 1 of `951-wifi-ath11k-disable-interrupts-during-firmware-crash-recovery.patch`, and to the `Subject:` line of both files (line 3 here, line 3 there), which is missing the `[PATCH]` prefix the other patches in this directory carry.

---

openwrt-ai (commented):
Re-reviewed at `49d670d`, diffing only what changed since `51dfe58`.

The previous blocker is resolved: `951` was regenerated and the `ath11k_cfr_deinit(ab);` context line is gone. Its hunk now matches stock ath11k exactly — `mutex_lock(&ab->core_lock);` / `ath11k_thermal_unregister(ab);` / `ath11k_dp_pdev_free(ab);` / `ath11k_spectral_deinit(ab);` / `ath11k_ce_cleanup_pipes(ab);` is what `ath11k_core_reconfigure_on_crash()` reads in 6.18.x, and the body is now in the tree's refreshed form. The commit message also gained the paragraph explaining the `ATH11K_FLAG_CRASH_FLUSH` behaviour of the CE half, which answers the earlier question on `950` — that thread can be considered addressed.

Re-checked the logic against 6.18.x with the regenerated context:

- `ab->is_reset` is set in `ath11k_core_reset()` before it powers the target up, and `ath11k_core_restart()` already branches on it, so the `if (!ab->is_reset)` guard does select the crash path and only the crash path.
- `__ath11k_ahb_ext_irq_disable()` is idempotent (it gates on `irq_grp->napi_enabled`), so an extra call on a path that already disabled cannot wedge on `napi_disable()`.
- Wiring `.ce_irq_enable` / `.ce_irq_disable` on `ath11k_ahb_hif_ops_ipq8074` also makes `ath11k_hif_ce_irq_{en,dis}able()` live in the suspend/resume paths, which were no-ops on AHB before. That looks like the intended behaviour rather than a regression — masking CE interrupts across suspend is what the pcic ops already do.
- `ath11k_ahb_ce_irqs_enable()` is defined well above the ops table, so no forward-declaration is needed.

Two things left, both on `950`, which was not touched in this push. Inline.

CI is green on this head (`Build Kernel` and `Build all core packages` both succeeded), so nothing CI-grounded.

---

JuliusBairaktaris:
Hi @robimarko — both patches are merged upstream now:

- [`a7ab5c835e42`](https://git.kernel.org/pub/scm/linux/kernel/git/ath/ath.git/commit/?id=a7ab5c835e4253b6577e4137367c05bfddd79fa2) wifi: ath11k: implement CE interrupt enable/disable for AHB
- [`f7a74e131d3f`](https://git.kernel.org/pub/scm/linux/kernel/git/ath/ath.git/commit/?id=f7a74e131d3f0de04335dbe278bfb82bdafcdb3e) wifi: ath11k: disable interrupts during firmware crash recovery

The two patch files here are those commits verbatim — same code, same commit
messages including the review trailers — rebased onto current main and
refreshed against backports 7.2. They drop out on the next backports bump that
picks them up.

Please merge, thanks as always!

openwrt-ai (commented):

---

package/kernel/mac80211/patches/ath11k/951-wifi-ath11k-disable-interrupts-during-firmware-crash-recovery.patch
openwrt-ai: Resolved by the backports 7.2 bump rather than by dropping the line: `ath11k_cfr_deinit()` is in the tree now, so the context is correct. Verified against the real backports-7.2 tarball (the hash in the Makefile) with the full patch series applied in order — both 950 and 951 land at zero offset and zero fuzz. Thanks.

---

package/kernel/mac80211/patches/ath11k/950-wifi-ath11k-implement-CE-interrupt-enable-disable-for-AHB.patch
openwrt-ai: fixed, thanks — 950 is in the tree's refreshed form now and its hunks land at zero offset against backports 7.2 with the full ath11k series applied.

---

robimarko:

PR #24580, opened 2026-08-05
Swconfig nla null deref fixes

SilentSobs:
## swconfig: fix missing netlink attribute checks in swlib

Fixes two NULL-pointer-dereference crashes in swconfig's swlib.c netlink
attribute parsing, found via fuzz testing:

- `add_switch()`: `SWITCH_ATTR_ALIAS` was read via `nla_get_string()`
  without checking the attribute was present, unlike every other
  attribute in the function.
- `store_link_val()`: same issue for `SWITCH_LINK_SPEED` via
  `nla_get_u32()`.

Also hardens `add_switch()`/`list_switch()`/`add_attr()`/`store_val()`
to parse with an explicit `nla_policy` instead of `NULL`, so libnl
enforces string typing/NUL-termination, and clamps `dev->ports`/
`dev->vlans` to sane upper bounds before they size later allocations.

Both crashes require a malformed/incomplete netlink reply that the
in-tree kernel switch driver never actually produces, so impact is
low included as defensive hardening.

### DIFF
--- a/package/network/config/swconfig/src/swlib.c
+++ b/package/network/config/swconfig/src/swlib.c
@@ -750,7 +750,7 @@ add_switch(struct nl_msg *msg, void *arg)
 	if (nla_parse(tb, SWITCH_ATTR_MAX, genlmsg_attrdata(gnlh, 0), genlmsg_attrlen(gnlh, 0), NULL) < 0)
 		goto done;
 
-	if (!tb[SWITCH_ATTR_DEV_NAME])
+	if (!tb[SWITCH_ATTR_DEV_NAME] || !tb[SWITCH_ATTR_ALIAS])
 		goto done;
 
 	name = nla_get_string(tb[SWITCH_ATTR_DEV_NAME]);
--- a/package/network/config/swconfig/src/swlib.c
+++ b/package/network/config/swconfig/src/swlib.c
@@ -36,6 +36,9 @@
 #define DPRINTF(fmt, ...) do {} while (0)
 #endif
 
+#define SWITCH_MAX_PORTS 256
+#define SWITCH_MAX_VLANS 4096
+
 static struct nl_sock *handle;
 static struct nl_cache *cache;
 static struct genl_family *family;
@@ -59,6 +62,15 @@ static struct nla_policy link_policy[SWITCH_LINK_ATTR_MAX] = {
 	[SWITCH_LINK_SPEED] = { .type = NLA_U32 },
 	[SWITCH_LINK_FLAG_EEE_100BASET] = { .type = NLA_FLAG },
 	[SWITCH_LINK_FLAG_EEE_1000BASET] = { .type = NLA_FLAG },
+};
+
+static struct nla_policy switch_policy[SWITCH_ATTR_MAX + 1] = {
+	[SWITCH_ATTR_DEV_NAME] = { .type = NLA_STRING },
+	[SWITCH_ATTR_ALIAS] = { .type = NLA_STRING },
+	[SWITCH_ATTR_NAME] = { .type = NLA_STRING },
+	[SWITCH_ATTR_OP_NAME] = { .type = NLA_STRING },
+	[SWITCH_ATTR_OP_VALUE_STR] = { .type = NLA_STRING },
+	[SWITCH_ATTR_OP_DESCRIPTION] = { .type = NLA_STRING },
 };
 
 static inline void *
@@ -232,7 +244,7 @@ store_link_val(struct nl_msg *msg, struct nlattr *nla, struct switch_val *val)
 	link->aneg = !!tb[SWITCH_LINK_FLAG_ANEG];
 	link->tx_flow = !!tb[SWITCH_LINK_FLAG_TX_FLOW];
 	link->rx_flow = !!tb[SWITCH_LINK_FLAG_RX_FLOW];
-	link->speed = nla_get_u32(tb[SWITCH_LINK_SPEED]);
+	link->speed = tb[SWITCH_LINK_SPEED] ? nla_get_u32(tb[SWITCH_LINK_SPEED]) : 0;
 	link->eee = 0;
 	if (tb[SWITCH_LINK_FLAG_EEE_100BASET])
 		link->eee |= SWLIB_LINK_FLAG_EEE_100BASET;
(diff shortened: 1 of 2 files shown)


package/network/config/swconfig/src/swlib.c
```diff
@@ -796,7 +813,7 @@ list_switch(struct nl_msg *msg, void *arg)
 {
 	struct genlmsghdr *gnlh = nlmsg_data(nlmsg_hdr(msg));
 
-	if (nla_parse(tb, SWITCH_ATTR_MAX, genlmsg_attrdata(gnlh, 0), genlmsg_attrlen(gnlh, 0), NULL) < 0)
+	if (nla_parse(tb, SWITCH_ATTR_MAX, genlmsg_attrdata(gnlh, 0), genlmsg_attrlen(gnlh, 0), switch_policy) < 0)
 		goto done;
 
 	if (!tb[SWITCH_ATTR_DEV_NAME] || !tb[SWITCH_ATTR_NAME])
```
openwrt-ai: `list_switch()` still has exactly the bug that a3052bb fixes in `add_switch()`. The guard checks `SWITCH_ATTR_DEV_NAME` and `SWITCH_ATTR_NAME`, but the very next statement passes `tb[SWITCH_ATTR_ALIAS]` straight to `nla_get_string()` — see [`printf("Found: %s - %s\n", ...)` at swlib.c:822-823](https://github.com/openwrt/openwrt/blob/cadff73321eac03df6d2617edb28eff9736e25ed/package/network/config/swconfig/src/swlib.c#L822-L823). With `tb[SWITCH_ATTR_ALIAS] == NULL`, `nla_get_string()` returns `(char *)NULL + NLA_HDRLEN` and `printf("%s")` dereferences it: the same near-NULL deref, reachable from `swlib_list()` (i.e. `swconfig list`). Since this hunk already touches the function, it seems worth covering here rather than leaving one of the two identical sites unfixed.

```suggestion
	if (!tb[SWITCH_ATTR_DEV_NAME] || !tb[SWITCH_ATTR_NAME] ||
			!tb[SWITCH_ATTR_ALIAS])
```

---

package/network/config/swconfig/src/swlib.c
```diff
@@ -232,7 +244,8 @@ store_link_val(struct nl_msg *msg, struct nlattr *nla, struct switch_val *val)
 	link->aneg = !!tb[SWITCH_LINK_FLAG_ANEG];
 	link->tx_flow = !!tb[SWITCH_LINK_FLAG_TX_FLOW];
 	link->rx_flow = !!tb[SWITCH_LINK_FLAG_RX_FLOW];
-	link->speed = nla_get_u32(tb[SWITCH_LINK_SPEED]);
+	if (tb[SWITCH_LINK_SPEED])
```
openwrt-ai: Guarding the read leaves `link->speed` uninitialized rather than zero: the link struct is allocated with plain `malloc()`, not `swlib_alloc()` — [`val->value.link = malloc(sizeof(struct switch_port_link));` at swlib.c:233-234](https://github.com/openwrt/openwrt/blob/cadff73321eac03df6d2617edb28eff9736e25ed/package/network/config/swconfig/src/swlib.c#L233-L234) — and `swlib_get_attr()` memsets only `val->value` (which just NULLs the pointer), so a fresh malloc happens on every call. Every other field in this block is assigned unconditionally (`link->link`, `link->eee = 0`, …); `speed` should follow that pattern so callers like `swconfig`'s link printing don't read stack/heap garbage on a reply that omits `SWITCH_LINK_SPEED`.

```suggestion
	link->speed = tb[SWITCH_LINK_SPEED] ? nla_get_u32(tb[SWITCH_LINK_SPEED]) : 0;
```

---

package/network/config/swconfig/src/swlib.c
```diff
@@ -61,6 +64,15 @@ static struct nla_policy link_policy[SWITCH_LINK_ATTR_MAX] = {
 	[SWITCH_LINK_FLAG_EEE_1000BASET] = { .type = NLA_FLAG },
 };
 
+static struct nla_policy switch_policy[SWITCH_ATTR_MAX + 1] = {
```
openwrt-ai, line 67: nit: `switch_policy` is sized correctly here (`SWITCH_ATTR_MAX + 1`), which highlights that the three pre-existing policy arrays just above are not, and they are in the same parsing path this PR is hardening:

- [`link_policy[SWITCH_LINK_ATTR_MAX]`](https://github.com/openwrt/openwrt/blob/cadff73321eac03df6d2617edb28eff9736e25ed/package/network/config/swconfig/src/swlib.c#L57) — 9 entries, but used as `nla_parse_nested(tb, SWITCH_LINK_ATTR_MAX, nla, link_policy)` in the `store_link_val()` hunk below.
- [`portmap_policy[SWITCH_PORTMAP_MAX]`](https://github.com/openwrt/openwrt/blob/cadff73321eac03df6d2617edb28eff9736e25ed/package/network/config/swconfig/src/swlib.c#L52) — 4 entries, used as `nla_parse_nested(tb, SWITCH_PORTMAP_MAX, p, portmap_policy)` in `add_port_map()`.

libnl's `validate_nla()` only rejects `type > maxtype`, so a nested attribute whose type equals `maxtype` indexes `policy[maxtype]`, one past the end. The matching `tb[]` arrays in both functions already use `+ 1`; only the policy arrays don't. Also [`port_policy[SWITCH_ATTR_MAX]`](https://github.com/openwrt/openwrt/blob/cadff73321eac03df6d2617edb28eff9736e25ed/package/network/config/swconfig/src/swlib.c#L45) appears to use the wrong enum (`SWITCH_PORT_ATTR_MAX` was presumably meant) — harmless today since it is oversized, but confusing. Given the PR's stated goal of making this parser safe against malformed replies, is fixing these in the same series in scope?

---

openwrt-ai (commented):
Reviewed 2 new commits. Both commit messages accurately describe their diffs, and the `swconfig_send_switch()` path in the in-tree kernel driver does emit `SWITCH_ATTR_ALIAS` unconditionally, so requiring it in `add_switch()` doesn't regress the normal case. Three inline notes: one identical missing-presence-check left in `list_switch()`, one uninitialized field introduced by the `SWITCH_LINK_SPEED` guard, and a nit about the neighbouring policy array sizes.

---

openwrt-ai (commented):

The amended commit only changes the two spots raised last round — `list_switch()` now requires `SWITCH_ATTR_ALIAS` before `nla_get_string()`, and `link->speed` is assigned unconditionally instead of left uninitialized when `SWITCH_LINK_SPEED` is absent. Both match the updated commit message, and the kernel side emits `SWITCH_ATTR_ALIAS` unconditionally in `swconfig_send_switch()` (the same message backs both the `GET_SWITCH` and list paths), so the stricter guard doesn't regress the normal case.

The open nit about the pre-existing `port_policy` / `portmap_policy` / `link_policy` array sizes is still unaddressed; not re-raising it here.

---

PR #24581, opened 2026-08-05
mediatek: add support for Cudy TR3600 v1

soapmancn:
## Hardware specifications

The Cudy TR3600 v1 is based on the MediaTek MT7987B SoC with 512 MiB DDR4 and an ESMT F50L2G41XA 2 Gbit (256 MiB) SPI-NAND.

- MediaTek MT7990AN dual-band radio: 2x2 on 2.4 GHz and 3x3 on 5 GHz
- Two 2.5GbE ports using the MT7987 integrated GMACs and a Realtek RTL8221B external PHY
- USB-A 3.0
- Red status LED on GPIO46 and blue status LED on GPIO48
- Mode button on GPIO0 and reset button on GPIO1
- PWM fan: GPIO6 fan supply enable and PWM0 output
- USB-C USB-PD input: 5 V/3 A, 9 V/2.2 A, or 12 V/1.6 A; 15 W maximum consumption
- Serial console: serial0, 115200 8N1. The vendor development files do not document the physical UART header location or pad pinout; UART is not required for installation.

## Flash instructions

1. From the stock Cudy web interface, upload cudy_tr3600-v1-sysupgrade_260715.bin from the official development files.
2. After the intermediate firmware boots, upload openwrt-mediatek-filogic-cudy_tr3600-v1-squashfs-sysupgrade.bin from the OpenWrt upgrade page.
3. This installation path does not require UART. Keep a stock firmware backup; the vendor recovery/dual-image revert procedure has not been validated by this change.

The intermediate firmware and the official device documentation are available from the [Cudy TR3600 download center](https://www.cudy.com/zh-cn/pages/download-center/tr3600-1-0).

The SPI-NAND layout follows the official Cudy R126 DTS and DTB: BL2, writable backup, Factory, bdinfo, FIP, and ubi. The vendor-defined backup label at 0x100000 is retained.

## MAC address layout

The base MAC is stored in bdinfo at offset 0xde00.

- LAN (eth1/gmac1): xx:xx:xx:xx:xx:xx, base + 0
- WAN (eth0/gmac0): xx:xx:xx:xx:xx:xx, base + 1
- 2.4 GHz (band@0): xx:xx:xx:xx:xx:xx, base + 3
- 5 GHz (band@1): xx:xx:xx:xx:xx:xx, base + 2

The 2.4 GHz offset is intentionally distinct from LAN to avoid duplicate MAC addresses.

The DTS follows the official Cudy intermediate firmware for the v1 identity, MT7990 calibration data, PWM0 fan output, red/blue status LEDs, and R126 NAND layout. The generic wifi@0,0 node follows the OpenWrt DT binding convention; the image recipe uses kmod-mt7990-firmware.

This update is pushed to the original PR branch so that the existing review context remains attached to this pull request.

### DIFF
--- a/target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
+++ b/target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
@@ -1,0 +1,265 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+/dts-v1/;
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+
+#include "mt7987b.dtsi"
+
+/ {
+	model = "Cudy TR3600 v1";
+	compatible = "cudy,tr3600-v1", "mediatek,mt7987b", "mediatek,mt7987";
+
+	aliases {
+		label-mac-device = &gmac1;
+		led-boot = &led_status_red;
+		led-failsafe = &led_status_blue;
+		led-running = &led_status_blue;
+		led-upgrade = &led_status_blue;
+		serial0 = &uart0;
+	};
+
+	chosen {
+		bootargs = "console=ttyS0,115200n1 earlycon=uart8250,mmio32,0x11000000 pci=pcie_bus_perf";
+		stdout-path = "serial0:115200n8";
+	};
+
+	memory@40000000 {
+		device_type = "memory";
+		reg = <0 0x40000000 0 0x20000000>;
+	};
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		led_status_blue: led-0 {
+			color = <LED_COLOR_ID_BLUE>;
+			function = LED_FUNCTION_STATUS;
+			gpios = <&pio 48 GPIO_ACTIVE_LOW>;
+		};
+
+		led_status_red: led-1 {
+			color = <LED_COLOR_ID_RED>;
+			function = LED_FUNCTION_STATUS;
+			gpios = <&pio 46 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+	gpio-keys {
+		compatible = "gpio-keys";
+
+		button-mode {
+			label = "mode";
+			linux,code = <BTN_0>;
+			linux,input-type = <EV_SW>;
+			gpios = <&pio 0 GPIO_ACTIVE_LOW>;
+			debounce-interval = <10>;
+		};
+
+		button-reset {
+			label = "reset";
+			linux,code = <KEY_RESTART>;
+			gpios = <&pio 1 GPIO_ACTIVE_LOW>;
+			debounce-interval = <10>;
+		};
+	};
+
+	fan_3p3v: regulator-fan-3p3v {
+		compatible = "regulator-fixed";
+		enable-active-high;
+		gpios = <&pio 6 GPIO_ACTIVE_HIGH>;
+		regulator-name = "fan";
+		regulator-min-microvolt = <3300000>;
+		regulator-max-microvolt = <3300000>;
+	};
+};
+
+&eth {
+	status = "okay";
+};
+
+&fan {
+	fan-supply = <&fan_3p3v>;
+	pwms = <&pwm 0 50000 0>;
+	status = "okay";
+};
+
+&gmac0 {
+	nvmem-cells = <&macaddr_bdinfo_de00 1>;
+	nvmem-cell-names = "mac-address";
+	phy-handle = <&phy1>;
+	phy-mode = "2500base-x";
+	status = "okay";
+};
+
+&gmac1 {
+	nvmem-cells = <&macaddr_bdinfo_de00 0>;
+	nvmem-cell-names = "mac-address";
+	phy-handle = <&phy15>;
+	phy-mode = "internal";
+	status = "okay";
+};
+
+&mdio {
+	phy1: phy@1 {
+		compatible = "ethernet-phy-ieee802.3-c45";
+		reg = <1>;
+		reset-gpios = <&pio 42 GPIO_ACTIVE_LOW>;
+		reset-assert-us = <100000>;
+		reset-deassert-us = <120000>;
+	};
+
+	phy15: phy@f {
+		compatible = "ethernet-phy-ieee802.3-c45";
+		reg = <15>;
+	};
+};
+
+&pcie0 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&pcie0_pins>;
+	reset-gpios = <&pio 38 GPIO_ACTIVE_HIGH>;
+	status = "okay";
+
+	pcie@0,0 {
+		reg = <0x0000 0 0 0 0>;
+		#address-cells = <3>;
+		#size-cells = <2>;
+		device_type = "pci";
+
+		wifi@0,0 {
+			compatible = "mediatek,mt76";
+			reg = <0x0000 0 0 0 0>;
+
+			#address-cells = <1>;
+			#size-cells = <0>;
+			ieee80211-freq-limit = <2400000 2500000>,
+					       <5170000 5835000>;
+
+			nvmem-cells = <&eeprom_factory_0>;
+			nvmem-cell-names = "eeprom";
+
+			band@0 {
+				reg = <0>;
+				nvmem-cells = <&macaddr_bdinfo_de00 3>;
+				nvmem-cell-names = "mac-address";
+			};
+
+			band@1 {
+				reg = <1>;
+				nvmem-cells = <&macaddr_bdinfo_de00 2>;
+				nvmem-cell-names = "mac-address";
+			};
+		};
+	};
+};
+
+&pwm {
+	pinctrl-names = "default";
+	pinctrl-0 = <&pwm_pins>;
+	status = "okay";
+};
+
+&spi0 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&spi0_flash_pins>;
+	status = "okay";
+
+	flash@0 {
+		compatible = "spi-nand";
+		reg = <0>;
+
+		spi-cal-enable;
+		spi-cal-mode = "read-data";
+		spi-cal-datalen = <7>;
+		spi-cal-data = /bits/ 8 <0x53 0x50 0x49 0x4e 0x41 0x4e 0x44>;
+		spi-cal-addrlen = <5>;
+		spi-cal-addr = /bits/ 32 <0x0 0x0 0x0 0x0 0x0>;
+
+		spi-max-frequency = <52000000>;
+		spi-tx-bus-width = <4>;
+		spi-rx-bus-width = <4>;
+
+		mediatek,nmbm;
+		mediatek,bmt-max-ratio = <1>;
+		mediatek,bmt-max-reserved-blocks = <64>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "BL2";
+				reg = <0x0 0x100000>;
+				read-only;
+			};
+
+			partition@100000 {
+				label = "backup";
+				reg = <0x100000 0x80000>;
+			};
+
+			partition@180000 {
+				label = "Factory";
+				reg = <0x180000 0x400000>;
+				read-only;
+
+				nvmem-layout {
+					compatible = "fixed-layout";
+					#address-cells = <1>;
+					#size-cells = <1>;
+
+					eeprom_factory_0: eeprom@0 {
+						reg = <0x0 0x1e00>;
+					};
+				};
+			};
+
+			partition@580000 {
+				label = "bdinfo";
+				reg = <0x580000 0x40000>;
+				read-only;
+
+				nvmem-layout {
+					compatible = "fixed-layout";
+					#address-cells = <1>;
+					#size-cells = <1>;
+
+					macaddr_bdinfo_de00: macaddr@de00 {
+						compatible = "mac-base";
+						reg = <0xde00 0x6>;
+						#nvmem-cell-cells = <1>;
+					};
+				};
+			};
+
+			partition@5c0000 {
+				label = "FIP";
+				reg = <0x5c0000 0x200000>;
+				read-only;
+			};
+
+			partition@7c0000 {
+				label = "ubi";
+				reg = <0x7c0000 0xe840000>;
+				compatible = "linux,ubi";
+			};
+		};
+	};
+};
+
+&ssusb {
+	status = "okay";
+};
+
+&tphyu3port0 {
+	status = "okay";
+};
+
+&uart0 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&uart0_pins>;
+	status = "okay";
+};
--- a/target/linux/mediatek/filogic/base-files/etc/board.d/02_network
+++ b/target/linux/mediatek/filogic/base-files/etc/board.d/02_network
@@ -165,6 +165,7 @@ mediatek_setup_interfaces()
 	cudy,tr3000-256mb-v1|\
 	cudy,tr3000-v1|\
 	cudy,tr3000-v1-ubootmod|\
+	cudy,tr3600-v1|\
 	glinet,gl-mt2500|\
 	glinet,gl-mt2500-airoha|\
 	glinet,gl-mt3000|\
--- a/target/linux/mediatek/image/filogic.mk
+++ b/target/linux/mediatek/image/filogic.mk
@@ -1405,6 +1405,25 @@ define Device/cudy_tr3000-v1-ubootmod
 endef
 TARGET_DEVICES += cudy_tr3000-v1-ubootmod
 
+define Device/cudy_tr3600-v1
+  DEVICE_VENDOR := Cudy
+  DEVICE_MODEL := TR3600
+  DEVICE_VARIANT := v1
+  DEVICE_DTS := mt7987b-cudy-tr3600-v1
+  DEVICE_DTS_DIR := ../dts
+  SUPPORTED_DEVICES += R126
+  DEVICE_PACKAGES := mt7987-2p5g-phy-firmware kmod-mt7990-firmware \
+	kmod-hwmon-pwmfan kmod-usb3
+  KERNEL_IN_UBI := 1
+  KERNEL_LOADADDR := 0x40000000
+  UBINIZE_OPTS := -E 5
+  BLOCKSIZE := 128k
+  PAGESIZE := 2048
+  IMAGE_SIZE := 237824k
+  IMAGE/sysupgrade.bin := sysupgrade-tar | append-metadata
+endef
+TARGET_DEVICES += cudy_tr3600-v1
+
 define Device/cudy_wr3000-v1
   DEVICE_VENDOR := Cudy
   DEVICE_MODEL := WR3000
--- a/target/linux/mediatek/base-files/lib/preinit/05_set_preinit_iface
+++ b/target/linux/mediatek/base-files/lib/preinit/05_set_preinit_iface
@@ -6,6 +6,7 @@ set_preinit_iface() {
 	cudy,tr3000-256mb-v1|\
 	cudy,tr3000-v1|\
 	cudy,tr3000-v1-ubootmod|\
+	cudy,tr3600-v1|\
 	glinet,gl-mt2500|\
 	glinet,gl-mt2500-airoha|\
 	glinet,gl-mt3000|\


target/linux/mediatek/dts/mt7987b-cudy-tr3600.dts
```diff
@@ -0,0 +1,259 @@
...
+	nvmem-cells = <&macaddr_bdinfo_de00 0>;
+	nvmem-cell-names = "mac-address";
+	phy-handle = <&phy15>;
+	phy-mode = "internal";
+	status = "okay";
+};
+
+&mdio {
+	phy1: phy@1 {
+		compatible = "ethernet-phy-id001c.c849", "ethernet-phy-id4f51.ea19";
```
openwrt-ai: These two IDs are two completely different chips from two different vendors, so they must not appear in one compatible list: `001c.c849` is the Realtek RTL8221B-VB-CG ([`RTL_8221B_VB_CG` in realtek_main.c:164](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/realtek/realtek_main.c#L164)), while `4f51.ea19` is the Motorcomm YT8821 ([`PHY_ID_YT8821` in motorcomm.c:19](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/motorcomm.c#L19)) — the ID used by [mt7981b-cudy-m3000-v2-yt8821.dtsi](https://github.com/openwrt/openwrt/blob/bfc367f758945aa34aaf42eb69367895e84ac535/target/linux/mediatek/dts/mt7981b-cudy-m3000-v2-yt8821.dtsi#L9). A device can't be both, and the `realtek,aldps-enable` below confirms the Realtek part. `realtek,rtl82xx.yaml` also declares `compatible` as a single-valued `enum`, so the second string is a binding violation too.

```suggestion
		compatible = "ethernet-phy-id001c.c849";
```

---

target/linux/mediatek/dts/mt7987b-cudy-tr3600.dts
```diff
@@ -0,0 +1,259 @@
...
+
+&eth {
+	status = "okay";
+};
+
+&fan {
+	fan-supply = <&fan_3p3v>;
+	pinctrl-names = "default";
+	pinctrl-0 = <&pwm_pins>;
+	pwms = <&pwm 0 50000 0>;
```
openwrt-ai: `&pwm` below (line 152-156) already claims `&pwm_pins`, and claiming the same pin group again from the `pwm-fan` consumer node makes the two devices fight over the same pins: `pin_request()` rejects a pin whose `owner` is already set by another device, so whichever of `pwm@10048000` / `pwm-fan` probes second fails with `-EBUSY` and fan control is lost. The pinmux belongs on the PWM controller only — no other in-tree board puts `pinctrl-*` on `&fan` (compare [mt7987a-hiveton-h5000m.dts](https://github.com/openwrt/openwrt/blob/bfc367f758945aa34aaf42eb69367895e84ac535/target/linux/mediatek/dts/mt7987a-hiveton-h5000m.dts#L96-L99)).

```suggestion
	fan-supply = <&fan_3p3v>;
	pwms = <&pwm 0 50000 0>;
```

---

target/linux/mediatek/dts/mt7987b-cudy-tr3600.dts
```diff
@@ -0,0 +1,259 @@
...
+	reset-gpios = <&pio 38 GPIO_ACTIVE_HIGH>;
+	status = "okay";
+
+	pcie@0,0 {
+		reg = <0x0000 0 0 0 0>;
+		#address-cells = <3>;
+		#size-cells = <2>;
+		device_type = "pci";
+
+		mt7996@0,0 {
```
openwrt-ai: Node name doesn't follow the generic-name convention, and it names the wrong chip: the image recipe pulls `kmod-mt7992-23-firmware`, i.e. this is an MT7992, not an MT7996. All three existing MT7987 boards with a PCIe MT799x radio use `wifi@0,0` ([routerich-be7200](https://github.com/openwrt/openwrt/blob/bfc367f758945aa34aaf42eb69367895e84ac535/target/linux/mediatek/dts/mt7987a-routerich-be7200.dts#L332), [gl-mt3600be](https://github.com/openwrt/openwrt/blob/bfc367f758945aa34aaf42eb69367895e84ac535/target/linux/mediatek/dts/mt7987a-glinet-gl-mt3600be.dts#L140), [tenda-be12-pro](https://github.com/openwrt/openwrt/blob/bfc367f758945aa34aaf42eb69367895e84ac535/target/linux/mediatek/dts/mt7987a-tenda-be12-pro.dts#L358)).

```suggestion
		wifi@0,0 {
```

---

target/linux/mediatek/dts/mt7987b-cudy-tr3600.dts
```diff
@@ -0,0 +1,259 @@
...
+			reg = <0x0000 0 0 0 0>;
+
+			ieee80211-freq-limit = <2400000 2500000>,
+					       <5170000 5835000>;
+
+			nvmem-cells = <&eeprom_factory_0>;
+			nvmem-cell-names = "eeprom";
+
+			#address-cells = <1>;
+			#size-cells = <0>;
```
openwrt-ai: `#address-cells`/`#size-cells` are declared here but no `band@N` children follow, so they have no effect. Every other MT7987 board with the same dual-band MT7992 and the same `ieee80211-freq-limit` pair adds `band@0`/`band@1` with per-band `mac-address` nvmem cells — see [mt7987a-routerich-be7200.dts:345-356](https://github.com/openwrt/openwrt/blob/bfc367f758945aa34aaf42eb69367895e84ac535/target/linux/mediatek/dts/mt7987a-routerich-be7200.dts#L345-L356) and [mt7987a-glinet-gl-mt3600be.dts:154-165](https://github.com/openwrt/openwrt/blob/bfc367f758945aa34aaf42eb69367895e84ac535/target/linux/mediatek/dts/mt7987a-glinet-gl-mt3600be.dts#L154-L165). Is the omission intentional (i.e. the `Factory` EEPROM already carries valid per-band MACs on this board), or should the two WiFi MACs be derived from `macaddr_bdinfo_de00` like the Ethernet ones? The PR body's MAC section doesn't say.

---

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,259 @@
...
+			partition@0 {
+				label = "BL2";
+				reg = <0x0 0x100000>;
+				read-only;
+			};
+
+			partition@100000 {
+				label = "backup";
+				reg = <0x100000 0x80000>;
+			};
```
openwrt-ai, line 217: This partition sits at exactly the same offset/size (`0x100000` / `0x80000`) that every other Cudy NAND board in this target labels `u-boot-env`, and the rest of the layout (BL2 / Factory / bdinfo / FIP / ubi) matches that family layout too — see [mt7981b-cudy-wr3000-nand.dtsi:100-104](https://github.com/openwrt/openwrt/blob/bfc367f758945aa34aaf42eb69367895e84ac535/target/linux/mediatek/dts/mt7981b-cudy-wr3000-nand.dtsi#L100-L104). Where does the `backup` name come from? If it really is the U-Boot environment it should be labelled `u-boot-env` and carry `read-only;` like the siblings, otherwise a stray write here can brick the bootloader config.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -1405,6 +1405,25 @@ define Device/cudy_tr3000-v1-ubootmod
 endef
 TARGET_DEVICES += cudy_tr3000-v1-ubootmod
 
+define Device/cudy_tr3600
+  DEVICE_VENDOR := Cudy
+  DEVICE_MODEL := TR3600
+  DEVICE_VARIANT := v1
+  DEVICE_DTS := mt7987b-cudy-tr3600
```
openwrt-ai: nit: the `v1` variant is not reflected anywhere else for this device — the DTS is `mt7987b-cudy-tr3600.dts`, the compatible is `cudy,tr3600` and `02_network` matches `cudy,tr3600`. Every other Cudy device that declares `DEVICE_VARIANT := v1` carries `-v1` through the DTS filename and compatible as well (e.g. [`cudy_wr3000u-v1` at filogic.mk:1612-1616](https://github.com/openwrt/openwrt/blob/bfc367f758945aa34aaf42eb69367895e84ac535/target/linux/mediatek/image/filogic.mk#L1612-L1616), and `cudy_tr3000-v1` just above). Either drop `DEVICE_VARIANT` or rename the DTS/compatible to `cudy,tr3600-v1` so the image name and the board compatible agree.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -1405,6 +1405,25 @@ define Device/cudy_tr3000-v1-ubootmod
...
+  DEVICE_DTS_DIR := ../dts
+  SUPPORTED_DEVICES += R126
+  DEVICE_PACKAGES := mt7987-2p5g-phy-firmware kmod-mt7996e \
+	kmod-mt7992-23-firmware kmod-hwmon-pwmfan kmod-usb3
```
openwrt-ai: nit: the PR body describes the radio as "an MT7996 PCIe radio" and the DTS names the node `mt7996@0,0`, but the firmware package selected here is `kmod-mt7992-23-firmware`. Please make the three agree — if the part is an MT7992 (which the `2400000-2500000` + `5170000-5835000` `ieee80211-freq-limit` pair also suggests), fix the PR body and the DTS node name rather than the package.

---

openwrt-ai (commented):
## Commit checks

- `bfc367f7` "mediatek: add support for Cudy TR3600" — this adds a new device (new `target/linux/mediatek/dts/mt7987b-cudy-tr3600.dts` plus a new `define Device/cudy_tr3600` in `target/linux/mediatek/image/filogic.mk`), so the commit message needs the three sections OpenWrt expects for new-device support, and currently has none of them:
  - **Hardware specification** — the message has a one-line summary, but no RAM/flash chip designations, no WiFi chip + band/chain layout, no Ethernet PHY/switch parts, no LED/button counts, no UART header location/pinout/baud, no power rating.
  - **Flash instructions** — no install path at all. At minimum one fully-described procedure (which image, how to get the device into a flashable state, verbatim button/IP/hostname strings) plus a revert-to-stock note. The PR body says the vendor dual-image U-Boot flow is unvalidated; whatever path is intended still has to be written down here.
  - **MAC address layout** — nothing, although the DTS derives both Ethernet MACs from `bdinfo@0xde00` with offsets 0 and 1. One line per interface (LAN/WAN/2.4G/5G) using `xx:xx:xx` for the unit-specific bytes.


---

RonKay:
Keep it up, looking forward to the firmware update for the TR3600.

soapmancn:
> Keep it up, looking forward to the firmware update for the TR3600.

PR要求太多了没空弄了已经分享出来了：https://www.right.com.cn/forum/thread-8484429-1-1.html
网盘：
链接：https://pan.quark.cn/s/f953599bd80a 提取码：BzmH

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,279 @@
...
+
+	chosen {
+		bootargs = "console=ttyS0,115200n1 earlycon=uart8250,mmio32,0x11000000 pci=pcie_bus_perf";
+		stdout-path = "serial0:115200n8";
+	};
+
+	/* Cudy TR3600 v1 has 512 MiB DDR4 at 0x40000000. */
+	memory {
+		reg = <0 0x40000000 0 0x20000000>;
+	};
```
openwrt-ai, line 32: Without `device_type = "memory"` this node is skipped entirely — [`early_init_dt_scan_memory()` at fdt.c:1035-1042](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/of/fdt.c#L1035-L1042) `continue`s on any top-level node whose `device_type` is not `"memory"`. The node it overrides in `mt7987a.dtsi` has the same defect, so the declared 512 MiB never reaches the kernel and RAM size ends up whatever the bootloader passes in. `mt7987a-routerich-be7200.dts` gets this right ([memory@40000000](https://github.com/openwrt/openwrt/blob/989da8c6a46b807ec6ab8f1610ccff7bb032e0b7/target/linux/mediatek/dts/mt7987a-routerich-be7200.dts#L110-L113)).

```suggestion
	memory {
		device_type = "memory";
		reg = <0 0x40000000 0 0x20000000>;
	};
```

---

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,279 @@
...
+	pcie@0,0 {
+		reg = <0x0000 0 0 0 0>;
+
+		wifi@0,0 {
+			compatible = "mediatek,mt76";
+			reg = <0x0000 0 0 0 0>;
+
+			#address-cells = <1>;
+			#size-cells = <0>;
+			device_type = "pci";
```
openwrt-ai: `device_type = "pci"` belongs on the root-port bridge `pcie@0,0`, not on the endpoint `wifi@0,0`, and `pcie@0,0` also needs `#address-cells = <3>`/`#size-cells = <2>` — the parent `pcie0` in [mt7987.dtsi:939-944](https://github.com/openwrt/openwrt/blob/989da8c6a46b807ec6ab8f1610ccff7bb032e0b7/target/linux/mediatek/dts/mt7987.dtsi#L939-L944) declares no `pcie@0,0` child, so this file is the only place they can come from. The `#address-cells = <1>`/`#size-cells = <0>` on `wifi@0,0` are for the `band@N` children and are correct where they are; combining them with `device_type = "pci"` makes the node look like a PCI bus with a 1-cell address. The previous revision of this PR had it in the correct shape, and it is what the bulk of the target does (e.g. [mt7988a-arcadyan-mozart.dts:309-313](https://github.com/openwrt/openwrt/blob/989da8c6a46b807ec6ab8f1610ccff7bb032e0b7/target/linux/mediatek/dts/mt7988a-arcadyan-mozart.dts#L309-L313)). Note [mt7987a-glinet-gl-mt3600be.dts:131-142](https://github.com/openwrt/openwrt/blob/989da8c6a46b807ec6ab8f1610ccff7bb032e0b7/target/linux/mediatek/dts/mt7987a-glinet-gl-mt3600be.dts#L131-L142) has the same layout as this PR, so if that arrangement was copied deliberately, say so — otherwise:

```suggestion
	pcie@0,0 {
		reg = <0x0000 0 0 0 0>;
		#address-cells = <3>;
		#size-cells = <2>;
		device_type = "pci";

		wifi@0,0 {
			compatible = "mediatek,mt76";
			reg = <0x0000 0 0 0 0>;

			#address-cells = <1>;
			#size-cells = <0>;
```

---

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,279 @@
...
+};
+
+&mdio {
+	phy1: phy@1 {
+		compatible = "ethernet-phy-id001c.c849";
+		reg = <1>;
+		reset-gpios = <&pio 42 GPIO_ACTIVE_LOW>;
+		reset-assert-us = <100000>;
+		reset-deassert-us = <120000>;
+		realtek,aldps-enable;
```
openwrt-ai: nit: this property has no effect on an RTL8221B. `realtek,aldps-enable` is only consumed by [`rtl8211f_config_aldps()`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/realtek/realtek_main.c#L650-L663), which is called exclusively from [`rtl8211f_config_init()` at realtek_main.c:677](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/realtek/realtek_main.c#L677). PHY id `001c.c849` is [`RTL_8221B_VB_CG`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/realtek/realtek_main.c#L164), which binds to the rtl822xb driver and never reads it. `mt7987a-hiveton-h5000m.dts` carries the same dead property on its RTL8221B, so this may just have been copied — but it can be dropped here.

```suggestion
```

---

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,279 @@
...
+	pinctrl-0 = <&pwm1_pins>;
+	status = "okay";
+};
+
+&spi0 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&spi0_flash_pins>;
+	status = "okay";
+	#address-cells = <1>;
+	#size-cells = <0>;
```
openwrt-ai: nit: redundant — the `spi0` node already sets both at [mt7987.dtsi:755-756](https://github.com/openwrt/openwrt/blob/989da8c6a46b807ec6ab8f1610ccff7bb032e0b7/target/linux/mediatek/dts/mt7987.dtsi#L755-L756). The other in-tree `&spi0` users (`mt7987a-routerich-be7200.dts`, `mt7987a-tenda-be12-pro.dts`) don't repeat them.

```suggestion
```

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -1405,6 +1405,29 @@ define Device/cudy_tr3000-v1-ubootmod
...
+  DEVICE_VARIANT := v1
+  DEVICE_DTS := mt7987b-cudy-tr3600-v1
+  DEVICE_DTS_DIR := ../dts
+  SUPPORTED_DEVICES += R126
+  DEVICE_PACKAGES := mt7987-2p5g-phy-firmware kmod-mt7990-firmware \
+	kmod-hwmon-pwmfan kmod-usb3
+  KERNEL = kernel-bin | lzma | \
+	fit lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb
+  KERNEL_INITRAMFS = kernel-bin | lzma | \
+	fit lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb with-initrd
```
openwrt-ai: nit: these four lines are a verbatim copy of `Device/Default` in [target/linux/mediatek/image/Makefile:20-23](https://github.com/openwrt/openwrt/blob/989da8c6a46b807ec6ab8f1610ccff7bb032e0b7/target/linux/mediatek/image/Makefile#L20-L23), so they can be dropped. The closest sibling recipe, `Device/glinet_gl-mt3600be`, doesn't set them either.

```suggestion
```

---

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,279 @@
...
+&gmac0 {
+	nvmem-cells = <&macaddr_bdinfo_de00 1>;
+	nvmem-cell-names = "mac-address";
+	phy-handle = <&phy1>;
+	phy-mode = "2500base-x";
+	status = "okay";
+};
+
+&gmac1 {
+	nvmem-cells = <&macaddr_bdinfo_de00 0>;
```
openwrt-ai, line 98: `gmac1` (LAN) and `band@0` (2.4 GHz, line 145) both take offset 0 from `macaddr_bdinfo_de00`, so the LAN interface and the 2.4 GHz AP will come up with the identical MAC. Both end up as ports of `br-lan` in the default config, which puts a duplicate address on the same L2 segment. Is that deliberate vendor behaviour that you verified on the device, or should 2.4 GHz use a distinct offset (base+3, say)? The commit message states the layout but doesn't note the collision, so it reads as unintentional.

For comparison, [mt7987a-glinet-gl-mt3600be.dts](https://github.com/openwrt/openwrt/blob/989da8c6a46b807ec6ab8f1610ccff7bb032e0b7/target/linux/mediatek/dts/mt7987a-glinet-gl-mt3600be.dts#L90-L104) keeps the Ethernet MACs and the radio MACs on two separate bases (`macaddr_factory_4000` vs `macaddr_factory_4006`).

---

target/linux/mediatek/filogic/base-files/etc/board.d/02_network
```diff
@@ -164,6 +164,7 @@ mediatek_setup_interfaces()
 	cudy,tr3000-256mb-v1|\
 	cudy,tr3000-v1|\
 	cudy,tr3000-v1-ubootmod|\
+	cudy,tr3600-v1|\
```
openwrt-ai, line 167: nit: every other `cudy,tr3000-*` board that lands in this `ucidef_set_interfaces_lan_wan eth1 eth0` branch is also listed in [target/linux/mediatek/base-files/lib/preinit/05_set_preinit_iface:3-16](``https://github.com/openwrt/openwrt/blob/989da8c6a46b807ec6ab8f1610ccff7bb032e0b7/target/linux/mediatek/base-files/lib/preinit/05_set_preinit_iface#L3-L16``) so failsafe comes up on the LAN port; `cudy,tr3600-v1` is not. Intentional? (`glinet,gl-mt3600be` is also absent there, so there is precedent either way — but for the Cudy family this is the odd one out.)

---

openwrt-ai (commented):
Re-reviewed after the force-push (`bfc367f7` → `989da8c6`). The commit message now carries the hardware / flash-instructions / MAC-layout sections that were missing before, so the earlier commit-check is resolved. Remaining notes are inline; CI on the head commit is green.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -1405,6 +1405,25 @@ define Device/cudy_tr3000-v1-ubootmod
...
+  KERNEL_IN_UBI := 1
```
openwrt-ai: This revision dropped the wrong four lines. `KERNEL_LOADADDR`, `UBINIZE_OPTS`, `BLOCKSIZE` and `PAGESIZE` are **not** in `Device/Default` — [target/linux/mediatek/image/Makefile:17-34](https://github.com/openwrt/openwrt/blob/245435461884225e018df32957e8797415f6ec51/target/linux/mediatek/image/Makefile#L17-L34) only sets `KERNEL_LOADADDR = $(loadaddr-y)`, which on filogic is `0x48000000`, and `BLOCKSIZE`/`PAGESIZE`/`UBINIZE_OPTS` default to empty in [include/image.mk:523-529](https://github.com/openwrt/openwrt/blob/245435461884225e018df32957e8797415f6ec51/include/image.mk#L523-L529).

So the removal silently moves the FIT load/entry address from `0x40000000` to `0x48000000` (`KERNEL_LOADADDR` feeds `-a`/`-e` in [`Build/fit-its`](https://github.com/openwrt/openwrt/blob/245435461884225e018df32957e8797415f6ec51/include/image-commands.mk#L448)) and leaves the UBI geometry unset. All three sibling MT7987 recipes keep them: [`glinet_gl-mt3600be`](https://github.com/openwrt/openwrt/blob/245435461884225e018df32957e8797415f6ec51/target/linux/mediatek/image/filogic.mk#L1853-L1857), [`routerich_be7200`](https://github.com/openwrt/openwrt/blob/245435461884225e018df32957e8797415f6ec51/target/linux/mediatek/image/filogic.mk#L3122-L3129), [`tenda_be12-pro`](https://github.com/openwrt/openwrt/blob/245435461884225e018df32957e8797415f6ec51/target/linux/mediatek/image/filogic.mk#L3221-L3224).

What *is* redundant here is `KERNEL` / `KERNEL_INITRAMFS` (lines 1417-1420), which are byte-identical to `Device/Default` and are still present. Swapping the two:

```suggestion
  KERNEL_IN_UBI := 1
  KERNEL_LOADADDR := 0x40000000
  UBINIZE_OPTS := -E 5
  BLOCKSIZE := 128k
  PAGESIZE := 2048
```

---

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,280 @@
...
+			};
+
+			band@1 {
+				reg = <1>;
+				nvmem-cells = <&macaddr_bdinfo_de00 2>;
+				nvmem-cell-names = "mac-address";
+			};
+
+			/* Keep the 2.4 GHz MAC distinct from the LAN MAC. */
+			/* MT7990 calibration data is provided by the Factory EEPROM. */
```
openwrt-ai: nit: the new comment explains the `3` offset on `band@0` but sits 12 lines below it, after `band@1`, where it reads as if it applied to the 5 GHz cell. Put it next to the node it describes:

```suggestion
			/* Offset 3 keeps the 2.4 GHz MAC distinct from LAN. */
			band@0 {
				reg = <0>;
				nvmem-cells = <&macaddr_bdinfo_de00 3>;
				nvmem-cell-names = "mac-address";
			};

			band@1 {
				reg = <1>;
				nvmem-cells = <&macaddr_bdinfo_de00 2>;
				nvmem-cell-names = "mac-address";
			};

			/* MT7990 calibration data is provided by the Factory EEPROM. */
```

---

openwrt-ai (commented):
Re-reviewed after the force-push (`989da8c6` → `24543546`). Most of the previous round is addressed: `device_type = "memory"`, the `pcie@0,0` / `wifi@0,0` cell split, the dead `realtek,aldps-enable`, the duplicate `&spi0` cell properties, the 2.4 GHz MAC collision (now base+3), and the `05_set_preinit_iface` entry. The commit message matches the diff.

One regression came out of my earlier `filogic.mk` nit — it was aimed at the redundant `KERNEL`/`KERNEL_INITRAMFS` lines, but the four UBI/loadaddr variables got removed instead. Details inline. Sorry for the ambiguous wording last time.

CI reports no checks on the head commit yet, so there was nothing to ground against there.

---

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,280 @@
...
+				nvmem-cell-names = "mac-address";
+			};
+
+			band@1 {
+				reg = <1>;
+				nvmem-cells = <&macaddr_bdinfo_de00 2>;
+				nvmem-cell-names = "mac-address";
+			};
+
+			/* MT7990 calibration data is provided by the Factory EEPROM. */
```
openwrt-ai: nit: the second comment is now the orphaned one — `/* MT7990 calibration data is provided by the Factory EEPROM. */` sits at the very end of `wifi@0,0`, 15 lines below the `nvmem-cells = <&eeprom_factory_0>;` / `nvmem-cell-names = "eeprom";` pair it describes, so it reads as a trailing note about `band@1`. Those two properties are already self-explanatory and no sibling annotates them ([mt7987a-glinet-gl-mt3600be.dts:150-151](``https://github.com/openwrt/openwrt/blob/a95d9635671a298bcba5544791b3d660081ebb43/target/linux/mediatek/dts/mt7987a-glinet-gl-mt3600be.dts#L150-L151)),`` so dropping it is simplest — move it directly above line 143 if you want to keep it.

Same hunk also restores the blank line between the node's properties and its first child, which the sibling has ([mt7987a-glinet-gl-mt3600be.dts:151-153](https://github.com/openwrt/openwrt/blob/a95d9635671a298bcba5544791b3d660081ebb43/target/linux/mediatek/dts/mt7987a-glinet-gl-mt3600be.dts#L151-L153)) and which my previous suggestion accidentally ate.

```suggestion
			nvmem-cell-names = "eeprom";

			/* Offset 3 keeps the 2.4 GHz MAC distinct from LAN. */
			band@0 {
				reg = <0>;
				nvmem-cells = <&macaddr_bdinfo_de00 3>;
				nvmem-cell-names = "mac-address";
			};

			band@1 {
				reg = <1>;
				nvmem-cells = <&macaddr_bdinfo_de00 2>;
				nvmem-cell-names = "mac-address";
			};
```

---

openwrt-ai (commented):
Re-reviewed after the force-push (`24543546` → `a95d9635`). The `filogic.mk` regression from last round is fixed correctly — `KERNEL`/`KERNEL_INITRAMFS` are gone and `KERNEL_LOADADDR`/`UBINIZE_OPTS`/`BLOCKSIZE`/`PAGESIZE` are back, so `Device/cudy_tr3600-v1` now matches [`Device/glinet_gl-mt3600be`](https://github.com/openwrt/openwrt/blob/a95d9635671a298bcba5544791b3d660081ebb43/target/linux/mediatek/image/filogic.mk#L1872-L1877) line for line, and `IMAGE_SIZE := 237824k` matches the `ubi` partition (`0xe840000`) exactly.

Checked the rest of the file against the target while I was here: every referenced dtsi label (`mdio`, `pio`, `pwm`, `fan`, `ssusb`, `tphyu3port0`, `spi0_flash_pins`, `pcie0_pins`, `uart0_pins`, `i2p5gbe_led0_pins`) resolves in `mt7987.dtsi`, the `pwms`/`button-mode` forms match the in-tree MT7987 boards, and both `mt7987-2p5g-phy-firmware` and `kmod-mt7990-firmware` exist. The commit message matches the diff. CI on the head commit is green.

One leftover comment-placement nit inline.

---

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,280 @@
...
+				nvmem-cell-names = "mac-address";
+			};
+
+			band@1 {
+				reg = <1>;
+				nvmem-cells = <&macaddr_bdinfo_de00 2>;
+				nvmem-cell-names = "mac-address";
+			};
+
+		};
```
openwrt-ai, line 159: nit: removing the trailing comment left a blank line as the last thing inside `wifi@0,0`. No other node in this file (or in [mt7987a-glinet-gl-mt3600be.dts:153-166](https://github.com/openwrt/openwrt/blob/df24eaa85a32874051629fdf20c8bf2a4fe7884b/target/linux/mediatek/dts/mt7987a-glinet-gl-mt3600be.dts#L153-L166)) ends with one.

```suggestion
		};
```

---

openwrt-ai (commented):
Re-reviewed after the force-push (`a95d9635` → `df24eaa8`). The only change is in `wifi@0,0`: the blank line between the node's properties and `band@0` is restored, and the orphaned `/* MT7990 calibration data ... */` comment is dropped — both as suggested last round. Commit message is byte-identical to the previous revision and still matches the diff, so no commit checks.

The other still-open threads look answered to me: the `backup` partition label is explained in the PR body ("The vendor-defined backup label at 0x100000 is retained"), `memory` now carries `device_type = "memory"`, and the 2.4 GHz MAC is on base+3. The `05_set_preinit_iface` question from the 08-07 round is the one item I don't see addressed either way.

One leftover whitespace nit inline. CI reports no checks on the head commit yet, so there was nothing to ground against there.

---

wangjie96998:
作者辛苦了，加油，期待tr3600的openwrt固件

openwrt-ai (commented):

Re-reviewed after the force-push (`df24eaa8` → `6065ef93`). The only code change is the removal of the stray blank line at the end of `wifi@0,0`, which was the one leftover nit from the previous round. The commit message gained a paragraph about the `05_set_preinit_iface` entry, which accurately describes the hunk that is already in the diff and closes out the open question from the 08-07 round — so no commit checks.

Spot-checked the rest of the final state while here: `Device/cudy_tr3600-v1` keeps `KERNEL_LOADADDR`/`UBINIZE_OPTS`/`BLOCKSIZE`/`PAGESIZE`, `IMAGE_SIZE := 237824k` still matches the `ubi` partition (`0xe840000`), `SUPPORTED_DEVICES += R126` follows the Cudy naming used by the other entries in the file, the recipe is in alphabetical position, and `pwm1_pins` selects the `pwm1_0` group that [mt7987.dtsi:614-615](https://github.com/openwrt/openwrt/blob/6065ef938cb5df700526e0ae409a83ea69d3b8ff/target/linux/mediatek/dts/mt7987.dtsi#L614-L615) documents for PWM1, matching [`mt7987a-glinet-gl-mt3600be.dts:178-182`](https://github.com/openwrt/openwrt/blob/6065ef938cb5df700526e0ae409a83ea69d3b8ff/target/linux/mediatek/dts/mt7987a-glinet-gl-mt3600be.dts#L178-L182). All three FormalityCheck runs on the head commit are green.

---

wangjie96998:
@dengbo @blogic @hauke All CI checks have passed and all review comments from openwrt-ai have been resolved. Could you please take a look when you have time? Thanks!

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,279 @@
...
+		led-upgrade = &led_status_blue;
+		serial0 = &uart0;
+	};
+
+	chosen {
+		bootargs = "console=ttyS0,115200n1 earlycon=uart8250,mmio32,0x11000000 pci=pcie_bus_perf";
+		stdout-path = "serial0:115200n8";
+	};
+
+	/* Cudy TR3600 v1 has 512 MiB DDR4 at 0x40000000. */
```
aiamadeus: This should be `memory@40000000`.
Also, remove redundant comments.

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,279 @@
...
+	nvmem-cells = <&macaddr_bdinfo_de00 0>;
+	nvmem-cell-names = "mac-address";
+	phy-handle = <&phy15>;
+	phy-mode = "internal";
+	status = "okay";
+};
+
+&mdio {
+	phy1: phy@1 {
+		compatible = "ethernet-phy-id001c.c849";
```
aiamadeus: Why hardcode the PHY ID? It should be C45.

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,279 @@
...
+		reset-gpios = <&pio 42 GPIO_ACTIVE_LOW>;
+		reset-assert-us = <100000>;
+		reset-deassert-us = <120000>;
+	};
+
+	phy15: phy@f {
+		compatible = "ethernet-phy-ieee802.3-c45";
+		reg = <15>;
+		pinctrl-names = "i2p5gbe-led";
+		pinctrl-0 = <&i2p5gbe_led0_pins>;
```
aiamadeus: Does this LED really exist?

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,279 @@
...
+		#size-cells = <2>;
+		device_type = "pci";
+
+		wifi@0,0 {
+			compatible = "mediatek,mt76";
+			reg = <0x0000 0 0 0 0>;
+
+			#address-cells = <1>;
+			#size-cells = <0>;
+			ieee80211-freq-limit = <2400000 2500000>,
```
aiamadeus, line 137: Why is `ieee80211-freq-limit` needed?

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,279 @@
...
+					       <5170000 5835000>;
+
+			nvmem-cells = <&eeprom_factory_0>;
+			nvmem-cell-names = "eeprom";
+
+			/* Offset 3 keeps the 2.4 GHz MAC distinct from LAN. */
```
aiamadeus: Drop redundant comments.

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,279 @@
...
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "BL2";
+				reg = <0x0 0x100000>;
+				read-only;
+			};
+
+			/* The official Cudy R126 layout names this writable partition "backup". */
```
aiamadeus: Drop redundant comments.

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
soapmancn: TR3600 is a dual-band device. Cudy specifies only 2.4 GHz and 5 GHz radios and FEMs, with no 6 GHz RF path or antenna. The limit prevents the MT7990 from exposing an unsupported 6 GHz band.

ahaodev (approved):

openwrt-ai (commented):

Re-reviewed after the force-push (`6065ef93` → `53f7fb12`, plus a rebase onto `f6bc2a3b`). Ignoring the rebase, the only code changes are in `mt7987b-cudy-tr3600-v1.dts` and all four address @aiamadeus's round:

- `memory` → `memory@40000000`. The `memory` node in [mt7987b.dtsi:13-15](https://github.com/openwrt/openwrt/blob/53f7fb12da7317a7bd01c89b3a9ffa9619d13c15/target/linux/mediatek/dts/mt7987b.dtsi#L13-L15) is no longer overridden, but it carries no `device_type`, so `early_init_dt_scan_memory()` skips it and only the new 512 MiB node is seen — same arrangement as [`mt7987a-routerich-be7200.dts:110`](https://github.com/openwrt/openwrt/blob/53f7fb12da7317a7bd01c89b3a9ffa9619d13c15/target/linux/mediatek/dts/mt7987a-routerich-be7200.dts#L110).
- `phy@1`: `ethernet-phy-id001c.c849` → `ethernet-phy-ieee802.3-c45`. The RTL8221B-VB-CG still binds via [`rtl8221b_vb_cg_c45_match_phy_device()` in realtek_main.c:1563-1566](https://github.com/gregkh/linux/blob/2c1a4a2c93a89b3f56b4936cd2ca4c8546a75e9e/drivers/net/phy/realtek/realtek_main.c#L1563-L1566) (v6.18.44, matching this target's `KERNEL_PATCHVER:=6.18`), and the `reset-gpios` on the node is honoured by the MDIO core before the bus scan. Same shape as [`mt7987a-hiveton-h5000m.dts:111-121`](https://github.com/openwrt/openwrt/blob/53f7fb12da7317a7bd01c89b3a9ffa9619d13c15/target/linux/mediatek/dts/mt7987a-hiveton-h5000m.dts#L111-L121).
- `i2p5gbe-led` pinctrl dropped from `phy@f` — this is the one that leaves the board differing from all four other MT7987 boards with the internal 2.5G PHY, but that was exactly the reviewer's "Does this LED really exist?" question, so I'm treating it as answered rather than re-raising it.
- Three explanatory comments removed.

The commit message is byte-identical to the previous revision and none of its claims are invalidated by the above, so no commit checks. GitHub reports no check runs on the head commit yet, so there was nothing to ground against on CI.

---

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,274 @@
...
+
+&eth {
+	status = "okay";
+};
+
+&fan {
+	fan-supply = <&fan_3p3v>;
+	pwms = <&pwm 1 50000 0>;
+	status = "okay";
+};
```
hauke, line 86: Does the fan really work?
Please try it like this: `echo 255 > /sys/class/hwmon/hwmon*/pwm1` and then `echo 0`, 
With the first it should run, with the second it should not run.

If it does not work please try: `pwms = <&pwm 0 50000 0>` with a `groups = "pwm0"` pin group.

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,274 @@
...
+			function = LED_FUNCTION_STATUS;
+			gpios = <&pio 48 GPIO_ACTIVE_LOW>;
+		};
+
+		led_status_white: led-1 {
+			color = <LED_COLOR_ID_WHITE>;
+			function = LED_FUNCTION_STATUS;
+			gpios = <&pio 46 GPIO_ACTIVE_LOW>;
+		};
+	};
```
hauke: Is this LED white or red?

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
soapmancn: The vendor R126 DTS uses PWM0 with the `pwm0` pin group (PIN13) and a 50,000 ns period; GPIO6 is the active-high `fan_switch`. I changed this to `pwms = <&pwm 0 50000 0>` and `pinctrl-0 = <&pwm_pins>`, while keeping GPIO6 modeled as the fan supply enable. This is based on the vendor DTS; I have not claimed a runtime hardware test.

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
soapmancn: It is red. The vendor R126 DTS names GPIO46 `red`; GPIO48 is `blue`. I changed `LED_COLOR_ID_WHITE` to `LED_COLOR_ID_RED` and updated the commit and PR descriptions.

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
hauke: Please do a runtime test too. Maybe the vendor DTS is for a different driver which needs a different dts format.

openwrt-ai (commented):

Re-reviewed after the force-push (`53f7fb12` → `6d99cc1e`). The only file touched is `mt7987b-cudy-tr3600-v1.dts`, and both changes answer @hauke's round:

- **LED white → red.** `led_status_white`/`LED_COLOR_ID_WHITE` → `led_status_red`/`LED_COLOR_ID_RED` on GPIO46, with the `led-boot` alias updated to match. The commit message and PR body were updated in the same push ("one red and one blue status LED"), so DTS / commit / body agree. The remaining alias shape (`led-failsafe`/`led-running`/`led-upgrade` all on the blue LED) is unchanged and matches [`mt7987a-glinet-gl-mt3600be.dts:19-22`](https://github.com/openwrt/openwrt/blob/6d99cc1ee93424cae10285f83f015fc8a7c31713/target/linux/mediatek/dts/mt7987a-glinet-gl-mt3600be.dts#L19-L22).
- **PWM1 → PWM0.** `pwms = <&pwm 0 50000 0>`, the local `pwm1_pins` override is dropped, and `&pwm` now takes the shared `&pwm_pins` from [mt7987.dtsi:610-621](https://github.com/openwrt/openwrt/blob/6d99cc1ee93424cae10285f83f015fc8a7c31713/target/linux/mediatek/dts/mt7987.dtsi#L610-L621), whose default mux is `groups = "pwm0"`. That group is [`mt7987_pwm0_pins[] = {13}`](``https://github.com/openwrt/openwrt/blob/6d99cc1ee93424cae10285f83f015fc8a7c31713/target/linux/mediatek/patches-6.18/360-pinctrl-add-pinctrl-driver-for-mt7987-from-sdk.patch#L513)`` (PWM0@PIN13), so the PWM index and the pin group now refer to the same channel — they didn't before. PIN13 is not claimed by any other node in this board (the GPIOs in use are 0, 1, 6, 38, 42, 46, 48), and the `&pwm { pinctrl-0 = <&pwm_pins>; }` form matches [`mt7987a-bananapi-bpi-r4-lite.dts:238-242`](``https://github.com/openwrt/openwrt/blob/6d99cc1ee93424cae10285f83f015fc8a7c31713/target/linux/mediatek/dts/mt7987a-bananapi-bpi-r4-lite.dts#L238-L242).`` The commit message's "PWM1" references were updated to "PWM0" in the same push, so it still matches the diff — no commit checks. (@hauke's request for a runtime `/sys/class/hwmon/hwmon*/pwm1` check is still the open item; the DTS is only verified against the vendor DTS here.)

Everything else in the file is byte-identical to the previous revision. All CI on the head commit is green — filogic/mt7622/mt7623/mt7629 kernel builds, kernel-patch check, `build-pr-profile`, and the three FormalityCheck runs (the one `cancelled` `build-pr-profile` job is a superseded duplicate run of the same workflow).

---

target/linux/mediatek/dts/mt7987b-cudy-tr3600-v1.dts
```diff
@@ -0,0 +1,265 @@
...
+		reg = <0 0x40000000 0 0x20000000>;
+	};
+
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		led_status_blue: led-0 {
+			color = <LED_COLOR_ID_BLUE>;
+			function = LED_FUNCTION_STATUS;
+			gpios = <&pio 48 GPIO_ACTIVE_LOW>;
```
xujiegb, line 39: It appears that my TR3600 is displaying a white LED.

e2720515-cloud:
Thanks for working on TR3600 v1 support. We have published a small, experimental source-only overlay and sanitized hardware notes to help compare observations, rather than open a duplicate device-support PR.

Our DTS comparison is against the 25.12 PR #24596 head `046aec0dccd90f5a156cb8e9725c121c81955dd3`, not a claim that every difference still applies to this main-branch PR. The notes derive from the vendor development configuration and a bounded comparison with one TR3600 v1:

| Item | Observation | Limit |
| --- | --- | --- |
| Memory | One 512 MiB memory node | Running FDT observation |
| External PHY at MDIO address 1 | Generic Clause 45; GPIO42 active-low reset, 100 ms assert/deassert; GPIO41 level-low IRQ | Configuration evidence; exact silicon identity unresolved |
| Fan | PWM1, 50,000 ns period; GPIO5 rising-edge tach input | Configuration evidence, not closed-loop validation |
| Mode control | GPIO0 configured active-high | Physical switch polarity not verified |
| Wi-Fi MAC cell offsets | Band 0 uses offset 0; band 1 uses offset 2 | Running FDT observation, not proof of uniqueness across all interfaces |

Pinned public notes: https://github.com/e2720515-cloud/openwrt-cudy-tr3600/blob/7aa17aba83838955bab282e1277ce3c08f49589d/docs/hardware-notes.md

Scope and limitations: https://github.com/e2720515-cloud/openwrt-cudy-tr3600/blob/7aa17aba83838955bab282e1277ce3c08f49589d/docs/known-limitations.md

There is historical isolated-build evidence, but a clean reproduction from the current public source and real RAMFS/UBI integration remain unqualified. The repository's separate experimental dual-slot updater is **not** being proposed for merge here: two-direction upgrades/recovery and release-image qualification are still outstanding. No flashable release is claimed.

This comment is discussion/observation feedback, not a signed patch submission. We have not published raw device dumps, calibration data, device MACs or user configuration. Would a narrowly scoped, separately tested hardware-correction series against main be useful, or would you prefer to incorporate relevant observations into this existing series first?

aiamadeus:
@soapmancn Do you plan to keep working on this? I just bought
a TR3600 last week and support it according to the [GPL source](https://www.cudy.com/zh-cn/pages/download-center/tr3600-1-0).
[All reported issues (fan/led/mac address) have been fixed](https://github.com/aiamadeus/openwrt/commit/5dafc0b17853cee01800f40e2bdc6118a390eb38).

ju5t4fun:
Runtime test results from a retail TR3600 v1. This should answer @hauke's fan question and the LED color question.

**Setup:** stock → Cudy intermediate firmware (`25.12-SNAPSHOT-CUDY r33005`, Linux 6.12.93) → sysupgrade to a local build of `main` @ `796b5590ce` + @aiamadeus's [5dafc0b](https://github.com/aiamadeus/openwrt/commit/5dafc0b17853cee01800f40e2bdc6118a390eb38) (Linux 6.18.44). The image boots and runs fine.

### Fan
- **Intermediate firmware** (vendor DTB: `pwms = <&pwm 1 50000 0>`, `pwm1_0` group): the fan spins at a constant speed. Writing `255` and `0` to `/sys/class/hwmon/hwmon*/pwm1` changes nothing.
- **aiamadeus's DTS** (`pwms = <&pwm 0 50000 0>`, `&pwm_pins`): `echo 0` stops the fan, and `echo 255` makes it clearly louder. The `step_wise` governor then puts it back to cooling level 1 (`pwm1=128`) within about 1 s (CPU ~57 °C, trip at 40 °C).

So **PWM0 is correct**. PWM1, as in the vendor DTB, doesn't control the fan.

### Status LEDs
I lit each LED on its own:
- **GPIO46 = red**
- **GPIO48 = white** (it can look slightly bluish-white)

Both the intermediate firmware's DTB (46 "white", 48 "blue") and `R126.dts` (46 red, 48 blue) are wrong. aiamadeus's DTS (46 red, 48 white) matches the hardware. #24581 currently has GPIO48 as blue.

### MAC addresses
The intermediate firmware's DTB uses these `bdinfo` 0xde00 offsets:
- gmac1/LAN +0
- gmac0/WAN +1
- `band@0` +0
- `band@1` +2

aiamadeus's DTS matches that. #24581 still uses +3 for `band@0`.

### Port LEDs: neither Ethernet port has one
**LAN (MT7987 internal 2.5GbE PHY)**
- Both vendor DTBs (stock Linux 5.4.281 `R126`, and the intermediate firmware) set `pinctrl-names = "i2p5gbe-led"; pinctrl-0 = <&i2p5gbe_led0_pins>;` (GPIO45) on `phy@f`.
- I restored that pinctrl in a test build:
  - The `Fail to set LED pins!` error went away.
  - debugfs confirms `pin 45 (2P5G_LED0): device mdio-bus:0f function led group 2p5gbe_led0`.
  - `LED0_ON_CTRL` reads `0x8087` (enabled, on for 10/100/1000/2500 link).
- With a 1 Gbps link up, nothing lit. I also wrote `0xc087` (inverted polarity) with mdio-tools, and nothing lit either.

So the answer to @aiamadeus's "Does this LED really exist?" is no. The vendor pinctrl is just the MT7987 RFB default.

**WAN (RTL8221B)**
- Neither vendor DT describes any LEDs on it.
- In a test build I described LED0–LED2 in a `leds` node under `phy@1` and forced each one on and off through `/sys/class/leds/mdio-bus:01::wan-N/brightness`, with a 1 Gbps link up. Nothing lit at either level.

Leaving port LEDs out of the DTS is correct. The only visible LEDs are the red/white status LEDs.

Nit: the node is `phy15: phy@15` with `reg = <15>`; the unit address should be `phy@f`.

### Flash layout / upgrade path
The vendor U-Boot is dual-image. One UBI device holds:
- `u-boot-env`
- `kernel`, `rootfs` (slot 0; the stock firmware was here)
- `kernel2`, `rootfs2` (slot 1; the intermediate firmware goes here)
- `rootfs_data` (shared)

U-Boot picks the slot from `dual_boot.current_slot` and passes `ubi.rootfs_volume=<vol>` plus `boot_param.*` on the command line.

When the OpenWrt image is flashed from the intermediate firmware:
- Its `nand.sh` writes `kernel`/`rootfs` and sets `dual_boot.current_slot=0`.
- U-Boot then passes `ubi.rootfs_volume=rootfs`. Upstream ignores that parameter but mounts `rootfs` by default anyway, so it boots.

`main` has no dual-boot handling for `cudy,tr3600-v1`, so later sysupgrades only rewrite slot 0 and never touch the env. That works as long as U-Boot stays on slot 0. If it ever falls back to slot 1, the intermediate firmware boots again, so it isn't a brick.

`fw_printenv` works with `/dev/ubi0_0 0x0 0x80000 0x80000 1`.

### uboot-envtools conflict (for @aiamadeus)
When cherry-picked onto current `main`, the `mediatek_filogic` hunk conflicts. `main` now has JioRouter's own corrected block:
```
jiorouter,ax6000-jidu6101|jiorouter,ax6000-jidu6j01
... "0x80000" "0x1f000" "5"
```
A naive resolution keeps a second, stale `cudy,tr3600-v1|jiorouter,ax6000-jidu6101` case above it, and that one would take precedence for JIDU6101. The TR3600 entry should be its own case.

### Other
- SPI-NAND probes as `Micron SPI NAND` 256 MiB with 0 bad blocks. The body of #24581 says ESMT F50L2G41XA.
- WAN PHY binds as `RTL8221B-VB-CG 2.5Gbps PHY` on `mdio-bus:01` (generic `ethernet-phy-ieee802.3-c45` compatible).
- MT7990 (EEPROM ID `0x7993`) comes up on both bands.
- USB RNDIS tethering works on the USB-A port.

untilscour:
> ### MAC addresses
> The intermediate firmware's DTB uses these `bdinfo` 0xde00 offsets:
> 
> * gmac1/LAN +0
> * gmac0/WAN +1
> * `band@0` +0
> * `band@1` +2
> 
> aiamadeus's DTS matches that. #24581 still uses +3 for `band@0`.

Is it a good idea to have the same MAC address for 2.4 GHz and LAN Ethernet? It's not mandatory to follow the vendor blindly; what are the respective MAC addresses on the stock firmware?

ju5t4fun:
> Is it a good idea to have the same MAC address for 2.4 GHz and LAN Ethernet? It's not mandatory to follow the vendor blindly; what are the respective MAC addresses on the stock firmware?

@untilscour Good question — I went back through the flash dumps for this, and the short answer is: on stock the Wi-Fi MACs do not come from the Cudy base MAC at all.

**Where the board keeps its MACs.** bdinfo 0xde00 holds a single Cudy-OUI base address (`d4:0d:ab:xx:xx:xx` on my unit, masked — full value off-thread if you want it). I searched the whole 240 MiB NAND dump raw plus both rootfs images after extracting them, and that byte sequence occurs exactly once. Separately, the MT7990's factory eeprom (the `Factory` partition) carries two per-unit addresses at the offsets upstream mt7996 calls `MT_EE_MAC_ADDR` (0x004) and `MT_EE_MAC_ADDR2` (0x00a) — both with MediaTek's OUI, distinct from each other and from the Cudy base:

| | Source | Value |
|---|---|---|
| LAN (`gmac1`) | bdinfo 0xde00 + 0 | `d4:0d:ab:xx:xx:xx` |
| WAN (`gmac0`) | bdinfo 0xde00 + 1 | `d4:0d:ab:…:xx+1` |
| 2.4 GHz | factory eeprom 0x004 | `00:0c:43:xx:xx:xx` |
| 5 GHz | factory eeprom 0x00a | `00:0c:43:xx:xx:xx` (different value) |

**Stock firmware** (Linux 5.4.281, DTB "R126"): its device tree has no `band@` nodes and no `mac-address`/`nvmem-cells` on the mt7996 node, i.e. nothing overrides the eeprom. In the driver that is exactly the path that keeps them: `mt7996_eeprom_init()` fills the 2.4 GHz address from eeprom 0x004, the second radio from 0x00a, and `mt76_eeprom_override()` only replaces them if a band node (or the device node) provides a valid `mac-address` — otherwise the eeprom values are what the phy registers with. So on stock, 2.4 GHz and LAN are different addresses, from different sources, and there is no duplication anywhere.

The duplication only appears in Cudy's own OpenWrt firmware: the intermediate build's DTB adds `band@0` = 0xde00 + 0 and `band@1` = 0xde00 + 2, which overrides the eeprom and puts the 2.4 GHz radio on the LAN address. That is a choice Cudy made when they moved this board to OpenWrt — not something stock ever did, and not something the hardware requires (nothing else in flash references those two eeprom fields or the band cells).

**So to your question: no, I don't think we should copy it.** Three options, roughly:

1. what #24581 has now — LAN +0 (label), WAN +1, 5G +2, 2.4G +3: four distinct addresses, Cudy OUI, and the only deviation from the vendor DTB is the value that would otherwise collide;
2. drop the `band@` cells entirely — then the phy keeps the eeprom addresses, which is literally what stock used and also has no duplicates. Precedent in main: hiveton-h5000m and bpi-r4-lite define no Wi-Fi MAC cells either;
3. `band@0` = +0 — matches the intermediate firmware byte for byte, but 2.4 GHz then bridges with the LAN port under an identical address, which is the thing you are asking about.

I'd take 1 (no code change, 5G keeps the vendor value), and 2 if the preference is "exactly what the factory firmware did". Either is fine with me; only 3 looks wrong. Happy to spell whichever one we pick out in the commit message.

One data point either way: the build I described above used aiamadeus's cells, i.e. option 3's layout, and it runs fine on my unit — so this is not a "something breaks" question, it is about which layout we want to encode.

untilscour:
> 2. drop the `band@` cells entirely — then the phy keeps the eeprom addresses, which is literally what stock used and also has no duplicates. Precedent in main: hiveton-h5000m and bpi-r4-lite define no Wi-Fi MAC cells either;
> 
> [...] 2 if the preference is "exactly what the factory firmware did". Either is fine with me; only 3 looks wrong. Happy to spell whichever one we pick out in the commit message.

In OpenWrt the standard practice is to match all the MAC addresses to the actual stock firmware as much as possible (not the vendor's signed intermediate OpenWrt)

someone should check and verify all 4 actual MAC addresses on a live unit running the stock firmware and work out where they come from and how to get them here

ju5t4fun:
> In OpenWrt the standard practice is to match all the MAC addresses to the actual stock firmware as much as possible (not the vendor's signed intermediate OpenWrt)
> 
> someone should check and verify all 4 actual MAC addresses on a live unit running the stock firmware and work out where they come from and how to get them here

@untilscour Both checks are done, and the live one corrects part of what I wrote above.

**1. A build with the `band@` cells removed** (upstream mt76, run on my unit): the driver then takes the Wi-Fi MACs from the MT7990 factory eeprom — `00:0c:43:26:f4:20` (2.4 GHz) and `00:0c:43:26:60:11` (5 GHz), i.e. the values at 0x004 / 0x00a, which upstream mt7996 defines as `MT_EE_MAC_ADDR` / `MT_EE_MAC_ADDR2`. Verified live, not inferred: `phy0` `macaddress` = `00:0c:43:26:f4:20`, `phy0` `addresses` = `[00:0c:43:26:f4:20 00:0c:43:26:60:11]`, the 5 GHz AP came up on `00:0c:43:26:60:11`, and netifd's wdev helper returned `00:0c:43:26:f4:20` for the first 2.4 GHz vif. So dropping the cells gives the eeprom addresses — it does **not** reproduce stock.

**2. A live stock run** on the same unit (I put the stock slot back and read it there): stock's Wi-Fi driver, the MTK SDK blob in `mt_wifi.ko`, logs its own BSS addresses at startup:

```
daemon.warn wireless: MSG_INFO: neighbor add bss ra0:Cudy-A2F5  (D4:0D:AB:9F:A2:F5)
daemon.warn wireless: MSG_INFO: neighbor add bss rai0:Cudy-A2F5-5G (D4:0D:AB:9F:A2:F7)
```

`ra0` is the 2.4 GHz AP and `rai0` the 5 GHz one; the SSID suffix `A2F5` is the last three bytes of the bdinfo MAC, and the stock UI's WAN page shows `D4:0D:AB:9F:A2:F6` (= +1) live. So stock's addresses come from bdinfo, exactly 0/+1/+2 — the eeprom values are only what the *upstream* driver picks when nothing overrides it.

**So I have to correct myself:** in the previous comment I said stock's Wi-Fi MACs are the eeprom ones. They are not. Stock does use the Cudy base MAC, and the 2.4 GHz radio **does** share the LAN address — the thing your question on 09-18 suspected. Sorry for the wrong lead.

That means "match stock" here is `band@0` = 0xde00 + 0 and `band@1` = 0xde00 + 2, i.e. the vendor DTB and aiamadeus's values, so **#24581's +3 is the deviation from stock, not the vendor's +0**. My option 2 from the previous comment (drop the cells) is wrong and should be ignored.

Two notes for whoever writes the commit message, because a reviewer will ask about the duplicate:

- it is real, it is what the stock firmware does, nothing in the hardware depends on it, and OpenWrt runs fine with it (my unit ran that layout for several days);
- only the four default addresses match stock byte for byte: for the additional BSSes OpenWrt derives its own addresses (local bit + index), while the SDK produced `D6:0D:AB:8F:A2:F5` / `D6:0D:AB:8F:A2:F7` for the guest BSSes.

All addresses above are from this unit (bdinfo 0xde00 = `D4:0D:AB:9F:A2:F5`); both runs are the same retail TR3600 v1, done back to back today.

Issue #12366, opened 2023-04-10
IGMP stopped working after commit that updates kernel to 5.15.86 on Turris Omnia


PedroM4:
### Describe the bug

IPTV channels stopped working after updating my Turris Omnia. I narrowed down the start of the issue to commit:

https://github.com/openwrt/openwrt/commit/738b04c88128d9963b81bff68426980f587f22d9

### OpenWrt version

738b04c88128d9963b81bff68426980f587f22d9

### OpenWrt target/subtarget

mvebu/cortexa9

### Device

Turris Omnia

### Image kind

Official downloaded image

### Steps to reproduce

update router firmware with a version that is above the commit that introduces the issue ( including newer kernel versions that that one ) keeping configuration.
IPTV channels stop working ( omcproxy, but same happens with igmproxy ).
Flash a version prior to that commit and IPTV channels work.

### Actual behaviour

IPTV channels stop working ( omcproxy, but same happens with igmproxy ).
Flash a version prior to that commit and IPTV channels work.

### Expected behaviour

IPTV channels should work

### Additional info

maybe these Omnia specific changes of kernel .86 may have something to do with it:
Pali Rohár (9):
ARM: dts: dove: Fix assigned-addresses for every PCIe Root Port
ARM: dts: armada-370: Fix assigned-addresses for every PCIe Root Port
ARM: dts: armada-xp: Fix assigned-addresses for every PCIe Root Port
ARM: dts: armada-375: Fix assigned-addresses for every PCIe Root Port
ARM: dts: armada-38x: Fix assigned-addresses for every PCIe Root Port
ARM: dts: armada-39x: Fix assigned-addresses for every PCIe Root Port
ARM: dts: turris-omnia: Add ethernet aliases
ARM: dts: turris-omnia: Add switch port 6 node
arm64: dts: armada-3720-turris-mox: Add missing interrupt for RTC

### Terms

- [X] I am reporting an issue for OpenWrt, not an unsupported fork.

stklcode:
First, I have no idea about the actual issue and don't even have any testcases for that.

But what comes to my mind is, if you already narrowed it down to a small set of potentially breaking changes:

> maybe these Omnia specific changes of kernel .86 may have something to do with it:
> ARM: dts: armada-38x: Fix assigned-addresses for every PCIe Root Port (https://github.com/gregkh/linux/commit/f27dd04e44d82a96e0e7be4d5fb48c84940a7d6e)
[...]

PR #24582, opened 2026-08-05
mvebu: fix IGMP on Turris Omnia by disabling switch port 6

nicorikken:
mvebu: fix IGMP on Turris Omnia by disabling switch port 6

Since Linux kernel 5.15.86 IGMP no longer worked on Turris Omnia.
This has been traced back to a change in the devicetree.
Port 6 is a second CPU port wired to eth0; DSA never selects it as
conduit, but its mere presence breaks IGMP.

It was pinpointed to this part of the devicetree.
PedroM4 successfully got it to work for kernel 6.1.38 and 5.15.120.

The patch has been updated for kernel 6.12 and now 6.18.
Instead of removing the new ethernet-port@6 node, it is disabled.
It has been tested for both OpenWrt 25.12.1 and OpenWrt git commit
e9a6e45556e66067f7359509a7866cdbb8fe3f29 and fixes the issue.
The IGMP config from OpenWrt 22.03 that remained now just worked.

Link: https://github.com/openwrt/openwrt/issues/12366
Signed-off-by: Nico Rikken <nico@nicorikken.eu>


nicorikken:
The commit was based on https://github.com/openwrt/openwrt/tree/v25.12.1 Seeing the merge issues, I'll have to update the branch to the latest repository state.

target/linux/mvebu/patches-6.18/922-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
```diff
@@ -0,0 +1,20 @@
+--- a/arch/arm/boot/dts/marvell/armada-385-turris-omnia.dts
```
openwrt-ai: This patch lands in a directory that is never used on the target branch, so the fix would be a no-op after rebase. `main` has mvebu on [`KERNEL_PATCHVER:=6.18`](https://github.com/openwrt/openwrt/blob/e9a6e45556e66067f7359509a7866cdbb8fe3f29/target/linux/mvebu/Makefile#L12) and there is no `target/linux/mvebu/patches-6.12/` directory there at all — only `patches-6.18/`. (The branch is still based on v25.12.1, where mvebu was on 6.12, which is why it looks right locally.) Move the file to `target/linux/mvebu/patches-6.18/`.

nit: also reconsider the number. In `patches-6.18/` the 9xx range is occupied by the IEI WT61P803 driver/doc patches (901–907, 913), while every DTS change lives in the 3xx range (300–326). A `3xx-` number would fit the existing layout better.

---

target/linux/mvebu/patches-6.18/922-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
```diff
@@ -0,0 +1,20 @@
...
++++ b/arch/arm/boot/dts/marvell/armada-385-turris-omnia.dts
+@@ -514,16 +514,7 @@
```
openwrt-ai: The hunk metadata was generated against a 6.12 tree and doesn't match 6.18: `ethernet-port@6` is at [armada-385-turris-omnia.dts:517 in v6.12.100](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/arch/arm/boot/dts/marvell/armada-385-turris-omnia.dts#L517) but at [line 518 in v6.18.41](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm/boot/dts/marvell/armada-385-turris-omnia.dts#L518), so this `@@ -514,16 +514,7 @@` header is off by one there.

Separately, the blank context line below (line 6 of this file) is completely empty — its leading space was stripped. GNU patch tolerates that, but it will not survive a refresh cleanly.

After moving the file into `patches-6.18/`, regenerate it with `make target/linux/refresh` (mvebu selected in `.config`) so the hunk header and context are correct for the kernel actually in use.

---

target/linux/mvebu/patches-6.18/922-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
```diff
@@ -0,0 +1,20 @@
...
+-			ethernet-port@6 {
+-				reg = <6>;
+-				ethernet = <&eth0>;
+-				phy-mode = "rgmii-id";
+-
+-				fixed-link {
+-					speed = <1000>;
+-					full-duplex;
+-				};
+-			};
```
openwrt-ai: Deleting the whole node is a larger and more fragile change than needed. DSA only walks *available* port nodes — [`for_each_available_child_of_node(ports, port)` in net/dsa/dsa.c:1294](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/net/dsa/dsa.c#L1294) — so adding `status = "disabled";` to `ethernet-port@6` makes the port disappear from the fabric exactly the same way, with a one-line patch that survives kernel bumps without context churn. That is also the established idiom in this target: [`310-linksys-use-eth0-as-cpu-port.patch`](``https://github.com/openwrt/openwrt/blob/e9a6e45556e66067f7359509a7866cdbb8fe3f29/target/linux/mvebu/patches-6.18/310-linksys-use-eth0-as-cpu-port.patch#L7-L8)`` disables an unwanted CPU path the same way rather than deleting nodes.

I'm not suggesting an inline replacement since the patch has to be regenerated anyway (see the comment on the hunk header).

Also worth stating in the commit message: the conduit for the user ports is not port 6. `dsa_tree_setup_cpu_ports()` assigns [the first CPU port in the tree](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/net/dsa/dsa.c#L422-L445), mv88e6xxx provides no `preferred_default_local_cpu_port`, and `ethernet-port@5` (`&eth1`) is declared first — so this change removes the *second, otherwise unused* CPU port, it does not move traffic off eth0. Making that explicit would help whoever revisits this at the next DSA change.

---

target/linux/mvebu/patches-6.18/922-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
```diff
@@ -0,0 +1,20 @@
...
++			/* port 6 is connected to eth0 */
```
openwrt-ai: nit: this leftover comment is the only in-file explanation, and it documents the wiring rather than the reason the node must be absent — a future reader bumping the kernel has no way to tell whether the removal is still needed. The filename is `Revert-…`-style, which normally implies a `From:`/`Subject:`/`Signed-off-by:` header (as in `101-arm64-dts-armada-3720-turris-mox-fix-usb3-phys.patch`, `321-…`, `913-…` in this same directory). Please add such a header describing the IGMP breakage, linking issue #12366, and naming the upstream commit that introduced `ethernet-port@6`, so the patch is self-explanatory and can be pointed at upstream.

---

openwrt-ai (commented):
## Commit checks

- `486b511c` "fix IGMP on Turris Omnia by patching devicetree" — the subject is missing the OpenWrt `<area>: ` prefix. The only file touched is under `target/linux/mvebu/`, so it should read something like `mvebu: fix IGMP on Turris Omnia by disabling switch port 6`. Trailer issues in the same message: `Reported-by: PedroM4` has no e-mail address (`Name <mail@example.com>` is the expected form), and `Co-developed-by: Stefan Kalscheuer <stefan@stklcode.de>` must be immediately followed by that person's own `Signed-off-by:` line. Minor typos: "successfuly" → "successfully", and "OpenWRT" → "OpenWrt" (twice).

The remaining review comments are inline on the new patch file. The most important one is that `target/linux/mvebu/patches-6.12/` does not exist on `main` — mvebu is on 6.18 there — so as-is the patch would never be applied once the branch is rebased.

---

target/linux/mvebu/patches-6.18/922-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
openwrt-ai: nit (carry-over): the directory problem from the earlier thread on this line is fixed, but the patch number still is not, and that comment's headline is now stale enough that the numbering point is easy to lose when the thread gets resolved.

In `target/linux/mvebu/patches-6.18/` the 9xx range is occupied by the IEI WT61P803 driver/documentation series ([`901-`…`907-`](https://github.com/openwrt/openwrt/tree/b22ff2414521bde0ff4883fa7384403348e9658e/target/linux/mvebu/patches-6.18)) plus `913-drivers_spi_add_mtpoe_device_compatible.patch` — i.e. driver additions. Every DTS change in this target lives in the 3xx range: `300-`…`326-`, plus `340-v7.2-arm64-dts-marvell-armada-37xx-mark-EIP97-as-dma-cohe.patch`. A name like `327-ARM-dts-turris-omnia-remove-switch-port-6-node.patch` would fit the existing layout.

---

target/linux/mvebu/patches-6.18/922-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
```diff
@@ -0,0 +1,20 @@
+--- a/arch/arm/boot/dts/marvell/armada-385-turris-omnia.dts
++++ b/arch/arm/boot/dts/marvell/armada-385-turris-omnia.dts
+@@ -515,16 +515,7 @@
```
openwrt-ai: nit: this hunk header is now correct for the kernel actually in use — `ethernet-port@6` is at [armada-385-turris-omnia.dts:518 in v6.18.41](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm/boot/dts/marvell/armada-385-turris-omnia.dts#L518), and `patch -p1 --dry-run` against that tree applies with no fuzz and no offset. So this is only about the surrounding prose being out of date.

The PR description was not updated along with the commit message and now contradicts it on the same facts: it still says *"The patch has been updated for kernel 6.12"* and *"It has been tested for OpenWRT 25.12.1"*, whereas the commit says 6.18 and names the tested git commit; it also still carries "successfuly" and the "OpenWRT" spelling that the commit message has since fixed. Worth refreshing the body so anyone landing on the PR page sees the current state rather than the pre-rebase one.

---

openwrt-ai (commented):
## Commit checks

- `b22ff241` "mvebu: fix IGMP on Turris Omnia by disabling switch port 6" — the rebase cleared both blockers from the previous round: the patch now lives in `target/linux/mvebu/patches-6.18/` (matching `KERNEL_PATCHVER:=6.18`) and the hunk header matches v6.18.41, which the green *Check Kernel patches* job confirms. The subject prefix and the typos are fixed too. Three trailer problems remain in the message:
  - `Reported-by: PedroM4` still has no e-mail address; the expected form is `Name <mail@example.com>`.
  - `Co-developed-by: Stefan Kalscheuer <stefan@stklcode.de>` must be immediately followed by that person's own `Signed-off-by:` line — a `Co-developed-by:` without a matching sign-off is incomplete.
  - `Fixes: https://github.com/openwrt/openwrt/issues/12366` misuses the trailer. `Fixes:` takes a commit reference (`Fixes: <12-char sha> ("<subject>")`); bug-tracker URLs go under `Link:`, which is what the rest of the tree uses for GitHub issues. If you want a genuine `Fixes:` here it should name the upstream kernel commit that added `ethernet-port@6`, which is also the commit this patch reverts.

The three earlier inline threads on the patch file are still open and only partly addressed — the missing `git am` header (also raised by the formality bot), the patch number, and the `status = "disabled";` alternative to deleting the node.

---

nicorikken:
I updated the patch to simplify it.

target/linux/mvebu/patches-6.18/922-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
nicorikken: Addressed by moving patch to number 341.

target/linux/mvebu/patches-6.18/922-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
nicorikken: Patch text updated to match commit message. Spell check and kernel update applied.

target/linux/mvebu/patches-6.18/341-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
```diff
@@ -0,0 +1,35 @@
+From b2907660e78b50f18796bc6dcdc358afcca7e331 Mon Sep 17 00:00:00 2001
+From: Nico Rikken <nico@nicorikken.eu>
+Date: Wed, 5 Aug 2026 18:31:04 +0200
+Subject: [PATCH] mvebu: fix IGMP on Turris Omnia by disabling switch port 6
```
openwrt-ai, line 4: This is a Linux kernel patch, so the `Subject:` should be a kernel subject, not the OpenWrt `<target>: ` area prefix — `mvebu:` is not a subsystem in the kernel tree. Every other git-am-style DTS patch in this directory follows kernel style: [`321-arm64-dts-marvell-reorder-ethernet-aliases-for-ESPRE.patch:4`](``https://github.com/openwrt/openwrt/blob/7458fe3cd05feb15d536f139aabe61538371f434/target/linux/mvebu/patches-6.18/321-arm64-dts-marvell-reorder-ethernet-aliases-for-ESPRE.patch#L4)`` uses `arm64: dts: marvell: ...`, [`326-…:4`](``https://github.com/openwrt/openwrt/blob/7458fe3cd05feb15d536f139aabe61538371f434/target/linux/mvebu/patches-6.18/326-arm64-dts-marvell-glinet-mv1000-fix-sdhci1.patch#L4)`` likewise. Omnia is 32-bit ARM, so the prefix is `ARM: dts:`.

```suggestion
Subject: [PATCH] ARM: dts: turris-omnia: disable switch port 6
```

Related, and the reason to fix the subject first: the filename is still `341-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch`, which describes the previous revision of this patch. Nothing is reverted or removed any more — the node stays and gains `status = "disabled";`. In this target the filename is derived from the subject, so renaming to something like `341-ARM-dts-turris-omnia-disable-switch-port-6.patch` keeps the two in sync.

---

target/linux/mvebu/patches-6.18/341-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
```diff
@@ -0,0 +1,35 @@
...
+
+It was pinpointed to this part of the devicetree.
+PedroM4 successfully got it to work for kernel 6.1.38 and 5.15.120.
+
+The patch has been updated for kernel 6.12 and now 6.18.
+It has been tested for both OpenWrt 25.12.1 and git commit
+e9a6e45556e66067f7359509a7866cdbb8fe3f29 and fixes the issue.
+The IGMP config from OpenWrt 22.03 that remained now just worked.
+
+Fixes: https://github.com/openwrt/openwrt/issues/12366
```
openwrt-ai, line 20: `Fixes:` takes a commit reference (`Fixes: <12-char sha> ("<subject>")`), not a bug-tracker URL. Tooling that parses this trailer — including stable-kernel scripts, since this header travels with the patch — will choke on a URL. OpenWrt's own convention for GitHub issues is `Link:`; a survey of the last 3000 commits on `main` finds 10 `Link: https://github.com/…` trailers and zero `Fixes: http…`.

```suggestion
Link: https://github.com/openwrt/openwrt/issues/12366
```

The same line is present in the git commit message and needs the same change there.

If you want a genuine `Fixes:` as well, it would name the upstream kernel commit that added `ethernet-port@6` to `armada-385-turris-omnia.dts` — that is the commit this patch neutralises, and naming it also tells a future reader when the workaround can be dropped.

---

target/linux/mvebu/patches-6.18/341-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
```diff
@@ -0,0 +1,35 @@
...
+
+Reported-by: PedroM4
+Co-developed-by: Stefan Kalscheuer <stefan@stklcode.de>
```
openwrt-ai: `Co-developed-by:` must be immediately followed by that person's own `Signed-off-by:` line — it is a co-authorship claim and is incomplete without the co-author's DCO sign-off. Only your `Signed-off-by:` follows on line 23.

This revision also introduced a mismatch: the trailer was dropped from the git commit message (`7458fe3c` ends at `Reported-by: PedroM4` / `Signed-off-by: Nico Rikken`) but is still here in the patch header, so the two now disagree about who authored the change.

Either get Stefan's `Signed-off-by:` and put it on both, or — if he only pointed you at the fix rather than writing it — drop the trailer here too, or downgrade it to `Suggested-by:` / `Helped-by:`, neither of which requires a sign-off.

---

target/linux/mvebu/patches-6.18/341-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
```diff
@@ -0,0 +1,35 @@
+From b2907660e78b50f18796bc6dcdc358afcca7e331 Mon Sep 17 00:00:00 2001
+From: Nico Rikken <nico@nicorikken.eu>
+Date: Wed, 5 Aug 2026 18:31:04 +0200
+Subject: [PATCH] mvebu: fix IGMP on Turris Omnia by disabling switch port 6
+
+Since Linux kernel 5.15.86 IGMP no longer worked on Turris Omnia.
+This has been traced back to a change in the devicetree.
+Port6 is connected to eth0, the interface used by the SoC for the LAN
+switch chip.
```
openwrt-ai, line 9: nit: this reads as if port 6 were the switch's conduit, which would make disabling it look alarming. It is not the conduit. `dsa_tree_setup_cpu_ports()` assigns [the first CPU port found in the tree](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/net/dsa/dsa.c#L417-L444), mv88e6xxx defines no `preferred_default_local_cpu_port`, and `ethernet-port@5` (`&eth1`) is declared first at [armada-385-turris-omnia.dts:507](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/arm/boot/dts/marvell/armada-385-turris-omnia.dts#L507). So user traffic already runs over eth1/port 5, and this patch removes a *second, otherwise unused* CPU port — nothing is moved off eth0.

Wording that says so ("port 6 is a second CPU port wired to eth0; DSA never selects it as conduit, but its mere presence breaks …") would both be accurate and explain why the fix is safe. It also matters for whoever revisits this at the next DSA change: as written, the message gives no clue what the actual failure mechanism is, only that removing the node makes IGMP work again.

---

target/linux/mvebu/patches-6.18/341-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
```diff
@@ -0,0 +1,35 @@
...
+
+It was pinpointed to this part of the devicetree.
+PedroM4 successfully got it to work for kernel 6.1.38 and 5.15.120.
+
+The patch has been updated for kernel 6.12 and now 6.18.
+It has been tested for both OpenWrt 25.12.1 and git commit
+e9a6e45556e66067f7359509a7866cdbb8fe3f29 and fixes the issue.
+The IGMP config from OpenWrt 22.03 that remained now just worked.
```
openwrt-ai, line 18: nit: the patch header and the git commit message are still not identical after the resync. `7458fe3c` carries an extra sentence right after line 14 — "Instead of removing the new ethernet-port@6 node, it is disabled." — which is the one sentence that actually describes what this patch does; it is missing here, so the header never states the change.

Two further points on this paragraph, given the header is a kernel-style one and reads as intended for netdev/upstream: "The patch has been updated for kernel 6.12 and now 6.18" is OpenWrt-workflow history that means nothing outside this tree, and "tested for OpenWrt 25.12.1 and git commit e9a6e455…" names an OpenWrt commit that no kernel reviewer can resolve. A kernel-facing header would say which kernel versions were tested on which hardware; the OpenWrt-specific provenance belongs in the git commit message only.

---

openwrt-ai (commented):
## Commit checks

- `7458fe3c` "mvebu: fix IGMP on Turris Omnia by disabling switch port 6" — the message now matches the diff, and the mechanical points from the last round are resolved: the patch is renumbered into the 3xx DTS range (`341-`, free in this directory), the hunk applies to v6.18.41 with no fuzz and no offset, and the blank context line has its leading space back. One trailer item is still open from the previous review: `Reported-by: PedroM4` has no e-mail address, and the expected form is `Name <mail@example.com>`. If PedroM4 does not want an address published, `Reported-by: PedroM4 (GitHub)` or simply crediting them in prose is preferable to a bare handle in a DCO trailer. The `Fixes:`-vs-`Link:` point is now an inline comment since it needs the same edit in the patch header.

On the approach itself: switching from deleting the node to `status = "disabled";` is the right call and behaves identically. `dsa_switch_touch_ports()` allocates a `dsa_port` for every index 0..`num_ports`-1 regardless of DT, and [`dsa_switch_parse_ports_of()` walks only available children](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/net/dsa/dsa.c#L1294), so port 6 stays `DSA_PORT_TYPE_UNUSED` exactly as it did when the node was absent — with a one-line patch that will survive kernel bumps far better than the old nine-line deletion.

Remaining inline comments are about the patch header being a kernel-facing artifact: subject style and filename, the `Fixes:` trailer, the unaccompanied `Co-developed-by:`, and two accuracy/sync nits in the body.

---

nicorikken:
Updated to fix remaining issues including remark about Reported-by footer.

target/linux/mvebu/patches-6.18/341-ARM-dts-turris-omnia-disable-switch-port-6.patch
```diff
@@ -0,0 +1,34 @@
...
+
+The patch has been updated for kernel 6.12 and now 6.18.
+Instead of removing the new ethernet-port@6 node, it is disabled.
+It has been tested for both OpenWrt 25.12.1 and git commit
+e9a6e45556e66067f7359509a7866cdbb8fe3f29 and fixes the issue.
+The IGMP config from OpenWrt 22.03 that remained now just worked.
+
+Fixes: https://github.com/openwrt/openwrt/issues/12366
+
+Signed-off-by: Nico Rikken <nico@nicorikken.eu>
```
openwrt-ai, line 22: nit: the PR description has drifted from the commit message again. `e26bef3d` drops `Reported-by: PedroM4` (correctly — a bare handle is not a valid DCO trailer, and the prose credit on line 12 covers it), but the PR body still carries that line right above the sign-off. Maintainers regularly take the PR body as the commit message when applying, so it is worth removing there too.

While editing the body: its last line renders as `Signed-off-by: Nico Rikken` with the address swallowed, because GitHub eats `<nico@nicorikken.eu>` as an HTML tag. Wrapping the trailer block in a fenced code block keeps the address visible.

---

openwrt-ai (commented):
Reviewed 1 new commit (`e26bef3d`), diffed against `7458fe3c` from the last round.

The revision closes the `Co-developed-by:` thread: the trailer is dropped from the patch header, so header and git commit message are byte-identical again, and both now carry the "Instead of removing the new ethernet-port@6 node, it is disabled." sentence — which was the missing half of the header/commit sync comment. `Reported-by: PedroM4` is gone from both as well; that matches the tree, where the last 3000 commits on `main` contain exactly one `Reported-by:` trailer and it has an address.

Nothing regressed mechanically: the hunk still applies to `arch/arm/boot/dts/marvell/armada-385-turris-omnia.dts` in v6.18.41 with no fuzz and no offset, the blank context line on line 32 keeps its leading space, and all three FormalityCheck runs on `e26bef3d` are green.

Four inline threads from the previous round are untouched by this push and still apply as written — the `Subject:` prefix plus the `Revert-…-Add-switch-port-6-node` filename that no longer describes the patch, `Fixes:` vs `Link:` for the issue URL, the "Port6 is connected to eth0" wording that implies port 6 is the conduit, and the OpenWrt-workflow provenance ("updated for kernel 6.12 and now 6.18", the OpenWrt commit hash) sitting in a kernel-style header.

---

target/linux/mvebu/patches-6.18/922-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
nicorikken: Patch migrated to kernel 6.18 and patch number changed to 341 to be in 3xx range.

target/linux/mvebu/patches-6.18/922-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
nicorikken: Patch regenerated. Should be fine now.

target/linux/mvebu/patches-6.18/922-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
nicorikken: Port is now disabled instead of deleting the node. Tested that it works.

target/linux/mvebu/patches-6.18/922-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
nicorikken: Patch has been updated to include `From:`/`Subject:`/`Signed-off-by:` headers.

target/linux/mvebu/patches-6.18/341-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
nicorikken: Removed since patch has changed significantly and because of lack of an email address.

target/linux/mvebu/patches-6.18/341-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
nicorikken: Advice implemented.

target/linux/mvebu/patches-6.18/341-Revert-ARM-dts-turris-omnia-Add-switch-port-6-node.patch
nicorikken: Updated patch to include a Fixes decorator referring to the Linux kernel commit https://github.com/torvalds/linux/commit/f87db2005f73876602211af0ee156817019b6bda

target/linux/mvebu/patches-6.18/341-ARM-dts-turris-omnia-disable-switch-port-6.patch
```diff
@@ -0,0 +1,35 @@
+From b2907660e78b50f18796bc6dcdc358afcca7e331 Mon Sep 17 00:00:00 2001
+From: Nico Rikken <nico@nicorikken.eu>
+Date: Wed, 5 Aug 2026 18:31:04 +0200
+Subject: [PATCH] mvebu: fix IGMP on Turris Omnia by disabling switch port 6
+
+Since Linux kernel 5.15.86 IGMP no longer worked on Turris Omnia.
+This has been traced back to a change in the devicetree.
+Port6 is connected to eth0, the interface used by the SoC for the LAN
+switch chip.
```
nicorikken: Addressed by adjusting the text.

target/linux/mvebu/patches-6.18/341-ARM-dts-turris-omnia-disable-switch-port-6.patch
```diff
@@ -0,0 +1,35 @@
...
+
+It was pinpointed to this part of the devicetree.
+PedroM4 successfully got it to work for kernel 6.1.38 and 5.15.120.
+
+The patch has been updated for kernel 6.12 and now 6.18.
+It has been tested for both OpenWrt 25.12.1 and git commit
+e9a6e45556e66067f7359509a7866cdbb8fe3f29 and fixes the issue.
+The IGMP config from OpenWrt 22.03 that remained now just worked.
```
nicorikken: Adjusted the text to be explicit that the commit ID is a OpenWRT commit.

target/linux/mvebu/patches-6.18/341-ARM-dts-turris-omnia-disable-switch-port-6.patch
nicorikken: Updated PR to match git commit.

target/linux/mvebu/patches-6.18/341-ARM-dts-turris-omnia-disable-switch-port-6.patch
```diff
@@ -0,0 +1,34 @@
+From b2907660e78b50f18796bc6dcdc358afcca7e331 Mon Sep 17 00:00:00 2001
+From: Nico Rikken <nico@nicorikken.eu>
+Date: Wed, 5 Aug 2026 18:31:04 +0200
+Subject: [PATCH] ARM: dts: turris-omnia: disable switch port 6
+
+Since Linux kernel 5.15.86 IGMP no longer worked on Turris Omnia.
+This has been traced back to a change in the devicetree.
+port 6 is a second CPU port wired to eth0; DSA never selects it as
```
openwrt-ai: nit: sentence starts lowercase.

```suggestion
Port 6 is a second CPU port wired to eth0; DSA never selects it as
```

The git commit message carries the same sentence and needs the same edit.

---

openwrt-ai (commented):
Reviewed 1 new commit; no new issues found beyond the inline nit.

---

target/linux/mvebu/patches-6.18/341-ARM-dts-turris-omnia-disable-switch-port-6.patch
nicorikken: Corrected.

openwrt-ai (commented):

---

target/linux/mvebu/patches-6.18/341-ARM-dts-turris-omnia-disable-switch-port-6.patch
```diff
@@ -0,0 +1,34 @@
+From b2907660e78b50f18796bc6dcdc358afcca7e331 Mon Sep 17 00:00:00 2001
+From: Nico Rikken <nico@nicorikken.eu>
+Date: Wed, 5 Aug 2026 18:31:04 +0200
+Subject: [PATCH] ARM: dts: turris-omnia: disable switch port 6
+
+Since Linux kernel 5.15.86 IGMP no longer worked on Turris Omnia.
+This has been traced back to a change in the devicetree.
+Port 6 is a second CPU port wired to eth0; DSA never selects it as
```
openwrt-ai, line 8: nit: the capitalisation fix landed here and in the commit message, but the PR description still reads `port 6 is a second CPU port wired to eth0;`, so body and commit message disagree again. Maintainers frequently apply the PR body verbatim as the commit message, so it is worth capitalising it there too.

---

openwrt-ai (commented):
Reviewed 1 new commit; no new issues found beyond the inline nit.

---

openwrt-ai (commented):

---

openwrt-ai (commented):

---

openwrt-ai (commented):

---

openwrt-ai (commented):

---

openwrt-ai (commented):

---

openwrt-ai (commented):

---

981213:
Hello! I asked GPT-5.6 to figure out the actual bug behind this problem and it came up with this patch: 
[401-net-dsa-mv88e6xxx-fix-global-CPU-destination-with-mu.patch](https://github.com/user-attachments/files/32428635/401-net-dsa-mv88e6xxx-fix-global-CPU-destination-with-mu.patch)
Could you test if it fixes the issue?

nicorikken:
@981213 I expect I should place the patch in `target/linux/mvebu/patches-6.18/`?

981213:
> @981213 I expect I should place the patch in `target/linux/mvebu/patches-6.18/`?

Yes.

BKPepe:
Ping @elkablo, he might be interested to look into it.

nicorikken:
I tried to build the suggested patch on 14651b96832a9f9cef3b7bdcfb3dfecf9161ac58, but failed:

```
SHELL= flock /home/user/openwrt/tmp/.root-copy.flock -c 'cp -fpR /home/user/openwrt/build_dir/target-arm_cortex-a9+vfpv3-d16_musl_eabi/linux-mvebu_cortexa9/packages/.pkgdir/kmod-leds-turris-omnia/. /home/user/openwrt/staging_dir/target-arm_cortex-a9+vfpv3-d16_musl_eabi/root-mvebu/'
touch /home/user/openwrt/staging_dir/target-arm_cortex-a9+vfpv3-d16_musl_eabi/root-mvebu/stamp/.kmod-leds-turris-omnia_installed
echo "kmod-leds-turris-omnia" >> /home/user/openwrt/staging_dir/target-arm_cortex-a9+vfpv3-d16_musl_eabi/pkginfo/linux.default.install
make[3]: Leaving directory '/home/user/openwrt/package/kernel/linux'
time: package/kernel/linux/compile#19.40#3.12#23.21
make[2]: Leaving directory '/home/user/openwrt'
make[1]: *** [package/Makefile:192: /home/user/openwrt/staging_dir/target-arm_cortex-a9+vfpv3-d16_musl_eabi/stamp/.package_compile] Error 2
make[1]: Leaving directory '/home/user/openwrt'
make: *** [/home/user/openwrt/include/toplevel.mk:233: world] Error 2
```

I'll try on a stable tag to see if that is better.

981213:
The patch is for main & 6.18 and I'm building just fine. It seems your error log indicates a problem in the package building step, which comes after kernel building. Could you build with -j1 V=s to see what exactly is causing the error?

nicorikken:
@981213 Thanks! There seems to be a problem with GCC. I found some clues on the forums, but nothing conclusive that I can use to fix it.
```
make[3]: Entering directory '/home/user/openwrt/toolchain/gcc/initial'
mkdir -p /home/user/openwrt/dl
SHELL= flock /home/user/openwrt/tmp/.gcc-14.3.0.tar.xz.flock -c '  	/home/user/openwrt/scripts/download.pl "/home/user/openwrt/dl" "gcc-14.3.0.tar.xz" "x" "" "@GNU/gcc/gcc-14.3.0"    '
Cannot find appropriate hash command, ensure the provided hash is either a MD5 or SHA256 checksum.
make[3]: *** [Makefile:29: /home/user/openwrt/dl/gcc-14.3.0.tar.xz] Error 255
make[3]: Leaving directory '/home/user/openwrt/toolchain/gcc/initial'
time: toolchain/gcc/initial/compile#0.08#0.01#0.10
    ERROR: toolchain/gcc/initial failed to build.
make[2]: *** [toolchain/Makefile:93: toolchain/gcc/initial/compile] Error 1
make[2]: Leaving directory '/home/user/openwrt'
make[1]: *** [toolchain/Makefile:89: /home/user/openwrt/staging_dir/toolchain-arm_cortex-a9+vfpv3-d16_gcc-14.3.0_musl_eabi/stamp/.toolchain_compile] Error 2
make[1]: Leaving directory '/home/user/openwrt'
make: *** [/home/user/openwrt/include/toplevel.mk:233: world] Error 2
```

PR #24583, opened 2026-08-05
ramips: add support for D-Link DIR-853 R3

grumworth:
Add support for the D-Link DIR-853 hardware revision R3.

Specifications:
- SoC: MediaTek MT7621AT
- Flash: 128 MiB NAND
- RAM: 128 MiB
- Wireless: MT7615 2.4/5GHz Dual-Band
- Ports: 4x LAN, 1x WAN, 1x USB 2.0

Hardware validated on both identified 128 MiB NAND variants:
- ESMT F59L1G81MB-25T
- AMD/Spansion S34ML01G2

### DIFF
--- a/target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
+++ b/target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
@@ -1,0 +1,197 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+#include "mt7621.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+
+/ {
+	compatible = "dlink,dir-853-r3", "mediatek,mt7621-soc";
+	model = "D-Link DIR-853 R3";
+
+	aliases {
+		label-mac-device = &gmac0;
+
+		led-boot = &led_wps;
+		led-failsafe = &led_wps;
+		led-running = &led_wps;
+		led-upgrade = &led_wps;
+	};
+
+	keys {
+		compatible = "gpio-keys";
+
+		reset {
+			label = "reset";
+			gpios = <&gpio 18 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
+
+		wps {
+			label = "wps";
+			gpios = <&gpio 14 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_WPS_BUTTON>;
+		};
+
+		wifi {
+			label = "wifi";
+			gpios = <&gpio 16 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RFKILL>;
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		usb {
+			function = LED_FUNCTION_USB;
+			color = <LED_COLOR_ID_GREEN>;
+			gpios = <&gpio 13 GPIO_ACTIVE_LOW>;
+			trigger-sources = <&ehci_port2>;
+			linux,default-trigger = "usbport";
+		};
+
+		led_wps: wps {
+			function = LED_FUNCTION_WPS;
+			color = <LED_COLOR_ID_GREEN>;
+			gpios = <&gpio 15 GPIO_ACTIVE_LOW>;
+		};
+	};
+};
+
+&nand {
+	status = "okay";
+
+	partitions {
+		compatible = "fixed-partitions";
+		#address-cells = <1>;
+		#size-cells = <1>;
+
+		partition@0 {
+			label = "u-boot";
+			reg = <0x0 0x80000>;
+			read-only;
+		};
+
+		partition@80000 {
+			label = "config";
+			reg = <0x80000 0x80000>;
+			read-only;
+		};
+
+		partition@100000 {
+			label = "factory";
+			reg = <0x100000 0x40000>;
+			read-only;
+
+			nvmem-layout {
+				compatible = "fixed-layout";
+				#address-cells = <1>;
+				#size-cells = <1>;
+
+				eeprom_factory_0: eeprom@0 {
+					reg = <0x0 0x4da8>;
+				};
+
+				macaddr_factory_4: macaddr@4 {
+					reg = <0x4 0x6>;
+				};
+
+				macaddr_factory_e006: macaddr@e006 {
+					compatible = "mac-base";
+					reg = <0xe006 0x6>;
+					#nvmem-cell-cells = <1>;
+				};
+			};
+		};
+
+		partition@140000 {
+			label = "kernel";
+			reg = <0x140000 0x360000>;
+		};
+
+		/* Kept intact for recovery to the vendor firmware. */
+		partition@4a0000 {
+			label = "stock-rootfs";
+			reg = <0x4a0000 0xb60000>;
+			read-only;
+		};
+
+		partition@1000000 {
+			label = "ubi";
+			reg = <0x1000000 0x4a20000>;
+		};
+
+		/* Contains vendor JFFS2 data and must not be overwritten. */
+		partition@5a20000 {
+			label = "stock-tail";
+			reg = <0x5a20000 0x2560000>;
+			read-only;
+		};
+
+		/* The final 512 KiB is also unmapped by the vendor firmware. */
+	};
+};
+
+&pcie {
+	status = "okay";
+};
+
+&pcie0 {
+	wifi@0,0 {
+		compatible = "mediatek,mt76";
+		reg = <0x0000 0 0 0 0>;
+		nvmem-cells = <&eeprom_factory_0>, <&macaddr_factory_4>;
+		nvmem-cell-names = "eeprom", "mac-address";
+	};
+};
+
+&gmac0 {
+	nvmem-cells = <&macaddr_factory_e006 1>;
+	nvmem-cell-names = "mac-address";
+};
+
+&gmac1 {
+	status = "okay";
+	label = "wan";
+	phy-handle = <&ethphy4>;
+
+	nvmem-cells = <&macaddr_factory_e006 0>;
+	nvmem-cell-names = "mac-address";
+};
+
+&ethphy4 {
+	/delete-property/ interrupts;
+};
+
+&switch0 {
+	ports {
+		port@0 {
+			status = "okay";
+			label = "lan4";
+		};
+
+		port@1 {
+			status = "okay";
+			label = "lan3";
+		};
+
+		port@2 {
+			status = "okay";
+			label = "lan2";
+		};
+
+		port@3 {
+			status = "okay";
+			label = "lan1";
+		};
+	};
+};
+
+&state_default {
+	gpio {
+		groups = "i2c", "uart2", "uart3", "jtag", "wdt";
+		function = "gpio";
+	};
+};
--- a/target/linux/ramips/image/mt7621.mk
+++ b/target/linux/ramips/image/mt7621.mk
@@ -1095,6 +1095,18 @@ define Device/dlink_dir-853-r1
 endef
 TARGET_DEVICES += dlink_dir-853-r1
 
+define Device/dlink_dir-853-r3
+  $(Device/nand)
+  IMAGE_SIZE := 75904k
+  KERNEL_SIZE := 3456k
+  DEVICE_VENDOR := D-Link
+  DEVICE_MODEL := DIR-853
+  DEVICE_VARIANT := R3
+  DEVICE_PACKAGES := kmod-mt7615-firmware kmod-usb3 \
+	kmod-usb-ledtrig-usbport -uboot-envtools
+endef
+TARGET_DEVICES += dlink_dir-853-r3
+
 define Device/dlink_dir-860l-b1
   $(Device/dsa-migration)
   $(Device/seama-lzma-loader)
--- a/target/linux/ramips/mt7621/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/mt7621/base-files/etc/board.d/01_leds
@@ -124,6 +124,10 @@ dlink,dir-853-a3)
 dlink,dir-853-r1)
 	ucidef_set_led_netdev "internet" "internet" "blue:net" "wan"
 	;;
+dlink,dir-853-r3)
+	ucidef_set_led_wlan "wlan2g" "WiFi 2.4GHz" "mt76-phy0" "phy0tpt"
+	ucidef_set_led_wlan "wlan5g" "WiFi 5GHz" "mt76-phy1" "phy1tpt"
+	;;
 dlink,dir-1935-a1|\
 dlink,dir-860l-b1|\
 dlink,dir-867-a1|\
--- a/target/linux/ramips/mt7621/base-files/etc/hotplug.d/ieee80211/10_fix_wifi_mac
+++ b/target/linux/ramips/mt7621/base-files/etc/hotplug.d/ieee80211/10_fix_wifi_mac
@@ -66,6 +66,13 @@ case "$board" in
 			macaddr_setbit_la "$base_mac" > /sys${DEVPATH}/macaddress
 		fi
 		;;
+	dlink,dir-853-r3)
+		if [ "$PHYNBR" = "0" ]; then
+			base_mac="$(mtd_get_mac_binary factory 0x4)"
+			macaddr_setbit_la "$(macaddr_setbit "$base_mac" 27)" \
+				> /sys${DEVPATH}/macaddress
+		fi
+		;;
 	edup,ep-rt2960s|\
 	haier,har-20s2u1|\
 	jcg,y2|\
--- a/target/linux/ramips/mt7621/base-files/lib/upgrade/platform.sh
+++ b/target/linux/ramips/mt7621/base-files/lib/upgrade/platform.sh
@@ -95,6 +95,7 @@ platform_do_upgrade() {
 	dlink,dir-3040-a1|\
 	dlink,dir-3060-a1|\
 	dlink,dir-853-a3|\
+	dlink,dir-853-r3|\
 	dlink,dir-x1860-b1|\
 	edup,ep-rt2960s|\
 	edup,ep-rt2983|\


target/linux/ramips/mt7621/base-files/etc/board.d/02_network
```diff
@@ -14,6 +14,9 @@ ramips_setup_interfaces()
 	buffalo,wsr-2533dhpls|\
 	confiabits,mt7621-v1|\
 	dlink,dir-x1860-b1|\
+	dlink,dir-853-r3)
+		ucidef_set_interfaces_lan_wan "lan1 lan2 lan3 lan4" "wan"
+		;;
```
openwrt-ai: This terminates a case list that used to run all the way down to `zyxel,wsm20`. Everything above the new entry — `alfa-network,ax1800rm`, `ampedwireless,ally-r1900k`, `asus,rt-ax53u`, `buffalo,wsr-2533dhpl2`, `buffalo,wsr-2533dhpls`, `confiabits,mt7621-v1`, `dlink,dir-x1860-b1` — now hits the new 4-LAN branch instead of [`ucidef_set_interfaces_lan_wan "lan1 lan2 lan3" "wan"`](https://github.com/openwrt/openwrt/blob/8f76fe33b2e2d1e3511e74a87ae3146ba28cae3a/target/linux/ramips/mt7621/base-files/etc/board.d/02_network#L43), silently regressing seven unrelated boards.

The hunk is also unnecessary: the [`*)` default](``https://github.com/openwrt/openwrt/blob/8f76fe33b2e2d1e3511e74a87ae3146ba28cae3a/target/linux/ramips/mt7621/base-files/etc/board.d/02_network#L203-L204``) already sets `ucidef_set_interfaces_lan_wan "lan1 lan2 lan3 lan4" "wan"`, which is exactly what this board needs. Dropping the hunk gives the correct result and leaves the other boards alone.

```suggestion
	dlink,dir-x1860-b1|\
```

---

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
```diff
@@ -0,0 +1,674 @@
...
+
+/ {
+    #address-cells = <0x1>;
+    #size-cells = <0x1>;
+    compatible = "dlink,dir-853-r3", "mediatek,mt7621-soc";
+    model = "D-Link DIR-853 R3";
+    aliases {
+        serial0 = "/palmbus@1e000000/uartlite@c00";
+        label-mac-device = "/ethernet@1e100000/mac@0";
+    };
```
openwrt-ai: This file is a `dtc -I dtb -O dts` dump of a built device tree, not a hand-written board DTS: the `// version:` / `// last_comp_version:` / `// boot_cpuid_phys:` lines are dtc output metadata, all phandles are raw integers (`<0x2>`, `<0x1A>`), every value is hex, indentation is 4 spaces instead of tabs, aliases are string paths, and the whole MT7621 SoC tree is inlined instead of coming from [`mt7621.dtsi`](https://github.com/openwrt/openwrt/blob/8f76fe33b2e2d1e3511e74a87ae3146ba28cae3a/target/linux/ramips/dts/mt7621.dtsi). The SPDX header every other file in this directory carries is missing too.

The consequence is not just style: the inlined SoC tree is a snapshot of an older tree and has drifted from the current `mt7621.dtsi` in several places (see the separate comments on the xhci, crypto and GIC nodes), so this board silently loses fixes that the shared dtsi already carries and will not pick up future ones.

Please rewrite it in the same shape as the sibling variants, e.g. [`mt7621_dlink_dir-853-r1.dts`](https://github.com/openwrt/openwrt/blob/8f76fe33b2e2d1e3511e74a87ae3146ba28cae3a/target/linux/ramips/dts/mt7621_dlink_dir-853-r1.dts): SPDX line, `#include "mt7621.dtsi"`, the `dt-bindings` includes for `gpio.h` / `input.h` / `leds/common.h`, then only the board-specific `/ { ... }` additions plus `&nand`, `&pcie`, `&pcie0`, `&gmac0`, `&gmac1`, `&switch0`, `&state_default` overrides. The remaining comments below are on issues that would survive that rewrite.

---

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
```diff
@@ -0,0 +1,674 @@
...
+            reg = <0x1>;
+            #trigger-source-cells = <0x0>;
+            phandle = <0x1F>;
+        };
+        port@2 {
+            reg = <0x2>;
+            #trigger-source-cells = <0x0>;
+            phandle = <0x20>;
+        };
+    };
```
openwrt-ai: USB will not work as configured. Three problems, all from the stale inlined SoC tree:

- `compatible = "mediatek,mt8173-xhci"` alone — the in-tree node uses [`"mediatek,mtk-xhci"`](https://github.com/openwrt/openwrt/blob/8f76fe33b2e2d1e3511e74a87ae3146ba28cae3a/target/linux/ramips/dts/mt7621.dtsi#L396). No other DTS in this tree uses the mt8173 string on its own; the two econet dtsi files that use it pair it with `"mediatek,mtk-xhci"` as fallback.
- No `phys` property and no `t-phy@1e1d0000` node anywhere in this file, so `u2port0` / `u3port0` / `u2port1` never get bound — see [mt7621.dtsi:408-410](https://github.com/openwrt/openwrt/blob/8f76fe33b2e2d1e3511e74a87ae3146ba28cae3a/target/linux/ramips/dts/mt7621.dtsi#L408-L410) and [the `t-phy` node at L435-L456](https://github.com/openwrt/openwrt/blob/8f76fe33b2e2d1e3511e74a87ae3146ba28cae3a/target/linux/ramips/dts/mt7621.dtsi#L435-L456).
- No `vbus-supply` / `vusb33-supply`, which the in-tree node has at [L412-L413](https://github.com/openwrt/openwrt/blob/8f76fe33b2e2d1e3511e74a87ae3146ba28cae3a/target/linux/ramips/dts/mt7621.dtsi#L412-L413).

The Makefile entry ships `kmod-usb3` and `kmod-usb-ledtrig-usbport`, so this is a functional gap, not a cosmetic one. Including `mt7621.dtsi` instead of inlining the node fixes all three at once.

---

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
```diff
@@ -0,0 +1,674 @@
...
+            };
+            partition@5a20000 {
+                label = "stock-tail";
+                reg = <0x5A20000 0x2560000>;
+                read-only;
+            };
+        };
+    };
+    crypto@1e004000 {
+        compatible = "mediatek,mtk-eip93";
```
openwrt-ai: `mediatek,mtk-eip93` is not a compatible string used anywhere in this tree — the EIP93 crypto engine is matched on [`"inside-secure,safexcel-eip93ies"`](https://github.com/openwrt/openwrt/blob/8f76fe33b2e2d1e3511e74a87ae3146ba28cae3a/target/linux/ramips/dts/mt7621.dtsi#L496) (same string in `an7581.dtsi` and `an7583.dtsi`). As written the driver will never bind on this board.

```suggestion
        compatible = "inside-secure,safexcel-eip93ies";
```

---

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
```diff
@@ -0,0 +1,674 @@
...
+        };
+        port@2 {
+            reg = <0x2>;
+            #trigger-source-cells = <0x0>;
+            phandle = <0x20>;
+        };
+    };
+    interrupt-controller@1fbc0000 {
+        compatible = "mti,gic";
+        reg = <0x1FBC0000 0x2000>;
```
openwrt-ai: The GIC register window is 16x too small: `mt7621.dtsi` maps [`<0x1fbc0000 0x20000>`](https://github.com/openwrt/openwrt/blob/8f76fe33b2e2d1e3511e74a87ae3146ba28cae3a/target/linux/ramips/dts/mt7621.dtsi#L458-L459), this says `0x2000`.

```suggestion
        reg = <0x1FBC0000 0x20000>;
```

Same class of drift on line 293: `mmc@1e130000` uses `<0x1E130000 0x4000>` where [mt7621.dtsi:367](https://github.com/openwrt/openwrt/blob/8f76fe33b2e2d1e3511e74a87ae3146ba28cae3a/target/linux/ramips/dts/mt7621.dtsi#L367) has `0x8000` (less harmful, that node is disabled).

---

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
```diff
@@ -0,0 +1,674 @@
...
+            gpios = <0x1A 0xD 0x1>;
+            trigger-sources = <0x1F 0x20>;
+            linux,default-trigger = "usbport";
+        };
+        wps_green {
+            function = "wps";
+            color = <0x2>;
+            gpios = <0x1A 0xF 0x1>;
+        };
+    };
```
openwrt-ai: No status/power LED is defined and the `aliases` node has no `led-boot` / `led-failsafe` / `led-running` / `led-upgrade`, so this board will give no boot, failsafe or sysupgrade indication at all. All three sibling variants define them — see [`mt7621_dlink_dir-853-r1.dts:17-21`](https://github.com/openwrt/openwrt/blob/8f76fe33b2e2d1e3511e74a87ae3146ba28cae3a/target/linux/ramips/dts/mt7621_dlink_dir-853-r1.dts#L17-L21). The R3 shares the DIR-853 chassis, so it almost certainly has the same bi-colour power LED; if it genuinely does not, please say so in the commit message.

Also `color = <0x2>` is a raw magic number here and on line 670 — use `LED_COLOR_ID_GREEN` from `<dt-bindings/leds/common.h>`, and `LED_FUNCTION_USB` / `LED_FUNCTION_WPS` for the `function` strings, once the file includes the bindings header.

---

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
```diff
@@ -0,0 +1,674 @@
...
+            };
+            partition@1000000 {
+                label = "ubi";
+                reg = <0x1000000 0x4A20000>;
+            };
+            partition@5a20000 {
+                label = "stock-tail";
+                reg = <0x5A20000 0x2560000>;
+                read-only;
+            };
```
openwrt-ai: `stock-rootfs` (0xB60000, 11.4 MiB) plus `stock-tail` (0x2560000, 37.4 MiB) fence off ~49 MiB of the 128 MiB NAND as read-only, leaving `ubi` only 74 MiB. Partitions also stop at 0x7F80000, so the last 512 KiB is unmapped entirely. Is keeping the OEM rootfs and the trailing region deliberate (a stock-recovery path?), or is this leftover from the vendor layout that could be folded into `ubi`? Nothing in the commit message explains it, and the sibling D-Link NAND devices using [`Device/dlink_dir_nand_128m`](https://github.com/openwrt/openwrt/blob/8f76fe33b2e2d1e3511e74a87ae3146ba28cae3a/target/linux/ramips/image/mt7621.mk#L970-L980) don't carry equivalent partitions.

---

target/linux/ramips/image/mt7621.mk
```diff
@@ -1186,6 +1186,18 @@ define Device/dlink_dra-1360-a1
...
+  $(Device/nand)
+  IMAGE_SIZE := 75904k
+  KERNEL_SIZE := 3456k
+  DEVICE_VENDOR := D-Link
+  DEVICE_MODEL := DIR-853
+  DEVICE_VARIANT := R3
+  DEVICE_PACKAGES := kmod-mt7615-firmware kmod-usb3 \
+	kmod-usb-ledtrig-usbport -uboot-envtools
+endef
+TARGET_DEVICES += dlink_dir-853-r3
```
openwrt-ai: nit: this block is inserted after `Device/dlink_dra-1360-a1`, but the file is ordered alphabetically and the other three variants live together at [lines 1073-1096](https://github.com/openwrt/openwrt/blob/8f76fe33b2e2d1e3511e74a87ae3146ba28cae3a/target/linux/ramips/image/mt7621.mk#L1073-L1096). It belongs directly after `TARGET_DEVICES += dlink_dir-853-r1`.

Separately: this is the only D-Link MT7621 NAND device that builds a plain `Device/nand` with no factory/recovery image, while the siblings go through [`Device/dlink_dir_nand_128m`](https://github.com/openwrt/openwrt/blob/8f76fe33b2e2d1e3511e74a87ae3146ba28cae3a/target/linux/ramips/image/mt7621.mk#L970-L980) which adds `uimage-sgehdr` and `recovery.bin`. Without a commit message describing the install path (see the review body) it isn't possible to tell whether the R3's bootloader accepts a bare sysupgrade image or whether the recovery image is simply missing — please state which.

---

openwrt-ai (commented):
## Commit checks

- `8f76fe3` "ramips: add support for D-Link DIR-853 R3" — this adds a new `.dts` under `target/linux/ramips/dts/` plus a new `define Device/dlink_dir-853-r3` in `target/linux/ramips/image/mt7621.mk`, so it is a new-device commit and the message is missing two of the three required sections:
  - **Flash instructions** — absent entirely. There is no described install path (recovery web UI / TFTP / U-Boot / serial), no verbatim button, IP or hostname strings, and no revert-to-stock note. This matters more than usual here because the Makefile entry builds no factory or recovery image (see the inline comment on `mt7621.mk`), so a reader has no way to guess how the first flash is meant to happen.
  - **MAC address layout** — absent entirely. The DTS reads MACs from three different nvmem cells (`macaddr@4`, `macaddr@e000`, `macaddr@e006` with `mac-base` offsets 0 and 1), and the message should spell out one line per interface (LAN / WAN / 2.4G / 5G) saying which cell and offset each comes from, with `xx:xx:xx` for the unit-specific bytes.
  - The **hardware specification** section is present but thin: no RAM or flash chip detail beyond the two NAND part numbers, no LED/button count, no UART header location/pinout/baud, no power rating.

  Also, the spec block is indented as a code block rather than a plain list, which renders as literal preformatted text in `git log`; the reference commits `986ca4c887f4088b6fbc703faa88884350e9274f` and `a2dcbd79a4460617bd42151555448f8bca8ca7be` show the usual form.

Beyond the commit message, the main blocker is inline: the `02_network` hunk changes the LAN port set for seven unrelated boards, and the DTS is a decompiled `dtb` rather than a board DTS built on `mt7621.dtsi`, which has left it with several stale/incorrect SoC nodes (xhci, crypto, GIC). Details in the inline comments.

---

target/linux/ramips/mt7621/base-files/etc/board.d/02_network
grumworth: Fixed by dropping the `02_network` hunk entirely. The R3 now uses the existing mt7621 default 4-LAN/1-WAN mapping, so the seven unrelated boards are no longer changed.

Updated in `e65bc1b`.

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
grumworth: Fixed. I replaced the decompiled 674-line DTB dump with a compact, hand-written board DTS that includes `mt7621.dtsi` plus the GPIO input/LED bindings and contains only board-specific overrides.

Updated in `e65bc1b`.

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
grumworth: Fixed by removing the stale USB controller/T-PHY/regulator copies. The board now inherits the current shared mt7621 USB definitions; its DTS only enables the verified USB LED trigger sources.

Updated in `e65bc1b`.

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
grumworth: Fixed. The board DTS no longer overrides the crypto node, so it inherits the current SoC-compatible definition from `mt7621.dtsi`.

Updated in `e65bc1b`.

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
grumworth: Fixed. The copied GIC and MMC nodes were removed along with the rest of the decompiled SoC description; both now come from the shared mt7621 DTSI.

Updated in `e65bc1b`.

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
grumworth: Addressed using tested hardware behavior. The R3 power LED is not software controllable, while the green WPS LED on GPIO15 is verified, so it is now used for boot/failsafe/running/upgrade aliases. Binding constants replace numeric GPIO/key values.

Updated in `e65bc1b`.

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
grumworth: The layout is deliberate and is now documented in the DTS and commit message. The stock rootfs is retained read-only for recovery; `stock-tail` contains vendor JFFS2 data at 0x5a20000 and 0x7f40000 and must be preserved; the final 512 KiB remains unmapped to match the stock MTD layout.

Updated in `e65bc1b`.

target/linux/ramips/image/mt7621.mk
grumworth: Fixed. The device entry is now next to DIR-853 R1. The R3 bootloader accepts a standard uImage through its UART/TFTP command line, so no SGE factory/recovery wrapper is used. The commit message now gives the exact tested initramfs -> sysupgrade procedure and emergency-web recovery steps.

Updated in `e65bc1b`.

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
```diff
@@ -0,0 +1,197 @@
...
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		usb {
+			function = LED_FUNCTION_USB;
+			color = <LED_COLOR_ID_GREEN>;
+			gpios = <&gpio 13 GPIO_ACTIVE_LOW>;
+			trigger-sources = <&xhci_ehci_port1>, <&ehci_port2>;
```
openwrt-ai: `ehci_port2` is a *second* physical USB slot, not the USB 2.0 half of the first one — see the comment above [`ehci_port2` in mt7621.dtsi:426-432](https://github.com/openwrt/openwrt/blob/ba8815b99a1e665540fff2f3229dad62c0457932/target/linux/ramips/dts/mt7621.dtsi#L426-L432) ("Only the second usb hub has a second port"). `xhci_ehci_port1` already covers both the EHCI/OHCI and XHCI sides of the single slot. The commit message lists "USB: 1x USB 3.0", and both siblings list only `xhci_ehci_port1` ([mt7621_dlink_dir-853-a3.dts:73](https://github.com/openwrt/openwrt/blob/ba8815b99a1e665540fff2f3229dad62c0457932/target/linux/ramips/dts/mt7621_dlink_dir-853-a3.dts#L73), [mt7621_dlink_dir-853-r1.dts:53](https://github.com/openwrt/openwrt/blob/ba8815b99a1e665540fff2f3229dad62c0457932/target/linux/ramips/dts/mt7621_dlink_dir-853-r1.dts#L53)), so the extra source refers to a port this board does not have.

```suggestion
			trigger-sources = <&xhci_ehci_port1>;
```

---

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
```diff
@@ -0,0 +1,197 @@
...
+&nand {
+	status = "okay";
+
+	partitions {
+		compatible = "fixed-partitions";
+		#address-cells = <1>;
+		#size-cells = <1>;
+
+		partition@0 {
+			label = "stock-bootloader";
```
openwrt-ai: nit: these two partitions have exactly the same offsets and sizes as the DIR-853 A3 (`0x0 0x80000` and `0x80000 0x80000`), where they are named `u-boot` and `config` — see [mt7621_dlink_dir-853-a3.dts:99-109](https://github.com/openwrt/openwrt/blob/ba8815b99a1e665540fff2f3229dad62c0457932/target/linux/ramips/dts/mt7621_dlink_dir-853-a3.dts#L99-L109). `u-boot` is the tree-wide name for the bootloader partition (437 ramips DTS files use it); `stock-bootloader` / `stock-config` are used nowhere in `target/linux/ramips/`. The `stock-` prefix is worth keeping only on `stock-rootfs` / `stock-tail`, where it genuinely signals "vendor data preserved for recovery".

```suggestion
			label = "u-boot";
```

Same for line 78: `label = "config";`.

---

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
```diff
@@ -0,0 +1,197 @@
...
+	wifi@0,0 {
+		compatible = "mediatek,mt76";
+		reg = <0x0000 0 0 0 0>;
+		nvmem-cells = <&eeprom_factory_0>, <&macaddr_factory_4>;
+		nvmem-cell-names = "eeprom", "mac-address";
+	};
+};
+
+&gmac0 {
+	nvmem-cells = <&macaddr_factory_e006 1>;
```
openwrt-ai, line 151: Deriving the LAN MAC as `0xe006 + 1` rather than reading it from `0xe000` is the one place this diverges from the A3, which has the identical `factory` partition geometry (`0x100000 0x40000`, eeprom at `0x0`, MACs at `0x4` / `0xe000` / `0xe006`) and reads LAN straight from `macaddr@e000` — [mt7621_dlink_dir-853-a3.dts:129-131 and 192-195](https://github.com/openwrt/openwrt/blob/ba8815b99a1e665540fff2f3229dad62c0457932/target/linux/ramips/dts/mt7621_dlink_dir-853-a3.dts#L129-L131). This DTS doesn't declare a `macaddr@e000` cell at all.

Did you check what's stored at `0xe000` on the R3? If it holds the sticker/LAN MAC, reading it directly is safer than `+1` arithmetic — `WAN + 1` only equals the vendor LAN MAC if the two happen to be consecutive in that order, and `label-mac-device = &gmac0` means a mismatch shows up as a wrong label MAC. If `0xe000` is blank or holds something else on this revision, a one-line note in the commit message's MAC section would settle it.

---

openwrt-ai (commented):
Re-reviewed after the force-push to `ba8815b`. The rewrite addresses everything from the previous round: the `02_network` hunk is gone (the mt7621 `*)` default already gives lan1-4/wan), the DTS is now a hand-written board file including `mt7621.dtsi` — so the stale xhci/crypto/GIC/mmc nodes are gone with it — LED bindings constants are used, boot/failsafe/running/upgrade aliases exist on `led_wps`, the Makefile entry sits next to `dlink_dir-853-r1`, and the commit message now carries all three required sections (hardware spec, flash instructions incl. revert-to-stock, MAC layout). No commit-message issues remain.

Three smaller things left, all inline: one likely-wrong USB trigger source, a partition-naming deviation from the sibling A3, and a question about the LAN MAC offset.

---

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
grumworth: The R3 differs from the A3/R1 here. Vendor specifies one USB 2.0 port, and hardware testing enumerates a connected drive as `usb 1-2` at high speed. Changed the trigger source to `ehci_port2` only and corrected the hardware specification from USB 3.0 to USB 2.0.

Updated in `7dc6aab`.

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
grumworth: Agreed. Renamed the partitions to the conventional `u-boot` and `config` labels.

Updated in `7dc6aab`.

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
grumworth: Checked on the R3 factory backup: offset `0xe000` contains `ff:ff:ff:ff:ff:ff`, while `0xe006` contains the WAN MAC. Stock uses the consecutive address (WAN + 1) for LAN. I kept the current derivation and documented this in the commit message.

Updated in `7dc6aab`.

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
```diff
@@ -0,0 +1,197 @@
...
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		usb {
+			function = LED_FUNCTION_USB;
+			color = <LED_COLOR_ID_GREEN>;
+			gpios = <&gpio 13 GPIO_ACTIVE_LOW>;
+			trigger-sources = <&ehci_port2>;
```
openwrt-ai, line 51: Thanks for checking on hardware — `trigger-sources = <&ehci_port2>;` alone is a well-established in-tree form for USB2-only mt7621 boards, including the D-Link siblings [`mt7621_dlink_dir-3040-a1.dts:20`](https://github.com/openwrt/openwrt/blob/7dc6aab847dd5305c272c37280946ed11ba04d7d/target/linux/ramips/dts/mt7621_dlink_dir-3040-a1.dts#L20) and [`mt7621_dlink_dir-882-r1.dts:15`](https://github.com/openwrt/openwrt/blob/7dc6aab847dd5305c272c37280946ed11ba04d7d/target/linux/ramips/dts/mt7621_dlink_dir-882-r1.dts#L15). Resolved from my side.

nit: the PR description still lists `- Ports: 4x LAN, 1x WAN, 1x USB 3.0` while the commit message now says `- USB: 1x USB 2.0` and adds the `ehci_port2` rationale. Worth updating the PR body so the two agree.

---

openwrt-ai (commented):
Re-reviewed the force-push from `ba8815b` to `7dc6aab`. The delta is three DTS lines plus a commit-message update, and both of the previous round's actionable items are addressed:

- `trigger-sources` narrowed to `<&ehci_port2>` with a hardware-verified rationale in the commit message; this matches how other USB2-only mt7621 D-Link boards are described in-tree.
- `stock-bootloader` / `stock-config` renamed to the conventional `u-boot` / `config`.
- The LAN-MAC question is answered in the commit message (`0xe000` is all-`0xff` on R3, so LAN is derived as `0xe006 + 1`).

Commit message still matches the diff and carries all three required sections; the Makefile sizes line up with the DTS (`KERNEL_SIZE 3456k` = `0x360000`, `IMAGE_SIZE 75904k` = `0x4a20000`), and the `01_leds` / `10_fix_wifi_mac` / `platform.sh` entries are placed consistently with their neighbours. FormalityCheck is green on `7dc6aab`.

One documentation nit left inline (PR body vs commit message on USB version).

---

target/linux/ramips/dts/mt7621_dlink_dir-853-r3.dts
grumworth: Fixed. Updated the PR description from USB 3.0 to USB 2.0 so it matches the hardware specification and commit message.

grumworth:
@untilscour, could you please review this? This is all about hardware-tested D-Link DIR-853 R3 support for ramips/mt7621. All formality, PR-profile, and ramips/mt7621 kernel builds pass, and the automated review findings have been addressed.

openwrt-ai (commented):

The move from `7dc6aab` to `70d3544` is a rebase onto current `main` only — the commit's patch content is byte-identical (`git diff` between the two commits' own diffs is empty) and the commit message is unchanged. The only `target/linux/ramips/` change that landed on `main` in between is `224a967ba1` ("ramips: mt7530: compile swconfig code conditionally"), which doesn't touch anything this PR modifies.

The last open item from the previous round — the PR body listing `1x USB 3.0` against the commit message's `1x USB 2.0` — is now fixed in the description. No CI checks are reporting on `70d3544` yet, and nothing is failing.

---

untilscour:
> @untilscour, could you please review this? This is all about hardware-tested D-Link DIR-853 R3 support for ramips/mt7621. All formality, PR-profile, and ramips/mt7621 kernel builds pass, and the automated review findings have been addressed.

looks good, just need to wait for maintainer feedback/merge

PR #24585, opened 2026-08-06
mediatek: filogic: add NMBM properties for netis NX30 V2

DH-41:
Add MediaTek NMBM (NAND Flash Bad Block Management) properties to ensure proper flash storage handling, matching the vendor firmware configuration.

Kernel Log
> [    0.997117] spi-nand spi0.0: ESMT SPI NAND was found.
[    1.002208] spi-nand spi0.0: 128 MiB, block size: 128 KiB, page size: 2048, OOB size: 64
[    1.011169] Signature found at block 1023 [0x07fe0000]
[    1.016303] NMBM management region starts at block 960 [0x07800000]
[    1.023658] First info table with writecount 0 found in block 960
[    1.032623] Second info table with writecount 0 found in block 963
[    1.038797] NMBM has been successfully attached
[    1.044282] 5 fixed-partitions partitions found on MTD device spi0.0

This change has been tested and verified on NX30 V2 in this PR. 
Please note that Netcore N30 Pro / netis NX35U (proposed by @bfdeh in #24316) is also expected to use this shared `.dtsi`, and runtime testing for that device will be required.

### DIFF
--- a/target/linux/mediatek/dts/mt7981b-netis-common.dtsi
+++ b/target/linux/mediatek/dts/mt7981b-netis-common.dtsi
@@ -110,6 +110,10 @@
 		spi-tx-bus-width = <4>;
 		spi-rx-bus-width = <4>;
 
+		mediatek,nmbm;
+		mediatek,bmt-max-ratio = <1>;
+		mediatek,bmt-max-reserved-blocks = <64>;
+
 		partitions {
 			compatible = "fixed-partitions";
 			#address-cells = <1>;


csharper2005:
This device use OpenWrt U-Boot with disabled NMBM. We must enable nmbm either everywhere or nowhere.

target/linux/mediatek/dts/mt7981b-netis-common.dtsi
```diff
@@ -110,6 +110,10 @@
 		spi-tx-bus-width = <4>;
 		spi-rx-bus-width = <4>;
 
+		mediatek,nmbm;
+		mediatek,bmt-max-ratio = <1>;
+		mediatek,bmt-max-reserved-blocks = <64>;
```
openwrt-ai, line 115: Enabling NMBM only on the kernel side looks asymmetric for this device, which backs up @csharper2005's point. Concrete evidence:

- This board installs OpenWrt's own bootloader — [`ARTIFACT/spim-nand-preloader.bin := mt7981-bl2 spim-nand-ddr3` / `ARTIFACT/spim-nand-bl31-uboot.fip := mt7981-bl31-uboot netis_nx30v2` at filogic.mk:2853-2855](https://github.com/openwrt/openwrt/blob/9ac433ccea34eef149dba6d5b11a00936e5d459c/target/linux/mediatek/image/filogic.mk#L2853-L2855).
- `package/boot/uboot-mediatek` carries the NMBM code (`100-0*`, `129-board-mediatek-wire-up-NMBM-support.patch`) but no defconfig in the tree sets `CONFIG_ENABLE_NAND_NMBM` / `CONFIG_NMBM_MTD`, and [`461-add-netis-nx30v2.patch`](https://github.com/openwrt/openwrt/blob/9ac433ccea34eef149dba6d5b11a00936e5d459c/package/boot/uboot-mediatek/patches/461-add-netis-nx30v2.patch) doesn't either — so the U-Boot and BL2 flashed on this device address the NAND raw.
- Because `mediatek,bmt-remap-range` is not set, the remap covers the *whole* device, not just the UBI area — see [`mtk_bmt.c:77-81`](https://github.com/openwrt/openwrt/blob/9ac433ccea34eef149dba6d5b11a00936e5d459c/target/linux/generic/files/drivers/mtd/nand/mtk_bmt.c#L77-L81) (`*start = 0; *end = bmtd.total_blks;` when the property is absent). BL2, `u-boot-env`, `Factory`, `FIP` and `ubi` all sit inside the remapped range, so as soon as one block there is retired the kernel's view and U-Boot's view diverge and the device can stop booting.

Cross-checking the rest of the target: of the ~80 filogic devices whose DTS sets `mediatek,nmbm`, only `bazis_ax3000wm` also ships OpenWrt U-Boot artifacts; every other one keeps the vendor bootloader (which does implement NMBM). So the in-tree pattern is "NMBM in DTS ⇔ vendor bootloader", and this PR would be the second exception.

Is enabling NMBM in the kernel alone intentional here, or should `CONFIG_ENABLE_NAND_NMBM`/`CONFIG_NMBM_MTD` be turned on in the `mt7981_netis_nx30v2` U-Boot config at the same time?

---

target/linux/mediatek/dts/mt7981b-netis-common.dtsi
```diff
@@ -110,6 +110,10 @@
 		spi-tx-bus-width = <4>;
 		spi-rx-bus-width = <4>;
 
+		mediatek,nmbm;
```
openwrt-ai, line 113: nit: the commit message (and PR body) justify this with "matching the vendor firmware configuration", but the supported OpenWrt install path for this board replaces the vendor bootloader with OpenWrt BL2 + U-Boot ([filogic.mk:2853-2855](https://github.com/openwrt/openwrt/blob/9ac433ccea34eef149dba6d5b11a00936e5d459c/target/linux/mediatek/image/filogic.mk#L2853-L2855)), neither of which has NMBM enabled. Whatever the vendor firmware did is not what the running system looks like, so that sentence doesn't really justify the change — it would help to state the actual reason (e.g. the flash layout already stops at 0x7800000, leaving the top 64 blocks free for the NMBM management region) instead.

---

openwrt-ai (commented):
Reviewed 1 new commit. The single commit's message matches its diff; comments are inline.

---

PR #20341, opened 2025-10-08
wifi-scripts: Add 60G band to mac80211.uc to support 802.11ad

Konstantin-Glukhov:
/etc/config/wireless is missing 60G band for routers supporting it. This support was enabled in OpenWrt v23 but somehow disappeared in v24. The suggested change was tested on TP-Link Talon AD7200 and it seems to work fine.

PR #24586, opened 2026-08-06
wifi-scripts: Add 60G band to mac80211.uc to support 802.11ad

Leo-PL:
This PR supersedes #20341.

/etc/config/wireless is missing 60G band for routers supporting it. This support was enabled in OpenWrt v23 but somehow disappeared in v24.

Run-tested-on: TP-Link Talon AD7200 v2

[Originally authored by Konstantin Glukhov, replaced SoB, because of lack of real email]
Backports to 25.12 and 24.10 will be needed as well, but this should cherry-pick cleanly.

### DIFF
--- a/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/hostapd.uc
+++ b/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/hostapd.uc
@@ -32,8 +32,10 @@ function set_device_defaults(config) {
 	/* validate band */
 	if (config.band == '2g')
 		config.hw_mode = 'g';
-	else if (config.band in [ '5g', '6g', '60g' ])
+	else if (config.band in [ '5g', '6g' ])
 		config.hw_mode = 'a';
+	else if (config.band == '60g')
+		config.hw_mode = 'ad';
 	else
 		switch (config.hw_mode) {
 		case 'a':
--- a/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/iface.uc
+++ b/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/iface.uc
@@ -118,7 +118,7 @@ export function parse_encryption(config, dev_config, phy_features) {
 
 	if (!config.wpa)
 		config.wpa_pairwise ??= null;
-	else if (config.hw_mode == 'ad')
+	else if (dev_config.band == '60g')
 		config.wpa_pairwise ??= 'GCMP';
 	else if (config.gcmp256 && phy_features?.cipher_gcmp256)
 		config.wpa_pairwise ??= 'GCMP-256 CCMP';
--- a/package/network/config/wifi-scripts/files/lib/wifi/mac80211.uc
+++ b/package/network/config/wifi-scripts/files/lib/wifi/mac80211.uc
@@ -2,7 +2,7 @@
 import { readfile } from "fs";
 import * as uci from 'uci';
 
-const bands_order = [ "6G", "5G", "2G" ];
+const bands_order = [ "60G", "6G", "5G", "2G" ];
 const htmode_order = [ "EHT", "HE", "VHT", "HT" ];
 
 let board = json(readfile("/etc/board.json"));


Leo-PL:
cc @hauke @robimarko

package/network/config/wifi-scripts/files/lib/wifi/mac80211.uc
```diff
@@ -2,7 +2,7 @@
 import { readfile } from "fs";
 import * as uci from 'uci';
 
-const bands_order = [ "6G", "5G", "2G" ];
+const bands_order = [ "60G", "6G", "5G", "2G" ];
```
openwrt-ai, line 5: With this change a 60G phy now gets an auto-generated section with `band='60g'` and no `hwmode`. On the ucode hostapd path (`WIFI_SCRIPTS_UCODE`, `default y`) that combination ends up as `hw_mode=a`, not `ad`:

[`set_device_defaults()` in hostapd.uc:33-36](``https://github.com/openwrt/openwrt/blob/77f5a598cdfae575c9b1ba545f6f26eae624142f/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/hostapd.uc#L33-L36)`` puts `'60g'` in the same bucket as `5g`/`6g` and assigns `hw_mode = 'a'`; the `'ad'` value is only reachable when the user explicitly writes `hwmode='11ad'`. Downstream code notices — [`iface.uc:121`](``https://github.com/openwrt/openwrt/blob/77f5a598cdfae575c9b1ba545f6f26eae624142f/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/iface.uc#L121)`` selects `wpa_pairwise = 'GCMP'` on `hw_mode == 'ad'`, which would never trigger for an auto-generated 60G radio. The shell path does the right thing here: [`netifd-wireless.sh:73`](``https://github.com/openwrt/openwrt/blob/77f5a598cdfae575c9b1ba545f6f26eae624142f/package/network/config/wifi-scripts/files/lib/netifd/netifd-wireless.sh#L73``) maps `60g) hwmode=ad`, and hostapd only accepts `ad` for 60 GHz (`config_file.c`, `hw_mode` parser).

Since the commit says run-tested on the Talon AD7200 v2, was that with `WIFI_SCRIPTS_UCODE` enabled? If so I may be missing where `hw_mode` gets corrected; if not, this looks like it needs a matching `'60g' -> hw_mode = 'ad'` fix in `hostapd.uc` for the generated config to actually be usable.

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the diff (band list extension + `PKG_RELEASE` bump, both described), and the `60G` name matches what [`wifi-detect.uc`](``https://github.com/openwrt/openwrt/blob/77f5a598cdfae575c9b1ba545f6f26eae624142f/package/network/config/wifi-scripts/files/usr/share/hostap/wifi-detect.uc#L136)`` writes into `board.json`, so the filter now resolves for 60 GHz phys. One question inline about the ucode hostapd path.

---

package/network/config/wifi-scripts/files/lib/wifi/mac80211.uc
robimarko: This looks potentially correct from the AI

Leo-PL: Indeed it does, I will check that on Monday. I've noticed that LuCI would show wrong frequency for the channel selected, as if A band was in place.
Edit: I made the fix, but have yet to test it on hardware, so not pushing yet.

Leo-PL:
@robimarko added the 2nd fix for `hw_mode`, it fixed frequency display in LuCI, so everything else should be fine too.

package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/hostapd.uc
```diff
@@ -32,8 +32,10 @@ function set_device_defaults(config) {
 	/* validate band */
 	if (config.band == '2g')
 		config.hw_mode = 'g';
-	else if (config.band in [ '5g', '6g', '60g' ])
+	else if (config.band in [ '5g', '6g' ])
 		config.hw_mode = 'a';
+	else if (config.band == '60g')
+		config.hw_mode = 'ad';
```
openwrt-ai, line 38: This fixes `hw_mode` on the *device* config, which covers the hostapd `hw_mode=ad` line and the LuCI frequency display, but the cipher part named in the commit message ("incorrect cipher selection") is still not reached.

[`iface.uc:119-126`](``https://github.com/openwrt/openwrt/blob/b55489cb3ff0a1993d43e5757df4c10efdeb3dd7/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/iface.uc#L119-L126)`` tests `config.hw_mode == 'ad'` before defaulting `wpa_pairwise = 'GCMP'`, but inside `parse_encryption()` the first argument is the *wifi-iface* config — the device config is the separate `dev_config` parameter, used a few lines up for `dev_config.band` / `dev_config.htmode`. See the call site [`ap.uc:569`](``https://github.com/openwrt/openwrt/blob/b55489cb3ff0a1993d43e5757df4c10efdeb3dd7/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/ap.uc#L569):`` `iface.parse_encryption(config, data.config, phy_features)`.

`hw_mode` only exists as a wifi-device option ([wireless.wifi-device.json:386](``https://github.com/openwrt/openwrt/blob/b55489cb3ff0a1993d43e5757df4c10efdeb3dd7/package/network/config/wifi-scripts/files-ucode/usr/share/schema/wireless.wifi-device.json#L386));`` `wireless.wifi-iface.json` has no such property, so `config.hw_mode` is `undefined` there and an encrypted 60 GHz BSS still falls through to `wpa_pairwise = 'CCMP'`. The shell path does set GCMP for `hwmode=ad` at [netifd-wireless.sh:222-223](``https://github.com/openwrt/openwrt/blob/b55489cb3ff0a1993d43e5757df4c10efdeb3dd7/package/network/config/wifi-scripts/files/lib/netifd/netifd-wireless.sh#L222-L223).``

Making the test `dev_config?.hw_mode == 'ad'` (or `dev_config?.band == '60g'`) in `iface.uc` would give the ucode path parity. This would not have shown up in the run test, since the auto-generated 60G section uses `encryption='none'`.

---

package/network/config/wifi-scripts/Makefile
```diff
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=wifi-scripts
 PKG_VERSION:=1.0
-PKG_RELEASE:=1
+PKG_RELEASE:=2
```
openwrt-ai, line 12: nit: this bump belongs to the first commit; the second commit (`wifi-scripts: set hw_mode = 'ad' ...`) changes installed package content (`files-ucode/usr/share/ucode/wifi/hostapd.uc`) without touching `PKG_RELEASE`. If the two are meant to be bisectable / backported separately, the second one should carry its own bump to 3 — intentional to keep a single bump for the series?

---

openwrt-ai (commented):
Reviewed 1 new commit (`b55489cb`). The `hw_mode = 'ad'` change matches its commit message and is the right shape — it also stops 60 GHz from picking up the 5 GHz OFDM rate lists in `device_cell_density_append()` and the VHT/HE/EHT `hw_mode == 'a'` branches. One inline note: the cipher-selection half of the described fix still doesn't trigger, because `iface.uc` reads `hw_mode` off the wifi-iface config rather than the device config. CI on the head commit is green.

---

Leo-PL:
@robimarko the last check mentioned by LLM was also right, so I fixed it, and while at that refactored PR bump into separate commit with a summary. What do you think?

package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/iface.uc
```diff
@@ -118,7 +118,7 @@ export function parse_encryption(config, dev_config, phy_features) {
 
 	if (!config.wpa)
 		config.wpa_pairwise ??= null;
-	else if (config.hw_mode == 'ad')
+	else if (dev_config.hw_mode == 'ad')
```
openwrt-ai: This fixes the AP path, but `hw_mode` is only ever populated by `set_device_defaults()`, which runs on the hostapd path only ([`set_device_defaults(config)` at hostapd.uc:507](``https://github.com/openwrt/openwrt/blob/7729d9589842d349fb9bd010b5dc10c524962af7/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/hostapd.uc#L507)).`` `supplicant.uc` never sets `hw_mode` anywhere, yet it calls this same function with the *device* config as `dev_config`: [`iface.parse_encryption(config, data)` at supplicant.uc:64](``https://github.com/openwrt/openwrt/blob/7729d9589842d349fb9bd010b5dc10c524962af7/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/supplicant.uc#L64),`` reached from [supplicant.uc:306](``https://github.com/openwrt/openwrt/blob/7729d9589842d349fb9bd010b5dc10c524962af7/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/supplicant.uc#L306)`` (`setup_sta(data.config, interface.config)`).

So for a 60 GHz `sta`/`mesh`/`adhoc` interface `dev_config.hw_mode` is still undefined and `wpa_pairwise` falls through to `CCMP`, which then becomes `pairwise=CCMP` in the network block — the GCMP special case at [supplicant.uc:195-200](``https://github.com/openwrt/openwrt/blob/7729d9589842d349fb9bd010b5dc10c524962af7/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/supplicant.uc#L195-L200)`` (which also forces `group=GCMP`) never fires, so a STA on 11ad still can't match a GCMP-only peer.

`band` is normalized to `'60g'` for both paths before either handler runs ([mac80211.sh:193](``https://github.com/openwrt/openwrt/blob/7729d9589842d349fb9bd010b5dc10c524962af7/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh#L193)),`` and with the `hw_mode = 'ad'` change in this series `band == '60g'` and `hw_mode == 'ad'` are equivalent on the AP path, so testing `band` here is a strict superset:

```suggestion
	else if (dev_config.band == '60g')
```

Was the supplicant path deliberately left out of scope?


---

package/network/config/wifi-scripts/Makefile
openwrt-ai: nit: the commit that carries this hunk is subjected `wifi-scripts: bump PR for 802.11ad fixes` — "PR" reads as "pull request" here, and in-tree standalone bumps consistently spell it out, e.g. [`ucode-mod-bpf: bump PKG_RELEASE`](https://github.com/openwrt/openwrt/commit/5b3b961f41e7d77242413d3e007336d94c445c82) and [`hostapd: bump PKG_RELEASE`](https://github.com/openwrt/openwrt/commit/89d705148553a19a9dc32191cc85aff7616cf536). Suggest `wifi-scripts: bump PKG_RELEASE for 802.11ad fixes` (and s/Bump PR/Bump PKG_RELEASE/ in the body).


---

openwrt-ai (commented):
Reviewed 2 new commits since `b55489cb` (`bb6bfeff`, `7729d958`); the first two commits were also rewritten (message reword on `641685ff`, `PKG_RELEASE` bump moved out of `2a74b0aa`), so the net tree change since the last review is just the `iface.uc` one-liner. Splitting the bump into its own commit resolves the earlier bisectability nit. One inline note: the `dev_config.hw_mode` test still misses the wpa_supplicant (`sta`/`mesh`/`adhoc`) path, which never populates `hw_mode`. CI on `7729d958` reports no failures.

## Commit checks
- `2a74b0aa` "wifi-scripts: add 60G band to mac80211.uc to support 802.11ad" — the bracketed annotation still claims `bumped PKG_RELEASE`, but the bump now lives in `7729d958`; this commit only touches `files/lib/wifi/mac80211.uc`. Drop that phrase from the bracket.


---

openwrt-ai (commented):

The force-push since `7729d958` is commit-message-only — `git diff 7729d958..85d75d6e` is empty, so the tree is byte-identical to what I reviewed last night. Both message findings from the previous round are addressed: `9c9f8508` no longer claims `bumped PKG_RELEASE` in its bracketed annotation, and `85d75d6e` now spells out `PKG_RELEASE` in subject and body instead of `PR`. All four messages match their diffs. CI on `85d75d6e` is green (FormalityCheck, malta/be and x86/64 package builds all succeeded).

The one open inline question from the previous review ([the `dev_config.hw_mode` test in `iface.uc` vs. the wpa_supplicant `sta`/`mesh`/`adhoc` path](https://github.com/openwrt/openwrt/pull/24586#discussion_r3755087667)) still stands, since the code is unchanged — not re-posting it here.

---

package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/iface.uc
Leo-PL: I'm not sure if it's the right fix, or if we should check for hw_mode also in wpa_supplicant path instead, and fix the actual root cause here. @hauke  what do you think?

hauke: I would prefer if you do this change.

As far as I understood `hw_mode` is only set in the hostapd script and not for wpa_supplicant.

Leo-PL: Nope, it doesn't work for STA mode, clanker confused this with old shell implementation:
```
Mon Jun 29 13:43:04 2026 daemon.notice netifd: radio2 (3872): Reference error: left-hand side expression is null
Mon Jun 29 13:43:04 2026 daemon.notice netifd: radio2 (3872): In parse_encryption(), file /usr/share/ucode/wifi/iface.uc, line 121, byte 22:
Mon Jun 29 13:43:04 2026 daemon.notice netifd: radio2 (3872):   called from function setup_sta (/usr/share/ucode/wifi/supplicant.uc:61:31)
Mon Jun 29 13:43:04 2026 daemon.notice netifd: radio2 (3872):   called from function generate (/usr/share/ucode/wifi/supplicant.uc:266:41)
Mon Jun 29 13:43:04 2026 daemon.notice netifd: radio2 (3872):   called from function setup (./mac80211.sh:254:61)
Mon Jun 29 13:43:04 2026 daemon.notice netifd: radio2 (3872):   called from anonymous function (./mac80211.sh:341:14)
Mon Jun 29 13:43:04 2026 daemon.notice netifd: radio2 (3872):
Mon Jun 29 13:43:04 2026 daemon.notice netifd: radio2 (3872):  `    else if (dev_config.band == '60g')`
Mon Jun 29 13:43:04 2026 daemon.notice netifd: radio2 (3872):   Near here --------------^
Mon Jun 29 13:43:04 2026 daemon.notice netifd: radio2 (3872):
Mon Jun 29 13:43:04 2026 daemon.notice netifd: radio2 (3872):
Mon Jun 29 13:43:04 2026 daemon.notice netifd: radio2 (3872): wifi-scripts: Configuring 'phy0' txantenna: 4294967295, rxantenna: 4294967295 distance: 0
Mon Jun 29 13:43:04 2026 daemon.notice netifd: radio2 (3872): wifi-scripts: Preparing interface: phy0-sta0 with MAC: 04:ce:14:0b:7c:cf
Mon Jun 29 13:43:04 2026 daemon.notice netifd: radio2 (3919): wifi-scripts: Tearing down phy0
```
I'll have to define either of them beforehand.

hauke:
When you want to backport this to 25.12 make sure it cherry pick 97e8911c9bda too.

package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/iface.uc
hauke: It fails becasue `dev_config` is not defined, you are probably running something without 97e8911c9bdaf73579cfd7696bde48b2bb314954.

Leo-PL: Hah. Nice catch, I was testing over 25.12 but doing my changes to main, just copying over the scripts. No wonder it failed, but with this change, switching to checking of `dev_config.band` now works. So, updated - this should be ready.

openwrt-ai (commented):

Since `85d75d6e` the series was rebased onto the current base and two commits were rewritten (`75dfa4cb` → `a2cdcf36`, `85d75d6e` → `e4fcc0bf`). `git diff 85d75d6e..e4fcc0bf` is a single line in `iface.uc`: `dev_config.hw_mode == 'ad'` → `dev_config.band == '60g'`, which closes the [open question from the previous round](https://github.com/openwrt/openwrt/pull/24586#discussion_r3755087667) about the wpa_supplicant path.

Verified the new test holds on both handlers:

- `band` is guaranteed populated before either handler runs — [`mac80211.sh:185-203`](``https://github.com/openwrt/openwrt/blob/e4fcc0bffd2e93bfa08ec51094f3ac8d9294e70a/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh#L185-L203)`` derives it from `hwmode` (`ad`/`11ad` → `60g`) when unset and then `delete`s `hwmode`, so `band == '60g'` is a strict superset of the old `hw_mode == 'ad'` test rather than a narrowing.
- AP path: [`ap.uc:569`](``https://github.com/openwrt/openwrt/blob/e4fcc0bffd2e93bfa08ec51094f3ac8d9294e70a/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/ap.uc#L569)`` passes `data.config` as `dev_config`; testing `dev_config.band` also matches the surrounding style, since the very next line already does `data.config.band == '6g'`.
- Supplicant path: [`setup_sta(data.config, interface.config)` at supplicant.uc:306](``https://github.com/openwrt/openwrt/blob/e4fcc0bffd2e93bfa08ec51094f3ac8d9294e70a/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/supplicant.uc#L306)`` → [`iface.parse_encryption(config, data)` at supplicant.uc:64](``https://github.com/openwrt/openwrt/blob/e4fcc0bffd2e93bfa08ec51094f3ac8d9294e70a/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/supplicant.uc#L64),`` so `dev_config` is the device config here too and `wpa_pairwise = 'GCMP'` now reaches the [`pairwise`/`group` GCMP branch at supplicant.uc:194-199](``https://github.com/openwrt/openwrt/blob/e4fcc0bffd2e93bfa08ec51094f3ac8d9294e70a/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/supplicant.uc#L194-L199).`` This also explains the `Reference error: left-hand side expression is null` seen on the 25.12 test tree — that branch predates the `dev_config` parameter, as @hauke noted.

`hw_mode = 'ad'` remains valid on the hostapd side (`hostapd/config_file.c` accepts `ad` for both `hw_mode` and `ieee80211ad`, hostap `7533e0a8`). All four commit messages match their diffs. CI on `e4fcc0bf` reports no failures (FormalityCheck and both package checks green, the two `Build` jobs still running).

---

openwrt-ai (commented):

The update since `e4fcc0bf` is a pure rebase from `f4a7ae10` onto `20d94d5a`. `git range-diff f4a7ae10..e4fcc0bf 20d94d5a..a7d11f1c` reports all four commits identical (`=`), and `git log 20d94d5a..f4a7ae10 -- package/network/config/wifi-scripts/` is empty, so nothing in the new base touches this package and the rebase changes neither the patches nor their context. The `PKG_RELEASE` bump is still `1` -> `2` against the new base, so no collision with an independent bump upstream.

All four commit messages match their diffs, and every earlier review thread is now resolved or answered — including the [`dev_config.band == '60g'` question](https://github.com/openwrt/openwrt/pull/24586#discussion_r3755087667), which @hauke's pointer to `97e8911c9bda` explained (the test tree lacked the `dev_config` parameter). Note for the backport: per that same comment, `97e8911c9bda` needs to land on 25.12 before this series, otherwise `iface.uc:121` hits `Reference error: left-hand side expression is null` on the supplicant path.

CI on `a7d11f1c` is green: FormalityCheck (Git & Commits, Code Patches, OpenWrt Makefiles), build-pr-profile, and both malta/be and x86/64 build + package checks all succeeded.

---

PR #24588, opened 2026-08-06
realtek: board: some family check cleanups

jameywine:
A relatively straightforward cleanup of family checks in the prom initialization by the power of match data.

Did the boot test on my rtl9607c machine and it is working.

### DIFF
--- a/target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
+++ b/target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
@@ -21,7 +21,7 @@ Signed-off-by: Markus Stockhausen <markus.stockhausen@gmx.de>
  load-$(CONFIG_MIPS_GENERIC)	+= 0xffffffff80100000
 --- a/arch/mips/generic/board-realtek.c
 +++ b/arch/mips/generic/board-realtek.c
-@@ -11,6 +11,344 @@
+@@ -11,6 +11,312 @@
  #include <asm/fw/fw.h>
  #include <asm/machine.h>
  
@@ -46,6 +46,16 @@ Signed-off-by: Markus Stockhausen <markus.stockhausen@gmx.de>
 +
 +#define soc_r32(reg)			__raw_readl(RTL_SOC_BASE + reg)
 +#define soc_w32(val, reg)		__raw_writel(val, RTL_SOC_BASE + reg)
++
++struct realtek_soc_data {
++	u32 model_info_reg;
++	u32 family;
++	int cpu_port;
++	void (*read_details)(u32 model);
++	void (*apply_quirks)(void);
++	void (*get_memory)(void);
++	void (*prepare_highmem)(void);
++};
 +
 +struct rtl83xx_soc_info soc_info;
 +
@@ -81,29 +91,27 @@ Signed-off-by: Markus Stockhausen <markus.stockhausen@gmx.de>
 +		};
 +		of_node_put(node);
 +	}
-+	
++
 +	return 0;
 +}
 +
 +arch_initcall(realtek_apply_dts_quirks);
 +
-+static void __init realtek_apply_early_quirks(void)
++static void __init rtl838x_apply_early_quirks(void)
 +{
-+	if (soc_info.family == RTL8380_FAMILY_ID) {
-+		/*
-+		 * Open up write protected registers. SDK opens/closes this whenever needed. For
-+		 * simplicity always work with an "open" register set.
-+		 */
-+		sw_w32(0x3, RTL838X_INT_RW_CTRL);
-+		/*
-+		 * Disable 4 byte address mode of flash controller. If this bit is not cleared
-+		 * the watchdog cannot reset the SoC. The SDK changes this short before restart.
-+		 * Until this quirk was implemented all RTL838x devices ran with this disabled
-+		 * because of a coding error. As no issues were detected keep the behaviour
-+		 * until more details are known.
-+		 */
-+		sw_w32_mask(BIT(30), 0, RTL838X_PLL_CML_CTRL);
-+	}
++	/*
++	 * Open up write protected registers. SDK opens/closes this whenever needed. For
++	 * simplicity always work with an "open" register set.
++	 */
++	sw_w32(0x3, RTL838X_INT_RW_CTRL);
++	/*
++	 * Disable 4 byte address mode of flash controller. If this bit is not cleared
++	 * the watchdog cannot reset the SoC. The SDK changes this short before restart.
++	 * Until this quirk was implemented all RTL838x devices ran with this disabled
++	 * because of a coding error. As no issues were detected keep the behaviour
++	 * until more details are known.
++	 */
++	sw_w32_mask(BIT(30), 0, RTL838X_PLL_CML_CTRL);
 +}
 +
 +static void __init rtl838x_read_details(u32 model)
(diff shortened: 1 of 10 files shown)


target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
```diff
@@ -366,36 +329,85 @@ Signed-off-by: Markus Stockhausen <markus.stockhausen@gmx.de>
...
-+	{ .compatible = "realtek,rtl930x-soc" },
-+	{ .compatible = "realtek,rtl9311-soc" },
-+	{ .compatible = "realtek,rtl9312-soc" },
-+	{ .compatible = "realtek,rtl9313-soc" },
-+	{ .compatible = "realtek,rtl931x-soc" },
+-	{ .compatible = "realtek,rtl9302-soc" },
++	{ .compatible = "realtek,rtl8380-soc", .data = &rtl838x_soc },
++	{ .compatible = "realtek,rtl8381-soc", .data = &rtl838x_soc },
++	{ .compatible = "realtek,rtl8382-soc", .data = &rtl838x_soc },
++	{ .compatible = "realtek,rtl838x-soc", .data = &rtl838x_soc },
```
openwrt-ai, line 398: Attaching `.data` to the match table makes the DT compatible string the *sole* source of SoC identity. Before, `realtek_read_model()` probed the real `*_MODEL_NAME_INFO` registers and derived the family from the hardware, so a sloppy/wrong SoC compatible in a board DTS was harmless. Now it decides everything, and [`mips_machine_is_compatible()` at machine.h:51-54](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/mips/include/asm/machine.h#L51-L54) returns the **first** table entry whose string appears anywhere in the root `compatible` list — so `realtek,rtl838x-soc` at index 3 wins over every 930x/931x entry below it.

Six in-tree RTL9302 boards name `realtek,rtl838x-soc` as their only SoC compatible:

- [`rtl9302_plasmacloud_psx8.dts:7`](https://github.com/openwrt/openwrt/blob/3d8e4e5bb4886c11f9d9488e3002339fab3ecbf1/target/linux/realtek/dts/rtl9302_plasmacloud_psx8.dts#L7)
- [`rtl9302_plasmacloud_psx10.dts:7`](https://github.com/openwrt/openwrt/blob/3d8e4e5bb4886c11f9d9488e3002339fab3ecbf1/target/linux/realtek/dts/rtl9302_plasmacloud_psx10.dts#L7)
- [`rtl9302_zyxel_xgs1210-12-a1.dts:7`](https://github.com/openwrt/openwrt/blob/3d8e4e5bb4886c11f9d9488e3002339fab3ecbf1/target/linux/realtek/dts/rtl9302_zyxel_xgs1210-12-a1.dts#L7) and `-b1`
- [`rtl9302_zyxel_xgs1250-12-a1.dts:7`](https://github.com/openwrt/openwrt/blob/3d8e4e5bb4886c11f9d9488e3002339fab3ecbf1/target/linux/realtek/dts/rtl9302_zyxel_xgs1250-12-a1.dts#L7) and `-b1`

Those six will bind to `&rtl838x_soc` after this change, meaning: `model` is read from `RTL838X_MODEL_NAME_INFO` (0x00d4) instead of `RTL93XX_MODEL_NAME_INFO` (0x0004), `soc_info.id`/`family` end up wrong (`RTL8380_FAMILY_ID`), `rtl838x_read_details()` pokes `RTL838X_CHIP_INFO`/`RTL838X_EXT_VERSION`, `rtl838x_apply_early_quirks()` writes `RTL838X_INT_RW_CTRL` and `RTL838X_PLL_CML_CTRL` on a 9300-series SoC, and `rtl930x_prepare_highmem()` never runs. Since `soc_info.family` also feeds the DSA/switch drivers, this is a boot-level regression on those boards.

Same trap one level up: [`rtl930x.dtsi:11`](https://github.com/openwrt/openwrt/blob/3d8e4e5bb4886c11f9d9488e3002339fab3ecbf1/target/linux/realtek/dts/rtl930x.dtsi#L11) and [`rtl931x.dtsi:12`](https://github.com/openwrt/openwrt/blob/3d8e4e5bb4886c11f9d9488e3002339fab3ecbf1/target/linux/realtek/dts/rtl931x.dtsi#L12) both declare `compatible = "realtek,rtl838x-soc"` at the root. Every board currently overrides it, but any new board that forgets to would silently land on `&rtl838x_soc` too.

Reordering the table doesn't help (those boards carry no 930x string at all) — the DTS compatibles need to be corrected as part of this series, and the two `*.dtsi` roots along with them.

---

target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
```diff
@@ -81,29 +91,27 @@ Signed-off-by: Markus Stockhausen <markus.stockhausen@gmx.de>
...
++	* simplicity always work with an "open" register set.
++	*/
++	sw_w32(0x3, RTL838X_INT_RW_CTRL);
++	/*
++	* Disable 4 byte address mode of flash controller. If this bit is not cleared
++	* the watchdog cannot reset the SoC. The SDK changes this short before restart.
++	* Until this quirk was implemented all RTL838x devices ran with this disabled
++	* because of a coding error. As no issues were detected keep the behaviour
++	* until more details are known.
++	*/
```
openwrt-ai: De-indenting these two comment blocks by one level dropped the space before the continuation `*`, so they no longer follow the kernel multi-line comment style used everywhere else in this file (e.g. the `rtl930x_prepare_highmem()` and `realtek_prom_init()` comments below).

```suggestion
+	 * Open up write protected registers. SDK opens/closes this whenever needed. For
+	 * simplicity always work with an "open" register set.
+	 */
+	sw_w32(0x3, RTL838X_INT_RW_CTRL);
+	/*
+	 * Disable 4 byte address mode of flash controller. If this bit is not cleared
+	 * the watchdog cannot reset the SoC. The SDK changes this short before restart.
+	 * Until this quirk was implemented all RTL838x devices ran with this disabled
+	 * because of a coding error. As no issues were detected keep the behaviour
+	 * until more details are known.
+	 */
```

---

target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
```diff
@@ -342,14 +294,25 @@ Signed-off-by: Markus Stockhausen <markus.stockhausen@gmx.de>
...
++	u32 model = sw_r32(cfg->model_info_reg);
 +
++	soc_info.id = model >> 16 & 0xffff;
++	soc_info.family = cfg->family;
++	soc_info.cpu_port = cfg->cpu_port;
++
++	if (cfg->apply_quirks)
++		cfg->apply_quirks();
++
++	cfg->read_details(model);
```
openwrt-ai, line 314: `cfg` is dereferenced unconditionally (`cfg->model_info_reg` two lines above, then `cfg->read_details()` / `cfg->get_memory()` here) with no NULL check, while `apply_quirks` and `prepare_highmem` are guarded. All 21 table entries currently carry `.data`, but this runs before any console is up, so a future entry added without `.data` fails as a silent hang rather than a diagnosable error. An early `if (!cfg) return;` (or at least keeping the two mandatory callbacks guarded the same way as the optional ones) would make that failure mode visible.

---

target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
```diff
@@ -366,36 +329,85 @@ Signed-off-by: Markus Stockhausen <markus.stockhausen@gmx.de>
...
++	.get_memory = rtl931x_get_system_memory,
++};
++
++static const struct realtek_soc_data rtl960x_soc __initconst = {
++	.model_info_reg = RTL96XX_MODEL_NAME_INFO,
++	.family = RTL9607_FAMILY_ID,
++	.cpu_port = RTL9607_CPU_PORT,
++	.read_details = rtl96xx_read_details,
++	.get_memory = realtek_get_system_memory,
++	.prepare_highmem = rtl930x_prepare_highmem,
```
openwrt-ai, line 390: nit: renaming `realtek_prepare_highmem()` to `rtl930x_prepare_highmem()` makes the shared use here slightly misleading — the function is also the RTL9607 path, and its `pr_info("highmem kernel on RTL930x with > 256 MB RAM, ...")` will print "RTL930x" on a 9607 board. The behaviour matches the old `family != RTL9300 && family != RTL9607` check so it's not a regression, but since the function now carries an SoC-specific name, either keep a neutral name or drop the SoC name from the message.

---

target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
```diff
@@ -366,36 +329,85 @@ Signed-off-by: Markus Stockhausen <markus.stockhausen@gmx.de>
...
++	{ .compatible = "realtek,rtl9301-soc", .data = &rtl930x_soc },
++	{ .compatible = "realtek,rtl9302-soc", .data = &rtl930x_soc },
++	{ .compatible = "realtek,rtl9302c-soc", .data = &rtl930x_soc },
++	{ .compatible = "realtek,rtl9303-soc", .data = &rtl930x_soc },
++	{ .compatible = "realtek,rtl930x-soc", .data = &rtl930x_soc },
++	{ .compatible = "realtek,rtl9311-soc", .data = &rtl931x_soc },
++	{ .compatible = "realtek,rtl9312-soc", .data = &rtl931x_soc },
++	{ .compatible = "realtek,rtl9313-soc", .data = &rtl931x_soc },
++	{ .compatible = "realtek,rtl931x-soc", .data = &rtl931x_soc },
++	{ .compatible = "realtek,rtl8198d-soc", .data = &rtl960x_soc },
```
openwrt-ai, line 412: nit: `realtek,rtl8198d-soc` has no user anywhere in the tree (no `.dts`/`.dtsi`, no binding, no other patch mentions `rtl8198` at all), and the commit message only says "add the rtl960x compatibles" — RTL8198D isn't an RTL960x part name, so it's not obvious it is covered by that sentence. Is it genuinely the same silicon family as RTL9607C, i.e. does it share `RTL96XX_MODEL_NAME_INFO` (0x10000), `RTL96XX_CHIP_SUB_INFO`, `RTL9607_FAMILY_ID` and `RTL9607_CPU_PORT` (9)? If so, worth a line in the commit message; if it's speculative, it's probably better added together with the board support that needs it.

---

openwrt-ai (commented):
Reviewed 1 new commit. The refactor itself is a clear improvement and the per-family split faithfully reproduces the old conditionals (838x quirks, 931x DRAM register, 9300+9607 highmem). Both patches also still apply cleanly against v6.18 with correct hunk metadata, and the previous trailing-whitespace warnings are gone.

The one blocking concern is that moving SoC identification from register probing to `of_device_id` matching makes the DT `compatible` authoritative, and six in-tree RTL9302 boards currently declare `realtek,rtl838x-soc` — details inline.

---

target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
jonasjelonek: The DTS need to be changed accordingly, probably even in a pre-commit.

target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
jameywine: Yes, it is the same silicon family as RTL9607C. I can add that detail to the commit just to be sure.

target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch
jameywine: Yeah, i was thinking that should be a separate commit before the main one.

PR #24589, opened 2026-08-06
urngd: enable optimization for Jitter RNG 3.x

dormancygrace:
Jitter RNG 2.x required its core to be built without compiler optimization. This restriction was removed in Jitter RNG 3.x, and the pinned `3rdparty/jitterentropy-rngd` submodule's standalone Makefile uses `-O2`. The urngd CMake file nevertheless still forces `-O0` for `jitterentropy-base.c`.

The CPU-intensive entropy initialization can also monopolize a slow single-core CPU while the remaining init scripts are still starting. Run urngd at nice level 19 so network and management services can initialize first without disabling entropy generation.

Together these changes bring the CMake build in line with the standalone build, prevent urngd from blocking startup on CPU-constrained devices and bump the OpenWrt package release.

The CMake change has also been submitted upstream as openwrt/urngd#1. The local patch remains until that change is merged and OpenWrt updates the pinned urngd revision.

Fixes #21573.

Validation:

- OpenWrt `checkpatch.pl`: 0 errors, 0 warnings.
- The patch applies cleanly to urngd `f17e33d9942750edbd73124e6f03118a98abb402`.
- `make package/urngd/clean` and `make package/urngd/compile -j12 V=s` succeeded with GCC 14.3.0 for ramips/mt7620; generated build rules and the compiler invocation end in `-O2` for `jitterentropy-base.c`, while retaining OpenWrt's `_FORTIFY_SOURCE=1` toolchain flag.
- A complete WE826 image build with `make -j12 V=s` succeeded.
- On MIPS32 QEMU with the same compiler and source, initialization plus one 64-byte block improved from about 0.40 s to 0.13 s; four blocks improved from about 0.81 s to 0.26 s.
- The resulting binary was run on an MT7620 WE826; urngd stayed healthy and completed its entropy initialization successfully.
- On the WE826, lowering urngd's priority reduced cold boot time to SSH and HTTP from about 117 s to 81 s with the unoptimized entropy core, and to about 80 s together with `-O2`.

### DIFF
--- a/package/system/urngd/files/urngd.init
+++ b/package/system/urngd/files/urngd.init
@@ -9,6 +9,7 @@ PROG=/sbin/urngd
 start_service() {
 	procd_open_instance
 	procd_set_param command "$PROG"
+	procd_set_param nice 19
 	procd_close_instance
 }
 
--- a/package/system/urngd/patches/100-jitterentropy-enable-optimization.patch
+++ b/package/system/urngd/patches/100-jitterentropy-enable-optimization.patch
@@ -1,0 +1,32 @@
+From f7f9185b1296c005efe58cd756d2646ca00acb8d Mon Sep 17 00:00:00 2001
+From: Gleb Pesin <dormancygrace@gmail.com>
+Date: Thu, 6 Aug 2026 19:46:43 +0000
+Subject: [PATCH] CMakeLists: enable optimization for Jitter RNG 3.x
+
+Jitter RNG 2.x prohibited compiler optimization. That restriction was
+removed in 3.x and the standalone Makefile builds the core with -O2.
+Bring the CMake build in line with it while retaining fortification.
+This depends on the pinned 3rdparty/jitterentropy-rngd submodule being 3.x.
+
+Upstream-Status: Submitted [https://github.com/openwrt/urngd/pull/1]
+
+Signed-off-by: Gleb Pesin <dormancygrace@gmail.com>
+---
+ CMakeLists.txt | 4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -22,10 +22,10 @@ ADD_EXECUTABLE(urngd
+ )
+ TARGET_LINK_LIBRARIES(urngd ${ubox})
+ 
+-# jitter RNG must not be compiled with optimizations, _FORTIFY_SOURCE needs optimizations
++# Jitter RNG 3.x supports optimization; match its standalone Makefile.
+ SET_PROPERTY(
+   SOURCE ${JTEN_DIR}/jitterentropy-base.c
+-  APPEND PROPERTY COMPILE_OPTIONS -U_FORTIFY_SOURCE -O0
++  APPEND PROPERTY COMPILE_OPTIONS -O2
+ )
+ 
+ INSTALL(TARGETS urngd RUNTIME DESTINATION ${CMAKE_INSTALL_SBINDIR})
--- a/package/system/urngd/Makefile
+++ b/package/system/urngd/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=urngd
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL=$(PROJECT_GIT)/project/urngd.git
--- a/package/system/urngd/Makefile
+++ b/package/system/urngd/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=urngd
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL=$(PROJECT_GIT)/project/urngd.git


package/system/urngd/patches/100-jitterentropy-enable-optimization.patch
```diff
@@ -0,0 +1,26 @@
...
+Jitter RNG 2.x prohibited compiler optimization. That restriction was
+removed in 3.x and the standalone Makefile builds the core with -O2.
+Bring the CMake build in line with it.
+
+Signed-off-by: Gleb Pesin <dormancygrace@gmail.com>
+---
+ CMakeLists.txt | 4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
```
neheb: make package/urngd/{clean,refresh}

dormancygrace: Done. I ran `make package/urngd/{clean,refresh}` with the OpenWrt host tools, folded the refreshed patch into the original commit, and force-pushed the branch. The refreshed patch applies cleanly.

package/system/urngd/patches/100-jitterentropy-enable-optimization.patch
```diff
@@ -0,0 +1,29 @@
...
+@@ -22,10 +22,10 @@ ADD_EXECUTABLE(urngd
+ )
+ TARGET_LINK_LIBRARIES(urngd ${ubox})
+ 
+-# jitter RNG must not be compiled with optimizations, _FORTIFY_SOURCE needs optimizations
++# Jitter RNG 3.x supports optimization; match its standalone Makefile.
+ SET_PROPERTY(
+   SOURCE ${JTEN_DIR}/jitterentropy-base.c
+-  APPEND PROPERTY COMPILE_OPTIONS -U_FORTIFY_SOURCE -O0
++  APPEND PROPERTY COMPILE_OPTIONS -U_FORTIFY_SOURCE -O2
```
openwrt-ai: `-U_FORTIFY_SOURCE` is kept, but its only reason was the `-O0` this patch removes. urngd added it in [`44365eb` "Deactivate _FORTIFY_SOURCE in jitterentropy-base.c"](https://github.com/openwrt/urngd/commit/44365eb1e1165f2a44cb31f404b04cf85031718e): *"_FORTIFY_SOURCE only works with compiler optimizations activated. We have to deactivate it when we set -O0"* — otherwise `-Werror=cpp` trips on `#warning _FORTIFY_SOURCE requires compiling with optimization (-O)`. The old comment being replaced says the same thing.

With `-O2` restored that warning can no longer fire, so the `-U` now only serves to strip OpenWrt's default `-D_FORTIFY_SOURCE=1` (from `include/hardening.mk`) off this one file — a hardening regression with no remaining justification. Upstream Jitter RNG's own Makefile keeps fortify on at `-O2` (`-U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 ... -O2` in jitterentropy-library v3.0.0).

```suggestion
++  APPEND PROPERTY COMPILE_OPTIONS -O2
```

Worth considering going further and dropping the whole `SET_PROPERTY` block: OpenWrt overrides `CMAKE_C_FLAGS_RELEASE` to just `-DNDEBUG` ([include/cmake.mk:101](https://github.com/openwrt/openwrt/blob/577a76c0965e7a6a15766829bafc34296a644360/include/cmake.mk#L101)), so without the block the file simply inherits `CONFIG_TARGET_OPTIMIZATION` like every other source in the tree, instead of hardcoding `-O2` over the user's chosen `-Os`.

---

package/system/urngd/patches/100-jitterentropy-enable-optimization.patch
```diff
@@ -0,0 +1,29 @@
+From f7f9185b1296c005efe58cd756d2646ca00acb8d Mon Sep 17 00:00:00 2001
```
openwrt-ai, line 1: Should this go into urngd.git directly rather than being carried as a local patch? urngd is an OpenWrt-hosted project (`PKG_SOURCE_URL=$(PROJECT_GIT)/project/urngd.git`), and none of the other `$(PROJECT_GIT)` core packages — ubus, ubox, procd, fstools, netifd, rpcd, uci — carry a `patches/` directory at all; `package/network/utils/iwinfo` is the only exception in the tree. The usual flow for these is a commit in the project repo followed by a `PKG_SOURCE_DATE` / `PKG_SOURCE_VERSION` / `PKG_MIRROR_HASH` bump here, which also avoids the patch going stale on the next urngd bump. If the intent is to land it upstream first and this patch is just an interim, saying so in the commit message would help.

---

package/system/urngd/patches/100-jitterentropy-enable-optimization.patch
```diff
@@ -0,0 +1,29 @@
...
+From: Gleb Pesin <dormancygrace@gmail.com>
+Date: Thu, 6 Aug 2026 19:46:43 +0000
+Subject: [PATCH] CMakeLists: enable optimization for Jitter RNG 3.x
+
+Jitter RNG 2.x prohibited compiler optimization. That restriction was
+removed in 3.x and the standalone Makefile builds the core with -O2.
```
openwrt-ai, line 7: nit: the patch text here is right ("*its* standalone Makefile", i.e. Jitter RNG's), but the PR description says "the standalone Makefile **bundled with urngd** now uses `-O2`", which is not the case — urngd's top-level `Makefile` is just a CMake wrapper with no optimization flags. The `-O2` comes from the `3rdparty/jitterentropy-rngd` submodule (pinned at jitterentropy-rngd v1.2.0, which carries Jitter RNG core 3.0.0). Naming the submodule in the PR body would make the claim checkable.

For the record, the premise verifies: jitterentropy-library `v2.2.0` Makefile has `CFLAGS +=-Wextra -Wall -pedantic -fPIC -O0`, `v3.0.0` has `CFLAGS +=-Wextra -Wall -pedantic -fPIC -O2 -fwrapv -Wconversion`, and `jitterentropy-base.c` in the pinned submodule reports `MAJVERSION 3 / MINVERSION 0`.

---

openwrt-ai (commented):
Reviewed 2 commits. Both commit messages match their diffs, and the refreshed patch applies cleanly to the pinned urngd source (`f17e33d`) — verified by dry-run. `procd_set_param nice 19` is a valid procd instance parameter (`INSTANCE_ATTR_NICE`, accepted range -20..20).

Three inline notes, the first one substantive.

---

package/system/urngd/patches/100-jitterentropy-enable-optimization.patch
dormancygrace: Fixed in 19a0a33dbf: the per-file options are now only , so OpenWrt's  remains enabled. A ramips/mt7620 cross-build succeeded and the compile rule contains both  and the final . I kept the explicit  because that is the tested Jitter RNG 3.x upstream setting and is also the change submitted to openwrt/urngd#1.

package/system/urngd/patches/100-jitterentropy-enable-optimization.patch
dormancygrace: Agreed. I submitted the source change upstream as openwrt/urngd#1. This local patch is now documented in the PR description as temporary until that change is merged and OpenWrt bumps the pinned urngd revision.

package/system/urngd/patches/100-jitterentropy-enable-optimization.patch
dormancygrace: Corrected the PR description: it now explicitly identifies the pinned  submodule and its standalone Makefile.

package/system/urngd/patches/100-jitterentropy-enable-optimization.patch
dormancygrace: Correction: the per-file options are now only -O2, so OpenWrt's -D_FORTIFY_SOURCE=1 remains enabled. The cross-build compile rule contains both flags.

package/system/urngd/patches/100-jitterentropy-enable-optimization.patch
dormancygrace: Correction: the PR description now names the pinned 3rdparty/jitterentropy-rngd submodule explicitly.

package/system/urngd/patches/100-jitterentropy-enable-optimization.patch
```diff
@@ -0,0 +1,29 @@
...
+Bring the CMake build in line with it.
```
openwrt-ai: The patch now does two things but the message only describes one. Besides `-O0` → `-O2`, it drops `-U_FORTIFY_SOURCE`, which re-enables OpenWrt's `-D_FORTIFY_SOURCE=1` ([include/hardening.mk:46-49](https://github.com/openwrt/openwrt/blob/639a940b793ba34950df56af67ff6e7df1a67ee6/include/hardening.mk#L46-L49)) on `jitterentropy-base.c`. That is a hardening behaviour change, and it is the part an urngd maintainer is most likely to question — worth one sentence here (and in the OpenWrt commit message).

```suggestion
Bring the CMake build in line with it.

Also drop -U_FORTIFY_SOURCE: it was only needed because _FORTIFY_SOURCE
emits a #warning at -O0. With -O2 restored, fortification can stay
enabled for this file.
```

---

package/system/urngd/patches/100-jitterentropy-enable-optimization.patch
```diff
@@ -0,0 +1,29 @@
+From f7f9185b1296c005efe58cd756d2646ca00acb8d Mon Sep 17 00:00:00 2001
+From: Gleb Pesin <dormancygrace@gmail.com>
+Date: Thu, 6 Aug 2026 19:46:43 +0000
+Subject: [PATCH] CMakeLists: enable optimization for Jitter RNG 3.x
```
openwrt-ai, line 4: Has openwrt/urngd#1 been updated to match this revision? The PR body states the plan is to drop this local patch once that PR lands and `PKG_SOURCE_VERSION` is bumped, but the patch content changed after urngd#1 was filed (`-U_FORTIFY_SOURCE -O2` → `-O2`). If urngd#1 merges with `-U_FORTIFY_SOURCE` still in place, the fortify re-enable made here silently disappears at the next revision bump, since the patch that carries it gets deleted at the same time.

(I can't read openwrt/urngd from this session, so this may already be handled — flagging only because the two must stay in sync for the interim-patch plan to work.)

---

openwrt-ai (commented):
Re-reviewed the delta since `577a76c`. Both commits were rewritten, but the only content change is dropping `-U_FORTIFY_SOURCE` from the per-file `COMPILE_OPTIONS` — that resolves my earlier comment, and the reasoning holds: with `-O2` appended after `CMAKE_C_FLAGS`, `-D_FORTIFY_SOURCE=1..3` from `include/hardening.mk` can no longer trip the "requires compiling with optimization" warning. The patch's diffstat (`4 ++--`, 2/2) still matches its hunk. CI is green on `639a940`.

## Commit checks

- `19a0a33` "urngd: enable optimization for Jitter RNG 3.x" — the diff now also removes `-U_FORTIFY_SOURCE`, restoring OpenWrt's fortify flag on `jitterentropy-base.c`. The message describes only the `-O0` → `-O2` switch. Undocumented hunk; see the inline note.

Two inline comments, both minor.

---

package/system/urngd/patches/100-jitterentropy-enable-optimization.patch
dormancygrace: Fixed in 11a0263. Both the embedded upstream patch message and the OpenWrt commit message now explain that removing -U_FORTIFY_SOURCE restores the toolchain fortification at -O2.

package/system/urngd/patches/100-jitterentropy-enable-optimization.patch
dormancygrace: Yes. openwrt/urngd#1 is already synchronized: its head 8503c52 changes the option to -O2 without -U_FORTIFY_SOURCE, and its commit message explicitly says fortification is retained.

dormancygrace:
Retargeted this PR from the now-stale master branch to main and rebased both commits without content changes.

namiltd:
Nice work, well validated. Small suggestion: a comment in the patch (or CMakeLists.txt) noting that dropping -O0 depends on the pinned 3rdparty/jitterentropy-rngd submodule being 3.x, with a link to openwrt/urngd#1, would help anyone reading the code later without this PR description. Not blocking.

dormancygrace:
Thanks, added this to the downstream patch header. It now explicitly states that the change depends on the pinned `3rdparty/jitterentropy-rngd` submodule being 3.x and carries `Upstream-Status: Submitted` with a link to openwrt/urngd#1. The patch body is unchanged; the branch was rewritten and CI has restarted.

package/system/urngd/files/urngd.init
```diff
@@ -9,6 +9,7 @@ PROG=/sbin/urngd
 start_service() {
 	procd_open_instance
 	procd_set_param command "$PROG"
+	procd_set_param nice 19
```
openwrt-ai, line 12: nit: this commit changes an installed file (`/etc/init.d/urngd`) but carries no `PKG_RELEASE` bump — the `1` → `2` bump lives only in the preceding commit `c70513c`, which covers the CMake patch.

That is harmless as long as the two land together, but this commit is the obvious backport candidate of the pair: it is a self-contained fix for #21573 and does not depend on the Jitter RNG 3.x patch. Cherry-picked alone onto a release branch it would produce an `urngd` with an unchanged version, so package managers would not offer the rebuilt init script as an upgrade. Bumping to `3` here keeps each commit self-sufficient.

Conflicting evidence: `.github/formalities.json` sets `check_pkg_release` to `warning` and `FormalityCheck / OpenWrt Makefiles` is green on `4a9585b`, so the bot is happy with the single PR-level bump. Intentional?

---

openwrt-ai, line 12: nit: (about this commit's message, no diff line to anchor on) — the `Fixes: #21573` trailer mixes two conventions. In this tree `Fixes:` is reserved for a commit reference, e.g. [`a3c3a18`](https://github.com/openwrt/openwrt/commit/a3c3a18cc3932965c973d84d3166501abb202224) uses `Fixes: e050bb7946 ("zynq: add Bitmain Antminer S9 control board support")`, while GitHub issues are referenced by full URL — [`f59200b`](https://github.com/openwrt/openwrt/commit/f59200bc56a4250028df060c893b64d82f561055) uses `Closes: https://github.com/openwrt/openwrt/issues/24454` and [`a3c3a18`](https://github.com/openwrt/openwrt/commit/a3c3a18cc3932965c973d84d3166501abb202224) also carries `Link: https://github.com/openwrt/openwrt/issues/24602`. A bare `#21573` also does not resolve for anyone reading the commit outside github.com, which matters since openwrt.git is the canonical tree and this repo is a mirror.

Suggest `Closes: https://github.com/openwrt/openwrt/issues/21573` instead. `FormalityCheck / Git & Commits` is green either way, so purely cosmetic.

(Checked the reference itself: issue #21573 "Long boot time on D-Link DWR-921 C3 starting from OpenWrt 22.03" is open and is ramips/mt7620 — consistent with the WE826 measurements in the message.)

---

openwrt-ai (commented):
Re-reviewed the delta since `639a940`. Both commits were rewritten again, but no code changed — the CMake hunk (`-O2`, no `-U_FORTIFY_SOURCE`) and `procd_set_param nice 19` are byte-identical to the previously reviewed revision. The only content changes are in messages:

- `c70513c` now documents the `-U_FORTIFY_SOURCE` removal ("it was only required to avoid its warning at -O0; the OpenWrt toolchain fortification remains enabled at -O2"). That resolves my previous Commit checks note — dropping the `## Commit checks` section this round.
- The embedded patch header gained the pinned-`3rdparty/jitterentropy-rngd`-is-3.x dependency note and `Upstream-Status: Submitted [openwrt/urngd#1]`, addressing @namiltd's request.
- `4a9585b` gained only a blank line before `Signed-off-by`.

Spot checks: the patch header diffstat (`CMakeLists.txt | 4 ++--`, 2 insertions / 2 deletions) still matches its single hunk. Against procd `60fdbf0`, `nice` is a real instance attribute — [`INSTANCE_ATTR_NICE` parsed and range-checked to -20..20 at instance.c:1349-1350](https://github.com/openwrt/procd/blob/60fdbf00e924fff0a510c4ec31c520d35e69bfe1/service/instance.c#L1349-L1350), applied via [`setpriority()` at instance.c:479-480](https://github.com/openwrt/procd/blob/60fdbf00e924fff0a510c4ec31c520d35e69bfe1/service/instance.c#L479-L480) — so `19` is valid and takes effect on the daemon itself. All 13 check runs are green on `4a9585b`, including the three FormalityCheck jobs.

Two inline notes, both cosmetic nits.

---

package/system/urngd/files/urngd.init
dormancygrace: Agreed. Fixed in 4ff9ebd5f8: this commit now bumps PKG_RELEASE from 2 to 3 independently, so a standalone backport upgrades the package correctly.

dormancygrace: Fixed in 4ff9ebd5f8: replaced the bare Fixes trailer with Closes: https://github.com/openwrt/openwrt/issues/21573.

openwrt-ai (commented):

Delta since `4a9585b` is confined to the second commit, which was rewritten. `c70513c` is unchanged (same SHA), so the CMake patch and its message carry over as previously reviewed.

The only code change is `PKG_RELEASE` `2` → `3` in `package/system/urngd/Makefile`, which resolves my earlier nit — the bumps now chain cleanly from the base (`1` on `2991205`) to `2` in `c70513c` (CMake patch) and `3` in `4ff9ebd` (init script), so either commit is self-sufficient if backported alone. The message documents the reason. The other nit is resolved too: the trailer is now `Closes: https://github.com/openwrt/openwrt/issues/21573`.

`4ff9ebd`'s message matches its diff (`Makefile` release bump + `procd_set_param nice 19`), so no Commit checks section. All 13 check runs are green on the head SHA, including the three FormalityCheck jobs.

---

PR #23981, opened 2026-06-28
mediatek: comfast cf-wr632ax: add all-in-UBI layout

andros-ua:
```
This commit introduces OpenWrt U-Boot UBI layout support
for the COMFAST CF-WR632AX, enabling:
- Prolonged device lifetime by allocating most of the flash
  to UBI (which takes care of wear-leveling)
- Maximum available storage space for OpenWrt
- Fully-featured U-Boot
- Effective recovery mechanisms

OpenWrt U-Boot UBI flash instructions
-------------------------------------

A device running stock firmware should be upgraded to the
latest standard OpenWrt firmware
(https://firmware-selector.openwrt.org/?target=mediatek%2Ffilogic&id=comfast_cf-wr632ax).

Back up critical data
---------------------

Console:
1. While the device is running OpenWrt with stock MTD partitions:
   mtd dump BL2 > /tmp/BL2.bin
   mtd dump u-boot-env > /tmp/u-boot-env.bin
   mtd dump Factory > /tmp/Factory.bin
   mtd dump FIP > /tmp/FIP.bin
2. Copy the backup files to your PC via SCP.

LuCI Web-UI:
"System" -> "Backup / Flash Firmware" -> "Save mtdblock contents"
Save mtdblock:
   BL2
   u-boot-env
   Factory
   FIP
Make sure the files were successfully downloaded to your downloads directory,
especially the Factory file.

Using the installer image
-------------------------
To simplify the installation process, this method uses a fork
of Daniel Golle's (@dangowrt) UBI Installer
https://github.com/dangowrt/owrt-ubi-installer

1. Ensure your router is running the latest generic OpenWrt firmware.
   Upgrade it if necessary.
2. Obtain the installer image:
   Build the installer from source
   https://github.com/andros-ua/owrt-ubi-installer/tree/cf-wr632ax
   or download a prebuilt image from the
   https://github.com/andros-ua/owrt-ubi-installer/releases
3. Flash the openwrt*-ubi-initramfs-recovery-installer.itb
   image using sysupgrade.
4. Wait for installation: the green status LED will blink rapidly,
   indicating that the all-in-UBI installer is running.
5. Once the installation finishes,
   the status LED will turn solid amber for 5 seconds.
6. After the device reboots, perform a final sysupgrade using the
   openwrt*-ubi-squashfs-sysupgrade.itb image.

Return to stock MTD
-------------------
1. Flash openwrt*comfast_cf-wr632ax-initramfs-kernel.bin
   via sysupgrade
[...]

PR #22379, opened 2026-03-11
mediatek: filogic: Add support for Comfast CF-WA933

grischard:
This is my first contribution to OpenWRT. I will inevitably get things wrong, and would appreciate gentle feedback.

The [Comfast CF-WA933](https://comfastgroup.com/product/cf-wa933/) is a big outdoor access point. They also OEM their routers for other providers. This PR backports the [patches](https://github.com/Wayru-Network/wayru-os/tree/main/profiles/prometheus) from one such white label client into ~~25.12~~ main.

The wifi works well once you set your country code, and the sysupgrade image flashes nicely in the boot loader.

Pictures of the inside of one of these:

7EC28E2A-A7D5-4143-BA5D-C81E3A457E8D_1_102_a
D723EDA3-6F8F-4ABB-977F-230D2F1BE6F1_1_105_c

PR #18689, opened 2025-05-03
mediatek: add support for ipTIME AX3000SM

NyankoTear:
This is a pull request in order to add support for ipTIME AX3000SM. The architecture is similar to Routerich AX3000 v1, so it is possible to run the Routerich AX3000 v1 kernel with minor issues(MAC addressing, LED indication, and LAN 4 is missing). This commit is based on pull request #18601, with some customizations to suit the ipTIME router.

Specification
-------------
- SoC       : MediaTek MT7981BA dual-core ARM Cortex-A53 1.3GHz
- RAM       : DDR3 256Mbytes, Nanya Technology NT5CC128M16IP
- Flash     : 128Mbytes NAND Flash, ESMT F50L1G41LB
- WLAN      : MediaTek MT7976CN dual-band Wi-Fi 6
  - 2.4GHz  : b/g/n/ax, MU-MIMO
  - 5GHz    : a/n/ac/ax, MU-MIMO
- Ethernet  : 
  - 10/100/1000 Mbps x4, LAN (MediaTek MT7531AE)
  - 10/100/1000 Mbps x1, WAN (MT7981 internal PHY)
- UART      : 1x4 pin header on PCB
  - [J500] 3.3V, TX, RX, GND (115200, 8N1)
- Buttons   : WPS, Reset
- LEDs      : 9 LEDs
  - 1x Power (Amber) 
  - 1x CPU (Amber)
  - 1x Wi-Fi 5GHz (Amber)
  - 1x Wi-Fi 2.4GHz (Amber)
  - 1x WAN activity (Amber)
  - 4x LAN activity (Amber)
- Power     : 12VDC, 1A (Center positive polarity)

MAC address
-----------

| Interface      | MAC                        | Algorithm |
|-------------------|---------------------------|----------------|
| WLAN 2.4G  | B0:38:6C:xx:xx:xx | label           |
| WLAN 5G     | B2:38:6C:4x:xx:xx |                     |
| WAN             | B0:38:6C:xx:xx:xx | label + 1     |
| LAN              | B0:38:6C:xx:xx:xx | label + 3     |

The WLAN 2.4G MAC address was found in 'Factory' partition, 0x4

Installation
------------
1. Download the *initramfs-kernel.bin file from the OpenWrt website
2. Attach UART to the router, and interrupt the boot process by pressing '0'
> If you successfully interrupt the boot process, a terminal prompt name should look like this:
```
   MT7981>
```
3. Connect the router(LAN port) to the PC
[...]

PR #24590, opened 2026-08-06
mediatek: comfast cf-wa933: replace 128M variant with all-in-UBI

grischard:
This PR forwards work by @andros-ua, who saw #22379 and contacted me to improve my original contribution. I tested the resulting series on my COMFAST CF-WA933. It is also similar to #23981.

The series first tightens the SPI NAND drive current strength to prevent [[overshooting](https://github.com/openwrt/openwrt/pull/18689#discussion_r2117701961)].

It then fixes the 5 GHz radio MAC address assignment so the device uses the expected, correct factory address.

Next, it names the dedicated GMAC interface `wan` and updates the default network and WAN LED configuration accordingly. This does not require a compatibility-version bump because the device has not shipped in a stable OpenWrt release.

The next commit cleans up the existing CF-WA933 definitions: it removes redundant memory, bootargs, crypto, and image settings, removes the misleading `64M` display suffix, and adds `cf-wa933` as a compatibility alias while keeping the existing targets and partition layouts intact.

Finally, the last commit replaces the expanded fixed-partition 128 MB variant with a dedicated all-in-UBI target. It adds a corresponding OpenWrt U-Boot configuration and environment, with UBI-backed production and recovery images, redundant environment volumes, and `rootfs_data`.

Flash partitioning and factory-data definitions are moved from the shared device-tree include into the variant-specific device trees. The image definitions, upgrade handling, board detection, and other runtime configuration are updated for the new UBI variant.

Converting an existing device to the all-in-UBI layout requires the installer described in the final commit; the sysupgrade image alone cannot convert the stock partition layout while preserving factory data.


andros-ua (approved):

package/boot/uboot-mediatek/patches/476-add-comfast_cf-wa933-ubi.patch
```diff
@@ -0,0 +1,321 @@
...
++serverip=192.168.1.254
++loadaddr=0x46000000
++console=earlycon=uart8250,mmio32,0x11002000 console=ttyS0
++bootcmd=if pstore check ; then run boot_recovery ; else run boot_ubi ; fi
++bootconf=config-1
++bootdelay=0
++bootfile=openwrt-mediatek-filogic-comfast-cf-wa933-ubi-initramfs-recovery.itb
++bootfile_bl2=openwrt-mediatek-filogic-comfast-cf-wa933-ubi-preloader.bin
++bootfile_fip=openwrt-mediatek-filogic-comfast-cf-wa933-ubi-bl31-uboot.fip
++bootfile_upg=openwrt-mediatek-filogic-comfast-cf-wa933-ubi-squashfs-sysupgrade.itb
```
openwrt-ai: These four TFTP filenames don't match the artifacts this PR builds. OpenWrt names images `openwrt-<target>-<subtarget>-$(DEVICE_NAME)-...`, and `DEVICE_NAME` here is `comfast_cf-wa933-ubi` (underscore between vendor and model), so the build produces e.g. `openwrt-mediatek-filogic-comfast_cf-wa933-ubi-initramfs-recovery.itb`. With `comfast-cf-wa933-ubi` every TFTP path in the boot menu (recovery boot, `boot_tftp_production`, FIP/BL2 write) will 404 unless the user renames the files by hand. Compare the sibling board's env at [`448-add-comfast_cf-wr632ax-ubi.patch:263-266`](``https://github.com/openwrt/openwrt/blob/44933d7c71f4e7dfbef75cbb1ddde805cc57d8f4/package/boot/uboot-mediatek/patches/448-add-comfast_cf-wr632ax-ubi.patch#L263-L266).``

```suggestion
+bootfile=openwrt-mediatek-filogic-comfast_cf-wa933-ubi-initramfs-recovery.itb
+bootfile_bl2=openwrt-mediatek-filogic-comfast_cf-wa933-ubi-preloader.bin
+bootfile_fip=openwrt-mediatek-filogic-comfast_cf-wa933-ubi-bl31-uboot.fip
+bootfile_upg=openwrt-mediatek-filogic-comfast_cf-wa933-ubi-squashfs-sysupgrade.itb
```

---

package/boot/uboot-mediatek/patches/476-add-comfast_cf-wa933-ubi.patch
```diff
@@ -0,0 +1,321 @@
...
+@@ -0,0 +1,136 @@
++// SPDX-License-Identifier: GPL-2.0-or-later
++
++/dts-v1/;
++#include "mt7981.dtsi"
++#include <dt-bindings/gpio/gpio.h>
++#include <dt-bindings/input/linux-event-codes.h>
++
++/ {
++	model = "Comfast CF-WA933 (UBI)";
```
openwrt-ai: nit: vendor casing differs from every other place this device is named — `NAME:=COMFAST CF-WA933 (UBI)` in `package/boot/uboot-mediatek/Makefile`, `DEVICE_VENDOR := COMFAST` in `filogic.mk`, and `model = "COMFAST CF-WA933 (UBI)"` in `mt7981a-comfast-cf-wa933-ubi.dts:6`. The sibling U-Boot DTS uses the uppercase form too ([`448-add-comfast_cf-wr632ax-ubi.patch:12`](``https://github.com/openwrt/openwrt/blob/44933d7c71f4e7dfbef75cbb1ddde805cc57d8f4/package/boot/uboot-mediatek/patches/448-add-comfast_cf-wr632ax-ubi.patch#L12)).``

```suggestion
+	model = "COMFAST CF-WA933 (UBI)";
```

---

target/linux/mediatek/dts/mt7981a-comfast-cf-wa933-common.dtsi
```diff
@@ -259,7 +200,7 @@
 
 	band@1 {
 		reg = <1>;
-		nvmem-cells = <&macaddr_factory_8000 1>;
+		nvmem-cells = <&macaddr_factory_8000 0>;
```
openwrt-ai, line 203: This changes the 5 GHz MAC of the *already supported* 64M device: the cell index goes from `1` (base+1) to `0` (base), so existing installs will come back with a different `phy1` address after an ordinary sysupgrade. It lives in the shared `.dtsi`, so it is not confined to the new UBI variant.

It matches the sibling board ([`mt7981b-comfast-cf-wr632ax-common.dtsi:250`](``https://github.com/openwrt/openwrt/blob/44933d7c71f4e7dfbef75cbb1ddde805cc57d8f4/target/linux/mediatek/dts/mt7981b-comfast-cf-wr632ax-common.dtsi#L250)),`` so it may well be the correct value, but neither the commit message nor the PR body mentions it. Is this an intentional fix of the OEM MAC layout? If so it deserves a line in the commit message (and arguably its own commit, since it is a user-visible change to an existing device).

---

target/linux/mediatek/dts/mt7981a-comfast-cf-wa933-common.dtsi
```diff
@@ -98,55 +91,25 @@
...
-					pause;
-				};
-			};
-		};
 	};
 };
 
-&crypto {
-	status = "okay";
-};
```
openwrt-ai, line 134: Dropping `&crypto { status = "okay"; }` disables the EIP-97 inline crypto engine for both variants, including the existing 64M device — that is a silent performance regression for anything using kernel crypto (WireGuard/IPsec offload). ~40 other mediatek DTS files still enable it, and nothing else in this diff replaces it.

If the removal is deliberate (e.g. the engine was never functional on this SoC revision), please say so in the commit message; otherwise keep the node.

---

target/linux/mediatek/dts/mt7981a-comfast-cf-wa933-common.dtsi
```diff
@@ -12,16 +12,10 @@
...
-		bootargs-override = "console=ttyS0,115200n8";
+	chosen: chosen {
 		stdout-path = "serial0:115200n8";
 	};
 
-	memory@40000000 {
-		reg = <0 0x40000000>, <0 0x10000000>; // 256MB
-		device_type = "memory";
-	};
-
```
openwrt-ai, line 24: Removing `memory@40000000` and `bootargs-override` here affects the existing 64M device as well as the new UBI one, and only the UBI one gets a replacement (`bootargs` in `mt7981a-comfast-cf-wa933-ubi.dts`). `mt7981b.dtsi` has no `memory` node ([kernel 6.18, `arch/arm64/boot/dts/mediatek/mt7981b.dtsi`](https://github.com/gregkh/linux/blob/v6.18.41/arch/arm64/boot/dts/mediatek/mt7981b.dtsi)), so after this change the 64M variant relies entirely on the bootloader fixing up both the memory node and the console bootargs.

That is fine for the OpenWrt U-Boot the UBI variant ships, but the 64M variant still boots on the OEM MediaTek U-Boot, and this PR also drops `DEVICE_DTC_FLAGS := --pad 4096` (see my comment on `filogic.mk`) which is exactly the headroom that fixup needs. Has the 64M variant been re-tested on stock U-Boot after this reorganisation? If it has not, keeping `memory@40000000` + `bootargs-override` in the common `.dtsi` and only overriding them from the UBI `.dts` would be the safe form.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -1012,33 +1012,39 @@ define Device/comfast_cf-wa933-common
   DEVICE_VENDOR := COMFAST
   DEVICE_MODEL := CF-WA933
   DEVICE_DTS_DIR := ../dts
-  DEVICE_DTC_FLAGS := --pad 4096
-  DEVICE_DTS_LOADADDR := 0x43f00000
   DEVICE_PACKAGES := kmod-ledtrig-network kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware
-  KERNEL_LOADADDR := 0x44000000
```
openwrt-ai, line 1018: These three were removed from the *common* block, so the existing 64M device loses them too. Its FIT kernel load address silently moves from `0x44000000` to the filogic default `0x48000000` ([`target/linux/mediatek/image/Makefile:14,24`](https://github.com/openwrt/openwrt/blob/44933d7c71f4e7dfbef75cbb1ddde805cc57d8f4/target/linux/mediatek/image/Makefile#L14-L24)), the FDT loses its explicit `0x43f00000` placement, and `--pad 4096` — the space the OEM bootloader needs for FDT fixups — is gone.

The 64M variant is still flashed from and booted by the stock COMFAST U-Boot, which is presumably why the original submission pinned these values. Copying them from `Device/comfast_cf-wr632ax-common` (a different board with a different OEM loader) isn't automatically safe here. Unless the 64M variant has been retested end-to-end, please keep `DEVICE_DTC_FLAGS`/`DEVICE_DTS_LOADADDR`/`KERNEL_LOADADDR` on `Device/comfast_cf-wa933` and only leave them out of the UBI variant.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -1012,33 +1012,39 @@ define Device/comfast_cf-wa933-common
...
+  KERNEL_INITRAMFS := kernel-bin | lzma | \
+	fit lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb with-initrd | pad-to 64k
+  IMAGE/sysupgrade.itb := append-kernel | \
+	fit gzip $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb external-static-with-rootfs | append-metadata
+  ARTIFACTS := preloader.bin bl31-uboot.fip
+  ARTIFACT/preloader.bin := mt7981-bl2 spim-nand-ubi-ddr3-1866
+  ARTIFACT/bl31-uboot.fip := mt7981-bl31-uboot comfast_cf-wa933-ubi
   $(call Device/comfast_cf-wa933-common)
 endef
-TARGET_DEVICES += comfast_cf-wa933-128m
```
openwrt-ai, line 1041: Removing `comfast_cf-wa933-128m` leaves existing 128M installs stranded: the board name changes from `comfast,cf-wa933-128m` to `comfast,cf-wa933-ubi`, so sysupgrade will refuse the new image, and reaching the UBI layout additionally requires writing a new BL2 + FIP and creating the UBI volumes. There is no `DEVICE_COMPAT_VERSION`/`DEVICE_COMPAT_MESSAGE` and no migration note anywhere in the diff.

Two things worth adding: a `DEVICE_COMPAT_VERSION` bump with a `DEVICE_COMPAT_MESSAGE` pointing at the manual procedure, and the actual procedure written out in the commit message (see the body of this review).

---

target/linux/mediatek/dts/mt7981a-comfast-cf-wa933-ubi.dts
```diff
@@ -0,0 +1,81 @@
...
+
+		partition@100000 {
+			label = "ubi";
+			reg = <0x100000 0x7f00000>;
+			compatible = "linux,ubi";
+
+			volumes {
+				ubi_factory: ubi-volume-factory {
+					volname = "factory";
+				};
```
openwrt-ai, line 29: Every nvmem consumer in this variant hangs off this volume — `eeprom_factory_0` for the MT7981 WiFi calibration data and all three `macaddr_factory_*` cells — but nothing in the diff creates or populates it. The shipped U-Boot env only creates `ubootenv`/`ubootenv2` (`ubi_create_env`), `fit`/`recovery` (`ubi_write_production`/`ubi_write_recovery`) and `rootfs_data` (`ubi_prepare_rootfs`); there is no `ubi create factory` and no step that copies the OEM `Factory` MTD partition (`0x180000`, `0x200000` bytes in the 64M layout) into it.

So installing this variant as-is gives a device with no WiFi calibration and random MACs unless the user dumps and restores `Factory` by hand. How is the `factory` volume meant to be created? That step needs to be in the flash instructions in the commit message.

---

openwrt-ai (commented):
## Commit checks

- `44933d7` "mediatek: comfast cf-wa933: replace 128M variant with all-in-UBI" — sign-off does not cover the author. The commit is authored by `Andrii Kuiukoff <andros.ua@gmail.com>` but the only trailer is `Signed-off-by: Guillaume RISCHARD <git@stereo.lu>`. Under the DCO the author needs their own `Signed-off-by:`; the person forwarding the patch adds theirs below it, not instead of it.
- `44933d7` — the message only describes replacing the 128M variant, moving partition/factory nodes out of the common `.dtsi`, and adding the UBI build/env/LED/network bits. The diff additionally changes the *already supported* 64M device in ways nothing in the message mentions: 5 GHz nvmem cell index `1` → `0`, `&crypto` node removed, SPI0 `drive-strength` `MTK_DRIVE_8mA` → `MTK_DRIVE_4mA`, `memory@40000000` and `bootargs-override` removed, `DEVICE_DTC_FLAGS` / `DEVICE_DTS_LOADADDR` / `KERNEL_LOADADDR` removed, `DEVICE_VARIANT := 64M` dropped, `SUPPORTED_DEVICES += cf-wa933` added, and the `comfast,cf-wa933` uboot-envtools entry folded into another case. The drive-strength change is explained in the PR body only — the commit message is what ends up in the tree. Please document these (or split the 64M-affecting changes into their own commit).
- `44933d7` — required commit-message sections are missing. This commit adds a new `.dts` under `target/linux/mediatek/dts/` and a new `define Device/comfast_cf-wa933-ubi` block in `target/linux/mediatek/image/filogic.mk`, so per `.github/llm-review-rules.md` it needs **Hardware specification**, **Flash instructions** and **MAC address layout** sections. None are present. Flash instructions are the critical one here: this variant replaces BL2 and the FIP, needs a `factory` UBI volume that nothing in the diff creates, and cannot be reached from the removed 128M variant by sysupgrade — a stranger cannot install it from this commit message.


---

target/linux/mediatek/image/filogic.mk
grischard: There are two 128M installs in the world, and I control both of them. This is therefore not an issue.

openwrt-ai (commented):
Re-reviewed at `6a44414`. The force-push fixes both points I raised on the U-Boot patch — the four `bootfile*` names now use `comfast_cf-wa933-ubi` (matching `DEVICE_NAME`), and `model` is `COMFAST CF-WA933 (UBI)`. The rest of the new `defenvs/comfast_cf-wa933-ubi_env` and `configs/mt7981_comfast_cf-wa933-ubi_defconfig` are byte-identical to the `cf-wr632ax-ubi` sibling apart from the board name, LED names and DRAM size, and `476-` is a free patch number. Dropping the `From:`/`Date:`/`Subject:` git header matches the in-tree convention for `uboot-mediatek` add-board patches (nearly all of them start straight at `--- /dev/null`), so the FormalityCheck bot warning about it looks like a false positive here. All CI checks on `6a44414` are green.

My earlier inline comments on `mt7981a-comfast-cf-wa933-common.dtsi`, `mt7981a-comfast-cf-wa933-ubi.dts` and `filogic.mk` are still open and unchanged by this push; not repeating them.

## Commit checks

- `6a44414` "mediatek: comfast cf-wa933: replace 128M variant with all-in-UBI" — the rewritten message is a large improvement on the install-procedure side, but it now describes *only* the new UBI variant. The sentence the previous revision had ("Reorganize device tree structure by moving partition and factory data definitions from the common dtsi to variant-specific dts files") was dropped, so the message is now strictly less descriptive of what the diff does. The diff still makes several user-visible changes to the *already supported* 64M device that nothing in the message mentions: 5 GHz nvmem cell index `1` → `0`, `&crypto` node removed, SPI0 `drive-strength` `MTK_DRIVE_8mA` → `MTK_DRIVE_4mA`, `memory@40000000` and `bootargs-override` removed, `DEVICE_DTC_FLAGS` / `DEVICE_DTS_LOADADDR` / `KERNEL_LOADADDR` dropped, `DEVICE_VARIANT := 64M` dropped, `SUPPORTED_DEVICES += cf-wa933` added, and the WAN netdev renamed `eth1` → `wan` (`label = "wan"` on `gmac1`, plus the matching `02_network` / `01_leds` edits). The drive-strength change is explained in the PR body only, and the PR body does not end up in the tree. Please document these in the commit message, or split the 64M-affecting changes into their own commit.

- `6a44414` — per `.github/llm-review-rules.md`, this commit adds a new `.dts` under `target/linux/mediatek/dts/` plus a new `define Device/comfast_cf-wa933-ubi`, so a **MAC address layout** section is required and is still absent. It is worth more than boilerplate here: the factory data moves out of the `Factory` MTD partition into a `factory` UBI volume, and the 5 GHz cell index changes at the same time, so one line per interface (LAN / WAN / 2.4G / 5G → `factory` offset + `#nvmem-cell-cells` index, `xx:xx:xx` for the unit bytes) would document both at once. Hardware specification is fairly considered covered by the original `comfast_cf-wa933` commit, and the flash instructions are now well covered.

- `6a44414` — the amend dropped `Signed-off-by: Guillaume RISCHARD <git@stereo.lu>`. The commit being authored *and* signed off by `Andrii Kuiukoff <andros.ua@gmail.com>` fixes the DCO gap I flagged on `44933d7`, but the fix swapped the problem rather than closing it: as the person forwarding someone else's patch you should add your own `Signed-off-by:` *below* the author's, not in place of it. Both trailers should be present.

---

andros-ua:
@jonasjelonek, please review the enhanced version of the initial support commit (cdc44e906ace23664ab5a19156d5f8470dddbaf1). It's been run-tested and doesn't affect any existing users.

openwrt-ai (commented):
Re-reviewed at `4704bd1`. The code diff is byte-identical to `6a44414` — this push is a rebase onto current `main` plus a rewritten commit message and the restored second sign-off. So I have no new inline comments; my earlier threads on `mt7981a-comfast-cf-wa933-common.dtsi`, `mt7981a-comfast-cf-wa933-ubi.dts` and `filogic.mk` are unchanged and not repeated here. All 22 checks on `4704bd1` are green (the FormalityCheck warning about the missing `git am` header on `476-add-comfast_cf-wa933-ubi.patch` is a non-blocking stylistic warning, and matches how nearly every other `uboot-mediatek` add-board patch is written).

The rewritten message closes most of what I raised on `6a44414`:

- Both `Signed-off-by:` trailers are present (author `Andrii Kuiukoff`, forwarder `Guillaume RISCHARD`) — the DCO gap is closed.
- The **MAC address layout** section is present and matches the DTS (`macaddr_factory_e000` idx 0/1 for LAN/WAN, `macaddr_factory_4` idx 0, `macaddr_factory_8000` idx 0).
- The 5 GHz cell-index change (`1` → `0`) is now explained, which answers my thread on `mt7981a-comfast-cf-wa933-common.dtsi`.
- The dropped `memory@40000000` / `bootargs-override` / `DEVICE_DTC_FLAGS` / `DEVICE_DTS_LOADADDR` / `KERNEL_LOADADDR` are documented, with an explicit statement that the standard layout was tested booting without them — that resolves both of my threads on the 64M variant's stock-U-Boot boot path.
- The drive-strength reduction, the `gmac1` `label = "wan"` rename, and the meaning of the old `64M` variant name are all documented.

## Commit checks

- `4704bd1` "mediatek: comfast cf-wa933: replace 128M with all-in-UBI, & clean up" — two hunks that touch the already-supported 64M device are still not mentioned anywhere in the message:
  - Removal of `&crypto { status = "okay"; }` from the common `.dtsi`. It disables the EIP-97 engine for both variants. 39 other mediatek DTS files still enable it, including the other COMFAST mt7981a board ([`mt7981a-comfast-cf-e393ax.dts:102`](https://github.com/openwrt/openwrt/blob/4704bd1cb902614bf582ec7060ea3f8fc9e5ca54/target/linux/mediatek/dts/mt7981a-comfast-cf-e393ax.dts#L102)) — though the CF-WR632AX common `.dtsi` this layout was modelled on does *not* have it, so the removal may well be deliberate parity. Either way it deserves a line ("drop the crypto node, the engine is unused/non-functional here" or similar), since it is a silent throughput change for existing installs.
  - `SUPPORTED_DEVICES += cf-wa933` on `Device/comfast_cf-wa933`. This mirrors [`SUPPORTED_DEVICES += cf-wr632ax` at `filogic.mk:1064`](https://github.com/openwrt/openwrt/blob/4704bd1cb902614bf582ec7060ea3f8fc9e5ca54/target/linux/mediatek/image/filogic.mk#L1064) so it looks like intentional sibling parity, but it is an unexplained hunk. One clause is enough.

- `4704bd1` — nit: the subject drops the subtarget. Every other recent commit touching `target/linux/mediatek/` uses `mediatek: filogic: <topic>: <change>` (`mediatek: filogic: fix sysupgrade for Wavlink WL-WNT100X3`, `mediatek: filogic: SDG-8733A: point label-mac-device at wan`, …), and the commit that added this board is `mediatek: filogic: add support for COMFAST CF-WA933`. `mediatek: filogic: cf-wa933: replace 128M variant with all-in-UBI` would match. The trailing `, & clean up` also reads as an admission of mixed scope — now that the message documents the cleanups it is less of a problem, but splitting the 64M-affecting changes into a separate preceding commit would still make both easier to revert independently.

- `4704bd1` — still open from my first review and not addressed by the new message: nothing in the tree creates or populates the `factory` UBI volume that every nvmem consumer of the UBI variant depends on (see my thread on `mt7981a-comfast-cf-wa933-ubi.dts`). The flash instructions now cover backing up the OEM `Factory` MTD and point at the `andros-ua/owrt-ubi-installer` fork, so I assume the installer does the `ubi create factory` + copy — but that is an out-of-tree image, and the message never states it. One sentence in the flash instructions saying which step creates and fills `factory` would close this.

---

target/linux/mediatek/dts/mt7981a-comfast-cf-wa933-common.dtsi
grischard: This does not disable EIP-97. The inherited crypto@10320000 node in mt7981b.dtsi has no status = "disabled", so it is available by default; the board-level override was redundant. I verified this on two running CF-WA933s. Both successfully probe the same hardware and register the same crypto implementations.

- stock/128M DT with the stanza: effective status is okay;
- UBI DT without the stanza: the status property is absent;

grischard:
Maintainers, what can we do to make this easier to merge? Would splitting into multiple PRs make it better or worse?

grischard:
Just rebased, but I'm going to split this into multiple smaller PRs to hopefully make it easier to review

package/boot/uboot-mediatek/patches/477-add-comfast_cf-wa933-ubi.patch
openwrt-ai, line 1: nit: `476-` is no longer free — the rebase brought in [`476-add-mercusys_mr85x.patch`](https://github.com/openwrt/openwrt/blob/dd45cbe981d9a6898fe9c809c4f786f37ac99dc7/package/boot/uboot-mediatek/patches/476-add-mercusys_mr85x.patch), so the head tree now has two `476-` patches. Renaming this one to `477-add-comfast_cf-wa933-ubi.patch` avoids the collision. Purely cosmetic — the tree already carries `472-add-bpi-r4-pro-8x.patch` alongside `472-add-globitel-bt-r320.patch` and both apply fine.

---

openwrt-ai (commented):
Reviewed 1 new commit.

---

package/boot/uboot-mediatek/patches/476-add-comfast_cf-wa933-ubi.patch
openwrt-ai: fixed, thanks

---

package/boot/uboot-mediatek/patches/476-add-comfast_cf-wa933-ubi.patch
openwrt-ai: fixed, thanks

---

target/linux/mediatek/dts/mt7981a-comfast-cf-wa933-common.dtsi
openwrt-ai: fixed, thanks — solved by documenting the change in the commit message rather than by splitting it out.

---

target/linux/mediatek/dts/mt7981a-comfast-cf-wa933-common.dtsi
openwrt-ai: fixed, thanks — answered by the commit message stating the standard layout was tested booting with these removed.

---

target/linux/mediatek/image/filogic.mk
openwrt-ai: fixed, thanks — answered by the commit message stating the standard layout was tested booting without these.

---

jonasjelonek:
Not needed as separate PRs, but split into some commits would be good. If I see correctly, this also does some cleanup and slightly related stuff. Similar to #24575 maybe.

target/linux/mediatek/filogic/base-files/etc/board.d/05_compat-version
andros-ua, line 1: I think using `05_compat-version` can be avoided here, since the device hasn't shipped in any stable builds yet.

package/boot/uboot-mediatek/patches/477-add-comfast_cf-wa933-ubi.patch
grischard: I picked 477

grischard:
@jonasjelonek yes, this is very similar to #24575. I've now split it into five commits, as requested.

target/linux/mediatek/dts/mt7981a-comfast-cf-wa933.dts
```diff
@@ -7,6 +7,69 @@
...
+
+		partition@380000 {
+			label = "FIP";
+			reg = <0x380000 0x0200000>;
+			read-only;
+		};
+	};
+};
+
 &partitions {
```
openwrt-ai, line 73: The `partitions` node is now declared 58 lines above in this same file, so overriding it by phandle here is redundant. Move `partition@580000` into that node and drop the `partitions:` label, which no longer has any out-of-file user.

---

target/linux/mediatek/dts/mt7981a-comfast-cf-wa933-ubi.dts
```diff
@@ -0,0 +1,87 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+#include "mt7981a-comfast-cf-wa933-common.dtsi"
+
+/ {
+	model = "COMFAST CF-WA933 (UBI)";
+	compatible = "comfast,cf-wa933-ubi", "mediatek,mt7981";
+};
+
+&{/chosen} {
```
openwrt-ai, line 10: nit: the sibling this layout mirrors labels the node in its common `.dtsi` and overrides it as [`&chosen` in `mt7981b-comfast-cf-wr632ax-ubi.dts:10`](https://github.com/openwrt/openwrt/blob/967111645d6c8f21c828fa9be016b6d6280e1856/target/linux/mediatek/dts/mt7981b-comfast-cf-wr632ax-ubi.dts#L10); in this directory `&{/chosen}` is otherwise used only by `.dtso` overlays. Adding `chosen: chosen {` in `mt7981a-comfast-cf-wa933-common.dtsi` and using `&chosen` here would match.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -1099,30 +1099,39 @@ define Device/comfast_cf-wa933-common
   DEVICE_DTC_FLAGS := --pad 4096
   DEVICE_DTS_LOADADDR := 0x43f00000
```
openwrt-ai, line 1100: These stay in the common block, so the new UBI variant inherits them: its FIT gets the FDT pinned at `0x43f00000` (via `mkits.sh -s`) while the kernel uses the filogic default load address. [`Device/comfast_cf-wr632ax-common`](https://github.com/openwrt/openwrt/blob/967111645d6c8f21c828fa9be016b6d6280e1856/target/linux/mediatek/image/filogic.mk#L1136-L1145), which `comfast_cf-wa933-ubi` otherwise mirrors, sets neither. Intentional, or should these move down into `Device/comfast_cf-wa933`?

---

openwrt-ai (commented):
## Commit checks

- `85d326a` "name dedicated WAN interface" — the message says it updates the network config for "both existing CF-WA933 variants", but the `02_network` hunk drops `comfast,cf-wa933-128m` (still a live target at this commit) and adds `comfast,cf-wa933-ubi` (which only exists from `9671116`). There is no default case in that `case`, so between this commit and `9671116` the 128M board gets no LAN/WAN config at all — swap those two board names so each commit stands alone.
- `886380f` "clean up legacy device definitions" — "Drop the DTB padding and explicit kernel/FDT load addresses" doesn't match the diff: only `KERNEL_LOADADDR` was removed, `DEVICE_DTC_FLAGS := --pad 4096` and `DEVICE_DTS_LOADADDR := 0x43f00000` are still in `Device/comfast_cf-wa933-common`.
- `9671116` — "Recognise the new compatible in ... runtime compatibility-version initialisation" describes a `05_compat-version` change that is not in the diff (it was dropped). Drop that clause.
- `9671116` — nit: stray double backtick in ``root=/dev/fit0 rootwait`` `` in the body.

---

target/linux/mediatek/dts/mt7981a-comfast-cf-wa933-ubi.dts
openwrt-ai: fixed, thanks — the flash instructions now state that the installer creates and populates the `factory` volume.

---

package/boot/uboot-mediatek/Makefile
```diff
@@ -317,6 +317,18 @@ define U-Boot/mt7981_cmcc_rax3000m-ubi-ddr3
   DEPENDS:=+trusted-firmware-a-mt7981-spim-nand-ubi-ddr3-1866
 endef
 
+define U-Boot/mt7981_comfast_cf-wa933-ubi
```
openwrt-ai: nit: this block lands between `U-Boot/mt7981_cmcc_rax3000m-ubi-ddr3` and `U-Boot/mt7981_cmcc_rax3000m-ubi-ddr4`, splitting that pair and breaking the file's alphabetical ordering (the `UBOOT_TARGETS` entry itself is correctly placed). Move it after the `endef` of `mt7981_cmcc_rax3000m-ubi-ddr4`, just before `define U-Boot/mt7981_comfast_cf-wr632ax`.

---

openwrt-ai (commented):
Reviewed 3 new commits.

---

openwrt-ai (commented):

---

grischard:
Just pushed a tiny additional cleanup. Still ready for review.

package/boot/uboot-mediatek/Makefile
openwrt-ai, line 320: This block again lands between `U-Boot/mt7981_cmcc_rax3000m-ubi-ddr3` and `U-Boot/mt7981_cmcc_rax3000m-ubi-ddr4`, splitting that pair and breaking the file's alphabetical order; this push moved the unrelated `ddr4` block down rather than moving this one. Put it after the `endef` of `mt7981_cmcc_rax3000m-ubi-ddr4`, just before `define U-Boot/mt7981_comfast_cf-wr632ax`, and leave the `cmcc` block untouched.

---

target/linux/mediatek/dts/mt7981a-comfast-cf-wa933.dts
openwrt-ai, line 73: Now that the `partitions` node is declared 58 lines above in this same file, overriding it by phandle here is redundant again. Fold `partition@580000` into that node and drop the `partitions:` label on line 15, which no longer has an out-of-file user.

---

package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic
```diff
@@ -85,6 +86,7 @@ nradio,c8-668gl)
 asiarf,ap7986-003|\
 cetron,ct3003|\
 comfast,cf-wr632ax|\
+comfast,cf-wa933|\
```
openwrt-ai, line 89: nit: this case list is alphabetically sorted, so `comfast,cf-wa933` belongs before `comfast,cf-wr632ax`.

```suggestion
comfast,cf-wa933|\
comfast,cf-wr632ax|\
```

---

openwrt-ai (commented):
## Commit checks

- `85d326a` "name dedicated WAN interface" — the message says it updates the network config for both existing variants, but the `02_network` hunk drops `comfast,cf-wa933-128m` (still a live target at this commit) and adds `comfast,cf-wa933-ubi` (only created by `fedaaa6`); with no default case the 128M board gets no LAN/WAN config between the two commits. Use `-128m` here and switch it to `-ubi` in `fedaaa6`.
- `ec6a50f` "clean up legacy device definitions" — "Drop the DTB padding and explicit kernel/FDT load addresses" no longer matches the diff: only `KERNEL_LOADADDR` is removed, `DEVICE_DTC_FLAGS := --pad 4096` and `DEVICE_DTS_LOADADDR := 0x43f00000` are back in `Device/comfast_cf-wa933-common`.
- `fedaaa6` — "Recognise the new compatible in ... runtime compatibility-version initialisation" still describes a `05_compat-version` change that is not in the diff; drop that clause.

---

PR #24595, opened 2026-08-07
unetmsg: fix TCP_USER_TIMEOUT socket option level

cmonroe:
The ucode socket module exports no SOL_TCP so setopt() was called with level 0 and silently failed. This left inter-node connections without the 5 second user timeout. A peer that stopped reading could then stall a blocking channel write forever, deadlocking every unetmsgd in the mesh along with their local clients.

Use IPPROTO_TCP and warn when setopt fails.

### DIFF
--- a/package/network/services/unetmsg/files/usr/share/ucode/unetmsg/unetmsgd-remote.uc
+++ b/package/network/services/unetmsg/files/usr/share/ucode/unetmsg/unetmsgd-remote.uc
@@ -301,7 +301,8 @@ function network_open_channel(net, name, peer)
 		return;
 
 	core.dbg(`Try to connect to ${name}\n`);
-	sock.setopt(socket.SOL_TCP, socket.TCP_USER_TIMEOUT, TCP_TIMEOUT);
+	if (!sock.setopt(socket.IPPROTO_TCP, socket.TCP_USER_TIMEOUT, TCP_TIMEOUT))
+		warn(`Failed to set TCP user timeout on channel socket: ${socket.error()}\n`);
 	sock.connect(addr);
 	let auth_data_cb = (msg) => {
 		if (!network_auth_valid(sock_data.name, sock_data.id, msg.token))
@@ -425,7 +426,8 @@ function network_open(name, info)
 	net.rx_channels = {};
 	net.tx_channels = {};
 
-	net.socket.setopt(socket.SOL_TCP, socket.TCP_USER_TIMEOUT, TCP_TIMEOUT);
+	if (!net.socket.setopt(socket.IPPROTO_TCP, socket.TCP_USER_TIMEOUT, TCP_TIMEOUT))
+		warn(`Failed to set TCP user timeout on listen socket: ${socket.error()}\n`);
 
 	let cb = () => {
 		let addr = {};
--- a/package/network/services/unetmsg/Makefile
+++ b/package/network/services/unetmsg/Makefile
@@ -8,7 +8,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=unetmsg
-PKG_RELEASE:=14
+PKG_RELEASE:=15
 
 PKG_LICENSE:=GPL-2.0
 PKG_MAINTAINER:=Felix Fietkau <nbd@nbd.name>


package/network/services/unetmsg/files/usr/share/ucode/unetmsg/unetmsgd-remote.uc
```diff
@@ -301,7 +301,8 @@ function network_open_channel(net, name, peer)
 		return;
 
 	core.dbg(`Try to connect to ${name}\n`);
-	sock.setopt(socket.SOL_TCP, socket.TCP_USER_TIMEOUT, TCP_TIMEOUT);
+	if (!sock.setopt(socket.IPPROTO_TCP, socket.TCP_USER_TIMEOUT, TCP_TIMEOUT))
+		warn(`Failed to set TCP user timeout: ${socket.error()}\n`);
```
openwrt-ai: nit: this warning string is byte-identical to the one added at line 430 (`network_open()`, on the listening socket), so a syslog line gives no hint which of the two call sites failed. The rest of the file distinguishes its messages by context (`Failed to open socket`, `Failed to generate auth reply token`, `Outgoing connection to ${name} established`). Naming the socket makes the two distinguishable:

```suggestion
		warn(`Failed to set TCP user timeout on channel socket: ${socket.error()}\n`);
```

and correspondingly something like `... on listen socket: ...` at line 430.

---

openwrt-ai (commented):
Reviewed 1 new commit; the fix itself checks out against the pinned ucode.

Confirmed the premise: `lib/socket.c` in ucode exports `IPPROTO_TCP` and `TCP_USER_TIMEOUT` but no `SOL_TCP` ([`ADD_CONST(SOL_SOCKET)` / `ADD_CONST(SOL_PACKET)` are the only `SOL_*` constants](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib/socket.c#L5256)), and the sockopt table keys `TCP_USER_TIMEOUT` on [`IPPROTO_TCP`](https://github.com/jow-/ucode/blob/81205a25437a83e3b9e1eb55ac169be84453d53b/lib/socket.c#L1719). A `null` `level` argument is rejected by `args_get_()` with `"Argument level is required"`, which returns `null` rather than raising — so the old call did fail silently, as the commit message says.

Also checked the second call site: setting `TCP_USER_TIMEOUT` on the listening socket is not a no-op — `icsk_user_timeout` is carried into accepted children by the whole-struct copy in `sk_clone_lock()` via [`inet_csk_clone_lock()`](https://github.com/gregkh/linux/blob/master/net/ipv4/inet_connection_sock.c), and nothing in the TCP child-creation path resets it (compared against v6.12.100). So accepted connections do pick up the 5s timeout.

Only one nit inline; no blocking issues found. CI on `a7e738c` is green/queued with no failures.

---

openwrt-ai (commented):

The only change since `a7e738c` is the warning-string differentiation (`... on channel socket:` / `... on listen socket:`), which resolves the nit from the previous review. Commit message is unchanged and still matches the diff, and `PKG_RELEASE` is bumped.

Re-confirmed the constants against the pinned ucode tree (`73ffef7`): [`ADD_CONST(IPPROTO_TCP)`](https://github.com/jow-/ucode/blob/73ffef75c653dc901c69678a27c67842d291e79e/lib/socket.c#L5341) and [`ADD_CONST(TCP_USER_TIMEOUT)`](https://github.com/jow-/ucode/blob/73ffef75c653dc901c69678a27c67842d291e79e/lib/socket.c#L5357) are exported, with no `SOL_TCP`. The `warn(...: ${socket.error()})` shape matches the existing [`Failed to open socket` handler at unetmsgd-remote.uc:420](``https://github.com/openwrt/openwrt/blob/7675e973d5675e162e6e5e5dc01454b56be6b064/package/network/services/unetmsg/files/usr/share/ucode/unetmsg/unetmsgd-remote.uc#L420).``

CI on `7675e97` is fully green (FormalityCheck, malta/be and x86/64 package builds).

---

danpawlik (approved):

PR #24601, opened 2026-08-07
mac80211: ath11k: fix warm-reboot hang on IPQ AHB SoCs

ddimension:
## Problem

On qualcommax IPQ AHB targets (IPQ8074/IPQ6018/IPQ5018) a plain `reboot`
hangs in the boot ROM (SBL1) during DDR training and never comes back —
only a physical power cycle recovers. Confirmed on a MikroTik Chateau 5G
R17 ax (IPQ6018); the serial log freezes at `sbl1_ddr_set_params`.

## Root cause

The ath11k AHB `shutdown()` handler was introduced for WCN6750 (which must
stop DMA before the SMMU is torn down) but is registered unconditionally
in the shared `ath11k_ahb_driver`, so it also runs on the IPQ SoCs, which
never had a `shutdown()` handler before. On reboot it tears the WCSS down
and stops the Q6 firmware; that stop corrupts shared WCSS state such that
SBL1 hangs in DDR training on the following **warm** reset (only a cold /
power-on reset re-initializes it).

Regression from kernel commit `ac41c2b642b1` ("wifi: ath11k: Register
shutdown handler for WCN6750"). The unconditional `.shutdown` is not new
to the current backports bump — it is already present in v6.12 — so every
OpenWrt branch whose ath11k backports carry that commit is affected, not
just main.

## Fix

Restrict the `shutdown()` teardown to WCN6750. On the IPQ AHB SoCs the SoC
reset re-initializes the WCSS from scratch, so no teardown is needed at
reboot. Module removal (`rmmod` / `remove()`) is unchanged.

Bisected by disabling teardown steps individually at reboot: skipping the
Q6 firmware stop (`ath11k_qmi_firmware_stop()`) is what makes the warm
reset train DDR and boot; none of the q6v5 power/reset/clock steps
reproduce it on their own.

## Testing

MikroTik Chateau 5G R17 ax (IPQ6018), kernel 6.18, serial console:
- **before:** `reboot` freezes in SBL1 at `sbl1_ddr_set_params`.
- **after:** SBL1 reaches `SBL1, End`, trains DDR and boots the kernel;
  repeated `reboot` are clean.

## Upstream

Same change submitted to linux-wireless / ath11k (carries `Fixes:` +
`Cc: stable`). This backport can be dropped once the mac80211 backports
bump pulls the upstream commit.

## Note

An explicit `rmmod ath11k_ahb` (or ath11k firmware crash-recovery) still
runs the full teardown via `remove()` and hits the same WCSS-stop issue;
that path is pre-existing and unrelated to this regression. Normal
`wifi down/up` is unaffected (it does not unload the module).


ddimension:
Submitted upstream to linux-wireless/ath11k: https://lore.kernel.org/linux-wireless/20260807094723.1546780-1-avalentin@marcant.net/ (added as a `Link:` in the commit). This backport can be dropped once the mac80211 backports bump pulls the upstream commit.

package/kernel/mac80211/patches/ath11k/950-wifi-ath11k-run-ahb-shutdown-teardown-only-on-WCN6750.patch
```diff
@@ -0,0 +1,52 @@
...
++	 * the WCSS Q6 firmware, which leaves shared state that hangs the boot ROM
++	 * (SBL1) during DDR training on the following warm reset -- only a cold
++	 * (power-on) reset recovers. Skip it on those targets; the SoC reset
++	 * re-initializes the WCSS from scratch.
++	 */
++	if (ab->hw_rev != ATH11K_HW_WCN6750_HW10)
++		return;
++
+ 	ath11k_ahb_remove_prepare(ab);
+
```
openwrt-ai: nit: this trailing context line is completely empty — it lost the leading space that marks it as context in a unified diff. GNU `patch` tolerates a bare empty line as blank context so this still applies, but none of the other 20 patches in `package/kernel/mac80211/patches/ath11k/` have an in-hunk empty line; they all use `" "`-prefixed blank context lines (the added blank lines on lines 42 and 49 are correctly `+`-prefixed, so this looks like trailing-whitespace stripping on the way in rather than a deliberate choice). Regenerating the patch so the context lines keep their leading space would bring it in line with the rest of the directory — for `package/kernel/mac80211/patches/...` that's `make package/mac80211/refresh`.

---

openwrt-ai (commented):
Reviewed 1 new commit (`5a79763`). The change itself checks out:

- `ATH11K_HW_WCN6750_HW10` exists in `drivers/net/wireless/ath/ath11k/core.h` and `ab->hw_rev` is set from the `of_match_table` `.data` in `ahb.c`, so the guard is well-formed.
- Hunk header `@@ -1358,6 +1358,18 @@` matches the backports tree exactly: upstream v6.18.41 has the comment at `ahb.c:1291`, and the in-tree ath11k patches 910/923/924/925 add 6+34+24+3 = 67 lines above it (1291+67 = 1358). Line counts (6 context, 12 added) are consistent too. None of the other ath11k patches touch `ath11k_ahb_shutdown()` or `ath11k_ahb_remove_prepare()`, so no ordering conflict.
- `PKG_RELEASE` bump is present.

One formatting nit inline. Note also that `kmod-ath11k-ahb` is `@TARGET_qualcommax`-only in `package/kernel/mac80211/ath.mk`, so on OpenWrt builds this effectively makes `.shutdown` a no-op everywhere — which is exactly the pre-`ac41c2b642b1` behaviour the commit message describes, just worth being explicit about.

---

ddimension:
Force-pushed to update the approach per the upstream ath11k review.

v1 used a direct `ab->hw_rev != ATH11K_HW_WCN6750_HW10` check inside `ath11k_ahb_shutdown()`. Jeff Johnson (ath11k maintainer) asked that hardware-specific behaviour be driven via `struct ath11k_hw_params` rather than a hw_rev comparison. v2 therefore adds a `shutdown_teardown` flag, sets it only for WCN6750, and runs the teardown only when it is set — same effect, expressed through hw_params, and framed as restoring the pre-regression behaviour for the IPQ AHB targets.

This now matches the patch on the list (v2):
https://lore.kernel.org/linux-wireless/20260808213551.4169863-1-avalentin@marcant.net/

(mac80211 PKG_RELEASE bumped accordingly; CI green.)

package/kernel/mac80211/patches/ath11k/950-wifi-ath11k-run-ahb-shutdown-teardown-only-on-WCN6750.patch
```diff
@@ -0,0 +1,89 @@
...
+Signed-off-by: André Valentin <avalentin@marcant.net>
+---
+ drivers/net/wireless/ath/ath11k/hw.h | 4 ++++
+ drivers/net/wireless/ath/ath11k/core.c | 1 +
+ drivers/net/wireless/ath/ath11k/ahb.c | 11 +++++++++++
+ 3 files changed, 16 insertions(+)
+
+--- a/drivers/net/wireless/ath/ath11k/hw.h
++++ b/drivers/net/wireless/ath/ath11k/hw.h
+@@ -233,6 +233,10 @@
```
openwrt-ai: nit: all three hunk headers in this patch are bare `@@ -a,b +c,d @@` with no function context, which is what plain `diff -u` (without `-p`) emits. Quilt — which is what `Quilt/Refresh/Package` in `package/kernel/mac80211/Makefile` drives for `patches/ath11k/` — emits the enclosing-symbol suffix. Across the other 20 patches in `package/kernel/mac80211/patches/ath11k/` only 4 of 73 hunks are bare, and those are hunks with genuinely no enclosing symbol; e.g. [949-wifi-ath11k-fix-monitor-rx-pktlen.patch:25](``https://github.com/openwrt/openwrt/blob/a66d28d0029c906fdb394e31d5a4af5ad5e4a2a4/package/kernel/mac80211/patches/ath11k/949-wifi-ath11k-fix-monitor-rx-pktlen.patch#L25)`` carries `@@ ... @@ ath11k_dp_rx_mon_merg_msdus(struct ath11`.

I re-generated the three hunks against the backports tree with the other ath11k patches applied; the refreshed headers are:

- L49 → `@@ -233,6 +233,10 @@ struct ath11k_hw_params {`
- L62 → `@@ -567,6 +567,7 @@ static struct ath11k_hw_params ath11k_hw`
- L72 → `@@ -1358,6 +1358,17 @@ static void ath11k_ahb_shutdown(struct p`

Suggestion below covers this line; please apply the other two at lines 62 and 72 as listed, or just regenerate the patch with `make package/mac80211/refresh`. (Offsets and context all verified correct — the patch applies with zero fuzz as-is, this is purely about matching the directory's generated-by-quilt form.)

```suggestion
@@ -233,6 +233,10 @@ struct ath11k_hw_params {
```

---

package/kernel/mac80211/Makefile
```diff
@@ -11,7 +11,7 @@ include $(INCLUDE_DIR)/kernel.mk
 PKG_NAME:=mac80211
 
 PKG_VERSION:=6.18.39
-PKG_RELEASE:=3
+PKG_RELEASE:=5
```
openwrt-ai: nit: this skips a release number. The base branch has `PKG_RELEASE:=3` and this PR is a single commit on top of it, so the bump should land on `4`. The `5` looks like an artifact of the force-push (v1 bumped 3→4, v2 bumped again on top of the already-bumped value). Harmless, but it leaves a gap and `PKG_RELEASE` feeds `ABI_VERSION:=$(PKG_VERSION)-$(PKG_RELEASE)` for `kmod-cfg80211`/`kmod-mac80211`, so it is nicer kept monotonic-by-one.

```suggestion
PKG_RELEASE:=4
```

---

package/kernel/mac80211/Makefile
```diff
@@ -11,7 +11,7 @@ include $(INCLUDE_DIR)/kernel.mk
 PKG_NAME:=mac80211
 
 PKG_VERSION:=6.18.39
```
openwrt-ai, line 13: nit: the PR body says the regression is something "which OpenWrt picked up via the current mac80211 backports", which reads as if only main is affected. The unconditional `.shutdown = ath11k_ahb_shutdown` is considerably older than that — it is already present in v6.12: [`ath11k_ahb_shutdown()` at ahb.c:1297-1315](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/drivers/net/wireless/ath/ath11k/ahb.c#L1297-L1315) (gregkh/linux v6.12.100), with `.shutdown` wired up unconditionally at ahb.c:1324.

More concretely for this repo: `openwrt-25.12` pins the *same* `PKG_VERSION:=6.18.39` backports tarball as this line does, and has no equivalent patch under `patches/ath11k/`, so IPQ8074/IPQ6018/IPQ5018 devices on 25.12 hit the identical warm-reboot hang. Is a `[25.12]` backport of this patch planned once it lands on main? Worth stating either way so it does not get lost.

---

openwrt-ai (commented):
Reviewed the force-pushed v2 (`a66d28d`, replacing `5a79763`). The switch from a `hw_rev` comparison to a `hw_params` flag is done correctly and the previously-reported empty-context-line issue is fixed.

Verification I ran against the backports tree (extracted `drivers/net/wireless/ath/ath11k/` from Linux v6.18.41 and applied all 20 in-tree `patches/ath11k/*.patch` in order — backports 6.18.39 vs. v6.18.41 do not diverge in ath11k):

- All three hunks apply with **zero fuzz and zero offset**. `hw.h` `@@ -233` lands on the `support_fw_mac_sequence`/`support_dual_stations`/`pdev_suspend` bool block (`pdev_suspend` is at 235 post-patch-stack, matching 233+2); `core.c` `@@ -567` lands on the `wcn6750 hw1.0` entry (`.name` at 568); `ahb.c` `@@ -1358` lands inside `ath11k_ahb_shutdown()` as before. The QCN6122 patches (920-925) insert their `ath11k_hw_params` entry *after* the wcn6750 one, so the core.c offset is unshifted.
- Diffstat is self-consistent: 4 + 1 + 11 = 16 insertions across 3 files.
- `ab->hw_params` is embedded by value in `struct ath11k_base` (core.h:977), so `ab->hw_params.shutdown_teardown` is valid at `shutdown()` time; `.shutdown_teardown` is set on exactly one hw_params entry.
- The zero-SHA `From 0000000000000000000000000000000000000000` header matches in-tree convention for not-yet-merged patches (949 does the same), so no concern there.
- The comment style (`/*` on its own logical first line, ` * ` continuation) matches netdev/ath11k style in both `hw.h` and `ahb.c`.

Three nits inline: quilt hunk-header form, the `PKG_RELEASE` gap, and a question about `openwrt-25.12` being affected by the same bug.

---

ddimension:
Rerolled to match the upstream ath11k review of v2.

Changes since v2 (Jeff Johnson's review):
- `ahb.c` / `hw.h`: fix block-comment style (`/*` on its own line) and drop the chip references from the generic code — the comments now describe the behaviour, not the chipset; only the `ath11k_hw_params` initializer ties the feature to WCN6750.
- `core.c`: explicitly initialize `shutdown_teardown` in every `ath11k_hw_params` record (`= false` on all but WCN6750), per the "initialize all members" convention.

Matches the patch now on the list (v3):
https://lore.kernel.org/linux-wireless/20260810192859.949348-1-avalentin@marcant.net/

(mac80211 PKG_RELEASE bumped accordingly; CI green.)

package/kernel/mac80211/patches/ath11k/950-wifi-ath11k-run-ahb-shutdown-teardown-only-on-WCN6750.patch
```diff
@@ -0,0 +1,170 @@
...
+Signed-off-by: André Valentin <avalentin@marcant.net>
+---
+ drivers/net/wireless/ath/ath11k/hw.h | 7 +++++++
+ drivers/net/wireless/ath/ath11k/core.c | 11 +++++++++++
+ drivers/net/wireless/ath/ath11k/ahb.c | 10 ++++++++++
+ 3 files changed, 28 insertions(+)
+
+--- a/drivers/net/wireless/ath/ath11k/hw.h
++++ b/drivers/net/wireless/ath/ath11k/hw.h
+@@ -233,6 +233,13 @@
```
openwrt-ai: nit: (carry-over from v2, now 13 hunks instead of 3) every hunk header in this patch is bare `@@ -a,b +c,d @@` with no enclosing-symbol suffix, which is what plain `diff -u` (without `-p`) emits. Quilt — what `Quilt/Refresh/Package` in `package/kernel/mac80211/Makefile` drives for `patches/ath11k/` — emits the suffix, and the other 20 patches in the directory carry it, e.g. [`949-wifi-ath11k-fix-monitor-rx-pktlen.patch:27`](``https://github.com/openwrt/openwrt/blob/b8f1987e78656769f34371c041c948beff0856c1/package/kernel/mac80211/patches/ath11k/949-wifi-ath11k-fix-monitor-rx-pktlen.patch#L27).``

I re-verified against the backports tree (v6.18.41 `drivers/net/wireless/ath/ath11k/` with all 20 in-tree ath11k patches applied in order — backports 6.18.39 and v6.18.41 do not diverge in ath11k): **all 13 hunks apply with zero fuzz and zero offset**, and every `+` line number is exact, so this is purely about matching the directory's generated-by-quilt form. Regenerating the file with `make package/mac80211/refresh` produces:

| line | refreshed header |
| --- | --- |
| 48 | `@@ -233,6 +233,13 @@ struct ath11k_hw_params {` |
| 64 | `@@ -42,6 +42,7 @@ MODULE_PARM_DESC(ftm_mode, "Boots up in` |
| 72 | `@@ -129,6 +130,7 @@ static struct ath11k_hw_params ath11k_hw` |
| 80 | `@@ -215,6 +217,7 @@ static struct ath11k_hw_params ath11k_hw` |
| 88 | `@@ -305,6 +308,7 @@ static struct ath11k_hw_params ath11k_hw` |
| 96 | `@@ -389,6 +393,7 @@ static struct ath11k_hw_params ath11k_hw` |
| 104 | `@@ -479,6 +484,7 @@ static struct ath11k_hw_params ath11k_hw` |
| 112 | `@@ -567,6 +573,7 @@ static struct ath11k_hw_params ath11k_hw` |
| 120 | `@@ -649,6 +656,7 @@ static struct ath11k_hw_params ath11k_hw` |
| 128 | `@@ -733,6 +741,7 @@ static struct ath11k_hw_params ath11k_hw` |
| 136 | `@@ -822,6 +831,7 @@ static struct ath11k_hw_params ath11k_hw` |
| 144 | `@@ -909,6 +919,7 @@ static struct ath11k_hw_params ath11k_hw` |
| 154 | `@@ -1358,6 +1358,16 @@ static void ath11k_ahb_shutdown(struct p` |

The suggestion below covers this line; with 12 more sites a `make package/mac80211/refresh` is probably easier than applying them one by one.

```suggestion
@@ -233,6 +233,13 @@ struct ath11k_hw_params {
```

---

package/kernel/mac80211/Makefile
```diff
@@ -11,7 +11,7 @@ include $(INCLUDE_DIR)/kernel.mk
...
-PKG_RELEASE:=3
+PKG_RELEASE:=6
```
openwrt-ai: nit: the gap widened with each reroll — `origin/main` still has `PKG_RELEASE:=3` and this is a single commit on top of it, so the bump belongs on `4`. v1 bumped 3→4, v2 bumped the already-bumped value to 5, v3 to 6, so `4` and `5` are now skipped release numbers that never existed on main.

Harmless for the build, but `PKG_RELEASE` feeds `ABI_VERSION:=$(PKG_VERSION)-$(PKG_RELEASE)` for `kmod-cfg80211`/`kmod-mac80211`, so it is nicer kept monotonic-by-one.

```suggestion
PKG_RELEASE:=4
```

---

openwrt-ai (commented):
Reviewed the force-pushed v3 (`b8f1987`, replacing `a66d28d`). The two substantive changes since v2 — the block-comment restyle and the explicit `.shutdown_teardown` initializers on every `ath11k_hw_params` record — both check out, and the commit message matches what the diff does.

Verification I ran against the backports tree (extracted `drivers/net/wireless/ath/ath11k/` from Linux v6.18.41 and applied all 20 in-tree `patches/ath11k/*.patch` in order; backports 6.18.39 vs. v6.18.41 do not diverge in ath11k):

- **All 13 hunks apply with zero fuzz and zero offset.** The 10 newly-added `core.c` hunks land exactly on the `.hw_rev` lines they claim: 42→IPQ8074, 129→IPQ6018, 215→QCA6390, 305→QCN9074, 389/479→WCN6855 hw2.0/2.1, 567→WCN6750, 649→IPQ5018, 733→QCA2066, 822→QCA6698AQ, 909→QCN6122. Note the QCN6122 entry is OpenWrt-only (added by in-tree patch 920), so that hunk is an adaptation on top of the upstream v3 patch rather than a copy of it — correctly done, and worth knowing since it is the piece that will need re-adapting when the upstream commit arrives via a backports bump.
- **Coverage is complete and consistent.** The patched `core.c` has exactly 11 `ath11k_hw_params` entries and all 11 now carry `.shutdown_teardown`, always placed immediately after `.hw_rev`. Exactly one is `true` (wcn6750 hw1.0).
- **Embedded diffstat is self-consistent** with the payload: hw.h 7, core.c 11, ahb.c 10 = 28 insertions, and each hunk header's new-length matches (233,13 = 6+7; 1358,16 = 6+10).
- `ab->hw_params` is embedded by value in `struct ath11k_base`, so `ab->hw_params.shutdown_teardown` is valid at `shutdown()` time; the early `return` sits after the pre-existing explanatory comment and before `ath11k_ahb_remove_prepare()`, leaving `remove()` untouched as the commit message states.
- On the block-comment restyle: I checked before flagging it. The netdev-specific "text on the first line" exception is no longer in `Documentation/process/coding-style.rst` (absent in both v6.12.100 and v6.18.41), so `/*` on its own line is the currently-documented kernel-wide preferred form and the reroll is moving toward it, not away. It does leave two adjacent comments in `ath11k_ahb_shutdown()` in different styles, but that is the pre-existing one being stale — not something to fix here.
- CI is green on `b8f1987` (Build Kernel, Build all core packages, Build PR Profile all `success`).

Two nits inline, both carried over from v2 and both still open. The earlier question about whether `openwrt-25.12` (same `PKG_VERSION:=6.18.39`, no equivalent patch) needs the same fix is still unanswered and is the one item here with user-visible impact — worth a line in the PR either way so it does not get lost.

---

openwrt-ai (commented):
Reviewed the force-pushed v4 (`11b6209`, replacing `b8f1987`). Both nits from the v3 review are fixed and nothing else changed — the patch payload, the commit message and the diffstat are byte-identical to v3.

Delta since v3 is exactly two things, both of them the requested fixes:

- **`PKG_RELEASE` 6 → 4.** `origin/main` is still at `PKG_RELEASE:=3` and this is a single commit on top of it, so `4` is the correct monotonic-by-one bump and the skipped `4`/`5` gap left by the earlier rerolls is gone.
- **Quilt hunk headers.** All 13 headers now carry the enclosing-symbol suffix, matching the rest of `package/kernel/mac80211/patches/ath11k/`.

Verification I re-ran (reconstructed `drivers/net/wireless/ath/ath11k/` and applied all 20 in-tree ath11k patches in order, then patch 950):

- **Every hunk header matches a freshly generated `diff -up` exactly**, symbol suffix included — `struct ath11k_hw_params {` for `hw.h`, `MODULE_PARM_DESC(ftm_mode, "Boots up in` / `static struct ath11k_hw_params ath11k_hw` for the eleven `core.c` hunks, `static void ath11k_ahb_shutdown(struct p` for `ahb.c`. No transcription errors in the suffixes.
- The patch applies with **zero fuzz**. Note my reference tree this round was Linux **v6.18.44** rather than the v6.18.41 I used for v2/v3, and `ahb.c` picked up one line upstream in between, so that one hunk reports `offset 1` against v6.18.44 while `-1358` remains correct for the 6.18.39 backports tarball this package actually pins (verified zero-offset at 1358 in the v2/v3 reviews; the patch text is unchanged there).
- Coverage is still complete and consistent: 11 `ath11k_hw_params` entries, all 11 carry `.shutdown_teardown` immediately after `.hw_rev`, exactly one `true` (wcn6750 hw1.0).
- Embedded diffstat is self-consistent with the payload (7 + 11 + 10 = 28 insertions), and each hunk header's new-length checks out (233,13 = 6+7; 1358,16 = 6+10).
- No in-hunk context lines lost their leading space this time — the v1 whitespace issue has not regressed.
- CI on `11b6209` is clean: all completed check runs `success`, remainder still running, nothing failed.

No new issues. The only item still open is the pre-existing question on line 13 about [`openwrt-25.12`](https://github.com/openwrt/openwrt/pull/24601#discussion_r3742528193) — I re-checked it against the branch just now and it is still accurate: `openwrt-25.12` pins the same `PKG_VERSION:=6.18.39` and its `patches/ath11k/` still ends at `949-`, so IPQ8074/IPQ6018/IPQ5018 on 25.12 hit the identical warm-reboot hang. Not a blocker for this PR, just worth a line so a `[25.12]` backport does not get lost once this lands.

---

ddimension:
Answering the still-open question about `openwrt-25.12`: yes, that branch is
affected in exactly the same way, and yes, a backport is planned.

Re-checked against the branch just now: `openwrt-25.12` pins the same
`PKG_VERSION:=6.18.39` (at `PKG_RELEASE:=1`) and its `patches/ath11k/` ends at
`948-wifi-ath11k-Fix-the-WMM-param-type.patch`, so there is no equivalent change
there. IPQ8074/IPQ6018/IPQ5018 on 25.12 hit the identical warm-reboot hang: a
plain `reboot` stops in SBL1 during DDR training and needs a physical power
cycle.

I will submit the `[25.12]` backport once this has landed on `main`, so the two
branches do not diverge in patch numbering. It applies unchanged apart from the
`PKG_RELEASE` bump.

Status of the upstream submission, for the record: v3 is on the list since
2026-08-10 and has had no reply yet
(<https://lore.kernel.org/linux-wireless/20260810192859.949348-1-avalentin@marcant.net/>,
patchwork 14741952, still `new`). It addresses all three points from Jeff
Johnson's v2 review — block-comment style, no chipset references in the generic
code, and explicit `.shutdown_teardown` initialisers in every `ath11k_hw_params`
record. I will ping the list. This OpenWrt backport can be dropped again as soon
as a backports bump carries the upstream commit.

package/kernel/mac80211/Makefile
ddimension: Good catch — the PR body was updated accordingly: it now states the unconditional `.shutdown` is already present in v6.12, so every OpenWrt branch carrying that ath11k commit is affected, not just main. Resolving.

ddimension:
Superseded by 1bf6bd801751b74dd47d512c957c62ed73347d4e ("qualcommax: fix IPQ6018 WCSS SSR name").

I re-tested on the Chateau 5G R17 ax (IPQ6018): a build with Robert's SSR-name fix applied and this ath11k backport reverted survives repeated `reboot` cleanly (4/4, no SBL1 hang). So the warm-reboot hang here was the invalid `q6wcss` SSR name that the 6.18 patch refresh (97f8b0878d62) left on IPQ6018 and that 1bf6bd80 corrects to `wcnss` — not something this ath11k change is needed for on OpenWrt.

Closing as redundant now that 1bf6bd80 is in master.

The upstream side is separate: mainline `qcom_q6v5_wcss.c` still hardcodes `q6wcss` with no `wcnss` workaround, so mainline IPQ6018 hits the same path — that's being handled on the linux-wireless/ath11k list, where fixing the SSR name in remoteproc is likely the cleaner root-cause fix than skipping the teardown.

PR #24606, opened 2026-08-07
airoha: an7583: fix ethernet breakage on Nokia XG-040G-MF

csharper2005:
Recent upstream cleanups moved switch port labels to the board level and disabled switch ports/PHYs by default on the AN7583 SoC dtsi. This broke ethernet on the Nokia XG-040G-MF since its board DTS was not updated to explicitly configure and enable them.

Fix this by explicitly enabling gsw_port2-4 along with their respective PHYs and restoring the "lan2", "lan3", and "lan4" port labels.

Fixes: ce55cdde26a9 ("airoha: an7583: move the switch port label from soc-level to board-level")
Fixes: d2698d325b68 ("airoha: an7583: disable all switch ports and PHYs by default")

Cc @danielschwierzeck 
Cc @robimarko please help fix regression

### DIFF
--- a/target/linux/airoha/dts/an7583-nokia_xg-040g-mf-common.dtsi
+++ b/target/linux/airoha/dts/an7583-nokia_xg-040g-mf-common.dtsi
@@ -207,15 +207,15 @@
 	pinctrl-0 = <&mdio0_pins>;
 };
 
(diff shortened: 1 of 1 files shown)


robimarko:

PR #24607, opened 2026-08-07
uboot-zynq: bump to 2026.04

wigyori:
While at it, replace the SPI MAC address read patch with a rebased series from the Digilent u-boot repository. It should be noted that this feature is still not upstreamed. Refresh patches as required.

As most of the zynq boards are merged into xilinx_zynq_virt, add UBOOT_CONFIG to the respective boards in the Makefile.

Compile-tested: all Zynq boards
Runtime-tested: Antminer S9


wigyori:
Ping @luaraneda

package/boot/uboot-zynq/patches/111-digilent-net-Set-QSPI-read-settings-via-Kconfig.patch
```diff
@@ -0,0 +1,97 @@
...
++++ b/board/digilent/Kconfig
+@@ -0,0 +1,48 @@
++# SPDX-License-Identifier: GPL-2.0
++#
++# Copyright (c) 2023, Ionel Bădișor <ionel.badisor@digilent.ro>
++
++
++if ARCH_ZYNQ || ARCH_ZYNQMP
++
++config ZYNQ_GEM_MAC_FROM_QSPI
```
openwrt-ai, line 57: Nothing in this PR ever turns `ZYNQ_GEM_MAC_FROM_QSPI` on, so the whole 110–114 series is dead code and the Zybo Z7 silently loses its MAC-from-SPI-OTP support.

The patch this replaces did enable the feature in a defconfig — [`110-zybo-z7-read-mac-address-from-SPI-flash-memory.patch:72-74`](``https://github.com/openwrt/openwrt/blob/29912055552480e5aedc9b81facc82709f1cf7d7/package/boot/uboot-zynq/patches/110-zybo-z7-read-mac-address-from-SPI-flash-memory.patch#L72-L74)`` added `CONFIG_MAC_ADDR_IN_SPI_FLASH=y`, `CONFIG_MAC_ADDR_SPI_FLASH_READ_CMD=0x4b` and `CONFIG_MAC_ADDR_SPI_FLASH_DATA_OFFSET=0x20` to `configs/zynq_zybo_z7_defconfig`. The new series only adds Kconfig symbols (all defaulting to `n`) plus `obj-$(CONFIG_ZYNQ_GEM_MAC_FROM_QSPI)` in the Makefile hunk of patch 112, and no patch here touches any defconfig. `zynq_zybo_z7_defconfig` is also gone upstream now that the board builds `xilinx_zynq_virt`.

Net effect on a Zybo Z7 booting 2026.04: `zynq_board_read_rom_ethaddr()` is never linked, the `__weak` stub returns `-ENOSYS`, and the GEM falls back to the DT/random MAC. That's a behaviour regression versus the current tree, and it's consistent with the file never being compiled (see the `<common.h>` comment on patch 112, which would be a hard build failure otherwise).

Since `xilinx_zynq_virt` is shared by zc702/zed/zybo/zybo_z7, enabling this in that defconfig affects all four — was leaving the feature disabled intentional, or is a defconfig hunk (`CONFIG_ZYNQ_GEM_MAC_FROM_QSPI=y`, `CONFIG_ZYNQ_GEM_MAC_IN_OTP=y`, `CONFIG_QSPI_IS_SM=y`, offset `0x20`) missing from the series?

---

package/boot/uboot-zynq/patches/112-digilent-net-Set-MAC-address-from-QPSI-flash.patch
```diff
@@ -0,0 +1,143 @@
...
+ 2 files changed, 116 insertions(+)
+ create mode 100644 board/digilent/common/eth_mac_from_qspi.c
+
+--- /dev/null
++++ b/board/digilent/common/eth_mac_from_qspi.c
+@@ -0,0 +1,113 @@
++
++#define LOG_CATEGORY UCLASS_ETH
++
++#include <common.h>
```
openwrt-ai, line 18: `<common.h>` no longer exists in U-Boot — it was removed tree-wide during the 2024 cleanup, and `include/common.h` is absent from v2026.04 (`grep -r 'include <common.h>'` over v2026.07 returns 0 hits). The moment `CONFIG_ZYNQ_GEM_MAC_FROM_QSPI=y` is set, this file fails to compile with "No such file or directory".

The reason the build is green today is that nothing enables that symbol (see the comment on patch 111), so `eth_mac_from_qspi.c` is never fed to the compiler — the "compile-tested: all Zynq boards" note doesn't cover this file.

The include is unused anyway; `<log.h>`, `<net.h>`, `<spi_flash.h>` and `<linux/errno.h>` already cover what the function needs, and `<bootdev.h>`/`<i2c.h>`/`<dm/uclass.h>` look unused too. Dropping lines changes the `@@ -0,0 +1,113 @@` hunk count, so please regenerate the patch afterwards (`make package/uboot-zynq/refresh`) rather than editing the header by hand.

---

package/boot/uboot-zynq/patches/112-digilent-net-Set-MAC-address-from-QPSI-flash.patch
```diff
@@ -0,0 +1,143 @@
...
++	u32 offset;
++	struct spi_flash *flash;
++	flash = spi_flash_probe(CONFIG_SF_DEFAULT_BUS, CONFIG_SF_DEFAULT_CS,
++	                        CONFIG_SF_DEFAULT_SPEED, CONFIG_SF_DEFAULT_MODE);
++
++	if (!flash) {
++		printf("no flash\n");
++		printf("SPI(bus:%u cs:%u) probe failed\n", CONFIG_SF_DEFAULT_BUS,
++		       CONFIG_SF_DEFAULT_CS);
++		return 0;
```
openwrt-ai, line 41: Returning `0` on a probe failure reports success to the caller while `ethaddr` is left untouched. In [`eth_post_probe()` at net/eth-uclass.c:600-606](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/net/eth-uclass.c#L599-L607) a zero return sets `source = "ROM"`, so a later MAC mismatch warning would blame a ROM address that was never read. It should propagate an error (`-ENODEV`, or the `ret` that `spi_flash_probe()` failure implies) so the caller falls through cleanly.

This line survives into the final state — patch 114 keeps it as context at line 14 — so fixing it here also needs that context line updated, i.e. edit and then `make package/uboot-zynq/refresh`.

Also note the two `printf()`s directly above are redundant: `"no flash\n"` adds nothing over the `"SPI(bus:%u cs:%u) probe failed\n"` line that follows.

---

package/boot/uboot-zynq/patches/114-digilent-net-Read-MAC-address-from-OTP-section-of-QS.patch
```diff
@@ -0,0 +1,114 @@
...
+-	 */
+-	offset = CONFIG_ZYNQ_GEM_ETH_MAC_QSPI_OFFSET;
++#if defined (CONFIG_ZYNQ_GEM_MAC_IN_OTP)
++
++	#if defined (CONFIG_QSPI_IS_SM) // For Spansion and Micron OPCODE
++		#define CMD_OTP_READ_ARRAY_FAST 0x4b
++	#elif defined (CONFIG_QSPI_IS_WB) // For Winbond OPCODE
++		#define CMD_OTP_READ_ARRAY_FAST 0x48
++	#else
++		#define CMD_OTP_READ_ARRAY_FAST "NOT IMPLEMENTED"
```
openwrt-ai, line 100: This fallback expands to `flash->read_opcode = "NOT IMPLEMENTED";` — assigning a `const char *` to the `u8 read_opcode` field of `struct spi_nor`, which is a hard compile error, not a diagnostic.

It is reachable from Kconfig: patch 111 defines `QSPI_IS_SM` and `QSPI_IS_WB` as two independent bools with no `default` and no enclosing `choice`, so `ZYNQ_GEM_MAC_IN_OTP=y` with both flash types left unset is a perfectly valid config. Use `#error "..."` here so the failure is legible, and/or make the two flash-vendor symbols a Kconfig `choice` in patch 111 so one is always selected when `ZYNQ_GEM_MAC_IN_OTP` is on.

---

package/boot/uboot-zynq/patches/112-digilent-net-Set-MAC-address-from-QPSI-flash.patch
```diff
@@ -0,0 +1,143 @@
...
++	}
++
++#if defined(ZYNQ_GEM_MAC_IN_OTP)
++
++	NOT VERIFIED
```
openwrt-ai, line 46: nit: patch-series hygiene — this imports a broken intermediate state that the next two patches then rewrite. `NOT VERIFIED` is a bare token that cannot compile, the guard above it is `#if defined(ZYNQ_GEM_MAC_IN_OTP)` (missing the `CONFIG_` prefix, so the block is dead), and the `#else` branch references `CONFIG_ZYNQ_GEM_ETH_MAC_QSPI_OFFSET`, a symbol patch 113 renames to `CONFIG_ZYNQ_GEM_MAC_QSPI_OFFSET`.

Same pattern in patch 111: `depends on !QPSI_IS_SM` (line 80) is a typo for `QSPI_IS_SM` that patch 113 fixes, and the three commented-out `#default` lines it adds are removed again by 113.

Since the series is explicitly not upstream and carries no bisectability value in this tree, squashing 110–114 into one `110-digilent-zynq-gem-mac-from-qspi.patch` reflecting only the final state would be easier to carry across the next bump. While at it, the stray blank line patch 112 adds to `board/xilinx/common/Makefile` is unrelated churn.

---

package/boot/uboot-zynq/patches/110-digilent-net-Add-entry-point-for-setting-zynq_get-MA.patch
```diff
@@ -0,0 +1,50 @@
...
++
++
+ static int zynq_gem_setup_mac(struct udevice *dev)
+ {
+ 	u32 i, macaddrlow, macaddrhigh;
+@@ -1036,6 +1052,7 @@ static const struct eth_ops zynq_gem_ops
+ 	.free_pkt		= zynq_gem_free_pkt,
+ 	.stop			= zynq_gem_halt,
+ 	.write_hwaddr		= zynq_gem_setup_mac,
++	.read_rom_hwaddr		= zynq_gem_read_rom_mac,
```
openwrt-ai, line 47: nit: the extra tab breaks the `=` alignment of the `eth_ops` initialiser. `.read_rom_hwaddr` is 16 chars, so with the leading tab it already reaches column 24 — one more tab lands on 32, matching `.free_pkt`/`.stop`/`.write_hwaddr` above; two tabs push it to 40.

```suggestion
+	.read_rom_hwaddr	= zynq_gem_read_rom_mac,
```

Also, the hunk above adds two consecutive blank lines before `zynq_gem_setup_mac()` (lines 38-39) — one is enough.

---

package/boot/uboot-zynq/Makefile
```diff
@@ -28,6 +27,7 @@ endef
 define U-Boot/zc702
   NAME:=Xilinx ZC702 Dev Board
   BUILD_DEVICES:=xlnx_zynq-zc702
+  UBOOT_CONFIG:=xilinx_zynq_virt
```
openwrt-ai, line 30: nit: with this hunk (and the matching ones for `zed`, `zybo`, `zybo_z7`) every one of the five targets now sets `UBOOT_CONFIG` explicitly, so `UBOOT_CONFIG:=zynq_$(1)` in `U-Boot/Default` (line 22) is dead — and worse, it points at `zynq_zc702_defconfig`-style names that no longer exist in U-Boot 2026.04. A new board added later would silently inherit a non-existent defconfig.

Either drop the line from `U-Boot/Default` or change it to `xilinx_zynq_virt`, which is what four of the five boards want anyway (and would then let these four `UBOOT_CONFIG` lines go away, leaving only `antminer-s9` to override).

---

openwrt-ai (commented):
Reviewed 1 new commit. The bump itself and the `UBOOT_CONFIG:=xilinx_zynq_virt` switch check out — `xilinx_zynq_virt_defconfig` and `bitmain_antminer_s9_defconfig` both exist in U-Boot 2026.04, while the old per-board `zynq_*_defconfig` files are gone. CI on `051a9a2` is green.

The main concern is the replacement MAC-from-QSPI series: nothing in the tree enables `CONFIG_ZYNQ_GEM_MAC_FROM_QSPI`, whereas the patch being removed enabled the equivalent option in `zynq_zybo_z7_defconfig`. Details inline, along with a `<common.h>` include that would break the build if the feature were ever switched on.

---

package/boot/uboot-zynq/patches/115-enable-read-mac-from-qspi.patch
```diff
@@ -0,0 +1,7 @@
+--- a/configs/xilinx_zynq_virt_defconfig
++++ b/configs/xilinx_zynq_virt_defconfig
+@@ -156,3 +156,4 @@ CONFIG_USB_GADGET_DOWNLOAD=y
+ CONFIG_SYS_TIMER_COUNTS_DOWN=y
+ CONFIG_MBEDTLS_LIB=y
+ CONFIG_SPL_GZIP=y
++CONFIG_ZYNQ_GEM_MAC_FROM_QSPI=y
```
openwrt-ai, line 7: Enabling only `ZYNQ_GEM_MAC_FROM_QSPI` makes the MAC be read from offset `0x0` of the main flash array — i.e. the first six bytes of `boot.bin` — not from the OTP area.

`ZYNQ_GEM_MAC_IN_OTP` stays `n`, so in [`113-...-Set-Kconfig-defaults-...patch:49-51`](``https://github.com/openwrt/openwrt/blob/4704f346669250e7ea9aebabfb654a2a32ccc775/package/boot/uboot-zynq/patches/113-digilent-net-Set-Kconfig-defaults-for-ZYNQ_GEM_MAC_Q.patch#L49-L51)`` the only matching default is `default 0x0 if ZYNQ_GEM_MAC_FROM_QSPI && !ZYNQ_GEM_MAC_IN_OTP`, and `#if defined(CONFIG_ZYNQ_GEM_MAC_IN_OTP)` in patch 114 is false so `flash->read_opcode` keeps the normal read command. The patch being removed did the opposite: [`110-zybo-z7-...patch:66-68`](``https://github.com/openwrt/openwrt/blob/652b2e90b3c4d7f5b82ea3c7e67033ee1500f3b1/package/boot/uboot-zynq/patches/110-zybo-z7-read-mac-address-from-SPI-flash-memory.patch#L66-L68)`` set `CONFIG_MAC_ADDR_SPI_FLASH_READ_CMD=0x4b` (Spansion/Micron OTP read) and `CONFIG_MAC_ADDR_SPI_FLASH_DATA_OFFSET=0x20`.

This isn't a silent no-op: on a Zynq the QSPI starts with the BootROM header (`0xEAFFFFFE` dummy words), so `spi_flash_read()` returns 0 with `ethaddr = fe:ff:ff:ea:fe:ff`, which passes `is_valid_ethaddr()` in [`eth_post_probe()` at net/eth-uclass.c:596-606](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/net/eth-uclass.c#L596-L606) and gets stored into `ethaddr` in the environment.

Setting `ZYNQ_GEM_MAC_IN_OTP` + `QSPI_IS_SM` restores the old semantics exactly — `CONFIG_ZYNQ_GEM_MAC_QSPI_OFFSET` then picks up `default 0x20` and patch 114 sets opcode `0x4b`, so no extra defconfig line is needed for the offset:

```suggestion
@@ -156,3 +156,6 @@ CONFIG_USB_GADGET_DOWNLOAD=y
 CONFIG_SYS_TIMER_COUNTS_DOWN=y
 CONFIG_MBEDTLS_LIB=y
 CONFIG_SPL_GZIP=y
+CONFIG_ZYNQ_GEM_MAC_FROM_QSPI=y
+CONFIG_ZYNQ_GEM_MAC_IN_OTP=y
+CONFIG_QSPI_IS_SM=y
```

Separately, note the scope change this file implies: `xilinx_zynq_virt` is shared by zc702, zed, zybo and zybo_z7 ([Makefile](https://github.com/openwrt/openwrt/blob/4704f346669250e7ea9aebabfb654a2a32ccc775/package/boot/uboot-zynq/Makefile#L26-L55)), whereas the removed patch touched `zynq_zybo_z7_defconfig` only. With the OTP variant above that is harmless for the other three (a blank OTP reads `ff:ff:ff:ff:ff:ff`, rejected as multicast, and boot falls through to DT/env); with the offset-`0x0` variant all four get a bogus-but-valid MAC. Was the wider scope intended?

---

package/boot/uboot-zynq/patches/114-digilent-net-Read-MAC-address-from-OTP-section-of-QS.patch
```diff
@@ -0,0 +1,114 @@
...
++	#endif
++
++	//change read opcode only when OTP is used
++	flash->read_opcode = CMD_OTP_READ_ARRAY_FAST;
+ #endif
+ 
++	printf("%s: Reading SPI MAC address from offset %08lxh\n", __func__,
++	       CONFIG_ZYNQ_GEM_MAC_QSPI_OFFSET);
```
openwrt-ai, line 108: nit: `%08lx` expects `unsigned long`, but `CONFIG_ZYNQ_GEM_MAC_QSPI_OFFSET` expands to a plain `0x20`/`0x0` integer constant, so this is a `-Wformat` warning. It works on arm32 where `long` is 32-bit, but `%08x` is the correct specifier (the same mismatch existed in the pre-114 `#else` branch, so it comes from the imported series rather than from you).

Worth mentioning because patch 115 makes `eth_mac_from_qspi.c` reach the compiler for the first time — until this revision `CONFIG_ZYNQ_GEM_MAC_FROM_QSPI` was never set, so `obj-$(CONFIG_ZYNQ_GEM_MAC_FROM_QSPI)` in patch 112 never pulled the file in and the `Compile-tested: all Zynq boards` line in the commit message didn't cover it. OpenWrt CI doesn't build zynq either (only x86/64 and malta/be ran on this head), so a fresh `xilinx_zynq_virt` build is the only thing that will catch warnings here.

---

openwrt-ai (commented):
Reviewed 1 new commit (`051a9a2` → `4704f34`). Two of the earlier findings are addressed: the `<common.h>` include is gone from patch 112, and the new `115-enable-read-mac-from-qspi.patch` turns `ZYNQ_GEM_MAC_FROM_QSPI` on so the series is no longer dead code.

The new defconfig patch introduces a follow-on problem, though — enabling only `ZYNQ_GEM_MAC_FROM_QSPI` resolves the offset to `0x0` and leaves the normal read opcode in place, so the MAC ends up being read from the start of the boot image rather than the OTP area the removed patch used. Details and a suggested defconfig hunk inline.

Still open from the previous round and unchanged in this revision: the `return 0` on `spi_flash_probe()` failure in patch 112, the `NOT IMPLEMENTED` fallback in patch 114 (now reachable in a valid Kconfig combination, and a hard compile error if hit), the series-hygiene note on the imported 110–114 patches, and the dead `UBOOT_CONFIG:=zynq_$(1)` in `U-Boot/Default`.

## Commit checks
- 4704f34 "uboot-zynq: bump to 2026.04" — the message covers the version bump, the replacement Digilent series and the `UBOOT_CONFIG` additions, but not the locally-written `115-enable-read-mac-from-qspi.patch`, which is the hunk with the actual behaviour change: MAC-from-QSPI goes from zybo_z7-only to all four `xilinx_zynq_virt` boards. Worth a sentence in the body. Also `Runtime-tested: Antminer S9` doesn't exercise that path, since the S9 builds `bitmain_antminer_s9_defconfig`.

---

peterwillcn (approved):

luaraneda:
> Ping @luaraneda

I just had the time to look at this. Is there anything (review, test) you wanted from my side?

I seems to be closed now, I don't know whether it was dropped, it will be merged, or a new PR will appear.

I don't use my board a lot these days. And I would be in favor of dropping the patch that was no upstreamed to keep things simple. We can always set the MAC to a specific value from U-Boot anyway.

PR #24608, opened 2026-08-07
uboot-sifiveu: bump to 2026.04 and upgrade opensbi

wigyori:
- opensbi: bump to v1.9

While at it, move away from using dated builds to release tags.
Compile-tested: sifiveu/all, d1/all
Runtime-tested: sifiveu/unleashed, sifiveu/unmatched, d1/nezha

 - uboot-sifiveu: bump to 2026.04

Remove upstreamed patches and refresh remaining ones as required.
Compile-tested: sifiveu/all, d1/all
Runtime-tested: sifiveu/unleashed, sifiveu/unmatched, d1/nezha


package/boot/opensbi/Makefile
```diff
@@ -6,13 +6,12 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=opensbi
-PKG_RELEASE:=1.6
+PKG_VERSION:=1.9
```
openwrt-ai, line 9: `PKG_RELEASE` is dropped entirely here: the old line was `PKG_RELEASE:=1.6` (a version string misfiled into `PKG_RELEASE`), and it is replaced by `PKG_VERSION:=1.9` with no `PKG_RELEASE` left. With both `PKG_VERSION` and `PKG_RELEASE` unset-vs-set, [`include/package-defaults.mk:24-32`](https://github.com/openwrt/openwrt/blob/ae7a050d10b4488cc68ada46a10325220a6c6856/include/package-defaults.mk#L24-L32) now yields `VERSION:=1.9` with no `-r<N>` suffix, so there is no way to bump the package for a packaging-only change. The sibling package in this same PR keeps [`PKG_RELEASE:=1`](https://github.com/openwrt/openwrt/blob/ae7a050d10b4488cc68ada46a10325220a6c6856/package/boot/uboot-sifiveu/Makefile#L9). This is also what the formality bot flagged on this PR.

Since it is not mentioned in the commit message ("move away from using dated builds to release tags" covers `PKG_SOURCE_DATE`/`PKG_SOURCE_VERSION`, not `PKG_RELEASE`), I assume it was unintended:

```suggestion
PKG_VERSION:=1.9
PKG_RELEASE:=1
```

---

package/boot/uboot-sifiveu/patches/110-no-kwbimage.patch
```diff
@@ -1,10 +0,0 @@
---- a/tools/Makefile
```
openwrt-ai, line 1: nit: the commit message says "Remove upstreamed patches", but this one and `130-fix-mkimage-host-build.patch` were OpenWrt-local workarounds that were never sent upstream — they are obsolete because upstream grew its own guards, not because they were merged:

- `110-no-kwbimage.patch`: upstream now has [`KWB_IMAGE_OBJS-$(CONFIG_TOOLS_LIBCRYPTO) := kwbimage.o`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/tools/Makefile#L107), so the `sed` that clears `CONFIG_TOOLS_LIBCRYPTO` in `Build/Configure` already drops the object.
- `130-fix-mkimage-host-build.patch`: `read_pub_key()` is now inside [`#if CONFIG_IS_ENABLED(FIT_SIGNATURE)`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/tools/image-host.c#L1186), and [`TOOLS_FIT_SIGNATURE depends on TOOLS_LIBCRYPTO`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/tools/Kconfig#L63-L67).

(The other two removals, `0006-...fu740-reduce-DDR-speed...` and `002-...SWIG_AppendOutput...`, genuinely are upstream.) Worth a word in the commit message so a future bisect does not read this as "upstream took our patch".

---

openwrt-ai (commented):
Reviewed 2 new commits. Both commit subjects match their diffs.

Verified against upstream:
- `v1.9` exists in `riscv/opensbi` (`e57b4a8b471192e479e4c7a438353f7ebbc10fac`), and `v2026.04` exists in `u-boot/u-boot`. The new `PKG_HASH` for u-boot 2026.04 matches the one already in-tree for `uboot-ath79`, `uboot-imx`, `uboot-mvebu` and `uboot-sunxi`.
- All four removed u-boot patches are indeed no longer needed at 2026.04 (details in the inline note).
- `arch/riscv/dts/fu740-c000.dtsi` still exists at `v2026.04`, so the remaining `0005-riscv-dts-Add-few-PMU-events.patch` still applies — note it disappears in `v2026.07` (moved to `dts/upstream/src/riscv/sifive/` by the OF_UPSTREAM migration), which will also invalidate `DTS_DIR:=arch/riscv/dts` in `Build/InstallDev` on the next bump.
- The refreshed hunk header in `100-mkimage-check-environment-for-dtc-binary-location.patch` (`@@ -1029,9`) lines up with `tools/fit_image.c` in current u-boot.

One real issue below (dropped `PKG_RELEASE`) plus one commit-message nit.

---

PR #24257, opened 2026-07-16
packages/boot: add arm-trusted-firmware-airoha + U-Boot sync

Ansuel:
Add support for compiling Airoha ATF, based on a modified version of the
TF-A based on 2.10.

The package will clone the base TF-A and apply the modified source and
pre-compiled objects (DDR calibration, eFUSE handling, TX/RX path...) to
correctly compile.

An helper script is added to implement the same format used for special BL2
handling. BL2 is composed of 3 sub-phase with a initial BL21, a BL22 that
loads a LZMA decompressor and BL23 that actually calibrate and loads BL31
ATF.

Also a special flash_table is used to handle SPI-NAND externally to the
BL2, a special Host target is added for that.

Move FIP handling to target now that we compile ATF. This is to have always
in sync new ATF and new U-Boot if one of the 2 is already compiled.

Also add HIDDEN ops to enable autoselection of the U-Boot for the required
target. This is needed to prevent user to deselect the U-Boot package
causing failure on ARTIFACTS generation.

PR #24609, opened 2026-08-07
airoha: add USB support for AN7583 and enable it on Nokia XG-040G-MF

VitaliySochniy:
The `airoha/an7583` subtarget had no USB support at all: `CONFIG_USB_SUPPORT` was unset, the SoC dtsi carried neither an xHCI nor a USB PHY node, and the Nokia XG-040G-MF device tree had its entire USB description commented out with a `AN7583 SoC doesn't support USB at this time` note.

This series adds the DT binding and the AN7583 USB PHY driver, describes the controller in the SoC dtsi, builds xHCI and its MediaTek glue into the subtarget kernel, and enables USB on the Nokia XG-040G-MF.

The AN7581 USB PHY driver cannot be reused here. AN7583 has a single xHCI controller fed by one PHY block that drives two USB 2.0 ports and one USB 3.0 port, and both the PLL init sequence and the U3 calibration differ from AN7581.

## Commits

| Commit | What |
|---|---|
| `airoha: an7583: add DT binding for AN7583 USB PHY` | New kernel patch `607-01-*`, documents `airoha,an7583-usb-phy` |
| `airoha: add support for AN7583 USB PHY` | New kernel patch `607-02-*` plus subtarget kernel configs |
| `airoha: an7583: add USB nodes to SoC dtsi` | xHCI + USB PHY nodes, left disabled |
| `airoha: an7583: build USB support into the kernel` | xHCI and the MediaTek glue built into the an7583 kernel |
| `airoha: an7583: enable USB on Nokia XG-040G-MF` | Board dtsi + image packages |

## Origin of the driver

The PHY driver comes from @Ansuel's [`main-airoha-6.12`](https://github.com/Ansuel/openwrt/tree/main-airoha-6.12) branch (`607-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch`, commit `ff57e25`). It is carried here with three changes:

1. **The PCIe PHY part was dropped.** The upstream file also matches `airoha,an7583-pcie-phy`, but no DT node for the AN7583 PCIe serdes exists in this tree yet, so that code would be dead. It can be restored as a separate patch when the PCIe serdes lands. The shared U3/HSGMII serdes code was kept — it belongs to the USB PHY, since AN7583 can repurpose the USB serdes for HSGMII via `airoha,an7583-pcs-usb`.
2. **Five bugs fixed.** Two found while carrying the patch: `an7583_usb_phy_parse_subnode()` passed an uninitialized `ret` to `dev_err_probe()`, and checked `syscon_regmap_lookup_by_phandle()` for `NULL` when it returns an `ERR_PTR` on failure. Three more found in review: `an7583_usb_phy_u2_set_mode()` was the only U2 helper not adding `instance->offset`, so it wrote the first U2 port's `U2PHYDTM1` regardless of which PHY it was called on; `parse_subnode()` bounded `index` with `>` instead of `>=`, allowing a write one past the end of `priv->phys[]`; and `an7583_usb_phy_init()` discarded the return values of both init helpers, swallowing the `-EINVAL` that `u3_init()` reports when the U3 KBand PLL fails to calibrate. All five are latent on this board but real; they are worth taking upstream rather than living only as a downstream delta.
3. **Naming adapted to this tree.** The donor branch is based on a later refactor of `drivers/phy/airoha/` (`PHY_AIROHA_AN7581_USB` / `phy-an7581-usb.o`); here the existing names are `PHY_AIROHA_USB` / `phy-airoha-usb.c`. The donor's rename of the AN7581 driver was deliberately **not** carried over, so AN7581 is untouched.

Three intentional deviations from the donor dtsi: the PHY node got a label so boards can enable it, `#address-cells`/`#size-cells` were added since the U2 subnodes carry `reg`, and the four `reg` regions are page-sized — review caught that `phy` (`0x11ff`) and `pma` (`0x3ff`) were smaller than the registers the driver actually touches, `0x136c` and `0x6f0` respectively.

## Testing

Tested on real hardware, on both affected subtargets.

### Nokia XG-040G-MF (AN7583) — USB now works

```
xhci-mtk 1fab0000.usb: xHCI Host Controller
xhci-mtk 1fab0000.usb: new USB bus registered, assigned bus number 1
xhci-mtk 1fab0000.usb: new USB bus registered, assigned bus number 2
xhci-mtk 1fab0000.usb: Host supports USB 3.2 Enhanced SuperSpeed
```

```
T:  Bus=01 ... Spd=480   MxCh= 2      <- two USB 2.0 ports
T:  Bus=02 ... Spd=10000 MxCh= 1      <- one USB 3.0 port
```

The PHY probes without errors. Verified with real devices:

| Device | Socket | Result |
|---|---|---|
| USB 2.0 stick | USB2 socket | `usb 1-2`, 480M, LED `usb-2` |
| USB 2.0 stick | USB3 socket | `usb 1-1`, 480M, LED `usb-1` |
| USB 3.0 stick | USB3 socket | `usb 2-1`, **5000M**, LED `usb-1` |
| both at once | both | both enumerate, both LEDs on |

SuperSpeed genuinely links up at 5 Gbps, so the U3 serdes calibration in the PHY driver works. The 5 V rail comes up automatically and the `usbport` LED triggers map correctly to the physical sockets. Verified both on an initramfs image and after a permanent sysupgrade.

Re-verified after the review fixes, with each fix exercised where it matters: the driver still binds to `1fac0000.phy` (all four `reg` regions map cleanly), a device on the second U2 port at offset `0x1000` still enumerates at 480M, and SuperSpeed still links at 5 Gbps, so `u3_init()` still reports success now that its return value is propagated.

Re-verified once more after the switch to built-in USB: every row of the table above still holds with the same LED mapping, the 5 V rail still comes up on its own, and xHCI now registers at 1.34 s instead of after the rootfs is mounted. `make kernel_oldconfig` reproduces `config-6.18` unchanged.

### Nokia XG-040G-MD (AN7581) — no regression

The board was booted with an image built from this series and compared against its previous firmware:

| | before | after |
|---|---|---|
| controllers | `1fab0000` + `1fad0000` | same |
| buses | 480 / 10000 / 480 / 10000 | identical |
| `usb-power` | enabled | enabled |

All three usable buses exercised: SuperSpeed 5000M on the blue port, 480M on the USB2-only port, and 480M on the U2 line of the blue port. The fourth bus is the SuperSpeed roothub of the second controller and has no usable port — that socket is USB 2.0 only (`mediatek,u3p-dis-msk = <0x1>` and a single `PHY_TYPE_USB2` in `phys`), so xhci-mtk registers the roothub but nothing can enumerate on it.

This is expected — the AN7581 kernel differs only by one `# CONFIG_PHY_AIROHA_AN7583_USB is not set` line, and the new driver is not compiled into it at all. The later switch to built-in USB touched only the an7583 subtarget config, so this result still stands.

## Notes for reviewers

- Following review feedback, xHCI and the MediaTek glue are built into the an7583 kernel rather than pulled in per board through `kmod-usb3`. The six config symbols are the same set `mediatek/mt7629` uses for this arrangement, and `KernelPackage/usb3` is left alone, so `package/kernel/linux/modules/usb.mk` is untouched by this series. The an7581 subtarget keeps the module route for now.
- `kmod-regulator-userspace-consumer` is required in `DEVICE_PACKAGES`, matching the XG-040G-MD. Without it the `regulator-output` node has no consumer and the kernel disables the 5 V rail as unused, so USB devices get no power. This was caught on hardware.


VitaliySochniy:
@Ansuel @csharper2005 @robimarko — could you take a look when you have a moment?

**@Ansuel** — you are the author of the AN7583 USB PHY driver, so I would value your review of how it is backported here. Three deviations from your version, all deliberate:

- The `airoha,an7583-pcie-phy` match was dropped: this tree has no DT node for the AN7583 PCIe serdes, so that code path would be unreachable. The shared U3/HSGMII serdes handling was kept, since it belongs to the USB PHY. If you would rather the file stayed byte-identical to ease a later PCIe patch, say so and I will restore it.
- Five bugs fixed — two found while backporting, three more raised in review on this PR: an uninitialized `ret` reaching `dev_err_probe()`; a `NULL` check on `syscon_regmap_lookup_by_phandle()`, which returns an `ERR_PTR`; a missing `instance->offset` in `an7583_usb_phy_u2_set_mode()`; an off-by-one bound in `parse_subnode()`; and a discarded return value in `phy_init()` that swallowed the U3 calibration failure. All are latent on this board, but they look worth having upstream rather than only here.
- Kconfig/Makefile naming follows this tree, which predates your `drivers/phy/airoha/` refactor: `PHY_AIROHA_USB` / `phy-airoha-usb.c` here versus `PHY_AIROHA_AN7581_USB` / `phy-an7581-usb.o` in yours. The new driver keeps your `PHY_AIROHA_AN7583_USB` symbol; only its file name follows the local `phy-airoha-*` convention. Your rename of the AN7581 driver was not carried over, so AN7581 stays untouched.

Three changes to the SoC dtsi on top: the PHY node got a label so boards can enable it, `#address-cells`/`#size-cells` were added since the U2 subnodes carry `reg`, and the four `reg` regions are page-sized — review caught that `phy` (`0x11ff`) and `pma` (`0x3ff`) were smaller than the registers the driver actually touches, `0x136c` and `0x6f0`.

**@csharper2005** — this enables the USB block you left commented out in the XG-040G-MF device tree. Two things worth your eyes:

- Port mapping was verified on hardware: the USB2 socket lands on roothub 1 port 2, and the USB3 socket on roothub 1 port 1 for its HS pins and roothub 2 port 1 for SuperSpeed. Both `usbport` LED triggers light the correct LED in every combination, including a USB3 device in the blue socket. Does that match what you see on your unit?
- `kmod-regulator-userspace-consumer` had to be added to `DEVICE_PACKAGES`, same as on the XG-040G-MD. Without it the `regulator-output` node has no consumer and the kernel disables the 5 V rail as unused at late init, so nothing gets power. I kept `kmod-usb-storage` out, matching the MD.

Unrelated to this series, but you may want to know: the MF logs a `_phy_state_machine` warning with `_phy_start_aneg: returned -22` from the EN8811H on lan1. It reproduces on builds predating this series, so I left it alone.

**@robimarko** — you merged the previous airoha commit this series is based on.

Tested on real hardware: USB 2.0 in both sockets and USB 3.0 linking at 5 Gbps on the XG-040G-MF, and no regression on the XG-040G-MD, where all three usable buses were exercised (HS and SuperSpeed on the USB3 port, HS on the USB2-only port) and the bus topology is identical before and after.

P.S. Regarding the PKG_RELEASE warning: package/kernel/linux/Makefile does not define PKG_RELEASE at all — kmod package versions are derived from LINUX_RELEASE via include/kernel.mk. There is nothing to bump here. The same kind of change for the sibling subtarget, commit f8a6798bf3 ("airoha: an7581: enable USB support"), also touched only usb.mk plus a subtarget config with no release bump.

target/linux/airoha/dts/an7583.dtsi
```diff
@@ -531,6 +532,53 @@
...
+
+			status = "disabled";
+		};
+
+		usb_phy: phy@1fac0000 {
+			compatible = "airoha,an7583-usb-phy";
+			reg = <0x0 0x1fac0000 0x0 0x11ff>,
+			      <0x0 0x1fa6f000 0x0 0xff>,
+			      <0x0 0x1fa6e000 0x0 0x3ff>,
+			      <0x0 0x1fa6a000 0x0 0x8ff>;
```
openwrt-ai: Two of these four regions are smaller than the register range the driver actually touches:

- `"phy"` is `0x11ff`, but the second U2 instance sits at `reg = <0x1000>` and the driver adds that offset to register offsets up to `AIROHA_USB_PHY_U2PHYDTM1` (`0x36c`), i.e. it reaches `0x136c`.
- `"pma"` is `0x3ff`, but PMA registers go up to `AIROHA_USB_PMA_RX_CTRL_50` (`0x6f0`).

`devm_platform_ioremap_resource_byname()` maps exactly `resource_size()`, and `devm_regmap_init_mmio()` is called without `max_register`, so nothing bounds-checks these accesses — they only land inside the mapping because `ioremap()` rounds up to a page, which is why hardware testing didn't catch it.

Sizing the regions to a full page matches how the sibling serdes blocks are described in this same file ([`pcs_ana` / `pcs_pma` on `eth_pcs`](https://github.com/openwrt/openwrt/blob/d042e9ec359b626e9c7224d44ba782c712fc85f9/target/linux/airoha/dts/an7583.dtsi#L645-L646) both use `0x1000`) and how [`usb0_phy` is described on an7581](https://github.com/openwrt/openwrt/blob/d042e9ec359b626e9c7224d44ba782c712fc85f9/target/linux/airoha/dts/an7581.dtsi#L603).

```suggestion
			reg = <0x0 0x1fac0000 0x0 0x2000>,
			      <0x0 0x1fa6f000 0x0 0x1000>,
			      <0x0 0x1fa6e000 0x0 0x1000>,
			      <0x0 0x1fa6a000 0x0 0x1000>;
```


---

target/linux/airoha/patches-6.18/607-02-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch
```diff
@@ -0,0 +1,1691 @@
...
++		break;
++	case PHY_MODE_USB_OTG:
++		break;
++	default:
++		return 0;
++	}
++
++	regmap_update_bits(priv->regs[AIROHA_USB_PHY_PHY], AIROHA_USB_PHY_U2PHYDTM1,
++			   AIROHA_USB_PHY_FORCE_IDDIG |
++			   AIROHA_USB_PHY_IDDIG, val);
```
openwrt-ai, line 1461: `an7583_usb_phy_u2_set_mode()` is the only U2 helper that does not add `instance->offset` to the register address — `an7583_usb_phy_u2_init()`, `_u2_power_on()`, `_u2_power_off()` and `_u2_slew_rate_calibration()` all do. On AN7583 the second U2 instance lives at offset `0x1000` (`usb1_u2: usb2-phy@1000` in the new dtsi), so a `set_mode()` on that PHY rewrites `U2PHYDTM1` of the *first* U2 port instead of its own. Host mode is the reset default for `FORCE_IDDIG`/`IDDIG`, which is why this stays invisible on the XG-040G-MF.

```suggestion
+static int an7583_usb_phy_u2_set_mode(struct phy *phy, enum phy_mode mode)
+{
+	struct an7583_usb_phy_instance *instance = phy_get_drvdata(phy);
+	struct an7583_usb_phy_priv *priv = dev_get_drvdata(phy->dev.parent);
+	u32 val = 0;
+
+	/*
+	 * For Device and Host mode, enable force IDDIG.
+	 * For Device set IDDIG, for Host clear IDDIG.
+	 * For OTG disable force and clear IDDIG bit while at it.
+	 */
+	switch (mode) {
+	case PHY_MODE_USB_DEVICE:
+		val |= AIROHA_USB_PHY_IDDIG;
+		fallthrough;
+	case PHY_MODE_USB_HOST:
+		val |= AIROHA_USB_PHY_FORCE_IDDIG;
+		break;
+	case PHY_MODE_USB_OTG:
+		break;
+	default:
+		return 0;
+	}
+
+	regmap_update_bits(priv->regs[AIROHA_USB_PHY_PHY],
+			   AIROHA_USB_PHY_U2PHYDTM1 + instance->offset,
+			   AIROHA_USB_PHY_FORCE_IDDIG |
+			   AIROHA_USB_PHY_IDDIG, val);
```


---

target/linux/airoha/patches-6.18/607-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch
```diff
@@ -0,0 +1,1691 @@
...
++{
++	const struct an7583_usb_phy_match_data *data;
++	struct an7583_usb_phy_instance  *instance;
++	struct phy *phy;
++	int ret;
++
++	data = of_device_get_match_data(dev);
++
++	if (index > data->max_instance)
++		return -EINVAL;
```
openwrt-ai: Off-by-one: `priv->phys[]` has `AIROHA_USB_PHY_MAX_INSTANCE` (3) entries and `.max_instance` is set to that same value, so `index == 3` passes this check and the `priv->phys[index] = phy` below writes one past the end of the array. `an7583_usb_phy_probe()` increments `port` for every child node with no upper bound, so a fourth subnode in a board DT is enough to trigger it. `an7583_usb_phy_xlate()` already uses the correct bound (`index < data->max_instance`).

The suggestion also drops the stray double space in the `instance` declaration.

```suggestion
+	struct an7583_usb_phy_instance *instance;
+	struct phy *phy;
+	int ret;
+
+	data = of_device_get_match_data(dev);
+
+	if (index >= data->max_instance)
+		return -EINVAL;
```


---

target/linux/airoha/patches-6.18/607-02-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch
```diff
@@ -0,0 +1,1691 @@
...
++		break;
++	case PHY_TYPE_USB3:
++		if (phy_get_mode(phy) == PHY_MODE_ETHERNET)
++			return 0;
++
++		an7583_usb_phy_u3_init(phy);
++		break;
++	default:
++		return -EINVAL;
++	}
```
openwrt-ai, line 1204: The return values of the two init helpers are discarded, so `phy_init()` always reports success. `an7583_usb_phy_u3_init()` deliberately ends with `return !an7583_usb_phy_u3_kband_is_calibrated(priv) ? -EINVAL : 0;`, i.e. it reports failure when the U3 KBand PLL is still uncalibrated after `AIROHA_USB_PHY_U3_MAX_CALIB_TRY` retries. As written that error is swallowed and xhci-mtk brings the SuperSpeed port up on an uncalibrated PLL with no diagnostic.

```suggestion
+	switch (instance->type) {
+	case PHY_TYPE_USB2:
+		return an7583_usb_phy_u2_init(phy);
+	case PHY_TYPE_USB3:
+		if (phy_get_mode(phy) == PHY_MODE_ETHERNET)
+			return 0;
+
+		return an7583_usb_phy_u3_init(phy);
+	default:
+		return -EINVAL;
+	}
```


---

target/linux/airoha/patches-6.18/607-02-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch
```diff
@@ -0,0 +1,1691 @@
+From 97de6398f93dbe0007674cd2699a29e055cf79f7 Mon Sep 17 00:00:00 2001
+From: Christian Marangi <ansuelsmth@gmail.com>
+Date: Fri, 19 Sep 2025 06:32:15 +0200
+Subject: [PATCH] phy: airoha: Add support for Airoha AN7583 USB PHY
```
openwrt-ai, line 4: nit: this adds the `airoha,an7583-usb-phy` compatible with no matching binding document. The AN7581 equivalent ships one — `target/linux/airoha/patches-6.18/220-05-dt-bindings-phy-Add-documentation-for-Airoha-AN7581-.patch` — and that is also where `AIROHA_USB2_MONCLK_SEL*` is defined, the header the new `an7583.dtsi` now includes. The AN7583 binding is not just "same as AN7581": it has four named reg regions (`phy`/`ana`/`pma`/`dig`), per-port child nodes with `#phy-cells`, and `airoha,scu` moved onto the U3 child. Was the binding patch intentionally left for a follow-up, or dropped along with the PCIe PHY part?

---

openwrt-ai (commented):
Reviewed 4 commits (fresh review, whole series). Commit messages match their diffs. Findings are inline: two in the carried PHY driver (missing per-instance offset in `u2_set_mode()`, off-by-one bound in `parse_subnode()`), one in the SoC dtsi (`reg` sizes smaller than the registers the driver accesses), plus a discarded init error path and a nit about the missing DT binding.

---

target/linux/airoha/dts/an7583.dtsi
VitaliySochniy: Confirmed, applying the suggestion. I checked both claims against the driver:

- `phy`: `AIROHA_USB_PHY_U2PHYDTM1` is `0x36c` and the second U2 instance sits at offset `0x1000`, so accesses reach `0x136c`, past the `0x11ff` end.
- `pma`: `AIROHA_USB_PMA_RX_CTRL_50` is `0x6f0`, past the `0x3ff` end, and it has two live call sites — `regmap_clear_bits()` in the U3 init path and `regmap_update_bits()` in the calibration path.

`ana` (max `0x40`) and `dig` (max `0x340`) do fit, but I'm taking the page-sized form for all four as suggested, matching `usb0_phy` on an7581 and the `pcs_ana`/`pcs_pma` regions in this same file.

Re-tested on hardware afterwards: the driver still binds to `1fac0000.phy`, which confirms all four regions map without a `request_mem_region` conflict.

csharper2005:
@VitaliySochniy thanks to continuing work on this device. Official USB support is a very important milestone for end users. Currently, some of them have to use third-party community builds.

> Unrelated to this series, but you may want to know: the MF logs a `_phy_state_machine` warning with `_phy_start_aneg: returned -22` from the EN8811H on lan1. It reproduces on builds predating this series, so I left it alone.

I can confirm this. This code used to work and still works in version 25.12:
```
[    1.518378] airoha_eth 1fb50000.ethernet eth0: entered promiscuous mode
[    3.388724] airoha-npu 1e900000.npu: NPU fw version: 1456.62
[   11.666864] airoha_eth 1fb50000.ethernet lan1: renamed from eth1
[   14.810990] airoha_eth 1fb50000.ethernet lan1: PHY [1fb00000.system-controller:mdio-bus@c8-mii:0f] driver [Airoha EN8811H] (irq=POLL)
[   14.823038] airoha_eth 1fb50000.ethernet lan1: configuring for phy/2500base-x link mode
[   33.216244] airoha_eth 1fb50000.ethernet lan1: PHY [1fb00000.system-controller:mdio-bus@c8-mii:0f] driver [Airoha EN8811H] (irq=POLL)
[   33.228357] airoha_eth 1fb50000.ethernet lan1: configuring for phy/2500base-x link mode
[   33.267286] airoha_eth 1fb50000.ethernet lan1: entered allmulticast mode
[   33.274330] airoha_eth 1fb50000.ethernet lan1: entered promiscuous mode
[   33.350995] airoha_eth 1fb50000.ethernet eth0: entered allmulticast mode
[  238.152381] airoha_eth 1fb50000.ethernet lan1: Link is Up - 1Gbps/Full - flow control rx/tx
```
A regression has likely occurred somewhere.

And one more topic regarding Nokia XG-040G-MF support: OpenWrt U-Boot. `GD5F2GM7UEYIG` SPI-NAND is absent in the PLAT. We should wait until the PR is merged - https://github.com/openwrt/openwrt/pull/24257. After that, we will be able to add `GD5F2GM7UEYIG`.

target/linux/airoha/patches-6.18/607-02-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch
VitaliySochniy: Confirmed, thanks. `_u2_init()`, `_u2_power_on()`, `_u2_power_off()` and `_u2_slew_rate_calibration()` each add `instance->offset`; `_u2_set_mode()` is the only one that doesn't.

It is latent on this board — `FORCE_IDDIG`/`IDDIG` are host-mode at reset and the XG-040G-MF is host-only, so the misdirected write lands on values that were already correct. Fixed, and re-tested with a device on the second U2 port (the one at offset `0x1000`, where the fix actually changes the address): enumerates at 480M as before.

target/linux/airoha/patches-6.18/607-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch
VitaliySochniy: Confirmed. `phys[]` is `struct phy *phys[AIROHA_USB_PHY_MAX_INSTANCE]` with `AIROHA_USB_PHY_MAX_INSTANCE == 3`, `.max_instance` is set to the same value, and `an7583_usb_phy_probe()` does `port++` for every child with no upper bound, so a fourth subnode reaches `priv->phys[3]`. `an7583_usb_phy_xlate()` uses `index < data->max_instance`, so the two bounds disagree.

Not reachable from any in-tree DT today — AN7583 has three PHYs and the SoC dtsi declares exactly three subnodes — but it is a real out-of-bounds write on a malformed board DT. Taking the fix and the whitespace cleanup.

target/linux/airoha/patches-6.18/607-02-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch
VitaliySochniy: Confirmed. `an7583_usb_phy_u3_init()` ends with

```c
return !an7583_usb_phy_u3_kband_is_calibrated(priv) ? -EINVAL : 0;
```

after retrying `AIROHA_USB_PHY_U3_MAX_CALIB_TRY` times, and `an7583_usb_phy_init()` drops it. Taking the suggestion, and also removing the `return 0;` that follows the switch — every arm including `default` now returns, so it became unreachable.

On the XG-040G-MF the calibration does succeed, so this changes nothing observable here — re-tested after the fix and SuperSpeed still links at 5 Gbps. It matters on a board where it doesn't, where the failure is currently silent.

target/linux/airoha/patches-6.18/607-02-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch
VitaliySochniy: Oversight rather than a deliberate deferral — I carried the driver patch and missed that the binding is a separate patch in the AN7581 series (`220-05`), which is also where `include/dt-bindings/phy/airoha,an7581-usb-phy.h` comes from, the header the new `an7583.dtsi` includes.

VitaliySochniy:
Thanks for confirming — I chased this down on my unit and the root cause turned out to be in the device tree, not in the PHY driver or the MDIO controller.

The PHY node is declared as:

```dts
compatible = "ethernet-phy-ieee802.3-c45";
```

In `fwnode_mdiobus_register_phy()` (`drivers/net/mdio/fwnode_mdio.c`) that forces the ID to be read from the hardware during the bus scan:

```c
is_c45 = fwnode_device_is_compatible(child, "ethernet-phy-ieee802.3-c45");
if (is_c45 || fwnode_get_phy_id(child, &phy_id))
	phy = get_phy_device(bus, addr, is_c45);   /* reads the ID off the wire */
else
	phy = phy_device_create(bus, addr, phy_id, 0, NULL);  /* ID from DT */
```

The EN8811H is not answering yet at that point, so the ID comes back as `0x00000000`. `air_en8811h` matches `PHY_ID_MATCH_MODEL(0x03a2a411)` and cannot match zero, the generic Clause 45 driver binds instead, `genphy_c45_pma_setup_forced()` returns `-EINVAL` for `2500base-x`, and `_phy_state_machine()` splats. lan1 never comes up.

The PHY itself is fine — it just answers later than it is probed. With mdio-tools on a running system it reports the correct ID over both C22 and C45:

```
mdio <bus> -> 0x0f              0x03a2a411
mdio <bus> phy 0x0f 2 / 3       0x03a2 / 0xa411     (C22)
mdio <bus> mmd 0x0f:1 2 / 3     0x03a2 / 0xa411     (C45 DEVID)
```

Also checked and ruled out: reset is deasserted (`gpio-34 PHY reset out hi ACTIVE LOW`), the firmware files are in place (`/lib/firmware/airoha/EthMD32.*`), and the MDIO driver implements both `bus->read` and `bus->read_c45`.

The fix is one line in the board DTS:

```diff
-		compatible = "ethernet-phy-ieee802.3-c45";
+		compatible = "ethernet-phy-id03a2.a411";
```

Note that simply dropping the c45 compatible is not enough — without an id-compatible, phylib still goes and reads the ID from the hardware.

Verified on a flashed board (not initramfs):

```
lan1: PHY [...mii:0f] driver [Airoha EN8811H]
MD32 firmware version: 25062302
phy_id = 0x03a2a411,  air_en8811h refcount 1
_phy_start_aneg warnings: 0,  Supports auto-negotiation: Yes
```

No autoneg patch is needed on top: `en8811h_probe()` loads the firmware and sets `phydev->c45_ids.mmds_present |= MDIO_DEVS_PMAPMD | MDIO_DEVS_AN` itself, so `get_features` sees the AN MMD and the `Autoneg` bit comes up on its own.

I have this ready as a separate commit and will send it as its own PR, since it is unrelated to the USB series.

openwrt-ai (commented):

The force-push since `d042e9e` touches only `target/linux/airoha/dts/an7583.dtsi` and the `607-*` patch, and resolves all four inline findings from the previous review:

- `reg` regions are now page-sized (`0x2000`/`0x1000`×3); `0x2000` covers the highest PHY register the driver touches (`AIROHA_USB_PHY_U3_PHYA_DA_REG19` at `0xc38`, and `AIROHA_USB_PHY_U2PHYDTM1 + 0x1000` = `0x136c`), and none of the four ranges overlap `pon_pcs` (`0x1fa8xxxx`) or `eth_pcs` (`0x1fa7xxxx`) in the same dtsi.
- `an7583_usb_phy_init()` now propagates both helpers' return values, with the dead trailing `return 0;` removed.
- `an7583_usb_phy_u2_set_mode()` adds `instance->offset`.
- `an7583_usb_phy_parse_subnode()` bounds `index` with `>=`.

Verified the patch is still well-formed and applies: the new-file hunk header `@@ -0,0 +1,1650 @@` matches the 1650 added lines, and `607-*` applies without fuzz on v6.18.41 with the airoha `220-06`/`220-07` prerequisites in place. CI on `d318f37` is green (no failed check runs).

The open thread about the missing `dt-bindings` patch for `airoha,an7583-usb-phy` still stands — acknowledged as an oversight but not yet addressed in this head.

---

target/linux/airoha/patches-6.18/607-01-dt-bindings-phy-Add-documentation-for-Airoha-AN7583-USB-PHY.patch
```diff
@@ -0,0 +1,163 @@
...
++    };
+--- a/MAINTAINERS
++++ b/MAINTAINERS
+@@ -776,6 +776,7 @@ M:	Christian Marangi <ansuelsmth@gmail.c
+ L:	linux-arm-kernel@lists.infradead.org (moderated for non-subscribers)
+ S:	Maintained
+ F:	Documentation/devicetree/bindings/phy/airoha,an7581-usb-phy.yaml
++F:	Documentation/devicetree/bindings/phy/airoha,an7583-usb-phy.yaml
+ F:	drivers/phy/airoha/phy-airoha-usb.c
+ F:	include/dt-bindings/phy/airoha,an7581-usb-phy.h
```
openwrt-ai, line 162: The new binding gets a `F:` entry, but the driver added by `607-*` does not. That patch creates `drivers/phy/airoha/phy-airoha-an7583-usb.c` and touches only `drivers/phy/airoha/Kconfig` and `drivers/phy/airoha/Makefile` — it never updates MAINTAINERS. The `AIROHA USB PHY DRIVER` block added by [`220-05`](``https://github.com/openwrt/openwrt/blob/0751764c26e3a72a9c1d511a204f38ce85e1f3c4/target/linux/airoha/patches-6.18/220-05-dt-bindings-phy-Add-documentation-for-Airoha-AN7581-.patch#L60-L66)`` / `220-07` lists `drivers/phy/airoha/phy-airoha-usb.c` as an exact path, not a glob, so after this series `phy-airoha-an7583-usb.c` is matched by no `F:` line and `get_maintainer.pl` returns nothing for it.

Adding it here keeps the MAINTAINERS churn in one patch (alphabetically it sorts before `phy-airoha-usb.c`); moving it into `607-*` instead is equally fine and is what upstream would expect, since that is the patch introducing the file.

```suggestion
@@ -776,6 +776,8 @@ M:	Christian Marangi <ansuelsmth@gmail.c
 L:	linux-arm-kernel@lists.infradead.org (moderated for non-subscribers)
 S:	Maintained
 F:	Documentation/devicetree/bindings/phy/airoha,an7581-usb-phy.yaml
+F:	Documentation/devicetree/bindings/phy/airoha,an7583-usb-phy.yaml
+F:	drivers/phy/airoha/phy-airoha-an7583-usb.c
 F:	drivers/phy/airoha/phy-airoha-usb.c
 F:	include/dt-bindings/phy/airoha,an7581-usb-phy.h
```


---

target/linux/airoha/patches-6.18/607-01-dt-bindings-phy-Add-documentation-for-Airoha-AN7583-USB-PHY.patch
```diff
@@ -0,0 +1,163 @@
+From: Vitaliy Sochnev <sochnev.v.74@gmail.com>
+Date: Sat, 8 Aug 2026 10:00:00 +0100
+Subject: [PATCH] dt-bindings: phy: Add documentation for Airoha AN7583 USB PHY
```
openwrt-ai, line 3: nit: the binding is numbered `608-*`, so it is applied *after* the driver in `607-*`. The AN7581 equivalent in this same target does it the other way round — [`220-05-dt-bindings-phy-Add-documentation-for-Airoha-AN7581-.patch`](``https://github.com/openwrt/openwrt/blob/0751764c26e3a72a9c1d511a204f38ce85e1f3c4/target/linux/airoha/patches-6.18/220-05-dt-bindings-phy-Add-documentation-for-Airoha-AN7581-.patch)`` precedes [`220-07-phy-airoha-Add-support-for-Airoha-AN7581-USB-PHY.patch`](``https://github.com/openwrt/openwrt/blob/0751764c26e3a72a9c1d511a204f38ce85e1f3c4/target/linux/airoha/patches-6.18/220-07-phy-airoha-Add-support-for-Airoha-AN7581-USB-PHY.patch)`` — and the kernel submission convention is bindings-before-driver too, which matters because you intend to send this upstream. Renumbering the pair (e.g. `607-01-dt-bindings-...` / `607-02-phy-airoha-...`) would keep the local order matching the eventual upstream series. Nothing breaks as-is; the ordering only affects how the series reads.

---

openwrt-ai (commented):
Reviewed 1 new commit since `d318f37` (`0751764`, "airoha: an7583: add DT binding for AN7583 USB PHY"). The commit message matches its diff.

The new `608-*` binding was checked against the driver in `607-*` and the node in `target/linux/airoha/dts/an7583.dtsi`, and it lines up: `reg`/`reg-names` match the four `devm_platform_ioremap_resource_byname()` lookups (`phy`/`ana`/`pma`/`dig`), the `^usb2-phy@[0-9a-f]+$` children match `usb2-phy@0` / `usb2-phy@1000` and require exactly the two properties `an7583_usb_phy_parse_subnode()` reads (`reg`, `airoha,usb2-monitor-clk-sel`), and `usb3-phy` requires `airoha,scu`, which is what the driver keys the `PHY_TYPE_USB3` branch on. `#phy-cells = <1>` on the children is right — the provider is registered on the parent and `an7583_usb_phy_xlate()` resolves by child `of_node` with the single cell used as the PHY type. The YAML parses, and the MAINTAINERS hunk applies cleanly on top of `220-05`/`220-07` against v6.18.41 (same `-8` offset as its siblings, no fuzz).

Two findings inline: the driver file added by `607-*` ends up with no `F:` entry, plus a nit about the binding patch being numbered after the driver patch.

---

target/linux/airoha/patches-6.18/607-01-dt-bindings-phy-Add-documentation-for-Airoha-AN7583-USB-PHY.patch
VitaliySochniy: Confirmed — `607-*` touches only `Kconfig`, `Makefile` and the new `.c`, so `drivers/phy/airoha/phy-airoha-an7583-usb.c` is matched by no `F:` line.

Taking the second option and putting it in the driver patch rather than here, since that is what the sibling does: `220-07-phy-airoha-Add-support-for-Airoha-AN7581-USB-PHY.patch` adds `F: drivers/phy/airoha/phy-airoha-usb.c` in the same patch that creates the file. That keeps the local series consistent with upstream convention and with what a later submission would look like.

target/linux/airoha/patches-6.18/607-01-dt-bindings-phy-Add-documentation-for-Airoha-AN7583-USB-PHY.patch
VitaliySochniy: Good catch, renumbering. The pair becomes `607-01-dt-bindings-phy-Add-documentation-for-Airoha-AN7583-USB-PHY.patch` and `607-02-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch`, so the binding is applied first, matching `220-05` -> `220-07` for AN7581 and the bindings-before-driver order this would be submitted in upstream.

target/linux/airoha/patches-6.18/607-02-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch
```diff
@@ -0,0 +1,1699 @@
...
++
++MODULE_DESCRIPTION("Airoha AN7583 USB PHY driver");
++MODULE_AUTHOR("Christian Marangi <ansuelsmth@gmail.com>");
++MODULE_LICENSE("GPL");
+--- a/MAINTAINERS
++++ b/MAINTAINERS
+@@ -777,6 +777,7 @@ L:	linux-arm-kernel@lists.infradead.org
+ S:	Maintained
+ F:	Documentation/devicetree/bindings/phy/airoha,an7581-usb-phy.yaml
+ F:	Documentation/devicetree/bindings/phy/airoha,an7583-usb-phy.yaml
```
openwrt-ai: This new MAINTAINERS hunk carries `F: Documentation/devicetree/bindings/phy/airoha,an7583-usb-phy.yaml` as a **context** line, but that line is only introduced by [`607-01-*` (its own MAINTAINERS hunk)](``https://github.com/openwrt/openwrt/blob/6e87ba7fb2022ba4f520d7c29b70f539bddec9a0/target/linux/airoha/patches-6.18/607-01-dt-bindings-phy-Add-documentation-for-Airoha-AN7583-USB-PHY.patch#L154-L162),`` which this series adds two commits later (`6e87ba7`). Within a full checkout of the PR head the order `607-01` -> `607-02` makes it apply, but at commit `663b0a5` ("airoha: add support for AN7583 USB PHY") only `607-02` exists, and it fails:

```
$ patch -p1 < 607-02-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch
patching file MAINTAINERS
Hunk #1 FAILED at 777.
1 out of 1 hunk FAILED -- saving rejects to file MAINTAINERS.rej
```

(reproduced on v6.18.41 with `180-01`/`200-02`/`220-05`/`220-07` applied first; the missing context line sits mid-hunk, so the default fuzz of 2 does not cover it, and kernel patches are applied with a bare `patch -f -p1` in [scripts/patch-kernel.sh:40](https://github.com/openwrt/openwrt/blob/6e87ba7fb2022ba4f520d7c29b70f539bddec9a0/scripts/patch-kernel.sh#L40)).

So the intermediate commit no longer builds — `git bisect` and per-commit CI break on it. Reordering the series so the binding commit comes before the driver commit fixes it, and matches both the new patch numbering (`607-01` before `607-02`) and the bindings-before-driver order you're aiming for upstream. Squashing the two commits works too.

---

target/linux/airoha/patches-6.18/607-02-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch
```diff
@@ -0,0 +1,1699 @@
...
++		.of_match_table = airoha_phy_id_table,
++	},
++};
++
++module_platform_driver(an7583_usb_driver);
++
++MODULE_DESCRIPTION("Airoha AN7583 USB PHY driver");
++MODULE_AUTHOR("Christian Marangi <ansuelsmth@gmail.com>");
++MODULE_LICENSE("GPL");
+--- a/MAINTAINERS
```
openwrt-ai: nit: the MAINTAINERS hunk is appended at the end, after `drivers/phy/airoha/phy-airoha-an7583-usb.c`. `git format-patch` emits files in path order, where `MAINTAINERS` sorts before `drivers/...`, which is what the sibling [`220-07-phy-airoha-Add-support-for-Airoha-AN7581-USB-PHY.patch`](``https://github.com/openwrt/openwrt/blob/6e87ba7fb2022ba4f520d7c29b70f539bddec9a0/target/linux/airoha/patches-6.18/220-07-phy-airoha-Add-support-for-Airoha-AN7581-USB-PHY.patch#L36)`` does (MAINTAINERS first, then Kconfig/Makefile/.c), and what `607-01` does (Documentation, then MAINTAINERS). It applies fine either way; it just means the file won't come back identical when you regenerate it from a real commit for the upstream submission.

---

openwrt-ai (commented):
Reviewed the delta since `0751764`. The series was re-pushed with new SHAs, but `git range-diff` shows the only content change is the one agreed in the previous round: both patches renumbered (`608-*` -> `607-01-*`, `607-*` -> `607-02-*`) and the `F: drivers/phy/airoha/phy-airoha-an7583-usb.c` entry added to the driver patch, squashed into the commit that introduces it (`663b0a5`) rather than into the binding commit. The other three commits are byte-identical to their predecessors, and both commit messages still match their diffs.

Simulated the full MAINTAINERS chain against v6.18.41 (`180-01` -> `200-02` -> `220-05` -> `220-07` -> `607-01` -> `607-02`): all hunks apply without fuzz and the resulting `AIROHA USB PHY DRIVER` block stays alphabetically sorted, with the new `.c` between the two yaml entries and `phy-airoha-usb.c`. No stale references to the old patch file names remain, and `607-0{1,2}` don't collide with any existing number in that directory. CI on `6e87ba7` shows no failed check runs.

One finding inline: with the MAINTAINERS entry in `607-02` depending on context that `607-01` adds, the driver commit now precedes the binding commit it needs, so the tree doesn't apply at that intermediate commit. Plus a nit about hunk placement inside `607-02`.

---

target/linux/airoha/patches-6.18/607-02-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch
VitaliySochniy: Confirmed and reproduced. On a v6.18.41 tree with `180-01`/`200-02`/`220-05`/`220-07` applied but not `607-01`:

```
checking file MAINTAINERS
Hunk #1 FAILED at 777.
1 out of 1 hunk FAILED
```

The context line the hunk needs is added by `607-01`, so at the driver commit alone the tree does not apply.

Fixed by reordering the series so the binding commit comes first, which also matches the `607-01` -> `607-02` numbering and the bindings-before-driver order this would be submitted in upstream. I kept them as separate commits rather than squashing, since they are logically distinct and that is how the AN7581 pair is carried.

target/linux/airoha/patches-6.18/607-02-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch
VitaliySochniy: Fixed — the MAINTAINERS diff now comes first, matching `git format-patch` path order and what `220-07` and `607-01` do.

target/linux/airoha/an7583/config-6.18
```diff
@@ -359,6 +360,7 @@ CONFIG_TREE_RCU=y
 CONFIG_TREE_SRCU=y
 CONFIG_UBIFS_FS=y
 # CONFIG_UNMAP_KERNEL_AT_EL0 is not set
+CONFIG_USB_SUPPORT=y
```
danielschwierzeck, line 373: although it is inconsistently used by various OpenWRT targets, but maybe use  `FEATURES:=... usb` in `target/linux/airoha/Makefile` and drop `CONFIG_USB_SUPPORT` from all kernel default configs?

robimarko: Please dont, just enable what you need directly in the kernel config, and then the usb feature will get added automatically after parsing the kernel config

danielschwierzeck:
I can confirm that at least the USB PHY is working on my board. I don't have USB ports but a EN8811 PHY connected to GDM4 via USB. With the multi-serdes PCS patches already in mainline, I just needed to add following DTS fragments to get that 2.5G LAN port working:

```
--- a/target/linux/airoha/dts/an7583.dtsi
+++ b/target/linux/airoha/dts/an7583.dtsi
@@ -657,6 +657,23 @@
            #pcs-cells = <0>;
        };
 
+       usb_pcs: pcs@1fa05000 {
+           compatible = "airoha,an7583-pcs-usb";
+           reg = <0x0 0x1fa05000 0x0 0x1000>,
+                 <0x0 0x1fa60000 0x0 0x60>,
+                 <0x0 0x1fa60a00 0x0 0x164>,
+                 <0x0 0x1fa64000 0x0 0x1000>,
+                 <0x0 0x1fa66000 0x0 0x300>;
+           reg-names = "pcs_mac", "hsgmii_an", "hsgmii_pcs",
+                   "multi_sgmii", "hsgmii_rate_adp";
+
+           phys = <&usb0_u3 PHY_TYPE_USB3>;
+
+           #pcs-cells = <0>;
+
+           status = "disabled";
+       };
+
        eth: ethernet@1fb50000 {
            compatible = "airoha,an7583-eth";
            reg = <0 0x1fb50000 0 0x2600>,
@@ -724,6 +741,22 @@
 
                status = "disabled";
            };
+
+           gdm4: ethernet@4 {
+               compatible = "airoha,eth-mac";
+               reg = <4>;
+
+               #address-cells = <1>;
+               #size-cells = <0>;
+
+               gdm4_usb: ethernet-port@1 {
+                   compatible = "airoha,eth-port";
+                   reg = <1>;
+                   pcs-handle = <&usb_pcs>;
+
+                   status = "disabled";
+               };
+           };
        };
 
        switch: switch@1fb58000 {
```

board.dts:
```
&mdio_0 {
	pinctrl-names = "default";
	pinctrl-0 = <&mdio_pins>;

	clock-frequency = <12500000>;

	en8811: ethernet-phy@d {
		compatible = "ethernet-phy-ieee802.3-c22";
		reg = <0xd>;

		reset-assert-us = <10000>;
		reset-deassert-us = <20000>;
		reset-gpios = <&an7583_pinctrl 5 GPIO_ACTIVE_LOW>;
	};

	as21xx: ethernet-phy@1d {
		compatible = "ethernet-phy-ieee802.3-c45";
		reg = <0x1d>;

		firmware-name = "as21x1x_fw.bin";

		reset-deassert-us = <350000>;
		reset-assert-us = <100>;
		reset-gpios = <&an7583_pinctrl 28 GPIO_ACTIVE_LOW>;
	};
};

&usb_pcs {
	status = "okay";
};

&gdm4_usb {
	status = "okay";

	phy-handle = <&en8811>;
	phy-mode = "2500base-x";

	openwrt,netdev-name = "lan3";
};
```

Thanks!

target/linux/airoha/patches-6.18/607-01-dt-bindings-phy-Add-documentation-for-Airoha-AN7583-USB-PHY.patch
```diff
@@ -0,0 +1,163 @@
+From: Vitaliy Sochnev <sochnev.v.74@gmail.com>
```
openwrt-ai, line 2: nit: this patch starts straight at `From:` and is missing the `From <hash> Mon Sep 17 00:00:00 2001` line that `git format-patch` emits, so it isn't `git am`-able as-is. The [openwrt bot flagged exactly this](https://github.com/openwrt/openwrt/pull/24609#issuecomment-5222235946) and it's still unaddressed at this head.

Its sibling in the same numbered pair has it — [`607-02-*` line 1](``https://github.com/openwrt/openwrt/blob/b21d5afc4cf5a1b56a9bc4ec407590ab1dbd1b0e/target/linux/airoha/patches-6.18/607-02-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch#L1)`` is `From 97de6398f93dbe0007674cd2699a29e055cf79f7 Mon Sep 17 00:00:00 2001` — as does [`220-05-*`](``https://github.com/openwrt/openwrt/blob/b21d5afc4cf5a1b56a9bc4ec407590ab1dbd1b0e/target/linux/airoha/patches-6.18/220-05-dt-bindings-phy-Add-documentation-for-Airoha-AN7581-.patch#L1),`` the AN7581 binding this one mirrors. Of the 144 patches in `patches-6.18/`, only 7 lack the line, and the other 6 are pre-existing.

No suggestion block since the hash has to come from the commit you generate it from — regenerating this file with `git format-patch` off the local commit that adds the binding will produce the header (and the `[PATCH]` subject line) in one go.

---

openwrt-ai (commented):
Reviewed the delta since `6e87ba7`. The series was re-pushed with new SHAs, but `git range-diff` shows only two changes, both the ones agreed in the previous round: the binding commit was moved to the front of the series (`161fc2e` now precedes `5386180`), and the MAINTAINERS hunk inside `607-02-*` was moved from the end of the patch to the front, matching `git format-patch` path order. The other three commits are byte-identical to their predecessors, and all five commit messages still match their diffs.

Both prior findings verified as fixed against `~/extra/linux-v6.18.41` (`2fe5967`):

- The intermediate-commit apply failure is gone. `607-01-*` now applies standalone on top of `200-02`/`220-04`/`220-05`, so the tree is consistent at `161fc2e`; `607-02-*` then applies on top of it with no fuzz, and the resulting `AIROHA USB PHY DRIVER` block stays alphabetically sorted (`...an7581-usb-phy.yaml`, `...an7583-usb-phy.yaml`, `phy-airoha-an7583-usb.c`, `phy-airoha-usb.c`, `airoha,an7581-usb-phy.h`).
- The MAINTAINERS hunk in `607-02-*` is now the first diff in the file, ahead of `Kconfig`/`Makefile`/the new `.c`, same as [`220-07-*`](``https://github.com/openwrt/openwrt/blob/b21d5afc4cf5a1b56a9bc4ec407590ab1dbd1b0e/target/linux/airoha/patches-6.18/220-07-phy-airoha-Add-support-for-Airoha-AN7581-USB-PHY.patch#L36)`` and `607-01-*`.

CI on `b21d5af` is green across all three workflows (Build Kernel, Build all core packages, Build PR Profile).

One nit inline, on the `607-01-*` header — the only item from the openwrt bot's formality check that is still open.

---

VitaliySochniy:
@danielschwierzeck, thanks — that is a useful data point, and it settles a question I raised in the description.

Your fragment exercises the path I could not test here: on both Nokia boards the U3 lane is wired to a real USB 3.0 socket, so `PHY_MODE_ETHERNET` never gets used. Your board drives the same lane as HSGMII, which means the `phy_get_mode(phy) == PHY_MODE_ETHERNET` early return in `an7583_usb_phy_init()` and the `AIROHA_SCU_SSUSB_HSGMII_SEL` switch in `an7583_usb_phy_set_mode()` are confirmed working on hardware.

That also justifies keeping the shared U3/HSGMII serdes handling when the PCIe part was dropped — had it gone too, your port would not come up.

Worth noting that `phys = <&usb0_u3 PHY_TYPE_USB3>` relies on the `usb0_u3` label this series adds to `an7583.dtsi`, so this PR is a prerequisite for those fragments.

The `usb_pcs` / `gdm4` nodes are out of scope here, since this series is only about enabling USB. Do you plan to submit them separately? Having `usb_pcs` in the SoC dtsi would make the HSGMII use case reachable for other boards without local patches.

csharper2005 (approved):
LGTM. Run tested.

openwrt-ai (commented):
Reviewed 5 new commits; no new issues found.

The series was re-pushed with new SHAs, but `git range-diff` against `b21d5af` shows exactly one content change, and it is the one from the previous round: `607-01-*` now carries the `From b1810200bbd5021c1a00643ea5dd8516ad9c1c1b Mon Sep 17 00:00:00 2001` header line. The other four commits are byte-identical to their predecessors, and all five commit messages still match their diffs.

The prior finding is fixed. The patch is now `git am`-able like its siblings — `Date: Sat, 8 Aug 2026` has the correct weekday, the `Subject: [PATCH] ...` line is intact, and the header shape matches [`607-02-*`](``https://github.com/openwrt/openwrt/blob/ed3464893ec8a729a8a7f628e4d430798d81a473/target/linux/airoha/patches-6.18/607-02-phy-airoha-Add-support-for-Airoha-AN7583-USB-PHY.patch#L1)`` and [`220-05-*`](``https://github.com/openwrt/openwrt/blob/ed3464893ec8a729a8a7f628e4d430798d81a473/target/linux/airoha/patches-6.18/220-05-dt-bindings-phy-Add-documentation-for-Airoha-AN7581-.patch#L1).`` `FormalityCheck / Code Patches` is green on this head, where it previously carried this item.

Re-verified the apply against `~/extra/linux-v6.18.41` (`2fe5967`): `607-01-*` and `607-02-*` both apply with no fuzz and no offsets after the airoha prerequisites, the new file hunk header (`@@ -0,0 +1,132 @@`) matches the added lines, and the resulting `AIROHA USB PHY DRIVER` block stays alphabetically sorted:

```
F:	Documentation/devicetree/bindings/phy/airoha,an7581-usb-phy.yaml
F:	Documentation/devicetree/bindings/phy/airoha,an7583-usb-phy.yaml
F:	drivers/phy/airoha/phy-airoha-an7583-usb.c
F:	drivers/phy/airoha/phy-airoha-usb.c
F:	include/dt-bindings/phy/airoha,an7581-usb-phy.h
```

CI on `ed34648` has no failed check runs across the three workflows (Build Kernel, Build all core packages, Build PR Profile) plus FormalityCheck. No open findings from previous rounds remain.

---

robimarko:
I am not a fan of the roundabout way needed to actually enable the required drivers here.
Instead of the whole mess with adding just CONFIG_USB_SUPPORT to the subtarget config, and then for some reason depending on kmod-usb3 per board to select the xhci-mtk why not just build the USB support and driver in?

That has been working for me for couple of weeks on AN7583 and avoids this whole roundabout way

VitaliySochniy:
@robimarko, sure. Two things before I redo it:

1. an7581 uses the same pattern (`+TARGET_airoha_an7581:kmod-usb-xhci-mtk`, `kmod-usb3` in
   MD/valyrian `DEVICE_PACKAGES`). Convert it here too, follow-up, or leave it?

2. Built-in pulls xHCI into the EVBs as well — fine by you?

`kmod-usb-ledtrig-usbport` and `kmod-regulator-userspace-consumer` stay, they're not xHCI.

If you have the config diff you've been running, I'll take it.

robimarko:
1. Leave it for now
2. It wont hurt boards that dont ship with USB ports, its not like they are space restrained.

This is what I have been using for some time, its just the required stuff selected:
https://github.com/robimarko/openwrt/commit/2868793a2b64f7db0f898da589899e2310130887

VitaliySochniy:
@robimarko, done — took your config verbatim, dropped the usb.mk commit and kmod-usb3 from the board. Left an7581 alone for now.

Retested on the XG-040G-MF: xHCI now comes up at 1.34s instead of after rootfs is mounted, both U2 ports and the SuperSpeed one enumerate (480M / 5000M), the 5V rail and the usbport LEDs behave as before. `make kernel_oldconfig` reproduces config-6.18 unchanged.

I left CONFIG_PCS_MTK_LYNXI out of it — unrelated to USB, and oldconfig doesn't ask for it here.

target/linux/airoha/an7583/config-6.18
```diff
@@ -359,6 +360,12 @@ CONFIG_TREE_RCU=y
 CONFIG_TREE_SRCU=y
 CONFIG_UBIFS_FS=y
 # CONFIG_UNMAP_KERNEL_AT_EL0 is not set
+CONFIG_USB=y
```
openwrt-ai, line 371: nit: the PR description no longer matches the series. Its **Commits** table still lists `kernel: usb3: select xhci-mtk on airoha/an7583` / "kmod-usb3 dependency", which was dropped in this push, and the intro still says the series "wires up the kmod dependency" — it now builds xHCI + glue into the kernel instead. The table also never picked up the `airoha: an7583: add DT binding for AN7583 USB PHY` commit added a few rounds ago, so it lists 4 of the 5 commits. Worth a refresh so the description matches what a reviewer sees.

---

openwrt-ai (commented):
Reviewed 5 new commits; no new issues found.

The series was re-pushed with new SHAs. `git range-diff` against `ed34648` shows exactly the rework agreed with @robimarko in the thread, and nothing else:

- `kernel: usb3: select xhci-mtk on airoha/an7583` is gone — `package/kernel/linux/modules/usb.mk` is untouched now.
- New commit `837b049` "airoha: an7583: build USB support into the kernel" adds the six symbols to `target/linux/airoha/an7583/config-6.18`.
- `CONFIG_USB_SUPPORT=y` moved out of the PHY commit (`505ead4`) into that new commit, with the commit message updated to match.
- `kmod-usb3` dropped from `DEVICE_PACKAGES` in `3ff51b8`, message updated to "the LED trigger and regulator kmods".

The other two commits are byte-identical to their predecessors, and all five commit messages match their diffs.

Checks on the new config, all clean:

- The six lines are exactly the set [`mediatek/mt7629` carries](https://github.com/openwrt/openwrt/blob/3ff51b80f20e0f76619848815c93ba7be6f1fed5/target/linux/mediatek/mt7629/config-6.18#L347-L352) for the same built-in xhci-mtk arrangement, in the same order, and they sort correctly between `# CONFIG_UNMAP_KERNEL_AT_EL0 is not set` and `CONFIG_USER_STACKTRACE_SUPPORT=y`.
- `USB_XHCI_MTK` is reachable on this SoC: its `depends on (MIPS && SOC_MT7621) || ARCH_MEDIATEK || COMPILE_TEST` in v6.18.41 is widened by the in-tree [`220-08-usb-host-add-ARCH_AIROHA-in-XHCI-MTK-dependency.patch`](``https://github.com/openwrt/openwrt/blob/3ff51b80f20e0f76619848815c93ba7be6f1fed5/target/linux/airoha/patches-6.18/220-08-usb-host-add-ARCH_AIROHA-in-XHCI-MTK-dependency.patch),`` and its `select MFD_SYSCON` is already satisfied (`CONFIG_MFD_SYSCON=y` at line 203).
- No symbol is left dangling by turning `CONFIG_USB=y`. Everything newly visible under `if USB` / `if USB_XHCI_HCD` is either already pinned in `target/linux/generic/config-6.18` (`USB_ANNOUNCE_NEW_DEVICES`, `USB_DEFAULT_PERSIST`, `USB_DYNAMIC_MINORS`, `USB_OTG`, `USB_AUTOSUSPEND_DELAY`, `USB_MON`, `USB_LED_TRIG`, `USB_LEDS_TRIGGER_USBPORT`, `USB_XHCI_DBGCAP`, `USB_XHCI_SIDEBAND`, `USB_XHCI_MVEBU`, `USB_XHCI_PCI_RENESAS`), arch-gated (`HISTB`/`RCAR`/`RZV2M`/`TEGRA`), invisible behind `# CONFIG_USB_PCI is not set`, or added here (`USB_XHCI_PLATFORM`).
- Dropping `kmod-usb3` and the `usb.mk` hunk doesn't break the kmod set on this subtarget: `KernelPackage` handles built-in symbols via the `modules.builtin` branch in [`include/kernel.mk:246`](https://github.com/openwrt/openwrt/blob/3ff51b80f20e0f76619848815c93ba7be6f1fed5/include/kernel.mk#L246), so `kmod-usb-xhci-hcd` / `kmod-usb-xhci-mtk` / `kmod-usb3` just become empty packages rather than failing on a missing `.ko`. `kmod-usb-ledtrig-usbport` and `kmod-regulator-userspace-consumer` stay real modules and are still in `DEVICE_PACKAGES`.
- Moving `CONFIG_USB_SUPPORT=y` out of `505ead4` leaves that intermediate commit self-consistent: `PHY_AIROHA_AN7583_USB` is `depends on ARCH_AIROHA || COMPILE_TEST` + `depends on OF` in the Kconfig hunk of `607-02-*`, with no USB dependency, so `CONFIG_PHY_AIROHA_AN7583_USB=y` is valid there without `USB_SUPPORT`.

Note the reworked push no longer touches `package/kernel/linux/modules/usb.mk`, so "Build all core packages" did not run this time; the `Build airoha/an7583` kernel job (which compiles the kmod packages) is green, as are `Build airoha/an7581`, `Check Kernel patches`, `build-pr-profile` and all three FormalityCheck jobs on `3ff51b8`. No failed check runs.

One nit inline about the PR description having gone stale.

---

robimarko:

PR #24610, opened 2026-08-08
IPQ8174-MX4300 Remove read only from safe partitions

wolfsilver00:
These 2 partitions were used by the device (in a previous iteration) for firmware for another company (fortinet).  That firmware is no longer in use and these 2 flags dont allow mounting mtd30 and 31 as rw, losing us about 500mb of storage.

More information on previous iteration here:
https://cdn.blueally.com/avfirewalls/datasheets/ds-linksys-homewrk.pdf

Ive tested using both mtd30 and 31 without any issues, and its a nice solution for a device that should have much more storage but its just locked behind non usable garbage

Signed-off-by: Ezequiel Scarcella scarcella.ezequiel@gmail.com

### DIFF
--- a/target/linux/qualcommax/dts/ipq8174-mx4300.dts
+++ b/target/linux/qualcommax/dts/ipq8174-mx4300.dts
@@ -244,13 +244,11 @@
 			partition@20c00000 {
 				label = "app2_data";
 				reg = <0x20c00000 0x16180000>;
-				read-only;
 			};
 
 			partition@36d80000 {
 				label = "app2";
 				reg = <0x36d80000 0x9280000>;
-				read-only;
 			};
 		};
 	};


target/linux/qualcommax/dts/ipq8174-mx4300.dts
```diff
@@ -244,13 +244,11 @@
 			partition@20c00000 {
 				label = "app2_data";
 				reg = <0x20c00000 0x16180000>;
```
openwrt-ai, line 246: The evidence offered for "these are unused" doesn't cover the MX4300. The linked datasheet is for the Linksys HomeWRK, which is a *separate* board in this tree, and its DTS doesn't even declare these regions — [`ipq8174-homewrk.dts` uses `compatible = "qcom,smem-part"`](https://github.com/openwrt/openwrt/blob/c1baddbc9362bff47b7af7e1c31c59a14c1854a6/target/linux/qualcommax/dts/ipq8174-homewrk.dts#L29-L30) and lets the bootloader supply the table, so it says nothing about what MX4300 stock firmware does with `app2`/`app2_data`.

This matters because `read-only` here is not recoverable at runtime (MTD clears `MTD_WRITEABLE` at partition creation), and once it's gone there is no in-tree way to restore the original contents: the MX4300 revert-to-stock path in [`platform.sh` (`linksys,mx4200v1|v2|mx4300` case)](``https://github.com/openwrt/openwrt/blob/c1baddbc9362bff47b7af7e1c31c59a14c1854a6/target/linux/qualcommax/ipq807x/base-files/lib/upgrade/platform.sh#L226-L232)`` only touches kernel/rootfs via `nand_do_upgrade`.

Can you back the claim with something MX4300-specific — e.g. a stock-firmware dump showing both regions erased/unreferenced, or a confirmed revert-to-stock after writing to them? Whatever the answer, it belongs in the commit message, since it is the entire basis for the change.

---

target/linux/qualcommax/dts/ipq8174-mx4300.dts
```diff
@@ -244,13 +244,11 @@
...
-				read-only;
 			};
 
 			partition@36d80000 {
 				label = "app2";
 				reg = <0x36d80000 0x9280000>;
```
openwrt-ai, line 251: nit: the commit message identifies these as "mtd30 and 31", but MTD numbering is assigned at probe time and is not a stable API — it shifts as soon as another MTD device is registered before this one. Please refer to the partitions by label (`app2_data` / `app2`) in the message instead.

Related: nothing else in the tree consumes these regions — the only occurrences of `app2` under `target/linux/` are these two `label` properties — so the reclaimed ~500 MiB is exposed as two bare MTD devices that the user still has to attach/format by hand. That's a fine enabling change, but worth stating explicitly in the commit message so it isn't read as "the space is now usable out of the box".

---

openwrt-ai (commented):
## Commit checks

- `c1baddb` "Remove read only from safe partitions" — the diff does what the message says, but the subject is missing the required area prefix. For a change to `target/linux/qualcommax/dts/ipq8174-mx4300.dts` it should read something like `qualcommax: mx4300: drop read-only from app2/app2_data`. The formality bot separately flags the GitHub noreply author/`Signed-off-by` address, the missing full author name, and three body lines over 100 chars — those all need fixing in the same amend.

Note the body/message also mixes the justification with editorial asides; keeping it to the factual claim (which regions, why they're unused, what was tested) will make it much easier to merge.

---

target/linux/qualcommax/dts/ipq8174-mx4300.dts
wolfsilver00: The evidence is not for the HomeWRK version on this repo, its for the base hardware that went unsold, then was resold to linksys. Linksys then put 2 different firmwares on them to sell them, but the hardware is the same, an app2 and app2_data are remains of that, one was the fortinet part (app2) and the other was a db, for the fortinet app. Both of which are unusable in the stock firmware  and of course in openwrt also.

I did revert to stock a couple times updating the firmware (it has dual boot and I still wanted the backup) and with those deleted, it worked perfectly. But as of 2 or 3 months ago I dont have stock anymore, both bootable options are openwrt, one to test new releases, one stable.

target/linux/qualcommax/dts/ipq8174-mx4300.dts
wolfsilver00: Yes, correct, I just did not think I had to explain that. But the main driver for this change is to be able to mount/extroot devices that have had their usb port break (or are using mesh in one of the various forks that permit it, but dont let you use the USB port).

Right now it works for me but I do have to I_want_a_brick on boot via script, which is not ideal for a long list of reasons

PR #24612, opened 2026-08-08
ath79: tplink: extract ath10k board.bin for TP-Link Deco M4R v1

artms:
Extract the device-specific board image from the ART/flash partition
and provide it as board.bin for ath10k.
    
With generic `board-2.bin`, the link speed fluctuates around 150-200 Mbps and
fails to reach the full 867 Mbps PHY rate even at close proximity, resulting
in real-world throughput capped at 60-80 Mbps.
    
Extracting `board.bin` directly from the device's ART partition
restores performance close to the vendor firmware, bringing throughput up
to 200-300 Mbps.

### DIFF
--- a/target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
+++ b/target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
@@ -11,11 +11,15 @@ case "$FIRMWARE" in
 	case $board in
 	dlink,dir-842-c1|\
 	dlink,dir-842-c2|\
-	dlink,dir-842-c3)
+	dlink,dir-842-c3|\
+	tplink,deco-m4r-v1|\
+	tplink,deco-m4r-v2)
 		caldata_extract "art" 0x5000 0x2f20
 		caldata_valid "202f" || caldata_extract "reserved" 0x15000 0x2f20
 		ln -sf /lib/firmware/ath10k/pre-cal-pci-0000\:00\:00.0.bin \
 			/lib/firmware/ath10k/QCA9888/hw2.0/board.bin
+		ln -sf /lib/firmware/ath10k/pre-cal-pci-0000\:00\:00.0.bin \
+			/lib/firmware/ath10k/QCA9888/hw2.0/board-2.bin
 		;;
 	esac
 	;;


target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
```diff
@@ -11,11 +11,15 @@ case "$FIRMWARE" in
...
 	dlink,dir-842-c2|\
-	dlink,dir-842-c3)
+	dlink,dir-842-c3|\
+	tplink,deco-m4r-v1|\
+	tplink,deco-m4r-v2)
 		caldata_extract "art" 0x5000 0x2f20
 		caldata_valid "202f" || caldata_extract "reserved" 0x15000 0x2f20
 		ln -sf /lib/firmware/ath10k/pre-cal-pci-0000\:00\:00.0.bin \
 			/lib/firmware/ath10k/QCA9888/hw2.0/board.bin
+		ln -sf /lib/firmware/ath10k/pre-cal-pci-0000\:00\:00.0.bin \
```
artms, line 21: As mentioned in PR description - ath10k drivers prefers board-2.bin instead of `board.bin` when loading

artms:
I'd like this change to be ported to stable 25.12 eventually

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
```diff
@@ -11,11 +11,15 @@ case "$FIRMWARE" in
 	case $board in
 	dlink,dir-842-c1|\
 	dlink,dir-842-c2|\
-	dlink,dir-842-c3)
+	dlink,dir-842-c3|\
+	tplink,deco-m4r-v1|\
```
artms, line 15: potentially this as well works with deco-s4r which just differs in shape but internals are the same, I do not have S4R to test though

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
```diff
@@ -11,11 +11,15 @@ case "$FIRMWARE" in
...
+	tplink,deco-m4r-v2)
```
openwrt-ai, line 16: `tplink,deco-m4r-v2` can never match: no device tree in the tree declares that compatible. The only Deco M4R in ath79 is [`compatible = "tplink,deco-m4r-v1", "qca,qca9563";`](https://github.com/openwrt/openwrt/blob/29034b57aad673447714b96ce62ce42a7b0394c4/target/linux/ath79/dts/qca9563_tplink_deco-m4r-v1.dts#L10), and the original support commit (`063e9047cc`, "ath79: add support for TP-Link Deco M4R v1 and v2") deliberately covers both hardware revisions with that single compatible and one `TPLINK_BOARD_ID := DECO-M4R-V1` image. So `board_name()` returns `tplink,deco-m4r-v1` on a v2 unit as well and this arm is dead.

```suggestion
	tplink,deco-m4r-v1)
```

---

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
```diff
@@ -11,11 +11,15 @@ case "$FIRMWARE" in
...
 		caldata_extract "art" 0x5000 0x2f20
 		caldata_valid "202f" || caldata_extract "reserved" 0x15000 0x2f20
```
openwrt-ai, line 18: This `reserved` fallback is DIR-842-specific and does not apply to the Deco M4R. `qca9563_dlink_dir-842-c.dtsi` defines a [`reserved` partition at 0xfe0000](https://github.com/openwrt/openwrt/blob/29034b57aad673447714b96ce62ce42a7b0394c4/target/linux/ath79/dts/qca9563_dlink_dir-842-c.dtsi#L119-L120); the Deco M4R v1 DTS has no such partition. If the ART blob at 0x5000 ever fails `caldata_valid`, `caldata_extract` will hit [`caldata_die "no mtd device found for partition $part"`](https://github.com/openwrt/openwrt/blob/29034b57aad673447714b96ce62ce42a7b0394c4/package/base-files/files/lib/functions/caldata.sh#L28) and abort the hotplug handler with a confusing message instead of degrading gracefully.

Would a separate case arm for `tplink,deco-m4r-v1` — with only the `art` extraction and the symlinks — be preferable to joining the D-Link arm? That would also keep the D-Link boards' behaviour unchanged (see the `board-2.bin` comment below).

---

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
```diff
@@ -11,11 +11,15 @@ case "$FIRMWARE" in
...
 		ln -sf /lib/firmware/ath10k/pre-cal-pci-0000\:00\:00.0.bin \
 			/lib/firmware/ath10k/QCA9888/hw2.0/board.bin
+		ln -sf /lib/firmware/ath10k/pre-cal-pci-0000\:00\:00.0.bin \
+			/lib/firmware/ath10k/QCA9888/hw2.0/board-2.bin
```
openwrt-ai, line 22: The symlink target is the raw 0x2f20 ART pre-calibration blob, which is not in the board-2 container format. ath10k validates the API-2 file against the `QCA-ATH10K-BOARD` magic and bails with [`ath10k_err(ar, "found invalid board magic\n")`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/ath/ath10k/core.c#L1518), after which [`ath10k_core_fetch_board_file()` falls through to `bd_api = 1`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/ath/ath10k/core.c#L1668-L1680) and loads `board.bin` — i.e. the desired result is reached only via the API-2 error path, and every boot logs a kernel error.

So this hunk works by making `board-2.bin` unparseable rather than by supplying valid board-2 data. That's worth a sentence in the commit message at minimum; if the goal is "don't use the generic `board-2.bin` on this device", masking the file (or shipping a proper board-2 container the way `package/firmware/ipq-wifi` does) expresses the intent without the error log. Which route do you prefer?

Note this also silently changes behaviour for `dlink,dir-842-c1/c2/c3`, which share this case arm — they currently use the generic `board-2.bin` and would switch to the ART blob too.

---

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
```diff
@@ -11,11 +11,15 @@ case "$FIRMWARE" in
...
 	dlink,dir-842-c1|\
 	dlink,dir-842-c2|\
-	dlink,dir-842-c3)
+	dlink,dir-842-c3|\
+	tplink,deco-m4r-v1|\
+	tplink,deco-m4r-v2)
 		caldata_extract "art" 0x5000 0x2f20
 		caldata_valid "202f" || caldata_extract "reserved" 0x15000 0x2f20
 		ln -sf /lib/firmware/ath10k/pre-cal-pci-0000\:00\:00.0.bin \
 			/lib/firmware/ath10k/QCA9888/hw2.0/board.bin
```
openwrt-ai, line 20: nit: the commit message / PR body says "Overwrite `board-2.bin` instead of `board.bin`", but this `board.bin` symlink is kept and `board-2.bin` is added alongside it. Please reword to match what the diff does (both symlinks are created), or drop the `board.bin` one if it is genuinely redundant now.

---

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
```diff
@@ -11,11 +11,15 @@ case "$FIRMWARE" in
 	case $board in
 	dlink,dir-842-c1|\
 	dlink,dir-842-c2|\
-	dlink,dir-842-c3)
+	dlink,dir-842-c3|\
+	tplink,deco-m4r-v1|\
+	tplink,deco-m4r-v2)
 		caldata_extract "art" 0x5000 0x2f20
```
openwrt-ai, line 17: nit: the subject says "add ath10k caldata extraction", but the Deco M4R already gets its pre-calibration data from the same ART offset via NVMEM — [`nvmem-cells = <&precal_art_5000>` / `nvmem-cell-names = "pre-calibration"`](https://github.com/openwrt/openwrt/blob/29034b57aad673447714b96ce62ce42a7b0394c4/target/linux/ath79/dts/qca9563_tplink_deco-m4r-v1.dts#L56-L57), backed by [`precal_art_5000: pre-calibration@5000 { reg = <0x5000 0x2f20>; }`](https://github.com/openwrt/openwrt/blob/29034b57aad673447714b96ce62ce42a7b0394c4/target/linux/ath79/dts/qca9563_tplink_deco-m4r-v1.dts#L139-L141). `ath10k_core_pre_cal_download()` tries NVMEM first, so this extraction does not change the calibration data the firmware receives; its only effect is to materialise a file for the symlinks below. Unless I'm missing something, the subject/body would be clearer as "override ath10k board data" rather than "add caldata extraction".

---

openwrt-ai (commented):
## Commit checks

- `29034b57` "ath79: add ath10k caldata extraction for TP-Link Deco M4R" — the subject scopes the change to the Deco M4R, but the new `board-2.bin` symlink is added inside the case arm shared with `dlink,dir-842-c1`, `dlink,dir-842-c2` and `dlink,dir-842-c3`, so it changes ath10k board-data selection on those three boards as well. Either give the Deco its own case arm or document the D-Link impact in the message.
- `29034b57` — the body states "Overwrite `board-2.bin` instead of `board.bin`", but the diff keeps the `board.bin` symlink and adds `board-2.bin` next to it; and the `tplink,deco-m4r-v2` compatible added here does not exist anywhere in the tree.


---

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
artms: Hmm, will investigate, right now it loads generic `board-2.bin` provided by ath10k firmware package and doesn't use precalibration.

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
artms: I had similar thoughts but didn't want to introduce code duplication but after your findings - it makes sense to have separate branch.

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
artms: Will move to separate branch and rewrite only `board-2.bin`

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
artms: Will check error log, I do not remember error message. As mentioned above - will move to separate branch so it doesn't touch dlink path at all.

artms: Logs indeed contains:
```
24.740519] ath10k_pci 0000:00:00.0: found invalid board magic
```

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
artms: Ok, it seems figured it out... Same blob is used as precalibration data and later loading necessary data to use precalibration data. Will rename commit title and description.

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
artms: Update to remove board-2.bin to avoid "error" in logs

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
artms: Removed v2 which I though existed but it relies on v1

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
artms: Moved deco to its' own branch

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
artms: Now I'm deleting board-2.bin

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
```diff
@@ -19,6 +19,17 @@ case "$FIRMWARE" in
 		;;
 	esac
 	;;
+"ath10k/pre-cal-pci-0000:00:00.0.bin")
+	case $board in
+	tplink,deco-m4r-v1)
+		caldata_extract "art" 0x5000 0x2f20
+		ln -sf /lib/firmware/ath10k/pre-cal-pci-0000\:00\:00.0.bin \
+			/lib/firmware/ath10k/QCA9888/hw2.0/board.bin
+		# remove board-2.bin, it contains bmi board which matche deco m4r which is not correct for this board
```
artms:
```suggestion
		# remove board-2.bin, it contains bmi board data which matches deco m4r which is not correct for this board
```

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
```diff
@@ -17,6 +17,14 @@ case "$FIRMWARE" in
 		ln -sf /lib/firmware/ath10k/pre-cal-pci-0000\:00\:00.0.bin \
 			/lib/firmware/ath10k/QCA9888/hw2.0/board.bin
 		;;
+
+	tplink,deco-m4r-v1)
+		caldata_extract "art" 0x5000 0x2f20
+		ln -sf /lib/firmware/ath10k/pre-cal-pci-0000\:00\:00.0.bin \
+			/lib/firmware/ath10k/QCA9888/hw2.0/board.bin
+		# remove board-2.bin, it contains bmi board which matche deco m4r which is not correct for this board
+		rm /lib/firmware/ath10k/QCA9888/hw2.0/board-2.bin
```
openwrt-ai, line 26: - Without `-f` this fails loudly when the file is already gone, and as the last command in the case arm its status becomes the hotplug handler's exit status.
- `board-2.bin` is owned by [`ath10k-board-qca9888`](https://github.com/openwrt/openwrt/blob/f0f80dad483c9f28a5ed8e4a419281e15e25f513/package/firmware/linux-firmware/qca_ath10k.mk#L56-L57), so reinstalling/upgrading that package restores it, while [the `[ -e /lib/firmware/$FIRMWARE ] && exit 0` guard](``https://github.com/openwrt/openwrt/blob/f0f80dad483c9f28a5ed8e4a419281e15e25f513/target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata#L3)`` stops this script from ever re-running — the override is then silently lost. Is that acceptable, or should the override be expressed as a device-specific board-2 container (as `package/firmware/ipq-wifi` does) instead?

```suggestion
		rm -f /lib/firmware/ath10k/QCA9888/hw2.0/board-2.bin
```

---

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
```diff
@@ -17,6 +17,14 @@ case "$FIRMWARE" in
 		ln -sf /lib/firmware/ath10k/pre-cal-pci-0000\:00\:00.0.bin \
 			/lib/firmware/ath10k/QCA9888/hw2.0/board.bin
 		;;
+
+	tplink,deco-m4r-v1)
+		caldata_extract "art" 0x5000 0x2f20
+		ln -sf /lib/firmware/ath10k/pre-cal-pci-0000\:00\:00.0.bin \
+			/lib/firmware/ath10k/QCA9888/hw2.0/board.bin
+		# remove board-2.bin, it contains bmi board which matche deco m4r which is not correct for this board
```
openwrt-ai, line 25: nit: grammar — "bmi board which matche". Your own suggestion in [#discussion_r3887355763](https://github.com/openwrt/openwrt/pull/24612#discussion_r3887355763) was never applied.

```suggestion
		# remove board-2.bin, it contains bmi board data which matches deco m4r but is not correct for this board
```

---

openwrt-ai (commented):
## Commit checks

- `f0f80dad` "ath79: tplink: extract ath10k board.bin for TP-Link Deco M4R v1" — the message only describes extracting and symlinking `board.bin`, but the change that actually makes ath10k use it is deleting the packaged `/lib/firmware/ath10k/QCA9888/hw2.0/board-2.bin`. State that removal and why it is required (API-2 lookup matches a wrong BMI entry for this board).

---

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
```diff
@@ -17,6 +17,17 @@ case "$FIRMWARE" in
...
 			/lib/firmware/ath10k/QCA9888/hw2.0/board.bin
 		;;
+
+	tplink,deco-m4r-v1)
+		caldata_extract "art" 0x5000 0x2f20
+		ln -sf /lib/firmware/ath10k/pre-cal-pci-0000\:00\:00.0.bin \
+			/lib/firmware/ath10k/QCA9888/hw2.0/board.bin
+		# overwite board-2.bin, it contains bmi board which matches deco m4r but is not correct for this board
+		# we need to keep file to survive firmware upgrades
+		# this will cause `found invalid board magic` during load unfortunetely
```
openwrt-ai, line 27: nit: typos — "overwite" and "unfortunetely".

```suggestion
		# overwrite board-2.bin, it contains bmi board data which matches deco m4r but is not correct for this board
		# we need to keep file to survive firmware upgrades
		# this will cause `found invalid board magic` during load unfortunately
```

---

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
```diff
@@ -17,6 +17,17 @@ case "$FIRMWARE" in
 		ln -sf /lib/firmware/ath10k/pre-cal-pci-0000\:00\:00.0.bin \
 			/lib/firmware/ath10k/QCA9888/hw2.0/board.bin
 		;;
+
+	tplink,deco-m4r-v1)
+		caldata_extract "art" 0x5000 0x2f20
+		ln -sf /lib/firmware/ath10k/pre-cal-pci-0000\:00\:00.0.bin \
+			/lib/firmware/ath10k/QCA9888/hw2.0/board.bin
+		# overwite board-2.bin, it contains bmi board which matches deco m4r but is not correct for this board
+		# we need to keep file to survive firmware upgrades
```
openwrt-ai, line 26: Nothing under `/lib/firmware` is preserved by sysupgrade — [`keep.d/base-files-essential`](https://github.com/openwrt/openwrt/blob/31e6d7ba68d0b0dc9666032299f95f54ae725d6a/package/base-files/files/lib/upgrade/keep.d/base-files-essential) lists only `/etc` files — so this symlink is wiped on upgrade and recreated by this handler on the fresh rootfs, exactly as the previous `rm -f` variant would be. Since the stated reason for preferring the symlink over removing the file doesn't hold, is the permanent `found invalid board magic` error on every boot still the tradeoff you want?

---

openwrt-ai (commented):
## Commit checks

- `31e6d7ba` "ath79: tplink: extract ath10k board.bin for TP-Link Deco M4R v1" — the message describes only extracting and symlinking `board.bin`, but the hunk that actually changes ath10k's behaviour is shadowing the packaged `board-2.bin` with the same non-board-2 blob so the API-2 parse fails and the driver falls back to `board.bin`. State that override and its side effect (a `found invalid board magic` error each boot) in the body.

---

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
openwrt-ai: solved by symlinking over `board-2.bin` instead of removing it, thanks

---

target/linux/ath79/generic/base-files/etc/hotplug.d/firmware/11-ath10k-caldata
openwrt-ai: fixed, thanks

---

PR #24616, opened 2026-08-08
uboot-mxs: bump to 2026.04

wigyori:
Compile-tested: all mxs boards
Runtime-tested: Olinuxino Micro

While at it, rework the Duckbill patch to compile on this u-boot version - no available imx287 hardware to test on.

### DIFF
--- a/package/boot/uboot-mxs/Makefile
+++ b/package/boot/uboot-mxs/Makefile
@@ -8,10 +8,9 @@
 include $(TOPDIR)/rules.mk
 include $(INCLUDE_DIR)/kernel.mk
 
-PKG_VERSION:=2020.04
-PKG_RELEASE:=6
-
-PKG_HASH:=fe732aaf037d9cc3c0909bad8362af366ae964bbdac6913a34081ff4ad565372
+PKG_VERSION:=2026.04
+PKG_HASH:=ac7c04b8b7004923b00a4e5d6699c5df4d21233bac9fda690d8cfbc209fff2fd
+PKG_RELEASE:=1
 
 include $(INCLUDE_DIR)/u-boot.mk
 include $(INCLUDE_DIR)/package.mk
--- a/package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
+++ b/package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
@@ -1,40 +1,35 @@
-From 83ee930c18b068c9a16b66c01aaa5d6e06570152 Mon Sep 17 00:00:00 2001
-From: Michael Heimpold <mhei@heimpold.de>
-Date: Sun, 19 Apr 2020 02:46:46 +0200
-Subject: [PATCH] arm: mxs: add support for I2SE's Duckbill boards
+From fc1143a78a45ae6ca707de7942a737bb7e9c20f9 Mon Sep 17 00:00:00 2001
+From: Zoltan HERPAI <wigyori@uid0.hu>
+Date: Fri, 21 Aug 2026 09:15:36 +0000
+Subject: [PATCH 15/15] ARM: imx: mxs: add support for I2SE Duckbill devices
 
-The Duckbill devices are small, pen-drive sized boards based on
-NXP's i.MX28 SoC. While the initial variants (Duckbill series) were
-equipped with a micro SD card slot only, the latest generation
-(Duckbill 2 series) have an additional internal eMMC onboard.
+This machine is an USB pen drive sized development board,
+based on NXP's i.MX28 CPU.
 
-Both device generations consist of four "family members":
+The changes are heavily based on Michael Heimpold's previous patchset,
+which was forward-ported to compile with u-boot 2026.04. No hardware
+was available to test, thus compile-tested only.
 
-- Duckbill/Duckbill 2: generic board, intended to be used as
-  baseboard for custom designs and/or as development board
-
-- Duckbill EnOcean/Duckbill 2 EnOcean: come with an EnOcean
-  daugther board equipped with the popular TCM310 module
-
-- Duckbill 485/Duckbill 2 485: as the name implies, these
-  devices are intended to be used as Ethernet - RS485 converters
-
-- Duckbill SPI/Duckbill 2 SPI: not sold separately, but used
-  in I2SE's development kits for Green PHY HomePlug Powerline
-  communication
-
-Signed-off-by: Michael Heimpold <mhei@heimpold.de>
-Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
+Signed-off-by: Zoltan HERPAI <wigyori@uid0.hu>
 ---
- arch/arm/mach-imx/mxs/Kconfig   |   5 +
- board/i2se/duckbill/Kconfig     |  15 +++
- board/i2se/duckbill/MAINTAINERS |   6 +
- board/i2se/duckbill/Makefile    |  10 ++
- board/i2se/duckbill/duckbill.c  | 189 ++++++++++++++++++++++++++++++++
- board/i2se/duckbill/iomux.c     | 157 ++++++++++++++++++++++++++
- configs/duckbill_defconfig      |  43 ++++++++
- include/configs/duckbill.h      | 172 +++++++++++++++++++++++++++++
- 8 files changed, 597 insertions(+)
+ arch/arm/dts/Makefile                     |   2 +
+ arch/arm/dts/imx28-duckbill-2-u-boot.dtsi |   2 +
+ arch/arm/dts/imx28-duckbill-2.dts         |  89 +++++++++++
+ arch/arm/dts/imx28-duckbill-u-boot.dtsi   |   2 +
+ arch/arm/dts/imx28-duckbill.dts           |  88 +++++++++++
+ arch/arm/mach-imx/mxs/Kconfig             |   6 +
+ board/i2se/duckbill/Kconfig               |  15 ++
+ board/i2se/duckbill/MAINTAINERS           |   7 +
+ board/i2se/duckbill/Makefile              |  10 ++
+ board/i2se/duckbill/duckbill.c            | 170 ++++++++++++++++++++++
+ board/i2se/duckbill/iomux.c               | 156 ++++++++++++++++++++
+ configs/duckbill_defconfig                |  90 ++++++++++++
+ include/configs/duckbill.h                |  93 ++++++++++++
+ 13 files changed, 730 insertions(+)
+ create mode 100644 arch/arm/dts/imx28-duckbill-2-u-boot.dtsi
+ create mode 100644 arch/arm/dts/imx28-duckbill-2.dts
+ create mode 100644 arch/arm/dts/imx28-duckbill-u-boot.dtsi
+ create mode 100644 arch/arm/dts/imx28-duckbill.dts
  create mode 100644 board/i2se/duckbill/Kconfig
  create mode 100644 board/i2se/duckbill/MAINTAINERS
  create mode 100644 board/i2se/duckbill/Makefile
@@ -43,27 +38,251 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
  create mode 100644 configs/duckbill_defconfig
  create mode 100644 include/configs/duckbill.h
 
+diff --git a/arch/arm/dts/Makefile b/arch/arm/dts/Makefile
+index 82ad3035308..d4b2f0eee59 100644
+--- a/arch/arm/dts/Makefile
++++ b/arch/arm/dts/Makefile
+@@ -706,6 +706,8 @@ dtb-$(CONFIG_TARGET_MX23_OLINUXINO) += \
+ 	imx23-olinuxino.dtb
+ 
+ dtb-$(CONFIG_MX28) += \
++	imx28-duckbill.dtb \
++	imx28-duckbill-2.dtb \
+ 	imx28-evk.dtb \
+ 	imx28-xea.dtb
+ 
+diff --git a/arch/arm/dts/imx28-duckbill-2-u-boot.dtsi b/arch/arm/dts/imx28-duckbill-2-u-boot.dtsi
+new file mode 100644
+index 00000000000..b43c10d7995
+--- /dev/null
++++ b/arch/arm/dts/imx28-duckbill-2-u-boot.dtsi
+@@ -0,0 +1,2 @@
++// SPDX-License-Identifier: GPL-2.0+
++#include "imx28-u-boot.dtsi"
+diff --git a/arch/arm/dts/imx28-duckbill-2.dts b/arch/arm/dts/imx28-duckbill-2.dts
+new file mode 100644
+index 00000000000..492e8c07f1b
+--- /dev/null
++++ b/arch/arm/dts/imx28-duckbill-2.dts
+@@ -0,0 +1,89 @@
++// SPDX-License-Identifier: GPL-2.0+
++/*
++ * I2SE Duckbill 2 board
++ *
++ * Copyright (C) 2014-2020 Michael Heimpold <mhei@heimpold.de>
++ * Copyright (C) 2026 Zoltan HERPAI <wigyori@uid0.hu>
++ */
++
++/dts-v1/;
++#include <dt-bindings/gpio/gpio.h>
++#include "imx28.dtsi"
++
++/ {
++	model = "I2SE Duckbill 2";
++	compatible = "i2se,duckbill-2", "fsl,imx28";
++
++	memory@40000000 {
++		device_type = "memory";
++		reg = <0x40000000 0x04000000>;
++	};
++
++	aliases {
++		serial0 = &duart;
++		ethernet0 = &mac0;
++	};
++
++	chosen {
++		stdout-path = "serial0:115200n8";
++	};
++};
++
++&duart {
++	pinctrl-names = "default";
++	pinctrl-0 = <&duart_pins_a>;
++	status = "okay";
++};
++
++&ssp0 {
++	compatible = "fsl,imx28-mmc";
++	pinctrl-names = "default";
++	pinctrl-0 = <&mmc0_8bit_pins_a
++		&mmc0_cd_cfg &mmc0_sck_cfg>;
++	bus-width = <8>;
++	status = "okay";
++};
++
++&mac0 {
++	phy-mode = "rmii";
++	pinctrl-names = "default";
++	pinctrl-0 = <&mac0_pins_a>, <&mac0_phy_reset_pin>;
++	phy-reset-gpios = <&gpio0 26 GPIO_ACTIVE_LOW>;
++	phy-reset-duration = <25>;
++	phy-handle = <&ethphy>;
++	status = "okay";
++
++	mdio {
++		#address-cells = <1>;
++		#size-cells = <0>;
++
++		ethphy: ethernet-phy@1 {
++			reg = <1>;
++		};
++	};
++};
++
++&pinctrl {
++	pinctrl-names = "default";
++	pinctrl-0 = <&hog_pins_a>;
++
++	hog_pins_a: hog@0 {
++		reg = <0>;
++		fsl,pinmux-ids = <
++			MX28_PAD_LCD_D17__GPIO_1_17	/* Revision detection */
++		>;
++		fsl,drive-strength = <MXS_DRIVE_4mA>;
++		fsl,voltage = <MXS_VOLTAGE_HIGH>;
++		fsl,pull-up = <MXS_PULL_DISABLE>;
++	};
++
++	mac0_phy_reset_pin: mac0-phy-reset@0 {
++		reg = <0>;
++		fsl,pinmux-ids = <
++			MX28_PAD_GPMI_ALE__GPIO_0_26	/* PHY Reset */
++		>;
++		fsl,drive-strength = <MXS_DRIVE_4mA>;
++		fsl,voltage = <MXS_VOLTAGE_HIGH>;
++		fsl,pull-up = <MXS_PULL_DISABLE>;
++	};
++};
+diff --git a/arch/arm/dts/imx28-duckbill-u-boot.dtsi b/arch/arm/dts/imx28-duckbill-u-boot.dtsi
+new file mode 100644
+index 00000000000..b43c10d7995
+--- /dev/null
++++ b/arch/arm/dts/imx28-duckbill-u-boot.dtsi
+@@ -0,0 +1,2 @@
++// SPDX-License-Identifier: GPL-2.0+
++#include "imx28-u-boot.dtsi"
+diff --git a/arch/arm/dts/imx28-duckbill.dts b/arch/arm/dts/imx28-duckbill.dts
+new file mode 100644
+index 00000000000..6b0e2219f20
+--- /dev/null
++++ b/arch/arm/dts/imx28-duckbill.dts
+@@ -0,0 +1,88 @@
++// SPDX-License-Identifier: GPL-2.0+
++/*
++ * I2SE Duckbill board
++ *
++ * Copyright (C) 2014-2020 Michael Heimpold <mhei@heimpold.de>
++ */
++
++/dts-v1/;
++#include <dt-bindings/gpio/gpio.h>
++#include "imx28.dtsi"
++
++/ {
++	model = "I2SE Duckbill";
++	compatible = "i2se,duckbill", "fsl,imx28";
++
++	memory@40000000 {
++		device_type = "memory";
++		reg = <0x40000000 0x04000000>;
++	};
++
++	aliases {
++		serial0 = &duart;
++		ethernet0 = &mac0;
++	};
++
++	chosen {
++		stdout-path = "serial0:115200n8";
++	};
++};
++
++&duart {
++	pinctrl-names = "default";
++	pinctrl-0 = <&duart_pins_a>;
++	status = "okay";
++};
++
++&ssp0 {
++	compatible = "fsl,imx28-mmc";
++	pinctrl-names = "default";
++	pinctrl-0 = <&mmc0_4bit_pins_a
++		&mmc0_cd_cfg &mmc0_sck_cfg>;
++	bus-width = <4>;
++	status = "okay";
++};
++
++&mac0 {
++	phy-mode = "rmii";
++	pinctrl-names = "default";
++	pinctrl-0 = <&mac0_pins_a>, <&mac0_phy_reset_pin>;
++	phy-reset-gpios = <&gpio2 7 GPIO_ACTIVE_LOW>;
++	phy-reset-duration = <25>;
++	phy-handle = <&ethphy>;
++	status = "okay";
++
++	mdio {
++		#address-cells = <1>;
++		#size-cells = <0>;
++
++		ethphy: ethernet-phy@1 {
++			reg = <1>;
++		};
++	};
++};
++
++&pinctrl {
++	pinctrl-names = "default";
++	pinctrl-0 = <&hog_pins_a>;
++
++	hog_pins_a: hog@0 {
++		reg = <0>;
++		fsl,pinmux-ids = <
++			MX28_PAD_LCD_D17__GPIO_1_17	/* Revision detection */
++		>;
++		fsl,drive-strength = <MXS_DRIVE_4mA>;
++		fsl,voltage = <MXS_VOLTAGE_HIGH>;
++		fsl,pull-up = <MXS_PULL_DISABLE>;
++	};
++
++	mac0_phy_reset_pin: mac0-phy-reset@0 {
++		reg = <0>;
++		fsl,pinmux-ids = <
++			MX28_PAD_SSP0_DATA7__GPIO_2_7	/* PHY Reset */
++		>;
++		fsl,drive-strength = <MXS_DRIVE_4mA>;
++		fsl,voltage = <MXS_VOLTAGE_HIGH>;
++		fsl,pull-up = <MXS_PULL_DISABLE>;
++	};
++};
+diff --git a/arch/arm/mach-imx/mxs/Kconfig b/arch/arm/mach-imx/mxs/Kconfig
+index 204252ed3c6..21cc44d69de 100644
 --- a/arch/arm/mach-imx/mxs/Kconfig
 +++ b/arch/arm/mach-imx/mxs/Kconfig
-@@ -50,6 +50,10 @@ config TARGET_APX4DEVKIT
- config TARGET_BG0900
- 	bool "Support bg0900"
+@@ -43,6 +43,11 @@ config TARGET_BTT
+ 	select PL01X_SERIAL
+ 	imply OF_UPSTREAM
  
 +config TARGET_DUCKBILL
 +	bool "Support duckbill"
++	select PL01X_SERIAL
 +	select BOARD_EARLY_INIT_F
 +
  config TARGET_MX28EVK
  	bool "Support mx28evk"
- 	select BOARD_EARLY_INIT_F
-@@ -70,6 +74,7 @@ config SYS_SOC
+ 	select PL01X_SERIAL
+@@ -75,6 +80,7 @@ config SPL_MXS_PMU_ENABLE_4P2_LINEAR_REGULATOR
+ 	  from VDD5V) - so the VDD4P2 power source is operational.
  
- source "board/bluegiga/apx4devkit/Kconfig"
- source "board/freescale/mx28evk/Kconfig"
+ source "board/nxp/mx28evk/Kconfig"
 +source "board/i2se/duckbill/Kconfig"
+ source "board/liebherr/btt/Kconfig"
  source "board/liebherr/xea/Kconfig"
- source "board/ppcag/bg0900/Kconfig"
- source "board/schulercontrol/sc_sps_1/Kconfig"
+ 
+diff --git a/board/i2se/duckbill/Kconfig b/board/i2se/duckbill/Kconfig
+new file mode 100644
+index 00000000000..98c1e4689f1
 --- /dev/null
 +++ b/board/i2se/duckbill/Kconfig
 @@ -0,0 +1,15 @@
@@ -82,15 +301,22 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
 +	default "duckbill"
 +
 +endif
+diff --git a/board/i2se/duckbill/MAINTAINERS b/board/i2se/duckbill/MAINTAINERS
+new file mode 100644
+index 00000000000..a36035b483c
 --- /dev/null
 +++ b/board/i2se/duckbill/MAINTAINERS
-@@ -0,0 +1,6 @@
+@@ -0,0 +1,7 @@
 +I2SE DUCKBILL BOARD
 +M:	Michael Heimpold <mhei@heimpold.de>
 +S:	Maintained
 +F:	board/i2se/duckbill/
 +F:	include/configs/duckbill.h
 +F:	configs/duckbill_defconfig
++F:	arch/arm/dts/imx28-duckbill*
+diff --git a/board/i2se/duckbill/Makefile b/board/i2se/duckbill/Makefile
+new file mode 100644
+index 00000000000..11bac98e4c3
 --- /dev/null
 +++ b/board/i2se/duckbill/Makefile
 @@ -0,0 +1,10 @@
@@ -104,9 +330,12 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
 +else
 +obj-y	:= iomux.o
 +endif
+diff --git a/board/i2se/duckbill/duckbill.c b/board/i2se/duckbill/duckbill.c
+new file mode 100644
+index 00000000000..a067f9dfb55
 --- /dev/null
 +++ b/board/i2se/duckbill/duckbill.c
-@@ -0,0 +1,189 @@
+@@ -0,0 +1,170 @@
 +// SPDX-License-Identifier: GPL-2.0+
 +/*
 + * I2SE Duckbill board
@@ -114,20 +343,21 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
 + * Copyright (C) 2014-2020 Michael Heimpold <mhei@heimpold.de>
 + */
 +
-+#include <common.h>
++#include <config.h>
++#include <dm/device.h>
++#include <env.h>
++#include <fdt_support.h>
++#include <fuse.h>
++#include <linux/delay.h>
++#include <linux/errno.h>
++#include <netdev.h>
++#include <asm/global_data.h>
 +#include <asm/gpio.h>
 +#include <asm/io.h>
 +#include <asm/arch/imx-regs.h>
 +#include <asm/arch/iomux-mx28.h>
 +#include <asm/arch/clock.h>
 +#include <asm/arch/sys_proto.h>
-+#include <asm/setup.h>
-+#include <fdt_support.h>
-+#include <linux/mii.h>
-+#include <miiphy.h>
-+#include <netdev.h>
-+#include <errno.h>
-+#include <fuse.h>
 +
 +DECLARE_GLOBAL_DATA_PTR;
 +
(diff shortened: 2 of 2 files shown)


wigyori:
Ping @mhei

package/boot/uboot-mxs/Makefile
```diff
@@ -8,10 +8,8 @@
 include $(TOPDIR)/rules.mk
 include $(INCLUDE_DIR)/kernel.mk
 
-PKG_VERSION:=2020.04
-PKG_RELEASE:=6
-
-PKG_HASH:=fe732aaf037d9cc3c0909bad8362af366ae964bbdac6913a34081ff4ad565372
+PKG_VERSION:=2026.04
```
openwrt-ai, line 11: `PKG_RELEASE` is removed here rather than reset. Every other u-boot package under `package/boot/` defines it — [`uboot-imx/Makefile:11`](https://github.com/openwrt/openwrt/blob/fa8e8dec1cceacf4222f60c2624fa96537dff2c2/package/boot/uboot-imx/Makefile#L11), [`uboot-mvebu/Makefile:12`](https://github.com/openwrt/openwrt/blob/fa8e8dec1cceacf4222f60c2624fa96537dff2c2/package/boot/uboot-mvebu/Makefile#L12), [`uboot-ath79/Makefile`](https://github.com/openwrt/openwrt/blob/fa8e8dec1cceacf4222f60c2624fa96537dff2c2/package/boot/uboot-ath79/Makefile) — and a version bump should reset it to 1, not drop it.

```suggestion
PKG_VERSION:=2026.04
PKG_RELEASE:=1
```

(`PKG_HASH` itself is correct — it matches the `u-boot-2026.04.tar.bz2` hash already used by uboot-imx/uboot-ath79/uboot-mvebu/uboot-sunxi.)

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
```diff
@@ -1,69 +1,248 @@
...
- 8 files changed, 597 insertions(+)
- create mode 100644 board/i2se/duckbill/Kconfig
- create mode 100644 board/i2se/duckbill/MAINTAINERS
- create mode 100644 board/i2se/duckbill/Makefile
- create mode 100644 board/i2se/duckbill/duckbill.c
- create mode 100644 board/i2se/duckbill/iomux.c
- create mode 100644 configs/duckbill_defconfig
- create mode 100644 include/configs/duckbill.h
-
+diff --git a/arch/arm/dts/Makefile b/arch/arm/dts/Makefile
```
openwrt-ai, line 1: The reworked patch starts directly at `diff --git`, dropping the whole git-format header the old version had: `From 83ee930c... Mon Sep 17 00:00:00 2001`, `From:`, `Date:`, `Subject: [PATCH] arm: mxs: add support for I2SE's Duckbill boards`, the body describing the Duckbill family, and the `Signed-off-by:` lines of Michael Heimpold and Stefan Wahren.

Without those the patch is no longer `git am`-able and the original authorship is lost. Please restore a git-format header (From:/Date:/Subject: plus a short body), keep the original `Signed-off-by:` lines and add your own for the 2026.04 rework.

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
```diff
@@ -275,11 +445,11 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
...
 +	if (system_rev == 1)
-+		led_red_gpio = MX28_PAD_AUART1_RX__GPIO_3_4;
++		dm_gpio_lookup_name("GPIO3_4", &desc);
 +	else
-+		led_red_gpio = MX28_PAD_SAIF0_LRCLK__GPIO_3_21;
-+	gpio_request(led_red_gpio, "led_red");
-+	gpio_direction_output(led_red_gpio, 1);
++		dm_gpio_lookup_name("GPIO3_21", &desc);
++	dm_gpio_request(&desc, "led_red");
++	dm_gpio_set_dir_flags(&desc, GPIOD_IS_OUT_ACTIVE);
```
openwrt-ai, line 452: `GPIOD_IS_OUT_ACTIVE` alone does not configure the pin as an output, so the red "bootloader running" LED is never driven — this is a behaviour regression versus the old `gpio_direction_output(led_red_gpio, 1)`.

`dm_gpio_set_dir_flags()` clears `GPIOD_MASK_DIR` and applies the given flags, and the driver-agnostic fallback in [`_dm_gpio_set_flags()`, gpio-uclass.c:711-724](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/drivers/gpio/gpio-uclass.c#L711-L724) only calls `direction_output()` when `GPIOD_IS_OUT` is set (`GPIOD_IS_OUT_ACTIVE` is just the value bit). [`gpio_mxs_ops`, mxs_gpio.c:228-233](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/drivers/gpio/mxs_gpio.c#L228-L233) has no `.set_flags` callback, so that fallback path is the one taken here and the call is a no-op.

```suggestion
+	dm_gpio_set_dir_flags(&desc, GPIOD_IS_OUT | GPIOD_IS_OUT_ACTIVE);
```

(Checked against u-boot master `ece349a`; v2026.04 was not in the reference clone, but this code path is long-standing.)

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
```diff
@@ -275,11 +445,11 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
...
 +	/* enable red LED to indicate a running bootloader */
 +	if (system_rev == 1)
-+		led_red_gpio = MX28_PAD_AUART1_RX__GPIO_3_4;
++		dm_gpio_lookup_name("GPIO3_4", &desc);
 +	else
-+		led_red_gpio = MX28_PAD_SAIF0_LRCLK__GPIO_3_21;
-+	gpio_request(led_red_gpio, "led_red");
-+	gpio_direction_output(led_red_gpio, 1);
++		dm_gpio_lookup_name("GPIO3_21", &desc);
++	dm_gpio_request(&desc, "led_red");
```
openwrt-ai, line 451: These two `dm_gpio_lookup_name()` calls ignore their return value, unlike the revision-detect lookup a few lines above which does `if (dm_gpio_lookup_name("GPIO1_17", &desc)) return -EINVAL;`.

Because the same `desc` is reused, a failing lookup here leaves `desc` still pointing at GPIO1_17 (which was just requested as `board_revision`), so `dm_gpio_request()`/`dm_gpio_set_dir_flags()` would then reconfigure the revision-detect pin as an output instead of the LED. Please check the return value on both branches, and consider `dm_gpio_free()`-ing the revision descriptor before reusing `desc` (or use a second `struct gpio_desc`).

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
```diff
@@ -456,31 +628,50 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
...
-+CONFIG_TARGET_DUCKBILL=y
-+CONFIG_SPL_SERIAL_SUPPORT=y
++CONFIG_NR_DRAM_BANKS=1
 +CONFIG_ENV_SIZE=0x20000
 +CONFIG_ENV_OFFSET=0x20000
-+CONFIG_NR_DRAM_BANKS=1
-+CONFIG_SPL=y
++CONFIG_IMX_CONFIG=""
++CONFIG_DM_GPIO=y
++CONFIG_DEFAULT_DEVICE_TREE="imx28-duckbill"
```
openwrt-ai, line 649: This defconfig enables `CONFIG_OF_CONTROL=y` with a single built-in DT and no `CONFIG_MULTI_DTB_FIT`, so u-boot itself always runs with `imx28-duckbill` — even on a Duckbill 2, which `misc_init_r()` detects at runtime and for which the board sets `fdt_file=imx28-duckbill-2.dtb` for Linux.

The two DTs added by this patch differ in ways u-boot's own drivers consume: `&ssp0` is `mmc0_4bit_pins_a` / `bus-width = <4>` vs `mmc0_8bit_pins_a` / `bus-width = <8>`, and `phy-reset-gpios` is `<&gpio2 7>` vs `<&gpio0 26>`. Under the old non-OF_CONTROL build both revisions were handled at runtime by `board_mmc_init()`/`board_eth_init()`; that runtime selection is gone now. So on a Duckbill 2 the eMMC would come up 4-bit and the PHY reset would toggle the wrong pin.

The sibling MX28 board in the same Kconfig solves exactly this with a multi-DTB FIT — see [`configs/imx28_btt3_defconfig` L96-L98](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/configs/imx28_btt3_defconfig#L96-L98) plus `board_fit_config_name_match()`. Given the PR notes no imx287 hardware was available to test, is the single-DT choice intentional, or should `CONFIG_MULTI_DTB_FIT` be wired up here?

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
```diff
@@ -1,69 +1,248 @@
...
-
+diff --git a/arch/arm/dts/Makefile b/arch/arm/dts/Makefile
+index 82ad3035..d4b2f0ee 100644
+--- a/arch/arm/dts/Makefile
++++ b/arch/arm/dts/Makefile
+@@ -706,6 +706,8 @@ dtb-$(CONFIG_TARGET_MX23_OLINUXINO) += \
+ 	imx23-olinuxino.dtb
+ 
+ dtb-$(CONFIG_MX28) += \
++	imx28-duckbill.dtb \
```
openwrt-ai, line 9: u-boot already ships both of these device trees via the devicetree-rebasing subtree: [`dts/upstream/src/arm/nxp/mxs/imx28-duckbill.dts`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/dts/upstream/src/arm/nxp/mxs/imx28-duckbill.dts) and `imx28-duckbill-2.dts`. The sibling MX28 board added to the same `mach-imx/mxs/Kconfig` this patch touches uses that route — [`TARGET_BTT ... imply OF_UPSTREAM`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/arch/arm/mach-imx/mxs/Kconfig#L44) with `CONFIG_DEFAULT_DEVICE_TREE="nxp/mxs/imx28-btt3-1"` — keeping only the small `*-u-boot.dtsi` files in `arch/arm/dts/`.

Was `OF_UPSTREAM` considered here? It would let you drop both hand-maintained stripped copies (and this Makefile hunk) and avoid them drifting from the kernel DTs.

nit: if the in-tree copies are kept, the entries are not in the alphabetical order the rest of this list uses — `imx28-duckbill-2.dtb` sorts before `imx28-duckbill.dtb`.

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
```diff
@@ -1,69 +1,248 @@
...
++#include <dt-bindings/gpio/gpio.h>
++#include "imx28.dtsi"
++
++/ {
++	model = "I2SE Duckbill";
++	compatible = "i2se,duckbill", "fsl,imx28";
++
++	memory@40000000 {
++		device_type = "memory";
++		reg = <0x40000000 0x04000000>;
```
openwrt-ai, line 148: nit: this says 64 MB, but the kernel DT for the same board declares 128 MB — [`dts/upstream/src/arm/nxp/mxs/imx28-duckbill.dts#L15-L18`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/dts/upstream/src/arm/nxp/mxs/imx28-duckbill.dts#L15-L18) uses `reg = <0x40000000 0x08000000>`. u-boot's own RAM size comes from `mxs_dram_init()` (SPL-passed value), so this is cosmetic today, but it contradicts the upstream DT.

```suggestion
+		reg = <0x40000000 0x08000000>;
```

Same applies to line 46 in the `imx28-duckbill-2.dts` hunk above (upstream `imx28-duckbill-2.dts` also uses `0x08000000`).

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
```diff
@@ -82,15 +261,22 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
...
 +++ b/board/i2se/duckbill/MAINTAINERS
-@@ -0,0 +1,6 @@
+@@ -0,0 +1,7 @@
 +I2SE DUCKBILL BOARD
 +M:	Michael Heimpold <mhei@heimpold.de>
 +S:	Maintained
 +F:	board/i2se/duckbill/
 +F:	include/configs/duckbill.h
 +F:	configs/duckbill_defconfig
++F:	arch/arm/dts/imx28-duckbill.dts
```
openwrt-ai, line 276: nit: the new `F:` entry covers only `imx28-duckbill.dts`, but this patch adds three more files under `arch/arm/dts/`: `imx28-duckbill-2.dts`, `imx28-duckbill-u-boot.dtsi` and `imx28-duckbill-2-u-boot.dtsi`. A glob would cover all of them:

```suggestion
+F:	arch/arm/dts/imx28-duckbill*
```

---

openwrt-ai (commented):
Reviewed 1 new commit (`fa8e8de`); the commit message matches the diff. Comments inline — the notable ones are the dropped `PKG_RELEASE`, the missing git header on the reworked patch, a red-LED GPIO that is no longer configured as an output, and the single built-in device tree for two board revisions that need different `ssp0`/PHY-reset settings.

---

package/boot/uboot-mxs/patches/0001-ARM-imx-mxs-add-support-for-I2SE-Duckbill-devices.patch
```diff
@@ -456,31 +410,51 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
...
-+CONFIG_SPL_SERIAL_SUPPORT=y
++CONFIG_NR_DRAM_BANKS=1
 +CONFIG_ENV_SIZE=0x20000
 +CONFIG_ENV_OFFSET=0x20000
-+CONFIG_NR_DRAM_BANKS=1
-+CONFIG_SPL=y
++CONFIG_IMX_CONFIG=""
++CONFIG_DM_GPIO=y
++CONFIG_DEFAULT_DEVICE_TREE="nxp/mxs/imx28-duckbill"
++CONFIG_OF_UPSTREAM=y
```
openwrt-ai, line 457: Switching to `OF_UPSTREAM` also dropped `arch/arm/dts/imx28-duckbill-u-boot.dtsi`, which leaves this board with **no** U-Boot `.dtsi` included at all.

The auto-include takes the first match of `<board>-u-boot.dtsi` / `<SYS_SOC>-u-boot.dtsi` / `<SYS_CPU>-u-boot.dtsi` / `<SYS_VENDOR>-u-boot.dtsi` / `u-boot.dtsi` under `arch/arm/dts/` ([`u_boot_dtsi_options`, scripts/Makefile.lib:226-230](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/scripts/Makefile.lib#L226-L230)). With `SYS_SOC="mxs"` / `SYS_VENDOR="i2se"`, none of `mxs-u-boot.dtsi`, `i2se-u-boot.dtsi` or `u-boot.dtsi` exist, and `imx28-u-boot.dtsi` is *not* picked up because the basename is `imx28-duckbill`. So nothing is pulled in, and two things regress:

1. **`gpio-ranges` is gone.** [`arch/arm/dts/imx28-u-boot.dtsi`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/arch/arm/dts/imx28-u-boot.dtsi) is the only place it is defined — upstream `imx28.dtsi` has none. [`mxs_of_to_plat()`, drivers/gpio/mxs_gpio.c:278-284](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/drivers/gpio/mxs_gpio.c#L278-L284) parses it to size each bank, so every bank will now print `mxs_of_to_plat: 'gpio-ranges' not defined - using default!` at boot and report 32 pins instead of the real 29/32/28/31/21.
2. **`chosen`/`stdout-path` and the `serial0 = &duart` override are gone.** Upstream [`imx28.dtsi:30`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/dts/upstream/src/arm/nxp/mxs/imx28.dtsi#L30) aliases `serial0` to `auart0`, and the upstream Duckbill DTS enables `&auart0`. Console selection therefore falls through `serial_check_stdout()` and `uclass_get_device_by_seq(UCLASS_SERIAL, 0)` down to the generic `uclass_first_device_err()` fallback in [`serial_find_console_or_panic()`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/drivers/serial/serial-uclass.c#L74). It happens to land on the DUART only because `CONFIG_MXS_AUART_SERIAL` is not set in this defconfig so `auart0` never binds — that is a fairly fragile thing to depend on.

The sibling MX28 board keeps a one-line stub for exactly this reason — [`arch/arm/dts/imx28-btt3-1-u-boot.dtsi`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/arch/arm/dts/imx28-btt3-1-u-boot.dtsi). Re-adding `arch/arm/dts/imx28-duckbill-u-boot.dtsi` with `#include "imx28-u-boot.dtsi"` plus the `aliases` / `chosen` nodes the removed in-tree copy carried would restore both.

---

package/boot/uboot-mxs/patches/0001-ARM-imx-mxs-add-support-for-I2SE-Duckbill-devices.patch
```diff
@@ -456,31 +410,51 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
...
-+CONFIG_TARGET_DUCKBILL=y
-+CONFIG_SPL_SERIAL_SUPPORT=y
++CONFIG_NR_DRAM_BANKS=1
 +CONFIG_ENV_SIZE=0x20000
 +CONFIG_ENV_OFFSET=0x20000
-+CONFIG_NR_DRAM_BANKS=1
-+CONFIG_SPL=y
++CONFIG_IMX_CONFIG=""
++CONFIG_DM_GPIO=y
++CONFIG_DEFAULT_DEVICE_TREE="nxp/mxs/imx28-duckbill"
```
openwrt-ai, line 456: With a single built-in DT and no `CONFIG_MULTI_DTB_FIT`, U-Boot always runs the Duckbill 1 device tree — including on a Duckbill 2, which `misc_init_r()` detects at runtime a few hunks above and for which the board sets `fdt_file=imx28-duckbill-2.dtb` for Linux.

The two upstream DTs differ in ways U-Boot's own drivers consume: `bus-width = <4>` / `mmc0_4bit_pins_a` ([`imx28-duckbill.dts:115-123`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/dts/upstream/src/arm/nxp/mxs/imx28-duckbill.dts#L115-L123)) vs `bus-width = <8>` / `mmc0_8bit_pins_a` ([`imx28-duckbill-2.dts:142-147`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/dts/upstream/src/arm/nxp/mxs/imx28-duckbill-2.dts#L142-L147)), and `phy-reset-gpios = <&gpio2 7>` vs [`<&gpio0 26>`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/dts/upstream/src/arm/nxp/mxs/imx28-duckbill-2.dts#L75). Under the old non-OF_CONTROL build both revisions were handled at runtime by `board_mmc_init()` / `board_eth_init()`; that runtime selection is gone. So on a Duckbill 2 the eMMC comes up 4-bit and the PHY reset toggles the wrong pin.

I raised this before the rework and the anchor went stale, but `OF_UPSTREAM` now makes it cheap: the sibling board wires it up with [`configs/imx28_btt3_defconfig:95-98`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/configs/imx28_btt3_defconfig#L95-L98) (`CONFIG_OF_LIST="nxp/mxs/imx28-btt3-0 ..."` + `CONFIG_MULTI_DTB_FIT`) plus `board_fit_config_name_match()`. Given the PR notes no imx287 hardware was available to test, is the single-DT choice intentional?

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
```diff
@@ -82,15 +43,22 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
...
```
openwrt-ai: nit: this `F:` entry no longer matches any file. The rework moved the board to `OF_UPSTREAM`, so the patch no longer adds `arch/arm/dts/imx28-duckbill.dts` — it now comes from `dts/upstream/src/arm/nxp/mxs/`. The sibling MX28 board that uses the same route lists no `arch/arm/dts` path at all: [`board/liebherr/btt/MAINTAINERS`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/board/liebherr/btt/MAINTAINERS).

(Supersedes my earlier suggestion here to glob the entry — that was written against the previous revision, which still added the DTs in-tree.)

Dropping the line also needs the hunk header on line 51 changed from `@@ -0,0 +1,7 @@` to `@@ -0,0 +1,6 @@`, so I've left it as prose rather than a suggestion block. Alternatively, if you re-add `arch/arm/dts/imx28-duckbill-u-boot.dtsi` per the defconfig comment, `F:	arch/arm/dts/imx28-duckbill*` becomes correct and keeps the line count at 7.

---

openwrt-ai (commented):
Re-reviewed at `80fd870` (force-push on top of `fa8e8de`). The new revision adds `PKG_RELEASE:=1` and switches the board to `OF_UPSTREAM`, which resolves the earlier `PKG_RELEASE` and hand-maintained-DTS comments — nice.

Three notes on the rework inline. The main one: dropping `imx28-duckbill-u-boot.dtsi` leaves the board with no U-Boot `.dtsi` included at all, which loses `gpio-ranges` and the `stdout-path` / `serial0 = &duart` overrides that the removed in-tree DTS copies carried.

Still open from the previous round and unchanged by this revision (not re-posted to avoid duplicate threads): the missing git-format header on the patch (also flagged by the formality bot on 2026-08-20), the two unchecked `dm_gpio_lookup_name()` return values in `misc_init_r()`, and `GPIOD_IS_OUT_ACTIVE` passed without `GPIOD_IS_OUT`.

Comparisons above are against u-boot `ece349a` (master in the reference clone); v2026.04 was not available locally, but all the code paths cited are long-standing.

---

package/boot/uboot-mxs/patches/0001-ARM-imx-mxs-add-support-for-I2SE-Duckbill-devices.patch
```diff
@@ -456,31 +668,57 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
...
++CONFIG_OF_UPSTREAM=y
++CONFIG_OF_LIST="nxp/mxs/imx28-duckbill nxp/mxs/imx28-duckbill-2"
++CONFIG_MULTI_DTB_FIT_GZIP=y
++CONFIG_MULTI_DTB_FIT_USER_DEFINED_AREA=y
++CONFIG_MULTI_DTB_FIT=y
```
openwrt-ai, line 461: `CONFIG_MULTI_DTB_FIT=y` on its own does not give you runtime DT selection — the board also has to provide `board_fit_config_name_match()`, and `board/i2se/duckbill/duckbill.c` in this patch does not define one.

The weak default returns `-EINVAL` ([`board_fit_config_name_match()`, boot/common_fit.c:24-27](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/boot/common_fit.c#L24-L27)), which is non-zero, so [`fit_find_config_node()`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/boot/common_fit.c#L70-L76) `continue`s past every configuration in the FIT and falls through to the default one. `fit-dtb.blob` is generated from `CONFIG_OF_LIST` in order ([Makefile:1485-1487](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/Makefile#L1485-L1487)), so the default is `nxp/mxs/imx28-duckbill`. Net effect: a Duckbill 2 still runs U-Boot with the Duckbill 1 DT (`bus-width = <4>`, `phy-reset-gpios = <&gpio2 7>`) — exactly the case this option was added to fix.

The sibling MX28 board implements it off the same kind of revision-GPIO read: [`board/liebherr/btt/btt.c:393-404`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/board/liebherr/btt/btt.c#L393-L404). Note it cannot reuse the `system_rev` computed in `misc_init_r()` — `fit_find_config_node()` runs out of `fdtdec_setup()` during `board_init_f`, long before `misc_init_r()`, so GPIO1_17 has to be sampled again there.

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
```diff
@@ -43,27 +38,251 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
...
+diff --git a/arch/arm/dts/Makefile b/arch/arm/dts/Makefile
+index 82ad3035308..d4b2f0eee59 100644
+--- a/arch/arm/dts/Makefile
++++ b/arch/arm/dts/Makefile
+@@ -706,6 +706,8 @@ dtb-$(CONFIG_TARGET_MX23_OLINUXINO) += \
+ 	imx23-olinuxino.dtb
+ 
+ dtb-$(CONFIG_MX28) += \
++	imx28-duckbill.dtb \
++	imx28-duckbill-2.dtb \
```
openwrt-ai: These two `dtb-` entries — and the `arch/arm/dts/imx28-duckbill.dts` / `imx28-duckbill-2.dts` files this patch re-adds — are never built for this board.

The defconfig sets `CONFIG_OF_UPSTREAM=y` (line 690), and [`dts/Makefile:13-25`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/dts/Makefile#L13-L25) then sets `dt_dir := dts/upstream/src/$(ARCH)`; the only DT build step is [`arch-dtbs: $(MAKE) $(build)=$(dt_dir) dtbs`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/dts/Makefile#L62-L64), so `arch/arm/dts/Makefile` is never descended into. `CONFIG_DEFAULT_DEVICE_TREE="nxp/mxs/imx28-duckbill"` and `CONFIG_OF_LIST="nxp/mxs/…"` confirm it: the board runs [`dts/upstream/src/arm/nxp/mxs/imx28-duckbill.dts`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/dts/upstream/src/arm/nxp/mxs/imx28-duckbill.dts).

So the two in-tree copies are dead code that will silently drift from the DT actually in use — they already disagree: `reg = <0x40000000 0x04000000>` (64 MB) here vs `0x08000000` (128 MB) upstream, and they drop the `leds`, `reg_3p3v` and `lradc` nodes.

They are not inert either: `dtb-$(CONFIG_MX28)` also applies to `mx28evk_defconfig`, which is *not* `OF_UPSTREAM` (`CONFIG_DEFAULT_DEVICE_TREE="imx28-evk"`), so that build would now compile two unused Duckbill DTBs.

Suggest dropping this hunk together with both `arch/arm/dts/imx28-duckbill*.dts` files — keeping only the two `-u-boot.dtsi` stubs, which *are* picked up — and adjusting the diffstat / `create mode` list in the patch header to match.

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
```diff
@@ -456,31 +668,57 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
...
 --- /dev/null
 +++ b/configs/duckbill_defconfig
-@@ -0,0 +1,43 @@
+@@ -0,0 +1,90 @@
 +CONFIG_ARM=y
 +CONFIG_ARCH_MX28=y
-+CONFIG_SYS_TEXT_BASE=0x40002000
-+CONFIG_SPL_GPIO_SUPPORT=y
++CONFIG_TEXT_BASE=0x40002000
++CONFIG_SYS_MALLOC_F_LEN=0x400
```
openwrt-ai: The newly enabled `CONFIG_MULTI_DTB_FIT_GZIP=y` (line 692) needs a far larger pre-relocation heap than the 1 KiB kept here.

[`uncompress_blob()`, lib/fdtdec.c:1204-1257](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/lib/fdtdec.c#L1204-L1257) calls `gunzip()` → `zunzip()`, which sets `s.zalloc = gzalloc` and calls `inflateInit2()`. [`gzalloc()` is a plain `malloc()` (lib/gunzip.c:32-42)](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/lib/gunzip.c#L32-L42) and [`inflateInit2()` allocates `sizeof(struct inflate_state)`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/lib/zlib/inflate.c#L48), which is over 8 KiB by itself because of `code codes[ENOUGH]` with [`#define ENOUGH 2048`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/lib/zlib/inftrees.h#L43). That is why `MULTI_DTB_FIT_GZIP` carries `depends on SYS_MALLOC_F`. With `SYS_MALLOC_F_LEN=0x400` the allocation cannot succeed and DT setup fails with `uncompress_blob: Unable to uncompress`.

`CONFIG_MULTI_DTB_FIT_USER_DEFINED_AREA` only moves the *output* buffer out of the heap; the inflate state still comes from the early pool. The sibling MX28 board using the identical GZIP + user-defined-area combination sets `CONFIG_SYS_MALLOC_F_LEN=0x8000` ([`configs/imx28_btt3_defconfig:5`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/configs/imx28_btt3_defconfig#L5)).

Since the PR notes this path was only compile-tested, worth bumping it:

```suggestion
+CONFIG_SYS_MALLOC_F_LEN=0x8000
```

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
```diff
@@ -1,40 +1,35 @@
-From 83ee930c18b068c9a16b66c01aaa5d6e06570152 Mon Sep 17 00:00:00 2001
-From: Michael Heimpold <mhei@heimpold.de>
-Date: Sun, 19 Apr 2020 02:46:46 +0200
-Subject: [PATCH] arm: mxs: add support for I2SE's Duckbill boards
+From fc1143a78a45ae6ca707de7942a737bb7e9c20f9 Mon Sep 17 00:00:00 2001
+From: Zoltan HERPAI <wigyori@uid0.hu>
+Date: Fri, 21 Aug 2026 09:15:36 +0000
+Subject: [PATCH 15/15] ARM: imx: mxs: add support for I2SE Duckbill devices
```
openwrt-ai: nit: `[PATCH 15/15]` is a leftover from the local series this was exported from — this is a standalone patch in `package/boot/uboot-mxs/patches/`, and the numbering will read as a missing 1..14 for anyone who looks at it later (or receives it upstream).

```suggestion
Subject: [PATCH] ARM: imx: mxs: add support for I2SE Duckbill devices
```

---

package/boot/uboot-mxs/patches/0001-ARM-imx-mxs-add-support-for-I2SE-Duckbill-devices.patch
```diff
@@ -1,40 +1,35 @@
...
-- Duckbill 485/Duckbill 2 485: as the name implies, these
-  devices are intended to be used as Ethernet - RS485 converters
-
-- Duckbill SPI/Duckbill 2 SPI: not sold separately, but used
-  in I2SE's development kits for Green PHY HomePlug Powerline
-  communication
-
-Signed-off-by: Michael Heimpold <mhei@heimpold.de>
-Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
+Signed-off-by: Zoltan HERPAI <wigyori@uid0.hu>
```
openwrt-ai, line 13: nit: thanks for restoring the git header — but it now carries only your `Signed-off-by:`. The version this replaces had `Signed-off-by: Michael Heimpold <mhei@heimpold.de>` and `Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>`.

Given the body says the change is "heavily based on Michael Heimpold's previous patchset", and the files added by this very patch still carry `Copyright (C) 2014-2020 Michael Heimpold` with him listed as `M:` in the new `board/i2se/duckbill/MAINTAINERS`, the original `Signed-off-by:` lines should be kept above yours (and `From:` set to the original author if the bulk is still his work, with your rework noted in the body).

---

package/boot/uboot-mxs/patches/0001-ARM-imx-mxs-add-support-for-I2SE-Duckbill-devices.patch
```diff
@@ -43,27 +38,251 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
...
- config TARGET_BG0900
- 	bool "Support bg0900"
+@@ -43,6 +43,11 @@ config TARGET_BTT
+ 	select PL01X_SERIAL
+ 	imply OF_UPSTREAM
  
 +config TARGET_DUCKBILL
 +	bool "Support duckbill"
++	select PL01X_SERIAL
 +	select BOARD_EARLY_INIT_F
```
openwrt-ai, line 55: nit: `TARGET_BTT`, two lines up in the very same hunk, carries `imply OF_UPSTREAM`; `TARGET_DUCKBILL` does not, which is why the defconfig has to spell out `CONFIG_OF_UPSTREAM=y` (line 690). Adding `imply OF_UPSTREAM` here would match the sibling and let `make savedefconfig` drop that line.

Left as prose rather than a suggestion block because adding the line also requires bumping the hunk header on line 263 from `@@ -43,6 +43,11 @@` to `@@ -43,6 +43,12 @@`.

---

openwrt-ai (commented):
Re-reviewed at `197f3c8` (force-push on top of `80fd870`); the only change since the last round is the Duckbill patch, and the commit message still matches the diff.

The rework addresses three earlier points: the git-format header is back (FormalityCheck is green now), `imx28-duckbill-u-boot.dtsi` / `imx28-duckbill-2-u-boot.dtsi` are added so `imx28-u-boot.dtsi` and its `gpio-ranges` are pulled in again, and `MULTI_DTB_FIT` is wired up for the two board revisions.

Three new things inline. The two that matter: `MULTI_DTB_FIT` without a `board_fit_config_name_match()` still always selects the Duckbill 1 DT, so the revision problem is not actually fixed yet; and `CONFIG_MULTI_DTB_FIT_GZIP` needs a bigger `SYS_MALLOC_F_LEN` than the 1 KiB currently set. The third is that the re-added `arch/arm/dts/imx28-duckbill*.dts` copies are never built under `OF_UPSTREAM`.

Still open from earlier rounds and unchanged here (not re-posted to avoid duplicate threads): the two unchecked `dm_gpio_lookup_name()` return values in `misc_init_r()`, and `GPIOD_IS_OUT_ACTIVE` passed without `GPIOD_IS_OUT`.

Comparisons are against the u-boot reference clone at `ece349a` (tag `v2026.07`); v2026.04 was not available locally, but the patch applies cleanly there and all cited code paths are long-standing.

---

openwrt-ai (commented):

---

package/boot/uboot-mxs/Makefile
openwrt-ai: fixed, thanks

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
openwrt-ai: fixed, thanks

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
openwrt-ai: fixed, thanks

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
openwrt-ai: solved by dropping the in-tree DTS copies entirely, thanks

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
openwrt-ai: fixed, thanks

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
openwrt-ai: solved by keeping the glob and re-adding the two `imx28-duckbill*-u-boot.dtsi` stubs, thanks

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
openwrt-ai: fixed, thanks

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
openwrt-ai: fixed, thanks

---

package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch
openwrt-ai: fixed, thanks

---

wigyori:
Looks ready to go, ping @mhei

mhei:
@wigyori Thanks, will have a look the next evenings.

mhei:
It does not boot. But I did not yet figured out the reason, still investigating...

mhei:
@wigyori : I finally managed to find the problem: it seems that U-Boot does not like the multi-fit stuff, so I just disabled it which works now like a charme on Duckbill v1 and v2 series. I placed the updated patch here: https://github.com/mhei/openwrt/blob/uboot-mxs-202604/package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch maybe you can pull it to update this PR here. Thanks

wigyori:
> @wigyori : I finally managed to find the problem: it seems that U-Boot does not like the multi-fit stuff, so I just disabled it which works now like a charme on Duckbill v1 and v2 series. I placed the updated patch here: https://github.com/mhei/openwrt/blob/uboot-mxs-202604/package/boot/uboot-mxs/patches/001-add-i2se-duckbill.patch maybe you can pull it to update this PR here. Thanks

Thanks a lot @mhei for checking this out - will update the patchset and add the Duckbill to the "Runtime-tested" section as well.

package/boot/uboot-mxs/patches/0001-ARM-imx-mxs-add-support-for-I2SE-Duckbill-devices.patch
```diff
@@ -43,27 +34,38 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
  create mode 100644 configs/duckbill_defconfig
  create mode 100644 include/configs/duckbill.h
 
+--- /dev/null
++++ b/arch/arm/dts/imx28-duckbill-2-u-boot.dtsi
```
openwrt-ai, line 38: nit: with `CONFIG_OF_LIST` / `CONFIG_MULTI_DTB_FIT` dropped in this revision, `nxp/mxs/imx28-duckbill` is the only DT built, so this stub is never included by anything — the auto-include only picks `<board>-u-boot.dtsi` for the board actually being built ([`u_boot_dtsi_options`, scripts/Makefile.lib:226-230](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/scripts/Makefile.lib#L226-L230)). Dropping the file plus its `create mode` and diffstat lines keeps the patch tidy.

---

package/boot/uboot-mxs/patches/0001-ARM-imx-mxs-add-support-for-I2SE-Duckbill-devices.patch
```diff
@@ -514,51 +531,17 @@ Signed-off-by: Stefan Wahren <stefan.wahren@i2se.com>
...
-+#define CONFIG_AUTOBOOT_STOP_STR	" "	/* stop autoboot with <Space> */
-+#define CONFIG_RESET_TO_RETRY			/* reset board to retry booting */
++#define CFG_SYS_SDRAM_BASE		PHYS_SDRAM_1
 +
 +/* Extra Environment */
-+#define CONFIG_EXTRA_ENV_SETTINGS \
++#define CFG_EXTRA_ENV_SETTINGS \
 +	"mmc_part2_offset=1000\0" \
 +	"mmc_part3_offset=19000\0" \
 +	"update_openwrt_firmware_filename=openwrt-mxs-root.ext4\0" \
```
openwrt-ai, line 547: This revision also dropped `CONFIG_USE_IPADDR` / `USE_SERVERIP` / `USE_NETMASK` / `USE_GATEWAYIP` from the defconfig, and [`env_default.h:55-60`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/include/env_default.h#L55-L60) only defines `ipaddr`/`serverip` when those are set. `update_openwrt_firmware` below calls bare `tftp`, so on a board with a clean environment it now aborts with ``*** ERROR: `serverip' not set`` instead of falling back to the 192.168.1.1 / 192.168.1.10 defaults the 2020.04 config carried in `include/configs/duckbill.h`. Intentional, or should the four `CONFIG_USE_*`/value pairs come back?

---

package/boot/uboot-mxs/patches/0001-ARM-imx-mxs-add-support-for-I2SE-Duckbill-devices.patch
```diff
@@ -1,40 +1,31 @@
...
-  devices are intended to be used as Ethernet - RS485 converters
-
-- Duckbill SPI/Duckbill 2 SPI: not sold separately, but used
-  in I2SE's development kits for Green PHY HomePlug Powerline
-  communication
+The changes are heavily based on Michael Heimpold's previous patchset,
+which was forward-ported to compile with u-boot 2026.04.
 
+Signed-off-by: Zoltan HERPAI <wigyori@uid0.hu>
+Tested-by: Michael Heimpold <mhei@heimpold.de>
```
openwrt-ai, line 13: nit: the PR description still reads "Runtime-tested: Olinuxino Micro" and "no available imx287 hardware to test on", which now contradicts this `Tested-by:` and the commit message's "Runtime-tested: Olinuxino Micro, I2SE Duckbill". Worth refreshing the PR body so the two agree.

---

openwrt-ai (commented):
Reviewed 1 new commit.

---

package/boot/uboot-mxs/patches/0001-ARM-imx-mxs-add-support-for-I2SE-Duckbill-devices.patch
openwrt-ai: solved by dropping `MULTI_DTB_FIT` entirely, thanks

---

package/boot/uboot-mxs/patches/0001-ARM-imx-mxs-add-support-for-I2SE-Duckbill-devices.patch
openwrt-ai: answered instead by @mhei runtime-testing the single-DT build on both Duckbill v1 and v2 (#issuecomment-5654570047) — treating this as intentional, thanks

---

PR #24618, opened 2026-08-08
netifd: add bandwidth autodetect capability

blocktrron:
Note: This is currently pretty bare, but features a working end to end functionality. A demo implementation with `sqm-scripts` can be found [here](https://github.com/blocktrron/sqm-scripts/tree/netifd-autorate)

---

This PR adds a new `bandwidth` property for netifd interfaces.

This can be used to configure the bandwidth of a connection for use with shaper-packages / mesh protocols etc.

Bandwidth can be configured in the `interface` section independent of the proto or set by the proto handler.

This is implemented for the PPPoE protocol to parse bandwidth information according to the [AVM specification](https://fritz.support/resources/PPP-Controlled_Settings_for_FRITZOS_v1.3.pdf)  as well as the proprietary United Internet format. This way, a OpenWrt router on a DSL line can automatically configure it's shaper without reading this information from a DSL modem. It also automatically adapts the configured rates after a contract change etc.

This is currently a niche feature used primarily in Germany, but a [proper specification for DHCPv4 / DHCPv6 based connections](https://github.com/GIC-de/draft-dhcp-rate-signaling ) is currently in the works.

### DIFF
--- a/package/network/config/netifd/patches/0001-interface-add-rate-propert.patch
+++ b/package/network/config/netifd/patches/0001-interface-add-rate-propert.patch
@@ -1,0 +1,316 @@
+From 5424f1bce1a57086e72b7573cb5cd9834676e962 Mon Sep 17 00:00:00 2001
+From: David Bauer <mail@david-bauer.net>
+Date: Sat, 25 Jul 2026 19:35:28 +0200
+Subject: [PATCH] interface: add rate propert
+
+Add a new interface property for advertised rate. This property can be
+used to autmatically configure bandwidth-dependent shapers according to
+the exact line-rate.
+
+This depends on support from the upstream ISP.
+
+Signed-off-by: David Bauer <mail@david-bauer.net>
+---
+ interface.c             | 21 +++++++++++++
+ interface.h             | 27 +++++++++++++++++
+ proto-ext.c             | 65 +++++++++++++++++++++++++++++++++++++++++
+ scripts/netifd-proto.sh | 35 ++++++++++++++++++++++
+ ubus.c                  | 29 +++++++++++++++++-
+ 5 files changed, 176 insertions(+), 1 deletion(-)
+
+diff --git a/interface.c b/interface.c
+index 61a18f9..bad92d2 100644
+--- a/interface.c
++++ b/interface.c
+@@ -77,6 +77,9 @@ enum {
+ 	IFACE_ATTR_IP6WEIGHT,
+ 	IFACE_ATTR_TAGS,
+ 	IFACE_ATTR_CARRIER_LOSS_DELAY,
++	IFACE_ATTR_BANDWIDTH_DOWNLINK,
++	IFACE_ATTR_BANDWIDTH_UPLINK,
++	IFACE_ATTR_BANDWIDTH_IGNORE_ANNOUNCEMENT,
+ 	IFACE_ATTR_MAX
+ };
+ 
+@@ -109,6 +112,9 @@ static const struct blobmsg_policy iface_attrs[IFACE_ATTR_MAX] = {
+ 	[IFACE_ATTR_IP6WEIGHT] = { .name = "ip6weight", .type = BLOBMSG_TYPE_INT32 },
+ 	[IFACE_ATTR_TAGS] = { .name = "tags", .type = BLOBMSG_TYPE_ARRAY },
+ 	[IFACE_ATTR_CARRIER_LOSS_DELAY] = { .name = "carrier_loss_delay", .type = BLOBMSG_TYPE_INT32 },
++	[IFACE_ATTR_BANDWIDTH_DOWNLINK] = { .name = "bandwidth_downlink", .type = BLOBMSG_TYPE_INT32 },
++	[IFACE_ATTR_BANDWIDTH_UPLINK] = { .name = "bandwidth_uplink", .type = BLOBMSG_TYPE_INT32 },
++	[IFACE_ATTR_BANDWIDTH_IGNORE_ANNOUNCEMENT] = { .name = "bandwidth_ignore_announcement", .type = BLOBMSG_TYPE_BOOL },
+ };
+ 
+ const struct uci_blob_param_list interface_attr_list = {
+@@ -1002,6 +1008,21 @@ interface_alloc(const char *name, struct blob_attr *config, bool dynamic)
+ 					   assign, iface->name);
+ 	}
+ 
++	if ((cur = tb[IFACE_ATTR_BANDWIDTH_IGNORE_ANNOUNCEMENT]))
++		iface->bandwidth.ignore_announcement = blobmsg_get_bool(cur);
++
++	if ((cur = tb[IFACE_ATTR_BANDWIDTH_DOWNLINK])) {
++		iface->bandwidth.downlink.rate_kbps = blobmsg_get_u32(cur);
++		iface->bandwidth.downlink.source = IFB_MANUAL;
++		iface->bandwidth.downlink.type = IFR_LAYER_2;
++	}
++
++	if ((cur = tb[IFACE_ATTR_BANDWIDTH_UPLINK])) {
++		iface->bandwidth.uplink.rate_kbps = blobmsg_get_u32(cur);
++		iface->bandwidth.uplink.source = IFB_MANUAL;
++		iface->bandwidth.uplink.type = IFR_LAYER_2;
++	}
++
+ 	/* defaults */
+ 	iface->assignment_iface_id_selection = IFID_FIXED;
+ 	iface->assignment_fixed_iface_id = in6addr_any;
+diff --git a/interface.h b/interface.h
+index 5c83d97..6d31810 100644
+--- a/interface.h
++++ b/interface.h
+@@ -98,6 +98,32 @@ struct interface_assignment_class {
+ 	char name[];
+ };
+ 
++enum interface_bandwidth_source {
++	IFB_NONE = 0,
++	IFB_PROTO,
++	IFB_MANUAL,
++};
++
++enum interface_rate_type {
++	IFR_NONE = 0,
++	IFR_LAYER_2,
++	IFR_LAYER_3,
++};
++
++struct interface_bandwidth_direction {
++	enum interface_rate_type type;
++	enum interface_bandwidth_source source;
++	uint32_t rate_kbps;
++	time_t last_update;
++};
++
++struct interface_bandwidth {
++	struct interface_bandwidth_direction downlink;
++	struct interface_bandwidth_direction uplink;
++
++	bool ignore_announcement;
++};
++
+ /*
+  * interface configuration
+  */
+@@ -147,6 +173,7 @@ struct interface {
+ 
+ 	/* interface that layer 3 communication will go through */
+ 	struct device_user l3_dev;
++	struct interface_bandwidth bandwidth;
+ 
+ 	struct blob_attr *config;
+ 	struct blob_attr *tags;
+diff --git a/proto-ext.c b/proto-ext.c
+index 2bedc0c..afd4e82 100644
+--- a/proto-ext.c
++++ b/proto-ext.c
+@@ -237,6 +237,65 @@ proto_ext_free(struct interface_proto_state *proto)
+ 	free(state);
+ }
+ 
++enum proto_ext_bandwidth_fields {
++	PROTO_EXT_BANDWIDTH_DOWNLINK_SOURCE,
++	PROTO_EXT_BANDWIDTH_DOWNLINK_RATE_TYPE,
++	PROTO_EXT_BANDWIDTH_DOWNLINK_RATE,
++	PROTO_EXT_BANDWIDTH_UPLINK_SOURCE,
++	PROTO_EXT_BANDWIDTH_UPLINK_RATE_TYPE,
++	PROTO_EXT_BANDWIDTH_UPLINK_RATE,
++	__PROTO_EXT_BANDWIDTH_MAX
++};
++
++static const struct blobmsg_policy proto_ext_bandwidth_policy[__PROTO_EXT_BANDWIDTH_MAX] = {
++	[PROTO_EXT_BANDWIDTH_DOWNLINK_SOURCE] = { .name = "downlink_source", .type = BLOBMSG_TYPE_STRING },
++	[PROTO_EXT_BANDWIDTH_DOWNLINK_RATE_TYPE] = { .name = "downlink_type", .type = BLOBMSG_TYPE_STRING },
++	[PROTO_EXT_BANDWIDTH_DOWNLINK_RATE] = { .name = "downlink_rate", .type = BLOBMSG_TYPE_INT32 },
++	[PROTO_EXT_BANDWIDTH_UPLINK_SOURCE] = { .name = "uplink_source", .type = BLOBMSG_TYPE_STRING },
++	[PROTO_EXT_BANDWIDTH_UPLINK_RATE_TYPE] = { .name = "uplink_type", .type = BLOBMSG_TYPE_STRING },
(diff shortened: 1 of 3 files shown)


package/network/services/ppp/files/lib/netifd/ppp-up
```diff
@@ -2,6 +2,48 @@
...
+
+	[ -n "$down" ] && [ -n "$up" ] || return 1
+
+	if [ "$down" -gt 0 ]; then
+		BW_UPLINK="$up"
+	fi
+
+	if [ "$up" -gt 0 ]; then
+		BW_DOWNLINK="$down"
+	fi
```
openwrt-ai, line 42: The guards and the assignments are crossed: the `down` check assigns `$up` to `BW_UPLINK`, and the `up` check assigns `$down` to `BW_DOWNLINK`. With an asymmetric line where one direction is announced as `0` (e.g. `SRU=0#SRD=100000#`) this sets the *wrong* direction and drops the valid one, so a shaper fed from this would be configured backwards.

```suggestion
	if [ "$down" -gt 0 ]; then
		BW_DOWNLINK="$down"
	fi

	if [ "$up" -gt 0 ]; then
		BW_UPLINK="$up"
	fi
```


---

package/network/services/ppp/files/lib/netifd/ppp-up
```diff
@@ -2,6 +2,48 @@
 PPP_IPPARAM="$6"
 
 . /lib/netifd/netifd-proto.sh
+
+BW_DOWNLINK=
+BW_UPLINK=
+function parse_bandwidth() {
```
openwrt-ai, line 8: `function name() { ... }` is a bashism. This script is `#!/bin/sh` and runs under busybox ash on target, whose parser has no `function` keyword — it will abort the whole `ip-up` script with `Syntax error: "(" unexpected`, so PPP interfaces would come up without any netifd update at all. Every other function in this file and in `/lib/netifd/netifd-proto.sh` uses the POSIX form.

```suggestion
parse_bandwidth() {
```


---

package/network/services/ppp/files/lib/netifd/ppp-up
```diff
@@ -2,6 +2,48 @@
...
+	local ppp_auth_ack_msg="$(cat /var/run/pppd/${IFNAME}.authackmsg)"
```
openwrt-ai, line 9: Two problems with reading this file unconditionally:

1. The file only exists when the peer authenticated with PAP *and* sent an auth-ack message (the new pppd patch writes it from `upap_rauthack()` only). For CHAP/MS-CHAP sessions or acks without a message, `cat` prints `No such file or directory` to stderr on every single connect, which lands in syslog via pppd.
2. Nothing ever removes the file. After a session where the ISP stops announcing rates (or a reconnect that uses CHAP), the stale message from the previous session is still on disk and gets re-applied as current bandwidth. Consider unlinking it in `ppp-down` (or right after reading it here) so a session that announces nothing does not inherit the old values.

For (1):

```suggestion
	local ppp_auth_ack_msg="$(cat /var/run/pppd/${IFNAME}.authackmsg 2>/dev/null)"
```


---

package/network/services/ppp/patches/600-pppd-write-auth-ack-message-to-file.patch
```diff
@@ -0,0 +1,50 @@
...
+ #include <string.h>
++#include <sys/param.h>
+ 
+ #include "pppd-private.h"
+ #include "options.h"
+@@ -478,6 +479,18 @@ upap_rauthack(upap_state *u, u_char *inp, int id, int len)
+ 	}
+     }
+ 
++    if (msg && ifname[0]) {
```
openwrt-ai, line 33: Is `msg` guaranteed to be initialised on every path that reaches here? In pppd's `upap_rauthack()` the locals are declared as `u_char msglen; char *msg;` and `msg` is only assigned inside the `if (msglen > 0)` branch of the message parsing — the `len < 1` path and the `msglen == 0` path fall through without touching it. Upstream never dereferences `msg` after that block, so it does not matter there, but this hunk is placed *after* the block closes, so `if (msg && ifname[0])` would read an indeterminate pointer and potentially `fwrite()` from it.

If that reading is right, either initialise (`char *msg = NULL;` / `msglen = 0;`) as part of this patch, or move the file write inside the `msglen > 0` branch where `msg`/`msglen` are known good.

---

package/network/config/netifd/patches/0001-interface-add-rate-propert.patch
```diff
@@ -0,0 +1,316 @@
+From 5424f1bce1a57086e72b7573cb5cd9834676e962 Mon Sep 17 00:00:00 2001
```
openwrt-ai, line 1: `package/network/config/netifd` has no `patches/` directory today — the package is pinned to an upstream commit via `PKG_SOURCE_VERSION`/`PKG_MIRROR_HASH`, and netifd changes normally land in `git.openwrt.org/project/netifd.git` first, followed by a `PKG_SOURCE_DATE`/`PKG_SOURCE_VERSION` bump here. Is carrying this as a local patch intended only for the draft/demo stage, or is it meant to be merged this way? If the former it would be worth saying so in the commit message so reviewers do not read it as the final shape.

Also, if the patch does stay, OpenWrt patch files use a three-digit ordering prefix (see the sibling `package/network/services/ppp/patches/` in this same PR: `600-...`), not `git format-patch`'s `0001-`.

---

package/network/config/netifd/patches/0001-interface-add-rate-propert.patch
```diff
@@ -0,0 +1,316 @@
...
+From: David Bauer <mail@david-bauer.net>
+Date: Sat, 25 Jul 2026 19:35:28 +0200
+Subject: [PATCH] interface: add rate propert
```
openwrt-ai, line 4: nit: `propert` -> `property` (the typo is also baked into the patch filename). The body on line 7 has `autmatically` -> `automatically` as well. Worth noting the subject says "rate property" while the properties actually added are `bandwidth_downlink` / `bandwidth_uplink` / `bandwidth_ignore_announcement`, and the OpenWrt commit that carries this is titled "add bandwidth property patch" — picking one term (`bandwidth`) everywhere would make this easier to grep for.

```suggestion
Subject: [PATCH] interface: add rate property
```


---

package/network/config/netifd/patches/0001-interface-add-rate-propert.patch
```diff
@@ -0,0 +1,316 @@
...
++		iface->bandwidth.downlink.rate_kbps = blobmsg_get_u32(cur);
++		iface->bandwidth.downlink.source = IFB_MANUAL;
++		iface->bandwidth.downlink.type = IFR_LAYER_2;
++	}
++
++	if ((cur = tb[IFACE_ATTR_BANDWIDTH_UPLINK])) {
++		iface->bandwidth.uplink.rate_kbps = blobmsg_get_u32(cur);
++		iface->bandwidth.uplink.source = IFB_MANUAL;
++		iface->bandwidth.uplink.type = IFR_LAYER_2;
++	}
```
openwrt-ai, line 62: The new config is parsed in `interface_alloc()` only, but nothing propagates it on a config reload. [`interface_change_config()`](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/interface.c#L1531-L1625) copies the non-proto interface fields one by one from `if_new` to `if_old` and then frees `if_new` — see [`if_old->carrier_loss_delay = if_new->carrier_loss_delay;`](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/interface.c#L1606) for the equivalent scalar. `bandwidth` is not in that list, and since these attributes are not part of any `proto_handler->config_params` the `uci_blob_check_equal()` check above will not force a reload either. Net effect: editing `bandwidth_downlink`/`bandwidth_uplink`/`bandwidth_ignore_announcement` in `/etc/config/network` and running `reload_config` silently keeps the old values until the interface is torn down and recreated.

`if_old->bandwidth = if_new->bandwidth;` in `interface_change_config()` would fix it, though see the separate note on `proto_ext_parse_bandwidth()` — the manual and proto-supplied values sharing one struct makes "what should a reload restore?" ambiguous.

---

package/network/config/netifd/patches/0001-interface-add-rate-propert.patch
```diff
@@ -0,0 +1,316 @@
...
++	struct blob_attr *tb[__PROTO_EXT_BANDWIDTH_MAX];
++	struct interface_bandwidth *bw = &iface->bandwidth;
++
++	blobmsg_parse(proto_ext_bandwidth_policy, __PROTO_EXT_BANDWIDTH_MAX, tb, blobmsg_data(attr), blobmsg_data_len(attr));
++
++	if (tb[PROTO_EXT_BANDWIDTH_DOWNLINK_RATE]) {
++		bw->downlink.rate_kbps = blobmsg_get_u32(tb[PROTO_EXT_BANDWIDTH_DOWNLINK_RATE]);
++		bw->downlink.source = IFB_PROTO;
++		bw->downlink.type = IFR_NONE;
++		bw->downlink.last_update = time(NULL);
```
openwrt-ai, line 151: A proto announcement overwrites a manually configured value in place, and there is no way back: `source` flips from `IFB_MANUAL` to `IFB_PROTO` and the UCI-configured `rate_kbps` is gone. Nothing resets the proto-sourced values when the interface goes down / the proto restarts either, so a subsequent session that announces nothing keeps serving the previous session's rates over ubus as if they were current — and, per the other comment, a config reload will not restore the manual value.

Keeping the config-supplied and proto-supplied values in separate members (and picking between them at dump time, preferring proto when fresh) would make both the reload and the teardown cases well-defined. `last_update` is already stored here but never read anywhere — it looks like it was meant for exactly this kind of staleness handling; either wire it up or drop it.

---

package/network/config/netifd/patches/0001-interface-add-rate-propert.patch
```diff
@@ -0,0 +1,316 @@
...
++	PROTO_EXT_BANDWIDTH_UPLINK_RATE_TYPE,
++	PROTO_EXT_BANDWIDTH_UPLINK_RATE,
++	__PROTO_EXT_BANDWIDTH_MAX
++};
++
++static const struct blobmsg_policy proto_ext_bandwidth_policy[__PROTO_EXT_BANDWIDTH_MAX] = {
++	[PROTO_EXT_BANDWIDTH_DOWNLINK_SOURCE] = { .name = "downlink_source", .type = BLOBMSG_TYPE_STRING },
++	[PROTO_EXT_BANDWIDTH_DOWNLINK_RATE_TYPE] = { .name = "downlink_type", .type = BLOBMSG_TYPE_STRING },
++	[PROTO_EXT_BANDWIDTH_DOWNLINK_RATE] = { .name = "downlink_rate", .type = BLOBMSG_TYPE_INT32 },
++	[PROTO_EXT_BANDWIDTH_UPLINK_SOURCE] = { .name = "uplink_source", .type = BLOBMSG_TYPE_STRING },
```
openwrt-ai, line 134: `downlink_source` and `uplink_source` are declared in the policy and are actually sent by `proto_add_bandwidth()` in `netifd-proto.sh`, but `proto_ext_parse_bandwidth()` never looks at `tb[...SOURCE]` — `source` is unconditionally set to `IFB_PROTO`. So the third field of the `rate/type/source` string that every proto handler has to pass is silently discarded (and `ppp-up` dutifully passes `proto`).

Either drop the two `*_source` policy entries and the corresponding `json_add_string`s from `proto_add_bandwidth()`, or define what a proto is allowed to claim there and honour it.

---

package/network/config/netifd/patches/0001-interface-add-rate-propert.patch
```diff
@@ -0,0 +1,316 @@
...
+diff --git a/scripts/netifd-proto.sh b/scripts/netifd-proto.sh
+index 4abc8f3..de30e1f 100644
+--- a/scripts/netifd-proto.sh
++++ b/scripts/netifd-proto.sh
+@@ -75,6 +75,8 @@ proto_init_update() {
+ 	PROTO_DNS_SEARCH=
+ 	PROTO_NEIGHBOR=
+ 	PROTO_NEIGHBOR6=
++	PROTO_BW_UP=
++	PROTO_BW_DOWN=
```
openwrt-ai, line 218: `PROTO_BW_UP` / `PROTO_BW_DOWN` are reset here but never assigned or read anywhere else in the patch — dead code. This looks like the leftover of the deferred-accumulator pattern that every other helper follows; see the comment on `proto_add_bandwidth()` below.

---

package/network/config/netifd/patches/0001-interface-add-rate-propert.patch
```diff
@@ -0,0 +1,316 @@
...
+@@ -310,6 +312,39 @@ _proto_push_route() {
+ 	json_close_object
+ }
+ 
++proto_add_bandwidth() {
++	local downlink_str="$1"
++	local uplink_str="$2"
++
++	json_add_object "bandwidth"
++
```
openwrt-ai, line 231: This is the only `proto_add_*` helper that writes into the JSON buffer directly. All the others — [`proto_add_ipv4_address()`](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/scripts/netifd-proto.sh#L127-L133), `proto_add_dns_server()`, `proto_add_ipv4_route()` … — just `append` to a `PROTO_*` variable and are serialised centrally in [`proto_send_update()`](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/scripts/netifd-proto.sh#L338-L344). That has two consequences here:

- Calling `proto_add_bandwidth` while a nested object is open (`proto_add_tunnel` / `proto_add_data`, which leave `PROTO_NESTED_OPEN` set) nests the `bandwidth` table *inside* `tunnel`/`data`, where `notify_attr` will never find it. The deferred helpers are immune to that by construction.
- `json_add_object "bandwidth"` runs unconditionally, so a call with two empty arguments still emits an empty `bandwidth: {}` in the update.

Storing the two strings in the already-declared `PROTO_BW_UP`/`PROTO_BW_DOWN` and emitting the table from `proto_send_update()` (only when at least one is set) would match the surrounding code and fix both.

Separately, packing `rate/type/source` into one slash-separated argument is the internal on-the-wire format of `_proto_push_*`, not the public helper signature — the public helpers take separate positional args. `proto_add_bandwidth <down_rate> <down_type> <up_rate> <up_type>` would read more like its neighbours and remove the manual re-splitting.

---

package/network/config/netifd/patches/0001-interface-add-rate-propert.patch
```diff
@@ -0,0 +1,316 @@
...
++			const char *rate_type = blobmsg_get_string(tb[PROTO_EXT_BANDWIDTH_UPLINK_RATE_TYPE]);
++			if (!strcmp(rate_type, "layer2"))
++				bw->uplink.type = IFR_LAYER_2;
++			else if (!strcmp(rate_type, "layer3"))
++				bw->uplink.type = IFR_LAYER_3;
++		}
++	}
++}
++
++
```
openwrt-ai, line 178: nit: two consecutive blank lines are added before `proto_ext_parse_route_list()`; the rest of `proto-ext.c` separates top-level functions with a single blank line. (Dropping one line also means the `@@ -237,6 +237,65 @@` hunk header and the diffstat need to change, so this is easiest to fix by regenerating the patch rather than editing it by hand.)

---

openwrt-ai (commented):
Reviewed 3 commits (first review of this PR). Findings are inline; the two that block a working end-to-end path on target are the `function` bashism in `ppp-up` (busybox ash cannot parse it) and the swapped up/down assignment in `parse_bandwidth()`.

Commit messages match their diffs, so no commit-vs-diff mismatches to report — the empty commit bodies, missing `Signed-off-by:` lines and missing `PKG_RELEASE` bumps are already covered by the formality bot above.

---

PR #24619, opened 2026-08-08
kernel: add mt7530 LED support

neheb:
While mt7530 switches support a GPIO mode for controlling LEDs, it doesn't integrate well with DSA. Add an LLM generated patch to add support for that. It mirrors qca8k in its implementation, so similar dts changes are needed.

Low chance this actually works. I pointed it to the datasheet and it says everything's fine but I doubt it.

I don't have the hardware to test this so putting this as a draft.


target/linux/generic/hack-6.18/740-net-dsa-mt7530-add-LED-controller-support.patch
```diff
@@ -0,0 +1,430 @@
...
+ config NET_DSA_MV88E6060
+ 	tristate "Marvell 88E6060 ethernet switch chip support"
+ 	select NET_DSA_TAG_TRAILER
+--- a/drivers/net/dsa/Makefile
++++ b/drivers/net/dsa/Makefile
+@@ -4,6 +4,7 @@ bcm-sf2-objs			:= bcm_sf2.o bcm_sf2_cfp.
+ obj-$(CONFIG_NET_DSA_LOOP)	+= dsa_loop.o
+ obj-$(CONFIG_NET_DSA_KS8995) 	+= ks8995.o
+ obj-$(CONFIG_NET_DSA_MT7530)	+= mt7530.o
++obj-$(CONFIG_NET_DSA_MT7530_LEDS_SUPPORT) += mt7530-leds.o
```
openwrt-ai, line 61: This builds `mt7530-leds.o` as a **separate** object rather than linking it into `mt7530.o`, which is not what qca8k does. `NET_DSA_MT7530_LEDS_SUPPORT` is a `bool`, so this expands to `obj-y += mt7530-leds.o` — the LED code always lands in vmlinux, while `mt7530.o` may be a module. Then `mt7530.c`'s call to `mt7530_setup_led_ctrl()` and `mt7530-leds.c`'s calls to `mt7530_read()`/`mt7530_rmw()` (neither is `EXPORT_SYMBOL`'d by this patch) cross the module boundary and fail to link with `CONFIG_NET_DSA_MT7530=m`.

qca8k does it as a composite object instead — [`drivers/net/dsa/qca/Makefile`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/dsa/qca/Makefile#L3-L7):

```
obj-$(CONFIG_NET_DSA_QCA8K)	+= qca8k.o
qca8k-y 			+= qca8k-common.o qca8k-8xxx.o
ifdef CONFIG_NET_DSA_QCA8K_LEDS_SUPPORT
qca8k-y				+= qca8k-leds.o
endif
```

The mt7530 case is slightly trickier because `mt7530.c` already produces `mt7530.o` directly, so the module has to be turned into a composite (rename the core file, or pick a different module name) before a second object can be added to it. Either way the end state needs `mt7530-leds.o` linked into the same module as `mt7530.o`, not standing alone. Same hunk at `target/linux/generic/hack-6.12/740-net-dsa-mt7530-add-LED-controller-support.patch:61`.

All the target configs this PR touches use `CONFIG_NET_DSA_MT7530=y`, which is why CI doesn't catch it.

---

target/linux/generic/hack-6.18/740-net-dsa-mt7530-add-LED-controller-support.patch
```diff
@@ -0,0 +1,430 @@
...
+ create mode 100644 drivers/net/dsa/mt7530-leds.c
+ create mode 100644 drivers/net/dsa/mt7530_leds.h
+
+--- a/drivers/net/dsa/Kconfig
++++ b/drivers/net/dsa/Kconfig
+@@ -63,6 +63,18 @@ config NET_DSA_MT7530_MMIO
+ 	  are directly mapped into the SoCs register space rather than being
+ 	  accessible via MDIO.
+ 
++config NET_DSA_MT7530_LEDS_SUPPORT
```
openwrt-ai, line 40: CI job *Check Kernel patches (realtek, rtl838x)* fails on this:

```
target/linux/realtek/patches-6.18/718-net-dsa-add-support-for-rtl838x-switch.patch
```

This generic hack inserts 12 lines into `drivers/net/dsa/Kconfig` and 1 into `drivers/net/dsa/Makefile` ahead of the target patches, which shifts the hunk offsets of [`718-net-dsa-add-support-for-rtl838x-switch.patch` (`@@ -86,6` / `@@ -23,5`)](``https://github.com/openwrt/openwrt/blob/aa483715aaa5bd7be40ccfaff9c0efe9d5c85c84/target/linux/realtek/patches-6.18/718-net-dsa-add-support-for-rtl838x-switch.patch#L25-L36).`` Per the project rules, regenerate it with `make target/linux/refresh` with realtek/rtl838x selected in `.config`, and include the refreshed patch in this PR.

---

target/linux/generic/hack-6.18/740-net-dsa-mt7530-add-LED-controller-support.patch
```diff
@@ -0,0 +1,430 @@
...
++	if (brightness) {
++		/* Drive the LED pin as a GPIO output. */
++		mt7530_rmw(led->priv, MT7530_LED_IO_MODE, bit, 0);
++		mt7530_rmw(led->priv, MT7530_LED_GPIO_DIR, bit, bit);
++		mt7530_rmw(led->priv, MT7530_LED_GPIO_DATA, bit, bit);
++		mt7530_rmw(led->priv, MT7530_LED_GPIO_OE, bit, bit);
++		mt7530_rmw(led->priv, MT7530_LED_EN, bit, bit);
++	} else {
++		mt7530_rmw(led->priv, MT7530_LED_EN, bit, 0);
++		mt7530_rmw(led->priv, MT7530_LED_GPIO_DATA, bit, 0);
```
openwrt-ai, line 92: The off path never takes the pin out of PHY mode, so it is not symmetric with the on path above. `MT7530_LED_IO_MODE` defaults to PHY mode ([mt7530.h:697-698](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/dsa/mt7530.h#L697-L698)), which means:

- an LED currently under hw control keeps `MT7530_LED_IO_MODE` set after userspace writes brightness 0, so `mt7530_cled_hw_control_status()` still reports "hardware controlled" for an LED the driver just took over;
- the `LEDS_DEFSTATE_OFF` probe path calls this with 0, so a freshly-probed LED is left in PHY mode and `mt7530_led_brightness_get()` then reads `MT7530_LED_GPIO_DATA`, which is meaningless while the pin is PHY-driven.

Switching to GPIO mode in both branches fixes both:

```suggestion
+	} else {
+		/* Drive the LED pin as a GPIO output. */
+		mt7530_rmw(led->priv, MT7530_LED_IO_MODE, bit, 0);
+		mt7530_rmw(led->priv, MT7530_LED_GPIO_DIR, bit, bit);
+		mt7530_rmw(led->priv, MT7530_LED_GPIO_OE, bit, bit);
+		mt7530_rmw(led->priv, MT7530_LED_GPIO_DATA, bit, 0);
+		mt7530_rmw(led->priv, MT7530_LED_EN, bit, 0);
```

---

target/linux/generic/hack-6.18/740-net-dsa-mt7530-add-LED-controller-support.patch
```diff
@@ -0,0 +1,430 @@
...
++
++static int
++mt7530_cled_hw_control_get(struct led_classdev *ldev, unsigned long *rules)
++{
++	struct mt7530_led *led = container_of(ldev, struct mt7530_led, cdev);
++
++	if (!mt7530_cled_hw_control_status(ldev))
++		return -EINVAL;
++
++	*rules = led->rules;
```
openwrt-ai, line 172: `rules` is only ever a software cache written by `mt7530_cled_hw_control_set()`, but `mt7530_cled_hw_control_status()` above derives hw-control from the actual `MT7530_LED_IO_MODE` register. Since PHY mode is the reset default, right after probe `hw_control_get()` succeeds (status is true) and reports `*rules = 0` — hardware control active with no trigger bits, which the netdev trigger will read back as "nothing enabled". qca8k reconstructs the rules from the register instead ([qca8k-leds.c hw_control_get](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/dsa/qca/qca8k-leds.c#L313-L349)). Since the only supported rule here is `TRIGGER_NETDEV_LINK`, returning `BIT(TRIGGER_NETDEV_LINK)` whenever the IO_MODE bit is set would keep the two in sync without a shadow field.

---

target/linux/airoha/an7581/config-6.18
```diff
@@ -236,6 +236,7 @@ CONFIG_NET_AIROHA_NPU=y
 CONFIG_NET_DEVLINK=y
 CONFIG_NET_DSA=y
 CONFIG_NET_DSA_MT7530=y
+CONFIG_NET_DSA_MT7530_LEDS_SUPPORT=y
```
openwrt-ai, line 239: Is enabling this on non-MT7530 subtargets intentional? `mt7530_setup_led_ctrl()` is only called from `mt7530_setup()`, which is the `sw_setup` for `ID_MT7530` and `ID_MT7621` only — `ID_MT7531`, `ID_MT7988`, `ID_EN7581` and `ID_AN7583` use `mt7531_setup`/`mt7988_setup` ([mt753x_table](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/dsa/mt7530.c#L3316-L3380)), and the LED/GPIO register block is documented as "MT7530 only" ([mt7530.h:685-704](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/dsa/mt7530.h#L685-L704)). So on airoha an7581/an7583, econet en7528, mediatek filogic/mt7622/mt7629 the LED code is compiled in but never reached. Only ramips/mt7621, mediatek/mt7623 (and rockchip if an external MT7530 is in play) can actually use it.

---

target/linux/generic/hack-6.18/740-net-dsa-mt7530-add-LED-controller-support.patch
```diff
@@ -0,0 +1,430 @@
...
++		/* Reg represent the led number of the port.
++		 * Each port can have at most 3 leds attached:
++		 * 1. is gigabit led
++		 * 2. is mbit led
++		 * 3. additional status led
++		 */
++		if (fwnode_property_read_u32(led, "reg", &led_num))
++			continue;
++
++		if (led_num >= MT7530_LED_NUM) {
```
openwrt-ai, line 217: nit: `led_num` is declared `int` (line 197) but filled by `fwnode_property_read_u32()`, so a DT `reg` above `INT_MAX` lands here as negative, passes this upper-bound-only check, and `MT7530_LED_PORT_INDEX()` then indexes `priv->leds->leds[]` out of bounds. Declaring it `u32` (and the same for `port_num` in `mt7530_setup_led_ctrl()`) makes the single bound check sufficient. qca8k carries the same pattern, so this is not a deviation from the model driver — just worth not copying.

---

target/linux/generic/hack-6.18/740-net-dsa-mt7530-add-LED-controller-support.patch
```diff
@@ -0,0 +1,430 @@
+From 8daf4f7664fecfd1742be155bb5399d3b31e0bcc Mon Sep 17 00:00:00 2001
```
openwrt-ai, line 1: nit: is `hack-6.18/` the right home for this? In-tree, `hack-*` is for changes that will never be submitted upstream, while new driver features intended for mainline live in `pending-*` — e.g. the qca8k equivalents this patch mirrors sit in `target/linux/generic/pending-6.18/711-0*-net-dsa-qca8k-*.patch`. If the intent is to eventually send this to netdev, `pending-6.18/` (and `pending-6.12/`) would match the convention.

---

openwrt-ai (commented):
Reviewed 1 new commit (`aa48371`). The commit message matches the diff, so no commit-check findings. Inline comments cover the Makefile linkage of the new LED object, the realtek patch-refresh CI failure this patch triggers, and a few driver-behaviour points; the Kconfig hunk offsets and the `MT7530_LED_BIT()` bit layout both check out against the register map in `mt7530.h`.

---

alexceltare2:
Thanks. Work is being done by these brilliant people @cjdelisle @naseef  @Sirherobrine23 @merbanan. MT7530 is more annoying than originally thought. The vendor DSA driver is inconsistent, needs calibration and is a MCM module that is interchangeable.

peterwillcn (approved):

merbanan:
IIRC switch leds already work without this type of change.

naseef:
Isn't this supposed to be in pinctrl and gphy drivers?

merbanan:
The mt7530/1 has support to drive gpios and other things. I dont think any of these are wired up on Airoha/Econet SoCs. This should not be applied until someone actually tests it.

cjdelisle:
So I can only speak for EcoNet, maybe other applications of MT7530 are different...

There are 4 lights on the unit. If you leave them alone, they just indicate port linkup and activity which is fine in pretty much every case.

If you want to control them manually, you can actually change them into GPIOs in the pinmux and they're then controllable from the Airoha-GPIO we already have.

Controlling them from inside of the MT7530 is "neat" but it doesn't really serve a purpose.

So unless there's a specific board that this solves a specific burning problem for, I think this should be send directly upstream and discussed as a patch to make the MT7530 driver "more capable". Otherwise it's just One More Patch that's going to need to be ported every time the kernel is updated.

Rondom:
I agree that working upstream is the way to go. There are designs using dedicated MT753x chips not integrated into the SoC, some of them even using non-MTK/EcoNet/Airoha chips such as BPI-R2 Pro comes to mind. So if it really does not work on Econet chips, that is a pity but nothing that speaks against the feature in itself.

> If you want to control them manually, you can actually change them into GPIOs in the pinmux and they're then controllable from the Airoha-GPIO we already have.
> 
> Controlling them from inside of the MT7530 is "neat" but it doesn't really serve a purpose.

From a user's perspective it is not the same. While it is indeed possible to achieve the same in device-tree for SoCs with integrated switch, this is not something that is done very quickly by a user. With the LEDs integrated into the LED subsystem, once can control the triggers from LUCI and use the LEDs for something else (useful for devices, where LAN LEDs are on the front panel and not integrated into the RJ45 ports themselves) or switch them off completely (router in bedroom situation).

Sirherobrine23:
Looking at it this way, it might be interesting for en751221 devices. the MCM switch isn't connected to the SoC GPIOs, so it would be nice to have that

cjdelisle:
I was not aware that the MCM switch's LEDs do not route back to the pinmux - though once you said it, it's quite obvious.

So it does make sense that we should have LED control in the MT7530, but I maintain that this really belongs in the kernel.

Sirherobrine23:
the pinmux only controls the FE switch LEDs. since many devices feature a cascaded switch, their LED controls differ aren't connected to the SoC pinmux and must instead be managed via the MDIO datapath. The question then arises. should this be handled by the dsa driver or the PHY?

cjdelisle:
> should this be handled by the dsa driver or the PHY

From a glance, it looks as though it requires hitting registers in the MT7530 so I guess it would be in the DSA driver, right?

PR #24620, opened 2026-08-09
tools/cmake: update to 4.4.2

zheshifandian:
Release notes:https://cmake.org/cmake/help/latest/release/4.4.html

No need refresh patches.

The --no-warn-unused-cli command-line option is deprecated. The new spelling is -Wno-unused-cli.

### DIFF
--- a/include/cmake.mk
+++ b/include/cmake.mk
@@ -93,7 +93,7 @@ define Build/Configure/Default
 		CXXFLAGS="$(TARGET_CXXFLAGS) $(EXTRA_CXXFLAGS) $(TARGET_CPPFLAGS) $(EXTRA_CPPFLAGS)" \
 		LDFLAGS="$(TARGET_LDFLAGS) $(EXTRA_LDFLAGS)" \
 		cmake \
-			--no-warn-unused-cli \
+			-Wno-unused-cli \
 			-DCMAKE_SYSTEM_NAME=Linux \
 			-DCMAKE_SYSTEM_VERSION=1 \
 			-DCMAKE_SYSTEM_PROCESSOR=$(ARCH) \
@@ -146,7 +146,7 @@ define Host/Configure/Default
 		CXXFLAGS="$(HOST_CFLAGS)" \
 		LDFLAGS="$(HOST_LDFLAGS)" \
 		cmake \
-			--no-warn-unused-cli \
+			-Wno-unused-cli \
 			-DCMAKE_BUILD_TYPE=Release \
 			-DCMAKE_C_COMPILER_LAUNCHER="$(CMAKE_C_COMPILER_LAUNCHER)" \
 			-DCMAKE_C_COMPILER="$(CMAKE_HOST_C_COMPILER)" \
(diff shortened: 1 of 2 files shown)


include/cmake.mk
```diff
@@ -93,7 +93,7 @@ define Build/Configure/Default
 		CXXFLAGS="$(TARGET_CXXFLAGS) $(EXTRA_CXXFLAGS) $(TARGET_CPPFLAGS) $(EXTRA_CPPFLAGS)" \
 		LDFLAGS="$(TARGET_LDFLAGS) $(EXTRA_LDFLAGS)" \
 		cmake \
-			--no-warn-unused-cli \
+			-Wno-unused-cli \
```
openwrt-ai, line 96: nit: this hunk makes `include/cmake.mk` hard-require host cmake >= 4.4 — `-Wno-unused-cli` is new in 4.4, and `--no-warn-unused-cli` is kept only as a compatibility synonym (`Help/manual/cmake.1.rst`: "`.. deprecated:: 4.4` / Compatibility synonym for `-Wno-unused-cli`"). That is fine on main since `tools/cmake` is unconditionally built ([`tools-y += cmake`](https://github.com/openwrt/openwrt/blob/4bbd50876718f763dc2c6bb0c72fa3bda5a92ed9/tools/Makefile#L38)), but it couples a build-system change to a tool version bump: if only the `tools/cmake` bump is cherry-picked to a stable branch, or only this hunk is, the other half breaks. Same applies to line 149.

Consider splitting this into its own commit (`build: use -Wno-unused-cli for cmake` or similar) — that is also how `include/cmake.mk` changes are normally scoped in-tree (e.g. [`build: pass CPPFLAGS to cmake build`](https://github.com/openwrt/openwrt/commit/7bf8d568f6c83411041e504e0ca231d4081c2534), [`build: suppress cmake warnings about unused vars`](https://github.com/openwrt/openwrt/commit/29a56e0412668f197353001254125ae40824db1e)).

---

openwrt-ai (commented):
## Commit checks

- `4bbd5087` "tools/cmake: update to 4.4.2" — subject scope says `tools/cmake:`, but the diff also changes `include/cmake.mk` (the treewide target and host cmake invocations). The body does mention the option rename, so it is not undocumented, but the build-system change is a separate concern from the version bump and is conventionally committed under a `build:` / `cmake.mk:` subject. See the inline note for the practical consequence.

Verified against upstream:
- `v4.4.2` exists in Kitware/CMake (`git ls-remote`, `f8b8df884725fb310aec9c03d71fd1877fe22f4f`).
- `PKG_HASH` matches the sha256 of the released `cmake-4.4.2.tar.gz` (`1db9e61e60b6e0874c86386340b910382f3c5e75b9fbfb44d122063129a2789d`).
- "No need refresh patches" holds: all six patches in `tools/cmake/patches/` apply to 4.4.2 with no fuzz or offset.
- The deprecation claim is accurate — `Help/release/4.4.rst`: "The `--no-warn-unused-cli` command-line option is deprecated. The new spelling is `-Wno-unused-cli`", and `Source/cmake.cxx` now calls `warnDeprecated("--no-warn-unused-cli", "-Wno-unused-cli")`, so the switch does silence a real new warning.

---

PR #24621, opened 2026-08-09
wifi-scripts: fix wdev fallback to use sprintf instead of printf

astrouski-google:
Previously, when printf was used, it didn't work because printf
returns the number of bytes written instead of the real content.
There was an attempt to fix that issue by using a loop, which
introduced several new bugs. It prevented orphan interfaces from
being correctly cleaned up since ubus wouldn't be called. Another
bug was that if we had several interfaces on the same radio, only
the last one would be applied because the loop called them
sequentially.

Update the command to use sprintf so it provides the correct
dictionary.

Signed-off-by: Alexander Astrovsky <astrouski@google.com>

### DIFF
--- a/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh
+++ b/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh
@@ -294,15 +294,7 @@ function setup() {
 		wdev_data[v.config.ifname] = config;
 	}
 
-	for (let ifname in active_ifnames) {
-		if (!wdev_data[ifname])
-			continue;
-
-		let if_config = {
-			[ifname]: wdev_data[ifname]
-		};
-		system(`ucode /usr/share/hostap/wdev.uc ${data.phy}${data.phy_suffix} set_config '${if_config}'`);
-	}
+	system([ "ucode", "/usr/share/hostap/wdev.uc", data.phy + data.phy_suffix, "set_config", sprintf("%J", wdev_data), ...active_ifnames ]);
 
 	if (fs.access('/usr/sbin/wpa_supplicant', 'x'))
 		supplicant.setup(supplicant_data, data);
--- a/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh
+++ b/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh
@@ -224,7 +224,7 @@ function setup() {
 		if (v.config.encryption == 'owe' && v.config.owe_transition) {
 			mode_idx = idx[mode]++;
 			v.config.owe_transition_ifname = data.ifname_prefix + mode + mode_idx;
-			push(active_ifnames, v.config.ifname);
+			push(active_ifnames, v.config.owe_transition_ifname);
 		}
 
 		switch (mode) {


package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh
```diff
@@ -294,15 +294,7 @@ function setup() {
...
-	for (let ifname in active_ifnames) {
-		if (!wdev_data[ifname])
-			continue;
-
-		let if_config = {
-			[ifname]: wdev_data[ifname]
-		};
-		system(`ucode /usr/share/hostap/wdev.uc ${data.phy}${data.phy_suffix} set_config '${if_config}'`);
-	}
+	system(`ucode /usr/share/hostap/wdev.uc ${data.phy}${data.phy_suffix} set_config '${sprintf("%J", wdev_data)}' ${join(' ', active_ifnames)}`);
```
openwrt-ai, line 297: The JSON is interpolated into a single-quoted `/bin/sh -c` string, so any value containing an apostrophe breaks out of the quoting. `%J` escapes `"` and `\` but not `'`, and the dict carries user-supplied SSIDs ([`ssid: v.config.ssid` at mac80211.sh:273](``https://github.com/openwrt/openwrt/blob/b3887cd93f3960a4191340ec1aa3ae9d642e034b/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh#L273)),`` so an adhoc/mesh/monitor SSID like `Bob's AP` yields a shell syntax error at best and command execution from UCI at worst. The shell implementation avoids this by passing the dump as a properly quoted argument ([`wdev_tool ... set_config "$(json_dump)" $active_ifnames`](``https://github.com/openwrt/openwrt/blob/b3887cd93f3960a4191340ec1aa3ae9d642e034b/package/network/config/wifi-scripts/files/lib/netifd/wireless/mac80211.sh#L1251)).``

Since this call is being rewritten anyway, use the array form of `system()` — ucode `execvp()`s it directly, so no quoting and no `join()` are needed. `wdev.uc` already uses that form for its own `iw` invocations ([wdev.uc:39-45](``https://github.com/openwrt/openwrt/blob/b3887cd93f3960a4191340ec1aa3ae9d642e034b/package/network/config/wifi-scripts/files/usr/share/hostap/wdev.uc#L39-L45)).``

```suggestion
	system([ "ucode", "/usr/share/hostap/wdev.uc", data.phy + data.phy_suffix, "set_config", sprintf("%J", wdev_data), ...active_ifnames ]);
```


---

package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh
```diff
@@ -294,15 +294,7 @@ function setup() {
 		wdev_data[v.config.ifname] = config;
 	}
```
openwrt-ai, line 296: nit: this change makes `active_ifnames` the `keep_devices` argument of `wdev.uc set_config` rather than just a loop source, so its contents now matter directly. In the OWE-transition branch the list gets `v.config.ifname` pushed a second time instead of the newly allocated name — [`push(active_ifnames, v.config.ifname);` at mac80211.sh:227](``https://github.com/openwrt/openwrt/blob/b3887cd93f3960a4191340ec1aa3ae9d642e034b/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh#L227),`` two lines after `v.config.owe_transition_ifname` is computed — so the transition ifname never reaches `keep_devices`. It looks like a copy/paste slip rather than intent; pre-existing and outside this diff, but worth a separate one-liner while you are in this area.

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the diff (single `set_config` call restores the per-radio dict and unconditional invocation, matching the shell implementation's `wdev_tool ... set_config "$(json_dump)" $active_ifnames`). Two inline notes, one about shell quoting of the interpolated JSON.

---

PolynomialDivision:
hey @astrouski-google whats the status? :)

astrouski-google:
@PolynomialDivision 
I've add another commit to address OWE copy/paste issue and also update ubus command to use [] instead of building string directly, pls take a look. 

I can update branch with latest main after review.

astrouski-google:
Also can you point out who can I add as reviewer? Looks like I need 6 approves to be able to submit that PR.

PolynomialDivision:
It is enough if I approve it. In the end I won't merge it via normal github button, instead we have our own scripts for it. Please fix the identation of the owe commit. Also please address the claude review.

openwrt-ai (commented):
Reviewed 2 new commits (the first was amended since the last review). Both inline comments from the previous round are addressed: the call now uses the array form of `system()` (no `/bin/sh -c` quoting, so an SSID containing `'` is safe), and the OWE-transition push was fixed. No new code issues found — the code change itself looks correct to me:

- `system([...])` with an array is `execvp()`'d directly by ucode and each element goes through `ucv_to_string()`, which returns strings verbatim (no JSON quoting), so `sprintf("%J", wdev_data)` arrives at `wdev.uc` as one intact argument.
- The remaining args become `keep_devices` in [`set_config` at wdev.uc:105-110](``https://github.com/openwrt/openwrt/blob/60cfeaf2be5f51594356b0a6a2687f8867ab4eb8/package/network/config/wifi-scripts/files/usr/share/hostap/wdev.uc#L105-L110),`` matching the shell implementation's [`wdev_tool "$phy$phy_suffix" set_config "$(json_dump)" $active_ifnames`](``https://github.com/openwrt/openwrt/blob/60cfeaf2be5f51594356b0a6a2687f8867ab4eb8/package/network/config/wifi-scripts/files/lib/netifd/wireless/mac80211.sh#L1251).``
- `data.phy + data.phy_suffix` is safe because [`phy_suffix()`](``https://github.com/openwrt/openwrt/blob/60cfeaf2be5f51594356b0a6a2687f8867ab4eb8/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh#L28-L32)`` always returns a string, and the same expression is already used two other places in this file.

## Commit checks

- `60cfeaf` "wifi-scripts: fix active_ifnames for OWE transition interfaces" — the whole message body, including the `Signed-off-by:` line, is indented by four extra spaces, so `git log` renders it double-indented and the S-o-b is no longer recognised as a trailer by `git interpret-trailers`. Please re-wrap the body flush left (PolynomialDivision asked for the same in the thread). The diff itself matches the message: the base ifname is still kept because it is pushed unconditionally a few lines earlier, so replacing the second push with `v.config.owe_transition_ifname` is the right fix.
- `e22c137` "wifi-scripts: fix wdev fallback to use sprintf instead of printf" — the body only describes switching to `sprintf`, but the amended diff also converts the invocation from a `/bin/sh -c` template string to the array form of `system()`. That second change is what removes the shell-quoting hazard for user-supplied SSIDs, so it deserves a sentence in the commit message rather than being implied by "update the command".

---

astrouski-google:
@PolynomialDivision Fix commit message and also address all Claude Review comments.

openwrt-ai (commented):

`git diff 60cfeaf..2589982` is empty — the only change since the last review is the commit message of the OWE commit, which is now flush left, so the `Signed-off-by:` line is picked up as a proper trailer again. The message still matches its diff (base ifname is pushed unconditionally a few lines earlier, so replacing the second push with `v.config.owe_transition_ifname` is the right fix). CI on the head SHA is green.

## Commit checks

- `e22c137` "wifi-scripts: fix wdev fallback to use sprintf instead of printf" — unchanged since the previous round, so the earlier note still stands: the body only describes the `printf` → `sprintf` switch, but the diff also converts the invocation from a `/bin/sh -c` template string to the array form of `system()`. That second change is the one that removes the shell-quoting hazard for user-supplied SSIDs; it deserves a sentence in the message (and in the PR description, which mirrors it) rather than being folded into "update the command". Not a blocker if you'd rather leave it.

---

astrouski-google:
@PolynomialDivision Should I close that CL or how that process goes? In another PR that I create it was submitted with button, so not sure should I leave it open or not?

PR #24622, opened 2026-08-09
dnsmasq: support MAC-only host entries

### LINKED ISSUE #22051: Dnsmasq: startup script doesn't add dhcp-host line to temp config when only using MAC address
### Describe the bug

When adding a static dhcp entry using only the MAC address, the dnsmasq init script does not add the dhcp-host line to the generated dnsmaq.conf file (in /var/etc).

In the start script, it's first checking for name, ip, and hostid, and exits if those are not found. However MAC address can also be used as a sole identifier for the host, according to the dnsmasq documentation. I also can confirm it works correctly when manually adding the config line to /etc/dnsmasq.conf.

I think this is the line of concern: https://github.com/openwrt/openwrt/blob/bc3b41767ddd7854067ed90341183491d88a5c06/package/network/services/dnsmasq/files/dnsmasq.init#L363

Further information can be found here: https://forum.openwrt.org/t/dnsmasq-set-tag-using-only-mac-address/246540

### OpenWrt version

r29087-d9c5716d1d

### OpenWrt release

24.10.5

### OpenWrt target/subtarget

mediatek/mt7622

### Device

Linksys E8450 (UBI)

### Image kind

Self-built image

### Steps to reproduce

Add config with uci using only MAC address:
```
uci add dhcp host
uci set "dhcp.@host[-1].mac"='11:22:33:44:55:66'
uci set "dhcp.@host[-1].tag"='mytag'
uci commit dhcp
```

Restart `dnsmasq` and check generated config:
```
service dnsmasq restart
cat /var/etc/dnsmasq.conf.cfg*
```

### Actual behaviour

See that dhcp-host line is not added to config

### Expected behaviour

dhcp-host line should be added to config.

### Additional info

Dnsmasq does not require that a hostname be provided in the dhcp-host line.

The image I'm using is generated by imagebuilder., not a fully custom build from source.

### Diffconfig

```text

```

### Terms

dhrm1k:
dnsmasq accepts dhcp-host entries without requiring a static IP
address. These entries can identify known clients or assign tags while
addresses continue to be allocated dynamically.

`dhcp_host_add()` currently discards an entry unless it contains a
name, IPv4 address, or IPv6 host ID. The check happens before the MAC
value is read, causing valid MAC-only entries to be ignored.

Read the client identifiers before the check and include them when
deciding whether the section is empty. This allows MAC-only entries to
reach dnsmasq configuration generation.

## Testing

Tested on OpenWrt 25.12.4 with the following MAC-only host entry:

```uci
config host
	option mac '11:22:33:44:55:66'
	option tag 'mytag'
```

Before the change, no `dhcp-host` section appeared in
the generated dnsmasq config.

After the change, the generated config contains:

```text
dhcp-host=11:22:33:44:55:66,set:mytag
```

Existing host entries containing a name and static IPv4 address
continue to be generated correctly.

Fixes #22051

### DIFF
--- a/package/network/services/dnsmasq/files/dnsmasq.init
+++ b/package/network/services/dnsmasq/files/dnsmasq.init
@@ -359,16 +359,17 @@ dhcp_host_add() {
 	config_get name "$cfg" name
 	config_get ip "$cfg" ip
 	config_get hostid "$cfg" hostid
+	config_get mac "$cfg" mac
+	config_get duid "$cfg" duid
 
-	[ -z "$ip" ] && [ -z "$name" ] && [ -z "$hostid" ] && return 0
+	[ -z "$ip" ] && [ -z "$name" ] && [ -z "$hostid" ] &&
+		[ -z "$mac" ] && [ -z "$duid" ] && return 0
 
 	config_get_bool dns "$cfg" dns 0
 	[ "$dns" = "1" ] && [ -n "$ip" ] && [ -n "$name" ] && {
 		echo "$ip $name${DOMAIN:+.$DOMAIN}" >> "$HOSTFILE_TMP"
 	}
 
-	config_get mac "$cfg" mac
-	config_get duid "$cfg" duid
 	config_get tag "$cfg" tag
 
 	add_tag() {
--- a/package/network/services/dnsmasq/Makefile
+++ b/package/network/services/dnsmasq/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=dnsmasq
 PKG_UPSTREAM_VERSION:=2.93
 PKG_VERSION:=$(subst test,~~test,$(subst rc,~rc,$(PKG_UPSTREAM_VERSION)))
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_UPSTREAM_VERSION).tar.xz
 PKG_SOURCE_URL:=https://thekelleys.org.uk/dnsmasq/


package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -359,16 +359,17 @@ dhcp_host_add() {
 	config_get name "$cfg" name
 	config_get ip "$cfg" ip
 	config_get hostid "$cfg" hostid
+	config_get mac "$cfg" mac
+	config_get duid "$cfg" duid
```
openwrt-ai, line 363: nit: the commit message says the old check caused "valid MAC-only **and DUID-only** entries to be ignored", but DUID-only sections aren't really fixed by this change — only MAC-only ones are.

I ran the post-patch `dhcp_host_add()` logic against a `config host` with only `option duid`:

* `DNSMASQ_DHCP_VER=4` — [`duids` is only populated under `[ $DNSMASQ_DHCP_VER -eq 6 ]`](https://github.com/openwrt/openwrt/blob/dd9c61447fb615a74737a02a2ff2a4c0edfe868c/package/network/services/dnsmasq/files/dnsmasq.init#L385-L389), so the section now passes this new guard but is still silently dropped a few lines later by `[ -n "$name" ] || return 0`. Same end result as before the patch (which is arguably right, since a DUID is meaningless for DHCPv4 — but then the commit message overstates the fix).
* `DNSMASQ_DHCP_VER=6` — the section is newly emitted, but as `dhcp-host=,id:00:03:00:01:12:00:00:01:02:03`. The leading empty field comes from [`${duids:+,$duids}` at dnsmasq.init:417](https://github.com/openwrt/openwrt/blob/dd9c61447fb615a74737a02a2ff2a4c0edfe868c/package/network/services/dnsmasq/files/dnsmasq.init#L417), which prepends a comma unconditionally even when `$macs` is empty. dnsmasq's `--dhcp-host` parser skips empty fields (an empty `arg` falls through every branch in the `case 'G'` loop in `src/option.c`), so it's harmless — and it was already reachable pre-patch via a `duid` + `name` section — but this change makes it reachable for a plain DUID-only section, which is the config the commit message advertises.

MAC-only works exactly as described (`dhcp-host=11:22:33:44:55:66,set:iot`), and existing name/ip sections are byte-identical, so nothing here blocks the fix. Either narrowing the commit message to MAC-only, or additionally guarding the comma so DUID-only produces a clean `dhcp-host=id:...`, would make the two agree.

---

openwrt-ai (commented):
Reviewed 1 new commit. The fix itself looks correct: I replayed the patched `dhcp_host_add()` logic over MAC-only, DUID-only, name+ip and empty `config host` sections for both `DNSMASQ_DHCP_VER=4` and `=6`, and confirmed MAC-only now emits `dhcp-host=11:22:33:44:55:66,set:iot` while every pre-existing case produces byte-identical output. Reading dnsmasq 2.93's `--dhcp-host` parser (`case 'G'` in `src/option.c`) also confirms a bare MAC plus `set:` tags with no address is accepted, and that `dhcp-range ... static` behaviour is unaffected since such a config carries no `CONFIG_ADDR`. `PKG_RELEASE` bump is present and correct.

One inline nit about the DUID-only claim in the commit message.

---

package/network/services/dnsmasq/files/dnsmasq.init
dhrm1k: i did overstate a little. narrowed down the commit message to address mac-only host entries.

package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -359,16 +359,17 @@ dhcp_host_add() {
...
 
-	[ -z "$ip" ] && [ -z "$name" ] && [ -z "$hostid" ] && return 0
+	[ -z "$ip" ] && [ -z "$name" ] && [ -z "$hostid" ] &&
+		[ -z "$mac" ] && [ -z "$duid" ] && return 0
```
openwrt-ai, line 366: nit: the commit message was narrowed to MAC-only in the force-push, but the PR description still carries the old wording — "causing valid MAC-only and DUID-only entries to be ignored" and "Read both identifiers before the check and only discard the section when all supported host identifiers are absent". A DUID-only section still doesn't reach the config generator on `DNSMASQ_DHCP_VER=4`: `duids` is only assigned under [the `-eq 6` guard at dnsmasq.init:386-390](https://github.com/openwrt/openwrt/blob/e4d59562de2d0a4717334babdb2eaa594deafe51/package/network/services/dnsmasq/files/dnsmasq.init#L386-L390), so it passes this new guard and is then dropped by [`[ -n "$name" ] || return 0` at dnsmasq.init:394](https://github.com/openwrt/openwrt/blob/e4d59562de2d0a4717334babdb2eaa594deafe51/package/network/services/dnsmasq/files/dnsmasq.init#L394). Worth syncing the PR body with the commit message so the two agree.

---

openwrt-ai (commented):
Reviewed 1 new commit; no new code issues found.

The force-push changed only the commit message (`git diff dd9c614..e4d5956` is empty) — the DUID overstatement flagged in my previous review is now resolved there. The code fix, `PKG_RELEASE` bump, and existing-config compatibility were verified in the earlier review and are unchanged; all three FormalityCheck runs on `e4d5956` are green.

One inline nit about the PR description still carrying the pre-force-push wording.

---

package/network/services/dnsmasq/files/dnsmasq.init
dhrm1k: PR description has been updated.

openwrt-ai (commented):

The update since my last review is a rebase only: `git show e4d5956` and `git show 73a38de` are byte-identical apart from the SHA line (same message, same diff, same context), and the new parent `f6bc2a3` carries no changes under `package/network/services/dnsmasq/` relative to the old base `831900b`, so the rebase is a no-op for this package. `PKG_RELEASE` still goes 2 -> 3 against the current base.

The PR body nit from the previous review is resolved — the description now matches the MAC-only-scoped commit message. All three FormalityCheck runs on `73a38de` are green.

---

dhrm1k:
gentle ping on this

PR #24623, opened 2026-08-09
mediatek: filogic: fix Keenetic phy1 MAC collision

lexfrei:
Changes the phy1 MAC in the Keenetic wifi fixup so the 5 GHz address can no longer collide with anything the VIF allocator derives. The current entry gives phy1 the base MAC with the LA bit set; the allocator derives the second interface on phy0 the same way, so a second 2.4 GHz VIF (mesh, STA, another AP) comes up with the 5 GHz AP address. On my KN-3811 with an 802.11s mesh this filled the log with `br-lan: received packet on bat0 with own address as source` every 10 seconds and kept the FDB entry for that address flapping.

Easy to check without any traffic:

```
# ucode /usr/share/hostap/wdev.uc phy0 get_macaddr id=1
d6:9c:53:xx:xx:xx   <- same address the fixup writes to phy1
```

The new entry reproduces the address mt76 derives on its own when the second calibration slot is empty: LA bit set, bit 7 of the first octet flipped. Neither allocator scheme can produce it, and it matches neither the base nor the base+1 ethernet address. The 5 GHz BSSID changes accordingly.

Dropping the override entirely does not work here: without it mt76 falls back to the second MAC slot in the calibration data (`MT_EE_MAC_ADDR2`, eeprom `0xa`), and on all these boards the DTS already uses that slot as the ethernet/label address. Verified on my KN-3811, the slot holds the WAN MAC, so phy1 would just copy a wired port; on KAP-630/NAP-630 that port even sits in the same bridge as the 5 GHz AP. Shifting the base is not enough either: the allocator's MBSSID scheme varies the last octet (`addr[5] ^= idx`), so any last-octet shift stays reachable at some index.

Tested on the KN-3811 with this exact change: phy1 comes up as the base with the LA bit and bit 7 flipped, no address overlaps on the device, and APs, mesh and clients re-associate fine.

KN-1012/NC-1012 have the same eeprom layout and the same collision, but they are not in the tree yet, so their names are added by their own board PR (#24820) instead of here. That PR adds the two names to the same case block this one rewrites, so whichever lands second needs a rebase.


lexfrei:
@Linaro1985 this override came in with the KN-3911 and the later boards inherited it, but I can't find the reason anywhere. The commit message and the discussion in #16830 don't mention it. Was there a driver problem back then, or does stock use base|LA for the 5 GHz side? If your KN-3911 is still alive: `grep rf-eeprom /proc/mtd`, then `dd if=/dev/mtdN bs=1 skip=10 count=6 | hexdump -C` shows whether the second MAC slot is factory-provisioned. That's the one board of the family I couldn't check.

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
```diff
@@ -176,14 +176,6 @@ case "$board" in
...
 		[ "$PHYNBR" = "1" ] && macaddr_add $addr 3 > /sys${DEVPATH}/macaddress
 		;;
-	keenetic,kap-630|\
-	keenetic,kn-3711|\
-	keenetic,kn-3811|\
-	keenetic,kn-3911|\
-	netcraze,nap-630)
-		[ "$PHYNBR" = "1" ] && \
-			macaddr_setbit_la "$(mtd_get_mac_binary rf-eeprom 0x4)" > /sys${DEVPATH}/macaddress
-		;;
```
openwrt-ai, line 186: Removing the override hands phy1 whatever sits in `MT_EE_MAC_ADDR2` (eeprom offset `0xa`), and on all five of these boards that offset is already consumed as an ethernet MAC — so the 5 GHz phy address stops colliding with the 2.4 GHz VIF range and starts colliding with a wired interface instead:

- KN-3811 / KN-3711 / KN-3911: `macaddr_factory_a` (= eeprom `0xa`) is the WAN gmac address — [kn-3811.dts gmac1](https://github.com/openwrt/openwrt/blob/e5111f502067914a4baa39a030d0fe17223ce223/target/linux/mediatek/dts/mt7981b-keenetic-kn-3811.dts#L135-L143), cell defined at [`macaddr@a`](https://github.com/openwrt/openwrt/blob/e5111f502067914a4baa39a030d0fe17223ce223/target/linux/mediatek/dts/mt7981b-keenetic-kn-3811.dts#L276-L279).
- KAP-630 / NAP-630: the same cell is the board's only ethernet MAC — [kap-630.dtsi gmac0](https://github.com/openwrt/openwrt/blob/e5111f502067914a4baa39a030d0fe17223ce223/target/linux/mediatek/dts/mt7981b-keenetic-kap-630.dtsi#L80-L88).
- All five also point `label-mac-device` at that same gmac, so `0xa` is the printed label MAC as well.

mt76 only computes the safe `base | 0x02 ^ BIT(7)` address when that slot is *invalid* ([mt7915/init.c:694-704](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/mediatek/mt76/mt7915/init.c#L694-L704), v6.18.41 — matches the `KERNEL_PATCHVER:=6.18` mt76 the target builds). Since the DTS guarantees `0xa` is populated on these boards, the derive path never runs and phy1 gets the wired/label MAC verbatim. On KAP-630/NAP-630 that eth port and the 5 GHz AP end up in the same `br-lan` ([02_network:137-147](``https://github.com/openwrt/openwrt/blob/e5111f502067914a4baa39a030d0fe17223ce223/target/linux/mediatek/filogic/base-files/etc/board.d/02_network#L137-L147)),`` i.e. two ports of one bridge with an identical address.

The bug you're fixing is real, but the in-tree remedy for exactly this collision on sibling filogic boards keeps an override and only changes its base — see the [`jcg,q30-pro` / `netcore,n60` case at lines 164-170](``https://github.com/openwrt/openwrt/blob/e5111f502067914a4baa39a030d0fe17223ce223/target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac#L164-L170),`` whose comment describes the same "conflicts with the other radio's VIFs" problem. Would reworking this entry to derive phy1 from an address that clashes with neither the VIF allocation range nor the ethernet/label MAC (mirroring mt76's own bit-7-flipped scheme, or the label-mac approach used above) be preferable to dropping it?

Related: because `0xa` is the label/ethernet MAC on these boards, a unit with an empty `MT_EE_MAC_ADDR2` would have no factory ethernet MAC either — was the "unit with an empty slot" you tested one of these five boards, or a different mt7981 device?

---

openwrt-ai (commented):
Reviewed 1 new commit; see the inline comment. The commit message accurately describes the diff.

---

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
lexfrei: You're right, and it reproduces on hardware: on my KN-3811 the `0xa` slot holds the WAN MAC (base+1), so with the override dropped phy1 comes up as a copy of the wired port. Works on this board only because wan is not bridged with the APs; on KAP-630/NAP-630 it would be two ports of one bridge again. Reworked the change to keep the override with a shifted base, `(base+2) | LA`, same idea as the q30-pro entry: last octet takes it out of the VIF range, and it matches neither the base nor the base+1 ethernet address.

The empty-slot unit I mentioned is an NC-4410 (same mt7981 family, not one of these five): its `0xa` is all zeroes and the DTS does not reference it, so the mt76 derive path is what that board actually runs.

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
```diff
@@ -181,8 +181,13 @@ case "$board" in
...
 	keenetic,kn-3911|\
 	netcraze,nap-630)
+		# Originally phy1 was the phy0 mac with only the LA bit set, which is
+		# also the address the VIF allocator gives the second 2.4 GHz
+		# interface. The eeprom second-MAC slot (0xa) is no alternative:
+		# the DTS already consumes it as the ethernet/label mac. Shift the
+		# base so the result collides with neither.
 		[ "$PHYNBR" = "1" ] && \
-			macaddr_setbit_la "$(mtd_get_mac_binary rf-eeprom 0x4)" > /sys${DEVPATH}/macaddress
+			macaddr_setbit_la "$(macaddr_add "$(mtd_get_mac_binary rf-eeprom 0x4)" 2)" > /sys${DEVPATH}/macaddress
```
openwrt-ai: `(base+2) | LA` escapes the default VIF range but not the MBSSID one. The allocator has two schemes ([`macaddr_generate` switch, common.uc:256-267](``https://github.com/openwrt/openwrt/blob/4b1c3753bfaaaa21f2e38bb6df35a04b9de4cdb6/package/network/config/wifi-scripts/files/usr/share/hostap/common.uc#L256-L267)):``

- `b1` (what you measured, since mt7915 sets neither `wiphy->addresses` nor an `addr_mask`, so `address_mask` reads back all-zero and is promoted to `ff:ff:ff:ff:ff:ff`): `addr[0] |= 2; addr[0] ^= idx << 2` — only octet 0 varies, octet 5 stays at `base[5]`. `base+2` is indeed outside this set. Good.
- `b5`, taken whenever `mbssid` is non-zero: `addr[0] |= 2; addr[5] ^= idx`. `multiple_bssid` is a plain UCI option and is passed straight through to the allocator ([mac80211.sh:583](``https://github.com/openwrt/openwrt/blob/4b1c3753bfaaaa21f2e38bb6df35a04b9de4cdb6/package/network/config/wifi-scripts/files/lib/netifd/wireless/mac80211.sh#L583)).`` For `idx = 2` this yields `base[0]|2 … base[5]^2`, and `base[5]^2 == base[5]+2` for every unit whose `base[5]` has bit 1 clear — i.e. the third MBSSID BSS on the 2.4 GHz radio lands on exactly the address this line writes to phy1, on roughly half the units. Any "shift the last octet" variant (including the `get_mac_label` = `base+1` form the `jcg,q30-pro` entry uses) has the same weakness at some `idx`.

mt76's own fallback avoids both schemes rather than one: `macaddr[0] |= 2; macaddr[0] ^= BIT(7)` ([mt7915/init.c:693-702](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/mediatek/mt76/mt7915/init.c#L693-L702), v6.18.41 — the series this target builds). Bit 7 of octet 0 is never touched by `b1` (`idx << 2` reaches bits 2-6 for `idx < 32`) nor by `b5`, and octets 1-5 stay equal to the base, so it collides with neither VIF range nor with the `0xa` ethernet/label MAC. Since the only reason this override exists is that mt76 cannot use that path here (`0xa` is valid on these boards), would reproducing mt76's scheme in the script be preferable to shifting the base? Note there is no XOR helper in `system.sh` — `macaddr_setbit`/`macaddr_unsetbit` only set or clear — so it would need the octet-0 bit flipped explicitly rather than a one-liner.

Either way the change is a clear improvement over the current `base | LA`; this is about whether the remaining MBSSID overlap is worth closing while the entry is being touched.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push replacing `e5111f5`); the reworked approach addresses the ethernet/label-MAC collision raised last round, and the commit message matches the diff. One follow-up inline about a residual overlap case.

---

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
lexfrei: Took the suggestion, the entry now reproduces mt76's scheme via `printf`, since `system.sh` has no XOR helper. Tested on the KN-3811: phy1 comes up as `56:xx:...` matching what the driver would derive itself, nothing overlaps, mesh and clients re-associated after the BSSID change.

openwrt-ai (commented):
Reviewed 1 new commit (force-push replacing `4b1c375`); no new issues found.

The reworked entry now matches mt76's own fallback (`macaddr[0] |= 2; macaddr[0] ^= BIT(7)`, [mt7915/init.c:697-702](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/mediatek/mt76/mt7915/init.c#L697-L702)), which closes the MBSSID overlap raised last round: octet 0 is the only octet that differs from the phy0 base, and neither allocator scheme reaches bit 7 there (`b1` varies bits 2-6 via `idx << 2`, `b5` varies octet 5 — [common.uc:256-276](``https://github.com/openwrt/openwrt/blob/fa0d1792e853bb268c7adfb5dec07baa6bc0b257/package/network/config/wifi-scripts/files/usr/share/hostap/common.uc#L256-L276)),`` so it also can't collide with the `0xa` ethernet/label MAC (which differs only in octet 5). The inline `printf` form mirrors the existing [`macaddr_unsetbit_mc`](https://github.com/openwrt/openwrt/blob/fa0d1792e853bb268c7adfb5dec07baa6bc0b257/package/base-files/files/lib/functions/system.sh#L270-L274) helper, and the missing trailing newline is fine — `macaddress_store` accepts a 17-byte write since kernfs NUL-terminates the buffer. Commit message matches the diff.

---

lexfrei:
Talked to @Linaro1985 out of band, he pulled rf-eeprom dumps from his units. The picture across the family:

- KN-3911, NAP-630 and his KN-3811: `0xa` holds base+1, same as my KN-3811. The slot is the ethernet address on all of them, so the mt76 fallback path is not available there and the override has to stay.
- KN-3411 (not in this list): `0xa` is empty, matching what its NC twin showed earlier. Those boards run the driver-derived address and need no entry here.

That covers four of the five boards with actual eeprom data; KAP-630 is the NAP-630 twin.

openwrt-ai (commented):
Reviewed 1 new commit (force-push/rebase replacing `fa0d1792`); no new issues found.

The only change since the last review is the two added board entries `keenetic,kn-1012` and `netcraze,nc-1012`; the phy1 address derivation (`printf "%02x:%s" $(( (0x${addr%%:*} | 0x2) ^ 0x80 )) "${addr#*:}"`) is unchanged from the previously validated version. Both new compatibles have no DTS/image board definition in the tree yet, but the commit message states this and its rationale explicitly, so the dead entries are intentional and harmless (they simply won't match until the boards land, at which point they inherit the family convention). Commit message matches the diff.

---

openwrt-ai (commented):
Reviewed 1 new commit (rebase of `7a648451` onto current `main`); no new issues found.

The commit is a rebase only — the single commit message and its one modified file (`package/base-files/files/etc/hotplug.d/ieee80211/11_fix_wifi_mac`) are byte-identical to the previously reviewed version. The larger range diff is entirely unrelated upstream commits that landed on `main` and were pulled in by the rebase, not changes authored by this PR. The phy1 address derivation and the `keenetic,kn-1012` / `netcraze,nc-1012` entries are unchanged from the last validated version. Commit message matches the diff.

---

openwrt-ai (commented):
Reviewed 1 new commit (rebase of `6f5e6dab` onto current `main`); no new issues found.

The commit is a rebase only: `git show` of `ecc204a2` and `6f5e6dab` produce byte-identical patches and byte-identical commit messages — only the parent moved (`f0204d78` → `1ad2e78e`). The wide range diff between the two bases is unrelated upstream work (siflower kernel 6.12→6.18 config move, uml patches, etc.), not changes authored by this PR.

Re-checked that the rebase did not invalidate the reasoning validated in earlier rounds:

- `target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac` was not touched on `main` between the two bases, so no board entry was added or removed under the `keenetic,*` / `netcraze,*` case arm and the hunk still lands in its original context.
- The userspace VIF allocator in [`common.uc`](https://github.com/openwrt/openwrt/blob/ecc204a22b81a423dcaaa0f3ec44532c4653c5c3/package/network/config/wifi-scripts/files/usr/share/hostap/common.uc) is unchanged across the two bases (the one wifi-scripts commit in that range, `a3357c90be7`, is a 6 GHz 40 MHz channel-config fix and does not touch address derivation), so the "neither allocator scheme reaches bit 7 of octet 0" argument still holds.
- `macaddr_setbit_la` and the surrounding helpers in [`system.sh`](https://github.com/openwrt/openwrt/blob/ecc204a22b81a423dcaaa0f3ec44532c4653c5c3/package/base-files/files/lib/functions/system.sh) are likewise untouched in that range.

The phy1 derivation (`printf "%02x:%s" $(( (0x${addr%%:*} | 0x2) ^ 0x80 )) "${addr#*:}"`) and the `keenetic,kn-1012` / `netcraze,nc-1012` entries are unchanged from the last validated version. Commit message matches the diff, and the `Signed-off-by:` trailer is present.

CI on `ecc204a2` is green across all 22 check runs (FormalityCheck, kernel patch check, and the four mediatek subtarget builds including `mediatek/filogic`).

---

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
```diff
@@ -177,12 +177,23 @@ case "$board" in
 		[ "$PHYNBR" = "1" ] && macaddr_add $addr 3 > /sys${DEVPATH}/macaddress
 		;;
 	keenetic,kap-630|\
+	keenetic,kn-1012|\
```
hauke: `keenetic,kn-1012` and `netcraze,nc-1012` are not supported by OpenWrt, please remove them here again.

lexfrei: @hauke Removed. They moved to the KN-1012 PR (#24820) and will land together with the boards.

openwrt-ai (commented):
## Commit checks
- `284fee82` "mediatek: filogic: fix Keenetic phy1 MAC collision" — the paragraph "KN-1012 and NC-1012 join the same list. Their support is not in the tree yet, ..." no longer matches the diff; both entries were dropped from the case arm. Drop that paragraph (the PR body's "The list now also carries KN-1012/NC-1012" is stale for the same reason).

---

openwrt-ai (commented):

---

openwrt-ai (commented):

---

openwrt-ai (commented):

---

lexfrei:
Rebased on current main.

This is one line of shell and it has been running on my own devices for about a month. It fixes something I hit daily rather than a theoretical collision.

It conflicts textually with #24820, which adds two board names to the same case block, so whichever of the two lands second needs a rebase.

openwrt-ai (commented):

---

lexfrei:
@jonasjelonek this is a single line in `11_fix_wifi_mac`, running on my own devices for about a month. It fixes an address collision I hit daily, not a theoretical one. Rebased on current main.

#24820 adds two board names to the same case block, so the two conflict. I know about it and will rebase whichever of them lands second.

openwrt-ai (commented):

---

jonasjelonek:

PR #24624, opened 2026-08-09
airoha: fix EN8811H PHY binding on Nokia XG-040G-MF and MD

VitaliySochniy:
The 2.5G port (`lan1`) on Nokia XG-040G-MF never came up. The wrong driver bound to the PHY, and every attempt to bring the link up ended in a `-EINVAL` and a WARNING backtrace:

```
airoha_eth 1fb50000.ethernet lan1: PHY [...:0f] driver [Generic Clause 45 PHY]
airoha_eth 1fb50000.ethernet lan1: configuring for phy/2500base-x link mode
_phy_start_aneg+0x0/0xa0: returned: -22
WARNING: CPU: 0 PID: 592 at _phy_state_machine+0x100/0x3e0
```

### Cause

The PHY node is described as `ethernet-phy-ieee802.3-c45`. That makes `fwnode_mdiobus_register_phy()` take the `get_phy_device()` path, which reads the PHY ID off the bus while scanning it. The EN8811H does not answer that early, the ID reads back as zero, so `air_en8811h` never matches and the generic Clause 45 driver binds instead.

The generic driver builds `phydev->supported` without the `Autoneg` bit, so `phy_probe()` clears `phydev->autoneg`. `_phy_start_aneg()` then lands in `genphy_c45_config_aneg()`, which takes the forced path and bails out of `genphy_c45_pma_setup_forced()` with `-EINVAL`. The state machine turns that into a `WARN`, on every attempt to bring the link up.

### Fix

Describe the PHY by ID instead. `phy_device_create()` then takes the ID from the device tree, no bus read happens during the scan, and the right driver binds:

```
- compatible = "ethernet-phy-ieee802.3-c45";
+ compatible = "ethernet-phy-id03a2.a411";
```

Dropping the c45 compatible alone does not help: phylib would still read the ID from the hardware and get zero. No `air_en8811h` change is needed either, `en8811h_probe()` sets up the AN MMD itself once the PHY firmware is loaded.

### Test

Nokia XG-040G-MF, OpenWrt SNAPSHOT, kernel 6.18.41:

```
airoha_eth 1fb50000.ethernet lan1: PHY [...:0f] driver [Airoha EN8811H]
Airoha EN8811H ...:0f: MD32 firmware version: 25062302
airoha_eth 1fb50000.ethernet lan1: Link is Up - 2.5Gbps/Full - flow control rx/tx
```

`phy_id` reads back `0x03a2a411`, `ethtool` reports `Supports auto-negotiation: Yes`, and the WARNING is gone. A 45 minute iperf3 run over the port moved 561 GByte at 1.78 Gbit/s without a single link event.

### The MD commit

Nokia XG-040G-MD carries the same EN8811H behind the same binding, and there LAN1 does come up: the PHY sits behind the switch MDIO bus and answers the MMD device ID reads in time, so `air_en8811h` is still matched through `phydev->c45_ids` even though the Clause 22 ID reads back as zero. Nothing guarantees that ordering — the MF is what happens when it does not hold. Describing the PHY by ID there as well removes the dependency on timing.

Verified on hardware: the ID now comes from the device tree, the driver binds and the port links at 2.5 Gbit/s. Since no breakage is observed on the MD, that commit carries no `Fixes:` tag.

### DIFF
--- a/target/linux/airoha/dts/an7583-nokia_xg-040g-mf-common.dtsi
+++ b/target/linux/airoha/dts/an7583-nokia_xg-040g-mf-common.dtsi
@@ -166,7 +166,7 @@
 &mdio_0 {
 	/* Airoha EN8811 2.5Gbps phy */
 	en8811: ethernet-phy@f {
-		compatible = "ethernet-phy-ieee802.3-c45";
+		compatible = "ethernet-phy-id03a2.a411";
 		reg = <0xf>;
 
 		reset-gpios = <&an7583_pinctrl 34 GPIO_ACTIVE_LOW>;
--- a/target/linux/airoha/dts/an7581-nokia_xg-040g-md-common.dtsi
+++ b/target/linux/airoha/dts/an7581-nokia_xg-040g-md-common.dtsi
@@ -134,7 +134,7 @@
 &mdio {
 	/* Airoha EN8811 2.5Gbps phy */
 	en8811: ethernet-phy@f {
-		compatible = "ethernet-phy-ieee802.3-c45";
+		compatible = "ethernet-phy-id03a2.a411";
 		reg = <0xf>;
 
 		reset-gpios = <&en7581_pinctrl 31 GPIO_ACTIVE_LOW>;


target/linux/airoha/dts/an7583-nokia_xg-040g-mf-common.dtsi
```diff
@@ -178,7 +178,7 @@
 &mdio_0 {
 	/* Airoha EN8811 2.5Gbps phy */
 	en8811: ethernet-phy@f {
-		compatible = "ethernet-phy-ieee802.3-c45";
+		compatible = "ethernet-phy-id03a2.a411";
```
openwrt-ai, line 181: nit: sister-device parity — the AN7583 EVB has the same EN8811H at `reg = <0xf>` on the same `&mdio_0` bus, but its node carries no `compatible` at all: [an7583-evb-emmc.dts:204-205](https://github.com/openwrt/openwrt/blob/04478e9d01da40c193557ca385436968645f1a86/target/linux/airoha/dts/an7583-evb-emmc.dts#L204-L205). With no `compatible`, `fwnode_mdiobus_register_phy()` still takes the `get_phy_device()` path and reads the ID off the bus during the scan — the same path this PR is working around — so the EVB looks exposed to the same "ID reads back as zero, generic driver binds" race, and it does ship the driver via [`DEVICE_PACKAGES := kmod-phy-airoha-en8811h`](https://github.com/openwrt/openwrt/blob/04478e9d01da40c193557ca385436968645f1a86/target/linux/airoha/image/an7583.mk#L43). Was leaving the EVB out intentional (e.g. it was observed to bind correctly), or should it get `compatible = "ethernet-phy-id03a2.a411";` in the same series?

---

openwrt-ai (commented):
Reviewed 2 new commits. Both commit messages match their diffs, the `Fixes: 42b8e0d9b781` tag resolves to the commit that introduced the c45 compatible on the MF, and `ethernet-phy-id03a2.a411` matches `EN8811H_PHY_ID` (0x03a2a411) and the upstream `airoha,en8811h.yaml` binding, which allows only that compatible. One inline question below.

---

target/linux/airoha/dts/an7583-nokia_xg-040g-mf-common.dtsi
VitaliySochniy: The EVB was left out because there is no EVB here; the series is limited to the two boards that could be verified on hardware. It was not a case of the EVB having been observed to bind correctly.

The analysis above holds. With no compatible at all, fwnode_mdiobus_register_phy() still takes the get_phy_device() path and reads the ID off the bus during the scan, so the EVB sits on the same path this series works around. One detail worth adding: apart from the reset GPIO number, the EVB node is identical to the MF one — same reset-assert-us = <10000>, same reset-deassert-us = <20000> — so there is no timing difference that would let the EVB reliably win the race the MF loses. It has simply not been reported.

Scope note: only an7583-evb-emmc.dts carries the PHY. an7583-evb.dts has no EN8811H node at all.

Adding compatible = "ethernet-phy-id03a2.a411"; to the EVB in this series is fine by me if maintainers are comfortable taking a change that has not been tested on the hardware in question. It is mechanical, the ID matches EN8811H_PHY_ID, and airoha,en8811h.yaml permits only that compatible. If confirmation on a real EVB is preferred, a separate patch would keep this series verified end to end.

@danielschwierzeck has been touching the AN7583 EVB device trees most recently and may have one on hand. @Ansuel @robimarko — whichever way you prefer the EVB handled, here or separately.

danielschwierzeck: I don't have a AN7583 EVB, but a custom AN7583 board with a EN8811 PHY. For me it probes with compatible string `compatible = "ethernet-phy-ieee802.3-c22";`.

VitaliySochniy: That data point fits the analysis: the failure is specific to the Clause 45 branch, not to reading the ID as such.

`fwnode_mdiobus_register_phy()` picks one of three paths. With `ethernet-phy-ieee802.3-c45` the scan calls `get_phy_device(bus, addr, true)`, which reads the devices-in-package MMD registers — those are what come back as zero on the MF. With `ethernet-phy-ieee802.3-c22`, `fwnode_get_phy_id()` fails, so the scan calls `get_phy_device(bus, addr, false)` and reads the Clause 22 ID registers instead, which the PHY evidently answers in time. With the ID compatible neither happens: `phy_device_create()` takes the ID from the device tree and the bus is not touched during the scan.

So c22 avoids the MMD read, but keeps the requirement that the PHY answer some ID read while the bus is being scanned. Describing it by ID drops that requirement entirely.

There is also a binding argument for the ID form — `Documentation/devicetree/bindings/net/airoha,en8811h.yaml` has:

```yaml
  compatible:
    enum:
      - ethernet-phy-id03a2.a411
```

so `ethernet-phy-ieee802.3-c22` on an EN8811H node would not validate against it.

Trying c22 on the XG-040G-MF is easy enough if it is useful for the record — it would confirm that the race is limited to the MMD path — but it would not change what this series does.

csharper2005 (approved):
LGTM. Run tested.

target/linux/airoha/dts/an7583-nokia_xg-040g-mf-common.dtsi
robimarko: PHY bindings always list the PHY ID, but it is fully valid to not hardcode a PHY ID in the compatible in DT cause they are probable devices.

So, if the PHY probes fine, then avoid hardcoding anything.

VitaliySochniy: Tested this on the XG-040G-MF, since the c22 data point suggested the ID hardcoding might be unnecessary here too. It is not: with `compatible = "ethernet-phy-ieee802.3-c22"` the board fails exactly as it does with the c45 compatible, on six consecutive boots:

```
airoha_eth 1fb50000.ethernet lan1: PHY [...:0f] driver [Generic Clause 45 PHY] (irq=POLL)
```

`phy_id` reads back `0x00000000`, two WARNING backtraces per boot, `lan1` stays `DOWN`. Not one boot in six bound `air_en8811h`, so on this board it is not a race that is sometimes lost — the PHY never answers the ID read during the bus scan.

The reason the generic **Clause 45** driver ends up bound even without a c45 compatible is in `get_phy_device()`:

```
if (!is_c45 && phy_id == 0 && bus->read_c45) {
    r = get_phy_c45_ids(bus, addr, &c45_ids);
    if (!r)
        return phy_device_create(bus, addr, phy_id, true, &c45_ids);
}
```

The Clause 22 read returns zero without an error, and the Airoha MDIO bus provides `read_c45`, so phylib retries in C45 space and creates a C45 device. Dropping the c45 compatible therefore lands in the same place by design.

So the ID form is not being used here to override a probeable device, but because this PHY is not probeable at bus-scan time on this board.

That reasoning does not carry over to the MD, though: that board binds `air_en8811h` through `phydev->c45_ids` on its own, so there is no failure to fix there and the change is hardening rather than a fix. @robimarko, should the MD commit be dropped from this series and the PR limited to the board that is actually broken?

danielschwierzeck: @VitaliySochniy could you check if the correct MDIO bus is used and if the according pin mux is activated (mdio_0 or mdio_1)? Also check if the PHY reset line is properly mapped (GPIO, high/low active) and working. I did not check the PHY driver code but I assume it behaves similarily to as21xxx where different PHY IDs are read back dependent on whether the PHY is in reset state or the firmware has booted. I had a similar issue with PHY reset and unstable probing on my AN7581 board with as21xxx due to the many pinctrl bugs which caused non-working PHY reset. But all known pinctrl bugs should be fixed by now in OpenWrt mainline.

VitaliySochniy: @danielschwierzeck, checked both on the board.

**MDIO bus**. The PHY node sits under `&mdio_0`, which is `mdio-bus@c8` in `an7583.dtsi` (`mdio_1` is `@cc`). The runtime path agrees: `1fb00000.system-controller:mdio-bus@c8-mii:0f`.

**Pin mux**. From the pinctrl debugfs on the running board:

```
pin 53 (mdc_0):  device 1fb58000.switch function mdio group mdio
pin 54 (mdio_0): device 1fb58000.switch function mdio group mdio
```

**Reset line.** Mapped and released:

```
gpio-34  (                    |PHY reset           ) out hi ACTIVE LOW
```

The stronger argument is that all three are identical in the failing and the working case. The DTS, the bus, the pins and the reset GPIO are the same; the only difference is the `compatible` string. With `ethernet-phy-id03a2.a411` the PHY answers, `air_en8811h` binds, the MD32 firmware loads and the port links at 2.5 Gbit/s. A wrong bus, a wrong mux or a broken reset could not produce that.

The intuition about the ID depending on reset state does look right, but it cannot be addressed from the device tree, because phylib reads the ID before it ever touches `reset-gpios`:

```
fwnode_mdiobus_register_phy()
  ├─ get_phy_device(bus, addr, is_c45)          <- reads the ID off the bus
  └─ fwnode_mdiobus_phy_device_register()
       └─ phy_device_register() → mdiobus_register_device()
            ├─ mdiobus_register_gpiod()          <- reset-gpios acquired here
            └─ mdio_device_reset(mdiodev, 1)
```

`reset-assert-us` and `reset-deassert-us` are parsed into `phy->mdio.reset_assert_delay` in the same function, but they only take effect at `mdiobus_register_device()`, which runs after `get_phy_device()` has already returned. So during the bus scan the PHY is in whatever state the bootloader left it, and no correct reset mapping in DT can change that ordering — it would take a phylib change to reset before scanning.

That is also why the pinctrl bugs behind the as21xxx trouble would not show up the same way here: reset works at runtime on this board, it is simply not applied yet at the point where the ID is read.

csharper2005: > So, if the PHY probes fine, then avoid hardcoding anything.

It worked before, but at some point it stopped. I think it was after the an7583 eth pcs patches.

itdoginfo:
Tested on Nokia XG-040G-MF. LAN1 now works correctly and negotiates the expected speed with both 1G and 2.5G NICs.

**1 Gbit NIC:**
```
~> sudo ethtool enp6s0
Settings for enp6s0:
	Supported ports: [ TP	 MII ]
	Supported link modes:   10baseT/Half 10baseT/Full
	                        100baseT/Half 100baseT/Full
	                        1000baseT/Full
	Supported pause frame use: Symmetric Receive-only
	Supports auto-negotiation: Yes
	Supported FEC modes: Not reported
	Advertised link modes:  10baseT/Half 10baseT/Full
	                        100baseT/Half 100baseT/Full
	                        1000baseT/Full
	Advertised pause frame use: Symmetric Receive-only
	Advertised auto-negotiation: Yes
	Advertised FEC modes: Not reported
	Link partner advertised link modes:  100baseT/Full
	                                     1000baseT/Full
	Link partner advertised pause frame use: Symmetric Receive-only
	Link partner advertised auto-negotiation: Yes
	Link partner advertised FEC modes: Not reported
	Speed: 1000Mb/s
	Duplex: Full
	Auto-negotiation: on
	master-slave cfg: preferred slave
	master-slave status: slave
	Port: Twisted Pair
	PHYAD: 0
	Transceiver: external
	MDI-X: Unknown
	Supports Wake-on: pumbg
	Wake-on: d
	Link detected: yes
```

**2.5 Gbit NIC:**

```
~> sudo ethtool enp4s0
Settings for enp4s0:
	Supported ports: [ TP	 MII ]
	Supported link modes:   10baseT/Half 10baseT/Full
	                        100baseT/Half 100baseT/Full
	                        1000baseT/Full
	                        2500baseT/Full
	Supported pause frame use: Symmetric Receive-only
	Supports auto-negotiation: Yes
	Supported FEC modes: Not reported
	Advertised link modes:  10baseT/Half 10baseT/Full
	                        100baseT/Half 100baseT/Full
	                        1000baseT/Full
	                        2500baseT/Full
	Advertised pause frame use: Symmetric Receive-only
	Advertised auto-negotiation: Yes
	Advertised FEC modes: Not reported
	Link partner advertised link modes:  100baseT/Full
	                                     1000baseT/Full
	                                     2500baseT/Full
	Link partner advertised pause frame use: Symmetric Receive-only
	Link partner advertised auto-negotiation: Yes
	Link partner advertised FEC modes: Not reported
	Speed: 2500Mb/s
	Duplex: Full
	Auto-negotiation: on
	master-slave cfg: preferred slave
	master-slave status: slave
	Port: Twisted Pair
	PHYAD: 0
	Transceiver: internal
	MDI-X: Unknown
	Supports Wake-on: pumbg
	Wake-on: d
	Link detected: yes
```

geokvant:
My MD Works correctly on 1 Gbpsimage
image
```
[    8.518001] airoha_eth 1fb50000.ethernet lan1: renamed from eth1
[    8.924957] airoha_eth 1fb50000.ethernet lan1: PHY [mt7530-0:0f] driver [Airoha EN8811H] (irq=POLL)
[    8.934294] airoha_eth 1fb50000.ethernet lan1: configuring for phy/2500base-x link mode
[   21.352823] airoha_eth 1fb50000.ethernet lan1: PHY [mt7530-0:0f] driver [Airoha EN8811H] (irq=POLL)
[   21.362214] airoha_eth 1fb50000.ethernet lan1: configuring for phy/2500base-x link mode
[   21.836312] br-lan: port 1(lan1) entered blocking state
[   21.841568] br-lan: port 1(lan1) entered disabled state
[   21.846913] airoha_eth 1fb50000.ethernet lan1: entered allmulticast mode
[   21.853862] airoha_eth 1fb50000.ethernet lan1: entered promiscuous mode
[  151.418084] airoha_eth 1fb50000.ethernet lan1: left allmulticast mode
[  151.424554] airoha_eth 1fb50000.ethernet lan1: left promiscuous mode
[  151.431061] br-lan: port 1(lan1) entered disabled state
[  151.544962] airoha_eth 1fb50000.ethernet lan1: PHY [mt7530-0:0f] driver [Airoha EN8811H] (irq=POLL)
[  151.554357] airoha_eth 1fb50000.ethernet lan1: configuring for phy/2500base-x link mode
[  162.397054] airoha_eth 1fb50000.ethernet lan1: Link is Up - 1Gbps/Full - flow control rx/tx
```
=-=-=-=-=-=-=-=-

2,5Gbps also works

telegram-cloud-photo-size-2-5199856612976301834-y


```
[    8.508135] airoha_eth 1fb50000.ethernet lan1: renamed from eth1
[    8.924753] airoha_eth 1fb50000.ethernet lan1: PHY [mt7530-0:0f] driver [Airoha EN8811H] (irq=POLL)
[    8.934093] airoha_eth 1fb50000.ethernet lan1: configuring for phy/2500base-x link mode
[   18.078459] airoha_eth 1fb50000.ethernet lan1: PHY [mt7530-0:0f] driver [Airoha EN8811H] (irq=POLL)
[   18.088289] airoha_eth 1fb50000.ethernet lan1: configuring for phy/2500base-x link mode
[   18.528253] br-lan: port 1(lan1) entered blocking state
[   18.533508] br-lan: port 1(lan1) entered disabled state
[   18.538818] airoha_eth 1fb50000.ethernet lan1: entered allmulticast mode
[   18.545845] airoha_eth 1fb50000.ethernet lan1: entered promiscuous mode
[   26.877056] airoha_eth 1fb50000.ethernet lan1: Link is Up - 2.5Gbps/Full - flow control rx/tx
[   26.885676] br-lan: port 1(lan1) entered blocking state
[   26.890949] br-lan: port 1(lan1) entered forwarding state
```


```
root@Nokia_XG-040G-MD:~# iperf3 -c 192.168.40.200
Connecting to host 192.168.40.200, port 5201
[  5] local 192.168.40.1 port 55410 connected to 192.168.40.200 port 5201
[ ID] Interval           Transfer     Bitrate         Retr  Cwnd
[  5]   0.00-1.00   sec   284 MBytes  2.37 Gbits/sec   36   1.14 MBytes       
[  5]   1.00-2.00   sec   280 MBytes  2.35 Gbits/sec    0   1.30 MBytes       
[  5]   2.00-3.00   sec   281 MBytes  2.36 Gbits/sec    3   1.09 MBytes       
[  5]   3.00-4.00   sec   280 MBytes  2.35 Gbits/sec    0   1.26 MBytes       
[  5]   4.00-5.00   sec   281 MBytes  2.35 Gbits/sec    3   1.05 MBytes       
[  5]   5.00-6.00   sec   281 MBytes  2.35 Gbits/sec    0   1.22 MBytes       
[  5]   6.00-7.00   sec   280 MBytes  2.35 Gbits/sec    2   1.01 MBytes       
[  5]   7.00-8.00   sec   279 MBytes  2.34 Gbits/sec    0   1.20 MBytes       
[  5]   8.00-9.00   sec   281 MBytes  2.36 Gbits/sec    1    984 KBytes       
[  5]   9.00-10.00  sec   281 MBytes  2.36 Gbits/sec    0   1.16 MBytes       
- - - - - - - - - - - - - - - - - - - - - - - - -
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec  2.74 GBytes  2.36 Gbits/sec   45            sender
[  5]   0.00-10.01  sec  2.74 GBytes  2.35 Gbits/sec                  receiver

iperf Done.
```

longnt2007:
> ```
> root@Nokia_XG-040G-MD:~# iperf3 -c 192.168.40.200
> Connecting to host 192.168.40.200, port 5201
> [  5] local 192.168.40.1 port 55410 connected to 192.168.40.200 port 5201
> [ ID] Interval           Transfer     Bitrate         Retr  Cwnd
> [  5]   0.00-1.00   sec   284 MBytes  2.37 Gbits/sec   36   1.14 MBytes       
> [  5]   1.00-2.00   sec   280 MBytes  2.35 Gbits/sec    0   1.30 MBytes       
> [  5]   2.00-3.00   sec   281 MBytes  2.36 Gbits/sec    3   1.09 MBytes       
> [  5]   3.00-4.00   sec   280 MBytes  2.35 Gbits/sec    0   1.26 MBytes       
> [  5]   4.00-5.00   sec   281 MBytes  2.35 Gbits/sec    3   1.05 MBytes       
> [  5]   5.00-6.00   sec   281 MBytes  2.35 Gbits/sec    0   1.22 MBytes       
> [  5]   6.00-7.00   sec   280 MBytes  2.35 Gbits/sec    2   1.01 MBytes       
> [  5]   7.00-8.00   sec   279 MBytes  2.34 Gbits/sec    0   1.20 MBytes       
> [  5]   8.00-9.00   sec   281 MBytes  2.36 Gbits/sec    1    984 KBytes       
> [  5]   9.00-10.00  sec   281 MBytes  2.36 Gbits/sec    0   1.16 MBytes       
> - - - - - - - - - - - - - - - - - - - - - - - - -
> [ ID] Interval           Transfer     Bitrate         Retr
> [  5]   0.00-10.00  sec  2.74 GBytes  2.36 Gbits/sec   45            sender
> [  5]   0.00-10.01  sec  2.74 GBytes  2.35 Gbits/sec                  receiver
> 
> iperf Done.
> ```

I see high retransmission rates when running iperf3 through the EN8811H PHY. I thought it was just a problem with my board, but it also happens with the XG-040G board. The fix is to switch TCP congestion control to BBR, which keeps `cwnd` below 1024 Kbytes and the retransmission count at zero. I don't know why a `cwnd` larger than 1024 Kbytes causes iperf3 retransmissions.

robimarko:

geokvant:
> iperf3 through the EN8811H PHY

tested it now from LAN4 few times - zero Retr
From LAN1 - they appear 

```
#0 
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec  1.08 GBytes   928 Mbits/sec   75            sender
[  5]   0.00-10.01  sec  1.08 GBytes   925 Mbits/sec                  receiver

#1
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec  1.08 GBytes   927 Mbits/sec   90            sender
[  5]   0.00-10.02  sec  1.08 GBytes   924 Mbits/sec                  receiver

#2
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-60.00  sec  6.56 GBytes   938 Mbits/sec   53            sender
[  5]   0.00-60.02  sec  6.55 GBytes   938 Mbits/sec                  receiver
```
Don't understand - should it be fixed with some configuration settings or should be included in firmware code?

PR #24626, opened 2026-08-09
arm-trusted-firmware-airoha: sync an7581-bl31 BUILD_DEVICES with bl2

peterwillcn:
arm-trusted-firmware-airoha: sync an7581-bl31 BUILD_DEVICES with bl2

an7581-bl31 BUILD_DEVICES list was incomplete and missed:
 - airoha_an7581-evb-emmc-eagle
 - airoha_an7581-evb-emmc-kite
 - nokia_valyrian
 - nokia_xg-040g-md-ubi

These devices are present in an7581-bl2 but missing for bl31.
When building these targets, an7581‑bl31 package is never built,
causes FIP failure "fopen ...an7581‑bl31.lzma: No such file or directory".

Sync an7581‑bl31 BUILD_DEVICES to exactly match an7581‑bl2.

Fixes: https://github.com/openwrt/openwrt/commit/467c5b90ba1140dcdabc2e2e203c1d91fca9963e

### DIFF
--- a/package/boot/arm-trusted-firmware-airoha/Makefile
+++ b/package/boot/arm-trusted-firmware-airoha/Makefile
@@ -50,7 +50,7 @@ define Trusted-Firmware-A/an7581-bl31
   $(call Trusted-Firmware-A/bl31/Default)
   BUILD_SUBTARGET:=an7581
   SOC:=EN7581
-  BUILD_DEVICES:=airoha_an7581-evb airoha_an7581-evb-emmc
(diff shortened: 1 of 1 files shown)


peterwillcn:
ping @robimarko

robimarko:
Now that I look at it, why are eagle and kite not in the BL31 list as well?

peterwillcn:
Good catch! Indeed eagle, kite and also nokia_valyrian are missing from an7581‑bl31 BUILD_DEVICES.
I will update the patch to fully sync the device list to match an7581‑bl2.

robimarko:

PR #24627, opened 2026-08-09
base-files: wait for DSA netdev during preinit

### LINKED ISSUE #17982: filogic: GL.iNet GL-MT6000 does not enter failsafe with button press
### Describe the bug

during boot, the normal 5-per-second 2-second flashing occurs signaling the failsafe window. but if the only button (reset) is pressed during this window, the command is ignored and the device boots normally. the hardware button itself is confirmed to be functional.

### OpenWrt version

r28427-6df0e3d02a

### OpenWrt release

24.10.0

### OpenWrt target/subtarget

mediatek/filogic

### Device

GL.iNet GL-MT6000

### Image kind

Official downloaded image

### Steps to reproduce

press reset during failsafe window.

### Actual behaviour

boots normally.

### Expected behaviour

boots in failsafe mode.

### Additional info

_No response_

### Diffconfig

```text

```

### Terms

Jose-Tejera:
The GL-MT6000 can enter failsafe while remaining unreachable on its recovery
network. This is caused by the DSA `lan1` interface appearing after preinit's
single `/proc/net/dev` check, so `preinit_ip_config()` returns before assigning
the failsafe address.

Retry the netdev check for a bounded interval before giving up. The regression
test uses deterministic host-side command shims to model late `lan1`
registration and also verifies the GL-MT6000 reset-key metadata (GPIO 9,
active-low, `KEY_RESTART`).

Validation:

    package/base-files/tests/preinit_ip_config.sh
    preinit_ip_config and GL-MT6000 reset metadata: PASS

    dash package/base-files/tests/preinit_ip_config.sh
    preinit_ip_config and GL-MT6000 reset metadata: PASS

Fixes: #17982

### DIFF
--- a/package/base-files/files/lib/preinit/10_indicate_preinit
+++ b/package/base-files/files/lib/preinit/10_indicate_preinit
@@ -2,7 +2,7 @@
 # Copyright (C) 2010 Vertical Communications
 
 preinit_ip_config() {
-	local netdev vid
+	local netdev vid i
 
 	netdev=${1%\.*}
 	vid=${1#*\.}
@@ -11,6 +11,13 @@ preinit_ip_config() {
 		vid=
 	fi
 
+	# DSA interfaces may be registered after preinit starts.  Give the
+	# network device a short window to appear before abandoning failsafe
+	# network setup.
+	for i in 1 2 3 4 5; do
+		grep -q "$netdev" /proc/net/dev && break
+		[ "$i" -eq 5 ] || sleep 1
+	done
 	grep -q "$netdev" /proc/net/dev || return
 
 	if [ -n "$vid" ]; then
--- a/package/base-files/tests/preinit_ip_config.sh
+++ b/package/base-files/tests/preinit_ip_config.sh
@@ -1,0 +1,64 @@
+#!/bin/sh
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Regression test for boards whose DSA LAN device is registered after preinit
+# starts.  The command shims make the race deterministic and keep the test
+# runnable on a regular Linux host without a target board or CAP_NET_ADMIN.
+
+set -eu
+
+ROOT=$(CDPATH= cd -- "$(dirname "$0")/../../.." && pwd)
+TMPDIR=$(mktemp -d)
+trap 'rm -rf "$TMPDIR"' EXIT
+
+mkdir "$TMPDIR/bin"
+cat > "$TMPDIR/bin/grep" <<'EOF'
+#!/bin/sh
+
+if [ "$#" -eq 3 ] && [ "$1" = "-q" ] &&
+   [ "$2" = "lan1" ] && [ "$3" = "/proc/net/dev" ]; then
+	count_file=${PREINIT_TEST_TMP}/grep-count
+	count=$(cat "$count_file" 2>/dev/null || echo 0)
+	count=$((count + 1))
+	echo "$count" > "$count_file"
+	[ "$count" -ge 3 ]
+	exit $?
+fi
+
+exec /usr/bin/grep "$@"
+EOF
+cat > "$TMPDIR/bin/ip" <<'EOF'
+#!/bin/sh
+echo "$*" >> "$PREINIT_TEST_TMP/ip.log"
+EOF
+cat > "$TMPDIR/bin/sleep" <<'EOF'
+#!/bin/sh
+# Avoid making the deterministic retry test take five seconds.
+:
+EOF
+chmod +x "$TMPDIR/bin/grep" "$TMPDIR/bin/ip" "$TMPDIR/bin/sleep"
+
+export PREINIT_TEST_TMP="$TMPDIR"
+export PATH="$TMPDIR/bin:$PATH"
+
+boot_hook_add() { :; }
+. "$ROOT/package/base-files/files/lib/preinit/10_indicate_preinit"
+
+pi_ip=192.168.1.1
+pi_netmask=255.255.255.0
+pi_broadcast=192.168.1.255
+preinit_ip_config lan1
+
+/usr/bin/grep -q '^link set dev lan1 up$' "$TMPDIR/ip.log"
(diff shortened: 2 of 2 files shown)


package/base-files/files/lib/preinit/10_indicate_preinit
```diff
@@ -11,6 +11,13 @@ preinit_ip_config() {
...
 	fi
 
+	# DSA interfaces may be registered after preinit starts.  Give the
+	# network device a short window to appear before abandoning failsafe
+	# network setup.
+	for i in 1 2 3 4 5; do
+		grep -q "$netdev" /proc/net/dev && break
+		[ "$i" -eq 5 ] || sleep 1
+	done
 	grep -q "$netdev" /proc/net/dev || return
```
openwrt-ai, line 21: The loop already knows whether the device showed up, but the pre-existing `grep -q ... || return` below runs a sixth probe on `/proc/net/dev` for a result that is already decided. The retry bound is also written twice (`1 2 3 4 5` and `-eq 5`), so changing the window means editing two places that must stay in sync.

A `while` loop keeps the exact same behaviour (5 probes, 4 x `sleep 1`, `return` on timeout) with one bound and one probe per iteration:

```suggestion
	# DSA interfaces may be registered after preinit starts.  Give the
	# network device a short window to appear before abandoning failsafe
	# network setup.
	i=0
	while ! grep -q "$netdev" /proc/net/dev; do
		i=$((i + 1))
		[ "$i" -lt 5 ] || return
		sleep 1
	done
```

I ran `package/base-files/tests/preinit_ip_config.sh` from this PR against that variant under `sh`, `dash` and `bash` — it still passes (the shim's `grep-count` reaches exactly 3, satisfying the `-ge 3` assertion), and it still fails when the retry is removed entirely.

---

package/base-files/tests/preinit_ip_config.sh
```diff
@@ -0,0 +1,64 @@
...
+   [ "$2" = "lan1" ] && [ "$3" = "/proc/net/dev" ]; then
+	count_file=${PREINIT_TEST_TMP}/grep-count
+	count=$(cat "$count_file" 2>/dev/null || echo 0)
+	count=$((count + 1))
+	echo "$count" > "$count_file"
+	[ "$count" -ge 3 ]
+	exit $?
+fi
+
+exec /usr/bin/grep "$@"
```
openwrt-ai, line 28: `/usr/bin/grep` is not a portable path. On Alpine and on non-usrmerged Debian/Devuan grep is `/bin/grep`, and on NixOS it only exists in the store — the shim's fallback then dies with "not found", and so do the six assertions that also hardcode it (lines 52, 54 and 60-62). Since the whole point of the shim is that `PATH` no longer resolves the real binary, resolve it once *before* line 42 rewrites `PATH` and export it, e.g. `REAL_GREP=$(command -v grep); export REAL_GREP`, then use `exec "$REAL_GREP" "$@"` here and `"$REAL_GREP"` at the other five call sites.

---

package/base-files/tests/preinit_ip_config.sh
```diff
@@ -0,0 +1,64 @@
...
+expected='-4 address add 192.168.1.1/255.255.255.0 broadcast 192.168.1.255 dev lan1'
+/usr/bin/grep -Fqx -- "$expected" "$TMPDIR/ip.log"
+[ "$(cat "$TMPDIR/grep-count")" -ge 3 ]
+
+# Keep the board contract covered alongside the preinit race: the reset key
+# must remain the active-low GPIO 9 KEY_RESTART input used by failsafe.
+DTS="$ROOT/target/linux/mediatek/dts/mt7986a-glinet-gl-mt6000.dts"
+/usr/bin/grep -Fq 'label = "reset";' "$DTS"
+/usr/bin/grep -Fq 'linux,code = <KEY_RESTART>;' "$DTS"
+/usr/bin/grep -Fq 'gpios = <&pio 9 GPIO_ACTIVE_LOW>;' "$DTS"
```
openwrt-ai, line 62: These three assertions exercise no base-files code at all — they string-match [`mt7986a-glinet-gl-mt6000.dts:49-53`](https://github.com/openwrt/openwrt/blob/d622b669cfedb9269860b3d1d2922255a40453e7/target/linux/mediatek/dts/mt7986a-glinet-gl-mt6000.dts#L49-L53), and the reset key was never part of the bug being fixed (the failsafe address assignment was). The coupling is also brittle in a way that will bite someone unrelated: reindenting the node, moving `keys` into a `.dtsi`, or writing the GPIO flag numerically all break a base-files test with no connection to the change.

Suggest dropping lines 56-62 so the test stays scoped to `preinit_ip_config()`. If the reset-key contract is worth pinning, it belongs in a mediatek-side check, not here.

---

package/base-files/tests/preinit_ip_config.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh
```
openwrt-ai, line 1: nit: nothing in the tree will ever run this file. There is no other `tests/` directory under `package/` (the only `test` dir is vendor source in `package/kernel/lantiq/vrx518_ep/src/test`), no workflow in `.github/workflows/` invokes shell tests, and `Package/base-files/install` only copies `./files/*`, so the script is neither packaged nor executed by CI — it runs only if a person remembers to invoke it by hand.

The test itself is sound (I ran it under `sh`, `dash` and `bash`: it passes with the patch and correctly fails with the retry removed), which makes it worth wiring up. Either add a CI job that runs it, or at least state the invocation in the commit message so it is discoverable — otherwise it will silently rot the first time `preinit_ip_config()` changes shape.

---

openwrt-ai (commented):
Reviewed 1 new commit; the commit message matches its diff. 4 inline comments.

Note on the bot's `PKG_RELEASE` warning above: `package/base-files/Makefile` sets `PKG_RELEASE:=$(COMMITCOUNT)`, so there is nothing to bump here — that suggestion looks like a false positive for this package.

---

PR #24628, opened 2026-08-09
odhcp6c: allow disabling RA address assignment

### LINKED ISSUE #22469: odhcp6c applies unroutable SLAAC addresses from WAN RA
### Describe the bug

When the upstream ISP/ONT sends Router Advertisements with the A (Autoconfiguration) flag set on prefixes that are not routable back to the CPE, odhcp6c + dhcpv6.script installs these addresses on the WAN interface via netifd. The kernel then selects one of these unroutable addresses as the source for router-originated traffic (ping, dns, apk, etc.), causing 100% packet loss for all IPv6 traffic originating from the router itself. LAN clients are unaffected because they use the delegated prefix which is properly routed by the ISP.

There is currently no UCI option to disable RA-derived address assignment on the WAN interface while keeping RA route processing and DHCPv6-PD functional.

### OpenWrt version

r32713-f919e7899d

### OpenWrt release

25.12.0

### OpenWrt target/subtarget

mediatek/filogic

### Device

GL.iNet GL-MT6000

### Image kind

Official downloaded image

### Steps to reproduce

1. Configure a standard wan6 interface with `proto dhcpv6` and `reqprefix auto`
2. ISP upstream sends RAs with Prefix Information Options that have the A (autoconfiguration) flag set
3. ISP also provides a delegated prefix via DHCPv6-PD
4. Observe that odhcp6c installs SLAAC addresses from the RA on the WAN interface (eth1)

### Actual behaviour

Pinging any IPv6 host from the router fails, but it works when the source is explicitly set to the delegated prefix assigned to br-lan.

### Expected behaviour

There should be a UCI option (e.g., `option noraaddr '1'`) to prevent dhcpv6.script from merging RA-derived addresses into the interface configuration.

### Additional info

In `/lib/netifd/dhcpv6.script`, the setup_interface function iterates over `$RA_ADDRESSES` and merges them into the `$ADDRESSES `variable, which is then pushed to netifd:

```sh
for entry in $RA_ADDRESSES; do
        local duplicate=0
        local addr="${entry%%/*}"
        for dentry in $ADDRESSES; do
                local daddr="${dentry%%/*}"
                [ "$addr" = "$daddr" ] && duplicate=1
        done
        [ "$duplicate" = "0" ] && ADDRESSES="$ADDRESSES $entry"
done
```

My workaround is commenting out the merge line :S

```
sed -i 's/\[ "$duplicate" = "0" \] && ADDRESSES="$ADDRESSES $entry"/# &/' /lib/netifd/dhcpv6.script
```

PS: My ISP is Netlife Ecuador and the ONT is VSOL V2802RH in bridge mode.

### Diffconfig

```text

```

### Terms

Jose-Tejera:
Some WAN Router Advertisements advertise SLAAC prefixes that are not routable
back to the CPE. `odhcp6c` currently merges those RA-derived addresses into
the interface unconditionally, allowing the kernel to select an unusable
source address and breaking router-originated IPv6 traffic.

Add the `noraaddr` DHCPv6 option. When enabled, `dhcpv6.script` skips merging
RA-derived addresses while retaining RA routes, DNS, and delegated prefixes;
the default behavior remains unchanged. The option is propagated from UCI
through the DHCPv6 protocol handler.

Validation:

    package/network/ipv6/odhcp6c/tests/test-noraaddr.sh
    noraaddr regression passed

Fixes: #22469

### DIFF
--- a/package/network/ipv6/odhcp6c/files/dhcpv6.script
+++ b/package/network/ipv6/odhcp6c/files/dhcpv6.script
@@ -64,16 +64,18 @@ setup_interface () {
 		proto_add_ipv6_prefix "$prefix"
 	done
 
-	# Merge addresses
-	for entry in $RA_ADDRESSES; do
-		local duplicate=0
-		local addr="${entry%%/*}"
-		for dentry in $ADDRESSES; do
-			local daddr="${dentry%%/*}"
-			[ "$addr" = "$daddr" ] && duplicate=1
+	# Merge addresses unless RA-derived addresses are disabled
+	if [ "$NORAADDR" != "1" ]; then
+		for entry in $RA_ADDRESSES; do
+			local duplicate=0
+			local addr="${entry%%/*}"
+			for dentry in $ADDRESSES; do
+				local daddr="${dentry%%/*}"
+				[ "$addr" = "$daddr" ] && duplicate=1
+			done
+			[ "$duplicate" = "0" ] && ADDRESSES="$ADDRESSES $entry"
 		done
-		[ "$duplicate" = "0" ] && ADDRESSES="$ADDRESSES $entry"
-	done
+	fi
 
 	for entry in $ADDRESSES; do
 		local addr="${entry%%/*}"
--- a/package/network/ipv6/odhcp6c/tests/test-noraaddr.sh
+++ b/package/network/ipv6/odhcp6c/tests/test-noraaddr.sh
@@ -1,0 +1,39 @@
+#!/bin/sh
+
+set -eu
+
+script=${1:-"$(dirname "$0")/../files/dhcpv6.script"}
+
+merge=$(awk '
+/^[[:space:]]*# Merge addresses/ { in_block = 1; next }
+/^[[:space:]]*for entry in \$ADDRESSES; do$/ { exit }
+in_block { print }
+' "$script")
+
+run_merge()
+{
+	eval "$merge"
+}
+
+original='2001:db8:1::1/64,3600,7200'
+ra_address='2001:db8:ffff::1/64,3600,7200'
+
+ADDRESSES="$original"
+RA_ADDRESSES="$ra_address"
+NORAADDR=1
+run_merge
+[ "$ADDRESSES" = "$original" ] || {
+	echo "noraaddr failed to suppress an RA-derived address" >&2
+	exit 1
+}
+
+ADDRESSES="$original"
+RA_ADDRESSES="$ra_address"
+NORAADDR=0
+run_merge
+case " $ADDRESSES " in
+	*" $ra_address "*) ;;
+	*) echo "RA-derived address was not merged by default" >&2; exit 1 ;;
+esac
+
+echo "noraaddr regression passed"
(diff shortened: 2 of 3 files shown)


package/network/ipv6/odhcp6c/files/dhcpv6.script
```diff
@@ -64,16 +64,18 @@ setup_interface () {
...
 
-	# Merge addresses
-	for entry in $RA_ADDRESSES; do
-		local duplicate=0
-		local addr="${entry%%/*}"
-		for dentry in $ADDRESSES; do
-			local daddr="${dentry%%/*}"
-			[ "$addr" = "$daddr" ] && duplicate=1
+	# Merge addresses unless RA-derived addresses are disabled
+	if [ "$NORAADDR" != "1" ]; then
```
openwrt-ai, line 68: Keeping RA addresses out of `$ADDRESSES` also removes them as *source* prefixes for RA route installation further down: [`for prefix in $PREFIXES $ADDRESSES` at dhcpv6.script:140-146](https://github.com/openwrt/openwrt/blob/858eaf8ae3ba2b811c7a0018506f1fbb41f43b0a/package/network/ipv6/odhcp6c/files/dhcpv6.script#L140-L146) installs a gatewayed RA route once per entry of that list, and only falls back to an unsourced route when `NOSOURCEFILTER=1` (i.e. `option sourcefilter '0'`).

So with `noraaddr=1` on an interface that ends up with neither an IA_NA address nor a delegated prefix (RA-only upstream, or `reqaddress none` together with no PD), `$PREFIXES $ADDRESSES` is empty and the RA default route is silently not installed at all — which doesn't match "while retaining RA routes" in the commit message. The [RFC 7278 `extendprefix` block at :96-105](https://github.com/openwrt/openwrt/blob/858eaf8ae3ba2b811c7a0018506f1fbb41f43b0a/package/network/ipv6/odhcp6c/files/dhcpv6.script#L96-L105) is similarly bypassed, since it only iterates `$ADDRESSES`.

For the reported case in #22469 (PD present) this is fine. Is the no-PD case intended to lose the default route, or should the route loop fall back to an unsourced route when `NORAADDR=1` leaves nothing to source from? Worth at least documenting the interaction with `sourcefilter` in the commit message.

---

package/network/ipv6/odhcp6c/tests/test-noraaddr.sh
openwrt-ai, line 1: This test never runs `dhcpv6.script`; it `awk`-extracts the source text between the `# Merge addresses` comment and `for entry in $ADDRESSES; do` and `eval`s that copy. Any refactor that reworks the comment or reorders the block leaves `$merge` empty, and the first assertion then passes vacuously — `run_merge` is a no-op, so `ADDRESSES` trivially still equals `$original`. The check is only kept honest by the second case.

There is also no precedent for package-local shell tests in this tree (`find package -type d -name tests` returns nothing) and nothing under `.github/workflows/` executes anything in a `tests/` directory, so this file is never run by CI and will rot unnoticed.

Suggest dropping it from the PR and keeping the validation note in the commit message / PR description, or wiring it into something that actually gets executed.

---

openwrt-ai (commented):
## Commit checks

- `858eaf8` "odhcp6c: allow disabling RA address assignment" — the PR body carries `Fixes: #22469`, but the commit message has no issue reference at all, so the link to the bug report is lost once the PR is squashed/merged. Recent in-tree practice for referencing an issue is the full URL form (`Fixes: https://github.com/openwrt/openwrt/issues/22469`, 26 occurrences in the last 5000 commits, vs 8 for the bare `Fixes: #NNNN`).


---

bayasdev:
Thanks @Jose-Tejera!

PR #24630, opened 2026-08-09
sdk: build apk host dependency

Jose-Tejera:
The SDK's APK package build can start before the APK host tools have been
built. APK's Meson configuration then cannot find the required `lua5.1`
program, even though the normal build graph provides it in host staging.

Make the APK package compile target depend on
`package/system/apk/host/compile` when `CONFIG_USE_APK` is enabled. Add a
Linux-runnable regression that inspects the generated SDK make graph and
requires the host prerequisite to precede the target compile.

Validation:

    scripts/test-sdk-apk-lua.sh
    SDK apk lua host dependency test: ok

Fixes: #21082

### DIFF
--- a/package/Makefile
+++ b/package/Makefile
@@ -88,7 +88,7 @@ else
 endif
 else
 ifneq ($(CONFIG_USE_APK),)
-  $(curdir)//compile += $(BUILD_KEY_APK_SEC) $(BUILD_KEY_APK_PUB)
+  $(curdir)//compile += $(curdir)/system/apk/host/compile $(BUILD_KEY_APK_SEC) $(BUILD_KEY_APK_PUB)
 endif
 endif
 
--- a/scripts/test-sdk-apk-lua.sh
+++ b/scripts/test-sdk-apk-lua.sh
@@ -1,0 +1,56 @@
+#!/bin/sh
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# Regression test for the SDK apk host-build dependency.
+#
+# Copyright (C) 2026 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+set -eu
+
+SCRIPTDIR=$(dirname "$0")
+BASEDIR=$(cd "$SCRIPTDIR/.." && pwd)
+TEST_TMP=$(mktemp -d "${TMPDIR:-/tmp}/sdk-apk-lua.XXXXXX")
+trap 'rm -rf "$TEST_TMP"' EXIT HUP INT TERM
+
+cat > "$TEST_TMP/.packagedeps" <<'EOF'
+package-m += system/apk
+EOF
+
+cat > "$TEST_TMP/Makefile" <<'EOF'
+TOPDIR := @TOPDIR@
+TMP_DIR := @TMP_DIR@
+INCLUDE_DIR := $(TOPDIR)/include
+SCRIPT_DIR := $(TOPDIR)/scripts
+BUILD_DIR := $(TOPDIR)/build_dir
+STAGING_DIR := $(TOPDIR)/staging_dir/target
+CONFIG_USE_APK := y
+SDK := 1
+.DEFAULT_GOAL := dump
+
+include $(TOPDIR)/rules.mk
+curdir := package
+include $(TOPDIR)/package/Makefile
+include $(TOPDIR)/include/subdir.mk
+$(eval $(call subdir,$(curdir)))
+
+dump:
+EOF
+
+sed -i \
+	-e "s|@TOPDIR@|$BASEDIR|" \
+	-e "s|@TMP_DIR@|$TEST_TMP|" \
+	"$TEST_TMP/Makefile"
+
+make -rnp -f "$TEST_TMP/Makefile" > "$TEST_TMP/make.db"
+
+APK_COMPILE='^package/system/apk/compile:.*package/system/apk/host/compile'
+if ! grep -Eq "$APK_COMPILE" "$TEST_TMP/make.db"; then
+	echo "SDK apk compile graph is missing its host build prerequisite" >&2
+	exit 1
+fi
+
+echo "SDK apk lua host dependency test: ok"


package/Makefile
```diff
@@ -88,7 +88,7 @@ else
 endif
 else
 ifneq ($(CONFIG_USE_APK),)
-  $(curdir)//compile += $(BUILD_KEY_APK_SEC) $(BUILD_KEY_APK_PUB)
+  $(curdir)//compile += $(curdir)/system/apk/host/compile $(BUILD_KEY_APK_SEC) $(BUILD_KEY_APK_PUB)
```
openwrt-ai, line 91: `package/system/apk` does not exist in an SDK tree, so this prerequisite has no rule there and every SDK build with `CONFIG_USE_APK` will abort.

The SDK tarball only copies `package/Makefile`, `package/libs/toolchain` and `package/kernel/linux` into `package/` ([target/sdk/Makefile:161-168](https://github.com/openwrt/openwrt/blob/8eb7abf2d3a0e61395a432f42211c3be750722eb/target/sdk/Makefile#L161-L168)). apk arrives via the base feed, and `scripts/feeds` symlinks feed packages to `package/feeds/<feed>/<name>` ([scripts/feeds:446-448](https://github.com/openwrt/openwrt/blob/8eb7abf2d3a0e61395a432f42211c3be750722eb/scripts/feeds#L446-L448)), so the builddir is `feeds/base/apk`. The traceback in #21082 shows exactly that: `make[1]: *** [package/Makefile:187: package/feeds/base/apk/compile] Error 1`.

Because `$(curdir)//compile` is appended to *every* package's compile target, the missing rule takes down unrelated packages too. Reproduced against this PR's head with an SDK-shaped graph (`SDK:=1`, `CONFIG_USE_APK:=y`):

```
make: *** No rule to make target 'package/system/apk/host/compile',
      needed by 'package/kernel/linux/compile'.  Stop.
```

The root cause is that the *target* apk build needs a build-host `lua5.1`, which only `lua/host` provides (in `staging_dir/hostpkg/bin`, a directory the SDK does not ship either). That is better expressed as a package-level build dependency, resolved by name through the metadata instead of by hardcoded subdir — e.g. add `PKG_BUILD_DEPENDS:=lua/host` alongside the existing [`HOST_BUILD_DEPENDS:=lua/host` at package/system/apk/Makefile:19](https://github.com/openwrt/openwrt/blob/8eb7abf2d3a0e61395a432f42211c3be750722eb/package/system/apk/Makefile#L19). Precedent for the pattern: [`PKG_BUILD_DEPENDS:=ruby/host` in arm-trusted-firmware-microchipsw](https://github.com/openwrt/openwrt/blob/8eb7abf2d3a0e61395a432f42211c3be750722eb/package/boot/arm-trusted-firmware-microchipsw/Makefile#L12). That form works identically in a full build and in the SDK, wherever the package directory happens to live.

---

scripts/test-sdk-apk-lua.sh
```diff
@@ -0,0 +1,56 @@
...
+TEST_TMP=$(mktemp -d "${TMPDIR:-/tmp}/sdk-apk-lua.XXXXXX")
+trap 'rm -rf "$TEST_TMP"' EXIT HUP INT TERM
+
+cat > "$TEST_TMP/.packagedeps" <<'EOF'
+package-m += system/apk
+EOF
+
+cat > "$TEST_TMP/Makefile" <<'EOF'
+TOPDIR := @TOPDIR@
+TMP_DIR := @TMP_DIR@
```
openwrt-ai, line 25: This `TMP_DIR` is discarded, so the test never exercises what it claims and currently fails on this PR's own head.

`include $(TOPDIR)/rules.mk` on line 34 pulls in `include/scan.mk`, which reassigns [`TMP_DIR:=$(TOPDIR)/tmp`](https://github.com/openwrt/openwrt/blob/8eb7abf2d3a0e61395a432f42211c3be750722eb/include/scan.mk#L3) unconditionally. `package/Makefile`'s `-include $(TMP_DIR)/.packagedeps` therefore reads `$(TOPDIR)/tmp/.packagedeps`, never the fixture written on lines 19-21. `system/apk` never enters `$(package/builddirs)`, `package/system/apk/compile` is never generated, and the grep on line 50 can never match. Confirmed in the generated `make.db`: the only mention is `# Not a target:` / `package/system/apk/host/compile:`.

Running the script unmodified on `8eb7abf` in a clean checkout:

```
$ scripts/test-sdk-apk-lua.sh
SDK apk compile graph is missing its host build prerequisite
$ echo $?
1
```

So the `SDK apk lua host dependency test: ok` quoted in the commit message and PR body is not reproducible — it presumably came from a tree that had a stale `tmp/.packagedeps` listing `system/apk`. A command-line assignment (`make TMP_DIR=...`) would survive, since command-line variables beat file assignments; setting it after the `rules.mk` include would too.

Note that fixing the plumbing alone is not enough: the assertion on line 50 hardcodes `package/system/apk/compile`, which is precisely the path that does *not* exist in a real SDK (it is `package/feeds/base/apk/compile` there — see the traceback in #21082). As written, the fixture models a layout the SDK never has, so a green result would not prove the SDK is fixed.

---

scripts/test-sdk-apk-lua.sh
```diff
@@ -0,0 +1,56 @@
...
+include $(TOPDIR)/rules.mk
+curdir := package
+include $(TOPDIR)/package/Makefile
+include $(TOPDIR)/include/subdir.mk
+$(eval $(call subdir,$(curdir)))
+
+dump:
+EOF
+
+sed -i \
```
openwrt-ai, line 43: nit: `sed -i` without a suffix argument is a GNU extension and fails on BSD/macOS hosts, which OpenWrt's build system still supports (hence the `$(SED)` wrapper used elsewhere in-tree). The placeholder round-trip is avoidable altogether here — the here-doc on line 23 is only quoted to protect the `$(...)` make expansions, so escaping those and interpolating `$BASEDIR`/`$TEST_TMP` directly drops lines 43-46 entirely.

nit: nothing in the tree invokes `scripts/test-sdk-apk-lua.sh` — there is no test harness under `scripts/` and no CI workflow picks it up, so it will silently rot. Worth either wiring it into a workflow or dropping it and describing the manual reproduction in the commit message instead.

---

openwrt-ai (commented):
## Commit checks

- `8eb7abf2` "sdk: build apk host dependency" — subject scope does not match the files touched: the diff changes `package/Makefile` and adds `scripts/test-sdk-apk-lua.sh`, nothing under `target/sdk/`. The in-tree prefix for `package/Makefile` changes is `build:` (cf. `10d9dbe8 build: add CycloneDX SBOM processing to apk`).
- `8eb7abf2` "sdk: build apk host dependency" — the body claims "Add a deterministic SDK make-graph regression", but the added script is not deterministic with respect to the tree: it depends on `$(TOPDIR)/tmp/.packagedeps` rather than the fixture it writes, and in a clean checkout it exits 1 on this very commit. The `Validation: ... ok` transcript in the PR body cannot be reproduced. See the inline comment on `scripts/test-sdk-apk-lua.sh`.

Two inline findings on the change itself; the `package/Makefile` one looks like it would break SDK builds rather than fix them.

---

PR #18961, opened 2025-05-30
ramips: mt7621: only include uboot-envtools on devices which require it

map-b:
Since commits from PR https://github.com/openwrt/openwrt/pull/12959 have fixed initramfs images, including correctly DEVICE_PACKAGES per device. This includes images generated by builbots, that is, using a config with:

 CONFIG_TARGET_MULTI_PROFILE=y
 CONFIG_TARGET_ALL_PROFILES=y
 CONFIG_TARGET_PER_DEVICE_ROOTFS=y

So, do not abuse of DEFAULT_PACKAGES anymore. I propose:
-Manually revert commit 79bd017 plus,
-Manually checking all new devices that have been added support since then(March 2023), for example following this git history: https://github.com/openwrt/openwrt/commits/main/package/boot/uboot-tools/uboot-envtools/files/ramips

Remove "uboot-envtools" from target.mk's DEFAULT_PACKAGES, remove about 158 occurrences of "-uboot-envtools" from DEVICE_PACKAGES and include 41 "uboot-envtools" only in the required recipes, approximately 50 devices (some of them use shared recipes).

While at here, make some clean up:
-wrap/unwrap lines to 80 characters,
-align indentations and
-sort the packages defined in DEVICE_PACKAGES alphabetically.

I've split the changes just trying to make easier the review. If it is better to only do the additions and removals of the package and do not change anything else, please, tell me.

-The indentations are mixed, two spaces but tabulations for the line wrappings. It seems a common practice with makefiles, Is it correct?
-Which max line size is preferred 80, 100?(I read about 80 characters limit being deprecated upstream)
-Should the packages in the DEVICE_PACKAGES lists be sorted alphabetically?

PS: My original intention was to fix the issue of some devices not including uboot-envtools (some Xiaomi devices using a shared recipe, ie:R3Gv1)

Fixes: 79bd017 ("mt7621: move uboot-envtools to DEFAULT_PACKAGES")
Fixes: https://github.com/openwrt/openwrt/issues/18960 ramips: mt7621: uboot-envtools is not included in some devices like Xiaomi R3Gv1

PR #24631, opened 2026-08-09
ramips: mt7621: include uboot-envtools for EdgeRouter X

Jose-Tejera:
The EdgeRouter X removes `uboot-envtools` from its device package list even
though it is supplied by the mt7621 target defaults. An installed
`uboot-envtools` package can therefore be absent from the generated manifest,
causing owut's installed-package selection to fail.

Stop subtracting `uboot-envtools` from the EdgeRouter X device packages and
add a deterministic manifest-selection regression. The target default remains
the single source of the package while the device manifest now includes it.

Validation:

    tests/owut-installed-package.sh
    (exited successfully)

Fixes: #22527

### DIFF
--- a/target/linux/ramips/image/mt7621.mk
+++ b/target/linux/ramips/image/mt7621.mk
@@ -3255,7 +3255,7 @@ define Device/ubnt_edgerouter_common
   FILESYSTEMS := squashfs
   KERNEL_SIZE := 6144k
   IMAGE/sysupgrade.bin := sysupgrade-tar | append-metadata
-  DEVICE_PACKAGES += -wpad-basic-mbedtls -uboot-envtools
+  DEVICE_PACKAGES += -wpad-basic-mbedtls
   DEVICE_COMPAT_VERSION := 2.0
   DEVICE_COMPAT_MESSAGE :=  Partition table has been changed due to kernel size restrictions. \
     Refer to the wiki page for instructions to migrate to the new layout: \
--- a/tests/owut-installed-package.sh
+++ b/tests/owut-installed-package.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Ensure the EdgeRouter X image manifest retains uboot-envtools, which is a
+# target default and may already be installed when owut selects packages.
+
+set -eu
+
+TOPDIR=$(CDPATH= cd -- "$(dirname -- "$0")/.." && pwd)
+device_file=$TOPDIR/target/linux/ramips/image/mt7621.mk
+target_file=$TOPDIR/target/linux/ramips/mt7621/target.mk
+
+device=$(sed -n '/^define Device\/ubnt_edgerouter_common$/,/^endef$/p' "$device_file")
+defaults=$(sed -n 's/^DEFAULT_PACKAGES += //p' "$target_file")
+
+# The device must not subtract a package supplied by the target defaults.
+test -n "$device"
+if printf '%s\n' "$device" | grep -q -- '-uboot-envtools'; then
+	exit 1
+fi
+printf '%s\n' "$defaults" | grep -qw uboot-envtools
+
+# Model owut's installed-package selection against the generated manifest.
+manifest="$defaults $(printf '%s\n' "$device" | sed -n 's/.*DEVICE_PACKAGES[^=]*=[[:space:]]*//p')"
+printf '%s\n' "$manifest" | grep -qw uboot-envtools


target/linux/ramips/image/mt7621.mk
```diff
@@ -3255,7 +3255,7 @@ define Device/ubnt_edgerouter_common
   FILESYSTEMS := squashfs
   KERNEL_SIZE := 6144k
   IMAGE/sysupgrade.bin := sysupgrade-tar | append-metadata
-  DEVICE_PACKAGES += -wpad-basic-mbedtls -uboot-envtools
+  DEVICE_PACKAGES += -wpad-basic-mbedtls
```
openwrt-ai, line 3258: The one-line change looks substantively right: the ER-X/ER-X-SFP `u-boot-env` partition is mtd1 ([`mt7621_ubnt_edgerouter-x.dtsi:84-86`](https://github.com/openwrt/openwrt/blob/948087367666058de41ed87dcedeb28f988986bd/target/linux/ramips/dts/mt7621_ubnt_edgerouter-x.dtsi#L84-L86)) and both boards already have a matching ubootenv config at [`uboot-envtools/files/ramips:79-80`](https://github.com/openwrt/openwrt/blob/948087367666058de41ed87dcedeb28f988986bd/package/boot/uboot-tools/uboot-envtools/files/ramips#L79-L80), so the subtraction was making a supported package uninstallable-by-manifest for no reason. The subtraction was introduced by [79bd017 "mt7621: move uboot-envtools to DEFAULT_PACKAGES"](https://github.com/openwrt/openwrt/commit/79bd0172ca5671f7cdf382dc80ec1630da057947).

However, the EdgeRouter X is not the only mt7621 device in that state. Cross-referencing every `TARGET_DEVICES` entry in this file that still carries `-uboot-envtools` against the board list in `uboot-envtools/files/ramips` gives 9 more devices with a ubootenv config but the package subtracted, i.e. all hitting the exact same owut failure from #22527:

`arcadyan,we420223-99`, `buffalo,wsr-1166dhp`, `buffalo,wsr-600dhp`, `netgear,eax12`, `netgear,wax202`, `xiaomi,mi-router-3g-v2`, `xiaomi,mi-router-4a-gigabit`, `yuncore,ax820`, `zbtlink,zbt-wg2626`

The precedent fix [6dffaf93 "ramips: mt7621: include uboot-envtools in some missed devices"](https://github.com/openwrt/openwrt/commit/6dffaf931675e70464f5cba98a1ccc707b0d784c) handled its set in one batch. Is limiting this to the EdgeRouter X intentional (e.g. only that one was reported), or would you fold the other nine in so the class of bug is actually closed?

---

tests/owut-installed-package.sh
openwrt-ai, line 1: This creates a new top-level `tests/` directory that does not exist in the tree today, and nothing will ever run the script: no workflow under `.github/workflows/` references `tests/`, and there is no top-level make target or CI harness that discovers shell scripts there. As committed it is dead code that will silently rot the first time `Device/ubnt_edgerouter_common` is renamed or the block is restructured.

It is also outside the scope declared by the commit subject (`ramips: mt7621: ...`) — the actual fix is the one-line `mt7621.mk` change, which is what should be backported to `openwrt-25.12` for #22527.

Suggest dropping this file from the commit. If you want a guard against re-introducing `-uboot-envtools` for boards that have a ubootenv config, it would need to be a generic check (walk every target's `image/*.mk` against `package/boot/uboot-tools/uboot-envtools/files/<target>`, not one hard-coded device block) and wired into a workflow — that is a separate PR.

---

tests/owut-installed-package.sh
```diff
@@ -0,0 +1,25 @@
...
+# The device must not subtract a package supplied by the target defaults.
+test -n "$device"
+if printf '%s\n' "$device" | grep -q -- '-uboot-envtools'; then
+	exit 1
+fi
+printf '%s\n' "$defaults" | grep -qw uboot-envtools
+
+# Model owut's installed-package selection against the generated manifest.
+manifest="$defaults $(printf '%s\n' "$device" | sed -n 's/.*DEVICE_PACKAGES[^=]*=[[:space:]]*//p')"
+printf '%s\n' "$manifest" | grep -qw uboot-envtools
```
openwrt-ai, line 25: This "manifest model" check can never fail, so it asserts nothing.

`manifest` is `"$defaults …"`, and line 21 already asserted `$defaults` matches `grep -qw uboot-envtools`. Since `manifest` is a strict superset of `defaults`, line 25 is true by construction regardless of what the device block contains.

It would not catch the regression even without line 21: `grep -w` treats `-` as a non-word character, so `grep -qw uboot-envtools` also matches the string `-uboot-envtools`. Concatenating the device's `DEVICE_PACKAGES` text onto the defaults does not model package subtraction at all — the only check doing real work here is the `-uboot-envtools` grep on line 18.

The `sed` on line 24 is additionally fragile: it stops at the first line, so it drops `DEVICE_PACKAGES` values continued with a trailing `\` (very common in this file) and ignores packages inherited via `$(Device/...)`.

---

openwrt-ai (commented):
## Commit checks

- `9480873` "ramips: mt7621: include uboot-envtools for EdgeRouter X" — subject and body name only the EdgeRouter X, but the hunk edits `Device/ubnt_edgerouter_common`, which is also inherited by `ubnt_edgerouter-x-sfp` (EdgeRouter X SFP / EdgePoint R6). Please name both devices, the way [6dffaf93](https://github.com/openwrt/openwrt/commit/6dffaf931675e70464f5cba98a1ccc707b0d784c) enumerated the devices it touched.
- `9480873` — "add a deterministic installed-package selection regression" does not describe what the diff actually adds: a brand-new top-level `tests/` directory containing a standalone, unreferenced shell script (see the inline comment on that file). Either the message should say so plainly, or the file should be dropped.
- `9480873` — nit: the `Fixes:` reference exists only in the PR body, not in the commit message. In-tree convention (again [6dffaf93](https://github.com/openwrt/openwrt/commit/6dffaf931675e70464f5cba98a1ccc707b0d784c)) is to carry it in the commit, e.g. `Fixes: 79bd0172ca56 ("mt7621: move uboot-envtools to DEFAULT_PACKAGES")` plus `Fixes: https://github.com/openwrt/openwrt/issues/22527`. Note the real subject of `79bd0172` is "mt7621: move uboot-envtools to DEFAULT_PACKAGES" — `6dffaf93` cites it under a different title, don't copy that.


---

FredApps:
Worth coordinating with #18961, which is going the other way for this device.

That PR drops `uboot-envtools` from the mt7621 `DEFAULT_PACKAGES` and adds it explicitly to ~41 recipes instead. The EdgeRouter X is not one of them — neither `Device/ubnt_edgerouter-x` nor the shared `Device/ubnt_edgerouter_common` gains it on that branch — so an ERX image would ship without the package and #22527 would come back for anyone who installs it by hand.

Both PRs cite the same underlying problem, so whichever lands first will want the other rebased on top rather than merged blind.

PR #24634, opened 2026-08-09
scripts: account for built-in module dependencies

### LINKED ISSUE #23094: Compilation problem on the `microchipsw/lan969x` architecture: Package kmod-crypto-xxhash is missing dependencies
### Describe the bug

I'm building a package for OpenWRT and I've encountered a problem that my package is not building on the `microchipsw/lan969x` architecture. Error is not related to my package, but it is related to the Linux kernel compilation itself:

```
(full logs are below in Actual behavior)
Package kmod-crypto-xxhash is missing dependencies for the following libraries:
xxhash.ko
````

I've tried to run this on `25.12.0` and `25.12.2`, I get the same error.

### OpenWrt version

Docker image:
`openwrt/sdk@sha256:a4dc7b3e0935d0abc17e800ae6e8421e1a13e91d09386eeeda10d191c03d65c0`
`openwrt/sdk:microchipsw-lan969x-25.12.2`

### OpenWrt release

25.12.2

### OpenWrt target/subtarget

microchipsw/lan969x

### Device

no device, it's SDK

### Image kind

Self-built image

### Steps to reproduce

Minimal reproduction:
```
docker run --rm -it openwrt/sdk:microchipsw-lan969x-25.12.2

# Inside docker container:
./scripts/feeds update -a
make defconfig
make package/kernel/linux/compile V=s
```

### Actual behaviour

```
...
/builder/staging_dir/host/bin/fakeroot /builder/staging_dir/host/bin/apk mkpkg --info "name:kmod-crypto-xxhash" --info "version:6.12.74-r1"  --info "description:xxHash non-cryptographic hash algorithm" --info "arch:aarch64_cortex-a53" --info "license:GPL-2.0-only" --info "origin:feeds/base/kernel/linux" --info "url:" --info "maintainer:" --info "provides:kmod-crypto-xxhash-any"  --script "post-install:/builder/build_dir/target-aarch64_cortex-a53_musl/linux-microchipsw_lan969x/packages/apk-aarch64_cortex-a53/kmod-crypto-xxhash/post-install" --script "post-upgrade:/builder/build_dir/target-aarch64_cortex-a53_musl/linux-microchipsw_lan969x/packages/apk-aarch64_cortex-a53/kmod-crypto-xxhash/post-upgrade" --script "pre-deinstall:/builder/build_dir/target-aarch64_cortex-a53_musl/linux-microchipsw_lan969x/packages/apk-aarch64_cortex-a53/kmod-crypto-xxhash/pre-deinstall" --info "depends:kernel=6.12.74~8d045d1f4b4010c040d03a4b02da92d4-r1 kmod-crypto-hash kmod-lib-xxhash" --files "/builder/build_dir/target-aarch64_cortex-a53_musl/linux-microchipsw_lan969x/packages/ipkg-aarch64_cortex-a53/kmod-crypto-xxhash" --output "/builder/bin/targets/microchipsw/lan969x/packages/kmod-crypto-xxhash-6.12.74-r1.apk"
[ -f /builder/bin/targets/microchipsw/lan969x/packages/kmod-crypto-xxhash-6.12.74-r1.apk ]
Package kmod-crypto-xxhash is missing dependencies for the following libraries:
xxhash.ko
make[2]: *** [modules/crypto.mk:1207: /builder/bin/targets/microchipsw/lan969x/packages/kmod-crypto-xxhash-6.12.74-r1.apk] Error 1
make[2]: Leaving directory '/builder/package/kernel/linux'
time: package/kernel/linux/compile#9.00#10.46#59.83
    ERROR: package/kernel/linux failed to build.
make[1]: *** [package/Makefile:188: package/kernel/linux/compile] Error 1
make[1]: Leaving directory '/builder'
make: *** [/builder/include/toplevel.mk:226: package/kernel/linux/compile] Error 2
```

Full logs: These logs are not from minimal reproduction, but from my package compilation, but error is the same:
[job-logs.txt](https://github.com/user-attachments/files/27086133/job-logs.2.txt)

### Expected behaviour

`make package/kernel/linux/compile` would complete successfully in official OpenWRT SDK image.

### Additional info

This problem occurs not only in Docker SDK, but also in pre-built sdk file from downloads.openwrt.org: 

https://downloads.openwrt.org/releases/25.12.2/targets/microchipsw/lan969x/openwrt-sdk-25.12.2-microchipsw-lan969x_gcc-14.3.0_musl.Linux-x86_64.tar.zst

### Diffconfig

```text

```

### Terms

Jose-Tejera:
scripts: account for built-in module dependencies

Fixes: #23094

Package dependency generation reported every module named by a kernel
module's `depends` metadata, even when that dependency was built into the
kernel and therefore had no corresponding kmod package.  This made SDK
package builds fail with a missing dependency for `xxhash.ko` on lan969x.

Pass `modules.builtin` to dependency generation and filter dependencies that
match built-in modules after normalizing their paths.  Add a deterministic
regression test proving that a module dependency is emitted normally and
suppressed when listed in `modules.builtin`.

Validation:

    ./scripts/test-gen-dependencies.sh

### DIFF
--- a/include/package-pack.mk
+++ b/include/package-pack.mk
@@ -276,6 +276,7 @@ ifneq ($(PKG_NAME),toolchain)
 			export \
 				READELF=$(TARGET_CROSS)readelf \
 				OBJCOPY=$(TARGET_CROSS)objcopy \
+				KERNEL_MODULES_BUILTIN="$(LINUX_DIR)/modules.builtin" \
 				XARGS="$(XARGS)"; \
 			$(SCRIPT_DIR)/gen-dependencies.sh "$$(IDIR_$(1))"; \
 		) | while read FILE; do \
--- a/scripts/gen-dependencies.sh
+++ b/scripts/gen-dependencies.sh
@@ -11,6 +11,7 @@ READELF="${READELF:-readelf}"
 OBJCOPY="${OBJCOPY:-objcopy}"
 TARGETS=$*
 XARGS="${XARGS:-xargs -r}"
+BUILTIN_MODULES="${KERNEL_MODULES_BUILTIN:-}"
 
 [ -z "$TARGETS" ] && {
   echo "$SELF: no directories / files specified"
@@ -25,9 +26,18 @@ find $TARGETS -type f -a -exec file {} \; | \
   sort -u
 
 tmp=$(mktemp $TMP_DIR/dep.XXXXXXXX)
+if [ -n "$BUILTIN_MODULES" ] && [ -f "$BUILTIN_MODULES" ]; then
+	builtin=$(mktemp $TMP_DIR/builtin.XXXXXXXX)
+	sed -e 's,.*/,,' "$BUILTIN_MODULES" > "$builtin"
+else
+	builtin=
+fi
 for kmod in $(find $TARGETS -type f -name \*.ko); do
 	$OBJCOPY -O binary -j .modinfo $kmod $tmp
 	sed -e 's,\x00,\n,g' $tmp | \
 		sed -ne '/^depends=.\+/ { s/^depends=//; s/,/.ko\n/g; s/$/.ko/p; q }'
-done | sort -u
-rm -f $tmp
+done | sort -u | while read dep; do
+	[ -n "$builtin" ] && grep -qxF "$dep" "$builtin" && continue
+	echo "$dep"
+done
+rm -f $tmp $builtin
--- a/scripts/test-gen-dependencies.sh
+++ b/scripts/test-gen-dependencies.sh
@@ -1,0 +1,29 @@
+#!/bin/sh
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Copyright (C) 2026 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+set -eu
+
+SCRIPT_DIR=$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)
+tmp=$(mktemp -d "${TMPDIR:-/tmp}/gen-dependencies.XXXXXX")
+trap 'rm -rf "$tmp"' EXIT
+
+printf '\0' > "$tmp/input"
+printf 'depends=xxhash\0' > "$tmp/modinfo"
+objcopy -I binary -O elf64-x86-64 -B i386 "$tmp/input" "$tmp/xxhash_generic.ko"
+objcopy --add-section .modinfo="$tmp/modinfo" \
+	--set-section-flags .modinfo=alloc,load,readonly,data \
+	"$tmp/xxhash_generic.ko"
+
+TMP_DIR="$tmp" "$SCRIPT_DIR/gen-dependencies.sh" "$tmp" > "$tmp/base"
+grep -qxF xxhash.ko "$tmp/base"
+
+printf 'kernel/lib/xxhash.ko\n' > "$tmp/modules.builtin"
+TMP_DIR="$tmp" KERNEL_MODULES_BUILTIN="$tmp/modules.builtin" \
+	"$SCRIPT_DIR/gen-dependencies.sh" "$tmp" > "$tmp/fixed"
+! grep -q . "$tmp/fixed"


include/package-pack.mk
```diff
@@ -276,6 +276,7 @@ ifneq ($(PKG_NAME),toolchain)
 			export \
 				READELF=$(TARGET_CROSS)readelf \
 				OBJCOPY=$(TARGET_CROSS)objcopy \
+				KERNEL_MODULES_BUILTIN="$(LINUX_DIR)/modules.builtin" \
```
openwrt-ai, line 279: `LINUX_DIR` is not defined for every package. It comes from [`include/kernel.mk:47`](https://github.com/openwrt/openwrt/blob/f27c655e631f44adf3a2471b183ce518077da947/include/kernel.mk#L47), and nothing in the `include/package.mk` include chain pulls `kernel.mk` in — only packages that explicitly `include $(INCLUDE_DIR)/kernel.mk` (or go through `image.mk`/`target.mk`) get it. `CheckDependencies` on the other hand runs for every non-toolchain package, so for a plain userspace package this line exports the literal `KERNEL_MODULES_BUILTIN="/modules.builtin"`.

It is harmless today — such packages ship no `.ko`, and the script guards with `[ -f "$BUILTIN_MODULES" ]` — but the bogus path means a genuinely wrong `LINUX_DIR` would also silently degrade to "no filtering" instead of being visible. Only export it when it is actually set:

```suggestion
				$(if $(LINUX_DIR),KERNEL_MODULES_BUILTIN="$(LINUX_DIR)/modules.builtin") \
```

---

scripts/gen-dependencies.sh
```diff
@@ -25,9 +26,18 @@ find $TARGETS -type f -a -exec file {} \; | \
...
 	$OBJCOPY -O binary -j .modinfo $kmod $tmp
 	sed -e 's,\x00,\n,g' $tmp | \
 		sed -ne '/^depends=.\+/ { s/^depends=//; s/,/.ko\n/g; s/$/.ko/p; q }'
-done | sort -u
-rm -f $tmp
+done | sort -u | while read dep; do
+	[ -n "$builtin" ] && grep -qxF "$dep" "$builtin" && continue
+	echo "$dep"
+done
+rm -f $tmp $builtin
```
openwrt-ai, line 43: nit: the `while read` loop forks a `grep` per dependency, and bare `read` (no `-r`) is a needless footgun. `grep` can do the whole filter in one pass, with `/dev/null` as the "no patterns" case so the unfiltered path needs no branch:

```suggestion
done | sort -u | grep -vxF -f "${builtin:-/dev/null}"
rm -f $tmp $builtin
```

`grep -v -f /dev/null` passes every line through, so this keeps the current behaviour when `KERNEL_MODULES_BUILTIN` is unset or the file is missing. `grep` exiting 1 on empty output does not leak into the script's status, since `rm -f` is the last command either way.

---

scripts/test-gen-dependencies.sh
```diff
@@ -0,0 +1,29 @@
...
+
+set -eu
+
+SCRIPT_DIR=$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)
+tmp=$(mktemp -d "${TMPDIR:-/tmp}/gen-dependencies.XXXXXX")
+trap 'rm -rf "$tmp"' EXIT
+
+printf '\0' > "$tmp/input"
+printf 'depends=xxhash\0' > "$tmp/modinfo"
+objcopy -I binary -O elf64-x86-64 -B i386 "$tmp/input" "$tmp/xxhash_generic.ko"
```
openwrt-ai, line 18: Hardcoding `elf64-x86-64`/`-B i386` makes the test x86-64-host-only. Distro binutils is built per target, not multi-target: on an x86-64 Ubuntu box `objcopy --info` lists only `elf64-x86-64`, `elf32-i386`, `pe*-*86*`; on an aarch64 host `elf64-x86-64` is simply not a supported output target, so this `objcopy` call exits non-zero and — with `set -eu` — the whole test fails for a reason unrelated to `gen-dependencies.sh`. OpenWrt supports aarch64 and macOS build hosts (the tools workflow even has a `Build tools with macos latest` job, where a bare `objcopy` may not exist at all).

Producing the stub with the host toolchain instead keeps it host-agnostic, e.g.:

```sh
echo 'static int unused;' | ${CC:-cc} -x c -c -o "$tmp/xxhash_generic.ko" -
```

Also, `gen-dependencies.sh` honours `$OBJCOPY`; the test invokes `objcopy` unqualified, so the two can disagree — worth using `${OBJCOPY:-objcopy}` here too.

---

scripts/test-gen-dependencies.sh
```diff
@@ -0,0 +1,29 @@
...
+	--set-section-flags .modinfo=alloc,load,readonly,data \
+	"$tmp/xxhash_generic.ko"
+
+TMP_DIR="$tmp" "$SCRIPT_DIR/gen-dependencies.sh" "$tmp" > "$tmp/base"
+grep -qxF xxhash.ko "$tmp/base"
+
+printf 'kernel/lib/xxhash.ko\n' > "$tmp/modules.builtin"
+TMP_DIR="$tmp" KERNEL_MODULES_BUILTIN="$tmp/modules.builtin" \
+	"$SCRIPT_DIR/gen-dependencies.sh" "$tmp" > "$tmp/fixed"
+! grep -q . "$tmp/fixed"
```
openwrt-ai, line 29: This assertion is too weak to catch the regression it is meant to guard. `! grep -q .` only says "the output is entirely empty", which also passes if the filter over-matches, if the `.ko` is no longer found by `find`, or if the kmod stage silently produced nothing at all — i.e. the test would still go green for a version of the script that suppresses *every* dependency.

The commit message says the test covers "emitted and suppressed dependencies", but the emitted case (line 24) is only exercised with `KERNEL_MODULES_BUILTIN` unset. Adding a second stub module with a dependency that is *not* in `modules.builtin`, and asserting it is still present in `$tmp/fixed` while `xxhash.ko` is gone, is what actually pins the filter's behaviour.

---

scripts/test-gen-dependencies.sh
openwrt-ai, line 1: nit: nothing in the tree runs this. There is no unit-test harness under `scripts/` (`deptest.sh` is a build driver, not a self-test), no `make check`-style target references it, and none of the workflows in `.github/workflows/` invoke it — the PR description runs it by hand. As added it will silently rot the first time `gen-dependencies.sh` changes shape.

Is the intent to follow up with a CI job, or would you rather drop the file and keep the manual reproduction in the commit message? Either is fine, but an unreferenced test script in `scripts/` is a maintenance liability on its own.

---

openwrt-ai (commented):
## Commit checks

- `f27c655e` "scripts: account for built-in module dependencies" — the subject scopes the change to `scripts:`, but the commit also changes `include/package-pack.mk`. Recent in-tree commits touching that file use the `build:` prefix (`b0b4bf6e`, `e36c2946`, `54d58b05`). `build: account for built-in module dependencies` would match the convention.

The root-cause analysis holds up: on lan969x `CONFIG_XXHASH=y`, so [`KernelPackage/install`](https://github.com/openwrt/openwrt/blob/f27c655e631f44adf3a2471b183ce518077da947/include/kernel.mk#L244-L246) skips the module, `kmod-lib-xxhash` ships no `.ko`, and its `.provides` therefore never lists `xxhash.ko` for `kmod-crypto-xxhash` to match against. Note that [`Package/kernel/extra_provides`](https://github.com/openwrt/openwrt/blob/f27c655e631f44adf3a2471b183ce518077da947/package/kernel/linux/Makefile#L71-L73) already emits exactly the same `sed -e 's,.*/,,' $(LINUX_DIR)/modules.builtin` list, but kmod packages reach `kernel` via `EXTRA_DEPENDS` rather than `DEPENDS`, so it never lands in their `IDEPEND`-derived `.provides` — filtering at the consumer side is the right layer. The basename matching is also sound: `modules.builtin` entries are `kernel/<modfile>.ko` and modpost's `depends=` uses `get_basename(module->name)`, so both sides use the same dash/underscore spelling.

Remaining comments are inline — one on `LINUX_DIR` not being defined for all packages, one simplification in the filter loop, and three on the new test script.

---

Issue #23834, opened 2026-06-16
TP-Link UE300 does not get recognized on a clean sysupgrade in 25.12-SNAPSHOT


fantom-x:
### Describe the bug

This commit works fine: https://git.openwrt.org/openwrt/openwrt/commit/?h=openwrt-25.12&id=949487e0900b92a87b5f5bc5db9861ce3480db6a.

The build by imagebuilder from yesterday did not recognize UE300: the module is loaded, but the generated "/etc/config/network" only includes the "lan" section (the onboard NIC), but no WAN. I did not make a copy of "/etc/board.json", but I suspect it is missing one NIC, which causes the generated "/etc/config/network" file to not have "wan".

There is this commit that changed when initial config is generated: https://git.openwrt.org/openwrt/openwrt/commit/?h=openwrt-25.12&id=2a8b27fab9c8e8bf67ff8a0c341b06a8c68453cf 

### OpenWrt version

https://git.openwrt.org/openwrt/openwrt/commit/?h=openwrt-25.12&id=73079fa3f1dcd6b473ea4e6fecbdbe5c33fbfc16

### OpenWrt release

25.12-SNAPSHOT

### OpenWrt target/subtarget

x64

### Device

86 x64

### Image kind

Official downloaded image

### Steps to reproduce

Build an image using imagebuilder and sysupgrade without preserving config.

### Actual behaviour

Generated "/etc/config/network" is missing "wan".

### Expected behaviour

Both "lan" and "wan" are present in "/etc/config/network"

### Diffconfig

```text

```

### Terms

- [x] I am reporting an issue for OpenWrt, not an unsupported fork.

fantom-x:
@feckert @robimarko 

On a successful boot UE300 comes up  later into the boot: is it possible that `/lib/preinit/82_config_generate` executes earlier than that and UE300 is not fully operational yet at that time?

```
dmesg | grep -E "eth0|eth1|8152"
[    1.629484] e1000e 0000:00:1f.6 eth0: (PCI Express:2.5GT/s:Width x1) 6c:4b:90:28:81:66
[    1.630857] e1000e 0000:00:1f.6 eth0: Intel(R) PRO/1000 Network Connection
[    1.632186] e1000e 0000:00:1f.6 eth0: MAC: 12, PHY: 12, PBA No: FFFFFF-0FF
[    5.101990] e1000e 0000:00:1f.6 eth0: NIC Link is Up 1000 Mbps Full Duplex, Flow Control: None
[...]

PR #24636, opened 2026-08-09
kernel: load r8152 USB network driver during boot

Jose-Tejera:
Issue: #23834  
Source commit: `16999bb4b1d817d8f5aca2be8affac41af29d838`

Submission status: HOLD — the issue is open but currently labeled `invalid`, and its target/version metadata was rejected by the issue bot. Reconfirm the report with an OpenWrt maintainer before submitting.

## What changed

Load the RTL8152 USB network driver during the early boot module phase by changing its autoload priority to:

```
AUTOLOAD:=$(call AutoProbe,r8152,1)
```

## Root cause and impact

Clean sysupgrades generate the initial network configuration before the late-loaded UE300 driver is available. The USB NIC is therefore absent from the generated `wan` configuration, even though it appears later once `r8152` loads. Priority-1 autoload makes the adapter visible during early configuration and failsafe boot.

## Validation

```
sh scripts/test-usb-net-rtl8152-boot.sh
```

The static regression test exits successfully only when the `usb-net-rtl8152` package uses the priority-1 autoload declaration. The commit also passes `git show --check`; no hardware test was run.

### DIFF
--- a/package/kernel/linux/modules/usb.mk
+++ b/package/kernel/linux/modules/usb.mk
@@ -1512,7 +1512,7 @@ define KernelPackage/usb-net-rtl8152
   DEPENDS:=+r8152-firmware +kmod-crypto-sha256 +kmod-mii +kmod-libphy
   KCONFIG:=CONFIG_USB_RTL8152
   FILES:=$(LINUX_DIR)/drivers/$(USBNET_DIR)/r8152.ko
-  AUTOLOAD:=$(call AutoProbe,r8152)
+  AUTOLOAD:=$(call AutoProbe,r8152,1)
   $(call AddDepends/usb)
 endef
 
--- a/scripts/test-usb-net-rtl8152-boot.sh
+++ b/scripts/test-usb-net-rtl8152-boot.sh
@@ -1,0 +1,29 @@
+#!/bin/sh
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Copyright (C) 2026 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+
+set -eu
+
+SCRIPT_DIR=$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)
+MODULES="$SCRIPT_DIR/../package/kernel/linux/modules/usb.mk"
+
+awk '
+/^define KernelPackage\/usb-net-rtl8152$/ {
+	in_block = 1
+	next
+}
+
+in_block && /^endef$/ {
+	if (!boot_autoload)
+		exit 1
+	exit 0
+}
+
+in_block && /AUTOLOAD:=\$\(call AutoProbe,r8152,1\)/ {
+	boot_autoload = 1
+}
+' "$MODULES"


package/kernel/linux/modules/usb.mk
```diff
@@ -1512,7 +1512,7 @@ define KernelPackage/usb-net-rtl8152
   DEPENDS:=+r8152-firmware +kmod-crypto-sha256 +kmod-mii +kmod-libphy
   KCONFIG:=CONFIG_USB_RTL8152
   FILES:=$(LINUX_DIR)/drivers/$(USBNET_DIR)/r8152.ko
-  AUTOLOAD:=$(call AutoProbe,r8152)
+  AUTOLOAD:=$(call AutoProbe,r8152,1)
```
openwrt-ai, line 1515: The second argument of `AutoProbe` is the **boot flag**, not a priority — the PR body and commit message both call this "priority-1 autoload", which is not what this does.

[`AutoProbe` in include/kernel.mk:285-289](https://github.com/openwrt/openwrt/blob/f2edc483fde094f26fce504feb94fa8360d54256/include/kernel.mk#L285-L289) forwards to [`AutoLoad(priority, list, boot)`](https://github.com/openwrt/openwrt/blob/f2edc483fde094f26fce504feb94fa8360d54256/include/kernel.mk#L278-L283) with an *empty* priority, so the priority stays `0`; the `1` lands in the boot field. The observable effect is in [`ModuleAutoLoad`](https://github.com/openwrt/openwrt/blob/f2edc483fde094f26fce504feb94fa8360d54256/include/kernel.mk#L187-L189): `/etc/modules.d/usb-net-rtl8152` gets symlinked into `/etc/modules-boot.d/`, which [initd runs before preinit](https://github.com/openwrt/procd/blob/60fdbf00e924fff0a510c4ec31c520d35e69bfe1/initd/init.c#L140) (procd `initd/init.c:140`). Please reword the description to match.

Two substantive concerns with the change itself:

1. **All ~50 other `usb-net-*` kmods stay on plain `AutoProbe`** — `asix` (line 1280), `ax88179_178a` (1296), `lan78xx` (1345), `rtl8150` (1499), `cdc_ether` (1437), … In `usb.mk` the boot flag is currently reserved for the stack itself (`usbcore`, HCDs) and for `usb-storage`/`uas`, i.e. things a rootfs can live on. This would make `r8152` the only USB *peripheral* NIC driver forced resident at boot for every user of the package, plus the `r8152-firmware` blob, whether or not an adapter is attached. If boot loading is the right answer for the UE300, why is it not equally right for the AX88179 family?

2. **It is not clear this actually closes the race.** `kmodloader` returns once the modules are inserted; USB bus enumeration and the `r8152` probe that registers the netdev happen asynchronously afterwards, while [`do_config_generate`](https://github.com/openwrt/openwrt/blob/f2edc483fde094f26fce504feb94fa8360d54256/package/base-files/files/lib/preinit/82_config_generate#L11) runs from the very next phase. Having the module *inserted* early does not guarantee `ethN` exists when `board_detect`/`config_generate` run. The PR body says no hardware test was performed — has this been confirmed to fix the reported symptom on an actual UE300, and if so is the win reliable or just a timing coincidence?

---

scripts/test-usb-net-rtl8152-boot.sh
openwrt-ai, line 1: This script does not belong in the tree. `scripts/` holds build-system helpers that the build actually invokes; the only test-shaped file there today is `deptest.sh`, and there is no per-package regression-test convention. Nothing calls this file — no Makefile target, no `.github/workflows/*` step — so it will never run and will silently rot.

Beyond placement, the assertion has no regression value: it greps `usb.mk` for the exact literal `AutoProbe,r8152,1` that the other hunk of this same commit just wrote. It cannot detect a behavioural regression, only a re-edit of that one string, and it breaks on any whitespace or reformatting of the line it pins.

Please drop this file and keep the PR to the one-line `usb.mk` change.

---

scripts/test-usb-net-rtl8152-boot.sh
```diff
@@ -0,0 +1,29 @@
...
+
+in_block && /^endef$/ {
+	if (!boot_autoload)
+		exit 1
+	exit 0
+}
+
+in_block && /AUTOLOAD:=\$\(call AutoProbe,r8152,1\)/ {
+	boot_autoload = 1
+}
```
openwrt-ai, line 28: The test passes when the thing it checks has disappeared. `in_block` is only set by the `define KernelPackage/usb-net-rtl8152` rule, so if that block is ever renamed or removed, no rule fires, awk reaches EOF with no `END` action and exits `0` — a green run reported for a package that no longer exists. `set -eu` cannot catch that, since the exit status is genuinely success.

If the script is kept despite the file-level comment, it needs a fallthrough failure. Note that a bare `END { exit 1 }` will not work: in awk, `exit` runs the `END` block, so it would clobber the successful `exit 0` above. Gate it on a flag:

```suggestion
in_block && /^endef$/ {
	found = 1
	if (!boot_autoload)
		exit 1
	exit 0
}

in_block && /AUTOLOAD:=\$\(call AutoProbe,r8152,1\)/ {
	boot_autoload = 1
}

END {
	if (!found)
		exit 1
}
```

---

openwrt-ai (commented):
## Commit checks

- `f2edc48` "kernel: load r8152 USB network driver during boot" — the body says the change uses "priority-1 autoload", but `AutoProbe`'s second argument is the boot flag, not a priority (`AutoProbe(list, boot)` calls `AutoLoad` with an empty priority). The change moves the module into `/etc/modules-boot.d/`; the priority stays `0`. Same wording error in the PR body ("changing its autoload priority to", "Priority-1 autoload makes the adapter visible"). Also, the `kernel:` subject scope does not cover `scripts/test-usb-net-rtl8152-boot.sh`, which is more than half the diff — if that file stays, it wants its own commit under a `scripts:` subject.

nit: the `openwrt[bot]` formality check already noted the missing `PKG_RELEASE` bump for `package/kernel/linux`; that still applies since this changes generated package content, not just a comment.

---

PR #24637, opened 2026-08-09
wifi-scripts: preserve zero txpower setting

### LINKED ISSUE #22855: wifi-scripts: txpower 0 treated as auto in ucode mac80211.sh — router transmits at maximum power
### Describe the bug

When setting option txpower '0' in /etc/config/wireless, the router ignores the value and transmits at maximum power. The bug is in the ucode version of mac80211.sh — txpower 0 is falsy in ucode/JavaScript, causing the script to set txpower to auto instead of fixed 000. The old shell version correctly handles 0 using [ -n "$txpower" ]. The bug was introduced during migration to ucode.

### OpenWrt version

r32802-f505120278

### OpenWrt release

25.12.2

### OpenWrt target/subtarget

Multiple targets affected: mediatek/filogic, ath79/generic, ath79/generic (ipq40xx)

### Device

Multiple devices affected — see additional info

### Image kind

Official downloaded image

### Steps to reproduce

uci set wireless.radio1.txpower='0'
uci commit wireless
wifi reload
iwinfo phy1-ap0 info | grep Tx-Power

### Actual behaviour

Tx-Power: 23 dBm — maximum power despite txpower=0

### Expected behaviour

Tx-Power: 0 dBm — or minimum supported by hardware

### Additional info

Root cause:
In package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh:

Bug — 0 is falsy in ucode, jumps to auto:
if (config.txpower)
    config.txpower = 'fixed ' + config.txpower + '00';
else
    config.txpower = 'auto';

Fix — explicit null check:
if (config.txpower != null)
    config.txpower = 'fixed ' + config.txpower + '00';
else
    config.txpower = 'auto';

Note: The old shell version uses [ -n "$txpower" ] which correctly handles 0.
The bug was introduced during migration to ucode.

Confirmed active version is ucode:
cat /lib/netifd/wireless/mac80211.sh | head -5
#!/usr/bin/ucode
'use strict';
import { set_default, log } from 'wifi.common';

Proof — direct iw command works correctly:
iw phy phy1 set txpower fixed 0
iwinfo phy1-ap0 info | grep Tx-Power
Tx-Power: 0 dBm

The kernel driver accepts 0 dBm correctly — the bug is in the ucode script, not the kernel driver.

Proof — after wifi reload, txpower is absent from configuration line:
netifd: radio1 (12995): wifi-scripts: Configuring 'phy1' txantenna: 4294967295, rxantenna: 4294967295 distance: 0

txpower is absent — confirming it was set to auto instead of fixed 000.

Also note:
netifd: radio1 (4936): command failed: Not supported (-95)
This error appears because the ucode script attempts to set txpower auto on a phy that does not support it.

Tested on five devices with three different chipsets, all running OpenWrt 25.12.2:

Device                  | Chipset                    | txpower=1 | txpower=0
Cudy WR3000S v1         | MediaTek Filogic           | 1 dBm     | 23 dBm
Buffalo WZR-600DHP      | Atheros AR7161             | 3 dBm     | 15 dBm
OpenWrt One             | MediaTek Filogic           | 1 dBm     | 23 dBm
TP-Link Archer C7 v2    | Qualcomm Atheros QCA9558   | 5 dBm     | 25 dBm
GL.iNet GL-MT6000       | MediaTek Filogic           | 1 dBm     | 23 dBm

All devices show identical behavior — this is not chipset specific.

Physical verification:
At txpower=0: -51 dBm measured by nearby router
At txpower=1: -72 dBm measured by nearby router

The 21 dBm difference confirms the router is transmitting at maximum power when txpower=0 is set.

UCI stores the value correctly:
uci show wireless.radio1
wireless.radio1.txpower='0'

UCI correctly stores and returns the value — the bug is exclusively in the ucode script.

Related issue in openwrt/luci: https://github.com/openwrt/luci/issues/8177

### Diffconfig

```text
Not applicable — official image
```

### Terms

Jose-Tejera:
wifi-scripts: preserve zero txpower setting

Fixes: #22855

The ucode mac80211 wireless script treated an explicitly configured
`txpower=0` as false and replaced it with `auto`.  On affected radios this
allowed the driver to use maximum transmit power instead of the requested
fixed value.

Use an explicit null check so zero remains a valid configured value while an
omitted value still selects automatic power.  Add a deterministic ucode
regression test covering zero, non-zero, and omitted values.

Validation:

    ./scripts/test-wifi-txpower.sh

### DIFF
--- a/scripts/test-wifi-txpower.sh
+++ b/scripts/test-wifi-txpower.sh
@@ -1,0 +1,81 @@
+#!/bin/sh
+# SPDX-License-Identifier: GPL-2.0-only
+#
+# Copyright (C) 2026 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+
+set -eu
+
+SCRIPT_DIR=$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)
+SCRIPT="$SCRIPT_DIR/../package/network/config/wifi-scripts"
+SCRIPT="$SCRIPT/files-ucode/lib/netifd/wireless/mac80211.sh"
+
+node - "$SCRIPT" <<'NODE'
+const fs = require('fs');
+
+const source = fs.readFileSync(process.argv[2], 'utf8');
+const start = source.indexOf('function setup_phy(');
+if (start < 0)
+	throw new Error('setup_phy() not found');
+
+const body = source.indexOf('{', start);
+let depth = 0;
+let quote = null;
+let end = -1;
+
+for (let i = body; i < source.length; i++) {
+	const ch = source[i];
+
+	if (quote) {
+		if (ch === '\\')
+			i++;
+		else if (ch === quote)
+			quote = null;
+		continue;
+	}
+
+	if (ch === "'" || ch === '"' || ch === '`') {
+		quote = ch;
+		continue;
+	}
+
+	if (ch === '{')
+		depth++;
+	else if (ch === '}' && --depth === 0) {
+		end = i + 1;
+		break;
+	}
+}
+
+if (end < 0)
+	throw new Error('unterminated setup_phy()');
+
+const netifd = { set_data() {} };
+const get_channel_frequency = () => null;
+const set_default = (object, key, value) => {
+	if (object[key] == null)
+		object[key] = value;
+};
+const log = () => {};
+const commands = [];
+const system = command => commands.push(command);
+const setup_phy = eval(`(${source.slice(start, end)})`);
+
+function check(txpower, expected) {
+	commands.length = 0;
+	const config = { txantenna: 1, rxantenna: 1, distance: 0 };
+	if (txpower !== undefined)
+		config.txpower = txpower;
+
+	setup_phy('phy-test', config, { txantenna: 1, rxantenna: 1 });
+	const command = commands.find(value => value.startsWith('iw phy phy-test set txpower '));
+	if (command !== `iw phy phy-test set txpower ${expected}`)
+		throw new Error(`txpower ${txpower}: got ${command}, expected ${expected}`);
+}
+
+check(0, 'fixed 000');
+check(1, 'fixed 100');
+check(undefined, 'auto');
+NODE
--- a/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh
+++ b/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh
@@ -95,7 +95,7 @@ function setup_phy(phy, config, data) {
 		rxantenna: config.rxantenna
 	});
 
-	if (config.txpower)
+	if (config.txpower != null)
 		config.txpower = 'fixed ' + config.txpower + '00';
 	else
 		config.txpower = 'auto';


scripts/test-wifi-txpower.sh
```diff
@@ -0,0 +1,81 @@
...
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+
+set -eu
+
+SCRIPT_DIR=$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)
+SCRIPT="$SCRIPT_DIR/../package/network/config/wifi-scripts"
+SCRIPT="$SCRIPT/files-ucode/lib/netifd/wireless/mac80211.sh"
+
+node - "$SCRIPT" <<'NODE'
```
openwrt-ai, line 15: This makes the test depend on Node.js, which is not an OpenWrt build prerequisite — there is no node/nodejs check anywhere in [`include/prereq-build.mk`](https://github.com/openwrt/openwrt/blob/f23db4a0f7e30c4e90c6b348f84b133e7ec6a833/include/prereq-build.mk), so on a stock build host this script just dies with `node: not found`. ucode, by contrast, is already built as a host tool ([`$(eval $(call HostBuild))` in package/utils/ucode/Makefile:202](https://github.com/openwrt/openwrt/blob/f23db4a0f7e30c4e90c6b348f84b133e7ec6a833/package/utils/ucode/Makefile#L202)).

Beyond availability, running ucode source through a JS engine does not test the thing this PR changes. ucode has no `undefined`: a missing property reads as `null`, and `!=`/`==` short-circuit to a plain identity check when either operand is null ([`ucv_compare()` in types.c](https://github.com/jow-/ucode/blob/73ffef75c653dc901c69678a27c67842d291e79e/types.c#L2236-L2239)). The harness instead exercises JS's `undefined != null`, which is a different rule that happens to give the same answer here. Please drive the real script with host `ucode` so the assertions reflect ucode semantics.

---

scripts/test-wifi-txpower.sh
```diff
@@ -0,0 +1,81 @@
...
+const netifd = { set_data() {} };
+const get_channel_frequency = () => null;
+const set_default = (object, key, value) => {
+	if (object[key] == null)
+		object[key] = value;
+};
+const log = () => {};
+const commands = [];
+const system = command => commands.push(command);
+const setup_phy = eval(`(${source.slice(start, end)})`);
```
openwrt-ai, line 64: Extracting `setup_phy()` by hand-rolled brace matching and then `eval`-ing it is fragile in a way that will bite silently. The scanner tracks string quotes but not comments, so a `//` or `/* */` comment inside `setup_phy()` containing `{`, `}`, `'` or `` ` `` desynchronises `depth`/`quote` and the extracted slice becomes garbage — the failure mode is a confusing `SyntaxError` from `eval`, not a clear "test harness broke" message. It also silently depends on `setup_phy()` never using anything that is valid ucode but not valid JS (e.g. `?:` on non-JS types, `delete` on array indices, ucode-only builtins).

A harness that imports/executes the actual script under `ucode` with stubbed `system`/`log` avoids both the parsing and the language-mismatch problem. If the file has to keep being source-scraped, at least fail loudly when the extracted text does not parse.

---

scripts/test-wifi-txpower.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/sh
```
openwrt-ai, line 1: nit: `scripts/` in this tree holds build-system helpers that the build actually invokes (image/header generators, `checkpatch.pl`, `dl_*`, `kconfig*`, …) — there is no unit-test harness there and nothing in CI will run `test-wifi-txpower.sh`, so a per-package one-off test dropped here is likely to bitrot unnoticed. Is there a reason not to keep it next to the code it tests (e.g. under `package/network/config/wifi-scripts/`), or to drop it and describe the manual verification in the commit message instead?

---

openwrt-ai (commented):
## Commit checks

- `f23db4a` "wifi-scripts: preserve zero txpower setting" — the message describes only the one-line ucode change ("Preserve an explicit zero txpower value when generating the mac80211 iw command"), but the commit also adds an 81-line test harness at `scripts/test-wifi-txpower.sh`. The `wifi-scripts:` subject scope does not cover `scripts/`, and the added file is not mentioned in the body at all. Either describe it in the message and split it into a separately-scoped commit, or drop it.
- `f23db4a` — the PR body carries `Fixes: #22855`, but the commit message has no reference to the issue. The trailer belongs in the commit message so it survives into the tree.

---

The one-line change in `mac80211.sh` itself looks correct to me: `!= null` matches the existing convention in the same file ([`config_add()`](``https://github.com/openwrt/openwrt/blob/f23db4a0f7e30c4e90c6b348f84b133e7ec6a833/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh#L155-L158)`` uses the same guard), and it brings the ucode path in line with the shell implementation, which already treats a configured `0` as a fixed value via [`[ -n "$txpower" ]`](``https://github.com/openwrt/openwrt/blob/f23db4a0f7e30c4e90c6b348f84b133e7ec6a833/package/network/config/wifi-scripts/files/lib/netifd/wireless/mac80211.sh#L1255-L1258).`` Since `txpower` has no `default` in `wireless.wifi-device.json`, an unset option stays absent and still resolves to `auto`. My comments are all on the new test script.

---

Issue #24341, opened 2026-07-21
package-metadata.pl emits PACKAGE_provider<PACKAGE_consumer expressions that newer Kconfig versions interpret as recursive self-dependencies


lianghu:
### Describe the bug

package-metadata.pl emits PACKAGE_provider<PACKAGE_consumer expressions that newer Kconfig versions interpret as recursive self-dependencies when virtual packages are used via DEPENDS:=+foo and foo has multiple PROVIDES implementations.

Collecting package info: done
tmp/.config-package.in:99809:error: recursive dependency detected!
tmp/.config-package.in:99809:   symbol PACKAGE_cache-domains-openssl depends on PACKAGE_cache-domains-openssl
For a resolution refer to Documentation/kbuild/kconfig-language.rst
subsection "Kconfig recursive dependency limitations"

tmp/.config-package.in:99824:error: recursive dependency detected!
tmp/.config-package.in:99824:   symbol PACKAGE_cache-domains-wolfssl depends on PACKAGE_cache-domains-wolfssl
For a resolution refer to Documentation/kbuild/kconfig-language.rst
subsection "Kconfig recursive dependency limitations"

tmp/.config-package.in:99794:error: recursive dependency detected!
tmp/.config-package.in:99794:   symbol PACKAGE_cache-domains-mbedtls depends on PACKAGE_cache-domains-mbedtls
For a resolution refer to Documentation/kbuild/kconfig-language.rst
subsection "Kconfig recursive dependency limitations"

### OpenWrt version

master

### OpenWrt release

25.12, 24.10

### OpenWrt target/subtarget

all

### Device

all

### Image kind

Official downloaded image

### Steps to reproduce

make  distclean
./scripts/feeds update -a
./scripts/feeds install -a
make defconfig

### Diffconfig

```text

```

### Terms

- [x] I am reporting an issue for OpenWrt, not an unsupported fork.

lianghu:
comment out the logic can solve the recursive self-dependencies

				if (@vdeps > 1) {
			#		$condition = ($condition ? "$condition && " : '') . join("&&", map { "PACKAGE_$_<PACKAGE_$pkgname" } @vdeps);
				} elsif (@vdeps > 0) {
			#		$condition = ($condition ? "$condition && " : '') . "PACKAGE_${vdeps[0]}<PACKAGE_$pkgname";
				}

PR #24638, opened 2026-08-09
build: avoid recursive virtual package dependencies

Jose-Tejera:
Issue: #24341  
Source commit: `2112c1b294d9bfd669214301dca622112c2e9b7c`

Submission status: HOLD — the issue is open but currently labeled `invalid`. Reconfirm the report and proposed Kconfig behavior with a maintainer before submitting.

## What changed

Remove the consuming package itself from the virtual-provider dependency list before emitting Kconfig `select` conditions.

## Root cause and impact

When a package provides a virtual name and also depends on that virtual name, `package-metadata.pl` included the package itself in `@vdeps`. The generated expression then contained a recursive dependency such as `PACKAGE_foo-self<PACKAGE_foo-self`, causing Kconfig to reject package metadata generation. Filtering the consumer preserves valid provider constraints while removing only the self-edge.

## Validation

```
sh scripts/package-metadata-test.sh
```

The test creates a minimal provider/consumer metadata set, verifies that no recursive self-dependency is emitted, and checks that the remaining provider condition is preserved. It prints `package-metadata.pl virtual package dependency test: ok` and exits successfully; the commit passes `git show --check`.

### DIFF
--- a/scripts/package-metadata-test.sh
+++ b/scripts/package-metadata-test.sh
@@ -1,0 +1,58 @@
+#!/bin/sh
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# Regression test for virtual package dependencies in package-metadata.pl.
+#
+# Copyright (C) 2026 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+set -eu
+
+SCRIPTDIR=$(dirname "$0")
+BASEDIR=$(cd "$SCRIPTDIR/.." && pwd)
+TMPDIR=$(mktemp -d "${TMPDIR:-/tmp}/package-metadata.XXXXXX")
+trap 'rm -rf "$TMPDIR"' EXIT HUP INT TERM
+
+cat > "$TMPDIR/packageinfo" <<'EOF'
+Source-Makefile: package/test/provider/Makefile
+
+Package: foo-default
+Depends:
+Provides: foo
+Default-Variant: 1
+Category: Test
+Title: Default foo provider
+Description: default
+@@
+
+Package: foo-self
+Depends: +foo
+Provides: foo
+Category: Test
+Title: Self-dependent foo provider
+Description: self
+@@
+
+Package: foo-other
+Depends:
+Provides: foo
+Category: Test
+Title: Other foo provider
+Description: other
+@@
+EOF
+
+"$BASEDIR/scripts/package-metadata.pl" config "$TMPDIR/packageinfo" > "$TMPDIR/config.in"
+
+if grep -Fq 'PACKAGE_foo-self<PACKAGE_foo-self' "$TMPDIR/config.in"; then
+	echo "package-metadata.pl emitted a recursive self-dependency" >&2
+	exit 1
+fi
+
+grep -Fq 'select PACKAGE_foo-default' "$TMPDIR/config.in"
+grep -Fq 'select PACKAGE_foo-default if PACKAGE_foo-other<PACKAGE_foo-self' "$TMPDIR/config.in"
+
+echo "package-metadata.pl virtual package dependency test: ok"
--- a/scripts/package-metadata.pl
+++ b/scripts/package-metadata.pl
@@ -194,6 +194,7 @@ sub mconf_depends {
 
 				$depend = shift @vdeps;
 
+				@vdeps = grep { $_ ne $pkgname } @vdeps;
 				if (@vdeps > 1) {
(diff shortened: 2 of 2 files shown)


scripts/package-metadata.pl
```diff
@@ -194,6 +194,7 @@ sub mconf_depends {
 
 				$depend = shift @vdeps;
 
+				@vdeps = grep { $_ ne $pkgname } @vdeps;
```
openwrt-ai, line 197: The filter runs *after* `shift`, so it only fixes the case where the consuming package is not the head of `@vdeps`. When the package is first in the list, `shift` hands it to `$depend` and the self-reference moves from the condition into the `select` itself.

The package is put at the head whenever it is the default variant (see [the `unshift` in the provider loop at package-metadata.pl:186-191](https://github.com/openwrt/openwrt/blob/bc0d1aef67314e75a9e1759c8a7be1df17287b7c/scripts/package-metadata.pl#L186-L191)), or simply when it is the first provider parsed and no default variant exists.

Running this branch with a fixture where `foo-self` carries `Default-Variant: 1`:

```
config PACKAGE_foo-self
	select PACKAGE_foo-self if PACKAGE_foo-other<PACKAGE_foo-self
```

and where `foo-self` is the sole provider of `foo`:

```
config PACKAGE_foo-self
	select PACKAGE_foo-self
```

Both are the same recursive-dependency class this commit is meant to remove. Filtering before the `shift` handles all three orderings, and the empty case (package is its own only provider) is genuinely satisfied by itself, so the dependency can be dropped:

```suggestion
				@vdeps = grep { $_ ne $pkgname } @vdeps;
				next unless @vdeps;

				$depend = shift @vdeps;
```

With that, the three fixtures above produce `select PACKAGE_foo-other`, `select PACKAGE_foo-other`, and no `select` respectively, and the fixture in `scripts/package-metadata-test.sh` is unchanged.

---

scripts/package-metadata-test.sh
```diff
@@ -0,0 +1,58 @@
...
+Description: default
+@@
+
+Package: foo-self
+Depends: +foo
+Provides: foo
+Category: Test
+Title: Self-dependent foo provider
+Description: self
+@@
```
openwrt-ai, line 37: This fixture only exercises the one ordering the current fix happens to handle: `foo-self` is neither the default variant nor the first provider, so it never reaches `$depend` via `shift`. Adding a second fixture where the self-dependent provider carries `Default-Variant: 1` (and a third where it is the only provider of `foo`) would cover the two orderings flagged in `scripts/package-metadata.pl` — both still emit a self-reference with this patch applied.

---

scripts/package-metadata-test.sh
```diff
@@ -0,0 +1,58 @@
...
+
+"$BASEDIR/scripts/package-metadata.pl" config "$TMPDIR/packageinfo" > "$TMPDIR/config.in"
+
+if grep -Fq 'PACKAGE_foo-self<PACKAGE_foo-self' "$TMPDIR/config.in"; then
+	echo "package-metadata.pl emitted a recursive self-dependency" >&2
+	exit 1
+fi
+
+grep -Fq 'select PACKAGE_foo-default' "$TMPDIR/config.in"
+grep -Fq 'select PACKAGE_foo-default if PACKAGE_foo-other<PACKAGE_foo-self' "$TMPDIR/config.in"
```
openwrt-ai, line 56: These two assertions rely on `set -e` to fail the script, so a regression aborts with exit 1 and no output at all — inconsistent with the explicit diagnostic in the `if` block just above. The first grep is also fully subsumed by the second (its pattern is a prefix of it).

```suggestion
if ! grep -Fq 'select PACKAGE_foo-default if PACKAGE_foo-other<PACKAGE_foo-self' "$TMPDIR/config.in"; then
	echo "package-metadata.pl did not emit the expected provider condition" >&2
	exit 1
fi
```

---

scripts/package-metadata-test.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/sh
```
openwrt-ai, line 1: nit: nothing in the tree invokes this script — `grep -rn package-metadata-test .` returns no hits, there is no test target in `include/toplevel.mk`, and no workflow under `.github/workflows/` runs it. `scripts/` currently has no unit-test harness at all, so as added the test can only be run by hand and will silently rot. Is the intent to wire it into a make target or a CI job, or to keep it as a manual reproducer? If the latter, it may be worth saying so in the commit message.

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the diff; findings are inline.

Main point: the new `grep` runs after `shift @vdeps`, so it does not cover the orderings where the self-providing package ends up as the selected provider rather than as a condition operand — verified by running `scripts/package-metadata.pl config` against modified fixtures.

---

Issue #24370, opened 2026-07-22
wifi-scripts: multi-radio phys sometimes assign wrong radio


andrewjlamarche:
### Describe the bug

This follows https://github.com/openwrt/openwrt/commit/2c642576270de2606de6d0890fb627066c3a1595. When setting up an ath12k device that exposes 3 radios via a single phy, radio index is also assigned in a non-deterministic manner. For example, the first boot may assign:

- radio0 -> 2.4g
- radio1 -> 5g
- radio2 -> 6g

While the next boot may have

- radio0 -> 6g
- radio1 -> 2.4g
- radio2 -> 5g

Or really any combination. `/etc/config/wireless` is not aware of this change and the affected radios will fail to come up. Example 02_network:

```
#
# Copyright (c) 2015 The Linux Foundation. All rights reserved.
# Copyright (c) 2011-2015 OpenWrt.org
#

. /lib/functions/uci-defaults.sh
. /lib/functions/system.sh

ipq95xx_setup_interfaces()
{
	local board="$1"

	case "$board" in
	8devices,kiwi-dvk)
		ucidef_set_interfaces_lan_wan "lan1 lan2" "wan"
		;;
	askey,sbe1v1k)
		ucidef_set_interfaces_lan_wan "lan1 lan2 lan3" "wan"
		;;
	qcom,ipq9574-ap-al02-c7)
		ucidef_set_interfaces_lan_wan "lan1 lan2 lan3 lan4 lan5" "wan"
		;;
	*)
		echo "Unsupported hardware. Network interfaces not initialized"
		;;
	esac
}

DEV_PATH_PCIE0=soc@0/10000000.pcie/pci0001:00/0001:00:00.0/0001:01:00.0
DEV_PATH_PCIE1=soc@0/18000000.pcie/pci0003:00/0003:00:00.0/0003:01:00.0
DEV_PATH_PCIE2=soc@0/20000000.pcie/pci0002:00/0002:00:00.0/0002:01:00.0

ipq95xx_setup_wlan()
{
	local board="$1"

	case "$board" in
	askey,sbe1v1k)
		ucidef_add_wlan "$DEV_PATH_PCIE0" "$DEV_PATH_PCIE1" "$DEV_PATH_PCIE2"
		;;
	esac
}

board_config_update
board=$(board_name)
ipq95xx_setup_interfaces $board
ipq95xx_setup_wlan $board
board_config_flush

exit 0
```

### OpenWrt version

[...]

PR #23840, opened 2026-06-17
wifi-scripts: add detect_band option for multi-radio phy

hurrian:
Add a detect_band option that binds a radio configuration to a phy + band pair rather than to a path or radio index.

Some devices (e.g. ath12k-wsi) enumerate their radios at non- deterministic PCI addresses, so neither the device path nor the radio index is stable across reboots. Matching by band keeps the per-band configuration attached to the correct radio regardless of enumeration order.

The behavior is opt-in: detection only switches to band matching when a 'wifi-config' section sets detect_band='1'. Devices seed it from a per-device uci-defaults script, e.g.:

```
. /lib/functions.sh

case "$(board_name)" in
askey,sbe1v1k)
	if ! uci -q get wireless.detect_band >/dev/null; then
		uci set wireless.detect_band=wifi-config
		uci set wireless.detect_band.detect_band='1'
		uci commit wireless
	fi
	;;
esac
```
Link: openwrt#22157
Signed-off-by: Kenneth Kasilag <kenneth@kasilag.me>

PR #24639, opened 2026-08-09
wifi-scripts: resolve multi-radio index by band

Jose-Tejera:
Issue: #24370  
Source commit: `a3b3bb572f4c32eb855fc076bf3108f1e964b904`

Submission status: UNSAFE — do not submit as a separate PR. Open PR [#21586](https://github.com/openwrt/openwrt/pull/21586) explicitly links this issue and is the active device-support change for the affected Askey SBE1V1K. Coordinate the radio fix with that PR’s author/maintainers first. The issue is also currently labeled `invalid`.

## What changed

Add a `wiphy_radio()` resolver to the ucode Wi-Fi helpers and call it from mac80211 setup to reconcile the configured radio index with the configured band.

## Root cause and impact

On single-phy multi-radio ath12k devices, PCIe/radio enumeration order can change between boots. Numeric `radioN` indices are therefore unstable, while supported frequency ranges identify the intended 2.4, 5, or 6 GHz radio. Resolving by band preserves the correct wireless configuration and prevents radios from being assigned to the wrong band.

## Validation

```
sh tests/wifi-radio-mapping.sh
```

The mocked regression models two boot orders and verifies that 5, 2.4, and 6 GHz bands resolve to their correct changed indices. It prints `radio mapping follows band identity across changed phy order` and exits successfully; the commit passes `git show --check`. No target hardware test was run.

### DIFF
--- a/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh
+++ b/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh
@@ -2,7 +2,7 @@
 
 'use strict';
 
-import { set_default, log } from 'wifi.common';
+import { set_default, log, wiphy_info, wiphy_radio } from 'wifi.common';
 import { validate, dump_options } from 'wifi.validate';
 import * as supplicant from 'wifi.supplicant';
 import * as hostapd from 'wifi.hostapd';
@@ -202,6 +202,9 @@ function setup() {
 		}
 	}
 	delete config.hwmode;
+
+	let phy_info = wiphy_info(data.phy);
+	data.config.radio = wiphy_radio(phy_info, data.config.band, data.config.radio);
 
 	validate('device', config);
 	setup_phy(data.phy, data.config, data.data);
--- a/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/common.uc
+++ b/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/common.uc
@@ -28,6 +28,43 @@ export function wiphy_band(info, band) {
 	return info.wiphy_bands[band_idx];
 };
 
+function freq_in_range(freq_ranges, freq) {
+	if (!freq_ranges)
+		return true;
+
+	freq *= 1000;
+	for (let range in freq_ranges)
+		if (freq >= range.start && freq <= range.end)
+			return true;
+}
+
+function radio_supports_band(radio, band) {
+	if (!radio.freq_ranges || !band?.freqs)
+		return false;
+
+	for (let freq in band.freqs)
+		if (!freq.disabled && freq_in_range(radio.freq_ranges, freq.freq))
+			return true;
+
+	return false;
+}
+
+export function wiphy_radio(info, band, current) {
+	let band_info = wiphy_band(info, band);
+	if (!info?.radios || !band_info)
+		return current;
+
+	for (let radio in info.radios)
+		if (radio.index == current && radio_supports_band(radio, band_info))
+			return current;
+
+	for (let radio in info.radios)
+		if (radio_supports_band(radio, band_info))
+			return radio.index;
+
+	return current;
+};
+
 export function log(msg) {
 	printf(`wifi-scripts: ${msg}\n`);
 };
--- a/tests/wifi-radio-mapping.sh
+++ b/tests/wifi-radio-mapping.sh
@@ -1,0 +1,64 @@
+#!/bin/sh
+# SPDX-License-Identifier: GPL-2.0-only
+
+set -eu
+
+TOPDIR=$(CDPATH= cd -- "$(dirname "$0")/.." && pwd)
+COMMON="$TOPDIR/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/common.uc"
+MAC80211="$TOPDIR/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh"
+
+grep -q '^export function wiphy_radio' "$COMMON" || {
+	echo "wiphy radio resolver is missing" >&2
+	exit 1
+}
+grep -q 'wiphy_radio(phy_info, data.config.band, data.config.radio)' "$MAC80211" || {
+	echo "mac80211 does not resolve radio by band" >&2
+	exit 1
+}
+
+# Model the nl80211 radio list returned by a mocked wiphy. The numeric index
+# for the 5 GHz radio changes from 1 to 2 between boots, while the frequency
+# range remains the stable identity.
+radio_for_band() {
+	band="$1"
+	current="$2"
+	radio_list="$3"
+
+	while IFS=: read -r index range; do
+		[ "$index" = "$current" ] || continue
+		[ "$range" = "$band" ] && {
+			echo "$current"
+			return
+		}
+	done < "$radio_list"
+
+	while IFS=: read -r index range; do
+		[ "$range" = "$band" ] && {
+			echo "$index"
+			return
+		}
+	done < "$radio_list"
+
+	echo "$current"
+}
+
+mock_sysfs=$(mktemp -d)
+trap 'rm -rf "$mock_sysfs"' EXIT
+
+cat > "$mock_sysfs/boot-a" <<EOF
+0:2g
+1:5g
+2:6g
+EOF
+cat > "$mock_sysfs/boot-b" <<EOF
+0:6g
+1:2g
+2:5g
+EOF
+
+[ "$(radio_for_band 5g 1 "$mock_sysfs/boot-a")" = 1 ]
+[ "$(radio_for_band 5g 1 "$mock_sysfs/boot-b")" = 2 ]
+[ "$(radio_for_band 2g 0 "$mock_sysfs/boot-b")" = 1 ]
+[ "$(radio_for_band 6g 2 "$mock_sysfs/boot-b")" = 0 ]
+
+echo "radio mapping follows band identity across changed phy order"


package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh
```diff
@@ -203,6 +203,9 @@ function setup() {
 	}
 	delete config.hwmode;
 
+	let phy_info = wiphy_info(data.phy);
+	data.config.radio = wiphy_radio(phy_info, data.config.band, data.config.radio);
```
openwrt-ai, line 207: The resolution happens too late to have the intended effect: [`data.phy_suffix` / `data.vif_phy_suffix` / `data.ifname_prefix` are already computed from `data.config.radio` at lines 174-178](``https://github.com/openwrt/openwrt/blob/a3b3bb572f4c32eb855fc076bf3108f1e964b904/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh#L174-L178),`` i.e. ~30 lines above this hunk.

Those suffixes are what actually selects the radio downstream — `data.phy + data.phy_suffix` is passed to `wdev.uc set_config` and to `iface.prepare()`, and `data.vif_phy_suffix` names the hostapd config file. So after this change the stale index still drives naming, while `setup_phy()` (called right below) stores the *new* index via `netifd.set_data({ radio: config.radio })`. Teardown then calls `reset_config(data.data.phy, data.data.radio)` with the resolved index against files created under the unresolved one, so a stale hostapd/wpa_supplicant config can be left behind.

Move the `wiphy_info()`/`wiphy_radio()` pair to just above line 174, after `data.phy` is known but before the suffixes are derived. Note `config.band` is defaulted from `hwmode` further down, so that block needs to move up with it (or the band fallback needs to run first).

Minor: the surrounding code in this function uses the local `config` alias (`let config = data.config;`); these two lines use `data.config` for the same object.

---

package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/common.uc
```diff
@@ -28,6 +28,43 @@ export function wiphy_band(info, band) {
...
+		if (!freq.disabled && freq_in_range(radio.freq_ranges, freq.freq))
+			return true;
+
+	return false;
+}
+
+export function wiphy_radio(info, band, current) {
+	let band_info = wiphy_band(info, band);
+	if (!info?.radios || !band_info)
+		return current;
```
openwrt-ai, line 55: This silently invents a radio index for whole-phy configurations. `radio` is optional and "no radio" is encoded as `null` or `-1` — see [`phy_suffix()` at mac80211.sh:28-32](``https://github.com/openwrt/openwrt/blob/a3b3bb572f4c32eb855fc076bf3108f1e964b904/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh#L28-L32)`` and [`set_default radio -1` in the shell variant](``https://github.com/openwrt/openwrt/blob/a3b3bb572f4c32eb855fc076bf3108f1e964b904/package/network/config/wifi-scripts/files/lib/netifd/wireless/mac80211.sh#L1144-L1147).``

On a multi-radio wiphy where the wifi-device section has no `radio` option, `info.radios` is present so this guard passes, the first loop never matches (`radio.index == null` is false), and the fallback loop returns a concrete `radio.index`. `phy_suffix` then flips from `""` to `":N"`, converting a whole-phy device into a single-radio one. Same for an explicit `radio -1`.

Keep the sentinel values untouched:

```suggestion
	if (!info?.radios || !band_info || current == null || current < 0)
		return current;
```

---

package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/common.uc
```diff
@@ -28,6 +28,43 @@ export function wiphy_band(info, band) {
...
 
+function freq_in_range(freq_ranges, freq) {
+	if (!freq_ranges)
+		return true;
+
+	freq *= 1000;
+	for (let range in freq_ranges)
+		if (freq >= range.start && freq <= range.end)
+			return true;
+}
```
openwrt-ai, line 39: This is a byte-for-byte duplicate of [`freq_in_range()` at supplicant.uc:254-262](``https://github.com/openwrt/openwrt/blob/a3b3bb572f4c32eb855fc076bf3108f1e964b904/package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/supplicant.uc#L254-L262)`` (and a third copy exists as [`freq_range_match()` in wifi-detect.uc:83-90](``https://github.com/openwrt/openwrt/blob/a3b3bb572f4c32eb855fc076bf3108f1e964b904/package/network/config/wifi-scripts/files/usr/share/hostap/wifi-detect.uc#L83-L90)).`` Since `common.uc` is already the shared module that `supplicant.uc` imports from, export this one and drop the copy in `supplicant.uc` rather than adding a second definition.

Also, the `!freq_ranges -> return true` fallback is dead code here: the only caller, `radio_supports_band()`, already returns early when `radio.freq_ranges` is falsy. It is meaningful in the `supplicant.uc` copy (where `ranges` stays undefined for a non-matching radio), so it should survive the merge, but it is worth noting the two call sites disagree about what a missing range list means.

---

package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/common.uc
```diff
@@ -28,6 +28,43 @@ export function wiphy_band(info, band) {
...
+	if (!info?.radios || !band_info)
+		return current;
+
+	for (let radio in info.radios)
+		if (radio.index == current && radio_supports_band(radio, band_info))
+			return current;
+
+	for (let radio in info.radios)
+		if (radio_supports_band(radio, band_info))
+			return radio.index;
```
openwrt-ai, line 63: The fallback picks the *first* radio with any enabled channel in the band, but an nl80211 band is not a unique radio identity. A wiphy can expose two radios that both fall inside the same band — e.g. a 5 GHz low/high split (5170-5330 and 5490-5730), which is exactly the kind of layout `freq_ranges` exists to describe. In that case `radio_supports_band()` is true for both and this loop resolves both wifi-device sections onto the same index, so one radio ends up unconfigured and the other double-configured.

The commit message says frequency ranges are the stable identity, but only band membership is actually compared here. If the reboot-reordering premise holds, matching against the radio's stored `freq_ranges` (start/end pair) rather than "has some channel in this band" would be both a real identity and a strict improvement. Was the split-band case considered, or is it out of scope for the ath12k hardware in #24370?

---

package/network/config/wifi-scripts/files-ucode/usr/share/ucode/wifi/common.uc
```diff
@@ -28,6 +28,43 @@ export function wiphy_band(info, band) {
...
+		return false;
+
+	for (let freq in band.freqs)
+		if (!freq.disabled && freq_in_range(radio.freq_ranges, freq.freq))
+			return true;
+
+	return false;
+}
+
+export function wiphy_radio(info, band, current) {
```
openwrt-ai, line 52: nit: this only lands in the `files-ucode/` tree, which is [installed conditionally on `CONFIG_WIFI_SCRIPTS_UCODE`](https://github.com/openwrt/openwrt/blob/a3b3bb572f4c32eb855fc076bf3108f1e964b904/package/network/config/wifi-scripts/Makefile#L45-L50). The default shell implementation keeps the same unresolved numeric index — [`files/lib/netifd/wireless/mac80211.sh:1144-1147`](``https://github.com/openwrt/openwrt/blob/a3b3bb572f4c32eb855fc076bf3108f1e964b904/package/network/config/wifi-scripts/files/lib/netifd/wireless/mac80211.sh#L1144-L1147)`` derives `phy_suffix`/`vif_phy_suffix` straight from the UCI `radio` value — so a device hitting the bug in #24370 with the default build is unaffected by this fix. Intentional (ucode variant only, on the assumption it becomes the default), or should the shell path be covered too?

---

tests/wifi-radio-mapping.sh
```diff
@@ -0,0 +1,64 @@
...
+
+	while IFS=: read -r index range; do
+		[ "$range" = "$band" ] && {
+			echo "$index"
+			return
+		}
+	done < "$radio_list"
+
+	echo "$current"
+}
```
openwrt-ai, line 43: This "regression test" never executes the code it claims to cover. `radio_for_band()` is a POSIX-shell reimplementation of `wiphy_radio()`'s two-pass algorithm run against a text file; `wiphy_radio()` itself is never called, so the assertions on lines 59-62 stay green no matter what the ucode does. Combined with the source-string greps on lines 10-17 — which only assert that a literal call expression is still present, and will break on any harmless refactor — the file gives no regression protection: both bugs flagged in this review (the sentinel-index case and the ordering relative to `phy_suffix`) pass it.

There is also no top-level `tests/` directory in the tree today and nothing in `.github/workflows/` or the build system would run it, so this adds an untracked entry point that will silently rot. Either exercise the real module (`ucode` is buildable for the host, so the resolver can be imported and driven with a hand-built `info` object including the `radio == null` / `radio == -1` cases), or drop the file and describe the manual verification in the commit message instead.

Minor: `mock_sysfs` holds a mocked nl80211 radio list, not a sysfs tree.

---

openwrt-ai (commented):
Reviewed 1 new commit; 6 inline comments. The commit message matches the diff.

Main concerns are that the new resolution runs after `phy_suffix`/`ifname_prefix` have already been derived from the old index, and that the guard does not preserve the `null` / `-1` "whole phy" sentinel.

---

motolav:
@Jose-Tejera nbd168 didn't want to match wireless interfaces by band that's why its not used https://github.com/openwrt/openwrt/pull/23840#issuecomment-4811879589

PR #24643, opened 2026-08-09
realtek: rtl930x: add PoE support for Hasivo F1100WP-4SX-4XGT

pktpls:
- split PoE variant off of Hasivo F1100W-4SX-4XGT
- add STC8 MCU support for Hasivo F1100W/WP-4SX-4XGT
- add PoE support for Hasivo F1100WP-4SX-4XGT

### DIFF
--- a/target/linux/realtek/dts/rtl9303_hasivo_f1100wp-4sx-4xgt-512mb.dts
+++ b/target/linux/realtek/dts/rtl9303_hasivo_f1100wp-4sx-4xgt-512mb.dts
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 /dts-v1/;
 
-#include "rtl9303_hasivo_f1100w-4sx-4xgt-common.dtsi"
+#include "rtl9303_hasivo_f1100wp-4sx-4xgt-common.dtsi"
 
 / {
 	compatible = "hasivo,f1100wp-4sx-4xgt-512mb", "realtek,rtl9303-soc";
--- a/target/linux/realtek/dts/rtl9303_hasivo_f1100wp-4sx-4xgt-512mb.dts
+++ b/target/linux/realtek/dts/rtl9303_hasivo_f1100wp-4sx-4xgt-512mb.dts
@@ -4,8 +4,8 @@
 #include "rtl9303_hasivo_f1100w-4sx-4xgt-common.dtsi"
 
 / {
-	compatible = "hasivo,f1100w-4sx-4xgt-512mb", "realtek,rtl9303-soc";
-	model = "Hasivo F1100W-4SX-4XGT (512MB)";
+	compatible = "hasivo,f1100wp-4sx-4xgt-512mb", "realtek,rtl9303-soc";
+	model = "Hasivo F1100WP-4SX-4XGT (512MB)";
 
 	memory@0 {
 		reg = <0x00000000 0x10000000>, /* 256 MiB lowmem */
--- a/target/linux/realtek/dts/rtl9303_hasivo_f1100wp-4sx-4xgt-common.dtsi
+++ b/target/linux/realtek/dts/rtl9303_hasivo_f1100wp-4sx-4xgt-common.dtsi
@@ -1,0 +1,86 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/dts-v1/;
+
+#include "rtl9303_hasivo_f1100w-4sx-4xgt-common.dtsi"
+
+&i2c_stc8 {
+	pse0: ethernet-pse@d {
+		compatible = "hasivo,hs104";
+		reg = <0x0d>;
+		poll-interval-ms = <500>;
+		status = "okay";
+
+		pse-pis {
+			#address-cells = <1>;
+			#size-cells = <0>;
+
+			pse0_pi0: pse-pi@0 {
+				reg = <0>;
+				#pse-cells = <0>;
+			};
+			pse0_pi1: pse-pi@1 {
+				reg = <1>;
+				#pse-cells = <0>;
+			};
+			pse0_pi2: pse-pi@2 {
+				reg = <2>;
+				#pse-cells = <0>;
+			};
+			pse0_pi3: pse-pi@3 {
+				reg = <3>;
+				#pse-cells = <0>;
+			};
+		};
+	};
+};
+
+&stc8 {
+	led_lan5_poe: led@1,0 {
+		compatible = "register-bit-led";
+		reg = <0x01 0x00>;
+		mask = <0x08>;
+		color = <LED_COLOR_ID_ORANGE>;
+		function = "poe";
+		function-enumerator = <5>;
+		linux,default-trigger = "pse-0-000d:port0:delivering";
+		default-state = "off";
+	};
+
+	led_lan6_poe: led@1,1 {
+		compatible = "register-bit-led";
+		reg = <0x01 0x01>;
+		mask = <0x04>;
+		color = <LED_COLOR_ID_ORANGE>;
+		function = "poe";
+		function-enumerator = <6>;
+		linux,default-trigger = "pse-0-000d:port1:delivering";
+		default-state = "off";
+	};
+
+	led_lan7_poe: led@1,2 {
+		compatible = "register-bit-led";
+		reg = <0x01 0x02>;
+		mask = <0x02>;
+		color = <LED_COLOR_ID_ORANGE>;
+		function = "poe";
+		function-enumerator = <7>;
+		linux,default-trigger = "pse-0-000d:port2:delivering";
+		default-state = "off";
+	};
+
+	led_lan8_poe: led@1,3 {
+		compatible = "register-bit-led";
+		reg = <0x01 0x03>;
+		mask = <0x01>;
+		color = <LED_COLOR_ID_ORANGE>;
+		function = "poe";
+		function-enumerator = <8>;
+		linux,default-trigger = "pse-0-000d:port3:delivering";
+		default-state = "off";
+	};
+};
+
+&phy24 { pses = <&pse0_pi0>; };
+&phy25 { pses = <&pse0_pi1>; };
+&phy26 { pses = <&pse0_pi2>; };
+&phy27 { pses = <&pse0_pi3>; };
--- a/target/linux/realtek/dts/rtl9303_hasivo_f1100wp-4sx-4xgt.dts
+++ b/target/linux/realtek/dts/rtl9303_hasivo_f1100wp-4sx-4xgt.dts
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 /dts-v1/;
 
-#include "rtl9303_hasivo_f1100w-4sx-4xgt-common.dtsi"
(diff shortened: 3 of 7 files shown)


target/linux/realtek/dts/rtl9303_hasivo_f1100wp-4sx-4xgt.dts
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/dts-v1/;
+
+#include "rtl9303_hasivo_f1100wp-4sx-4xgt-common.dtsi"
+
+/ {
+	compatible = "hasivo,f1100wp-4sx-4xgt", "realtek,rtl930x-soc";
```
openwrt-ai: Giving the WP variant its own compatible drops it out of the two realtek base-files case statements that currently match it through `hasivo,f1100w-4sx-4xgt*`, and neither is updated in this PR:

- [`board.d/02_network:102-103`](https://github.com/openwrt/openwrt/blob/1e014b68fbb1314555699aac19071f05d6a9e465/target/linux/realtek/base-files/etc/board.d/02_network#L102-L103) — reads `ethaddr` from `u-boot-env`, substitutes a random MAC when it is empty/dummy, and sets `eth0_mac`/`lan_mac`/`lan_mac_start`. With no matching arm, F1100WP boards fall through the whole `case` and get no `eth0` MAC, no bridge MAC and no per-port MAC assignment.
- [`uci-defaults/99_fwenv-store-ethaddr:44-45`](``https://github.com/openwrt/openwrt/blob/1e014b68fbb1314555699aac19071f05d6a9e465/target/linux/realtek/base-files/etc/uci-defaults/99_fwenv-store-ethaddr#L44-L45)`` — persists that generated address into the U-Boot env so it stays stable across reboots.

Both lists need `hasivo,f1100wp-4sx-4xgt` and `hasivo,f1100wp-4sx-4xgt-512mb` (the comment block above the `99_fwenv-store-ethaddr` arm already says "F1100W-4SX-4XGT **and its variants**", and the PoE sibling `hasivo,s1300wp-8xgt-4s-plus` is in both lists).

---

target/linux/realtek/image/rtl930x.mk
```diff
@@ -47,11 +45,34 @@ define Device/hasivo_f1100w-4sx-4xgt-512mb
...
+    kmod-hasivo-mcu-sensor kmod-hasivo-mcu-wdt kmod-mfd-hasivo-stc8 \
+    kmod-phy-realtek kmod-pse-hasivo-hs104 rtl826x-firmware
+  IMAGE_SIZE := 29696k
+  $(Device/kernel-lzma)
+endef
+
+define Device/hasivo_f1100wp-4sx-4xgt
+  $(Device/hasivo_f1100wp-4sx-4xgt-common)
+endef
+TARGET_DEVICES += hasivo_f1100wp-4sx-4xgt
```
openwrt-ai, line 67: F1100WP owners are today running the `hasivo_f1100w-4sx-4xgt` image (it claimed them via `DEVICE_ALT1_*`), so their board reports `hasivo,f1100w-4sx-4xgt`. `SUPPORTED_DEVICES` defaults to the new device name only ([image.mk:551](https://github.com/openwrt/openwrt/blob/1e014b68fbb1314555699aac19071f05d6a9e465/include/image.mk#L551)), so after this split `sysupgrade` to the new WP image is rejected as unsupported and every existing install needs `-F`. Carrying the old compat string keeps the migration path — same pattern as [rtl838x.mk:396](https://github.com/openwrt/openwrt/blob/1e014b68fbb1314555699aac19071f05d6a9e465/target/linux/realtek/image/rtl838x.mk#L396).

```suggestion
define Device/hasivo_f1100wp-4sx-4xgt
  $(Device/hasivo_f1100wp-4sx-4xgt-common)
  SUPPORTED_DEVICES += hasivo,f1100w-4sx-4xgt
endef
TARGET_DEVICES += hasivo_f1100wp-4sx-4xgt
```

The same applies to `hasivo_f1100wp-4sx-4xgt-512mb` below (lines 69-74), which needs `SUPPORTED_DEVICES += hasivo,f1100w-4sx-4xgt-512mb`.

---

target/linux/realtek/dts/rtl9303_hasivo_f1100wp-4sx-4xgt-common.dtsi
```diff
@@ -0,0 +1,86 @@
...
+};
+
+&stc8 {
+	led_lan5_poe: led@1,0 {
+		compatible = "register-bit-led";
+		reg = <0x01 0x00>;
+		mask = <0x08>;
+		color = <LED_COLOR_ID_ORANGE>;
+		function = "poe";
+		function-enumerator = <1>;
```
openwrt-ai: nit: the PoE-capable ports on this board are lan5-lan8 (the RJ45 ports, per `SWITCH_PORT_LED(24..27, 5..8, ...)` in the common dtsi), and the node labels agree — `led_lan5_poe` … `led_lan8_poe`. But the enumerators are 1-4, so the LEDs surface as `orange:poe-1` … `orange:poe-4`, which does not line up with the port they sit next to. The sibling Hasivo boards keep `function-enumerator` equal to the LAN port number, including for the second HS104 covering lan5-8: [rtl9303_hasivo_s1100wp-8xgt-se.dts:142](https://github.com/openwrt/openwrt/blob/1e014b68fbb1314555699aac19071f05d6a9e465/target/linux/realtek/dts/rtl9303_hasivo_s1100wp-8xgt-se.dts#L142) and [rtl9313_hasivo_s1300wp-8xgt-4s-plus.dts:218](``https://github.com/openwrt/openwrt/blob/1e014b68fbb1314555699aac19071f05d6a9e465/target/linux/realtek/dts/rtl9313_hasivo_s1300wp-8xgt-4s-plus.dts#L218).`` Was the 1-4 numbering deliberate?

```suggestion
		function-enumerator = <5>;
```

Same applies to lines 55, 66 and 77 below (→ `<6>`, `<7>`, `<8>`).

---

openwrt-ai (commented):
Reviewed 3 new commits; 3 inline comments below. All three commit messages match the changes they carry, and the series is bisectable (the WP `.dts` files switch to the new `-common.dtsi` in the same commit that adds it).

---

target/linux/realtek/dts/rtl9303_hasivo_f1100wp-4sx-4xgt.dts
pktpls: Very good point about board.d and uci-defaults

target/linux/realtek/image/rtl930x.mk
pktpls: This is fine, the device hasn't been part of an official release yet.

target/linux/realtek/dts/rtl9303_hasivo_f1100wp-4sx-4xgt-512mb.dts
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/dts-v1/;
+
+#include "rtl9303_hasivo_f1100w-4sx-4xgt-common.dtsi"
+
+/ {
+	compatible = "hasivo,f1100wp-4sx-4xgt-512mb", "realtek,rtl930x-soc";
```
plappermaul: Real SOC please.

plappermaul (changes requested):

target/linux/realtek/dts/rtl9303_hasivo_f1100wp-4sx-4xgt-512mb.dts
pktpls: Is this a cleanup that I could do for all devices?

carlosz1986:
love it @pktpls !

target/linux/realtek/dts/rtl9303_hasivo_f1100wp-4sx-4xgt-512mb.dts
plappermaul: the  files you Are working on Are enough.

target/linux/realtek/dts/rtl9303_hasivo_f1100w-4sx-4xgt.dts
```diff
@@ -4,7 +4,7 @@
 #include "rtl9303_hasivo_f1100w-4sx-4xgt-common.dtsi"
 
 / {
-	compatible = "hasivo,f1100w-4sx-4xgt", "realtek,rtl930x-soc";
+	compatible = "hasivo,f1100w-4sx-4xgt", "realtek,rtl9303-soc";
```
openwrt-ai, line 7: The 512MB sibling of this same device was left behind: [`rtl9303_hasivo_f1100w-4sx-4xgt-512mb.dts:7`](https://github.com/openwrt/openwrt/blob/19797e886bedd40d2e237bbe661cf9374704af10/target/linux/realtek/dts/rtl9303_hasivo_f1100w-4sx-4xgt-512mb.dts#L7) still reads `"hasivo,f1100w-4sx-4xgt-512mb", "realtek,rtl930x-soc"`. It shares this `-common.dtsi` and the same SoC, and both new WP `.dts` files in this commit already use `realtek,rtl9303-soc`, so it is the only file in the F1100W family left on the generic string — which doesn't match "be explicit about the real SoC in the root compatible" in the commit message.

Fix is the one-line change in that file:

```
	compatible = "hasivo,f1100w-4sx-4xgt-512mb", "realtek,rtl9303-soc";
```

(`realtek,rtl9303-soc` is matched by the SoC ID table added in [`300-02-enhance-realtek-board-setup.patch`](``https://github.com/openwrt/openwrt/blob/19797e886bedd40d2e237bbe661cf9374704af10/target/linux/realtek/patches-6.18/300-02-enhance-realtek-board-setup.patch#L406),`` so this is behaviour-neutral.)

---

openwrt-ai (commented):
Re-reviewed after the force-push (1e014b6 → 19797e8). All three findings from the previous round are resolved: the two base-files case statements now carry `hasivo,f1100wp-4sx-4xgt{,-512mb}`, the PoE LED `function-enumerator`s were renumbered to 5-8 to match lan5-lan8, and the `SUPPORTED_DEVICES` point was answered (device not in a release yet).

All three commit messages still match their diffs, and the series remains bisectable. One inline comment below on a leftover from the new `realtek,rtl9303-soc` change.

---

openwrt-ai (commented):
Re-reviewed after the force-push (19797e8 → dc6ea2c). The only content change in the range is the one-line fix in `rtl9303_hasivo_f1100w-4sx-4xgt-512mb.dts` (`realtek,rtl930x-soc` → `realtek,rtl9303-soc`), which resolves the finding from the previous round — all four F1100W/WP family `.dts` files now carry the explicit SoC compatible, matching the "be explicit about the real SoC in the root compatible" claim in f0d6a62.

All three commit messages still match their diffs, and all CI checks on the head commit are green (rtl838x/rtl839x/rtl930x/rtl930x_nand/rtl931x/rtl931x_nand kernel builds, kernel patch check, formality checks).

---

jonasjelonek:

PR #24644, opened 2026-08-09
airoha: an7581: enable NET_AIROHA_FLOW_STATS for flow offload statistics

peterwillcn:
Enable CONFIG_NET_AIROHA_FLOW_STATS in an7581 kernel config.
This option exposes hardware flow‑offload counter statistics for the
Airoha PPE flow table, allowing users to inspect offloaded packet/byte
counters.

Enables `NET_AIROHA_FLOW_STATS` to expose PPE hardware flow offload packet/byte counters over devlink. This only enables statistics reporting and does not toggle hardware flow‑offload functionality itself.


peterwillcn:
ping @robimarko

danpawlik:
AFAIR by enabling that, NPU would not work anymore.

target/linux/airoha/an7581/config-6.18
```diff
@@ -231,7 +231,7 @@ CONFIG_MUTEX_SPIN_ON_OWNER=y
 CONFIG_NEED_DMA_MAP_STATE=y
 CONFIG_NEED_SG_DMA_LENGTH=y
 CONFIG_NET_AIROHA=y
-# CONFIG_NET_AIROHA_FLOW_STATS is not set
+CONFIG_NET_AIROHA_FLOW_STATS=y
```
openwrt-ai, line 234: This is not a free "statistics only" switch — it halves the SRAM flow table. `airoha_ppe_hw_init()` subtracts the stats entries from the SRAM data entries: [`sram_ppe_num_data_entries -= sram_num_stats_entries` at airoha_ppe.c:99-101](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/ethernet/airoha/airoha_ppe.c#L99-L101), and the constants are [`PPE_SRAM_NUM_ENTRIES` = 8192 / `PPE_STATS_NUM_ENTRIES` = 4096 in airoha_eth.h:50-51](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/ethernet/airoha/airoha_eth.h#L50-L51) (v6.18.41, the tree this target builds — `KERNEL_PATCHVER:=6.18`). So SRAM-resident offloaded flows drop from 8K to 4K per PPE, plus a coherent DMA allocation for the stats array.

That capacity trade-off is the reason a maintainer would want to keep the symbol off despite its `default y` in Kconfig. Please state it explicitly in the commit message so the decision is on record, rather than describing the change as purely additive ("only enables statistics reporting").

---

openwrt-ai, line 234: @danpawlik's comment ("by enabling that, NPU would not work anymore") has a concrete mechanism in the driver, so it deserves a hardware check before this lands rather than a wording tweak.

With the symbol on, [`airoha_ppe_offload_setup()` at airoha_ppe.c:1363-1369](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/ethernet/airoha/airoha_ppe.c#L1363-L1369) calls `npu->ops.ppe_init_stats()` and jumps to `error_npu_put` on *any* error — it does not degrade to "offload without stats". That error propagates out of [`airoha_ppe_setup_tc_block_cb()` at airoha_ppe.c:1389-1405](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/ethernet/airoha/airoha_ppe.c#L1389-L1405), so the flowtable/tc block registration fails and you get *no* hardware offload at all. The callback in question sends the `PPE_FUNC_SET_WAIT_FLOW_STATS_SETUP` mailbox message and then ioremaps the address the firmware writes back ([`airoha_npu_ppe_stats_setup()` at airoha_npu.c:382-410](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/ethernet/airoha/airoha_npu.c#L382-L410)) — i.e. it only works if the loaded NPU blob implements that command.

The blob OpenWrt ships for this target is `airoha/en7581_npu_rv32.bin` / `en7581_npu_data.bin` from `airoha-en7581-npu-firmware` (`package/firmware/linux-firmware/airoha.mk`). Can you confirm on real AN7581 hardware, with that firmware, that after this change offloaded flows still bind — e.g. entries still show up in `/sys/kernel/debug/ppe/entries` and the counters are non-zero — and that no `airoha_ppe_offload_setup` failure appears in dmesg? A pasted before/after would settle it; if the shipped firmware doesn't implement the command, this one-liner turns offload off instead of adding stats.

---

openwrt-ai (commented):
## Commit checks

- 4619cda "airoha: an7581: enable NET_AIROHA_FLOW_STATS for flow offload statistics" — the message claims the counters are exposed "via devlink/netlink", and the PR body repeats it ("over devlink"). The airoha ethernet driver registers no devlink instance at all — `grep -r devlink drivers/net/ethernet/airoha/` on v6.18.41 returns nothing. The counters actually surface through the flow-offload stats callback ([`airoha_ppe_flow_offload_stats()` at airoha_ppe.c:1253-1276](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/ethernet/airoha/airoha_ppe.c#L1253-L1276), i.e. `FLOW_CLS_STATS` for tc/netfilter flowtable users) and through debugfs ([`airoha_ppe_debugfs.c:155-158`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/ethernet/airoha/airoha_ppe_debugfs.c#L155-L158), `/sys/kernel/debug/ppe/entries`). Please correct both the commit message and the PR body to name the real interfaces.
- 4619cda — the message also frames the change as costless ("does not toggle hardware flow-offload functionality itself"). It does change offload behaviour in two ways; see the inline comments on the changed line.


---

PR #24651, opened 2026-08-10
hostapd: ucode: fix setup of a BSS added at run time

cmonroe:
The 2026-07-09 update changed the type of the first parameter of hostapd_setup_bss() from int to bool. Before the update, add_bss passed -1. The value -1 selected the setup path for a secondary BSS and told hostapd to adopt an existing netdev. Converted to bool, -1 becomes true. As a result, hostapd does not do the setup of a secondary BSS. The new BSS uses the driver state of the first BSS.

Pass false so the non-first block runs again. In this path, use_existing is true, and hostapd adopts a netdev that already exists.

Fixes: 345404476ac8 ("hostapd: update to 2026-07-09")

### DIFF
--- a/package/network/services/hostapd/src/src/ap/ucode.c
+++ b/package/network/services/hostapd/src/src/ap/ucode.c
@@ -470,7 +470,7 @@ uc_hostapd_iface_add_bss(uc_vm_t *vm, size_t nargs)
 
 	if (iface->state == HAPD_IFACE_ENABLED) {
 		hapd->conf->start_disabled = 0;
-		if (hostapd_setup_bss(hapd, -1, true))
+		if (hostapd_setup_bss(hapd, false, true))
 			goto deinit_ctrl;
 	}
 
--- a/package/network/services/hostapd/Makefile
+++ b/package/network/services/hostapd/Makefile
@@ -5,7 +5,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=hostapd
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_URL:=https://w1.fi/hostap.git
 PKG_SOURCE_PROTO:=git


package/network/services/hostapd/src/src/ap/ucode.c
```diff
@@ -470,7 +470,7 @@ uc_hostapd_iface_add_bss(uc_vm_t *vm, size_t nargs)
 
 	if (iface->state == HAPD_IFACE_ENABLED) {
 		hapd->conf->start_disabled = 0;
-		if (hostapd_setup_bss(hapd, -1, true))
+		if (hostapd_setup_bss(hapd, false, true))
```
openwrt-ai, line 473: The `int first` -> `bool first` conversion in 345404476ac8 also rewrote the `use_existing` computation inside `hostapd_setup_bss()` itself, and that half doesn't appear to be addressed here. Before the update `601-ucode_support.patch` carried `bool use_existing = first == -1;`; it is now [`bool use_existing = !first;`](``https://github.com/openwrt/openwrt/blob/9a548030f3646b0d0f37be5fee106477fedec4bc/package/network/services/hostapd/patches/601-ucode_support.patch#L285),`` which lives inside the `if (!first)` block and is therefore unconditionally `true` — which also makes the later [`use_existing = true;`](``https://github.com/openwrt/openwrt/blob/9a548030f3646b0d0f37be5fee106477fedec4bc/package/network/services/hostapd/patches/601-ucode_support.patch#L304``) in the `mld_ap` branch dead.

Net effect: a secondary BSS brought up the normal way via [`__uc_hostapd_bss_start()`](https://github.com/openwrt/openwrt/blob/9a548030f3646b0d0f37be5fee106477fedec4bc/package/network/services/hostapd/src/src/ap/ucode.c#L246) (`first == 0`) now also gets `use_existing = 1`, where it previously got `0`. Upstream passes a literal `1` at that `hostapd_if_add()` call site, so matching it may well be deliberate — but in that case `use_existing` no longer distinguishes anything and could be dropped. Was that path considered as part of this fix, or is it intended as a separate change?

(The call-site fix itself looks right: `false` re-enters the `!first` block and, with `use_existing` true there, hostapd adopts the pre-existing netdev, which is what `-1` used to select.)

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message accurately describes the diff, the `Fixes:` trailer resolves to a real commit with the quoted subject, and `PKG_RELEASE` is bumped. One question inline about the other half of the same `int` -> `bool` conversion.

---

package/network/services/hostapd/src/src/ap/ucode.c
cmonroe: It seems like the intended behavior of the -1 was to inherit the hostapd default behavior so I'll leave this alone for now. Perhaps @nbd168 can provide some adult supervision. If there's a good reason to change it I certainly can but don't want to force a behavior change without fully understanding the consequences.

danpawlik (approved):

PolynomialDivision:

PR #24652, opened 2026-08-10
lantiq: xrx200: fix fiber port support on AVM 5490/5491

abajk:
The fiber port on the AVM 5490/5491 is only partially working. Frames are sent correctly. Most received frames are dropped due to an invalid FCS. It turns out that delays are needed between the MAC and the PHY. This commit enables RGMII delays on the PHY side. Traffic now flows in both directions.

Fiber port topology:
SWITCH<-----RGMII----->PHY<-----SGMII----->SFP/SFF cage

Tested on the AVM 5490. On the other end is a network card with a 1000Base-BX10 SFP module (HiKivision HK-SFP-1.25G-20-1550).

The fiber port on the AVM5491 should also work. The AVM 5490 and 5491 share the same PCB and are almost identical. The only difference is that the AVM5491 has a standard SFP cage with an external GPON module, while the 5490 has a soldered 1000Base-BX SFF module.

Fixes: 48b2df5a4164 ("lantiq: add support for AVM Fritzbox 5490/5491")

### DIFF
--- a/target/linux/lantiq/dts/vr9_avm_fritz5490.dtsi
+++ b/target/linux/lantiq/dts/vr9_avm_fritz5490.dtsi
@@ -107,7 +107,7 @@
 	port@0 {
 		reg = <0>;
 		label = "wan";
-		phy-mode = "rgmii";
+		phy-mode = "rgmii-rxid";
 		phy-handle = <&phy6>;
 	};
 
--- a/target/linux/lantiq/xrx200/base-files/etc/board.d/02_network
+++ b/target/linux/lantiq/xrx200/base-files/etc/board.d/02_network
@@ -74,6 +74,10 @@ lantiq_setup_dsl()
 	avm,fritz7490-micron)
 		annex="b"
 		;;
+	avm,fritz5490|\
+	avm,fritz5490-micron)
+		return
+		;;
 	esac
 
 	lantiq_setup_dsl_helper "$annex"


abajk:
@jschwartzenberg ping :)

openwrt-ai (commented):

Verified against the in-tree/kernel sources: with `phy-mode = "rgmii"` the AT803x driver takes the `else` branch in [`at803x_config_init()`](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/drivers/net/phy/qcom/at803x.c#L433-L446) and explicitly *disables* the RX delay that is on by default after HW reset, which matches the reported RX-side FCS errors; `rgmii-rxid` re-enables it, and [`lantiq_gswip.c`](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/drivers/net/dsa/lantiq_gswip.c#L1714-L1727) correspondingly zeroes the MAC-side PCDU RX delay, so the delay ends up only on the PHY as the commit message states. Same behaviour in 6.18 ([`drivers/net/dsa/lantiq/lantiq_gswip.c`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/dsa/lantiq/lantiq_gswip.c#L1508-L1521)), and `CONFIG_AT803X_PHY=y` is set in both `target/linux/lantiq/xrx200/config-6.12` and `config-6.18`.

The value is also consistent with the sibling AVM boards sharing `vr9_avm_fritzxx90.dtsi` ([`vr9_avm_fritz7490.dtsi:60`](https://github.com/openwrt/openwrt/blob/82745231633e28e93ca15190f86f34683bdf66e7/target/linux/lantiq/dts/vr9_avm_fritz7490.dtsi#L56-L67), `vr9_avm_fritz3490.dtsi:61`, `vr9_avm_fritz3390.dts:177`), and the `Fixes:` trailer resolves correctly to 48b2df5a4164 with a matching subject.

---

openwrt-ai (commented):

`30afe28` "lantiq: xrx200: skip DSL configuration on AVM 5490/5491" — message matches the diff, and the described failure mode checks out in-tree: `lantiq_setup_dsl_helper()` ends with [`ucidef_set_interface_wan "dsl0" "pppoe"` at lantiq.sh:17](https://github.com/openwrt/openwrt/blob/30afe28b7e6971e76cb7dd31abb1cc0b6e050c6f/target/linux/lantiq/base-files/lib/functions/lantiq.sh#L17), which clobbers the `"wan"` port that [`lantiq_setup_interfaces()`](https://github.com/openwrt/openwrt/blob/30afe28b7e6971e76cb7dd31abb1cc0b6e050c6f/target/linux/lantiq/xrx200/base-files/etc/board.d/02_network#L22-L29) assigns to `avm,fritz5490` / `avm,fritz5490-micron` via `ucidef_set_interfaces_lan_wan`. The early `return` is placed in a branch neither board could otherwise match (they are absent from the `annex="b"` list), so `lantiq_setup_dsl_helper` is simply never reached and the `"wan"` (fiber) port survives. Consistent with the image recipes, which already strip every DSL package for both variants ([`vr9.mk:179-215`](https://github.com/openwrt/openwrt/blob/30afe28b7e6971e76cb7dd31abb1cc0b6e050c6f/target/linux/lantiq/image/vr9.mk#L179-L215): `-kmod-ltq-vdsl-vr9`, `-kmod-ltq-ptm-vr9`, `-ltq-vdsl-vr9-app`, …), so the `dsl` section this used to emit was dead config.

Both board compatibles are correct (`avm,fritz5490` from `vr9_avm_fritz5490.dtsi`, `avm,fritz5490-micron` from `vr9_avm_fritz5490-micron.dts:6`), and the `Fixes:` trailer resolves to 48b2df5a4164 with a matching subject — that commit is indeed what added both the `ucidef_set_interfaces_lan_wan` entry and the boards themselves. All 19 checks on the head commit are green.

---

PR #24653, opened 2026-08-10
qualcommax: fix SoC hang after the CMN PLL probe on IPQ5018

kuncy7:
The `ipq-cmn-pll` driver enables the GCC AHB & SYS clocks of the CMN block
through a runtime PM reference taken in probe, registers its clocks and then
drops that reference, so `pm_clk` gates both clocks a few milliseconds after
probe returns.

IPQ5018 does not survive that gate. Boards die silently within milliseconds of
the CMN PLL probe - before userspace exists, and with no consumer of the PLL
outputs active. Depending on the binary layout (micro-timing) this ranges from
an occasional hang to a fully reproducible boot loop.

**This currently affects released images:** the snapshot image for the GL.iNet
GL-B3000 boot loops on my unit. Booting the very same image with
`initcall_blacklist=ipq_cmn_pll_clk_driver_init` brings the board up, and so
does this patch - that is how it was isolated.

### Isolation on a GL-B3000 (UART, all four on the same unit)

| what was booted | result |
|---|---|
| official snapshot image (6.18) | boot loop, dies ~0.39 s, last line `bsg` |
| same image + `initcall_blacklist=ipq_cmn_pll_clk_driver_init` | boots fully |
| own build, no patch | same boot loop |
| own build **with this patch** | boots fully, `clk_summary` shows the CMN block bus clocks held by `9b000.clock-controller` |

### On the mechanism

This is *not* about the driver's own register accesses: the common clock
framework already wraps the provider ops in `clk_pm_runtime_get()/put()`, so
accesses through CCF are safe with or without this change. The failure has
since been isolated as a race between the asynchronous `pm_clk` gate and
early-boot bus activity, on three boards from three vendors:

- Cudy P5 (isolation by @MayorBug): the probe completes in ~628 us and the
  board dies before the next initcall starts; stretching the end of probe by
  ~15 ms makes the same kernel boot reliably; an enabled UNIPHY0 node is what
  arms the failure there (MDIO0/1, GMAC0/1, QCA8337 do not trigger it).
- GL-B3000: neither a 15 ms nor a 2 s stretch helps (8 of 8 boots die with
  either delay), while gating the same clocks on an idle, fully booted system
  is harmless. A delay is a diagnostic, not a workaround: the gate lands in
  the window between the CMN PLL probe and the first consumer reference, and
  no consumer-side scheme can cover that window.

The patch is submitted upstream - v4 carries the isolation above:
https://lore.kernel.org/linux-clk/20260813093351.178419-1-kuncy7@gmail.com/

I am opening this PR separately from the upstream discussion because the
boards are unbootable without it today. Once the fix (or an equivalent) lands
upstream and reaches a 6.18 stable release, the patch can be dropped.

### Testing

- GL.iNet GL-B3000 (IPQ5018): boot loop before, boots reliably after; NAND,
  both radios and ethernet working.
- TP-Link Archer AX55 v1 (IPQ5018): in daily use with this patch.
- Cudy P5 (IPQ5018): hangs during boot without the patch, boots reliably
  with it (Tested-by: Georg Seema, carried in the patch header).
- No effect on other qualcommax subtargets - the gate only ever hurt IPQ5018
  and the change merely keeps a clock enabled while the driver is bound.

### DIFF
--- a/target/linux/qualcommax/patches-6.18/0191-clk-qcom-ipq-cmn-pll-keep-the-CMN-block-bus-clocks-enabled.patch
+++ b/target/linux/qualcommax/patches-6.18/0191-clk-qcom-ipq-cmn-pll-keep-the-CMN-block-bus-clocks-enabled.patch
@@ -1,0 +1,67 @@
+From bd6a3f8a0ee98d465d83863b4e80efcae0b1d24c Mon Sep 17 00:00:00 2001
+From: Stanislaw Pal <kuncy7@gmail.com>
+Date: Thu, 13 Aug 2026 11:38:12 +0200
+Subject: [PATCH] clk: qcom: ipq-cmn-pll: keep the CMN block bus clocks enabled
+
+The probe function takes a runtime PM reference to enable the GCC AHB &
+SYS clocks of the CMN PLL block, registers the clocks, and then drops
+the reference, letting pm_clk gate both clocks asynchronously a few
+milliseconds after probe has returned.
+
+On IPQ5018 that gate races with early-boot activity on the bus and can
+hang the SoC: boards die silently right after the CMN PLL probe, before
+the next initcall gets to run, and the watchdog resets them. Whether a
+given kernel binary survives depends on micro-timing, ranging from an
+occasional hang to a 100% reproducible boot loop. The failure has been
+reported independently on three boards from three vendors (TP-Link
+Archer AX55 v1, GL.iNet GL-B3000, Cudy P5).
+
+Isolation on the Cudy P5 (by Georg Seema) shows the failure is a
+matter of timing against boot activity, not a steady-state clock
+dependency: the probe completes in ~628 us and the board dies before
+the next initcall starts; stretching the end of probe by ~15 ms makes
+the same kernel boot reliably; an enabled UNIPHY0 node is what arms
+the failure there. The armed configuration differs per board (on the
+GL-B3000 the failure persists with UNIPHY0 disabled), and so does the
+window: the same 15 ms stretch - or a 2 s one - still dies 8 of 8
+boots on the GL-B3000, so a delay is a diagnostic, not a workaround.
+Gating the same clocks on an idle, fully booted system is harmless.
+The window between the CMN PLL probe and the first reference taken by
+any consumer is exactly where the gate lands, so no consumer-side
+scheme can cover it.
+
+Take a devres-managed runtime PM reference in probe, so the bus clocks
+stay enabled for as long as the driver is bound and the reference is
+released again on unbind.
+
+Upstream submission (v4):
+https://lore.kernel.org/linux-clk/20260813093351.178419-1-kuncy7@gmail.com/
+
+Fixes: f81715a4c87c ("clk: qcom: Add CMN PLL clock controller driver for IPQ SoC")
+Signed-off-by: Stanislaw Pal <kuncy7@gmail.com>
+Tested-by: Georg Seema <georgseema@gmail.com>
+---
+ drivers/clk/qcom/ipq-cmn-pll.c | 12 ++++++++++++
+ 1 file changed, 12 insertions(+)
+
+--- a/drivers/clk/qcom/ipq-cmn-pll.c
++++ b/drivers/clk/qcom/ipq-cmn-pll.c
+@@ -433,6 +433,18 @@ static int ipq_cmn_pll_clk_probe(struct
+ 	if (ret)
+ 		return dev_err_probe(dev, ret, "Failed to add SYS clock\n");
+ 
++	/*
++	 * Gating the CMN block AHB & SYS clocks is only safe on an idle
++	 * system: without this reference the gate lands asynchronously a
++	 * few milliseconds after probe, in the middle of the early-boot
++	 * probe activity, and on IPQ5018 that races with other bus
++	 * traffic and hangs the SoC. Hold the reference for as long as
++	 * the driver is bound so that the bus clocks stay enabled.
++	 */
++	ret = devm_pm_runtime_get_noresume(dev);
++	if (ret)
++		return ret;
++
+ 	ret = pm_runtime_resume_and_get(dev);
+ 	if (ret)
+ 		return ret;


openwrt-ai (commented):

Checks performed against `KERNEL_PATCHVER:=6.18` (compared with `v6.18.41`):

- `devm_pm_runtime_get_noresume()` exists in 6.18 (`include/linux/pm_runtime.h`, `drivers/base/power/runtime.c`) and its devres action is `pm_runtime_put_noidle()`, so the reference is dropped on unbind before `devm_pm_runtime_enable()`'s `pm_runtime_disable()` and before `devm_pm_clk_create()`'s teardown — release ordering is correct.
- The new hunk applies with `--fuzz=0` and zero offset on top of `0082-v7.1-…-Add-IPQ6018-SoC-support.patch` and `0083-v7.1-…-Add-IPQ8074-SoC-support.patch`, the only other in-tree patches touching `drivers/clk/qcom/ipq-cmn-pll.c`. The `@@ -433,6 +433,16 @@` header and the `10 insertions(+)` diffstat both match the post-patch tree, so no refresh is needed, and `0191-` sorts after `0082`/`0083` as required.
- Patch file is tab-indented with no trailing whitespace/CRLF, and the commit subject scope (`qualcommax: …`) matches the single file touched.


---

openwrt-ai (commented):

The only change since `eb205fb` is the hunk-header context of the single hunk, shortened from `static int ipq_cmn_pll_clk_probe(struct platform_device *pdev)` to `static int ipq_cmn_pll_clk_probe(struct`. That is the correct in-tree form, and it addresses the one thing that was previously non-canonical:

- Re-generating the diff against `v6.18.41` + `0082-v7.1-…-Add-IPQ6018-SoC-support.patch` + `0083-v7.1-…-Add-IPQ8074-SoC-support.patch` (the only other in-tree patches touching `drivers/clk/qcom/ipq-cmn-pll.c`) reproduces the patch in this revision byte-for-byte, truncated hunk header included. So the file now matches what `make target/linux/refresh` emits and needs no further refresh.
- The truncation is GNU `diff -p`'s 40-character function-context limit, which is the dominant convention here: of the hunk headers carrying context in `target/linux/qualcommax/patches-6.18/`, 134 sit at exactly 40 characters. The new string is 39 characters (40 with the trailing space `diff` strips).
- Patch still applies with `--fuzz=0` and zero offset; `@@ -433,6 +433,16 @@` and the `10 insertions(+)` diffstat both match the post-`0082`/`0083` tree.

Carried over from the previous revision and re-confirmed against `KERNEL_PATCHVER:=6.18` (`gregkh/linux` at `v6.18.41`):

- `devm_pm_runtime_get_noresume()` exists in 6.18 — declared at [`include/linux/pm_runtime.h:100`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/include/linux/pm_runtime.h#L100) and defined at [`drivers/base/power/runtime.c:1655`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/base/power/runtime.c#L1655) — and its devres action is `pm_runtime_put_noidle()`, so the reference is dropped on unbind before `devm_pm_runtime_enable()`'s `pm_runtime_disable()` and before `devm_pm_clk_create()`'s teardown. Release ordering is correct.
- Taking the reference `_noresume` before the existing `pm_runtime_resume_and_get()` is the right order: the resume that follows brings the clocks up, and the extra reference keeps `usage_count > 0` afterwards so the device never idles back down.
- `0191-` is free (in-tree jumps `0189-` → `0305-`) and sorts after `0082`/`0083`, as required.
- Commit subject scope `qualcommax: …` matches the single file touched; the message body accurately describes the change. CI on the head SHA is green across all three FormalityCheck runs.

---

MayorBug:
Thanks for figuring this out and for the patch. I actually found it first through your forum post, then started using it in my ongoing Cudy P5 bring up.

On the Cudy P5 IPQ5018, the device hangs during boot without this patch. With it applied, the board boots reliably.
Tested-by: Georg Seema <georgseema@gmail.com>

kuncy7:
Thanks for testing, and for the tag - I've added it to the commit here, and I'll carry it upstream in the next revision so it doesn't get lost.

Good to know it isn't just my two boards. For the record that now makes three IPQ5018 boards from three vendors (Archer AX55 v1, GL-B3000, Cudy P5) where gating the CMN block bus clocks after probe kills the boot.

One thing worth knowing while you continue the P5 bring up: what exactly dies when those clocks get gated is still not pinned down. The clock framework already takes a runtime PM reference around the provider ops, so it is not this driver's own register accesses. My working theory is that the CMN block AHB clock gates more than the CMN PLL registers - on IPQ5018 that block sits in the same CSR region as the MDIO controllers and the uniphys - but that is a theory, not a result, and the upstream discussion is still open:

https://lore.kernel.org/linux-clk/20260805193638.15327-1-kuncy7@gmail.com/

If you ever feel like helping to name the offending access, a third board would genuinely help: the experiment is to boot with the fix reversed but with mdio/uniphy/gmac disabled in DT (expected: survives), then re-enable one of them at a time until it dies. No pressure at all - the fix stands on its own either way.

kuncy7:
For reference, there is a forum thread collecting test reports for this, and describing how to confirm the symptom on a board that hangs before you can get a shell:

https://forum.openwrt.org/t/ipq5018-current-snapshots-hang-during-boot-cmn-pll-bus-clocks-fix-in-pr-24653/252633

It also carries the earlier discussion that led here, for anyone who wants the full history:

https://forum.openwrt.org/t/does-anyone-have-kernel-6-18-booting-reliably-on-an-ipq50xx-board/252240

openwrt-ai (commented):

The only change since `8df7e2d` is the `Tested-by: Georg Seema <georgseema@gmail.com>` trailer — added both to the OpenWrt commit message and to the carried patch's own trailer block — plus the accompanying "Independently confirmed on a second board: … Cudy P5" paragraph in the OpenWrt commit body. No code hunk changed.

- The trailer matches [the tester's comment](https://github.com/openwrt/openwrt/pull/24653#issuecomment-5251508694) verbatim (name and address), and the added body paragraph accurately reports what that comment says (Cudy P5, IPQ5018, hangs without / boots with). Commit message and diff still agree; subject scope `qualcommax: …` still matches the single file touched.
- Trailer placement is correct on both sides: it sits below `Signed-off-by:` inside the patch's own trailer block, above the `---` separator, so it survives as part of the kernel commit message rather than being stripped as a patch comment.
- Re-verified that the patch is unaffected: applying `0082-v7.1-…-Add-IPQ6018-SoC-support.patch`, `0083-v7.1-…-Add-IPQ8074-SoC-support.patch` and then this patch on top of `v6.18.41`'s `drivers/clk/qcom/ipq-cmn-pll.c` succeeds with `--fuzz=0` and zero offset. `@@ -433,6 +433,16 @@` and the `10 insertions(+)` diffstat still match the post-`0082`/`0083` tree, so no refresh is needed. `0191-` is still free on `main` (in-tree jumps `0189-` → `0305-`) and still sorts after `0082`/`0083`.

Carried over from the previous revisions and still valid against `KERNEL_PATCHVER:=6.18` (`gregkh/linux` at `v6.18.41`):

- `devm_pm_runtime_get_noresume()` exists in 6.18 — declared at [`include/linux/pm_runtime.h:100`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/include/linux/pm_runtime.h#L100), defined at [`drivers/base/power/runtime.c:1655`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/base/power/runtime.c#L1655) — and its devres action is `pm_runtime_put_noidle()`, so the reference is dropped on unbind before `devm_pm_runtime_enable()`'s `pm_runtime_disable()` and before `devm_pm_clk_create()`'s teardown. Release ordering is correct.
- Taking the reference `_noresume` ahead of the existing `pm_runtime_resume_and_get()` is the right order: the resume that follows brings the clocks up, and the extra reference keeps `usage_count > 0` afterwards so the device never idles back down.
- The truncated hunk-header context (`static int ipq_cmn_pll_clk_probe(struct`) is GNU `diff -p`'s 40-character limit and matches what `make target/linux/refresh` emits — unchanged in this revision.
- CI on the head SHA is green across all three FormalityCheck runs (Code Patches, Git & Commits, OpenWrt Makefiles).

---

MayorBug:
I put the Cudy P5 through the isolation tests.

- The board boots without the fix when UNIPHY0 is disabled.
- MDIO0, MDIO1, GMAC0, GMAC1, and QCA8337 are not enough to trigger the reset.
- Enabling only UNIPHY0 triggers a watchdog reset.
- Removing the UNIPHY0 assigned-clock properties changes nothing.
- pm_runtime_put() returns, and the CMN probe completes.
- The board dies before the next init call starts.

This is a proper Heisenbug: looking at it changes the timing enough to hide it.
Debug messages around pm_runtime_put() added about 13 ms to the CMN probe.
That was enough to make the bug vanish. UNIPHY0 then completed its probe.
I replaced the messages with: usleep_range(15000, 16000);
The board booted again. The CMN probe took 16.8 ms, and UNIPHY0 completed normally.
Without the delay, the CMN probe takes about 628 us and the board resets.With a 15-16 ms delay, it boots.
So UNIPHY0 makes the issue reproducible on the Cudy P5, and delaying the end of the CMN probe by ~15 ms also avoids it.

kuncy7:
This is excellent work - thank you. It lines up exactly with what I saw isolating the same thing on a GL-B3000, and your delay experiment nails the mechanism better than I managed to.

The picture your data paints: `ipq_cmn_pll_clk_probe()` ends with `pm_runtime_put()`, `pm_clk` then gates `GCC_CMN_BLK_AHB`/`SYS` asynchronously from a workqueue a few ms later, and on IPQ5018 the CMN block shares a CSR fabric segment with UNIPHY0 (0x98000, right next to CMN at 0x9b000). When that gate lands *during* a UNIPHY0 register transaction - and UNIPHY0 probes right after CMN - the fabric hangs mid-access and the watchdog fires. That explains why it's UNIPHY0 specifically, why it's a timing race rather than a permanent clock dependency (your 15 ms delay separates the gate from UNIPHY0's access in time), and why a rebuild with a few extra `pr_info`s flips the outcome.

On the B3000 I saw the same thing from the other side: gating the clocks on an *idle* system, long after boot, is completely harmless (WiFi keeps serving clients) - it only kills when the gate collides with early-boot bus activity. Your delay and my hold-the-reference fix both work by making sure that collision never happens; holding the reference just does it deterministically instead of relying on a timing margin.

Would you be OK with me quoting your isolation on the linux-clk thread? It's exactly the concrete evidence the maintainer asked for, and a third board from a third vendor makes it much stronger.

MayorBug:
Yes ofc, you can use my testing as evidence

target/linux/qualcommax/patches-6.18/0191-clk-qcom-ipq-cmn-pll-keep-the-CMN-block-bus-clocks-enabled.patch
```diff
@@ -0,0 +1,68 @@
...
+https://lore.kernel.org/linux-clk/20260813093351.178419-1-kuncy7@gmail.com/
+
+Fixes: f81715a4c87c ("clk: qcom: Add CMN PLL clock controller driver for IPQ SoC")
+Signed-off-by: Stanislaw Pal <kuncy7@gmail.com>
+Tested-by: Georg Seema <georgseema@gmail.com>
+---
+ drivers/clk/qcom/ipq-cmn-pll.c | 12 ++++++++++++
+ 1 file changed, 12 insertions(+)
+
+
```
openwrt-ai: nit: this revision added a second blank line between the diffstat and the `--- a/...` header. Every other patch in `target/linux/qualcommax/patches-6.18/` uses either one blank line (67 files) or none (14 files); none uses two, and the previous revision of this file (`8df7e2d`) had one. It sits in the part quilt treats as the patch header, so `make target/linux/refresh` will not normalise it away.

```suggestion
 1 file changed, 12 insertions(+)

```

---

target/linux/qualcommax/patches-6.18/0191-clk-qcom-ipq-cmn-pll-keep-the-CMN-block-bus-clocks-enabled.patch
```diff
@@ -0,0 +1,68 @@
...
+The window between the CMN PLL probe and the first reference taken by
+any consumer is exactly where the gate lands, so no consumer-side
+scheme can cover it.
+
+Take a devres-managed runtime PM reference in probe, so the bus clocks
+stay enabled for as long as the driver is bound and the reference is
+released again on unbind.
+
+Upstream submission (v4):
+https://lore.kernel.org/linux-clk/20260813093351.178419-1-kuncy7@gmail.com/
```
openwrt-ai, line 38: nit: the PR description is now out of step with this revision. It still links the first upstream posting (`.../20260805193638.15327-1-kuncy7@gmail.com/`) where the patch header and the OpenWrt commit message both link v4 (`.../20260813093351.178419-1-kuncy7@gmail.com/`), and it still states "Which access is the fatal one has not been pinned down yet" while both messages now name the UNIPHY0 race and the Cudy P5 timing isolation. Worth refreshing the description so a reader landing on the PR gets the same account as the commit that will be merged.

---

openwrt-ai (commented):
Reviewed 1 new commit; two nits inline, no blocking issues.

The change since `9d9107b` is a rewrite of the carried patch's kernel commit message (v4 wording, the three-board evidence and the Cudy P5 timing isolation), a rewrite of the in-code comment, and the corresponding rewrite of the OpenWrt commit body. The code change itself is unchanged in effect.

Checks against `KERNEL_PATCHVER:=6.18` (in-tree `LINUX_VERSION-6.18 = .44`, compared with `gregkh/linux` at `v6.18.44`):

- The hunk grew from 10 to 12 insertions (the comment is two lines longer). Both the diffstat (`12 ++++++++++++` / `1 file changed, 12 insertions(+)`) and the hunk header (`@@ -433,6 +433,18 @@`, 6 → 18 with 3 lines of context each side) match the new body exactly, so the metadata is self-consistent.
- Applying `0082-v7.1-…-Add-IPQ6018-SoC-support.patch`, `0083-v7.1-…-Add-IPQ8074-SoC-support.patch` and then this patch on top of `v6.18.44`'s `drivers/clk/qcom/ipq-cmn-pll.c` succeeds with `--fuzz=0` and zero offset — no refresh needed. `0191-` is still free on `main` (numbering jumps `0189-` → `0305-`) and still sorts after `0082`/`0083`.
- The two new comment lines are tab-indented like the rest of the block, stay under 80 columns, and carry no trailing whitespace. The only over-80 line in the file is the `Fixes:` tag, which is conventionally not wrapped.
- Commit-message accuracy: the patch body's Cudy P5 figures track the tester's report verbatim (~628 us probe, dies before the next initcall, ~15 ms stretch makes it boot, UNIPHY0 is what arms it), and the OpenWrt commit body summarises the same without contradicting it. Subject scope `qualcommax: …` still matches the single file touched.

Carried over and still valid:

- `devm_pm_runtime_get_noresume()` exists in 6.18 — declared at [`include/linux/pm_runtime.h:100`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/include/linux/pm_runtime.h#L100), defined at [`drivers/base/power/runtime.c:1655`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/base/power/runtime.c#L1655) — and its devres action is `pm_runtime_put_noidle()`, so the reference is released on unbind before `devm_pm_runtime_enable()`'s `pm_runtime_disable()` and before `devm_pm_clk_create()`'s teardown. Release ordering is correct.
- Taking the reference `_noresume` ahead of the existing `pm_runtime_resume_and_get()` is the right order: the resume that follows brings the clocks up, and the extra reference keeps `usage_count > 0` afterwards so the device never idles back down.
- The truncated hunk-header context (`static int ipq_cmn_pll_clk_probe(struct`) is GNU `diff -p`'s 40-character limit and is what `make target/linux/refresh` emits.
- CI on `3547d86` is green: all three FormalityCheck runs plus the ipq50xx/ipq60xx/ipq807x kernel builds and the ipq50xx patch check.

---

target/linux/qualcommax/patches-6.18/0191-clk-qcom-ipq-cmn-pll-keep-the-CMN-block-bus-clocks-enabled.patch
kuncy7: Fixed in 73e14542 - back to a single blank line between the diffstat and the `---` header (an artifact of how this revision was assembled; the previous revisions were clean). The public copy of the patch was fixed the same way.

target/linux/qualcommax/patches-6.18/0191-clk-qcom-ipq-cmn-pll-keep-the-CMN-block-bus-clocks-enabled.patch
kuncy7: Description refreshed: the mechanism section now carries the v4 account (the Cudy P5 timing isolation, the delay-is-not-a-workaround measurements from the GL-B3000) and links the v4 posting; Cudy P5 added under Testing.

openwrt-ai (commented):
Reviewed 1 new commit; no new issues found. Both nits from the previous review are resolved.

The only change since `3547d86` is the removal of the extra blank line between the diffstat and the `--- a/drivers/clk/qcom/ipq-cmn-pll.c` header. The OpenWrt commit message is byte-identical to the previous revision, and no code hunk changed.

- The patch header now has a single blank line before the `---`/`+++` block, which is the dominant in-tree form: across the 81 patches in `target/linux/qualcommax/patches-6.18/`, 67 use one blank line, 14 use none, and none uses two. Matches the state the file was in at `8df7e2d`.
- The PR description was refreshed as suggested: the mechanism section now carries the v4 account (Cudy P5 timing isolation, the GL-B3000 delay measurements showing a delay is a diagnostic rather than a workaround) and links `20260813093351.178419-1-kuncy7@gmail.com`, the same posting the patch header and the commit message link. Cudy P5 is listed under Testing. The description and the commit that will be merged now tell the same story.
- Re-verified against `KERNEL_PATCHVER:=6.18` (`gregkh/linux` at `v6.18.44`): applying `0082-v7.1-…-Add-IPQ6018-SoC-support.patch`, `0083-v7.1-…-Add-IPQ8074-SoC-support.patch` and then this patch on top of `v6.18.44`'s `drivers/clk/qcom/ipq-cmn-pll.c` succeeds with `--fuzz=0` and zero offset. `@@ -433,6 +433,18 @@` and the `12 insertions(+)` diffstat still match the post-`0082`/`0083` tree, so no refresh is needed. `0191-` is still free on `main` (numbering jumps `0189-` → `0305-`) and still sorts after `0082`/`0083`.
- Whitespace is clean: tab-indented body, no CRLF, no trailing whitespace outside the diff's own blank-context space, and the only over-80 line is the `Fixes:` tag, which is conventionally not wrapped.

Carried over and still valid:

- `devm_pm_runtime_get_noresume()` exists in 6.18 — declared at [`include/linux/pm_runtime.h:100`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/include/linux/pm_runtime.h#L100), defined at [`drivers/base/power/runtime.c:1655`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/base/power/runtime.c#L1655) — and its devres action is `pm_runtime_put_noidle()`, so the reference is released on unbind before `devm_pm_runtime_enable()`'s `pm_runtime_disable()` and before `devm_pm_clk_create()`'s teardown. Release ordering is correct.
- Taking the reference `_noresume` ahead of the existing `pm_runtime_resume_and_get()` is the right order: the resume that follows brings the clocks up, and the extra reference keeps `usage_count > 0` afterwards so the device never idles back down.
- The truncated hunk-header context (`static int ipq_cmn_pll_clk_probe(struct`) is GNU `diff -p`'s 40-character limit and is what `make target/linux/refresh` emits.
- Commit subject scope `qualcommax: …` matches the single file touched, and the message body accurately describes the change.
- CI on `73e1454` is green across all three FormalityCheck runs (Code Patches, Git & Commits, OpenWrt Makefiles).

---

AugustoAmaral:
Another data point, on a Xiaomi AX6000 (IPQ5018 + QCN9074).

The stock snapshot boots fine here, but my NSS-enabled build for this board hit exactly what you describe:
silent death within a few ms of the CMN PLL probe, and, matching the micro-timing note, it ranged from
an occasional hang to a fully reproducible boot loop depending on the binary layout. It was the single
thing blocking NSS bring-up on the AX6000.

With this patch applied, it boots reliably to shell every time, and the NSS core comes up clean
(`NSS core 0 booted successfully`). Confirmed over UART, both RAM-booted and flashed.
Great find and a clear explanation of the mechanism, thanks @kuncy7.

Tested by me today! :)

wallystech-Gg:
I can confirm the same issue on the Wallys DR5018S (IPQ5018+qcn6122, 512 MB DDR3L, with UNIPHY0 and QCA8081 enabled).

Without this patch, Linux 6.18.44 shows a timing-dependent boot failure. The same image may sometimes reach userspace, while other boots silently hang or reset through the watchdog before userspace starts. Some hangs do not recover automatically and require a full power cycle. I observed this with both TFTP-booted initramfs images and kernels booted from NAND.

With the current revision of this PR applied, I performed five complete power-cycle tests and all five boots reached OpenWrt reliably. Ethernet and all three ath11k radios (2.4 GHz, 5 GHz and 6 GHz) worked correctly.

This confirms that the issue also affects the Wallys DR5018S and that its behavior is consistent with the early-boot race caused by gating the IPQ5018 CMN PLL bus clocks.

Tested-by: Alan

robimarko:
Manually merged for now as upstream is slow:
https://github.com/openwrt/openwrt/commit/86b584bd09949f14231d373c46563cc9b8b0666d

Please keep it moving upstream, as it looks like a nice bug fix

namiltd:
I think this patch can be replaced by https://lore.kernel.org/linux-clk/20260918-ipq5018-mark-gpll0_main-critical-v1-1-fbe8f27a0106@gmail.com/

Issue #19702, opened 2025-08-07
qualcommax - ipq60xx: Kernel memory leak on axt1800


flipreverse:
### Describe the bug

The Linux kernel 6.12 suffers from a memory leak. It is most likely located in the ath11k driver.
A quick check with kmemleak do not reveal any memory leaks. Given that, the memory seems to be referenced somewhere.

### OpenWrt version

r0-d802e63

### OpenWrt release

SNAPSHOT

### OpenWrt target/subtarget

qualcommax/ipq60xx

### Device

GL.iNet GL-AXT1800

### Image kind

Self-built image

### Steps to reproduce

- Load the `ath11kk-ahb` kernel module
- Start at least one hostapd for either 2,4 GHz or 5 GHz
- Just wait
- Memory usage starts to increase

### Actual behaviour

- Kernel eats the memory. Presumably the ath11k driver.
- Behavior does not occur when the module is not loaded or no

### Diffconfig

```text

```

### Terms

- [x] I am reporting an issue for OpenWrt, not an unsupported fork.

brada4:
How much it grows compared to v24.10.2?

flipreverse:
> How much it grows compared to v24.10.2?

I cannot relate it to the growth in v24.10.2.  I, however, can show use the memory usage over the past six hours.

<img width="1171" height="261" alt="Image" src="https://github.com/user-attachments/assets/722d7872-2048-4812-83e4-c7d4383561c1" />

brada4:
No operational problem whatsoever?

flipreverse:
> No operational problem whatsoever?

No, except thevfact that my system runs out of memory. I added a cronjob that reboots every two days.

brada4:
please add zram-swap and keep it longer than usual. Once something lands in compressed RAM chech whether RAM is spent by slab (kernel) or anon(userland)
If you could log all /proc/meminfo into your graphs

flipreverse:
> please add zram-swap and keep it longer than usual. Once something lands in compressed RAM chech whether RAM is spent by slab (kernel) or anon(userland) If you could log all /proc/meminfo into your graphs

Do I need zram-swap or is normal swap sufficient? I already have a swap partition attached.

brada4:
Normal swap is ok but slow for casual usage.

flipreverse:
@brada4 How long shall I record the memory usage?
A scripts writes the content of `/proc/meminfo` every 10 mins to a file.

[...]

PR #24658, opened 2026-08-10
mac80211: ath11k: fix use after free in ath11k_dp_rx_msdu_coalesce()

sonpython:
Backport of upstream commit [72b8654e3b83](https://github.com/torvalds/linux/commit/72b8654e3b83) ("wifi: ath11k: fix use after free in ath11k_dp_rx_msdu_coalesce()"), currently in ath-next and not yet present in stable 6.12.y (checked up to 6.12.102) nor in the backports snapshot shipped by `package/kernel/mac80211`.

**Bug:** in `ath11k_dp_rx_msdu_coalesce()` the A-MSDU dequeue loop reads `rxcb->is_continuation` after `dev_kfree_skb_any()` has freed the skb that rxcb points into. Under memory pressure the freed skb gets recycled quickly, so the stale read corrupts A-MSDU chain processing, leaking kernel pages. The lower free memory gets, the more likely the misread: the leak is self-accelerating and ends in OOM.

**Field data (9x Linksys MX4200, IPQ8074, qualcommax/ipq807x):**
- Unpatched: MemAvailable declines 81-490 MB/day (scales with RX error rate), recoverable only by reloading the ath11k module; worst node needed 143 reloads over 10.7 days and eventually OOM-rebooted. This matches long-standing reports in #19702.
- With this patch: canary node flat for 40+ hours (no measurable decline, zero reload interventions) under identical load, while unpatched control nodes on the same build needed up to 11 reloads/day. Since validated and rolled out to all 9 nodes.

Related: I previously opened #24417 / #24416 with a wrong diagnosis and closed them with an explanation; this PR is the actual fix in the correct location (ath11k is built from mac80211 backports, not the target kernel tree).

Tested on: Linksys MX4200v1 (qualcommax/ipq807x), 9 nodes, kernel 6.12.94, backports 6.18.39.


package/kernel/mac80211/patches/ath11k/950-wifi-ath11k-fix-use-after-free-in-dp_rx_msdu_coalesce.patch
```diff
@@ -0,0 +1,53 @@
...
+Link: https://patch.msgid.link/20260505171709.547274-1-willmar@met-dubbel-l.nl
+Signed-off-by: Jeff Johnson <jeff.johnson@oss.qualcomm.com>
+---
+ drivers/net/wireless/ath/ath11k/dp_rx.c | 6 ++++--
+ 1 file changed, 4 insertions(+), 2 deletions(-)
+
+diff --git a/drivers/net/wireless/ath/ath11k/dp_rx.c b/drivers/net/wireless/ath/ath11k/dp_rx.c
+index 72d5d933656ddd..2a413e3a07a78c 100644
+--- a/drivers/net/wireless/ath/ath11k/dp_rx.c
++++ b/drivers/net/wireless/ath/ath11k/dp_rx.c
```
openwrt-ai, line 26: This is raw `git format-patch` output. The patches under `package/kernel/mac80211/patches/` are quilt-managed: none of the 164 of them carry a `diff --git` or an `index` line — the diffstat is followed directly by `--- a/...` (see the same directory's [947-wifi-ath11k-fix-rssi-station-dump-for-IPQ5018-and-QC.patch:19-24](``https://github.com/openwrt/openwrt/blob/168fb653baf5b169fb8044d563d2dd940b8d76a9/package/kernel/mac80211/patches/ath11k/947-wifi-ath11k-fix-rssi-station-dump-for-IPQ5018-and-QC.patch#L19-L24)).`` The hunk headers here also keep git's untruncated function context (`@@ -1761,6 +1761,7 @@ static int ath11k_dp_rx_msdu_coalesce(struct ath11k *ar,`), whereas every other patch in the tree has it truncated the way GNU diff emits it.

Dropping the two extended-header lines is enough to make the patch match the tree; regenerating it with `make package/mac80211/refresh` normalizes the hunk headers as well.

```suggestion
--- a/drivers/net/wireless/ath/ath11k/dp_rx.c
+++ b/drivers/net/wireless/ath/ath11k/dp_rx.c
```


---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the diff, and the patch context lines up with `drivers/net/wireless/ath/ath11k/dp_rx.c` as of Linux 6.18.41 (`rem_len = msdu_len - buf_first_len;` is at line 1810 there, matching the second hunk header), so it should apply cleanly against the backports-6.18.39 tarball; it also does not overlap the hunks of the existing `949-wifi-ath11k-fix-monitor-rx-pktlen.patch`, which is the only other patch touching that file.

One inline comment on patch formatting. The Signed-off-by / author-email mismatch and the missing `PKG_RELEASE` bump for `package/kernel/mac80211` are already covered by the formality-check bot, so they are not repeated here.

I was not able to independently verify upstream commit `72b8654e3b83` from this session (only `openwrt/openwrt` is reachable, and the commit is not in the 6.18.41 tree), so the provenance of the patch header is taken at face value.

---

openwrt-ai (commented):

The force-push adds the `PKG_RELEASE` bump for `package/kernel/mac80211` (3 -> 4), which was the outstanding formality item; the patch file itself is byte-identical to the previously reviewed revision, and the commit message still matches the diff.

Re-verified the patch context against `drivers/net/wireless/ath/ath11k/dp_rx.c` in Linux v6.18.41 (the series backports-6.18.39 tracks): both hunk headers still line up (`rem_len = msdu_len - buf_first_len;` at line 1810, `bool is_continuation` insertion point at 1761), and no other patch in `patches/ath11k/` touches those lines, so it should apply cleanly.

The existing inline comment about the raw `git format-patch` headers (`diff --git` / `index`) is unaddressed but still applies — not repeated here.

Upstream commit `72b8654e3b83` remains unverifiable from this session: it is in ath-next, not in the v6.18.41 tree available here, and only `openwrt/openwrt` is reachable via the GitHub connector. Its provenance is again taken at face value.

---

PR #17172, opened 2024-12-05
mediatek: filogic: add support for Tenbay WR3000K

jian-yu:
Tenbay WR3000K is an 802.11ax (Wi-Fi 6) router, based on MediaTek MT7981B. 

## Specifications:
- SoC: MetiaTek MT7981B
- RAM: Hynex H5TQ2G863GFR 512MiB
- Flash: Winbond W25N01GVZEIG 128MiB
- Wi-Fi: MediaTek MT7976C (2.4GHz/5GHz, 802.11ax, 2x2 MIMO, AX3000)
- MediaTek MT7915E: 2.4GHz and 5GHz
- Ethernet: 1x 10/100/1000 Mbps WAN + 3x 10/100/1000 Mbps LAN
- Switch: MediaTek MT7531AE
- UART: J4 (115200 baud)
- LEDs: Power
- Buttons: Reset, WPS
- PWR: 12V/1A DC, 5.5×2.1 connector

## MAC Addresses:
| Vendor  | OpenWrt Interface | Address       | Notes                                          |
|---------|-------------------|---------------|------------------------------------------------|
| WAN     | wan            | Stored MAC - 2     |  Label MAC |
| LAN     | br-lan             | Stored MAC - 1   |    |
| 2.4GHz  | phy0-ap0          | Stored MAC     |   Stored in  Factory + offset 4        |
| 5GHz    | phy1-ap0          | Stored MAC +1    |              |

## MTD Partitions:
- 0x000000000000-0x000000100000 : "BL2"
- 0x000000100000-0x000000180000 : "u-boot-env"
- 0x000000180000-0x000000380000 : "Factory"
- 0x000000380000-0x000000580000 : "FIP"
- 0x000000580000-0x000003580000 : "ubi"
- 0x000003580000-0x000006580000 : "ubi1"
- 0x000006580000-0x0000065a0000 : "Product"
- 0x0000065a0000-0x000007580000 : "Custom"

#### Note:
- The original partition-Ubi partition-Ubi1 is an AB dual system, and Openwrt only uses Ubi. So flash requires modifying the uboot variable `boot_from=ubi` to ensure that it only starts from Ubi.

- The `Product` and `Custom` partitions are original and only exist to align with the original layout; they are not used by OpenWrt.

## UBI Partitions (Dynamic):
- id: 0, kernel
- id: 1, rootfs
- id: 2, rootfs_data

##### Hardware Requirements
- **USB-to-TTL Serial Adapter** (e.g., CH340 or CP2102).
[...]

PR #24662, opened 2026-08-11
mediatek: filogic: add Tenbay WR3000K OpenWrt U-Boot layout variant

unisequence:
This commit adds OpenWrt U-Boot layout (ubootmod) variant for the Tenbay WR3000K #17172 (MediaTek MT7981B, 512MB DDR3, 128MB SPI-NAND).

The stock device (tenbay,wr3000k) uses the vendor U-Boot with a redundant flash layout: ubi (48MB) + ubi1 (48MB) + Product (128KB) + Custom (16MB). After kernel and filesystem overhead, this leaves only ~29MB of overlay space for packages.

This variant replaces the vendor BL2/FIP with OpenWrt mainline U-Boot and merges those four partitions into a single 112MB ubi volume (0x580000-0x75a0000, matching their combined span exactly), up from the 48MB stock ubi. This leaves ~85MB of overlay space for packages.

The DTS is refactored into a shared mt7981b-tenbay-wr3000k-common.dtsi included by both variants. NMBM stays in the stock DTS only: the OpenWrt U-Boot for this board has no NMBM support (the defconfig added by 504-add-tenbay_wr3000k.patch doesn't set CONFIG_ENABLE_NAND_NMBM, same as 501/502/503), so the ubootmod variant must see raw flash for U-Boot and Linux to agree on the layout. This matches the pattern used by CF-WR632AX and Cudy WR3000S.

Flashing instructions:

1. From the stock OpenWrt firmware, install kmod-mtd-rw and unlock MTD:
```
apk add kmod-mtd-rw && insmod mtd-rw i_want_a_brick=1
```

2. Step 3 below overwrites the vendor BL2/FIP and erases the Product and Custom partitions, which hold per-unit vendor data that cannot be regenerated. There is no path back to stock afterwards, so back them up first, e.g. using the mtd numbers from `cat /proc/mtd`:
```
cat /dev/mtd0 > /tmp/BL2.bin
cat /dev/mtd3 > /tmp/FIP.bin
cat /dev/mtd6 > /tmp/Product.bin
cat /dev/mtd7 > /tmp/Custom.bin
```

3. Upload preloader.bin and bl31-uboot.fip to /tmp, then write:
```
mtd write /tmp/openwrt-mediatek-filogic-tenbay_wr3000k-ubootmod-preloader.bin BL2
mtd write /tmp/openwrt-mediatek-filogic-tenbay_wr3000k-ubootmod-bl31-uboot.fip FIP
mtd erase ubi && mtd erase ubi1 && mtd erase Product && mtd erase Custom
reboot
```

4. Set a static IP on your PC: 192.168.1.254/24 and serve initramfs-recovery.itb via TFTP. The new U-Boot will automatically attempt TFTP boot when no system is found.

5. From the recovery shell, perform sysupgrade:
```
sysupgrade -n /tmp/openwrt-mediatek-filogic-tenbay_wr3000k-ubootmod-squashfs-sysupgrade.itb
```

Link: https://github.com/openwrt/openwrt/pull/17172

### DIFF
--- a/package/boot/uboot-mediatek/patches/504-add-tenbay_wr3000k.patch
+++ b/package/boot/uboot-mediatek/patches/504-add-tenbay_wr3000k.patch
@@ -1,0 +1,333 @@
+--- /dev/null
++++ b/configs/mt7981_tenbay_wr3000k_defconfig
+@@ -0,0 +1,111 @@
++CONFIG_ARM=y
++CONFIG_ARM_SMCCC=y
++CONFIG_SYS_HAS_NONCACHED_MEMORY=y
++CONFIG_POSITION_INDEPENDENT=y
++CONFIG_ARCH_MEDIATEK=y
++CONFIG_TEXT_BASE=0x41e00000
++CONFIG_SYS_MALLOC_F_LEN=0x4000
++CONFIG_NR_DRAM_BANKS=1
++CONFIG_DEFAULT_DEVICE_TREE="mt7981-tenbay-wr3000k"
++CONFIG_OF_LIBFDT_OVERLAY=y
++CONFIG_TARGET_MT7981=y
++CONFIG_SYS_LOAD_ADDR=0x46000000
++CONFIG_PRE_CON_BUF_ADDR=0x4007ef00
++CONFIG_DEBUG_UART_BASE=0x11002000
++CONFIG_DEBUG_UART_CLOCK=40000000
++CONFIG_DEBUG_UART=y
++CONFIG_FIT=y
++CONFIG_BOOTDELAY=30
++CONFIG_AUTOBOOT_KEYED=y
++CONFIG_AUTOBOOT_MENU_SHOW=y
++CONFIG_DEFAULT_FDT_FILE="mediatek/mt7981-tenbay-wr3000k.dtb"
++CONFIG_LOGLEVEL=7
++CONFIG_PRE_CONSOLE_BUFFER=y
++CONFIG_LOG=y
++# CONFIG_BOARD_INIT is not set
++CONFIG_BOARD_LATE_INIT=y
++CONFIG_HUSH_PARSER=y
++CONFIG_SYS_PROMPT="MT7981> "
++CONFIG_CMD_CPU=y
++CONFIG_CMD_LICENSE=y
++CONFIG_CMD_BOOTMENU=y
++CONFIG_CMD_ASKENV=y
++CONFIG_CMD_ERASEENV=y
++CONFIG_CMD_ENV_FLAGS=y
++CONFIG_CMD_STRINGS=y
++CONFIG_CMD_DM=y
++CONFIG_CMD_GPIO=y
++CONFIG_CMD_GPT=y
++CONFIG_CMD_MTD=y
++CONFIG_CMD_PART=y
++CONFIG_CMD_TFTPSRV=y
++CONFIG_CMD_RARP=y
++CONFIG_CMD_CDP=y
++CONFIG_CMD_SNTP=y
++CONFIG_CMD_LINK_LOCAL=y
++CONFIG_CMD_DHCP=y
++CONFIG_CMD_DNS=y
++CONFIG_CMD_PING=y
++CONFIG_CMD_PXE=y
++CONFIG_CMD_CACHE=y
++CONFIG_CMD_PSTORE=y
++CONFIG_CMD_PSTORE_MEM_ADDR=0x42ff0000
++CONFIG_CMD_UUID=y
++CONFIG_CMD_HASH=y
++CONFIG_CMD_SMC=y
++CONFIG_CMD_UBI=y
++CONFIG_CMD_UBI_RENAME=y
++CONFIG_OF_EMBED=y
++CONFIG_ENV_OVERWRITE=y
++CONFIG_ENV_IS_IN_UBI=y
++CONFIG_ENV_REDUNDANT=y
++CONFIG_ENV_UBI_PART="ubi"
++CONFIG_ENV_UBI_VOLUME="ubootenv"
++CONFIG_ENV_UBI_VOLUME_REDUND="ubootenv2"
++CONFIG_ENV_RELOC_GD_ENV_ADDR=y
++CONFIG_ENV_USE_DEFAULT_ENV_TEXT_FILE=y
++CONFIG_ENV_DEFAULT_ENV_TEXT_FILE="defenvs/tenbay_wr3000k_env"
++CONFIG_ENV_VARS_UBOOT_RUNTIME_CONFIG=y
++CONFIG_VERSION_VARIABLE=y
++CONFIG_NETCONSOLE=y
++CONFIG_USE_IPADDR=y
++CONFIG_IPADDR="192.168.1.1"
++CONFIG_USE_SERVERIP=y
++CONFIG_SERVERIP="192.168.1.254"
++CONFIG_NET_RANDOM_ETHADDR=y
++CONFIG_BUTTON=y
++CONFIG_BUTTON_GPIO=y
++CONFIG_CLK=y
++CONFIG_GPIO_HOG=y
++CONFIG_LED=y
++CONFIG_LED_BLINK=y
++CONFIG_LED_GPIO=y
++# CONFIG_MMC is not set
++CONFIG_MTD=y
++CONFIG_DM_MTD=y
++CONFIG_MTD_SPI_NAND=y
++CONFIG_MTD_UBI_FASTMAP=y
++CONFIG_PHY_FIXED=y
++CONFIG_MEDIATEK_ETH=y
++CONFIG_PHY=y
++CONFIG_PHY_MTK_TPHY=y
++CONFIG_PINCTRL=y
++CONFIG_PINCONF=y
++CONFIG_PINCTRL_MT7981=y
++CONFIG_POWER_DOMAIN=y
++CONFIG_MTK_POWER_DOMAIN=y
++CONFIG_DM_REGULATOR=y
++CONFIG_DM_REGULATOR_FIXED=y
++CONFIG_DM_REGULATOR_GPIO=y
++CONFIG_RAM=y
++CONFIG_DM_RNG=y
++CONFIG_RNG_MTK_V2=y
++CONFIG_CMD_RNG=y
++CONFIG_DM_SERIAL=y
++CONFIG_SERIAL_RX_BUFFER=y
++CONFIG_MTK_SERIAL=y
++CONFIG_SPI=y
++CONFIG_DM_SPI=y
++CONFIG_MTK_SPIM=y
++CONFIG_ZSTD=y
++CONFIG_HEXDUMP=y
+--- /dev/null
++++ b/arch/arm/dts/mt7981-tenbay-wr3000k.dts
+@@ -0,0 +1,160 @@
++// SPDX-License-Identifier: GPL-2.0
++/*
++ * Tenbay WR3000K - U-Boot device tree
++ *
++ * GPIOs match the Linux device tree
++ * (mt7981b-tenbay-wr3000k-common.dtsi): pio 0 = wps button,
++ * pio 1 = reset button, pio 10 = red status LED,
++ * pio 11 = green status LED, pio 12 = blue status LED.
++ */
++
++/dts-v1/;
++#include "mt7981.dtsi"
++#include <dt-bindings/gpio/gpio.h>
++#include <dt-bindings/input/linux-event-codes.h>
++
++/ {
++	#address-cells = <1>;
++	#size-cells = <1>;
++	model = "Tenbay WR3000K";
++	compatible = "tenbay,wr3000k", "mediatek,mt7981";
++
++	chosen {
++		stdout-path = &uart0;
++		tick-timer = &timer0;
++	};
++
++	memory@40000000 {
++		device_type = "memory";
++		reg = <0x40000000 0x20000000>;
++	};
++
++	keys {
++		compatible = "gpio-keys";
++
++		button-0 {
++			label = "wps";
++			linux,code = <KEY_WPS_BUTTON>;
++			gpios = <&pio 0 GPIO_ACTIVE_LOW>;
++		};
++
++		button-1 {
++			label = "reset";
++			linux,code = <KEY_RESTART>;
++			gpios = <&pio 1 GPIO_ACTIVE_LOW>;
++		};
++	};
++
++	leds {
++		compatible = "gpio-leds";
++
++		led-0 {
++			label = "red:status";
++			gpios = <&pio 10 GPIO_ACTIVE_LOW>;
++		};
++
++		led-1 {
++			label = "green:status";
++			gpios = <&pio 11 GPIO_ACTIVE_LOW>;
++		};
++
++		led-2 {
++			label = "blue:status";
++			gpios = <&pio 12 GPIO_ACTIVE_LOW>;
++		};
++	};
++};
++
++&uart0 {
++	status = "okay";
++};
++
++&eth {
++	status = "okay";
++	mediatek,gmac-id = <0>;
++	phy-mode = "2500base-x";
++	mediatek,switch = "mt7531";
++	reset-gpios = <&pio 39 GPIO_ACTIVE_HIGH>;
++
++	fixed-link {
++		speed = <2500>;
++		full-duplex;
++	};
++};
++
++&pio {
++	spi_flash_pins: spi0-pins-func-1 {
++		mux {
++			function = "flash";
++			groups = "spi0", "spi0_wp_hold";
++		};
++
++		conf-pu {
++			pins = "SPI0_CS", "SPI0_HOLD", "SPI0_WP";
++			drive-strength = <MTK_DRIVE_8mA>;
++			bias-pull-up = <MTK_PUPD_SET_R1R0_11>;
++		};
++
++		conf-pd {
++			pins = "SPI0_CLK", "SPI0_MOSI", "SPI0_MISO";
++			drive-strength = <MTK_DRIVE_8mA>;
++			bias-pull-down = <MTK_PUPD_SET_R1R0_11>;
++		};
++	};
++};
++
++&spi0 {
++	#address-cells = <1>;
++	#size-cells = <0>;
++	pinctrl-names = "default";
++	pinctrl-0 = <&spi_flash_pins>;
++	status = "okay";
++	must_tx;
++	enhance_timing;
++	dma_ext;
++	ipm_design;
++	support_quad;
++	tick_dly = <2>;
++	sample_sel = <0>;
++
++	spi_nand@0 {
++		compatible = "spi-nand";
++		reg = <0>;
++		spi-max-frequency = <52000000>;
++
++		partitions {
++			compatible = "fixed-partitions";
++			#address-cells = <1>;
++			#size-cells = <1>;
++
++			partition@0 {
++				label = "BL2";
++				reg = <0x0 0x100000>;
++			};
++
++			partition@100000 {
++				label = "u-boot-env";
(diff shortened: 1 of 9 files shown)


target/linux/mediatek/dts/mt7981b-tenbay-wr3000k-ubootmod.dts
```diff
@@ -0,0 +1,24 @@
...
+	compatible = "tenbay,wr3000k-ubootmod", "mediatek,mt7981";
+
+	chosen {
+		bootargs = "root=/dev/fit0 rootwait";
+		rootdisk = <&ubi_rootdisk>;
+	};
+};
+
+&ubi {
+	reg = <0x580000 0x6e80000>;
```
openwrt-ai: The merged volume is smaller than the four partitions it replaces, so ~1.6 MiB of flash ends up owned by nothing.

The vendor partitions ([`mt7981b-tenbay-wr3000k.dts:11-25`](https://github.com/openwrt/openwrt/blob/213db16e0b8f11cbc6f780c7ddd7ef5f7054c1a9/target/linux/mediatek/dts/mt7981b-tenbay-wr3000k.dts#L11-L25) plus `ubi` from the shared dtsi) span `0x580000`–`0x75a0000`:

    ubi     0x3000000
    ubi1    0x3000000
    Product 0x0020000
    Custom  0x1000000
    -----------------
            0x7020000

This sets `0x6e80000`, i.e. `0x1a0000` (1.6 MiB = 13 erase blocks) short of the merged span, and no other partition covers the tail.

There is even a bit more headroom available: with `mediatek,bmt-max-ratio = <1>` / `mediatek,bmt-max-reserved-blocks = <64>` NMBM reserves 64 of the 1024 blocks ([`mtk_bmt_nmbm.c:1631`](https://github.com/openwrt/openwrt/blob/213db16e0b8f11cbc6f780c7ddd7ef5f7054c1a9/target/linux/generic/files/drivers/mtd/nand/mtk_bmt_nmbm.c#L1631), `NMBM_MGMT_DIV` is 16), so the upper MTD ends at `0x7800000` and a size of `0x7280000` would still fit.

Is `0x6e80000` deliberate (bad-block headroom you want on top of NMBM's reserve), or should it be `0x7020000`? Whatever the answer, the same value in `package/boot/uboot-mediatek/patches/504-add-tenbay_wr3000k.patch:269` has to move with it, otherwise U-Boot and Linux disagree on where the UBI device ends.

---

package/boot/uboot-mediatek/patches/504-add-tenbay_wr3000k.patch
```diff
@@ -0,0 +1,333 @@
...
++			#address-cells = <1>;
++			#size-cells = <1>;
++
++			partition@0 {
++				label = "BL2";
++				reg = <0x0 0x100000>;
++			};
++
++			partition@100000 {
++				label = "u-boot-env (unused)";
```
openwrt-ai: nit: the `(unused)` marker is on the wrong side. In-tree the convention is to put it in the *Linux* DTS of the ubootmod variant — see [`mt7981b-teralink-tl3020-256mb.dts:187`](https://github.com/openwrt/openwrt/blob/213db16e0b8f11cbc6f780c7ddd7ef5f7054c1a9/target/linux/mediatek/dts/mt7981b-teralink-tl3020-256mb.dts#L187), and the same in `mt7981b-snr-snr-cpe-ax2.dts`, `mt7981b-netis-n6-v2.dts`, `mt7981b-netis-nx31.dts`. This PR inverts it: the U-Boot DTS gets `(unused)` while the Linux ubootmod DTS inherits a plain `u-boot-env` from the shared dtsi. No other board DTS under `package/boot/uboot-mediatek/patches/` uses a parenthesised partition label.

```suggestion
+				label = "u-boot-env";
```

If you want the marker, override the label in `mt7981b-tenbay-wr3000k-ubootmod.dts` instead — it can't live in the shared dtsi because the stock variant really does use that MTD partition for its environment (`tenbay,wr3000k` is still in the non-UBI block of `package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic`).

---

target/linux/mediatek/dts/mt7981b-tenbay-wr3000k-common.dtsi
```diff
@@ -0,0 +1,233 @@
...
+		};
+	};
+};
+
+&spi0 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&spi0_flash_pins>;
+	status = "okay";
+
+	spi_nand: flash@0 {
```
openwrt-ai, line 149: nit: the `spi_nand` label added here is never referenced — neither variant uses `&spi_nand`. The `partitions:` and `ubi:` labels added by the same refactor are both used, so this one looks like a leftover.

```suggestion
	flash@0 {
```

---

openwrt-ai (commented):
## Commit checks

- 213db16 "mediatek: filogic: add Tenbay WR3000K OpenWrt U-Boot layout variant" — the capacity figures don't add up. The message says the stock layout leaves "only ~48MB usable for packages" and that this variant produces a "single 110MB ubi volume, freeing ~84MB". 48 MB + 84 MB = 132 MB, not 110 MB. What the diff actually does is grow the `ubi` partition from `0x3000000` (48 MiB) to `0x6e80000` (110.5 MiB), i.e. roughly +62 MiB. Same wording in the PR body — please make the two consistent with each other and with the DTS.
- 213db16 — the flash procedure is one-way and doesn't say so. Step 2 overwrites `BL2`/`FIP` with the OpenWrt preloader/FIP and erases `ubi1`, `Product` and `Custom`; `Product`/`Custom` hold per-unit vendor data that cannot be regenerated, and the vendor U-Boot is gone afterwards. Please add a "dump `BL2`, `FIP`, `Product` and `Custom` before erasing" step, and either a revert-to-stock path or an explicit note that there isn't one.

Everything else lines up with the in-tree conventions I checked: patch number 504 is free and the patch has the same three-file shape (`configs/`, `arch/arm/dts/`, `defenvs/`) as 501–503; `trusted-firmware-a-mt7981-spim-nand-ddr3-1866` exists; the new `U-Boot/mt7981_tenbay_wr3000k` entry and `UBOOT_TARGETS` addition are alphabetically placed and match the `NAME:=` convention (layout suffix omitted, as for `wavlink_wl-wnt100x3` / `routerich_ax3000`); `tenbay,wr3000k-ubootmod` lands in the `ubootenv_add_ubi_default`, `fit_do_upgrade` and `fit_check_image` blocks, all correctly sorted; the `Device/tenbay_wr3000k-ubootmod` block matches `teralink_tl3020-256mb`; and the split into `mt7981b-tenbay-wr3000k-common.dtsi` is a faithful move of the existing node content.

<sub>Automated review. CI on 213db16 is green.</sub>

---

target/linux/mediatek/dts/mt7981b-tenbay-wr3000k-common.dtsi
```diff
@@ -0,0 +1,233 @@
...
+		spi-cal-datalen = <7>;
+		spi-cal-data = /bits/ 8 <0x53 0x50 0x49 0x4E 0x41 0x4E 0x44>;
+		spi-cal-addrlen = <5>;
+		spi-cal-addr = /bits/ 32 <0x0 0x0 0x0 0x0 0x0>;
+
+		spi-tx-bus-width = <4>;
+		spi-rx-bus-width = <4>;
+		mediatek,nmbm;
+		mediatek,bmt-max-ratio = <1>;
+		mediatek,bmt-max-reserved-blocks = <64>;
```
openwrt-ai: Keeping the NMBM properties in the *shared* dtsi turns NMBM on for the `-ubootmod` variant too, and that variant's bootloader does not speak NMBM: the defconfig added by this PR does not set `CONFIG_ENABLE_NAND_NMBM`, which is the guard around [`board_nmbm_init()` in 129-board-mediatek-wire-up-NMBM-support.patch](``https://github.com/openwrt/openwrt/blob/4398c4862457c1f97746be6a1b561da577ba551d/package/boot/uboot-mediatek/patches/129-board-mediatek-wire-up-NMBM-support.patch).`` So OpenWrt U-Boot would read/write the `ubi` MTD raw while Linux goes through NMBM's remapping layer — the two agree only for as long as no block is ever remapped.

Every one of the 74 in-tree DTS/DTSI files that set `mediatek,nmbm` belongs to a stock layout; none is a `-ubootmod` or `-ubi` variant. The directly analogous refactor is COMFAST CF-WR632AX, which splits exactly this way: the common dtsi keeps a [`spi_nand:` label on the flash node](``https://github.com/openwrt/openwrt/blob/4398c4862457c1f97746be6a1b561da577ba551d/target/linux/mediatek/dts/mt7981b-comfast-cf-wr632ax-common.dtsi#L139),`` the NMBM properties live only in the stock DTS ([`mt7981b-comfast-cf-wr632ax.dts:10-15`](https://github.com/openwrt/openwrt/blob/4398c4862457c1f97746be6a1b561da577ba551d/target/linux/mediatek/dts/mt7981b-comfast-cf-wr632ax.dts#L10-L15)), and [`mt7981b-comfast-cf-wr632ax-ubootmod.dts`](https://github.com/openwrt/openwrt/blob/4398c4862457c1f97746be6a1b561da577ba551d/target/linux/mediatek/dts/mt7981b-comfast-cf-wr632ax-ubootmod.dts) inherits none of them.

Suggest restoring the `spi_nand:` label that was dropped on the last push and moving these three properties into `mt7981b-tenbay-wr3000k.dts` as a `&spi_nand { ... }` block. Note that without NMBM the 64 reserved blocks at the top of the flash are free, so the ubootmod `ubi` could then run to `0x8000000` instead of stopping at `0x75a0000` — comfast does exactly that.

---

package/boot/uboot-mediatek/patches/504-add-tenbay_wr3000k.patch
```diff
@@ -0,0 +1,333 @@
...
++bootdelay=0
++bootfile=openwrt-mediatek-filogic-tenbay_wr3000k-ubootmod-initramfs-recovery.itb
++bootfile_bl2=openwrt-mediatek-filogic-tenbay_wr3000k-ubootmod-preloader.bin
++bootfile_fip=openwrt-mediatek-filogic-tenbay_wr3000k-ubootmod-bl31-uboot.fip
++bootfile_upg=openwrt-mediatek-filogic-tenbay_wr3000k-ubootmod-squashfs-sysupgrade.itb
++bootled_status=blue:status
++bootmenu_confirm_return=askenv - Press ENTER to return to menu ; bootmenu 60
++bootmenu_default=0
++bootmenu_delay=0
++bootmenu_title=      [0;34m( ( ( [1;39mOpenWrt[0;34m ) ) )
```
openwrt-ai: The ANSI colour sequences here lost their ESC (0x1B) bytes — the file literally contains `[0;34m`, not `<ESC>[0;34m` (`grep bootmenu_title 504-add-tenbay_wr3000k.patch | cat -A` shows no `^[`). As written the boot menu prints the raw `[0;34m` / `[1;39m` text instead of colouring the title.

Same on lines 304 and 305 (`[31m` … `[0m` around the two "write to NAND" entries) and line 333 (`[33m$ver[0m`).

For reference, [`502-add-bazis-ax3000wm.patch:296`](``https://github.com/openwrt/openwrt/blob/4398c4862457c1f97746be6a1b561da577ba551d/package/boot/uboot-mediatek/patches/502-add-bazis-ax3000wm.patch#L296``) carries the real ESC bytes (`^[[0;34m` under `cat -A`); copying the four lines from there is the easiest fix. (`503-add-wavlink-wl-wnt100x3.patch` is not a good template — it has U+241B "␛" symbol characters instead of real ESC.)

---

package/boot/uboot-mediatek/patches/504-add-tenbay_wr3000k.patch
```diff
@@ -0,0 +1,333 @@
...
++CONFIG_ZSTD=y
++CONFIG_HEXDUMP=y
+--- /dev/null
++++ b/arch/arm/dts/mt7981-tenbay-wr3000k.dts
+@@ -0,0 +1,160 @@
++// SPDX-License-Identifier: GPL-2.0
++/*
++ * Tenbay WR3000K - U-Boot device tree
++ *
++ * GPIOs match the already-merged Linux device tree
```
openwrt-ai: nit: `mt7981b-tenbay-wr3000k-common.dtsi` is not "already-merged" — it is added by this same commit. Either drop the qualifier or point at the merged `mt7981b-tenbay-wr3000k.dts` instead.

```suggestion
+ * GPIOs match the Linux device tree
```

---

openwrt-ai (commented):
Re-reviewed at 4398c48 (previous review was against 213db16).

Both earlier commit-message findings are resolved: the capacity figures now add up and match the DTS (`0x580000`–`0x75a0000` = `0x7020000` is exactly `ubi` + `ubi1` + `Product` + `Custom`, and the `mtdN` numbers in the backup step match the stock partition order), and the one-way nature of the procedure is now stated explicitly with a backup step. No new commit-message issues.

The remaining points are inline. The NMBM one is the substantive one; the other two are cosmetic.

<sub>Automated review. The one failing check on this head (`Determine Changed Files`) is a GitHub infrastructure failure downloading `tj-actions/changed-files` (429/503), unrelated to this PR.</sub>

---

unisequence:
Addressed the review feedback, hardware-tested:

- **NMBM**: moved out of the shared `-common.dtsi` into the stock (non-ubootmod) `.dts` only, via a restored `spi_nand:` label, matching the CF-WR632AX pattern. The `-ubootmod` U-Boot has no NMBM support, so U-Boot and Linux now agree on the raw flash layout.
- **Bootmenu colors**: restored the missing ESC bytes in the four `bootmenu_*` lines.
- **UBI size**: tried extending `ubi` to the full `0x8000000` as suggested. Broke on hardware — UBI's reserved-PEB accounting leaves no margin when a volume fills 100% of a partition; after one sysupgrade cycle, cold boot hit `not enough PEBs` and U-Boot force-erased the partition. Reverted to `0x7020000`, stable across repeated sysupgrade+reboot on hardware.
- **Comment wording**: dropped "already-merged".

Hardware-verified: 512MB DRAM confirmed, colored bootmenu, `sysupgrade -n` creates the `fit` volume correctly, boots to a running system.

unisequence:
Tested extending \`ubi\` to the full \`0x8000000\` (matching netis N6 V2 / netis NX32U / Cudy WR3000S, which all use this same pattern on 128MB flash). It survives one sysupgrade cycle fine, but fails on the second: replacing the \`fit\` volume (remove + recreate) leaves UBI without enough spare PEBs to re-attach cleanly on the next cold boot, and the device falls back to booting \`initramfs-recovery.itb\` over TFTP instead of the flashed system. Not a brick, but not stable either.

Keeping \`ubi\` at \`0x7020000\` (the size in this PR) — verified stable across repeated sysupgrade+reboot cycles on hardware.

target/linux/mediatek/dts/mt7981b-tenbay-wr3000k.dts
```diff
@@ -1,247 +1,32 @@
...
-		interrupt-controller;
-		#interrupt-cells = <1>;
-		interrupt-parent = <&pio>;
-		interrupts = <38 IRQ_TYPE_LEVEL_HIGH>;
-	};
+&spi_nand {
+	mediatek,nmbm;
+	mediatek,bmt-max-ratio = <1>;
+	mediatek,bmt-max-reserved-blocks = <64>;
 };
```
openwrt-ai, line 15: nit: this asymmetry is undocumented in the commit message, which only says "the DTS is refactored into a shared `mt7981b-tenbay-wr3000k-common.dtsi` included by both variants". A reader diffing the two variants can't tell whether NMBM being stock-only is deliberate or an oversight. One sentence in the message — the OpenWrt U-Boot for this board has no NMBM support (the defconfig added by `504-add-tenbay_wr3000k.patch` doesn't set `CONFIG_ENABLE_NAND_NMBM`, same as `501`/`502`/`503`), so the ubootmod variant must see raw flash for U-Boot and Linux to agree — would make it self-explanatory.

The code itself is right and matches [`mt7981b-comfast-cf-wr632ax.dts:10-15`](https://github.com/openwrt/openwrt/blob/ec86dbe67e6f48711b8b0b9ca910500a78adc0f3/target/linux/mediatek/dts/mt7981b-comfast-cf-wr632ax.dts#L10-L15) exactly.

---

openwrt-ai (commented):
Re-reviewed at ec86dbe (previous review was against 4398c48). The new push is three changes, and all three resolve earlier findings:

- **NMBM split** — `spi_nand:` label restored on the flash node in the common dtsi, and `mediatek,nmbm` / `mediatek,bmt-max-ratio` / `mediatek,bmt-max-reserved-blocks` moved into `mt7981b-tenbay-wr3000k.dts` as a `&spi_nand { ... }` override. Byte-identical to the three properties the stock DTS carried on `main`, and the block placement matches the COMFAST CF-WR632AX precedent exactly. The ubootmod variant now inherits none of them, which is consistent with its U-Boot defconfig not enabling `CONFIG_ENABLE_NAND_NMBM` (neither do `501`/`502`/`503`).
- **Bootmenu colours** — the four `bootmenu_*` lines now carry real ESC bytes; `grep -P '\[[0-9;]+m'` finds no remaining sequence in the patch that isn't preceded by `\x1b`.
- **Comment wording** — "already-merged" dropped.

The `ubi` size stays at `0x580000 0x7020000`, and the U-Boot DTS in `504-add-tenbay_wr3000k.patch:269` carries the same `reg`, so the two still agree. I'm taking the hardware test result on `0x8000000` at face value and not re-raising it.

Commit message is unchanged from 4398c48 and still matches the diff; no new commit-message issues. One cosmetic nit inline.

For the record, the FormalityCheck bot warning about missing `From:`/`Date:`/`Subject:` git headers on `504-add-tenbay_wr3000k.patch` looks like a false positive here — none of `500`–`503` in that directory have them either.

<sub>Automated review. CI on ec86dbe is fully green (22 checks).</sub>

---

openwrt-ai (commented):

Re-reviewed at 2fdd657 (previous review was against ec86dbe). The tree is byte-identical to ec86dbe (`git diff ec86dbe..2fdd657` is empty) — the only change is the commit message, which gained the NMBM rationale paragraph. That resolves the last outstanding nit.

Spot-checked the three factual claims the new paragraph makes:

- "NMBM stays in the stock DTS only" — `mediatek,nmbm` / `mediatek,bmt-max-ratio` / `mediatek,bmt-max-reserved-blocks` appear only in [`mt7981b-tenbay-wr3000k.dts:11-15`](https://github.com/openwrt/openwrt/blob/2fdd657e5f7ef20e39ad7a9b29a72a586d9cfb5e/target/linux/mediatek/dts/mt7981b-tenbay-wr3000k.dts#L11-L15); neither the common dtsi nor the ubootmod DTS carries them.
- "the defconfig added by `504-add-tenbay_wr3000k.patch` doesn't set `CONFIG_ENABLE_NAND_NMBM`, same as 501/502/503" — none of `501`–`504` mention NMBM at all.
- "matches the pattern used by CF-WR632AX and Cudy WR3000S" — both split the same way: NMBM only in `mt7981b-comfast-cf-wr632ax.dts` / `mt7981b-cudy-wr3000s-v1.dts`, with the shared dtsi and the `-ubootmod` DTS carrying none.

PR body and commit message agree on this paragraph too, so no drift there.

<sub>Automated review. CI on 2fdd657 is fully green (23 checks, no failures).</sub>

---

unisequence:
Gentle ping @hauke — this one's ready too, formality check feedback addressed and hardware-tested. Thanks!

unisequence:
@dangowrt  could you take a look at this one when you have time? Thanks!

openwrt-ai (commented):

---

unisequence:
@jonasjelonek Can you review this PR?

PR #24528, opened 2026-08-02
qualcommax: ipq807x: 301w: read the port MAC addresses from ART

JuliusBairaktaris:
Since the PPE conversion the 301w brings `lan1`, `lan2`, `10g-1` and `10g-2` up with no MAC
address of their own. As DSA user ports they inherit the conduit's, so four ports and
`br-lan` all answer to the same address. Only `lan4` and `lan3` still get one patched in,
though the board names all six for the bootloader, in
`target/linux/qualcommax/dts/ipq8072-301w.dts`:

```
	ethernet0 = &lan4;
	ethernet1 = &lan3;
	ethernet2 = &lan2;
	ethernet3 = &lan1;
	ethernet4 = &port_10g_1;
	ethernet5 = &port_10g_2;
```

The addresses it hands out sit in a table at the start of `0:art`, one 6-byte slot per alias.
This describes those six slots as nvmem cells and points each port at its own, so the ports
read them whatever the bootloader does. `ipq8071-ax3600.dtsi` describes its own ART table the
same way.

@xlighting2017 hexdumped `0:art` on a 301w: slots 0 to 5 hold six consecutive addresses, the
rest of the first 48 bytes `ff`. The same unit on `r35301-d47ce71a89`, the last snapshot
before f50435627d37, brought lan4 up on slot 0 through 10g-2 on slot 5
(openwrt/openwrt#19121). That answers the open question from #24512, where this commit first
sat. It is split out here because it does not depend on that change.

I have no 301w here, so that is a user's dump and a user's boot log, not a board on my bench.
Nobody has flashed this branch on a 301w yet.

This does not change a port that works today: the nvmem cell is the last source consulted,
and an all-`ff` slot is not a valid address, so an unprovisioned port keeps the behaviour it
has now. `net/core/of_net.c, of_get_mac_address():`

```c
	ret = of_get_mac_addr(np, "mac-address", addr);
	if (!ret)
		return 0;

	ret = of_get_mac_addr(np, "local-mac-address", addr);
	if (!ret)
		return 0;

	ret = of_get_mac_addr(np, "address", addr);
	if (!ret)
		return 0;

	return of_get_mac_address_nvmem(np, addr);
```

PR #24663, opened 2026-08-11
qualcommax: fix the RT-AX89X coming up with no ethernet ports

### LINKED ISSUE #24604: OpenWRT Snapshot:  |    Deferred probe pending: qca-edma: failed to get MAC address
### Describe the bug

Since #24512 the RT-AX89X is not getting MAC-Addresses for 10g-copper, VLAN-ports or any of the then LAN-Ports thus these are non-functional.

Also tried **r35691-224a967ba1** for good measure but issue is the same. 

Don't know if #24528 (also) fixes (this) issue for the RT-AX89X, but am skeptical  about that.
 I don' t see any  "0:ART" partition for the RT-AX89X 

### OpenWrt version

r35210-8614a2ba68

### OpenWrt release

SNAPSHOT

### OpenWrt target/subtarget

qualcommax/ipq807x

### Device

Asus RT-AX89X

### Image kind

Official downloaded image

### Steps to reproduce

Install Snapshot on RT-AX89X and , reboot.

Notice that _only_ the WiFi-radio's and BR-LAN have a MAC-address now....  

### Actual behaviour

The logic described in #24512 apparently does not hold for the RT-AX89X  
(or i might be wrongly assuming causation between #24512 an the issue I'm seeing here...) 

For my RT-AX89X  since "qca_edma" fails to get a mac-address: 
`[kern.warn: [   25.779011] platform 3ab00000.edma: deferred probe pending: qca-edma: failed to get MAC address`

The ten RT-AX89X's LAN-ports are coupled by 90000.mdio-1:10 and are inoperable since #24512 was pushed to  main   for OpenWRT Snapshot about four days ago. 

### Expected behaviour

Working 10g-copper, LAN, WAN (and VLAN) ports that are assigned (their) MAC-Address.

### Additional info

I thought this might be broader then the RT-AX89X and tried my MX4200v1 which is also built on Qualcomm's IP8074 platform (with the addition that the RT-AX89X apparently uses a newer revision but I digress), However, I tried the snapshot dated august 5th and  the MX4200V1 is _not_ affected by this issue: it works fine. 

So perhaps the board file for the RT-AX89X has a oddity that 

### Diffconfig

```text

```

### Terms

JuliusBairaktaris:
`edma_probe()` propagates `-EPROBE_DEFER` out of the conduit's MAC address
lookup, but a DSA port address is optional, so the whole datapath ends up
waiting on a supplier that may never arrive. Once an nvmem provider is
registered whose layout driver never bound, that lookup defers for good:

    drivers/nvmem/core.c, nvmem_layout_module_get_optional():

        if (!nvmem->layout->dev.driver ||
            !try_module_get(nvmem->layout->dev.driver->owner))
                return -EPROBE_DEFER;

On the RT-AX89X every port points at one `u-boot,env` cell in `0:appsblenv`,
and the `fw_printenv` output @Razerwire posted here reports a bad CRC over
that whole partition, which is the span the layout driver reads. My
understanding is the layout probe fails on that CRC and adds no cells, so the
conduit never probes: no `eth0`, no DSA, no ports.

Patch 1 treats every lookup error the same and falls back to a random address,
with a warning that names the failed lookup. A conduit is not a port: losing
it loses `eth0` and every tree behind it. That gives up the transient case, a
provider not yet registered, since the two states look the same at this API.

Patch 2 restores the `ethernet0..5` aliases
[f50435627d](https://github.com/openwrt/openwrt/commit/f50435627d370a1bb07d455d7081207b7b6a744e)
dropped, now pointing at the ports that replaced the dp nodes, so the
bootloader can patch each address in. The `nvmem-cells` pairs and the
`u-boot,env` layout node go with them, because the six qca8337 ports have no
alias and nothing else to read, and since 679b73b1884a a deferring port takes
its whole tree down:

    target/linux/generic/pending-6.18/
    704-net-dsa-support-EPROBE_DEFER-when-reading-the-port-M.patch:

        err = dsa_port_setup(dp);
        if (err == -EPROBE_DEFER)
                goto teardown;

Patch 1 ran on my AX3600 (ipq8071): `eth0` comes up with the `0:art` address
the WAN port uses, ports and WAN unaffected. Patch 2 has not run on any board
and I have no RT-AX89X. Which `ethNaddr` variables the environment holds is
unknown: on a bad CRC `fw_printenv` prints U-Boot's sandbox default
environment, so that dump cannot say.

Fixes https://github.com/openwrt/openwrt/issues/24604

I used AI to help me write this change. @Razerwire, could you test a build with
these two patches on your RT-AX89X and post the boot dmesg?

### DIFF
--- a/target/linux/qualcommax/dts/ipq8074-rt-ax89x.dts
+++ b/target/linux/qualcommax/dts/ipq8074-rt-ax89x.dts
@@ -17,6 +17,12 @@
 	aliases {
 		serial0 = &blsp1_uart5;
 		mdio-gpio0 = &mdio1;
+		ethernet0 = &port_switch;
+		ethernet1 = &port_lan2;
+		ethernet2 = &port_lan1;
+		ethernet3 = &port_wan;
+		ethernet4 = &port_10g_sfp;
+		ethernet5 = &port_10g_copper;
 		led-boot = &led_pwr;
 		led-failsafe = &led_pwr;
 		led-running = &led_pwr;
@@ -684,7 +690,7 @@
 			nvmem-cell-names = "mac-address";
 		};
 
-		port@2 {
+		port_lan2: port@2 {
 			reg = <2>;
 			label = "lan2";
 			phy-handle = <&qca8075_9>;
@@ -694,7 +700,7 @@
 			nvmem-cell-names = "mac-address";
 		};
 
-		port@3 {
+		port_lan1: port@3 {
 			reg = <3>;
 			label = "lan1";
 			phy-handle = <&qca8075_a>;
@@ -704,7 +710,7 @@
 			nvmem-cell-names = "mac-address";
 		};
 
-		port@4 {
+		port_wan: port@4 {
 			reg = <4>;
 			label = "wan";
 			phy-handle = <&qca8075_b>;
@@ -714,7 +720,7 @@
 			nvmem-cell-names = "mac-address";
 		};
 
-		port@5 {
+		port_10g_sfp: port@5 {
 			reg = <5>;
 			label = "10g-sfp";
 			sfp = <&sfp>;
@@ -725,7 +731,7 @@
 			nvmem-cell-names = "mac-address";
 		};
 
-		port@6 {
+		port_10g_copper: port@6 {
 			reg = <6>;
 			label = "10g-copper";
 			phy-handle = <&aqr113c>;
--- a/target/linux/qualcommax/dts/ipq8074-rt-ax89x.dts
+++ b/target/linux/qualcommax/dts/ipq8074-rt-ax89x.dts
@@ -17,6 +17,12 @@
 	aliases {
 		serial0 = &blsp1_uart5;
 		mdio-gpio0 = &mdio1;
+		ethernet0 = &port_switch;
+		ethernet1 = &port_lan2;
+		ethernet2 = &port_lan1;
+		ethernet3 = &port_wan;
+		ethernet4 = &port_10g_sfp;
+		ethernet5 = &port_10g_copper;
 		led-boot = &led_pwr;
 		led-failsafe = &led_pwr;
 		led-running = &led_pwr;
@@ -451,12 +457,6 @@
 			partition@3e0000 {
 				label = "0:appsblenv";
 				reg = <0x003e0000 0x20000>;
-
-				nvmem-layout {
-					compatible = "u-boot,env";
-
-					macaddr_ubootenv_ethaddr: ethaddr {};
-				};
 			};
 
 			partition@400000 {
@@ -572,40 +572,30 @@
 				reg = <1>;
 				label = "lan7";
 				phy-handle = <&qca8337_0>;
-				nvmem-cells = <&macaddr_ubootenv_ethaddr>;
-				nvmem-cell-names = "mac-address";
 			};
 
 			port@2 {
 				reg = <2>;
 				label = "lan6";
 				phy-handle = <&qca8337_1>;
-				nvmem-cells = <&macaddr_ubootenv_ethaddr>;
-				nvmem-cell-names = "mac-address";
 			};
 
 			port@3 {
 				reg = <3>;
 				label = "lan5";
 				phy-handle = <&qca8337_2>;
-				nvmem-cells = <&macaddr_ubootenv_ethaddr>;
-				nvmem-cell-names = "mac-address";
 			};
 
 			port@4 {
 				reg = <4>;
 				label = "lan4";
 				phy-handle = <&qca8337_3>;
-				nvmem-cells = <&macaddr_ubootenv_ethaddr>;
-				nvmem-cell-names = "mac-address";
 			};
 
 			port@5 {
 				reg = <5>;
 				label = "lan3";
 				phy-handle = <&qca8337_4>;
-				nvmem-cells = <&macaddr_ubootenv_ethaddr>;
-				nvmem-cell-names = "mac-address";
 			};
 
 			port@6 {
@@ -615,8 +605,6 @@
 				phy-handle = <&qca8033>;
 				managed = "in-band-status";
 				qca,sgmii-enable-pll;
-				nvmem-cells = <&macaddr_ubootenv_ethaddr>;
-				nvmem-cell-names = "mac-address";
 			};
 		};
 	};
@@ -680,60 +668,48 @@
 			phy-handle = <&qca8075_8>;
 			phy-mode = "qsgmii";
 			pcs-handle = <&uniphy0 0>;
-			nvmem-cells = <&macaddr_ubootenv_ethaddr>;
-			nvmem-cell-names = "mac-address";
 		};
 
-		port@2 {
+		port_lan2: port@2 {
 			reg = <2>;
 			label = "lan2";
 			phy-handle = <&qca8075_9>;
 			phy-mode = "qsgmii";
 			pcs-handle = <&uniphy0 1>;
-			nvmem-cells = <&macaddr_ubootenv_ethaddr>;
-			nvmem-cell-names = "mac-address";
 		};
 
-		port@3 {
+		port_lan1: port@3 {
 			reg = <3>;
 			label = "lan1";
 			phy-handle = <&qca8075_a>;
 			phy-mode = "qsgmii";
 			pcs-handle = <&uniphy0 2>;
-			nvmem-cells = <&macaddr_ubootenv_ethaddr>;
-			nvmem-cell-names = "mac-address";
 		};
 
-		port@4 {
+		port_wan: port@4 {
 			reg = <4>;
 			label = "wan";
 			phy-handle = <&qca8075_b>;
 			phy-mode = "qsgmii";
 			pcs-handle = <&uniphy0 3>;
-			nvmem-cells = <&macaddr_ubootenv_ethaddr>;
-			nvmem-cell-names = "mac-address";
 		};
 
-		port@5 {
+		port_10g_sfp: port@5 {
 			reg = <5>;
 			label = "10g-sfp";
 			sfp = <&sfp>;
 			phy-mode = "10gbase-r";
 			managed = "in-band-status";
 			pcs-handle = <&uniphy1 0>;
-			nvmem-cells = <&macaddr_ubootenv_ethaddr>;
-			nvmem-cell-names = "mac-address";
 		};
 
-		port@6 {
+		port_10g_copper: port@6 {
 			reg = <6>;
 			label = "10g-copper";
 			phy-handle = <&aqr113c>;
 			phy-mode = "usxgmii";
 			managed = "in-band-status";
 			pcs-handle = <&uniphy2 0>;
-			nvmem-cells = <&macaddr_ubootenv_ethaddr>;
-			nvmem-cell-names = "mac-address";
 		};
 	};
 };
--- a/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
+++ b/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
@@ -1123,16 +1123,18 @@ static const struct regmap_config edma_regmap_cfg = {
  * whose ports carry the board's addresses either from DT or patched in by
  * the bootloader. DSA user ports without one of their own inherit whatever
  * ends up here.
+ *
+ * A port address is optional here, so no lookup error is fatal: a cell whose
+ * layout driver never binds defers forever, and dsa_port_setup() reads these
+ * same addresses without acting on the error either.
  */
 static int edma_get_mac_address(struct net_device *netdev,
 				struct device_node *np)
 {
 	struct device_node *cpu_port;
-	int ret;
 
-	ret = of_get_ethdev_address(np, netdev);
-	if (!ret || ret == -EPROBE_DEFER)
-		return ret;
+	if (!of_get_ethdev_address(np, netdev))
+		return 0;
 
 	for_each_node_with_property(cpu_port, "ethernet") {
 		struct device_node *conduit __free(device_node) =
@@ -1142,10 +1144,9 @@ static int edma_get_mac_address(struct net_device *netdev,
 			continue;
 
 		for_each_available_child_of_node_scoped(cpu_port->parent, port) {
-			ret = of_get_ethdev_address(port, netdev);
-			if (!ret || ret == -EPROBE_DEFER) {
+			if (!of_get_ethdev_address(port, netdev)) {
 				of_node_put(cpu_port);
-				return ret;
+				return 0;
 			}
 		}
 	}
@@ -1195,11 +1196,10 @@ static int edma_probe(struct platform_device *pdev)
 	priv->pdev = pdev;
 	priv->soc = device_get_match_data(dev);
 
-	ret = edma_get_mac_address(netdev, dev->of_node);
-	if (ret == -EPROBE_DEFER)
-		return dev_err_probe(dev, ret, "failed to get MAC address\n");
-	if (ret)
+	if (edma_get_mac_address(netdev, dev->of_node)) {
+		dev_warn(dev, "failed to get MAC address from DT, using a random one\n");
 		eth_hw_addr_random(netdev);
+	}
 
 	ret = edma_page_pool_create(priv);
 	if (ret)
--- a/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
+++ b/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
@@ -1227,16 +1227,17 @@ static const struct regmap_config edma_regmap_cfg = {
  * whose ports carry the board's addresses either from DT or patched in by
  * the bootloader. DSA user ports without one of their own inherit whatever
  * ends up here.
+ *
+ * No lookup error is fatal here: a cell whose layout driver never binds
+ * defers forever, and the conduit carries every tree behind it.
  */
 static int edma_get_mac_address(struct net_device *netdev,
 				struct device_node *np)
 {
 	struct device_node *cpu_port;
-	int ret;
 
-	ret = of_get_ethdev_address(np, netdev);
-	if (!ret || ret == -EPROBE_DEFER)
-		return ret;
+	if (!of_get_ethdev_address(np, netdev))
+		return 0;
 
 	for_each_node_with_property(cpu_port, "ethernet") {
 		struct device_node *conduit __free(device_node) =
@@ -1246,10 +1247,9 @@ static int edma_get_mac_address(struct net_device *netdev,
 			continue;
 
 		for_each_available_child_of_node_scoped(cpu_port->parent, port) {
-			ret = of_get_ethdev_address(port, netdev);
-			if (!ret || ret == -EPROBE_DEFER) {
+			if (!of_get_ethdev_address(port, netdev)) {
 				of_node_put(cpu_port);
-				return ret;
+				return 0;
 			}
 		}
 	}
@@ -1299,11 +1299,10 @@ static int edma_probe(struct platform_device *pdev)
 	priv->pdev = pdev;
 	priv->soc = device_get_match_data(dev);
 
-	ret = edma_get_mac_address(netdev, dev->of_node);
-	if (ret == -EPROBE_DEFER)
-		return dev_err_probe(dev, ret, "failed to get MAC address\n");
-	if (ret)
+	if (edma_get_mac_address(netdev, dev->of_node)) {
+		dev_warn(dev, "failed to get MAC address from DT, using a random one\n");
 		eth_hw_addr_random(netdev);
+	}
 
 	priv->rx_page_order = edma_rx_page_order(netdev->mtu);
 	priv->rx_buffer_size = edma_rx_buffer_size(priv->rx_page_order);


target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
```diff
@@ -1123,16 +1123,18 @@ static const struct regmap_config edma_regmap_cfg = {
  * whose ports carry the board's addresses either from DT or patched in by
  * the bootloader. DSA user ports without one of their own inherit whatever
  * ends up here.
+ *
+ * A port address is optional here, so no lookup error is fatal: a cell whose
+ * layout driver never binds defers forever, and dsa_port_setup() reads these
+ * same addresses without acting on the error either.
```
openwrt-ai: The nvmem analysis checks out — [`nvmem_layout_module_get_optional()` at core.c:1420-1431](https://github.com/gregkh/linux/blob/52a355b23cd2dd77b683ed1a1bbe2b7408bd9c74/drivers/nvmem/core.c#L1420-L1431) does return `-EPROBE_DEFER` unconditionally once a layout device exists whose driver never bound, so that state is genuinely permanent.

What the change also loses, though, is the transient case: an nvmem/MTD provider that simply hasn't registered yet now yields a random conduit address instead of a retry, silently and permanently. The justification in the commit message is a timing measurement on one board ("MTD partitions register at 1.20s, conduit probes at 2.37s"), but probe ordering isn't guaranteed across targets/subtargets, and this driver serves all of ipq60xx/ipq807x — a board whose SPI-NAND/QPIC stack comes up later would regress from "correct MAC after a deferral" to "random MAC forever", with no message.

Is that trade considered acceptable, or would it be worth narrowing to the permanent case — e.g. keep propagating `-EPROBE_DEFER` but only until the nvmem provider device for the referenced cell actually exists? At minimum the fallback should be logged (see the comment on the probe hunk).

---

target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
```diff
@@ -1195,10 +1196,7 @@ static int edma_probe(struct platform_device *pdev)
 	priv->pdev = pdev;
 	priv->soc = device_get_match_data(dev);
 
-	ret = edma_get_mac_address(netdev, dev->of_node);
-	if (ret == -EPROBE_DEFER)
-		return dev_err_probe(dev, ret, "failed to get MAC address\n");
-	if (ret)
+	if (edma_get_mac_address(netdev, dev->of_node))
 		eth_hw_addr_random(netdev);
```
openwrt-ai: nit: the driver now falls back to a random address with no trace at all. The old code at least emitted `deferred probe pending: qca-edma: failed to get MAC address`, which is exactly the string that made #24604 diagnosable. After this change the same class of breakage (a cell that never resolves) shows up only as "my MACs are wrong", with nothing in the log to point at DT/nvmem.

```suggestion
	if (edma_get_mac_address(netdev, dev->of_node)) {
		dev_warn(dev, "no MAC address from DT, using a random one\n");
		eth_hw_addr_random(netdev);
	}
```

---

target/linux/qualcommax/dts/ipq8074-rt-ax89x.dts
```diff
@@ -17,6 +17,12 @@
 	aliases {
 		serial0 = &blsp1_uart5;
 		mdio-gpio0 = &mdio1;
+		ethernet0 = &port_switch;
```
openwrt-ai, line 20: nit: the commit message (and the PR body) state that `fdt_fixup_ethernet()` "walks /aliases from ethernet0 and stops at the first absent ethNaddr". It doesn't stop — a missing env var is skipped and the walk carries on to the next alias: [`boot/fdt_support.c:692-694`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/boot/fdt_support.c#L692-L694) does `tmp = env_get(mac); if (!tmp) continue;`. Same in the older U-Boot bases the QCA vendor loaders derive from.

The outcome you describe holds as long as the `eth*addr` vars are contiguous, so this doesn't change the alias mapping — but if this board's bootloader sets e.g. `ethaddr` and `eth3addr` only, ports 4/5/6 would get real addresses too, not inherit the conduit's. Worth rewording so git history doesn't carry the "stops at" claim.

---

target/linux/qualcommax/dts/ipq8074-rt-ax89x.dts
```diff
@@ -684,7 +690,7 @@
 			nvmem-cell-names = "mac-address";
```
openwrt-ai: The commit message says both "that cell resolves to nothing on this board" and "the nvmem cells stay as the fallback for a bootloader that patches nothing". Those can't both be true for this board — if `macaddr_ubootenv_ethaddr` never resolves, it isn't a fallback, it's a dead reference on all twelve ports (here and on the qca8337 ports at lines 581/589/597/605/613/624).

Has the underlying layout failure been chased down? `CONFIG_NVMEM_LAYOUT_U_BOOT_ENV=y` is set for this target ([config-6.12:293](https://github.com/openwrt/openwrt/blob/f669d5278d4b0d6d5ae134dbec60319fb549b29e/target/linux/qualcommax/config-6.12#L293)), so the layout driver is built in and would normally bind — a permanent `-EPROBE_DEFER` therefore means the `u-boot-env` driver's *probe failed* on [`0:appsblenv`](https://github.com/openwrt/openwrt/blob/f669d5278d4b0d6d5ae134dbec60319fb549b29e/target/linux/qualcommax/dts/ipq8074-rt-ax89x.dts#L456-L467), most likely a CRC/format mismatch (ASUS storing a redundant env, i.e. `u-boot,redundant-env`, or the env not spanning the whole 128K partition). Fixing that would make the cells actually work and give the non-`ethaddr` ports their own vendor addresses, rather than relying on the bootloader's fdt fixups; the `dmesg` from #24604 should show the `u_boot_env` probe error. If it turns out the layout genuinely can't work here, the honest move is to drop the dead `nvmem-cells`/`nvmem-cell-names` pairs rather than describe them as a fallback.

---

openwrt-ai (commented):
Reviewed 2 new commits. The alias-to-port mapping in the second patch checks out against the pre-PPE DTS (`dp1`/`qca8075_8` → `port@1`, ... `dp6_syn`/`aqr113c` → `port@6`), and the `of_nvmem_cell_get()` reasoning in the first one matches 6.12 (the kernel series this target builds). Four comments inline: one question on dropping the deferral for the transient-provider case, one on the still-unexplained `u-boot,env` layout failure that is the actual root cause, plus two nits.

---

Razerwire:
> `edma_probe()` propagates `-EPROBE_DEFER` out of the conduit's MAC address lookup, but a DSA port address is optional, so the whole datapath ends up hostage to a supplier that may never arrive. `of_nvmem_cell_get()` answers `-EPROBE_DEFER` both while a provider is missing and, permanently, once a provider is registered whose layout driver never bound or never produced the named cell:
> 
> ```c
> 	ret = nvmem_layout_module_get_optional(nvmem);	/* layout driver never bound */
> 	...
> 	cell_entry = nvmem_find_cell_entry_by_node(nvmem, cell_np);
> 	if (!cell_entry) {
> 		ret = nvmem->layout ? -EPROBE_DEFER : -ENOENT;	/* named cell never produced */
> ```
> 
> On the RT-AX89X every port points at one `u-boot,env` cell that lands in that state, so the conduit never probes: no `eth0`, no DSA, no ports (#24604). The board boots off the same MTD, so its partitions did register and the provider does exist - the layout is what failed.
> 
> Nothing else in the chain treats these addresses as mandatory. `dsa_port_setup()` reads the very same nodes and acts on no error:
> 
> ```c
> 	case DSA_PORT_TYPE_USER:
> 		of_get_mac_address(dp->dn, dp->mac);
> 		err = dsa_user_create(dp);
> ```
> 
> and the nss-dp driver this one replaced dropped the return value before falling back to a random address. The first patch does the same. The removed branch bought nothing either way: on an AX3600 the MTD partitions carrying the cells register at 1.20s and the conduit probes at 2.37s, so the lookup has never had to wait.
> 
> That restores the ports, but on a random address. The second patch gets the board a real one. It carried an `ethernet0..5` alias for each of the six ESS ports so the bootloader could patch their addresses in; [f504356](https://github.com/openwrt/openwrt/commit/f50435627d370a1bb07d455d7081207b7b6a744e) dropped the block and described one shared cell on every port instead. The aliases now point at the ports that replaced the dp nodes, matched on the PHY each one drives:
> alias 	was 	PHY 	now
> ethernet0 	`dp1` "switch" 	`qca8075_8` 	`port@1`
> ethernet1 	`dp2` "lan2" 	`qca8075_9` 	`port@2`
> ethernet2 	`dp3` "lan1" 	`qca8075_a` 	`port@3`
> ethernet3 	`dp4` "wan" 	`qca8075_b` 	`port@4`
> ethernet4 	`dp5_syn` "10g-sfp" 	`sfp` 	`port@5`
> ethernet5 	`dp6_syn` "10g-copper" 	`aqr113c` 	`port@6`
> 
> `fdt_fixup_ethernet()` walks `/aliases` from `ethernet0` and stops at the first absent `ethNaddr`, writing `local-mac-address` with create set. A port it reaches gets the vendor address; the rest inherit the conduit's, which is now whatever the bootloader patched into `port@1` rather than a random value. Same shape as the 301w in #24528, where the walk stops after two. The nvmem cells stay as the fallback.
> 
> Verified on an AX3600 (ipq8071): `eth0` still comes up as the `0:art` address the WAN port uses, no deferred probe pending, ports and WAN unaffected.
> 
> I have no RT-AX89X. @Razerwire - this needs a test on the device, and `fw_printenv ethaddr` would tell us how many ports the bootloader reaches.
> 
> @Ansuel @robimarko

I'll have a look at my uboot output tomorrow, and report back👌

Razerwire:
> I have no RT-AX89X. @Razerwire - this needs a test on the device, and `fw_printenv ethaddr` would tell us how many ports the bootloader reaches.

@JuliusBairaktaris 
Alright, I've had a look on both  RT-AX89X's.
 U-boot only (directly)  references _two_ MAC-address that are used for recovery.
 These MAC-Addresses have **nothing** to  to do with the actual MAC-addresses of physical ports: 
 ```
root@RT-AX89X:~# fw_printenv
Warning: Bad CRC, using default environment
bootcmd=run distro_bootcmd
bootdelay=2
baudrate=115200
loadaddr=0x0
mtdids=
mtdparts=
bootm_size=0x10000000
eth6addr=02:00:11:22:33:47
ethaddr=02:00:11:22:33:44
fdt_addr_r=0xc00000
ipaddr=192.0.2.1
ipaddr2=192.0.2.3
ipaddr3=192.0.2.4
ipaddr5=192.0.2.6
ipaddr6=192.0.2.7
ipaddr7=192.0.2.8
kernel_addr_r=0x1000000
pxefile_addr_r=0x2000
ramdisk_addr_r=0x2000000
scriptaddr=0x1000
stderr=serial,vidconsole
stdin=serial
stdout=serial,vidconsole

 

``` 
@robimarko @Ansuel 
I am very much 'out of my league here', however, If someone was to ask me about my opinion:

The QCA-SSDK is very much needed for the reading of MAC-Addresses for the RT-AX89X.

(As an example: the OEM-Firmware [from Asus] contains a Shell Script, for PHY-debugging, the first thing the script does is call the SSDK......):
```
set_reg_data() {
	ret=`ssdk_sh debug phy set $1 $2 $3`
	[ "$?" != "0" ] && echo "fail to set data:$3 to reg:$2"
}

get_reg_data() {
	reg_data=`ssdk_sh debug phy get $1 $2 | grep [Data] | awk -F ':' '{print $2}'`
	#echo $reg_data | awk '{print int($0)}'
	echo $reg_data | awk '{printf("%d\n", $0)}'
}

printfmt() {
	printf "0x%x" $1
}

#read reg rangs $1: phy_addr $2: first_regaddr, $3:increment, $4:last_regaddr
get_reg_rang() {
	phy_addr=$1
	for reg_addr in $(seq $2 $3 $4)
	do
		reg_data=$(get_reg_data $phy_addr $reg_addr)
		echo "the reg_data[$(printfmt $reg_addr)]: $(printfmt $reg_data)"
	done
}
```


Please let me know if further info is needed, I could hook up a serial cable if needed,  etc....

**Comment by Razerwire** [comment]
> These MAC-Addresses have **nothing** to to do with the actual MAC-addresses of physical ports:

@sppmasterspp

It seems OpenWrt might be overwriting  the 'stock' ?  u-boot-environment-variables for the RT-AX89X (including, but not limited to, the list of OEM/factory MAC-Addresses) . 


Stock firmware _does seem_  to print  a list of valid (Asus) MAC-Addresses for the uboot-env, (just  like the 301W does and what @JuliusBairaktaris was referring to)  
([Source](https://www.snbforums.com/threads/asus-rt-ax89x-b2-is-boot-looping-bricked-or-hardware-fault.91293/))

(note to myself:  remember when you tried to do TFTP recovery via the blue WAN-port and it doing nothing, _perhaps_ there is a relation > is this  related to #23633 ?)


Haven't verified this for myself, will look into this and report back.

File: target/linux/qualcommax/dts/ipq8074-rt-ax89x.dts [hunk 5]
```diff
@@ -17,6 +17,12 @@
 	aliases {
 		serial0 = &blsp1_uart5;
 		mdio-gpio0 = &mdio1;
+		ethernet0 = &port_switch;
+		ethernet1 = &port_lan2;
+		ethernet2 = &port_lan1;
+		ethernet3 = &port_wan;
+		ethernet4 = &port_10g_sfp;
+		ethernet5 = &port_10g_copper;
```
openwrt-ai, line 25: Before this mapping gets validated against the `fw_printenv` dump posted in the thread: that output is not the board's environment. Every variable in it is U-Boot's **sandbox** default environment, verbatim and in order — [`board/sandbox/sandbox.env:8-10`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/board/sandbox/sandbox.env#L8-L10) is exactly `ethaddr=02:00:11:22:33:44` / `eth6addr=02:00:11:22:33:47` / `ipaddr=192.0.2.1`, and the rest of that same file supplies `ipaddr2/3/5/6/7`, `stdin`/`stdout`/`stderr`, `bootm_size`, `kernel_addr_r`, `fdt_addr_r`, `ramdisk_addr_r`, `scriptaddr` and `pxefile_addr_r`. That is the fallback `fw_printenv` prints when it cannot validate what it read, which is what its `Warning: Bad CRC, using default environment` line is saying.

So it neither confirms nor refutes the `ethernet0..5` mapping, and the `eth6addr` in it is a sandbox artifact — not evidence of a seventh port needing an `ethernet6` alias here. The only thing that establishes what `fdt_fixup_ethernet()` will actually patch is `printenv` at the board's own U-Boot prompt; @Razerwire offered a serial cable, and that is the measurement worth taking before this patch's premise can be checked.

What it *is* good evidence for is the root cause behind patch 1 — the question still open on the `port@1` hunk. The env config for this board is not missing or guessed: [`qualcommax_ipq807x:38-41`](``https://github.com/openwrt/openwrt/blob/9f5a574c7924db138d29171a11f8827e4288d9fe/package/boot/uboot-tools/uboot-envtools/files/qualcommax_ipq807x#L38-L41``) maps `asus,rt-ax89x` to `0:appsblenv` at offset 0 with size `0x20000`, i.e. the whole partition — the same span and the same single-image assumption the kernel layout driver makes for [`compatible = "u-boot,env"` at ipq8074-rt-ax89x.dts:461-464](https://github.com/openwrt/openwrt/blob/9f5a574c7924db138d29171a11f8827e4288d9fe/target/linux/qualcommax/dts/ipq8074-rt-ax89x.dts#L461-L464). If userspace gets a CRC mismatch over that range, [`u-boot-env.c:151-155`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/nvmem/layouts/u-boot-env.c#L151-L155) gets the same one and bails with `-EINVAL`, so no cells are ever added and [`core.c:1476`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/nvmem/core.c#L1476) hands back a permanent `-EPROBE_DEFER` — exactly the state patch 1 works around. (Checked against 6.18, the target's current `KERNEL_PATCHVER`; identical on 6.12.)

The `#24604` dmesg should therefore carry a literal `Invalid calculated CRC32: 0x... (expected: 0x...)` from `u_boot_env`. If it does, that is worth putting in the commit message in place of "that cell resolves to nothing" — and it points at a real fix, since ASUS storing a redundant env (`u-boot,env-redundant-bool`/`-count`, one flag byte before the data) would produce precisely this mismatch under the plain `u-boot,env` compatible, and correcting it would make the cells work and give the non-`ethaddr` ports vendor addresses without depending on bootloader fixups at all.

---

target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
```diff
@@ -1195,11 +1196,10 @@ static int edma_probe(struct platform_device *pdev)
 	priv->pdev = pdev;
 	priv->soc = device_get_match_data(dev);
 
-	ret = edma_get_mac_address(netdev, dev->of_node);
-	if (ret == -EPROBE_DEFER)
-		return dev_err_probe(dev, ret, "failed to get MAC address\n");
-	if (ret)
+	if (edma_get_mac_address(netdev, dev->of_node)) {
+		dev_warn(dev, "no MAC address in DT, using a random one\n");
```
openwrt-ai: nit: "no MAC address in DT" describes the one case this warning will rarely fire for. The breakage the commit message is about — and the RT-AX89X specifically — is a DT that *does* describe an address through `nvmem-cells`, which then fails to resolve; `edma_get_mac_address()` collapses both into the same non-zero return, so this string sends the reader looking for a missing property when the property is right there. Naming the operation rather than guessing the cause keeps the trace true to both paths:

```suggestion
		dev_warn(dev, "failed to get MAC address from DT, using a random one\n");
```

---

openwrt-ai (commented):
Reviewed the changes since f669d52. The only code delta is the `dev_warn` on the fallback path, and the second commit's `fdt_fixup_ethernet()` paragraph was reworded — both address the earlier nits, and both commit messages still match their diffs. CI is green on 9f5a574.

The substantive finding this round comes from the thread rather than the diff: the `fw_printenv` dump posted on Aug 15 is U-Boot's sandbox default environment verbatim, not this board's, so it can't be used to validate the alias mapping — but the bad-CRC condition behind it is good evidence for the root cause. Details inline on the aliases block, plus one nit on the new warning string. My two earlier questions (the transient-deferral trade-off, and the `nvmem-cells` described as a fallback while said to resolve to nothing) are still open.

---

target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
JuliusBairaktaris: > What the change also loses, though, is the transient case: an nvmem/MTD provider that simply hasn't registered yet now yields a random conduit address instead of a retry, silently and permanently.

I agree, and I take the trade. A provider that really is late now costs a random
address instead of a retry.

The caller cannot tell the two states apart. `of_nvmem_cell_get()` returns the
same `-EPROBE_DEFER` for "the layout driver has not probed yet" and "the layout
driver probed and failed":

    drivers/nvmem/core.c, nvmem_layout_module_get_optional():

        if (!nvmem->layout->dev.driver ||
            !try_module_get(nvmem->layout->dev.driver->owner))
                return -EPROBE_DEFER;

DSA has since gone the other way for ports, in 679b73b1884a:

    target/linux/generic/pending-6.18/
    704-net-dsa-support-EPROBE_DEFER-when-reading-the-port-M.patch

        err = of_get_mac_address(dp->dn, dp->mac);
        if (err == -EPROBE_DEFER)
                break;

My understanding is that this is right for a port and wrong for a conduit. A
port's address is its own, so a deferral costs that one port until its supplier
appears. A conduit borrows an address from ports it does not own, so a deferral
there costs `eth0` and every DSA tree behind it while it waits on one port's
supplier - and in the permanent state that supplier never appears.

I added the log line you asked for in the probe-hunk comment, using the later
correction to the string: `failed to get MAC address from DT, using a random
one`.

target/linux/qualcommax/dts/ipq8074-rt-ax89x.dts
JuliusBairaktaris: > If it turns out the layout genuinely can't work here, the honest move is to drop the dead `nvmem-cells`/`nvmem-cell-names` pairs rather than describe them as a fallback.

You are right that the two statements cannot both hold, so patch 2 now drops all
twelve `nvmem-cells`/`nvmem-cell-names` pairs and the `u-boot,env`
`nvmem-layout` node with them.

The failure is the CRC you pointed at: the environment in `0:appsblenv` fails
its check, so the layout driver adds no cells at all.

    drivers/nvmem/layouts/u-boot-env.c, u_boot_env_parse():

        calc = crc32_le(~0, buf + crc32_data_offset, crc32_data_len) ^ ~0L;
        if (calc != crc32) {
                dev_err(dev, "Invalid calculated CRC32: 0x%08x (expected: 0x%08x)\n", calc, crc32);
                err = -EINVAL;

The bad CRC itself is measured: `fw_printenv` on the device prints
`Warning: Bad CRC, using default environment` over the same span the kernel
layout covers, `0:appsblenv` offset 0 size `0x20000`. I have not seen the
kernel-side half yet, so I have asked for the dmesg that would carry
`Invalid calculated CRC32` from `u_boot_env`.

Dropping the cells is also what 679b73b1884a now requires. A user port whose
address lookup answers `-EPROBE_DEFER` tears its DSA tree down, so the six
qca8337 ports - which have no alias and nothing else to read - would take that
switch down permanently.

If the environment turns out to be a redundant image rather than genuinely
invalid, the cells can come back under the right compatible in a follow-up.

target/linux/qualcommax/dts/ipq8074-rt-ax89x.dts
JuliusBairaktaris: > Every variable in it is U-Boot's **sandbox** default environment, verbatim and in order

Confirmed, and it falls out of how OpenWrt builds the tool:

    package/boot/uboot-tools/Makefile:

        define Build/Configure
                $(call Build/Compile/Default,tools-only_defconfig)
        [...]
        MAKE_FLAGS += \
                ARCH="sandbox" \

That configuration compiles the sandbox board's default environment into
`fw_printenv`, which is what it prints after `Bad CRC` - so the dump carries no
information about this board, `eth6addr` included.

What it does establish is the bad CRC itself. The DTS commit keeps "on this
board it resolves to nothing" and now gives the CRC as the reason for it:
`fw_printenv` reports a bad CRC over the whole of `0:appsblenv`, the span the
layout driver reads, so it adds no cells. The driver commit does not mention the
CRC; it says only that the cell lands in the permanent defer state.

The partition can still be read from the running system without a cable, and I
have asked @Razerwire for that in a separate comment. It narrows the question
rather than settling it: if the board's own U-Boot fails the same CRC, the
variable names in the partition say nothing about what it patches.

JuliusBairaktaris:
@Razerwire - that dump is U-Boot's sandbox default environment, which is what
`fw_printenv` falls back to when the CRC check fails, so it says nothing about
this board. Two things are still open, and both can be answered from the running
OpenWrt: whether the CRC failure is a format mismatch fixable in DT - a
redundant environment carries one flag byte before the data, which shifts the
image under the plain `u-boot,env` compatible - or an environment that is simply
not valid; and which `ethNaddr` variables the environment holds.

Could you paste the output of these, as root? None of them prints a MAC address.

    cat /proc/mtd

    cat /etc/fw_env.config

    { dmesg; logread; } | grep -iE 'u_boot_env|crc32'

    hexdump -C -n 5 /dev/mtd$(sed -n 's/^mtd\([0-9]*\):.*"0:appsblenv".*/\1/p' /proc/mtd)

    strings /dev/mtd$(sed -n 's/^mtd\([0-9]*\):.*"0:appsblenv".*/\1/p' /proc/mtd) | grep -oE 'eth[0-9]*addr'

The third should carry a literal `Invalid calculated CRC32: 0x... (expected:
0x...)` from `u_boot_env` if the kernel-side layout fails the same check
userspace does. The fourth is the environment header: four bytes of CRC, then
either the flag byte of a redundant image or the first character of a variable
name.

One caveat: if the header comes back blank, the board's own U-Boot fails that
CRC too and boots on its compiled-in defaults, in which case the variable names
in the partition say nothing about what `fdt_fixup_ethernet()` will patch. If it
lands that way I will send you a test build with the aliases in it, and
`find /sys/firmware/devicetree/base -name mac-address` on the running system
answers it directly.

What is not settled: the bootloader's own `printenv` has not been read, so which
`ethNaddr` variables exist is unknown, and a port the bootloader does not name
inherits the conduit's address.

target/linux/qualcommax/dts/ipq8074-rt-ax89x.dts
openwrt-ai, line 20: nit: the commit message says the bootloader ends up "writing mac-address and local-mac-address on each port it does find one for", but `mac-address` is fixed up with `create = 0` — [`do_fixup_by_path(fdt, path, "mac-address", &mac_addr, 6, 0)` at boot/fdt_support.c:703-706](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/boot/fdt_support.c#L703-L706) — so it is only rewritten where the property already exists. After this patch no port node in this DTS carries one, so `local-mac-address` is the only property actually created; worth correcting so git history doesn't carry the stronger claim.

---

openwrt-ai (commented):
Reviewed 2 new commits.

---

PR #24664, opened 2026-08-11
uboot-airoha: drop AN7583 NPU reserved-memory nodes

FireEmblem-purple:
This adds a follow-up patch for the AN7583 U-Boot device tree.

The current device tree reserves five regions for NPU firmware and
buffers. U-Boot does not use these regions, so the patch removes the
NPU reserved-memory nodes while retaining the region reserved for
TF-A.

The removed regions total 39 MiB.

Testing:

- Target: airoha/an7583
- U-Boot variant: an7583_rfb
- `package/boot/uboot-airoha/compile`: passed
- Hardware boot test: passed on Airoha AN7583 EVB

Upstream status:

- Not yet submitted to U-Boot upstream

### DIFF
--- a/package/boot/uboot-airoha/patches/908-arm-dts-an7583-drop-npu-reserved-memory.patch
+++ b/package/boot/uboot-airoha/patches/908-arm-dts-an7583-drop-npu-reserved-memory.patch
@@ -1,0 +1,50 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Li Xiao <Li.Xiao@airoha.com>
+Date: Tue, 11 Aug 2026 15:57:43 +0800
+Subject: [PATCH] arm: dts: an7583: drop NPU reserved-memory nodes
+
+U-Boot does not use the AN7583 NPU firmware and buffer regions.
+
+Keeping these entries in the U-Boot control device tree unnecessarily
+reserves 39 MiB of DRAM. Drop the NPU reserved-memory nodes while
+retaining the memory region reserved for TF-A.
+
+Signed-off-by: Li Xiao <Li.Xiao@airoha.com>
+---
+ arch/arm/dts/an7583.dtsi |   25 -------------------------
+ 1 file changed, 25 deletions(-)
+
+--- a/arch/arm/dts/an7583.dtsi
++++ b/arch/arm/dts/an7583.dtsi
+@@ -21,31 +21,6 @@
+ 			no-map;
+ 			reg = <0x0 0x80000000 0x0 0x40000>;
+ 		};
+-
+-		npu-binary@84000000 {
+-			no-map;
+-			reg = <0x0 0x84000000 0x0 0xa00000>;
+-		};
+-
+-		npu-flag@84b0000 {
+-			no-map;
+-			reg = <0x0 0x84b00000 0x0 0x100000>;
+-		};
+-
+-		npu-pkt@85000000 {
+-			no-map;
+-			reg = <0x0 0x85000000 0x0 0x1a00000>;
+-		};
+-
+-		npu-phyaddr@86b00000 {
+-			no-map;
+-			reg = <0x0 0x86b00000 0x0 0x100000>;
+-		};
+-
+-		npu-rxdesc@86d00000 {
+-			no-map;
+-			reg = <0x0 0x86d00000 0x0 0x100000>;
+-		};
+ 	};
+ 
+ 	psci {


danielschwierzeck:
Reviewed-by: Daniel Schwierzeck <daniel.schwierzeck@gmail.com>

package/boot/uboot-airoha/patches/908-arm-dts-an7583-drop-npu-reserved-memory.patch
```diff
@@ -0,0 +1,46 @@
...
+U-Boot does not use the AN7583 NPU firmware and buffer regions.
+
+Keeping these entries in the U-Boot control device tree unnecessarily
+reserves 39 MiB of DRAM. Drop the NPU reserved-memory nodes while
+retaining the memory region reserved for TF-A.
+
+Signed-off-by: Li Xiao <Li.Xiao@airoha.com>
+--- a/arch/arm/dts/an7583.dtsi
++++ b/arch/arm/dts/an7583.dtsi
+@@ -21,31 +21,6 @@
```
openwrt-ai, line 19: Inserting this patch at slot `302` puts it ahead of the five other patches in the series that touch `arch/arm/dts/an7583.dtsi`, and it deletes 25 lines near the top of that file — so every one of them now applies with a stale offset. I reconstructed `an7583.dtsi` from `300-airoha-add-support-for-Airoha-AN7583-SoC.patch` and replayed the series with `patch --fuzz=0`:

```
=== 308-airoha-add-Ethernet-node-in-AN7583-dtsi ===
Hunk #1 succeeded at 105 (offset -25 lines).
=== 309-airoha-add-MMC-node-for-Airoha-AN7583 ===
Hunk #1 succeeded at 80 (offset -25 lines).
Hunk #2 succeeded at 249 (offset -25 lines).
=== 510-arm-dts-an7583-add-mdio-child-node-to-switch-node ===
Hunk #1 succeeded at 141 (offset -25 lines).
=== 521-airoha-add-PCS-node-for-AN7583 ===
Hunk #1 succeeded at 120 (offset -25 lines).
Hunk #2 succeeded at 179 (offset -25 lines).
=== 531-an7583-fix-system-controller-mdio-nodes ===
Hunk #1 succeeded at 235 (offset -25 lines).
```

Nothing fails (no fuzz), but the hunk headers of [308 `@@ -130,6 +130,29 @@`](``https://github.com/openwrt/openwrt/blob/294bd0aa6fa44a43778e8549d61390e8d3ba8e2f/package/boot/uboot-airoha/patches/308-airoha-add-Ethernet-node-in-AN7583-dtsi.patch#L18),`` 309, 510, 521 and 531 are now wrong. Two ways out:

- Refresh the series so the downstream hunk headers are regenerated (`make package/uboot-airoha/refresh`), or
- Renumber this patch to sit after the existing `an7583.dtsi` patches (e.g. `533-`), which leaves the rest of the series untouched. That also keeps it out of the `300`–`309` block, which currently holds the imported `[PATCH nn/29]` AN7583 series — slot `302` only became free because `302-reset-airoha-convert-to-regmap-API.patch` was dropped in 6b7fb8e9c2 when it landed upstream. Given the PR body says this one is "Not yet submitted to U-Boot upstream", the local `5xx`/`9xx` range looks like the better home.

---

package/boot/uboot-airoha/patches/908-arm-dts-an7583-drop-npu-reserved-memory.patch
```diff
@@ -0,0 +1,46 @@
...
+Subject: [PATCH] arm: dts: an7583: drop NPU reserved-memory nodes
+
+U-Boot does not use the AN7583 NPU firmware and buffer regions.
+
+Keeping these entries in the U-Boot control device tree unnecessarily
+reserves 39 MiB of DRAM. Drop the NPU reserved-memory nodes while
+retaining the memory region reserved for TF-A.
+
+Signed-off-by: Li Xiao <Li.Xiao@airoha.com>
+--- a/arch/arm/dts/an7583.dtsi
```
openwrt-ai, line 17: nit: the `---` separator and diffstat are missing between the `Signed-off-by:` line and the diff body. Every other patch in this directory carries them, including the locally-authored ones such as [`998-airoha-add-snfi-label.patch`](``https://github.com/openwrt/openwrt/blob/294bd0aa6fa44a43778e8549d61390e8d3ba8e2f/package/boot/uboot-airoha/patches/998-airoha-add-snfi-label.patch#L7-L11).``

```suggestion
---
 arch/arm/dts/an7583.dtsi | 25 -------------------------
 1 file changed, 25 deletions(-)

--- a/arch/arm/dts/an7583.dtsi
```

Related, on line 1: the `From` line carries an all-zero SHA. Every other patch here (e.g. `301`, `308`, `510`, `531`, `998`) has the real author-commit hash, which is what `git format-patch` emits. If there is no backing commit, dropping the `From <sha> Mon Sep 17 ...` line altogether is cleaner than a placeholder hash.

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the diff (5 NPU nodes dropped, `0xa00000 + 0x100000 + 0x1a00000 + 0x100000 + 0x100000` = 39 MiB, `atf-reserved-memory@80000000` retained), and the rationale checks out: U-Boot's LMB does consume the *control* DT's `reserved-memory` (`lmb_init()` → `lmb_reserve_common((void *)gd->fdt_blob)`), and the OS device tree's own reservations are re-applied at `bootm` time via `boot_fdt_add_mem_rsv_regions()`, so freeing this window in the U-Boot DT does not lose the kernel-side carveouts (`target/linux/airoha/dts/an7583.dtsi` still reserves `npu-binary@84000000` and the qdma buffers).

Two comments inline, both about patch-series hygiene rather than the DT change itself.

---

package/boot/uboot-airoha/patches/908-arm-dts-an7583-drop-npu-reserved-memory.patch
FireEmblem-purple: Moved the patch to slot 908 after the existing AN7583 DTS patches, so the downstream patches no longer apply with stale offsets.

package/boot/uboot-airoha/patches/908-arm-dts-an7583-drop-npu-reserved-memory.patch
FireEmblem-purple: Regenerated the patch，It now contains the separator and diffstat

FireEmblem-purple: The all-zero SHA is intentionally retained because this is a locally
generated quilt patch without a backing U-Boot Git commit. The OpenWrt
formality checks accept this header format.

peterwillcn (approved):

openwrt-ai (commented):

Since `294bd0a` the only changes are the rename `302-` → `908-` and the added `---` separator + diffstat, which address both earlier comments. Re-verified:

- **Patch order.** Reconstructed `arch/arm/dts/an7583.dtsi` from `300-airoha-add-support-for-Airoha-AN7583-SoC.patch` and replayed every patch touching it in series order (`300`, `308`, `309`, `510`, `521`, `531`, `908`) with `patch --fuzz=0`. All hunks apply at their stated line numbers with no offsets and no fuzz — slot `908` sits after all existing `an7583.dtsi` patches, so the downstream hunk headers are no longer stale. `908` is also free between `904` and `998`.
- **Diffstat.** `25 deletions(-)` matches the single hunk (`@@ -21,31 +21,6 @@` → 31 − 6 = 25) and the dash run is 25 wide.
- **Commit message.** Identical to the previously reviewed one apart from the added `Reviewed-by:` trailer; subject scope `uboot-airoha:` matches the only file touched, and `an7583_rfb` is a real variant (`define U-Boot/an7583_rfb`, `UBOOT_CONFIG:=an7583_evb`).
- The all-zero `From` SHA is left as-is per the author's explanation that this is a locally authored patch with no backing U-Boot commit.

The bot's `PKG_RELEASE` warning looks like a false positive here: `package/boot/uboot-airoha/Makefile` has never carried a `PKG_RELEASE`, same as `uboot-mediatek`, `uboot-ath79`, `uboot-sunxi` and `uboot-zynq`.

CI on `3cae5aa` is green (all three FormalityCheck runs succeeded).

---

YaleiZang (approved):

PR #24665, opened 2026-08-11
mediatek: update ASUS TUF-AX6000 LED and device tree cleanups

DH-41:
### Summary
This PR includes LED trigger updates and minor device tree cleanups for the ASUS TUF-AX6000.

### Changes
1. LED Improvement:
   * Replaced the temporary single-phy workaround with the OpenWrt `"network"` LED trigger.
   * Added `kmod-ledtrig-network` to DEVICE_PACKAGES in filogic.mk.
   * The WLAN LED now correctly reflects activity on both `phy0` and `phy1`.
   * Refactored the LED definitions to use the modern LED naming convention.
   (Note: this renames the sysfs LED from `white:system` to `white:status`)
2. Device Tree Cleanups:
   * Formatted `wf_dbdc_pins` to match the alignment of `wf_2g_5g-pins`.
   * Removed unnecessary `#address-cells` and `#size-cells` properties from `flash@0`.
   * Removed the unused `&pcie_phy` status override (`status = "okay";`).
   * Cleaned up unused labels and consolidated items into their respective nodes.

### DIFF
--- a/target/linux/mediatek/dts/mt7986a-asus-tuf-ax6000.dts
+++ b/target/linux/mediatek/dts/mt7986a-asus-tuf-ax6000.dts
@@ -54,11 +54,12 @@
 			function = LED_FUNCTION_WLAN;
 			color = <LED_COLOR_ID_WHITE>;
 			gpios = <&pio 2 GPIO_ACTIVE_HIGH>;
-			linux,default-trigger = "phy1tpt";
+			linux,default-trigger = "network";
 		};
 
 		led_system: system {
-			label = "white:system";
+			function = LED_FUNCTION_STATUS;
+			color = <LED_COLOR_ID_WHITE>;
 			gpios = <&pio 11 GPIO_ACTIVE_HIGH>;
 		};
 
@@ -69,17 +70,20 @@
 		};
 
 		cover-blue {
-			label = "blue:cover";
+			function = "cover";
+			color = <LED_COLOR_ID_BLUE>;
 			gpios = <&pio 20 GPIO_ACTIVE_HIGH>;
 		};
 
 		cover-red {
-			label = "red:cover";
+			function = "cover";
+			color = <LED_COLOR_ID_RED>;
 			gpios = <&pio 21 GPIO_ACTIVE_HIGH>;
 		};
 
 		cover-green {
-			label = "green:cover";
+			function = "cover";
+			color = <LED_COLOR_ID_GREEN>;
 			gpios = <&pio 22 GPIO_ACTIVE_HIGH>;
 		};
 	};
@@ -125,7 +129,7 @@
 		};
 	};
 
-	gmac1: mac@1 {
+	mac@1 {
 		/* WAN */
 		compatible = "mediatek,eth-mac";
 		reg = <1>;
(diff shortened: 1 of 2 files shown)


target/linux/mediatek/dts/mt7986a-asus-tuf-ax6000.dts
```diff
@@ -54,7 +54,7 @@
 			function = LED_FUNCTION_WLAN;
 			color = <LED_COLOR_ID_WHITE>;
 			gpios = <&pio 2 GPIO_ACTIVE_HIGH>;
-			linux,default-trigger = "phy1tpt";
+			linux,default-trigger = "network";
```
openwrt-ai, line 57: nit: `mt7986a-asus-tuf-ax4200q.dts` is, after this PR, byte-for-byte identical to this file apart from the model/compatible, the LAN port label ordering, the extra `cover-*` LEDs, and this trigger — it still carries `linux,default-trigger = "phy1tpt";` on its single WLAN LED ([mt7986a-asus-tuf-ax4200q.dts:57](https://github.com/openwrt/openwrt/blob/cf55e40dcd4a28ec774001bc0d3e5c156e501529/target/linux/mediatek/dts/mt7986a-asus-tuf-ax4200q.dts#L57)) and has the same dual-band MT7986A. Is leaving that board on the single-phy workaround intentional, or should it get the same treatment in this series?

(For the record, the mechanism checks out: with no `label`/`family`/`mode` on the node the LED classdev name is `white:wlan`, the trigger takes the substring after the last `:` and resolves it to `NET_TRIG_WLAN`, and [`name_matches_type()`](``https://github.com/openwrt/openwrt/blob/cf55e40dcd4a28ec774001bc0d3e5c156e501529/target/linux/generic/files/drivers/leds/trigger/ledtrig-network.c#L199-L233``) accepts the `phyN-*` netdev names OpenWrt uses, so both radios are covered. `CONFIG_LEDS_TRIGGER_NETWORK` is unset in `config-6.18`, so the added `kmod-ledtrig-network` dependency is required.)

---

target/linux/mediatek/dts/mt7986a-asus-tuf-ax6000.dts
```diff
@@ -54,7 +54,7 @@
...
 		};
 
 		led_system: system {
```
openwrt-ai, line 60: nit: while you are cleaning up this `leds` node, four LEDs still use the deprecated `label = "<color>:<function>"` form and were left untouched: `led_system` (line 61, `"white:system"`), `cover-blue` (line 72), `cover-red` (line 77) and `cover-green` (line 82). The preferred form is `color = <LED_COLOR_ID_*>;` + `function = ...;` with the constants from `dt-bindings/leds/common.h`, which the `wlan` and `wan-red` nodes right above/below already use.

The near-identical sibling has already migrated the same LED — [mt7986a-asus-tuf-ax4200q.dts:60-63](https://github.com/openwrt/openwrt/blob/cf55e40dcd4a28ec774001bc0d3e5c156e501529/target/linux/mediatek/dts/mt7986a-asus-tuf-ax4200q.dts#L60-L63) uses `function = LED_FUNCTION_STATUS;` + `color = <LED_COLOR_ID_WHITE>;`. For the `cover-*` LEDs there is no matching `LED_FUNCTION_*` constant, so `color = <LED_COLOR_ID_BLUE>;` + `function = "cover";` is the equivalent. Note this renames the sysfs LED from `white:system` to `white:status`; the `led-boot`/`led-running`/`led-failsafe`/`led-upgrade` aliases go through the `&led_system` phandle so they are unaffected.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -532,7 +532,7 @@ define Device/asus_tuf-ax6000
   DEVICE_DTS := mt7986a-asus-tuf-ax6000
   DEVICE_DTS_DIR := ../dts
   DEVICE_DTS_LOADADDR := 0x47000000
-  DEVICE_PACKAGES := kmod-usb3 kmod-mt7915e kmod-mt7986-firmware mt7986-wo-firmware
+  DEVICE_PACKAGES := kmod-usb3 kmod-mt7915e kmod-mt7986-firmware mt7986-wo-firmware kmod-ledtrig-network
```
openwrt-ai, line 535: nit: this hunk is the only functional change outside the DTS and it is not mentioned in the commit message, which makes the mixed scope of the single commit harder to review. The DTS diff is 341 lines, but a normalized property-level comparison of the old and new file shows the only semantic deltas are: the LED trigger swap, the removal of the dead `&pcie_phy { status = "okay"; }` override, and the removal of `#address-cells`/`#size-cells` from `flash@0` — everything else is pure re-indentation of the `&mdio`/`&switch`/`&ubi_factory` overrides into their parent nodes.

Consider splitting into two commits so the functional change stays bisectable and reviewable on its own:

1. `mediatek: use "network" LED trigger for ASUS TUF-AX6000` — the `linux,default-trigger` change plus this `DEVICE_PACKAGES` addition.
2. `mediatek: clean up ASUS TUF-AX6000 device tree` — the inlining, the `&pcie_phy` and `#address-cells`/`#size-cells` removals, and the `wf_dbdc_pins` re-alignment.

(The three semantic DTS deltas themselves look correct: `pcie` is `status = "disabled"` in the kernel's `mt7986a.dtsi` and is never enabled by this board, so the `pcie_phy` override was dead; and the sibling `mt7986a-asus-tuf-ax4200q.dts` carries the same `u-boot-dont-touch-spi-nand` + bait-node layout with no cells on `flash@0`.)

---

openwrt-ai (commented):
Reviewed 1 new commit. No functional regressions found — the DTS restructure is property-for-property equivalent to the original, and the three real changes (LED trigger, dead `&pcie_phy` override, `flash@0` cells) all check out against the kernel 6.18 `mt7986a.dtsi` and the sibling `mt7986a-asus-tuf-ax4200q.dts`. Three nits inline.

---

openwrt-ai (commented):

The only change since the last review is the `label = "<color>:<function>"` migration of `led_system` and the three `cover-*` LEDs, which resolves the earlier nit; the commit message now also mentions the `filogic.mk` `DEVICE_PACKAGES` hunk. Verified:

- `color`/`function` ordering matches the `wlan` / `wan-red` nodes already in this file and the sibling [mt7986a-asus-tuf-ax4200q.dts:60-63](https://github.com/openwrt/openwrt/blob/3c8b26c8afef71627bed1bc340c92388ad99ea91/target/linux/mediatek/dts/mt7986a-asus-tuf-ax4200q.dts#L60-L63); `dt-bindings/leds/common.h` is already included at line 6.
- The `cover-*` LEDs keep their exact sysfs names — `LED_COLOR_ID_BLUE/RED/GREEN` are 3/1/2 in [common.h](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/include/dt-bindings/leds/common.h#L25-L28) and `function = "cover"` has no matching `LED_FUNCTION_*`, so the composed names stay `blue:cover`, `red:cover`, `green:cover`.
- The `white:system` -> `white:status` rename is safe: nothing in-tree references that string (the mediatek `01_leds` entry for `asus,tuf-ax6000` only touches the `mdio-bus:0*` PHY LEDs), and the `led-boot`/`led-failsafe`/`led-running`/`led-upgrade` aliases resolve through [`get_dt_led()`](https://github.com/openwrt/openwrt/blob/3c8b26c8afef71627bed1bc340c92388ad99ea91/package/base-files/files/lib/functions/leds.sh#L44-L56), which falls back to composing `color:function` when no `label` is present.

CI on this head reports no failures (no completed checks yet).

---

PR #24667, opened 2026-08-11
[24.10] wireguard-tools: bump to 20260223

sanderpqr:
```
% git shortlog v1.0.20210914..v1.0.20260223

Daniel Grober (1):
      wg-quick: run PreUp hook after creating interface

Dmitry Selivanov (1):
      show: fix show all endpoints output

Doug Freed (1):
      wg-quick@.service: add deps on wg-quick.target

Ilia Shipitsin (1):
      config: handle strdup failure

Jason A. Donenfeld (17):
      ipc: use more clever PnP enumerator
      reresolve-dns: use $EPOCHSECONDS instead of $(date +%s)
      wg-quick: android: use right regex for host-vs-IP
      global: dual license core files as MIT for FreeBSD
      show: apply const to right part of pointer
      man: set private key in PreUp rather than PostUp
      ipc: linux: enforce IFNAMSIZ limit
      ipc: add stub for allowedips flags on other platforms
      version: bump
      wg-quick: linux: use smallest mtu, not largest
      syncconf: account for psks removed from config file
      wg-quick: linux: deal with resolvconf migration more gracefully
      wg-quick: use addconf instead of setconf
      wg-quick: linux: do not unnecessarily set sysctl
      config: preserve const correctness
      syncconf: account for persistent keepalive removed from config file
      version: bump

Jordan Rife (1):
      ipc: linux: support incremental allowed ips updates

Kyle Evans (4):
      ipc: freebsd: move if_wg path to reflect new in-tree location
      ipc: freebsd: avoid leaking memory in kernel_get_device()
      ipc: freebsd: NULL out some freed memory in kernel_set_device()
      ipc: freebsd: use AF_LOCAL for the control socket

Mikael Magnusson (1):
      embeddable-wg-library: add named wg_endpoint union

Robyn Kosching (1):
      wg-quick: pass on # comments to {Pre,Post}{Up,Down}

Tom Yan (1):
      wg-quick: linux: prevent traffic from momentarily leaking into tunnel

TriangleSnake (1):
      wg-quick: linux: add 'dev' to 'ip link add' to avoid keyword conflicts
```

Build system: x86/64
Build-tested: generic

Link: https://github.com/openwrt/openwrt/pull/22190

(cherry picked from commit eda2630ceb7d98293fcc3cf9df1c27e1d8ed803a)


package/network/utils/wireguard-tools/Makefile
```diff
@@ -10,12 +10,12 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=wireguard-tools
 
-PKG_VERSION:=1.0.20210914
-PKG_RELEASE:=4
+PKG_VERSION:=1.0.20260223
+PKG_RELEASE:=1
```
openwrt-ai, line 15: This is not the diff of the commit named in the `(cherry picked from commit eda2630ceb7d…)` trailer. Upstream [eda2630 on main](https://github.com/openwrt/openwrt/commit/eda2630ceb7d98293fcc3cf9df1c27e1d8ed803a) goes `1.0.20250521` → `1.0.20260223` and leaves `PKG_RELEASE:=1` untouched; openwrt-24.10 is still at `1.0.20210914`, so this commit also silently folds in the intermediate main bump [c374deca "wireguard-tools: update to v1.0.20250521"](https://github.com/openwrt/openwrt/commit/c374deca7ea3c9ccca0059382fc37b54e1a2da6d) and resets `PKG_RELEASE` 4 → 1.

The `PKG_RELEASE` reset itself is fine and matches branch practice, but the provenance should reflect what actually landed — either cherry-pick c374deca first as its own commit and then eda2630 (each with its own `-x` trailer), or keep the squash and say so explicitly in the message instead of relying on a single cherry-pick trailer, the way [cd8d701c7f ("backport: update dnsmasq to v2.93 from master")](https://github.com/openwrt/openwrt/commit/cd8d701c7fc0efe9b8170b44354945a21474c066) does on this branch ("Wholesale drop … / not a cherry pick"). As it stands, `git log --grep` on the trailer will attribute a two-release jump to a one-release commit.

---

package/network/utils/wireguard-tools/Makefile
openwrt-ai, line 1: nit: the commit message and the PR body disagree about how this was tested. The message carries the upstream author's `Build-tested: x86/64-glibc` / `Run-tested: x86/64-glibc` lines verbatim, while the PR body says only `Build-tested: generic` and claims no run-test.

That distinction matters more than usual here, because the run-test on main exercised a different proto handler: main rewrote it in ucode ([41bc4546](https://github.com/openwrt/openwrt/commit/41bc454602f15ac81f7cc9933d079323614d6492), `files/wireguard.uc`), whereas openwrt-24.10 still ships the shell version — [`wg setconf` at wireguard.sh:165](https://github.com/openwrt/openwrt/blob/ce0a747105e70d9339cc802b8f2ed0fa8a15db25/package/network/utils/wireguard-tools/files/wireguard.sh#L165) and [`wg show "${config}" endpoints` at wireguard.sh:199](https://github.com/openwrt/openwrt/blob/ce0a747105e70d9339cc802b8f2ed0fa8a15db25/package/network/utils/wireguard-tools/files/wireguard.sh#L199). Nothing in the shortlog looks like it changes per-interface `wg show … endpoints` output or `setconf` semantics (the `show:` fix is about `wg show all endpoints`, and the `syncconf` changes don't apply), so I don't see a concrete breakage — but this jumps 4.5 years of wireguard-tools onto a handler nobody exercised upstream. Was it run-tested on 24.10, and if not, could the inherited `Run-tested:` line be adjusted to reflect what was actually done for this backport?

---

openwrt-ai (commented):
## Commit checks

- ce0a747 "wireguard-tools: bump to 20260223" — the `(cherry picked from commit eda2630ceb7d…)` trailer doesn't describe this diff: upstream eda2630 bumps `1.0.20250521` → `1.0.20260223` with `PKG_RELEASE` unchanged, while this commit goes `1.0.20210914` → `1.0.20260223` and resets `PKG_RELEASE` 4 → 1, i.e. it also absorbs main's intermediate bump c374deca ("wireguard-tools: update to v1.0.20250521"). The rewritten shortlog in the body correctly covers `v1.0.20210914..v1.0.20260223`, so only the provenance line is out of sync — see the inline comment for the two ways this branch usually handles it.

`PKG_HASH` matches the tarball hash carried by eda2630 on main, and no other file on openwrt-24.10 references the wireguard-tools version.

---

sanderpqr:
Ping @hauke @robimarko

PR #24670, opened 2026-08-11
wifi-scripts: recognise dyn. hostapd vlan ifaces

knarrff:
If using hostapd with dynamic vlans, it is the task of hostapd to create the relevant interfaces, may they be specified in a file or received from radius. netifd, so far, does not know about and ignores them. Among not being able to see and query statistics for those devices from command line, this especially leads to luci not knowing about those interfaces either, as luci mostly relies on netifd for that type of information. This by itself might not sound like such a big thing to miss, as luci will not be able to manage those interfaces directly, but it also leads to luci ignoring those interfaces when displaying associated wireless clients.
The effect in the end is, that up until before this patch, luci will not display any wireless client associated via such a dynamic vlan, often leading to a completely empty list.

See https://github.com/openwrt/netifd/issues/26 for a bug report about this issue.

hostapd does not report these vlan interfaces using the netlink kernel interface as I believe it should (see
https://lists.infradead.org/pipermail/hostap/2026-July/045328.html) for an unanswered question regarding this issue. netifd might (not checked since not implemented) pickup those interfaces otherwise, but it is also unlikely that this is changed any time soon, if at all.

While netifd does not receive notification about those interfaces through the netlink kernel interface, it still receives notifications through hotplug, so the patch essentially does this whenever a new device is reported through hotplug (simplified):

- see if this is a dynamic vlan interface to be added
- see if this interface belongs to the current wifi device (radio) by querying nl80211 for all known interfaces of the "current" radio

If so: add this device to the known devices and from then on (until it is removed regularly), provide reports about it, especially its existence and the related radio.

Note that this patch alone will not solve the issue with luci mentioned above, only the one on the command line. However, this patch can stand on its own and solves a problem regardless of the relevance to luci.

A second patch will be provided to luci in its repository that handles its recognition and display of associated clients to those interfaces.

hostapd could, and possibly should, directly report those interfaces to the kernel, which it currently does not do. If that would be the case, this patch might not be necessary and/or might look differently. However, it is unclear to me whether this would be a correct solution and a query towards this remained unanswered
(https://lists.infradead.org/pipermail/hostap/2026-July/045328.html).

A second possibility would be to implement this logic into luci (and I initially did). This, however, would be, IMHO, not the the right place to do this, and would not solve the same issue on the netifd command line.

### DIFF
--- a/package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
+++ b/package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
@@ -1,6 +1,7 @@
 'use strict';
 import * as ubus from "ubus";
 import * as uloop from "uloop";
+import * as nl80211 from "nl80211";
 import { is_equal } from "./utils.uc";
 import { access } from "fs";
 
@@ -559,12 +560,73 @@ function hotplug(name, add)
 	if (m)
 		name = m[1];
 
+	let suspect_more_ifaces = false;
+	let handled = false;
+	let known_phys = {};
 	for (let section, data in this.handler_data) {
+		if (data.type == "vif" && data?.config?.dynamic_vlan)
+			suspect_more_ifaces = true;
+		// remember all phy we see and assume this handler is responsible for those (and no others)
+		if (data.type == "device" && data.phy)
+			// remember both phy and radio(s), to be able to match those also on multi-radio phys
+			known_phys[data.phy] = { section, radio: +(data.radio ?? -1) };
 		if (data.ifname != name ||
 		    data.type != "vif" && data.type != "vlan")
 			continue;
 
 		handle_link(dev, data, add);
+		if (!add && "dynamic_vlan_iface" in data)
+			delete this.handler_data[dev];
+		handled = true;
+	}
+	// Return if the device was found, if it is not about to be added, and also if there is no indication for more devices
+	if (handled || !add || !suspect_more_ifaces)
+		return;
+
+	// We got a wireless device we don't know about yet: find it
+
+	// Get all radios on this device
+	let phys = nl80211.request(nl80211.const.NL80211_CMD_GET_WIPHY, nl80211.const.NLM_F_DUMP, { split_wiphy_dump: true });
+	// Get all known interfaces on this device: this should contain the one about to be added
+	let nl80211interfaces = nl80211.request(nl80211.const.NL80211_CMD_GET_INTERFACE, nl80211.const.NLM_F_DUMP);
+
+	// Loop through all radios found
+	for (let phy in phys) {
+		let phys_data = known_phys[phy.wiphy_name];
+		// skip if this radio is not part of this handlers'
+		if (!phys_data)
+			continue;
+		// loop through all interfaces found
+		for (let k, v in nl80211interfaces) {
+			// check if this interface has the requested name, skip otherwise
+			if (v.ifname != name)
+				continue;
+			// check if this wiphy is this handler's, skip otherwise
+			if (v.wiphy != phy.wiphy)
+				continue;
+			// in a multi-radio setup, check radios and skip if they do not participate
+			if (phys_data.radio >= 0 && !(v.vif_radio_mask & (1 << phys_data.radio)))
+				continue;
+			// Only handle vlans here
+			if (v.iftype != nl80211.const.NL80211_IFTYPE_AP_VLAN)
+				continue;
+			// Add this interface to our known list
+			let ifdata = {
+				"config": {
+					"device": [ phys_data.section ],
+					"mode": "ap",
+				},
+				"dynamic_vlan_iface": true,
+				"ifname": dev,
+				"name" : dev,
+				"sta" : [],
+				"type" : "vif",
+				"vlan" : [],
+			};
+
+			handle_link(dev, ifdata, add);
+			this.handler_data[dev] = ifdata;
+		}
 	}
 }
 
@@ -603,7 +665,7 @@ function get_status_stations(wdev, vif)
 function status()
 {
 	let interfaces = [];
-	for (let vif in this.data.vif) {
+	for (let vif in [...this.data.vif, ...filter(values(this.handler_data), (vif) => "dynamic_vlan_iface" in vif)]) {
 		let vlans = get_status_vlans(this, vif);
 		let stations = get_status_stations(this, vif);
 		let data = get_status_data(this, vif);


knarrff:
I am hesitant to increment PKG_RELEASE, but lean towards it: please advise if you find this addition to be large enough to warrant it.

package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
```diff
@@ -559,12 +560,67 @@ function hotplug(name, add)
...
+	// Return if the device was found, if it is not about to be added, and also if there is no indication for more devices
+	if (handled || !add || !suspect_more_ifaces)
+		return;
+
+	// We got a wireless device we don't know about yet: find it
+
+	// Get all radios on this device
+	phys = nl80211.request(nl80211.const.NL80211_CMD_GET_WIPHY, nl80211.const.NLM_F_DUMP, { split_wiphy_dump: true });
+	// Get all known interfaces on this device: this should contain the one about to be added
+	nl80211interfaces = nl80211.request(nl80211.const.NL80211_CMD_GET_INTERFACE, nl80211.const.NLM_F_DUMP);
```
openwrt-ai: `phys` and `nl80211interfaces` are assigned without `let`, i.e. they are stores to undeclared globals. netifd loads these scripts with [`strict_declarations = true` (netifd `ucode.c:535-536`)](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/ucode.c#L535-L540), and the file additionally opts in with `'use strict'` on line 1, so a store to an undeclared name raises `Reference error: access to undeclared variable phys` ([ucode `vm.c:1412-1415`](https://github.com/jow-/ucode/blob/73ffef75c653dc901c69678a27c67842d291e79e/vm.c#L1409-L1420)). The new code path therefore throws on the very first dynamic-VLAN hotplug event and never reaches the matching loop.

```suggestion
	let phys = nl80211.request(nl80211.const.NL80211_CMD_GET_WIPHY, nl80211.const.NLM_F_DUMP, { split_wiphy_dump: true });
	// Get all known interfaces on this device: this should contain the one about to be added
	let nl80211interfaces = nl80211.request(nl80211.const.NL80211_CMD_GET_INTERFACE, nl80211.const.NLM_F_DUMP);
```


---

package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
```diff
@@ -559,12 +560,67 @@ function hotplug(name, add)
...
+				"check_vlan": false,
+				"config": {
+					"device": [ radio ],
+					"mode": "ap",
+					"vlan_bridge": dev,
+				},
+				"ifname": dev,
+				"name" : dev,
+				"type" : "vif",
+			};
```
openwrt-ai, line 625: This object is pushed into `this.data.vif` below, but it lacks the `vlan` and `sta` arrays that every other entry of that array carries (see the vif construction in [`wireless.uc:196-202`](``https://github.com/openwrt/openwrt/blob/e2175dcf1ee2ade3bdda4c648a86eae8b7d89ef0/package/network/config/wifi-scripts/files/lib/netifd/wireless.uc#L196-L202)).`` [`wdev_update_disabled_vifs()` does `push(vifs, vif, ...vif.vlan)` at wireless-device.uc:339](``https://github.com/openwrt/openwrt/blob/e2175dcf1ee2ade3bdda4c648a86eae8b7d89ef0/package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc#L336-L339),`` and spreading a non-array in ucode raises `TypeError: (null) is not iterable` ([`vm.c:651-656`](https://github.com/jow-/ucode/blob/73ffef75c653dc901c69678a27c67842d291e79e/vm.c#L648-L660)) — so the next `check()`/`start()` on this radio will throw. `vlan` is the one that actually crashes; add `sta` too so the entry matches the shape the rest of the file assumes (`get_status_stations()` iterates it).

```suggestion
			let ifdata = {
				"check_vlan": false,
				"config": {
					"device": [ radio ],
					"mode": "ap",
					"vlan_bridge": dev,
				},
				"ifname": dev,
				"name" : dev,
				"type" : "vif",
				"vlan": [],
				"sta": [],
			};
```


---

package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
```diff
@@ -559,12 +560,67 @@ function hotplug(name, add)
...
+
+			handle_link(dev, ifdata, add);
+			this.handler_data[dev] = ifdata;
+			push(this.data.vif, ifdata);
```
openwrt-ai: Injecting a synthetic entry into `this.data.vif` mutates the config netifd handed down, and that array is not just a status list — it is the input to [`wdev_config_init()` (`for (let vif in data.vif)` at wireless-device.uc:92, building `handler_config.interfaces` at L120-L124)](``https://github.com/openwrt/openwrt/blob/e2175dcf1ee2ade3bdda4c648a86eae8b7d89ef0/package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc#L91-L129),`` which is serialized and passed to the setup script. On the next `config_change`/retry the hostapd-created VLAN interface would come back as a configured BSS (`mode=ap`, no ssid/encryption) handed to `mac80211.sh`.

There is also no removal path: on `hotplug(name, false)` the loop above matches the synthetic entry, calls `handle_link(..., false)` and returns with `handled = true`, but nothing deletes it from `this.handler_data` or from `this.data.vif`. `status()` keeps advertising the interface after hostapd has torn it down, until the next full config `update()` replaces `this.data`.

Would keeping the entry only in `this.handler_data` (which is wiped by `wdev_teardown_cb()`) and teaching `status()` to also walk `handler_data` for dynamic-VLAN entries avoid both problems? Otherwise the remove side needs an explicit `delete this.handler_data[dev]` plus a splice out of `this.data.vif`.

---

package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
```diff
@@ -559,12 +560,67 @@ function hotplug(name, add)
...
+		// skip if this radio is not part of this handlers'
+		if (!known_phys[phy.wiphy_name])
+			continue;
+		// loop through all interfaces found
+		for (let k, v in nl80211interfaces) {
+			// check if this interface has the requested name, skip otherwise
+			if (v.ifname != name)
+				continue;
+			// check if this interface is on one of this handler's radios, skip otherwise
+			if (v.wiphy != phy.wiphy)
```
openwrt-ai: Ownership is decided purely on the wiphy, but a wiphy can back several `wifi-device` sections: the schema documents [`radio` as "Index of the phy radio (for multi-radio PHYs)"](``https://github.com/openwrt/openwrt/blob/e2175dcf1ee2ade3bdda4c648a86eae8b7d89ef0/package/network/config/wifi-scripts/files-ucode/usr/share/schema/wireless.wifi-device.json#L574-L578),`` and [`mac80211.sh:1213`](``https://github.com/openwrt/openwrt/blob/e2175dcf1ee2ade3bdda4c648a86eae8b7d89ef0/package/network/config/wifi-scripts/files/lib/netifd/wireless/mac80211.sh#L1213)`` records `phy=` and `radio=` together in the handler data. On such a device every `wdev` handler sharing `phy0` will see `known_phys["phy0"]` set, so each of them claims the same AP_VLAN interface, calls `handle_link()` on it and pushes it into its own `data.vif` — the VLAN gets attributed to an arbitrary radio (whichever `wifi-device` section was iterated last into `known_phys`).

Is that intentional / considered out of scope for now? `data.radio` is available in the handler data, and ucode's nl80211 binding exposes `vif_radio_mask` on the interface dump, so the pair could be matched instead of the phy alone.

---

package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
```diff
@@ -559,12 +560,67 @@ function hotplug(name, add)
...
+			if (v.iftype != nl80211.const.NL80211_IFTYPE_AP_VLAN)
+				continue;
+			// Add this interface to our known list
+			let radio = known_phys[phy.wiphy_name];
+			let ifdata = {
+				"check_vlan": false,
+				"config": {
+					"device": [ radio ],
+					"mode": "ap",
+					"vlan_bridge": dev,
```
openwrt-ai: nit: `vlan_bridge` is a hostapd option meaning "bridge to attach dynamic VLAN interfaces to" ([`hostapd.sh:1082-1083`](``https://github.com/openwrt/openwrt/blob/e2175dcf1ee2ade3bdda4c648a86eae8b7d89ef0/package/network/config/wifi-scripts/files/lib/netifd/hostapd.sh#L1082-L1083),`` [schema entry](``https://github.com/openwrt/openwrt/blob/e2175dcf1ee2ade3bdda4c648a86eae8b7d89ef0/package/network/config/wifi-scripts/files-ucode/usr/share/schema/wireless.wifi-iface.json#L1130)),`` not the VLAN interface name. `handle_link()` never reads it, but `status()` publishes this whole `config` object over ubus, so consumers get a `vlan_bridge` that is really the netdev name. Dropping the key (or naming it something clearly synthetic) would avoid the false signal.

---

package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
```diff
@@ -1,6 +1,7 @@
 'use strict';
 import * as ubus from "ubus";
 import * as uloop from "uloop";
+import * as nl80211 from "nl80211";
```
openwrt-ai, line 4: nit: answering your question about the bot's `PKG_RELEASE` warning — this is a behaviour change to installed runtime files (`files/lib/netifd/wireless-device.uc`), not a cosmetic edit, so [`package/network/config/wifi-scripts/Makefile:12`](https://github.com/openwrt/openwrt/blob/e2175dcf1ee2ade3bdda4c648a86eae8b7d89ef0/package/network/config/wifi-scripts/Makefile#L12) should go `PKG_RELEASE:=1` → `2`; otherwise sysupgrade/opkg users will not pick the change up. The "do not increment for minor changes" carve-out is for comments, whitespace, `PKG_MAINTAINER` and source-URL churn.

---

openwrt-ai (commented):
## Commit checks

- `e2175dc` "netifd: recognise dynamic hostapd vlan interfaces" — the subject is scoped to `netifd:`, but the commit only touches `package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc`; no netifd source or package version is changed. Every recent commit on that path uses the `wifi-scripts:` prefix (`0e371752ae`, `e18424ddd8`, `105dd239ac`, …), so `wifi-scripts: recognise dynamic hostapd VLAN interfaces` would match the tree convention. Also, "recognision" in the body (and in the PR description) should be "recognition".


---

package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
knarrff: I am hesitant because it is still release 1 and version 1.0, despite obvious changes in the past that would have been even more reason for an increase, suggesting that there was and still is a reason not to.

package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
knarrff: Dear AI, this sounds good, but has a few problems I would need advise on:
1. vif_radio_mask (nl80211) is on my devices always set to 0 (meaning unset), regardless of which radio a device is on. The nl80211 data only exposes "dev" and "ifname" (same value), other than "wiphy", and both are already matched. It does expose a few radio frequencies, but matching those is likely wrong: who is to say that two radios couldn't be on the same frequencies?
2. Even if vif_radio_mask>0 (which looks would be the case on multi-link setups, but which I don't use and which I can't test), what would I have to match this index/mask against? All I get from the handler are radio names ("radio0", "radio1"): I could assume an index from those (0, 1), but that would likely break at some point or somehwere when different names are used. The handler data does link interfaces to devices by name, but those device sections then (at least for me) all contain `"radio": -1` entries, i.e., not a radio index.

package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
knarrff: Should be handled by 49e173e56d1075117a4d4d5f6ca38a33d1c89630

package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
```diff
@@ -559,12 +560,68 @@ function hotplug(name, add)
...
+				"dynamic_vlan": true,
+				"ifname": dev,
+				"name" : dev,
+				"sta" : [],
+				"type" : "vif",
+				"vlan" : [],
+			};
+
+			handle_link(dev, ifdata, add);
+			this.handler_data[dev] = ifdata;
```
openwrt-ai, line 628: The entry is added to `this.handler_data` here but there is still no path that removes it. On `hotplug(dev, false)` the [loop at wireless-device.uc:566-578](``https://github.com/openwrt/openwrt/blob/49e173e56d1075117a4d4d5f6ca38a33d1c89630/package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc#L566-L578)`` matches this synthetic entry (`data.ifname == name`, `data.type == "vif"`), calls `handle_link(dev, data, false)`, sets `handled = true` and returns — nothing ever does `delete this.handler_data[dev]`.

Dropping the `push(this.data.vif, ...)` did not fix that; it moved the leak to the place `status()` now reads from. Two consequences that are new with this commit:

- `status()` keeps advertising the AP_VLAN netdev over ubus after hostapd has torn it down, until `wdev_teardown_cb()` / `wdev_config_init()` wipe `handler_data`.
- [`wdev_mark_up()` replays `handle_link(data.ifname, data, true)` over every `handler_data` entry (wireless-device.uc:476-478)](``https://github.com/openwrt/openwrt/blob/49e173e56d1075117a4d4d5f6ca38a33d1c89630/package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc#L476-L478),`` so a setup retry re-registers netdevs that no longer exist.

So the commit body's claim that this "should also remove the problem that the added entries in `this.data.vif` might have not been deleted until a new `update()`" doesn't hold yet. The removal side needs an explicit delete in the first loop — e.g. after `handle_link(dev, data, add)`, when `!add && data.dynamic_vlan`, `delete this.handler_data[section]`.

---

package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
```diff
@@ -559,12 +560,68 @@ function hotplug(name, add)
...
+				continue;
+			// Add this interface to our known list
+			let radio = known_phys[phy.wiphy_name];
+			let ifdata = {
+				"check_vlan": false,
+				"config": {
+					"device": [ radio ],
+					"mode": "ap",
+				},
+				"dynamic_vlan": true,
```
openwrt-ai: nit: this key reuses a name that already has a different meaning in the same function. Ten lines earlier, [`data?.config?.dynamic_vlan` at wireless-device.uc:567](``https://github.com/openwrt/openwrt/blob/49e173e56d1075117a4d4d5f6ca38a33d1c89630/package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc#L567)`` is the hostapd `dynamic_vlan` *option* ("this BSS may spawn VLAN interfaces", [schema](``https://github.com/openwrt/openwrt/blob/49e173e56d1075117a4d4d5f6ca38a33d1c89630/package/network/config/wifi-scripts/files-ucode/usr/share/schema/wireless.wifi-iface.json#L1123)),`` whereas the new top-level `dynamic_vlan` means "this entry *is* a hostapd-created VLAN netdev". A reader (and `status()` at line 663, which tests `"dynamic_vlan" in vif`) has to keep the two apart by nesting level alone. Something like `dynamic_vlan_iface` for the synthetic marker would make both sites unambiguous.

---

openwrt-ai (commented):
Reviewed the 4 new commits since `e2175dc`.

## Commit checks

- `8404a05` "netifd: declare locally used variables as local" and `058230f`… sorry, `0582302` "netifd: fully populate vif entry" — diffs match their messages, but both are still prefixed `netifd:` while touching only `package/network/config/wifi-scripts/`. Same point as on `e2175dc`.
- `6a387e8` "wireless-devices: remove wrong config entry" and `49e173e` "wireless-devices: remove mutation of this.data.vif" — diffs match, but `wireless-devices:` is neither the package (`wifi-scripts`) nor the file (`wireless-device.uc`, singular). The series now carries three different prefixes across five commits (`netifd:` ×3, `wireless-devices:` ×2) while the PR title says `wifi-scripts:`; please settle on `wifi-scripts:` for all of them.
- `49e173e` "wireless-devices: remove mutation of this.data.vif" — the body says the patch "should also remove the problem that the added entries in this.data.vif might have not been deleted until a new update()". The entries are still never deleted; they only moved from `this.data.vif` to `this.handler_data`, which `status()` now reads. Details inline on line 623.
- Series hygiene: `8404a05`, `0582302`, `6a387e8` and `49e173e` only amend `e2175dc`, which is not upstream yet — each one fixes a defect introduced by the first commit. Please squash them into a single commit before merge so the history doesn't carry an introduce-then-fix chain (the intermediate states, e.g. the `phys`/`nl80211interfaces` undeclared-global writes, are not bisectable).

---

package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
```diff
@@ -559,12 +560,70 @@ function hotplug(name, add)
...
 
+	let suspect_more_ifaces = false;
+	let handled = false;
+	let known_phys = {};
 	for (let section, data in this.handler_data) {
+		if (data.type == "vif" && data?.config?.dynamic_vlan)
+			suspect_more_ifaces = true;
+		// remember all phy we see and assume this handler is responsible for those (and no others)
+		if (data.type == "device" && data.phy)
+			known_phys[data.phy] = section;
```
openwrt-ai: Following up on your two questions in the [thread on line 603](https://github.com/openwrt/openwrt/pull/24670#discussion_r3764706617) — I went and checked, and `vif_radio_mask` does look usable, but only on genuinely multi-radio wiphys. Two answers:

**1. Why you always see `0`.** cfg80211 initialises [`wdev->radio_mask = BIT(wdev->wiphy->n_radio) - 1` at core.c:1464](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/net/wireless/core.c#L1464), and `n_radio` is 0 on every classic one-radio-per-wiphy driver, so the mask is `BIT(0) - 1 == 0`. [`nl80211_send_iface()` emits the attribute unconditionally](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/net/wireless/nl80211.c#L4217) (kernel v6.18.44, which is `KERNEL_PATCHVER` on main), so `0` here means "this wiphy is not radio-split" — precisely the case where matching on the phy alone is already unambiguous. The ambiguity you'd want to resolve only arises when `n_radio > 0`, and there the mask is non-zero.

**2. What to match it against.** The same `wireless_set_data` that gives you `data.phy` also gives you `data.radio`: [`mac80211.sh:1213`](``https://github.com/openwrt/openwrt/blob/54aab695313e8f6bf9f74b791090dc9093b27c03/package/network/config/wifi-scripts/files/lib/netifd/wireless/mac80211.sh#L1213)`` does `wireless_set_data phy="$phy" radio="$radio"`. That `radio` is the phy radio index (`-1` when the phy is not split, see [`mac80211.sh:1144`](``https://github.com/openwrt/openwrt/blob/54aab695313e8f6bf9f74b791090dc9093b27c03/package/network/config/wifi-scripts/files/lib/netifd/wireless/mac80211.sh#L1144))`` — and it is exactly the bit index hostapd uses: [`hostapd.uc:1231-1236`](https://github.com/openwrt/openwrt/blob/54aab695313e8f6bf9f74b791090dc9093b27c03/package/network/services/hostapd/files/hostapd.uc#L1231-L1236) builds `radio_mask |= 1 << r` from `config.radios`. OpenWrt's [`370-preserve_radio_mask.patch`](``https://github.com/openwrt/openwrt/blob/54aab695313e8f6bf9f74b791090dc9093b27c03/package/network/services/hostapd/patches/370-preserve_radio_mask.patch``) then propagates the parent BSS's mask into `wpa_driver_nl80211_if_add()`, which is the path `hostapd_vlan_if_add()` → `hostapd_if_add(WPA_IF_AP_VLAN, ...)` takes for dynamic-VLAN netdevs. So the AP_VLAN inherits the mask of the BSS that created it.

So pairing the phy with the radio index would work and degrades to today's behaviour on your hardware. Note `data.radio` is a *string* — [`_wdev_add_variables()` uses `json_add_string`](``https://github.com/openwrt/openwrt/blob/54aab695313e8f6bf9f74b791090dc9093b27c03/package/network/config/wifi-scripts/files/lib/netifd/netifd-wireless.sh#L142)`` — so it needs coercing:

```
	if (data.type == "device" && data.phy)
		known_phys[data.phy] = { section, radio: +(data.radio ?? -1) };
```

and then at line 595/603:

```
	let pd = known_phys[phy.wiphy_name];
	if (!pd)
		continue;
	...
	if (pd.radio >= 0 && !(v.vif_radio_mask & (1 << pd.radio)))
		continue;
	...
	let radio = pd.section;
```

Agreed that matching on frequencies would be wrong.

---

package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
```diff
@@ -559,12 +560,70 @@ function hotplug(name, add)
...
+			// check if this interface is on one of this handler's radios, skip otherwise
+			if (v.wiphy != phy.wiphy)
+				continue;
+			// Only handle vlans here
+			if (v.iftype != nl80211.const.NL80211_IFTYPE_AP_VLAN)
+				continue;
+			// Add this interface to our known list
+			let radio = known_phys[phy.wiphy_name];
+			let ifdata = {
+				"check_vlan": false,
```
openwrt-ai: nit: this key is never read. `handle_link()` does not look at `data.check_vlan` — it hardcodes `check_vlan: false` in its own `dev_data` at [wireless-device.uc:58](``https://github.com/openwrt/openwrt/blob/54aab695313e8f6bf9f74b791090dc9093b27c03/package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc#L58),`` and that is the only occurrence in the whole package. It is also the only consumer that matters here: `check_vlan` is what stops netifd from treating a dotted name like `wlan0.4` (hostapd `vlan_naming=1`) as a VLAN of `wlan0` in [`__device_get()` (netifd device.c:923)](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/device.c#L917-L924), and `handle_link()` already covers that. Carrying a copy in `ifdata` suggests it is what disables the VLAN-chain lookup, which it isn't.

```suggestion
```

---

openwrt-ai (commented):
Reviewed the squashed head `54aab69`. The two open code findings from the previous round are addressed: the `dynamic_vlan` → `dynamic_vlan_iface` rename removes the collision with the hostapd `dynamic_vlan` BSS option, and the new `delete this.handler_data[section]` on the remove path closes the leak (deleting the current key during a `for (k, v in obj)` loop is safe in ucode — the iterator has already advanced, and `ucv_free_object_entry()` fixes up live iterators anyway). Squashing the series also resolves the introduce-then-fix hygiene point.

## Commit checks

- `54aab69` "wifi-scripts: recognise dynam. hostapd vlan ifaces" — the diff matches the message and the `wifi-scripts:` prefix is now right, but the subject abbreviates for no reason: it is 49 characters, and the unabbreviated form the PR title already uses — `wifi-scripts: recognise dynamic hostapd vlan interfaces` — is 54, still well inside the usual limit. Having the commit subject and the PR title disagree makes the change harder to find later. Also a typo in the body: "would be, IMHO, not the the right place" → "not the right place".

Still open from earlier rounds, not repeated inline: `PKG_RELEASE` in `package/network/config/wifi-scripts/Makefile` is still `1` ([thread](https://github.com/openwrt/openwrt/pull/24670#discussion_r3763629891)). CI on `54aab69` is green.

---

package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
```diff
@@ -559,12 +560,73 @@ function hotplug(name, add)
...
+	for (let phy in phys) {
+		let phys_data = known_phys[phy.wiphy_name];
+		// skip if this radio is not part of this handlers'
+		if (!phys_data)
+			continue;
+		// in a multi-radio setup, check participating radios and skip if they do not
+		if (phys_data.radio >=0 && !(v.vif_radio_mask & (1 << phys_data.radio)))
+			continue;
+		// loop through all interfaces found
+		for (let k, v in nl80211interfaces) {
```
openwrt-ai: `v` is not in scope here — it is declared by the inner `for (let k, v in nl80211interfaces)` on the next line, and ucode confines for-in declarations to the loop construct itself ([`uc_compiler_compile_for_in()` enters a scope at compiler.c:2692 and leaves it at :2789](https://github.com/jow-/ucode/blob/73ffef75c653dc901c69678a27c67842d291e79e/compiler.c#L2683-L2692)). So `v.vif_radio_mask` compiles to a global load, and netifd runs these scripts with [`strict_declarations = true` (netifd `ucode.c:536`)](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/ucode.c#L535-L540) on top of the `'use strict'` on line 1, so the load raises `Reference error: access to undeclared variable v` ([`uc_vm_insn_load_var()`, vm.c:1213-1219](https://github.com/jow-/ucode/blob/73ffef75c653dc901c69678a27c67842d291e79e/vm.c#L1210-L1220)).

Short-circuiting hides this on single-radio phys (`phys_data.radio` is `-1`), so it throws on exactly the multi-radio hardware this hunk is meant to support: the first dynamic-VLAN hotplug aborts before the matching loop ever runs.

`vif_radio_mask` is a per-interface attribute of the `NL80211_CMD_GET_INTERFACE` dump anyway ([`nl80211.c:1095` in the ucode binding](https://github.com/jow-/ucode/blob/73ffef75c653dc901c69678a27c67842d291e79e/lib/nl80211.c#L1095)), not a wiphy attribute, so the test belongs inside the interface loop:

```suggestion
		// loop through all interfaces found
		for (let k, v in nl80211interfaces) {
			// in a multi-radio setup, check participating radios and skip if they do not
			if (phys_data.radio >= 0 && !(v.vif_radio_mask & (1 << phys_data.radio)))
				continue;
```

(also fixes the missing space in `>=0`; you may prefer to move the check down below the `v.wiphy != phy.wiphy` test so it only runs for interfaces of this phy)

---

openwrt-ai (commented):
Reviewed the one new commit since `54aab69`. The `check_vlan` nit is addressed, the `delete this.handler_data[dev]` change is correct (the synthetic entries are keyed by `dev`, so the `.staN` path now no longer risks removing a live VLAN entry), and `known_phys[data.phy] = { section, radio: ... }` picks up the `radio=` value `mac80211.sh:1213` puts into the handler data as intended. One blocking problem in the new `vif_radio_mask` check — details inline on line 600.

The commit message matches the diff; no commit-message findings this round.

Still open from earlier rounds, not repeated inline: `PKG_RELEASE` in `package/network/config/wifi-scripts/Makefile` ([thread](https://github.com/openwrt/openwrt/pull/24670#discussion_r3763629891)). No CI results on `0c9b865` yet.

---

package/network/config/wifi-scripts/files/lib/netifd/wireless-device.uc
```diff
@@ -559,12 +560,73 @@ function hotplug(name, add)
...
+			// check if this interface has the requested name, skip otherwise
+			if (v.ifname != name)
+				continue;
+			// check if this wiphy is this handler's, skip otherwise.
+			if (v.wiphy != phy.wiphy)
+				continue;
+			// in a multi-radio setup, check radios and skip if they do not participate.
```
openwrt-ai: nit: these two comments end with a full stop, while every other comment in this function — including the neighbours added by this same commit (`// loop through all interfaces found`, `// check if this interface has the requested name, skip otherwise`, `// Only handle vlans here`, `// Add this interface to our known list`) — does not.

```suggestion
			// check if this wiphy is this handler's, skip otherwise
			if (v.wiphy != phy.wiphy)
				continue;
			// in a multi-radio setup, check radios and skip if they do not participate
```

---

openwrt-ai (commented):
Reviewed the squashed head `72ec99c` (diffed against `0c9b865`). The blocking finding from the last round is fixed correctly: the `vif_radio_mask` test now sits inside `for (let k, v in nl80211interfaces)`, so `v` is a real loop-local rather than an undeclared-global load, and putting it after the `v.wiphy != phy.wiphy` test means it only runs for interfaces of this phy. The `>=0` spacing was picked up too.

I also chased down the one way this could still have misbehaved, and it holds up. `phys_data.radio` comes from `+(data.radio ?? -1)`, and `data.radio` is a *string* from `wireless_set_data`. In ucode `+("")` is `0`, not NaN — [`uc_number_parse_common()` accepts the empty string and returns `ucv_uint64_new(0)` (vallist.c:69-133)](https://github.com/jow-/ucode/blob/73ffef75c653dc901c69678a27c67842d291e79e/vallist.c#L69-L133) — so an empty `radio=` would have made `phys_data.radio` `0`, and since `vif_radio_mask` is `0` on every non-radio-split wiphy ([`wdev->radio_mask = BIT(wdev->wiphy->n_radio) - 1`, core.c:1464](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/net/wireless/core.c#L1464), kernel v6.18.44 = `KERNEL_PATCHVER` on main), `!(0 & 1)` would have skipped every AP_VLAN on ordinary single-radio hardware. It does not happen, because [`mac80211_set_suffix()` blanks `radio` only to build the suffixes and then restores it with `set_default radio -1`](``https://github.com/openwrt/openwrt/blob/72ec99c02612a6a74d57299f692fbafe4f029149/package/network/config/wifi-scripts/files/lib/netifd/wireless/mac80211.sh#L1143-L1148)`` before [`wireless_set_data phy="$phy" radio="$radio"` at mac80211.sh:1213](``https://github.com/openwrt/openwrt/blob/72ec99c02612a6a74d57299f692fbafe4f029149/package/network/config/wifi-scripts/files/lib/netifd/wireless/mac80211.sh#L1213).`` So `data.radio` is `"-1"` there and the check is skipped, as intended. Handlers that never call `wireless_set_data phy=…` don't populate `known_phys` at all, so they are unaffected.

The commit message matches the diff, the `wifi-scripts:` prefix and the PR title now agree, and the "the the" typo is gone — no commit-message findings this round.

Still open from earlier rounds, not repeated inline: `PKG_RELEASE` in `package/network/config/wifi-scripts/Makefile` is still `1` ([thread](https://github.com/openwrt/openwrt/pull/24670#discussion_r3763629891)). No CI results on `72ec99c` yet.

---

openwrt-ai (commented):

Diffed `72ec99c..70c9fe4`: the head is the same single commit rebased onto current `main`, with two trailing-period removals in the new comments inside `hotplug()` and the subject de-abbreviated to `wifi-scripts: recognise dynamic hostapd vlan ifaces`. No functional change since the last round, so the `vif_radio_mask`/`known_phys` analysis from that review still stands.

Still open from earlier rounds, not repeated inline: `PKG_RELEASE` in `package/network/config/wifi-scripts/Makefile` is still `1` ([thread](https://github.com/openwrt/openwrt/pull/24670#discussion_r3763629891)) — this adds new runtime behaviour to a shipped file, so it looks like a genuine bump rather than a cosmetic edit. CI on `70c9fe4` is green (13 checks, no failures).

---

PR #24671, opened 2026-08-11
mediatek: gl-mt2500: fix GPY211 rate adaptation

wpower28:
## Problem

The GL.iNet GL-MT2500 v1 using the MaxLinear GPY211C shows packet loss and
unstable Ethernet operation with vanilla OpenWrt when the copper link operates
below 2.5Gbps.

## Root cause

The GL-MT2500 v1 uses a fixed 2.5Gbps MAC-side connection between the MT7981
and GPY211C while the copper side may negotiate at a lower rate. The PHY
performs rate adaptation internally and uses PAUSE frames for backpressure.

## Fix

Add an explicit firmware-node opt-in for the GL-MT2500 v1. When enabled, it:

- sets `FIXED2G5` and clears `ANEN`;
- keeps the MAC-side interface at 2500BASE-X; and
- reports `RATE_MATCH_PAUSE`.

Without the opt-in, existing driver behaviour is unchanged. The property is
documented in the MaxLinear GPY2xx Device Tree binding.

## Scope

The opt-in is enabled only on the GL.iNet GL-MT2500 v1 MaxLinear GPY211C.
The callback is restricted to the GPY211C driver entry. Other GPY models and
the GL-MT2500 Airoha revision are unchanged.

## Attribution

The fixed-rate-adaptation mechanism is based on work by Charles Eidsness /
CCX Technologies:

- https://github.com/ccxtechnologies/linux-ccx-lts-6.18/commit/0a835f4b9c444ff9d42e56ea6635926adb841f3f
- https://github.com/ccxtechnologies/linux-ccx-lts-6.18/commit/ad3b47ccd6f4a842e3b3e81515b0def6178b0fcb

## Testing

Hardware validation was performed on the equivalent OpenWrt 25.xx / Linux
6.12 implementation on the GL.iNet GL-MT2500 v1 with a MaxLinear GPY211C.

Results:

- MMD30:0x08 = 0x24fa (`FIXED2G5=1`, `ANEN=0`);
- iperf3: 947 Mbps RX, 949 Mbps TX;
- simultaneous bidirectional: 931 Mbps / 816 Mbps;
- RX physical errors: 0;
- TCP retransmissions in the tested TX direction: 0;
- PAUSE counter delta: +7339.

The OpenWrt main / Linux 6.18 version was validated by patch preparation,
formal checks and kernel compilation. No additional hardware validation of the
development kernel is planned.

### DIFF
--- a/target/linux/generic/pending-6.18/970-net-phy-mxl-gpy-add-fixed-2.5G-rate-adaptation-opt-in.patch
+++ b/target/linux/generic/pending-6.18/970-net-phy-mxl-gpy-add-fixed-2.5G-rate-adaptation-opt-in.patch
@@ -1,0 +1,156 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Erick Mesquita <richerick@gmail.com>
+Date: Wed, 12 Aug 2026 12:00:00 -0300
+Subject: net: phy: mxl-gpy: add fixed 2.5G rate adaptation opt-in
+
+The GL.iNet GL-MT2500 v1 uses a MaxLinear GPY211C PHY with a fixed
+2.5Gbps MAC-side connection while copper may negotiate at
+10/100/1000/2500Mbps. In this design the PHY performs rate adaptation
+internally and uses PAUSE frames for MAC-side backpressure.
+
+Add a firmware-node opt-in for this hardware configuration. The property is
+accepted only for the GPY211C model. When the opt-in is present, keep the
+MAC-side interface at 2500BASE-X, set FIXED2G5, clear ANEN, and report
+RATE_MATCH_PAUSE. Without it, retain the existing dynamic GPY interface
+selection and rate-matching behavior.
+
+Based on fixed-rate-adaptation work by Charles Eidsness / CCX Technologies,
+adapted for the GL-MT2500 v1.
+
+Link: https://github.com/ccxtechnologies/linux-ccx-lts-6.18/commit/0a835f4b9c444ff9d42e56ea6635926adb841f3f
+Link: https://github.com/ccxtechnologies/linux-ccx-lts-6.18/commit/ad3b47ccd6f4a842e3b3e81515b0def6178b0fcb
+Signed-off-by: Erick Mesquita <richerick@gmail.com>
+
+---
+--- a/Documentation/devicetree/bindings/net/maxlinear,gpy2xx.yaml
++++ b/Documentation/devicetree/bindings/net/maxlinear,gpy2xx.yaml
+@@ -14,6 +14,15 @@ allOf:
+   - $ref: ethernet-phy.yaml#
+ 
+ properties:
++  maxlinear,fixed-2p5g-rate-adaptation:
++    description: |
++      Keep the MAC-side interface fixed at 2.5Gbps while the copper-side
++      link may negotiate at a lower rate. The PHY performs rate adaptation
++      internally and uses PAUSE frames for MAC-side backpressure.
++
++      This property is valid only for GPY211C PHYs.
++    type: boolean
++
+   maxlinear,use-broken-interrupts:
+     description: |
+       Interrupts are broken on some GPY2xx PHYs in that they keep the
+--- a/drivers/net/phy/mxl-gpy.c
++++ b/drivers/net/phy/mxl-gpy.c
+@@ -102,6 +102,7 @@
+ #define VSPEC1_SGMII_CTRL	0x08
+ #define VSPEC1_SGMII_CTRL_ANEN	BIT(12)		/* Aneg enable */
+ #define VSPEC1_SGMII_CTRL_ANRS	BIT(9)		/* Restart Aneg */
++#define VSPEC1_SGMII_CTRL_FIXED2G5	BIT(5)		/* Fixed 2.5G rate adaptation */
+ #define VSPEC1_SGMII_ANEN_ANRS	(VSPEC1_SGMII_CTRL_ANEN | \
+ 				 VSPEC1_SGMII_CTRL_ANRS)
+ 
+@@ -134,6 +135,7 @@ struct gpy_priv {
+ 	u8 fw_major;
+ 	u8 fw_minor;
+ 	u32 wolopts;
++	bool fixed_2p5g_rate_adaptation;
+ 
+ 	/* It takes 3 seconds to fully switch out of loopback mode before
+ 	 * it can safely re-enter loopback mode. Record the time when
+@@ -337,10 +339,35 @@ static int gpy_config_init(struct phy_de
+ 
+ static int gpy21x_config_init(struct phy_device *phydev)
+ {
++	struct gpy_priv *priv = phydev->priv;
++	int ret;
++
+ 	__set_bit(PHY_INTERFACE_MODE_2500BASEX, phydev->possible_interfaces);
+-	__set_bit(PHY_INTERFACE_MODE_SGMII, phydev->possible_interfaces);
++	if (priv->fixed_2p5g_rate_adaptation)
++		__clear_bit(PHY_INTERFACE_MODE_SGMII, phydev->possible_interfaces);
++	else
++		__set_bit(PHY_INTERFACE_MODE_SGMII, phydev->possible_interfaces);
++
++	ret = gpy_config_init(phydev);
++	if (ret < 0)
++		return ret;
++
++	if (!priv->fixed_2p5g_rate_adaptation) {
++		phydev->rate_matching = RATE_MATCH_NONE;
++		return 0;
++	}
++
++	/* Keep the MAC-side link at 2.5G and let the PHY rate match copper. */
++	ret = phy_modify_mmd(phydev, MDIO_MMD_VEND1, VSPEC1_SGMII_CTRL,
++			     VSPEC1_SGMII_CTRL_FIXED2G5 |
++			     VSPEC1_SGMII_CTRL_ANEN,
++			     VSPEC1_SGMII_CTRL_FIXED2G5);
++	if (ret < 0)
++		return ret;
++
++	phydev->rate_matching = RATE_MATCH_PAUSE;
+ 
+-	return gpy_config_init(phydev);
++	return 0;
+ }
+ 
+ static int gpy_probe(struct phy_device *phydev)
+@@ -361,6 +388,10 @@ static int gpy_probe(struct phy_device *
+ 		return -ENOMEM;
+ 	phydev->priv = priv;
+ 	mutex_init(&priv->mbox_lock);
++	if (phydev->drv->phy_id == PHY_ID_GPY211C)
++		priv->fixed_2p5g_rate_adaptation =
++			device_property_read_bool(dev,
++						  "maxlinear,fixed-2p5g-rate-adaptation");
+ 
+ 	if (!device_property_present(dev, "maxlinear,use-broken-interrupts"))
+ 		phydev->dev_flags |= PHY_F_NO_IRQ;
+@@ -577,6 +608,7 @@ static int gpy_update_mdix(struct phy_de
+ 
+ static int gpy_update_interface(struct phy_device *phydev)
+ {
++	struct gpy_priv *priv = phydev->priv;
+ 	int ret;
+ 
+ 	/* Interface mode is fixed for USXGMII and integrated PHY */
+@@ -584,6 +616,11 @@ static int gpy_update_interface(struct p
+ 	    phydev->interface == PHY_INTERFACE_MODE_INTERNAL)
+ 		return 0;
+ 
++	if (priv->fixed_2p5g_rate_adaptation) {
++		phydev->interface = PHY_INTERFACE_MODE_2500BASEX;
++		return 0;
++	}
++
+ 	/* Automatically switch SERDES interface between SGMII and 2500-BaseX
+ 	 * according to speed. Disable ANEG in 2500-BaseX mode.
+ 	 */
+@@ -1067,6 +1104,18 @@ static int gpy_config_inband(struct phy_
+ 			      VSPEC1_SGMII_ANEN_ANRS);
+ }
+ 
++static int gpy21x_get_rate_matching(struct phy_device *phydev,
++				    phy_interface_t interface)
++{
++	struct gpy_priv *priv = phydev->priv;
++
++	if (priv->fixed_2p5g_rate_adaptation &&
++	    interface == PHY_INTERFACE_MODE_2500BASEX)
++		return RATE_MATCH_PAUSE;
++
++	return RATE_MATCH_NONE;
++}
++
+ static struct phy_driver gpy_drivers[] = {
+ 	{
+ 		PHY_ID_MATCH_MODEL(PHY_ID_GPY2xx),
+@@ -1189,6 +1238,7 @@ static struct phy_driver gpy_drivers[] =
+ 		.led_hw_control_get = gpy_led_hw_control_get,
+ 		.led_hw_control_set = gpy_led_hw_control_set,
+ 		.led_polarity_set = gpy_led_polarity_set,
++		.get_rate_matching = gpy21x_get_rate_matching,
+ 	},
+ 	{
+ 		.phy_id		= PHY_ID_GPY212B,
--- a/target/linux/mediatek/dts/mt7981b-glinet-gl-mt2500-v1.dts
+++ b/target/linux/mediatek/dts/mt7981b-glinet-gl-mt2500-v1.dts
@@ -16,5 +16,6 @@
 	phy5: ethernet-phy@5 {
 		reg = <5>;
 		compatible = "ethernet-phy-ieee802.3-c45";
+		maxlinear,fixed-2p5g-rate-adaptation;
 	};
 };


target/linux/generic/pending-6.18/970-net-phy-mxl-gpy-add-fixed-2.5G-rate-adaptation-opt-in.patch
openwrt-ai, line 1: Wrong patch directory: `drivers/net/phy/mxl-gpy.c` is a generic PHY driver, not mediatek-specific code, so this doesn't belong in `target/linux/mediatek/patches-6.18/`. Every other in-tree change to this driver lives under `target/linux/generic/` — e.g. [730-11-v7.0-net-phy-mxl-gpy-implement-SGMII-in-band-configuratio.patch](``https://github.com/openwrt/openwrt/blob/ebb15bb13791daf695c671442eaeb56bf7b414ec/target/linux/generic/backport-6.18/730-11-v7.0-net-phy-mxl-gpy-implement-SGMII-in-band-configuratio.patch)`` and [761-v6.19-net-phy-mxl-gpy-add-support-for-mxl86252-and-mxl86282.patch](``https://github.com/openwrt/openwrt/blob/ebb15bb13791daf695c671442eaeb56bf7b414ec/target/linux/generic/backport-6.18/761-v6.19-net-phy-mxl-gpy-add-support-for-mxl86252-and-mxl86282.patch).`` Since this is not upstream yet, `target/linux/generic/pending-6.18/` is the conventional home, and it also makes the fix reachable for the other targets that ship GPY2xx PHYs.

Two follow-ons if you move it:

* `pending-6.18/` patches all use the trimmed quilt header — `From:` / `Date:` / `Subject:` only, no `From <sha> Mon Sep 17 00:00:00 2001` line and no `---` diffstat block (all 60+ files in that directory start with `From:`). See [742-net-phy-realtek-add-5G-and-10G-PHY-support.patch](``https://github.com/openwrt/openwrt/blob/ebb15bb13791daf695c671442eaeb56bf7b414ec/target/linux/generic/pending-6.18/742-net-phy-realtek-add-5G-and-10G-PHY-support.patch)`` for the expected shape.
* The commit message carries no `Link:` to a netdev posting. This adds a *new* DT binding property; if it lands upstream under a different name later, the DTS in this PR becomes wrong and has to be churned. Has `maxlinear,fixed-2p5g-rate-adaptation` been submitted to netdev / the DT maintainers?

---

target/linux/mediatek/patches-6.18/970-net-phy-mxl-gpy-add-fixed-2.5G-rate-adaptation-opt-in.patch
```diff
@@ -0,0 +1,147 @@
...
+-	/* Interface mode is fixed for USXGMII and integrated PHY */
++	if (priv->fixed_2p5g_rate_adaptation) {
++		phydev->interface = PHY_INTERFACE_MODE_2500BASEX;
++		return 0;
++	}
++
++	/* Interface mode is fixed for USXGMII and integrated PHY. */
+ 	if (phydev->interface == PHY_INTERFACE_MODE_USXGMII ||
+ 	    phydev->interface == PHY_INTERFACE_MODE_INTERNAL)
+ 		return 0;
```
openwrt-ai: The new early return is placed *above* the USXGMII/INTERNAL guard, so it now overrides the fixed-interface case: a PHY running USXGMII or as an integrated PHY that happens to have the property set gets `phydev->interface` rewritten to 2500BASEX, which the original code explicitly refuses to do. Put the opt-in after the guard so the existing invariant still holds. This also avoids the unrelated punctuation change to the existing comment.

```suggestion
 	/* Interface mode is fixed for USXGMII and integrated PHY */
 	if (phydev->interface == PHY_INTERFACE_MODE_USXGMII ||
 	    phydev->interface == PHY_INTERFACE_MODE_INTERNAL)
 		return 0;
+
+	if (priv->fixed_2p5g_rate_adaptation) {
+		phydev->interface = PHY_INTERFACE_MODE_2500BASEX;
+		return 0;
+	}
```

(The hunk header counts stay valid — this only reorders lines within the hunk.)

---

target/linux/generic/pending-6.18/970-net-phy-mxl-gpy-add-fixed-2.5G-rate-adaptation-opt-in.patch
```diff
@@ -0,0 +1,147 @@
...
++}
++
+ static struct phy_driver gpy_drivers[] = {
+ 	{
+ 		PHY_ID_MATCH_MODEL(PHY_ID_GPY2xx),
+@@ -1189,6 +1227,7 @@ static struct phy_driver gpy_drivers[] =
+ 		.led_hw_control_get = gpy_led_hw_control_get,
+ 		.led_hw_control_set = gpy_led_hw_control_set,
+ 		.led_polarity_set = gpy_led_polarity_set,
++		.get_rate_matching = gpy21x_get_rate_matching,
```
openwrt-ai, line 153: `.get_rate_matching` is wired up on the GPY211C entry only, but nothing else in the patch is scoped to GPY211C:

* `gpy_probe()` (line 82 of this patch) reads the property for **every** GPY model — it is the shared `.probe` for all 13 driver entries.
* `gpy21x_config_init()` (line 71) writes `FIXED2G5` / clears `ANEN`, and is the `.config_init` of GPY211B, GPY211C, GPY212B, GPY215B and the GPY24x entries — see [mxl-gpy.c:1094](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/mxl-gpy.c#L1094), [:1139](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/mxl-gpy.c#L1139), [:1161](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/mxl-gpy.c#L1161), [:1184](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/mxl-gpy.c#L1184), [:1206](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/mxl-gpy.c#L1206) (v6.18.41).
* `gpy_update_interface()` (line 96) is on the shared `gpy_read_status()` path for every entry.
* The binding hunk documents the property under `maxlinear,gpy2xx.yaml` with no per-model restriction.

So on any GPY that is not a GPY211C, setting this property pins the SERDES to fixed 2.5G and forces `phydev->interface = 2500BASEX`, while phylink still gets `RATE_MATCH_NONE` and therefore never enables the TX pause backpressure the design relies on — i.e. exactly the packet-loss failure mode this PR is fixing, silently. The PR body states the restriction is deliberate but not why. Either add `.get_rate_matching` to the other `gpy21x_config_init` users, or gate the property itself on GPY211C in `gpy_probe()` and say so in the binding.

---

target/linux/mediatek/patches-6.18/970-net-phy-mxl-gpy-add-fixed-2.5G-rate-adaptation-opt-in.patch
```diff
@@ -0,0 +1,147 @@
...
+ drivers/net/phy/mxl-gpy.c | 48 +++++++++++++++++++++++++++++++++++++--
+ 1 file changed, 46 insertions(+), 2 deletions(-)
+
+--- a/drivers/net/phy/mxl-gpy.c
++++ b/drivers/net/phy/mxl-gpy.c
+@@ -1,6 +1,7 @@
+ // SPDX-License-Identifier: GPL-2.0+
+ /* Copyright (C) 2021 Maxlinear Corporation
+  * Copyright (C) 2020 Intel Corporation
++ * Copyright (C) 2026 CCX Technologies
```
openwrt-ai: nit: this asserts copyright for a third party (CCX Technologies) while the only sign-off on the patch is `Signed-off-by: Erick Mesquita`. The commit body already credits the work as "based on" Charles Eidsness / CCX Technologies with two `Link:` trailers, which is the appropriate attribution for a re-implementation. If code was actually taken from those commits, the right form is `Co-developed-by:` + `Signed-off-by:` from Charles Eidsness (which needs his agreement); otherwise drop the copyright line — it will not survive netdev review as-is.

---

openwrt-ai (commented):
## Commit checks

- `ebb15bb` "mediatek: filogic: gl-mt2500: fix GPY211 rate adaptation" — the `filogic:` scope is too narrow for what the diff touches. The DTS change is filogic-only, but `target/linux/mediatek/patches-6.18/` is shared by every mediatek subtarget (mt7622, mt7623, mt7629 and filogic), so the kernel patch is applied target-wide. `mediatek: ...` would match the files touched; see also the inline note suggesting the patch belongs in `target/linux/generic/pending-6.18/` instead, which would widen the scope further.

Verified locally that `970-net-phy-mxl-gpy-add-fixed-2.5G-rate-adaptation-opt-in.patch` applies cleanly on top of the existing `target/linux/generic/backport-6.18/730-10`, `730-11` and `761` mxl-gpy patches, and that the `.get_rate_matching` hunk lands in the GPY211C driver entry as intended. Hunk headers and counts are correct. CI on `ebb15bb` is green.

---

target/linux/generic/pending-6.18/970-net-phy-mxl-gpy-add-fixed-2.5G-rate-adaptation-opt-in.patch
```diff
@@ -0,0 +1,142 @@
...
++			      VSPEC1_SGMII_CTRL_ANEN,
++			      VSPEC1_SGMII_CTRL_FIXED2G5);
+ }
+ 
+ static int gpy_probe(struct phy_device *phydev)
+@@ -361,6 +377,10 @@ static int gpy_probe(struct phy_device *
+ 		return -ENOMEM;
+ 	phydev->priv = priv;
+ 	mutex_init(&priv->mbox_lock);
++	if ((phydev->phy_id & PHY_ID_MATCH_MODEL_MASK) == PHY_ID_GPY211C)
```
openwrt-ai: `phydev->phy_id` is `0` for a PHY probed as Clause 45, so this guard never matches on the board the property targets — the opt-in becomes dead code.

[`mt7981b-glinet-gl-mt2500-v1.dts:18`](https://github.com/openwrt/openwrt/blob/79ba5306c430960c38dedc5bdcb6fba7383dc8aa/target/linux/mediatek/dts/mt7981b-glinet-gl-mt2500-v1.dts#L18) declares `compatible = "ethernet-phy-ieee802.3-c45"`. That makes [`fwnode_mdiobus_register_phy()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/mdio/fwnode_mdio.c#L136-L138) call `get_phy_device(bus, addr, true)`, where `phy_id` is left at its [`u32 phy_id = 0;`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/phy_device.c#L1110) initialiser and stored verbatim by [`phy_device_create()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/phy_device.c#L828). The GPY211C driver entry still binds because C45 matching goes through `phydev->c45_ids.device_ids[]` in [`genphy_match_phy_device()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/phy_device.c#L565-L586) — but `phydev->phy_id` stays `0`, so `priv->fixed_2p5g_rate_adaptation` is never set and neither the `FIXED2G5` write, the forced 2500BASE-X, nor `RATE_MATCH_PAUSE` ever take effect.

`phydev->drv` is already assigned before `.probe` runs ([`phy_probe()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/phy_device.c#L3401)), and the GPY211C entry uses `PHY_ID_MATCH_MODEL(PHY_ID_GPY211C)` (i.e. `.phy_id = PHY_ID_GPY211C`), so keying off the matched driver works for both C22- and C45-probed instances and keeps the line count of the hunk unchanged:

```suggestion
+	if (phydev->drv->phy_id == PHY_ID_GPY211C)
```

---

target/linux/generic/pending-6.18/970-net-phy-mxl-gpy-add-fixed-2.5G-rate-adaptation-opt-in.patch
```diff
@@ -0,0 +1,142 @@
...
++				    phy_interface_t interface)
++{
++	struct gpy_priv *priv = phydev->priv;
++
++	if (priv->fixed_2p5g_rate_adaptation &&
++	    interface == PHY_INTERFACE_MODE_2500BASEX)
++		return RATE_MATCH_PAUSE;
++
++	return RATE_MATCH_NONE;
++}
```
openwrt-ai, line 144: Implementing `.get_rate_matching` alone does not put phylink into rate-matching mode at runtime — the driver must also set the `phydev->rate_matching` field.

phylink reads the field, not the callback, on every link change: [`pl->phy_state.rate_matching = phydev->rate_matching;` in `phylink_phy_change()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/phylink.c#L1973), and [`phylink_resolve()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/phylink.c#L1665-L1670) only substitutes the fixed interface speed / forced RX pause when `pl->phy_state.rate_matching` is non-zero. The callback is consulted only during validation ([`phylink_validate_phy()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/phylink.c#L2052) and `phylink_validate_one()`).

So with this patch as written, `mac_link_up()` still receives the copper speed (e.g. `SPEED_1000`) together with `interface = 2500base-x` and un-forced pause — exactly the MAC/SerDes rate mismatch the patch is meant to remove.

All three in-tree `RATE_MATCH_PAUSE` PHY drivers set the field in addition to providing the callback:
- [`realtek_main.c:1236-1244`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/realtek/realtek_main.c#L1236-L1244)
- [`motorcomm.c:2786-2796`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/motorcomm.c#L2786-L2796) — the YT8821 `FORCE_BX2500` case is structurally identical to this one
- [`air_en8811h.c:1086`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/air_en8811h.c#L1086)

Please set `phydev->rate_matching = RATE_MATCH_PAUSE;` on the opt-in path in `gpy21x_config_init()` (leaving it `RATE_MATCH_NONE` otherwise). That changes the `@@ -337,10 +339,24 @@` hunk's line count, so the patch will need regenerating afterwards — `make target/linux/refresh` with a mediatek/filogic `.config`.

Since the PR body states the 6.18 variant was only compile-validated and the +7339 PAUSE-delta figures come from the 6.12 implementation, it would be worth confirming whether that earlier version set `phydev->rate_matching` — if it did, those numbers don't carry over to this one. (Compared against gregkh/linux v6.18.44.)

---

target/linux/generic/pending-6.18/970-net-phy-mxl-gpy-add-fixed-2.5G-rate-adaptation-opt-in.patch
```diff
@@ -0,0 +1,142 @@
...
+ 	 * it can safely re-enter loopback mode. Record the time when
+@@ -337,10 +339,24 @@ static int gpy_config_init(struct phy_de
+ 
+ static int gpy21x_config_init(struct phy_device *phydev)
+ {
++	struct gpy_priv *priv = phydev->priv;
++	int ret;
++
+ 	__set_bit(PHY_INTERFACE_MODE_2500BASEX, phydev->possible_interfaces);
+ 	__set_bit(PHY_INTERFACE_MODE_SGMII, phydev->possible_interfaces);
```
openwrt-ai: nit: with the opt-in active the SerDes never leaves 2500BASE-X (`gpy_update_interface()` returns early), yet SGMII is still advertised in `possible_interfaces`, so phylink validates against a mode this configuration cannot reach. The closest in-tree analogue splits this: motorcomm only adds SGMII in the `AUTO_BX2500_SGMII` case and not in `FORCE_BX2500` ([`motorcomm.c:2786-2796`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/motorcomm.c#L2786-L2796)).

It looks harmless for the supported-linkmode union here (`RATE_MATCH_PAUSE` already widens caps down to 10M), so is leaving SGMII advertised intentional?

---

openwrt-ai (commented):
## Commit checks

- `79ba530` "mediatek: gl-mt2500: fix GPY211 rate adaptation" — the patch moved from `target/linux/mediatek/patches-6.18/` to `target/linux/generic/pending-6.18/` since `ebb15bb` (good — that addresses the earlier note), but the subject scope wasn't updated to follow. A `generic/pending-6.18/` patch is applied to every target building 6.18, so `mediatek:` is now narrower than what the commit does. In-tree convention for commits touching `target/linux/generic/pending-*` is a `kernel:` prefix — e.g. `0256e6a1e47` "kernel: backport quirk for Horaco SFP+ Copper module", `7926c03b4da` "kernel: pse-pd: add poll helper and LED trigger support". Consider splitting into `kernel: net: phy: mxl-gpy: ...` for the driver patch and `mediatek: filogic: gl-mt2500: ...` for the DTS opt-in.

---

Re-verified against the reworked patch: it applies cleanly (no fuzz, no offset) on top of the existing `backport-6.18/730-10`, `730-11` and `761` mxl-gpy patches over v6.18.44, the hunk headers and counts are correct, `PHY_ID_MATCH_MODEL_MASK` exists in `include/linux/phy.h`, and `.get_rate_matching` does land in the GPY211C driver entry as intended. CI on `79ba530` has no failures.

Two functional issues are flagged inline: the `phydev->phy_id` guard in `gpy_probe()` never matches for this C45-declared PHY, and `phydev->rate_matching` is never assigned so phylink's rate-matching path stays inactive at runtime.

---

target/linux/generic/pending-6.18/970-net-phy-mxl-gpy-add-fixed-2.5G-rate-adaptation-opt-in.patch
```diff
@@ -0,0 +1,158 @@
...
+adapted for the GL-MT2500 v1.
+
+Link: https://github.com/ccxtechnologies/linux-ccx-lts-6.18/commit/0a835f4b9c444ff9d42e56ea6635926adb841f3f
+Link: https://github.com/ccxtechnologies/linux-ccx-lts-6.18/commit/ad3b47ccd6f4a842e3b3e81515b0def6178b0fcb
+Signed-off-by: Erick Mesquita <richerick@gmail.com>
+
+Index: linux-6.18.44/Documentation/devicetree/bindings/net/maxlinear,gpy2xx.yaml
+===================================================================
+--- linux-6.18.44.orig/Documentation/devicetree/bindings/net/maxlinear,gpy2xx.yaml
++++ linux-6.18.44/Documentation/devicetree/bindings/net/maxlinear,gpy2xx.yaml
```
openwrt-ai: These `Index:` / `===` headers with version-pinned paths were not produced by OpenWrt's patch tooling and will be rewritten the first time anyone refreshes this patch. The project refreshes with [`QUILT_DIFF_OPTS="-p" $(QUILT_CMD) refresh -p ab --no-index --no-timestamps` at `include/quilt.mk:173`](https://github.com/openwrt/openwrt/blob/c0470ce7f8c634c0b5765f17818f0955ce78c258/include/quilt.mk#L173) — `-p ab` produces `a/` / `b/` prefixes and `--no-index` explicitly suppresses the `Index:` line. Consistently so in-tree: 0 of the 187 patches in `target/linux/generic/pending-6.18/` (and 0 anywhere under `target/linux/`) carry an `Index:` header, while 185 use `--- a/` / `+++ b/`.

Hard-coding `linux-6.18.44` also dates the patch against a specific stable release it does not actually depend on.

It still applies cleanly at `-p1`, so this is churn rather than breakage, but it should go back to the refreshed form:

```suggestion
--- a/Documentation/devicetree/bindings/net/maxlinear,gpy2xx.yaml
+++ b/Documentation/devicetree/bindings/net/maxlinear,gpy2xx.yaml
```

Same applies to lines 43-46 for `drivers/net/phy/mxl-gpy.c`. Easiest route is `make target/linux/refresh` with a mediatek/filogic `.config`, which will regenerate both file headers correctly.

---

openwrt-ai (commented):
## Commit checks

- `c0470ce` "mediatek: gl-mt2500: fix GPY211 rate adaptation" — subject scope still doesn't match the files touched: the driver patch lives in `target/linux/generic/pending-6.18/`, which applies to every target building 6.18, so `mediatek:` is narrower than what the commit does. Same note as on `79ba530`; repeating only because the commit was amended without picking it up. `kernel: net: phy: mxl-gpy: ...` (plus a separate `mediatek: filogic: gl-mt2500: ...` for the DTS opt-in) matches in-tree convention.

---

Both functional issues from the previous round are fixed, verified against `gregkh/linux` v6.18.44 with `backport-6.18/730-10`, `730-11` and `761` applied first:

- `gpy_probe()` now keys off `phydev->drv->phy_id == PHY_ID_GPY211C`, which resolves for the C45-probed PHY on this board (the GPY211C entry uses `PHY_ID_MATCH_MODEL(PHY_ID_GPY211C)`, so `drv->phy_id` is exactly `PHY_ID_GPY211C`).
- `phydev->rate_matching = RATE_MATCH_PAUSE;` is now assigned on the opt-in path, so phylink's `phylink_phy_change()` / `phylink_resolve()` path actually engages.
- The `possible_interfaces` question from the earlier nit is addressed too: SGMII is now cleared when the opt-in is active, matching the motorcomm `FORCE_BX2500` shape.

Also re-checked for interactions with the other `VSPEC1_SGMII_CTRL` writers, since `FIXED2G5` has to survive them: `gpy_inband_caps()` returns `LINK_INBAND_DISABLE` for 2500BASE-X so `gpy_config_inband()` only clears `ANEN|ANRS` and leaves `FIXED2G5` set, and `gpy_config_aneg()` bails at `!gpy_sgmii_aneg_en()` once `ANEN` is cleared, so the SGMII AN retrigger never runs. `priv` is safe to dereference unconditionally in `gpy_update_interface()` — all 14 driver entries set `.probe = gpy_probe`. All nine hunk headers have correct line counts and the patch applies with no fuzz or offset. CI on `c0470ce` has no failures.

One packaging issue is flagged inline.

---

target/linux/generic/pending-6.18/970-net-phy-mxl-gpy-add-fixed-2.5G-rate-adaptation-opt-in.patch
```diff
@@ -0,0 +1,154 @@
+From: Erick Mesquita <richerick@gmail.com>
```
openwrt-ai, line 2: The `Index:`/`===` headers are gone now — good, that resolves the previous round's note. But the file still trips the project's own formality check, which ran on this exact commit:

```
Commit 390d715 - mediatek: gl-mt2500: fix GPY211 rate adaptation:
 - Embedded patch file 'target/linux/generic/pending-6.18/970-...-opt-in.patch'
   violates standard guidelines. Missing required Git header parameters
   ('From <hash> Mon Sep 17 00:00:00 2001' / 'From:' / 'Date:' / 'Subject:')
   to ensure 'git am' application compatibility
```

`From:`, `Date:` and `Subject:` are present, so the missing one is the leading `From <hash> Mon Sep 17 00:00:00 2001` line. `check_patch_headers` is enabled in [`.github/formalities.json`](https://github.com/openwrt/openwrt/blob/390d71571cf61037f1aaf2eff766fa45012f9d89/.github/formalities.json#L21).

Correcting my own earlier advice on this file: I previously said `pending-6.18/` uses the trimmed `From:`-first header. That was wrong as a majority claim — of the 188 patches in that directory, 110 start with the full `From <sha> Mon Sep 17 00:00:00 2001` line and only 76 start with `From:`. Recent locally-authored additions use the full form, e.g. [`897-01-net-pse-pd-add-devm_pse_poll_helper.patch:1`](``https://github.com/openwrt/openwrt/blob/390d71571cf61037f1aaf2eff766fa45012f9d89/target/linux/generic/pending-6.18/897-01-net-pse-pd-add-devm_pse_poll_helper.patch#L1).`` Since the bot flags the trimmed form, the full form is the one to use.

The hash is arbitrary as far as `git am` is concerned (it is only the "from" line of a mailbox message); `git format-patch` on your local kernel commit emits it for you. Replace the placeholder below with whatever `format-patch` produces:

```suggestion
From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
From: Erick Mesquita <richerick@gmail.com>
```

Optional, for full `format-patch` shape: the same reference patch also puts a bare `---` line between the `Signed-off-by:` trailer and the first `--- a/` file header ([line 50](``https://github.com/openwrt/openwrt/blob/390d71571cf61037f1aaf2eff766fa45012f9d89/target/linux/generic/pending-6.18/897-01-net-pse-pd-add-devm_pse_poll_helper.patch#L50)).`` The bot does not check for it, and `make target/linux/refresh` preserves everything above the first file header either way.

---

openwrt-ai (commented):
## Commit checks

- `390d715` "mediatek: gl-mt2500: fix GPY211 rate adaptation" — subject scope still doesn't match the files touched. Third round on the same point, so this is the last time I raise it: the driver patch lives in `target/linux/generic/pending-6.18/`, which applies to every target building 6.18, while `mediatek:` claims a single target. Of the last 25 commits touching that directory, every one uses `kernel:` or `generic:` — `53a0d54` "kernel: bump 6.18 to 6.18.44", `0256e6a` "kernel: backport quirk for Horaco SFP+ Copper module", `7926c03` "kernel: pse-pd: add poll helper and LED trigger support", `96ecf8d` "generic: pse-pd: add patches for module-based PSE drivers". This commit is the only outlier. A split into `kernel: net: phy: mxl-gpy: ...` (driver patch) plus `mediatek: filogic: gl-mt2500: ...` (DTS opt-in) matches convention; a single `kernel: ...` commit would at least not misstate the scope.

---

The only change since `c0470ce` is the patch-header rework, which drops the `Index:`/`===` block and restores `--- a/` / `+++ b/` — that was the issue flagged last round, and it's fixed.

Re-verified the patch against `gregkh/linux` v6.18.44 with `backport-6.18/730-10`, `730-11` and `761` applied first: it applies with no fuzz and no offset on both files, and the hunk headers and counts are unchanged from the previous round (the edit touched only the mail header). The functional fixes from earlier rounds — `phydev->drv->phy_id == PHY_ID_GPY211C` in `gpy_probe()`, `phydev->rate_matching = RATE_MATCH_PAUSE` on the opt-in path, and clearing SGMII from `possible_interfaces` — are all still in place, so the earlier GPY211C-scoping thread is addressed. CI on `390d715` has no failures.

One packaging issue is flagged inline, along with a correction to my own earlier advice on that file.

---

openwrt-ai (commented):

The only change since `390d715` is the patch mail header: the `From 000…0 Mon Sep 17 00:00:00 2001` line and the bare `---` separator before the first file header were added. That is exactly what was flagged last round, and `FormalityCheck / Code Patches` is now green on `1153fc8` (all three FormalityCheck runs pass). The all-zeros placeholder hash is fine — 16 other in-tree patches under `target/linux/` use it, including [`pending-6.18/711-04-net-dsa-qca8k-no-program-unicast-host-FDB-entries.patch`](``https://github.com/openwrt/openwrt/blob/1153fc89940a950446d42f1cef4c41062796bb7c/target/linux/generic/pending-6.18/711-04-net-dsa-qca8k-no-program-unicast-host-FDB-entries.patch#L1).``

Re-verified the patch body against `gregkh/linux` v6.18.44 (`1efe5d0`) with `backport-6.18/730-10`, `730-11` and `761` applied first: it applies with zero fuzz and zero offset on both files, and `git apply --check` accepts it, so all nine hunk headers and counts are exact. Nothing in the diff below the header changed, so the functional fixes from earlier rounds are all still in place — `phydev->drv->phy_id == PHY_ID_GPY211C` in `gpy_probe()`, `phydev->rate_matching = RATE_MATCH_PAUSE` on the opt-in path, SGMII cleared from `possible_interfaces`, and the `gpy_update_interface()` early return sitting below the USXGMII/INTERNAL guard rather than above it.

The commit subject scope point is unchanged, but I said last round I would stop raising it, so it is not repeated here. The one substantive question still open from the first round is whether `maxlinear,fixed-2p5g-rate-adaptation` has been posted to netdev / the DT maintainers — the DTS in this PR bakes in the property name, so it needs churning if upstream lands it under a different one. That thread is still unresolved rather than newly raised.

---

PR #24672, opened 2026-08-11
mold: update to 2.42.0

schuettecarsten:
mold: Update to 2.42.0

Changelog:
https://github.com/rui314/mold/releases/tag/v2.42.0

### DIFF
--- a/tools/mold/Makefile
+++ b/tools/mold/Makefile
@@ -3,13 +3,13 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mold
-PKG_VERSION:=2.41.0
+PKG_VERSION:=2.42.0
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL_FILE:=v$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://github.com/rui314/mold/archive/refs/tags
-PKG_HASH:=0a61abac85d818437b425df856822e9d6e9982baeae5a93bcb02fe6c0060c61a
+PKG_HASH:=6c0f3308c5b3159a369202d970922ad819bab1bfcb5a3b3c06a723d19f65373e
 
 include $(INCLUDE_DIR)/host-build.mk
 include $(INCLUDE_DIR)/cmake.mk


openwrt-ai (commented):
## Commit checks

- 2d3f5bb "toolchain: mold: bump to 2.42.0" — the subject scope doesn't match the files touched. The diff only changes [`tools/mold/Makefile`](https://github.com/openwrt/openwrt/blob/2d3f5bb931ab3f82a62c8d357a7546f63b7fc52b/tools/mold/Makefile), while [`toolchain/mold/Makefile`](https://github.com/openwrt/openwrt/blob/2d3f5bb931ab3f82a62c8d357a7546f63b7fc52b/toolchain/mold/Makefile) is a separate, untouched in-tree package, so the `toolchain:` prefix points at the wrong directory. The last bump of this same file used `mold: Update to 2.41.0` (de41c05e), and the prevailing `tools/` convention is `tools/<name>: update to <ver>` (e.g. d70761c2 `tools/isl: update to 0.28`) — `tools/mold: update to 2.42.0` would be unambiguous. (There is one older precedent, 06b81f5d `toolchain: mold: Update to 2.33.0`, that also bumped `tools/mold/Makefile`, so this is a readability point rather than a hard rule.)
- 2d3f5bb — the commit body's first line repeats the subject verbatim, so the message reads `toolchain: mold: bump to 2.42.0` twice before the changelog link. The 2.41.0 commit body used a distinct sentence instead. Also note the PR title says "update to 2.42.0" while the commit subject says "bump to 2.42.0"; worth settling on one wording.

Verification of the version bump itself (no issues found): tag `v2.42.0` exists upstream and is the newest `v2.4x` tag (`git ls-remote https://github.com/rui314/mold.git`), and the new `PKG_HASH` reproduces exactly from that tag's tree, so the checksum is correct. The three CMake options this Makefile sets (`MOLD_LTO`, `MOLD_USE_SYSTEM_MIMALLOC`, `MOLD_USE_SYSTEM_TBB`) all still exist in 2.42.0's `CMakeLists.txt`, the release adds no git submodules (bundled deps stay vendored in the tarball), and the bundled zlib's `cmake_minimum_required` moved from `2.4.4...3.15.0` to `3.12...3.31`, which is friendlier to the tree's CMake 4.x — so no build-side follow-up appears needed.

---

schuettecarsten:
Tested with GCC 15 and GCC 16 (PR pending) on bcm27xx/bcm2711 and bcm27xx/bcm2712

schuettecarsten:
@robimarko Can we get this merged? You also merged the 2.41.0 changes 4 month ago... ;-)

robimarko:

PR #24673, opened 2026-08-11
airoha: configure network activity LEDs for Nokia XG-040G-MD

csharper2005:
Set up network device LEDs for lan2, lan3, and lan4 interfaces on the Nokia XG-040G-MD board to display network TX and RX activity.

### DIFF
--- a/target/linux/airoha/an7581/base-files/etc/board.d/01_leds
+++ b/target/linux/airoha/an7581/base-files/etc/board.d/01_leds
@@ -19,6 +19,12 @@ gemtek,w1700k-ubi)
 	ucidef_set_led_netdev "lan4-yellow" "lan4" "yellow:lan-4" "lan4" "link_10 link_100"
 	ucidef_set_led_netdev "lan4-green" "lan4" "green:lan-4" "lan4" "link_1000 tx rx"
 	;;
+nokia,xg-040g-md|\
+nokia,xg-040g-md-ubi)
+	ucidef_set_led_netdev "lan2" "lan2" "mt7530-0:0a:green:lan-2" "lan2" "link tx rx"
+	ucidef_set_led_netdev "lan3" "lan3" "mt7530-0:0b:green:lan-3" "lan3" "link tx rx"
+	ucidef_set_led_netdev "lan4" "lan4" "mt7530-0:0c:green:lan-4" "lan4" "link tx rx"
+	;;
 esac
 
 board_config_flush


target/linux/airoha/an7581/base-files/etc/board.d/01_leds
```diff
@@ -19,6 +19,11 @@ gemtek,w1700k-ubi)
 	ucidef_set_led_netdev "lan4-yellow" "lan4" "yellow:lan-4" "lan4" "link_10 link_100"
 	ucidef_set_led_netdev "lan4-green" "lan4" "green:lan-4" "lan4" "link_1000 tx rx"
 	;;
+nokia,xg-040g-md)
```
openwrt-ai: The UBI variant is a separate board and won't match this case, so it gets no LED configuration. [`an7581-nokia_xg-040g-md-ubi.dts`](https://github.com/openwrt/openwrt/blob/afd4aeb1f688f8e3eca6e89d77e7e7c5c9e78f50/target/linux/airoha/dts/an7581-nokia_xg-040g-md-ubi.dts#L8-L10) declares `compatible = "nokia,xg-040g-md-ubi", ...` and is built as its own device ([`an7581.mk:190`](https://github.com/openwrt/openwrt/blob/afd4aeb1f688f8e3eca6e89d77e7e7c5c9e78f50/target/linux/airoha/image/an7581.mk#L190)), yet it includes the same `an7581-nokia_xg-040g-md-common.dtsi` and therefore has the identical `gsw_phy2/3/4` LEDs. The sibling script [`02_network`](https://github.com/openwrt/openwrt/blob/afd4aeb1f688f8e3eca6e89d77e7e7c5c9e78f50/target/linux/airoha/an7581/base-files/etc/board.d/02_network#L23-L26) already lists both compatibles in one case arm; matching that here keeps the two variants in sync.

```suggestion
nokia,xg-040g-md |\
nokia,xg-040g-md-ubi)
```

---

openwrt-ai (commented):
Reviewed 1 new commit. Verified the LED class-device names against the DTS and kernel 6.18 sources: `gsw_phy2/3/4` sit at MDIO addresses `0x0a`/`0x0b`/`0x0c` on the switch bus, whose id is `mt7530-0` (`snprintf(bus->id, ..., KBUILD_MODNAME "-%d", idx++)` in `drivers/net/dsa/mt7530.c`), and `led_compose_name()` yields `<devname>:<color>:<function>-<enumerator>` — so `mt7530-0:0a:green:lan-2` etc. are correct, and they match the entries already in tree for the an7583 `nokia,xg-040g-mf` sibling. One inline note about the UBI variant.

---

peterwillcn (approved):

target/linux/airoha/an7581/base-files/etc/board.d/01_leds
csharper2005: Sure

robimarko:

PR #24674, opened 2026-08-11
realtek: board: restore masking of unused R4K timer on CPU1

luizluca:
Commit 7cc31af7bdd4 ("realtek: convert to generic machine initialization") dropped the legacy Realtek setup code (prom.c/setup.c). During this transition, a critical SMP quirk was lost: masking the MIPS CP0 timer interrupt (IRQ 7) on secondary VPEs when the kernel doesn't use the native CEVT_R4K timer.

Realtek SoCs use their own timer block (timer@3100) instead of the MIPS internal timer. Without masking IP7 in the CP0 Status register, the secondary CPU comes up with the timer interrupt active by default. Since no handler is registered for IRQ 7, this immediately results in an IRQ storm and a "irq 7: nobody cared" kernel panic.

Restore this behavior by hooking into the smp_finish operation of the VSMP ops, similar to what the legacy rtlsmp_finish() did. We use an early_initcall to ensure our operations override the generic MIPS setup, safeguarded by a board check. This quirk is only applied if the kernel is built without CONFIG_CEVT_R4K.

Fixes: 7cc31af7bdd4 ("realtek: convert to generic machine initialization")
Assisted-by: Gemini:gemini-3.1-pro

### DIFF
--- a/target/linux/realtek/patches-6.18/300-04-realtek-board-fix-smp-irq7-storm.patch
+++ b/target/linux/realtek/patches-6.18/300-04-realtek-board-fix-smp-irq7-storm.patch
@@ -1,0 +1,99 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Luiz Angelo Daros de Luca <luizluca@gmail.com>
+Date: Tue, 11 Aug 2026 14:36:22 -0300
+Subject: [PATCH] realtek: board: restore masking of unused R4K timer on CPU1
+
+The transition to generic machine initialization introduced by
+patch 300-02-enhance-realtek-board-setup.patch dropped the legacy
+Realtek setup code (prom.c/setup.c). During this transition, a critical
+SMP quirk was lost: masking the MIPS CP0 timer interrupt (IRQ 7)
+on secondary VPEs when the kernel doesn't use the native CEVT_R4K timer.
+
+Realtek SoCs use their own timer block (timer@3100) instead of the MIPS
+internal timer. Without masking IP7 in the CP0 Status register, the
+secondary CPU comes up with the timer interrupt active by default. Since
+no handler is registered for IRQ 7, this immediately results in an IRQ
+storm and a "irq 7: nobody cared" kernel panic.
+
+Restore this behavior by hooking into the smp_finish operation of the
+currently registered SMP ops (which covers both VSMP and CPS setups),
+similar to what the legacy rtlsmp_finish() did. We use an early_initcall
+to ensure our operations override the generic MIPS setup, safeguarded by
+a board check. This quirk is only applied if the kernel is built without
+CONFIG_CEVT_R4K.
+
+Note: As a side effect of using register_smp_ops() in an initcall, the
+kernel will print "Overriding previously set SMP ops" during boot. This
+is expected and not a bug.
+
+Assisted-by: Gemini:gemini-3.1-pro
+Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>
+
+--- a/arch/mips/generic/board-realtek.c
++++ b/arch/mips/generic/board-realtek.c
+@@ -10,6 +10,8 @@
+ 
+ #include <asm/fw/fw.h>
+ #include <asm/machine.h>
++#include <asm/mipsregs.h>
++#include <asm/smp-ops.h>
+ 
+ #include <mach-rtl-otto.h>
+ 
+@@ -280,6 +282,47 @@ static void rtl930x_prepare_highmem(void
+ 	__sync();
+ }
+ 
++static bool is_realtek_board;
++
++#ifdef CONFIG_SMP
++extern const struct plat_smp_ops *mp_ops;
++static const struct plat_smp_ops *orig_smp_ops;
++static struct plat_smp_ops realtek_smp_ops;
++
++static void realtek_smp_finish(void)
++{
++	/*
++	 * Realtek SoCs do not use the native MIPS R4K timer. The secondary VPE
++	 * comes up with IP7 (timer interrupt) active. Mask it to prevent
++	 * an unhandled IRQ storm.
++	 */
++	write_c0_status(read_c0_status() & ~STATUSF_IP7);
++
++	/* Call upstream smp_finish if it exists */
++	if (orig_smp_ops && orig_smp_ops->smp_finish)
++		orig_smp_ops->smp_finish();
++	else
++		local_irq_enable();
++}
++
++static int __init realtek_smp_quirk(void)
++{
++	if (IS_ENABLED(CONFIG_CEVT_R4K))
++		return 0;
++	if (!is_realtek_board || !mp_ops)
++		return 0;
++
++	orig_smp_ops = mp_ops;
++	realtek_smp_ops = *mp_ops;
++	realtek_smp_ops.smp_finish = realtek_smp_finish;
++	pr_info("realtek: masking IP7 on CPU1 (the following 'Overriding previously set SMP ops' message is expected and can be ignored)\n");
++	register_smp_ops(&realtek_smp_ops);
++
++	return 0;
++}
++early_initcall(realtek_smp_quirk);
++#endif
++
+ static void __init realtek_prom_init(const void *match_data)
+ {
+ 	const struct realtek_soc_data *cfg = match_data;
+@@ -307,6 +350,8 @@ static void __init realtek_prom_init(con
+ 
+ 	pr_info("%s SoC with %d MB\n", get_system_type(), soc_info.memory_size >> 20);
+ 
++	is_realtek_board = true;
++
+ 	/*
+ 	 * fw_arg2 is the pointer to the environment. Some devices (e.g. HP JG924A) hand
+ 	 * over other than expected kernel boot arguments. Something like 0xfffdffff looks


luizluca:
This PR fixes this IRQ loop error while booting realtek SMP switches:

```
[   37.040538] Disabling IRQ #7
[   37.153540] irq 7: nobody cared (try booting with the "irqpoll" option)
[   37.160888] CPU: 1 UID: 0 PID: 0 Comm: swapper/1 Not tainted 6.18.41 #0 NONE 
[   37.160918] Hardware name: Edgecore ECS2100-52T
[   37.160927] Stack : 81813cf4 80119fe8 80aa8040 00000001 00000000 00000000 00000000 00000000
[   37.160976]         00000000 00000000 00000000 00000000 00000000 00000001 81813cb0 818792c0
[   37.161016]         00000000 00000000 809bc258 81813b48 00000000 ffffefff 00000001 00000dce
[   37.161058]         00000dd0 81813b0c 00000dd0 0000a5c0 00000000 00000000 809bc258 0000001f
[   37.161100]         812e0000 812e0000 00000007 80119fe8 00000000 80ac5470 00000004 812b0004
[   37.161143]         ...
[   37.161152] Call Trace:
[   37.161157] [<80112128>] show_stack+0x28/0xf0
[   37.161187] [<8010be44>] dump_stack_lvl+0x70/0xb0
[   37.161219] [<801020f4>] __report_bad_irq+0x5c/0x11c
[   37.161236] [<801ae310>] note_interrupt+0x348/0x398
[   37.161260] [<801aa730>] handle_irq_event_percpu+0x54/0x6c
[   37.161277] [<801b0730>] handle_percpu_irq+0x84/0xb4
[   37.161302] [<801a9c30>] handle_irq_desc+0x58/0x94
[   37.161318] [<808da484>] do_domain_IRQ+0x28/0x3c
[   37.161345] [<804d143c>] plat_irq_dispatch+0x90/0xdc
[   37.161368] [<8010d7c8>] except_vec_vi_end+0xc4/0xd0
```

target/linux/realtek/patches-6.18/300-04-realtek-board-fix-smp-irq7-storm.patch
```diff
@@ -0,0 +1,93 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Luiz Angelo Daros de Luca <luizluca@gmail.com>
+Date: Tue, 11 Aug 2026 14:36:22 -0300
+Subject: [PATCH] realtek: board: restore masking of unused R4K timer on CPU1
+
+The transition to generic machine initialization introduced by
+patch 300-02-enhance-realtek-board-setup.patch dropped the legacy
+Realtek setup code (prom.c/setup.c). During this transition, a critical
+SMP quirk was lost: masking the MIPS CP0 timer interrupt (IRQ 7)
+on secondary VPEs when the kernel doesn't use the native CEVT_R4K timer.
```
jonasjelonek, line 10: I'd drop this from the patch message. Once this is getting upstreamed, this OpenWrt-only history needs to be removed anyway.

luizluca: This is a fixup patch. It would probably get squashed without a commit message before being upstreamed. If it is OK, I would keep it to explain why it is in the 300- series.

luizluca:
I was avoiding #ifdefs but the plat_smp_ops is not defined if the SMP is not enabled.

target/linux/realtek/patches-6.18/300-04-realtek-board-fix-smp-irq7-storm.patch
```diff
@@ -0,0 +1,94 @@
...
++static struct plat_smp_ops realtek_smp_ops;
++
++static void realtek_smp_finish(void)
++{
++	/* 
++	 * Realtek SoCs do not use the native MIPS R4K timer. The secondary VPE
++	 * comes up with IP7 (timer interrupt) active. Mask it to prevent
++	 * an unhandled IRQ storm.
++	 */
++	write_c0_status(read_c0_status() & ~CAUSEF_IP7);
```
openwrt-ai: `CAUSEF_IP7` is the Cause-register bit name, but this writes the Status register, whose interrupt-mask bit is `STATUSF_IP7`. Both are bit 15, so the generated code is identical — but the naming is misleading, and every in-tree Status-register user spells it `STATUSF_IP7` (e.g. [`vsmp_init_secondary()` at smp-mt.c:115-118](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/mips/kernel/smp-mt.c#L115-L118), [smp-cps.c:713](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/mips/kernel/smp-cps.c#L713)). Since this is headed upstream, better to use the right constant.

```suggestion
+	write_c0_status(read_c0_status() & ~STATUSF_IP7);
```

---

target/linux/realtek/patches-6.18/300-04-realtek-board-fix-smp-irq7-storm.patch
```diff
@@ -0,0 +1,94 @@
...
+ 
++static bool is_realtek_board;
++
++#ifdef CONFIG_MIPS_MT_SMP
++extern const struct plat_smp_ops vsmp_smp_ops;
++static struct plat_smp_ops realtek_smp_ops;
++
++static void realtek_smp_finish(void)
++{
++	/*
```
openwrt-ai: nit: trailing whitespace after `/*` on this added line (`checkpatch: trailing whitespace`). It is the only such line in the patch.

```suggestion
+	/*
```

---

target/linux/realtek/patches-6.18/300-04-realtek-board-fix-smp-irq7-storm.patch
```diff
@@ -0,0 +1,94 @@
...
++	if (IS_ENABLED(CONFIG_CEVT_R4K))
++		return 0;
++	if (!is_realtek_board)
++		return 0;
++	if (!cpu_has_mipsmt)
++		return 0;
++
++	realtek_smp_ops = vsmp_smp_ops;
++	realtek_smp_ops.smp_finish = realtek_smp_finish;
++	register_smp_ops(&realtek_smp_ops);
```
openwrt-ai: This unconditionally replaces whatever ops are already registered with a copy of `vsmp_smp_ops`, but VSMP is not necessarily what is registered on every Realtek subtarget.

`rtl931x` and `rtl931x_nand` build with [`CONFIG_MIPS_CPS=y`](https://github.com/openwrt/openwrt/blob/726fe64251009b7778e4cf3a50c7d619536ff053/target/linux/realtek/rtl931x/config-6.18#L143) *and* `CONFIG_MIPS_MT_SMP=y` (300-01 makes `RTL931X` select both `SYS_SUPPORTS_MIPS_CPS` and `MIPS_GIC`, and it inherits `SYS_SUPPORTS_MULTITHREADING` from `RTL930X`). [`device_tree_init()` at generic/init.c:111-120](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/mips/generic/init.c#L111-L120) tries `register_cps_smp_ops()` *before* `register_vsmp_smp_ops()`, so with a CM+GIC present `mp_ops` is `cps_smp_ops` there. Two outcomes, both wrong:

- `cpu_has_mipsmt` true (interAptiv has the MT ASE): this initcall swaps CPS ops out for VSMP ops — a full SMP-model downgrade, not just an `smp_finish` override.
- `cpu_has_mipsmt` false: the quirk is skipped and rtl931x keeps the storm, since [`cps_init_secondary()` at smp-cps.c:713](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/mips/kernel/smp-cps.c#L713) unmasks `STATUSF_IP7` just like the VSMP one does.

Deriving from the ops that are actually registered covers both, and works for CPS as well:

```c
extern const struct plat_smp_ops *mp_ops;	/* arch/mips/kernel/smp.c */
static const struct plat_smp_ops *orig_smp_ops;
...
	orig_smp_ops = mp_ops;
	realtek_smp_ops = *mp_ops;
	realtek_smp_ops.smp_finish = realtek_smp_finish;
	register_smp_ops(&realtek_smp_ops);
```

with `#ifdef CONFIG_SMP` in place of `CONFIG_MIPS_MT_SMP` (that also answers the `plat_smp_ops`-undefined problem you mentioned in the PR comment) and `orig_smp_ops->smp_finish()` in `realtek_smp_finish()`. `mp_ops` is `EXPORT_SYMBOL`'d but only declared inside the `plat_smp_setup()` inline in `smp-ops.h`, so a local `extern` is still needed. The `cpu_has_mipsmt` guard can then go away.

Has this been booted on an rtl931x board, or only on the rtl839x/rtl930x ones?

Minor side effect either way: `register_smp_ops()` prints `Overriding previously set SMP ops` at boot ([smp.c:164-170](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/mips/kernel/smp.c#L164-L170)) — worth a word in the patch description so it is not mistaken for a bug.

---

target/linux/realtek/patches-6.18/300-04-realtek-board-fix-smp-irq7-storm.patch
```diff
@@ -0,0 +1,94 @@
...
++	 * an unhandled IRQ storm.
++	 */
++	write_c0_status(read_c0_status() & ~CAUSEF_IP7);
++	write_c0_compare(read_c0_count() - 1);
++
++	/* Call upstream vsmp_finish if it exists */
++	if (vsmp_smp_ops.smp_finish)
++		vsmp_smp_ops.smp_finish();
++	else
++		local_irq_enable();
```
openwrt-ai: The `write_c0_compare()` here is immediately undone: [`vsmp_smp_finish()` at smp-mt.c:121-133](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/mips/kernel/smp-mt.c#L121-L133) opens with `write_c0_compare(read_c0_count() + (8 * mips_hpt_frequency / HZ))`. With `CSRC_R4K`/`CEVT_R4K` deselected by 300-01, `mips_hpt_frequency` is 0, so Compare ends up equal to Count and the R4K timer re-arms right away — harmless only because IP7 is masked by the line above, which means the ack here buys nothing. Either drop it as redundant (the Status mask is what actually fixes the storm) or move it after the `vsmp_smp_finish()` call if the intent is to leave the counter disarmed.

Also, `vsmp_smp_ops.smp_finish` is a static initializer ([smp-mt.c:236](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/arch/mips/kernel/smp-mt.c#L236)) and can never be NULL, so the `if`/`else` is decided at build time and the `local_irq_enable()` fallback is dead code — a plain unconditional call reads clearer.

---

openwrt-ai (commented):
Reviewed 1 commit (`726fe64`). The commit message matches the diff, and CI on this head is green, so nothing to raise there. Four inline notes: one likely functional problem on the rtl931x subtargets (CPS vs. VSMP ops), a wrong-register constant name, a redundant/undone `write_c0_compare()`, and a trailing-whitespace nit. Compared against the `v6.18.41` kernel tree, matching `KERNEL_PATCHVER:=6.18`.

---

target/linux/realtek/patches-6.18/300-04-realtek-board-fix-smp-irq7-storm.patch
jameywine: @jonasjelonek could you check this on rtl931x? seems like a regression to me and it might affect rtl960x stuff as well.

target/linux/realtek/patches-6.18/300-04-realtek-board-fix-smp-irq7-storm.patch
jonasjelonek: Then why is this a separate patch at all and not directly editing the original patch file?

plappermaul:
With all the refactoring of the last months nowadays I tend to just revert the CEVT_R4K removal from Patch 300-01.

target/linux/realtek/patches-6.18/300-04-realtek-board-fix-smp-irq7-storm.patch
luizluca: OK

luizluca:
I just updated to deal with Claude review points. It still works in my RTL839X.

target/linux/realtek/patches-6.18/300-04-realtek-board-fix-smp-irq7-storm.patch
luizluca: I was just mimicking the 300-03 patch, that is also a fixup for 300-02.

I'll keep the current form during this review to avoid extra noise and I can apply it over 300-02 if we decide to merge it.

luizluca:
> With all the refactoring of the last months nowadays I tend to just revert the CEVT_R4K removal from Patch 300-01.

It would work around the issue. I do not like hijacking the smp_ops either, but I didn't find any other solution other than modifying the MIPS generic code to add a callback for that situation.

Do you all agree with re-enabling the CEVT_R4K timer? If so, we can take that route and we can just close this PR.

plappermaul:
yes please and check that IRQ7 Interrupt is not firing like crazy afterwards.

luizluca:
> yes please and check that IRQ7 Interrupt is not firing like crazy afterwards.

This one-line patch seems to be enough:

```diff
diff --git a/target/linux/realtek/patches-6.18/300-01-enhance-realtek-platform-config.patch b/target/linux/realtek/patches-6.18/300-01-enhance-realtek-platform-config.patch
index 70d4f7990e..0b6e48935e 100644
--- a/target/linux/realtek/patches-6.18/300-01-enhance-realtek-platform-config.patch
+++ b/target/linux/realtek/patches-6.18/300-01-enhance-realtek-platform-config.patch
@@ -11,12 +11,12 @@ Submitted-by: Birger Koblitz <git@birger-koblitz.de>
 ---
 --- a/arch/mips/Kconfig
 +++ b/arch/mips/Kconfig
-@@ -700,19 +700,22 @@ config MACH_REALTEK_RTL
+@@ -700,19 +700,23 @@ config MACH_REALTEK_RTL
        select MACH_GENERIC_CORE
        select DMA_NONCOHERENT
        select IRQ_MIPS_CPU
 -      select CSRC_R4K
--      select CEVT_R4K
+       select CEVT_R4K
        select SYS_HAS_CPU_MIPS32_R1
        select SYS_HAS_CPU_MIPS32_R2
        select SYS_SUPPORTS_BIG_ENDIAN
```

And timer is very quiet:

```
root@OpenWrt:~# uptime
 00:05:38 up 5 min,  load average: 0.62, 0.38, 0.19
root@OpenWrt:~# cat /proc/interrupts 
           CPU0       CPU1       
  7:         27         28     MIPS   7  timer
  8:      10250      11193     MIPS   0  IPI call
  9:        560        767     MIPS   1  IPI resched
 12:        664          0 realtek-otto-gpio  12  sfp-p49-los
 14:          0          0 realtek-otto-gpio  14  sfp-p50-los
 19:          0          0 realtek-rtl-intc  19  realtek-otto-wdt
 20:          4          0 realtek-rtl-intc  20  rtldsa-link-state
 21:          0          0 realtek-otto-gpio  21  sfp-p51-los
 22:          0          0 realtek-otto-gpio  22  sfp-p52-los
 24:         56          0 realtek-rtl-intc  24  eth0
 28:          0     230402 realtek-rtl-intc  28  timer@3100
 29:     319127          0 realtek-rtl-intc  29  timer@3100
 31:      86919          0 realtek-rtl-intc  31  ttyS0
ERR:         52
```

It might only be triggered when the counter overflows and reach again the compare. I'll create a new PR that closes this one.

target/linux/realtek/patches-6.18/300-04-realtek-board-fix-smp-irq7-storm.patch
```diff
@@ -0,0 +1,99 @@
...
++static struct plat_smp_ops realtek_smp_ops;
++
++static void realtek_smp_finish(void)
++{
++	/*
++	 * Realtek SoCs do not use the native MIPS R4K timer. The secondary VPE
++	 * comes up with IP7 (timer interrupt) active. Mask it to prevent
++	 * an unhandled IRQ storm.
++	 */
++	write_c0_status(read_c0_status() & ~STATUSF_IP7);
```
openwrt-ai, line 61: MIPS already has a helper for exactly this: `clear_c0_status()`, generated by [`__BUILD_SET_C0(status)` at mipsregs.h:3056](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/include/asm/mipsregs.h#L3056). It expands to the same read/modify/write but compiles to a single `ins` on MIPS32r2, and it is what in-tree Status-IM masking uses (e.g. [`clear_c0_status(STATUSF_IP4)` at octeon-irq.c:1363](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/cavium-octeon/octeon-irq.c#L1363)). Worth matching since this is headed upstream.

```suggestion
+	clear_c0_status(STATUSF_IP7);
```

---

target/linux/realtek/patches-6.18/300-04-realtek-board-fix-smp-irq7-storm.patch
```diff
@@ -0,0 +1,99 @@
...
++
++	/* Call upstream smp_finish if it exists */
++	if (orig_smp_ops && orig_smp_ops->smp_finish)
```
openwrt-ai, line 64: nit: `orig_smp_ops` can never be NULL here — `realtek_smp_quirk()` bails out on `!mp_ops` and only then does `orig_smp_ops = mp_ops`, so the first half of the condition is always true. The `->smp_finish` half is the only one that can matter, and even that is set unconditionally by both candidates ([`.smp_finish = vsmp_smp_finish` at smp-mt.c:236](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/kernel/smp-mt.c#L236), [`.smp_finish = cps_smp_finish` at smp-cps.c:945](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/kernel/smp-cps.c#L945)), so the `local_irq_enable()` fallback stays unreachable in practice — keeping it as a safety net is fine, but the NULL-pointer half of the test just adds noise.

```suggestion
+	if (orig_smp_ops->smp_finish)
```

---

target/linux/realtek/patches-6.18/300-04-realtek-board-fix-smp-irq7-storm.patch
```diff
@@ -0,0 +1,99 @@
...
++{
++	if (IS_ENABLED(CONFIG_CEVT_R4K))
++		return 0;
++	if (!is_realtek_board || !mp_ops)
++		return 0;
++
++	orig_smp_ops = mp_ops;
++	realtek_smp_ops = *mp_ops;
++	realtek_smp_ops.smp_finish = realtek_smp_finish;
++	pr_info("realtek: masking IP7 on CPU1 (the following 'Overriding previously set SMP ops' message is expected and can be ignored)\n");
```
openwrt-ai, line 80: nit: two small mismatches with the surrounding code. The hook is installed for *every* secondary VPE (and runs again on each CPU hotplug-online), not just CPU1, so "on CPU1" under-describes what it does. And `board-realtek.c` has no `pr_fmt` define and its other messages print without a manual prefix ([`pr_info("Adding initrd info from environment\n")` at board-realtek.c:31](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/generic/board-realtek.c#L31)), so the hand-written `realtek: ` prefix here is inconsistent.

```suggestion
+	pr_info("masking unused R4K timer IP7 on secondary CPUs (the following 'Overriding previously set SMP ops' message is expected)\n");
```

---

openwrt-ai (commented):
Re-reviewed the delta since `726fe64`. The three functional points from the last round are addressed: `STATUSF_IP7` instead of `CAUSEF_IP7`, the redundant `write_c0_compare()` is gone, and deriving the ops from `mp_ops` under `#ifdef CONFIG_SMP` now covers the rtl931x CPS case as well as VSMP. I re-checked the init ordering against v6.18.44 and it holds: `smp_prepare_cpus()` runs before `do_pre_smp_initcalls()`, so `prepare_cpus` has already run against the original ops when the copy is taken, and `smp_init()` runs after, so `mp_ops->smp_finish` is the replacement by the time any secondary reaches it. Hunk offsets in the patch are consistent. CI on this head is still queued, so nothing to report there.

Three minor inline notes below, none blocking.

## Commit checks

- `ba5bf4f` "realtek: board: restore masking of unused R4K timer on CPU1" — the git commit message still describes the *previous* implementation: "hooking into the `smp_finish` operation of the VSMP ops". The diff no longer does that, and the patch-file header inside `300-04-...patch` was updated to "currently registered SMP ops (which covers both VSMP and CPS setups)" plus the new "Overriding previously set SMP ops" note. The two texts are otherwise verbatim copies of each other, so the git message looks like it just missed the last rewrite.

Compared against the `v6.18.44` kernel tree, matching `KERNEL_PATCHVER:=6.18`.

---

PR #24677, opened 2026-08-11
mac80211: add RTL8192SU USB support

rdmitry0911:
# mac80211: add RTL8192SU USB support

This series restores mac80211 support for RTL8192SU, RTL8191SU and
RTL8188SU USB adapters and adds the AP firmware required by the driver.

The driver is based on the former rtlwifi RTL8192SU implementation and has
been ported to the current mac80211 backports API. The OpenWrt package glue
builds the shared RTL8192S code and the RTL8192SU USB module as separate
kmods, selects their rtlwifi dependencies and installs the AP-specific
firmware as `rtlwifi/rtl8192sufw-ap.bin`.

AP operation also needs three RTL8192SU-specific fixes:

- a dedicated USB command URB path for firmware download before the normal
  USB data path is started, with command completion kept out of mac80211 TX
  status handling;
- beacon and buffered broadcast delivery through the USB transport;
- using the robust 6 Mbps OFDM rate for AP downlink data, because HT
  downlink selected by mac80211 stalls this firmware.

The rate workaround is deliberately limited to RTL8192SU in AP mode;
station mode and other rtlwifi devices retain their normal rate control.
Station power-save state is left to mac80211, so TIM and buffered broadcast
handling remain reachable.

Source: https://github.com/chunkeey/rtl8192su at
`1f09c7a6eceffc92e63ddc636c174e8f3770eeb9`.

Tested on x86/64:

- OpenWrt main at `0256e6a1e470`, Linux 6.18.41, backports 6.18.39:
  `make -j8 package/kernel/mac80211/clean` followed by
  `make -j8 package/kernel/mac80211/compile`, with both `kmod-rtl8192su`
  and `kmod-rtl8192cu` selected;
- `rtl8192su-firmware`, `kmod-rtl8192s-common` and `kmod-rtl8192su` APKs
  were generated successfully;
- the packaged AP firmware SHA-256 is
  `db101f23116464059e5f8f9bd5412a2c7133af4479fb9e8318afe1c3fdbbfafb`;
- clean runtime test on the same snapshot and Linux 6.18.41 with a Realtek
  `0bda:8172` AP and MediaTek `0e8d:7610` client: WPA2 AP on channel 6,
  successful authentication and association, DHCP lease and DNS through
  the AP, 100/100 ICMP packets to the AP and 100/100 through routed NAT to
  the Internet, with no firmware-download or USB-command errors and no TX
  retries or failures at the AP;
- the MediaTek client rejects enabling power save with `-EOPNOTSUPP`, so
  the sleeping-station DTIM/buffered-broadcast case was not exercised by
  this runtime test;
- OpenWrt `checkpatch.pl` on the backports patch: 0 errors, 0 warnings
  (516 informational checks inherited mostly from the legacy driver).

This is intended as an RFC because the driver is a substantial downstream
backports addition and maintainers may prefer it to be accepted in Linux or
the backports project first.


package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
```diff
@@ -0,0 +1,12012 @@
...
++	switch (rtlefuse->eeprom_regulatory) {
++	case 3:
++		/* The following is for calculation
++		 * of the power diff for Ant-B to Ant-A.
++		 */
++		if (rtlphy->current_chan_bw == HT_CHANNEL_WIDTH_20_40) {
++			p_final_pwridx[0] += rtlefuse->pwrgroup_ht40
++						[RF90_PATH_A][
++						chnl - 1];
++			p_final_pwridx[1] +=
```
openwrt-ai: This breaks the build for every target that builds rtlwifi-usb:

```
drivers/net/wireless/realtek/rtlwifi/rtl8192cu/sw.c:148:23: error: initialization of
'int (*)(struct ieee80211_hw *, struct sk_buff *)' from incompatible pointer type
'void (*)(struct ieee80211_hw *, struct sk_buff *)' [-Wincompatible-pointer-types]
  148 |         .usb_rx_hdl = rtl8192cu_rx_hdl,
      |                       ^~~~~~~~~~~~~~~~
make[13]: *** [scripts/Makefile.build:287: .../rtl8192cu/sw.o] Error 1
```

(`Build x86/64` and `Build malta/be` both fail with exactly this error; the PR description reports a clean `package/kernel/mac80211/compile` on the same base, so that result is probably from a tree where rtl8192cu was deselected.)

The hunks at the end of this patch retype `usb_rx_hdl` from `void (*)` to `int (*)` in both `usb.h` and `wifi.h`, but [`rtl8192cu_rx_hdl` in rtl8192cu/trx.c:423](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/realtek/rtlwifi/rtl8192cu/trx.c#L423) and its prototype in `rtl8192cu/trx.h:381` are not converted. The patch needs to update the only other implementer as well.

Beyond the compile error, this new dispatch is also a behaviour regression for rtl8192cu. Upstream `usb_rx_hdl` is stored in `rtl_usb` but never invoked — this line is the first call site. [`rtl8192cu_rx_hdl`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/realtek/rtlwifi/rtl8192cu/trx.c#L423-L426) calls `_rtl_rx_process()`, which ends in [`ieee80211_rx(hw, skb)`](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/realtek/rtlwifi/rtl8192cu/trx.c#L420) and hands the skb to mac80211. Once it is made to return `int`, whatever it returns for the normal case will be `0`, so the code falls through to `_rtl_usb_rx_process_noagg(hw, skb)` on an skb that is already owned by mac80211 — a use-after-free on every rtl8192cu RX frame.

Please scope the new hook to the driver that needs it, e.g. add a separate callback for the RTL8192SU event path rather than repurposing `usb_rx_hdl`, or set `.usb_rx_hdl = NULL` in rtl8192cu as part of this patch.

---

openwrt-ai: This power-save workaround is dead code for RTL8192SU. It is added to `_rtl_usb_rx_process_agg()`, which is only reached from the segregate path — [usb.c:466-475](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/realtek/rtlwifi/usb.c#L466-L475) calls it only when `rtlusb->usb_rx_segregate_hdl` is non-NULL. This patch sets `.usb_rx_segregate_hdl = NULL` for rtl8192su (line 11014 of this file), so RX always goes through `_rtl_usb_rx_process_noagg()` instead.

On top of that, `_rtl_usb_rx_process_agg()` never delivers the skb to mac80211 at all (no `ieee80211_rx*()` call), so clearing `IEEE80211_FCTL_PM` there could not influence mac80211's PS bookkeeping even if the path were taken.

The hunk belongs in `_rtl_usb_rx_process_noagg()`, before the frame is passed up.

---

openwrt-ai, line 11013: `rtlpriv->rtlhal.pfirmware` is leaked on this error path. `rtl_usb_probe()` jumps to `error_out` when `init_sw_vars()` fails, and that label sits *after* the `deinit_sw_vars()` call, so `rtl92su_deinit_sw_vars()` never runs and the `vzalloc(sizeof(struct rt_firmware))` above is never freed.

The equivalent upstream code in [rtl8192se/sw.c:191-196](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/realtek/rtlwifi/rtl8192se/sw.c#L191-L196) frees it explicitly before returning. Please add the `vfree()` + `NULL` assignment here too (a patch refresh via `make package/mac80211/refresh` will fix up the hunk counts).

---

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
```diff
@@ -0,0 +1,12012 @@
...
+ EXPORT_SYMBOL(rtl_rx_ampdu_apply);
+ 
++void rtl_send_buffered_bc(struct ieee80211_hw *hw)
++{
++	struct rtl_priv *rtlpriv = rtl_priv(hw);
++	struct rtl_mac *mac = rtl_mac(rtlpriv);
++	struct rtl_tcb_desc tcb_desc;
++	struct sk_buff *skb;
++
++	while ((skb = ieee80211_get_buffered_bc(hw, mac->vif))) {
```
openwrt-ai: `mac->vif` is not checked for NULL before being handed to `ieee80211_get_buffered_bc()`, which dereferences it unconditionally via `vif_to_sdata()`.

This runs from `rtl92su_rx_hdl()` in the RX tasklet, driven by a firmware event (`0x15`). Nothing guarantees a vif is still attached at that point — a firmware event that arrives during/after `remove_interface()` will oops. Please bail out early when `mac->vif` is NULL (and ideally when `mac->opmode` is not AP/mesh/adhoc, since buffered BC only makes sense there).

---

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
openwrt-ai, line 11830: Is `offs.tim_offset` still correct here? `ieee80211_beacon_get_template()` returns it relative to the start of the 802.11 beacon frame, but the `fill_tx_desc` call on the previous line runs `rtl92su_tx_fill_desc()`, which does `skb_push(skb, RTL_TX_HEADER_SIZE)` (see line 11535 of this file). The buffer handed to the firmware on the next line is `skb->data`, i.e. it now starts `RTL_TX_HEADER_SIZE` bytes earlier than the frame `tim_offset` was measured against.

If the firmware expects the offset relative to the buffer it receives, this needs `offs.tim_offset + RTL_TX_HEADER_SIZE`; if it expects it relative to the 802.11 header, the current value is right. Worth a comment either way, since the two are silently off by a constant.

---

openwrt-ai: nit: `int rtl92su_update_beacon(struct ieee80211_hw *hw);` is already declared 16 lines above in this same header (line 11851 of this file). Drop this duplicate and regenerate the patch with `make package/mac80211/refresh` so the hunk line counts stay consistent.

Same file also declares `rtl92su_rx_segregate_hdl()`, which has no definition anywhere in the series (`sw.c` sets `.usb_rx_segregate_hdl = NULL` with the name commented out) — that declaration can go too.

---

openwrt-ai: nit: `-D__CHECK_ENDIAN__` has been a no-op since sparse endianness checking became the default (kernel 4.20); no rtlwifi sub-Makefile carries it any more — compare [rtl8192du/Makefile](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/realtek/rtlwifi/rtl8192du/Makefile). The same line is also added to the new `rtl8192s/Makefile` (line 146 of this file).

Both new Makefiles are also missing the `# SPDX-License-Identifier: GPL-2.0` first line that every sibling rtlwifi Makefile has.

---

openwrt-ai: nit: `rtlpriv->cfg->mod_params->sw_crypto = rtlpriv->cfg->mod_params->sw_crypto;` is a self-assignment with no effect. Neither [rtl8192se/sw.c](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/wireless/realtek/rtlwifi/rtl8192se/sw.c#L150-L178) nor any other in-tree rtlwifi `init_sw_vars()` has it. Please drop the statement.

---

package/firmware/rtl8192su-firmware/Makefile
```diff
@@ -0,0 +1,39 @@
...
+
+define Build/Compile
+	true
+endef
+
+define Package/rtl8192su-firmware/install
+	$(INSTALL_DIR) $(1)/lib/firmware/RTL8192SU
+	$(INSTALL_DATA) \
+		$(PKG_BUILD_DIR)/firmwares/rtl8192sufw-ap.bin \
+		$(1)/lib/firmware/RTL8192SU/rtl8192sfw.bin
```
openwrt-ai: nit: the install path deviates from the rtlwifi firmware convention. Every rtlwifi firmware in linux-firmware lives under `/lib/firmware/rtlwifi/` — the sibling package in this tree installs [`rtlwifi/rtl8192sefw.bin`](https://github.com/openwrt/openwrt/blob/7eff35b578431cf4ccb8b8ac2ea1c1c837d91a10/package/firmware/linux-firmware/realtek.mk#L84), and upstream's `rtl8192se` requests `"rtlwifi/rtl8192sefw.bin"`. This package creates a new uppercase `RTL8192SU/` directory, matching `RTL8192SU_FW_NAME` in the driver patch (line 10889 of `990-rtlwifi-add-rtl8192su-support.patch`).

Also, since this is an AP-specific blob and not the stock `rtl8192sfw.bin`, the generic basename is misleading — something like `rtlwifi/rtl8192sufw-ap.bin` would say what it is. Both the `MODULE_FIRMWARE()`/`RTL8192SU_FW_NAME` in the driver patch and this install rule need to change together.

---

package/firmware/rtl8192su-firmware/Makefile
```diff
@@ -0,0 +1,39 @@
...
+PKG_SOURCE_DATE:=2017-03-31
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_URL:=https://github.com/chunkeey/rtl8192su.git
+PKG_SOURCE_VERSION:=1f09c7a6eceffc92e63ddc636c174e8f3770eeb9
+PKG_MIRROR_HASH:=dca07ce948db9b51bc327d424a6a0eee8fd19b0208aeb1822ee8edc707093aa1
+
+PKG_MAINTAINER:=Dmitry R <rdmitry0911@gmail.com>
+PKG_LICENSE:=Firmware
```
openwrt-ai: nit: `Firmware` is not an SPDX identifier and is not used by any other package in the tree — every other `package/firmware/*/Makefile` either sets a real identifier (`GPL-2.0-only`, `ISC`, `LA_OPT_NXP_Software_License`) or omits `PKG_LICENSE` and only sets `PKG_LICENSE_FILES`. Since the redistribution terms here come from `LICENCE.rtlwifi_firmware.txt`, dropping this line and keeping only `PKG_LICENSE_FILES` matches what `cypress-firmware` and `ath11k-firmware` do.

```suggestion
```

---

openwrt-ai (commented):
Reviewed 2 new commits; findings posted inline.

---

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
rdmitry0911: Fixed: the RTL8192SU AP PM-bit workaround now runs in _rtl_usb_rx_process_noagg(), before delivery to mac80211.

rdmitry0911: Fixed by keeping the existing void usb_rx_hdl callback untouched and adding a separate bool usb_rx_event_hdl used only by RTL8192SU firmware events. A clean build with both kmod-rtl8192cu and kmod-rtl8192su selected now succeeds.

rdmitry0911: Fixed: the request_firmware_nowait() error path now vfree()s pfirmware and clears the pointer.

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
rdmitry0911: Fixed: rtl_send_buffered_bc() snapshots/checks the vif and only proceeds for AP, adhoc, or mesh modes.

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
rdmitry0911: Fixed: both new Makefiles now carry SPDX-License-Identifier: GPL-2.0 and the obsolete __CHECK_ENDIAN__ flags are gone.

package/firmware/rtl8192su-firmware/Makefile
rdmitry0911: Fixed: the package now installs rtlwifi/rtl8192sufw-ap.bin and RTL8192SU_FW_NAME/MODULE_FIRMWARE request that same AP-specific path.

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
rdmitry0911: Verified against the archived Realtek vendor driver: timoffset is measured from the start of beaconbuf (the 802.11 frame), while the TX descriptor is prepended separately. The value remains offs.tim_offset; added a comment documenting this.

rdmitry0911: Fixed: removed the no-op self-assignment.

rdmitry0911: Fixed: removed the duplicate rtl92su_update_beacon declaration and the undefined rtl92su_rx_segregate_hdl declaration.

package/firmware/rtl8192su-firmware/Makefile
rdmitry0911: The non-SPDX Firmware value is gone. OpenWrt's current FormalityCheck requires PKG_LICENSE for every new package, so the package now uses the valid custom SPDX identifier LicenseRef-Realtek-Firmware and retains PKG_LICENSE_FILES:=firmwares/LICENCE.rtlwifi_firmware.txt.

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
```diff
@@ -0,0 +1,12027 @@
...
++	/* The following is for Antenna
++	 * diff from Ant-B to Ant-A
++	 */
++	p_final_pwridx[0] = pwrlevel[0];
++	p_final_pwridx[1] = pwrlevel[1];
++
++	switch (rtlefuse->eeprom_regulatory) {
++	case 3:
++		/* The following is for calculation
++		 * of the
```
openwrt-ai: Moving this hunk into `_rtl_usb_rx_process_noagg()` makes it live — and now that it actually runs, it makes the buffered-broadcast path added by this same series unreachable.

rtlwifi does not set `IEEE80211_HW_AP_LINK_PS`, so mac80211 derives STA power-save state purely from the PM bit: [`ieee80211_rx_h_sta_process()` at rx.c:1833-1847](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/net/mac80211/rx.c#L1833-L1847) only calls `sta_ps_start()` when `ieee80211_has_pm(hdr->frame_control)` is true. Clearing PM on *every* data/mgmt frame in AP mode means `sta_ps_start()` is never reached and `ps->num_sta_ps` stays 0 forever. Consequences:

- [`ieee80211_tx_h_multicast_ps_buf()` at tx.c:419-421](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/net/mac80211/tx.c#L419-L421) returns `TX_CONTINUE` when `num_sta_ps == 0` and `bc_buf` is empty, so `ps->bc_buf` is never filled. The `rtl_send_buffered_bc()` call driven by firmware event `0x15` ([line 11782 of this file](``https://github.com/openwrt/openwrt/blob/71e7144e7a09e6f5de02162cb2c136260eb3aca7/package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch#L11782))`` will always find an empty queue.
- [`__ieee80211_beacon_add_tim()` at tx.c:5024-5030](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/net/mac80211/tx.c#L5024-L5030) only builds a TIM bitmap when `num_sta_ps > 0`, so the beacon TIM pushed by `rtl92su_update_beacon()` never carries AID bits either.

So the two workarounds in this series cancel each other out: with PM always cleared, no client is ever tracked as sleeping, downlink for genuinely sleeping clients is transmitted while they are off-channel, and the beacon/buffered-BC machinery added here is dead code. The PR description presents the two as independent fixes; it does not mention that the PS workaround disables the BC path.

Could the workaround be narrowed so PS tracking keeps working — e.g. clear PM only when the firmware's PS state is known to be stale (a per-STA inactivity heuristic), rather than unconditionally for all AP-mode frames? Or, if the buffered-BC/TIM path is known not to work on this firmware anyway, it would be worth saying so explicitly so a future reader does not try to fix a path that is intentionally unreachable.

(Compared against Linux 6.18.41; the package builds backports 6.18.39, these mac80211 paths are unchanged between the two.)

---

openwrt-ai (commented):
Reviewed the changes since `7eff35b`. The fixes from the previous round all check out: the `usb_rx_hdl` retype is gone (replaced by a separate `usb_rx_event_hdl`, so rtl8192cu is untouched), the firmware leak on the `request_firmware_nowait()` error path is fixed, `rtl_send_buffered_bc()` now guards on vif/opmode, the duplicate and undefined declarations in `trx.h` are removed, both new Makefiles carry SPDX headers without `-D__CHECK_ENDIAN__`, and the firmware now installs and is requested as `rtlwifi/rtl8192sufw-ap.bin`. Hunk offsets and line counts in the regenerated patch are internally consistent, and the hunks against existing files (`base.c`, `base.h`, `core.c`, `usb.c`, `usb.h`, `wifi.h`) apply cleanly with zero fuzz against Linux 6.18.41.

One new finding posted inline about the PS workaround now being live.

---

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
rdmitry0911: Confirmed against the packaged backports 6.18.39 paths. Because rtlwifi does not advertise `IEEE80211_HW_AP_LINK_PS`, unconditionally clearing PM before mac80211 RX prevents `sta_ps_start()`, keeps `num_sta_ps` at zero, and makes the TIM/buffered-broadcast paths unreachable.

Removed the RTL8192SU PM-bit rewrite entirely rather than adding an unvalidated per-STA heuristic. The original runtime test had this code in the unreachable aggregate RX path, so it was not evidence that the workaround was needed. Power-save tracking now remains with mac80211 while the RTL8192SU event callback and buffered-broadcast delivery stay intact.

Clean x86/64 rebuild on base `0256e6a1e470` succeeded with `kmod-rtl8192su`, `kmod-rtl8192cu`, `kmod-rtl8192s-common`, and `rtl8192su-firmware` selected; both driver APKs and the firmware APK were generated. Fixed in `90e16846`; the PR description was updated as well.

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
```diff
@@ -0,0 +1,12009 @@
...
++	/* The following is for Antenna
++	 * diff from Ant-B to Ant-A
++	 */
++	p_final_pwridx[0] = pwrlevel[0];
++	p_final_pwridx[1] = pwrlevel[1];
++
++	switch (rtlefuse->eeprom_regulatory) {
++	case 3:
++		/* The following is for calculation
++		 * of the power diff for Ant-B
```
openwrt-ai: This is a real behaviour fix, not a cleanup, and it deserves a mention in the commit message — plus a fresh runtime test, because it invalidates the test evidence in the PR body.

The previous condition was `GET_RX_STATUS_DEC_MACID(rxdesc) == 0x1ff`. That macro is [`SHIFT_AND_MASK_LE(__pdesc + 4, 0, 5)`](``https://github.com/openwrt/openwrt/blob/ffabf696ab687f438c2e2c1f23c22d9008371b41/package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch#L522-L523),`` i.e. a 5-bit field masked with `RTL92S_BIT_LEN_MASK_32(5)` = `0x1f`, so it could never equal `0x1ff`. `rtl92su_rx_event_hdl()` therefore returned `false` for every frame in every build up to `71e7144`, and `rtl_send_buffered_bc()` was never reached. The "WPA2 AP on channel 6, DHCP lease, 10/10 ICMP, routed Internet access" result in the PR description was produced against that dead path, and it is also from before the PM-clearing hunk was dropped in this same push — so with this head, TIM/buffered-broadcast delivery is exercised for the first time. Worth re-running the AP test against `ffabf69` and saying so, otherwise the series has no runtime coverage of the one path it exists to add.

The new event-ID extraction looks right, for what it's worth: bits 16..23 of the first dword after the RX descriptor matches the CMD ID position that [`_rtl92s_fill_h2c_cmd()` writes for the H2C direction](``https://github.com/openwrt/openwrt/blob/ffabf696ab687f438c2e2c1f23c22d9008371b41/package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch#L2108-L2119)`` (len at 0..15, ID at 16..23, seq at 24..30), and the `macid == 0x1f && tid == 0xf` sentinel is consistent with the field widths. The read itself is in bounds — [`_rtl_rx_completed()` at usb.c:554](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/usb.c#L554-L558) rejects anything shorter than `RTL_RX_DESC_SIZE + sizeof(struct ieee80211_hdr)` = 48 bytes, so the `__le32` at offset 24 is always present.

A short comment above the `if` naming the sentinel (`macid 0x1f / tid 0xf marks a C2H event packet`) and the header layout would save the next reader the same archaeology.

---

openwrt-ai: nit: `RTL_RX_DRV_INFO_UNIT` is defined here and never referenced anywhere in the 12k-line patch (it is the only occurrence of the identifier). Please drop it and regenerate the patch so the `@@ -0,0 +1,499 @@` header for `rtl8192su/trx.c` stays consistent — `make package/mac80211/refresh`.

---

openwrt-ai (commented):
Reviewed the changes since `71e7144`. The PS workaround is fully gone — no `IEEE80211_FCTL_PM` rewrite remains anywhere in the patch, the in-patch commit message and the outer commit message were both updated to drop the power-save claim, and the PR description now states that PS state is left to mac80211. With that removed, `sta_ps_start()` / `num_sta_ps` work normally again, so the TIM and buffered-broadcast paths this series adds are reachable as intended.

I re-validated the regenerated patch mechanically: every hunk header's old/new line counts match the actual hunk bodies (checked all hunks in the file, including the renumbered `usb.c` offsets after the 12-line removal and the `498` → `499` bump for `rtl8192su/trx.c`). The hunks against existing files still apply against Linux 6.18.44 with the only rejects being the expected `CONFIG_` → `CPTCFG_` differences in `rtlwifi/Kconfig` and `rtlwifi/Makefile`, which is what backports substitutes. CI on `ffabf69` is still queued, so nothing to ground there.

One substantive comment posted inline on the C2H event-detection fix, plus one nit.

---

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
rdmitry0911: Confirmed: the old 0x1ff comparison could never match the 5-bit MACID field, so the earlier runtime result did not exercise C2H delivery. The driver commit now records this behavioural fix, and the code documents the MACID 0x1f/TID 0xf sentinel plus the command-ID layout. A clean x86/64 build on base 0256e6a1e470 succeeded with kmod-rtl8192su, kmod-rtl8192cu, and rtl8192su-firmware selected. I did not claim a fresh hardware result: this task provides only the build host and explicitly excludes touching host Wi-Fi, so the thread remains open pending an RTL8192SU AP runtime retest. Fixed documentation in 7381a0bd.

rdmitry0911: Confirmed and fixed: removed the unused RTL_RX_DRV_INFO_UNIT definition and refreshed the rtl8192su/trx.c hunk count. The clean x86/64 build with both RTL8192SU and RTL8192CU selected succeeded. Fixed in 7381a0bd.

rdmitry0911: Fresh hardware retest completed on `609d2100` using a clean x86/64 snapshot at base `0256e6a1e470` (Linux 6.18.41, backports 6.18.39).

The retest exposed an additional startup bug: firmware/init skbs were going through the regular USB data path while it was still marked stopped. This head now uses a dedicated command URB path before `SET_USB_START()` and a separate completion that does not report command skbs as mac80211 data frames.

Runtime setup and results:
- Realtek `0bda:8172` as the WPA2 AP on channel 6;
- MediaTek `0e8d:7610` in a second OpenWrt guest as the client;
- successful authentication/association, DHCP lease, DNS, and routed NAT;
- 100/100 ICMP packets to the AP and 100/100 to `1.1.1.1`;
- no firmware-download, command-URB, or USB-stop errors; AP station counters showed zero TX retries and failures.

The MediaTek driver rejects `iw ... set power_save on` with `-EOPNOTSUPP`, so this validates the current WPA/AP/data path but still does not directly exercise the sleeping-station DTIM/C2H buffered-broadcast case. I updated the PR body to state that limitation and am leaving this thread open.

rdmitry0911: Follow-up for `eed0dd7a`: this update changes only the `usb.c` hunk context so the OpenWrt patch applies with `--fuzz=0`. A fresh clean build on `192.168.40.111` completed without fuzz or failed hunks and produced `kmod-rtl8192su`, `kmod-rtl8192s-common`, `kmod-rtl8192cu`, and `rtl8192su-firmware`. The packaged `rtl8192su.ko` (`a02f01e101d8...`) and `rtl_usb.ko` (`083ba8a7ad34...`) are byte-identical to the modules used in the hardware test reported above, so the runtime evidence remains applicable. The thread stays open for direct sleeping-station DTIM/C2H coverage.

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
```diff
@@ -0,0 +1,12006 @@
...
++	struct rtl_tcb_desc *cb_desc;
++	struct sk_buff *skb;
++	u32	len;
++
++	len = _rtl92s_get_h2c_cmdlen(MAX_TRANSMIT_BUFFER_SIZE, 1, &cmd_len);
++	skb = dev_alloc_skb(len);
++	if (!skb)
++		return -ENOMEM;
++	cb_desc = (struct rtl_tcb_desc *)(skb->cb);
++	cb_desc->queue_index = TXCMD_QUEUE;
```
openwrt-ai: Now that `rtl92su_cmd_send_packet()` feeds `tcb_desc->queue_index` straight into `rtl_usb_transmit_cmd()`, which does `ep_num = rtlusb->ep_map.ep_mapping[qnum]`, this line decides which USB endpoint every H2C command goes out on — and it uses a PCI ring index for that.

`TXCMD_QUEUE` is [`#define TXCMD_QUEUE 5` in pci.h](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/pci.h#L28), which is unrelated to [`enum rtl_txq` in usb.h](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/usb.h#L25-L40) where 5 happens to be `RTL_TXQ_MGT`. So on RTL8192SU all H2C commands (`H2C_UPDATE_BCN_CMD`, pwrmode, joinbss, …) are submitted to `ep_mapping[RTL_TXQ_MGT]` — 0x0d, or 0x0c in the 8-out-ep case per [`rtl92su_endpoint_mapping()`](``https://github.com/openwrt/openwrt/blob/eed0dd7af1ee4c1b51fe93f607eed888d13ba362/package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch#L11364-L11390)`` — while the firmware download in `_rtl92s_firmware_downloadcode()` deliberately uses [`_rtl92s_get_fw_queue()`](``https://github.com/openwrt/openwrt/blob/eed0dd7af1ee4c1b51fe93f607eed888d13ba362/package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch#L1755-L1763),`` which returns `RTL_TXQ_VO` (0x04) for SU and `TXCMD_QUEUE` only for SE.

The function is also internally inconsistent about it: [seven lines below](``https://github.com/openwrt/openwrt/blob/eed0dd7af1ee4c1b51fe93f607eed888d13ba362/package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch#L2184)`` the poll uses `_rtl92s_get_fw_queue(rtlpriv)`, i.e. the skb is queued as TXCMD_QUEUE but the poll targets the SU FW queue.

Is the MGT endpoint the intended H2C endpoint for RTL8192SU, or should this be `_rtl92s_get_fw_queue(rtlpriv)` like the download path and the `tx_polling()` call below? If it is intentional, please say so in a comment — as written it is only correct by the numeric coincidence `TXCMD_QUEUE == RTL_TXQ_MGT == 5` between two unrelated enums, which will break silently if either is ever renumbered.

---

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
```diff
@@ -0,0 +1,12006 @@
...
++	/* The following is for Antenna
++	 * diff from Ant-B to Ant-A
++	 */
++	p_final_pwridx[0] = pwrlevel[0];
++	p_final_pwridx[1] = pwrlevel[1];
++
++	switch (rtlefuse->eeprom_regulatory) {
++	case 3:
++		/* The following is for calculation
++		 * of the power diff for Ant-B
```
openwrt-ai, line 12051: The comment scopes this to firmware download, but the code path is wider than that: the only caller is the `.cmd_send_packet` op, and `rtl92s_firmware_set_h2c_cmd()` routes *every* runtime H2C command through it too (update-beacon, pwrmode, joinbss, …). So dropping the `IS_USB_STOP()` guard that [`_rtl_usb_transmit()` still has at usb.c:852-856](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/usb.c#L852-L856) applies to runtime commands as well, not just to the pre-`SET_USB_START()` download.

That is reachable during teardown: [`rtl_usb_stop()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/usb.c#L737-L768) calls `SET_USB_STOP(rtlusb)` at line 748 and only reaches `cancel_work_sync(&rtlpriv->works.update_beacon_work)` at line 755, so an already-running `update_beacon_work` will now submit a command URB after the device has been marked stopped, where it was previously dropped with an error print.

Could the bypass be narrowed to the case it is documented for — e.g. keep the `IS_USB_STOP()` check unless `tcb_desc->cmd_or_init == RTL92S_DESC_PACKET_TYPE_INIT`? Alternatively, if submitting during stop is harmless here (the URBs are anchored on `tx_submitted` and killed in `rtl_usb_cleanup()`), please widen the comment so it says the bypass covers all H2C commands rather than only firmware download.

---

openwrt-ai: nit: this is the only hunk header in the whole patch without the trailing function context — every other one carries it (`@@ -767,7 +773,7 @@ static void rtl_usb_stop(struct ieee8021`, `@@ -783,6 +789,7 @@ static void _rtl_submit_tx_urb(struct ie`, …), and the previous revision of this very hunk had `static struct urb *_rtl_usb_tx_urb_setup`. It applies fine either way, but the missing context is a sign the hunk was hand-edited rather than regenerated; a `make package/mac80211/refresh` would restore it.

```suggestion
@@ -841,6 +848,41 @@ static struct urb *_rtl_usb_tx_urb_setup
```

---

openwrt-ai (commented):
Reviewed the changes since `ffabf69` (two commits, `7381a0bd`/`609d2100`/`eed0dd7a` squashed into the current head).

The regenerated patch checks out mechanically: I re-verified every hunk header in the file — all 60+ old/new line counts match their hunk bodies, including the `@@ -0,0 +1,500 @@` bump for `rtl8192su/trx.c` after `RTL_RX_DRV_INFO_UNIT` was dropped and the renumbered `usb.c` offsets. A `patch -p1 --dry-run -F0` against Linux 6.18.44 (the version `x86` builds, `KERNEL_PATCHVER:=6.18`) now applies the `usb.c`, `usb.h`, `wifi.h`, `base.c`, `base.h` and `core.c` hunks with zero fuzz; the only rejects are the expected `CONFIG_` → `CPTCFG_` differences in `rtlwifi/Kconfig` and `rtlwifi/Makefile` that backports substitutes. So the fuzz fix in `eed0dd7a` does what it claims.

On the new command path itself: `_rtl_usb_tx_urb_setup()` does not free the skb on `usb_alloc_urb()` failure, so the `kfree_skb()` in `rtl_usb_transmit_cmd()` is correct and there is no double free; `_rtl_install_trx_info()` takes no extra skb reference, so the single `dev_kfree_skb_any()` in `_rtl_usb_cmd_tx_complete()` balances; the submit-failure path in `_rtl_submit_tx_urb()` frees the skb and the error is propagated. The C2H comment requested last round is in place, and the outer commit message now records the sentinel fix as a behaviour change. CI on `eed0dd7a` is green (all three FormalityCheck runs), so nothing to ground there.

Two substantive comments posted inline (H2C endpoint selection, and the scope of the new `IS_USB_STOP()` bypass), plus one nit on the hand-edited hunk header.

---

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
rdmitry0911: Confirmed the numeric-domain bug. `_rtl92s_get_fw_queue()` is not correct for runtime H2C: the vendor RTL8712 USB mapping sends H2C on endpoint 0x0d, while RTL8192SU firmware download uses VO/0x04. I added a dedicated `RTL_TXQ_CMD`, mapped it to 0x0d for all supported USB endpoint layouts, and kept `TXCMD_QUEUE` for the shared RTL8192SE path. The clean x86/64 build covered both RTL8192SU and RTL8192CU. Fixed in eba8d104.

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
rdmitry0911: Confirmed the teardown race. `rtl_usb_transmit_cmd()` now checks both `IS_USB_STOP()` and the HAL stopped state, frees and rejects runtime command skbs with `-ENODEV`, while only firmware initialization packets explicitly bypass the pre-start stopped state. The clean x86/64 build covered both RTL8192SU and RTL8192CU. Fixed in eba8d104.

rdmitry0911: Refreshed the patch with `make package/mac80211/refresh`; the hunk header now includes the function context. A clean `package/mac80211/prepare` applies it without fuzz. Fixed in eba8d104.

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
```diff
@@ -0,0 +1,12027 @@
...
++				pwrbase1;
++	*mcsbase = pwrbase1;
++
++	/* The following is for Antenna
++	 * diff from Ant-B to Ant-A
++	 */
++	p_final_pwridx[0] = pwrlevel[0];
++	p_final_pwridx[1] = pwrlevel[1];
++
++	switch (rtlefuse->eeprom_regulatory
```
openwrt-ai, line 12073: Adding `RTL_TXQ_CMD` inside `enum rtl_txq` bumps `__RTL_TXQ_NUM` from 7 to 8, and that breaks probe for **rtl8192cu and rtl8192du**, which this patch does not touch.

`_rtl_usb_init_tx()` validates the whole map and rejects any zero entry:

```c
for (i = 0; i < __RTL_TXQ_NUM; i++) {
	u32 ep_num = rtlusb->ep_map.ep_mapping[i];

	if (!ep_num) {
		rtl_dbg(rtlpriv, COMP_INIT, DBG_DMESG,
			"Invalid endpoint map setting!\n");
		return -EINVAL;
	}
}
```
([`_rtl_usb_init_tx()` at usb.c:184-192](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/usb.c#L184-L192); this is the only consumer of `__RTL_TXQ_NUM` in the tree.)

`rtl_usb` lives in the zeroed `ieee80211_alloc_hw()` private area, and only [`rtl92su_endpoint_mapping()` in this patch](``https://github.com/openwrt/openwrt/blob/eba8d10457da3d8f15e2905653118760880bf45f/package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch#L11369)`` fills index 7. The other two USB drivers stop at `RTL_TXQ_HI`:

- [`_rtl_twoout_epmapping()` at rtl8192cu/trx.c:89-95](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/rtl8192cu/trx.c#L89-L95) (and the sibling one/three-out-ep helpers in the same file)
- [`_rtl92du_one_out_ep_mapping()` … `_rtl92du_three_out_ep_mapping()` at rtl8192du/trx.c:267-300](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/rtl8192du/trx.c#L267-L300)

So on any tree with `kmod-rtl8192cu` or `kmod-rtl8192du` installed, `ep_mapping[RTL_TXQ_CMD] == 0` → `_rtl_usb_init_tx()` returns `-EINVAL` → `rtl_usb_probe()` fails and the adapter never comes up. It compiles fine, which is why the "clean x86/64 build with both RTL8192SU and RTL8192CU selected" doesn't catch it — the failure is at probe time, not build time.

Two ways out, either is fine:

- Exempt the new command queue from the check in `_rtl_usb_init_tx()` (e.g. `if (i == RTL_TXQ_CMD) continue;`, since only RTL8192SU has a dedicated command endpoint), or
- Add `ep_map->ep_mapping[RTL_TXQ_CMD] = ...` alongside the existing `RTL_TXQ_HI` assignments in `rtl8192cu/trx.c` and `rtl8192du/trx.c` so every driver still populates a full map.

Whichever you pick, please regenerate the patch afterwards (`make package/mac80211/refresh`) so the hunk headers stay consistent.

---

openwrt-ai (commented):
Reviewed the changes since `eed0dd7a` (the delta is confined to `990-rtlwifi-add-rtl8192su-support.patch`).

The two findings from last round are addressed: H2C commands no longer ride on the `TXCMD_QUEUE`/`RTL_TXQ_MGT` numeric coincidence but go through a dedicated `RTL_TXQ_CMD` mapped to endpoint 0x0d, and the `IS_USB_STOP()` bypass is now narrowed to firmware init packets (`is_init`), with runtime H2C rejected during teardown via `IS_USB_STOP() || is_hal_stop()`. I checked that no H2C *command* skb is emitted while `is_hal_stop()` is still true during `hw_init()` — `rtl92s_phy_set_fw_cmd()` goes through `REG_WFM5` register writes, and the only `rtl92s_firmware_set_h2c_cmd()` callers (pwrmode, joinbss, update-beacon) all run post-start — so the new rejection does not break bring-up. `0x0d` is also a plausible unconditional choice: it is the one endpoint present in all three of the 3/5/8-out-ep layouts the mapping function handles.

Mechanically the patch is still sound: I re-verified all 48 hunk headers (old/new line counts against the hunk bodies, plus the running offsets in `usb.c` and `usb.h`), including the `@@ -0,0 +1,668 @@` and `@@ -0,0 +1,502 @@` bumps — all consistent, and the hand-edited `usb.c` hunk header now carries its function context again. `patch -p1 --dry-run -F0` against Linux 6.18.44 applies `base.c`, `base.h`, `core.c`, `usb.c`, `usb.h` and `wifi.h` with zero fuzz; the only rejects remain the expected `CONFIG_` → `CPTCFG_` differences in `rtlwifi/Kconfig` and `rtlwifi/Makefile` that backports substitutes. CI on `eba8d104` is still pending, so nothing to ground there.

One blocking issue posted inline: the new `RTL_TXQ_CMD` enumerator widens `__RTL_TXQ_NUM`, which makes `_rtl_usb_init_tx()` reject the (now incomplete) endpoint maps of rtl8192cu and rtl8192du, breaking probe for those devices.

---

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
rdmitry0911: Confirmed the probe regression: RTL_TXQ_CMD widened __RTL_TXQ_NUM while RTL8192CU and RTL8192DU still leave that optional entry unset, so shared USB endpoint validation returned -EINVAL. The refreshed patch now skips only the dedicated command queue during shared validation; all existing queues remain mandatory and RTL8192SU still maps its command endpoint to 0x0d. A clean x86/64 build on base 0256e6a1e470 with kmod-rtl8192su, kmod-rtl8192cu, and rtl8192su-firmware enabled produced all three APKs. Fixed in 60713484.

openwrt-ai (commented):

The delta since `eba8d104` is confined to `990-rtlwifi-add-rtl8192su-support.patch` and is exactly the fix for the probe regression flagged last round — a new `usb.c` hunk skipping `RTL_TXQ_CMD` in the shared endpoint-map validation, plus the renumbered offsets it forces on the four later `usb.c` hunks.

What I checked:

- `_rtl_usb_init_tx()`'s loop is validation-only — nothing else in the function consumes `ep_num` per-iteration, and `usb.c:184` is still the only consumer of `__RTL_TXQ_NUM` in the tree — so skipping index 7 is sufficient and side-effect-free. rtl8192cu and rtl8192du, which stop populating at `RTL_TXQ_HI`, now pass validation again.
- rtl8192su's own command endpoint is not left unvalidated in practice: `rtl92su_endpoint_mapping()` sets `ep_mapping[RTL_TXQ_CMD] = 0x0d` unconditionally before the `out_ep_nums` switch, and `_rtl_usb_init()` runs `usb_endpoint_mapping()` before `_rtl_usb_init_tx()`, so `rtl_usb_transmit_cmd()` can never build a bulk URB on endpoint 0.
- `RTL_TXQ_CMD` is still appended last in `enum rtl_txq`, so no existing enumerator is renumbered and `struct rtl_ep_map` growth stays additive.
- Mechanically the patch is still sound: all 49 hunk headers' old/new line counts match their bodies, including the renumbered `219 → 222`, `489 → 494`, `767 → 776`, `783 → 792` and `841 → 851` offsets in `usb.c`. `patch -p1 --dry-run -F0` against Linux 6.18.44 applies `base.c`, `base.h`, `core.c`, `usb.c`, `usb.h` and `wifi.h` with zero fuzz; the only rejects remain the expected `CONFIG_` → `CPTCFG_` differences in `rtlwifi/Kconfig` and `rtlwifi/Makefile` that backports substitutes. The new hunk header carries its function context, so it was regenerated rather than hand-edited.
- The commit message picked up a matching sentence ("Keep the command queue optional during shared endpoint-map validation so existing rtlwifi USB devices continue to probe"), which accurately describes the diff.

CI on `60713484` is green across the FormalityCheck, `Check Kernel patches`, `Build Kernel` and `Build all core packages` jobs I sampled, so nothing to ground there.

---

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
rdmitry0911: Follow-up for d43a790b: the earlier NULL/opmode guard did not keep mac->vif alive across concurrent remove_interface(). The C2H buffered-broadcast reader now holds an RCU read-side critical section over ieee80211_get_buffered_bc(), with RCU publication/clear of mac->vif; beacon work now reads vif only while holding conf_mutex, the same lock used by remove_interface(). A clean x86_64 build on base 0256e6a1e470 succeeded with kmod-rtl8192su, kmod-rtl8192cu, and rtl8192su-firmware selected; rtl8192su.ko, rtl8192cu.ko, and the firmware APK were produced. This thread was already resolved, so I leave its resolution state unchanged.

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
```diff
@@ -0,0 +1,12083 @@
...
+ void rtl_c2hcmd_enqueue(struct ieee80211_hw *hw, struct sk_buff *skb);
+ 
+--- a/drivers/net/wireless/realtek/rtlwifi/core.c
++++ b/drivers/net/wireless/realtek/rtlwifi/core.c
+@@ -290,7 +290,7 @@ static int rtl_op_add_interface(struct ieee80211_hw *hw,
+ 		rtlpriv->cfg->ops->set_hw_reg(hw, HW_VAR_BASIC_RATE,
+ 				(u8 *)(&mac->basic_rates));
+ 	}
+-	mac->vif = vif;
++	rcu_assign_pointer(mac->vif, vif);
```
openwrt-ai: `mac->vif` is still declared as a plain pointer, so all three RCU macros this push adds will trip sparse. [`struct rtl_mac::vif` at wifi.h:1448](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/wifi.h#L1448) is `struct ieee80211_vif *vif;`, and the `wifi.h` hunks in this patch (lines 12056-12083) only touch `rtl_tcb_desc`, `rtl_hal_ops` and `rtl_hal_usbint_cfg` — the declaration is unchanged. `rcu_assign_pointer()` here, `RCU_INIT_POINTER()` at line 145 and `rcu_dereference()` at line 100 each expand `rcu_check_sparse(p, __rcu)`, which emits "incompatible types in comparison expression (different address spaces)" against a non-`__rcu` field. It builds fine (the check is `__CHECKER__`-only), so CI is green, but for a patch aimed at backports/upstream this is the kind of thing that gets bounced.

Making it consistent means annotating the field and converting the remaining readers to `rcu_dereference_protected(mac->vif, lockdep_is_held(&rtlpriv->locks.conf_mutex))` (or `rcu_access_pointer()` for the bare NULL test). The affected sites outside this hunk are small in number: [`core.c:207-209`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/core.c#L207-L209), [`core.c:563`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/core.c#L563), [`base.c:2194`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/base.c#L2194), [`pci.c:1037`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/pci.c#L1037), [`rtl8192ce/trx.c:374`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/rtl8192ce/trx.c#L374), [`wifi.h:3028`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/wifi.h#L3028), plus the new `vif = rtlpriv->mac80211.vif;` at line 157 of this patch and the `mac->vif` reads in the new driver ([line 10240](``https://github.com/openwrt/openwrt/blob/d43a790b659fcfa7a1003092e233feedb112fde9/package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch#L10240)`` and [line 11797](``https://github.com/openwrt/openwrt/blob/d43a790b659fcfa7a1003092e233feedb112fde9/package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch#L11797)).``

Alternatively, if touching the other rtlwifi drivers is unwanted here, keep the `rcu_read_lock()` (which is what actually provides the lifetime guarantee) and drop the annotation-requiring macros in favour of plain `WRITE_ONCE()`/`READ_ONCE()` — same generated code, no sparse noise.

---

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
```diff
@@ -0,0 +1,12083 @@
...
+ int rtl_rx_agg_stop(struct ieee80211_hw *hw,
+ 		    struct ieee80211_sta *sta, u16 tid);
+ void rtl_rx_ampdu_apply(struct rtl_priv *rtlpriv);
++void rtl_send_buffered_bc(struct ieee80211_hw *hw);
+ void rtl_c2hcmd_launcher(struct ieee80211_hw *hw, int exec);
+ void rtl_c2hcmd_enqueue(struct ieee80211_hw *hw, struct sk_buff *skb);
+ 
+--- a/drivers/net/wireless/realtek/rtlwifi/core.c
++++ b/drivers/net/wireless/realtek/rtlwifi/core.c
+@@ -290,7 +290,7 @@ static int rtl_op_add_interface(struct ieee80211_hw *hw,
```
openwrt-ai: nit: this hunk header was not produced by the same tool as the rest of the file. GNU `diff -p` (what quilt, and therefore `make package/mac80211/refresh`, uses) truncates the function context to 40 characters — every other one of the 52 hunk headers in this patch obeys that, e.g. `@@ -1731,7 +1733,8 @@ static int rtl_op_set_tim(struct ieee802` at line 171. The three core.c headers added by this push carry 57, 61 and 63 characters (this one, plus line 140 `static void rtl_op_remove_interface(struct ieee80211_hw *hw,` and line 149 `void rtl_update_beacon_work_callback(struct work_struct *work)`), which is `git diff` output.

No functional impact — I re-verified all 52 headers and every old/new line count matches its hunk body, the running offsets are consistent (`992` stays `992`, `1731` → `1733`), and `patch -p1 --dry-run -F0` against Linux 6.18.44 applies `base.c`, `base.h`, `core.c`, `usb.c`, `usb.h` and `wifi.h` with zero fuzz. But it does mean these hunks were hand-assembled rather than regenerated; a `make package/mac80211/refresh` pass would make the whole file uniform again and rule out the class of mismatch that had to be fixed in `eed0dd7a`.

---

openwrt-ai (commented):
Reviewed the changes since `60713484`. The delta is confined to `990-rtlwifi-add-rtl8192su-support.patch` and does three things: RCU-protects the C2H buffered-broadcast reader, moves the beacon-work vif read under `conf_mutex`, and reserves `RTL_TX_HEADER_SIZE` headroom on the two command skbs.

The lifetime fix looks sound to me. Writes to `mac->vif` in [`rtl_op_add_interface()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/core.c#L293) and [`rtl_op_remove_interface()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/core.c#L332) both already happen under `conf_mutex`, so the beacon-work reader is now correctly serialised against removal, and there is no new deadlock — the only `cancel_work_sync(&rtlpriv->works.update_beacon_work)` is in [`rtl_usb_stop()` at usb.c:755](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/realtek/rtlwifi/usb.c#L755), which does not hold that mutex. On the tasklet side, no explicit `synchronize_rcu()` is needed in `rtl_op_remove_interface()` because mac80211 guarantees a grace period between `drv_remove_interface()` and freeing the vif: `unregister_netdevice_many_notify()` runs `dev_close_many()` (which reaches `ieee80211_do_stop()`) before its `synchronize_net()`, and [`ieee80211_if_remove()` at iface.c:2342](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/net/mac80211/iface.c#L2342) adds another for the non-netdev case. The `goto out` restructuring keeps `adapter_tx()` inside the read-side section, which is fine — it is already an atomic-context path (`usb_alloc_urb`/`usb_submit_urb` with `GFP_ATOMIC`).

The publish/clear ordering also fails safe: `rcu_assign_pointer(mac->vif, vif)` runs one line before `mac->opmode = vif->type`, so a reader that catches the window sees a fresh vif with a stale opmode and either bails on the opmode test or hits `ieee80211_get_buffered_bc()` on a non-AP vif, which returns NULL via the `sdata->bss` check.

The skb headroom change is defensive rather than a crash fix — `RTL_TX_HEADER_SIZE` is 32 and `dev_alloc_skb()` already reserves `NET_SKB_PAD` = `max(32, L1_CACHE_BYTES)`, so `rtl92su_cmd_send_packet()`'s `skb_push()` was fitting by coincidence. Reserving it explicitly is the right call; tailroom is unchanged in both sites, so `_rtl92s_fill_h2c_cmd()` and the firmware-fragment `skb_put()` still have exactly the space they had.

Mechanically the patch still checks out: all 52 hunk headers' old/new counts match their bodies, offsets chain correctly, and `patch -p1 --dry-run -F0` against Linux 6.18.44 applies everything with zero fuzz except the expected `CONFIG_` → `CPTCFG_` rejects in `rtlwifi/Kconfig` and `rtlwifi/Makefile` that backports substitutes. The commit message accurately describes the RCU/`conf_mutex` change, and all three FormalityCheck runs on `d43a790b` are green, so nothing to ground from CI.

One issue posted inline (missing `__rcu` annotation for the newly RCU-managed `mac->vif`), plus one nit on the hunk headers.

---

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
rdmitry0911: Fixed in 763d7d69. The tasklet still holds rcu_read_lock() for vif lifetime, but the shared, unannotated mac->vif field now uses READ_ONCE()/WRITE_ONCE(). This avoids sparse __rcu diagnostics without changing unrelated rtlwifi readers. Regenerated with make package/kernel/mac80211/refresh and built on base 0256e6a1e470 with kmod-rtl8192su, kmod-rtl8192cu, and rtl8192su-firmware enabled.

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
rdmitry0911: Fixed in 763d7d69. Regenerated with make package/kernel/mac80211/refresh; the core.c hunk contexts now use quilt/diff -p 40-character form. The same patch built on base 0256e6a1e470 with kmod-rtl8192su, kmod-rtl8192cu, and rtl8192su-firmware enabled.

openwrt-ai (commented):

---

hauke:
Did you send the patch to upstream Linux? It would be nice to have this also in upstream Linux instead of shipping this always in OpenWrt.

openwrt-ai (commented):

---

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
rdmitry0911: Follow-up in `c205b88b`: an independent audit found and fixed two additional code-level blockers in this C2H path.

First, `_rtl_rx_completed()` required an RX descriptor plus a full 802.11 header before invoking `usb_rx_event_hdl`, so short command-only C2H transfers were discarded before the handler could see them. Event-capable drivers now admit a descriptor plus the C2H command word; `_rtl_rx_work()` retains the full 802.11 minimum-length guard before normal frame processing.

Second, the Realtek/r8712 event enum defines `0x15` as ADDBA and C2HBCN as `0x16`; the handler now uses `0x16`.

The patch was quilt-refreshed without a byte change and built cleanly on the pinned base `0256e6a1e470` with `kmod-rtl8192su`, `kmod-rtl8192cu`, and `rtl8192su-firmware` enabled. All three FormalityCheck jobs on this head also pass. I did not run the direct sleeping-station DTIM/C2H hardware test, so I am leaving this thread unresolved for that remaining runtime coverage.

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
```diff
@@ -0,0 +1,12114 @@
...
++ */
++
++#include "../wifi.h"
++#include "reg_common.h"
++#include "def_common.h"
++#include "phy_common.h"
++#include "rf_common.h"
++#include "dm_common.h"
++
++stat
```
openwrt-ai: The C2H beacon event ID changed from `0x15` to `0x16` in this push, and neither the in-patch nor the outer commit message records it — unlike the sentinel fix, which does. Per the still-open [C2H thread](https://github.com/openwrt/openwrt/pull/24677#discussion_r3772631179) this path has no hardware coverage yet, so neither value has runtime evidence behind it: what is the source for `0x16`, and can it become a named constant next to the `H2C_*` enum in `fw_common.h` instead of a bare literal?

---

openwrt-ai: nit: `sizeof(__le32)` is exactly the number of bytes [`rtl92su_rx_event_hdl()`](``https://github.com/openwrt/openwrt/blob/c205b88b1f9eb1bf4086bda300d080fe0db1ad17/package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch#L11843)`` reads past `RTL_RX_DESC_SIZE`, but nothing in this shared file says so, so widening that read later turns this into a silent out-of-bounds access. A named constant in `usb.h` (or a one-line comment naming the C2H header word) would tie the bound to its reader.

---

openwrt-ai (commented):
Reviewed 2 new commits.

---

package/kernel/mac80211/patches/rtl/990-rtlwifi-add-rtl8192su-support.patch
rdmitry0911: Agreed. `usb.h` now defines the packed four-byte `struct rtl_usb_c2h_header`; `_rtl_rx_completed()` uses `sizeof(struct rtl_usb_c2h_header)` for the event-capable minimum, while `rtl92su_rx_event_hdl()` reads `c2h->id` from that same type. The bound and reader therefore change together if the header layout is widened. The pinned-base x86/64 build succeeded with `kmod-rtl8192su`, `kmod-rtl8192cu`, and `rtl8192su-firmware` enabled. Fixed in `9e5e1023`.

rdmitry0911: Confirmed. `0x16` comes from the Realtek/r8712 C2H event table: ADDBA is event `0x15`, and C2HBCN/HBCN is the following `0x16` entry (also preserved as `C2H_HBCN_EVENT` in the archived source at https://github.com/chunkeey/rtl8192su/blob/master/r92su/h2cc2h.h). The literal is now the named `C2H_HBCN_EVENT = 0x16` beside the H2C enum, and both the in-patch and outer commit messages record the correction. The pinned-base x86/64 build succeeded with `kmod-rtl8192su`, `kmod-rtl8192cu`, and `rtl8192su-firmware` enabled. Fixed in `9e5e1023`.

rdmitry0911 (commented):
Addressed both inline findings in 9e5e1023; details and pinned-base build verification are in the replies.

openwrt-ai (commented):

---

openwrt-ai (commented):

---

PR #24679, opened 2026-08-11
mediatek: filogic: add Adtran SmartRG Mount Rainier series

cmonroe:
Common specifications:
 * MediaTek MT7987A (4x Cortex-A53)
 * 4 GiB eMMC
 * 1 GiB DDR4 RAM
 * 1x 2500M/1000M/100M/10M WAN port (MaxLinear MXL86211C PHY)
 * 2x 2500M/1000M/100M/10M LAN ports (MaxLinear MXL86211C PHY, MT7987 built-in 2.5G PHY)
 * 3x LED (green/amber/white) for each Ethernet port
 * 2 buttons (Reset, Mesh/WPS)
 * uC-controlled RGB LED via I2C
 * USB 2.0 type A port
 * 3.3V-level 115200 baud UART console
 * 12V 3A DC barrel jack power input

SDG-8712: MT7992 dual-band 802.11be Wi-Fi (2.4G 4T4R, 5G 4T5R)
SDG-8732: MT7996 tri-band 802.11be Wi-Fi (2.4G 2T2R, 5G 3T3R, 6G 3T3R)

Both models are also available in versions including 1x FXS POTS interface for analog phones. Those interfaces are not supported by OpenWrt.

Installation:

1. Prepare a TFTP server at 192.168.1.10 and serve the OpenWrt initramfs image, renamed to "openwrt.<engnum>" for the model being installed:
        openwrt.440: SDG-8712
        openwrt.442: SDG-8732
2. Connect to LAN2.
3. Connect to the UART port.
4. Power on the device and type "st" to enter the U-Boot command line. The window to do that might be very short (below 1s), so type it as soon as the "U-Boot ..." string appears.
5. Run "setenv boot_mode openwrt; saveenv" to boot from OpenWrt partition by default.
6. Run "bootmenu" and select option 6 (TFTP openwrt image and boot); "run boot5" is a shortcut for the same. This downloads openwrt.<engnum> from the TFTP server and boots it.
7. After OpenWrt finishes booting, download the OpenWrt sysupgrade image and write it to eMMC using "sysupgrade".
8. After sysupgrade finishes writing, the device reboots into OpenWrt. When the front LED is solid white the device is ready for configuration.


target/linux/mediatek/dts/mt7987a-smartrg-mt-rainier.dtsi
```diff
@@ -0,0 +1,336 @@
...
+	nvmem-cell-names = "mac-address";
+	label = "lan1";
+	phy-mode = "2500base-x";
+	phy-handle = <&phy5>;
+
+	status = "okay";
+};
+
+&gmac1 {
+	nvmem-cells = <&macaddr 0x1>;
```
openwrt-ai, line 108: `gmac1` (lan2) takes `<&macaddr 0x1>`, the same index `gmac0` (lan1) uses on line 99, so both LAN netdevs come up with an identical MAC address. The Adtran layout adds the cell index to the last nibble of `MFG_MAC` and reserves 16 addresses per device ([adtran_mac_address_pp in 450-nvmem-add-layout-for-Adtran-devices.patch](``https://github.com/openwrt/openwrt/blob/de800d05210ef0f68312cc4c39535bd733850cc1/target/linux/mediatek/patches-6.18/450-nvmem-add-layout-for-Adtran-devices.patch)),`` and this board only consumes 0x0, 0x1, 0x4, 0x6 and 0xa — 0x2/0x3/0x5 are free.

On the sibling Mt. Stuart boards each interface gets its own index ([`&gmac1`/`&gmac2` in mt7988a-smartrg-mt-stuart-phase1.dtsi:107-118](``https://github.com/openwrt/openwrt/blob/de800d05210ef0f68312cc4c39535bd733850cc1/target/linux/mediatek/dts/mt7988a-smartrg-mt-stuart-phase1.dtsi#L107-L118)),`` but there lan2-lan4 are MT7530 switch ports that legitimately share the conduit MAC, whereas here lan1 and lan2 are two independent GMACs. Is the shared address deliberate (matching OEM firmware), or should lan2 get its own index?

---

target/linux/mediatek/dts/mt7987a-smartrg-mt-rainier.dtsi
```diff
@@ -0,0 +1,336 @@
...
+			label = "sync";
+			linux,code = <KEY_WPS_BUTTON>;
+			gpios = <&pio 0 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+	leds: gpio-leds {
+		compatible = "gpio-leds";
+
+		lan2_white {
```
openwrt-ai: nit: device-tree node names use hyphens, not underscores, and the other recent mt7987a boards name `gpio-leds` children `led-N` (e.g. [`led_status_blue: led-0` in mt7987a-routerich-be7200.dts:58](https://github.com/openwrt/openwrt/blob/de800d05210ef0f68312cc4c39535bd733850cc1/target/linux/mediatek/dts/mt7987a-routerich-be7200.dts#L58)). The node name has no effect on the resulting LED name here (`color` + `function` produce `white:lan2`), so renaming is safe.

```suggestion
		led-0 {
```

---

target/linux/mediatek/dts/mt7987a-smartrg-mt-rainier.dtsi
```diff
@@ -0,0 +1,336 @@
...
+		};
+	};
+
+	phy15: ethernet-phy@f {
+		/* MT7987 - LAN2 */
+		pinctrl-names = "i2p5gbe-led";
+		pinctrl-0 = <&i2p5gbe_led0_pins>, <&i2p5gbe_led1_1_pins>;
+		compatible = "ethernet-phy-ieee802.3-c45";
+		reg = <15>;
+		phy-mode = "internal";
```
openwrt-ai: nit: `phy-mode` is a MAC-side property (`ethernet-controller` binding), not part of `ethernet-phy`; it is already set on `&gmac1` at line 112. No other in-tree mediatek DTS puts it on a PHY node — the same internal PHY on GL-MT3600BE carries only compatible/reg/pinctrl ([mt7987a-glinet-gl-mt3600be.dts:123-128](``https://github.com/openwrt/openwrt/blob/de800d05210ef0f68312cc4c39535bd733850cc1/target/linux/mediatek/dts/mt7987a-glinet-gl-mt3600be.dts#L123-L128)).`` It is silently ignored, so it can just be dropped.

```suggestion
```

---

target/linux/mediatek/filogic/base-files/etc/board.d/02_network
```diff
@@ -193,6 +193,10 @@ mediatek_setup_interfaces()
 	yuncore,ax835)
 		ucidef_set_interfaces_lan_wan lan wan
 		;;
+	smartrg,sdg-8712|\
```
openwrt-ai: nit: the case blocks in `mediatek_setup_interfaces()` are ordered alphabetically by their first compatible, but this one lands between the `comfast,cf-wr632ax|…|yuncore,ax835)` block and `keenetic,kn-1812)`. It belongs after the `mercusys,mr90x-v1-ubi)` block and before `tenda,be12-pro)` (lines 220/221 on this branch). The 01_leds, platform.sh and uboot-envtools hunks in this PR are all placed correctly, so this looks like an oversight rather than intent.

---

openwrt-ai (commented):
## Commit checks

- `de800d05` "mediatek: filogic: add Adtran SmartRG Mount Rainier series" — new-device commit. Hardware specification and flash instructions are both present and detailed, but the **MAC address layout** section is missing. Please add one line per interface (wan / lan1 / lan2 / 2.4G / 5G / 6G) saying where each address comes from — here that is the `mfginfo` partition's `MFG_MAC` plus the per-interface index used in the DTS (`0x0`, `0x1`, `0x4`, `0xa`, `0x6`) — using `xx:xx:xx` for the unit-specific bytes.


---

target/linux/mediatek/dts/mt7987a-smartrg-mt-rainier.dtsi
cmonroe: this is intentional since both are LAN ports which end up in br-lan.

target/linux/mediatek/dts/mt7987a-smartrg-mt-rainier.dtsi
```diff
@@ -0,0 +1,335 @@
...
+			gpios = <&pio 29 GPIO_ACTIVE_HIGH>;
+		};
+	};
+};
+
+&eth {
+	pinctrl-0 = <&mdio0_pins>;
+	pinctrl-names = "default";
+	status = "okay";
+};
```
openwrt-ai, line 95: nit: the SoC dtsi already sets exactly this default pinctrl state on the `eth` node ([`pinctrl-names` / `pinctrl-0 = <&mdio0_pins>` at mt7987.dtsi:1118-1119](https://github.com/openwrt/openwrt/blob/961a9129800d1f42f90a85bea05a1c2e1b974261/target/linux/mediatek/dts/mt7987.dtsi#L1118-L1119)), so repeating it here has no effect. The other mt7987a board that overrides `&eth` only flips the status ([mt7987a-hiveton-h5000m.dts:93-95](https://github.com/openwrt/openwrt/blob/961a9129800d1f42f90a85bea05a1c2e1b974261/target/linux/mediatek/dts/mt7987a-hiveton-h5000m.dts#L93-L95)).

```suggestion
&eth {
	status = "okay";
};
```

---

openwrt-ai (commented):
Reviewed 1 new commit (force-push: rebase onto current main plus the fixes from the previous round — `led-0` rename, stray `phy-mode` on the internal PHY dropped, `02_network` case moved into alphabetical order, and the MAC address layout section added to the commit message). The shared `lan1`/`lan2` MAC is answered as intentional, so that is dropped.

One new nit inline; nothing blocking. CI on `961a912` is green/queued at the time of review.

---

openwrt-ai (commented):

The only change since the last review (`961a912` → `2eb0612`) is dropping the redundant `pinctrl-names`/`pinctrl-0 = <&mdio0_pins>` from `&eth`, which was the previous round's nit — the SoC dtsi already sets that default state. Commit message is unchanged and still carries the hardware spec, flash instructions and MAC address layout sections. CI on `2eb0612` is green.

---

target/linux/mediatek/dts/mt7987a-smartrg-mt-rainier.dtsi
```diff
@@ -0,0 +1,333 @@
...
+		mux {
+			function = "pcie";
+			groups = "pcie0_pereset", "pcie0_clkreq";
+		};
+	};
+};
+
+&ssusb {
+	status = "okay";
+};
```
hauke, line 335: Is USB3 working?
I think you also need:
```
&tphyu3port0 {
	status = "okay";
};
```

target/linux/mediatek/dts/mt7987a-smartrg-mt-rainier.dtsi
```diff
@@ -0,0 +1,333 @@
...
+
+&uart0 {
+	status = "okay";
```
hauke, line 340: i think you have to activate the UART pin control too:
```
	pinctrl-names = "default";
	pinctrl-0 = <&uart0_pins>;
```

target/linux/mediatek/dts/mt7987a-smartrg-mt-rainier.dtsi
```diff
@@ -0,0 +1,333 @@
...
+			gpios = <&pio 0 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+	leds: gpio-leds {
+		compatible = "gpio-leds";
+
+		led-0 {
+			color = <LED_COLOR_ID_WHITE>;
+			function = "lan2";
```
hauke: I would prefer instead of the label:
```
function = LED_FUNCTION_LAN;
function-enumerator = <2>;
```

It then gets the name: `white:lan-2`

target/linux/mediatek/dts/mt7987a-smartrg-mt-rainier.dtsi
cmonroe: It is not.. because we only wired up USB 2.0 on this project :)

I fixed my bad note in the description to match the hardware correctly.

target/linux/mediatek/dts/mt7987a-smartrg-mt-rainier.dtsi
cmonroe: Nice catch, thank you! I've been working on getting all of these static labels/function names out of our tree as well but this one slipped through.

openwrt-ai (commented):

Changes since the last review (`2eb0612` → `5fe8767`, force-push = rebase onto current main plus the fixes from @hauke's round):

- `led-0`: `function = "lan2"` → `LED_FUNCTION_LAN` + `function-enumerator = <2>`, and `function-enumerator` added to the `phy5` (LAN1) and `phy15` (LAN2) PHY LEDs. Checked against `led_compose_name()` in 6.18 — with `func_enum_present` the name becomes `<color>:<function>-<enum>`, and for PHY LEDs `devname_mandatory` is set, so these resolve to `mdio-bus:05:{green,amber,white}:lan-1`, `mdio-bus:0f:{green,amber}:lan-2` and `white:lan-2`. That matches the updated `01_leds` block exactly, and `mdio-bus:04:{green,amber,white}:wan` is unchanged and still correct.
- `&uart0` gained `pinctrl-names`/`pinctrl-0 = <&uart0_pins>`. `uart0_pins` exists in `mt7987.dtsi` and every other in-tree mt7987a board sets it, so this matches convention. `uart0` maps to pins 31/32, which don't collide with GPIO 29 (white LAN2 LED), `i2c0_2` (pins 43/44) or the 2.5G PHY LED pins (45/49).
- Commit message: "USB 3.0 type A port" → "USB 2.0 type A port". Consistent with leaving `&tphyu3port0` disabled; `phy-mtk-tphy` registers all `usb-phy@*` children with `for_each_child_of_node_scoped()` regardless of `status`, so the USB2 port still comes up, and `kmod-usb3` is still the right package (it pulls `kmod-usb-xhci-mtk` on `TARGET_mediatek`).

Commit message still carries the hardware spec, flash instructions and MAC address layout sections. CI on `5fe8767` is all success/skipped.

---

jonasjelonek:
nit: keep the "SDG" in the DTS file names as lowercase, to stay consistent with the usual convention

cmonroe:
> nit: keep the "SDG" in the DTS file names as lowercase, to stay consistent with the usual convention

I was torn on this since all of our other devices in OpenWrt use "SDG" .. shall I change the new one here to lower case and then follow up with a PR to sync the others to lower case?

jonasjelonek:
A quick sweep over existing recipes shows that all I've seen use full lowercase, so I would stick to that. On other targets, this is the same. Also because the DTS name isn't set there per device, rather just assembled automatically based on other variables.

So I would prefer to make that lowercase, and fix those other SDG occurences either here as another commit or as a separate PR.

openwrt-ai (commented):

---

jonasjelonek:

PR #24680, opened 2026-08-11
uhttpd: update to Git HEAD (2026-08-03)

### LINKED ISSUE #3: Incorrect chunk size parsing behavior
uhttpd has some overly-permissive behavior when parsing HTTP chunk sizes. Due to the minor risk of request smuggling attacks when uhttpd is deployed behind a reverse proxy with corresponding parsing issues, these should probably be fixed.

The RFC specifies that chunked sizes must consist of one or more hex digits. Further, it specifies that the final chunk must consist of one or more `0`s. uhttpd doesn't enforce either of these rules.

For example, the following requests are erroneously accepted by uhttpd:
```
GET / HTTP/1.1\r\n
Host: whatever\r\n
Transfer-Encoding: chunked\r\n
\r\n
0_0\r\n
\r\n
```
```
GET / HTTP/1.1\r\n
Host: whatever\r\n
Transfer-Encoding: chunked\r\n
\r\n
\r0\r\n
\r\n
```
```
GET / HTTP/1.1\r\n
Host: whatever\r\n
Transfer-Encoding: chunked\r\n
\r\n
0x1\r\n
Z\r\n
0\r\n
\r\n
```

### LINKED ISSUE #5: uhttpd incorrectly accepts empty header names.
uhttpd accepts requests with empty header names. These are invalid according to the RFCs.

For example, uhttpd will accept the following request:
```
GET / HTTP/1.1\r\n
:\r\n
\r\n
```

### LINKED ISSUE #10: uhttpd incorrectly terminates chunked message bodies on `\r\n\r\n` instead of `0\r\n\r\n`.
When uhttpd receives a request with a chunked message body that is missing the final chunk size of 0, it treats the request as valid even though it should be rejected.

For example, uhttpd incorrectly accepts the following request:
```
POST / HTTP/1.1\r\n
Host: whatever\r\n
Transfer-Encoding: chunked\r\n
\r\n
1\r\n
Z\r\n
\r\n
\r\n
```

### LINKED ISSUE #11: uhttpd doesn't strip spaces and tabs from the ends of header values
The HTTP RFCs specify that spaces and tabs on the beginnings and ends of header values are to be stripped. uhttpd correctly handles this for the beginnings of header values, but not for the ends.

For example, if you send the following request to uhttpd:
```
GET / HTTP/1.1\r\n
Host: whatever\r\n
Test: \t abc \t \r\n
\r\n
```
...it sees a `Test` header value of `abc \t `, but it should see `abc`.

### LINKED ISSUE #14: Dereference symlinks when determining file type
It would be good if the code that [determines filetypes](https://github.com/openwrt/uhttpd/blob/34a8a74dbdec3c0de38abc1b08f6a73c51263792/file.c#L281-L300) dereferenced symlinks.

As it stands now, if you create a symlink (`ln -s symlink real.html`), the symlink will be served as an `application/octet-stream` and not be displayed by the browser.

### LINKED ISSUE #17: Add method to use buffered IO (mod-lua)
Right now uhttpd.send() just calls write() to stdout, thus potentially doing lots of small writes.
Using buffered io for prometheus-node-exporter-lua reduces the netclass collector scrape time from 230ms to 170ms
https://github.com/openwrt/packages/pull/26860/files#diff-bfae82b02aa972d6c7a6848c12c61fe5af113318ddaacde9db8d89692112fe4bR90

Is there a reason not to use buffered io I'm missing ?
Should uhttpd.send() use buffered io by default ?

hauke:
https://github.com/openwrt/uhttpd/commit/f6c2fcfa539d client: cap cumulative header count and bytes per request
https://github.com/openwrt/uhttpd/commit/4f307a2156f3 client: verify chunk data is followed by CRLF before resyncing
https://github.com/openwrt/uhttpd/commit/628a073fef44 listen: set IP_FREEBIND so binding to a not-yet-present address works
https://github.com/openwrt/uhttpd/commit/0910904b75a3 client: reject empty chunk size lines in chunked transfer parsing
https://github.com/openwrt/uhttpd/commit/893fed8577d7 client: use strict chunk size parsing to fix issue #3
https://github.com/openwrt/uhttpd/commit/8867514f51fa client: fix parse_chunksize() return type and stale comment
https://github.com/openwrt/uhttpd/commit/daa5078b959b client: reject invalid header names to prevent request smuggling
https://github.com/openwrt/uhttpd/commit/69f9cabd0d7d http: use chartypes table consistently, drop ctype.h
https://github.com/openwrt/uhttpd/commit/003cd188b8f9 http: further consolidate character classification and header parsing
https://github.com/openwrt/uhttpd/commit/f682ccabee9e client: parse and discard chunked trailer sections, use enum chunked_state
https://github.com/openwrt/uhttpd/commit/e76736e5676f ubus: guard params type in uh_ubus_send_request()
https://github.com/openwrt/uhttpd/commit/42e30caa704e ubus: set d->free unconditionally to fix url_path leak
https://github.com/openwrt/uhttpd/commit/d6205aad6176 client: move unread-body guard into uh_request_done()
https://github.com/openwrt/uhttpd/commit/3c48b9e17d20 ubus: release client refcount on successful /ubus/subscribe
https://github.com/openwrt/uhttpd/commit/d0b89eb7c70b utils: strip trailing whitespace from header field-values
https://github.com/openwrt/uhttpd/commit/067466662368 file: dereference symlinks when looking up MIME type
https://github.com/openwrt/uhttpd/commit/c54ba119cf54 lua, ucode: use buffered stdio for script output
https://github.com/openwrt/uhttpd/commit/60f64bec40c8 tls: add SNI-based redirect support

Fixes: https://github.com/openwrt/uhttpd/issues/3
Fixes: https://github.com/openwrt/uhttpd/issues/5
Fixes: https://github.com/openwrt/uhttpd/issues/10
Fixes: https://github.com/openwrt/uhttpd/issues/11
Fixes: https://github.com/openwrt/uhttpd/issues/14
Fixes: https://github.com/openwrt/uhttpd/issues/17

Changes: https://github.com/openwrt/uhttpd/compare/7b1bec45826bd78c8afc993435bdc0f1df2fe399...60f64bec40c8113cf09815ec377761b1f4f95f22


package/network/services/uhttpd/Makefile
```diff
@@ -9,9 +9,9 @@ PKG_RELEASE:=1
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL=$(PROJECT_GIT)/project/uhttpd.git
-PKG_SOURCE_DATE:=2026-06-16
-PKG_SOURCE_VERSION:=7b1bec45826bd78c8afc993435bdc0f1df2fe399
-PKG_MIRROR_HASH:=ee52821b04cc78f46875e9870b6ed41670606d6b90fb31b31312666c4ed809dd
+PKG_SOURCE_DATE:=2026-08-03
+PKG_SOURCE_VERSION:=60f64bec40c8113cf09815ec377761b1f4f95f22
```
openwrt-ai, line 13: nit: four of the 18 commits in this range fix published upstream security advisories, but the commit message gives no hint of that — it only carries `Fixes:` links for issues 3/5/10/11/14/17. The advisory-fixing commits and their trailers:

- `e76736e5676f ubus: guard params type in uh_ubus_send_request()` — `Fixes: GHSA-2mpg-6wp5-435p`
- `42e30caa704e ubus: set d->free unconditionally to fix url_path leak` — `Fixes: GHSA-83vv-qrc6-h3hx`
- `d6205aad6176 client: move unread-body guard into uh_request_done()` — `Fixes: GHSA-c2wg-hcff-hqrm`
- `3c48b9e17d20 ubus: release client refcount on successful /ubus/subscribe` — `Fixes: GHSA-wvgh-cm54-q6f6`

The previous uhttpd bump, [ecbb5f6e9f6f "uhttpd: update to Git HEAD (2026-06-16)"](https://github.com/openwrt/openwrt/commit/ecbb5f6e9f6ffe871aa63dc8b1548278ffc5d385), listed its advisories as `Link: https://github.com/openwrt/uhttpd/security/advisories/GHSA-...` trailers. Adding the same here makes it visible that this bump is security-relevant when the stable branches are triaged.

Also in the same message: `60f64bec40c8` carries `Closes: #19`, but issue 19 is missing from the `Fixes:` list.

---

package/network/services/uhttpd/Makefile
```diff
@@ -9,9 +9,9 @@ PKG_RELEASE:=1
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL=$(PROJECT_GIT)/project/uhttpd.git
-PKG_SOURCE_DATE:=2026-06-16
-PKG_SOURCE_VERSION:=7b1bec45826bd78c8afc993435bdc0f1df2fe399
-PKG_MIRROR_HASH:=ee52821b04cc78f46875e9870b6ed41670606d6b90fb31b31312666c4ed809dd
+PKG_SOURCE_DATE:=2026-08-03
```
openwrt-ai, line 12: `60f64bec40c8 tls: add SNI-based redirect support` adds a new repeatable command-line option `-Q host=url`, but nothing in this package exposes it: [`append_bool "$cfg" redirect_https "-q" 0` at uhttpd.init:215](https://github.com/openwrt/openwrt/blob/7d5401cbb0e3bd659cfba87968c099e1d5b5479c/package/network/services/uhttpd/files/uhttpd.init#L215) is the only TLS-redirect mapping, and `files/uhttpd.config` has no matching option either. Is leaving the UCI plumbing to a follow-up intentional, or should the init script gain a list-style mapping here — the shape used for [`ucode_prefix` at uhttpd.init:159](https://github.com/openwrt/openwrt/blob/7d5401cbb0e3bd659cfba87968c099e1d5b5479c/package/network/services/uhttpd/files/uhttpd.init#L159) would fit a repeatable `host=url` option?

---

openwrt-ai (commented):
Reviewed 1 new commit. Verified against the upstream repository: `60f64bec40c8113cf09815ec377761b1f4f95f22` is current `refs/heads/master` of `openwrt/uhttpd`, its commit date is 2026-08-03 (matches `PKG_SOURCE_DATE`), and the 18 commit lines in the message match `7b1bec45..60f64bec` exactly, in order, with correct subjects. No `CMakeLists.txt` change upstream in that range, so no new build dependencies; `PKG_RELEASE` correctly stays at 1 for a source-version bump. Two nits inline.

---

PR #24682, opened 2026-08-11
ramips: mt7621: add TL-R605 v1 as alternative name for TP-Link ER605 v1

kolavar:
This was the original name for the TP-Link ER605 v1 before rebranding.
Hardware is identical. Adds DEVICE_ALT0 entries for the original branding.

### DIFF
--- a/target/linux/ramips/image/mt7621.mk
+++ b/target/linux/ramips/image/mt7621.mk
@@ -3129,6 +3129,9 @@ define Device/tplink_er605-v1
   DEVICE_VENDOR := TP-Link
   DEVICE_MODEL := ER605
   DEVICE_VARIANT := v1
+  DEVICE_ALT0_VENDOR := TP-Link
+  DEVICE_ALT0_MODEL := TL-R605
+  DEVICE_ALT0_VARIANT := v1
   DEVICE_PACKAGES := -wpad-basic-mbedtls -uboot-envtools
   IMAGE_SIZE := 13760k
 endef


target/linux/ramips/image/mt7621.mk
```diff
@@ -3129,6 +3129,9 @@ define Device/tplink_er605-v1
   DEVICE_VENDOR := TP-Link
   DEVICE_MODEL := ER605
   DEVICE_VARIANT := v1
+  DEVICE_ALT0_VENDOR := TP-Link
+  DEVICE_ALT0_MODEL := TL-R605
```
openwrt-ai, line 3133: nit: sister-device parity question — the block immediately below, [`Device/tplink_er605-v2`](https://github.com/openwrt/openwrt/blob/8ea25a9b6060ea6048d09a613e640ac5f4d8a2cd/target/linux/ramips/image/mt7621.mk#L3139-L3148), already carries a rebrand alias (`DEVICE_ALT0_MODEL := FR205`, `DEVICE_ALT0_VARIANT := v1`). If TL-R605 is the alternative branding for the ER605 line as a whole, does a TL-R605 v2 exist that should be added there as `DEVICE_ALT1_*`? If the TL-R605 name only ever applied to the v1 hardware, this is fine as-is — just want to make sure the omission is deliberate rather than an oversight.

---

openwrt-ai (commented):
## Commit checks

- `8ea25a9` "ramips: mt7621: add support for TP-Link TL-R605 v1" — the subject claims new device support, but the diff adds no DTS, no `define Device/...` block and no `TARGET_DEVICES` entry; it only attaches `DEVICE_ALT0_*` branding metadata to the already-supported `tplink_er605-v1`. `DEVICE_ALT*_*` is consumed purely for title generation in [`include/image.mk:439`](https://github.com/openwrt/openwrt/blob/8ea25a9b6060ea6048d09a613e640ac5f4d8a2cd/include/image.mk#L439) and the profiles JSON, so nothing about device support changes. Consider a subject that matches the diff, e.g. `ramips: mt7621: add TL-R605 v1 as alternative name for TP-Link ER605 v1`.

Otherwise the change looks correct: the `DEVICE_ALT0_VENDOR`/`MODEL`/`VARIANT` ordering after `DEVICE_VARIANT` matches the sibling `tplink_er605-v2` block and other ramips entries, and since this is an alias-only addition the usual new-device hardware/flashing/MAC sections are not expected.

---

target/linux/ramips/image/mt7621.mk
kolavar: I cannot definitively answer that question. TP-Link offers firmware downloads for "TL-R605 v2" which is suggestive that TL-R605 v2 does exist. However, I have never seen one in commerce, suggesting the rebranding took place before the v2 variant was shipped. If it does exist, it would be an alternate name for the ER605 v2, which is outside the scope of this pull request.

kolavar:
Reworded subject and rebased.

jonasjelonek:

PR #24683, opened 2026-08-12
dnsmasq: fix where startup would fail in the absence of DHCP

findlayfeng:
The `dhcp-name-match` option in the `dhcpbogushostname.conf` configuration file prevents the software from starting if DHCP support is missing.

### DIFF
--- a/package/network/services/dnsmasq/files/dnsmasq.init
+++ b/package/network/services/dnsmasq/files/dnsmasq.init
@@ -1194,7 +1194,7 @@ dnsmasq_start()
 	echo >> "$CONFIGFILE_TMP"
 
 	config_get_bool dhcpbogushostname "$cfg" dhcpbogushostname 1
-	[ "$dhcpbogushostname" -gt 0 ] && {
+	[ "$dhcpbogushostname" -gt 0 ] && ! dnsmasq_ignore_opt dhcp-name-match && {
 		xappend "--dhcp-ignore-names=tag:dhcp_bogus_hostname"
 		[ -r "$DHCPBOGUSHOSTNAMEFILE" ] && xappend "--conf-file=$DHCPBOGUSHOSTNAMEFILE"
 	}
--- a/package/network/services/dnsmasq/Makefile
+++ b/package/network/services/dnsmasq/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=dnsmasq
 PKG_UPSTREAM_VERSION:=2.93
 PKG_VERSION:=$(subst test,~~test,$(subst rc,~rc,$(PKG_UPSTREAM_VERSION)))
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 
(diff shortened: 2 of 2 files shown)


package/network/services/dnsmasq/files/dnsmasq.init
```diff
@@ -1194,7 +1194,7 @@ dnsmasq_start()
 	echo >> "$CONFIGFILE_TMP"
 
 	config_get_bool dhcpbogushostname "$cfg" dhcpbogushostname 1
-	[ "$dhcpbogushostname" -gt 0 ] && {
+	[ "$dhcpbogushostname" -gt 0 ] && ! dnsmasq_ignore_opt dhcp-name-match && {
```
openwrt-ai, line 1197: nit: unrelated to this line, but the PR description still contains the submission-template boilerplate ("Thanks for your contribution to OpenWrt! ... Please remove this message before posting the pull request."). Worth trimming the description down to just the first paragraph, which already reads as a good commit body.

---

openwrt-ai (commented):
Reviewed 1 new commit; no functional issues found.

Verified the premise against the dnsmasq sources: `dhcp-name-match` is parsed inside the `#ifdef HAVE_DHCP` block in `src/option.c` ([option table entry](https://github.com/dnsmasq/dnsmasq/blob/8d8ad76d896f8eff3bd59dcdb207062c7466a49f/src/option.c#L330), handler at [`case LOPT_NAME_MATCH`](https://github.com/dnsmasq/dnsmasq/blob/8d8ad76d896f8eff3bd59dcdb207062c7466a49f/src/option.c#L4353)), so with `-DNO_DHCP` (i.e. `dnsmasq-full` with `PACKAGE_dnsmasq_full_dhcp=n`) it falls through to the [`unsupported option ...`](https://github.com/dnsmasq/dnsmasq/blob/8d8ad76d896f8eff3bd59dcdb207062c7466a49f/src/option.c#L5496) fatal error — pulling in `dhcpbogushostname.conf` really does abort startup. The `--dhcp-ignore-names=` line on the next line was already filtered by [`dnsmasq_ignore_opt`'s `dhcp-*` case](https://github.com/openwrt/openwrt/blob/f0ace37f8d9f6d9ddbd62d26ad1fc3ceea1789f8/package/network/services/dnsmasq/files/dnsmasq.init#L44-L47) inside `xappend`, so the new guard's only behavioural effect is suppressing the `--conf-file=` include, which is the right scope. The approach also mirrors the existing DNSSEC handling, where [`--conf-file=$TRUSTANCHORSFILE`](https://github.com/openwrt/openwrt/blob/f0ace37f8d9f6d9ddbd62d26ad1fc3ceea1789f8/package/network/services/dnsmasq/files/dnsmasq.init#L1143-L1146) is likewise tied to feature availability.

Also checked that no other feature-conditional include remains broken in a `NO_DHCP` build: `rfc6761.conf` is DNS-only, and `--enable-ra`/`--quiet-ra` (not covered by `dnsmasq_ignore_opt`'s patterns) are only emitted when `DNSMASQ_DHCP_VER -gt 4`, which cannot happen without DHCPv6 support. `PKG_RELEASE` bump is present.

---

PR #24689, opened 2026-08-12
ramips: mt7621: add support for Cudy LT700 Outdoor v1

carlosz1986:
These commits add support for the cudy LT700 outdoor poe version ...

The cudy team was so nice to send me a intermediate signed firmware quick. It can be downloaded here:

https://drive.google.com/file/d/1u3w1LsK3sPd2raSTqnkPso5a5VTc3pzx/view?usp=drive_link

The kernel changes I will send upstream as well and I will add the lore link here.
edit: https://lore.kernel.org/all/20260812104228.802711-1-github@szelinsky.de/

Includes three comgt ncm fixes the EG060W needs: CR stripping on teardown (disconnect was never sent), a linksettle wait so DHCP doesn't race the modem's connect flap, and per-model ncm.json selection so the EG060W uses AT+CGDCONT while EC25/EG25 keep AT+QICSGP.

On linksettle (because I expect questions here): I defaulted it to 3s because the EG060W's post-connect carrier flap changes the PDP address, and DHCP started before it settles keeps a lease the network no longer routes - netifd removes the dynamic interface on carrier loss, so nothing recovers it. directip.sh already waits for carrier for a similar firmware quirk and qmi.sh sleeps unconditionally while waiting for registration, so a wait here seemed in keeping, but it does cost up to 3s on every NCM bring-up for modems that don't need it. 
Would you prefer the default to be 0 (opt-in per device), or is 3s acceptable as a default? Any other ideas? 

Driving the signal strength[ LEDs is covered here... 
](https://github.com/openwrt/openwrt/pull/24845)

Technically everything works, but I will do the next days more tests on production.

### DIFF
--- a/package/network/utils/comgt/files/ncm.sh
+++ b/package/network/utils/comgt/files/ncm.sh
@@ -263,7 +263,7 @@ proto_ncm_teardown() {
 	json_get_vars manufacturer
 	[ $? -ne 0 -o -z "$manufacturer" ] && {
 		# Fallback to direct detect, for proper handle device replug.
-		manufacturer=$(gcom -d "$device" -s /etc/gcom/getcardinfo.gcom | awk 'NF && $0 !~ /AT\+CGMI/ { sub(/\+CGMI: /,""); print tolower($1); exit; }')
+		manufacturer=$(gcom -d "$device" -s /etc/gcom/getcardinfo.gcom | awk -v RS='\r?\n' 'NF && $0 !~ /AT\+CGMI/ { sub(/\+CGMI: /,""); print tolower($1); exit; }')
 		[ $? -ne 0 -o -z "$manufacturer" ] && {
 			echo "Failed to get modem information"
 			proto_notify_error "$interface" GETINFO_FAILED
--- a/target/linux/generic/pending-6.18/781-usb-serial-option-add-Quectel-EG060W.patch
+++ b/target/linux/generic/pending-6.18/781-usb-serial-option-add-Quectel-EG060W.patch
@@ -1,0 +1,57 @@
+From d88de38ea3d41a5b22e93bfcab2bf42ceaf34931 Mon Sep 17 00:00:00 2001
+From: Carlo Szelinsky <github@szelinsky.de>
+Date: Wed, 12 Aug 2026 13:24:00 +0300
+Subject: [PATCH] USB: serial: option: add Quectel EG060W
+
+Add Quectel EG060W with product ID 0x6004.
+The three vendor-specific interfaces are DIAG, NMEA and AT; the CDC-NCM
+data interfaces are handled by the cdc_ncm driver.
+
+T:  Bus=01 Lev=01 Prnt=01 Port=01 Cnt=01 Dev#=  2 Spd=480  MxCh= 0
+D:  Ver= 2.10 Cls=ef(misc ) Sub=02 Prot=01 MxPS=64 #Cfgs=  1
+P:  Vendor=2c7c ProdID=6004 Rev= 3.18
+S:  Manufacturer=Quectel
+S:  Product=EG060W-EAAA
+S:  SerialNumber=123456789ABCD
+C:* #Ifs= 5 Cfg#= 1 Atr=e0 MxPwr=  2mA
+A:  FirstIf#= 0 IfCount= 2 Cls=02(comm.) Sub=0d Prot=00
+I:* If#= 0 Alt= 0 #EPs= 1 Cls=02(comm.) Sub=0d Prot=00 Driver=cdc_ncm
+E:  Ad=82(I) Atr=03(Int.) MxPS=  16 Ivl=32ms
+I:  If#= 1 Alt= 0 #EPs= 0 Cls=0a(data ) Sub=00 Prot=01 Driver=cdc_ncm
+I:* If#= 1 Alt= 1 #EPs= 2 Cls=0a(data ) Sub=00 Prot=01 Driver=cdc_ncm
+E:  Ad=81(I) Atr=02(Bulk) MxPS= 512 Ivl=0ms
+E:  Ad=01(O) Atr=02(Bulk) MxPS= 512 Ivl=0ms
+I:* If#= 2 Alt= 0 #EPs= 2 Cls=ff(vend.) Sub=00 Prot=00 Driver=option
+E:  Ad=83(I) Atr=02(Bulk) MxPS= 512 Ivl=0ms
+E:  Ad=02(O) Atr=02(Bulk) MxPS= 512 Ivl=0ms
+I:* If#= 3 Alt= 0 #EPs= 2 Cls=ff(vend.) Sub=00 Prot=00 Driver=option
+E:  Ad=84(I) Atr=02(Bulk) MxPS= 512 Ivl=0ms
+E:  Ad=03(O) Atr=02(Bulk) MxPS= 512 Ivl=0ms
+I:* If#= 4 Alt= 0 #EPs= 2 Cls=ff(vend.) Sub=00 Prot=00 Driver=option
+E:  Ad=85(I) Atr=02(Bulk) MxPS= 512 Ivl=0ms
+E:  Ad=04(O) Atr=02(Bulk) MxPS= 512 Ivl=0ms
+
+Cc: stable@vger.kernel.org
+Signed-off-by: Carlo Szelinsky <github@szelinsky.de>
+---
+ drivers/usb/serial/option.c | 2 ++
+ 1 file changed, 2 insertions(+)
+
+--- a/drivers/usb/serial/option.c
++++ b/drivers/usb/serial/option.c
+@@ -281,6 +281,7 @@ static void option_instat_callback(struc
+ #define QUECTEL_PRODUCT_EC200U			0x0901
+ #define QUECTEL_PRODUCT_EG912Y			0x6001
+ #define QUECTEL_PRODUCT_EC200S_CN		0x6002
++#define QUECTEL_PRODUCT_EG060W			0x6004
+ #define QUECTEL_PRODUCT_EC200A			0x6005
+ #define QUECTEL_PRODUCT_EG916Q			0x6007
+ #define QUECTEL_PRODUCT_EM061K_LWW		0x6008
+@@ -1270,6 +1271,7 @@ static const struct usb_device_id option
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EC200U, 0xff, 0, 0) },
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EC200S_CN, 0xff, 0, 0) },
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EC200T, 0xff, 0, 0) },
++	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EG060W, 0xff, 0, 0) },
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EG912Y, 0xff, 0, 0) },
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EG916Q, 0xff, 0x00, 0x00) },
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_RM500K, 0xff, 0x00, 0x00) },
--- a/package/network/utils/comgt/Makefile
+++ b/package/network/utils/comgt/Makefile
@@ -9,7 +9,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=comgt
 PKG_VERSION:=0.32
-PKG_RELEASE:=36
+PKG_RELEASE:=37
 
 PKG_SOURCE:=$(PKG_NAME).$(PKG_VERSION).tgz
 PKG_SOURCE_URL:=@SF/comgt
--- a/target/linux/generic/hack-6.18/780-usb-net-MeigLink_modem_support.patch
+++ b/target/linux/generic/hack-6.18/780-usb-net-MeigLink_modem_support.patch
@@ -44,7 +44,7 @@ Subject: [PATCH] net/usb/qmi_wwan: add MeigLink modem support
  
  #define QUECTEL_VENDOR_ID			0x2c7c
  /* These Quectel products use Quectel's vendor ID */
-@@ -1159,6 +1164,11 @@ static const struct usb_device_id option
+@@ -1160,6 +1165,11 @@ static const struct usb_device_id option
  	{ USB_DEVICE(QUALCOMM_VENDOR_ID, 0x0023)}, /* ONYX 3G device */
  	{ USB_DEVICE(QUALCOMM_VENDOR_ID, 0x9000), /* SIMCom SIM5218 */
  	  .driver_info = NCTRL(0) | NCTRL(1) | NCTRL(2) | NCTRL(3) | RSVD(4) },
@@ -56,7 +56,7 @@ Subject: [PATCH] net/usb/qmi_wwan: add MeigLink modem support
  	/* Quectel products using Qualcomm vendor ID */
  	{ USB_DEVICE(QUALCOMM_VENDOR_ID, QUECTEL_PRODUCT_UC15)},
  	{ USB_DEVICE(QUALCOMM_VENDOR_ID, QUECTEL_PRODUCT_UC20),
-@@ -1200,6 +1210,11 @@ static const struct usb_device_id option
+@@ -1201,6 +1211,11 @@ static const struct usb_device_id option
  	  .driver_info = ZLP },
  	{ USB_DEVICE(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_BG96),
  	  .driver_info = RSVD(4) },


target/linux/generic/backport-6.18/782-v7.3-USB-serial-option-add-Quectel-EG060W.patch
```diff
@@ -0,0 +1,57 @@
...
+ drivers/usb/serial/option.c | 2 ++
+ 1 file changed, 2 insertions(+)
+
+--- a/drivers/usb/serial/option.c
++++ b/drivers/usb/serial/option.c
+@@ -281,6 +281,7 @@ static void option_instat_callback(struc
+ #define QUECTEL_PRODUCT_EC200U			0x0901
+ #define QUECTEL_PRODUCT_EG912Y			0x6001
+ #define QUECTEL_PRODUCT_EC200S_CN		0x6002
++#define QUECTEL_PRODUCT_EG060W			0x6004
```
openwrt-ai, line 47: This added line shifts every later line of `drivers/usb/serial/option.c` by one, which invalidates the hunk headers of the patch applied after it. All `Check Kernel patches` jobs for 6.18 targets fail because of it ([ath79/generic job](https://github.com/openwrt/openwrt/actions/runs/31589700839/job/94100271009), and identically for mediatek/filogic, starfive, rockchip, mvebu, sunxi, zynq, sifiveu, mpc85xx):

```
target/linux/generic/hack-6.18/780-usb-net-MeigLink_modem_support.patch
Kernel patches for ath79/generic require refresh. (run 'make target/linux/refresh' and force push this pr)
```

[hack-6.18/780's option.c hunks at old lines 1159 and 1200](``https://github.com/openwrt/openwrt/blob/1f6105eaff81b547522f9719010de166dba7f409/target/linux/generic/hack-6.18/780-usb-net-MeigLink_modem_support.patch#L47-L59)`` sit below the `#define` inserted here, so they now start one line later. Please regenerate with `make target/linux/refresh` (ramips/mt7621 selected) and include the updated `hack-6.18/780-usb-net-MeigLink_modem_support.patch` in this series.

---

target/linux/generic/backport-6.18/782-v7.3-USB-serial-option-add-Quectel-EG060W.patch
```diff
@@ -0,0 +1,57 @@
...
+ #define QUECTEL_PRODUCT_EC200A			0x6005
+ #define QUECTEL_PRODUCT_EG916Q			0x6007
+ #define QUECTEL_PRODUCT_EM061K_LWW		0x6008
+@@ -1269,6 +1270,7 @@ static const struct usb_device_id option
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EC200A, 0xff, 0, 0) },
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EC200U, 0xff, 0, 0) },
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EC200S_CN, 0xff, 0, 0) },
++	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EG060W, 0xff, 0, 0) },
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EC200T, 0xff, 0, 0) },
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EG912Y, 0xff, 0, 0) },
```
openwrt-ai, line 57: nit: the `EG060W` entry is inserted in the middle of the `EC200*` group; the surrounding block groups the EC200 IDs first and the EG* IDs after. Moving it one line down keeps the grouping (and the hunk header stays valid, the hunk still covers the same six context lines). Worth fixing for the upstream v2 too.

```suggestion
 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EC200S_CN, 0xff, 0, 0) },
 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EC200T, 0xff, 0, 0) },
+	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EG060W, 0xff, 0, 0) },
 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EG912Y, 0xff, 0, 0) },
```

---

target/linux/ramips/dts/mt7621_cudy_lt700-outdoor-v1.dts
```diff
@@ -0,0 +1,239 @@
...
+
+		signal2 {
+			gpios = <&gpio 45 GPIO_ACTIVE_LOW>;
+			label = "green:signal2";
+		};
+
+		signal3 {
+			gpios = <&gpio 44 GPIO_ACTIVE_LOW>;
+			label = "green:signal3";
+		};
```
openwrt-ai, line 89: These three LEDs still use the deprecated `label = "<color>:<function>"` syntax, while every other LED in this file already uses `color`/`function`. There is no `LED_FUNCTION_SIGNAL` constant, so use the plain string plus `function-enumerator`, as done in [mt7628an_zbtlink_zbt-we2426-b.dts](https://github.com/openwrt/openwrt/blob/1f6105eaff81b547522f9719010de166dba7f409/target/linux/ramips/dts/mt7628an_zbtlink_zbt-we2426-b.dts#L34-L39) and [qca9563_dlink_dap-1720-a1.dts](https://github.com/openwrt/openwrt/blob/1f6105eaff81b547522f9719010de166dba7f409/target/linux/ath79/dts/qca9563_dlink_dap-1720-a1.dts#L57-L62). Nothing in `01_leds` references these names, so the rename from `green:signalN` to `green:signal-N` is harmless.

```suggestion
		signal1 {
			gpios = <&gpio 42 GPIO_ACTIVE_LOW>;
			color = <LED_COLOR_ID_GREEN>;
			function = "signal";
			function-enumerator = <1>;
		};

		signal2 {
			gpios = <&gpio 45 GPIO_ACTIVE_LOW>;
			color = <LED_COLOR_ID_GREEN>;
			function = "signal";
			function-enumerator = <2>;
		};

		signal3 {
			gpios = <&gpio 44 GPIO_ACTIVE_LOW>;
			color = <LED_COLOR_ID_GREEN>;
			function = "signal";
			function-enumerator = <3>;
		};
```

---

target/linux/ramips/mt7621/base-files/etc/board.d/02_network
```diff
@@ -133,6 +133,11 @@ ramips_setup_interfaces()
 	mikrotik,routerboard-750gr3)
 		ucidef_set_interfaces_lan_wan "lan2 lan3 lan4 lan5" "wan"
 		;;
+	cudy,lt700-outdoor-v1)
+		ucidef_set_interface_lan "lan"
+		ucidef_set_interface "wwan" device "/dev/ttyUSB2" protocol "ncm" auto 0
```
openwrt-ai: `auto 0` is silently dropped and never reaches `/etc/config/network`. `ucidef_set_interface` stores it in `/etc/board.json`, but [`generate_network()` in config_generate](https://github.com/openwrt/openwrt/blob/1f6105eaff81b547522f9719010de166dba7f409/package/base-files/files/bin/config_generate#L94-L104) only reads `device macaddr metric protocol ipaddr netmask vlan` plus `ports`, and the `ncm` branch further down only emits `proto` and `pdptype`. So the generated `wwan` interface will come up automatically regardless.

If the intent is to keep the modem down until the user configures an APN, it needs an explicit `uci set network.wwan.auto='0'` next to the `add_list` below; otherwise drop the `auto 0` so the line doesn't suggest behaviour that isn't there. Note the sibling Cudy LTE boards (`cudy,lt300-v3`, `cudy,lt400e-v1`, `cudy,lt500-outdoor-v1` in `mt76x8/base-files/etc/board.d/02_network`) don't set it either.

---

target/linux/ramips/dts/mt7621_cudy_lt700-outdoor-v1.dts
```diff
@@ -0,0 +1,239 @@
...
+			gpio-export,name = "pwr";
+			gpio-export,output = <1>;
+			gpios = <&gpio 41 GPIO_ACTIVE_LOW>;
+		};
+
+		usb4g {
+			gpio-export,name = "4g";
+			gpio-export,output = <0>;
+			gpios = <&gpio 46 GPIO_ACTIVE_HIGH>;
+		};
```
openwrt-ai, line 116: Is the modem really meant to be left de-asserted at boot? `gpio-export,output = <0>` with `GPIO_ACTIVE_HIGH` drives GPIO 46 to its inactive state, whereas the neighbouring `pwr` export asserts its line (`output = <1>` on an active-low GPIO). If GPIO 46 is the enable/power line for the internal EG060W, the modem would be off after a fresh boot and no `/dev/ttyUSB*` would appear — which would also explain why `02_network` references `/dev/ttyUSB2`. If it is a reset-n or a "disable" line instead, a short comment in the DTS would help, since neither the node name (`usb4g`) nor the exported name (`4g`) says which polarity means "on".

---

target/linux/ramips/image/mt7621.mk
```diff
@@ -783,6 +783,20 @@ define Device/cudy_c200p
 endef
 TARGET_DEVICES += cudy_c200p
 
+define Device/cudy_lt700-outdoor-v1
```
openwrt-ai, line 786: nit: the PR description calls this "the cudy LT700 outdoor poe version", but the commit message's hardware section has no power entry at all (no PoE note, no voltage/current), and the serial console line gives only the baud rate, not the header location/pinout. Both are part of the expected hardware specification for a new-device commit — please add them so the two descriptions agree.

---

openwrt-ai (commented):
Reviewed 2 commits (fresh review). Both commit messages match their diffs; comments are inline.

Verified alongside the review: the new pending patch applies cleanly to linux 6.18.41 (`target/linux/generic/kernel-6.18`) with no fuzz, the pinctrl `gpio` group list (`i2c`, `uart3`, `uart2`, `jtag`, `sdhci`) covers exactly GPIOs 3/7/8/10/12/15/41-46 used by the DTS, `IMAGE_SIZE` matches the `firmware` partition, and the MAC layout in the DTS matches the table in the commit message.

---

target/linux/generic/pending-6.18/781-usb-serial-option-add-Quectel-EG060W.patch
```diff
@@ -0,0 +1,57 @@
...
+ #define QUECTEL_PRODUCT_EG916Q			0x6007
+ #define QUECTEL_PRODUCT_EM061K_LWW		0x6008
+@@ -1269,6 +1270,7 @@ static const struct usb_device_id option
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EC200A, 0xff, 0, 0) },
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EC200U, 0xff, 0, 0) },
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EC200S_CN, 0xff, 0, 0) },
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EC200T, 0xff, 0, 0) },
++	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EG060W, 0xff, 0, 0) },
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EG912Y, 0xff, 0, 0) },
+ 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EG916Q, 0xff, 0x00, 0x00) },
```
openwrt-ai: `Check Kernel patches` still fails on this head, now on this patch itself:

```
target/linux/generic/pending-6.18/781-usb-serial-option-add-Quectel-EG060W.patch
Kernel patches for x86/64 require refresh. (run 'make target/linux/refresh' and force push this pr)
```

([x86/64 job](https://github.com/openwrt/openwrt/actions/runs/31740277394/job/94587562690); identical in apm821xx/nand, mpc85xx/p1010, d1, zynq, omap, ixp4xx and gemini.) The hack-6.18/780 refresh in the other commit is correct — this is a new failure introduced by hand-moving the `+` line here.

The `+` line was moved one line down but the context window was not, leaving 4 lines of leading and 2 of trailing context. `quilt refresh` emits 3/3, so the stored patch no longer round-trips, and applying it against 6.18.44 needs `fuzz 2` (the pre-force-push version applied at fuzz 0):

```
$ patch -p1 --dry-run < 781-usb-serial-option-add-Quectel-EG060W.patch
Hunk #1 succeeded at 281.
Hunk #2 succeeded at 1270 with fuzz 2.
```

`make target/linux/refresh` (ramips/mt7621 selected) produces the hunk below — note the header also becomes `-1270,6 +1271,7`, and `QUECTEL_PRODUCT_RM500K` joins as the third trailing context line. Hunk #1 is unaffected.

```suggestion
@@ -1270,6 +1271,7 @@ static const struct usb_device_id option
 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EC200U, 0xff, 0, 0) },
 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EC200S_CN, 0xff, 0, 0) },
 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EC200T, 0xff, 0, 0) },
+	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EG060W, 0xff, 0, 0) },
 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EG912Y, 0xff, 0, 0) },
 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_EG916Q, 0xff, 0x00, 0x00) },
 	{ USB_DEVICE_AND_INTERFACE_INFO(QUECTEL_VENDOR_ID, QUECTEL_PRODUCT_RM500K, 0xff, 0x00, 0x00) },
```

---

target/linux/ramips/mt7621/base-files/etc/board.d/02_network
```diff
@@ -133,6 +133,12 @@ ramips_setup_interfaces()
 	mikrotik,routerboard-750gr3)
 		ucidef_set_interfaces_lan_wan "lan2 lan3 lan4 lan5" "wan"
 		;;
+	cudy,lt700-outdoor-v1)
+		ucidef_set_interface_lan "lan"
+		ucidef_set_interface "wwan" device "/dev/ttyUSB2" protocol "ncm"
+		uci set network.wwan.auto='0'
```
openwrt-ai: Sorry — my earlier comment pointed you at the wrong place. This `uci set` still doesn't take effect, for two reasons, and it's a no-op rather than a fix:

1. `/etc/board.d/*` runs *before* `/etc/config/network` exists. [`82_config_generate`](https://github.com/openwrt/openwrt/blob/705f9c3ee344d97d38bb0df21b73a0c7a4632ee9/package/base-files/files/lib/preinit/82_config_generate#L2-L12) calls `board_detect` (which sources these scripts) and only then `config_generate`; base-files ships no `/etc/config/network`. So at this point `network.wwan` does not exist and `uci set network.wwan.auto` fails — silently, since `board_detect` sources each script in a plain subshell.
2. Even if it were staged, [`generate_network()` starts with `delete network.$1`](https://github.com/openwrt/openwrt/blob/705f9c3ee344d97d38bb0df21b73a0c7a4632ee9/package/base-files/files/bin/config_generate#L153-L159), which would drop it again.

The neighbouring `uci add_list firewall.@zone[1].network='wwan'` works only because `/etc/config/firewall` is a shipped file and `config_generate` ends with a global `uci commit`. Consistently, no in-tree `board.d` script does `uci set network.*` — that pattern lives in `etc/uci-defaults/`, which `/etc/init.d/boot` runs after `config_generate`.

Either move it to a `target/linux/ramips/mt7621/base-files/etc/uci-defaults/` script guarded on `$(board_name)` (alongside `01_enable_packet_steering`), or drop the line so it doesn't suggest behaviour that isn't there — the sibling Cudy LTE boards in `mt76x8/base-files/etc/board.d/02_network` don't set it either.

---

openwrt-ai (commented):
Reviewed the 2 new commits (force-push of 1f6105e). Both commit messages match their diffs — the added Power/serial-console lines and the `hack-6.18/780` refresh note are accurate, so no commit-message issues.

Resolved since the last review: the LED `label` → `color`/`function` + `function-enumerator` conversion, and the DTS comments documenting the `pwr`/`4g` GPIO polarities.

Two things left, inline. Verified alongside: the `hack-6.18/780` hunk-header shift (+1 on both sides) is correct, and it is unaffected by the 781 hunk-2 fix since its hunks sit above line 1270.

---

openwrt-ai (commented):

Both commit messages still match their diffs. Everything flagged in the previous two reviews is now addressed:

- **`pending-6.18/781` hunk header** — now `@@ -1270,6 +1271,7 @@` with 3 lines of leading and 3 of trailing context (`RM500K` added as the third trailing line). Verified against 6.18.44 (`target/linux/generic/kernel-6.18`): `781` applies at fuzz 0 with no offset, and the refreshed `hack-6.18/780` then applies on top at fuzz 0 as well (pending applies before hack). All three workflow runs on `16c413d` are green, including `Build Kernel` / `Check Kernel patches`, so the refresh complaint is gone.
- **Signal LEDs** — `function = LED_FUNCTION_MOBILE` + `function-enumerator` is the right in-tree pattern for a cellular indicator (cf. `mt7621_asus_4g-ax56.dts`, `mt7621_teltonika_rutm30.dts`); nothing in `01_leds` references the old `green:signalN` names, and the two names it does use (`green:lan`, `green:wan-online`) are unaffected.
- **`uci set network.wwan.auto='0'`** — dropped, which matches the sibling Cudy LTE boards.

Also spot-checked: `0x6004` collides with no existing Quectel product ID (`CMOTECH_PRODUCT_6004` is a different vendor ID), and the new `#define` sits in numeric order between `EC200S_CN` (0x6002) and `EC200A` (0x6005).

---

package/network/utils/comgt/files/ncm.sh
```diff
@@ -178,10 +224,17 @@ proto_ncm_setup() {
 
 	json_get_vars finalize
 
+	[ -n "$linksettle" ] || linksettle=3
+	[ "$linksettle" -gt 0 ] && {
+		ncm_wait_link "$ifname" "$linksettle" 30 || \
+			echo "Link did not settle, continuing anyway"
+	}
```
openwrt-ai, line 234: The wait is inserted after `connect` but before `finalize`, and for the `mtk1` profile `finalize` is what actually enters data state: [`"connect": "AT+CGACT=1,${profile}"` / `"finalize": "AT+CGDATA=\"M-MBIM\",${profile},1"`](https://github.com/openwrt/openwrt/blob/b2479bf33819fb4e64d09735bfd026d3c3420dbb/package/network/utils/comgt/files/ncm.json#L67-L77), and [`finalize` only runs at the very end of `proto_ncm_setup()`](https://github.com/openwrt/openwrt/blob/b2479bf33819fb4e64d09735bfd026d3c3420dbb/package/network/utils/comgt/files/ncm.sh#L277-L283). If the netdev carrier on those modems only rises at `CGDATA`, this loop can never see carrier and every `mtk1` bring-up pays the full hardcoded 30 s before printing "Link did not settle, continuing anyway".

Since the default is `3` (i.e. enabled), this hits every ncm modem, not just the EG060W. Is `mtk1` known to raise carrier already at `CGACT`? If not, the wait probably needs to run after `finalize`, or be skipped when a `finalize` command is set.

---

package/network/utils/comgt/files/ncm.sh
```diff
@@ -26,16 +27,57 @@ proto_ncm_init_config() {
...
+# connection is finalised, and the PDP address can change across that flap.
+# Running DHCP too early yields a lease belonging to the dead session.
+ncm_wait_link() {
+	local ifname="$1" settle="$2" timeout="$3"
+	local stable=0 waited=0 carrier
+
+	/sbin/ip link set dev "$ifname" up 2>/dev/null
+
+	while [ "$waited" -lt "$timeout" ]; do
+		carrier=$(cat "/sys/class/net/$ifname/carrier" 2>/dev/null)
```
openwrt-ai: `/sbin/ip` is not guaranteed to exist here: it is an alternatives symlink owned by `ip-tiny`/`ip-full` ([`ALTERNATIVES:=200:/sbin/ip:/usr/libexec/ip-tiny`](https://github.com/openwrt/openwrt/blob/b2479bf33819fb4e64d09735bfd026d3c3420dbb/package/network/utils/iproute2/Makefile#L36-L42)), neither is in [`DEFAULT_PACKAGES`](https://github.com/openwrt/openwrt/blob/b2479bf33819fb4e64d09735bfd026d3c3420dbb/include/target.mk#L21-L35) nor `DEFAULT_PACKAGES.router`, and [`Package/comgt-ncm` doesn't depend on it](https://github.com/openwrt/openwrt/blob/b2479bf33819fb4e64d09735bfd026d3c3420dbb/package/network/utils/comgt/Makefile#L46-L50). The existing `/sbin/ip` use in this file is inside the optional `mtu` block, so a missing binary was previously harmless; here it is on the default path.

With `2>/dev/null` swallowing the failure, the netdev stays admin-down, `/sys/class/net/$ifname/carrier` returns `-EINVAL` (the sysfs attribute errors unless `netif_running()`), `carrier` comes back empty every iteration, and the loop silently burns all 30 s on every bring-up. Same outcome for drivers that don't export a usable carrier at all.

Either add `+ip-tiny` to `Package/comgt-ncm`'s `DEPENDS`, or bail out of the wait when the carrier attribute cannot be read at all (as opposed to reading `0`), so an unsupported setup degrades to the old behaviour instead of a 30 s stall.

---

package/network/utils/comgt/files/ncm.sh
```diff
@@ -26,16 +27,57 @@ proto_ncm_init_config() {
...
+		if [ "$carrier" = "1" ]; then
+			stable=$((stable + 1))
+			[ "$stable" -ge "$settle" ] && return 0
+		else
+			stable=0
+		fi
+		sleep 1
+		waited=$((waited + 1))
+	done
```
openwrt-ai: nit: `settle` counts samples, not seconds, and the `sleep 1` is skipped on the sample that satisfies the check — so `linksettle 3` returns after 3 reads spanning only 2 s of observed carrier, not the 3 s the option name, the PR description and the commit message all imply. `linksettle 1` returns after a single read with no settling at all. Comparing `stable -ge settle` after the sleep (or seeding `stable=0` and incrementing post-sleep) makes the unit match the name.

Separately, `timeout` is hardcoded to `30` at the call site while `linksettle` is user-configurable, so any `linksettle` >= 30 can never succeed and always falls through to the "did not settle" path. Worth deriving the timeout from `linksettle`, or clamping it.

---

package/network/utils/comgt/files/ncm.sh
```diff
@@ -273,7 +326,7 @@ proto_ncm_teardown() {
 	}
```
openwrt-ai, line 335: The replug fallback re-detects `manufacturer` but not `model`, so on that path `model` stays empty and `ncm_select_modem` falls back to the vendor-level entry — exactly the case this commit exists to distinguish. An EG060W that was replugged would get `quectel`'s `disconnect` instead of `quectel-eg060w`'s.

Today the two disconnect strings happen to be identical, so nothing breaks yet, but any future model entry whose `disconnect` differs from its vendor entry silently sends the wrong command on ifdown. Adding a `getmodel.gcom` call alongside the `getcardinfo.gcom` one here (with the same `[ "$model" = "error" ] && model=""` / `${model%%-*}` handling as setup) keeps the two paths in sync.

---

openwrt-ai (commented):
Reviewed the 3 new commits since `16c413d` (`f5084b1`, `c0c243c`, `6dffc23`); the ramips commit is unchanged apart from the rebase. All three commit messages match their diffs, so no commit-message issues — comments are inline, all on `ncm.sh`.

Verified alongside the review:

- **`f5084b1`** — the claim about the `"zte` key is confirmed by the file itself: `ncm.json` stores that entry as `"\"zte"` (leading quote, no trailing one), which is exactly `$1` of `"ZTE INCORPORATED"` lowercased, so the CR does sit past `$1` there while `"\"marvell\""` and `"\"mikrotik\""` are single-token replies that were being broken by it. Setup already used `RS='\r?\n'`, so the two paths now agree.
- **`c0c243c`** — the netifd reference checks out: [`interface_check_state()` sets `IFC_REMOVE` for dynamic interfaces when `link_state` drops](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/interface.c#L394-L408), and the DHCP sub-interface is only added once in `proto_ncm_setup()`, so the described failure mode is real. On the question in the PR body about the default: the risk isn't the 3 s itself, it's the 30 s fallback — see the two inline comments about the `finalize` ordering and `/sbin/ip`. Both make a non-settling modem cost 30 s per bring-up rather than 3, and both apply to every ncm modem because the default is enabled.
- **`6dffc23`** — `json_is_a` is available at runtime (used in [`uci-defaults.sh`](https://github.com/openwrt/openwrt/blob/b2479bf33819fb4e64d09735bfd026d3c3420dbb/package/base-files/files/lib/functions/uci-defaults.sh#L819)) and its `<field> <type>` argument order is right; `getmodel.gcom` mirrors `getcardinfo.gcom` exactly and is installed in the same `Package/comgt/install` block; `ncm.json` has no alphabetical ordering convention, so the placement of `quectel-eg060w` is fine, and the lowercase `AT+qnetdevctl` matches the neighbouring `quectel` entry rather than deviating from it. `model=${model%%-*}` correctly yields `eg060w` from `EG060W-EAAA`.

All workflow runs on `b2479bf` are green (`Build Kernel`, `Build all core packages`, `Build PR Profile`, FormalityCheck), so nothing CI-grounded to add.

---

package/network/utils/comgt/files/ncm.sh
carlosz1986: Fixed ... the wait is skipped when a profile sets finalize (mtk1 enters data state later, at AT+CGDATA), and an unreadable carrier returns immediately instead of stalling. settle now counts seconds, and the timeout follows linksettle rather than being hardcoded.

carlosz1986:
@jonasjelonek as you are pretty active - would you mind having a look here?
best

jonasjelonek:
I wouldn't call "vanishing without a trace for more than a week" "pretty active" :D ... but I'll take a look soon ;)

carlosz1986:
> I wouldn't call "vanishing without a trace for more than a week" "pretty active" :D ... but I'll take a look soon ;)

haha, no worries take your time ! once in a while everyone deserves a vacation;-)

package/network/utils/comgt/files/ncm.sh
```diff
@@ -178,10 +225,18 @@ proto_ncm_setup() {
 
 	json_get_vars finalize
 
+	# mtk1 only enters data state at finalize, further down, so skip it there
+	[ -n "$linksettle" ] || linksettle=3
+	[ "$linksettle" -gt 0 ] && [ -z "$finalize" ] && {
```
openwrt-ai: nit: the `"mikrotik"` profile also sets [`"finalize": "AT+ZDHCPLEASE=0"`](https://github.com/openwrt/openwrt/blob/1958ca404e2aa015c188eb4558e4b9a1a531cc7a/package/network/utils/comgt/files/ncm.json#L132-L143), which is a lease tweak rather than the command that enters data state — its `connect` (`AT+ZGACT=1`) already does that. So this guard silently drops the settle wait there for an unrelated reason, while the comment above only mentions `mtk1`; intentional?

---

openwrt-ai (commented):
## Commit checks

- `3510649` "comgt: ncm: wait for the data link to settle before configuring it" — the message says `linksettle 0` keeps the previous behaviour, but the diff also skips the wait entirely for any profile that defines `finalize` (`mtk1`, `"mikrotik"`), and derives the timeout as `max(3 * linksettle, 30)` instead of the fixed 30 s. Both are user-visible carve-outs and should be in the message.

---

package/network/utils/comgt/files/ncm.sh
openwrt-ai: solved by skipping the wait entirely when the profile sets `finalize`, thanks.

---

package/network/utils/comgt/files/ncm.sh
openwrt-ai: solved by returning from the wait when the carrier attribute cannot be read at all, so a missing `/sbin/ip` degrades to the old behaviour instead of stalling, thanks.

---

package/network/utils/comgt/files/ncm.sh
openwrt-ai: fixed, thanks

---

package/network/utils/comgt/files/ncm.sh
carlosz1986: ... the skip should usenow a linkatfinalize flag set only on mtk1, so mikrotik keeps the wait.

achimbo:
Tested successfully on Cudy LT700 Outdoor.

- Compiled from source
- Installation and sysupgrade work
- Ethernet works
- LTE modem works

Tested-by: achim <achim@cloudystars.net>

FredApps:
Tested the `comgt` half of this on a second vendor's modem, since the `ncm.sh` rework affects every NCM user in the tree and you only have Quectel hardware to hand.

**Setup:** ASUS 4G-AX56 (ramips/mt7621), Fibocom FG621-EA, OpenWrt 25.12-SNAPSHOT. Built your `ncm.sh`, `getmodel.gcom` and the Makefile install line onto a tree that also carries the plain `fibocom` key from #24974, then installed `comgt`/`comgt-ncm` r39 over r37 on the running device.

**Model detection works on a modem that answers without the prefix.** This one replies to `AT+CGMM` with a bare model string:

```
00000000  41 54 2b 43 47 4d 4d 0d  0d 0a 46 47 36 32 31 2d  |AT+CGMM...FG621-|
00000010  45 41 0d 0a 0d 0a 4f 4b  0d 0a                    |EA....OK..|
```

There is no `+CGMM: ` for your `sub()` to strip, and the awk still yields `fg621-ea` correctly. `gcom -d /dev/ttyUSB0 -s /etc/gcom/getmodel.gcom` returns `FG621-EA`.

**The fallback in `ncm_select_modem()` behaves.** With no `fibocom-fg621-ea` entry present it falls through to the plain `fibocom` object, and the profile's own commands are used — visible in the log as `AT+CGDCONT=1,"IP"` and `AT+GTRNDIS=1,1`. The profile sets no `linkatfinalize`, so the new wait path is exercised rather than skipped.

**Connect and traffic are fine**, DHCP lease obtained normally:

```
netifd: wwan (4768): Connecting modem
netifd: wwan (4768): sending -> AT+GTRNDIS=1,1
netifd: wwan (4768): Setting up usb0
netifd: wwan_4 (4958): udhcpc: lease of 95.197.69.244 obtained from 95.0.0.1
```

5/5 ICMP replies at 1400 bytes and a 274 kB HTTP transfer afterwards.

**Cost of the settle wait**, same unit, ifdown/ifup, address acquired:

| `ncm.sh` | connect |
| --- | --- |
| current r37 | 14 s |
| this PR | 17 s, twice |

Exactly the default `linksettle=3`, so it is the wait doing what it says rather than anything unexpected.

**One thing worth coordinating:** this collides with #24974, which is open and adds the `fibocom` key. Both bump `comgt`'s `PKG_RELEASE` from 36 — yours to 39, mine to 37 — and both add a key to `ncm.json`. Whichever lands second will need a trivial rebase; no objection from me if that is yours, I just did not want it to be a surprise at merge time.

carlosz1986:
* added tested tags...
* rebased
* and added the backport that was now merged: https://git.kernel.org/pub/scm/linux/kernel/git/johan/usb-serial.git/commit/?id=cd243401994880b8b6810c53177bce56eebaece5
*  changed the version of comgt to 37, so one increase... 


pinging @jonasjelonek again kindly for a review :-)

jonasjelonek (approved):
Overall looks good to me, but I'm not the expert regarding this modem stuff. I just recently tried to use one and mostly failed.

carlosz1986:
thx @jonasjelonek!

I ping some other guys for review that have been active on modem stuff: @patrakov , @jlabuz , @Leo-PL , @blocktrron @RavRavioli ... Maybe someone has time :-) 
cheers

patrakov:
> Modems raise carrier as soon as the data call starts and drop it again briefly while the connection is finalised, and the PDP address changes across that flap.

It would be interesting (but optional) to see where the old address comes from. It almost looks like the modem connects based on internally stored information before it is fully configured and needs some extra initialization command to suppress this spurious early connection. Or is it just supplying the previous address?

Having said that, waiting for a stable link is the next best thing that can be done within the limits of the current fire-and-forget architecture of `comgt`. To me, it looks like a very special case of working around the lack of any support for dynamically following disconnections and provider-initiated address changes - something that ModemManager and https://github.com/ddimension/wwand attempt to do.

I have no objections to adding a 3-second stable-link wait to every modem.

openwrt-ai (commented):
Reviewed 5 new commits; no new issues found.

---

FredApps:
> thx @jonasjelonek!
> 
> I ping some other guys for review that have been active on modem stuff: @patrakov , @jlabuz , @Leo-PL , @blocktrron @RavRavioli ... Maybe someone has time :-) cheers

Me too please :)

carlosz1986:
@patrakov measured it, it's neither guess. The address is the LTE default bearer - the network sets it up at attach, before comgt does anything.

I booted three times with the wwan interface disabled so comgt sent literally zero AT commands. Every time:

```
+CGDCONT: 1,"IP","internet.tel.....","10.76.AA.AA"
+CGACT: 1,1
+QNETDEVCTL: 0,1,0,0     <- netdev off
```

Context up with an address while the netdev is still off, and it's a different address on every attach (10.24.BB.BB, 10.213.CC.CC, 10.76.AA.AA), so nothing is cached in the modem. The APN isn't mine either - the last thing comgt wrote to cid 1 was `AT+CGDCONT=1,"IP",""`, empty, and the network filled that in.

So no init command can stop it. Attaching isn't optional, and attaching brings this context up.

I also found what actually causes the carrier drop, and it isn't our CGDCONT. Context already up, sending only `AT+qnetdevctl=1,1,1` and nothing else:

```
10:58:01  qnetdevctl sent
10:58:02  carrier=1
10:58:03  carrier=0     <- drop
10:58:04  carrier=1
```

Same both runs: about a second long, about two seconds after the data call is enabled. So it comes from the data call coming up, not from reconfiguring anything. (I first thought there was no drop here - Maybe I was polling too slow... )

And the wait itself, 3 runs each, alternating:

```
linksettle=0        carrierloss=1  lease=0  wwan4down=1  route=0  ping=0/3
linksettle=default  carrierloss=0  lease=1  wwan4down=0  route=1  ping=3/3
```

3/3 dead, 3/3 fine. With the wait netifd logs no carrier loss at all - the drop still happens, it just happens before the interface is handed over, so netifd never tears the sub-interface down.

So I will change my commit msgs... 

One limit worth saying: carrier doesn't mean the data call works. cdc_ncm raises it as soon as the netdev is enabled. I hit  his when the modem was in limited service - carrier 1, but `CGACT: 1,0` and `CGPADDR` empty, so udhcpc had nothing to talk to. That fails the same way with or without this patch though, so it's a gap that was already there, not something the wait breaks.

I did look at waiting for `AT+CGPADDR` instead. It's worse here: the default bearer means CGPADDR already has an  address at attach, before we send qnetdevctl, so on a fresh boot that wait would just return straight away and do nothing. Carrier at least starts at 0 and changes, so there's something to wait for. 

We could add a proper readiness check on top if you think it's worth it, I just didn't want another AT command in the hot path. wdyt @patrakov .. I'm not sure.

carlosz1986:
I rebased, changed the commit msg (after newest learning) and fixed one change in the DTS:

* pwr changed to gpio-export,output = <0> - since 8f3285dbd0 ("kernel: gpio-export: restore prior direction setup") the initial output is applied raw, ignoring GPIO_ACTIVE_LOW, so <1> parked the modem power enable high and the modem was never detected.

FredApps:
> We could add a proper readiness check on top if you think it's worth it, I just didn't want another AT command in the hot path.

Checked the CGPADDR half of that on the other vendor, since it is a claim about the network rather than about Quectel. It holds.

`AT+CFUN=4`, then `AT+CFUN=1` and nothing else — no context configuration, no data call:

```
+CGATT: 1
+CGDCONT: 0,"IP","online.telia.se","95.192.6.232",0,0,0,0
+CGDCONT: 1,"IP","online.telia.se","95.192.6.232",0,0,0,0
+CGACT: 1,1
+CGPADDR: 1,"95.192.6.232"
+GTRNDIS: 0                          <- data call off
```

That APN is not one we set either. So the default bearer is up with an address before anything is dialled here too, and a CGPADDR wait would return straight away, exactly as you found. One extra wrinkle on this modem: `AT+CGPADDR` lists every context, inactive ones as `0.0.0.0`, so even a per-cid "does this one have an address" test is satisfied at attach.

**A scope note on the settle commit, not an objection.** The carrier flap does not happen on this modem. Three ifdown/ifup cycles, sampling `/sys/class/net/usb0/carrier` in a tight loop at roughly 90 samples/s:

| run | samples | `carrier=0` samples | carrier up after ifup |
| --- | --- | --- | --- |
| 1 | 3623 | 0 | — |
| 2 | 3866 | 0 | +5.21 s |
| 3 | 3874 | 0 | +5.23 s |

Carrier goes from unreadable straight to `1` and never drops. A one-second flap would have left about ninety consecutive zeros; there are none. The PDP address still differs every session — 95.192.6.232, 95.192.13.216, 95.192.108.58 — but it changes at attach, not across a flap, because there is no flap.

So "Modems raise carrier as soon as the data call starts and drop it again briefly" reads as a property of the EG060W rather than of modems generally, and on this one the wait is three seconds of pure cost, consistent with the 14 s vs 17 s I measured earlier.

I still think the patch is right as it stands. The failure it prevents is permanent rather than transient, three seconds is cheap against that, and `linksettle 0` is there for anyone who measures no flap on their hardware. It just seems worth knowing that the benefit is modem-specific while the cost is not, in case anyone asks why it applies to every modem.

Also re-tested model detection against the current head, since `${model%%-*}` landed after my last run: `AT+CGMM` returns `FG621-EA`, the strip gives `fg621`, `ncm_select_modem` tries `fibocom-fg621`, misses, and falls back to the plain `fibocom` entry. Still correct.

target/linux/ramips/dts/mt7621_cudy_lt700-outdoor-v1.dts
```diff
@@ -0,0 +1,245 @@
...
+	};
+
+	gpio-export {
+		compatible = "gpio-export";
+
+		pwr {
+			gpio-export,name = "pwr";
+			gpio-export,output = <0>;
+			gpios = <&gpio 41 GPIO_ACTIVE_LOW>;
+		};
```
openwrt-ai, line 109: nit: `GPIO_ACTIVE_LOW` together with `output = <0>` reads as "leave it deasserted", and only the commit message says otherwise — the value is written with [`gpiod_direction_output_raw()`](``https://github.com/openwrt/openwrt/blob/16603667c097d012e1fce8d20e9d3696bb49579c/target/linux/generic/hack-6.18/800-GPIO-add-named-gpio-exports.patch#L55),`` so `<0>` is what powers the modem. This revision dropped the polarity comments that were here, which makes it the form most likely to get "corrected" back to `<1>`; please keep a short comment, on `usb4g` below as well.

```suggestion
		/* Modem power enable, active low; gpio-export writes the value raw, so <0> powers the modem */
		pwr {
			gpio-export,name = "pwr";
			gpio-export,output = <0>;
			gpios = <&gpio 41 GPIO_ACTIVE_LOW>;
		};
```

---

openwrt-ai (commented):
Reviewed 2 new commits.

---

patrakov:
I have no objections to adding as many AT commands as needed to resolve this bug.

carlosz1986:
@FredApps thx for your tremendous help :-) cheers!

carlosz1986:
> I have no objections to adding as many AT commands as needed to resolve this bug.

 @patrakov cool, then I'll try to add a real check :-)
CGPADDR won't do it though: FredApps saw the same on his Fibocom, it has an address right at attach, so it'd always pass.

My idea: wait for AT+CGATT? = 1 before connecting. It's somehow what others like  directip.sh and 3g.sh already do it, and it's exactly what was wrong when my link died (+CGATT: 0). If it never attaches, fail and let netifd retry, like qmi.sh does.

linksettle stays, that's a different problem. I'd send it as a separate PR after testing it on my LT700. This PR works, and the changes are somehow "independent" ...  Fine for you ?

openwrt-ai (commented):

---

carlosz1986:
@jonasjelonek wdyt? From my pov we are ready to go ... best :-)

jonasjelonek:

PR #24692, opened 2026-08-12
kirkwood: dns325: drive Marvell PHY LEDs via LED trigger API

SimplyKot:
Drive the two PHY LEDs of the D-Link DNS-325 (Marvell 88E1116R) through the standard LED trigger API with hardware offload.

Two patches:

1. `net: phy: marvell: add LED support for 88E1116R PHY` — the 88E1116R exposes its LEDs through the same page 3 register 0x10 layout as the 88E1318S (4 bits of function-select per LED for LED[0..2]). This wires the existing `m88e1318_led_*` ops into the 88E1116R `phy_driver` entry and adds a `led_polarity_set` callback (`m88e1318_led_polarity_set()`) so the LED polarity can be described through the standard PHY LED DT binding (`active-high`/`active-low`). The polarity support is based on Aleksander Jan Bajkowski's patch (`Co-developed-by`).
2. `ARM: dts: kirkwood: dns325: configure Marvell PHY LEDs via LED trigger API` — describes the two PHY LEDs in the DT (`led@0` = LED0, left 1G link indicator, `LED_FUNCTION_LAN`; `led@2` = LED2, right rx/tx activity, `LED_FUNCTION_ACTIVITY`) with `active-high` polarity, and adds the netdev default triggers in `board.d/01_leds`.

Verified on a DNS-325 rev A2 (kernel 6.18): both LEDs hardware-offloaded (`offloaded=1`), page 3 reg 0x11 = `0x4411`, reg 0x10 = `0x1117` (LED0 = 1G link, LED2 = LINK|RX|TX, matching stock). The left LED lights on the 1G link; the right LED is on while the link is up and blinks on traffic.

### DIFF
--- a/target/linux/kirkwood/patches-6.18/119-dns325-add-phy-led-reg-init.patch
+++ b/target/linux/kirkwood/patches-6.18/119-dns325-add-phy-led-reg-init.patch
@@ -1,0 +1,22 @@
+From 910160930ed479aed825cd676882461bc92ff1e7 Mon Sep 17 00:00:00 2001
+From: Konstantin Meshkov <skyline@mail.ru>
+Date: Wed, 12 Aug 2026 14:08:30 +0300
+Subject: [PATCH] kirkwood: dns325: configure Marvell PHY LEDs via reg-init
+
+---
+ arch/arm/boot/dts/marvell/kirkwood-dns325.dts | 5 +++++
+ 1 file changed, 5 insertions(+)
+
+diff --git a/arch/arm/boot/dts/marvell/kirkwood-dns325.dts b/arch/arm/boot/dts/marvell/kirkwood-dns325.dts
+index aa7e43e..4706f54 100644
+--- a/arch/arm/boot/dts/marvell/kirkwood-dns325.dts
++++ b/arch/arm/boot/dts/marvell/kirkwood-dns325.dts
+@@ -113,3 +113,8 @@
+ &fan0 {
+ 	#cooling-cells = <2>;
+ };
++
++&ethphy0 {
++	marvell,reg-init = <3 0x10 0xffff 0x1107>,
++	                   <3 0x11 0xffff 0x4411>;
++};


SimplyKot:
Please run formality check

target/linux/kirkwood/patches-6.18/119-dns325-add-phy-led-reg-init.patch
```diff
@@ -0,0 +1,22 @@
...
+index aa7e43e..4706f54 100644
+--- a/arch/arm/boot/dts/marvell/kirkwood-dns325.dts
++++ b/arch/arm/boot/dts/marvell/kirkwood-dns325.dts
+@@ -113,3 +113,8 @@
+ &fan0 {
+ 	#cooling-cells = <2>;
+ };
++
++&ethphy0 {
++	marvell,reg-init = <3 0x10 0xffff 0x1107>,
```
abajk: The correct solution would be to implement an LED API for the 88E1116R. Take the 88E1510 as an example[1].

1. https://elixir.bootlin.com/linux/v7.2-rc7/source/drivers/net/phy/marvell.c#L3945

abajk (changes requested):

target/linux/kirkwood/patches-6.18/119-dns325-add-phy-led-reg-init.patch
abajk: and then describe it in dts like this:
~~~
		phy6: phy@6 {
			compatible = "ethernet-phy-ieee802.3-c45";
			reg = <6>;

			leds {
				#address-cells = <1>;
				#size-cells = <0>;

				led@0 {
					reg = <0>;
					color = <LED_COLOR_ID_GREEN>;
					function = LED_FUNCTION_WAN;
				};
			};
		};
~~~
source: https://github.com/openwrt/openwrt/blob/main/target/linux/mediatek/dts/mt7986a-zyxel-ex5601-t0-common.dtsi#L236C1-L251C5

target/linux/kirkwood/patches-6.18/119-dns325-add-phy-led-reg-init.patch
```diff
@@ -0,0 +1,22 @@
...
++	                   <3 0x11 0xffff 0x4411>;
```
openwrt-ai: `mask = 0xffff` does not write these registers — it turns the write into a pure OR that can never clear a bit, so the stock values 0x1107 / 0x4411 will not actually be programmed.

In `marvell,reg-init` the third cell is the set of bits to **preserve** from the current register value, not the bits to change ([binding doc at marvell.c:520-530](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/marvell.c#L520-L530)):

```
 * mask: if non-zero, ANDed with existing register value.
 * value: ORed with the masked value and written to the regiser.
```

[The implementation at marvell.c:565-576](https://github.com/gregkh/linux/blob/2fe596715f840d053aed5cee5455f701bdcd2b50/drivers/net/phy/marvell.c#L565-L576) does `val &= mask; val |= val_bits;`. With `mask = 0xffff` the entire existing value is kept, so the result is `old | 0x1107` — and the premise of this patch is that `old` is the driver's *default* LED config, i.e. it has bits set that 0x1107 does not. Those bits stay set.

To write the value verbatim the mask must be `0`. That is what every other in-tree user of these page-3 LED registers does — e.g. [`armada-370-buffalo-ls220de.dts:231-233`](``https://github.com/openwrt/openwrt/blob/efff5c88a11a90c8366be3d7a7b4ef18dfad8b7a/target/linux/mvebu/files/arch/arm/boot/dts/marvell/armada-370-buffalo-ls220de.dts#L231-L233),`` which programs the same 0x10 (LED function) / 0x11 (LED polarity) pair and uses mask `0x0000` for the exact write, and [`armada-385-nas1dual.dts:201`](``https://github.com/openwrt/openwrt/blob/efff5c88a11a90c8366be3d7a7b4ef18dfad8b7a/target/linux/mvebu/files/arch/arm/boot/dts/marvell/armada-385-nas1dual.dts#L201)`` (`<3 16 0 0x1017>`).

```suggestion
+	marvell,reg-init = <3 0x10 0 0x1107>,
+	                   <3 0x11 0 0x4411>;
```


---

target/linux/kirkwood/patches-6.18/119-dns325-add-phy-led-reg-init.patch
```diff
@@ -0,0 +1,22 @@
...
+Date: Wed, 12 Aug 2026 14:08:30 +0300
+Subject: [PATCH] kirkwood: dns325: configure Marvell PHY LEDs via reg-init
+
+---
+ arch/arm/boot/dts/marvell/kirkwood-dns325.dts | 5 +++++
+ 1 file changed, 5 insertions(+)
+
+diff --git a/arch/arm/boot/dts/marvell/kirkwood-dns325.dts b/arch/arm/boot/dts/marvell/kirkwood-dns325.dts
+index aa7e43e..4706f54 100644
+--- a/arch/arm/boot/dts/marvell/kirkwood-dns325.dts
```
openwrt-ai: nit: the `git format-patch` preamble doesn't match the rest of this directory — all 23 patches in `patches-6.18` (and all 23 in `patches-6.12`) start directly at `--- a/...`, with no `From`/`Date`/`Subject`, no diffstat, and no `diff --git`/`index` lines. Compare [`108-dns325-add-led-labels-and-thermal-zones.patch:1-2`](``https://github.com/openwrt/openwrt/blob/efff5c88a11a90c8366be3d7a7b4ef18dfad8b7a/target/linux/kirkwood/patches-6.18/108-dns325-add-led-labels-and-thermal-zones.patch#L1-L2),`` the patch this one builds on.

Also worth noting: `From 910160930ed479aed825cd676882461bc92ff1e7` is a SHA from your local branch rather than an upstream Linux commit, so leaving it in reads as if this hunk were an upstream backport when it isn't.

```suggestion
--- a/arch/arm/boot/dts/marvell/kirkwood-dns325.dts
```


---

target/linux/kirkwood/patches-6.18/119-dns325-add-phy-led-reg-init.patch
openwrt-ai, line 1: nit: `target/linux/kirkwood/` carries both `patches-6.12` and `patches-6.18`, and the two directories currently hold an identical set of 23 patches (including the `108-`/`109-` dns325 patches this one depends on). This PR adds `119-` only to `patches-6.18`, which makes the two diverge.

`KERNEL_PATCHVER:=6.18` and there's no `KERNEL_TESTING_PATCHVER`, so 6.12 may just be on its way out — is skipping it intentional, or should the patch be added to `patches-6.12` as well?

---

openwrt-ai (commented):
Reviewed 1 new commit; see inline comments. The commit message matches the diff, and the patch applies cleanly on top of `108-`/`109-` (verified against v6.18.41 — context at line 113 and the `&fan0` block resolve correctly).

---

target/linux/kirkwood/patches-6.18/119-dns325-add-phy-led-reg-init.patch
abajk: This comment is wrong. Patches should have a header so they can be applied using the `git am` command.

target/linux/kirkwood/patches-6.18/119-dns325-add-phy-led-reg-init.patch
SimplyKot: The DNS-325 is just a basic NAS; it has a single LAN port on the back with two built-in LEDs.

photo_2026-08-31_11-02-26


Furthermore, I cannot find the register specifications for the 88E1116R that control LED behavior, so creating a full-fledged LED API won't be possible.

I will try to configure it based on what I know, but I certainly don't claim the result will be comprehensive.

target/linux/kirkwood/patches-6.18/119-net-phy-marvell-add-88e1116r-led-api.patch
```diff
@@ -0,0 +1,255 @@
...
++	case 1:
++		if (!marvell_find_led_rules(rules, marvell_88e1116r_led1,
++					    ARRAY_SIZE(marvell_88e1116r_led1),
++					    mode))
++			return 0;
++		*rules = BIT(TRIGGER_NETDEV_RX) | BIT(TRIGGER_NETDEV_TX);
++		return 0;
++	default:
++		return -EINVAL;
++	}
```
openwrt-ai: Both cases throw away the `-EOPNOTSUPP` that [`marvell_find_led_rules()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/marvell.c#L3483-L3497) returns for a mode that isn't in the table and report a hard-coded rule instead, so a LED parked at OFF (0x9) or ON (0x1e/0xe) by `led_brightness_set` still reads back as link-1000 / rx+tx hardware control. Return the helper's result directly, the way [`m88e1318_led_hw_control_get()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/marvell.c#L3570-L3585) does. Since that shortens the hunk, regenerate the patch with `make target/linux/refresh` so the hunk headers stay correct.

---

target/linux/kirkwood/patches-6.18/119-net-phy-marvell-add-88e1116r-led-api.patch
```diff
@@ -0,0 +1,255 @@
...
+ 		return err;
+ 
++	/* Set LED polarity/mode baseline. reg17=0x4411 is required for
++	 * correct LED behavior on boards with active-low LED wiring
++	 * (e.g. D-Link DNS-325). Without this, the function codes
++	 * produce inverted or incorrect LED states.
++	 */
++	err = phy_write_paged(phydev, MII_MARVELL_LED_PAGE, 0x11, 0x4411);
++	if (err < 0)
++		return err;
```
openwrt-ai: This writes a DNS-325-specific LED polarity value from `m88e1116r_config_init()`, which runs for every 88E1116R board in the tree (dockstar, pogoplug, nsa3xx, ...) and clobbers whatever their bootloader or `marvell,reg-init` set. Polarity has a dedicated hook: implement `led_polarity_set` and put `active-low` in the DNS-325 `led@N` nodes — see [`of_phy_led()` at phy_device.c:3223-3241](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/phy_device.c#L3223-L3241) — rather than hard-coding it in the shared init path.

---

target/linux/kirkwood/patches-6.18/119-net-phy-marvell-add-88e1116r-led-api.patch
```diff
@@ -0,0 +1,255 @@
+From a0349e22f9592df2b8b610d200af781aac376a61 Mon Sep 17 00:00:00 2001
+From: Konstantin Meshkov <skyline@mail.ru>
+Date: Thu, 27 Aug 2026 10:46:01 +0300
+Subject: [PATCH] net: phy: marvell: add LED API support for 88E1116R
+
+Signed-off-by: Konstantin Meshkov <skyline@mail.ru>
```
openwrt-ai: nit: the patch body is empty — only a Signed-off-by. Move the rationale from the OpenWrt commit message (register layout, function codes, why the 88E1116R needs its own handlers) in here so the patch stands on its own when it goes upstream. Same for `120-dns325-add-phy-led-config.patch`.

---

target/linux/kirkwood/patches-6.12/119-net-phy-marvell-add-88e1116r-led-api.patch
```diff
@@ -0,0 +1,255 @@
...
+
+---
+ drivers/net/phy/marvell.c | 219 ++++++++++++++++++++++++++++++++++++++
+ 1 file changed, 219 insertions(+)
+
+diff --git a/drivers/net/phy/marvell.c b/drivers/net/phy/marvell.c
+index c248c90..c8d919d 100644
+--- a/drivers/net/phy/marvell.c	2026-08-09 21:25:30.000000000 +0300
++++ b/drivers/net/phy/marvell.c	2026-08-28 13:21:40.301511058 +0300
+@@ -1263,6 +1263,14 @@
```
openwrt-ai: nit: this file is byte-identical to the 6.18 copy and was generated against 6.18 sources — old line 1263 sits in `m88e1116r_config_init()` on 6.18 but in `m88e1510_config_init()` on 6.12, where the real target is line 1220; the other two hunks are off by 30. It only lands correctly because patch searches outward for matching context. Regenerate this copy against 6.12 with `make target/linux/refresh`.

---

target/linux/kirkwood/patches-6.18/120-dns325-add-phy-led-config.patch
```diff
@@ -0,0 +1,58 @@
...
++		#size-cells = <0>;
++
++		/* left PHY LED: link status indicator (stock behavior) */
++		led@0 {
++			reg = <0>;
++			color = <LED_COLOR_ID_GREEN>;
++			function = LED_FUNCTION_STATUS;
++			linux,default-trigger = "netdev";
++			netdev,dev = "eth0";
++			netdev,link-1000;
```
openwrt-ai: `netdev,dev` and `netdev,link-1000` are not read by anything in the kernel — [`leds/common.yaml`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/leds/common.yaml#L76-L118) only defines `netdev` as a value of `linux,default-trigger`, and `ledtrig-netdev.c` has no DT parsing at all; it takes the interface from `hw_control_get_device()` and the modes from `hw_control_get()`. Drop them. Same for `netdev,dev`/`netdev,rx`/`netdev,tx` on lines 50-52.

---

target/linux/kirkwood/base-files/etc/uci-defaults/dns325_led
```diff
@@ -0,0 +1,27 @@
...
+	lan=$(uci add system led)
+	uci set system.$lan.sysfs="f1072004.mdio-bus-mii:08:green:status"
+	uci set system.$lan.trigger="netdev"
+	uci set system.$lan.dev="eth0"
+	uci add_list system.$lan.mode="link_1000"
+
+	act=$(uci add system led)
+	uci set system.$act.sysfs="f1072004.mdio-bus-mii:08:green:activity"
+	uci set system.$act.trigger="netdev"
+	uci set system.$act.dev="eth0"
```
openwrt-ai: Per-board LED defaults belong in [`target/linux/kirkwood/base-files/etc/board.d/01_leds`](https://github.com/openwrt/openwrt/blob/b5ab3b2fc5925179bb11de4a4c9e850b6c82b213/target/linux/kirkwood/base-files/etc/board.d/01_leds), which already has a `case "$board"` block for this target; [`ucidef_set_led_netdev`](https://github.com/openwrt/openwrt/blob/b5ab3b2fc5925179bb11de4a4c9e850b6c82b213/package/base-files/files/lib/functions/uci-defaults.sh#L470-L482) emits exactly these `system led` sections. Hand-rolling `uci add system led` in a uci-defaults script duplicates that machinery, including the `uci show | grep` guard it needs to stay idempotent. Adding two `ucidef_set_led_netdev` lines there also drops the need for this file (whose name lacks the `NN_` prefix every other uci-defaults script uses).

---

openwrt-ai (commented):
## Commit checks

- `213584a` "net: phy: marvell: add LED API support for 88E1116R" — Linux-style subject on a commit that only touches `target/linux/kirkwood/patches-*`; use the target prefix, e.g. `kirkwood: add LED API support for Marvell 88E1116R PHY`.
- `213584a` — the message covers only the LED handlers, but the diff also adds an unconditional `phy_write_paged(..., 0x11, 0x4411)` to `m88e1116r_config_init()`. Document that hunk in the message or drop it.

---

target/linux/kirkwood/patches-6.18/119-dns325-add-phy-led-reg-init.patch
openwrt-ai: solved by dropping the `marvell,reg-init` approach entirely, thanks

---

target/linux/kirkwood/patches-6.18/119-dns325-add-phy-led-reg-init.patch
openwrt-ai: fixed, thanks — both new patches are now present in `patches-6.12` and `patches-6.18`

---

SimplyKot:
Thanks for the reviews. I reworked the series based on all the feedback and re-verified on hardware (DNS-325 rev A2, kernel 6.18). Summary of the changes in the updated commits (`9114021`):

**Commit 1 — `kirkwood: marvell: add LED API support for 88E1116R PHY`** (patches 119)
- Dropped the unconditional `phy_write_paged(..., 0x11, 0x4411)` from `m88e1116r_config_init()`. The per-board LED polarity/function baseline is no longer touched by the shared driver (the DNS-325 sets it via `marvell,reg-init` in its DTS), so dockstar/pogoplug/nsa3xx etc. are unaffected.
- Removed the `led_polarity_set` handler entirely: since the baseline is board-provided and not a per-LED `active-low` polarity, the handler was wrong and is gone.
- `led_hw_control_get()` now returns the result of `marvell_find_led_rules()` directly (no more swallowed `-EOPNOTSUPP` / hard-coded rules).
- Commit subject now uses the target prefix; the patch body carries the full rationale (register layout, function codes, why dedicated handlers are needed).

**Commit 2 — `kirkwood: dns325: configure Marvell PHY LEDs via LED trigger API`** (patches 120)
- The 88E1116R LED polarity baseline (page 3, reg 0x11 = 0x4411) is applied board-specifically via `marvell,reg-init = <3 0x11 0 0x4411>;` in the DNS-325 ethernet PHY node. This mirrors how the stock D-Link firmware programs it (`mv_netdev.c`, "control phy led") and the in-tree `m88e1118`/DNS-323 precedent.
- Removed the unparsed `netdev,dev` / `netdev,link-1000` / `netdev,rx/tx` properties from the LED nodes.
- Moved LED defaults out of the hand-rolled uci-defaults script into `board.d/01_leds` using `ucidef_set_led_netdev()` (deleted `uci-defaults/dns325_led`).

**patches-6.12**: 6.12 is not built in this tree (kirkwood `KERNEL_PATCHVER:=6.18`, no 6.12 kernel details exist), so I kept both directories in sync by mirroring the 6.18 copies rather than leaving them to drift.

**Hardware verification** (DNS-325, kernel 6.18): after a clean boot both PHY LEDs are hardware-offloaded (`offloaded=1`), page-3 reg 0x11 reads `0x4411`, and physically the left LED is lit on a 1 Gbit/s link while the right LED blinks on traffic.

target/linux/kirkwood/patches-6.18/119-net-phy-marvell-add-88e1116r-led-api.patch
SimplyKot: Fixed in the reworked patch: `led_hw_control_get()` now returns `marvell_find_led_rules()` directly for both LED indices, matching `m88e1318_led_hw_control_get()`. The hard-coded fallback rules were removed.

target/linux/kirkwood/patches-6.18/119-net-phy-marvell-add-88e1116r-led-api.patch
SimplyKot: Agreed. The unconditional `phy_write_paged(..., 0x11, 0x4411)` was dropped from `m88e1116r_config_init()`. The DNS-325 applies its LED polarity/function baseline board-specifically via `marvell,reg-init = <3 0x11 0 0x4411>;` in the ethernet PHY node, so other 88E1116R boards (dockstar, pogoplug, nsa3xx, ...) are left untouched. The `led_polarity_set` handler was removed as well, since the baseline is not a per-LED active-low polarity.

target/linux/kirkwood/patches-6.18/119-net-phy-marvell-add-88e1116r-led-api.patch
SimplyKot: Fixed: both patches now carry the full rationale (register layout, function codes, why the 88E1116R needs dedicated handlers) in their bodies.

target/linux/kirkwood/patches-6.12/119-net-phy-marvell-add-88e1116r-led-api.patch
SimplyKot: 6.12 is not built in this tree (kirkwood sets KERNEL_PATCHVER:=6.18 and there are no 6.12 kernel version details), so I kept `patches-6.12` in sync with `patches-6.18` rather than letting the two diverge. If 6.12 support is expected to return I can regenerate against a real 6.12 tree.

target/linux/kirkwood/patches-6.18/120-dns325-add-phy-led-config.patch
SimplyKot: Fixed: the `netdev,dev` / `netdev,link-1000` / `netdev,rx/tx` properties were removed from both LED nodes.

target/linux/kirkwood/base-files/etc/uci-defaults/dns325_led
SimplyKot: Fixed: the hand-rolled `uci-defaults/dns325_led` script was deleted and the LED defaults now live in `board.d/01_leds` via `ucidef_set_led_netdev()`.

target/linux/kirkwood/patches-6.18/119-net-phy-marvell-add-88e1116r-led-api.patch
```diff
@@ -0,0 +1,242 @@
...
++ *   [3:0]   = LED0 (left)  function select
++ *   [7:4]   = LED0 (left)  blink control
++ *   [11:8]  = LED1 (right) function select
++ *   [15:12] = LED1 (right) blink control
++ * Function codes (verified with reg17=0x4411):
++ *   OFF=0x9, ON=0xe, Activity=0x1, 1G_Link=0x7
++ */
++#define MII_88E1116R_PHY_LED_FUNC_OFF		0x9
++#define MII_88E1116R_PHY_LED_FUNC_ON		0xe
++#define MII_88E1116R_PHY_LED_FUNC_BLINK		0xb
```
openwrt-ai: The `[7:4]`/`[15:12]` blink-control split doesn't look like this family's layout — page 3 reg 0x10 is 4 bits of *function select* per LED for LED[2:0], which is why [`m88e1318_led_brightness_set()` uses `4 * index` for indices 0-2](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/marvell.c#L3275-L3286) on the same page/register — and the codes you measured fit that exactly: 0x7 is [`marvell_led0` mode 7 = LINK_1000](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/marvell.c#L3356-L3359) (bits [3:0]) and 0x1 is [`marvell_led2` mode 1 = LINK|RX|TX](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/marvell.c#L3408-L3413) (bits [11:8]), while the stock `0x10 = 0x1107` from this PR's first revision decodes as LED0=0x7 / LED1=0x0 / LED2=0x1. Likewise `OFF=0x9` is the family's *Force On* value and 0xe is not a force code at all ([`_OFF` is 0x8, `_ON` 0x9, `_BLINK` 0xb](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/marvell.c#L148-L151)); an inverted-looking result here is more likely the reg 0x11 polarity you program than a different code table. If the right-hand LED is in fact LED[2] rather than LED[1], this whole patch is unnecessary: add the existing `m88e1318_led_*` ops to the 88E1116R `phy_driver` entry and use `reg = <0>` / `reg = <2>` in the DTS `leds` node.

---

target/linux/kirkwood/patches-6.18/120-dns325-add-phy-led-config.patch
```diff
@@ -0,0 +1,66 @@
+From 755fec58873781b57f90deb3e3b6d30e19d803bb Mon Sep 17 00:00:00 2001
+From: Konstantin Meshkov <skyline@mail.ru>
+Date: Thu, 27 Aug 2026 10:48:36 +0300
+Subject: [PATCH] kirkwood: dns325: configure Marvell PHY LEDs via LED trigger API
```
openwrt-ai: nit: `kirkwood:` is the OpenWrt target prefix; inside a `git am`-able header the subject is a Linux one, and for `arch/arm/boot/dts/marvell/` that is `ARM: dts: kirkwood:`. Same for `119-…`, where the subject prefix for `drivers/net/phy/marvell.c` would be `net: phy: marvell:`.

```suggestion
Subject: [PATCH] ARM: dts: kirkwood: dns325: configure Marvell PHY LEDs via LED trigger API
```

---

openwrt-ai (commented):
Reviewed 2 new commits.

---

target/linux/kirkwood/patches-6.18/119-net-phy-marvell-add-88e1116r-led-api.patch
```diff
@@ -0,0 +1,242 @@
+From 213584a702492eef6e367ef991cbf453b8fb4e82 Mon Sep 17 00:00:00 2001
```
abajk, line 1: Great job that you managed to add LED API support for this PHY. The next step is to submit this patch upstream :)
https://docs.kernel.org/process/submitting-patches.html

abajk:
> * Removed the `led_polarity_set` handler entirely: since the baseline is board-provided and not a per-LED `active-low` polarity, the handler was wrong and is gone.

What is the issue with the implementation of ‘led_polarity_set’?
This is the proper way to set the polarity. This is how we do it on the MediaTek[1] and Lantiq targets for the MaxLinear and Lantiq-Xway PHYs[2].

1. https://github.com/openwrt/openwrt/blob/main/target/linux/mediatek/dts/mt7986a-smartrg-bonanza-peak.dtsi#L230
2. https://elixir.bootlin.com/linux/v7.2.2/source/drivers/net/phy/mxl-gpy.c#L1025

target/linux/kirkwood/patches-6.18/119-net-phy-marvell-add-88e1116r-led-api.patch
SimplyKot: You're right — thank you for the detailed analysis. I verified the register layout on hardware and reworked the series:

- The 88E1116R uses the standard page 3 reg 0x10 layout: 4 bits of function-select per LED for LED[0..2] (shift = 4 * index), with the family function codes. The stock `0x10 = 0x1107` decodes as LED0=0x7 (LINK_1000) + LED2=0x1 (LINK|RX|TX), LED1 unused — the DNS-325's right-hand LED is physically LED[2], not LED[1]. The previous revision's custom shift table and "blink nibble" theory were wrong.
- Patch 119 is now minimal: it wires the existing `m88e1318_led_*` ops into the 88E1116R `phy_driver` entry (same pattern as 88E1510/88E1540). All custom 88E1116R LED handlers, tables and the bogus OFF=0x9/ON=0xe constants are removed.
- Patch 120 now describes `led@0` (LED0, left) and `led@2` (LED2, right); the board-specific `marvell,reg-init = <3 0x11 0 0x4411>` polarity baseline stays in the DTS.

Verified on the device (kernel 6.18.44): with the netdev trigger on `eth0` both LEDs are hardware-offloaded (`offloaded=1`), page 3 reg 0x11 reads `0x4411` and reg 0x10 reads `0x1117` (LED0=7 LINK_1000 / LED2=1 LINK|RX|TX) — matching stock. The left LED lights on the 1G link; the right LED is on while the link is up and blinks on traffic (the `activity` LED is configured with the `link rx tx` netdev mode to reproduce that stock LINK|RX|TX behaviour).

target/linux/kirkwood/patches-6.18/120-dns325-add-phy-led-config.patch
SimplyKot: Fixed — both patch subjects now use the Linux-style prefixes: `net: phy: marvell: add LED support for 88E1116R PHY` (patch 119) and `ARM: dts: kirkwood: dns325: configure Marvell PHY LEDs via LED trigger API` (patch 120).

target/linux/kirkwood/patches-6.18/120-dns325-add-phy-led-config.patch
```diff
@@ -0,0 +1,66 @@
...
+ #include <dt-bindings/thermal/thermal.h>
+ 
+ / {
+@@ -113,3 +114,28 @@
+ &fan0 {
+ 	#cooling-cells = <2>;
+ };
++
++&ethphy0 {
++	marvell,reg-init = <3 0x11 0 0x4411>;
```
openwrt-ai, line 44: nit: the PR description still describes the abandoned revision — it promises the LEDs are driven "instead of a one-off `marvell,reg-init`" (this line reintroduces one), says patch 1 implements dedicated 88E1116R brightness/blink/hw-control handlers with "LED1 in [11:8]" (it now just reuses the `m88e1318_led_*` ops, and the second LED is LED2), and says patch 2 "adds an uci-defaults script" (replaced by `board.d/01_leds`). Worth refreshing so it matches what is actually being merged.

---

target/linux/kirkwood/patches-6.18/120-dns325-add-phy-led-config.patch
```diff
@@ -0,0 +1,66 @@
...
++
++	leds {
++		#address-cells = <1>;
++		#size-cells = <0>;
++
++		/* left PHY LED: LED0 link status indicator (stock behavior) */
++		led@0 {
++			reg = <0>;
++			color = <LED_COLOR_ID_GREEN>;
++			function = LED_FUNCTION_STATUS;
```
openwrt-ai: nit: `LED_FUNCTION_STATUS` is the board-status function, which on this board is already the white power LED (`led-running` alias in `108-dns325-add-led-labels-and-thermal-zones.patch`); in-tree PHY LED nodes label port LEDs with `LED_FUNCTION_LAN`/`LED_FUNCTION_WAN` instead — e.g. [`ipq8072-ax9000.dts:404-414`](https://github.com/openwrt/openwrt/blob/8b0d29bf188e1518be81cbf8919c180e1e9e7da4/target/linux/qualcommax/dts/ipq8072-ax9000.dts#L404-L414) and [`mt7986a-zyxel-ex5601-t0-common.dtsi:236-251`](``https://github.com/openwrt/openwrt/blob/8b0d29bf188e1518be81cbf8919c180e1e9e7da4/target/linux/mediatek/dts/mt7986a-zyxel-ex5601-t0-common.dtsi#L236-L251).`` `LED_FUNCTION_LAN` fits the single rear LAN port better; the sysfs name on line 13 of `base-files/etc/board.d/01_leds` and the `patches-6.12` copy need the same change.

---

openwrt-ai (commented):
Reviewed 2 new commits.

---

target/linux/kirkwood/patches-6.18/120-dns325-add-phy-led-config.patch
SimplyKot: Fixed — the PR description has been rewritten to match what is actually merged: it no longer mentions a one-off `marvell,reg-init`, describes patch 1 as reusing the existing `m88e1318_led_*` ops (LED0/LED2, not "LED1 in [11:8]"), and patch 2 as adding the DT nodes plus `board.d/01_leds` (no uci-defaults script).

target/linux/kirkwood/patches-6.18/120-dns325-add-phy-led-config.patch
SimplyKot: Fixed — `LED_FUNCTION_STATUS` replaced with `LED_FUNCTION_LAN` on `led@0`; the sysfs name in `board.d/01_leds` (and the patches-6.12 copy) now references `...:green:lan`. Verified on the device: the LED registers as `f1072004.mdio-bus-mii:08:green:lan`, hardware-offloaded, with reg 0x10 = 0x1117 as before.

openwrt-ai (commented):

---

abajk:
In general, it looks good. In the meantime, I've implemented an LED polarity setting that should also work for the 88E1116R[1].

1. https://patchwork.kernel.org/project/netdevbpf/patch/20260914200847.445375-1-olek2@wp.pl/

SimplyKot:
@abajk Thank you for the review and for sharing your patch!

I reworked the series to use your LED polarity implementation instead of the board-specific `marvell,reg-init` baseline:

- Patch 1 (`kirkwood: marvell: add LED support for 88E1116R PHY`) now adds `m88e1318_led_polarity_set()` together with the `MII_88E1318S_PHY_LED_POL` / `MII_88E1318S_PHY_LED_POL_ACT_HIGH()` defines and wires the callback into the 88E1116R driver entry. Your work is reused with attribution:

  `Co-developed-by: Aleksander Jan Bajkowski <olek2@wp.pl>`

- Patch 2 (`kirkwood: dns325: ...`) drops `marvell,reg-init = <3 0x11 0 0x4411>` and describes the polarity through the standard DT binding (`active-high`) on both LED nodes.

Verified on hardware (DNS-325, kernel 6.18): both PHY LEDs are registered and hardware-offloaded (`offloaded=1`), and page 3 registers read back `0x10 = 0x1117`, `0x11 = 0x4411` — matching the stock baseline.

Thanks again for the pointer!

target/linux/kirkwood/patches-6.18/120-dns325-add-phy-led-config.patch
```diff
@@ -0,0 +1,66 @@
...
+driver programs via the led_polarity_set callback.
+
+Signed-off-by: Konstantin Meshkov <skyline@mail.ru>
+
+---
+ arch/arm/boot/dts/marvell/kirkwood-dns325.dts | 26 +
+ 1 file changed, 26 insertions(+)
+
+--- a/arch/arm/boot/dts/marvell/kirkwood-dns325.dts	2026-09-15 12:31:04.199657547 +0300
++++ b/arch/arm/boot/dts/marvell/kirkwood-dns325.dts	2026-09-15 12:30:34.739732588 +0300
```
openwrt-ai: `Check Kernel patches (kirkwood, generic)` fails at this head:

```
target/linux/kirkwood/patches-6.18/119-net-phy-marvell-add-88e1116r-led-api.patch
target/linux/kirkwood/patches-6.18/120-dns325-add-phy-led-config.patch
Kernel patches for kirkwood/generic require refresh. (run 'make target/linux/refresh' and force push this pr)
```

The mtime suffixes here are one visible symptom — no patch anywhere under `target/linux/` carries them, which is what a hand-run `diff -u` produces rather than quilt. Run `make target/linux/refresh` with kirkwood selected and carry the result into the `patches-6.12` copies too.

---

target/linux/kirkwood/patches-6.12/119-net-phy-marvell-add-88e1116r-led-api.patch
```diff
@@ -0,0 +1,104 @@
...
++
++	if (index >= M88E1318S_MAX_LEDS)
++		return -EINVAL;
++
++	for_each_set_bit(mode, &modes, __PHY_LED_MODES_NUM) {
++		switch (mode) {
++		case PHY_LED_ACTIVE_LOW:
++			force_active_low = true;
++			break;
++		case PHY_LED_ACTIVE_HIGH:
```
openwrt-ai: `PHY_LED_ACTIVE_HIGH` does not exist on 6.12 — [`enum phy_led_modes` at phy.h:886-893](https://github.com/gregkh/linux/blob/064531c7e30cd67b79c0c694ac97f00c7796f4d6/include/linux/phy.h#L886-L893) has only `PHY_LED_ACTIVE_LOW` and `PHY_LED_INACTIVE_HIGH_IMPEDANCE` — so this copy would not build against a 6.12 tree. Relatedly, [`of_phy_led()` at phy_device.c:3388-3391](https://github.com/gregkh/linux/blob/064531c7e30cd67b79c0c694ac97f00c7796f4d6/drivers/net/phy/phy_device.c#L3388-L3391) never parses `active-high` there, so the `active-high;` in the 6.12 copy of `120-…` is a silent no-op and the polarity never gets programmed. Since `KERNEL_PATCHVER:=6.18` and 6.12 is not built for this target, dropping both `patches-6.12` copies is cleaner than shipping ones that cannot work.

---

target/linux/kirkwood/patches-6.18/119-net-phy-marvell-add-88e1116r-led-api.patch
```diff
@@ -0,0 +1,104 @@
...
++}
++
+ static int marvell_probe(struct phy_device *phydev)
+ {
+ 	struct marvell_priv *priv;
+@@ -3942,6 +3981,12 @@
+ 		.get_stats = marvell_get_stats,
+ 		.get_tunable = m88e1011_get_tunable,
+ 		.set_tunable = m88e1011_set_tunable,
++		.led_polarity_set = m88e1318_led_polarity_set,
```
openwrt-ai, line 97: nit: `m88e1318_led_polarity_set()` is named for and implements the 1318S-family page-3 reg 0x11 layout, but it is wired only into the 88E1116R entry. The five entries that already register the sibling `m88e1318_led_*` ops — [88E1318S](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/marvell.c#L3855-L3861), 88E1510, 88E1540, 88E1545, 88E1548P — keep rejecting `active-low`/`active-high` in DT with `-EINVAL`. Intentional scope for now, or should they get the callback in the same patch?

---

openwrt-ai (commented):
Reviewed 2 new commits; see inline comments.

---

target/linux/kirkwood/patches-6.18/119-net-phy-marvell-add-88e1116r-led-api.patch
```diff
@@ -0,0 +1,104 @@
...
+
+Wire up the existing 88E1318S LED handlers in the 88E1116R driver entry
+so the PHY LEDs can be described in the device tree and driven through
+the LED trigger API with hardware offload.  LED polarity (active-high /
+active-low) is configured through the PHY LED DT binding using the
+led_polarity_set callback.
+
+The DNS-325 device tree change follows in a separate commit.
+
+Co-developed-by: Aleksander Jan Bajkowski <olek2@wp.pl>
```
openwrt-ai, line 21: `Co-developed-by:` denotes authorship and must be immediately followed by a `Signed-off-by:` from that same co-author — see [submitting-patches.rst:503-511](https://github.com/gregkh/linux/blob/8f3741e6feb045da5b406df0a80b42a1adfb289b/Documentation/process/submitting-patches.rst#L503-L511). Without it netdev will bounce the patch when it goes upstream. Same in the `patches-6.12` copy and in the `00a577d` commit message.

```suggestion
Co-developed-by: Aleksander Jan Bajkowski <olek2@wp.pl>
Signed-off-by: Aleksander Jan Bajkowski <olek2@wp.pl>
```


---

openwrt-ai (commented):
## Commit checks

- `f0741f7` "kirkwood: dns325: configure Marvell PHY LEDs via LED trigger API" — the message's last paragraph says it adds the netdev default trigger configuration to `board.d/01_leds` (with the `link rx tx` activity mode), but the commit only adds the two `120-dns325-add-phy-led-config.patch` copies; the `dlink,dns-325-a1` block that the previous revision had in `01_leds` is gone at this head. Restore that hunk or drop the paragraph — and the matching claim in the PR body.

---

target/linux/kirkwood/patches-6.18/120-dns325-add-phy-led-config.patch
openwrt-ai: fixed, thanks

---

SimplyKot:
Addressed the review from the latest push (`75390c2`):

- **`01_leds` restored.** The `dlink,dns-325-a1` block was accidentally dropped when the commits were re-created during the patch refresh; it is back, so the commit message / PR body claim is accurate again.
- **Co-author `Signed-off-by` added.** `Signed-off-by: Aleksander Jan Bajkowski <olek2@wp.pl>` now immediately follows the `Co-developed-by:` line, in both the patch header and the commit message.
- **`patches-6.12` copies dropped.** The polarity binding used here (`active-high` + `led_polarity_set`) does not exist in 6.12 (`enum phy_led_modes` has no `PHY_LED_ACTIVE_HIGH` and `of_phy_led()` does not parse `active-high`), and `KERNEL_PATCHVER:=6.18` means 6.12 is not built for this target. Shipping only `patches-6.18` avoids dead patches that cannot work.
- **Re: `m88e1318_led_polarity_set()` wired only into the 88E1116R entry** — intentional scope. The DNS-325 uses the 88E1116R; the other 1318S-family entries can pick up the callback separately (as in the upstream series) without affecting this board.

The patches are the quilt-refreshed output from `make target/linux/refresh`, not hand-run `diff -u`.

PR #24694, opened 2026-08-12
mediatek: filogic: add support for Teralink TL3020 128mb

unisequence:
This commit adds support for the 128MB SPI-NAND variant of the Teralink TL3020, alongside the already merged 256MB variant (#24518 / 63bf79af2a4b).

The board is identical hardware except for NAND capacity. The DTS is split into a shared `mt7981b-teralink-tl3020.dtsi` with two thin per-variant `.dts` files that only override the `ubi` partition size.

## Hardware specification
- SoC: MediaTek MT7981B dual-core ARM Cortex-A53
- RAM: DDR3 256 MiB
- Flash: SPI-NAND 128 MiB (Winbond W25N01KV)
- WLAN: MediaTek MT7976 2.4/5 GHz Wi-Fi 6
- Ethernet: MediaTek MT7531 switch, 1x WAN + 3x LAN 10/100/1000 Mbps (2.5GbE uplink via gmac0)
- USB: 1x USB 2.0 Type-A (power enable on GPIO 14)
- Buttons: WPS, Reset
- LEDs: 7x blue, 1x red
- UART: 4-pin header on PCB, 115200n8
- Power: 12 VDC, 1.5 A (barrel jack)

Partition layout is identical to the 256mb variant up to and including `config2`; only `ubi` is reduced from `0xf980000` to `0x7980000` bytes to fit the 128 MiB chip.

Installation and MAC address handling are identical to the 256mb variant (see that commit message / PR #24518).

## Testing
Build-tested and hardware-tested on a real unit: bootloader (bl2+fip) written in-place from a running Linux userspace, booted the initramfs recovery image via TFTP, then installed the squashfs sysupgrade image. WiFi radio calibration (stored in the untouched `factory` partition, unaffected by the smaller NAND) and the USB host controller both confirmed working.


target/linux/mediatek/image/filogic.mk
```diff
@@ -3254,6 +3254,32 @@ define Device/teralink_tl3020-256mb
...
+  DEVICE_VENDOR := Teralink
+  DEVICE_MODEL := TL3020
+  DEVICE_VARIANT := 128mb
+  DEVICE_DTS := mt7981b-teralink-tl3020-128mb
+  DEVICE_DTS_DIR := ../dts
+  UBINIZE_OPTS := -E 5
+  BLOCKSIZE := 128k
+  PAGESIZE := 2048
+  KERNEL_IN_UBI := 1
+  UBOOTENV_IN_UBI := 1
```
openwrt-ai: `UBOOTENV_IN_UBI := 1` requires a matching entry in `package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic`, but that file still only lists [`teralink,tl3020-256mb` at mediatek_filogic:59](https://github.com/openwrt/openwrt/blob/2b2c4ec2be736743986fcc603d7ae24354e63564/package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic#L59). Without `teralink,tl3020-128mb|\` added to that `ubootenv_add_ubi_default` case list, `/etc/fw_env.config` is never generated on this board and `fw_printenv`/`fw_setenv` will not work — while they do on the 256mb sibling.

That file isn't part of this diff, so no suggestion block; it needs `teralink,tl3020-128mb|\` inserted immediately above the 256mb line.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -3254,6 +3254,32 @@ define Device/teralink_tl3020-256mb
 endef
 TARGET_DEVICES += teralink_tl3020-256mb
 
+define Device/teralink_tl3020-128mb
```
openwrt-ai, line 3253: nit: `filogic.mk` keeps `define Device/...` blocks in alphabetical order (`teltonika_rutc50` → `tenbay_wr3000k` → `tenda_be12-pro` → `teralink_tl3020-256mb` → `totolink_x6000r`), so the `-128mb` block belongs *before* the `-256mb` one, not after it. Every other file in this PR (`01_leds`, `02_network`, `platform.sh`, `uboot-mediatek/Makefile`) already places `-128mb` first, so this is the one spot that's out of order.

---

target/linux/mediatek/dts/mt7981b-teralink-tl3020-128mb.dts
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+/*
+ * Copyright (C) 2026 Murad Rabadanov <thetwertyone@gmail.com>
```
openwrt-ai: nit: this address (`thetwertyone@gmail.com`) doesn't match the address used everywhere else in this PR for the same person — the commit's `Signed-off-by: Murad Rabadanov <the21.21@mail.ru>` and the `From:`/`Signed-off-by:` lines in `package/boot/uboot-mediatek/patches/476-add-teralink_tl3020-128mb.patch`. Intentional, or should the copyright line use the same address as the sign-off?

---

package/boot/uboot-mediatek/patches/476-add-teralink_tl3020-128mb.patch
```diff
@@ -0,0 +1,371 @@
+From 0000000000000000000000000000000000000001 Mon Sep 17 00:00:00 2001
```
openwrt-ai: nit: `0000000000000000000000000000000000000001` is a hand-written placeholder hash. The sibling patch this one is modelled on, [475-add-teralink_tl3020-256mb.patch:1](``https://github.com/openwrt/openwrt/blob/2b2c4ec2be736743986fcc603d7ae24354e63564/package/boot/uboot-mediatek/patches/475-add-teralink_tl3020-256mb.patch#L1),`` carries a real commit hash, as does essentially every other patch in this directory. Regenerating the header from an actual commit would keep the metadata meaningful.

---

target/linux/mediatek/dts/mt7981b-teralink-tl3020.dtsi
```diff
@@ -0,0 +1,368 @@
...
+
+	band@1 {
+		reg = <1>;
+		nvmem-cells = <&macaddr_config_60a 0>;
+		nvmem-cell-names = "mac-address";
+	};
+};
+
+&xhci {
+	mediatek,u3p-dis-msk = <0x1>;
```
openwrt-ai, line 366: nit: `mediatek,u3p-dis-msk = <0x1>` disables the USB3 (SS) port, which matches the commit message's hardware spec line `USB : 1x USB 2.0 Type-A` — but the testing paragraph of the same commit message (and the PR body) claims "the USB3 host controller ... confirmed working". Those two statements disagree; the test note probably wants to read "USB host controller confirmed working".

---

openwrt-ai (commented):
Reviewed 1 new commit. The DTS split into `mt7981b-teralink-tl3020.dtsi` is a clean move of the 256mb content (only addition being the `ubi:` label on `partition@680000`), and the U-Boot patch mirrors `475-add-teralink_tl3020-256mb.patch` with only the name and ubi size changed. Inline comments below — one functional gap plus a few nits.

---

target/linux/mediatek/dts/mt7981b-teralink-tl3020-128mb.dts
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+/*
+ * Copyright (C) 2026 Murad Rabadanov <the21.21@mail.ru>
+ */
+
+/dts-v1/;
+#include "mt7981b-teralink-tl3020.dtsi"
```
openwrt-ai, line 7: nit: the commit message describes the model for this split as "the Zbtlink ZBT-Z8103AX 128mb/256mb devices", but no such variants exist in tree — the two devices are `zbtlink_zbt-z8103ax` (64M NAND, see [`mt7981b-zbtlink-zbt-z8103ax.dts:12-15`](https://github.com/openwrt/openwrt/blob/1ab10bb606a1c72dd806030c131dd3b45e977473/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8103ax.dts#L12-L15)) and `zbtlink_zbt-z8103ax-c`, neither carrying a `DEVICE_VARIANT` of `128mb`/`256mb`. The PR body describes the same thing as the "ZBT-Z8103AX 64mb/full-size variants", which is accurate; the commit message is the one that's off. Worth aligning the commit message with the PR body before merge, since the commit message is what stays in the history.

---

openwrt-ai (commented):
Reviewed the force-pushed commit (`2b2c4ec` → `1ab10bb`). All five findings from the previous round are addressed:

- `teralink,tl3020-128mb` added to the `ubootenv_add_ubi_default` case in `uboot-envtools/files/mediatek_filogic`
- `Device/teralink_tl3020-128mb` moved ahead of the `-256mb` block in `filogic.mk` (the `-256mb` block is otherwise byte-identical to before — pure reorder)
- DTS copyright address now matches the `Signed-off-by:`
- placeholder `0000...0001` hash in `476-add-teralink_tl3020-128mb.patch` replaced with a real one (same locally-generated form as the sibling `475-` patch, which is the in-tree norm here)
- commit message "USB3 host controller" → "the USB host controller", consistent with `mediatek,u3p-dis-msk = <0x1>` and the USB 2.0 spec line

Registration parity with the 256mb sibling is now complete — every list the `-256mb` device appears in (`filogic.mk`, `01_leds`, `02_network`, `platform.sh`, `uboot-mediatek/Makefile` incl. `UBOOT_TARGETS`, `uboot-envtools`) has a matching `-128mb` entry. CI is green on `1ab10bb`.

One remaining nit inline.

---

openwrt-ai (commented):
Reviewed 1 new commit (`1ab10bb` → `3dc8fee`); no new issues found.

The amend is commit-message-only — `git diff 1ab10bb..3dc8fee` is empty, so every file in the PR is byte-identical to the version reviewed in the previous round.

The one wording change resolves the last outstanding nit: "Zbtlink ZBT-Z8103AX 128mb/256mb devices" → "Zbtlink ZBT-Z8103AX 64mb/full-size devices". That now matches the tree — [`mt7981b-zbtlink-zbt-z8103ax.dts:13-16`](https://github.com/openwrt/openwrt/blob/3dc8fee24201d38f47228895a017bebd821c3231/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8103ax.dts#L13-L16) overrides `&ubi` to fit 64M NAND while [`mt7981b-zbtlink-zbt-z8103ax-c.dts`](https://github.com/openwrt/openwrt/blob/3dc8fee24201d38f47228895a017bebd821c3231/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8103ax-c.dts) inherits the full-size partition from the shared `.dtsi`, which is exactly the pattern this PR applies to the TL3020 pair.

CI on `3dc8fee` is pending (nothing failed), so no CI-grounded comments.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -3228,6 +3228,32 @@ define Device/tenda_be12-pro
 endef
 TARGET_DEVICES += tenda_be12-pro
 
+define Device/teralink_tl3020-128mb
```
aiamadeus: Can a shared Device definition be created with the 256mb variant?

unisequence: Fixed

target/linux/mediatek/image/filogic.mk
```diff
@@ -3250,6 +3250,14 @@ define Device/teralink_tl3020-256mb
   DEVICE_PACKAGES := kmod-usb3 kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware
   ARTIFACTS := preloader.bin bl31-uboot.fip
   ARTIFACT/preloader.bin := mt7981-bl2 spim-nand-ddr3-1866
+  ARTIFACT/bl31-uboot.fip := mt7981-bl31-uboot teralink_tl3020-128mb
+endef
+TARGET_DEVICES += teralink_tl3020-128mb
+
+define Device/teralink_tl3020-256mb
+  $(Device/teralink_tl3020-128mb)
```
openwrt-ai: nit: this deduplicates by inheriting from a device that is itself registered in `TARGET_DEVICES`, so `teralink_tl3020-256mb` transiently carries `DEVICE_VARIANT := 128mb` and `ARTIFACT/bl31-uboot.fip := ... teralink_tl3020-128mb` until the following lines override them, and it hard-couples the 256mb device to the 128mb one — if the 128mb device is ever dropped or renamed, `$(Device/teralink_tl3020-128mb)` expands to nothing and the 256mb device silently loses `DEVICE_VENDOR`/`DEVICE_MODEL`/all image recipes with no make error.

The convention in this very target is a shared base that is *not* added to `TARGET_DEVICES`, with thin per-variant blocks inheriting from it: [`Device/buffalo_wsr` at mt7622.mk:126](https://github.com/openwrt/openwrt/blob/69dd95f270b72f405157d150a4dfd4df648f61ab/target/linux/mediatek/image/mt7622.mk#L126), [`Device/dlink_eagle-pro-ai-ax3200-a1` at mt7622.mk:186](https://github.com/openwrt/openwrt/blob/69dd95f270b72f405157d150a4dfd4df648f61ab/target/linux/mediatek/image/mt7622.mk#L186), [`Device/elecom_wrc-gst` at mt7622.mk:237](https://github.com/openwrt/openwrt/blob/69dd95f270b72f405157d150a4dfd4df648f61ab/target/linux/mediatek/image/mt7622.mk#L237). That is also exactly the shape you already chose for the DTS in this PR — a shared `mt7981b-teralink-tl3020.dtsi` plus two thin `.dts` files — so mirroring it here would keep the two halves consistent.

Concretely: a `define Device/teralink_tl3020` holding everything common (`DEVICE_VENDOR`, `DEVICE_MODEL`, `DEVICE_DTS_DIR`, the UBI/kernel/image recipes, `DEVICE_PACKAGES`, `ARTIFACTS`, `ARTIFACT/preloader.bin`) with no `TARGET_DEVICES +=` line, and `-128mb` / `-256mb` each reduced to `$(Device/teralink_tl3020)` + `DEVICE_VARIANT` + `DEVICE_DTS` + `ARTIFACT/bl31-uboot.fip`. That satisfies the shared-definition request from the earlier review without either variant depending on the other.

No suggestion block here because the restructuring spans lines outside this hunk.

---

openwrt-ai (commented):
Reviewed 1 new commit (`3dc8fee` → `69dd95f`).

`git diff 3dc8fee..69dd95f` touches only `target/linux/mediatek/image/filogic.mk`: the `Device/teralink_tl3020-256mb` block is collapsed to `$(Device/teralink_tl3020-128mb)` plus the three variant-specific settings, addressing the "can a shared Device definition be created" request. Every other file in the PR — both `.dts`, the `.dtsi`, `01_leds`, `02_network`, `platform.sh`, `uboot-mediatek/Makefile`, `uboot-envtools/files/mediatek_filogic`, and `476-add-teralink_tl3020-128mb.patch` — is byte-identical to the previously reviewed version, and the commit message is unchanged too.

The dedup itself is functionally equivalent: `include/image.mk` expands the block via `$(call Device/$(1),$(1))`, and nesting `$(Device/...)` inside another `define` does not add an expansion level, so the `$$(KDIR)` / `$$(DEVICE_DTS)` escaping in the inherited recipes still resolves the same way. The three overrides (`DEVICE_VARIANT`, `DEVICE_DTS`, `ARTIFACT/bl31-uboot.fip`) come after the inherited assignments, so the 256mb device ends up with the same values it had before. Inheriting from a concrete `TARGET_DEVICES` entry also has in-tree precedent (e.g. `Device/linksys_e7350` ← `Device/belkin_rt1800` in `ramips/image/mt7621.mk`).

One nit inline about matching the mt7622.mk shared-base shape instead.

The single failing check (`build-pr-profile / check-and-build`) is infrastructure, not this PR — the runner got HTTP 429 fetching `peter-evans/find-comment` and never started the build:

```
##[warning]Failed to download action 'https://codeload.github.com/peter-evans/find-comment/tar.gz/...'. Error: Response status code does not indicate success: 429 (Too Many Requests).
##[error]Failed to download archive '...' after 3 attempts.
```

All build/kernel/formality checks on `69dd95f` are green.

---

unisequence:
Applied the shared-base restructuring as suggested.

Caught one issue while build-testing: `KERNEL_INITRAMFS`/`IMAGE/sysupgrade.itb` in the base use `:=`, which captures `$(DEVICE_DTS)` immediately at that line, not lazily. With `$(Device/teralink_tl3020)` placed before the `DEVICE_DTS` override, it captured an empty value and broke the build (`image-.dtb`).

Fixed by moving `DEVICE_DTS` above the inherited block in both variants. Build-verified locally — both `-128mb` and `-256mb` produce correct DTB names now.

target/linux/mediatek/image/filogic.mk
```diff
@@ -3250,6 +3248,20 @@ define Device/teralink_tl3020-256mb
   DEVICE_PACKAGES := kmod-usb3 kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware
   ARTIFACTS := preloader.bin bl31-uboot.fip
   ARTIFACT/preloader.bin := mt7981-bl2 spim-nand-ddr3-1866
+endef
+
+define Device/teralink_tl3020-128mb
+  DEVICE_DTS := mt7981b-teralink-tl3020-128mb
+  $(Device/teralink_tl3020)
```
openwrt-ai: nit: the diagnosis in [your comment](https://github.com/openwrt/openwrt/pull/24694#issuecomment-5324650068) is right — `Device/Build` eval's the expanded block top-to-bottom, so the `:=` on `KERNEL_INITRAMFS` / `IMAGE/sysupgrade.itb` in the base snapshots `$(DEVICE_DTS)` at that point — but the resulting ordering is now load-bearing and silent: `DEVICE_DTS` above `$(Device/teralink_tl3020)` reads like a stray line, and anyone later "tidying" it below the include to match every other shared-base user in this target gets `image-.dtb` with no make error. Same applies to lines 3262-3263 in the `-256mb` block.

The in-tree idiom for exactly this case is to make the affected recipes recursively expanded in the base instead, and keep the base include first: [`KERNEL_INITRAMFS = kernel-bin | lzma | fit lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb ...` at mt7622.mk:137-139](https://github.com/openwrt/openwrt/blob/d7cd84c321407262f80a2ab2c973d6de0604d161/target/linux/mediatek/image/mt7622.mk#L137-L139) in `Device/buffalo_wsr` uses `=` precisely because its variants set `DEVICE_DTS` *after* `$(Device/buffalo_wsr)`; [`IMAGE/factory.bin =` at mt7622.mk:140](https://github.com/openwrt/openwrt/blob/d7cd84c321407262f80a2ab2c973d6de0604d161/target/linux/mediatek/image/mt7622.mk#L140) does the same for `$$(DEVICE_MODEL)`, while the recipes in that base that only reference already-set variables stay `:=`.

Switching `KERNEL_INITRAMFS :=` (line 3243) and `IMAGE/sysupgrade.itb :=` (line 3245) to `=` would let both variant blocks read `$(Device/teralink_tl3020)` first, then `DEVICE_VARIANT` / `DEVICE_DTS` / `ARTIFACT/bl31-uboot.fip` — matching `buffalo_wsr`, `dlink_eagle-pro-ai-ax3200-a1` and `elecom_wrc-gst`, and removing the ordering trap. No suggestion block since the `:=` lines sit outside this hunk.

---

openwrt-ai (commented):
Reviewed 1 new commit (`69dd95f` → `d7cd84c`).

`git diff 69dd95f..d7cd84c` touches only `target/linux/mediatek/image/filogic.mk`, and the commit message is byte-identical. Every other file in the PR — both `.dts`, the shared `.dtsi`, `01_leds`, `02_network`, `platform.sh`, `uboot-mediatek/Makefile`, `uboot-envtools/files/mediatek_filogic`, and `476-add-teralink_tl3020-128mb.patch` — is unchanged from the previously reviewed version.

The change adopts the shared-base shape from the last round's nit: a new `define Device/teralink_tl3020` holding the common settings with no `TARGET_DEVICES +=` line, and `-128mb` / `-256mb` reduced to the include plus `DEVICE_VARIANT`, `DEVICE_DTS` and `ARTIFACT/bl31-uboot.fip`. The 256mb device no longer depends on the 128mb one, which was the substance of that comment. The `DEVICE_DTS`-before-include ordering is explained in the PR conversation and is genuinely required given the `:=` recipes in the base; one nit inline about expressing that more robustly.

All 22 checks on `d7cd84c` are green, including the mediatek/filogic kernel + package builds and all three FormalityCheck jobs, so no CI-grounded comments.

---

openwrt-ai (commented):
Reviewed 1 new commit (`d7cd84c` → `6946ce2`); no new issues found.

`git diff d7cd84c..6946ce2` touches only `target/linux/mediatek/image/filogic.mk` (4 lines changed), and the commit message is byte-identical. Every other file in the PR — both `.dts`, the shared `.dtsi`, `01_leds`, `02_network`, `platform.sh`, `uboot-mediatek/Makefile`, `uboot-envtools/files/mediatek_filogic`, and `476-add-teralink_tl3020-128mb.patch` — is unchanged from the previously reviewed version.

The change resolves the last round's nit exactly as the in-tree idiom does it: `KERNEL_INITRAMFS` and `IMAGE/sysupgrade.itb` in `define Device/teralink_tl3020` switch from `:=` to `=`, and both variant blocks now read `$(Device/teralink_tl3020)` first, then `DEVICE_VARIANT` / `DEVICE_DTS` / `ARTIFACT/bl31-uboot.fip`. That matches [`Device/buffalo_wsr` at mt7622.mk:137-140](https://github.com/openwrt/openwrt/blob/6946ce2f8236ad4c60b748fc755ae6dc4288158a/target/linux/mediatek/image/mt7622.mk#L137-L140) and removes the silent ordering dependency.

Deferred expansion is safe for both recipes: `KERNEL_INITRAMFS` is exported per device via [`Device/ExportVar`](https://github.com/openwrt/openwrt/blob/6946ce2f8236ad4c60b748fc755ae6dc4288158a/include/image.mk#L627-L630) as a target-specific `:=` assignment, and `IMAGE/sysupgrade.itb` is expanded with a single `$` inside [`Device/Build/image` at image.mk:784](https://github.com/openwrt/openwrt/blob/6946ce2f8236ad4c60b748fc755ae6dc4288158a/include/image.mk#L784) — both after `$(call Device/$(1),$(1))` has set `DEVICE_DTS`, so `$(firstword $(DEVICE_DTS))` resolves to the right per-variant value. `KERNEL`, `ARTIFACT/preloader.bin` and the remaining `:=` lines in the base reference nothing set later, so they correctly stay immediate.

Also re-checked the refactor against the pre-PR definition: the flattened `teralink_tl3020-256mb` on `main` (`cf198cc`, filogic.mk:3253-3277) and the inherited one here end up with the same value for every variable, so the 256mb device is unaffected.

All 22 checks on `6946ce2` are green, including the mediatek/filogic kernel + package builds and all three FormalityCheck jobs, so no CI-grounded comments.

---

unisequence:
Hi @hauke, gentle ping on this one. All review feedback addressed, CI is green (22/22), and it's hardware-tested on a real unit. Let me know if there's anything else you'd like me to address before merge.

unisequence:
@dangowrt could you take a look at this one when you have time? Thanks!

unisequence:
maybe @jonasjelonek ?
Can you review this PR?

PR #24699, opened 2026-08-12
mpc85xx: red-15w: use nvmem for wifi MAC

neheb:
Userspace handling is deprecated.

ping @Shine- 

No idea what the proper configuration is for the uboot environment. Other devices using a size of 0x100000 for the environment tend to have a redundant one of size 0x20000.

### DIFF
--- a/target/linux/mpc85xx/files/arch/powerpc/boot/dts/red-15w-rev1.dts
+++ b/target/linux/mpc85xx/files/arch/powerpc/boot/dts/red-15w-rev1.dts
@@ -13,6 +13,7 @@
 		led-failsafe = &system_red;
 		led-running = &system_green;
 		led-upgrade = &system_red;
+		label-mac-device = &enet0;
 	};
 
 	memory {
@@ -157,6 +158,14 @@
 				partition@100000 {
 					reg = <0x100000 0x100000>;
 					label = "u-boot-env";
+
+					nvmem-layout {
+						compatible = "u-boot,env";
+						env-size = <0x4000>;
+
+						macaddr_uboot_ethaddr: ethaddr {
+						};
+					};
 				};
 
 				partition@200000 {
@@ -222,6 +231,13 @@
 				 <0x1000000 0x0 0x0
 				  0x1000000 0x0 0x0
 				  0x0 0x100000>;
+
+			wifi@0,0 {
+				compatible = "pci168c,0030";
+				reg = <0x0000 0 0 0 0>;
+				nvmem-cells = <&macaddr_uboot_ethaddr>;
+				nvmem-cell-names = "mac-address";
+			};
 		};
 	};
 };
--- a/target/linux/mpc85xx/files/arch/powerpc/boot/dts/red-15w-rev1.dts
+++ b/target/linux/mpc85xx/files/arch/powerpc/boot/dts/red-15w-rev1.dts
@@ -157,6 +157,14 @@
 				partition@100000 {
 					reg = <0x100000 0x100000>;
 					label = "u-boot-env";
+
+					nvmem-layout {
+						compatible = "u-boot,env-redundant-bool";
+						env-size = <0x20000>;
+
+						macaddr_uboot_ethaddr: ethaddr {
+						};
+					};
 				};
 
 				partition@200000 {
@@ -222,6 +230,13 @@
 				 <0x1000000 0x0 0x0
 				  0x1000000 0x0 0x0
 				  0x0 0x100000>;
+
+			wifi@0,0 {
+				compatible = "pci168c,0030";
+				reg = <0x0000 0 0 0 0>;
+				nvmem-cells = <&macaddr_uboot_ethaddr>;
+				nvmem-cell-names = "mac-address";
+			};
 		};
 	};
 };
--- a/target/linux/mpc85xx/base-files/etc/hotplug.d/ieee80211/10-fix-wifi-mac
+++ b/target/linux/mpc85xx/base-files/etc/hotplug.d/ieee80211/10-fix-wifi-mac
@@ -23,7 +23,4 @@ extreme-networks,ws-ap3825i)
 ocedo,panda)
 	mtd_get_mac_ascii uboot-env0 wmac$(($PHYNBR + 1)) > /sys${DEVPATH}/macaddress
 	;;
-sophos,red-15w-rev1)
-	mtd_get_mac_ascii u-boot-env ethaddr > /sys${DEVPATH}/macaddress
-	;;
 esac


target/linux/mpc85xx/files/arch/powerpc/boot/dts/red-15w-rev1.dts
```diff
@@ -223,6 +231,13 @@
 				  0x1000000 0x0 0x0
 				  0x0 0x100000>;
 		};
+
+		wifi@0,0 {
+			compatible = "pci168c,0030";
+			reg = <0x0000 0 0 0 0>;
+			nvmem-cells = <&macaddr_uboot_ethaddr>;
+			nvmem-cell-names = "mac-address";
+		};
```
openwrt-ai: `wifi@0,0` is a sibling of `pcie@0`, i.e. a direct child of the host-bridge node, so it never gets associated with the endpoint. The card sits on the root port's secondary bus, so the node has to live inside `pcie@0` — that is what every other mpc85xx board does, e.g. [`br200-wp.dts:293`](``https://github.com/openwrt/openwrt/blob/169daf04a2a676006662043768e746df7165808c/target/linux/mpc85xx/files/arch/powerpc/boot/dts/br200-wp.dts#L285-L301``) and [`ws-ap3715i.dts:187`](``https://github.com/openwrt/openwrt/blob/169daf04a2a676006662043768e746df7165808c/target/linux/mpc85xx/files/arch/powerpc/boot/dts/ws-ap3715i.dts#L180-L192).`` As written, `of_node` is never attached to the ath9k device and the `nvmem-cells` phandle is simply ignored — combined with the removal of the hotplug entry the wifi MAC silently falls back to the card EEPROM value.

```suggestion

			wifi@0,0 {
				compatible = "pci168c,0030";
				reg = <0x0000 0 0 0 0>;
				nvmem-cells = <&macaddr_uboot_ethaddr>;
				nvmem-cell-names = "mac-address";
			};
		};
```

---

target/linux/mpc85xx/files/arch/powerpc/boot/dts/red-15w-rev1.dts
```diff
@@ -157,6 +157,14 @@
 				partition@100000 {
 					reg = <0x100000 0x100000>;
 					label = "u-boot-env";
+
+					nvmem-layout {
+						compatible = "u-boot,env-redundant-bool";
+						env-size = <0x20000>;
```
openwrt-ai: Re. your question in the PR description — this pair is a guess that fails silently if wrong, so it is worth confirming against the actual device before this leaves draft.

`u-boot,env-redundant-bool` selects `struct u_boot_env_image_redundant`, i.e. a 5-byte header (crc32 + flags) versus 4 bytes for plain `u-boot,env`, and the CRC is computed over `env-size - 5` bytes. If either the format or `env-size` is wrong, [`u_boot_env_parse()` bails out with `-EINVAL` on the CRC mismatch](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/nvmem/layouts/u-boot-env.c#L151-L156) and registers no cells at all. Since this commit also drops the `mtd_get_mac_ascii` fallback, the failure mode is "wifi keeps its EEPROM MAC and nothing obvious in the log points at the DT" — a regression relative to today's behaviour.

Unlike the other mpc85xx boards there is no entry for `sophos,red-15w-rev1` in [`package/boot/uboot-tools/uboot-envtools/files/mpc85xx`](https://github.com/openwrt/openwrt/blob/169daf04a2a676006662043768e746df7165808c/package/boot/uboot-tools/uboot-envtools/files/mpc85xx), so there is no in-tree cross-check for the env size or for whether the env is redundant at all. Note also that the original layout comment in this file calls the partition a single "NAND U-Boot Environment" region with no second copy, which would argue for plain `u-boot,env`. A `fw_printenv`/hexdump of `/dev/mtd1` from a real unit would settle both the format and the size.

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the change; two inline notes on the DTS.

---

neheb:
given that the erasesize is 0x20000, this sounds accurate:

```
dev:    size   erasesize  name
mtd0: 00100000 00020000 "u-boot"
mtd1: 00100000 00020000 "u-boot-env"
mtd2: 00100000 00020000 "provisioning"
mtd3: 01900000 00020000 "sophos-os1"
mtd4: 00800000 00020000 "kernel"
mtd5: 01100000 00020000 "ubi"
mtd6: 01900000 00020000 "sophos-os2"
```

openwrt-ai (commented):

The only change since the last review moves `wifi@0,0` inside `pcie@0`, which resolves the earlier node-placement finding — the structure now matches [`msm460.dts:196-203`](``https://github.com/openwrt/openwrt/blob/659d71e5c97490620f24005540cbaeef9a9a3ec7/target/linux/mpc85xx/files/arch/powerpc/boot/dts/msm460.dts#L196-L203``) and [`br200-wp.dts:285-300`](``https://github.com/openwrt/openwrt/blob/659d71e5c97490620f24005540cbaeef9a9a3ec7/target/linux/mpc85xx/files/arch/powerpc/boot/dts/br200-wp.dts#L285-L300).`` The commit message matches the change.

The earlier thread about `u-boot,env-redundant-bool` vs plain `u-boot,env` is still open: your `/proc/mtd` dump confirms the `0x20000` erase size (so `env-size` is plausible), but it does not tell us whether U-Boot writes the 5-byte redundant header. If it does not, [`u_boot_env_parse()` fails the CRC check and registers no cells at all](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/nvmem/layouts/u-boot-env.c#L148-L153), and since this commit also drops the `mtd_get_mac_ascii` fallback the wifi would silently keep its EEPROM MAC. A `hexdump -C -n 16 /dev/mtd1` from a real unit settles it.

---

Shine-:
Sorry for delay, just returned from vacation. Will review/test within the week, please bear with me.

Shine-:
Checked my `mtd1` dump from the stock unit, the Uboot Env partition is zero-padded up to `0x3fff`, the rest is filled with `0xff`. There's a 4-byte header before the env variables start. So I suspect there's no redundant env and the env size is `0x4000` (16k).
I can check for "wrong CRC" message with Uboot Envtools ~~tomorrow~~ later using various configurations.

Shine-:
/etc/fw_env.config
```
/dev/mtd1 0x0 0x4000 0x4000 1
```
seems to be the correct config (at least this doesn't give a CRC error when using `fw_printenv`).

Edit (due to erase size, otherwise writing the env will throw an error):
```
/dev/mtd1 0x0 0x4000 0x20000 1
```

neheb:
Changed to non redundant u-boot,env and fixed the size.

target/linux/mpc85xx/files/arch/powerpc/boot/dts/red-15w-rev1.dts
```diff
@@ -157,6 +157,14 @@
 				partition@100000 {
 					reg = <0x100000 0x100000>;
 					label = "u-boot-env";
+
+					nvmem-layout {
+						compatible = "u-boot,env";
+						env-size = <0x4000>;
```
openwrt-ai, line 164: nit: now that the env geometry is pinned down (`u-boot,env` single format, 0x4000 payload — matching @Shine-'s working `/dev/mtd1 0x0 0x4000 0x4000 1`), should `sophos,red-15w-rev1` also get an entry in [`package/boot/uboot-tools/uboot-envtools/files/mpc85xx`](https://github.com/openwrt/openwrt/blob/4868a59368cfb6fdd73770e6ef85ac3d9b2ffdfe/package/boot/uboot-tools/uboot-envtools/files/mpc85xx)? Every other mpc85xx board with a U-Boot env partition is listed there (`enterasys,ws-ap3715i`, `extreme-networks,ws-ap3825i`, `ocedo,panda`, `watchguard,firebox-t10`/`t15`, `watchguard,xtm330`, `aerohive,hiveap-330`), and this board is the only one that isn't — which is why there was no in-tree cross-check for the size in the first place. It would look like:

```
sophos,red-15w-rev1)
	ubootenv_add_uci_config "$(find_mtd_part 'u-boot-env')" "0x0" "0x4000" "0x20000"
	;;
```

(erase size `0x20000` per your `/proc/mtd` dump.) Out of scope for a commit titled "use nvmem for wifi MAC", so a follow-up patch in the same series is probably the cleaner shape — just flagging it so the information gathered in this thread doesn't get lost.

---

openwrt-ai (commented):
Reviewed 1 new commit; no new issues found. The commit message still matches the change.

The only delta since the last review is `u-boot,env-redundant-bool`/`0x20000` → `u-boot,env`/`0x4000`, which resolves the open format/size thread. @Shine-'s findings line up with the kernel side: a 4-byte header is exactly `struct u_boot_env_image_single` (`crc32` only, vs. 5 bytes for the redundant variants), and a CRC-clean `fw_printenv` with `/dev/mtd1 0x0 0x4000 0x4000 1` confirms U-Boot computes the CRC over `0x4000 - 4` bytes — which is what [`u_boot_env_parse()` will now do](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/nvmem/layouts/u-boot-env.c#L131-L146) given `env-size = <0x4000>` from the [backported `env-size` support](``https://github.com/openwrt/openwrt/blob/4868a59368cfb6fdd73770e6ef85ac3d9b2ffdfe/target/linux/generic/backport-6.18/010-v6.19-nvmem-layouts-u-boot-env-add-optional-env-size-property.patch).`` `CONFIG_NVMEM_LAYOUT_U_BOOT_ENV=y` is set in [`target/linux/mpc85xx/config-6.18:188`](https://github.com/openwrt/openwrt/blob/4868a59368cfb6fdd73770e6ef85ac3d9b2ffdfe/target/linux/mpc85xx/config-6.18#L188), so the layout driver is present.

One nit inline about `uboot-envtools`. CI on `4868a59` is green.

---

Shine-:
Tested 4868a59368cfb6fdd73770e6ef85ac3d9b2ffdfe against current Main:
* eth0 MAC is correct (`ethaddr` env variable)
* eth1 MAC (+1) is correct (`eth1addr` env variable)
* WiFi MAC (same as eth0) is correct

LGTM.

Shine-:
>  so a follow-up patch in the same series is probably the cleaner shape — just flagging it so the information gathered in this thread doesn't get lost.

Was planning to do that independently anyway, so here it is: #24939

neheb:
great. you think there's any point in adding nvmem overrides to eth0/1 as well?

Shine-:
Don't think so, they're already properly populated as-is, no need to double-assign them.

neheb:
It's not really double assigned. It's a single of_get_mac_address call in the driver. My thinking is that it's clearer to specify them anyway.

Also if using label-mac-device (missing here), nvmem is required. Speaking of which, is this MAC the one printed on the device?

Shine-:
Both are printed on the device, `eth1addr` as WAN MAC, `ethaddr` as LAN MAC.

Shine-:
If you want to reassign them from NVMEM, you should use `eth1addr` for eth1, instead of "`ethaddr` + 1".
Though I still think there's no need to reassign the MACs via DTS, when they're already correct in the first place.

neheb:
The uboot-env driver does not support variables other than ethaddr.

Shine-:
Then that's even more reason to leave the eth0/eth1 MACs as-is, since they're already correctly assigned.

neheb:
But then label-mac-device wouldn't work.

Shine-:
Devil's advocate: Two different MACs are printed on the label. Which is the correct "label-mac-device" then? WAN MAC (`eth1addr`) is printed first, even.

target/linux/mpc85xx/files/arch/powerpc/boot/dts/red-15w-rev1.dts
```diff
@@ -13,6 +13,9 @@
 		led-failsafe = &system_red;
 		led-running = &system_green;
 		led-upgrade = &system_red;
+		label-mac-device = &enet0;
+		/delete-property/ ethernet0;
+		/delete-property/ ethernet1;
```
openwrt-ai: `label-mac-device` does not need the aliases gone: U-Boot's [`fdt_fixup_ethernet()`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/boot/fdt_support.c#L703-L706) already writes `mac-address` into the aliased node, and that is exactly the property [`get_mac_label_dt()`](https://github.com/openwrt/openwrt/blob/126157723c45e26735e99c37afc15143a7b3243a/package/base-files/files/lib/functions/system.sh#L25-L26) reads back out of `/proc/device-tree`, so it works on this board as-is. Deleting them drops the bootloader-supplied MAC as a fallback and makes *both* ethernet ports depend on the env CRC — if `u_boot_env_parse()` ever fails the layout registers no cells and [`of_nvmem_cell_get()` returns `-EPROBE_DEFER` permanently](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/nvmem/core.c#L1474-L1478), so gianfar never probes and the unit has no network at all rather than just a wrong wifi MAC. Is giving up that fallback intentional?

---

openwrt-ai (commented):
Reviewed 1 new commit.

---

target/linux/mpc85xx/files/arch/powerpc/boot/dts/red-15w-rev1.dts
openwrt-ai: handled separately in #24939, thanks

---

Shine-:
If you want to consider LAN (`eth0`) the `label-mac-device` - which is coherent, given that `eth0` is the one that gets assigned its value from the `ethaddr` env variable, IMO you should only assign the `eth0` MAC via DTS, and leave `eth1` alone. Especially since you say that the `eth1addr` var isn't readable from DTS either way - so keep `eth1` as pre-assigned (correctly) by Uboot. In addition, make sure also `eth0` can fall back to whatever Uboot assigns.

What we want in the DTS in the first place is `label-mac-device`, right? All eth MACs themselves are already correct.

neheb:
OK.

target/linux/mpc85xx/files/arch/powerpc/boot/dts/red-15w-rev1.dts
```diff
@@ -13,6 +13,8 @@
 		led-failsafe = &system_red;
 		led-running = &system_green;
 		led-upgrade = &system_red;
+		label-mac-device = &enet0;
+		/delete-property/ ethernet0;
```
openwrt-ai: This disables the very feature the commit adds: [`get_mac_label_dt()`](https://github.com/openwrt/openwrt/blob/55a0abc6ddecfaa1bd10b90ae33961fd4c4cb4db/package/base-files/files/lib/functions/system.sh#L25-L26) reads `mac-address`/`local-mac-address` out of `/proc/device-tree`, and the only writer of those properties on this board is U-Boot's [`fdt_fixup_ethernet()`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/boot/fdt_support.c#L703-L706), which walks the `ethernet*` aliases — an `nvmem-cells` phandle is resolved in-kernel and never materialises as a DT property. With no `ucidef_set_label_macaddr` for `sophos,red-15w-rev1` in `target/linux/mpc85xx/base-files/` either, `get_mac_label` ends up empty. Keeping the alias also restores the bootloader fallback for eth0 that @Shine- asked for in [#issuecomment-5462006720](https://github.com/openwrt/openwrt/pull/24699#issuecomment-5462006720).

```suggestion
```

---

openwrt-ai (commented):
## Commit checks

- `55a0abc` "mpc85xx: red-15w: use nvmem for MAC addresses" — the body says "Remove ethernet aliases" (plural) and that the bootloader "passing them no longer works and nvmem does", but only `ethernet0` is deleted; `ethernet1` stays and `enet1` has no nvmem cell, so its MAC still comes from U-Boot. Reword to state that only the `ethernet0` alias is dropped and that eth1 keeps the bootloader-assigned MAC.

---

target/linux/mpc85xx/files/arch/powerpc/boot/dts/red-15w-rev1.dts
neheb: go home. you're drunk. on certain platforms, this need deletion to allow nvmem overrides.

Shine-: In fact, this line caused the MAC addresses to reverse (WAN<->LAN) and ethernet connectivity broke completely for _both_ I/F. Had to recover via serial.

Removing that line, however, results in what we are trying to achieve:
* Ethernet is working again
* MAC addresses are correct
* `get-label-mac` returns the correct (LAN) MAC address

So yes, I suggest **not** using the `/delete-property/` line. Then everything is fine - MACs are correctly assigned by U-Boot and Label-MAC is correct, too.

neheb: OK.

openwrt-ai (commented):

---

target/linux/mpc85xx/files/arch/powerpc/boot/dts/red-15w-rev1.dts
openwrt-ai: fixed, thanks

---

target/linux/mpc85xx/files/arch/powerpc/boot/dts/red-15w-rev1.dts
```diff
@@ -92,6 +93,9 @@
 		enet0: ethernet@b0000 {
 			phy-connection-type = "rgmii-id";
 
+			nvmem-cells = <&macaddr_uboot_ethaddr>;
+			nvmem-cell-names = "mac-address";
+
```
Shine-: These could then also go away, since they're ignored with `/delete-property/` gone - Uboot-assigned MAC takes precedence, right?
WiFi and Label-MAC is still fine with these lines removed (just tested).

neheb: Sure.

Shine-:
Thanks!

For now at least, I'd say this is the minimum required to get rid of rid of user space assignment for the WiFi MAC, as well as define the previously missing label MAC.

Note that your commit message ("The ethaddr MAC shows up on the label on the device.") might be a bit misleading - in fact, both MACs are printed on the label, with WAN (`eth1addr`) first, even. But - confirmed for both of my devices - LAN (`ethaddr`) is the lower numbered one and also the one used for WiFi.

LGTM!

neheb:
ping @jonasjelonek

Shine-:
Also related, #24939, please. Thanks.

openwrt-ai (commented):

---

jonasjelonek:

PR #24700, opened 2026-08-12
wolfssl: update to 5.9.2

sanderpqr:
Update WolfSSL to the latest version, 5.9.2

For changes, see:
https://github.com/wolfSSL/wolfssl/releases/tag/v5.9.2-stable

Ping @PolynomialDivision

PR #24701, opened 2026-08-12
[25.12] wolfssl: update to 5.9.2

sanderpqr:
Update WolfSSL to the latest version, 5.9.2

For changes, see:
https://github.com/wolfSSL/wolfssl/releases/tag/v5.9.2-stable

(cherry picked from commit 6bd6eb4a0764c7e271ab5accb889d3288c88fb50 See https://github.com/openwrt/openwrt/pull/24700)

Ping @PolynomialDivision


openwrt-ai (commented):
Reviewed 1 new commit. The backported hunks themselves match PR #24700 verbatim — the `PKG_VERSION`/`PKG_HASH` bump and the refreshed `@@ -4281,7 +4281,7 @@` hunk header in `100-disable-hardening-check.patch` are byte-identical to `6bd6eb4a`, and the remaining differences between the two trees (`--enable-riscv-asm`, the `riscv64` dependency, the `https://` URL) are pre-existing `main`-vs-`openwrt-25.12` divergence, not something this cherry-pick introduced. CI on `c7dd358` is green.

## Commit checks

- `c7dd358` "wolfssl: update to 5.9.2" — the `(cherry picked from commit 6bd6eb4a0764c7e271ab5accb889d3288c88fb50)` trailer does not reference a commit in `openwrt/openwrt`. `6bd6eb4a` is the current head of `sanderpqr/openwrt:wolfssl592`, i.e. the still-open PR #24700; `git merge-base --is-ancestor 6bd6eb4a <main>` is false, and [`PKG_VERSION:=5.9.1` in package/libs/wolfssl/Makefile:11](https://github.com/openwrt/openwrt/blob/fe716b909e63eea7cf3923024141663f64c22a1e/package/libs/wolfssl/Makefile#L11) confirms `main` has not taken the update yet. Since OpenWrt lands PRs through staging trees rather than merging the fork commit as-is, the SHA that ends up on `main` will differ from `6bd6eb4a`, leaving this trailer pointing at a commit that exists nowhere in the repo. Worth re-running `git cherry-pick -x` against the real `main` commit once #24700 is merged, and holding this backport until then so the branches don't diverge in ordering.

- `c7dd358` — the message carries `Signed-off-by: Sander van Deijck <sander@vandeijck.com>` twice, once before and once after the cherry-pick trailer. The second one is redundant for a self-authored cherry-pick; dropping it would leave the usual `S-o-b` / `(cherry picked from …)` shape.

Note: I could not verify `PKG_HASH:=2f4ef3d4fd387a9b3191d36a6316d69116c46ff69bb9583b6c82b36d7b8ca114` against the upstream `v5.9.2-stable` tarball from this session (network access is scoped to `openwrt/openwrt`). It is identical to the hash in #24700, so it needs checking once, on that PR rather than here.

---

PR #24702, opened 2026-08-13
realtek: fix SMP IRQ 7 storm by restoring CEVT_R4K

luizluca:
The transition to generic machine initialization dropped the legacy Realtek setup code. Without it, disabling CEVT_R4K results in an unhandled IRQ 7 (MIPS CP0 timer) storm on secondary VPEs, leading to an "irq 7: nobody cared" kernel panic.

The performance impact is negligible. Once the Realtek timer (timer@3100) is loaded, it seamlessly takes over the tick duties due to its higher rating. The MIPS R4K timer goes completely dormant and only triggers harmlessly when the 32-bit counter overflows (a couple of times per minute), avoiding the panic without core modifications.

Fixes: 7cc31af7bdd4 ("realtek: convert to generic machine initialization")
Assisted-by: Gemini:gemini-3.1-pro

Closes #24674


target/linux/realtek/patches-6.18/300-01-enhance-realtek-platform-config.patch
```diff
@@ -11,12 +11,12 @@ Submitted-by: Birger Koblitz <git@birger-koblitz.de>
...
 --- a/arch/mips/Kconfig
 +++ b/arch/mips/Kconfig
-@@ -700,19 +700,22 @@ config MACH_REALTEK_RTL
+@@ -700,19 +700,23 @@ config MACH_REALTEK_RTL
  	select MACH_GENERIC_CORE
  	select DMA_NONCOHERENT
  	select IRQ_MIPS_CPU
 -	select CSRC_R4K
--	select CEVT_R4K
+ 	select CEVT_R4K
```
openwrt-ai, line 19: nit: this flips `CONFIG_CEVT_R4K` from n to y for all six realtek subtargets, but none of the `target/linux/realtek/*/config-6.18` files gained a `CONFIG_CEVT_R4K=y` line. Other MIPS targets that get the symbol purely via `select` do carry it in their kernel config — [`target/linux/malta/config-6.18:25`](https://github.com/openwrt/openwrt/blob/23352efd6c9ebda50ac77815f96e85f16f76dfb2/target/linux/malta/config-6.18#L25) and [`target/linux/bcm47xx/config-6.18:29`](https://github.com/openwrt/openwrt/blob/23352efd6c9ebda50ac77815f96e85f16f76dfb2/target/linux/bcm47xx/config-6.18#L29) — and realtek already lists comparable prompt-less select-only symbols, e.g. [`CONFIG_MIPS_EXTERNAL_TIMER=y`](https://github.com/openwrt/openwrt/blob/23352efd6c9ebda50ac77815f96e85f16f76dfb2/target/linux/realtek/rtl838x/config-6.18#L146) and `CONFIG_NO_EXCEPT_FILL=y` / `CONFIG_RESET_CONTROLLER=y`.

Nothing breaks at build time (the symbol has no prompt and is forced on by the `select`, which is why the six kernel builds are green), so this is config-file hygiene only.

Worth noting the realtek configs are arguably already stale in the same way: `MACH_REALTEK_RTL` has selected `MIPS_GENERIC` since 7cc31af7bdd4, while [`target/linux/generic/config-6.18:3830`](https://github.com/openwrt/openwrt/blob/23352efd6c9ebda50ac77815f96e85f16f76dfb2/target/linux/generic/config-6.18#L3830) says `# CONFIG_MIPS_GENERIC is not set` and no realtek subtarget config overrides it. So — is omitting these deliberate for this target, or would you rather fold a `kernel_oldconfig` refresh in here?

---

openwrt-ai (commented):
Reviewed 1 new commit.

Verified against `gregkh/linux:v6.18.44` (`target/linux/realtek` is on `patches-6.18`):

- The edited hunk headers are exact — reapplying the patch to a pristine v6.18.44 `arch/mips/Kconfig` and re-diffing reproduces `@@ -700,19 +700,23 @@` and `@@ -1056,6 +1060,37 @@` byte-for-byte, and it applies at fuzz 0.
- The commit message's causal claim holds up: the code dropped by 7cc31af7bdd4 was `rtlsmp_finish()` in `target/linux/realtek/files-6.18/arch/mips/rtl-otto/prom.c`, which explicitly masked IP7 and defused CP0 compare on secondary VPEs *only* when `!IS_ENABLED(CONFIG_CEVT_R4K)` — so restoring the symbol addresses exactly the case that workaround covered.
- The "higher rating takes over" argument checks out: `timer-rtl-otto.c` registers its per-CPU clockevents at rating 400 (via `CPUHP_AP_REALTEK_TIMER_STARTING`, which fires for the boot CPU inside `timer_probe()`, i.e. before `r4k_clockevent_init()`), versus rating 300 for cevt-r4k. Leaving `CSRC_R4K` deselected alongside `MIPS_EXTERNAL_TIMER` is consistent, and `cevt-r4k.c` carries its own `CONFIG_CPU_FREQ` notifier, so the cpufreq-scaling targets (rtl838x/rtl839x) are covered.

One config-hygiene nit inline.

---

plappermaul:
Thanks. Please run the above mentioned config file fixeup and we are good to go.

luizluca:
> Thanks. Please run the above mentioned config file fixeup and we are good to go.

I guess it is OK now. However, there are other missing non-related symbols that I left as it is.

target/linux/realtek/patches-6.18/300-01-enhance-realtek-platform-config.patch
jameywine, line 19: Does this have to be selected globally for everyone even though it is a solution to a problem exclusively on rtl839x? selecting it under `config RTL839X` would make more sense for me.

openwrt-ai (commented):

The only change since `23352ef` is the kernel-config refresh raised in my previous review, and it looks complete:

- `CONFIG_CEVT_R4K=y` was added to all six realtek subtarget configs (`rtl838x`, `rtl839x`, `rtl930x`, `rtl930x_nand`, `rtl931x`, `rtl931x_nand`) — no subtarget missed.
- Each line lands in the correct alphabetical slot for its file (after `BMIPS_CPUFREQ` / `BOARD_SCACHE`, before `CLKSRC_MIPS_GIC` / `CLONE_BACKWARDS`), matching what `kernel_oldconfig` would emit.
- `CSRC_R4K` correctly stays deselected and absent from the configs; the patch hunk headers are unchanged from the version I verified against `gregkh/linux:v6.18.44`.

The remaining staleness I mentioned (`MIPS_GENERIC`) is explicitly deferred in [your comment](https://github.com/openwrt/openwrt/pull/24702#issuecomment-5288657450), so nothing further from me there. @jameywine's open question about scoping the `select` to `config RTL839X` rather than `MACH_REALTEK_RTL` is the one item still outstanding — I have no evidence to add either way, so I'm leaving that to you two.

---

target/linux/realtek/patches-6.18/300-01-enhance-realtek-platform-config.patch
luizluca: I'm not sure IRQ 7 storm is a hardware quirk specific to the rtl839x silicon. When the generic MIPS SMP code (whether VSMP or CPS) brings up a secondary core or VPE, it inherently unmasks IP7, assuming the native MIPS CP0 timer is available. If we build the kernel without CONFIG_CEVT_R4K, no handler is ever registered for this interrupt. Consequently, any Realtek SoC running in SMP mode (which includes rtl930x and rtl931x) is exposed to the exact same "irq 7: nobody cared" kernel panic. I don't have any of those devices to test why they work today. Are they using SMP?

Although the issue was prominently observed on devices like the Zyxel GS1920-24HPv1 and Edgecore ECS-2100, handling this at the subtarget level hides the true nature of the problem. The generic MIPS SMP code fundamentally relies on CEVT_R4K being present to gracefully handle the secondary CPU bring-up.

Keeping it selected globally for the target safely satisfies this generic MIPS requirement for all current and future SMP devices with practically zero runtime overhead (as the timer just goes dormant anyway) and a little bit of extra code.

jameywine: how much of this is LLM talking points?

I can't speak for rtl930x but i am very much sure rtl931x will have the same result as mine on rtl9607c.
It is SMP, has 4 CPUs, all brought up without any issues and there is no such panic. Here is even the interrupt stats for you to look at.

```

root@OpenWrt:~# cat /proc/interrupts 
           CPU0       CPU1       CPU2       CPU3       
  8:      13547          0          0          0 MIPS GIC   7  timer@3200
  9:          0       2516          0          0 MIPS GIC   8  timer@3200
 10:          0          0       2583          0 MIPS GIC   9  timer@3200
 11:          0          0          0       2668 MIPS GIC  10  timer@3200
 12:       1475          0          0          0 MIPS GIC  63  IPI call
 13:          0        725          0          0 MIPS GIC  64  IPI call
 14:          0          0       1365          0 MIPS GIC  65  IPI call
 15:          0          0          0        808 MIPS GIC  66  IPI call
 16:        405          0          0          0 MIPS GIC  67  IPI resched
 17:          0        486          0          0 MIPS GIC  68  IPI resched
 18:          0          0        319          0 MIPS GIC  69  IPI resched
 19:          0          0          0        412 MIPS GIC  70  IPI resched
 23:       1703          0          0          0 MIPS GIC  30  ttyS0
 24:       5154          0          0          0 MIPS GIC  42  rtl-snand
 25:          0          0          0          0 MIPS GIC  40  realtek-otto-wdt
 26:          0          0          0          0 realtek-otto-gpio   5  keys
 27:          0          0          0          0 realtek-otto-gpio   2  keys
 28:          0          0          0          0 realtek-otto-gpio   4  keys
ERR:          1
root@OpenWrt:~#
```

Your comment did not convince me so my previous comment still stands.

target/linux/realtek/patches-6.18/300-01-enhance-realtek-platform-config.patch
```diff
@@ -55,13 +55,15 @@ Submitted-by: Birger Koblitz <git@birger-koblitz.de>
...
 +	select SYS_SUPPORTS_MULTITHREADING
++	select CEVT_R4K
 +
 +config RTL930X
 +	bool "Realtek RTL930X based platforms"
 +	depends on MACH_REALTEK_RTL
 +	select MIPS_CPU_SCACHE
 +	select MIPS_EXTERNAL_TIMER
 +	select SYS_SUPPORTS_MULTITHREADING
++	select CEVT_R4K
```
openwrt-ai: This `select` also reaches the rtl931x subtargets, so removing `CONFIG_CEVT_R4K=y` from their config files leaves them out of sync with the config the build actually produces.

`RTL931X` is [`depends on RTL930X`](``https://github.com/openwrt/openwrt/blob/a1dbc9ae45988f7e2660f6cfae5fd6848de61a24/target/linux/realtek/patches-6.18/300-01-enhance-realtek-platform-config.patch#L70),`` not a sibling of it, and both rtl931x configs carry `CONFIG_RTL930X=y` — [`rtl931x/config-6.18:237`](https://github.com/openwrt/openwrt/blob/a1dbc9ae45988f7e2660f6cfae5fd6848de61a24/target/linux/realtek/rtl931x/config-6.18#L237) and [`rtl931x_nand/config-6.18:245`](https://github.com/openwrt/openwrt/blob/a1dbc9ae45988f7e2660f6cfae5fd6848de61a24/target/linux/realtek/rtl931x_nand/config-6.18#L245). `CEVT_R4K` is a promptless bool ([`arch/mips/Kconfig:1116-1117`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/Kconfig#L1116-L1117) in v6.18.44), so it is forced on wherever `RTL930X=y` — including rtl931x and rtl931x_nand. Yet this revision drops the `CONFIG_CEVT_R4K=y` line from both of those config files while keeping it in the three `RTL930X`-only ones.

That is the same config-file staleness that the previous round added the lines to fix, just moved to two other subtargets; it also silently contradicts @jameywine's rtl9607c data point, since rtl931x still ends up with `CEVT_R4K=y` after this change.

Two coherent ways out, depending on what you intend:
- rtl931x is meant to get it (your "all SMP realtek platforms" argument) — then restore `CONFIG_CEVT_R4K=y` in `rtl931x/config-6.18` and `rtl931x_nand/config-6.18`.
- rtl931x is meant to stay without it (it boots fine on GIC-based CPS SMP per the `/proc/interrupts` dump above) — then the `select` belongs on `RTL839X` only, or `RTL930X` needs a `!RTL931X` guard, since `depends on RTL930X` makes inheritance unavoidable otherwise.

Either way the build stays green (promptless symbol), so CI won't catch this.

---

target/linux/realtek/patches-6.18/300-01-enhance-realtek-platform-config.patch
```diff
@@ -55,13 +55,15 @@ Submitted-by: Birger Koblitz <git@birger-koblitz.de>
 +	select CPU_SUPPORTS_CPUFREQ
 +	select MIPS_EXTERNAL_TIMER
 +	select SYS_SUPPORTS_MULTITHREADING
++	select CEVT_R4K
```
openwrt-ai: nit: the `select` list in the `RTL838X`, `RTL839X` and `RTL930X` blocks this patch adds is alphabetically sorted; appending `CEVT_R4K` at the end breaks that in both blocks it touches.

```suggestion
+	select CEVT_R4K
+	select CPU_SUPPORTS_CPUFREQ
+	select MIPS_EXTERNAL_TIMER
+	select SYS_SUPPORTS_MULTITHREADING
```

Same applies to line 66 in the `RTL930X` block (`CEVT_R4K` before `MIPS_CPU_SCACHE`). Conflicting evidence, so take it or leave it: the `RTL931X` block right below and the `MACH_REALTEK_RTL` block in the first hunk are both unsorted already.

---

openwrt-ai (commented):
Reviewed the force-push (`506dafc` → `a1dbc9a`), which rescopes the fix from a target-wide `select` to a per-platform one.

What I could verify against `gregkh/linux:v6.18.44` (`target/linux/realtek` is on `patches-6.18`):

- The rewritten hunk headers are exact. `@@ -700,19 +700,22 @@` is unchanged from base, and `@@ -1056,6 +1059,39 @@` is arithmetically right (6 context + 33 added; start shifted +3 by the first hunk). Reapplying the patch to a pristine v6.18.44 `arch/mips/Kconfig` succeeds at fuzz 0.
- The rtl838x exclusion is coherent: `rtl838x/config-6.18` has no `CONFIG_SMP=y` and `RTL838X` doesn't select `SYS_SUPPORTS_MULTITHREADING`, so there are no secondary VPEs to unmask IP7 there. Dropping its `CONFIG_CEVT_R4K=y` line returns that file to its base state.

One issue inline: the `RTL930X` select leaks into the rtl931x subtargets via `depends on RTL930X`, which makes the two `CONFIG_CEVT_R4K=y` removals in `rtl931x/config-6.18` and `rtl931x_nand/config-6.18` inconsistent with the resulting kernel config. Plus one ordering nit.

## Commit checks

- `a1dbc9a` "realtek: fix SMP IRQ 7 storm by restoring CEVT_R4K" — the message body is unchanged from the previous revision and still reads as an unconditional restore ("Without it, disabling CEVT_R4K results in an unhandled IRQ 7 ... on secondary VPEs"), but the diff now restores the symbol only for `RTL839X`/`RTL930X` (and `RTL931X` transitively), deliberately leaving `RTL838X` without it. Since the scoping is the contested point of this review, worth a line in the body saying which platforms get it and why UP rtl838x is excluded.

---

plappermaul:
Please go back to initial Version of this PR. This was what I asked for and exactly describes what all the hardware has built in.

jameywine:
> exactly describes what all the hardware has built in.

how exactly does it describe all of the hardware if only one part of it suffer from the issue? Is it *really* descriptive of all of them then? Nobody seems to know why rtl839x is the only one with the irq 7 storm issue and all i've got was llm talking about which doesn't help.

plappermaul:
If the Hardware is fixed with an upstream Solution, why dig around in this? Why should one advise Solutions for Hardware one does not own?

jameywine:
> If the Hardware is fixed with an upstream Solution, why dig around in this?

I am not against restoring the config or whatever else they want to fix their problem. My request is solely to not leak that solution onto other hardware that doesn't need it at all.

plappermaul:
Living under the same target is about compromises. Use a common base where possible. As of now I See no clear Argument that is against this harmonization - Besides some half-baked vague comment from myself years ago. 

Your invest in this is just a test if it works for RTL960x

luizluca:
@jameywine , @plappermaul ,

I must have clicked "resolved" before adding the comment and github ate my long comment. Sorry. I'll try to rewrite the important parts:

I dig into the subject. There are two SMP models: MT and CPS. The rtl839x and rtl930x are quite close subtargets and both use SMP while rtl931x uses CPS.

For MT, it unconditionally enable the STATUSF_IP7:

```C
static void vsmp_init_secondary(void)
{
        /* This is Malta specific: IPI,performance and timer interrupts */
        if (mips_gic_present())
                change_c0_status(ST0_IM, STATUSF_IP2 | STATUSF_IP3 |
                                         STATUSF_IP4 | STATUSF_IP5 |
                                         STATUSF_IP6 | STATUSF_IP7);
        else
                change_c0_status(ST0_IM, STATUSF_IP0 | STATUSF_IP1 |
                                         STATUSF_IP6 | STATUSF_IP7);
}
```

So, no matter what (if it has mips_gic_present or cpu_has_veic), it will trigger the IRQ 7 by default and we would either need to masquerade it or add something to handle it. In that case, CEVT_R4K is necessary for rtl839x and rtl930x (this one not tested).

For rtl931x, the CPS code only enables STATUSF_IP7 conditionally:

```c
static void cps_init_secondary(void)
{
        int core = cpu_core(&current_cpu_data);

        (...)
        if (cpu_has_veic)
                clear_c0_status(ST0_IM);
        else
                change_c0_status(ST0_IM, STATUSF_IP2 | STATUSF_IP3 |
                                         STATUSF_IP4 | STATUSF_IP5 |
                                         STATUSF_IP6 | STATUSF_IP7);
}
```

If cpu_has_veic, it does not enable the STATUSF_IP7. cpu_has_veic is defined as:

```C
#if defined(CONFIG_CPU_MIPSR2_IRQ_EI) && !defined(cpu_has_veic)
# define cpu_has_veic           __opt(MIPS_CPU_VEIC)
#elif !defined(cpu_has_veic)
# define cpu_has_veic                   0
#endif
```

All targets have CONFIG_CPU_MIPSR2_IRQ_EI, leaving the SoC to define if it has the feature. rtl931x seems to have GIC (Global Interrupt Controller) and, that way, cpu_has_veic is true. This is why the issue didn't appear in all rtl931x devices.

It would be good to confirm that rtl930x is also affected in real HW, but the code (future kernel v7.3) makes it mostly certain it is. I only have access to rtl838x and rtl839x devices.

I also don't expect Realtek to create a newer family that does not have an IRQ Controller that would also justify adding CEVT_R4K by default for all targets.

@plappermaul, with all that, it is your call. Should I enable CEVT_R4K for all targets or only for rtl839x and rtl930x?

(Meanwhile, I'll deal with the Claude review)

plappermaul:
With active CONFIG_CEVT_R4K kernel nicely checks for timer availability in [r4k_clockevent_init()](https://elixir.bootlin.com/linux/v6.19.13/C/ident/r4k_clockevent_init). 

Use upstream generic config and enable it for all targets. Running a quick check on RTL838x and RTL931x should be a no-brainer.

jameywine:
> rtl931x seems to have GIC (Global Interrupt Controller) and, that way, cpu_has_veic is true. This is why the issue didn't appear in all rtl931x devices.

yeah and so does rtl960x (well rtl9607c to be exact) which is also why i didn't encounter any issues on my board.

> With active CONFIG_CEVT_R4K kernel nicely checks for timer availability in [r4k_clockevent_init()](https://elixir.bootlin.com/linux/v6.19.13/C/ident/r4k_clockevent_init).
> 
> Use upstream generic config and enable it for all targets. Running a quick check on RTL838x and RTL931x should be a no-brainer.

I am cool with this resolution then.

plappermaul:
Works nicely on RLT931x

plappermaul:
Works on RTL838x

luizluca:
> With active CONFIG_CEVT_R4K kernel nicely checks for timer availability in [r4k_clockevent_init()](https://elixir.bootlin.com/linux/v6.19.13/C/ident/r4k_clockevent_init).
> 
> Use upstream generic config and enable it for all targets. Running a quick check on RTL838x and RTL931x should be a no-brainer.

I rebase 506dafc89dfec3ffd590d663c1ccbbaf851a7264 and force-pushed. It was already reviewed (https://github.com/openwrt/openwrt/pull/24702#pullrequestreview-4934055360).

In practice, setting either globally or only for RTL839x and RTL930x are currently equivalent as RTL931x depends on RTL930x. We would first need to decouple them before adding a config specific for RTL930x. I wouldn't be hard as the code is built for all subtargets unconditionally and there is no #ifdef mentioning RTL930x (except for the REALTEK_THERMAL)

plappermaul:

PR #4915, opened 2022-01-03
ramips: add support for ipTIME A3004NS-M

mans0n:
ipTIME A3004NS-M is an 802.11ac (Wi-Fi 5) router, based on MediaTek
MT7621A.

Specifications:
* SoC: MT7621A
* RAM: 256 MiB
* Flash: SPI NOR 16 MiB
* Wi-Fi:
  * MT7615D: 2.4/5 GHz (DBDC)
* Ethernet: 5x 1GbE
  * Switch: SoC built-in
* USB: 1x 3.0
* UART: J4 (57600 baud)
  * Pinout: [3V3] (TXD) (RXD) (GND)

MAC addresses:

| interface |        MAC        |     source     | comment
|-----------|-------------------|----------------|---------
|       LAN | 88:XX:XX:E9:XX:BF |                | [1]
|       WAN | 88:XX:XX:E9:XX:BD | u-boot 0x1fc40 |
|   WLAN 2G | 8A:XX:XX:C9:XX:BC |                |
|   WLAN 5G | 88:XX:XX:E9:XX:BC | factory 0x4    |
|           |                   |                |
|           | 88:XX:XX:E9:XX:BC | u-boot 0x1fc20 | [2]
|           | 88:XX:XX:E9:XX:BE | factory 0x8004 |

[1] Used in this patch as WLAN 2G MAC address with the local bit set
[2] LAN MAC address on older stock firmware (e.g. version 10.050)

Known issues:
* DBDC is not turned on automatically after boot.

Notes:
* phy0/phy1 should be set to wlan2g/wlan5g respectively.

Installation via web interface:
1.  Flash **initramfs** image through the stock web interface.
2.  Boot into OpenWrt and perform sysupgrade with sysupgrade image.

Revert to stock firmware:
1.  Perform sysupgrade with stock image.

PR #24707, opened 2026-08-13
ramips: add support for ipTIME A3004NS-M

hwkim3330:
This revives [#4915](https://github.com/openwrt/openwrt/pull/4915), which added
this device in 2021 and was closed unmerged by its author. The reason it was
closed is fixed here, so the port is proposed again rather than left in a fork.

@mans0n wrote the device support and the commit is still his, with his
`Signed-off-by` intact; the bracketed note in it records what the rebase changed.
He closed the original because the MT7615D on this board is wired for DBDC while
its factory EEPROM reports `MT_EE_DUAL_BAND` at 0x3e, so mt76 registered a single
dual-band phy and only 2.4 GHz worked. In his words there was *"no way to
override it in dts"*, and he would not merge it with the debugfs workaround — "a
major flaw".

The first commit is that override, carried as
`package/kernel/mt76/patches/100-mt7615-allow-forcing-DBDC-from-device-tree.patch`.
The same change is proposed against openwrt/mt76 in
[openwrt/mt76#1119](https://github.com/openwrt/mt76/pull/1119); this
patch should be dropped at whichever `mt76: update to Git HEAD` bump includes it,
and the patch header says so. Carrying it rather than waiting keeps the two
reviews independent — without it the port still builds and still works on
2.4 GHz, which is the state that got the original closed.

### Rebased onto current conventions

- `nvmem-layout` with `fixed-layout` rather than a bare `nvmem-cells` partition
- a `mac-base` cell instead of `mac-address-increment`
- LED nodes without the deprecated `label` property
- no `dsa-migration`: this device never shipped a swconfig release, so there is no
  configuration to migrate from and adding it would only force `-F` on users
- the WLAN LED gained `linux,default-trigger = "phy0radio"`; without a trigger it
  never lit. a3002mesh and a3004t are the same chip driving the same GPIO 17 for
  the same purpose and both use it

`label-mac-device` is deliberately absent. The stock firmware used the un-offset
u-boot value for its LAN address, which under OpenWrt belongs to the 5 GHz phy
while LAN is that value + 3 — so it is likely no netdev carries the label MAC.
Nobody has read the sticker on the underside to confirm it, so rather than guess
the property is left out; happy to add it if a maintainer would prefer a
specific interface named.

### Tested on hardware

ipTIME A3004NS-M, kernel 6.18.41. Flashed through the stock web interface, then
sysupgraded to the persistent image; everything below survived a power-cycle
reboot with the overlay on flash (`/dev/mtdblock6`, jffs2).

| | result |
|---|---|
| Wi-Fi | `phy0` 2412 MHz and `phy1` 5180 MHz, both Master **at the same time** — ch1 HT20 and ch36 VHT80 |
| a 5 GHz client | 866.7 Mbit/s, VHT-MCS 9, 80 MHz, VHT-NSS 2, took a DHCP lease |
| antenna split | `0x34` = `0x44` → phy0 `0x3`, phy1 `0xc`, i.e. 2x2 + 2x2 |
| EEPROM 0x3e | `0x0a`, matching what was dumped on this model in 2021 |
| Ethernet | lan1 links at 1000 Mbps; lan2-4 and wan correctly down when unplugged |
| USB 3.0 | a UVC camera enumerates at 5000 Mbps |
| LEDs, buttons | CPU LED on boot, WLAN LED follows phy0, reset and WPS both report |
| flash layout | partitions tile 0→16 MiB exactly; `IMAGE_SIZE` 16128k == the firmware partition |
| driver log | no mt76 or MCU errors |

MAC assignment, matching the table in the commit message:

| | measured | rule |
|---|---|---|
| LAN | `70:5d:cc:77:4c:03` | u-boot 0x1fc20 **+3** |
| WAN | `70:5d:cc:77:4c:01` | u-boot 0x1fc40 |
| WLAN 5 GHz | `70:5d:cc:77:4c:00` | factory 0x4 |
| WLAN 2.4 GHz | `72:5d:cc:77:4c:03` | LAN with the local bit set |

### Installation

Flash the **initramfs** image through the stock web interface, boot it, then
`sysupgrade` with the squashfs image. Reverting is a `sysupgrade` with an official
ipTIME image (`a3004nm_kr_*.bin`), which carries the same `a3004nm` uImage name
the stock updater checks.

### Release notes

> ramips: add support for the ipTIME A3004NS-M (MT7621A, 256 MiB RAM, 16 MiB
> SPI NOR, 5x 1GbE, MT7615D DBDC Wi-Fi 5, 1x USB 3.0). Requires the accompanying
> mt7615 change to bring up both bands; without it only 2.4 GHz is available.

### DIFF
--- a/target/linux/ramips/dts/mt7621_iptime_a3004ns-m.dts
+++ b/target/linux/ramips/dts/mt7621_iptime_a3004ns-m.dts
@@ -1,0 +1,183 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+#include "mt7621.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+
+/ {
+	compatible = "iptime,a3004ns-m", "mediatek,mt7621-soc";
+	model = "ipTIME A3004NS-M";
+
+	aliases {
+		led-boot = &led_cpu;
+		led-failsafe = &led_cpu;
+		led-running = &led_cpu;
+		led-upgrade = &led_cpu;
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		led_cpu: led-0 {
+			color = <LED_COLOR_ID_AMBER>;
+			function = LED_FUNCTION_CPU;
+			gpios = <&gpio 18 GPIO_ACTIVE_LOW>;
+		};
+
+		led-1 {
+			color = <LED_COLOR_ID_AMBER>;
+			function = LED_FUNCTION_WLAN;
+			gpios = <&gpio 17 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "phy0radio";
+		};
+	};
+
+	keys {
+		compatible = "gpio-keys";
+
+		reset {
+			label = "reset";
+			gpios = <&gpio 4 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
+
+		wps {
+			label = "wps";
+			gpios = <&gpio 3 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_WPS_BUTTON>;
+		};
+	};
+};
+
+&spi0 {
+	status = "okay";
+
+	flash@0 {
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <50000000>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "u-boot";
+				reg = <0x0 0x20000>;
+				read-only;
+
+				nvmem-layout {
+					compatible = "fixed-layout";
+					#address-cells = <1>;
+					#size-cells = <1>;
+
+					macaddr_uboot_1fc20: macaddr@1fc20 {
+						compatible = "mac-base";
+						reg = <0x1fc20 0x6>;
+						#nvmem-cell-cells = <1>;
+					};
+
+					macaddr_uboot_1fc40: macaddr@1fc40 {
+						reg = <0x1fc40 0x6>;
+					};
+				};
+			};
+
+			partition@20000 {
+				label = "config";
+				reg = <0x20000 0x10000>;
+				read-only;
+			};
+
+			partition@30000 {
+				label = "factory";
+				reg = <0x30000 0x10000>;
+				read-only;
+
+				nvmem-layout {
+					compatible = "fixed-layout";
+					#address-cells = <1>;
+					#size-cells = <1>;
+
+					eeprom_factory_0: eeprom@0 {
+						reg = <0x0 0x4da8>;
+					};
+				};
+			};
+
+			partition@40000 {
+				compatible = "denx,uimage";
+				label = "firmware";
+				reg = <0x40000 0xfc0000>;
+			};
+		};
+	};
+};
+
+&state_default {
+	gpio {
+		groups = "i2c", "jtag", "wdt";
+		function = "gpio";
+	};
+};
+
+&gmac0 {
+	nvmem-cells = <&macaddr_uboot_1fc20 3>;
+	nvmem-cell-names = "mac-address";
+};
+
+&switch0 {
+	ports {
+		port@0 {
+			status = "okay";
+			label = "lan1";
+		};
+
+		port@1 {
+			status = "okay";
+			label = "lan2";
+		};
+
+		port@2 {
+			status = "okay";
+			label = "lan3";
+		};
+
+		port@3 {
+			status = "okay";
+			label = "lan4";
+		};
+
+		port@4 {
+			status = "okay";
+			label = "wan";
+			nvmem-cells = <&macaddr_uboot_1fc40>;
+			nvmem-cell-names = "mac-address";
+		};
+	};
+};
+
+&pcie {
+	status = "okay";
+};
+
+&pcie0 {
+	wifi@0,0 {
+		compatible = "mediatek,mt76";
+		reg = <0x0000 0 0 0 0>;
+		nvmem-cells = <&eeprom_factory_0>;
+		nvmem-cell-names = "eeprom";
+
+		/*
+		 * The MT7615D is wired for DBDC, but the factory EEPROM
+		 * describes it as a single dual-band phy (MT_EE_WIFI_CONF
+		 * band_sel == MT_EE_DUAL_BAND), so the driver cannot detect
+		 * DBDC on its own.  Force it here, otherwise only one band
+		 * comes up at a time.
+		 */
+		mediatek,dbdc;
+	};
+};
--- a/target/linux/ramips/image/mt7621.mk
+++ b/target/linux/ramips/image/mt7621.mk
@@ -1897,6 +1897,15 @@ define Device/iptime_a3004ns-dual
 endef
 TARGET_DEVICES += iptime_a3004ns-dual
 
+define Device/iptime_a3004ns-m
+  IMAGE_SIZE := 16128k
+  UIMAGE_NAME := a3004nm
+  DEVICE_VENDOR := ipTIME
+  DEVICE_MODEL := A3004NS-M
+  DEVICE_PACKAGES := kmod-mt7615-firmware kmod-usb3 -uboot-envtools
+endef
+TARGET_DEVICES += iptime_a3004ns-m
+
 define Device/iptime_a3004t
   $(Device/nand)
   $(Device/uimage-lzma-loader)
--- a/target/linux/ramips/mt7621/base-files/etc/hotplug.d/ieee80211/10_fix_wifi_mac
+++ b/target/linux/ramips/mt7621/base-files/etc/hotplug.d/ieee80211/10_fix_wifi_mac
@@ -95,6 +95,7 @@ case "$board" in
 			macaddr_unsetbit "$(cat /sys${DEVPATH}/macaddress)" 6 > /sys${DEVPATH}/macaddress
 		;;
 	iptime,a3002mesh|\
+	iptime,a3004ns-m|\
 	iptime,a3004t)
 		hw_mac_addr="$(mtd_get_mac_binary factory 0x4)"
 		[ "$PHYNBR" = "0" ] && \
--- a/package/kernel/mt76/Makefile
+++ b/package/kernel/mt76/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mt76
-PKG_RELEASE=1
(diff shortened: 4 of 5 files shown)


target/linux/ramips/mt7621/base-files/etc/hotplug.d/ieee80211/10_fix_wifi_mac
```diff
@@ -102,6 +102,13 @@ case "$board" in
 		[ "$PHYNBR" = "1" ] && echo -n "$hw_mac_addr" > /sys${DEVPATH}/macaddress
 		;;
 	iptime,ax2002m|\
+	iptime,a3004ns-m)
+		hw_mac_addr="$(mtd_get_mac_binary factory 0x4)"
+		[ "$PHYNBR" = "0" ] && \
+			macaddr_setbit_la "$(macaddr_add $hw_mac_addr 3)" > /sys${DEVPATH}/macaddress
+		[ "$PHYNBR" = "1" ] && \
```
neheb: can this be implemented in dts?

target/linux/ramips/mt7621/base-files/etc/hotplug.d/ieee80211/10_fix_wifi_mac
```diff
@@ -102,6 +102,13 @@ case "$board" in
...
+			echo -n "$hw_mac_addr" > /sys${DEVPATH}/macaddress
+		;;
```
openwrt-ai: This regresses `iptime,ax2002m`. The new case label was inserted between `iptime,ax2002m|\` and `iptime,ax2004m)`, so the `|\` continuation now joins `ax2002m` to the *new* pattern list. After this patch `ax2002m` executes the a3004ns-m body (which also writes phy0, and writes phy1 unmodified) instead of the `ax2004m` body that only touches phy1 — i.e. both radios on ax2002m get different MACs than before.

Moving the new case above the `ax2002m|` label keeps the existing pair intact and stays alphabetical:

```suggestion
	iptime,a3004ns-m)
		hw_mac_addr="$(mtd_get_mac_binary factory 0x4)"
		[ "$PHYNBR" = "0" ] && \
			macaddr_setbit_la "$(macaddr_add $hw_mac_addr 3)" > /sys${DEVPATH}/macaddress
		[ "$PHYNBR" = "1" ] && \
			echo -n "$hw_mac_addr" > /sys${DEVPATH}/macaddress
		;;
	iptime,ax2002m|\
```

Note the body is byte-for-byte the same rule as the existing [`iptime,a3002mesh|iptime,a3004t)` case at 10_fix_wifi_mac:97-103](``https://github.com/openwrt/openwrt/blob/31ec78c93c50bd859c70457ed04cd8beac9bf35d/target/linux/ramips/mt7621/base-files/etc/hotplug.d/ieee80211/10_fix_wifi_mac#L97-L103),`` so adding `iptime,a3004ns-m|\` as a third label on that case would be even smaller and avoids duplicating the logic.

---

target/linux/ramips/image/mt7621.mk
```diff
@@ -1897,6 +1897,16 @@ define Device/iptime_a3004ns-dual
 endef
 TARGET_DEVICES += iptime_a3004ns-dual
 
+define Device/iptime_a3004ns-m
+  IMAGE_SIZE := 16128k
+  UIMAGE_NAME := a3004nm
+  DEVICE_VENDOR := ipTIME
+  DEVICE_MODEL := A3004NS-M
+  DEVICE_PACKAGES := kmod-mt7615e kmod-mt7615-firmware kmod-usb3 \
+	-uboot-envtools
```
openwrt-ai: nit: `kmod-mt7615e` is redundant — `KernelPackage/mt7615-firmware` already carries `DEPENDS+=+kmod-mt7615e` ([package/kernel/mt76/Makefile](https://github.com/openwrt/openwrt/blob/31ec78c93c50bd859c70457ed04cd8beac9bf35d/package/kernel/mt76/Makefile)). Every other ipTIME MT7615 recipe in this file (`iptime_a3002mesh`, `iptime_a3004t`, `iptime_a6004ns-m`, `iptime_a6ns-m`) lists only the firmware package, and the line then fits without the continuation, matching `iptime_a3004t` right below.

```suggestion
  DEVICE_PACKAGES := kmod-mt7615-firmware kmod-usb3 -uboot-envtools
```

---

target/linux/ramips/dts/mt7621_iptime_a3004ns-m.dts
```diff
@@ -0,0 +1,183 @@
...
+			gpios = <&gpio 18 GPIO_ACTIVE_LOW>;
+		};
+
+		led-1 {
+			color = <LED_COLOR_ID_AMBER>;
+			function = LED_FUNCTION_WLAN;
+			gpios = <&gpio 17 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "phy0radio";
+		};
+	};
```
openwrt-ai, line 35: nit: no USB LED here, although the board has USB 3.0 (`kmod-usb3` is in `DEVICE_PACKAGES`). The two sibling ipTIME boards with a USB port both have one — [mt7621_iptime_a3004ns-dual.dts:29-35](https://github.com/openwrt/openwrt/blob/31ec78c93c50bd859c70457ed04cd8beac9bf35d/target/linux/ramips/dts/mt7621_iptime_a3004ns-dual.dts#L29-L35) drives GPIO 7 with `trigger-sources = <&xhci_ehci_port1>` / `linux,default-trigger = "usbport"` and frees the `uart3` pin group for it, and `iptime_a6004ns-m` pulls in `kmod-usb-ledtrig-usbport`. This DTS neither frees `uart3` nor declares such an LED. The commit message covers the deliberately-omitted second WLAN LED but says nothing about a USB LED — is there no USB LED on the case, or was it just not probed?

---

openwrt-ai (commented):
## Commit checks

- `31ec78c9` "ramips: add support for ipTIME A3004NS-M" — new-device commits are expected to carry a complete hardware specification; this one lists SoC, RAM, Flash, Wi-Fi, Ethernet, USB and UART but has no **LEDs/Buttons** line (count + type — the DTS adds a CPU LED on GPIO 18, a WLAN LED on GPIO 17, and reset/WPS buttons) and no **Power** line (voltage + current from the barrel-jack label). Worth adding while the hardware is on the bench.

Everything else in the two commit messages matches the diffs (the `PKG_RELEASE` bump and the carried patch are both described in `3007750`).


---

hwkim3330:
@neheb — not the part that is left, no.

The device tree already takes both ethernet MACs from nvmem cells: `&gmac0` uses
`<&macaddr_uboot_1fc20 3>` and the WAN port `<&macaddr_uboot_1fc40>`. What remains in
the script is only the wifi derivation, which needs the locally-administered bit set on
top of the increment, and there is no device-tree property for that — `mac-base` gives an
increment and nothing else. That is why `macaddr_setbit_la` appears 30 times across the
ramips and filogic hotplug scripts, including for the two siblings that share this board's
factory layout.

The rule this board needs is byte-for-byte the one `iptime,a3002mesh` and `iptime,a3004t`
already use, so I have folded it into their case rather than adding a block of my own. The
change to that file is now one line:

```
 	iptime,a3002mesh|\
+	iptime,a3004ns-m|\
 	iptime,a3004t)
```

That also fixes a regression, which the AI review caught and I had missed: the new label had
been inserted between `iptime,ax2002m|\` and `iptime,ax2004m)`, so the `|\` continuation
joined `ax2002m` to the new pattern list. `ax2002m` would have run the new body — which
writes phy0 and writes phy1 unmodified — instead of the `ax2004m` body that only touches
phy1, changing both of its radios' MACs. Thank you for that one.

The rest of the review is addressed:

- `kmod-mt7615e` dropped from `DEVICE_PACKAGES`. Verified rather than assumed: it still
  appears in the built image's manifest, pulled in by `kmod-mt7615-firmware`'s
  `DEPENDS+=+kmod-mt7615e`.
- `LEDs`, `Buttons` and `Power` lines added to the specification: 2x amber (cpu, wlan),
  2x tactile (reset, wps), 12 VDC / 2 A from the barrel-jack label.
- **No USB LED on the case.** It was checked, not skipped. `a3004ns-dual` is a different
  design — MT76x2 with two separate radios, and it does have a USB LED on GPIO 7 — but this
  board has none, so the DTS is deliberately without one and `uart3` stays as it is.
- The carried mt76 patch already has its `From:` / `Date:` / `Subject:` headers for
  `git am`, and `PKG_RELEASE` is bumped. I also reverted an accidental `=` → `:=` operator
  change on that line so the mt76 diff is a single character.

@mans0n — thank you, and thanks for the original work; everything here except the DBDC fix
and the rebase is yours. I have cut the bracket down as you suggested.

hwkim3330:
Two corrections to my comment above.

The changes it described were not actually in the branch when I posted it — an
`--amend` with nothing staged rewrote only the commit message, so the first push
carried the new text and none of the code. The formality checks passed on it,
which hid that. They are in now, at 55f3d13:

- one label added to the `iptime,a3002mesh|iptime,a3004t` case, duplicated block
  removed, `iptime,ax2002m` back beside `iptime,ax2004m`
- `kmod-mt7615e` dropped from `DEVICE_PACKAGES`

And I was wrong about `PKG_RELEASE`. I had "reverted an accidental `=` → `:=`
operator change to keep the diff to one character"; `:=` is what the guidelines
require, the bot said so, and it is back. Apologies for the noise.

hwkim3330:
@namiltd — happy to, but I think one of us is looking at a different branch, so let
me state what I see rather than guess.

This branch is already pushed and current: `upstream/a3004ns-m` is at `55f3d13`,
which is what this PR points at, and all three formality checks pass on it. A
further force push is a no-op — git says `Everything up-to-date`. The bot comment
above about `PKG_RELEASE=2` was against the previous push; `:=` went back in at
`55f3d13` and the Makefiles check is green now.

On reopening #4915: that PR is @mans0n's, from `mans0n:pr-a3004nm`, so it is not
mine to reopen and it does not track this branch — reopening it would surface the
2021 version of the patch rather than this one. If you would rather the work land
in #4915, @mans0n would need to reopen it and I am glad to rebase onto whatever
branch it points at, or to close this one. If continuing here is fine, there is
nothing outstanding that I know of.

Either way, worth saying what changed since your last look, since it includes a
bug I introduced:

- `10_fix_wifi_mac`: the new label had been inserted between `iptime,ax2002m|\` and
  `iptime,ax2004m)`, so `ax2002m` would have run the new body and both its radios
  would have changed MAC. It is now one label on the existing
  `iptime,a3002mesh|iptime,a3004t` case, which has a byte-identical rule — the diff
  to that file is a single line.
- `kmod-mt7615e` dropped from `DEVICE_PACKAGES`; checked it still lands in the
  image via `kmod-mt7615-firmware`'s `DEPENDS`.
- `LEDs`, `Buttons` and `Power` (12 VDC, 2 A) added to the specification, and no
  USB LED on this case — that was checked rather than skipped.

openwrt-ai (commented):

---

PR #24368, opened 2026-07-22
kernel: add kmod-dm-thin-pool related modules and refactor kmod-dm / kmod-md-mod

davidandreoletti:
**Env:**
- Compiled for : Rockchip / armv8
- Run tested: NanoPi R4S (with RAID1 + LVM thin pool accessible)

**Context**

Add snapshot/thin provisioning LVM support in the linux kernel in order to use it via the lvm2 package.

**Core problem**

kmod-dm brings in many dependencies not required to conform to its linux kernel's KConfig equivalent. 
Adding new kmod-dm-xxxx to implement new CONFIG_DM_xxxx should pull in only the dependencies chain described by the linux kernel's KConfig equivalent.

**Sub problems**

- kmod-dm sets additional CONFIG_DM_xxx=n that should be explicitly marked as "CONFIG_DM_xxx not set" in a linux generic kernel config
- kmod-dm sets CONFIG_DM_MIRROR/CONFIG_DM_CRYPT while a stricter kmod-dm implementation should only set CONFIG_BLK_DEV_DM to closely match its linux's KConfig equivalent.

**This PR in a nutshell:**
- Addresses all problems in order to simplifying kmod-dm
- Adds snapshot/thin provisioning new kernal package relying on the simplified kmod-dm
- Aims not to break OpenWrt package/images during the transition

**Transition Stage 1/2 details**
- Refactor block.mk's RAID and MD KernelPackages to move some ancillary CONFIG_xxx values to linux kernel configs by default
- Introduce CONFIG_DM_MIRROR/CONFIG_DM_CRYPT as new individual KernelPackages
- Refactor block.mk's kmod-dm to provide dm-mod.ko via a transitional kmod-dm2 kernel package.
- Existing block.mk kmod-xxx depending on kmod-dm will now referenecs kmod-dm2 (for the duration of the transition)
-- Required to not break existing package/image during the transition. 
- Add 2 new KernelPackages (kernel package for LVM snapshots, a kernel package for LVM thin pool) and dependencies)

**Notes**
- The current STYLISTIC WARNINGS & SUGGESTIONS seems irrelevant for the change made to KernelPackages 

**Objectives for the next PR: Transition stage 2/2**
- Transition **all packages** (in this repo and packages.git) to add +kmod-dm-mirror +kmod-dm-crypt when they used +kmod-dm
- Transition **all images** (in this repo) to add +kmod-dm-mirror +kmod-dm-crypt when they used +kmod-dm
-- How to notify users of the change ?
[...]

PR #24709, opened 2026-08-13
kernel: add kmod-nvme-tcp

davidandreoletti:
Env:

Compiled for : Rockchip / armv8
Run tested: NanoPi R4S with a remote NVME tcp target without TLS

**Context**

Add support for nvme over tcp initiator.

**Problem and solution**

`kmod-nvme` currently requires PCI support while the future `kmod-nvme-tcp` does not. 

Therefore, `kmod-nvme` will continue to require PCI support while the common config required by `kmod-nvme` and `kmod-nvme-tcp` will be moved into a new kmod-nvme-core

Note:
This PR is built upon https://github.com/openwrt/openwrt/pull/24368 only to test NVME over tcp on a MD raid settings. Once https://github.com/openwrt/openwrt/pull/24368 is merged, all commits but nvme related ones will be deleted.


package/kernel/linux/modules/block.mk
```diff
@@ -489,20 +659,34 @@ endef
...
 define KernelPackage/nvme
   SUBMENU:=$(BLOCK_MENU)
   TITLE:=NVM Express block device
-  DEPENDS:=@PCI_SUPPORT +kmod-hwmon-core
+  DEPENDS:=@PCI_SUPPORT +kmod-nvme-core +kmod-hwmon-core
   KCONFIG:= \
-	CONFIG_NVME_CORE \
 	CONFIG_BLK_DEV_NVME \
 	CONFIG_NVME_MULTIPATH=n \
 	CONFIG_NVME_HWMON=y
```
openwrt-ai: `CONFIG_NVME_HWMON` and `CONFIG_NVME_MULTIPATH` don't affect `nvme.ko` at all — both select objects that are linked into `nvme-core.ko`:

```
nvme-core-$(CONFIG_NVME_MULTIPATH)	+= multipath.o
...
nvme-core-$(CONFIG_NVME_HWMON)		+= hwmon.o
```

([`drivers/nvme/host/Makefile:16-19`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/nvme/host/Makefile#L16-L19), v6.18; identical in v6.12.)

So after this split, with `CONFIG_NVME_HWMON=y` still declared here, `nvme-core.ko` pulls in [`hwmon_device_register_with_info()` at hwmon.c:252](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/nvme/host/hwmon.c#L252) from `hwmon.ko`, while `kmod-nvme-core` (lines 662-669) declares no `DEPENDS` — `+kmod-hwmon-core` sits on `kmod-nvme` instead. [`CheckDependencies` in include/package-pack.mk:286](https://github.com/openwrt/openwrt/blob/ecbbfd88fabec8fc05ad09ec698747ede7086d47/include/package-pack.mk#L286) resolves that per-package, so this should fail with `Package kmod-nvme-core is missing dependencies for the following libraries: hwmon.ko` — the same failure mode you documented for `dm-mod.ko` in 144b560a.

Move both KCONFIG symbols and `+kmod-hwmon-core` into `KernelPackage/nvme-core`, so the package that ships `nvme-core.ko` owns the config that shapes it (a side effect is that `kmod-nvme-tcp` then pulls `kmod-hwmon-core` in too; if that is unwanted, the alternative is `CONFIG_NVME_HWMON=n`, but that is a behaviour change for PCIe users).

```suggestion
  DEPENDS:=@PCI_SUPPORT +kmod-nvme-core
  KCONFIG:=CONFIG_BLK_DEV_NVME
```

---

package/kernel/linux/modules/block.mk
```diff
@@ -513,6 +697,42 @@ endef
...
+ (TCP/RDMA/FC).
+endef
+
+$(eval $(call KernelPackage,nvme-fabrics))
+
+
+define KernelPackage/nvme-tcp
+  SUBMENU:=$(BLOCK_MENU)
+  TITLE:=NVM Express over Fabrics TCP host driver
+  DEPENDS:=+kmod-nvme-fabrics +kmod-crypto-crc32c
```
openwrt-ai: `+kmod-crypto-crc32c` is only correct for 6.12. On 6.18 `nvme-tcp` no longer uses the crypto ahash API — it was converted to the CRC32c library:

- v6.12: [`crypto_alloc_ahash("crc32c", ...)` at tcp.c:1420](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/drivers/nvme/host/tcp.c#L1420), and `NVME_TCP` `select CRYPTO_CRC32C` — so the runtime dependency on `crc32c_generic.ko` is real here.
- v6.18: [`crc32c()` at tcp.c:471](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/nvme/host/tcp.c#L471) via `<linux/crc32.h>`, and [`NVME_TCP` selects `CRC32`/`NET_CRC32C`, not `CRYPTO_CRC32C`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/nvme/host/Kconfig#L83-L89) — those are built-in libs, so `kmod-crypto-crc32c` (and the `kmod-crypto-hash` it drags in) is dead weight on the default kernel.

Guard it the same way `KernelPackage/dm-persistent-data` does at line 521 in this PR:

```suggestion
  DEPENDS:=+kmod-nvme-fabrics +LINUX_6_12:kmod-crypto-crc32c
```

---

package/kernel/linux/modules/block.mk
```diff
@@ -489,20 +659,34 @@ endef
 
 $(eval $(call KernelPackage,nbd))
 
+define KernelPackage/nvme-core
```
openwrt-ai, line 663: nit: every other `define KernelPackage/...` block in this file is separated from the preceding `$(eval ...)` by two blank lines (including the new `nvme-fabrics`/`nvme-tcp` blocks added in the same commit); this one only has one.

```suggestion


define KernelPackage/nvme-core
```

---

package/kernel/linux/modules/block.mk
```diff
@@ -513,6 +697,42 @@ endef
 $(eval $(call KernelPackage,nvme))
 
 
+define KernelPackage/nvme-fabrics
+  SUBMENU:=$(BLOCK_MENU)
+  TITLE:=NVM Express over Fabrics support
+  DEPENDS:=+kmod-nvme-core
```
openwrt-ai, line 706: nit: `kmod-nvme-core` gets `HIDDEN:=1` because it is "not directly useful on its own", but `kmod-nvme-fabrics` — whose own description says it is "common code shared by the NVMe over Fabrics host transport drivers" — is left user-selectable, even though `nvme-fabrics.ko` does nothing without a transport module either. Is exposing it in menuconfig intentional (e.g. anticipating out-of-tree RDMA/FC users), or should it be `HIDDEN:=1` for parity with `kmod-nvme-core`?

---

package/kernel/linux/modules/block.mk
```diff
@@ -513,6 +697,42 @@ endef
...
+
+define KernelPackage/nvme-fabrics/description
+ Common code shared by the NVMe over Fabrics host transport drivers
+ (TCP/RDMA/FC).
+endef
+
+$(eval $(call KernelPackage,nvme-fabrics))
+
+
+define KernelPackage/nvme-tcp
```
openwrt-ai, line 720: nit: the commit that adds this block is titled `kernel: added kmod-nvme-tcp and dependencies` — past tense, whereas the PR title (`kernel: add kmod-nvme-tcp`) and all 17 other commits in this series use the imperative mood (`add`, `move`, `mark`, `drop`, `remove`, `introduce`, `keep`). Worth rewording to `kernel: add kmod-nvme-tcp and dependencies` before this leaves draft.

---

openwrt-ai (commented):
Reviewed 18 commits. The 16 device-mapper commits are carried from #24368 per the PR description and are reviewed there, so I focused on the two NVMe commits (`d0e4fa6`, `ecbbfd8`); both commit messages match their diffs.

Main finding is inline on `KernelPackage/nvme`: the `nvme-core.ko` split left `CONFIG_NVME_HWMON`/`CONFIG_NVME_MULTIPATH` and `+kmod-hwmon-core` on `kmod-nvme`, but those symbols shape `nvme-core.ko`, which is now owned by `kmod-nvme-core`. Second finding is the unconditional `+kmod-crypto-crc32c` on `kmod-nvme-tcp`, which is only needed on 6.12.

Compared against the pre-cloned Linux v6.12.103 and v6.18.44 trees. CI on `ecbbfd8` reports no failed checks (status pending, nothing red), so nothing CI-grounded to add.

---

openwrt-ai (commented):
Reviewed 18 new commits; no new issues found.

The branch was force-pushed since my last review (`ecbbfd8` → `7cc24e0`). I compared patch-ids across the two series: the 16 device-mapper commits carried from #24368 are byte-identical, so the only new content is in `package/kernel/linux/modules/block.mk`. Both commit messages (`38c7084`, `7cc24e0`) still match their diffs.

Both findings from my previous review are addressed, and I re-verified the fixes against the pre-cloned Linux v6.12.103 and v6.18.44 trees:

- `CONFIG_NVME_HWMON=y` / `CONFIG_NVME_MULTIPATH=n` and `+kmod-hwmon-core` now sit on `kmod-nvme-core`. Correct: `drivers/nvme/host/Makefile` puts `hwmon.o` and `multipath.o` into `nvme-core-y`, and `NVME_HWMON` depends on `NVME_CORE`, not `BLK_DEV_NVME`.
- `+LINUX_6_12:kmod-crypto-crc32c` is right, and `kmod-crypto-crc32c` is the correct package rather than the more common in-tree `kmod-lib-crc32c`: on 6.12 `tcp.c` does `crypto_alloc_ahash("crc32c", ...)` (crypto API, matching `select CRYPTO_CRC32C`), whereas on 6.18 it calls the `crc32c()` library directly under `select CRC32`/`NET_CRC32C`, and `CONFIG_CRC32=y` is already built in via `target/linux/generic/config-6.18`.

Two other things I checked and found fine: `HIDDEN:=1` on `kmod-nvme-core` and `kmod-nvme-fabrics` is necessary, not just cosmetic — both `NVME_CORE` and `NVME_FABRICS` are promptless `tristate` symbols, so selecting either package standalone would leave kconfig computing them as `n` and the listed `.ko` would never be built. Adding `# CONFIG_NVME_TCP_TLS is not set` to both generic configs is also needed, since `NVME_TCP=m` makes that bool visible.

CI on `7cc24e06` reports no failed checks (pending, nothing red), so nothing CI-grounded to add. The bot's `PKG_RELEASE` note is worth acting on here beyond the usual formality — `kmod-nvme` loses `nvme-core.ko` in this split, so a bump is what gets existing installs to pick up the new `kmod-nvme-core`.

---

PR #24710, opened 2026-08-13
kernel: fix mdio series kernel modules dependencies for econet target

DragonBluep:
`econet` is a source-only target so we did not notice them before.

### DIFF
--- a/package/kernel/linux/modules/netdevices.mk
+++ b/package/kernel/linux/modules/netdevices.mk
@@ -254,7 +254,7 @@ $(eval $(call KernelPackage,mdio-devres))
 define KernelPackage/mdio-gpio
   SUBMENU:=$(NETWORK_DEVICES_MENU)
   TITLE:= Supports GPIO lib-based MDIO busses
-  DEPENDS:=+kmod-libphy @GPIO_SUPPORT +(TARGET_armsr||TARGET_bcm27xx_bcm2708||TARGET_loongarch64||TARGET_malta||TARGET_tegra):kmod-of-mdio
+  DEPENDS:=+kmod-libphy @GPIO_SUPPORT +(TARGET_armsr||TARGET_bcm27xx_bcm2708||TARGET_econet||TARGET_loongarch64||TARGET_malta||TARGET_tegra):kmod-of-mdio
   KCONFIG:= \
 	CONFIG_MDIO_BITBANG \
 	CONFIG_MDIO_GPIO
--- a/package/kernel/linux/modules/netdevices.mk
+++ b/package/kernel/linux/modules/netdevices.mk
@@ -235,7 +235,7 @@ $(eval $(call KernelPackage,mii))
 define KernelPackage/mdio-devres
   SUBMENU:=$(NETWORK_DEVICES_MENU)
   TITLE:=Supports MDIO device registration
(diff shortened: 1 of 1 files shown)


package/kernel/linux/modules/netdevices.mk
```diff
@@ -254,7 +254,7 @@ $(eval $(call KernelPackage,mdio-devres))
 define KernelPackage/mdio-gpio
   SUBMENU:=$(NETWORK_DEVICES_MENU)
   TITLE:= Supports GPIO lib-based MDIO busses
-  DEPENDS:=+kmod-libphy @GPIO_SUPPORT +(TARGET_armsr||TARGET_bcm27xx_bcm2708||TARGET_loongarch64||TARGET_malta||TARGET_tegra):kmod-of-mdio
+  DEPENDS:=+kmod-libphy @GPIO_SUPPORT +(TARGET_armsr||TARGET_bcm27xx_bcm2708||TARGET_econet||TARGET_loongarch64||TARGET_malta||TARGET_tegra):kmod-of-mdio
```
openwrt-ai, line 257: nit: `kmod-switch-rtl8366-smi` carries the exact same target list and looks like it needs `TARGET_econet` too, but it is not touched by this series.

[`netdevices.mk:1092`](https://github.com/openwrt/openwrt/blob/0e691dbababa9f0464666c3a7c7fb3b13a41e3b2/package/kernel/linux/modules/netdevices.mk#L1092) is the third and last occurrence of `+(TARGET_armsr||TARGET_bcm27xx_bcm2708||TARGET_loongarch64||TARGET_malta||TARGET_tegra):kmod-of-mdio` in the tree, and its module genuinely pulls in `of_mdio` symbols — [`of_mdiobus_register()` at rtl8366_smi.c:1033](https://github.com/openwrt/openwrt/blob/0e691dbababa9f0464666c3a7c7fb3b13a41e3b2/target/linux/generic/files/drivers/net/phy/rtl8366_smi.c#L1033) and [`of_mdio_find_bus()` at rtl8366_smi.c:1466](https://github.com/openwrt/openwrt/blob/0e691dbababa9f0464666c3a7c7fb3b13a41e3b2/target/linux/generic/files/drivers/net/phy/rtl8366_smi.c#L1466). It also gates on `@GPIO_SUPPORT` just like `kmod-mdio-gpio`, and econet gets `GPIO_SUPPORT` auto-added from the kernel config by [`include/target.mk`](https://github.com/openwrt/openwrt/blob/0e691dbababa9f0464666c3a7c7fb3b13a41e3b2/include/target.mk#L293), so nothing should exclude it from an econet all-kmods build.

Since the same "source-only target, never noticed" reasoning applies, was it left out deliberately (e.g. `CONFIG_RTL8366_SMI` not reachable in your build), or is it just outside the "mdio series" scope of this PR?

---

openwrt-ai (commented):
Reviewed 2 new commits. Both commit messages match their diffs (one target added per package, list kept alphabetical). One parity question inline.

Note on the bot's `PKG_RELEASE` warning: `package/kernel/linux/Makefile` has no `PKG_RELEASE`/`PKG_VERSION` to bump, so that suggestion looks like a false positive here.

---

robimarko:

PR #24711, opened 2026-08-13
mediatek: filogic: add support for ZBT Z8105AX-C

LS3434:
Add support for the ZBT Z8105AX-C router based on the MediaTek MT7981B SoC.

Specifications:

- SoC: MediaTek MT7981B
- RAM: 256 MiB
- Flash: Winbond SPI-NAND, 128 MiB
- Switch: MediaTek MT7531, 1 WAN + 3 LAN (Gigabit)
- Button: Reset, Mesh
- Power: DC 12 V / 2.5 A
- Wi-Fi: MT7981B 2.4 GHz + 5 GHz
- M.2 slot for a WWAN modem
- 2 full-size SIM slots, controlled via GPIO

The ZBT Z8105AX-C is externally identical to the ZBT Z8103AX-C and uses the same SoC, RAM and flash configuration.

LED layout (top to bottom):

- LAN 1: green, hardware controlled
- LAN 2: green, hardware controlled
- LAN 3: green, hardware controlled
- WAN: green, user controllable
- Mobile: green, user controllable
- Mesh: red/green/blue, used for OpenWrt status
- Power: green, hardware controlled

There is no Wi-Fi LED on the case and none was defined in the original firmware.

SIM slots:

The SIM slots are controlled via GPIOs named SIM1 and SIM2.

Installation:

A. Through U-Boot menu:

- Configure the computer with a static IP address in the 192.168.1.0/24 subnet, for example 192.168.1.10/24.
- Power off the router and hold the Reset button.
- Power on the router while keeping the Reset button pressed.
- Hold the button for approximately 10 seconds, then release it.
- Open 192.168.1.1 in a web browser.
- If a firmware flashing interface is displayed, the router is in the U-Boot recovery interface.
- Upload the factory image.

The U-Boot web interface can also be used to recover the router after an incorrect firmware flash.

B. Through the OpenWrt dashboard:

If the router comes with a vendor-modified OpenWrt installation, the firmware can be upgraded through the OpenWrt dashboard at 192.168.1.1:

System -> Backup / Flash Firmware

Flash the OpenWrt firmware.

Important: Do not keep the vendor configuration. Disable the "Keep settings" option when flashing. Vendor settings are incompatible with OpenWrt 24.10 and 25.12.

MAC addresses:

MAC addresses were found in the factory partition. The layout is identical to the one used by the ZBT Z8106AX-S:

- offset 0x4:  F8:5E:3C:xx:xx:aa — Router Label -2
- offset 0xa:  F8:5E:3C:xx:xx:bb — Router Label -1
- offset 0x24: F8:5E:3C:xx:xx:cc — Router Label +1
- offset 0x2a: F8:5E:3C:xx:xx:yy — MAC address printed on the router label

Untested:

- Flashing OpenWrt through the original firmware has not been tested. My unit came with a vendor-modified OpenWrt installation running kernel 5.4.
- Hardware watchdog support has not been tested. A hardware watchdog is defined in the stock firmware.

Notes:

The ZBT Z8105AX-C can be ordered from the vendor with different WWAN modem configurations. My unit came without a modem, but versions with 5G/LTE modems were available.

### DIFF
--- a/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dts
+++ b/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dts
@@ -1,0 +1,15 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+/dts-v1/;
+
+#include "mt7981b-zbtlink-zbt-z8105ax-c.dtsi"
+
+/ {
+	model = "Zbtlink ZBT-Z8105AX-C";
+	compatible = "zbtlink,zbt-z8105ax-c", "mediatek,mt7981b";
+};
+
+&port_4 {
+	/* The 4th port must be defined but isn't usable in model C, so we set it to disabled. */
+	status = "disabled";
+};
--- a/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dts
+++ b/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dts
@@ -1,0 +1,318 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+/dts-v1/;
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+#include <dt-bindings/pinctrl/mt65xx.h>
+
+#include "mt7981b.dtsi"
+
+/ {
+	model = "Zbtlink ZBT-Z8105AX-C";
+	compatible = "zbtlink,zbt-z8105ax-c", "mediatek,mt7981b";
+
+	aliases {
+		serial0 = &uart0;
+		led-boot = &led_status_green;
+		led-failsafe = &led_status_red;
+		led-running = &led_status_green;
+		led-upgrade = &led_status_red;
+		label-mac-device = &gmac1;
+	};
+
+	chosen {
+		stdout-path = "serial0:115200n8";
+		bootargs = "earlycon=uart8250,mmio32,0x11002000 console=ttyS0,115200n8";
+	};
+
+	gpio-keys {
+		compatible = "gpio-keys";
+
+		button-mesh {
+			label = "mesh";
+			linux,code = <BTN_0>;
+			linux,input-type = <EV_SW>;
+			gpios = <&pio 0 GPIO_ACTIVE_HIGH>;
+		};
+
+		button-reset {
+			label = "reset";
+			linux,code = <KEY_RESTART>;
+			gpios = <&pio 1 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		4g {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_MOBILE;
+			gpios = <&pio 34 GPIO_ACTIVE_LOW>;
+		};
+
+		wan {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WAN;
+			gpios = <&pio 8 GPIO_ACTIVE_LOW>;
+		};
+
+		led_status_blue: mesh-blue {
+			color = <LED_COLOR_ID_BLUE>;
+			function = LED_FUNCTION_STATUS;
+			gpios = <&pio 23 GPIO_ACTIVE_LOW>;
+			default-state = "off";
+		};
+
+		led_status_green: mesh-green {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_STATUS;
+			gpios = <&pio 24 GPIO_ACTIVE_LOW>;
+		};
+
+		led_status_red: mesh-red {
+			color = <LED_COLOR_ID_RED>;
+			function = LED_FUNCTION_STATUS;
+			gpios = <&pio 25 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+	watchdog {
+		compatible = "linux,wdt-gpio";
+		gpios = <&pio 2 GPIO_ACTIVE_HIGH>;
+		hw_algo = "toggle";
+		hw_margin_ms = <1000>;
+	};
+
+	gpio-export {
+		compatible = "gpio-export";
+
+		4g {
+			gpio-export,name = "4g";
+			gpio-export,output = <1>;
+			gpios = <&pio 4 GPIO_ACTIVE_HIGH>;
+		};
+
+		sim1 {
+			gpio-export,name = "sim1";
+			gpio-export,output = <0>;
+			gpios = <&pio 6 GPIO_ACTIVE_LOW>;
+		};
+
+		sim2 {
+			gpio-export,name = "sim2";
+			gpio-export,output = <0>;
+			gpios = <&pio 7 GPIO_ACTIVE_LOW>;
+		};
+	};
+};
+
+&eth {
+	status = "okay";
+
+	gmac0: mac@0 {
+		compatible = "mediatek,eth-mac";
+		reg = <0>;
+		phy-mode = "2500base-x";
+
+		nvmem-cell-names = "mac-address";
+		nvmem-cells = <&macaddr_factory_4 0>;
+
+		fixed-link {
+			speed = <2500>;
+			full-duplex;
+			pause;
+		};
+	};
+
+	gmac1: mac@1 {
+		compatible = "mediatek,eth-mac";
+		reg = <1>;
+		phy-mode = "gmii";
+		phy-handle = <&int_gbe_phy>;
+
+		nvmem-cell-names = "mac-address";
+		nvmem-cells = <&macaddr_factory_2a 0>;
+	};
+};
+
+&mdio_bus {
+	switch: switch@1f {
+		compatible = "mediatek,mt7531";
+		reg = <31>;
+		reset-gpios = <&pio 39 GPIO_ACTIVE_HIGH>;
+		interrupt-controller;
+		#interrupt-cells = <1>;
+		interrupt-parent = <&pio>;
+		interrupts = <38 IRQ_TYPE_LEVEL_HIGH>;
+
+		ports {
+			#address-cells = <1>;
+			#size-cells = <0>;
+
+			port@0 {
+				reg = <0>;
+				label = "lan1";
+			};
+
+			port@1 {
+				reg = <1>;
+				label = "lan2";
+			};
+
+			port@2 {
+				reg = <2>;
+				label = "lan3";
+			};
+
+			port@6 {
+				reg = <6>;
+				label = "cpu";
+				ethernet = <&gmac0>;
+				phy-mode = "2500base-x";
+
+				fixed-link {
+					speed = <2500>;
+					full-duplex;
+					pause;
+				};
+			};
+		};
+	};
+};
+
+&spi0 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&spi0_flash_pins>;
+	status = "okay";
+
+	spi_nand@0 {
+		compatible = "spi-nand";
+		reg = <0>;
+
+		spi-max-frequency = <52000000>;
+		spi-tx-bus-width = <4>;
+		spi-rx-bus-width = <4>;
+
+		mediatek,nmbm;
+		mediatek,bmt-max-ratio = <1>;
+		mediatek,bmt-max-reserved-blocks = <64>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "bl2";
+				reg = <0x0000000 0x0100000>;
+				read-only;
+			};
+
+			partition@100000 {
+				label = "u-boot-env";
+				reg = <0x100000 0x80000>;
+			};
+
+			partition@180000 {
+				label = "Factory";
+				reg = <0x180000 0x200000>;
+				read-only;
+
+				nvmem-layout {
+					compatible = "fixed-layout";
+					#address-cells = <1>;
+					#size-cells = <1>;
+
+					eeprom_factory: eeprom@0 {
+						reg = <0x0 0x1000>;
+					};
+
+					macaddr_factory_4: macaddr@4 {
+						reg = <0x4 0x6>;
+						compatible = "mac-base";
+						#nvmem-cell-cells = <1>;
+					};
+
+					macaddr_factory_a: macaddr@a {
+						reg = <0xa 0x6>;
+						compatible = "mac-base";
+						#nvmem-cell-cells = <1>;
+					};
+
+					macaddr_factory_2a: macaddr@2a {
+						reg = <0x2a 0x6>;
+						compatible = "mac-base";
+						#nvmem-cell-cells = <1>;
+					};
+				};
+			};
+
+			partition@380000 {
+				label = "FIP";
+				reg = <0x380000 0x200000>;
+				read-only;
+			};
+
+			nand_rootfs: partition@580000 {
+				label = "ubi";
+				reg = <0x580000 0x7280000>;
+			};
+		};
+	};
+};
+
+&pio {
+	spi0_flash_pins: spi0-pins {
+		mux {
+			function = "spi";
+			groups = "spi0", "spi0_wp_hold";
+		};
+
+		conf-pu {
+			pins = "SPI0_CS", "SPI0_HOLD", "SPI0_WP";
+			drive-strength = <MTK_DRIVE_8mA>;
+			bias-pull-up = <MTK_PUPD_SET_R1R0_11>;
+		};
+
+		conf-pd {
+			pins = "SPI0_CLK", "SPI0_MOSI", "SPI0_MISO";
+			drive-strength = <MTK_DRIVE_8mA>;
+			bias-pull-down = <MTK_PUPD_SET_R1R0_11>;
+		};
+	};
+};
+
+&uart0 {
+	status = "okay";
+};
+
+&usb_phy {
+	status = "okay";
+};
+
+&xhci {
+	status = "okay";
+};
+
+&wifi {
+	status = "okay";
+	#address-cells = <1>;
+	#size-cells = <0>;
+	nvmem-cells = <&eeprom_factory>;
+	nvmem-cell-names = "eeprom";
+
+	band@0 {
+		reg = <0>;
+		nvmem-cells = <&macaddr_factory_4 0>;
+		nvmem-cell-names = "mac-address";
+	};
+
+	band@1 {
+		reg = <1>;
+		nvmem-cells = <&macaddr_factory_a 1>;
+		nvmem-cell-names = "mac-address";
+	};
+};
--- a/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
+++ b/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
@@ -1,0 +1,326 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+/dts-v1/;
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+#include <dt-bindings/pinctrl/mt65xx.h>
+
+#include "mt7981b.dtsi"
+
+/ {
+	aliases {
+		serial0 = &uart0;
+		led-boot = &led_status_green;
+		led-failsafe = &led_status_red;
+		led-running = &led_status_green;
+		led-upgrade = &led_status_red;
+		label-mac-device = &gmac1;
+	};
+
+	chosen {
+		stdout-path = "serial0:115200n8";
+		bootargs = "earlycon=uart8250,mmio32,0x11002000 console=ttyS0,115200n8";
+	};
+
+	gpio-keys {
+		compatible = "gpio-keys";
+
+		button-mesh {
+			label = "mesh";
+			linux,code = <BTN_0>;
+			linux,input-type = <EV_SW>;
+			gpios = <&pio 0 GPIO_ACTIVE_HIGH>;
+		};
+
+		button-reset {
+			label = "reset";
+			linux,code = <KEY_RESTART>;
+			gpios = <&pio 1 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		4g {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_MOBILE;
+			gpios = <&pio 34 GPIO_ACTIVE_LOW>;
+		};
+
+		wan {
+			gpios = <&pio 8 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WAN;
+		};
+
+		led_status_blue: mesh-blue {
+			gpios = <&pio 23 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_BLUE>;
+			function = LED_FUNCTION_STATUS;
+			default-state = "off";
+		};
+
+		led_status_green: mesh-green {
+			gpios = <&pio 24 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_STATUS;
+		};
+
+		led_status_red: mesh-red {
+			gpios = <&pio 25 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_RED>;
+			function = LED_FUNCTION_STATUS;
+		};
+	};
+
+	watchdog {
+		compatible = "linux,wdt-gpio";
+		gpios = <&pio 2 GPIO_ACTIVE_HIGH>;
+		hw_algo = "toggle";
+		hw_margin_ms = <1000>;
+	};
+
+	gpio-export {
+		compatible = "gpio-export";
+
+		4g {
+			gpio-export,name = "4g";
+			gpio-export,output = <1>;
+			gpios = <&pio 4 GPIO_ACTIVE_HIGH>;
+		};
+
+		sim1 {
+			gpio-export,name = "sim1";
+			gpio-export,output = <0>;
+			gpios = <&pio 6 GPIO_ACTIVE_LOW>;
+		};
+
+		sim2 {
+			gpio-export,name = "sim2";
+			gpio-export,output = <0>;
+			gpios = <&pio 7 GPIO_ACTIVE_LOW>;
+		};
+	};
+};
+
+&eth {
+	status = "okay";
+
+	gmac0: mac@0 {
+		compatible = "mediatek,eth-mac";
+		reg = <0>;
+		phy-mode = "2500base-x";
+
+		nvmem-cell-names = "mac-address";
+		nvmem-cells = <&macaddr_factory_4 0>;
+
+		fixed-link {
+			speed = <2500>;
+			full-duplex;
+			pause;
+		};
+	};
+
+	gmac1: mac@1 {
+		compatible = "mediatek,eth-mac";
+		reg = <1>;
+		phy-mode = "gmii";
+		phy-handle = <&int_gbe_phy>;
+
+		nvmem-cell-names = "mac-address";
+		nvmem-cells = <&macaddr_factory_2a 0>;
+	};
+};
+
+&mdio_bus {
+	switch: switch@1f {
+		compatible = "mediatek,mt7531";
+		reg = <31>;
+		reset-gpios = <&pio 39 GPIO_ACTIVE_HIGH>;
+		interrupt-controller;
+		#interrupt-cells = <1>;
+		interrupt-parent = <&pio>;
+		interrupts = <38 IRQ_TYPE_LEVEL_HIGH>;
+	};
+};
+
+&switch {
+	ports {
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		port@0 {
+			reg = <0>;
+			label = "lan1";
+		};
+
+		port@1 {
+			reg = <1>;
+			label = "lan2";
+		};
+
+		port@2 {
+			reg = <2>;
+			label = "lan3";
+		};
+
+		port_4: port@4 {
+			reg = <4>;
+			label = "wwan";
+		};
+
+		port@6 {
+			reg = <6>;
+			label = "cpu";
+			ethernet = <&gmac0>;
+			phy-mode = "2500base-x";
+
+			fixed-link {
+				speed = <2500>;
+				full-duplex;
+				pause;
+			};
+		};
+	};
+};
+
+&spi0 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&spi0_flash_pins>;
+	status = "okay";
+
+	spi_nand@0 {
+		compatible = "spi-nand";
+		reg = <0>;
+
+		spi-max-frequency = <52000000>;
+		spi-tx-bus-width = <4>;
+		spi-rx-bus-width = <4>;
+
+		mediatek,nmbm;
+		mediatek,bmt-max-ratio = <1>;
+		mediatek,bmt-max-reserved-blocks = <64>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "bl2";
+				reg = <0x0000000 0x0100000>;
+				read-only;
+			};
+
+			partition@100000 {
+				label = "u-boot-env";
+				reg = <0x100000 0x80000>;
+			};
+
+			partition@180000 {
+				label = "Factory";
+				reg = <0x180000 0x200000>;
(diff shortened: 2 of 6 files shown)


target/linux/mediatek/filogic/base-files/etc/board.d/02_network
```diff
@@ -91,6 +91,9 @@ mediatek_setup_interfaces()
 	wavlink,wl-wn536ax6-a|\
 	zbtlink,zbt-z8102ax|\
 	zbtlink,zbt-z8102ax-v2|\
+	zbtlink,zbt-z8105ax-c)
+		ucidef_set_interfaces_lan_wan "lan1 lan2 lan3" "eth1"
+		;;
 	zbtlink,zbt-z8106ax-s)
 		ucidef_set_interfaces_lan_wan "lan1 lan2 lan3 lan4" "eth1"
 		;;
```
openwrt-ai: This terminates the pattern list that was previously ended by `zbtlink,zbt-z8106ax-s)`. The preceding `zbtlink,zbt-z8102ax-v2|\` (and the whole list above it) now falls through into the new `zbtlink,zbt-z8105ax-c)` case, so `asus,tuf-ax4200`, `iptime,ax3000sm`, `iptime,ax7800m-6e`, `jdcloud,re-cp-03`, `mediatek,mt7981-rfb`, `netcore,n60`, `netcore,n60-pro`, `nradio,c8-668gl`, `ruijie,rg-x60-pro`, `unielec,u7981-01*`, `wavlink,wl-wn536ax6-a`, `zbtlink,zbt-z8102ax` and `zbtlink,zbt-z8102ax-v2` all silently lose `lan4` and get `"lan1 lan2 lan3"` instead of `"lan1 lan2 lan3 lan4"`.

Give the new board its own case and restore the old group terminator:

```suggestion
	zbtlink,zbt-z8106ax-s)
		ucidef_set_interfaces_lan_wan "lan1 lan2 lan3 lan4" "eth1"
		;;
	zbtlink,zbt-z8105ax-c)
		ucidef_set_interfaces_lan_wan "lan1 lan2 lan3" "eth1"
		;;
```

---

target/linux/mediatek/filogic/base-files/etc/board.d/03_gpio_switches
```diff
@@ -26,6 +26,11 @@ zbtlink,zbt-z8106ax-t)
 	ucidef_add_gpio_switch "4g" "Power modem" "4g" "1"
 	ucidef_add_gpio_switch "sim" "SIM" "sim" "1"
 	;;
+zbtlink,zbt-z8105ax-c)
+	ucidef_add_gpio_switch "4g" "Power modem" "4g" "1"
+	ucidef_add_gpio_switch "sim1" "SIM 1" "sim1" "1"
+    ucidef_add_gpio_switch "sim2" "SIM 2" "sim2" "1"
+    ;;
```
openwrt-ai: These two lines are indented with 4 spaces while every other line in this file (including the two lines directly above) uses a tab.

```suggestion
	ucidef_add_gpio_switch "sim2" "SIM 2" "sim2" "1"
	;;
```

nit: the new case is also appended after `zbtlink,zbt-z8106ax-t)`, which breaks the ascending board-name order of this `case` (z8102ax, z8106ax-s/t) — `zbtlink,zbt-z8105ax-c)` belongs before the `zbtlink,zbt-z8106ax-s|` block.

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
```diff
@@ -0,0 +1,326 @@
...
+			label = "reset";
+			linux,code = <KEY_RESTART>;
+			gpios = <&pio 1 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+        4g {
```
openwrt-ai: Indented with 8 spaces instead of two tabs, unlike every other node in this file.

```suggestion
		4g {
```

The same applies to the `sim2 {` node at line 100.

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
```diff
@@ -0,0 +1,326 @@
...
+		reg = <1>;
+		nvmem-cells = <&macaddr_factory_a 1>;
+		nvmem-cell-names = "mac-address";
+	};
+};
+
+&factory {
+	compatible = "nvmem-cells";
+	#address-cells = <1>;
+	#size-cells = <1>;
```
openwrt-ai: `compatible = "nvmem-cells"` on the partition node is the legacy MTD nvmem path (`add_legacy_fixed_of_cells` in the kernel). No file in `target/linux/mediatek/dts/` uses it any more — all 167 nvmem-providing DTs there use an `nvmem-layout { compatible = "fixed-layout"; ... }` child of the partition instead.

Please move these cells into the `Factory` partition the way the sibling board does — see [`nvmem-layout` in mt7981b-zbtlink-zbt-z8106ax.dtsi:250](https://github.com/openwrt/openwrt/blob/3700289ffc12a20a3c9ea9d7dcf1887cb23016ee/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8106ax.dtsi#L245-L275) — and drop this whole `&factory` override (the `factory:` label on `partition@180000` then becomes unnecessary too).

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
```diff
@@ -0,0 +1,326 @@
...
+			pins = "SPI0_CS", "SPI0_HOLD", "SPI0_WP";
+			drive-strength = <8>;
+			mediatek,pull-up-adv = <0>; /* bias-disable */
+		};
+
+		conf-pd {
+			pins = "SPI0_CLK", "SPI0_MOSI", "SPI0_MISO";
+			drive-strength = <8>;
+			mediatek,pull-up-adv = <0>; /* bias-disable */
+		};
```
openwrt-ai: Two problems here:

1. `conf-pd` sets `mediatek,pull-up-adv` — a pull-*up* property in the pull-*down* group. Combined with the value `0` (bias-disable) both groups end up with no bias at all, whereas every other mt7981 board in this target biases these SPI-NAND pins (`bias-pull-up`/`bias-pull-down` with `MTK_PUPD_SET_R1R0_11`). `mediatek,pull-up-adv` appears in only 3 files under `target/linux/mediatek/dts/` and in none of them for `spi0-pins`.
2. `<dt-bindings/pinctrl/mt65xx.h>` is included at line 8 but never used; `drive-strength = <8>` should be `MTK_DRIVE_8mA`.

Matching [mt7981b-zbtlink-zbt-z8106ax.dtsi:300-311](``https://github.com/openwrt/openwrt/blob/3700289ffc12a20a3c9ea9d7dcf1887cb23016ee/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8106ax.dtsi#L300-L311):``

```suggestion
		conf-pu {
			pins = "SPI0_CS", "SPI0_HOLD", "SPI0_WP";
			drive-strength = <MTK_DRIVE_8mA>;
			bias-pull-up = <MTK_PUPD_SET_R1R0_11>;
		};

		conf-pd {
			pins = "SPI0_CLK", "SPI0_MOSI", "SPI0_MISO";
			drive-strength = <MTK_DRIVE_8mA>;
			bias-pull-down = <MTK_PUPD_SET_R1R0_11>;
		};
```

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
```diff
@@ -0,0 +1,326 @@
...
+	};
+
+	gpio-keys {
+		compatible = "gpio-keys";
+
+		wps {
+			label = "wps";
+			linux,code = <KEY_WPS_BUTTON>;
+			gpios = <&pio 0 GPIO_ACTIVE_LOW>;
+		};
```
openwrt-ai: The commit message and PR body both list only `Button: Reset` for this board, yet a WPS button is declared here. On the ZBT-Z8103AX — which the commit message says this board is externally identical to — GPIO 0 is not a WPS button but the mesh slide switch, declared as `EV_SW`/`BTN_0`: [`button-mesh` at mt7981b-zbtlink-zbt-z8103ax.dtsi:35](https://github.com/openwrt/openwrt/blob/3700289ffc12a20a3c9ea9d7dcf1887cb23016ee/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8103ax.dtsi#L35-L40) (and it is `GPIO_ACTIVE_HIGH` there, not `GPIO_ACTIVE_LOW`).

Was this GPIO actually verified as a momentary WPS button on the Z8105AX-C, or is it the same mesh switch? If it's the switch, the polarity and `linux,code`/`linux,input-type` here are wrong; if there is genuinely a WPS button, please add it to the button list in the commit message.

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
```diff
@@ -0,0 +1,326 @@
...
+	status = "okay";
+};
+
+&usb_phy {
+	status = "okay";
+};
+
+&xhci {
+	status = "okay";
+};
```
openwrt-ai: The sibling M.2/WWAN board wires the USB controller to two fixed regulators — [`vusb33-supply`/`vbus-supply` at mt7981b-zbtlink-zbt-z8106ax.dtsi:346-350](``https://github.com/openwrt/openwrt/blob/3700289ffc12a20a3c9ea9d7dcf1887cb23016ee/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8106ax.dtsi#L346-L350),`` backed by the `reg_3p3v`/`reg_5v` nodes at lines 81-98 — and this board also ships `kmod-usb3 kmod-usb-net-qmi-wwan kmod-usb-serial-option` for an M.2 modem on the same USB port. Here `&xhci` is enabled with no supplies and no regulator nodes at all.

Was the omission deliberate (e.g. verified working without them on a unit with no modem fitted), or should the `reg_3p3v`/`reg_5v` regulators be carried over so modem-equipped units enumerate? Since your unit came without a modem this may not have been exercised.

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
```diff
@@ -0,0 +1,326 @@
...
+			gpios = <&pio 24 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_STATUS;
+		};
+
+		led_status_red: red {
+			gpios = <&pio 25 GPIO_ACTIVE_LOW>;
+			color = <LED_COLOR_ID_RED>;
+			function = LED_FUNCTION_STATUS;
+		};
```
openwrt-ai: nit: the node names `blue`/`green`/`red` say nothing about which physical LED they are. Both sibling files name the RGB status LED's sub-nodes after the LED itself (`mesh-blue`/`mesh-green`/`mesh-red`, see [mt7981b-zbtlink-zbt-z8103ax.dtsi:58-75](``https://github.com/openwrt/openwrt/blob/3700289ffc12a20a3c9ea9d7dcf1887cb23016ee/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8103ax.dtsi#L58-L75)),`` which matches the "Mesh (red, green, blue)" entry in your own LED layout.

```suggestion
		led_status_blue: mesh-blue {
			gpios = <&pio 23 GPIO_ACTIVE_LOW>;
			color = <LED_COLOR_ID_BLUE>;
			function = LED_FUNCTION_STATUS;
			default-state = "off";
		};

		led_status_green: mesh-green {
			gpios = <&pio 24 GPIO_ACTIVE_LOW>;
			color = <LED_COLOR_ID_GREEN>;
			function = LED_FUNCTION_STATUS;
		};

		led_status_red: mesh-red {
			gpios = <&pio 25 GPIO_ACTIVE_LOW>;
			color = <LED_COLOR_ID_RED>;
			function = LED_FUNCTION_STATUS;
		};
```

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
```diff
@@ -0,0 +1,326 @@
...
+
+		port@2 {
+			reg = <2>;
+			label = "lan3";
+		};
+
+		port_4: port@4 {
+			reg = <4>;
+			label = "wwan";
+		};
```
openwrt-ai: nit: the commit message says "lan5 was removed from the DTS, while lan4 was retained ... Removing lan5 and disabling lan4 in the DTS works correctly", but no node here is labelled `lan4` or `lan5`. What actually happened is that `port@3` (labelled `lan4` on the sibling, [mt7981b-zbtlink-zbt-z8106ax.dtsi:184-187](https://github.com/openwrt/openwrt/blob/3700289ffc12a20a3c9ea9d7dcf1887cb23016ee/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8106ax.dtsi#L184-L187)) was dropped entirely, and `port@4` — labelled `wwan`, not `lan5` — is the one kept and disabled in the `.dts`. Please reword the paragraph in terms of the port numbers/labels that are actually in the DTS, so a later reader can match the text to the file.

---

openwrt-ai (commented):
Reviewed 1 new commit.

The commit message covers all three required sections for a new device (hardware spec, flash instructions, MAC layout) — thanks for the thorough write-up.

One blocking issue: the `02_network` change accidentally splits an existing `case` pattern list and regresses the LAN port set for ~12 unrelated boards. Details inline, along with a few DTS deviations from the sibling ZBT-Z8106AX/Z8103AX files this board is clearly derived from.

## Commit checks

- `3700289` "mediatek: filogic: add support for ZBT Z8105AX-C" — the message states the board has only a Reset button, but the DTS also declares a WPS button on GPIO 0 (the mesh switch pin on the otherwise-identical Z8103AX). The message's `lan4`/`lan5` narrative also does not match the DTS, which contains neither label. Both flagged inline.

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
LS3434: Z8106AX-S has a USB port, the Z8105AX-C does not. This is for internal m.2 slot.

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
```diff
@@ -0,0 +1,327 @@
...
+			gpio-export,output = <0>;
+			gpios = <&pio 6 GPIO_ACTIVE_LOW>;
+		};
+
+        sim2 {
+			gpio-export,name = "sim2";
+			gpio-export,output = <0>;
+			gpios = <&pio 7 GPIO_ACTIVE_LOW>;
+		};
+
```
openwrt-ai: The `4g` node was re-indented in this push but `sim2` still opens with 8 spaces instead of two tabs — every other node in the file uses tabs. There is also a stray blank line at 106 before the closing brace of `gpio-export`, which no other block in this file has.

```suggestion
		sim2 {
			gpio-export,name = "sim2";
			gpio-export,output = <0>;
			gpios = <&pio 7 GPIO_ACTIVE_LOW>;
		};
```

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
```diff
@@ -0,0 +1,327 @@
...
+			linux,code = <BTN_0>;
+			linux,input-type = <EV_SW>;
+			gpios = <&pio 0 GPIO_ACTIVE_HIGH>;
+		};
+
+		reset {
+			label = "reset";
+			linux,code = <KEY_RESTART>;
+			gpios = <&pio 1 GPIO_ACTIVE_LOW>;
+		};
```
openwrt-ai: Thanks for switching this to the mesh slide switch — the code/type/polarity now match the sibling exactly.

One thing left: the node names. `gpio-keys` child nodes must match [`"^(button|event|key|switch|(button|event|key|switch)-[a-z0-9-]+|[a-z0-9-]+-(button|event|key|switch))$"` in gpio-keys.yaml:26](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/input/gpio-keys.yaml#L26); plain `mesh` and `reset` match neither alternative and would be flagged by `dtbs_check`. Both siblings use the prefixed form — [`button-mesh` / `button-reset` at mt7981b-zbtlink-zbt-z8103ax.dtsi:35-47](https://github.com/openwrt/openwrt/blob/2df5ba9d8706c4391ce6d19dda7ad23ce929f1ae/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8103ax.dtsi#L35-L47).

```suggestion
		button-mesh {
			label = "mesh";
			linux,code = <BTN_0>;
			linux,input-type = <EV_SW>;
			gpios = <&pio 0 GPIO_ACTIVE_HIGH>;
		};

		button-reset {
			label = "reset";
			linux,code = <KEY_RESTART>;
			gpios = <&pio 1 GPIO_ACTIVE_LOW>;
		};
```

nit: the commit message was updated to `Buttons: Reset, Mesh`, but the PR description still lists only `- Button: Reset` and still carries the old `lan5 was removed from the DTS, while lan4 was retained...` paragraph that the commit message dropped. Worth syncing the PR body so the two agree.

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
```diff
@@ -0,0 +1,327 @@
...
+
+					macaddr_factory_2a: macaddr@2a {
+						reg = <0x2a 0x6>;
+						compatible = "mac-base";
+						#nvmem-cell-cells = <1>;
+					};
+
+					eeprom_factory: eeprom@0 {
+						reg = <0x0 0x1000>;
+					};
```
openwrt-ai: nit: `eeprom@0` is listed after `macaddr@2a`, so the `nvmem-layout` children are no longer in ascending unit-address order. The sibling this layout was copied from lists them `eeprom@0`, `macaddr@4`, `macaddr@a`, `macaddr@2a` — see [mt7981b-zbtlink-zbt-z8106ax.dtsi:255-275](``https://github.com/openwrt/openwrt/blob/2df5ba9d8706c4391ce6d19dda7ad23ce929f1ae/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8106ax.dtsi#L255-L275).``

```suggestion
					eeprom_factory: eeprom@0 {
						reg = <0x0 0x1000>;
					};

					macaddr_factory_4: macaddr@4 {
						reg = <0x4 0x6>;
						compatible = "mac-base";
						#nvmem-cell-cells = <1>;
					};

					macaddr_factory_a: macaddr@a {
						reg = <0xa 0x6>;
						compatible = "mac-base";
						#nvmem-cell-cells = <1>;
					};

					macaddr_factory_2a: macaddr@2a {
						reg = <0x2a 0x6>;
						compatible = "mac-base";
						#nvmem-cell-cells = <1>;
					};
```

---

openwrt-ai (commented):
Reviewed the force-push `3700289` → `2df5ba9`.

Every point from the previous round is addressed or answered: the `02_network` fall-through that regressed ~12 unrelated boards is fixed, the legacy `&factory { compatible = "nvmem-cells"; }` override became a proper `nvmem-layout` inside the `Factory` partition, the SPI-NAND pin group now uses `MTK_DRIVE_8mA` + `bias-pull-up`/`bias-pull-down`, the bogus WPS button is now the mesh slide switch matching the Z8103AX, the status LED sub-nodes are named `mesh-*`, and the `03_gpio_switches` indentation is tabs. Thanks for the quick turnaround — and for the note that the M.2 slot is internal, which explains the missing USB regulators.

CI on the head commit is green (mediatek/filogic kernel build, patch check and formality checks all pass).

Three small leftovers inline: one indentation slip that survived the fix-up, a `gpio-keys` node-naming issue that would trip `dtbs_check`, and a nit about node ordering in the new `nvmem-layout`.

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
```diff
@@ -0,0 +1,327 @@
...
+		sim2 {
+			gpio-export,name = "sim2";
+			gpio-export,output = <0>;
+			gpios = <&pio 7 GPIO_ACTIVE_LOW>;
+		};
+	};
+};
+
+
+&eth {
```
openwrt-ai: nit: removing the stray blank line inside `gpio-export` left a double blank line here instead — this is the only place in the file with two consecutive empty lines, and neither sibling DTS separates the root node from the first `&label` override with more than one.

```suggestion
&eth {
```

---

openwrt-ai (commented):
Reviewed the force-push `2df5ba9` → `70f8dc4` (DTS-only, 8 lines changed; commit message unchanged).

All three leftovers from the previous round are fixed: `sim2` is tab-indented, the `gpio-keys` children are renamed to `button-mesh` / `button-reset` so they match the `gpio-keys.yaml` node-name pattern, and `eeprom@0` moved to the front of the `nvmem-layout` so the children are in ascending unit-address order again.

CI on the head commit is green (mediatek filogic/mt7622/mt7623/mt7629 kernel builds, kernel-patch check and all three formality checks pass).

Only one cosmetic leftover inline. The PR description still lists `- Button: Reset` while the commit message now says `Buttons: Reset, Mesh` — worth syncing before merge (already noted on the previous round, not repeating inline).

---

openwrt-ai (commented):

The force-push `70f8dc4` → `271ecf5` is a single-line change: the double blank line before `&eth` in `mt7981b-zbtlink-zbt-z8105ax-c.dtsi` is gone, which was the last leftover from the previous round. Commit message is unchanged, and the PR description now matches it (`Buttons: Reset, Mesh` and the same `lan4`/`lan5` note).

CI on the head commit is green — mediatek filogic/mt7622/mt7623/mt7629 kernel builds, the filogic kernel-patch check and all three FormalityCheck runs pass.

---

LS3434:
Can someone please review this ? I think this is ready for final reviews. Ping @jonasjelonek

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
```diff
@@ -0,0 +1,326 @@
...
+		4g {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_MOBILE;
+			gpios = <&pio 34 GPIO_ACTIVE_LOW>;
+		};
+
+		wan {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WAN;
+                       gpios = <&pio 8 GPIO_ACTIVE_LOW>;
```
jonasjelonek: Tabs instead of spaces, please.

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
```diff
@@ -0,0 +1,326 @@
...
+		};
+
+		led_status_blue: mesh-blue {
+			color = <LED_COLOR_ID_BLUE>;
+			function = LED_FUNCTION_STATUS;
+                       gpios = <&pio 23 GPIO_ACTIVE_LOW>;
```
jonasjelonek: same here

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
```diff
@@ -0,0 +1,326 @@
...
+			default-state = "off";
+		};
+
+		led_status_green: mesh-green {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_STATUS;
+                       gpios = <&pio 24 GPIO_ACTIVE_LOW>;
```
jonasjelonek: same here

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
```diff
@@ -0,0 +1,326 @@
...
+		};
+
+		led_status_red: mesh-red {
+			color = <LED_COLOR_ID_RED>;
+			function = LED_FUNCTION_STATUS;
+                       gpios = <&pio 25 GPIO_ACTIVE_LOW>;
```
jonasjelonek: same here

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
```diff
@@ -0,0 +1,326 @@
...
+			gpio-export,output = <0>;
+			gpios = <&pio 6 GPIO_ACTIVE_LOW>;
+		};
+
+		sim2 {
+			gpio-export,name = "sim2";
+			gpio-export,output = <0>;
+			gpios = <&pio 7 GPIO_ACTIVE_LOW>;
+		};
+	};
```
jonasjelonek: Please check if that still works as expected with recent changes from https://github.com/openwrt/openwrt/commit/8f3285dbd0ad396f62b9815a71244ea650a9dd1b

jonasjelonek (changes requested):
Why is there a DTSI + DTS at this stage? Do you plan to add support for another device after this?

openwrt-ai (commented):

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
openwrt-ai: fixed, thanks

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
openwrt-ai: fixed, thanks

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
openwrt-ai: fixed, thanks

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
openwrt-ai: fixed, thanks

---

LS3434:
@jonasjelonek  ZBT also offers Z8105AX-M2-T, which has different case and i think  GPIO's for LED's also may differ. If someone has that device then i think it would be easier to add it. Both devices have the same partition layout and network configuration.

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
LS3434: I included this commit in my build and it works correctly. Checked after reboot:

```
root@OpenWrt:~# cat /sys/class/gpio/sim1/value
1
root@OpenWrt:~# cat /sys/class/gpio/sim2/value
1
```

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
LS3434: Done

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
LS3434: Done

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
LS3434: Done

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dtsi
LS3434: Done

openwrt-ai (commented):

---

jonasjelonek:
To be honest, that contains too much "if". I'd prefer to have the single DTS for that single device right now, and IF later someone wants to add support for that device, it is better to split it then knowing the exact differences between both.

LS3434:
@jonasjelonek  Ok, i will merge both dts files, and i will push an update later this evening or ealier if i find time.

EDIT: Done.

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dts
```diff
@@ -0,0 +1,334 @@
...
+
+		port@2 {
+			reg = <2>;
+			label = "lan3";
+		};
+
+		port_4: port@4 {
+			reg = <4>;
+			label = "wwan";
+		};
```
openwrt-ai: Now that the DTSI was folded in, the `port_4:` label and the `&port_4 { status = "disabled"; }` block at lines 331-334 override a node 150 lines above in the same file. Set the status inline and drop both the label and the trailing override — that is what other in-tree boards do for an unusable port, e.g. [`port@3` in mt7988d-tplink-be450-common.dtsi:355-358](``https://github.com/openwrt/openwrt/blob/0f2dcea2bf68323d0a08a016478af13bb6e07a55/target/linux/mediatek/dts/mt7988d-tplink-be450-common.dtsi#L355-L358).``

```suggestion
		/* the 4th port must be defined but isn't usable on this model */
		port@4 {
			reg = <4>;
			label = "wwan";
			status = "disabled";
		};
```

---

openwrt-ai (commented):
Reviewed 1 new commit.

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dts
LS3434: I deleted the 4th port entirely and router boots fine. So it's done.

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dts
```diff
@@ -0,0 +1,324 @@
...
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+#include <dt-bindings/pinctrl/mt65xx.h>
+
+#include "mt7981b.dtsi"
+
+/ {
+    model = "Zbtlink ZBT-Z8105AX-C";
```
openwrt-ai: This push re-indented the line with 4 spaces; every other line in the file uses a tab.

```suggestion
	model = "Zbtlink ZBT-Z8105AX-C";
```

---

openwrt-ai (commented):
Reviewed 1 new commit.

---

openwrt-ai (commented):

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dts
openwrt-ai: fixed, thanks

---

openwrt-ai (commented):

---

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dts
```diff
@@ -0,0 +1,324 @@
...
+			phy-mode = "2500base-x";
+
+			fixed-link {
+				speed = <2500>;
+				full-duplex;
+				pause;
+			};
+		};
+	};
+};
```
jonasjelonek, line 184: I'd prefer to merge those, it's not like this will end up in hilarious nesting. Better have the `ports` subnode directly defined within the switch node.

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dts
```diff
@@ -0,0 +1,324 @@
...
+
+&uart0 {
+	status = "okay";
+};
+
+&watchdog {
+	status = "okay";
+};
```
jonasjelonek: I think the upstream mt7981b.dtsi doesn't disable it by default so this shouldn't be needed.

jonasjelonek (changes requested):

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dts
LS3434: Done :)

target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8105ax-c.dts
LS3434: Done :)

jonasjelonek:

PR #24254, opened 2026-07-16
mac80211: ath11k: isolate RXDMA page-frag caches

eisbaw:
## Problem

On a Linksys MX4200 v1 / IPQ8074, stock ath11k repeatedly exhausted
available memory under ordinary traffic, eventually causing OOM kills of
`hostapd` and `wpa_supplicant`.

Generation-safe tracing found bounded RXDMA ring and IDR occupancy while
page-frag backing continued to grow. In two captures, 86–87% of the final
tracker-visible pages contained current or historical fragments from
different allocation origins.

This points to page-frag lifetime amplification from the shared per-CPU
cache, rather than a missing RX buffer free.

## Changes

This series:

- gives each RXDMA ring its own `page_frag_cache`;
- uses it for normal RX and monitor-status skb heads;
- aligns empty skb heads with `skb_reserve()` instead of the ineffective
  `skb_pull()`;
- drains each cache after ring-owned skbs and IDR entries are released;
- quiesces external IRQ and NAPI processing before crash teardown;
- makes external IRQ lifecycle transitions idempotent on AHB and
  multi-MSI PCI.

**No RX ring sizes or capacities are changed.**

Patch 951 depends on patch 950. Crash reconfiguration can disable HIF
IRQ/NAPI after reset or suspend has already done so. Patch 950 keeps the
repeated transition from adding extra Linux IRQ disable depth.

## Testing

Tested on a Linksys MX4200 v1 with IPQ8074 hw2.0, Linux 6.12.87, and
firmware `WLAN.HK.2.9.0.1-02146-QCAHKSWPL_SILICONZ-1`.

The integrated candidate completed:

- a guarded 30-minute natural-traffic test;
- a separate 110-minute resident run;
- sustained large-packet and 4,000 packets/s UDP pressure tests.

Memory and page-frag backing remained bounded, with no OOM, ath11k fatal
error, memguard intervention, or traffic-generator error.

The test modules contained an additional validation-only metadata marker;
their data-path code matched this branch. Only the AHB hardware path was
runtime-tested.

## Prior art and investigation

OpenWrt PR #21495 and
[Enspect's IPQ807x small-buffer work](https://github.com/enspect/ath11k-ipq807x-rx-buffer-oom-fix)
reduce RX ring sizes to lower memory consumption. This series instead
keeps ring capacity unchanged and separates page-frag allocation
lifetimes.

Detailed investigation:
https://blog.radix63.dk/posts/router-out-of-memory/

Capture summaries, simulator, and standalone patches:
[...]

PR #24713, opened 2026-08-13
mac80211: ath11k: use kmalloc-backed skbs for rx ring buffers

sonpython:
## What

Switch ath11k's three DMA rx ring buffer allocation sites (`ath11k_dp_rxbufs_replenish`, `ath11k_dp_rx_alloc_mon_status_buf`, `ath11k_ce_rx_post_pipe`) from `dev_alloc_skb()` to `alloc_skb(..., GFP_ATOMIC)` + explicit `skb_reserve(NET_SKB_PAD)`.

## Why

`dev_alloc_skb()` carves sub-page buffers from the shared per-CPU **page_frag cache**: 32KB compound pages filled append-only, released only when *every* skb sharing the chunk is freed. ath11k posts long-lived 2KB buffers to its DMA rings from this cache, interleaved with short-lived allocations. Because each recycled ring buffer lands in a *new* frag chunk, every surviving buffer can pin a whole 32KB page. Under steady PPDU/event churn (just an AP with associated stations, no data traffic needed) the pinned set climbs toward `ring_entries * 32KB` per ring.

On a low-RAM IPQ8074 device (Linksys MX4200, 512MB) this drained `MemAvailable` at **80-490 MB/day**, invisible to `slabinfo`/`vmstat`, only recoverable by reloading the module, and ending in OOM reboots. `page_owner` pinned the growth on order-3 `__netdev_alloc_skb` chunks from the mon-status-replenish and CE-rx-post paths specifically.

`alloc_skb()` uses kmalloc-backed heads that are freed independently; the slab allocator refills partial slabs, so pinning cannot compound. `NET_SKB_PAD` is reserved explicitly to keep the buffer layout identical to `dev_alloc_skb()`.

## Field validation

9x Linksys MX4200 (qualcommax/ipq807x, kernel 6.12.94, backports 6.18.39):
- **Before:** MemAvailable -80 to -490 MB/day, module reloads required, eventual OOM.
- **After:** MemAvailable flat (-2 to -6 MB/day, within noise) over 54h, **zero** reload interventions across all 9 nodes.

## Notes

Independent of #24658 (the `dp_rx_msdu_coalesce` use-after-free), which fixes a different bug. This is the allocation-strategy change for the page_frag pinning leak. Happy to route this to linux-wireless as well if maintainers prefer the driver change land upstream first.


package/kernel/mac80211/patches/ath11k/951-wifi-ath11k-use-kmalloc-backed-skbs-for-rx-ring-buffers.patch
```diff
@@ -0,0 +1,88 @@
...
++++ b/drivers/net/wireless/ath/ath11k/dp_rx.c
+@@ -371,11 +371,13 @@ int ath11k_dp_rxbufs_replenish(struct at
+ 	num_remain = req_entries;
+ 
+ 	while (num_remain > 0) {
+-		skb = dev_alloc_skb(DP_RX_BUFFER_SIZE +
+-				    DP_RX_BUFFER_ALIGN_SIZE);
++		skb = alloc_skb(DP_RX_BUFFER_SIZE +
++				DP_RX_BUFFER_ALIGN_SIZE +
++				NET_SKB_PAD, GFP_ATOMIC);
```
openwrt-ai, line 47: The headroom is identical, but the *tailroom* is not, so "Layout is kept identical to `dev_alloc_skb()`" in the commit message isn't quite right — and the difference is large enough to be worth stating explicitly.

`__netdev_alloc_skb()` builds the skb around the exact page-frag slice it asked for ([`len = SKB_HEAD_ALIGN(len); ... __build_skb(data, len)` at skbuff.c:740](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/net/core/skbuff.c#L740-L765)), whereas `__alloc_skb()` lets `kmalloc_reserve()` write the *rounded-up* bucket size back into `size` and then builds around that ([skbuff.c:670-685](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/net/core/skbuff.c#L670-L685)). With `DP_RX_BUFFER_SIZE` 2048 + `DP_RX_BUFFER_ALIGN_SIZE` 128 + `NET_SKB_PAD` 64 on arm64:

- before: request 2240 → page-frag slice of `SKB_HEAD_ALIGN(2240)` = 2560 B, `skb->end` = 2240, `skb_tailroom()` = 2176
- after: request 2240 → 2240 + 320 = 2560 rounded to the kmalloc-4096 bucket, `skb->end` = `SKB_WITH_OVERHEAD(4096)` = 3776, `skb_tailroom()` = 3712

Two consequences: each ring buffer now costs 4096 B of slab instead of a 2560 B frag slice (+60%), and every `dma_map_single(..., skb->len + skb_tailroom(skb), ...)` in these paths (dp_rx.c:387, :2853, ce.c) now maps 3712 B instead of 2176 B, with the matching `dma_unmap_single()` invalidating the same larger range per received buffer. With `DP_RXDMA_BUF_RING_SIZE` 4096 plus a 1024-entry mon-status ring per pdev that is roughly +10 MB steady state on a 3-radio IPQ8074, and this patch lands on every ath11k target, including the 256 MB IPQ5018/IPQ6018 boards, not just the 512 MB MX4200 you measured.

Still a clear net win against a leak that grows without bound, but could you note the fixed footprint increase in the commit message (and ideally a `MemFree` delta right after boot, before/after) so maintainers can weigh it for the small-RAM targets? Nothing here is a correctness problem as far as I can tell — the extra tailroom only relaxes the `space_extra`/`extra_space` checks in the coalesce and defrag paths, and map/unmap stay symmetric.

---

package/kernel/mac80211/patches/ath11k/951-wifi-ath11k-use-kmalloc-backed-skbs-for-rx-ring-buffers.patch
```diff
@@ -0,0 +1,88 @@
+From: Michael Phan <sonunix@gmail.com>
+Date: Wed, 13 Aug 2026 23:50:00 +0700
```
openwrt-ai, line 2: nit: two header problems in one place. The `From <hash> Mon Sep 17 00:00:00 2001` line is missing (every other patch in this directory has it — see `949-wifi-ath11k-fix-monitor-rx-pktlen.patch`, which uses an all-zero hash for a not-yet-upstream fix, and `948-wifi-ath11k-Fix-the-WMM-param-type.patch` for the upstream case), which is what the formality bot flagged. And the weekday in `Date:` is wrong: 2026-08-13 is a Thursday, and the commit itself is dated `Thu Aug 13 23:51:05 2026 +0700`.

```suggestion
From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
From: Michael Phan <sonunix@gmail.com>
Date: Thu, 13 Aug 2026 23:50:00 +0700
```

---

package/kernel/mac80211/patches/ath11k/951-wifi-ath11k-use-kmalloc-backed-skbs-for-rx-ring-buffers.patch
```diff
@@ -0,0 +1,88 @@
...
+
+Validated on a 9-node MX4200 fleet: MemAvailable flat (-2 to -6 MB/day,
+within noise) over 54h with zero driver-reload interventions, versus
+80-490 MB/day and eventual OOM before the change.
+
+Signed-off-by: Michael Phan <sonunix@gmail.com>
+---
+ drivers/net/wireless/ath/ath11k/ce.c    | 3 ++-
+ drivers/net/wireless/ath/ath11k/dp_rx.c | 9 ++++++---
+ 2 files changed, 8 insertions(+), 4 deletions(-)
```
openwrt-ai, line 35: nit: the diffstat doesn't match the diff below it. `dp_rx.c` has two hunks of +4/-2 each, i.e. 12 changed lines, not 9, and the totals are 10 insertions / 5 deletions, not 8/4. (The stated 8/4 happens to be the dp_rx.c-only figure.) The body also lists `dp_rx.c` before `ce.c` while the stat lists them alphabetically, which `git format-patch` would not produce.

```suggestion
 drivers/net/wireless/ath/ath11k/ce.c    |  3 ++-
 drivers/net/wireless/ath/ath11k/dp_rx.c | 12 ++++++++----
 2 files changed, 10 insertions(+), 5 deletions(-)
```

Relatedly, the second hunk header `@@ -2839,12 +2841,14 @@` points 3 lines past where that context actually starts in the ath11k tree I checked (`struct sk_buff *skb;` in `ath11k_dp_rx_alloc_mon_status_buf()` is at dp_rx.c:2836 in v6.18.44, and no earlier patch in `patches/ath11k/` touches dp_rx.c before line 4929), while the first hunk header lands exactly. Quilt will apply it with an offset either way, but running `make package/mac80211/refresh` would regenerate both the hunk headers and the stat from the real tree.

---

openwrt-ai (commented):
Reviewed 1 new commit. The change itself looks correct to me: all three ath11k rx-ring allocation sites are covered (the only remaining `dev_alloc_skb()` users in the driver are `ath11k_htc_alloc_skb()` and the HTC control buffer, which are tx-side and short-lived), `GFP_ATOMIC` is right for both call sites since they run under `srng->lock` / `ce.ce_lock`, and `alloc_skb(len + NET_SKB_PAD) + skb_reserve(NET_SKB_PAD)` reproduces `dev_alloc_skb()`'s headroom exactly. `PKG_RELEASE` is bumped. The premise checks out too: at 2240 bytes these requests fall squarely in `__netdev_alloc_skb()`'s page-frag window (above `SKB_WITH_OVERHEAD(SKB_SMALL_HEAD_CACHE_SIZE)`, below `SKB_WITH_OVERHEAD(PAGE_SIZE)`).

Three comments inline: one on a footprint side effect that the commit message currently understates, and two nits on the patch header. Nothing blocking.

---

robimarko:
Please submit upstream, ideally they will review and merge it first

neheb:
I asked LLM about this:

```
Genuine high-confidence matches, worth fixing (active, maintained):
- ath/ath12k/ce.c:72, dp_rx.c:117, dp_mon.c — copy-paste twins of ath11k; same 2048B buffers, same rings. This is the obvious companion patch.
- ath/ath10k/htt_rx.c:161 + pci.c:774/snoc.c:503 — same HTT/CE rings, 2048B.
```

I am genuinely confused. For ath10k, there are lowmem versions in the tree. Does this mean they would become redundant?

castiel652:
Looks similar to #24254

package/kernel/mac80211/patches/ath11k/951-wifi-ath11k-use-kmalloc-backed-skbs-for-rx-ring-buffers.patch
LGA1150: You can create a separate slab to allocate fixed-size buffers.

PR #24714, opened 2026-08-13
realtek: pcs: several improvements

jonasjelonek:
Another series with several improvements to the PCS driver, covering:
- some cleanups to improve readability or drop redundancies
- slight functional changes regarding the power lifecycling during SerDes configuration
- a new mask-based write helper intended to assist in giving several register writes real meaning

### DIFF
--- a/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
+++ b/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
@@ -3383,12 +3383,12 @@ static void rtpcs_931x_sds_rx_reset(struct rtpcs_serdes *sds)
 		return;
 
 	rtpcs_sds_write(sds, PAGE_ANA_10G, 0x12, 0x2740);
-	rtpcs_sds_write(sds, PAGE_ANA_10G_EXT, 0x0, 0x0);	/* [11:6] DFE_TAP3_ODD | [5:0] DFE_TAP3_EVEN */
+	rtpcs_sds_write_bits(sds, PAGE_ANA_10G_EXT, 0x0, 15, 12, 0x0);
 	rtpcs_sds_write(sds, PAGE_ANA_10G_EXT, 0x2, 0x2010);
 	rtpcs_sds_write(sds, PAGE_ANA_MISC, 0x0, 0xc10);
 
 	rtpcs_sds_write(sds, PAGE_ANA_10G, 0x12, 0x27c0);
-	rtpcs_sds_write(sds, PAGE_ANA_10G_EXT, 0x0, 0xc000); /* [11:6] DFE_TAP3_ODD | [5:0] DFE_TAP3_EVEN */
+	rtpcs_sds_write_bits(sds, PAGE_ANA_10G_EXT, 0x0, 15, 12, 0xc);
 	rtpcs_sds_write(sds, PAGE_ANA_10G_EXT, 0x2, 0x6010);
 	rtpcs_sds_write(sds, PAGE_ANA_MISC, 0x0, 0xc30);
 
@@ -3793,7 +3793,7 @@ static int rtpcs_931x_sds_config_attachment(struct rtpcs_serdes *sds,
 	 *   while rx_reset doesn't do this.
 	 */
 	rtpcs_sds_write(sds, PAGE_ANA_10G, 0x12, 0x2740);
-	rtpcs_sds_write(sds, PAGE_ANA_10G_EXT, 0x0, 0x0);	/* [11:6] DFE_TAP3_ODD | [5:0] DFE_TAP3_EVEN */
+	rtpcs_sds_write_bits(sds, PAGE_ANA_10G_EXT, 0x0, 15, 12, 0x0);
 	rtpcs_sds_write(sds, PAGE_ANA_10G_EXT, 0x2, 0x2010);
 	rtpcs_sds_write(sds, PAGE_ANA_MISC, 0x0, 0xcd1); /* from 930x: [7:6] POWER_DOWN OF ?? */
 
@@ -3846,7 +3846,7 @@ static int rtpcs_931x_sds_config_attachment(struct rtpcs_serdes *sds,
 
 	if (is_10g) {
 		rtpcs_sds_write(sds, PAGE_ANA_10G, 0x12, 0x27c0);
-		rtpcs_sds_write(sds, PAGE_ANA_10G_EXT, 0x0, 0xc000); /* [11:6] DFE_TAP3_ODD | [5:0] DFE_TAP3_EVEN */
+		rtpcs_sds_write_bits(sds, PAGE_ANA_10G_EXT, 0x0, 15, 12, 0xc);
 		rtpcs_sds_write(sds, PAGE_ANA_10G_EXT, 0x2, 0x6010);
 	}
 
--- a/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
+++ b/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
@@ -3824,42 +3824,12 @@ static int rtpcs_931x_sds_config_tx_amps(struct rtpcs_serdes *sds, u8 pre_amp, u
 	return rtpcs_sds_write_bits(sds, PAGE_ANA_10G, 0x0, 1, 0, en_val);
 }
 
-/**
- * rtpcs_931x_sds_config_tx - Configure static TX path parameters
- */
-static int rtpcs_931x_sds_config_tx(struct rtpcs_serdes *sds,
-				    enum rtpcs_sds_attachment attachment)
-{
-	const struct rtpcs_sds_tx_config *tx_cfg;
-
-	switch (attachment) {
-	case RTPCS_SDS_ATTACH_DAC_SHORT:
-		tx_cfg = &rtpcs_931x_sds_tx_cfg_sdac;
-		break;
-
-	case RTPCS_SDS_ATTACH_DAC_LONG:
-		tx_cfg = &rtpcs_931x_sds_tx_cfg_ldac;
-		break;
-
-	default:
-		if (sds->ctrl->chip_version == RTPCS_CHIP_V2)
-			/* consider 9311 vs. 9313 here too, see SDK */
-			tx_cfg = &rtpcs_931x_sds_tx_cfg_v2[sds->id - 2];
-		else
-			tx_cfg = &rtpcs_931x_sds_tx_cfg_v1[sds->id - 2];
-
-		break;
-	}
-
-	return rtpcs_931x_sds_config_tx_amps(sds, tx_cfg->pre_amp, tx_cfg->main_amp,
-					     tx_cfg->post_amp);
-}
-
 static int rtpcs_931x_sds_config_attachment(struct rtpcs_serdes *sds,
 					    enum rtpcs_sds_attachment attachment,
 					    enum rtpcs_sds_mode hw_mode)
 {
 	struct rtpcs_serdes *even_sds = rtpcs_sds_get_even(sds);
+	const struct rtpcs_sds_tx_config *tx_cfg;
 	bool is_dac, is_10g;
 	int ret;
 
@@ -3884,11 +3854,6 @@ static int rtpcs_931x_sds_config_attachment(struct rtpcs_serdes *sds,
 	if (attachment == RTPCS_SDS_ATTACH_NONE)
 		return 0;
 
-	/* config SerDes TX path (amps, impedance, etc.) */
-	ret = rtpcs_931x_sds_config_tx(sds, attachment);
-	if (ret < 0)
-		return ret;
-
 	is_dac = (attachment == RTPCS_SDS_ATTACH_DAC_SHORT ||
 		  attachment == RTPCS_SDS_ATTACH_DAC_LONG);
 	is_10g = (hw_mode == RTPCS_SDS_MODE_10GBASER ||
@@ -3904,6 +3869,9 @@ static int rtpcs_931x_sds_config_attachment(struct rtpcs_serdes *sds,
 	switch (attachment) {
 	case RTPCS_SDS_ATTACH_DAC_SHORT:
 	case RTPCS_SDS_ATTACH_DAC_LONG:
+		tx_cfg = (attachment == RTPCS_SDS_ATTACH_DAC_SHORT) ? &rtpcs_931x_sds_tx_cfg_sdac :
+								      &rtpcs_931x_sds_tx_cfg_ldac;
+
 		rtpcs_sds_write(sds, PAGE_ANA_COM, 0x19, 0xf0a5);
 		rtpcs_sds_write(even_sds, PAGE_ANA_10G, 0x8, 0x02a0); /* [10:7] TX impedance */
 		break;
@@ -3914,10 +3882,20 @@ static int rtpcs_931x_sds_config_attachment(struct rtpcs_serdes *sds,
 
 		fallthrough;
 	default:
+		if (sds->ctrl->chip_version == RTPCS_CHIP_V2)
+			tx_cfg = &rtpcs_931x_sds_tx_cfg_v2[sds->id - 2];
+		else
+			tx_cfg = &rtpcs_931x_sds_tx_cfg_v1[sds->id - 2];
+
 		rtpcs_sds_write(sds, PAGE_ANA_COM, 0x19, 0xf0f0);
 		rtpcs_sds_write(even_sds, PAGE_ANA_10G, 0x8, 0x0294); /* [10:7] TX impedance */
 		break;
 	}
+
+	ret = rtpcs_931x_sds_config_tx_amps(sds, tx_cfg->pre_amp, tx_cfg->main_amp,
+					    tx_cfg->post_amp);
+	if (ret)
+		return ret;
 
 	rtpcs_sds_write_mask(sds, PAGE_TGR_PRO_0, 0xd, RTL93XX_LINKDW_SEL,
 			     is_dac ? RTL93XX_LINKDW_SEL_DAC : RTL93XX_LINKDW_SEL_NON_DAC);
--- a/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
+++ b/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
@@ -3832,9 +3832,6 @@ static int rtpcs_931x_sds_config_tx(struct rtpcs_serdes *sds,
 {
 	const struct rtpcs_sds_tx_config *tx_cfg;
 
-	if (sds->type != RTPCS_SDS_TYPE_10G)
-		return 0;
-
 	switch (attachment) {
 	case RTPCS_SDS_ATTACH_DAC_SHORT:
 		tx_cfg = &rtpcs_931x_sds_tx_cfg_sdac;
@@ -3865,6 +3862,9 @@ static int rtpcs_931x_sds_config_attachment(struct rtpcs_serdes *sds,
 	struct rtpcs_serdes *even_sds = rtpcs_sds_get_even(sds);
 	bool is_dac, is_10g;
 	int ret;
+
+	if (sds->type != RTPCS_SDS_TYPE_10G)
+		return 0;
 
 	rtpcs_sds_write_bits(sds, PAGE_ANA_10G, 0xe, 13, 11, 0x0);
 	if (hw_mode != RTPCS_SDS_MODE_XSGMII)
--- a/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
+++ b/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
@@ -3480,10 +3480,12 @@ static void rtpcs_931x_sds_rx_reset(struct rtpcs_serdes *sds)
 		return;
 
 	rtpcs_931x_sds_10g_ana_pre(sds);
-	rtpcs_sds_write(sds, PAGE_ANA_MISC, 0x0, 0xc10);
+	rtpcs_sds_write_mask(sds, PAGE_ANA_MISC, 0x0, RTL93XX_FRC_RX_EN_MASK,
+			     RTL93XX_FRC_RX_EN_OFF);
 
 	rtpcs_931x_sds_10g_ana_post(sds);
-	rtpcs_sds_write(sds, PAGE_ANA_MISC, 0x0, 0xc30);
+	rtpcs_sds_write_mask(sds, PAGE_ANA_MISC, 0x0, RTL93XX_FRC_RX_EN_MASK,
+			     RTL93XX_FRC_RX_EN_ON);
 	msleep(50);
 }
 
--- a/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
+++ b/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
@@ -168,6 +168,10 @@ enum rtpcs_page {
 #define DIGI_1(page)	((page) + 0x40)
 #define DIGI_2(page)	((page) + 0x80)
 
+#define RTL93XX_FRC_CMU_EN_MASK		GENMASK(11, 10)
+#define RTL93XX_FRC_CMU_EN_UNFORCED	FIELD_PREP(RTL93XX_FRC_CMU_EN_MASK, 0x0)
+#define RTL93XX_FRC_CMU_EN_FORCE_OFF	FIELD_PREP(RTL93XX_FRC_CMU_EN_MASK, 0x1)
+#define RTL93XX_FRC_CMU_EN_FORCE_ON	FIELD_PREP(RTL93XX_FRC_CMU_EN_MASK, 0x3)
 #define RTL93XX_FRC_PDOWN_MASK		GENMASK(7, 6)
 #define RTL93XX_FRC_PDOWN_DOWN		FIELD_PREP(RTL93XX_FRC_PDOWN_MASK, 0x3)
 #define RTL93XX_FRC_PDOWN_UNFORCED	FIELD_PREP(RTL93XX_FRC_PDOWN_MASK, 0x0)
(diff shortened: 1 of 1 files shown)


target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -3374,24 +3450,50 @@ static int rtpcs_931x_sds_deactivate(struct rtpcs_serdes *sds)
...
+	if (sds->hw_mode != RTPCS_SDS_MODE_USXGMII &&
+	    sds->hw_mode != RTPCS_SDS_MODE_10GBASER) {
+		ret = rtpcs_sds_write_mask(sds, DIGI_1(PAGE_WDIG), 0x1, RTL931X_STOP_GLI_CLK, 0x0);
+		if (ret)
+			return ret;
+	}
+
+	ret = rtpcs_sds_write_mask(sds, PAGE_ANA_MISC, 0x0,
+				   RTL93XX_FRC_PDOWN_MASK | RTL93XX_FRC_RX_EN_MASK,
+				   RTL93XX_FRC_PDOWN_UNFORCED | RTL93XX_FRC_RX_EN_UNFORCED);
```
openwrt-ai: `RTL93XX_FRC_RX_EN_UNFORCED` looks like a mis-decomposition of the register value this replaces.

Before the series, `config_attachment` ended with `rtpcs_sds_write(sds, PAGE_ANA_MISC, 0x0, 0xc30)` followed by `rtpcs_sds_write_bits(sds, PAGE_ANA_MISC, 0x0, 9, 0, 0x30)`. In both, bits [5:4] are `0b11`, i.e. `RTL93XX_FRC_RX_EN_ON` — not unforced. `4dbc16f5` translated that write into `RTL93XX_FRC_RX_EN_UNFORCED`, and `e27ecf12` moved the (now wrong) value here, so after this series a SerDes finishes bring-up with FRC_RX_EN released instead of forced on.

Note that `bf71deb5` maps the very same encodings correctly in `rx_reset`: `0xc10` → `RX_EN_OFF`, `0xc30` → `RX_EN_ON`.

Nothing puts it back either: [`rtpcs_931x_sds_post_config()`](``https://github.com/openwrt/openwrt/blob/34d91763e20e2e0b30619ffeb30b08f361dc2f84/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L3928-L3948``) only runs `rx_reset()` via the PHY/10GBASE-R-fiber calibration paths, so DAC and non-10GBASE-R fiber attachments never hit it.

This is also what the commit message of `e27ecf12` promises — "the same FRC_PDOWN/FRC_RX_EN forcing 930x's set_power() already does" — and [`rtpcs_930x_sds_set_power()`](``https://github.com/openwrt/openwrt/blob/34d91763e20e2e0b30619ffeb30b08f361dc2f84/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L1894-L1905``) uses `RTL93XX_FRC_RX_EN_ON` for `on == true`.

```suggestion
	ret = rtpcs_sds_write_mask(sds, PAGE_ANA_MISC, 0x0,
				   RTL93XX_FRC_PDOWN_MASK | RTL93XX_FRC_RX_EN_MASK,
				   RTL93XX_FRC_PDOWN_UNFORCED | RTL93XX_FRC_RX_EN_ON);
```

If leaving RX_EN unforced is deliberate (i.e. hardware-tested to be equivalent), please say so in the commit message, since it is a functional change rather than the pure move the message describes.

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -3365,6 +3431,16 @@ static int rtpcs_931x_sds_deactivate(struct rtpcs_serdes *sds)
 {
 	int ret;
 
+	ret = rtpcs_sds_write_mask(sds, PAGE_ANA_MISC, 0x0,
+				   RTL93XX_FRC_PDOWN_MASK | RTL93XX_FRC_RX_EN_MASK,
+				   RTL93XX_FRC_PDOWN_DOWN | RTL93XX_FRC_RX_EN_OFF);
+	if (ret)
+		return ret;
+
+	ret = rtpcs_sds_write_mask(sds, DIGI_1(PAGE_WDIG), 0x1, RTL931X_STOP_GLI_CLK, 0x1);
```
openwrt-ai: The `set` argument is a raw `0x1` while the mask is a `BIT()` macro. It only happens to be correct because `STOP_GLI_CLK` is bit 0 — for any other bit position this pattern silently writes the wrong field, and it reads inconsistently against every other `write_mask()` call in this series, which all pass a pre-shifted value. Pass the bit itself (the clear in `rtpcs_931x_sds_activate()` on line 3457 can stay `0x0`).

```suggestion
	ret = rtpcs_sds_write_mask(sds, DIGI_1(PAGE_WDIG), 0x1, RTL931X_STOP_GLI_CLK,
				   RTL931X_STOP_GLI_CLK);
```

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -3722,151 +3823,100 @@ static int rtpcs_931x_sds_config_tx_amps(struct rtpcs_serdes *sds, u8 pre_amp, u
...
+		rtpcs_sds_write(even_sds, PAGE_ANA_10G, 0x8, 0x0294);
 		break;
 	}
 
-	/* LINKDW_SEL? (same semantics as 930x) */
-	rtpcs_sds_write_bits(sds, PAGE_TGR_PRO_0, 0xd, 6, 6, is_dac ? 0x0 : 0x1);
+	ret = rtpcs_931x_sds_config_tx_amps(sds, tx_cfg->pre_amp, tx_cfg->main_amp,
+					    tx_cfg->post_amp);
+	if (ret)
+		return ret;
```
openwrt-ai, line 3898: This moves the TX amp programming, not just its enclosing function. Previously `rtpcs_931x_sds_config_tx()` was called right after the `RTPCS_SDS_ATTACH_NONE` early return, i.e. *before* the `FRC_CMU_EN` unforce/force-on toggle around the `PAGE_ANA_5G0 0x7` write and before the `PAGE_ANA_COM 0x19` / TX-impedance writes. Now the amps (`PAGE_ANA_10G` regs 0x1/0x2/0x0) are written after all of those.

`34d91763`'s message ("drop it to make the code a little bit simpler") reads as a pure refactor, so the reordering is easy to miss. Was the new ordering verified on hardware, or should the `rtpcs_931x_sds_config_tx_amps()` call stay at the old position (right after the `ATTACH_NONE` return, with the `tx_cfg` selection kept in the switch)? Either way, please call the reorder out in the commit message.

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -3722,151 +3823,100 @@ static int rtpcs_931x_sds_config_tx_amps(struct rtpcs_serdes *sds, u8 pre_amp, u
...
+		rtpcs_sds_write(even_sds, PAGE_ANA_10G, 0x8, 0x02a0); /* [10:7] TX impedance */
+		break;
+
 	case RTPCS_SDS_ATTACH_DAC_LONG:
-		rtpcs_sds_write(sds, PAGE_ANA_COM, 0x19, 0xf0a5);	/* from XS1930-10 SDK */
-		rtpcs_sds_write(even_sds, PAGE_ANA_10G, 0x8, 0x02a0); /* [10:7] impedance */
+		tx_cfg = &rtpcs_931x_sds_tx_cfg_ldac;
+		rtpcs_sds_write(sds, PAGE_ANA_COM, 0x19, 0xf0a5);
+		rtpcs_sds_write(even_sds, PAGE_ANA_10G, 0x8, 0x02a0); /* [10:7] TX impedance */
 		break;
```
openwrt-ai, line 3877: nit: splitting the previously shared `DAC_SHORT`/`DAC_LONG` case duplicates the two register writes verbatim, which works against the "simpler and more condense" goal in the commit message. Keeping the cases merged and picking `tx_cfg` from `attachment` avoids the copy, e.g. a single case label with `tx_cfg = (attachment == RTPCS_SDS_ATTACH_DAC_SHORT) ? &rtpcs_931x_sds_tx_cfg_sdac : &rtpcs_931x_sds_tx_cfg_ldac;` before the two writes.

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -237,6 +267,9 @@ struct rtpcs_sds_ops {
 		    int bithigh, int bitlow);
 	int (*write)(struct rtpcs_serdes *sds, enum rtpcs_page page, int regnum,
 		     int bithigh, int bitlow, u16 value);
+	/* write an arbitrary, possibly non-contiguous bitmask in one transaction */
```
openwrt-ai: nit: "in one transaction" is inaccurate for the partial-mask path — `__rtpcs_sds_write_mask()` falls through to `mdiobus_c45_modify()`, which is a read-modify-write, i.e. two bus transactions ([mdio_bus.c:880](https://github.com/gregkh/linux/blob/v6.18.44/drivers/net/phy/mdio_bus.c#L880)). The actual benefit — which the commit message states correctly — is that one call covers a non-contiguous field set instead of N separate read-modify-writes.

```suggestion
	/* write an arbitrary, possibly non-contiguous bitmask in a single call */
```

---

openwrt-ai (commented):
## Commit checks

- `4dbc16f5` "realtek: pcs: add write_mask primitive" — the message says the decomposed writes only omit bitfields that are already at their reset value, but the decomposition of `rtpcs_sds_write(sds, PAGE_ANA_MISC, 0x0, 0xc30)` also *changes* one: `0xc30` has bits [5:4] = `0b11` (`RTL93XX_FRC_RX_EN_ON`), yet it is translated to `RTL93XX_FRC_RX_EN_UNFORCED`. See the inline comment on `rtpcs_931x_sds_activate()`.
- `e27ecf12` "realtek: pcs: rtl931x: move FRC_PDOWN/FRC_RX_EN out of config_attachment" — the message promises "the same FRC_PDOWN/FRC_RX_EN forcing 930x's `set_power()` already does", but `rtpcs_930x_sds_set_power(sds, true)` writes `RTL93XX_FRC_RX_EN_ON` while the new `activate()` writes `RTL93XX_FRC_RX_EN_UNFORCED`. Message and code disagree.
- `f9e111bf` "realtek: pcs: rtl931x: narrow register write for RX reset" — the subject scopes the change to the RX reset, but two of the three hunks are in `config_attachment` (the body even says "Stop clobbering them from config_attachment"). Either widen the subject or split the `config_attachment` hunks out.
- `4f05fd81` "realtek: pcs: rtl931x: add defines for FRC_CMU_EN fields" — besides the announced write_bits→write_mask conversion, the commit also *deletes* the `rtpcs_sds_write_bits(sds, PAGE_ANA_MISC, 0x0, 11, 10, 0x3); /* FRC_CMU_EN */` before the `FRC_V2ANALOG` write. It does look redundant (CMU_EN is already forced on at that point), but a dropped register write should be stated in the message rather than hidden in a rename commit.
- `34d91763` "realtek: pcs: rtl931x: merge TX config into caller" — the message reads as a pure inline of the helper, but the TX amp programming also moves from before the `FRC_CMU_EN` toggle block to after the `PAGE_ANA_COM`/TX-impedance writes. See the inline comment.
- nit — typos in commit bodies: `e2fbe534` "halves.Extract" (missing space); `0e4efe49` "the comment pointing already pointing to this"; `cc7fd414` "Since the is no apparent cycling" → "there is"; `c58bd659` "its call side" → "call site"; `34d91763` "more condense" → "more concise".

The `write_mask` primitive itself looks correct: `mdiobus_c45_modify()` takes `(mask, set)` in that order and the full-register shortcut for `mask == 0xffff` preserves the previous `mdiobus_c45_write()` behaviour, including through `__rtpcs_sds_write_raw()`. All four `rtpcs_sds_ops` instances are wired up, and gating `rtpcs_931x_sds_config_attachment()` on `RTPCS_SDS_TYPE_10G` matches what `rtpcs_930x_sds_config_attachment()` already does.

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
jonasjelonek: These are static settings not really being tied to any order or flow, being at set various places in the SDK at different times during a SerDes bringup lifecycle. The reorder here happens just out of that merge and is irrelevant to functionality. Though, it has been verified on hardware. Now, it matches the SDK even more.

Added this to the commit message.

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
jonasjelonek: Damn, shitty AI messed this up.

Fixed in the next push.

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -3722,151 +3823,98 @@ static int rtpcs_931x_sds_config_tx_amps(struct rtpcs_serdes *sds, u8 pre_amp, u
...
 
 	switch (attachment) {
 	case RTPCS_SDS_ATTACH_DAC_SHORT:
 	case RTPCS_SDS_ATTACH_DAC_LONG:
-		rtpcs_sds_write(sds, PAGE_ANA_COM, 0x19, 0xf0a5);	/* from XS1930-10 SDK */
-		rtpcs_sds_write(even_sds, PAGE_ANA_10G, 0x8, 0x02a0); /* [10:7] impedance */
+		tx_cfg = (attachment == RTPCS_SDS_ATTACH_DAC_SHORT) ? &rtpcs_931x_sds_tx_cfg_sdac :
+								      &rtpcs_931x_sds_tx_cfg_ldac;
+
+		tx_cfg = &rtpcs_931x_sds_tx_cfg_ldac;
```
openwrt-ai: The leftover unconditional assignment overwrites the ternary, so `RTPCS_SDS_ATTACH_DAC_SHORT` now gets `rtpcs_931x_sds_tx_cfg_ldac` — the short-DAC TX amp set (`rtpcs_931x_sds_tx_cfg_sdac`) becomes unreachable and dead.

This is a regression introduced by the last force-push: before it, the `DAC_SHORT` case had its own `tx_cfg = &rtpcs_931x_sds_tx_cfg_sdac;`, matching the pre-series [`rtpcs_931x_sds_config_tx()`](``https://github.com/openwrt/openwrt/blob/f9e111bf1bbbdd4b34c7c66be478b5bdd03e8c24/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L3729-L3737``) switch, which selected `sdac` for `DAC_SHORT` and `ldac` for `DAC_LONG`.

```suggestion
		tx_cfg = (attachment == RTPCS_SDS_ATTACH_DAC_SHORT) ? &rtpcs_931x_sds_tx_cfg_sdac :
								      &rtpcs_931x_sds_tx_cfg_ldac;

```

---

openwrt-ai (commented):
Re-reviewed the force-push (`34d91763` → `5975ef00`). The `FRC_RX_EN_UNFORCED` mis-decomposition, the raw `0x1` for `STOP_GLI_CLK`, the "one transaction" comment, and the missing reorder note in the TX-merge commit message are all resolved. The merged `DAC_SHORT`/`DAC_LONG` case introduced a new bug — see the inline comment.

## Commit checks

- `5975ef00` "realtek: pcs: rtl931x: merge TX config into caller" — the message describes an inline of the helper plus a TX-amp reorder, but the diff also silently changes short-DAC behaviour: the stray `tx_cfg = &rtpcs_931x_sds_tx_cfg_ldac;` left behind after the ternary makes `RTPCS_SDS_ATTACH_DAC_SHORT` use the long-DAC amps. See the inline comment.
- `589b8451` "realtek: pcs: rtl931x: add defines for FRC_CMU_EN fields" — unchanged from the previous round, so restating with the new SHA: besides the announced `write_bits`→`write_mask` conversion, the commit also *deletes* the `rtpcs_sds_write_bits(sds, PAGE_ANA_MISC, 0x0, 11, 10, 0x3); /* FRC_CMU_EN */` that preceded the `FRC_V2ANALOG` write. It does look redundant (CMU_EN is already forced on at that point), but a dropped register write is worth a line in the message rather than being folded into a rename commit.

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -1830,11 +1871,14 @@ static int rtpcs_930x_sds_wait_clock_ready(struct rtpcs_serdes *sds)
...
+	int power_down, rx_enable;
 
-	rtpcs_sds_write_bits(sds, PAGE_ANA_MISC, 0x00, 7, 6, power_down);
-	rtpcs_sds_write_bits(sds, PAGE_ANA_MISC, 0x00, 5, 4, rx_enable);
+	power_down = on ? RTL93XX_FRC_PDOWN_UNFORCED : RTL93XX_FRC_PDOWN_DOWN;
+	rx_enable = on ? RTL93XX_FRC_RX_EN_ON : RTL93XX_FRC_RX_EN_OFF;
+
+	rtpcs_sds_write_mask(sds, PAGE_ANA_MISC, 0x00,
+			     RTL93XX_FRC_PDOWN_MASK | RTL93XX_FRC_RX_EN_MASK,
+			     power_down | rx_enable);
```
Djfe: this cannot work, unless I'm wrong. `power_down` needs to be shifted by two.

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -3795,7 +3839,12 @@ static int rtpcs_931x_sds_config_attachment(struct rtpcs_serdes *sds,
 
 	/* SDK: media none behavior - baseline applied regardless of attachment */
 	rtpcs_931x_sds_10g_ana_pre(sds);
-	rtpcs_sds_write(sds, PAGE_ANA_MISC, 0x0, 0xcd1); /* from 930x: [7:6] POWER_DOWN OF ?? */
+
+	rtpcs_sds_write_mask(sds, PAGE_ANA_MISC, 0x0,
+			     RTL93XX_FRC_PDOWN_MASK | RTL93XX_FRC_RX_EN_MASK,
+			     RTL93XX_FRC_PDOWN_DOWN | RTL93XX_FRC_RX_EN_OFF);
```
Djfe: same as above

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -3847,7 +3896,11 @@ static int rtpcs_931x_sds_config_attachment(struct rtpcs_serdes *sds,
 	if (is_10g)
 		rtpcs_931x_sds_10g_ana_post(sds);
 
-	rtpcs_sds_write(sds, PAGE_ANA_MISC, 0x0, 0xc30);			/* from 930x: [7:6] POWER_DOWN OF ?? */
+	rtpcs_sds_write_mask(sds, PAGE_ANA_MISC, 0x0,
+			     RTL93XX_FRC_PDOWN_MASK | RTL93XX_FRC_RX_EN_MASK,
+			     RTL93XX_FRC_PDOWN_UNFORCED | RTL93XX_FRC_RX_EN_ON);
+	rtpcs_sds_write_bits(sds, PAGE_ANA_MISC, 0x0, 11, 10, 0x3); /* FRC_CMU_EN */
```
Djfe: same line as 3873 above (repetition that could be dropped, no?)

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -167,6 +168,34 @@ enum rtpcs_page {
...
+#define RTL93XX_FRC_V2ANALOG_MASK	GENMASK(1, 0)
+#define RTL93XX_FRC_V2ANALOG_UNFORCED	FIELD_PREP(RTL93XX_FRC_V2ANALOG_MASK, 0x0)
+#define RTL93XX_FRC_V2ANALOG_FORCE_OFF	FIELD_PREP(RTL93XX_FRC_V2ANALOG_MASK, 0x1)
+
+/*
+ * Gates a digital clock inside the SerDes. The exact block is unknown — it
+ * appears to be used by all modes except USXGMII and 10GBASE-R. The bit name
+ * is inherited from an earlier SerDes generation and has not been verified
+ * on RTL931x.
+ */
```
openwrt-ai, line 197: nit: this is the only define group in the block that doesn't name its register. The two groups directly above carry `/* TGR_PRO_0, reg 0x0d */` (line 171) and `/* ANA_MISC, reg 0x00 */` (line 176), and since `RTL931X_STOP_GLI_CLK` sits immediately after the ANA_MISC list a reader can easily take it for another ANA_MISC bit — it is actually `DIGI_1(PAGE_WDIG)` reg `0x1`, per its two call sites in [`rtpcs_931x_sds_deactivate()`](``https://github.com/openwrt/openwrt/blob/856a367384d17b8c63b8ec3065bf5acbcb89876e/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L3439-L3440``) and [`rtpcs_931x_sds_activate()`](``https://github.com/openwrt/openwrt/blob/856a367384d17b8c63b8ec3065bf5acbcb89876e/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L3457).``

```suggestion
/*
 * DIGI_1(WDIG), reg 0x01
 *
 * Gates a digital clock inside the SerDes. The exact block is unknown — it
 * appears to be used by all modes except USXGMII and 10GBASE-R. The bit name
 * is inherited from an earlier SerDes generation and has not been verified
 * on RTL931x.
 */
```

---

openwrt-ai (commented):
Re-reviewed the force-push (`5975ef00` → `856a3673`). The only code change in the range is the removal of the stray `tx_cfg = &rtpcs_931x_sds_tx_cfg_ldac;` in the merged `DAC_SHORT`/`DAC_LONG` case — `RTPCS_SDS_ATTACH_DAC_SHORT` now correctly resolves to `rtpcs_931x_sds_tx_cfg_sdac` again, matching the pre-series `rtpcs_931x_sds_config_tx()` switch. Everything else in the range is commit-message rewording, which covers the previously flagged items: the `narrow register write for baseline` subject now matches its two-function scope, the `add defines for FRC_CMU_EN fields` message now states the dropped `FRC_CMU_EN` write, the TX-merge message notes the reorder, and the typos are gone. All newly-added commit messages match their diffs, so no commit checks this round.

Two things I re-verified while walking the current head:

- `rtpcs_931x_sds_activate()` reads `sds->hw_mode` for the `STOP_GLI_CLK` decision, and `rtpcs_pcs_config()` assigns `sds->hw_mode` at [line 4239](https://github.com/openwrt/openwrt/blob/856a367384d17b8c63b8ec3065bf5acbcb89876e/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L4239) before calling `activate()` at [line 4242](``https://github.com/openwrt/openwrt/blob/856a367384d17b8c63b8ec3065bf5acbcb89876e/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L4242),`` so the new lifecycle placement sees the mode being brought up, not the previous one. The clear condition (`!= USXGMII && != 10GBASER`) is equivalent to the old `!is_10g || hw_mode == XSGMII` at the end of `config_attachment`.
- The `RTPCS_SDS_TYPE_10G` gate added to `rtpcs_931x_sds_config_attachment()` also removes a latent out-of-bounds path: on 931x only `sds->id >= 2` is typed `RTPCS_SDS_TYPE_10G` ([`rtpcs_931x_sds_probe()`](``https://github.com/openwrt/openwrt/blob/856a367384d17b8c63b8ec3065bf5acbcb89876e/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L4034-L4038)),`` and the function indexes `rtpcs_931x_sds_tx_cfg_v1/v2[sds->id - 2]` and `BIT(sds->id - 2)`, which would underflow for ids 0/1.

`tx_cfg` is assigned on every switch path (the `FIBER` case falls through to `default`), so no uninitialized use. CI is green on `856a3673`.

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -175,6 +175,15 @@ enum rtpcs_page {
...
 #define RTL93XX_FRC_RX_EN_OFF		FIELD_PREP(RTL93XX_FRC_RX_EN_MASK, 0x1)
 
+/* DIGI_1(WDIG), reg 0x01 */
+/*
+ * Gates a digital clock inside the SerDes. The exact block is unknown — it
+ * appears to be used by all modes except USXGMII and 10GBASE-R. The bit name
+ * is inherited from an earlier SerDes generation and has not been verified
+ * on RTL931x.
+ */
+#define RTL931X_STOP_GLI_CLK		BIT(0)
```
plappermaul, line 198: From https://www.svanheule.net/realtek/longan/register/wrap_sds_intf_ctrl0 this sounds like "GMII Line Interface / Gigabit Line Interface".

plappermaul (approved):

openwrt-ai (commented):
Reviewed 7 new commits; no new issues found.

The `856a3673` → `49cc09db` range is a rebase in which the only content change is a single added line — the `/* DIGI_1(WDIG), reg 0x01 */` header above `RTL931X_STOP_GLI_CLK` in `50232bd7` (`git diff` over the range reports `1 file changed, 1 insertion(+)`). That resolves the previously flagged nit, and the annotation matches both call sites, which write `DIGI_1(PAGE_WDIG)` reg `0x1` in [`rtpcs_931x_sds_deactivate()`](``https://github.com/openwrt/openwrt/blob/49cc09dbb84e81efd46b821de80437c95d29be7f/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L3440-L3441``) and [`rtpcs_931x_sds_activate()`](``https://github.com/openwrt/openwrt/blob/49cc09dbb84e81efd46b821de80437c95d29be7f/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L3458).`` The `reg 0xNN` form also matches the neighbouring `/* TGR_PRO_0, reg 0x0d */` and `/* ANA_MISC, reg 0x00 */` group headers.

All other commits in the range are unchanged apart from hunk line offsets, and every commit message still matches its diff, so no commit checks this round. CI reports no failures on `49cc09db`.

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
jonasjelonek: Thanks for the hint. Makes sense, so added this to the comment.

PR #24717, opened 2026-08-13
px5g-mbedtls: port to Mbed TLS 4.x

sanderpqr:
Mbed TLS 4.0 removes the legacy RNG interface and moves the low-level RSA and ECP key generation functions out of the public API. Key generation now goes through psa_generate_key(), with the result copied into a PK context via mbedtls_pk_copy_from_psa() for use with the X.509 writing API, which still takes mbedtls_pk_context.

psa_crypto_init() is now required before any crypto operation. The _urandom() callback is gone: the X.509 and PK write functions no longer take an RNG argument, and the certificate serial is generated with getrandom() directly and set with
mbedtls_x509write_crt_set_serial_raw().

Curve names are resolved against a local table mapping to PSA curve families, replacing mbedtls_ecp_curve_info_from_name(). x25519 and x448 are no longer accepted: they are key agreement curves and cannot sign a certificate. secp192r1, secp224r1, secp192k1 and secp224k1 were removed upstream.

The -3 option (RSA exponent 3) is rejected by rsakey with an error: psa_generate_key() always uses 65537.

Link tfpsacrypto rather than the legacy mbedcrypto alias.

_This PR is interdependent with https://github.com/openwrt/openwrt/pull/24705_

### DIFF
--- a/package/utils/px5g-mbedtls/px5g-mbedtls.c
+++ b/package/utils/px5g-mbedtls/px5g-mbedtls.c
@@ -32,14 +32,11 @@
 #include <stdbool.h>
 #include <errno.h>
 
-#include <mbedtls/bignum.h>
-#include <mbedtls/entropy.h>
 #include <mbedtls/x509_crt.h>
-#include <mbedtls/ecp.h>
-#include <mbedtls/rsa.h>
 #include <mbedtls/pk.h>
 #include <mbedtls/asn1.h>
 #include <mbedtls/oid.h>
+#include <psa/crypto.h>
 
 #define SET_OID(x, oid) \
 	do { x.len = MBEDTLS_OID_SIZE(oid); x.p = (unsigned char *) oid; } while (0)
@@ -49,17 +46,6 @@
 #define PX5G_LICENSE "Licensed under the GNU Lesser General Public License v2.1"
 
 static char buf[16384];
-
-static int _urandom(void *ctx, unsigned char *out, size_t len)
-{
-	ssize_t ret;
-
-	ret = getrandom(out, len, 0);
-	if (ret < 0 || (size_t)ret != len)
-		return MBEDTLS_ERR_ENTROPY_SOURCE_FAILED;
-
-	return 0;
-}
 
 static void write_file(const char *path, size_t len, bool pem, bool cert)
 {
@@ -102,21 +88,51 @@ static void write_file(const char *path, size_t len, bool pem, bool cert)
 		close(fd);
 }
 
-static mbedtls_ecp_group_id ecp_curve(const char *name)
-{
-	const mbedtls_ecp_curve_info *curve_info;
+static const struct {
+	const char *name;
+	psa_ecc_family_t family;
+	size_t bits;
+} ecp_curves[] = {
+#if defined(PSA_WANT_ECC_SECP_R1_256)
+	{ "P-256", PSA_ECC_FAMILY_SECP_R1, 256 },
+	{ "secp256r1", PSA_ECC_FAMILY_SECP_R1, 256 },
+#endif
+#if defined(PSA_WANT_ECC_SECP_R1_384)
+	{ "P-384", PSA_ECC_FAMILY_SECP_R1, 384 },
+	{ "secp384r1", PSA_ECC_FAMILY_SECP_R1, 384 },
+#endif
+#if defined(PSA_WANT_ECC_SECP_R1_521)
+	{ "P-521", PSA_ECC_FAMILY_SECP_R1, 521 },
+	{ "secp521r1", PSA_ECC_FAMILY_SECP_R1, 521 },
+#endif
+#if defined(PSA_WANT_ECC_SECP_K1_256)
+	{ "secp256k1", PSA_ECC_FAMILY_SECP_K1, 256 },
+#endif
+#if defined(PSA_WANT_ECC_BRAINPOOL_P_R1_256)
+	{ "brainpoolP256r1", PSA_ECC_FAMILY_BRAINPOOL_P_R1, 256 },
+#endif
+#if defined(PSA_WANT_ECC_BRAINPOOL_P_R1_384)
+	{ "brainpoolP384r1", PSA_ECC_FAMILY_BRAINPOOL_P_R1, 384 },
+#endif
+#if defined(PSA_WANT_ECC_BRAINPOOL_P_R1_512)
+	{ "brainpoolP512r1", PSA_ECC_FAMILY_BRAINPOOL_P_R1, 512 },
+#endif
+	{ NULL, 0, 0 },
+};
 
-	if (!strcmp(name, "P-256"))
-		return MBEDTLS_ECP_DP_SECP256R1;
-	else if (!strcmp(name, "P-384"))
-		return MBEDTLS_ECP_DP_SECP384R1;
-	else if (!strcmp(name, "P-521"))
-		return MBEDTLS_ECP_DP_SECP521R1;
-	curve_info = mbedtls_ecp_curve_info_from_name(name);
-	if (curve_info == NULL)
-		return MBEDTLS_ECP_DP_NONE;
-	else
-		return curve_info->grp_id;
+static int ecp_curve(const char *name, psa_ecc_family_t *family, size_t *bits)
+{
+	size_t i;
+
+	for (i = 0; ecp_curves[i].name; i++) {
+		if (!strcmp(name, ecp_curves[i].name)) {
+			*family = ecp_curves[i].family;
+			*bits = ecp_curves[i].bits;
+			return 0;
+		}
+	}
+
+	return -1;
 }
 
 static void write_key(mbedtls_pk_context *key, const char *path, bool pem)
@@ -135,40 +151,66 @@ static void write_key(mbedtls_pk_context *key, const char *path, bool pem)
 	write_file(path, len, pem, false);
 }
 
-static void gen_key(mbedtls_pk_context *key, bool rsa, int ksize, int exp,
-		    mbedtls_ecp_group_id curve, bool pem)
+static void gen_key(mbedtls_pk_context *key, bool rsa, int ksize,
+		    psa_ecc_family_t family, size_t bits)
 {
-	mbedtls_pk_init(key);
+	psa_key_attributes_t attr = PSA_KEY_ATTRIBUTES_INIT;
+	mbedtls_svc_key_id_t key_id = MBEDTLS_SVC_KEY_ID_INIT;
+	psa_status_t status;
+
+	if (!rsa && !family) {
+		fprintf(stderr, "error: no EC curves enabled in this build\n");
+		exit(1);
+	}
+
 	if (rsa) {
 		fprintf(stderr, "Generating RSA private key, %i bit long modulus\n", ksize);
-		mbedtls_pk_setup(key, mbedtls_pk_info_from_type(MBEDTLS_PK_RSA));
-		if (!mbedtls_rsa_gen_key(mbedtls_pk_rsa(*key), _urandom, NULL, ksize, exp))
-			return;
+		psa_set_key_type(&attr, PSA_KEY_TYPE_RSA_KEY_PAIR);
+		psa_set_key_bits(&attr, ksize);
+		psa_set_key_algorithm(&attr, PSA_ALG_RSA_PKCS1V15_SIGN(PSA_ALG_SHA_256));
 	} else {
 		fprintf(stderr, "Generating EC private key\n");
-		mbedtls_pk_setup(key, mbedtls_pk_info_from_type(MBEDTLS_PK_ECKEY));
-		if (!mbedtls_ecp_gen_key(curve, mbedtls_pk_ec(*key), _urandom, NULL))
-			return;
+		psa_set_key_type(&attr, PSA_KEY_TYPE_ECC_KEY_PAIR(family));
+		psa_set_key_bits(&attr, bits);
+		psa_set_key_algorithm(&attr, PSA_ALG_ECDSA(PSA_ALG_SHA_256));
 	}
-	fprintf(stderr, "error: key generation failed\n");
-	exit(1);
+	psa_set_key_usage_flags(&attr, PSA_KEY_USAGE_SIGN_HASH | PSA_KEY_USAGE_EXPORT);
+
+	status = psa_generate_key(&attr, &key_id);
+	if (status == PSA_ERROR_NOT_SUPPORTED) {
+		fprintf(stderr, "error: key type or size not supported in this build\n");
+		exit(1);
+	}
+	if (status != PSA_SUCCESS) {
+		fprintf(stderr, "error: key generation failed (%d)\n", (int) status);
+		exit(1);
+	}
+
+	mbedtls_pk_init(key);
+	if (mbedtls_pk_copy_from_psa(key_id, key) != 0) {
+		fprintf(stderr, "error: key import failed\n");
+		psa_destroy_key(key_id);
+		exit(1);
+	}
+	psa_destroy_key(key_id);
 }
 
 int dokey(bool rsa, char **arg)
 {
 	mbedtls_pk_context key;
 	unsigned int ksize = 512;
-	int exp = 65537;
 	char *path = NULL;
 	bool pem = true;
-	mbedtls_ecp_group_id curve = MBEDTLS_ECP_DP_SECP256R1;
+	psa_ecc_family_t family = ecp_curves[0].family;
+	size_t bits = ecp_curves[0].bits;
 
 	while (*arg && **arg == '-') {
 		if (!strcmp(*arg, "-out") && arg[1]) {
 			path = arg[1];
 			arg++;
-		} else if (!strcmp(*arg, "-3")) {
-			exp = 3;
+		} else if (rsa && !strcmp(*arg, "-3")) {
+			fprintf(stderr, "error: -3 is not supported, RSA keys use exponent 65537\n");
+			return 1;
 		} else if (!strcmp(*arg, "-der")) {
 			pem = false;
 		}
@@ -178,18 +220,15 @@ int dokey(bool rsa, char **arg)
 	if (*arg && rsa) {
 		ksize = (unsigned int)atoi(*arg);
 	} else if (*arg) {
-		curve = ecp_curve((const char *)*arg);
-		if (curve == MBEDTLS_ECP_DP_NONE) {
+		if (ecp_curve((const char *)*arg, &family, &bits)) {
 			fprintf(stderr, "error: invalid curve name: %s\n", *arg);
 			return 1;
 		}
 	}
 
-	gen_key(&key, rsa, ksize, exp, curve, pem);
+	gen_key(&key, rsa, ksize, family, bits);
 	write_key(&key, path, pem);
-
 	mbedtls_pk_free(&key);
-
 	return 0;
 }
 
@@ -197,7 +236,7 @@ int selfsigned(char **arg)
 {
 	mbedtls_pk_context key;
 	mbedtls_x509write_cert cert;
-	mbedtls_mpi serial;
+	unsigned char serial[8];
 	mbedtls_x509_san_list *san_list = NULL, *san_prev = NULL, *san_cur = NULL;
 	/*support
 	- MBEDTLS_X509_SAN_DNS_NAME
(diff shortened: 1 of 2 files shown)


package/utils/px5g-mbedtls/px5g-mbedtls.c
```diff
@@ -361,23 +371,21 @@ int selfsigned(char **arg)
...
-		sprintf(sstr + len*2, "%02x", (unsigned char) buf[len]);
-
-	mbedtls_mpi_init(&serial);
-	mbedtls_mpi_read_string(&serial, 16, sstr);
-	mbedtls_x509write_crt_set_serial(&cert, &serial);
+	if (getrandom(serial, sizeof(serial), 0) != (ssize_t) sizeof(serial)) {
+		fprintf(stderr, "error: failed to generate serial number\n");
+		return 1;
+	}
+	mbedtls_x509write_crt_set_serial_raw(&cert, serial, sizeof(serial));
```
openwrt-ai, line 407: The old path hex-printed the 8 random bytes and fed them through `mbedtls_mpi_read_string()` + `mbedtls_x509write_crt_set_serial()`, which encodes a non-negative MPI and inserts a leading `0x00` when the top bit is set. Handing 8 raw random bytes to `set_serial_raw()` means `serial[0]` has its MSB set roughly half the time, and RFC 5280 §4.1.2.2 requires a positive serial number (`serial[0] == 0x00`, 1-in-256, would also give a non-minimal DER INTEGER).

I could not reach the Mbed TLS 4.x sources from this environment to confirm whether `mbedtls_x509write_crt_set_serial_raw()`/the CRT DER writer prepends the `0x00` itself — if it does, this is a non-issue. If it doesn't, forcing the top byte into `0x01..0x7f` (e.g. `serial[0] = (serial[0] & 0x7f) | 0x01;` right after `getrandom()`) keeps the encoding valid without touching the rest of the entropy. Was the generated certificate checked with `openssl x509 -noout -serial` (or an ASN.1 dumper) to confirm the serial stays positive?

Separately, `mbedtls_x509write_crt_set_serial_raw()` returns an `int` that is discarded here; the other `mbedtls_x509write_crt_set_*` calls above ignore theirs too, so that is consistent with the file, but worth noting.

---

package/utils/px5g-mbedtls/px5g-mbedtls.c
```diff
@@ -102,21 +88,36 @@ static void write_file(const char *path, size_t len, bool pem, bool cert)
...
+	{ "P-256", PSA_ECC_FAMILY_SECP_R1, 256 },
+	{ "secp256r1", PSA_ECC_FAMILY_SECP_R1, 256 },
+	{ "P-384", PSA_ECC_FAMILY_SECP_R1, 384 },
+	{ "secp384r1", PSA_ECC_FAMILY_SECP_R1, 384 },
+	{ "P-521", PSA_ECC_FAMILY_SECP_R1, 521 },
+	{ "secp521r1", PSA_ECC_FAMILY_SECP_R1, 521 },
+	{ "secp256k1", PSA_ECC_FAMILY_SECP_K1, 256 },
+	{ "brainpoolP256r1", PSA_ECC_FAMILY_BRAINPOOL_P_R1, 256 },
+	{ "brainpoolP384r1", PSA_ECC_FAMILY_BRAINPOOL_P_R1, 384 },
+	{ "brainpoolP512r1", PSA_ECC_FAMILY_BRAINPOOL_P_R1, 512 },
```
openwrt-ai, line 118: The static table accepts a curve name regardless of whether the library was actually built with that curve, which moves a clean argument-parsing error into a hard `exit(1)` much later.

`mbedtls_ecp_curve_info_from_name()` only reported curves that were compiled in, so `px5g eckey brainpoolP256r1` used to fail immediately with `error: invalid curve name: brainpoolP256r1` and a plain `return 1` from `dokey()`. The companion PR #24705 leaves `PSA_WANT_ECC_BRAINPOOL_P_R1_256/384/512` at `default n` in `package/libs/mbedtls/Config.in`, so with the stock config those three entries now parse fine and then die inside [`gen_key()`](https://github.com/openwrt/openwrt/blob/32787c6912e88a824d5b035bdd664c60d38b24df/package/utils/px5g-mbedtls/px5g-mbedtls.c#L158-L161) with the generic `error: key generation failed` and `exit(1)` — no indication that the curve was the problem.

Either guard the table entries with the matching `PSA_WANT_ECC_*` macros so unbuilt curves are rejected at parse time, or have `gen_key()` print the `psa_status_t` and the requested curve so the failure is diagnosable. Same concern applies to `secp256k1` on line 102 for configs that turn `PSA_WANT_ECC_SECP_K1_256` off.

---

package/utils/px5g-mbedtls/px5g-mbedtls.c
```diff
@@ -386,14 +394,21 @@ int selfsigned(char **arg)
...
 
 int main(int argc, char *argv[])
 {
+	if (argv[1] && (!strcmp(argv[1], "eckey") || !strcmp(argv[1], "rsakey") ||
+			!strcmp(argv[1], "selfsigned"))) {
+		if (psa_crypto_init() != PSA_SUCCESS) {
+			fprintf(stderr, "error: failed to initialize PSA crypto\n");
+			return 1;
+		}
+	}
```
openwrt-ai: nit: this whitelist duplicates the dispatch chain immediately below it, so the two can drift — a subcommand added to the dispatch later will run with PSA uninitialised and every crypto call in it will fail with `PSA_ERROR_BAD_STATE`. `psa_crypto_init()` is cheap enough to run unconditionally before the dispatch, which removes the duplicated list:

```suggestion
	if (psa_crypto_init() != PSA_SUCCESS) {
		fprintf(stderr, "error: failed to initialize PSA crypto\n");
		return 1;
	}
```

---

package/utils/px5g-mbedtls/px5g-mbedtls.c
```diff
@@ -135,40 +136,52 @@ static void write_key(mbedtls_pk_context *key, const char *path, bool pem)
...
+	psa_ecc_family_t family = PSA_ECC_FAMILY_SECP_R1;
+	size_t bits = 256;
 
 	while (*arg && **arg == '-') {
 		if (!strcmp(*arg, "-out") && arg[1]) {
 			path = arg[1];
 			arg++;
-		} else if (!strcmp(*arg, "-3")) {
-			exp = 3;
 		} else if (!strcmp(*arg, "-der")) {
```
openwrt-ai, line 214: nit: dropping `-3` here makes it silently accepted-and-ignored rather than rejected — this parse loop skips any unrecognised `-` option, so `px5g rsakey -3 -out key.pem` still succeeds and just returns a 65537-exponent key with no diagnostic. Meanwhile px5g-wolfssl, which also carries `PROVIDES:=px5g` and is meant to be interchangeable with this build, still honours the option at [px5g-wolfssl.c:324](https://github.com/openwrt/openwrt/blob/32787c6912e88a824d5b035bdd664c60d38b24df/package/utils/px5g-wolfssl/px5g-wolfssl.c#L324-L325). The two `px5g` variants now disagree on the CLI depending on which one the image happens to install. Nothing in tree passes `-3` (uhttpd.init does not), so this is cosmetic, but it may be worth either keeping the option as an explicitly-rejected no-op or dropping it from px5g-wolfssl in the same series.

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message accurately describes the diff (PSA key generation, removed `_urandom()` callback, curve table, dropped `-3`, `tfpsacrypto` link), so no commit-message findings.

Four inline comments, the main ones being the certificate serial encoding and curve names that are now accepted regardless of the library's build config. Note that `mbedcrypto` → `tfpsacrypto` and the removed `mbedtls_ecp_*`/`mbedtls_rsa_*` APIs mean this cannot build against the in-tree Mbed TLS 3.6.7, so it does need to land together with #24705.

---

package/utils/px5g-mbedtls/px5g-mbedtls.c
sanderpqr: The DER writer handles it. `set_serial_raw()` stores the bytes verbatim, but `x509write_crt.c` prepends 0x00 when the MSb is set. Verified with 20 certs (11 with high bit set), all parse fine.

package/utils/px5g-mbedtls/px5g-mbedtls.c
```diff
@@ -135,40 +150,54 @@ static void write_key(mbedtls_pk_context *key, const char *path, bool pem)
...
 int dokey(bool rsa, char **arg)
 {
 	mbedtls_pk_context key;
 	unsigned int ksize = 512;
-	int exp = 65537;
 	char *path = NULL;
 	bool pem = true;
-	mbedtls_ecp_group_id curve = MBEDTLS_ECP_DP_SECP256R1;
+	psa_ecc_family_t family = PSA_ECC_FAMILY_SECP_R1;
+	size_t bits = 256;
```
openwrt-ai: The new `PSA_WANT_ECC_*` guards make the table config-aware, but the implicit default is still an unguarded hard-coded P-256, so one path into `gen_key()` bypasses the new check. `PSA_WANT_ECC_SECP_R1_256` is a user-visible option in [#24705's Config.in](https://github.com/openwrt/openwrt/blob/5b9edc604ee6c0596bd3b27d0dc13c68ec163751/package/libs/mbedtls/Config.in#L65-L67) (`default y`, but selectable); with it unset, `px5g eckey -out key.pem` — no curve argument, so `ecp_curve()` never runs — still reaches [`gen_key()`](https://github.com/openwrt/openwrt/blob/5ec57c088be5b01b1a04f03d69a3fc8940a6fa0a/package/utils/px5g-mbedtls/px5g-mbedtls.c#L169-L174) and exits with the generic `error: key generation failed (...)`, which is the failure mode the guards were added to remove. Same two lines in `selfsigned()` at 247-248.

Deriving both defaults from `ecp_curves[0]` (which the guards keep in preference order) or `#error`-ing out when no curve is compiled in would close it. The second form also covers the other edge of the guards: if all seven are off the initializer becomes `{ }`, which ISO C doesn't allow (GCC takes it as a zero-length-array extension) and leaves `ecp_curve()` rejecting every name.

---

openwrt-ai (commented):
Reviewed the force-pushed commit (32787c6 → 5ec57c0). The delta addresses the earlier comments: the curve table entries are now guarded by the matching `PSA_WANT_ECC_*` macros — which line up exactly with the symbols #24705 drives through `scripts/config.py` — `psa_generate_key()` now reports its `psa_status_t`, and `psa_crypto_init()` runs unconditionally. The commit message still matches the diff, so no commit-message findings.

One inline comment on a gap left by the new guards: the implicit P-256 default in `dokey()`/`selfsigned()` is not guarded, so the no-curve-argument path can still reach the generic key-generation failure.

---

package/utils/px5g-mbedtls/px5g-mbedtls.c
```diff
@@ -102,21 +88,58 @@ static void write_file(const char *path, size_t len, bool pem, bool cert)
 		close(fd);
 }
 
-static mbedtls_ecp_group_id ecp_curve(const char *name)
+#if !defined(PSA_WANT_ECC_SECP_R1_256) && !defined(PSA_WANT_ECC_SECP_R1_384) && \
+	!defined(PSA_WANT_ECC_SECP_R1_521) && !defined(PSA_WANT_ECC_SECP_K1_256) && \
+	!defined(PSA_WANT_ECC_BRAINPOOL_P_R1_256) && \
+	!defined(PSA_WANT_ECC_BRAINPOOL_P_R1_384) && \
+	!defined(PSA_WANT_ECC_BRAINPOOL_P_R1_512)
+#error "px5g requires at least one PSA_WANT_ECC_* curve to be enabled"
```
openwrt-ai: This turns an EC-only limitation into a hard build failure for configurations that only need RSA. px5g's only in-tree consumer is uhttpd, and its default key type is RSA — [`local KEY_OPTS="rsa:${bits:-2048}"` at uhttpd.init:52-54](https://github.com/openwrt/openwrt/blob/9b092c761f86163c6e3124b6975c6496f790ffe1/package/network/services/uhttpd/files/uhttpd.init#L52-L54) — the EC path is only taken when `key_type` is explicitly set to `ec`. All seven `PSA_WANT_ECC_*` symbols are plain user-selectable bools with no `depends` in [#24705's Config.in:65-91](https://github.com/openwrt/openwrt/blob/a63a4003f79470fff6e7e0979eedda19fd229aed/package/libs/mbedtls/Config.in#L65-L91), so menuconfig can legitimately produce a curve-less mbedtls. Before this patch px5g still built and generated RSA keys in that config; now the package stops here at build time and takes anything that selects it down with it.

Would putting the EC support (table, `ecp_curve()`, the `-newkey ec` / `eckey` branches) under `#if` and failing at runtime be preferable? That keeps the RSA path buildable while still giving a clear diagnostic when an EC key is actually requested, which is what the guards were added for.

nit: the seven macros listed here duplicate the guards on the table below, so the two can drift — adding an eighth entry with a new `PSA_WANT_ECC_*` guard leaves this condition unaware of it and the `#error` then fires even though the table is non-empty. A `_Static_assert(sizeof(ecp_curves) / sizeof(ecp_curves[0]) > 0, "...")` placed after the table derives the same check from the table itself.

---

package/utils/px5g-mbedtls/px5g-mbedtls.c
```diff
@@ -135,40 +158,57 @@ static void write_key(mbedtls_pk_context *key, const char *path, bool pem)
...
+	size_t bits = ecp_curves[0].bits;
 
 	while (*arg && **arg == '-') {
 		if (!strcmp(*arg, "-out") && arg[1]) {
 			path = arg[1];
 			arg++;
 		} else if (!strcmp(*arg, "-3")) {
-			exp = 3;
+			fprintf(stderr, "error: -3 is not supported, RSA keys use exponent 65537\n");
+			return 1;
```
openwrt-ai: nit: `dokey()` backs both `eckey` and `rsakey`, so `px5g eckey -3` now exits 1 with "RSA keys use exponent 65537" — a diagnostic about an exponent for a command that has none. px5g-wolfssl, which is interchangeable via `PROVIDES:=px5g`, accepts it there and ignores it ([`exp = 3` at px5g-wolfssl.c:325-326](https://github.com/openwrt/openwrt/blob/9b092c761f86163c6e3124b6975c6496f790ffe1/package/utils/px5g-wolfssl/px5g-wolfssl.c#L325-L326) is only consumed on the RSA branch), so the two variants diverge on a command where the option has no meaning either way. Gating on `rsa` keeps the error where it says something:

```suggestion
		} else if (rsa && !strcmp(*arg, "-3")) {
			fprintf(stderr, "error: -3 is not supported, RSA keys use exponent 65537\n");
			return 1;
```

Also worth noting the commit message's "The -3 option (RSA exponent 3) is rejected with an error" only holds for `rsakey`: `selfsigned()`'s option loop starting at line 261 has no `-3` branch and its `while` silently skips unrecognised `-` options, so `px5g selfsigned -3 -newkey rsa:2048 ...` still succeeds with exponent 65537 and no diagnostic. That matches the pre-patch behaviour (the old `selfsigned()` hard-coded `exp = 65537` as well), but the two subcommands of the same binary now disagree about the option.

---

openwrt-ai (commented):
Reviewed the force-pushed commit (5ec57c0 → 9b092c7). The delta does three things: derives the implicit key defaults from `ecp_curves[0]` instead of a hard-coded P-256, adds an `#error` when no `PSA_WANT_ECC_*` curve is enabled, and rejects `-3` in `dokey()`. The first closes the unguarded-default gap from the last round, and I confirmed the guards are sound — [#24705](https://github.com/openwrt/openwrt/pull/24705) rewrites `tf-psa-crypto/include/psa/crypto_config.h` via `scripts/config.py` and installs `usr/include/psa` in `Build/InstallDev`, so the `PSA_WANT_ECC_*` macros px5g tests are the ones the build actually used.

## Commit checks
- 9b092c7 "px5g-mbedtls: port to Mbed TLS 4.x" — the message doesn't mention that px5g now hard-requires at least one EC curve at build time (the new `#error`); that's a new constraint on the package's config space, not just a port detail. The `-3` paragraph also over-states the change: `selfsigned()` still ignores the option (details inline).

Two inline comments, the substantive one being that the `#error` makes an RSA-only mbedtls config fail to build px5g.

---

package/utils/px5g-mbedtls/px5g-mbedtls.c
sanderpqr: I've taken a different approach, as this is a follow-up on the previous review comment(s). Dropped the #error and kept only the per-entry table guards. With all seven PSA_WANT_ECC_* unset, px5g still builds (RSA-only).

package/utils/px5g-mbedtls/px5g-mbedtls.c
sanderpqr: selfsigned() ignored the option before this change and continues to do so.

package/utils/px5g-mbedtls/px5g-mbedtls.c
```diff
@@ -135,40 +150,61 @@ static void write_key(mbedtls_pk_context *key, const char *path, bool pem)
...
 int dokey(bool rsa, char **arg)
 {
 	mbedtls_pk_context key;
 	unsigned int ksize = 512;
-	int exp = 65537;
 	char *path = NULL;
 	bool pem = true;
-	mbedtls_ecp_group_id curve = MBEDTLS_ECP_DP_SECP256R1;
+	psa_ecc_family_t family = ecp_curves[0].family;
+	size_t bits = ecp_curves[0].bits;
```
openwrt-ai, line 205: Dropping the `#error` leaves these two reads out of bounds in exactly the config it was protecting. With all seven `PSA_WANT_ECC_*` unset, every entry in [`ecp_curves[]` (lines 94-120)](https://github.com/openwrt/openwrt/blob/1e8055ab23ab0c343d682c3b571061bf952f445b/package/utils/px5g-mbedtls/px5g-mbedtls.c#L94-L120) is preprocessed away and the initializer becomes `{ }`, so `ecp_curves` is a zero-length array and `ecp_curves[0]` reads past its end. Same two lines in `selfsigned()` at [254-255](https://github.com/openwrt/openwrt/blob/1e8055ab23ab0c343d682c3b571061bf952f445b/package/utils/px5g-mbedtls/px5g-mbedtls.c#L254-L255).

It does still build — I checked that GCC 13 accepts the empty initializer and the OOB read silently under the package's `-Wall -Werror` ([Makefile:56](https://github.com/openwrt/openwrt/blob/1e8055ab23ab0c343d682c3b571061bf952f445b/package/utils/px5g-mbedtls/Makefile#L56)); it only becomes a hard error under `-std=cNN -pedantic` (`error: zero or negative size array 'ecp_curves'`). So the RSA-only path works by relying on a GNU zero-length-array extension plus a read of whatever `.rodata` follows, rather than by construction.

A terminating sentinel keeps the array non-empty and gives the defaults a defined value, which the new `PSA_ERROR_NOT_SUPPORTED` branch above then turns into a sensible message when `px5g eckey` is run on a curve-less build:

```c
	{ NULL, 0, 0 },
};
```

with `ecp_curve()`'s loop stopping on `!ecp_curves[i].name` instead of running to `sizeof(ecp_curves) / sizeof(ecp_curves[0])` (otherwise it `strcmp()`s against `NULL`). Guarding the initialisers directly (`sizeof(ecp_curves) ? ecp_curves[0].family : 0`) works too, but leaves the non-ISO-C empty initializer in place.

---

openwrt-ai (commented):
Reviewed the force-pushed commit (9b092c7 → 1e8055a). The delta drops the `#error`, adds a dedicated `PSA_ERROR_NOT_SUPPORTED` message in `gen_key()`, and gates `-3` on `rsa` — the last two close the previous round's comments, and the commit message now matches the code on `-3` ("rejected by rsakey"), so no commit-message findings.

One inline comment: removing the `#error` without touching the `ecp_curves[0]` defaults means the curve-less RSA-only config now indexes a zero-length array instead of failing the build.

The `mbedtls_x509write_crt_der`/`_pem` "too few arguments" CI failures are the in-tree Mbed TLS 3.6 API, i.e. the #24705 interdependency already noted in the PR description — not something to fix here.

---

openwrt-ai (commented):

The delta (1e8055a → 49510f9) is the sentinel fix from the last round: `{ NULL, 0, 0 }` terminates `ecp_curves[]`, `ecp_curve()` now stops on `!ecp_curves[i].name` instead of `sizeof(...)`, and `gen_key()` rejects `!rsa && !family` with a dedicated message. That closes the out-of-bounds `ecp_curves[0]` read in the curve-less RSA-only config — the array is never empty now, `dokey()`/`selfsigned()` default to a defined `family = 0`, and no PSA ECC family constant is `0` (`PSA_ECC_FAMILY_SECP_R1` 0x12, `SECP_K1` 0x17, `BRAINPOOL_P_R1` 0x30), so `!family` is a safe "no curve compiled in" test. The non-ISO-C empty initializer is gone as well. Both `ecp_curves[0]` sites (lines 204-205 and 260-265) are covered by the single `gen_key()` check they both funnel through.

The commit message is unchanged and still matches the diff, so no commit-message findings.

The remaining `mbedtls_x509write_crt_der`/`_pem` "too few arguments" build failures on x86/64 and malta/be are the in-tree Mbed TLS 3.6 API, i.e. the #24705 interdependency already noted in the PR description — nothing to fix here.

---

sanderpqr:
Ready for review. Ping @jow- as maintainer of this package.

PR #24718, opened 2026-08-13
realtek: pcs: discover SerDes MDIO bus via phandle

jonasjelonek:
The driver located its SerDes MDIO bus by scanning the whole DT for a node with the generic "realtek,otto-serdes-mdio" compatible. That compatible exists only as a search key for this driver, doesn't describe real hardware, and can't disambiguate more than one match.

Add an explicit "mdio-parent-bus" phandle from the pcs node to the mdio-serdes node instead, following the same property already used by mdio-mux-mmioreg and the ksz DSA driver for referencing a secondary MDIO controller. Use dev_err_probe() for the lookup failures, which also quiets the log during the expected -EPROBE_DEFER retries while the MDIO bus driver comes up. Adjust the naming of local variables and function a bit.

This also drops the now stale compatibles from SoC DTSIs and fixes a pre-existing bug where the MDIO bus device hasn't been 'put' on driver unload/remove. Thus, refcount isn't properly reduced as expected by of_mdio_find_bus.

### DIFF
--- a/target/linux/realtek/dts/rtl838x.dtsi
+++ b/target/linux/realtek/dts/rtl838x.dtsi
@@ -244,13 +244,15 @@
 		};
 
 		mdio_serdes: mdio-serdes {
-			compatible = "realtek,rtl8380-serdes-mdio", "realtek,otto-serdes-mdio";
+			compatible = "realtek,rtl8380-serdes-mdio";
 		};
 
 		pcs {
 			compatible = "realtek,rtl8380-pcs", "realtek,otto-pcs";
 			#address-cells = <1>;
 			#size-cells = <0>;
+
+			mdio-parent-bus = <&mdio_serdes>;
 
 			serdes0: serdes@0 {
 				reg = <0>;
--- a/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
+++ b/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
@@ -4264,31 +4264,39 @@ static int rtpcs_pcs_config(struct phylink_pcs *pcs, unsigned int neg_mode,
 	return ret;
 }
 
-static struct mii_bus *rtpcs_probe_serdes_bus(struct rtpcs_ctrl *ctrl)
+static void rtpcs_mdio_bus_put(void *data)
 {
-	struct device_node *np;
-	struct mii_bus *bus;
+	struct mii_bus *mdio_bus = data;
 
-	np = of_find_compatible_node(NULL, NULL, "realtek,otto-serdes-mdio");
-	if (!np) {
-		dev_err(ctrl->dev, "SerDes mdio bus not found in DT");
-		return ERR_PTR(-ENODEV);
+	put_device(&mdio_bus->dev);
+}
+
+static struct mii_bus *rtpcs_find_mdio_bus(struct rtpcs_ctrl *ctrl)
+{
+	struct device_node *mdio_np;
+	struct mii_bus *mdio_bus;
+	int ret;
+
+	mdio_np = of_parse_phandle(ctrl->dev->of_node, "mdio-parent-bus", 0);
+	if (!mdio_np)
+		return ERR_PTR(dev_err_probe(ctrl->dev, -ENODEV, "MDIO parent bus not found\n"));
+
+	if (!of_device_is_available(mdio_np)) {
+		of_node_put(mdio_np);
+		return ERR_PTR(dev_err_probe(ctrl->dev, -ENODEV, "MDIO parent bus not usable\n"));
 	}
 
-	if (!of_device_is_available(np)) {
-		dev_err(ctrl->dev, "SerDes mdio bus not usable");
-		of_node_put(np);
-		return ERR_PTR(-ENODEV);
-	}
+	mdio_bus = of_mdio_find_bus(mdio_np);
+	of_node_put(mdio_np);
+	if (!mdio_bus)
+		return ERR_PTR(dev_err_probe(ctrl->dev, -EPROBE_DEFER,
+					     "MDIO parent bus not (yet) active\n"));
 
-	bus = of_mdio_find_bus(np);
-	of_node_put(np);
-	if (!bus) {
-		dev_warn(ctrl->dev, "SerDes mdio bus not (yet) active");
-		return ERR_PTR(-EPROBE_DEFER);
-	}
+	ret = devm_add_action_or_reset(ctrl->dev, rtpcs_mdio_bus_put, mdio_bus);
+	if (ret)
+		return ERR_PTR(ret);
 
-	return bus;
+	return mdio_bus;
 }
 
 static void rtpcs_sds_put_fwnode(void *data)
@@ -4454,7 +4462,7 @@ static int rtpcs_probe(struct platform_device *pdev)
 	if (IS_ERR(ctrl->map))
 		return PTR_ERR(ctrl->map);
 
-	ctrl->bus = rtpcs_probe_serdes_bus(ctrl);
+	ctrl->bus = rtpcs_find_mdio_bus(ctrl);
 	if (IS_ERR(ctrl->bus))
 		return PTR_ERR(ctrl->bus);
 
--- a/target/linux/realtek/dts/rtl839x.dtsi
+++ b/target/linux/realtek/dts/rtl839x.dtsi
@@ -257,13 +257,15 @@
 		};
 
 		mdio_serdes: mdio-serdes {
-			compatible = "realtek,rtl8392-serdes-mdio", "realtek,otto-serdes-mdio";
+			compatible = "realtek,rtl8392-serdes-mdio";
 		};
 
 		pcs {
 			compatible = "realtek,rtl8392-pcs", "realtek,otto-pcs";
 			#address-cells = <1>;
 			#size-cells = <0>;
+
+			mdio-parent-bus = <&mdio_serdes>;
 
 			serdes0: serdes@0 {
 				reg = <0>;
--- a/target/linux/realtek/dts/rtl930x.dtsi
+++ b/target/linux/realtek/dts/rtl930x.dtsi
@@ -244,13 +244,15 @@
 		};
 
 		mdio_serdes: mdio-serdes {
-			compatible = "realtek,rtl9301-serdes-mdio", "realtek,otto-serdes-mdio";
+			compatible = "realtek,rtl9301-serdes-mdio";
 		};
 
 		pcs {
 			compatible = "realtek,rtl9301-pcs", "realtek,otto-pcs";
 			#address-cells = <1>;
 			#size-cells = <0>;
+
+			mdio-parent-bus = <&mdio_serdes>;
 
 			serdes0: serdes@0 {
 				reg = <0>;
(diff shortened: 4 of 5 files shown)


target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -4216,31 +4216,27 @@ static int rtpcs_pcs_config(struct phylink_pcs *pcs, unsigned int neg_mode,
...
+	struct mii_bus *mdio_bus;
 
-	np = of_find_compatible_node(NULL, NULL, "realtek,otto-serdes-mdio");
-	if (!np) {
-		dev_err(ctrl->dev, "SerDes mdio bus not found in DT");
-		return ERR_PTR(-ENODEV);
-	}
+	mdio_np = of_parse_phandle(ctrl->dev->of_node, "mdio-parent-bus", 0);
+	if (!mdio_np)
+		return ERR_PTR(dev_err_probe(ctrl->dev, -ENODEV, "MDIO parent bus not found"));
```
openwrt-ai: `dev_err_probe()` does not append a newline — it formats as `"error %pe: %pV"` ([`__dev_probe_failed()` at core.c:5093](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/base/core.c#L5087-L5093)), so these messages get concatenated with the next log line. Every other `dev_err_probe()`/`dev_err()` in this file terminates with `\n` (e.g. [pcs-rtl-otto.c:4300](``https://github.com/openwrt/openwrt/blob/4df9be14c5d6c1e12f01838473947d9961dd1328/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L4300));`` the old `dev_err()` calls replaced here were the odd ones out.

```suggestion
		return ERR_PTR(dev_err_probe(ctrl->dev, -ENODEV, "MDIO parent bus not found\n"));
```

Same applies to the "MDIO parent bus not usable" string on line 4230 and "MDIO parent bus not (yet) active" on line 4237.

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -4216,31 +4216,27 @@ static int rtpcs_pcs_config(struct phylink_pcs *pcs, unsigned int neg_mode,
...
+		return ERR_PTR(dev_err_probe(ctrl->dev, -ENODEV, "MDIO parent bus not usable"));
 	}
 
-	bus = of_mdio_find_bus(np);
-	of_node_put(np);
-	if (!bus) {
-		dev_warn(ctrl->dev, "SerDes mdio bus not (yet) active");
-		return ERR_PTR(-EPROBE_DEFER);
-	}
+	mdio_bus = of_mdio_find_bus(mdio_np);
```
openwrt-ai, line 4289: `of_mdio_find_bus()` returns the bus with `get_device()` already taken and [documents that the caller must put it](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/mdio_bus.c#L393-L397). `rtpcs_probe()` stores the result in `ctrl->bus` and the driver has no `.remove` and no devm action for it, so the reference is never dropped on unbind or on a later probe failure — the SerDes MDIO bus device stays pinned. This is pre-existing (the old `rtpcs_probe_serdes_bus()` had the same issue), but since this function is being rewritten anyway, a `devm_add_action_or_reset(ctrl->dev, ..., &mdio_bus->dev)` doing `put_device()` would close it; compare [`mdio_mux_uninit()` at mdio-mux.c:264](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/mdio/mdio-mux.c#L259-L264), which is the mdio-mux counterpart of the same phandle lookup.

---

target/linux/realtek/dts/rtl838x.dtsi
```diff
@@ -252,6 +252,8 @@
 			#address-cells = <1>;
 			#size-cells = <0>;
 
+			mdio-parent-bus = <&mdio_serdes>;
```
openwrt-ai, line 255: nit: with the phandle in place, `"realtek,otto-serdes-mdio"` on [`mdio_serdes` at rtl838x.dtsi:247](https://github.com/openwrt/openwrt/blob/4df9be14c5d6c1e12f01838473947d9961dd1328/target/linux/realtek/dts/rtl838x.dtsi#L247) has no remaining consumer in-tree: the MDIO driver matches only the SoC-specific compatibles ([`rtsds_of_match[]` at mdio-realtek-otto-serdes.c:517](``https://github.com/openwrt/openwrt/blob/4df9be14c5d6c1e12f01838473947d9961dd1328/target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-serdes.c#L517)),`` and this PR removes the only `of_find_compatible_node()` lookup for it. The commit message argues the compatible "exists only as a search key for this driver [and] doesn't describe real hardware", which reads like it should go away here too (same for rtl839x/rtl930x/rtl931x.dtsi). Is keeping it intentional? Note the sibling generic strings are not in the same position — `"realtek,otto-mdio"` and `"realtek,otto-switch"` are still live search keys in [common.c:260](``https://github.com/openwrt/openwrt/blob/4df9be14c5d6c1e12f01838473947d9961dd1328/target/linux/realtek/files-6.18/drivers/net/dsa/rtl83xx/common.c#L260-L276).``

---

openwrt-ai (commented):
Reviewed 1 new commit; 3 inline comments. Commit message matches the diff (DTS phandle + driver lookup rewrite). Verified `mdio-parent-bus` is the property used by mdio-mux and the ksz DSA driver in Linux v6.18.44, and that all four SoC dtsi provide the `mdio_serdes` label the new phandle references.

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -4216,31 +4216,37 @@ static int rtpcs_pcs_config(struct phylink_pcs *pcs, unsigned int neg_mode,
 	return ret;
 }
 
-static struct mii_bus *rtpcs_probe_serdes_bus(struct rtpcs_ctrl *ctrl)
+static void rtpcs_mdio_bus_put(void *data)
 {
-	struct device_node *np;
-	struct mii_bus *bus;
+	put_device(&((struct mii_bus *)data)->dev);
+}
```
openwrt-ai, line 4272: nit: the two other devres action callbacks in this file unpack `void *data` into a typed local first — [`rtpcs_sds_put_fwnode()` at pcs-rtl-otto.c:4252-4257](``https://github.com/openwrt/openwrt/blob/dd609e025001012bcbbfee0fefc8ca6c6d3139a5/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L4252-L4257``) and [`rtpcs_del_provider_action()` at pcs-rtl-otto.c:4259](``https://github.com/openwrt/openwrt/blob/dd609e025001012bcbbfee0fefc8ca6c6d3139a5/target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c#L4259-L4271``) — rather than casting inline. Matching them keeps the file consistent.

```suggestion
static void rtpcs_mdio_bus_put(void *data)
{
	struct mii_bus *mdio_bus = data;

	put_device(&mdio_bus->dev);
}
```

---

target/linux/realtek/files-6.18/drivers/net/pcs/pcs-rtl-otto.c
```diff
@@ -4216,31 +4216,37 @@ static int rtpcs_pcs_config(struct phylink_pcs *pcs, unsigned int neg_mode,
...
 	}
 
-	return bus;
+	mdio_bus = of_mdio_find_bus(mdio_np);
+	of_node_put(mdio_np);
+	if (!mdio_bus)
+		return ERR_PTR(dev_err_probe(ctrl->dev, -EPROBE_DEFER,
+					     "MDIO parent bus not (yet) active\n"));
+
+	ret = devm_add_action_or_reset(ctrl->dev, rtpcs_mdio_bus_put, mdio_bus);
```
openwrt-ai, line 4295: nit: this is a behavioural fix — it drops the `get_device()` reference `of_mdio_find_bus()` takes, which the old `rtpcs_probe_serdes_bus()` leaked — but the commit message only mentions the phandle lookup, `dev_err_probe()` and renames. Worth a sentence in the body so the refcount fix isn't invisible to `git log` readers (and so it's clear it's intentional scope creep rather than an accident).

Ordering is correct as written: registering the put action here, before the `rtpcs_sds_put_fwnode` / `rtpcs_del_provider_action` actions in `rtpcs_probe()`, means devres LIFO releases the bus reference last.

---

target/linux/realtek/dts/rtl838x.dtsi
```diff
@@ -244,14 +244,16 @@
 		};
 
 		mdio_serdes: mdio-serdes {
-			compatible = "realtek,rtl8380-serdes-mdio", "realtek,otto-serdes-mdio";
+			compatible = "realtek,rtl8380-serdes-mdio";
```
openwrt-ai, line 247: nit: the commit-message paragraph covering this hunk has mismatched quotes — `Drop the now stale 'realtek,otto-serdes-mdio" compatibles` opens with `'` and closes with `"`. Should be `"realtek,otto-serdes-mdio"`.

Also, the PR description still only describes the phandle change and doesn't mention dropping the compatibles at all, so body and commit message now disagree about the scope of the change. Same hunk in rtl839x.dtsi:260, rtl930x.dtsi:247 and rtl931x.dtsi:280.

Verified the removal is safe: [`rtsds_of_match[]` at mdio-realtek-otto-serdes.c:517-536](``https://github.com/openwrt/openwrt/blob/dd609e025001012bcbbfee0fefc8ca6c6d3139a5/target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-serdes.c#L517-L536)`` lists all four SoC-specific compatibles, and `realtek,otto-serdes-mdio` has no remaining occurrence anywhere in the tree at this head.

---

openwrt-ai (commented):
Re-reviewed after the force-push (4df9be1 → dd609e0); 3 inline comments, all nits.

All three findings from the previous review are addressed: the missing `\n` on the three `dev_err_probe()` strings, the leaked `of_mdio_find_bus()` reference (now released via `devm_add_action_or_reset()`), and the stale `realtek,otto-serdes-mdio` compatibles (now dropped from all four SoC dtsi). Commit message matches the diff, including the new paragraph about the DTS change. CI is green on the head commit across all six realtek subtargets.

---

openwrt-ai (commented):

Delta since the last review (dd609e0 → f0ac49e) is the `rtpcs_mdio_bus_put()` cleanup plus two commit-message changes, and all three nits from that review are addressed: the devres callback now unpacks `void *data` into a typed `struct mii_bus *` like the other two callbacks in the file, the commit message gained a paragraph documenting the `of_mdio_find_bus()` refcount fix, the quoting around `'realtek,otto-serdes-mdio'` is now balanced, and the PR body mentions both the dropped compatibles and the refcount fix. Commit message still matches the diff. Verified at head that all four SoC dtsi carry `mdio-parent-bus = <&mdio_serdes>;` and that `realtek,otto-serdes-mdio` has no remaining occurrence in the tree. CI is green on the head commit across all six realtek subtargets (the one `cancelled` `build-pr-profile` job is the superseded run from the force-push).

---

plappermaul (approved):
Nice

PR #24720, opened 2026-08-14
kernel: kmod-ltq-tapi: use hrtimers for validation timers

dangowrt:
The TAPI hook state machine validates pulse-dial make and break segments with jiffies timers (HZ*nTime/1000), quantising every validation window to 10 ms ticks at HZ=100. Mechanical rotary dials at 10 pps survive this; fast electronic pulse dialers (around 20 pps, makes of 15-20 ms, found in 1980s button pads retrofitted onto rotary-era phones) race the timer on every pulse, so consecutive pulses merge pairwise and dialled digits become unreliable. No IFX_TAPI_LINE_HOOK_VT_SET configuration can compensate, as the error lies in the timer mechanism rather than in the validation windows.

This replaces the jiffies timer with an hrtimer (CLOCK_MONOTONIC, HRTIMER_MODE_REL_SOFT, keeping expiry in softirq context). The expiry handler still defers the TAPI callback to the workqueue, and periodical timers keep re-arming from the work item after the callback has run, preserving the old defer-but-never-drop tick semantics; the only periodical user is the 5 s power-save poll in drv_tapi_opcontrol.c, which does not need drift-free periods, while all pulse-dial validation timers are one-shot and gain hrtimer precision. TAPI_Delete_Timer() additionally cancels the work item and re-cancels the hrtimer to catch a concurrent periodical re-arm. The new patch supersedes all code hunks of 617-adapt-to-the-6.18-kernel-timer-API.patch, so that patch is folded in, carrying over its <linux/version.h> include.

Run-tested on lantiq/xrx200 (o2 Box 6431, kernel 6.12): an affected fast pulse dialer that previously miscounted every digit now counts reliably. The hrtimer_setup() path for kernel 6.13 and later is build-tested.


package/kernel/lantiq/ltq-tapi/patches/620-use-hrtimers-for-validation-timers.patch
```diff
@@ -0,0 +1,203 @@
...
+ #ifdef LINUX_2_6
+@@ -3519,6 +3503,13 @@
+ #else
+    schedule_task (&(Timer->timerTask));
+ #endif /* LINUX_2_6 */
++
++   if (Timer->bPeriodical)
++   {
++      hrtimer_forward_now(hrt, Timer->period);
++      return HRTIMER_RESTART;
```
openwrt-ai: Periodic timers change from "never lose a tick, only run late" to "silently drop ticks".

Before, the re-arm lived in the work item and ran *after* `pTimerEntry()` had returned, so a new expiry could never be scheduled while a callback was still outstanding — one work execution per period, guaranteed. Now the hrtimer re-arms from softirq every `period` regardless of whether the previous work item has run yet, and `queue_work()`/`schedule_work()` on an already-queued `work_struct` returns `false` and does nothing. On a loaded lantiq SoC with the 15–20 ms windows this patch targets, that expiry's `pTimerEntry()` call is dropped outright rather than deferred.

The commit message presents the deferral as unchanged ("The expiry handler still defers the TAPI callback to the workqueue"), so this looks like an unintended side effect rather than a deliberate trade. `hrtimer_forward_now()` returns the overrun count — could the work item replay it, or could the periodic re-arm stay in the work item so only the one-shot precision win is taken?

---

package/kernel/lantiq/ltq-tapi/patches/620-use-hrtimers-for-validation-timers.patch
```diff
@@ -0,0 +1,203 @@
...
+-#endif
++   hrtimer_cancel(&(Timer->hrt));
+    return (IFX_TRUE);
+ }
+ 
+@@ -3452,6 +3447,7 @@
+    if (Timer)
+    {
+       TAPI_Stop_Timer(Timer);
++      cancel_work_sync(&(Timer->timerTask));
```
openwrt-ai, line 122: Two things about this new line:

1. It is the only unguarded reference to `timerTask` in the file — `timerTask` is a `struct work_struct` only under `#ifdef LINUX_2_6`, otherwise a `struct tq_struct` (see the struct definition in the `drv_tapi_linux.h` hunk above, and the `#ifdef LINUX_2_6` / `schedule_task()` guard you keep at the bottom of this patch). In practice the package always configures `--enable-linux-26`, so it builds, but the surrounding style wraps every such access.
2. `cancel_work_sync()` deadlocks if it is ever reached from inside the work item it is cancelling. Worth confirming no TAPI path calls `TAPI_Delete_Timer()` from a `pTimerEntry()` callback.

(No suggestion block here: adding `#ifdef`/`#endif` lines would also change this hunk's `@@ -3452,6 +3447,7 @@` line counts, so it needs a refresh rather than a hand edit.)

---

package/kernel/lantiq/ltq-tapi/patches/620-use-hrtimers-for-validation-timers.patch
```diff
@@ -0,0 +1,203 @@
...
++static enum hrtimer_restart TAPI_hrtimer_cb (struct hrtimer *hrt);
+ #if (LINUX_VERSION_CODE < KERNEL_VERSION(2,6,20))
+ static IFX_void_t TAPI_tqueue (IFX_void_t *pWork);
+ #else /* for Kernel newer or equal 2.6.20 */
+@@ -3352,7 +3352,14 @@
+    pTimerData->nArgument = nArgument;
+    pTimerData->bStopped = IFX_FALSE;
+ 
+-   timer_setup(&(pTimerData->Timer_List), TAPI_timer_call_back, 0);
++#if LINUX_VERSION_CODE < KERNEL_VERSION(6,13,0)
```
openwrt-ai, line 75: Is `6.13` the right boundary for `hrtimer_setup()`? I could only verify the two endpoints OpenWrt actually builds: `hrtimer_setup()` is absent from `include/linux/hrtimer.h` in v6.12.103 and present in v6.18.44 (where `hrtimer_init()` is gone entirely, so the guard is definitely needed). I could not confirm it already exists in 6.13/6.14.

Since [`target/linux/lantiq/Makefile:12-13`](https://github.com/openwrt/openwrt/blob/8a18af0fa4867369169edad13706e58a1898e2da/target/linux/lantiq/Makefile#L12-L13) pins `KERNEL_PATCHVER:=6.12` / `KERNEL_TESTING_PATCHVER:=6.18`, nothing in tree hits the 6.13–6.17 range, so this is harmless as-is — but if the boundary is actually later, both this `#if` and the commit message's "hrtimer_setup() path for kernel 6.13 and later" would be off.

---

package/kernel/lantiq/ltq-tapi/patches/620-use-hrtimers-for-validation-timers.patch
```diff
@@ -0,0 +1,203 @@
...
+ #ifdef LINUX_2_6
+@@ -3395,21 +3402,14 @@
+           ("\nDRV_ERROR: TAPI_SetTime_Timer (Timer == 0)\n"));
+      return IFX_FALSE;
+    }
+-   Timer->Periodical_Time = HZ*nTime/1000;
+    /* prevent restart of driver */
+    Timer->bPeriodical = IFX_FALSE;
+-   /* remove driver from list */
+-#if LINUX_VERSION_CODE < KERNEL_VERSION(6,16,0)
```
openwrt-ai: nit: this patch reverts every code hunk of [`617-adapt-to-the-6.18-kernel-timer-API.patch`](``https://github.com/openwrt/openwrt/blob/8a18af0fa4867369169edad13706e58a1898e2da/package/kernel/lantiq/ltq-tapi/patches/617-adapt-to-the-6.18-kernel-timer-API.patch)`` — the `del_timer_sync`/`timer_delete_sync` guards in `TAPI_SetTime_Timer()`, `TAPI_Stop_Timer()` and `TAPI_tqueue()`, plus the `from_timer`/`timer_container_of` guard in the callback. After 620 the only surviving hunk of 617 is its `#include <linux/version.h>`, which is still needed here.

Leaving an introduce-then-revert pair in the series is a bit confusing to read. Consider folding: drop 617 and carry the `<linux/version.h>` include in this patch instead, or say in the commit message why 617 is retained.

---

package/kernel/lantiq/ltq-tapi/patches/620-use-hrtimers-for-validation-timers.patch
```diff
@@ -0,0 +1,203 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Daniel Golle <daniel@makrotopia.org>
```
openwrt-ai, line 2: nit: the all-zero `From <sha>` line is a placeholder with no commit behind it; the sibling patches here start straight at `From:` (e.g. [`617-adapt-to-the-6.18-kernel-timer-API.patch:1`](``https://github.com/openwrt/openwrt/blob/8a18af0fa4867369169edad13706e58a1898e2da/package/kernel/lantiq/ltq-tapi/patches/617-adapt-to-the-6.18-kernel-timer-API.patch#L1)).``

```suggestion
From: Daniel Golle <daniel@makrotopia.org>
```

Related, and not fixable by the suggestion above: the `---`/`+++` headers on lines 36-37 and 70-71 carry local mtime timestamps, and none of the hunk headers carry function context (`@@ -3395,21 +3402,14 @@` here vs. `@@ -3398,7 +3399,11 @@ IFX_boolean_t TAPI_SetTime_Timer(Timer_I` in 617) — i.e. this was generated with plain `diff -u` rather than through quilt. `make package/ltq-tapi/refresh` normalizes both.

---

openwrt-ai (commented):
Reviewed 1 new commit (`8a18af0`); comments inline. The commit message matches what the diff does, and the `PKG_RELEASE` bump is there. Hunk offsets and line counts in the new patch all add up, and every kernel symbol it uses (`hrtimer_forward_now`, `ms_to_ktime`, `HRTIMER_MODE_REL_SOFT`, `hrtimer_cancel`, `hrtimer_start`) exists in both v6.12.103 and v6.18.44, with `hrtimer_init()` present only in 6.12 — so the version guard is genuinely needed. Main substantive point is the change in periodic re-arm semantics; the rest are nits.

---

package/kernel/lantiq/ltq-tapi/patches/620-use-hrtimers-for-validation-timers.patch
```diff
@@ -0,0 +1,206 @@
...
+
+Replace the jiffies timer with an hrtimer (CLOCK_MONOTONIC,
+HRTIMER_MODE_REL_SOFT, so expiry runs in softirq context like the old
+timer callback did). The expiry handler defers the actual TAPI callback
+to the workqueue exactly like before, and periodical timers keep
+re-arming from the work item after the callback has run, so a slow
+callback defers, but never drops, a tick. The only periodical user is
+the 5 s power-save poll in drv_tapi_opcontrol.c, which does not need
+drift-free periods; all pulse-dial validation timers are one-shot and
+now expire with hrtimer precision.
```
openwrt-ai, line 31: nit: the PR description still describes the previous revision and now contradicts this commit message on two points:

- Description: *"periodic re-arm moves into the hrtimer callback via `hrtimer_forward_now()`, removing the drift the old code accrued by re-arming only after the work item had run."* This revision does the opposite — the re-arm stays in the work item (`TAPI_tqueue`, further down this patch) and `hrtimer_forward_now()` no longer appears in the patch at all.
- Description: *"The `hrtimer_setup()` path for kernel 6.13 and later is untested; ... Draft until the 6.13+ path has seen a build."* The commit message now says that path is build-tested, which also retires the stated reason for the PR still being a draft.

Worth syncing the description (and un-drafting) so anyone reviewing from the PR page isn't evaluating the design that was already replaced.

---

package/kernel/lantiq/ltq-tapi/patches/620-use-hrtimers-for-validation-timers.patch
```diff
@@ -0,0 +1,206 @@
+From b3cb8745da4fab948200e50c5886ee291e7eeb57 Mon Sep 17 00:00:00 2001
+From: Daniel Golle <daniel@makrotopia.org>
```
openwrt-ai: nit: `b3cb8745da4fab948200e50c5886ee291e7eeb57` resolves to nothing — `git cat-file -t` on it fails against openwrt/openwrt, and `drv_tapi` is imported as a tarball ([`Makefile:15`](https://github.com/openwrt/openwrt/blob/11f0e325bddb42c1ebae1e3c645406228e4040b2/package/kernel/lantiq/ltq-tapi/Makefile#L15)), so there is no git history for a patch-header SHA to point into either. The previous revision's all-zero placeholder at least read as "no commit behind this"; a plausible-looking hex string reads as a genuine upstream reference and will send someone hunting for it.

Every sibling patch in this directory starts straight at `From:` — [616](``https://github.com/openwrt/openwrt/blob/11f0e325bddb42c1ebae1e3c645406228e4040b2/package/kernel/lantiq/ltq-tapi/patches/616-fix-missing-prototypes-warnings.patch#L1),`` [618](``https://github.com/openwrt/openwrt/blob/11f0e325bddb42c1ebae1e3c645406228e4040b2/package/kernel/lantiq/ltq-tapi/patches/618-convert-Makefile-EXTRA_CFLAGS-to-ccflags-y.patch#L1),`` and the [617](``https://github.com/openwrt/openwrt/blob/3cdb8b59d9aa7569b3b9ae6df5f1b05ce93b50d2/package/kernel/lantiq/ltq-tapi/patches/617-adapt-to-the-6.18-kernel-timer-API.patch#L1)`` this patch folds in. Dropping the line matches them:

```suggestion
From: Daniel Golle <daniel@makrotopia.org>
```

---

openwrt-ai (commented):
Reviewed 1 new commit (`11f0e32`); this revision addresses every substantive point from the previous round — the periodic re-arm is back in the work item (no more silently dropped ticks), `cancel_work_sync()` is now `#ifdef LINUX_2_6`-guarded like the rest of the file, 617 is folded in with its `<linux/version.h>` include carried over, and the patch is properly quilt-generated (function context in every hunk header, no local mtimes).

Re-verified against the rebase onto a 617-less tree: all eight `drv_tapi_linux.c` hunk headers and both `drv_tapi_linux.h` ones have correct line counts and cumulative offsets, and they line up exactly with the pre-image line numbers 617 used to target (`del_timer_sync` at 3401 / 3428 / 3481, `from_timer` at 3497). The diffstat (38/23) matches the hunks. `hrtimer_setup(timer, function, clock_id, mode)` in v6.18.44 matches the call's argument order, and the v6.12 `hrtimer_init()` + `timer->function = ...` idiom is the correct pre-6.13 form. `TAPI_Delete_Timer()`'s stop → `cancel_work_sync()` → re-`hrtimer_cancel()` → `kfree()` ordering does close the use-after-free window a work item re-arming a periodical timer could otherwise open. Premise of the fix also holds on this target: `CONFIG_HIGH_RES_TIMERS=y` with `CONFIG_HZ=100`, so the 10 ms quantisation is real and hrtimers do escape it.

Only nits left, inline.

---

package/kernel/lantiq/ltq-tapi/patches/620-use-hrtimers-for-validation-timers.patch
dangowrt: Dropping the `From <sha>` line entirely turned out to trip the formality check: openwrt[bot] requires the full mbox header quartet (`From <hash> Mon Sep 17 00:00:00 2001` / `From:` / `Date:` / `Subject:`) on embedded patches — see `validateEmbeddedPatches()` in openwrt-bot-worker's `cloudflare-worker/src/validators.js`. Restored the line with the all-zero hash, which the check accepts and which honestly reads as "no commit behind this" rather than a fake upstream reference. Verified the patch still applies with `git am` on top of the extracted drv_tapi-3.13.0 tree with all preceding patches applied.

openwrt-ai (commented):

---

PR #24721, opened 2026-08-14
busybox: sysntpd: restart on ifup until first sync

dangowrt:
busybox ntpd retries failed server hostname lookups with a backoff that saturates at 252 seconds, so an instance started before WAN connectivity exists can take minutes to notice that its servers became resolvable. The existing procd interface trigger does deliver the ifup event, but it only performs a reload, which procd treats as a no-op service set unless the computed command line changed; that only happens when the DHCP lease actually delivered option 42 NTP servers. With a lease not carrying option 42, or a statically configured interface providing the default route, the command line stays byte-identical and the running ntpd is left untouched. Devices without an RTC therefore keep a wildly wrong clock for minutes after connectivity arrives, which also stalls protocols requiring monotonic wall time across reboots, such as WireGuard handshakes.

This adds an iface hotplug script restarting sysntpd on ifup while the clock has not been synchronised yet, no matter which interface provides connectivity, and records the first synchronisation through the hotplug.d/ntp mechanism, following the dnsmasq 25-dnsmasqsec precedent. Once synchronised, ntpd recovers on its own and is left untouched on routine interface events.

An unconditional restart trigger was considered and rejected: it would bounce ntpd on every interface update, including DHCP renews and route changes, discarding its clock discipline state each time.

An equivalent hotplug script pair has been in production on RTC-less lantiq/xrx200 CPEs behind WiFi uplinks, and the exact scripts from this PR were verified end-to-end on a 25.12.5 device: ifup while unsynchronised restarts ntpd, the first synchronisation records the marker through the jailed ntpd → ubus (ACL-gated) → procd hotplug.ntp path, and subsequent ifup events leave the synchronised instance untouched.

### DIFF
--- a/package/utils/busybox/files/ntpd.iface-hotplug
+++ b/package/utils/busybox/files/ntpd.iface-hotplug
@@ -1,0 +1,6 @@
+[ "$ACTION" = ifup ] || exit 0
+[ "$INTERFACE" = loopback ] && exit 0
+[ -f /var/state/ntpd-synced ] && exit 0
+/etc/init.d/sysntpd enabled || exit 0
+/etc/init.d/sysntpd running || exit 0
+/etc/init.d/sysntpd restart >/dev/null 2>&1 &
--- a/package/utils/busybox/Makefile
+++ b/package/utils/busybox/Makefile
@@ -6,7 +6,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=busybox
 PKG_VERSION:=1.38.0
-PKG_RELEASE:=2
+PKG_RELEASE:=3
 PKG_FLAGS:=essential
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.bz2
@@ -166,6 +166,9 @@ ifneq ($(CONFIG_BUSYBOX_$(BUSYBOX_SYM)_NTPD),)
 	$(INSTALL_DIR) $(1)/etc/capabilities $(1)/usr/share/acl.d
 	$(INSTALL_DATA) ./files/ntpd.capabilities $(1)/etc/capabilities/ntpd.json
 	$(INSTALL_DATA) ./files/ntpd_acl.json $(1)/usr/share/acl.d/ntpd.json
+	$(INSTALL_DIR) $(1)/etc/hotplug.d/iface $(1)/etc/hotplug.d/ntp
+	$(INSTALL_DATA) ./files/ntpd.iface-hotplug $(1)/etc/hotplug.d/iface/20-ntpd
+	$(INSTALL_DATA) ./files/ntpd.ntp-hotplug $(1)/etc/hotplug.d/ntp/10-ntpd-sync-state
 endif
 	-rm -rf $(1)/lib64
 endef
--- a/package/utils/busybox/files/ntpd.ntp-hotplug
+++ b/package/utils/busybox/files/ntpd.ntp-hotplug
@@ -1,0 +1,5 @@
+case "$ACTION" in
+	step|stratum)
+		touch /var/state/ntpd-synced
+	;;
+esac


package/utils/busybox/files/ntpd.iface-hotplug
```diff
@@ -0,0 +1,13 @@
+# Restart sysntpd when an interface comes up while the clock has not been
+# synchronised yet. busybox ntpd backs off failed server hostname lookups
+# by up to 252 seconds, so an ntpd started before connectivity was
+# available may otherwise take several minutes to notice that its servers
+# became resolvable. Once the clock has been set, ntpd recovers on its
+# own; avoid discarding its discipline state on routine interface events.
+
+[ "$ACTION" = ifup ] || exit 0
```
openwrt-ai, line 1: There is no `$INTERFACE` filter here, so this fires for *every* interface netifd brings up, including `loopback` — netifd queues a hotplug call for all interfaces unconditionally ([`interface_queue_event()` at interface-event.c:122-134](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/interface-event.c#L122-L134), only `IFEV_LINK_UP` is filtered out).

Two consequences that the commit message's "no matter which interface provides connectivity" argument does not seem to cover:

- On a cold boot `loopback`'s `ifup` arrives from netifd (`START=20`) long before sysntpd's own `START=98` slot, so `sysntpd` gets started early by this script and then started again by rc.d. Likewise a plain `/etc/init.d/network restart` bounces ntpd via `loopback`/`lan` even though no uplink state changed.
- netifd runs hotplug handlers strictly serialized ([`task_complete()` → `call_hotplug()` at interface-event.c:105-112](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/interface-event.c#L105-L112)), so the blocking `/etc/init.d/sysntpd restart` on the last line holds up every other queued interface event for the duration of a full init-script stop+start.

Was skipping interfaces that cannot possibly provide connectivity (e.g. an explicit `loopback` exclusion, or requiring a default route / `$DEVICE`) considered and rejected, or is the unfiltered form intentional?

---

openwrt-ai (commented):
Reviewed 1 new commit; one inline question, no commit-message issues found.

Verified against busybox `networking/ntpd.c`: `HOSTNAME_INTERVAL 4` × `DNS_ERRORS_CAP 0x3f` = 252 s, matching the commit message; `run_script("step", ...)` (from `step_time()`) and `run_script("stratum", ...)` (only when `G.stratum` is set to `lastpkt_stratum + 1`) are both emitted after a valid peer reply, so the `step|stratum` filter in `10-ntpd-sync-state` cannot latch on a still-unsynchronised ntpd. `unsync` is a separate action and is correctly not matched. Install paths, `PKG_RELEASE` bump, `INSTALL_DATA` (scripts are sourced, not executed) and the `hotplug.d/ntp` directory creation all match the dnsmasq precedent. CI on the head SHA is green/running with no failures.

---

package/utils/busybox/files/ntpd.iface-hotplug
```diff
@@ -0,0 +1,6 @@
...
+[ "$INTERFACE" = loopback ] && exit 0
+[ -f /var/state/ntpd-synced ] && exit 0
+/etc/init.d/sysntpd enabled || exit 0
+/etc/init.d/sysntpd running || exit 0
+/etc/init.d/sysntpd restart >/dev/null 2>&1 &
```
openwrt-ai, line 6: Detaching the restart removes the serialization that was previously making concurrent restarts impossible, and nothing else here takes its place.

netifd dedups hotplug events *per interface* only ([`interface_queue_event()` at interface-event.c:122-134](https://github.com/openwrt/netifd/blob/e97e36f9b5b545866c40d9803bd39e257389647e/interface-event.c#L122-L134) — "An interface can only have one event in the queue and one event waiting for completion"), so on a dual-stack uplink `wan` and `wan6` each get their own `ifup`. Both pass lines 1-5 (the marker file cannot exist yet, that is the whole point), and because the handler now returns immediately, netifd fires the second one while the first `restart` is still in flight.

`restart` is `stop; start` ([rc.common:23-27](https://github.com/openwrt/openwrt/blob/098d7f5a88106cf3f8ae61b4235f773ceb31b5d2/package/base-files/files/etc/rc.common#L23-L27)) and `procd_lock` is taken only inside [`stop()` at rc.common:156-157](https://github.com/openwrt/openwrt/blob/098d7f5a88106cf3f8ae61b4235f773ceb31b5d2/package/base-files/files/etc/rc.common#L156-L157), never in [`start()` at rc.common:137-143](https://github.com/openwrt/openwrt/blob/098d7f5a88106cf3f8ae61b4235f773ceb31b5d2/package/base-files/files/etc/rc.common#L137-L143). So the lock serializes the two `procd_kill`s against each other but not against the other invocation's `procd_close_service "set"`: invocation B's kill can land after invocation A has already registered and started the instance.

The end state converges (B's `start` re-registers), but the practical result is that an unsynchronised ntpd gets torn down and rebuilt twice in a row exactly when it is trying to make its first query — which is the churn the commit message rejects the unconditional trigger for. Would a single-flight guard around the restart (a lock file, or reusing `procd_lock`'s lockfile) be worth adding, or is the double restart considered acceptable here?

---

package/utils/busybox/files/ntpd.iface-hotplug
```diff
@@ -0,0 +1,6 @@
+[ "$ACTION" = ifup ] || exit 0
+[ "$INTERFACE" = loopback ] && exit 0
```
openwrt-ai, line 2: nit: this guard, plus the `running` guard on line 5 and the detached restart on line 6, were added after the PR body was written, but the body still claims "the exact scripts from this PR were verified end-to-end on a 25.12.5 device" and still describes only the two-condition version ("restarting sysntpd on ifup while the clock has not been synchronised yet"). The commit message has the new third paragraph; the body does not. Worth syncing the body so the verification claim matches what is actually being merged — the boot-ordering behaviour in particular changed (`running` now suppresses the pre-`START=98` restart entirely).

For the guard itself: `loopback` is the right name to match, `config_generate` emits `network.loopback` unconditionally ([config_generate:42-46](https://github.com/openwrt/openwrt/blob/098d7f5a88106cf3f8ae61b4235f773ceb31b5d2/package/base-files/files/bin/config_generate#L42-L46)), so this is correct for stock configs.

---

openwrt-ai (commented):
Re-reviewed after the force-push (`675b0b7d` → `098d7f5a`). The amended commit addresses my earlier question on the unfiltered `$INTERFACE`: loopback is now excluded, the pre-`START=98` case is handled by the `running` guard, and the restart is detached so it no longer stalls netifd's serialized handler chain. The commit message body accurately describes all three additions, so no commit-message issues.

Two follow-ups inline: the detached restart reintroduces a concurrency window that the previous serialized form did not have (line 6), and a body/commit-message sync nit (line 2).

Re-verified for this head: `running` is a valid command for `USE_PROCD=1` init scripts ([rc.common:122](https://github.com/openwrt/openwrt/blob/098d7f5a88106cf3f8ae61b4235f773ceb31b5d2/package/base-files/files/etc/rc.common#L122)); `/var/state` is created by [`/etc/init.d/boot`](https://github.com/openwrt/openwrt/blob/098d7f5a88106cf3f8ae61b4235f773ceb31b5d2/package/base-files/files/etc/init.d/boot#L32) at `START=10`, well before anything here runs, and lives on tmpfs so the marker is correctly cleared each boot; `hotplug-call` sources each script in its own subshell, so the `exit 0` guards cannot short-circuit other `hotplug.d/iface` scripts; and the sync marker is written outside the ujail — `ntpd-hotplug` only makes an ACL-gated `ubus call hotplug.ntp call`, procd runs `hotplug.d/ntp/*` as root. CI on the head SHA is fully green.

---

pwned-pixel:
Just saw this by accident and wonder if it is possible to get this merged. I'm facing same issue. Due to the fact that time is not synced the wireguard interfaces are dead after boot/start and continue to be dead for a long time. Sometimes it recovers after time is synced. Sometimes it does not. So I have to restart the wireguard interfaces itself.

It's not like its urgent. But I could get rid of a long standing workaround script regarding this timesync issue.

I'll test the patch the next few days and report back.

EDIT: I tested this patch for a while. Time is now synced before my "manual" rc.local timesync is called. Despite it fails 2 times.

```
Thu Sep 17 05:29:51 2026 daemon.notice procd: /etc/rc.d/S95done: ntpd: send failed: Network unreachable
Thu Sep 17 05:29:51 2026 daemon.notice procd: /etc/rc.d/S95done: ntpd: send failed: Network unreachable
Thu Sep 17 05:30:01 2026 daemon.notice procd: /etc/rc.d/S95done: Alarm clock
Fri Sep 18 06:33:04 2026 daemon.notice procd: /etc/rc.d/S95done: ntpd: setting time to 2026-09-18 06:33:04.507385 (offset +90175.809472s)
Fri Sep 18 06:33:04 2026 daemon.notice procd: /etc/rc.d/S95done: Time synchronized successfully
Fri Sep 18 06:33:14 2026 daemon.notice procd: /etc/rc.d/S95done: No CPU core saturated for 10s <- rc.local grace period for executing its stuff. After this line sysntpd gets restarted by the script.
Fri Sep 18 06:33:14 2026 daemon.notice procd: /etc/rc.d/S95done: Command failed: ubus call service delete { "name": "sysntpd" } (Not found)
```

I know this are edgecases. But I wondered always why this takes so long. WAN has its own DNS set (in my case DNS on upstream router). I've set my upstream router to provide NTP also now. But I cannot explicit set "42,IP" on the providers box. So I don't know if this has an influence and is picked up by OpenWrt at all.

PR #24724, opened 2026-08-14
ramips: mt7620: reset network blocks on restart

dormancygrace:
## Summary

Reset the MT7620 frame engine, embedded switch and PPE together with the system during restart.

A system-only warm reset can leave PDMA unable to consume its first TX descriptor on the next boot. The state survives driver reprobes and separate FE, ESW or PPE reset attempts. The combined reset is the sequence that recovered the hardware during diagnosis.

MT7628 and MT7688 expose the same FE and ESW reset bits, but this failure and the required reset mask have not been verified on those SoCs, and they do not expose the MT7620 PPE reset block. The workaround is therefore deliberately limited to tested MT7620 hardware.

## Testing

Tested on MT7620N rev 2 eco 6 with repeated ordinary warm reboots. Ethernet returned after every reboot. This patch was also part of the previously green CI history of #24493 before being split out at reviewer request.

Extracted from: #24493

### DIFF
--- a/target/linux/ramips/patches-6.18/335-mt7620-reset-network-blocks-on-restart.patch
+++ b/target/linux/ramips/patches-6.18/335-mt7620-reset-network-blocks-on-restart.patch
@@ -1,0 +1,58 @@
+From: Gleb Pesin <dormancygrace@gmail.com>
+Date: Thu, 30 Jul 2026 16:00:00 +0000
+Subject: [PATCH] MIPS: ralink: reset MT7620 network blocks on restart
+
+The MT7620 system reset does not reliably return the frame engine to a
+usable state when it is asserted on its own.  After a warm restart PDMA can
+accept the first TX producer index update but never consume the descriptor,
+leaving TX_CTX_IDX0 at 1 and TX_DTX_IDX0 at 0.
+
+Assert the FE, ESW and PPE resets together with the system reset.  This
+matches the state produced by a cold reset and prevents the stale network
+blocks from surviving into the next kernel.
+
+This was reproduced on an MT7620N rev 2 eco 6.  Repeated boots with the
+existing system-only reset hit a NETDEV watchdog on the first packet, while
+the combined reset restored both PDMA RX and TX.
+
+MT7628 and MT7688 expose the same FE and ESW reset bits, but the failure
+has not been reproduced or tested on those SoCs and their reset set does
+not include the MT7620 PPE block.  Keep this recovery scoped to MT7620 until
+the same failure and required reset mask are verified there.
+
+Signed-off-by: Gleb Pesin <dormancygrace@gmail.com>
+---
+ arch/mips/ralink/reset.c | 11 ++++++++++-
+ 1 file changed, 10 insertions(+), 1 deletion(-)
+
+--- a/arch/mips/ralink/reset.c
++++ b/arch/mips/ralink/reset.c
+@@ -18,18 +18,27 @@
+ /* Reset Control */
+ #define SYSC_REG_RESET_CTRL	0x034
+ 
++#define RSTCTL_RESET_PPE	BIT(31)
+ #define RSTCTL_RESET_PCI	BIT(26)
++#define RSTCTL_RESET_ESW	BIT(23)
++#define RSTCTL_RESET_FE		BIT(21)
+ #define RSTCTL_RESET_SYSTEM	BIT(0)
+ 
+ static void ralink_restart(char *command)
+ {
++	u32 reset = RSTCTL_RESET_SYSTEM;
++
+ 	if (IS_ENABLED(CONFIG_PCI)) {
+ 		rt_sysc_m32(0, RSTCTL_RESET_PCI, SYSC_REG_RESET_CTRL);
+ 		mdelay(50);
+ 	}
+ 
+ 	local_irq_disable();
+-	rt_sysc_w32(RSTCTL_RESET_SYSTEM, SYSC_REG_RESET_CTRL);
++	if (ralink_soc == MT762X_SOC_MT7620A ||
++	    ralink_soc == MT762X_SOC_MT7620N)
++		reset |= RSTCTL_RESET_PPE | RSTCTL_RESET_ESW | RSTCTL_RESET_FE;
++
++	rt_sysc_w32(reset, SYSC_REG_RESET_CTRL);
+ 	unreachable();
+ }
+ 


target/linux/ramips/patches-6.18/335-mt7620-reset-network-blocks-on-restart.patch
```diff
@@ -0,0 +1,54 @@
...
+ 	if (IS_ENABLED(CONFIG_PCI)) {
+ 		rt_sysc_m32(0, RSTCTL_RESET_PCI, SYSC_REG_RESET_CTRL);
+ 		mdelay(50);
+ 	}
+ 
+ 	local_irq_disable();
+-	rt_sysc_w32(RSTCTL_RESET_SYSTEM, SYSC_REG_RESET_CTRL);
++	if (ralink_soc == MT762X_SOC_MT7620A ||
++	    ralink_soc == MT762X_SOC_MT7620N)
++		reset |= RSTCTL_RESET_PPE | RSTCTL_RESET_ESW | RSTCTL_RESET_FE;
```
openwrt-ai, line 53: nit: MT7628AN/MT7688 use the exact same FE and ESW reset bits as MT7620 — [`resets = <&sysc 21>, <&sysc 23>; reset-names = "fe", "esw";` in mt7628an.dtsi:465-466](https://github.com/openwrt/openwrt/blob/4ad5ccbef6b6a9be39970ad8c6da2dd49be343f0/target/linux/ramips/dts/mt7628an.dtsi#L465-L466), identical to [mt7620a.dtsi:484-485](https://github.com/openwrt/openwrt/blob/4ad5ccbef6b6a9be39970ad8c6da2dd49be343f0/target/linux/ramips/dts/mt7620a.dtsi#L484-L485). Since `ralink_restart()` is shared across all ralink SoCs, an mt76x8 board hitting the same PDMA-stuck-after-warm-reboot condition would not be covered here.

Is restricting this to `MT762X_SOC_MT7620A`/`MT762X_SOC_MT7620N` deliberate (e.g. mt76x8 is known not to need it), or just a consequence of only having MT7620N hardware to test on? A one-line note in the patch body would save the next person from re-diagnosing it. Note `is_mt76x8()` already exists in [`arch/mips/include/asm/mach-ralink/mt7620.h`](https://github.com/gregkh/linux/blob/v6.18.44/arch/mips/include/asm/mach-ralink/mt7620.h#L55-L58) if you do decide to widen the condition.

---

target/linux/ramips/patches-6.18/335-mt7620-reset-network-blocks-on-restart.patch
```diff
@@ -0,0 +1,54 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
openwrt-ai: nit: this is a locally-authored patch, not a real cherry-pick, so the all-zero `git format-patch` header line is a placeholder that carries no information. Locally-authored ramips patches in this directory (e.g. `330-mt7620-avoid-direct-jal-in-random-helper.patch`, `331-...`) start straight at `From:` and omit this line.

```suggestion
From: Gleb Pesin <dormancygrace@gmail.com>
```
(drop the now-duplicated `From:` line below.)

---

openwrt-ai (commented):
Reviewed 1 new commit. The patch applies cleanly against the v6.18 `arch/mips/ralink/reset.c` (hunk header `@@ -18,18 +18,27 @@` and the diffstat both check out), `ralink_soc` / `MT762X_SOC_MT7620A` / `MT762X_SOC_MT7620N` are already reachable via the existing `asm/mach-ralink/ralink_regs.h` include, and the FE (bit 21) / ESW (bit 23) bit assignments match `resets = <&sysc 21>, <&sysc 23>` in `target/linux/ramips/dts/mt7620a.dtsi`. Commit message matches the diff. Two nits inline.

---

target/linux/ramips/patches-6.18/335-mt7620-reset-network-blocks-on-restart.patch
dormancygrace: The restriction is deliberate. The failure and required reset mask were reproduced only on MT7620, and the recovery also includes the MT7620 PPE reset bit. MT7628/MT7688 expose matching FE/ESW bits but have not been verified for this failure. I added that rationale to both the patch description and PR body rather than widening an untested hardware workaround.

target/linux/ramips/patches-6.18/335-mt7620-reset-network-blocks-on-restart.patch
dormancygrace: Removed the all-zero format-patch placeholder. The revised locally authored patch now starts directly with the From header.

dormancygrace:
Addressed the Formality suggestion. The embedded reset patch now has a complete `git format-patch` mail header with a real Linux v6.18.44 commit ID, and its description was wrapped to the kernel style limit. `checkpatch.pl` reports zero errors and zero warnings; the code payload is unchanged.

target/linux/ramips/patches-6.18/335-mt7620-reset-network-blocks-on-restart.patch
```diff
@@ -0,0 +1,59 @@
...
+ 	if (IS_ENABLED(CONFIG_PCI)) {
+ 		rt_sysc_m32(0, RSTCTL_RESET_PCI, SYSC_REG_RESET_CTRL);
+ 		mdelay(50);
+ 	}
+ 
+ 	local_irq_disable();
+-	rt_sysc_w32(RSTCTL_RESET_SYSTEM, SYSC_REG_RESET_CTRL);
++	if (ralink_soc == MT762X_SOC_MT7620A ||
++	    ralink_soc == MT762X_SOC_MT7620N)
++		reset |= RSTCTL_RESET_PPE | RSTCTL_RESET_ESW | RSTCTL_RESET_FE;
```
DragonBluep, line 53: Have you tried assert them at driver `.probe()`/`.remove()`?

target/linux/ramips/patches-6.18/335-mt7620-reset-network-blocks-on-restart.patch
```diff
@@ -0,0 +1,59 @@
+From 88d7abfc45ff00f0a0a33bb25e74587717ec5216 Mon Sep 17 00:00:00 2001
+From: Gleb Pesin <dormancygrace@gmail.com>
```
openwrt-ai: `88d7abfc45ff00f0a0a33bb25e74587717ec5216` does not exist in Linux. `https://github.com/gregkh/linux/commit/88d7abfc45ff00f0a0a33bb25e74587717ec5216` returns 404, as does the same path under `torvalds/linux` (control check: [`1efe5d048a391de3ead2804b2e7f86376c356cc5` resolves to "Linux 6.18.44"](https://github.com/gregkh/linux/commit/1efe5d048a391de3ead2804b2e7f86376c356cc5) on the same host, so this is not a fetch problem).

The [PR comment](https://github.com/openwrt/openwrt/pull/24724#issuecomment-5296840535) describes this as "a real Linux v6.18.44 commit ID", which is not the case. This is a bigger problem than the all-zero placeholder it replaced: an all-zero SHA is visibly a placeholder, whereas a plausible-looking hex string on the `From <sha> Mon Sep 17 ...` line asserts an upstream provenance that does not exist, and `git am` will carry it into the commit-id hint.

This patch is locally authored — the change is absent from [`arch/mips/ralink/reset.c` in v6.18.44](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/ralink/reset.c#L18-L33) — so no SHA belongs on that line at all. The sibling locally-authored patches in this directory (`330-mt7620-avoid-direct-jal-in-random-helper.patch`, `331-...`, `332-...`) all start straight at `From:` and omit the line:

```suggestion
From: Gleb Pesin <dormancygrace@gmail.com>
```

---

target/linux/ramips/patches-6.18/335-mt7620-reset-network-blocks-on-restart.patch
```diff
@@ -0,0 +1,59 @@
...
+blocks from surviving into the next kernel.
+
+This was reproduced on an MT7620N rev 2 eco 6.  Repeated boots with the
+existing system-only reset hit a NETDEV watchdog on the first packet, while
+the combined reset restored both PDMA RX and TX.
+
+MT7628 and MT7688 expose the same FE and ESW reset bits, but the failure
+has not been reproduced or tested on those SoCs and their reset set does
+not include the MT7620 PPE block. Keep this recovery scoped to MT7620 until
+the same failure and required reset mask are verified there.
```
openwrt-ai, line 21: nit: this new paragraph uses a single space after the sentence-ending period ("PPE block. Keep this"), while the three paragraphs above it all use two ("on its own.  After a warm restart", "system reset.  This matches", "eco 6.  Repeated boots"). Worth matching for consistency within the same message.

```suggestion
MT7628 and MT7688 expose the same FE and ESW reset bits, but the failure
has not been reproduced or tested on those SoCs and their reset set does
not include the MT7620 PPE block.  Keep this recovery scoped to MT7620
until the same failure and required reset mask are verified there.
```

---

openwrt-ai (commented):
Reviewed the changes since 4ad5ccb. The code payload is unchanged; the delta is the `From <sha>` header line and a new commit-message paragraph.

The new paragraph resolves the earlier MT7628/MT7688 scoping question well. The header-line change went the wrong direction though — see inline. Commit message still matches the diff, the hunk header (`@@ -18,18 +18,27 @@`) and diffstat (10 insertions, 1 deletion) both still check out against [v6.18.44 `arch/mips/ralink/reset.c`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/ralink/reset.c#L18-L33), and CI is green on `04d4145`.

---

target/linux/ramips/patches-6.18/335-mt7620-reset-network-blocks-on-restart.patch
dormancygrace: Yes. During diagnosis I tried driver reprobes and individual FE, ESW and PPE resets; the PDMA failure survived them. A driver .remove() callback is also not part of the normal reboot path: device_shutdown() calls the bus/driver .shutdown callback, not .remove(). More importantly, recovery required FE+ESW+PPE to be asserted atomically with the system reset. Doing that from one driver probe/remove would cross ownership into the other live blocks and still would not preserve the required combined-reset handoff. That is why this remains in the MT7620 platform restart path.

target/linux/ramips/patches-6.18/335-mt7620-reset-network-blocks-on-restart.patch
dormancygrace: Fixed in 912dd4af2c. The nonexistent patch-envelope SHA has been removed; the locally authored patch now starts at the From: author line, matching the neighboring MT7620 patches.

target/linux/ramips/patches-6.18/335-mt7620-reset-network-blocks-on-restart.patch
dormancygrace: Fixed in 912dd4af2c; the paragraph now uses the same two-space sentence separation as the rest of the patch description.

openwrt-ai (commented):

The delta since `04d4145` is exactly the two points raised last round: the nonexistent `From 88d7abfc... Mon Sep 17 00:00:00 2001` envelope line is gone, and the MT7628/MT7688 paragraph now uses the same two-space sentence separation as the rest of the description. The code payload is unchanged.

Re-verified against head:
- The patch applies cleanly to [`arch/mips/ralink/reset.c` in v6.18.44](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/ralink/reset.c#L18-L33) (`patch -p1 --dry-run`, no fuzz), and the inner diffstat (`1 file changed, 10 insertions(+), 1 deletion(-)`) matches the hunk.
- `ralink_soc`, `MT762X_SOC_MT7620A` and `MT762X_SOC_MT7620N` are all reachable through the existing `asm/mach-ralink/ralink_regs.h` include ([ralink_regs.h:24-30](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/include/asm/mach-ralink/ralink_regs.h#L24-L30)).
- `ramips` only carries `patches-6.18` and `KERNEL_PATCHVER:=6.18`, so there is no second patch series needing the same change.
- Commit message matches the diff.
- CI is green on `912dd4a` (all five ramips subtarget kernel builds plus `Check Kernel patches` succeeded).

One heads-up so it doesn't cause a second round-trip: the `FormalityCheck` bot [now warns](https://github.com/openwrt/openwrt/pull/24724#issuecomment-5301263609) that the patch is "Missing required Git header parameters (`From <hash> Mon Sep 17 00:00:00 2001` ...)". The check itself still reports `success`, and the current form matches how locally-authored patches in this directory are written — `120-1-mips-pci-mt7620-fix-bridge-register-access.patch`, `330-mt7620-avoid-direct-jal-in-random-helper.patch`, `331-...`, `332-...`, `808-...`, `809-...` all start straight at `From:` (21 of the 46 patches in `patches-6.18` omit that line). Re-adding a fabricated SHA to silence the warning would be worse than the warning; leaving it as-is looks right to me, but the maintainer may want to weigh in.

---

PR #24725, opened 2026-08-14
E87n

ZJJCKA:

To help keep the codebase consistent and readable,
and to help people review your contribution,
we ask you to follow the rules you find in the wiki at this link
https://openwrt.org/submitting-patches

Please remove this message before posting the pull request.


target/linux/mediatek/dts/mt7987a-edgepi-e87n.dts
```diff
@@ -0,0 +1,265 @@
...
+	phy3: phy@3 {
+		compatible = "ethernet-phy-ieee802.3-c45";
+		reg = <3>;
+		reset-assert-us = <100000>;
+		reset-deassert-us = <100000>;
+		reset-gpios = <&pio 42 GPIO_ACTIVE_LOW>;
+		interrupt-parent = <&pio>;
+		interrupts = <41 IRQ_TYPE_LEVEL_LOW>;
+		realtek,aldps-enable;
+		realtek,led-link-select = <0xa7 0xa7 0x0>;
```
abajk, line 136: Please use the upstream LED API for the RTL8221B
https://git.kernel.org/pub/scm/linux/kernel/git/netdev/net-next.git/commit/drivers/net/phy/realtek?id=d3aae4d954f92a273388439ab015763e0cdea1e0

abajk (changes requested):

feeds.conf.default
```diff
@@ -1,4 +1,7 @@
-src-git packages https://git.openwrt.org/feed/packages.git
+src-git passwall_packages https://github.com/Openwrt-Passwall/openwrt-passwall-packages.git^b4f719f44343b6f62d5c21e14da0c844b21a58ed
+src-git passwall_luci https://github.com/Openwrt-Passwall/openwrt-passwall.git^ba02ff03af0150ddc1974d47e5dfd2f9c298717d
+src-git openclash https://github.com/vernesong/OpenClash.git^c3a33c1d3407956fdf8f0e0b7c1a4c52e6ad9593
+src-git packages https://github.com/openwrt/packages.git^95875218de25e6340251910a8c996e5a811c69b8
```
openwrt-ai, line 4: This replaces the canonical `packages` feed (`https://git.openwrt.org/feed/packages.git`, floating) with a GitHub mirror pinned to a fixed commit, and adds three unaffiliated third-party feeds. `feeds.conf.default` is global — this change affects every target and every builder, not just E87N, and freezing `packages` at `95875218…` would stop all package updates tree-wide. Local feed changes belong in your own `feeds.conf`, which is already gitignored.

```suggestion
src-git packages https://git.openwrt.org/feed/packages.git
```

---

.config
```diff
@@ -0,0 +1,61 @@
+CONFIG_TARGET_mediatek=y
```
openwrt-ai, line 1: `.config` is the buildroot's generated configuration and is listed in `.gitignore` for exactly this reason — committing it makes every subsequent `make menuconfig`/`make defconfig` in a fresh clone start from an E87N-only config. `.config.e87n` added in the same commit is byte-identical to this file (`diff .config .config.e87n` reports no differences), so it is also a duplicate. Both files, plus `e87n-upstream-state.txt`, `update-openclash-passwall-all.sh` and `update-openclash-passwall-sources.sh` at the repo root, are personal build-harness state and should be dropped from the series.

---

scripts/feeds
```diff
@@ -700,11 +700,44 @@ sub refresh_config {
...
+sub e87n_target_selected {
+	open(my $config, '<', '.config') or return 0;
+	while (my $line = <$config>) {
+		if ($line =~ /^CONFIG_TARGET_mediatek_filogic_DEVICE_edgepi_e87n=y\s*$/) {
+			close($config);
+			return 1;
+		}
+	}
+	close($config);
+	return 0;
```
openwrt-ai, line 712: `scripts/feeds` is a generic, target-agnostic tool; hard-coding a single device symbol (`CONFIG_TARGET_mediatek_filogic_DEVICE_edgepi_e87n`) into it, plus the `e87n_unselected_feed_sources()`/`e87n-prune-unselected-feeds.sh` hooks below, makes the shared tooling behave differently for one board. Note also that `e87n_unselected_feed_sources()` calls `die` if `scripts/e87n-unselected-feed-packages` is unreadable, so a partial checkout aborts `./scripts/feeds install -a` outright.

Package selection is already expressible per-device via `DEVICE_PACKAGES` in `filogic.mk`; the feed installer should not be special-cased.

---

package/base-files/files/etc/rc.button/reset
```diff
@@ -3,12 +3,17 @@
 . /lib/functions.sh
 
 OVERLAY="$( grep ' /overlay ' /proc/mounts )"
+RESET_HOLD_SECONDS=5
+
+# E87N requires an intentional eight-second hold. Keep the established
+# five-second reset threshold unchanged on every other board.
+[ "$(board_name)" = 'edgepi,e87n' ] && RESET_HOLD_SECONDS=8
```
openwrt-ai, line 10: Board-specific behaviour does not belong in `package/base-files`, which ships on every OpenWrt device — this adds a `board_name` call and a compatible-string comparison to the reset handler for all targets. If E87N needs a different hold time, override `/etc/rc.button/reset` from the target's `base-files` (`target/linux/mediatek/filogic/base-files/etc/rc.button/reset`) or make the threshold a UCI/`/etc/config` value rather than a hard-coded `edgepi,e87n` branch here.

---

target/linux/mediatek/filogic/base-files/etc/uci-defaults/99-e87n-defaults
```diff
@@ -0,0 +1,140 @@
+. /lib/functions.sh
+
+case "$(board_name)" in
+	edgepi,e87n)
+		# Requested factory login for fresh images: root / password.
+		printf '%s\n%s\n' 'password' 'password' | passwd root >/dev/null 2>&1
```
openwrt-ai, line 6: Shipping a known default root password is a security regression. OpenWrt deliberately boots with an empty root password so that dropbear refuses password logins until the user sets one; presetting `password` gives every E87N unit built from this tree the same credential, and `uhttpd`/LuCI will accept it over the network from first boot. Drop this and let the user set the password on first login.

```suggestion
```

---

target/linux/mediatek/filogic/base-files/etc/uci-defaults/99-e87n-defaults
```diff
@@ -0,0 +1,140 @@
...
+				-e '\|/passwall_packages\(/packages\.adb\)\?$|d' \
+				"$feed_file"
+		done
+
+		# The pinned OpenClash feed registers its Services menu at order 50.
+		# Put it ahead of Passwall only on E87N; other boards keep feed defaults.
+		openclash_controller='/usr/lib/lua/luci/controller/openclash.lua'
+		if [ -f "$openclash_controller" ]; then
+			sed -i 's/_("OpenClash"), 50)/_("OpenClash"), -5)/' "$openclash_controller"
+			rm -f /tmp/luci-indexcache
```
openwrt-ai, line 136: This `sed -i`-patches a file owned by another package (`luci-app-openclash`) inside the running rootfs. It silently breaks on any upstream string change (the substitution just no-ops), it desynchronises the file from the package manager's checksum so a later `apk`/`opkg` upgrade reverts it, and on a read-only squashfs-without-overlay boot the write fails. Menu ordering is configurable through the LuCI menu JSON (`/usr/share/luci/menu.d/`), which this PR already uses in `package/e87n-luci-overview/files/zz-e87n-menu-overrides.json` — use that mechanism here too instead of rewriting a third-party controller.

---

target/linux/mediatek/dts/mt7987a-edgepi-e87n.dts
```diff
@@ -0,0 +1,269 @@
...
+			color = <LED_COLOR_ID_GREEN>;
+			gpios = <&pio 4 GPIO_ACTIVE_LOW>;
+		};
+
+		led_wan: led-wan {
+			label = "left";
+			function = LED_FUNCTION_WAN;
+			color = <LED_COLOR_ID_GREEN>;
+			gpios = <&pio 3 GPIO_ACTIVE_LOW>;
+		};
```
openwrt-ai, line 77: `label` and `color`/`function` are mutually exclusive: when `label` is present the LED core uses it verbatim as the class-device name and never composes one from color/function — see [`led_compose_name()` in led-core.c:540-551](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/leds/led-core.c#L540-L551). So these LEDs register as `mid` and `left`, and the `function`/`color` properties are dead weight. That is also why `01_leds` has to reference the non-standard names `"mid"`/`"left"` instead of the usual `green:lan`/`green:wan`.

Drop the `label` properties and let color+function name the LEDs; `01_leds` then uses `"green:lan"` / `"green:wan"`.

```suggestion
		led_lan: led-lan {
			function = LED_FUNCTION_LAN;
			color = <LED_COLOR_ID_GREEN>;
			gpios = <&pio 4 GPIO_ACTIVE_LOW>;
		};

		led_wan: led-wan {
			function = LED_FUNCTION_WAN;
			color = <LED_COLOR_ID_GREEN>;
			gpios = <&pio 3 GPIO_ACTIVE_LOW>;
		};
```

---

target/linux/mediatek/filogic/base-files/etc/board.d/02_network
```diff
@@ -290,6 +291,7 @@ mediatek_setup_macs()
 		lan_mac=$(macaddr_add "$wan_mac" 1)
 		label_mac=$wan_mac
 		;;
+	edgepi,e87n|\
 	hiveton,h5000m)
 		lan_mac=$(macaddr_generate_from_mmc_cid mmcblk0)
```
openwrt-ai, line 296: This derives LAN/WAN MACs from the eMMC CID, but `mt7987a-edgepi-e87n.dts` assigns the actual interface MACs from real factory nvmem cells (`macaddr_factory_24` on `&gmac0`, `macaddr_factory_2a` on `&gmac1`). The two sources will not agree, so the label MAC and `board.json` values reported here would not match what the interfaces actually use. `hiveton,h5000m` uses the CID path because it has no factory MAC — E87N appears to have one. Is the CID derivation intentional here, or should E87N read the same `factory` offsets the DTS uses?

---

target/linux/generic/hack-6.18/204-module_strip.patch
```diff
@@ -130,6 +130,15 @@ Signed-off-by: Felix Fietkau <nbd@nbd.name>
...
 +++ b/scripts/mod/modpost.c
+@@ -1611,8 +1611,6 @@ static void read_symbols(const char *mod
+ 			add_namespace(&mod->imported_namespaces, namespace);
+ 		}
+ 
+-		if (!get_modinfo(&info, "description"))
+-			warn("missing MODULE_DESCRIPTION() in %s\n", modname);
+ 	}
+ 
+ 	for (sym = info.symtab_start; sym < info.symtab_stop; sym++) {
```
openwrt-ai, line 141: This removes modpost's `missing MODULE_DESCRIPTION()` warning for the entire tree — every target, every kernel module — in order to silence warnings for one board. That warning is an upstream correctness check; suppressing it globally hides genuinely undescribed modules everywhere. The correct fix for a module that legitimately has a description is to keep `MODULE_DESCRIPTION()`, not to delete the check.

Two further problems with the hunk itself: it leaves a stray blank line before the closing `}`, and it invalidates the diffstat in the patch header (still says `scripts/mod/modpost.c | 12 ++++++++++--`, and the rest of the header was rewritten by hand in `d4e1de4b54f`). If the patch really does need to change, regenerate it with `make target/linux/refresh` so the metadata and hunk offsets are consistent.

---

package/kernel/gpio-button-hotplug/src/gpio-button-hotplug.c
```diff
@@ -722,8 +722,21 @@ static void __exit gpio_button_exit(void)
...
+		_tag "=" _value
+
+GPIO_BUTTON_MODINFO(gpio_button_author_gabor, "author",
+			"Gabor Juhos <juhosg@openwrt.org>");
+GPIO_BUTTON_MODINFO(gpio_button_author_felix, "author",
+			"Felix Fietkau <nbd@nbd.name>");
+GPIO_BUTTON_MODINFO(gpio_button_description, "description",
+			"Polled GPIO Buttons hotplug driver");
+GPIO_BUTTON_MODINFO(gpio_button_alias, "alias", "platform:" DRV_NAME);
 MODULE_LICENSE("GPL v2");
```
openwrt-ai, line 742: Hand-rolling `.modinfo` records duplicates what `MODULE_AUTHOR`/`MODULE_DESCRIPTION`/`MODULE_ALIAS` already do and loses their guarantees: `MODULE_ALIAS` is what `depmod` consumes for `modules.alias` autoloading, and the `__section(".modinfo")` strings here are not `MODULE_INFO`-tagged, so the built-in (`CONFIG_..=y`) case and the module-namespacing logic no longer behave the same. It also silently diverges from every other module in the tree.

The stated reason — that `204-module_strip.patch` elides these wrappers when `CONFIG_MODULE_STRIPPED` is set — is the intended behaviour of that hack, and the same commit series also deletes the modpost warning that motivated this. One of the two changes is redundant; neither is needed if the module keeps the standard macros.

---

package/fancontrol/src/fancontrol.c
```diff
@@ -0,0 +1,591 @@
...
+    fp = fopen(path ,"r");
+    if (fp == NULL)
+        return -1;
+
+    if (( read = getline(&line ,&len ,fp) ) != -1) {
+        if (size != 0)
+            memcpy(result ,line ,size);
+        else
+            memcpy(result ,line ,read - 1);
+    }
```
openwrt-ai, line 97: Both `memcpy()` branches write an unbounded amount into the caller's buffer, and neither NUL-terminates.

- `size != 0`: copies exactly `size` bytes from `line` even when `getline()` returned fewer — heap over-read past the end of the glibc/musl-allocated `line`.
- `size == 0`: copies `read - 1` bytes with no knowledge of the destination size. Callers pass small stack buffers — `get_temperature()` and `get_fanspeed()` both use `char buf[8]` and `get_device_model()` uses `char buf[64]` — while the paths are user-controlled via `-T`/`-F` and the `fancontrol.settings.*` UCI values, so a longer first line is a stack smash.

Also, the function returns 0 when `getline()` fails, so callers cannot tell "read nothing" from "read a value"; `get_temperature()` then `atoi()`s a zeroed buffer and reports 0 °C, which drives the fan to a stop.

Pass the destination size unconditionally and clamp:

```suggestion
static int read_file(const char* path ,char* result ,size_t size) {
    FILE* fp;
    char* line = NULL;
    size_t len = 0;
    ssize_t read;
    int ret = -1;

    if (size == 0)
        return -1;

    fp = fopen(path ,"r");
    if (fp == NULL)
        return -1;

    if (( read = getline(&line ,&len ,fp) ) != -1) {
        size_t copy;

        while (read > 0 && ( line[read - 1] == '\n' || line[read - 1] == '\r' ))
            read--;

        copy = (size_t)read;
        if (copy > size - 1)
            copy = size - 1;

        memcpy(result ,line ,copy);
        result[copy] = '\0';
        ret = 0;
    }
```
(Callers currently passing `0` must be updated to pass `sizeof(buf)`.)

---

package/luci-app-display/Makefile
```diff
@@ -0,0 +1,26 @@
...
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=luci-app-display
+PKG_VERSION:=1.0.0
+PKG_RELEASE:=10
+
+LUCI_TITLE:=LuCI support for display control
+LUCI_DESCRIPTION:=Configure display backlight brightness and screen layout from LuCI
+LUCI_DEPENDS:=+luci-base +kmod-backlight-pwm +kmod-fb-tft-nv3007 +libfreetype +libstdcpp +libgcc
+LUCI_PKGARCH:=$(ARCH_PACKAGES)
```
openwrt-ai, line 12: This package ships `root/usr/sbin/display` as a committed 2.1 MB pre-built, unstripped aarch64 ELF (`ELF 64-bit LSB executable, ARM aarch64, dynamically linked, interpreter /lib/ld-musl-aarch64.so.1, with debug_info, not stripped`) with no corresponding source in the tree. That is not buildable or auditable, it defeats `PKG_ARCH`/toolchain matching (`LUCI_PKGARCH:=$(ARCH_PACKAGES)` here claims the package is arch-generic while the binary is aarch64-only), and it bloats the git history permanently. Commit `0f2b1d3a72e` is titled "build cyberpunk display from source" but the binary is still present at HEAD.

The same applies to `package/luci-app-fanctrol/root/usr/lib/lua/luci/i18n/fancontrol-bundled.zh-cn.lmo` — `.lmo` files are compiled from `.po` at build time by `luci.mk`, they should not be committed. Together with the five `display/screen*.png` screenshots (~5 MB) this series adds roughly 12 MB of binaries.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -1726,6 +1726,32 @@ define Device/edgecore_eap111
...
+	luci-app-argon-config luci-app-display \
+	luci-app-firewall luci-app-openclash luci-app-package-manager \
+	hysteria luci-app-passwall luci-app-samba4 luci-app-ttyd luci-app-upnp luci-base \
+	luci-i18n-argon-config-zh-cn luci-i18n-base-zh-cn \
+	luci-i18n-firewall-zh-cn luci-i18n-nlbwmon-zh-cn \
+	luci-i18n-package-manager-zh-cn luci-i18n-passwall-zh-cn luci-i18n-samba4-zh-cn \
+	luci-i18n-ttyd-zh-cn luci-i18n-upnp-zh-cn \
+	luci-mod-admin-full luci-proto-ipv6 luci-proto-ppp luci-theme-argon mkf2fs \
+	kmod-fs-ext4 \
+	mhz miniupnpd-nftables mt7987-2p5g-phy-firmware openclash-core-meta pciutils \
```
openwrt-ai: `DEVICE_PACKAGES` should list the packages a board needs to *function* (kmods for its flash/PHY/USB, firmware, `mkf2fs`/`f2fsck`, etc.), not a curated end-user distribution. Several entries here cannot build against the official feeds at all — `luci-app-openclash`, `openclash-core-meta`, `luci-app-passwall`, `hysteria` come only from the third-party feeds this PR bolts onto `feeds.conf.default`, so any builder using the stock feed set gets an unresolvable device. Others are policy choices that upstream leaves to the user: `-dnsmasq`/`dnsmasq-full`, `-wpad-basic-mbedtls`, `samba4-server`, `ttyd`, `btop`, the `luci-theme-argon` swap, and the whole `luci-i18n-*-zh-cn` set.

Trim this to the hardware-required list (`kmod-backlight-pwm kmod-fb-tft-nv3007 kmod-hwmon-pwmfan kmod-nvme kmod-usb3 kmod-fs-ext4 mt7987-2p5g-phy-firmware mkf2fs f2fsck` and similar) and leave the rest to a local `.config`. Note `luci-base`, `luci-mod-admin-full` and `uhttpd` are already pulled in by the default LuCI selection and don't need listing.

---

target/linux/mediatek/dts/mt7987.dtsi
```diff
@@ -1087,6 +1087,7 @@
 				     <GIC_SPI 197 IRQ_TYPE_LEVEL_HIGH>,
 				     <GIC_SPI 198 IRQ_TYPE_LEVEL_HIGH>,
 				     <GIC_SPI 199 IRQ_TYPE_LEVEL_HIGH>;
+			interrupt-names = "fe0", "fe1", "fe2", "fe3";
```
openwrt-ai, line 1090: nit: this touches the shared MT7987 SoC `.dtsi` — it affects every MT7987 board, not just E87N — yet it arrives inside `a34b4fc51c7 "mediatek: e87n: finalize defaults and warning fixes"` with no explanation of why `interrupt-names` is needed or which driver consumes it. The same commit also modifies `mt7981b-netis-n6-v2.dts`, `mt7981b-netis-nx32u.dts` and six `mt7987a-rfb-*.dtso` overlays, none of which are E87N. Please split the SoC/RFB/other-board DT fixes into their own commits with their own rationale so they can be reviewed and backported independently.

---

package/kernel/linux/modules/video.mk
```diff
@@ -34,6 +34,20 @@ $(eval $(call KernelPackage,media-controller))
...
+# expressing both as conditional Kconfig selects creates a recursive select.
+BACKLIGHT_DEPENDS:=@DISPLAY_SUPPORT
+FB_DEPENDS:=@DISPLAY_SUPPORT
+
+ifeq ($(KERNEL_PATCHVER),6.12)
+  BACKLIGHT_DEPENDS += +kmod-fb
+endif
+
+ifeq ($(KERNEL_PATCHVER),6.18)
+  FB_DEPENDS += +kmod-backlight
```
openwrt-ai, line 48: Keying the dependency direction off `KERNEL_PATCHVER` string-matches only the two series that happen to exist today: any target on a different `KERNEL_PATCHVER` (or the next series bump) falls through both `ifeq`s and gets `kmod-backlight` and `kmod-fb` with no relationship at all, which silently reintroduces the missing-dependency the change is meant to fix rather than failing loudly. A `$(if $(CONFIG_LINUX_6_12),…)` style test, or an `else` branch that errors on an unhandled version, would degrade more safely.

The comment claims a recursive select, but the actual upstream relationship is one-directional per series — worth stating in the commit message which config triggered the recursion, since this hunk changes `kmod-fb`/`kmod-backlight` for every target in the tree.

---

target/linux/mediatek/patches-6.18/999990-add-nv3007-fbtft.patch
```diff
@@ -0,0 +1,299 @@
+Index: linux-6.18/drivers/staging/fbtft/Kconfig
+===================================================================
+--- linux-6.18.orig/drivers/staging/fbtft/Kconfig
++++ linux-6.18/drivers/staging/fbtft/Kconfig
```
openwrt-ai, line 4: Two convention problems with this patch file:

1. It uses `Index:` / `linux-6.18/` style headers instead of the `--- a/… +++ b/…` form every other patch in `target/linux/mediatek/patches-6.18/` uses, and it carries no `From:`/`Subject:`/`Signed-off-by:` header at all. Regenerating it with `make target/linux/refresh` (E87N selected in `.config`) produces the correct quilt-style output.
2. `999990-` is outside the numbering scheme used in this directory (existing patches run 7xx–9xx, highest is `966-`). Pick the next free number in the appropriate range, e.g. `967-`.

Separately: `drivers/staging/fbtft` is deprecated upstream and slated for removal; a new panel added there will need re-porting to a DRM `tiny` driver on the next kernel bump. Worth mentioning in the commit message that this is knowingly a staging-fbtft driver.

---

target/linux/mediatek/filogic/base-files/etc/board.d/01_leds
```diff
@@ -130,6 +130,10 @@ glinet,gl-xe3000)
 	ucidef_set_led_netdev "wlan2g" "WLAN2G" "green:wifi2g" "phy0-ap0"
 	ucidef_set_led_netdev "wlan5g" "WLAN5G" "green:wifi5g" "phy1-ap0"
 	;;
+edgepi,e87n)
```
openwrt-ai, line 133: nit: this case arm is inserted after `glinet,gl-xe3000)`, breaking the alphabetical ordering the rest of the file follows — it belongs before `edgecore,eap111`/after it but ahead of the `glinet,*` block. The same misplacement occurs in `lib/upgrade/platform.sh` at lines 208, 442 and 490, where `edgepi,e87n|\` is added after `glinet,gl-xe3000|\` / `globitel,bt-r320|\`. `02_network` and `image/filogic.mk` got the ordering right, so this is just an oversight.

---

.gitattributes
```diff
@@ -2,3 +2,8 @@
 *.patch whitespace=-indent-with-non-tab,-space-before-tab,-tab-in-indent,-trailing-space
 *.dts text eol=lf
 *.dts[io] text eol=lf
+files/usr/share/passwall/app.sh text eol=lf whitespace=-trailing-space
+files/etc/init.d/passwall text eol=lf
+files/etc/hotplug.d/iface/98-passwall text eol=lf
+/update-openclash-passwall-*.sh text eol=lf
+/e87n-upstream-state.txt text eol=lf
```
openwrt-ai, line 9: These rules exist to support files that should not be in the repository in the first place. `files/` is the buildroot's *user* custom-files overlay — anything placed there is copied verbatim into the rootfs of whatever the local builder happens to be building, so committing `files/usr/share/passwall/app.sh` (2103 lines), `files/etc/init.d/passwall` and `files/etc/hotplug.d/iface/98-passwall` injects Passwall into every image built from this tree, for every target. Those files are also a vendored copy of a third-party project rather than an OpenWrt package.

`/update-openclash-passwall-*.sh` and `/e87n-upstream-state.txt` are likewise personal update-harness artifacts at the repo root. Drop all of it, and this `.gitattributes` addition along with it.

```suggestion
```

---

package/luci-app-fanctrol/Makefile
```diff
@@ -0,0 +1,178 @@
+include $(TOPDIR)/rules.mk
```
openwrt-ai, line 1: This series adds three overlapping fan-control packages that all ship a `/etc/config/fancontrol` and a fan daemon: `package/fancontrol` (C daemon + init), `package/luci-app-fancontrol` (LuCI, `luasrc/controller/fancontrol.lua`) and `package/luci-app-fanctrol` (LuCI + its *own* `root/usr/bin/fancontrol` shell daemon + `root/etc/init.d/fancontrol`). `luci-app-fancontrol` and `luci-app-fanctrol` are near-duplicates of each other down to the CSS, and only `luci-app-fanctrol` is referenced by `DEVICE_PACKAGES`. Two of the three should be removed rather than merged into the tree.

Also, `package/luci-app-fanctrol/tools/` ships `build_ipk.py` (535 lines) plus four `test_*.py` files that are development scaffolding for an out-of-tree build — those don't belong in an OpenWrt package directory, where packaging is `luci.mk`'s job.

---

openwrt-ai (commented):
## Commit checks

- **All 44 commits** — none carries a `Signed-off-by:` trailer, and every one has a subject line with an empty body. OpenWrt requires a DCO sign-off on every commit (see https://openwrt.org/submitting-patches). The PR description is also still the unmodified template text, which it explicitly asks you to remove.
- `0c9870f56e3` **"mediatek: add EdgePi E87N support"** — this is the new-device commit (adds `mt7987a-edgepi-e87n.dts` + `define Device/edgepi_e87n`), so its message must document the hardware specification (SoC/RAM/flash/WiFi/Ethernet/LEDs/buttons/UART/power), flash instructions (at least one fully-described install path plus a revert-to-stock note), and the MAC address layout (one line per interface, with the source offset). It currently has no body at all. The subject also doesn't match the diff: this commit rewrites `feeds.conf.default`, which is not `mediatek:` scope.
- `a34b4fc51c7` **"mediatek: e87n: finalize defaults and warning fixes"** — subject claims E87N scope but the diff changes the shared `mt7987.dtsi`, two unrelated MT7981 boards (`mt7981b-netis-n6-v2.dts`, `mt7981b-netis-nx32u.dts`), six `mt7987a-rfb-*.dtso` overlays and `mt7987a-tenda-be12-pro.dts`/`mt7987a-hiveton-h5000m.dts`. Please split those into separate commits.
- `681bfcae285` **"mediatek: e87n: complete proxy cores and reset handling"** — the `mediatek:` prefix is wrong for a change to `package/base-files/files/etc/rc.button/reset`, which is generic base-files affecting every target.
- `a903aae1ad3` **"Add direction-safe PWM fan package"** — subject has no `<area>: ` prefix; the project convention would be something like `fancontrol: add direction-safe PWM fan package`.
- `d4e1de4b54f` **"mediatek: fix RTL8221 LED patch metadata"** and `6bb3919a539` **"fancontrol: fix GCC 14 build errors"** are introduce-then-fix follow-ups to `9d040905ca3` and `a903aae1ad3` respectively; likewise `b9248026e6f` fixes `43b475728da`, `96b6e42df2d` fixes `0f2b1d3a72e`, `085d1057cac` fixes `bb4db3e31fe`, and the feed-pruning work spans six commits (`3b6c25a88e8`, `4e021597e71`, `0d37af0a2d1`, `d45f8bfca6e`, `a2ea3da20b5`, `ad2ada9532d`, `c61eaa68bae`, `da9b646f026`). Please squash each fix into the commit that introduced the problem.
- `57754e73510` **"kernel: remove stripped module description warning"** and `25c82e89b98` **"mediatek: silence stripped module description warnings"** solve the same problem twice, in contradictory ways (see the inline notes on `204-module_strip.patch` and `gpio-button-hotplug.c`). One of them should go.

---

Beyond the per-commit points, the series as a whole reads as a full downstream firmware distribution rather than a device-support submission: it commits the buildroot `.config`, rewrites `feeds.conf.default` to pin the official `packages` feed and add three third-party feeds, adds a `files/` rootfs overlay, patches `scripts/feeds` with device-specific logic, modifies generic `package/base-files`, and ships ~12 MB of pre-built binaries (including a 2.1 MB unstripped aarch64 ELF with no source). Extracting the genuine E87N hardware enablement — the DTS, the `Device/edgepi_e87n` block, the `board.d`/`platform.sh` entries and the NV3007 fbtft patch — into a small standalone series would make this reviewable; the inline comments mark what I'd expect to be dropped.

Note @abajk has already pointed out that the RTL8221B LED configuration should use the upstream LED API rather than the custom `realtek,led-link-select` property, which would remove `790-net-phy-realtek-add-led-link-select-for-RTL8221.patch` entirely — I haven't duplicated that comment.

---

.github/workflows/e87n-build.yml
```diff
@@ -0,0 +1,653 @@
...
+          done
+          find "$work/rootfs/lib/modules" -name 'fb_nv3007.ko' -print -quit | grep -q .
+          find "$work/rootfs/lib/modules" -name 'hwmon.ko' -print -quit | grep -q .
+          find "$work/rootfs/lib/modules" -name 'nvme-core.ko' -print -quit | grep -q .
+          find "$work/rootfs/lib/modules" -name 'nf_flow_table.ko' -print -quit | grep -q .
+
+          file "$work/rootfs/usr/sbin/display" | grep -q 'ARM aarch64'
+          echo "4edc1c9be957d5d80ea7e90b6807e273c23ca3c860407a5353279d566ea33e15  $work/rootfs/usr/sbin/display" | sha256sum -c -
+          for screen in "$work/rootfs/www/luci-static/resources/display/"screen{0,1,2,3,4}.png; do
+            file "$screen" | grep -Fq 'PNG image data, 428 x 142'
```
openwrt-ai, line 621: This assertion can never pass at HEAD, so the workflow is self-contradictory. `f9ed6c7` downscaled the five `screen*.png` to 428x142 and added this rootfs check; `8bd0563` then restored the original images (`Bin 1282 -> 214546` etc.) and updated only the *source-tree* copy of the check at line 201 to `2006 x 784`, leaving this one behind. The packaged files are the same bytes — `luci-app-display/Makefile` has no image-processing step and `luci.mk` copies `root/` verbatim into the rootfs — so `$work/rootfs/www/luci-static/resources/display/screen0.png` is still 2006x784. Verified against the blobs at HEAD:

```
screen0.png: PNG image data, 2006 x 784, 8-bit/color RGB, non-interlaced
... (screen1..4 identical dimensions)
```

Either bring the assertion in line with the images actually shipped, or re-apply the downscale to the source PNGs (which is what the new `aspect-ratio: 428 / 142` CSS expects — see the comment on `index.js`).

```suggestion
            file "$screen" | grep -Fq 'PNG image data, 2006 x 784'
```

---

package/luci-app-display/htdocs/luci-static/resources/view/display/index.js
```diff
@@ -0,0 +1,2100 @@
...
+				height: 10px;
+				border: solid #fff;
+				border-width: 0 2px 2px 0;
+				transform: rotate(45deg);
+			}
+
+			.display-screen-image {
+				display: block;
+				width: 100%;
+				aspect-ratio: 428 / 142;
```
openwrt-ai, line 700: The static screen thumbnails are now geometrically distorted. This box is forced to 428/142 (≈3.01), the `screen*.png` files restored by `8bd0563` are 2006x784 (≈2.56), and the image inside is styled `object-fit: fill` (changed from `cover` in the same series, line 572). Filling a 3.01 box with a 2.56 source stretches every thumbnail ~18% horizontally, which is visible on text-heavy mock-ups like these.

`f9ed6c7` made this consistent by downscaling the PNGs to 428x142; `8bd0563` reverted the PNGs but kept the CSS. Either restore the 428x142 assets (and fix the rootfs CI assertion, see the comment on `e87n-build.yml:621`) or use `object-fit: contain` so the un-resized mock-ups are letterboxed rather than stretched.

---

target/linux/mediatek/filogic/base-files/etc/uci-defaults/99-e87n-defaults
```diff
@@ -0,0 +1,162 @@
...
+				uci -q set "samba4.$samba_share.name=share"
+				uci -q set "samba4.$samba_share.path=/mnt/nvme0n1p1"
+				uci -q set "samba4.$samba_share.browseable=yes"
+				uci -q set "samba4.$samba_share.read_only=no"
+				uci -q set "samba4.$samba_share.force_root=1"
+				uci -q set "samba4.$samba_share.guest_ok=yes"
+				uci -q set "samba4.$samba_share.guest_only=no"
+				uci -q set "samba4.$samba_share.inherit_owner=no"
+				uci -q set "samba4.$samba_share.create_mask=0666"
+				uci -q set "samba4.$samba_share.dir_mask=0777"
```
openwrt-ai, line 81: This preseeds an unauthenticated, world-writable, root-privileged SMB export on every fresh E87N image. The combination is the problem: `guest_ok=yes` accepts anonymous SMB sessions, `read_only=no` makes them writable, `force_root=1` runs every operation as uid 0, and `create_mask=0666`/`dir_mask=0777` strips any residual protection. Any host on the LAN — including an untrusted Wi-Fi client — can read, overwrite and delete the entire NVMe contents with no credential, and files land owned by root.

That is a much stronger default than OpenWrt ships anywhere else, and it is not implied by the commit subject ("preseed NVMe share"). If a first-boot share is wanted at all, preseed it as authenticated (`guest_ok=no`, drop `force_root`) and leave the guest/root-squash decision to the user in LuCI. Note this compounds the preset `root:password` credential already flagged at line 6 — together the device ships network-reachable with both a known login and an open root-writable share.

---

package/luci-app-display/root/usr/sbin/display-control
```diff
@@ -0,0 +1,558 @@
...
+	fb_height="$2"
+	fb_bpp="$3"
+	fb_stride="$4"
+	frame_bytes=$(( fb_stride * fb_height ))
+
+	[ -r /dev/fb0 ] || json_error "Framebuffer is not readable" 3
+	dd if=/dev/fb0 of="$snapshot_tmp" bs="$frame_bytes" count=1 2>/dev/null || {
+		rm -f "$snapshot_tmp"
+		json_error "Failed to capture framebuffer" 4
+	}
```
openwrt-ai, line 442: `bs=$frame_bytes count=1` asks `dd` for the whole frame (~121 KB) in a single `read(2)`. A short read is not an error for `dd` — it writes what it got and exits 0 — so the failure surfaces four lines down as the hard `"Incomplete framebuffer snapshot"` error, and the whole live preview falls back to the static image with no retry. Reading in stride-sized records makes the transfer robust and is the natural unit here, since `fb_stride`/`fb_height` are already computed:

```suggestion
	dd if=/dev/fb0 of="$snapshot_tmp" bs="$fb_stride" count="$fb_height" 2>/dev/null || {
		rm -f "$snapshot_tmp"
		json_error "Failed to capture framebuffer" 4
	}
```

Two smaller points on this function: `snapshot_tmp="${SNAPSHOT_PATH}.$$"` is a predictable name in world-writable `/tmp`, and `dd of=` follows symlinks — a pre-created `/tmp/display-framebuffer.raw.<pid>` symlink turns this root-run helper into an arbitrary-file overwrite. `mktemp` avoids that. And `chmod 0644` below publishes the panel contents (network status, addresses) to every local user; `0600` is enough, since the rpcd ACL read happens as root.

---

package/luci-app-display/root/usr/sbin/display-control
```diff
@@ -0,0 +1,558 @@
...
+	set -- $(fb_geometry)
+	fb_width="$1"
+	fb_height="$2"
+	fb_bpp="$3"
+	fb_stride="$4"
+	frame_bytes=$(( fb_stride * fb_height ))
+
+	[ -r /dev/fb0 ] || exit 3
+	dd if=/dev/fb0 bs="$frame_bytes" count=1 2>/dev/null
+}
```
openwrt-ai, line 425: `frame` is dead code at HEAD. `59760db` added it for `fs.exec_direct(HELPER, [ 'frame' ], 'blob')`, then `a3a1a2e` switched the UI to `frame-snapshot` + `fs.read_direct`; nothing calls `frame` any more (the only remaining references are the `case` arm at line 492 and the usage string). It is also the one subcommand that breaks the helper's contract — every other path prints JSON, this one writes raw binary to stdout and uses a bare `exit 3` instead of `json_error`.

Please drop the function, the `frame)` case arm and the `frame|` token in the usage string. Note that `e87n-build.yml:187` currently asserts `grep -Fq 'dd if=/dev/fb0 bs="$frame_bytes" count=1'`, which matches only this dead function — that assertion needs to go with it.

---

package/luci-app-display/htdocs/luci-static/resources/view/display/index.js
```diff
@@ -0,0 +1,2100 @@
+'use strict';
+'require dom';
+'require fs';
+'require uci';
+'require ui';
+'require view';
+
+var HELPER = '/usr/sbin/display-control';
+var FRAME_SNAPSHOT = '/tmp/display-framebuffer.raw';
+var FRAME_REFRESH_MS = 100;
```
openwrt-ai, line 10: A 100 ms poll is very expensive for what it does on this class of hardware. Each tick performs two round trips: `runHelper(['frame-snapshot'])` forks a shell that reads four sysfs files, runs `dd`, `wc`, `chmod` and `mv`, then `fs.read_direct()` spawns `cgi-download` to ship ~121 KB (428 × 142 × 2 bytes, base64-inflated over HTTP). That is roughly 20 process spawns and well over 1 MB/s of HTTP traffic per second, sustained for as long as the page is open, on a router whose CPU is also meant to be forwarding packets — and `framebuffersDiffer()` additionally walks all 121 KB byte-by-byte in JS on every comparison.

The panel is a status display; 2-5 fps would look equally "live". Consider raising this to ~500 ms for the idle poll and keeping the fast cadence only inside `waitForFramebufferChange()`, where the short latency actually matters.

---

package/luci-app-display/htdocs/luci-static/resources/view/display/index.js
```diff
@@ -0,0 +1,2100 @@
...
+	},
+
+	renderPreviewVisual: function() {
+		var canvas;
+
+		if (this.state && this.state.enabled && this.frameBuffer) {
+			canvas = E('canvas', {
+				class: 'display-live-frame',
+				width: '428',
+				height: '142',
```
openwrt-ai, line 1176: The canvas is pinned to 428x142 even though `readState()` now stores the real geometry in `this.frameInfo` from the helper's `fb_width`/`fb_height`. `paintFramebuffer()` decodes at `info.width`×`info.height` and then `drawImage(source, 0, 0, canvas.width, canvas.height)`, so any panel whose framebuffer is not 428x142 gets resampled into the wrong-shaped bitmap. Use the value you just read instead of the literal:

```suggestion
				width: String((this.frameInfo && this.frameInfo.width) || 428),
				height: String((this.frameInfo && this.frameInfo.height) || 142),
```

---

package/luci-app-display/root/usr/sbin/display-control
```diff
@@ -0,0 +1,558 @@
...
+	if [ -s "$pidfile" ]; then
+		pid="$(cat "$pidfile" 2>/dev/null)"
+		[ -n "$pid" ] && kill "$pid" >/dev/null 2>&1 || true
+		rm -f "$pidfile"
+	fi
+
+	for pid in $(ps w | awk '/\/usr\/sbin\/display-control daemon/ && !/awk/ { print $1 }'); do
+		[ -n "$pid" ] && kill "$pid" >/dev/null 2>&1 || true
+	done
+}
```
openwrt-ai, line 281: nit: this is a near-verbatim copy of `kill_daemons()` plus the pidfile handling from `stop()` in `root/etc/init.d/display-control` — same `ps w | awk '/\/usr\/sbin\/display-control daemon/ && !/awk/'` pattern, same `/var/run/display-control.pid`. Two copies of the same PID discovery will drift.

It is also asymmetric with the enable path added in the same commit, which correctly delegates to `/etc/init.d/display-control start` (line 296). The disable branch at line 300 can just call `/etc/init.d/display-control stop`, which already does the pidfile kill, `kill_daemons()` with its `kill -9` second pass, `$HELPER off` and `killall display` — i.e. the whole `stop_control_daemon` + `turn_off` + `killall` sequence below it.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -1726,6 +1726,32 @@ define Device/edgecore_eap111
...
+	luci-app-argon-config luci-app-display \
+	luci-app-firewall luci-app-openclash luci-app-package-manager \
+	hysteria luci-app-passwall luci-app-samba4 luci-app-ttyd luci-app-upnp luci-base \
+	luci-i18n-argon-config-zh-cn luci-i18n-base-zh-cn \
+	luci-i18n-firewall-zh-cn \
+	luci-i18n-package-manager-zh-cn luci-i18n-passwall-zh-cn luci-i18n-samba4-zh-cn \
+	luci-i18n-ttyd-zh-cn luci-i18n-upnp-zh-cn \
+	luci-mod-admin-full luci-proto-ipv6 luci-proto-ppp luci-theme-argon lsblk mkf2fs \
+	mhz miniupnpd-nftables mt7987-2p5g-phy-firmware openclash-core-meta pciutils \
+	rpcd-mod-rrdns samba4-server ttyd uhttpd uhttpd-mod-ubus usbutils nvme-cli
```
openwrt-ai, line 1749: nit: `nvme-cli` is appended after `usbutils`, breaking the alphabetical ordering the rest of this `DEVICE_PACKAGES` list follows — it belongs between `mt7987-2p5g-phy-firmware` and `openclash-core-meta` on the previous line. The other five additions from `a288b92` (`block-mount`, `e2fsprogs`, `kmod-fs-exfat`, `kmod-fs-ntfs3`, `kmod-fs-vfat`, `lsblk`) were all placed correctly, so this looks like an oversight.

---

openwrt-ai (commented):
Re-reviewed the 10 commits added since `300010fe` (`7c71fc3`..`a3a1a2e`). Earlier findings on the series as a whole still stand and are not repeated here.

## Commit checks

- **All 10 new commits** — still no `Signed-off-by:` trailer and still a bare subject with an empty body (`git log --format='%b' 300010fe..a3a1a2e` finds zero sign-offs). OpenWrt requires DCO sign-off on every commit; `git commit -s` / `git rebase --signoff` will add it. The PR description is also still the unmodified template text.
- `f9ed6c7` **"mediatek: e87n: align display previews and preseed NVMe share"** — the subject admits two unrelated changes, and they are unrelated: resizing the LuCI preview assets has nothing to do with preseeding a Samba share in `99-e87n-defaults`. Please split. The Samba half in particular needs its own message explaining the chosen share semantics (see the inline comment).
- `8bd0563` **"luci-app-display: fix live preview switching"** — the subject describes a JS state-machine fix, but the diff also reverts the five `screen*.png` back to their pre-`f9ed6c7` 2006x784 versions (`Bin 1282 -> 214546`, …). That revert is undocumented, it re-adds ~5 MB of binaries to the history one commit after they were shrunk, and it leaves the CI self-check contradicting itself (inline on `e87n-build.yml:621`). Squashing `f9ed6c7`'s image change with this revert would keep both out of the history entirely.
- `7c71fc3` **"mediatek: e87n: remove bandwidth monitor"** — this removes the nlbwmon support added earlier in the same series by `f6ae305c` "mediatek: e87n: add persistent nlbwmon". Squash the pair away rather than shipping add-then-remove.
- `6af4d58` **"luci-theme-argon: pin unsaved changes indicator"** and `b0242e8` **"luci-theme-argon: restore responsive header indicators"** — introduce-then-fix pair: the first adds 18 lines to `less/layout.less`, the second removes 16 of them and re-adds equivalent rules in `less/responsive.less`. Fold into one.
- `59760db` / `8bd0563` / `a3a1a2e` — three consecutive commits rewriting the same live-preview feature (`fs.read_direct('/dev/fb0')` → `fs.exec_direct(HELPER,['frame'])` → `frame-snapshot` + `fs.read_direct`), each bumping `PKG_RELEASE` and rewriting the same CI assertions. Only the final design survives; the intermediate two are churn and one of them left the now-dead `frame` subcommand behind (inline on `display-control:411`).

---

Substantively, the two things I would not merge as-is are the guest-writable root-privileged Samba share preseeded on first boot (`99-e87n-defaults:74-81`) and the CI check that cannot pass (`e87n-build.yml:621`). The rest of the inline notes are the framebuffer helper's robustness and the 10 Hz poll cost.

---

PR #24726, opened 2026-08-14
feature/snr advance ax2 and bo rt622 g31 g2

akosmakov:
Add support SNR ADVANCE AX2 (SNR-CPE-AX2 with factory bootloader)
Add support BO RT622-G31-G2

### DIFF
--- a/package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic
+++ b/package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic
@@ -82,6 +82,7 @@ nradio,c8-668gl)
 	ubootenv_add_mmc "u-boot-env" "" "0x0" "0x80000"
 	;;
 asiarf,ap7986-003|\
+bo,rt622-g31-g2|\
 cetron,ct3003|\
 comfast,cf-wr632ax|\
 edgecore,eap111|\
--- a/package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic
+++ b/package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic
@@ -86,6 +86,7 @@ cetron,ct3003|\
 comfast,cf-wr632ax|\
 edgecore,eap111|\
 netgear,wax220|\
+snr,advance-ax2|\
 zbtlink,zbt-z8102ax|\
 zbtlink,zbt-z8103ax)
 	ubootenv_add_uci_config "/dev/mtd1" "0x0" "0x20000" "0x20000"
--- a/target/linux/mediatek/dts/mt7981b-bo-rt622-g31-g2.dts
+++ b/target/linux/mediatek/dts/mt7981b-bo-rt622-g31-g2.dts
@@ -1,0 +1,302 @@
+// SPDX-License-Identifier: GPL-2.0-only OR MIT
+/*
+ * Copyright (C) 2026 LLC NAGTECH
+ */
+
+/dts-v1/;
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+
+#include "mt7981b.dtsi"
+
+/ {
+	model = "BO RT622-G31-G2";
+	compatible = "bo,rt622-g31-g2", "mediatek,mt7981";
+
+	aliases {
+		label-mac-device = &gmac0;
+
+		led-boot = &led_sys;
+		led-failsafe = &led_sys;
+		led-running = &led_sys;
+		led-upgrade = &led_sys;
+
+		serial0 = &uart0;
+	};
+
+	chosen {
+		stdout-path = "serial0:115200n8";
+	};
+
+	memory@40000000 {
+		device_type = "memory";
+		reg = <0 0x40000000 0 0x10000000>;
+	};
+
+	gpio-keys {
+		compatible = "gpio-keys";
+
+		button-0 {
+			label = "mesh";
+			gpios = <&pio 0 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RFKILL>;
+			debounce-interval = <60>;
+		};
+
+		button-1 {
+			label = "reset";
+			gpios = <&pio 1 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+			debounce-interval = <60>;
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		led_sys: led_sys {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_STATUS;
+			gpios = <&pio 8 GPIO_ACTIVE_LOW>;
+			default-state = "on";
+		};
+
+		led_wan {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WAN;
+			gpios = <&pio 4 GPIO_ACTIVE_LOW>;
+		};
+
+		led_mesh {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_INDICATOR;
+			gpios = <&pio 5 GPIO_ACTIVE_LOW>;
+		};
+
+		led_wlan_2g {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WLAN_2GHZ;
+			gpios = <&pio 6 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "phy0tpt";
+		};
+
+		led_wlan_5g {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WLAN_5GHZ;
+			gpios = <&pio 7 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "phy1tpt";
+		};
+	};
+};
+
+&eth {
+	status = "okay";
+
+	gmac0: mac@0 {
+		compatible = "mediatek,eth-mac";
+		reg = <0>;
+		phy-mode = "2500base-x";
+
+		nvmem-cell-names = "mac-address";
+		nvmem-cells = <&macaddr_factory_e000 0>;
+
+		fixed-link {
+			speed = <2500>;
+			full-duplex;
+			pause;
+		};
+	};
+};
+
+&mdio_bus {
+	switch: switch@1f {
+		compatible = "mediatek,mt7531";
+		reg = <0x1f>;
+		reset-gpios = <&pio 39 GPIO_ACTIVE_HIGH>;
+		interrupt-controller;
+		#interrupt-cells = <1>;
+		interrupt-parent = <&pio>;
+		interrupts = <38 IRQ_TYPE_LEVEL_HIGH>;
+	};
+};
+
+&switch {
+	ports {
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		port@1 {
+			reg = <1>;
+			label = "lan3";
+		};
+
+		port@2 {
+			reg = <2>;
+			label = "lan2";
+		};
+
+		port@3 {
+			reg = <3>;
+			label = "lan1";
+		};
+
+		port@4 {
+			reg = <4>;
+			label = "wan";
+			nvmem-cell-names = "mac-address";
+			nvmem-cells = <&macaddr_factory_e006 0>;
+		};
+
+		port@6 {
+			reg = <6>;
+			ethernet = <&gmac0>;
+			phy-mode = "2500base-x";
+
+			fixed-link {
+				speed = <2500>;
+				full-duplex;
+				pause;
+			};
+		};
+	};
+};
+
+&spi0 {
+	pinctrl-names = "default";
+	pinctrl-0 = <&spi0_flash_pins>;
+	status = "okay";
+
+	spi_nand@0 {
+		compatible = "spi-nand";
+		reg = <0>;
+
+		spi-max-frequency = <52000000>;
+		spi-tx-bus-width = <4>;
+		spi-rx-bus-width = <4>;
+
+		spi-cal-enable;
+		spi-cal-mode = "read-data";
+		spi-cal-datalen = <7>;
+		spi-cal-data = /bits/ 8 <0x53 0x50 0x49 0x4e 0x41 0x4e 0x44>;
+		spi-cal-addrlen = <5>;
+		spi-cal-addr = /bits/ 32 <0x0 0x0 0x0 0x0 0x0>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "BL2";
+				reg = <0x0 0x100000>;
+				read-only;
+			};
+
+			partition@100000 {
+				label = "u-boot-env";
+				reg = <0x100000 0x80000>;
+			};
+
+			partition@180000 {
+				label = "Factory";
+				reg = <0x180000 0x200000>;
+				read-only;
+
+				nvmem-layout {
+					compatible = "fixed-layout";
+					#address-cells = <1>;
+					#size-cells = <1>;
+
+					eeprom_factory_0: eeprom@0 {
+						reg = <0x0 0x1000>;
+					};
+
+					macaddr_factory_4: macaddr@4 {
+						compatible = "mac-base";
+						reg = <0x4 0x6>;
+						#nvmem-cell-cells = <1>;
+					};
+
+					macaddr_factory_e000: macaddr@e000 {
+						compatible = "mac-base";
+						reg = <0xe000 0x6>;
+						#nvmem-cell-cells = <1>;
+					};
+
+					macaddr_factory_e006: macaddr@e006 {
+						compatible = "mac-base";
+						reg = <0xe006 0x6>;
+						#nvmem-cell-cells = <1>;
+					};
+
+					macaddr_factory_100004: macaddr@100004 {
+						compatible = "mac-base";
+						reg = <0x100004 0x6>;
+						#nvmem-cell-cells = <1>;
+					};
+				};
+			};
+
+			partition@380000 {
+				label = "FIP";
+				reg = <0x380000 0x200000>;
+				read-only;
+			};
+
+			partition@580000 {
+				label = "ubi";
+				reg = <0x580000 0x7280000>;
+			};
+		};
+	};
+};
+
+&pio {
+	spi0_flash_pins: spi0-pins {
+		mux {
+			function = "spi";
+			groups = "spi0", "spi0_wp_hold";
+		};
+
+		conf-pu {
+			pins = "SPI0_CS", "SPI0_HOLD", "SPI0_WP";
+			drive-strength = <MTK_DRIVE_8mA>;
+			bias-pull-up = <MTK_PUPD_SET_R1R0_11>;
+		};
+
+		conf-pd {
+			pins = "SPI0_CLK", "SPI0_MOSI", "SPI0_MISO";
+			drive-strength = <MTK_DRIVE_8mA>;
+			bias-pull-down = <MTK_PUPD_SET_R1R0_11>;
+		};
+	};
+};
+
+&uart0 {
+	status = "okay";
+};
+
+&watchdog {
+	status = "okay";
+};
+
+&wifi {
+	status = "okay";
+	#address-cells = <1>;
+	#size-cells = <0>;
+	nvmem-cell-names = "eeprom";
+	nvmem-cells = <&eeprom_factory_0>;
+
+	band@0 {
+		reg = <0>;
+		nvmem-cells = <&macaddr_factory_4 0>;
+		nvmem-cell-names = "mac-address";
+	};
+
+	band@1 {
+		reg = <1>;
+		nvmem-cells = <&macaddr_factory_100004 0>;
+		nvmem-cell-names = "mac-address";
+	};
+};
--- a/target/linux/mediatek/dts/mt7981b-snr-advance-ax2.dts
+++ b/target/linux/mediatek/dts/mt7981b-snr-advance-ax2.dts
@@ -1,0 +1,317 @@
+// SPDX-License-Identifier: GPL-2.0-only OR MIT
+/*
+ * Copyright (C) 2026 LLC NAGTECH
+ */
+
+/dts-v1/;
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+
+#include "mt7981b.dtsi"
+
+/ {
+	model = "SNR ADVANCE AX2";
+	compatible = "snr,advance-ax2", "mediatek,mt7981";
+
+	aliases {
+		label-mac-device = &gmac0;
+
+		led-boot = &led_fail;
+		led-failsafe = &led_fail;
+		led-running = &led_sys;
+		led-upgrade = &led_sys;
+
+		serial0 = &uart0;
+	};
+
+	chosen {
+		stdout-path = "serial0:115200n8";
+	};
+
+	memory@40000000 {
+		device_type = "memory";
+		reg = <0 0x40000000 0 0x10000000>;
+	};
+
+	gpio-keys {
+		compatible = "gpio-keys";
+
+		button-0 {
+			label = "mesh";
+			gpios = <&pio 0 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RFKILL>;
+			debounce-interval = <60>;
+		};
+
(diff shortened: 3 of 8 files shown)


target/linux/mediatek/image/filogic.mk
```diff
@@ -3152,6 +3173,27 @@ define Device/ruijie_rg-x60-pro
 endef
 TARGET_DEVICES += ruijie_rg-x60-pro
 
+define Device/snr_advance-ax2
+  DEVICE_VENDOR := SNR
+  DEVICE_MODEL := ADAVANCE AX2
```
openwrt-ai: Typo: `ADAVANCE`. This string ends up in `/etc/board.json`, in the image filename and on the download page, so it needs to read `ADVANCE`.

```suggestion
  DEVICE_MODEL := ADVANCE AX2
```

---

target/linux/mediatek/dts/mt7981b-snr-advance-ax2.dts
```diff
@@ -0,0 +1,332 @@
...
+
+/dts-v1/;
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+
+#include "mt7981b.dtsi"
+
+/ {
+	model = "ADVANCE AX2";
```
openwrt-ai: nit: the `model` string drops the vendor, so it disagrees with `DEVICE_VENDOR`/`DEVICE_MODEL` in `filogic.mk` (which compose to "SNR ADVANCE AX2") and with the sibling board, which uses `model = "SNR SNR-CPE-AX2";` in [`mt7981b-snr-snr-cpe-ax2.dts:11`](https://github.com/openwrt/openwrt/blob/39dac631affbe5698c4a346f071a52c4d7e3735d/target/linux/mediatek/dts/mt7981b-snr-snr-cpe-ax2.dts#L11).

```suggestion
	model = "SNR ADVANCE AX2";
```

---

target/linux/mediatek/dts/mt7981b-bo-rt622-g31-g2.dts
```diff
@@ -0,0 +1,298 @@
...
+	gpio-leds {
+		compatible = "gpio-leds";
+
+		led_sys: led-0 {
+			color = <LED_COLOR_ID_GREEN>;
+			label = "green:system";
+			function = LED_FUNCTION_STATUS;
+			gpios = <&pio 8 GPIO_ACTIVE_LOW>;
+			default-state = "on";
+		};
```
openwrt-ai: `label` is the deprecated LED naming form and it is not merely redundant here — it silently disables the `color`/`function` properties you added. `led_parse_fwnode_props()` returns as soon as it sees a `label`, so `color` and `function` are never parsed: [`drivers/leds/led-core.c:485-490`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/leds/led-core.c#L485-L490) (v6.18.44, the target's `KERNEL_PATCHVER`). Drop `label` and let the name be composed from `color` + `function`.

```suggestion
		led_sys: led-0 {
			color = <LED_COLOR_ID_GREEN>;
			function = LED_FUNCTION_STATUS;
			gpios = <&pio 8 GPIO_ACTIVE_LOW>;
			default-state = "on";
		};
```

Same applies to `led-1` (line 68) and `led-2` (line 75) here, and to all seven LED nodes in `mt7981b-snr-advance-ax2.dts` (lines 60, 67, 74, 82, 90, 97, 104). The sibling board already does it this way — [`mt7981b-snr-snr-cpe-ax2.dts`](https://github.com/openwrt/openwrt/blob/39dac631affbe5698c4a346f071a52c4d7e3735d/target/linux/mediatek/dts/mt7981b-snr-snr-cpe-ax2.dts#L48-L92) has no `label` at all.

Note the resulting class-device names change for two of them: `green:system` becomes `green:status` and `green:mesh` becomes `green:indicator`. `green:wan` is unchanged, so `01_leds` for this board still matches. See the separate note on `01_leds` for the SNR board.

---

target/linux/mediatek/dts/mt7981b-snr-advance-ax2.dts
```diff
@@ -0,0 +1,332 @@
...
+			function = LED_FUNCTION_WAN;
+			gpios = <&pio 10 GPIO_ACTIVE_LOW>;
+		};
+
+		led-6 {
+			color = <LED_COLOR_ID_BLUE>;
+			label = "blue:mesh";
+			function = LED_FUNCTION_STATUS;
+			gpios = <&pio 11 GPIO_ACTIVE_LOW>;
+		};
```
openwrt-ai: The mesh LED is declared as `LED_FUNCTION_STATUS`, which is the same color+function pair as `led_sys` (`led-4`, line 88). Today the `label` hides that, but once the deprecated `label` properties are dropped both nodes compose to the name `blue:status` and the second registration fails. The BO board in this same PR uses `LED_FUNCTION_INDICATOR` for its mesh LED, so use that here too.

```suggestion
		led-6 {
			color = <LED_COLOR_ID_BLUE>;
			function = LED_FUNCTION_INDICATOR;
			gpios = <&pio 11 GPIO_ACTIVE_LOW>;
		};
```

---

target/linux/mediatek/filogic/base-files/etc/board.d/01_leds
```diff
@@ -313,6 +316,11 @@ smartrg,sdg-8733a)
 	ucidef_set_led_netdev "wan-orange" "WAN" "mdio-bus:08:orange:wan" "wan" "link_100 link_1000"
 	ucidef_set_led_netdev "wan-white" "WAN" "mdio-bus:08:white:wan" "wan" "link_10000"
 	;;
+snr,advance-ax2)
+	ucidef_set_led_netdev "wan" "wan" "blue:wan" "wan"
+	ucidef_set_led_netdev "wlan2g" "WLAN2G" "blue:wlan2g" "phy0-ap0" "link tx rx"
+	ucidef_set_led_netdev "wlan5g" "WLAN5G" "blue:wlan5g" "phy1-ap0" "link tx rx"
```
openwrt-ai: These two LEDs already carry `linux,default-trigger = "phy0tpt"` / `"phy1tpt"` in the DTS (lines 77 and 85 of `mt7981b-snr-advance-ax2.dts`). Configuring a netdev trigger here on top of that means the DTS trigger is overwritten on first boot, so one of the two is dead configuration — pick one. The sibling [`snr,snr-cpe-ax2` entry](https://github.com/openwrt/openwrt/blob/39dac631affbe5698c4a346f071a52c4d7e3735d/target/linux/mediatek/filogic/base-files/etc/board.d/01_leds#L30-L33) keeps only the DTS `phyXtpt` triggers and configures just the WAN LED here.

Also, if the deprecated `label` properties are removed from the DTS as noted there, these sysfs names have to become `blue:wlan-2ghz` and `blue:wlan-5ghz` (from `LED_FUNCTION_WLAN_2GHZ`/`_5GHZ`), not `blue:wlan2g`/`blue:wlan5g`. `blue:wan` on line 320 stays valid either way.

---

target/linux/mediatek/dts/mt7981b-bo-rt622-g31-g2.dts
```diff
@@ -0,0 +1,298 @@
...
+
+		port@2 {
+			reg = <2>;
+			label = "lan2";
+		};
+
+		port@3 {
+			reg = <3>;
+			label = "lan1";
```
openwrt-ai, line 141: This board only has `lan1`..`lan3`, but no case was added to `target/linux/mediatek/filogic/base-files/etc/board.d/02_network`, so it falls through to the [default `*)` branch](``https://github.com/openwrt/openwrt/blob/39dac631affbe5698c4a346f071a52c4d7e3735d/target/linux/mediatek/filogic/base-files/etc/board.d/02_network#L344-L346),`` which does `ucidef_set_interfaces_lan_wan "lan1 lan2 lan3 lan4" wan`. The resulting default `br-lan` will list a non-existent `lan4` port. Please add

```
	bo,rt622-g31-g2)
		ucidef_set_interfaces_lan_wan "lan1 lan2 lan3" wan
		;;
```

to `mediatek_setup_interfaces()`. (`snr,advance-ax2` has four LAN ports, so the default is correct for it.)

---

package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic
```diff
@@ -163,6 +163,10 @@ smartrg,sdg-8733a|\
 smartrg,sdg-8734)
 	ubootenv_add_mmc "u-boot-env" "mmcblk0" "0x0" "0x8000" "0x8000"
 	;;
+bo,rt622-g31-g2|\
+snr,advance-ax2)
+	ubootenv_add_uci_config "/dev/mtd1" "0x0" "0x20000" "0x20000" "1"
```
openwrt-ai: Two things here:

1. This block is inserted between `smartrg,...` and `teltonika,...`, which breaks the alphabetical ordering the rest of this `case` follows (acer, asiarf, asus, bananapi, buffalo, comfast, dlink, gatonetworks, glinet, jiorouter, openembed, smartrg, teltonika, tplink, ubnt, xiaomi, zyxel). `bo,rt622-g31-g2` sorts between `bananapi` and `buffalo`.
2. The parameters are identical to the existing group at [line 91](``https://github.com/openwrt/openwrt/blob/39dac631affbe5698c4a346f071a52c4d7e3735d/package/boot/uboot-tools/uboot-envtools/files/mediatek_filogic#L83-L92``) (`asiarf,ap7986-003|cetron,ct3003|...`) — the trailing `"1"` is the default sector count, so it adds nothing. Adding the two new boards to that existing group instead removes the duplication and keeps the sorting intact.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -3152,6 +3173,27 @@ define Device/ruijie_rg-x60-pro
...
+  DEVICE_DTS := mt7981b-snr-advance-ax2
+  DEVICE_DTS_DIR := ../dts
+  SUPPORTED_DEVICES += snr-cpe-ax2
```
openwrt-ai, line 3205: `SUPPORTED_DEVICES` already defaults to the device name with `_` turned into `,` ([`include/image.mk:551`](https://github.com/openwrt/openwrt/blob/39dac631affbe5698c4a346f071a52c4d7e3735d/include/image.mk#L551)), i.e. `snr,advance-ax2`, and it is matched against `board_name` (the DT compatible) during sysupgrade. `snr-cpe-ax2` can never match any `board_name`, so this alias is dead — and it is confusingly close to the unrelated in-tree `snr_snr-cpe-ax2` device. Since this is brand-new device support there are no legacy images to stay compatible with. Was this intended for something specific? Same question for `SUPPORTED_DEVICES += bo-rt622-g31-g2` on line 862.

Related: the commit message says this is "similar to the SNR-CPE-AX2 but uses the factory bootloader". For that situation filogic normally keeps one device name with `-stock` / `-ubootmod` suffixes (`xiaomi,mi-router-wr30u-stock` / `-ubootmod`, `acer,predator-w6x-stock` / `-ubootmod`, ...). Is `snr,advance-ax2` a genuinely different retail product, or is this the stock-bootloader variant of `snr,snr-cpe-ax2`?

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -854,6 +854,27 @@ define Device/bazis_ax3000wm
...
+define Device/bo_rt622-g31-g2
+  DEVICE_VENDOR := BO
+  DEVICE_MODEL := RT622-G31-G2
+  DEVICE_DTS := mt7981b-bo-rt622-g31-g2
+  DEVICE_DTS_DIR := ../dts
+  SUPPORTED_DEVICES += bo-rt622-g31-g2
+  DEVICE_PACKAGES := kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware
+  KERNEL_LOADADDR := 0x48080000
+  KERNEL := kernel-bin | lzma | \
+	fit-with-netgear-top-level-rootfs-node lzma $$(KDIR)/image-$$(firstword $$(DEVICE_DTS)).dtb
```
openwrt-ai, line 867: `fit-with-netgear-top-level-rootfs-node` is a NETGEAR-specific workaround — it shells out to [`scripts/gen_netgear_rootfs_node.sh`](https://github.com/openwrt/openwrt/blob/39dac631affbe5698c4a346f071a52c4d7e3735d/target/linux/mediatek/image/filogic.mk#L11-L18) and until now had exactly one user, `netgear_eax17` (line 2766). Neither commit message explains why an SNR/BO device with the MTK stock bootloader needs a top-level rootfs node in the FIT. Could you add a sentence about the bootloader requirement to the commit message, and — if this is a generic stock-bootloader requirement rather than a NETGEAR quirk — rename the macro to something vendor-neutral? Same line in the `snr_advance-ax2` block (line 3185).

---

target/linux/mediatek/dts/mt7981b-snr-advance-ax2.dts
```diff
@@ -0,0 +1,332 @@
...
+		compatible = "gpio-leds";
+
+		led-0 {
+			color = <LED_COLOR_ID_RED>;
+			label = "red:wan";
+			function = LED_FUNCTION_WAN;
+			gpios = <&pio 4 GPIO_ACTIVE_LOW>;
+		};
+
+		led_fail: led-1 {
```
openwrt-ai: nit: the `led_fail` phandle label is defined but never referenced — all four aliases (lines 20-23) point at `&led_sys`. Either drop the label, or wire the failsafe/upgrade states to this dedicated red status LED the way the sibling board does (`led-failsafe = &led_fail; led-upgrade = &led_fail;` in [`mt7981b-snr-snr-cpe-ax2.dts:17-20`](https://github.com/openwrt/openwrt/blob/39dac631affbe5698c4a346f071a52c4d7e3735d/target/linux/mediatek/dts/mt7981b-snr-snr-cpe-ax2.dts#L17-L20)); a red LED for failsafe/upgrade is much more visible than reusing the blue system LED that is also `led-running`.

---

openwrt-ai (commented):
## Commit checks

Both commit subjects match what their diffs actually do, but both messages are missing a section that `.github/llm-review-rules.md` requires for new-device support:

- `4591b16` "mediatek: filogic: add support for SNR ADVANCE AX2" — no **MAC address layout** section. The DTS sources MACs from the `Factory` partition at `0x4` (2.4G), `0xe000` (gmac0/label), `0xe006` (WAN) and `0x100004` (5G); please document one line per interface with `xx:xx:xx` for the unit-specific bytes. The hardware spec also omits the LED list, and "DC power interface" gives no voltage/current.
- `39dac63` "mediatek: filogic: add support for BO RT622-G31-G2" — same missing **MAC address layout** section, same missing LED list and power rating.

The flash instructions are fine as far as they go; a note on how to revert to stock (if a procedure exists) would round them out.


---

target/linux/mediatek/dts/mt7981b-snr-advance-ax2.dts
```diff
@@ -0,0 +1,321 @@
...
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		led_wan_red {
+			color = <LED_COLOR_ID_RED>;
+			function = LED_FUNCTION_INDICATOR;
+			gpios = <&pio 4 GPIO_ACTIVE_LOW>;
+		};
```
openwrt-ai: This LED and `led_fail` right below both declare `LED_COLOR_ID_RED` + `LED_FUNCTION_INDICATOR`, so both class devices resolve to the same sysfs name `red:indicator`. The LED core resolves the collision by appending a suffix to whichever registers second, which makes the name unstable and unusable from `board.d`. The previous push of this file had this node as `label = "red:wan"` / `function = LED_FUNCTION_WAN`, which also matches the node name — restoring it keeps the two red LEDs distinct.

```suggestion
		led-0 {
			color = <LED_COLOR_ID_RED>;
			function = LED_FUNCTION_WAN;
			gpios = <&pio 4 GPIO_ACTIVE_LOW>;
		};
```
(The `led-0` rename is the separate node-naming point raised below.)

---

target/linux/mediatek/dts/mt7981b-snr-advance-ax2.dts
```diff
@@ -0,0 +1,321 @@
...
+			color = <LED_COLOR_ID_BLUE>;
+			function = LED_FUNCTION_WLAN_2GHZ;
+			gpios = <&pio 6 GPIO_ACTIVE_LOW>;
+		};
+
+		wlan_5g {
+			color = <LED_COLOR_ID_BLUE>;
+			function = LED_FUNCTION_WLAN_5GHZ;
+			gpios = <&pio 7 GPIO_ACTIVE_LOW>;
+		};
```
openwrt-ai: `wlan_2g` and `wlan_5g` do not contain the substring `led`, so they match neither branch of the `gpio-leds` child pattern [`"(^led-[0-9a-f]$|led)"` in leds-gpio.yaml:24](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/leds/leds-gpio.yaml#L24); with `additionalProperties: false` on the parent these nodes are rejected by dtbs_check. The whole block was renamed from `led-0`…`led-6` to underscore names in the last push — the in-tree sibling [`mt7981b-snr-snr-cpe-ax2.dts` LEDs](https://github.com/openwrt/openwrt/blob/27031edac617fe0125e2a967e4b74e1b544ed815/target/linux/mediatek/dts/mt7981b-snr-snr-cpe-ax2.dts#L49-L93) still uses the `led-N` form, and underscores in DT node names are against convention generally.

```suggestion
		led-2 {
			color = <LED_COLOR_ID_BLUE>;
			function = LED_FUNCTION_WLAN_2GHZ;
			gpios = <&pio 6 GPIO_ACTIVE_LOW>;
		};

		led-3 {
			color = <LED_COLOR_ID_BLUE>;
			function = LED_FUNCTION_WLAN_5GHZ;
			gpios = <&pio 7 GPIO_ACTIVE_LOW>;
		};
```

The same rename applies to `led_fail:` (line 64 → `led_fail: led-1`), `led_sys:` (83 → `led_sys: led-4`), `led_wan` (89 → `led-5`), `led_mesh` (95 → `led-6`), and to all five LED nodes in `mt7981b-bo-rt622-g31-g2.dts` (lines 58-87) — keep the `led_fail:`/`led_sys:` labels, only the node names change.

---

target/linux/mediatek/dts/mt7981b-snr-advance-ax2.dts
```diff
@@ -0,0 +1,321 @@
...
+	status = "okay";
+};
+
+&watchdog {
+	status = "okay";
+};
+
+&xhci {
+	status = "okay";
+};
```
openwrt-ai: nit: `&xhci` is enabled here (and at line 282 of `mt7981b-bo-rt622-g31-g2.dts`), but neither commit message lists a USB port in the hardware specification and neither `Device/snr_advance-ax2` nor `Device/bo_rt622-g31-g2` pulls in `kmod-usb3`, so nothing would be usable even if a port exists. The in-tree sibling [`mt7981b-snr-snr-cpe-ax2.dts`](https://github.com/openwrt/openwrt/blob/27031edac617fe0125e2a967e4b74e1b544ed815/target/linux/mediatek/dts/mt7981b-snr-snr-cpe-ax2.dts) leaves `&xhci` disabled. Do these boards actually have USB? If yes, please add the port to the spec section and `kmod-usb3` to `DEVICE_PACKAGES`; if not, drop the node.

---

target/linux/mediatek/dts/mt7981b-bo-rt622-g31-g2.dts
```diff
@@ -0,0 +1,304 @@
...
+			function = LED_FUNCTION_WLAN_2GHZ;
+			gpios = <&pio 6 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "phy0tpt";
+		};
+		led_wlan_5g {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WLAN_5GHZ;
+			gpios = <&pio 7 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "phy1tpt";
+		};
```
openwrt-ai, line 89: nit: two things in this hunk. The blank-line separator between LED nodes is missing before `led_wlan_2g` and before `led_wlan_5g`, unlike every other node in this block. And the WLAN LEDs here are driven from the DTS with `linux,default-trigger = "phy0tpt"/"phy1tpt"`, while the sibling board added in the same series wires its WLAN LEDs up through `board.d/01_leds` netdev triggers (`ucidef_set_led_netdev "wlan2g" ... "blue:wlan-2ghz" "phy0-ap0" "link tx rx"`). Picking one approach for both devices would be more consistent — the `board.d` netdev form is the one used by most recent filogic boards.

```suggestion
		};

		led_wlan_2g {
			color = <LED_COLOR_ID_GREEN>;
			function = LED_FUNCTION_WLAN_2GHZ;
			gpios = <&pio 6 GPIO_ACTIVE_LOW>;
			linux,default-trigger = "phy0tpt";
		};

		led_wlan_5g {
			color = <LED_COLOR_ID_GREEN>;
			function = LED_FUNCTION_WLAN_5GHZ;
			gpios = <&pio 7 GPIO_ACTIVE_LOW>;
			linux,default-trigger = "phy1tpt";
		};
```

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -854,6 +854,29 @@ define Device/bazis_ax3000wm
 endef
 TARGET_DEVICES += bazis_ax3000wm
 
+define Device/bo_rt622-g31-g2
+  DEVICE_VENDOR := BO
+  DEVICE_MODEL := RT622-G31-G2
+  DEVICE_DTS := mt7981b-bo-rt622-g31-g2
+  DEVICE_DTS_DIR := ../dts
+  SUPPORTED_DEVICES += bo-rt622-g31-g2
```
openwrt-ai: `SUPPORTED_DEVICES` already defaults to the comma form derived from the device name (`SUPPORTED_DEVICES := $(subst _,$(comma),$(1))` in `include/image.mk`), i.e. `bo,rt622-g31-g2`, which matches this board's `compatible`. Adding the hyphenated `bo-rt622-g31-g2` on top only makes sense if some OEM bootloader or stock firmware reports that exact string — which is what the `snr_advance-ax2` block documents for its own extra entry ("snr-cpe-ax2" is needed to pass the OEM U-boot model compliance check). Is this entry actually required here, or is it a copy of the SNR block? If required, please add the matching one-line comment; otherwise drop it.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -854,6 +854,29 @@ define Device/bazis_ax3000wm
...
+  DEVICE_PACKAGES := kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware
+  KERNEL_LOADADDR := 0x48080000
+  # fit-with-netgear-top-level-rootfs-node is need for top-level /rootfs node, which the
+  # OEM U-Boot verifies at every boot.
```
openwrt-ai: nit: "is need for" → "is needed for". The identical comment at lines 3209-3210 in the `snr_advance-ax2` block has the same typo.

```suggestion
  # fit-with-netgear-top-level-rootfs-node is needed for the top-level /rootfs node, which the
  # OEM U-Boot verifies at every boot.
```

---

openwrt-ai (commented):
Re-review after the force-push (previous review was against `39dac63`). Both commits were rebased onto current main; the delta versus what I reviewed last is the LED rework in both DTS files, the `flash@0` → `spi_nand@0` / NMBM changes, the `ADAVANCE` → `ADVANCE` typo fix, the two new explanatory comments in `filogic.mk`, and the uboot-envtools entries being folded into the existing `/dev/mtd1` group. CI on `27031ed` is green.

## Commit checks

Both subjects and both file scopes match their diffs, and the `ADAVANCE AX2` → `ADVANCE AX2` fix and the `model = "SNR ADVANCE AX2"` alignment from the last round look good. Still outstanding from the previous review, per `.github/llm-review-rules.md`:

- `80e4734` "mediatek: filogic: add support for SNR ADVANCE AX2" — no **MAC address layout** section. The DTS sources MACs from the `Factory` partition at `0x4` (2.4G), `0xe000` (gmac0/label), `0xe006` (WAN) and `0x100004` (5G); one line per interface with `xx:xx:xx` for the unit-specific bytes would cover it. The hardware specification also still omits the LED list (seven LEDs are described in the DTS) and gives no voltage/current for "DC power interface".
- `27031ed` "mediatek: filogic: add support for BO RT622-G31-G2" — same missing **MAC address layout** section, same missing LED list (five LEDs) and power rating.

---

PR #24728, opened 2026-08-14
ramips: mt7530: use IS_REACHABLE for swconfig code

namiltd:
Commit 224a967ba113 ("ramips: mt7530: compile swconfig code conditionally") used IS_ENABLED(CONFIG_SWCONFIG) to guard swconfig-specific code.

When CONFIG_ALL_KMODS=y is selected, kmod-swconfig's KCONFIG override forces CONFIG_SWCONFIG=m during the kernel config merge.

An explicit CONFIG_SWCONFIG=y is protected from being downgraded by the kmod override, but an absent/unset symbol is not, so the override's =m wins instead.

IS_ENABLED() evaluates to 1 both for built-in (=y) and module (=m), so the swconfig code is still compiled into the built-in mt7530 driver even though register_switch() lives in the external swconfig.ko module and is not available during vmlinux linking, resulting in:

  drivers/net/ethernet/ralink/mt7530.c: undefined reference to `register_switch'

Switch to IS_REACHABLE(CONFIG_SWCONFIG) which returns true only when the symbol is actually reachable from the current compilation unit. This correctly handles the case where mt7530 is built-in and swconfig is a module, while still allowing both to be built as modules. Also update the diagnostic message in the fallback branch, since it is now also reached when swconfig is built as an unreachable module, not only when support is absent entirely.

Fixes: 224a967ba113 ("ramips: mt7530: compile swconfig code conditionally")

### DIFF
--- a/target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c
+++ b/target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c
@@ -160,7 +160,7 @@ enum {
 	MT7530_ATTR_ENABLE_VLAN,
 };
 
-#if IS_ENABLED(CONFIG_SWCONFIG)
+#if IS_REACHABLE(CONFIG_SWCONFIG)
 struct mt7530_port_entry {
 	u16	pvid;
 	bool	mirror_rx;
@@ -177,7 +177,7 @@ struct mt7530_vlan_entry {
 struct mt7530_priv {
 	void __iomem		*base;
 	struct mii_bus		*bus;
-#if IS_ENABLED(CONFIG_SWCONFIG)
+#if IS_REACHABLE(CONFIG_SWCONFIG)
 	struct switch_dev	swdev;
 
 	u8			mirror_dest_port;
@@ -188,7 +188,7 @@ struct mt7530_priv {
 #endif /* CONFIG_SWCONFIG */
 };
 
-#if IS_ENABLED(CONFIG_SWCONFIG)
+#if IS_REACHABLE(CONFIG_SWCONFIG)
 struct mt7530_mapping {
 	char	*name;
 	u16	pvids[MT7530_NUM_PORTS];
@@ -330,7 +330,7 @@ mt7530_w32(struct mt7530_priv *priv, u32 reg, u32 val)
 	iowrite32(val, priv->base + reg);
 }
 
-#if IS_ENABLED(CONFIG_SWCONFIG)
+#if IS_REACHABLE(CONFIG_SWCONFIG)
 static void
 mt7530_vtcr(struct mt7530_priv *priv, u32 cmd, u32 val)
 {
@@ -1019,7 +1019,7 @@ mt7530_probe(struct device *dev, void __iomem *base, struct mii_bus *bus, int vl
 	bool swconfig_active = false;
 	const char *name = bus ? "mt7530" : "mt7620";
 	int i;
-#if IS_ENABLED(CONFIG_SWCONFIG)
+#if IS_REACHABLE(CONFIG_SWCONFIG)
 	struct switch_dev *swdev;
 	struct mt7530_mapping *map;
 	int ret;
@@ -1031,7 +1031,7 @@ mt7530_probe(struct device *dev, void __iomem *base, struct mii_bus *bus, int vl
 
 	mt7530->base = base;
 	mt7530->bus = bus;
-#if IS_ENABLED(CONFIG_SWCONFIG)
+#if IS_REACHABLE(CONFIG_SWCONFIG)
 	mt7530->global_vlan_enable = vlan;
 
 	swdev = &mt7530->swdev;
@@ -1058,7 +1058,7 @@ mt7530_probe(struct device *dev, void __iomem *base, struct mii_bus *bus, int vl
 		dev_info(dev, "swconfig disabled, MAC learning disabled on all ports\n");
 	}
 #else
-	dev_info(dev, "swconfig support not built in, MAC learning disabled on all ports\n");
+	dev_info(dev, "swconfig support not reachable, MAC learning disabled on all ports\n");
 #endif /* CONFIG_SWCONFIG */
 
 	if (!swconfig_active) {


target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c
```diff
@@ -160,7 +160,7 @@ enum {
 	MT7530_ATTR_ENABLE_VLAN,
 };
 
-#if IS_ENABLED(CONFIG_SWCONFIG)
+#if IS_REACHABLE(CONFIG_SWCONFIG)
```
openwrt-ai, line 163: `IS_REACHABLE()` is the right idiom here (`__or(IS_BUILTIN(option), __and(IS_MODULE(option), __is_defined(MODULE)))` — [kconfig.h:65-66](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/include/linux/kconfig.h#L65-L66)), and all six guard sites are converted consistently. But I can't reproduce the premise the commit message gives for it:

- The only ramips subtarget that compiles `mt7530.c` is mt7620 — [`obj-$(CONFIG_NET_RALINK_GSW_MT7620) += gsw_mt7620.o mt7530.o`](https://github.com/openwrt/openwrt/blob/bda951764e14958aac1cece575c96449d17261f9/target/linux/ramips/files/drivers/net/ethernet/ralink/Makefile#L17), and `CONFIG_NET_RALINK_GSW_MT7620=y` only appears in [mt7620/config-6.18:139](https://github.com/openwrt/openwrt/blob/bda951764e14958aac1cece575c96449d17261f9/target/linux/ramips/mt7620/config-6.18#L139). That same subtarget config already carries [`CONFIG_SWCONFIG=y` at line 208](https://github.com/openwrt/openwrt/blob/bda951764e14958aac1cece575c96449d17261f9/target/linux/ramips/mt7620/config-6.18#L208).
- kmod-swconfig's `CONFIG_SWCONFIG` (=m) should not be able to downgrade that: `Kernel/Configure/Default` merges with [`kconfig.pl 'm+' '+' .config.target /dev/null .config.override`](https://github.com/openwrt/openwrt/blob/bda951764e14958aac1cece575c96449d17261f9/include/kernel-defaults.mk#L121), and [`config_add()`](https://github.com/openwrt/openwrt/blob/bda951764e14958aac1cece575c96449d17261f9/scripts/kconfig.pl#L66-L84) with `mod_plus` does `next if $config{$config} eq "y"` for the override file. Checked it directly:

```
$ printf 'CONFIG_SWCONFIG=y\n' > t.target; printf 'CONFIG_SWCONFIG=m\n' > t.override
$ scripts/kconfig.pl 'm+' '+' t.target /dev/null t.override
CONFIG_SWCONFIG=y
```

Which target/config actually produces the `undefined reference to 'register_switch'` link error? If it needs something beyond a stock `CONFIG_ALL_KMODS=y` ramips/mt7620 build (a `CONFIG_KERNEL_*` override, a custom kernel config, a downstream target), it would help to say so in the commit message — as written it reads like `CONFIG_ALL_KMODS=y` alone breaks ramips/mt7620 on main, which the config merge above seems to rule out. The change itself is strictly safer either way, so this is about the justification, not the code.

---

target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c
```diff
@@ -1031,7 +1031,7 @@ mt7530_probe(struct device *dev, void __iomem *base, struct mii_bus *bus, int vl
 
 	mt7530->base = base;
 	mt7530->bus = bus;
-#if IS_ENABLED(CONFIG_SWCONFIG)
+#if IS_REACHABLE(CONFIG_SWCONFIG)
```
openwrt-ai, line 1034: nit: the `#else` branch this hunk feeds into now has a slightly wrong message. [`dev_info(dev, "swconfig support not built in, ...")` at mt7530.c:1061](https://github.com/openwrt/openwrt/blob/bda951764e14958aac1cece575c96449d17261f9/target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c#L1061) was accurate under `IS_ENABLED()` (only reached when `CONFIG_SWCONFIG=n`), but with `IS_REACHABLE()` it is also reached in exactly the configuration this commit targets — `CONFIG_SWCONFIG=m` with a built-in mt7530 — where swconfig *is* built, just not linkable from vmlinux. Something like "swconfig support not reachable" would keep the dmesg line diagnostic for the case that motivated the change.

---

openwrt-ai (commented):
## Commit checks

- `bda95176` "ramips: mt7530: use IS_REACHABLE for swconfig code" — the diff (6 × `IS_ENABLED` → `IS_REACHABLE`) matches the subject, but the message's causal claim ("kmod-swconfig forces `CONFIG_SWCONFIG=m` in the kernel configuration") does not appear to hold for the only ramips subtarget that builds `mt7530.c`: mt7620 sets `CONFIG_SWCONFIG=y` in its subtarget config, and `scripts/kconfig.pl`'s `m+` merge keeps `=y` over the kmod override. Details and a reproduction of the merge behaviour are in the inline comment.
- Same commit — `Fixes: commit 224a967ba113 ("...")` carries an extra `commit ` word; the in-tree convention (and `git log --format=... --abbrev=12`) is `Fixes: 224a967ba113 ("ramips: mt7530: compile swconfig code conditionally")`. The referenced SHA itself is correct (224a967ba1139bdab687d54dd60126e57159e52a).

No CI results on `bda95176` yet (status pending, 0 checks), so nothing CI-grounded to report.

---

target/linux/ramips/files/drivers/net/ethernet/ralink/mt7530.c
namiltd: Any config where the built-in mt7530 driver's file is compiled while CONFIG_SWCONFIG is left unset, combined with CONFIG_ALL_KMODS=y. The kmod-swconfig override elevates the unset symbol to =m (only an explicit =y is protected from that), and IS_ENABLED() then treats =m as reachable at link time for a built-in file. Stock mt7620 is unaffected since it already carries CONFIG_SWCONFIG=y.

jonasjelonek:

PR #24547, opened 2026-08-03
kernel: xhci: fix USB3 port events with a single roothub

strrs:
USB3 stopped working on bcm53xx in 24.10. Nothing appears when a USB3 device is plugged into a USB3 port, while the same device works in a USB2 port and a USB2 device works in the USB3 port.

USB3 port events are dropped whenever the controller has only one roothub. The check that guards them treats a missing shared hcd as proof the hcd was removed, which stopped being true with upstream commit 4736ebd7fcaf ("usb: host: xhci-plat: omit shared hcd if either root hub has no ports") in 5.19. A controller whose USB2 root hub has no ports gets a single roothub and never has a shared one, so every SuperSpeed event is thrown away.

Broadcom Northstar is such a controller, which is why the port reports the device while nothing enumerates:

```
0x006a1203 Powered Connected Enabled Link:U0 PortSpeed:4 Change: CSC WRC PRC PLC
```

Connected, enabled, link up at SuperSpeed, and the connect change bit never cleared because nothing services the event.

Not target specific. Any xhci-plat controller whose USB2 root hub has no ports is affected.

Tested on an Asus RT-N18U on main with kernel 6.18, front USB3 port:

```
usb 3-1: new SuperSpeed USB device number 5 using xhci-hcd
usb 3-1: new SuperSpeed USB device number 6 using xhci-hcd
usb 3-1: new SuperSpeed USB device number 7 using xhci-hcd
usb 3-1: new SuperSpeed USB device number 8 using xhci-hcd
```

Seven SuperSpeed enumerations over repeated insert and remove cycles, seven high speed enumerations on the USB2 ports in the same session, and one case of a USB2 and a USB3 device enumerating a second apart. No kernel warnings. After the fix the `Change:` field is empty, so the events are being serviced.

Fixes: https://github.com/openwrt/openwrt/issues/19592

Issue #19592, opened 2025-07-29
USB3 since 24.10 not working on Asus rt-ac88u | bcm53xx


stevob99:
### Describe the bug

Since  24.10.xx the USB3 port of the router Asus rt-ac88u (bcm53xx) is not working anymore.
(22.03.xx and 23.05.xx has been working absolutely fine)

Nothing occurs at dmesg or logread when inserting the USB3-stick in USB3 port (in the front).
Looks like USB3 has been deactivated.

Just to make sure, I also tested 4 different usb-sticks 3.0 and also 3 different Asus Router rt-ac88u.
 I said, I tested the same device: with openwrt 22.xx this device has been working with the same USB-stick (SanDisk 3.2Gen1) and same usb-port.

### OpenWrt version

r28739-d9340319c6

### OpenWrt release

24.10.2

### OpenWrt target/subtarget

bcm53xx/generic

### Device

Asus RT-AC88U

### Image kind

Official downloaded image

### Steps to reproduce

_just insert any usb3 stick in the front usb slot_

### Actual behaviour

_nothing appears in dmesg / logread_

### Expected behaviour

```
248.730091] kmodloader: loading kernel modules from /etc/modules.d/*
[  248.742054] SCSI subsystem initialized
[  248.751359] usb-storage 4-1:1.0: USB Mass Storage device detected
[  248.758568] scsi host0: usb-storage 4-1:1.0
[  248.763241] usbcore: registered new interface driver usb-storage
[  248.770833] usbcore: registered new interface driver uas
[  248.776751] kmodloader: done loading kernel modules from /etc/modules.d/*
[  248.819225] kmodloader: loading kernel modules from /etc/modules.d/*
[  248.827797] kmodloader: done loading kernel modules from /etc/modules.d/*
[  248.869240] kmodloader: loading kernel modules from /etc/modules.d/*
[  248.877710] kmodloader: done loading kernel modules from /etc/modules.d/*
[  249.793568] scsi 0:0:0:0: Direct-Access      USB      SanDisk 3.2Gen1 1.00 PQ: 0 ANSI: 6
[  249.804186] sd 0:0:0:0: [sda] 60088320 512-byte logical blocks: (30.8 GB/28.7 GiB)
[  249.812063] sd 0:0:0:0: [sda] Write Protect is off
[  249.816897] sd 0:0:0:0: [sda] Mode Sense: 43 00 00 00
[...]

PR #24729, opened 2026-08-14
[24.10] kernel: xhci: fix USB3 port events with a single roothub

GoetzGoerisch:
USB3 port events are dropped whenever the controller has only one roothub. The check that guards them treats a missing shared hcd as proof the hcd was removed, which stopped being true with upstream commit 4736ebd7fcaf ("usb: host: xhci-plat: omit shared hcd if either root hub has no ports"). A controller whose USB2 root hub has no ports gets a single roothub and never has a shared one, so every SuperSpeed event is thrown away and nothing enumerates even though the port reports the device:

  0x006a1203 Powered Connected Enabled Link:U0 PortSpeed:4

Broadcom Northstar is such a controller. USB3 works there up to 5.15 and stops from 5.19 onwards.

Tested on an Asus RT-N18U with kernel 6.18, seven SuperSpeed enumerations, USB2 unaffected on both ports.

Link: https://github.com/openwrt/openwrt/issues/19592
Link: https://github.com/openwrt/openwrt/pull/24547
(cherry picked from commit f6bc2a3b294b3d6b298cf5a625e9923459d64edd)

### DIFF
--- a/target/linux/generic/backport-6.6/891-v7.3-usb-xhci-handle-port-events-when-there-is-one-roothub.patch
+++ b/target/linux/generic/backport-6.6/891-v7.3-usb-xhci-handle-port-events-when-there-is-one-roothub.patch
@@ -1,6 +1,6 @@
-From eb8cfe27aef67b30299aaf2a1ac538cf905135f2 Mon Sep 17 00:00:00 2001
+From 3e91ec3e7d80a327fb558207613c80415d3bf756 Mon Sep 17 00:00:00 2001
 From: Semih Baskan <strst.gs@gmail.com>
-Date: Mon, 3 Aug 2026 07:08:11 +0300
+Date: Thu, 6 Aug 2026 17:21:12 +0300
 Subject: [PATCH] usb: xhci: Handle USB3 port events when there is one roothub
 
 handle_port_status() drops every USB3 port event when xhci->shared_hcd is
@@ -12,9 +12,11 @@ either root hub has no ports") that is no longer true. A controller whose
 USB2 root hub has no ports gets a single roothub, the USB3 rhub is served
 by the main hcd, and shared_hcd stays NULL for the lifetime of the device.
 Every SuperSpeed port event is then thrown away as bogus behind a debug
-message, so devices never enumerate even though the port reports them:
+message, so devices never enumerate even though the port sees the device
+and its change bits stay set:
 
   0x006a1203 Powered Connected Enabled Link:U0 PortSpeed:4
+  Change: CSC WRC PRC PLC
 
 Broadcom Northstar is such a controller. USB3 works there up to 5.15 and
 stops working from 5.19 onwards.
@@ -23,9 +25,21 @@ Ask xhci_get_usb3_hcd() instead. It returns the shared hcd when there is
 one, the main hcd when the USB2 root hub has no ports, and NULL once the
 shared hcd is gone, which keeps the original meaning of the check.
 
+Tested on an Asus RT-N18U (BCM47081), which has a single roothub. Before
+the change nothing enumerates on the USB3 port; after it SuperSpeed
+devices enumerate normally over repeated connect and disconnect cycles,
+the change bits shown above clear, and USB2 is unaffected on both ports.
+
 Fixes: 4736ebd7fcaf ("usb: host: xhci-plat: omit shared hcd if either root hub has no ports")
+Cc: stable@vger.kernel.org
 Signed-off-by: Semih Baskan <strst.gs@gmail.com>
+Signed-off-by: Mathias Nyman <mathias.nyman@linux.intel.com>
+Link: https://patch.msgid.link/20260806142113.2436238-17-mathias.nyman@linux.intel.com
+Signed-off-by: Greg Kroah-Hartman <gregkh@linuxfoundation.org>
 ---
+ drivers/usb/host/xhci-ring.c | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
 --- a/drivers/usb/host/xhci-ring.c
 +++ b/drivers/usb/host/xhci-ring.c
 @@ -2010,7 +2010,7 @@ static void handle_port_status(struct xh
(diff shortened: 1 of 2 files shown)


target/linux/generic/backport-6.6/891-v7.3-usb-xhci-handle-port-events-when-there-is-one-roothub.patch
openwrt-ai, line 1: The patch lands in `pending-6.12/`, but `openwrt-24.10` has no 6.12 kernel — the branch only carries 6.6 (`target/linux/generic/` on the base commit contains `backport-6.6`, `config-6.6`, `hack-6.6`, `pending-6.6` and nothing else; every `target/linux/*/Makefile` on this branch sets `KERNEL_PATCHVER := 6.6`, and there is no `KERNEL_TESTING_PATCHVER` anywhere). As committed, this file is dead weight: nothing will ever apply it, so the USB3 fix does not reach 24.10.

CI on this head confirms it — the "Build all affected Kernels / Set targets" job resolved the change to 6.12 and then found no target to build:

```
env:
  AFFECTED_KERNEL_VERSIONS: 6.12
---- targets to build ----
[]
---- targets to check patch ----
[]
```

That is also why the green checks here are not evidence the patch works; the kernel was never built with it.

The file should be `target/linux/generic/pending-6.6/891-usb-xhci-handle-port-events-when-there-is-one-roothub.patch`. The `891-` prefix is free on this branch — [`pending-6.6/` on the base commit](https://github.com/openwrt/openwrt/tree/a1ea57bd050c172fdc2b851824ebd9782aafb055/target/linux/generic/pending-6.6) tops out at `890-usb-serial-add-support-for-CH348.patch`.

Two things worth confirming once it is moved:

1. The hunk context is taken verbatim from the 6.12 tree — in v6.12.103 the `/* We might get interrupts after shared_hcd is removed */` guard sits at [`drivers/usb/host/xhci-ring.c:1988-1992`](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/drivers/usb/host/xhci-ring.c#L1988-L1992), matching the `@@ -1986,7 +1986,7 @@` header. Please verify it still applies cleanly against 6.6.151 and, if the surrounding context differs, regenerate the patch with `make target/linux/refresh` so the hunk header and context match.
2. `xhci_get_usb3_hcd()` was introduced alongside 4736ebd7fcaf (v5.19), so it should be present in 6.6, but the build after the move is the real check.

Everything else about the cherry-pick is clean: the `(cherry picked from commit f6bc2a3b294b3d6b298cf5a625e9923459d64edd)` trailer is present and points at the real commit on `main`, and the patch body is byte-identical to the upstream one. Note that the upstream commit added the patch twice, under `pending-6.12/` and `pending-6.18/`; for 24.10 only a single `pending-6.6/` copy is needed.

---

openwrt-ai (commented):
Reviewed 1 new commit. The cherry-pick metadata is correct and the patch content matches upstream verbatim, but the patch is filed under a kernel directory that does not exist on this branch — see the inline comment.

---

target/linux/generic/backport-6.6/891-v7.3-usb-xhci-handle-port-events-when-there-is-one-roothub.patch
GoetzGoerisch: fixed

openwrt-ai (commented):

The earlier finding is resolved: the patch now lands in `target/linux/generic/pending-6.6/`, `891-` is still free on this branch (`pending-6.6/` goes 890 -> 900), and the hunk was regenerated for the 6.6 tree (`@@ -2010,7 +2010,7 @@` with `}` as leading context instead of the 6.12/6.18 `vdev = xhci->devs[port->slot_id];`). Apart from that hunk header and its first context line, the file is byte-identical to both copies added by [f6bc2a3b](https://github.com/openwrt/openwrt/commit/f6bc2a3b294b3d6b298cf5a625e9923459d64edd) on `main`, and the `(cherry picked from commit ...)` trailer points at that commit. CI on this head now resolves the change to 6.6 and ran the kernel patch-check and build jobs across the affected targets with no failures, which is the real confirmation that the refreshed context applies.

---

target/linux/generic/backport-6.6/891-v7.3-usb-xhci-handle-port-events-when-there-is-one-roothub.patch
openwrt-ai, line 1: nit: `894-` is already taken in this directory by [`894-v6.8-net-ethtool-implement-ethtool_puts.patch`](``https://github.com/openwrt/openwrt/blob/54903d3a28252ab12b0e4bb5a1ec719ccba3f18e/target/linux/generic/backport-6.6/894-v6.8-net-ethtool-implement-ethtool_puts.patch),`` and every other repeated prefix in `backport-6.6/` is a numbered sub-series of one upstream patchset (`791-v6.11-01/02`, `793-01..03`, `798-v6.10-01..03`), so this would be the only unrelated pair sharing a number. Upstream on `main` the number was free in `backport-6.12`/`backport-6.18`; on this branch `881`–`893` are unused, so e.g. `893-v7.3-usb-xhci-handle-port-events-when-there-is-one-roothub.patch` avoids the clash. Was keeping the filename byte-identical to `main` the intent here?

---

openwrt-ai (commented):
## Commit checks

- `54903d3a` "kernel: 6.6: move xhci single-roothub fix to backports" — body still says "Move the patch from pending to backport in both kernels", but only the single 6.6 copy exists on this branch; the subject was adapted to `kernel: 6.6:`, the body should say 6.6 alone.

---

openwrt-ai (commented):

---

target/linux/generic/backport-6.6/891-v7.3-usb-xhci-handle-port-events-when-there-is-one-roothub.patch
openwrt-ai: fixed, thanks — solved by renaming to `891-` rather than `893-`, which is equally free in `backport-6.6/`, and the commit message now records the deviation from `main`.

---

strrs:
@GoetzGoerisch Thank you. Patch merged in: 6.18.49, 6.12.108, 6.6.156.

GoetzGoerisch:
Upstream in https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=d27599347b691f7b279949adc7652b7e0ba9bbb7
Comes with the kernel bump to `6.6.156`
Closing.

PR #24730, opened 2026-08-14
tools: zip: fix command injection in -T

polonec-dev:
Backport Debian's fix for command injection in Info-ZIP 3.0.

Info-ZIP builds a shell command with the archive name when -T tests an
archive. A crafted archive name can inject additional shell commands.

The patch quotes the archive name before it is added to the test command.
It was adjusted to apply after the existing OpenWrt zip patches.

Tested with the published proof of concept. The command injection creates
the "poc" file before the fix. It does not create the file after the fix.

Advisory:
https://sintonen.fi/advisories/infozip-test-option-command-injection.txt

Debian bug:
https://bugs.debian.org/1143866

### DIFF
--- a/tools/zip/patches/013-fix-command-injection.patch
+++ b/tools/zip/patches/013-fix-command-injection.patch
@@ -1,0 +1,238 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Paul Marquess <pmqs@outlook.com>
+Date: Fri, 14 Aug 2026 18:58:00 +0200
+Subject: [PATCH] Fix command injection issue
+
+Bug-Debian: https://bugs.debian.org/1143866
+X-Debian-version: 3.0-16
+
+[Peter Polonec: adjust patch context for OpenWrt]
+Signed-off-by: Peter Polonec <polonec.devel@pm.me>
+
+---
+--- a/zip.c
++++ b/zip.c
+@@ -122,6 +122,7 @@ ZCONST uLongf *crc_32_tab;
+ 
+ local void freeup  OF((void));
+ local int  finish  OF((int));
++local char *quote_arg(char *instring);
+ #if (!defined(MACOS) && !defined(WINDLL))
+ local void handler OF((int));
+ local void license OF((void));
+@@ -1323,6 +1324,134 @@ local int check_unzip_version(unzippath)
+   return 1;
+ }
+ 
++
++/* quote_arg()
++ *
++ * Add quotation and/or escapes to a shell (VMS: DCL) argument string
++ * appropriate to the local operating system or shell  (Unix, Windows,
++ * etc.).  This is mainly used to build the command line to pass to
++ * UnZip (or other application when -TT used) to test an archive.
++ * Return malloc()'d result.
++ *
++ *    All:     Add " at beginning and end.
++ *    MSDOS:   % -> "^%"
++ *             " -> \""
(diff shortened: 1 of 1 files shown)


robimarko:
Please just refresh the patch and force push

tools/zip/patches/013-fix-command-injection.patch
```diff
@@ -0,0 +1,238 @@
...
+Signed-off-by: Peter Polonec <polonec.devel@pm.me>
+
+---
+--- a/zip.c
++++ b/zip.c
+@@ -122,6 +122,7 @@ ZCONST uLongf *crc_32_tab;
+ 
+ local void freeup  OF((void));
+ local int  finish  OF((int));
++local char *quote_arg(char *instring);
```
openwrt-ai, line 19: This forward declaration is an unconditional ANSI prototype, but the definition it declares (added further down in this patch) carries a `#ifndef NO_PROTO` / K&R fallback pair — so on a `NO_PROTO` build the declaration is exactly the construct that guard exists to avoid, and it contradicts its own definition.

The two neighbouring declarations in this same hunk (`local void freeup  OF((void));`, `local int  finish  OF((int));`) wrap the argument list in `OF()` for that reason, and so does upstream Info-ZIP. Use the same form here:

```suggestion
+local char *quote_arg OF((char *instring));
```


---

tools/zip/patches/013-fix-command-injection.patch
```diff
@@ -0,0 +1,238 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Paul Marquess <pmqs@outlook.com>
+Date: Fri, 14 Aug 2026 18:58:00 +0200
```
openwrt-ai, line 3: nit: `From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001` is a fabricated commit hash, and the `Date:` is this PR's submission date (2026-08-14) rather than anything belonging to Paul Marquess. Together they make the header look like a real upstream git commit that doesn't exist.

The other Debian-derived patches in this directory — [007-fclose-in-file-not-fclose-x.patch:1-4](https://github.com/openwrt/openwrt/blob/25ac5f0688c5f9ccaa4a91fcb7d616651fb59c2a/tools/zip/patches/007-fclose-in-file-not-fclose-x.patch#L1-L4) and [008-hardening-build-fix-1.patch:1-4](https://github.com/openwrt/openwrt/blob/25ac5f0688c5f9ccaa4a91fcb7d616651fb59c2a/tools/zip/patches/008-hardening-build-fix-1.patch#L1-L4) — use a plain `From:` / `Subject:` / `Bug-Debian:` / `X-Debian-version:` header with no `git format-patch` line, which is the convention to follow here.

```suggestion
+From: Paul Marquess <pmqs@outlook.com>
```

Related: the bare `---` on line 12 is the `git format-patch` diffstat separator with nothing following it — it can go along with the header line.


---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the change, and the patch's hunk headers are internally consistent (line counts and cumulative offsets across all six hunks check out). The `quote_arg()` buffer sizing (`QA_FACTOR * len + QA_INCR + 1`) is exact for the worst case on each platform branch, and the two `malloc()` adjustments in `check_zipfile()` remain sufficient now that the surrounding `'`-quoting is gone. Two inline comments below.

---

robimarko:

PR #21078, opened 2025-12-07
generic, x86, packages: add support for kernel 6.18

graysky2:
Enable kernel 6.18 as the testing kernel for x86.

Build system: x86/64
Build-tested: x86/64-glibc
Build-tested with external toolchains:  x86/64, x86/generic, x86/legacy
Run-tested: x86/64-glibc

PR #24732, opened 2026-08-14
generic: make the threaded NAPI poll loop yield on every exit

VitaliySochniy:
On boards whose driver runs NAPI in a kernel thread, the poll loop can hold a CPU without ever yielding it. Everything that waits for deferred work on that CPU then blocks — in practice `ip link` and network reconfiguration hang for tens of seconds while the board is passing traffic at line rate.

`napi_threaded_poll_loop()` reaches `cond_resched()` only on the path where `repoll` is true. When the poll drains the queue within its budget the loop breaks first, `napi_thread_wait()` returns without scheduling because work has already been queued again, and the loop is re-entered. Under `CONFIG_PREEMPT_NONE` — which every affected target uses — that path has no preemption point at all.

Three different waits were seen blocking on it, all from `ip link del`:

| wait | why it blocks |
| :-- | :-- |
| `synchronize_net()` | the grace period never completes |
| `flush_all_backlogs()` -> `flush_work()` | the per-CPU worker never runs |
| `netdev_run_todo()` -> `rcu_barrier()` | callbacks never execute |

Which one appears varies by board. They share one cause, so the single `cond_resched()` move covers all three.

## Measurements

Nokia XG-040G-MF, 60 minutes of terminating `iperf3` into a DSA switch port, timing `ip link del` in a loop:

| | without | with |
| :-- | --: | --: |
| mean | 1.22 s | **0.10 s** |
| worst | **60.32 s** | **0.19 s** |
| over 1 s | 7 of 172 | **0 of 179** |
| RCU stalls | 2 | **0** |
| receive | 985 Mbit/s | 985 Mbit/s |
| packet rate | 82 094 p/s | 82 073 p/s |

The packet rate is the control: the same work is done in both runs, so this is not a lighter load. Every A/B pair recorded so far agrees — on XG-040G-MF from other testers (62.70 s -> 0.22 s and 64.90 s -> 0.25 s), here, and on XG-040G-MD where the worst case went from 241.55 s to 0.41 s.

Both halves of every pair come from one build session and record the hash of every kernel patch in the tree with and without the one under test, so each pair is verifiably one patch apart.

Without the patch the kernel names the thread itself. This run is on a build with the out-of-tree GPIO button module removed, so nothing taints it:

```
rcu: INFO: rcu_sched self-detected stall on CPU
rcu:  0-....: (5999 ticks this GP) ... (t=6000 jiffies g=913 q=1218 ncpus=4)
CPU: 0 UID: 0 PID: 203 Comm: napi/qdma_eth-0 Not tainted 6.18.44 #0
Hardware name: Nokia XG-040G-MD (UBI) (DT)
pc : __dma_sync_single_for_device+0x8/0xfc
```

That board has four cores. Three of them sat around 65% idle throughout while `ip link del` blocked for 151.92 s, because the deferred work it waits for is tied to the CPU the poll loop holds — spare CPUs do not help.

## Scope

This is narrower than "any threaded NAPI driver", and worth stating plainly:

* `airoha_eth` has no interrupt moderation and re-enters the loop over 12 000 times a second — it triggers reliably.
* `mtk_eth_soc` moderates through `net_dim` and stayed clean for an hour on MT7621 at the same packet rate, so MediaTek boards are not affected in practice.
* The load shape matters as well. The same board stays clean when the queue drains completely on each poll, and when packets arrive in batches large enough that the thread sleeps between them. On the MF the defect appears at roughly 4–7 packets per poll; four `iperf3` streams land inside that band, sixteen do not.

## Preemption models

`cond_resched()` only does anything where the kernel is not preemptible, so the model decides whether this patch matters at all. Since kernel **7.0** upstream has gated the non-preempting models:

```
kernel/Kconfig.preempt
  config PREEMPT_NONE       depends on ARCH_NO_PREEMPT
  config PREEMPT_VOLUNTARY  depends on !ARCH_HAS_PREEMPT_LAZY

kernel/sched/core.c, sched_dynamic_mode()
  "none" and "voluntary" are compiled out under !(PREEMPT_RT || ARCH_HAS_PREEMPT_LAZY)
```

`ARCH_NO_PREEMPT` is selected by m68k, hexagon and alpha; `ARCH_HAS_PREEMPT_LAZY` by x86, arm64, powerpc, s390, riscv and loongarch. On those six there is no way to build, boot or switch into a model where `cond_resched()` is live — `preempt=none` is rejected at boot and at runtime. What keeps `PREEMPT_VOLUNTARY` is arc, arm, csky, microblaze, mips, nios2, openrisc, parisc, sh, sparc and xtensa.

None of that applies to this PR as it stands: 6.18 has no such gating, and `KERNEL_PREEMPT_NONE` is OpenWrt's default on every target.

It matters at the next bump, so the same A/B was run under `PREEMPT_VOLUNTARY` as well — the model the affected architectures keep on 7.x. Same board, untainted, 30 minutes per half:

| | baseline | patched |
| :-- | --: | --: |
| mean `ip link del` | 13.38 s | **0.19 s** |
| worst | **135.95 s** | **0.51 s** |
| over 1 s | 10 of 55 | **0 of 89** |
| RCU stalls | 32 | **0** |
| packet rate | 80 312 p/s | 81 532 p/s |

So the defect is not specific to `PREEMPT_NONE`, and on targets without `PREEMPT_LAZY` no preemption model avoids it short of full `PREEMPT`.

Worth flagging separately, unrelated to this patch: on 7.x [`default KERNEL_PREEMPT_NONE`](https://github.com/openwrt/openwrt/blob/aa66786f38505c80daaec59ed922d420261467fe/config/Config-kernel.in#L1656-L1658) stops being selectable for every target except m68k/hexagon/alpha. kconfig will settle on `PREEMPT_LAZY` for arm64 and x86, and on `PREEMPT_VOLUNTARY` for mips and arm.

## Upstream

Sent to netdev as a fix to `net/core/dev.c`, with `Fixes: 29863d41bb6e ("net: implement threaded-able napi poll loop support")`. **It was not applied**, and the maintainer asked for no further versions — `cond_resched()` is a nop on the architectures netdev cares about, so there is nothing there for the change to fix.

The version sent there is shaped differently, because the loop has since gained a busy-poll parameter and the tail was reordered around it. Both forms make the same change — yield on both exits instead of only on the iterating one — and both were run on this board for an hour against the same baseline:

| | baseline | this patch | upstream form |
| :-- | --: | --: | --: |
| worst `del` | 60.32 s | 0.19 s | 0.16 s |
| over 1 s | 7 of 172 | 0 of 179 | 0 of 179 |
| RCU stalls | 2 | 0 | 0 |
| packet rate | 82 094 p/s | 82 073 p/s | 82 027 p/s |

The form kept here is the one with the hardware behind it — every A/B pair above used it.

This patch is `pending-6.18`. It should not be dropped waiting for an upstream landing that will not come; the point to revisit it is the next kernel bump, where arm64 and x86 targets settle on `PREEMPT_LAZY` and no longer need it, while mips and arm settle on `PREEMPT_VOLUNTARY` and still do. Neither form applies to a kernel that already carries the reordered loop, so the bump is where it has to be re-made or removed anyway.

### DIFF
--- a/target/linux/generic/pending-6.18/684-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
+++ b/target/linux/generic/pending-6.18/684-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
@@ -1,0 +1,74 @@
+From 72bc090b453fbc425a2f107eb22b24947fa56042 Mon Sep 17 00:00:00 2001
+From: Vitaliy Sochnev <sochnev.v.74@gmail.com>
+Date: Fri, 14 Aug 2026 23:04:27 +0100
+Subject: [PATCH] net: yield the CPU on every exit of the threaded NAPI poll loop
+
+napi_threaded_poll_loop() only reaches cond_resched() when it is about to
+iterate again. When __napi_poll() clears repoll the loop breaks first, so
+the last iteration returns without any voluntary preemption point.
+
+Control then goes back to napi_threaded_poll(), which calls
+napi_thread_wait(). If work is already pending that helper returns without
+ever calling schedule(). Under a receive load that keeps arriving at least
+as fast as it is drained, this repeats indefinitely and the kthread holds
+its CPU without a single reschedule. On CONFIG_PREEMPT_NONE - which every
+affected OpenWrt target uses - nothing else on that CPU gets to run.
+
+Everything that waits for deferred work on that CPU then blocks: RCU grace
+periods, per-CPU work items and RCU callbacks. Deleting a network device
+hits all three - synchronize_net(), flush_all_backlogs() -> flush_work()
+and rcu_barrier() from netdev_run_todo(). In practice this shows up as
+"ip link" and network reconfiguration hanging for tens of seconds while
+the board forwards or terminates traffic at line rate.
+
+rcu_softirq_qs_periodic() does not help here. It reports a quiescent state
+but does not schedule, so the work items and the callbacks still wait, and
+it is skipped on the exit path anyway.
+
+Measured on a Nokia XG-040G-MF (Airoha AN7583) terminating a 985 Mbit/s
+TCP receive load, 60 minute runs, timing "ip link del":
+
+                    before        after
+  mean               1.22 s       0.10 s
+  worst             60.32 s       0.19 s
+  over 1 s        7 of 172      0 of 179
+  RCU stalls             2            0
+  packet rate   82094 p/s    82073 p/s
+
+The packet rate is the control: the same work is done in both runs. A
+separate pair of 60 minute runs on an AN7581 board took the worst case
+from 241.55 s to 0.41 s.
+
+Reproducing it needs the loop to be re-entered tens of thousands of times
+a second, so it depends on the driver and on the shape of the load. It
+did not reproduce on mtk_eth_soc, whose net_dim moderation coalesces the
+same packet rate into far fewer interrupts.
+
+This is against the 6.18 code, where the loop breaks before reaching
+rcu_softirq_qs_periodic() and cond_resched(). Upstream has since gained a
+busy-poll parameter and reordered the tail, so the version sent to netdev
+looks different while making the same change - the yield happens on both
+exits instead of only on the iterating one. That version was run on the
+same board for an hour against the same baseline and came out equally
+clean (worst 0.16 s). Drop this patch once it lands in net.
+
+Link: https://lore.kernel.org/netdev/20260814220427.623427-1-sochnev.v.74@gmail.com/
+Fixes: 29863d41bb6e ("net: implement threaded-able napi poll loop support")
+Signed-off-by: Vitaliy Sochnev <sochnev.v.74@gmail.com>
+
+--- a/net/core/dev.c
++++ b/net/core/dev.c
+@@ -7775,11 +7775,12 @@ static void napi_threaded_poll_loop(stru
+ 		bpf_net_ctx_clear(bpf_net_ctx);
+ 		local_bh_enable();
+ 
++		cond_resched();
++
+ 		if (!repoll)
+ 			break;
+ 
+ 		rcu_softirq_qs_periodic(last_qs);
+-		cond_resched();
+ 	}
+ }
+ 
--- a/target/linux/generic/pending-6.18/684-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
+++ b/target/linux/generic/pending-6.18/684-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
@@ -1,0 +1,75 @@
+From 72bc090b453fbc425a2f107eb22b24947fa56042 Mon Sep 17 00:00:00 2001
+From: Vitaliy Sochnev <sochnev.v.74@gmail.com>
+Date: Fri, 14 Aug 2026 23:04:27 +0100
+Subject: [PATCH] net: yield the CPU on every exit of the threaded NAPI poll loop
+
+napi_threaded_poll_loop() only reaches cond_resched() when it is about to
+iterate again. When __napi_poll() clears repoll the loop breaks first, so
+the last iteration returns without any voluntary preemption point.
+
+Control then goes back to napi_threaded_poll(), which calls
+napi_thread_wait(). If work is already pending that helper returns without
+ever calling schedule(). Under a receive load that keeps arriving at least
+as fast as it is drained, this repeats indefinitely and the kthread holds
+its CPU without a single reschedule. On CONFIG_PREEMPT_NONE - which every
+affected OpenWrt target uses - nothing else on that CPU gets to run.
+
+Everything that waits for deferred work on that CPU then blocks: RCU grace
+periods, per-CPU work items and RCU callbacks. Deleting a network device
+hits all three - synchronize_net(), flush_all_backlogs() -> flush_work()
+and rcu_barrier() from netdev_run_todo(). In practice this shows up as
+"ip link" and network reconfiguration hanging for tens of seconds while
+the board forwards or terminates traffic at line rate.
+
+rcu_softirq_qs_periodic() does not help here. It reports a quiescent state
+but does not schedule, so the work items and the callbacks still wait, and
+it is skipped on the exit path anyway.
+
+Measured on a Nokia XG-040G-MF (Airoha AN7583) terminating a 985 Mbit/s
+TCP receive load, 60 minute runs, timing "ip link del":
+
+                    before        after
+  mean               1.22 s       0.10 s
+  worst             60.32 s       0.19 s
+  over 1 s        7 of 172      0 of 179
+  RCU stalls             2            0
+  packet rate   82094 p/s    82073 p/s
+
+The packet rate is the control: the same work is done in both runs. A
+separate pair of 60 minute runs on an AN7581 board took the worst case
+from 241.55 s to 0.41 s.
+
+Reproducing it needs the loop to be re-entered tens of thousands of times
+a second, so it depends on the driver and on the shape of the load. It
+did not reproduce on mtk_eth_soc, whose net_dim moderation coalesces the
+same packet rate into far fewer interrupts.
+
+This is against the 6.18 code, where the loop breaks before reaching
+rcu_softirq_qs_periodic() and cond_resched(). Upstream has since gained a
+busy-poll parameter and reordered the tail, so the version sent to netdev
+looks different while making the same change - the yield happens on both
+exits instead of only on the iterating one. That version was run on the
+same board for an hour against the same baseline and came out equally
+clean (worst 0.16 s), but was not applied. Revisit this patch at the next
+kernel bump rather than waiting for it to land in net.
+
+Link: https://lore.kernel.org/netdev/20260902210053.263070-1-sochnev.v.74@gmail.com/
+Fixes: 29863d41bb6e ("net: implement threaded-able napi poll loop support")
+Signed-off-by: Vitaliy Sochnev <sochnev.v.74@gmail.com>
+
+--- a/net/core/dev.c
++++ b/net/core/dev.c
+@@ -7775,11 +7775,12 @@ static void napi_threaded_poll_loop(stru
+ 		bpf_net_ctx_clear(bpf_net_ctx);
+ 		local_bh_enable();
+ 
++		cond_resched();
++
+ 		if (!repoll)
+ 			break;
+ 
+ 		rcu_softirq_qs_periodic(last_qs);
+-		cond_resched();
+ 	}
+ }
+ 


target/linux/generic/pending-6.18/684-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
openwrt-ai, line 1: nit: the patch number puts a `net/core/dev.c` change in the driver range. [`PATCHES.md`'s pending numbering table](https://github.com/openwrt/openwrt/blob/25b117b863c66ca7d7ea729bd88f85d1d0dc88a4/target/linux/generic/PATCHES.md#L58-L59) maps `6xx` to "Generic network" and `7xx` to "Network / PHY", and that split holds in `pending-6.18`: every `7xx` patch touches a driver or a DSA/PCS file, while the closest precedent for a `net/core/` change is [`683-of_net-add-mac-address-to-of-tree.patch`](``https://github.com/openwrt/openwrt/blob/25b117b863c66ca7d7ea729bd88f85d1d0dc88a4/target/linux/generic/pending-6.18/683-of_net-add-mac-address-to-of-tree.patch)`` in `6xx`. Nothing is used between 683 and 700, so e.g. `684-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch` would fit — and it would also stop this patch from landing in the middle of the `760-xx` mxl862xx series.

---

target/linux/generic/pending-6.18/767-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
```diff
@@ -0,0 +1,74 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
```
openwrt-ai: nit: all-zero placeholder SHA. Of the 188 patches in `pending-6.18`, 111 carry a real commit SHA on this line and 75 omit the line entirely and start at `From:` — only one uses zeros. Your own [`airoha/patches-6.18/921-net-airoha-npu-load-firmware-without-the-sysfs-fallb.patch`](``https://github.com/openwrt/openwrt/blob/25b117b863c66ca7d7ea729bd88f85d1d0dc88a4/target/linux/airoha/patches-6.18/921-net-airoha-npu-load-firmware-without-the-sysfs-fallb.patch#L1)`` carries the real one. Harmless to quilt, but either keep the real SHA of your local commit or drop the line and let the header start at `From:` (line 2).

---

target/linux/generic/pending-6.18/684-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
```diff
@@ -0,0 +1,74 @@
...
+ 		bpf_net_ctx_clear(bpf_net_ctx);
+ 		local_bh_enable();
+ 
++		cond_resched();
++
+ 		if (!repoll)
+ 			break;
+ 
+ 		rcu_softirq_qs_periodic(last_qs);
+-		cond_resched();
```
openwrt-ai, line 72: The hunk does two things, and the commit message only justifies one of them. Besides adding the missing yield on the `break` path, it also swaps the two calls on the repoll path: [upstream runs `rcu_softirq_qs_periodic(last_qs)` and then `cond_resched()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/net/core/dev.c#L7778-L7782), after this patch the iterating path runs `cond_resched()` and then the QS report.

Moving the existing two lines up instead keeps upstream's ordering, is the same size, and yields on both exits just the same:

```
 		local_bh_enable();
 
 		rcu_softirq_qs_periodic(last_qs);
 		cond_resched();
 
 		if (!repoll)
 			break;
 	}
```

That form additionally reports the quiescent state on the exit path. The commit message notes `rcu_softirq_qs_periodic()` "is skipped on the exit path anyway", but it is rate-limited to `HZ/10`, so running it there costs essentially nothing and removes one more difference from the version you sent to netdev. Was the swap deliberate, or a side effect of placing the new `cond_resched()` above the `break`?

---

target/linux/generic/pending-6.18/684-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
```diff
@@ -0,0 +1,74 @@
...
+same board for an hour against the same baseline and came out equally
+clean (worst 0.16 s). Drop this patch once it lands in net.
+
+Link: https://lore.kernel.org/netdev/20260814220427.623427-1-sochnev.v.74@gmail.com/
+Fixes: 29863d41bb6e ("net: implement threaded-able napi poll loop support")
+Signed-off-by: Vitaliy Sochnev <sochnev.v.74@gmail.com>
+
+--- a/net/core/dev.c
++++ b/net/core/dev.c
+@@ -7775,11 +7775,12 @@ static void napi_threaded_poll_loop(stru
```
openwrt-ai, line 62: 6.12 carries the identical defect — [`napi_threaded_poll_loop()` breaks before ever reaching `cond_resched()` there too](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/net/core/dev.c#L7030-L7034) — and `pending-6.12` still ships `702-net-ethernet-mtk_eth_soc-enable-threaded-NAPI.patch`. Nothing on `KERNEL_PATCHVER:=6.12` builds a threaded-NAPI driver by default today (airoha, ramips, mediatek, ipq40xx and qualcommax are all 6.18), so restricting this to 6.18 is defensible, but neither the PR body nor the commit message says why. Is the 6.12 omission intentional? If so a one-liner in the commit message would save the next reader the check. Note it cannot be the same file verbatim: 6.12 has `skb_defer_free_flush(sd);` in the context, so it would need its own copy.

---

openwrt-ai (commented):
Reviewed 1 commit (`25b117b8`). The commit message matches the change, and the analysis holds up against the 6.18 source: [`napi_thread_wait()` returns 0 without ever calling `schedule()` when `NAPI_STATE_SCHED_THREADED` is already set](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/net/core/dev.c#L7723-L7745), and `airoha_eth` does set `napi_dev->threaded = true` unconditionally, so the described re-entry loop with no preemption point is real. The patch also applies to v6.18.44 with `--fuzz=0`, and CI is green on the head commit.

Four inline notes, none blocking: one on the `rcu_softirq_qs_periodic()` / `cond_resched()` reordering that comes along with the fix, one question about 6.12, and two nits (patch number range, placeholder SHA in the header).

---

rsalvaterra:
I'm… not really happy with this, regardless if it's accepted upstream or not. Sprinkling `cond_resched()` to deal with CPU hogging in a `PREEMPT_NONE` kernel is a game of whack-a-mole. Since Linux 6.13, we have `PREEMPT_LAZY`, which should be the default, where available, as mentioned here: https://github.com/openwrt/openwrt/pull/21078#issuecomment-3691425611. Could you please retest as such?

rsalvaterra:
> I'm… not really happy with this, regardless if it's accepted upstream or not. Sprinkling `cond_resched()` to deal with CPU hogging in a `PREEMPT_NONE` kernel is a game of whack-a-mole. Since Linux 6.13, we have `PREEMPT_LAZY`, which should be the default, where available, as mentioned here: [#21078 (comment)](https://github.com/openwrt/openwrt/pull/21078#issuecomment-3691425611). Could you please retest as such?

Unforunately, MT7621 is MIPS, and `CONFIG_PREEMPT_LAZY` isn't available there yet, but the point stands.

VitaliySochniy:
Retested on Nokia XG-040G-MD (Airoha AN7581, quad A53). Both images are plain `main` without the patch, only the preemption model differs — verified in the built kernel `.config`, and `#0 SMP` vs `#0 SMP PREEMPT` at runtime. Same load, same script.

| | `PREEMPT_NONE` (30 min) | `PREEMPT_LAZY` (60 min) |
| :-- | --: | --: |
| worst `ip link del` | 248.48 s | 0.39 s |
| samples > 1 s | 8 of 29 | 0 of 177 |
| RCU stalls | 8 classic + 48 expedited | 0 |
| packet rate | 80 862 p/s | 80 036 p/s |
| receive | 970 Mbit/s | 960 Mbit/s |

A second 10 min `PREEMPT_NONE` run reproduced as well (153.91 s worst, 4+43 stalls). So you're right: LAZY prevents it here.

Interestingly not by preempting the NAPI kthread — `nonvoluntary_ctxt_switches` on it is 4.6/s under LAZY vs 13.9/s under `PREEMPT_NONE`. What does change is the dynamics: 211k vs 127k interrupts/s, 30.5 vs 6.7 packets per poll-loop entry. I can't pin the mechanism from this pair.

Still, every target is built `PREEMPT_NONE` today and `config/Config-kernel.in` has no LAZY option at all — I had to add one locally just to run this test. Adding that option and evaluating a change of default is a separate piece of work with its own trade-offs, quite apart from this fix: the PR here only moves one statement above the loop's exit check so that both exits reach it, and fixes the configuration OpenWrt actually ships right now.

On "sprinkling": the loop already calls `cond_resched()`, just not on one of its two exits. This removes the asymmetry rather than adding a new call site, so both go together whenever those calls get audited away.

rsalvaterra:
> Still, every target is built `PREEMPT_NONE` today and `config/Config-kernel.in` has no LAZY option at all

Indeed, I'm well aware of that and I should have mentioned it. It would be really nice to add it, but only when 6.12 support is removed in main, as 6.12 doesn't have it.

target/linux/generic/pending-6.18/684-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
VitaliySochniy: Renumbered to `684-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch`

target/linux/generic/pending-6.18/767-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
VitaliySochniy: Dropped, the header now starts at `From:`. There is no real SHA to put there: the commit sent to netdev carries a different diff, because the loop was reshaped upstream after 6.18

target/linux/generic/pending-6.18/684-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
VitaliySochniy: Side effect originally, kept deliberately: this is the shape with the hardware data behind it — four A/B pairs on two boards.

Your form doesn't actually reduce the divergence from netdev, though. That version keeps `rcu_softirq_qs_periodic()` under `if (repoll || busy_poll_last_qs)` and lifts only `cond_resched()` out, so reporting the QS on the exit path would be a new difference rather than one less.

I did test the upstream-shaped variant on 6.18 separately — `if (repoll) rcu_softirq_qs_periodic(last_qs); cond_resched(); if (!repoll) break;` — worst `ip link del` 0.16 s against 0.19 s for the form here, against the same baseline. So if you'd prefer this copy to match upstream exactly, I'll swap to that one. I'd rather not ship a third shape that nothing has run

target/linux/generic/pending-6.18/684-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
VitaliySochniy: Intentional, and for exactly the reason you give — nothing on 6.12 builds a threaded-NAPI driver by default. Now stated in the commit message. Agreed it would need its own copy of the patch if that ever changes

target/linux/generic/pending-6.18/684-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
```diff
@@ -0,0 +1,74 @@
...
+The packet rate is the control: the same work is done in both runs. A
+separate pair of 60 minute runs on an AN7581 board took the worst case
+from 241.55 s to 0.41 s.
+
+Reproducing it needs the loop to be re-entered tens of thousands of times
+a second, so it depends on the driver and on the shape of the load. It
+did not reproduce on mtk_eth_soc, whose net_dim moderation coalesces the
+same packet rate into far fewer interrupts.
+
+This is against the 6.18 code, where the loop breaks before reaching
```
openwrt-ai, line 47: Anchoring here because it is the 6.18-scope paragraph, but this is about the new 6.12 paragraph in the OpenWrt commit message:

> no target on KERNEL_PATCHVER:=6.12 builds a driver with threaded NAPI enabled by default - pending-6.12's 702-net-ethernet-mtk_eth_soc-enable-threaded-NAPI.patch covers mtk_eth_soc, and every target using that driver is on 6.18

The `mtk_eth_soc` half checks out (mediatek and ramips are both `KERNEL_PATCHVER:=6.18`), but it is not the only threaded-NAPI user that reaches a 6.12 target. mt76 turns it on unconditionally in `mt76_dma_init()` — [`dev->napi_dev->threaded = 1;` at mt76/dma.c:966 (v6.12.103)](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/drivers/net/wireless/mediatek/mt76/dma.c#L966), and the same line is still there in v6.18.44 — and lantiq is on 6.12 and ships mt76 in `DEVICE_PACKAGES` by default on e.g. [`buffalo_wbmr-300hpd` (`kmod-mt7603`)](https://github.com/openwrt/openwrt/blob/d8d6165791bb43024c4e4023b00b0e08e4556956/target/linux/lantiq/image/vr9.mk#L331) and [`tplink_vr200` / `tplink_vr200v` (`kmod-mt76x0e`)](https://github.com/openwrt/openwrt/blob/d8d6165791bb43024c4e4023b00b0e08e4556956/target/linux/lantiq/image/tp-link.mk#L29). OpenWrt builds mt76 out of tree from `openwrt/mt76` rather than from the kernel copy, so I could not check the pinned `5967691` directly — if that revision dropped the line, ignore this.

The conclusion is probably still fine: your own exposure argument (12000 re-entries/s on `airoha_eth`, nothing on `mtk_eth_soc` with net_dim) makes a wifi driver on an xRX200 an unlikely reproducer. It is the stated reason that is too strong. Would you reword it to the exposure argument — no 6.12 target runs a threaded-NAPI driver at a re-entry rate that reproduces this — instead of claiming none builds one at all?

---

openwrt-ai (commented):
Reviewed 1 new commit (`d8d6165`), which replaces `25b117b8` and addresses both nits from the previous round: the patch is renumbered `767-` → `684-` (684 is free, follows `683-of_net-add-mac-address-to-of-tree.patch`, and no longer lands inside the `760-xx` mxl862xx series), and the all-zero placeholder SHA is replaced with a real one. The patch payload itself is unchanged and still applies to v6.18.44 with `--fuzz=0`; no CI checks have reported on the head commit yet.

The `cond_resched()` / `rcu_softirq_qs_periodic()` ordering question from last round is answered — keeping the shape that the four A/B pairs actually ran is a reasonable call, and I have nothing to add to it.

## Commit checks
- `d8d6165` "generic: make the threaded NAPI poll loop yield on every exit" — the newly added 6.12 paragraph asserts that no `KERNEL_PATCHVER:=6.12` target builds a threaded-NAPI driver by default. mt76 sets `napi_dev->threaded = 1` unconditionally and lantiq (6.12) ships it by default on three devices, so the claim as written does not hold, even though the practical conclusion likely does. Detail inline.

---

openwrt-ai (commented):
Reviewed 1 new commit (`823f816`, replacing `d8d6165`). The patch file is byte-identical between the two — `git diff d8d6165..823f816` is empty — so the only change is the commit message, and it resolves the one finding from last round.

The reworked 6.12 paragraph now rests on exposure rather than absence, and each of its claims holds:

- *"6.12 would need its own copy, as the context differs there"* — confirmed: v6.12.103 has [`skb_defer_free_flush(sd);` in the context](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/net/core/dev.c#L7026) where [v6.18.44 has `skb_defer_free_flush();`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/net/core/dev.c#L7774), so the hunk cannot be shared.
- *"mtk_eth_soc gets threaded NAPI from an OpenWrt patch and every target using it is on 6.18"* — `702-net-ethernet-mtk_eth_soc-enable-threaded-NAPI.patch` is present in both `pending-6.12` and `pending-6.18`, and the two targets that build the driver (`mediatek`, `ramips`) are both `KERNEL_PATCHVER:=6.18`.
- *"mt76 enables it unconditionally and does reach 6.12 targets — lantiq ships `kmod-mt76*` by default on a few devices"* — matches what I found last round, and correctly stops short of claiming no 6.12 target builds a threaded-NAPI driver at all.

No new inline issues. CI has not reported yet on the head commit (status `pending`, no checks recorded), so nothing to ground there.

---

VitaliySochniy:
@rsalvaterra — upstream has since closed off the non-preempting models, which changes what this patch is for.

Since v7.0:

```
kernel/Kconfig.preempt
  config PREEMPT_NONE       depends on ARCH_NO_PREEMPT
  config PREEMPT_VOLUNTARY  depends on !ARCH_HAS_PREEMPT_LAZY

kernel/sched/core.c, sched_dynamic_mode()
  "none" and "voluntary" are compiled out under !(PREEMPT_RT || ARCH_HAS_PREEMPT_LAZY)
```

`ARCH_NO_PREEMPT` is selected by m68k, hexagon and alpha; `ARCH_HAS_PREEMPT_LAZY` by x86, arm64, powerpc, s390, riscv and loongarch. On those six there is no way to build, boot or switch into a model where `cond_resched()` is anything but a nop — `preempt=none` is rejected at boot and at runtime. On an x86 guest built from `net/main`, `/sys/kernel/debug/sched/preempt` offers `full (lazy)`. So this patch is a no-op there, and your LAZY point is now enforced upstream rather than a preference.

MIPS is the case that survives. It still does not select `ARCH_HAS_PREEMPT_LAZY`, and on 7.x it loses `PREEMPT_NONE` as well, leaving `PREEMPT_VOLUNTARY` — as for arc, arm, csky, microblaze, nios2, openrisc, parisc, sh, sparc and xtensa.

Every number in the commit message came from `PREEMPT_NONE`, so I measured `VOLUNTARY` too. Same board, 6.18 where it is still selectable, untainted, no out-of-tree module, 30 min per half, halves differing only in this patch:

| | baseline | patched |
| --- | ---: | ---: |
| mean `ip link del` | 13.38 s | 0.19 s |
| worst | 135.95 s | 0.51 s |
| over 1 s | 10 of 55 | 0 of 89 |
| RCU stalls, classic | 12 | 0 |
| RCU stalls, expedited | 20 | 0 |
| receive | 964 Mbit/s | 978 Mbit/s |
| packet rate | 80312 p/s | 81532 p/s |

`add`, `mtu` and a no-op netlink call stayed between 0.01 and 0.41 s in both halves, so the delay is the grace period, not rtnl or scheduling. The packet rate is 1.5% higher in the patched half. The splat states the configuration:

```
rcu: INFO: rcu_sched self-detected stall on CPU
CPU: 0 PID: 200 Comm: napi/qdma_eth-0 Not tainted 6.18.44 #0 VOLUNTARY
```

The `PREEMPT_NONE` pair gave 151.92 s against 1.00 s. `VOLUNTARY` re-enters the loop a third less often — 8163 polls/s against 11545, 9.8 packets per poll against 6.9 — at the same packet rate and with the same result, so it is not specific to one shape of load. On targets without `PREEMPT_LAZY` there is no preemption model that avoids this short of full `PREEMPT`.

Nothing changes for this PR on 6.18: `KERNEL_PREEMPT_NONE` is still the default and still selectable, so the patch does what the commit message says. The branch is untouched.

Two separate points:

- It is not going into net. The maintainer asked for no further versions, on the basis above: there is no configuration of the tree it targets in which the change does anything. So the commit message here needs its "drop this patch once it lands in net" replaced with a revisit-at-the-next-bump note, which I will push if you want the patch kept.
- On 7.x [`default KERNEL_PREEMPT_NONE`](https://github.com/openwrt/openwrt/blob/aa66786f38505c80daaec59ed922d420261467fe/config/Config-kernel.in#L1656-L1658) stops being selectable for every target except m68k/hexagon/alpha. kconfig will settle on `PREEMPT_LAZY` for arm64 and x86, and on `PREEMPT_VOLUNTARY` for mips and arm.

target/linux/generic/pending-6.18/684-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
```diff
@@ -0,0 +1,74 @@
...
+rcu_softirq_qs_periodic() and cond_resched(). Upstream has since gained a
+busy-poll parameter and reordered the tail, so the version sent to netdev
+looks different while making the same change - the yield happens on both
+exits instead of only on the iterating one. That version was run on the
+same board for an hour against the same baseline and came out equally
+clean (worst 0.16 s). Drop this patch once it lands in net.
```
openwrt-ai: This contradicts the new OpenWrt commit message ("This will not go upstream... Revisit at the next kernel bump rather than dropping it"), and the patch header is the copy that survives in the tree, so it is the one a future maintainer will act on. The PR body's `## Upstream` section still says the same thing ("should be dropped once the fix lands in `net`") and needs the same update.

```suggestion
clean (worst 0.16 s). That version was not applied; revisit this patch at
the next kernel bump rather than waiting for it to land in net.
```

---

openwrt-ai (commented):
## Commit checks
- `446fbdc` "generic: make the threaded NAPI poll loop yield on every exit" — "there is no configuration of mainline where the change does anything" is contradicted two sentences later by "mips and arm settle on `PREEMPT_VOLUNTARY` and still do". Scope that claim to arm64/x86, or attribute it as the netdev maintainer's basis for declining.

---

target/linux/generic/pending-6.18/767-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
openwrt-ai: fixed, thanks

---

target/linux/generic/pending-6.18/684-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
openwrt-ai: solved by dropping the claim entirely — the message now just says the context differs on 6.12, which holds. Thanks

---

openwrt-ai (commented):

---

target/linux/generic/pending-6.18/684-net-yield-on-both-exits-of-threaded-NAPI-poll-loop.patch
openwrt-ai: fixed, thanks — patch header and the PR body's `## Upstream` section both updated.

---

willmo:
Re: upstream:

> It was not applied, and the maintainer asked for no further versions — cond_resched() is a nop on the architectures netdev cares about, so there is nothing there for the change to fix.

Is the upstream discussion just https://lore.kernel.org/netdev/20260818190420.21204-1-sochnev.v.74@gmail.com/T/ ? I don't see any explicit mention of the impacts on different architectures...perhaps the maintainers would feel differently if it was clarified that this should still be relevant on some architectures on mainline (although it would be best to also have numbers showing that)?

Also, does the Airoha hardware support interrupt moderation?

VitaliySochniy:
> Is the upstream discussion just https://lore.kernel.org/netdev/20260818190420.21204-1-sochnev.v.74@gmail.com/T/ ?

@willmo, that is v1. v2 is a separate thread, as netdev requires:

https://lore.kernel.org/netdev/20260902210053.263070-1-sochnev.v.74@gmail.com/

> perhaps the maintainers would feel differently if it was clarified that this should still be relevant on some architectures on mainline (although it would be best to also have numbers showing that)

That is what the last message in that thread does:

https://lore.kernel.org/netdev/20260903231253.292576-1-sochnev.v.74@gmail.com/

It names the architectures that keep a non-preempting model on 7.x (`PREEMPT_VOLUNTARY` on arc, arm, csky, microblaze, mips, nios2, openrisc, parisc, sh, sparc, xtensa; `PREEMPT_NONE` on alpha, hexagon, m68k) and carries the `PREEMPT_VOLUNTARY` A/B for exactly that reason: worst `ip link del` 135.95 s against 0.51 s, 12 classic and 20 expedited RCU stalls against none, at the same packet rate. The answer to it was "No, do not send v3."

> Also, does the Airoha hardware support interrupt moderation?

The hardware does, the driver switches it off. There is a per-RX-ring `REG_RX_DELAY_INT_IDX` with a 16-bit `RX_DELAY_INT_MASK`, and `airoha_eth` clears it for every ring during init:

```c
/* disable qdma rx delay interrupt */
for (i = 0; i < ARRAY_SIZE(qdma->q_rx); i++) {
	if (!qdma->q_rx[i].ndesc)
		continue;

	airoha_qdma_clear(qdma, REG_RX_DELAY_INT_IDX(i),
			  RX_DELAY_INT_MASK);
}
```

`airoha_ethtool_ops` has no `get_coalesce`/`set_coalesce`, so it cannot be turned back on from userspace either. That is the difference from `mtk_eth_soc`, which drives the same kind of knob through `net_dim` and does not reproduce this.

willmo:
> > Is the upstream discussion just https://lore.kernel.org/netdev/20260818190420.21204-1-sochnev.v.74@gmail.com/T/ ?
> 
> @willmo, that is v1. v2 is a separate thread, as netdev requires:
> 
> https://lore.kernel.org/netdev/20260902210053.263070-1-sochnev.v.74@gmail.com/

I see, thank you.
 
> > perhaps the maintainers would feel differently if it was clarified that this should still be relevant on some architectures on mainline (although it would be best to also have numbers showing that)
> 
> That is what the last message in that thread does:
> 
> https://lore.kernel.org/netdev/20260903231253.292576-1-sochnev.v.74@gmail.com/
> 
> It names the architectures that keep a non-preempting model on 7.x (`PREEMPT_VOLUNTARY` on arc, arm, csky, microblaze, mips, nios2, openrisc, parisc, sh, sparc, xtensa; `PREEMPT_NONE` on alpha, hexagon, m68k) and carries the `PREEMPT_VOLUNTARY` A/B for exactly that reason: worst `ip link del` 135.95 s against 0.51 s, 12 classic and 20 expedited RCU stalls against none, at the same packet rate. The answer to it was "No, do not send v3."

Well, that repro was on 6.18, and the next paragraph says it didn't repro on `net/main` with `veth` on x86 (forcibly selecting voluntary preemption), right? So maybe that muddied the message a bit?

In any case, if it turns out that this patch would be controversial even here downstream in OpenWrt, perhaps it makes sense to focus on this next question first:

> > Also, does the Airoha hardware support interrupt moderation?
> 
> The hardware does, the driver switches it off. There is a per-RX-ring `REG_RX_DELAY_INT_IDX` with a 16-bit `RX_DELAY_INT_MASK`, and `airoha_eth` clears it for every ring during init:
> 
> ```c
> /* disable qdma rx delay interrupt */
> for (i = 0; i < ARRAY_SIZE(qdma->q_rx); i++) {
> 	if (!qdma->q_rx[i].ndesc)
> 		continue;
> 
> 	airoha_qdma_clear(qdma, REG_RX_DELAY_INT_IDX(i),
> 			  RX_DELAY_INT_MASK);
> }
> ```
> 
> `airoha_ethtool_ops` has no `get_coalesce`/`set_coalesce`, so it cannot be turned back on from userspace either. That is the difference from `mtk_eth_soc`, which drives the same kind of knob through `net_dim` and does not reproduce this.

Maybe this is something to raise with @LorenzoBianconi ? I assume that interrupt moderation would be desirable if feasible...

VitaliySochniy:
> Well, that repro was on 6.18, and the next paragraph says it didn't repro on `net/main` with `veth` on x86 (forcibly selecting voluntary preemption), right? So maybe that muddied the message a bit?

Probably, yes, and the wording was too weak. That VM run was never a test of stock `net/main`: on x86 there is no way to build a kernel in the configuration where this patch does anything. `PREEMPT_NONE` needs `ARCH_NO_PREEMPT` and `PREEMPT_VOLUNTARY` needs `!ARCH_HAS_PREEMPT_LAZY`, so to get voluntary preemption at all I had to drop the `ARCH_HAS_PREEMPT_LAZY` select from `arch/x86/Kconfig` first. What was under test was a configuration that does not exist upstream on that architecture.

On top of that the vehicle was weak: `veth` is a poor stand-in for a NIC with no interrupt moderation, and I could not hold the load inside the band that provokes it, since below it the thread sleeps between batches and above it `repoll` stays set so the existing `cond_resched()` runs. So it is not a negative result, it is an attempt that could not be made fair. Reporting it was honest but it gave an easy place to stop reading.

> Maybe this is something to raise with @LorenzoBianconi ? I assume that interrupt moderation would be desirable if feasible...

Agreed, and worth asking rather than patching blind. Since that comment I looked into why the register is cleared:

- It is not a driver-side decision. The vendor SDK disables it at the same point in init, with the same comment, so the line was carried over as-is.
- The SDK does expose an API to configure it, with a frame-count threshold and a time threshold, and its self-test sweeps a range of values for both. Nothing mentions an erratum.
- Nothing outside those tests ever calls that API, so there are no vendor defaults to copy either.

That makes "is there a known reason this is disabled?" the right first question for Lorenzo. If the answer is no reason, `get_coalesce`/`set_coalesce` looks straightforward; the field layout is documented on the vendor side.

For scale, all at roughly 80k pkt/s of terminated traffic: `airoha_eth` runs 72k-163k interrupts/s with cpu0 at 0.2-1.2% idle, against about a twelfth of that interrupt rate on MT7621 at a comparable packet rate. Two caveats on how much that is worth. It is terminated traffic; forwarded traffic goes through the PPE and does not load the CPU the same way, so the gain would mostly show for VPN, SQM and similar. And an opt-in ethtool knob changes nothing by default, so the real benefit needs either a changed default or `net_dim`, which is the part that would need justifying with measurements

VitaliySochniy:
@hauke @jonasjelonek @robimarko @nbd168 @dangowrt @plappermaul — merge or close?

- **Adds no `cond_resched()` call.** The hunk is `+1/-1`; `napi_threaded_poll_loop()` has exactly one before and after. It changes which loop exit reaches it, so the `break` path yields too.
- Inert on `PREEMPT`/`PREEMPT_LAZY`, where `_cond_resched()` returns 0. Live on `KERNEL_PREEMPT_NONE`, the default on every 6.18 target.
- A/B pairs on two boards: worst `ip link del` 151.92 s -> 1.00 s under `PREEMPT_NONE`, 135.95 s -> 0.51 s under `PREEMPT_VOLUNTARY`, same packet rate, untainted kernels. Two other people reproduced the defect on the XG-040G-MF; twain ran the patched half too.
- Cost where it is live: none found. Throughput within 3% and `cpu0` idle unchanged across pairs. Ping RTT under a saturating load improves rather than degrades: p50 1.17 -> 1.10 ms, p99 2.60 -> 1.93 ms, max 88.0 -> 3.5 ms, 2/3000 lost -> 0.
- Declined upstream, structurally: since 7.0 `PREEMPT_NONE` and `PREEMPT_VOLUNTARY` are gated off x86 and arm64, so it is a nop on the architectures netdev cares about. @rsalvaterra's whack-a-mole objection stands as a concern about direction.

The commit message now says to revisit at the next kernel bump rather than wait for an upstream landing that will not come.

robimarko:
This honestly is not something that we should carry downstream as far as I am concerned

PR #24734, opened 2026-08-15
ramips: mt7620: disable broken CPU frequency scaling

dormancygrace:
This revives Shiji Yang's 2023 patch on current `main` and adds hardware validation on an MT7620A WE826.

The MT7620-specific systick path enabled CPU frequency scaling after `SI_Sleep` and raised systick above the stable MIPS clocksource. The clock framework does not observe the resulting CPU clock change, breaking timing assumptions for delay calibration and users such as the Jitter RNG.

This revision removes the unused autosleep callback and its elevated rating, uses the generic systick compatible for MT7620A/N, and explicitly clears `CLK_LUT_CFG.SLEEP_EN` during early MT7620 initialization. The result is independent of reset state and of what a bootloader left behind.

Hardware results on OpenWrt 25.12.5 / Linux 6.12.94:

* before: `SLEEP_EN=1`, systick clocksource, urngd consumed 92-100% of otherwise-idle CPU for several minutes and reported startup at about 204 seconds;
* after: `SLEEP_EN=0`, MIPS clocksource, urngd initialized at 41.5 seconds and slept normally;
* five-run Ethernet throughput remained about 93.9 Mbit/s;
* WireGuard, warm boot, mixed-load stability, and a 30-minute clock-drift run showed no regression.

An isolation build kept systick selected while preventing only `SLEEP_EN`. urngd still consumed 83-91% CPU. Switching that running system to the MIPS clocksource and restarting urngd made it initialize in about six seconds and sleep normally. The effective fix is therefore the combination of a stable MIPS timebase and no untracked CPU-frequency changes.

The change is rebased onto current `main` / Linux 6.18.44. The full ramips patch stack applies cleanly, `git diff --check` is clean, and the modified `timer-ralink.o` and `mt7620.o` compile with the OpenWrt MIPS toolchain. Full 6.18 target builds are left to CI; hardware validation currently remains on 6.12.94.

Original submission:
https://lists.openwrt.org/pipermail/openwrt-devel/2023-July/041232.html

Relationship to openwrt/openwrt#24589: compiling Jitter RNG 3.x with `-O2` is an independent urngd performance fix, while running urngd at `nice=19` is a cross-platform scheduling workaround that keeps a CPU-bound entropy bootstrap from blocking other init services. This PR addresses the MT7620-specific platform root cause: an unstable systick timebase combined with untracked `SI_Sleep` CPU-frequency changes. It therefore does not make the urngd changes obsolete on other slow targets; `nice=19` remains useful defense in depth.

### DIFF
--- a/target/linux/ramips/dts/mt7620a.dtsi
+++ b/target/linux/ramips/dts/mt7620a.dtsi
@@ -295,7 +295,7 @@
 		};
 
 		systick: systick@d00 {
-			compatible = "ralink,mt7620a-systick", "ralink,cevt-systick";
+			compatible = "ralink,cevt-systick";
 			reg = <0xd00 0x10>;
 
 			interrupt-parent = <&cpuintc>;
--- a/target/linux/ramips/dts/mt7620n.dtsi
+++ b/target/linux/ramips/dts/mt7620n.dtsi
@@ -241,7 +241,7 @@
 		};
 
 		systick: systick@d00 {
-			compatible = "ralink,mt7620a-systick", "ralink,cevt-systick";
+			compatible = "ralink,cevt-systick";
 			reg = <0xd00 0x10>;
 
 			interrupt-parent = <&cpuintc>;
--- a/target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch
+++ b/target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch
@@ -1,60 +1,36 @@
-From: John Crispin <blogic@openwrt.org>
-Date: Sun, 28 Jul 2013 16:26:41 +0200
-Subject: [PATCH 2/2] MIPS: ralink: add cpu frequency scaling
+From: Gleb Pesin <dormancygrace@gmail.com>
+Date: Sat, 15 Aug 2026 19:30:00 +0300
+Subject: [PATCH] MIPS: ralink: improve systick timer handling
 
-This feature will break udelay() and cause the delay loop to have longer delays
-when the frequency is scaled causing a performance hit.
+Keep the system counter running when the clockevent is shut down, program
+the next event with wrap-safe arithmetic, and register the clocksource and
+clockevent with the modern helpers.
+
+Do not implement MT7620 CPU autosleep here. Its untracked CPU frequency
+changes break delay calibration and timekeeping.
+
+Clear SLEEP_EN during early MT7620 initialization so the result does not
+depend on reset state or on what a bootloader left behind.
+
+This is based on the original MT7620 CPU frequency scaling patch by John
+Crispin and the later systick timer improvements by Michael Lee.
 
 Signed-off-by: John Crispin <blogic@openwrt.org>
+Signed-off-by: Michael Lee <igvtee@gmail.com>
+Signed-off-by: Gleb Pesin <dormancygrace@gmail.com>
 ---
- drivers/clocksource/timer-ralink.c | 117 ++++++++++++++++++++++-------
- 1 file changed, 89 insertions(+), 28 deletions(-)
-
 --- a/drivers/clocksource/timer-ralink.c
 +++ b/drivers/clocksource/timer-ralink.c
-@@ -5,6 +5,7 @@
-  * Copyright (C) 2013 by John Crispin <john@phrozen.org>
-  */
- 
-+#include <asm/mach-ralink/ralink_regs.h>
- #include <linux/clockchips.h>
- #include <linux/clocksource.h>
- #include <linux/interrupt.h>
-@@ -26,6 +27,10 @@
- /* enable the counter */
- #define CFG_CNT_EN		0x1
- 
-+/* mt7620 frequency scaling defines */
-+#define CLK_LUT_CFG	0x40
-+#define SLEEP_EN	BIT(31)
-+
- struct systick_device {
+@@ -30,24 +30,38 @@ struct systick_device {
  	void __iomem *membase;
  	struct clock_event_device dev;
-@@ -33,21 +38,53 @@ struct systick_device {
- 	int freq_scale;
+ 	int irq_requested;
+-	int freq_scale;
  };
  
-+static void (*systick_freq_scaling)(struct systick_device *sdev, int status);
-+
  static int systick_set_oneshot(struct clock_event_device *evt);
  static int systick_shutdown(struct clock_event_device *evt);
  
-+static inline void mt7620_freq_scaling(struct systick_device *sdev, int status)
-+{
-+	if (sdev->freq_scale == status)
-+		return;
-+
-+	sdev->freq_scale = status;
-+
-+	pr_info("%s: %s autosleep mode\n", sdev->dev.name,
-+			(status) ? ("enable") : ("disable"));
-+	if (status)
-+		rt_sysc_w32(rt_sysc_r32(CLK_LUT_CFG) | SLEEP_EN, CLK_LUT_CFG);
-+	else
-+		rt_sysc_w32(rt_sysc_r32(CLK_LUT_CFG) & ~SLEEP_EN, CLK_LUT_CFG);
-+}
-+
 +static inline unsigned int read_count(struct systick_device *sdev)
 +{
 +	return ioread32(sdev->membase + SYSTICK_COUNT);
@@ -90,7 +66,7 @@ Signed-off-by: John Crispin <blogic@openwrt.org>
  }
  
  static void systick_event_handler(struct clock_event_device *dev)
-@@ -57,20 +94,25 @@ static void systick_event_handler(struct
+@@ -57,20 +71,25 @@ static void systick_event_handler(struct
  
  static irqreturn_t systick_interrupt(int irq, void *dev_id)
  {
@@ -124,37 +100,16 @@ Signed-off-by: John Crispin <blogic@openwrt.org>
  		.features		= CLOCK_EVT_FEAT_ONESHOT,
  		.set_next_event		= systick_next_event,
  		.set_state_shutdown	= systick_shutdown,
-@@ -88,7 +130,13 @@ static int systick_shutdown(struct clock
+@@ -88,7 +107,7 @@ static int systick_shutdown(struct clock
  	if (sdev->irq_requested)
  		free_irq(systick.dev.irq, &systick.dev);
  	sdev->irq_requested = 0;
 -	iowrite32(0, systick.membase + SYSTICK_CONFIG);
 +	iowrite32(CFG_CNT_EN, systick.membase + SYSTICK_CONFIG);
-+
-+	if (systick_freq_scaling)
-+		systick_freq_scaling(sdev, 0);
-+
-+	if (systick_freq_scaling)
-+		systick_freq_scaling(sdev, 1);
  
  	return 0;
  }
-@@ -113,20 +161,37 @@ static int systick_set_oneshot(struct cl
- 	return 0;
- }
- 
-+static const struct of_device_id systick_match[] = {
-+	{ .compatible = "ralink,mt7620a-systick", .data = mt7620_freq_scaling},
-+	{},
-+};
-+
- static int __init ralink_systick_init(struct device_node *np)
- {
- 	int ret;
-+	const struct of_device_id *match;
-+	int rating = 200;
- 
- 	systick.membase = of_iomap(np, 0);
+@@ -121,12 +140,6 @@ static int __init ralink_systick_init(st
  	if (!systick.membase)
  		return -ENXIO;
  
@@ -164,48 +119,56 @@ Signed-off-by: John Crispin <blogic@openwrt.org>
 -	systick.dev.max_delta_ticks = 0x7fff;
 -	systick.dev.min_delta_ns = clockevent_delta2ns(0x3, &systick.dev);
 -	systick.dev.min_delta_ticks = 0x3;
-+	match = of_match_node(systick_match, np);
-+	if (match) {
-+		systick_freq_scaling = match->data;
-+		/*
-+		 * cevt-r4k uses 300, make sure systick
-+		 * gets used if available
-+		 */
-+		rating = 310;
-+	}
-+
-+	/* enable counter than register clock source */
-+	iowrite32(CFG_CNT_EN, systick.membase + SYSTICK_CONFIG);
-+	clocksource_mmio_init(systick.membase + SYSTICK_COUNT, np->name,
-+			SYSTICK_FREQ, rating, 16, clocksource_mmio_readl_up);
-+
-+	/* register clock event */
  	systick.dev.irq = irq_of_parse_and_map(np, 0);
  	if (!systick.dev.irq) {
  		pr_err("%pOFn: request_irq failed", np);
-@@ -134,21 +199,16 @@ static int __init ralink_systick_init(st
+@@ -134,13 +147,19 @@ static int __init ralink_systick_init(st
  		goto err_iounmap;
  	}
  
--	ret = clocksource_mmio_init(systick.membase + SYSTICK_COUNT, np->name,
++	/* enable counter then register clock source */
++	iowrite32(CFG_CNT_EN, systick.membase + SYSTICK_CONFIG);
+ 	ret = clocksource_mmio_init(systick.membase + SYSTICK_COUNT, np->name,
 -				    SYSTICK_FREQ, 301, 16,
--				    clocksource_mmio_readl_up);
--	if (ret)
--		goto err_free_irq;
--
++				    SYSTICK_FREQ, 200, 16,
+ 				    clocksource_mmio_readl_up);
+ 	if (ret)
+ 		goto err_free_irq;
+ 
 -	clockevents_register_device(&systick.dev);
++	/* register clock event */
 +	systick.dev.name = np->name;
-+	systick.dev.rating = rating;
++	systick.dev.rating = 200;
 +	systick.dev.cpumask = cpumask_of(0);
 +	clockevents_config_and_register(&systick.dev, SYSTICK_FREQ, 0x3, 0x7fff);
  
  	pr_info("%pOFn: running - mult: %d, shift: %d\n",
  			np, systick.dev.mult, systick.dev.shift);
+--- a/arch/mips/ralink/mt7620.c
++++ b/arch/mips/ralink/mt7620.c
+@@ -36,6 +36,10 @@
+ #define PMU1_CFG		0x8C
+ #define DIG_SW_SEL		BIT(25)
  
- 	return 0;
++/* CPU sleep */
++#define CLK_LUT_CFG		0x40
++#define SLEEP_EN		BIT(31)
++
+ /* EFUSE bits */
+ #define EFUSE_MT7688		0x100000
  
--err_free_irq:
--	irq_dispose_mapping(systick.dev.irq);
- err_iounmap:
- 	iounmap(systick.membase);
- 	return ret;
+@@ -228,6 +232,14 @@ void __init prom_soc_init(struct ralink_
+ 		"MediaTek %s ver:%u eco:%u",
+ 		name, mt7620_get_soc_ver(), mt7620_get_soc_eco());
+ 
++	if (!is_mt76x8()) {
++		u32 clk_lut_cfg;
++
++		clk_lut_cfg = __raw_readl(MT7620_SYSC_BASE + CLK_LUT_CFG);
++		__raw_writel(clk_lut_cfg & ~SLEEP_EN,
++			     MT7620_SYSC_BASE + CLK_LUT_CFG);
++	}
++
+ 	cfg0 = __raw_readl(MT7620_SYSC_BASE + SYSC_REG_SYSTEM_CONFIG0);
+ 	if (is_mt76x8()) {
+ 		dram_type = cfg0 & DRAM_TYPE_MT7628_MASK;


target/linux/ramips/dts/mt7620a.dtsi
```diff
@@ -295,7 +295,7 @@
 		};
 
 		systick: systick@d00 {
-			compatible = "ralink,mt7620a-systick", "ralink,cevt-systick";
+			compatible = "ralink,cevt-systick";
```
openwrt-ai, line 298: After this change nothing in the tree ever writes `SLEEP_EN` any more. The only writer is [`mt7620_freq_scaling()` in 312-MIPS-ralink-add-cpu-frequency-scaling.patch](``https://github.com/openwrt/openwrt/blob/8243e045589b31b8f864d728225cd81e12430692/target/linux/ramips/patches-6.18/312-MIPS-ralink-add-cpu-frequency-scaling.patch#L43-L56),`` reachable only via the [`ralink,mt7620a-systick` match entry](``https://github.com/openwrt/openwrt/blob/8243e045589b31b8f864d728225cd81e12430692/target/linux/ramips/patches-6.18/312-MIPS-ralink-add-cpu-frequency-scaling.patch#L147)`` that this line removes — so the previous `mt7620_freq_scaling(sdev, 0)` call that *cleared* the bit is gone too.

The observation in the commit message that "`SLEEP_EN` remained clear" is therefore a statement about the WE826's bootloader, not something the kernel now guarantees. If any MT7620 bootloader hands over with `SLEEP_EN` already set, the outcome is arguably worse than today: the CPU clock still scales, but the selected clocksource/clockevent becomes the r4k one, whose rate is latched once at boot ([`mips_hpt_frequency = clk_get_rate(clk) / 2` in arch/mips/ralink/clk.c:83](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/ralink/clk.c#L83)) and never updated — i.e. untracked frequency changes against a timebase that cannot follow them.

Have you checked the reset/bootloader value of `CLK_LUT_CFG[31]` on more than the one WE826? If it can't be assumed clear, it would be safer to also make the driver clear `SLEEP_EN` unconditionally at init rather than rely on boot state.

---

target/linux/ramips/dts/mt7620n.dtsi
```diff
@@ -241,7 +241,7 @@
 		};
 
 		systick: systick@d00 {
-			compatible = "ralink,mt7620a-systick", "ralink,cevt-systick";
+			compatible = "ralink,cevt-systick";
```
openwrt-ai, line 244: With this hunk and the mt7620a.dtsi one, `ralink,mt7620a-systick` has no remaining user in the tree — `git grep` finds it only inside 312-MIPS-ralink-add-cpu-frequency-scaling.patch itself. That leaves `mt7620_freq_scaling()`, the `systick_match` table, `struct systick_device.freq_scale` and the `rating = 310` branch as unreachable code, and it leaves a trap: any future MT7620 board DTS copied from an older tree (or from the pre-change `mt7620a.dtsi`) silently re-enables the exact bug this commit fixes, with no build-time or runtime warning.

Would you consider dropping the frequency-scaling half of that patch in the same series instead of (or in addition to) disabling it from DT? A DT-only opt-out relies on every board file staying correct forever, whereas removing the match entry makes the bad configuration unrepresentable.

Nothing wrong with the two DTS lines themselves — for what it's worth they also bring the node in line with the upstream binding, which is [`const: ralink,cevt-systick`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/timer/ralink%2Ccevt-systick.yaml#L14) with no SoC-specific string allowed, and with [rt5350.dtsi:242](https://github.com/openwrt/openwrt/blob/8243e045589b31b8f864d728225cd81e12430692/target/linux/ramips/dts/rt5350.dtsi#L242), whose SoC-specific string is likewise unmatched and which therefore already runs at rating 200.

---

target/linux/ramips/dts/mt7620a.dtsi
```diff
@@ -295,7 +295,7 @@
 		};
```
openwrt-ai, line 295: nit: the PR body states the hardware validation was done on "OpenWrt 25.12.5 / Linux 6.12.94", but this PR targets `main`, where ramips is on [`KERNEL_PATCHVER:=6.18`](https://github.com/openwrt/openwrt/blob/8243e045589b31b8f864d728225cd81e12430692/target/linux/ramips/Makefile#L13) and the only patch directory is `patches-6.18`. The body says the DTS change was rebased and run through `checkpatch.pl` against `main`, but not that the behaviour was re-confirmed on 6.18. Worth stating explicitly which kernel the boot/urngd/throughput numbers come from, since the timer and urngd behaviour is exactly what changed between those series.

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the diff; comments are inline.

---

target/linux/ramips/dts/mt7620a.dtsi
dormancygrace: Agreed. I checked the custom WE826 bootloader source: it fixes the CPU PLL at 600 MHz and disables SSC, but it does not write `CLK_LUT_CFG`, so relying on its hand-off state would be unsafe. The updated commit now clears `SLEEP_EN` explicitly in early `prom_soc_init()` for MT7620A/N (while excluding MT76x8), independent of reset or bootloader state.

target/linux/ramips/dts/mt7620n.dtsi
dormancygrace: Agreed. The updated commit removes the frequency-scaling mechanism itself: the `freq_scale` state, callback, match table, MT7620-specific match, and rating-310 branch are gone. The generic systick remains at rating 200, so copying an old DTS compatible can no longer re-enable the autosleep path.

target/linux/ramips/dts/mt7620a.dtsi
dormancygrace: Clarified in both the commit message and PR body. The measured boot, urngd, throughput, stability, and drift results are from OpenWrt 25.12.5 / Linux 6.12.94. On current `main` / Linux 6.18.44, the full ramips patch stack applies, checkpatch is clean for the diff, and the modified `timer-ralink.o` and `mt7620.o` compile with the OpenWrt MIPS toolchain; full target builds are left to CI. I am not claiming 6.18 hardware validation yet.

target/linux/ramips/patches-6.18/312-MIPS-ralink-add-cpu-frequency-scaling.patch
```diff
@@ -1,60 +1,34 @@
+From b4db6d29d3873ebf71bdef82303c2399eb2d89b5 Mon Sep 17 00:00:00 2001
```
openwrt-ai: `b4db6d29d3873ebf71bdef82303c2399eb2d89b5` is not a Linux commit. `git fetch --depth=1 https://github.com/torvalds/linux.git b4db6d29d3873ebf71bdef82303c2399eb2d89b5` and the same against `gregkh/linux` both return `upload-pack: not our ref` (the identical command with `1efe5d048a391de3ead2804b2e7f86376c356cc5`, i.e. v6.18.44, succeeds, so the method is sound). It resolves in `openwrt/openwrt` instead — to the 2015 commit *"ramips: improve systick timer"* (`SVN-Revision: 47122`).

In a `target/linux/*/patches-*/` file the `From <sha> Mon Sep 17 00:00:00 2001` line is read as the upstream Linux commit id (compare [316-arch-mips-do-not-select-illegal-access-driver-by-def.patch:1](``https://github.com/openwrt/openwrt/blob/009a87f4107c8e578b078ea3a4a037d371318c5a/target/linux/ramips/patches-6.18/316-arch-mips-do-not-select-illegal-access-driver-by-def.patch#L1)),`` so this hash will send anyone who tries to look it up nowhere. It is also not the origin of the content in this file: at that openwrt.git commit the ramips systick changes live in `patches-3.18/0021-MIPS-ralink-add-cpu-frequency-scaling.patch` and `patches-3.18/0063-cevt-rt3352.patch`, against `arch/mips/ralink/cevt-rt3352.c`, in a substantially different form.

The file carried no `From <sha>` line before this PR — simplest is to drop it again and leave the provenance to the commit message, which already has the `Link:` trailer.

```suggestion
```

---

target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch
```diff
@@ -1,60 +1,34 @@
+From b4db6d29d3873ebf71bdef82303c2399eb2d89b5 Mon Sep 17 00:00:00 2001
 From: John Crispin <blogic@openwrt.org>
 Date: Sun, 28 Jul 2013 16:26:41 +0200
-Subject: [PATCH 2/2] MIPS: ralink: add cpu frequency scaling
+Subject: [PATCH] MIPS: ralink: improve systick timer handling
```
openwrt-ai, line 3: The patch is now titled *"MIPS: ralink: improve systick timer handling"* and no longer adds CPU frequency scaling — it removes it — but the file is still `312-MIPS-ralink-add-cpu-frequency-scaling.patch`. In this directory the filename tracks the subject: [311-MIPS-use-set_mode-to-enable-disable-the-cevt-r4k-irq.patch:3](``https://github.com/openwrt/openwrt/blob/009a87f4107c8e578b078ea3a4a037d371318c5a/target/linux/ramips/patches-6.18/311-MIPS-use-set_mode-to-enable-disable-the-cevt-r4k-irq.patch#L3)`` and [316-arch-mips-do-not-select-illegal-access-driver-by-def.patch:4](``https://github.com/openwrt/openwrt/blob/009a87f4107c8e578b078ea3a4a037d371318c5a/target/linux/ramips/patches-6.18/316-arch-mips-do-not-select-illegal-access-driver-by-def.patch#L4)`` both match their filenames. As it stands the filename now advertises exactly the behaviour the patch deletes, which is the worst case for someone grepping the patch stack later. Please `git mv` it to something like `312-MIPS-ralink-improve-systick-timer-handling.patch`.

nit: relatedly, the header keeps `From: John Crispin` / `Date: Sun, 28 Jul 2013` / `Signed-off-by: John Crispin` (lines 2, 3, 16) while the Subject and the entire description have been replaced with a summary of changes that are not his — the systick modernisation (`read_count()`/`write_compare()` helpers, the `STATUSF_IP7`-gated ISR, `clockevents_config_and_register()`) predates this PR and traces to openwrt.git `b4db6d29`, signed off by Michael Lee; the `SLEEP_EN` clearing is yours. Keeping the original subject and description and putting the new text under the existing `[Gleb: ...]` annotation would keep attribution straight.

---

target/linux/ramips/patches-6.18/312-MIPS-ralink-add-cpu-frequency-scaling.patch
```diff
@@ -209,3 +152,30 @@ Signed-off-by: John Crispin <blogic@openwrt.org>
...
+ 	snprintf(soc_info->sys_type, RAMIPS_SYS_TYPE_LEN,
+ 		"MediaTek %s ver:%u eco:%u",
+ 		name, mt7620_get_soc_ver(), mt7620_get_soc_eco());
++	if (!is_mt76x8()) {
++		u32 clk_lut_cfg;
++
++		clk_lut_cfg = __raw_readl(MT7620_SYSC_BASE + CLK_LUT_CFG);
++		__raw_writel(clk_lut_cfg & ~SLEEP_EN,
++			     MT7620_SYSC_BASE + CLK_LUT_CFG);
++	}
```
openwrt-ai: nit: the new block butts directly against the `snprintf()` with no blank line, while every other logical step in [`prom_soc_init()`](``https://github.com/openwrt/openwrt/blob/009a87f4107c8e578b078ea3a4a037d371318c5a/target/linux/ramips/patches-6.18/312-MIPS-ralink-add-cpu-frequency-scaling.patch#L168-L181)`` (`cfg0` read, `soc_info->mem_base`, the `pmu0`/`pmu1` reads) is separated by one. Suggestion below also bumps the hunk's new-side count from 13 to 14 so the patch still applies without fuzz; if you'd rather regenerate it, `make target/linux/refresh` with ramips selected does the same.

```suggestion
@@ -227,6 +231,14 @@ void __init prom_soc_init(struct ralink_
 	snprintf(soc_info->sys_type, RAMIPS_SYS_TYPE_LEN,
 		"MediaTek %s ver:%u eco:%u",
 		name, mt7620_get_soc_ver(), mt7620_get_soc_eco());
+
+	if (!is_mt76x8()) {
+		u32 clk_lut_cfg;
+
+		clk_lut_cfg = __raw_readl(MT7620_SYSC_BASE + CLK_LUT_CFG);
+		__raw_writel(clk_lut_cfg & ~SLEEP_EN,
+			     MT7620_SYSC_BASE + CLK_LUT_CFG);
+	}
```

---

openwrt-ai (commented):
Reviewed the force-pushed revision (`8243e04` → `009a87f`). The new approach addresses both points from the previous round — the frequency-scaling mechanism is gone from the timer patch rather than just unhooked from DT, and `SLEEP_EN` is now cleared in early SoC init.

Verified locally against Linux 6.18.44 (`KERNEL_PATCHVER:=6.18`): the rewritten `312-*.patch` applies to `drivers/clocksource/timer-ralink.c` and `arch/mips/ralink/mt7620.c` with zero fuzz, all seven hunk headers have correct line counts, and the resulting `timer-ralink.c` has no dangling references to `freq_scale`, `rt_sysc_*`, `CLK_LUT_CFG` or `ralink_regs.h` — so dropping the `#include <asm/mach-ralink/ralink_regs.h>` hunk is safe. `read_c0_cause()`/`STATUSF_IP7` remain reachable via `linux/clocksource.h` → `linux/timex.h` → `asm/timex.h` → `asm/mipsregs.h`. The `ramips: mt7620:` scope also checks out: `rt5350.dtsi` uses `ralink,rt5350-systick`, which never matched the removed table, so RT3352/RT5350 were already at rating 200 and are unaffected by the 310 → 200 change.

Remaining comments are inline and all concern the patch file's metadata rather than the code.

---

target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch
dormancygrace: Addressed in c29e83f4c8: the file is now named 312-MIPS-ralink-improve-systick-timer-handling.patch, the nonexistent Linux envelope SHA is gone, and the embedded patch metadata now attributes the downstream rework to me while documenting John Crispin and Michael Lee as the original sources and retaining their sign-offs. The requested blank separator and refreshed hunk count are included as well.

namiltd:
Add to commit message for clarity:
"With CPU frequency scaling removed, the MIPS Count Register is again a reliable clock source. Lower systick rating back to 200 so cevt-r4k (rating 300) is preferred as the clockevent, while systick continues to serve as the clocksource."

dormancygrace:
Added to the commit message as suggested. Thanks — this makes the clocksource/clockevent split much clearer.

namiltd (approved):

target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch
```diff
@@ -164,26 +119,16 @@ Signed-off-by: John Crispin <blogic@openwrt.org>
...
-+		 * gets used if available
-+		 */
-+		rating = 310;
-+	}
-+
 +	/* enable counter than register clock source */
 +	iowrite32(CFG_CNT_EN, systick.membase + SYSTICK_CONFIG);
 +	clocksource_mmio_init(systick.membase + SYSTICK_COUNT, np->name,
-+			SYSTICK_FREQ, rating, 16, clocksource_mmio_readl_up);
++			SYSTICK_FREQ, 200, 16, clocksource_mmio_readl_up);
```
openwrt-ai: This lowers the **clocksource** rating from 301 to 200 (line 143 below does the same for the clockevent), which contradicts the paragraph added to the commit message in this revision:

> Lower the systick rating back to 200 so cevt-r4k (rating 300) is preferred as the clockevent, while systick continues to serve as the clocksource.

The cevt-r4k half checks out — [`cd->rating = 300` at cevt-r4k.c:315](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/kernel/cevt-r4k.c#L315). The clocksource half does not. The MIPS r4k clocksource registers with [`rating = 200 + clamp(mips_hpt_frequency / 10000000, 0, 99)` at csrc-r4k.c:126-127](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/kernel/csrc-r4k.c#L126-L127), and ralink sets [`mips_hpt_frequency = clk_get_rate(clk) / 2` at clk.c:83](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/ralink/clk.c#L83) — 290 MHz on a 580 MHz MT7620A, i.e. rating 229. Since 229 > 200, after this change the **MIPS** count register wins the clocksource selection and systick is demoted to a fallback.

That is exactly what the PR body reports as the desired end state ("before: `SLEEP_EN=1`, systick clocksource … after: `SLEEP_EN=0`, MIPS clocksource") and what the isolation experiment identified as half the fix — so the code is right and the sentence is wrong. Suggest rewording to something like: systick drops to 200 for both the clocksource and the clockevent, so cevt-r4k (300) takes the clockevent and the now-reliable MIPS count register (rating ~229 here) takes the clocksource.

---

target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch
```diff
@@ -1,60 +1,36 @@
-From: John Crispin <blogic@openwrt.org>
-Date: Sun, 28 Jul 2013 16:26:41 +0200
-Subject: [PATCH 2/2] MIPS: ralink: add cpu frequency scaling
+From: Gleb Pesin <dormancygrace@gmail.com>
+Date: Fri, 15 Aug 2026 19:30:00 +0300
```
openwrt-ai: nit: 2026-08-15 was a Saturday, not a Friday (`date -d 2026-08-15 +%A`). `git am` won't care, but a mismatched weekday is the kind of thing that makes a hand-written RFC-822 date look synthesised.

```suggestion
Date: Sat, 15 Aug 2026 19:30:00 +0300
```

---

openwrt-ai (commented):
Reviewed the delta since `009a87f` (rename + patch-header rework + the `prom_soc_init()` hunk adjustment).

## Commit checks
- `0c9a6b8` "ramips: mt7620: disable broken CPU frequency scaling" — the paragraph added in this revision states "systick continues to serve as the clocksource", but the diff also drops the systick *clocksource* rating from 301 to 200, which puts it below the MIPS r4k clocksource's computed rating (~229 on a 580 MHz MT7620A). The PR body's own measurement ("after: … MIPS clocksource") says the same. Code is right, sentence is wrong — details inline.

Everything else in this revision verified clean:

- The renamed file applies to pristine v6.18.44 with `patch -p1 -F0` and **zero fuzz**; no other ramips or generic patch touches `arch/mips/ralink/mt7620.c` or `drivers/clocksource/timer-ralink.c`, so the offsets stand on their own. The corrected hunk header `@@ -228,6 +232,14 @@` matches: `prom_soc_init()` starts at line 220 upstream, so the three context lines begin at 228, and 228 + 4 (the `CLK_LUT_CFG`/`SLEEP_EN` define block added by the earlier hunk) = 232 on the new side.
- The new blank-line separator before `cfg0 = __raw_readl(...)` resolves the previous spacing comment.
- The bogus `From b4db6d29… Mon Sep 17 00:00:00 2001` envelope is gone, and the bare `From:`/`Date:`/`Subject:` form matches 17 of the 46 patches already in `target/linux/ramips/patches-6.18/` (e.g. `120-1-mips-pci-mt7620-fix-bridge-register-access.patch`, `330-mt7620-avoid-direct-jal-in-random-helper.patch`), so the bot's stylistic warning about the missing hash line looks benign — `FormalityCheck / Code Patches` itself is green.
- The new attribution is accurate: openwrt.git `b4db6d29` ("ramips: improve systick timer") carries `Signed-off-by: Michael Lee <igvtee@gmail.com>`, and its `patches-3.18/0066-cevt.patch` is where the `read_count()`/`write_compare()` helpers, the `STATUSF_IP7`-gated ISR, the `CFG_CNT_EN`-on-shutdown change and `clockevents_config_and_register()` originate — this patch is a faithful forward-port of that minus the frequency-scaling half, so `Signed-off-by: John Crispin` + `Signed-off-by: Michael Lee` + your own is the right chain.

All 25 check runs on `0c9a6b8` are green (ramips mt7620/mt76x8/mt7621/rt305x/rt3883 kernel builds and `Check Kernel patches` all pass), so no CI-grounded comments.

---

namiltd:
"Unfortunately" the bot is more attentive than the human. 
Change to:
"With CPU frequency scaling removed, the MIPS Count Register is again a reliable clock source. Systick drops to rating 200 for both the clocksource and the clockevent, so cevt-r4k (rating 300) takes the clockevent and the now-reliable MIPS count register (rating ~229 here) takes the clocksource."

target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch
dormancygrace: Fixed in 849319bd7c. The commit message now states that systick is rated 200 for both roles, cevt-r4k becomes the clockevent, and the MIPS Count register becomes the clocksource.

target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch
dormancygrace: Fixed in 849319bd7c: corrected the embedded patch date to Sat, 15 Aug 2026.

dormancygrace:
Thanks, corrected in 849319bd7c. The message now describes the actual selection: systick is rating 200 for both roles, cevt-r4k wins the clockevent, and the now-reliable MIPS Count register wins the clocksource.

target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch
openwrt-ai: The reply on the previous thread says "corrected the embedded patch date to Sat, 15 Aug 2026" in `849319bd7c`, but `849319b` only amends the commit message — `git diff 0c9a6b8..849319b` is empty, and this line still reads `Fri`. 2026-08-15 was a Saturday. The fix appears not to have made it into the amend.

```suggestion
Date: Sat, 15 Aug 2026 19:30:00 +0300
```

---

target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch
openwrt-ai: Two error-handling regressions relative to pristine v6.18.44 come in with this reordering (raising them late — they were present in earlier revisions too and I missed them):

1. `clocksource_mmio_init()`'s return value is now discarded. Upstream checks it — [`ret = clocksource_mmio_init(...); if (ret) goto err_free_irq;` at timer-ralink.c:136-140](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/clocksource/timer-ralink.c#L136-L140) — so a failed registration now goes unreported.
2. The clocksource is registered (and `CFG_CNT_EN` written) *before* `irq_of_parse_and_map()`, but the `if (!systick.dev.irq)` path below still does `goto err_iounmap`, which `iounmap()`s `systick.membase`. That leaves a live, registered clocksource whose `clocksource_mmio_readl_up()` reads through the just-unmapped address. Upstream cannot hit this because it registers the clocksource only after the IRQ is mapped.

The simplest fix that keeps the new ordering is to capture the return and unregister/bail before `iounmap()`:

```
	ret = clocksource_mmio_init(systick.membase + SYSTICK_COUNT, np->name,
			SYSTICK_FREQ, 200, 16, clocksource_mmio_readl_up);
	if (ret)
		goto err_iounmap;
```

…and, in the `!systick.dev.irq` branch, either drop the counter/clocksource again or move the `iowrite32(CFG_CNT_EN, …)` + `clocksource_mmio_init()` pair back below the IRQ mapping (which also restores upstream ordering). Either way the `@@ -121,12 +140,12 @@` hunk header needs its new-side count updated — `make target/linux/refresh` with ramips selected does that.

---

target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch
```diff
@@ -164,26 +119,16 @@ Signed-off-by: John Crispin <blogic@openwrt.org>
...
-+	if (match) {
-+		systick_freq_scaling = match->data;
-+		/*
-+		 * cevt-r4k uses 300, make sure systick
-+		 * gets used if available
-+		 */
-+		rating = 310;
-+	}
-+
 +	/* enable counter than register clock source */
```
openwrt-ai: nit: "than" should be "then" — the comment describes the ordering (enable the counter, then register the clocksource). Inherited from the 3.18-era openwrt patch, but since this line is being introduced fresh into `patches-6.18/` it may as well land correct.

```suggestion
+	/* enable counter then register clock source */
```

---

openwrt-ai (commented):
Reviewed the delta since `0c9a6b8`. `git diff 0c9a6b8..849319b` is **empty** — this revision only amends the commit message, replacing the clocksource/clockevent paragraph with namiltd's wording. That paragraph now matches the code: the patch sets rating 200 for both the clocksource ([line 125](``https://github.com/openwrt/openwrt/blob/849319bd7c43d424c81cf51e2f65544e41a3a897/target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch#L125))`` and the clockevent ([line 143](``https://github.com/openwrt/openwrt/blob/849319bd7c43d424c81cf51e2f65544e41a3a897/target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch#L143)),`` so the previous round's inline finding is resolved. Commit message and diff otherwise agree.

Two things inline: the embedded `Date:` fix that was reported as landed in `849319bd7c` is not actually in the tree, and — raising these late, they predate this revision — the reordered `ralink_systick_init()` drops upstream's `clocksource_mmio_init()` error check and leaves an `iounmap()`-after-registration path.

CI on `849319b` is queued/pending with nothing failed, so no CI-grounded comments.

---

target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch
dormancygrace: Fixed in a45b4888b5: the embedded patch date now matches Saturday, 15 August 2026.

target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch
dormancygrace: Fixed in a45b4888b5. IRQ mapping now happens before enabling/registering the clocksource; clocksource_mmio_init() is assigned to ret and failures go through the existing irq_dispose_mapping() cleanup path.

target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch
dormancygrace: Fixed the typo ("then") in a45b4888b5.

namiltd:
Noticed the duplicate init block from the earlier revision (the one with the "than" typo and no error handling before irq_of_parse_and_map()) is gone now — good catch cleaning that up. Single init path with proper err_free_irq handling looks correct. LGTM on this part.

openwrt-ai (commented):

The delta since `849319b` is confined to the embedded patch: the `Fri` → `Sat` date fix and the `ralink_systick_init()` reordering. Both previous findings are resolved, and I re-verified the result rather than taking the reordering on trust:

- **Applies with zero fuzz.** Applied `312-MIPS-ralink-improve-systick-timer-handling.patch` with `patch -p1 -F0` onto pristine v6.18.44 (`1efe5d0`) — both files patch clean. The two reworked hunk headers are arithmetically correct: `@@ -121,12 +140,6 @@` (6 context + 6 removed = 12 old, 6 new) and `@@ -134,13 +147,19 @@` (11 context + 2 removed = 13 old, 11 + 8 added = 19 new).
- **Error handling restored.** The resulting `ralink_systick_init()` now maps the IRQ first, then does `iowrite32(CFG_CNT_EN, …)` + `ret = clocksource_mmio_init(…)` with `if (ret) goto err_free_irq;`, matching upstream's ordering at [timer-ralink.c:130-146](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/clocksource/timer-ralink.c#L130-L146). Because the patch no longer deletes the `err_free_irq:` label, the `goto` target exists and `irq_dispose_mapping()` runs before `iounmap()` — no registered-clocksource-over-unmapped-memory window, and no unused-label warning.
- **Rating change unchanged and consistent.** Systick is 200 for both roles ([clocksource, line 133](``https://github.com/openwrt/openwrt/blob/a45b4888b5874003d0c2a46f04a7a8561c2e951e/target/linux/ramips/patches-6.18/312-MIPS-ralink-improve-systick-timer-handling.patch#L133)`` and clockevent), which is what the amended commit message now describes.
- **No leftovers.** `ralink,mt7620a-systick` has no user anywhere in the tree after this commit, and `SLEEP_EN` is written only by the new `prom_soc_init()` block; `CLK_LUT_CFG`/`SLEEP_EN` do not collide with any existing define in `arch/mips/ralink/mt7620.c`.
- **Message claims hold.** `checkpatch.pl --no-tree` on the resulting kernel diff reports 0 errors / 0 warnings (only the synthetic "missing commit description" for the hand-built diff), and `git diff --check` against the base is clean.

All 25 check runs on `a45b488` are green — ramips mt7620/mt76x8/mt7621/rt305x/rt3883 kernel builds, `Check Kernel patches`, and all three `FormalityCheck` jobs — so no CI-grounded comments.

---

namiltd:
Ping @robimarko @jonasjelonek

robimarko:

PR #24735, opened 2026-08-15
 kernel: backport fix for Macronix MX25L128xx series Flash

DragonBluep:
We need this patch to fix support for the early generation MX25L128xx series Flash as it doesn't have SFDP.

Link: https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/patch/?id=01739ccea21b69818dc071096a249e6f84a6b13d

### DIFF
--- a/target/linux/generic/pending-6.18/489-01-mtd-spi-nor-Add-support-for-MX25L12833F-and-MX25L128.patch
+++ b/target/linux/generic/pending-6.18/489-01-mtd-spi-nor-Add-support-for-MX25L12833F-and-MX25L128.patch
@@ -1,0 +1,77 @@
+From 733d1d1e820b64e6643bfaa9a3680bdd851fbfb2 Mon Sep 17 00:00:00 2001
+From: Cheng Ming Lin <chengminglin@mxic.com.tw>
+Date: Fri, 5 Jun 2026 16:48:36 +0800
+Subject: [PATCH 1/2] mtd: spi-nor: Add support for MX25L12833F and MX25L12845G
+
+Add support for Macronix MX25L12833F and MX25L12845G SPI NOR flashes.
+These parts share the same JEDEC ID (0xc2, 0x20, 0x18) as the legacy
+MX25L12805D.
+
+The newer flashes support SFDP and 1-4-4 Page Program in 3-byte address
+mode, but this 4PP capability is not defined in their SFDP tables.
+Conversely, the legacy MX25L12805D lacks SFDP support entirely and does
+not support 4PP.
+
+To safely enable 4PP for the newer flashes without breaking the legacy
+part, introduce a post_sfdp fixup. Since the legacy MX25L12805D does
+not support SFDP, it falls back to static parameters and will never
+execute the post_sfdp hook. The newer flashes will successfully parse
+the SFDP, trigger the hook, and safely append the SNOR_HWCAPS_PP_1_4_4
+capability.
+
+Signed-off-by: Cheng Ming Lin <chengminglin@mxic.com.tw>
+---
+ drivers/mtd/spi-nor/macronix.c | 26 +++++++++++++++++++++++++-
+ 1 file changed, 25 insertions(+), 1 deletion(-)
+
+--- a/drivers/mtd/spi-nor/macronix.c
++++ b/drivers/mtd/spi-nor/macronix.c
+@@ -83,6 +83,25 @@ mx25l3255e_late_init_fixups(struct spi_n
+ 	return 0;
+ }
+ 
++static int
++mx25l12805d_4pp3b_post_sfdp_fixups(struct spi_nor *nor)
++{
++	struct spi_nor_flash_parameter *params = nor->params;
++
++	/*
++	 * JEDEC ID 0xc22018 is shared by MX25L12805D (no SFDP, no 4PP) and
++	 * MX25L12833F/MX25L12845G (support SFDP and 4PP in 3-byte mode).
++	 * The legacy 05D lacks SFDP and will not execute this hook. For
++	 * the newer flashes, 3-byte 1-4-4 PP is not defined in SFDP, so
++	 * we safely enable it here.
++	 */
++	params->hwcaps.mask |= SNOR_HWCAPS_PP_1_4_4;
++	spi_nor_set_pp_settings(&params->page_programs[SNOR_CMD_PP_1_4_4],
++				SPINOR_OP_PP_1_4_4, SNOR_PROTO_1_4_4);
++
++	return 0;
++}
++
+ static const struct spi_nor_fixups mx25l25635_fixups = {
+ 	.post_bfpt = mx25l25635_post_bfpt_fixups,
+ 	.post_sfdp = macronix_qpp4b_post_sfdp_fixups,
+@@ -96,6 +115,10 @@ static const struct spi_nor_fixups mx25l
+ 	.late_init = mx25l3255e_late_init_fixups,
+ };
+ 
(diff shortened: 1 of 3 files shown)


target/linux/generic/pending-6.18/489-02-mtd-spi-nor-macronix-Restore-fallback-parameters-for.patch
```diff
@@ -0,0 +1,41 @@
...
+--- a/drivers/mtd/spi-nor/macronix.c
++++ b/drivers/mtd/spi-nor/macronix.c
+@@ -155,7 +155,9 @@ static const struct flash_info macronix_
+ 	}, {
+ 		/* MX25L12805D, MX25L12833F, MX25L12845G */
+ 		.id = SNOR_ID(0xc2, 0x20, 0x18),
++		.size = SZ_16M,
+ 		.flags = SPI_NOR_HAS_LOCK | SPI_NOR_4BIT_BP,
++		.no_sfdp_flags = SECT_4K,
+ 		.fixups = &mx25l12805d_4pp3b_fixups,
```
openwrt-ai: With `.size` restored here, JEDEC 0xc22018 stops parsing SFDP entirely — which makes the `post_sfdp` fixup added by 489-01 dead code.

[`spi_nor_needs_sfdp()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mtd/spi-nor/core.h#L697-L705) is just `!nor->info->size`, so once `.size = SZ_16M` is set probing falls into the `spi_nor_init_params_deprecated()` branch of [`spi_nor_init_params()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mtd/spi-nor/core.c#L3117-L3127). That helper only re-parses SFDP when `no_sfdp_flags` carries one of DUAL/QUAD/OCTAL read ([core.c:3012-3016](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mtd/spi-nor/core.c#L3012-L3016)), and this entry only gets `SECT_4K`.

Net effect of 489-01 + 489-02 applied together:

- `mx25l12805d_4pp3b_post_sfdp_fixups()` can never run — its in-code rationale ("The newer flashes will successfully parse the SFDP, trigger the hook") no longer holds after 489-02.
- MX25L12833F / MX25L12845G regress relative to 6.18.44 as shipped today (where the entry has no `.size`, so SFDP is parsed): they lose SFDP-derived dual/quad read, page size and 4B opcode setup and fall back to `SECT_4K` only.

MX25L12805D itself is fixed as intended, so the trade may be deliberate — but since the series is still only on the list and not in mainline, is it worth carrying the newer-part regression now? If SFDP should be kept for the newer parts, adding e.g. `SPI_NOR_DUAL_READ | SPI_NOR_QUAD_READ` to `no_sfdp_flags` would route through `spi_nor_sfdp_init_params_deprecated()`, which rolls back to the static parameters when the legacy part fails to answer RDSFDP.

---

target/linux/generic/pending-6.18/489-01-mtd-spi-nor-Add-support-for-MX25L12833F-and-MX25L128.patch
openwrt-ai, line 1: nit: the commit subject says "backport fixes", but neither patch is upstream yet — a commit search of `torvalds/linux` for `MX25L12833F` returns nothing, and the `Link:` points at the June 2026 lore submission. The `pending-6.18/` placement is the right one; it's the subject wording that disagrees with it, since in OpenWrt "backport" means an upstream-accepted commit (`backport-6.18/`). Consider something like `kernel: add pending fixes for Macronix MX25L128xx series Flash`.

---

openwrt-ai (commented):
Reviewed 1 new commit. The refresh of `target/linux/ath79/patches-6.18/450-mtd-spi-nor-macronix-locking-support-for-mx25l64-ser.patch` is correct — I applied 489-01, 489-02 and 450 in build order against v6.18.44 and all three land with no offset or fuzz. Two inline notes, one of them about the combined runtime effect of the two new patches.

---

openwrt-ai (commented):
Reviewed 1 new commit (force-pushed rework of the previous head); no new issues found.

The rework resolves both points from the previous review:

- `489-01-...-Add-support-for-MX25L12833F-and-MX25L128.patch` is dropped, so the `post_sfdp` hook that could never fire once `.size` is set is gone with it. Only upstream 2/2 remains.
- The subject no longer says "backport" for a patch that sits in `pending-6.18/`.

Dropping 489-01 also makes the `450-mtd-spi-nor-macronix-locking-support-for-mx25l64-ser.patch` refresh unnecessary, and reverting it was correct: I applied `pending-6.18/489-...` then `ath79/patches-6.18/450-...` in build order (generic before platform, per [`include/quilt.mk:103-105`](https://github.com/openwrt/openwrt/blob/03a3b5bfc274c15c742848de9f4915cedf0d4ddd/include/quilt.mk#L103-L105)) against v6.18.44 `drivers/mtd/spi-nor/macronix.c` — both land with no offset and no fuzz.

One thing I flagged last time and am explicitly *not* re-raising: with `.size = SZ_16M` set, [`spi_nor_needs_sfdp()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mtd/spi-nor/core.h#L697-L705) goes false, so JEDEC 0xc22018 takes the `spi_nor_init_params_deprecated()` path and the SFDP-capable MX25L12833F / MX25L12845G lose SFDP-derived quad read. But the resulting entry is byte-for-byte what `macronix.c` already carries on 6.12 (`.size = SZ_16M`, `.flags = SPI_NOR_HAS_LOCK | SPI_NOR_4BIT_BP`, `.no_sfdp_flags = SECT_4K`), so this restores 6.12 parity rather than introducing a divergence — and no `pending-6.12/` counterpart is needed.

---

openwrt-ai (commented):

---

target/linux/generic/pending-6.18/489-02-mtd-spi-nor-macronix-Restore-fallback-parameters-for.patch
openwrt-ai: Solved by dropping 489-01 — the unreachable `post_sfdp` hook is gone with it, thanks.

---

target/linux/generic/pending-6.18/489-01-mtd-spi-nor-Add-support-for-MX25L12833F-and-MX25L128.patch
openwrt-ai: fixed, thanks

---

openwrt-ai (commented):

---

openwrt-ai (commented):
## Commit checks

- 337b6d0 "kernel: backport fix for Macronix MX25L128xx series Flash" — the rebase set the committer to `DragonBluep <70847398+DragonBluep@users.noreply.github.com>`; on the previous head it was `Shiji Yang <yangshiji66@outlook.com>`, and the formality check fails on both the name format and the noreply address. Redo the rebase locally with `user.name`/`user.email` set to a real name and a non-noreply address, then force-push.

---

openwrt-ai (commented):

---

openwrt-ai (commented):

---

openwrt-ai (commented):

---

PR #24432, opened 2026-07-26
qualcommbe: add support for IPQ95XX based Jio Router AX6000 JIDU6J11 [6111, 6411, 6611, 6811 and 6911]

Sandiep:
# JioRouter AX6000 (JIDU6J11 series)

## Hardware

| Component | Details |
|---|---|
| **SoC** | Qualcomm IPQ9554 (4× ARM Cortex-A73 @ 1.5 GHz) |
| **RAM** | 512 MB |
| **Flash** | 256 MB NAND – UBI volume layout |
| **Ethernet** | 5× 10/100/1000 Mbps (1 WAN + 4 LAN) via QCA8075 switch |
| **WLAN 2.4 GHz** | Integrated IPQ95XX AHB — 802.11b/g/n/ax, 4×4 MIMO (ath11k) |
| **WLAN 5 GHz** | QCN9074 PCIe — 802.11a/n/ac/ax, 4×4 MIMO (ath11k) |
| **LEDs** | 3× GPIO-controlled (Red/Green/Blue) |
| **Button** | 1× Reset |
| **USB** | 1× USB 3.0 |

## MAC Address Assignment

| Interface | Source |
|---|---|
| WAN / Label | ART partition, offset `0x00` |
| LAN | WAN + 1 |
| 2.4 GHz | WAN+2|
| 5 GHz | WAN+3|

### 1. Prepare TFTP server

- Set a static IP on your computer (e.g. `192.168.1.2`).
- Place the initramfs image (`openwrt-qualcommbe-ipq95xx-jiorouter_ax6000-jidu6j11-6111-initramfs-uImage.itb`) in the TFTP root.

### 2. Interrupt U-Boot

- Connect a serial console (115200 baud).
- Power on the device and press any key when prompted to stop autoboot.

### 3. Load and run the initramfs image

```sh
setenv ipaddr 192.168.1.1
setenv serverip 192.168.1.2
tftpboot 0x44000000 openwrt-qualcommbe-ipq95xx-jiorouter_ax6000-jidu6j11-6111-initramfs-uImage.itb
bootm 0x44000000
```

### 4. Flash sysupgrade image

- Transfer the sysupgrade image to the running system (e.g. via SCP or HTTP).
- Execute:

```sh
sysupgrade /tmp/openwrt-qualcommbe-ipq95xx-jiorouter_ax6000-jidu6j11-6111-squashfs-sysupgrade.bin
```

## Notes

[...]

PR #24736, opened 2026-08-15
qualcommbe: add Linksys LN6001 support

ikm-san:
This adds fail-closed, RAM-boot-only support for the IPQ9554-based
Linksys Velop WRT Pro 7 (LN6001/MBE70).

## Included

- IPQ9554 DTS/profile using the shared `ipq9574.dtsi` and
  `qcom,ipq9574` family fallback
- 1 GiB DDR4 RAM and Samsung 8GTF4R 8 GB eMMC
- board 48 MHz XO and 24 MHz board-XO clock setup required by eMMC,
  MDIO and PWM
- device-local disable of PCS1/UNIPHY1, which is absent on IPQ9554
- QCA8075 PSGMII wiring for LAN1-LAN4
- QCA8081 SGMII wiring for the 2.5 GbE WAN port, using its exact public
  PHY ID compatible
- stable wired MAC addresses read from the existing `devinfo` partition
- reset and WPS GPIO keys, plus red/green/blue/white PWM status LEDs
- default LAN/WAN interface grouping
- device-local PWM LED support
- an initramfs-only LN6001 profile
- a non-forceable sysupgrade refusal

The reset node describes the physical key wired to GPIO 40 and maps it to
`KEY_RESTART`. It does not attempt to reproduce an OEM-specific hold timer.

## RAM boot and stock return

The generated development artifact is:

`openwrt-qualcommbe-ipq95xx-linksys_ln6001-initramfs-uImage.itb`

Copy or rename it to `ln6001-initramfs.itb` in a TFTP server root. Connect
a serial console at 115200 8N1, interrupt the three-second U-Boot
autoboot, and use a TFTP server at 192.168.1.254/24:

```
setenv ipaddr 192.168.1.1
setenv serverip 192.168.1.254
setenv netmask 255.255.255.0
tftp 0x4d100000 ln6001-initramfs.itb
bootm 0x4d100000#config@1
```

The `setenv` values are temporary. Do not run `saveenv` and do not write
the FIT to eMMC. Reset or power-cycle the router and allow its normal
`bootipq` command to run to return to stock firmware.

The U-Boot command, complete transfer, FIT verification and stock-return
behavior were checked on retail hardware.

## MAC address and partition safety

Retail runtime evidence shows that `devinfo` contains `hw_mac_addr`.
Stock `br-lan` and WAN both use its base value. The DTS models only that
named GPT partition through the read-only block NVMEM framework and assigns
offset zero to the four OpenWrt LAN members and WAN. This gives each member
a permanent address while retaining the observed stock bridge identity.
No unit address is recorded here.

The stock GPT remains runtime-discovered. The tested unit reports an invalid
alternate GPT and uses its primary GPT; this proposal does not repair or
rewrite either copy. It adds no fixed partition offsets and no code that
formats, resizes or writes `devinfo` or the partition table.

## Image and upgrade safety

The profile sets `IMAGES :=`, so it generates no factory, sysupgrade, disk,
GPT, partition-table or persistent-rootfs image.

For this device, `platform_check_image()` returns the non-forceable failure
code 74. `platform_do_upgrade()` returns before reaching either an eMMC or
default storage writer. Automatic sysupgrade remains unsupported.

The initramfs root and configuration are RAM-backed and intentionally do not
survive a power cycle. No persistent overlay, installation path or eMMC write
was tested.

## Hardware validation and rebase provenance

The selected LN6001 initramfs profile was built from a clean tree at the
pre-rebase commit `f2fbe5c802ce625d76b393696b159a9b82feee8f`, based directly
on upstream commit `ce22f3ba6cbfea047a48f9d512b2247bcb6e38f4`.

The current PR head `cc6db21d22c42d13e67260a238d1c85674f719d9` is the same
one-commit device proposal rebased onto upstream commit
`3809dfe5c69b43bc9b39f7b75fa493bbc2ad57ac`. The 20 upstream commits
between the preceding PR base `928bb37350d20a2a7737e3a3b5329261572ca533`
and this base do not touch any of the five paths changed by this proposal.
Within the LN6001 patch, the only source-tree adjustments after hardware
testing are the review-requested public clock-rationale comments and moving
the unchanged clock override group to match the sibling IPQ9574 file layout.
No DT property value, image definition or runtime behavior changed. The
artifact was not rebuilt or booted after this base rebase and non-functional
DTS cleanup, so the hardware results below apply to the same effective LN6001
configuration.

- artifact size: 10,848,128 bytes
- artifact SHA-256:
  `75d4af083ef976a7521f7f39a2cf25af5e4c6e5406d29aa1f29985d81a357b88`
- U-Boot received the exact byte count and verified the FIT before starting
  the kernel
- socinfo reported IPQ9554, SoC ID 512, revision 1.1
- runtime clocks showed a 48 MHz XO, 24 MHz board XO and approximately
  100 MHz PWM clock
- eMMC initialized with CQE and enumerated in HS400 Enhanced Strobe mode;
  its main device reported `ro=0`, while boot0/boot1 retained `force_ro=1`
- an early SDHCI software-reset timeout was followed immediately by successful
  controller initialization and enumeration; no later MMC I/O, read-only or
  timeout error was observed
- `/` was RAM-backed `tmpfs`; there was no overlay or eMMC mount and no write
  was performed
- the named `devinfo` NVMEM provider registered; LAN1-LAN4/WAN each reported
  permanent rather than random assignment, and their ordered fingerprint
  matched the preceding cold RAM boot
- four QCA8075 PHYs attached; the WAN PHY at `90000.mdio-1:1c` bound as
  `Qualcomm QCA8081`, rather than the generic Clause 45 fallback seen before
  the exact-ID correction
- WAN established a 1 Gbps full-duplex link, reported a clean cable-down
  event, then renegotiated 1 Gbps full duplex after replug without a PHY
  error, timeout or reset
- LAN4 established a 1 Gbps link on the tested exact-ID artifact and was the
  TFTP path; LAN1-LAN3 physical wiring was exercised on the preceding
  otherwise-identical image, where LAN1/LAN2 showed the known PSGMII
  speed-reporting dependency and LAN3 established 1 Gbps full duplex
- all four PWM channels and LED class entries registered; red, green, blue and
  white were each selected individually and visually confirmed, then the
  original blue-on state was restored
- reset and WPS DT nodes registered with GPIO 40 and GPIO 37 as pulled-up
  inputs; short physical presses produced complete high-low-high transitions

A diagnostic DT containing the MP5496 L5 regulator was rejected by RPM with
`-ENXIO` on the tested unit. The earlier eMMC, MDIO and PWM failures were
instead caused by missing board clock properties. With those clocks described
and L5 omitted, the final DTS reaches the recovered HS400 state above. No
enabled LN6001 node references L5.

The early `uboot-envtools` fallback still has no environment NVMEM/configuration
mapping. That is deliberate: this proposal does not guess the stock A/B
environment layout. It is separate from the working read-only `devinfo` NVMEM
provider used for wired addresses.

Static validation on the rebased head
`cc6db21d22c42d13e67260a238d1c85674f719d9`:

- OpenWrt checkpatch: 0 errors, 0 warnings
- shell syntax checks
- target/device metadata parsing, confirming no flashable image metadata
- LN6001 no-write upgrade dispatch checks
- Git whitespace and repository integrity checks
- scans for credentials, raw MACs/serials, private paths, excluded packages,
  proprietary references, logs, artifacts and binary blobs
- GitHub FormalityCheck and the qualcommbe/ipq95xx kernel-patch and kernel-build
  jobs passed on the preceding PR head; the green Build PR Profile result
  skipped checkout and compilation, so CI did not build an LN6001 image.
  Current-head status is recorded by GitHub Checks.

Only the selected initramfs profile was built. No factory, sysupgrade,
persistent-rootfs or wireless artifact was generated, and no router storage
was written.

## Explicit exclusions and dependency

- no persistent rootfs, factory or sysupgrade image
- no wireless nodes, calibration data, board data or firmware blobs
- no A/B writer, boot-environment mutation or confirmation service
- no proprietary source, private packages, credentials or raw logs
- no generic PSGMII/phylink patch


target/linux/qualcommbe/ipq95xx/base-files/lib/upgrade/platform.sh
```diff
@@ -5,11 +5,23 @@ RAMFS_COPY_BIN='fw_printenv fw_setenv head'
...
+		return 0
+		;;
+	esac
 }
 
 platform_do_upgrade() {
 	case "$(board_name)" in
+	linksys,ln6001)
+		echo "Refusing automatic sysupgrade on this device"
+		exit 1
```
openwrt-ai: `exit 1` here terminates `do_stage2` itself, so the script never reaches its tail — see [`do_stage2:9-23`](https://github.com/openwrt/openwrt/blob/4b299bc732d153574a4e76889b0d6a93cb6a7a91/package/base-files/files/lib/upgrade/do_stage2#L9-L23), where `platform_do_upgrade` is called unconditionally and the `umount -a` / `reboot -f` that recovers the box comes after it. By the time this runs, stage2 has already killed every service and pivoted into the ramfs with the real rootfs unmounted, so the device would sit in a dead ramdisk until it is power-cycled — the opposite of fail-closed.

`return 1` lets `do_stage2` fall through to the reboot, and since nothing was written the board comes back up on the untouched firmware.

```suggestion
		return 1
```


---

target/linux/qualcommbe/image/ipq95xx.mk
```diff
@@ -14,6 +14,19 @@ define Device/8devices_kiwi-dvk
 endef
 TARGET_DEVICES += 8devices_kiwi-dvk
 
+define Device/linksys_ln6001
```
openwrt-ai, line 34: This is a new `.dts` plus a new `define Device/...`, which puts the commit under the project's new-device commit-message requirements (hardware spec / flash instructions / MAC layout). All three headings are present, but two of them are placeholders rather than content:

- *Installation*: "exact public loading and stock-recovery instructions remain to be documented" — no image name, no way into a flashable state, no revert-to-stock path.
- *MAC address layout*: "source and derivation not yet established" for both LAN and WAN.
- *Hardware specification*: RAM chip, flash capacity/chip, UART header location and pinout, and power ratings are all left open.

The bar is that a stranger could flash the board from the commit message alone, and that is not reachable today. Fully aware this is a draft and that the PR body lists these as remaining work — flagging so it is on the checklist for the un-draft pass. Two in-tree commits at the expected level of detail: [`986ca4c`](https://github.com/openwrt/openwrt/commit/986ca4c887f4088b6fbc703faa88884350e9274f) (terse) and [`a2dcbd7`](https://github.com/openwrt/openwrt/commit/a2dcbd79a4460617bd42151555448f8bca8ca7be) (fuller, U-Boot + OEM-dashboard flash paths).


---

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
```diff
@@ -0,0 +1,308 @@
...
+	pinctrl-names = "default";
+	status = "okay";
+};
+
+&qcom_ppe {
+	ethernet-ports {
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		lan1: port@1 {
```
openwrt-ai, line 145: nit: none of the five port nodes carry a MAC source — no `local-mac-address`, no `nvmem-cells`/`nvmem-cell-names` — and `02_network` does not call `ucidef_set_network_device_mac` either. The EDMA netdev probe falls back to [`eth_hw_addr_random()`](``https://github.com/openwrt/openwrt/blob/4b299bc732d153574a4e76889b0d6a93cb6a7a91/target/linux/qualcommbe/patches-6.18/0332-net-ethernet-qualcomm-Add-netdevice-support-for-QCOM.patch#L275)`` in that case, so every interface gets a fresh random MAC on each boot; DHCP leases, bridge FDB entries and anything MAC-pinned will churn across reboots.

The commit already lists the MAC layout as unresolved, so this is only to record the concrete runtime consequence for whoever tests the draft — a note in the commit message that random MACs are expected until the factory-partition offsets are established would save some confusion.


---

openwrt-ai (commented):
Reviewed 1 new commit. Verified against the in-tree qualcommbe patch stack: the `qcom,qca8075-package` compatible, the `psgmii`/`sgmii` phy-modes, the `pcs0_ch0..ch3`/`pcs2_ch0` handles, the `pwm` node and its TLMM `pwm` function on gpio44-47, and the `74` return convention in `validate_firmware_image` (non-forceable) all check out. Three comments inline.

---

target/linux/qualcommbe/ipq95xx/base-files/lib/upgrade/platform.sh
ikm-san: Fixed in d2ad5b08c957: LN6001 now uses return 1, so the common stage-2 path reaches cleanup and forced reboot without invoking a storage writer. I also removed the persistent/sysupgrade image definition; the profile now defines no flashable image.

target/linux/qualcommbe/image/ipq95xx.mk
ikm-san: Agreed. I narrowed the profile to RAM-boot/initramfs-only, removed the staging persistent image, and documented the known 8 GB eMMC capacity and 12 V / 2.5 A supply. The public RAM-boot/recovery procedure, RAM/eMMC part identification, and UART header details are still incomplete. I am leaving this thread open and the PR in Draft until those facts are documented.

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
ikm-san: Agreed. There is no verified public MAC source or derivation yet, and the current fallback may be random on each boot. I documented that behavior in the amended commit and PR description and am leaving this thread open until stable LAN/WAN MAC provisioning is implemented from verified data.

target/linux/qualcommbe/ipq95xx/base-files/lib/upgrade/platform.sh
```diff
@@ -5,11 +5,23 @@ RAMFS_COPY_BIN='fw_printenv fw_setenv head'
...
+		;;
+	esac
 }
 
 platform_do_upgrade() {
 	case "$(board_name)" in
+	linksys,ln6001)
+		echo "Refusing automatic sysupgrade on this device"
+		return 1
+		;;
```
openwrt-ai: nit: this case is inserted ahead of `8devices,kiwi-dvk)`, while the other two board lists this PR touches keep their entries alphabetical — `linksys,ln6001)` goes after `8devices,kiwi-dvk)` in [`02_network:17`](``https://github.com/openwrt/openwrt/blob/d2ad5b08c957db9e98493088a9a9e1e19b2be2b3/target/linux/qualcommbe/ipq95xx/base-files/etc/board.d/02_network#L17),`` and `define Device/linksys_ln6001` goes after `define Device/8devices_kiwi-dvk` in [`ipq95xx.mk:17`](https://github.com/openwrt/openwrt/blob/d2ad5b08c957db9e98493088a9a9e1e19b2be2b3/target/linux/qualcommbe/image/ipq95xx.mk#L17). Moving this block below the `8devices,kiwi-dvk)` case keeps all three consistent. No functional impact — `case` matching is exact, not ordered.

---

openwrt-ai (commented):
Reviewed the amended commit (`4b299bc` → `d2ad5b0`); the only non-rebase changes are the two hunks in `ipq95xx.mk` and `platform.sh`.

Both prior blocking points check out:

- `exit 1` → `return 1` in `platform_do_upgrade()` is the right fix. [`do_stage2:9-23`](https://github.com/openwrt/openwrt/blob/d2ad5b08c957db9e98493088a9a9e1e19b2be2b3/package/base-files/files/lib/upgrade/do_stage2#L9-L23) now falls through to `umount -a` / `reboot -f`, so a defensively-reached upgrade lands back on the untouched firmware instead of stalling in the ramdisk.
- `IMAGES :=` does what the commit message claims. It overrides the target `Device/Default`'s `IMAGES := sysupgrade.bin`, and the initramfs FIT is emitted independently of `IMAGES` via [`Device/Build/initramfs`](https://github.com/openwrt/openwrt/blob/d2ad5b08c957db9e98493088a9a9e1e19b2be2b3/include/image.mk#L633-L646). Since `qualcommbe/Makefile` carries the `ramdisk` feature, which [selects `USES_INITRAMFS`](https://github.com/openwrt/openwrt/blob/d2ad5b08c957db9e98493088a9a9e1e19b2be2b3/scripts/target-metadata.pl#L40) and therefore makes `TARGET_ROOTFS_INITRAMFS` default `y`, a stock build of this profile does produce `openwrt-qualcommbe-ipq95xx-linksys_ln6001-initramfs-uImage.itb` — the exact name in the Installation section. `image_prepare` still depends on `compile-dtb`, so the new DTS is compile-validated even with no images. Device-level empty `IMAGES` has in-tree precedent (`imx`).

Commit message matches the diff, and the `Signed-off-by:` trailer is well-formed. All three FormalityCheck runs are green on `d2ad5b0`.

Not re-raising the two open threads (installation/recovery procedure and MAC layout) — they are still accurate and already acknowledged as Draft blockers. One ordering nit inline.

---

target/linux/qualcommbe/ipq95xx/base-files/lib/upgrade/platform.sh
ikm-san: Fixed in b7b414450e9e: moved the LN6001 case below 8devices,kiwi-dvk for alphabetical consistency; no functional change.

ikm-san:
CI note for `d2ad5b08c957`: the requested [Build PR Profile run](https://github.com/openwrt/openwrt/actions/runs/32495002304) did not reach checkout or compilation. It parsed `build qualcommbe/ipq95xx/linksys_ln6001`, then failed while creating its initial status comment because the fork workflow token was read-only; GitHub returned HTTP 403 (`Resource not accessible by integration`).

The [Build Kernel workflow](https://github.com/openwrt/openwrt/actions/runs/32494916972) also performed no kernel compilation. `qualcommbe` is marked `source-only`, so target discovery produced empty build and patch matrices. These are CI workflow/permission outcomes, not LN6001 compiler failures.

openwrt-ai (commented):

Range `d2ad5b0..b7b4144` restricted to `target/linux/qualcommbe/` contains a single hunk: the `linksys,ln6001)` case in `platform_do_upgrade()` moved below `8devices,kiwi-dvk)`. That resolves the ordering nit from the previous review, and `platform_check_image()` / `platform_do_upgrade()` / `platform_copy_config()` in [`platform.sh`](https://github.com/openwrt/openwrt/blob/b7b414450e9ee61a5b6fdf5d54a58513cff85460/target/linux/qualcommbe/ipq95xx/base-files/lib/upgrade/platform.sh) are now consistently ordered with `02_network` and `ipq95xx.mk`. No functional change — `case` matching is exact. Everything else in the range is rebase noise from `main`.

The commit message is byte-identical to the one on `d2ad5b0`, still matches the diff, and the `Signed-off-by:` trailer is well-formed.

Not re-raising the two open threads (public RAM-boot/stock-recovery procedure + remaining hardware-spec details, and the LAN/WAN MAC source) — they are still accurate and already tracked as Draft blockers in the PR body.

---

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
```diff
@@ -0,0 +1,319 @@
...
+};
+
+&rpm_requests {
+	regulators {
+		compatible = "qcom,rpm-mp5496-regulators";
+
+		ipq9574_s1: s1 {
+			regulator-min-microvolt = <725000>;
+			regulator-max-microvolt = <1075000>;
+		};
```
openwrt-ai, line 258: nit: `725000` is not the rail's real floor, and without the rationale the next person to touch this node has no way to know that. It is the SVS voltage chosen so the regulator core does not bring the APC rail up below what 800 MHz needs, because at registration time the kernel does not know the bootloader already set 875 mV and would otherwise start from the true minimum. Upstream [`ipq9574-rdp-common.dtsi:95-106`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/arm64/boot/dts/qcom/ipq9574-rdp-common.dtsi#L95-L106) and both in-tree ipq9574 boards ([`ipq9570-kiwi-dvk.dts:231-242`](https://github.com/openwrt/openwrt/blob/4f48bd8a286f1159b79acc62a53f92e79457236b/target/linux/qualcommbe/dts/ipq9570-kiwi-dvk.dts#L231-L242), [`ipq9570-sbe1v1k.dts:178-189`](https://github.com/openwrt/openwrt/blob/4f48bd8a286f1159b79acc62a53f92e79457236b/target/linux/qualcommbe/dts/ipq9570-sbe1v1k.dts#L178-L189)) carry the comment verbatim, odd indentation included; copying it keeps the three boards textually identical here and keeps future upstream syncs diff-clean.

```suggestion
		ipq9574_s1: s1 {
		/*
		 * During kernel bootup, the SoC runs at 800MHz with 875mV set by the bootloaders.
		 * During regulator registration, kernel not knowing the initial voltage,
		 * considers it as zero and brings up the regulators with minimum supported voltage.
		 * Update the regulator-min-microvolt with SVS voltage of 725mV so that
		 * the regulators are brought up with 725mV which is sufficient for all the
		 * corner parts to operate at 800MHz
		 */
			regulator-min-microvolt = <725000>;
			regulator-max-microvolt = <1075000>;
		};
```

---

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
```diff
@@ -0,0 +1,319 @@
...
+	};
```
openwrt-ai, line 259: nit: the block stops at `s1`, while every other ipq9574 board also describes the 1.8 V `l5` rail with `regulator-always-on` + `regulator-boot-on` — [`ipq9570-kiwi-dvk.dts:244-249`](https://github.com/openwrt/openwrt/blob/4f48bd8a286f1159b79acc62a53f92e79457236b/target/linux/qualcommbe/dts/ipq9570-kiwi-dvk.dts#L244-L249), [`ipq9570-sbe1v1k.dts:191-196`](https://github.com/openwrt/openwrt/blob/4f48bd8a286f1159b79acc62a53f92e79457236b/target/linux/qualcommbe/dts/ipq9570-sbe1v1k.dts#L191-L196), and upstream [`ipq9574-rdp-common.dtsi:115-120`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/arm64/boot/dts/qcom/ipq9574-rdp-common.dtsi#L115-L120) (which adds `l2` as well). Is the omission deliberate?

Evidence cuts both ways, which is why this is a question rather than an assertion:

- Argues it is fine: `l5` has no consumer on this board. `kiwi-dvk` references it from the PCIe PHYs' `vdda-pll-supply`, but LN6001 enables no PCIe, and `sbe1v1k` declares it with no consumer at all. With the rail absent from DT, `qcom_smd-regulator` never registers it and Linux casts no RPM vote, so whatever the bootloader voted stays. Adding it *without* `always-on` would actually be worse — the regulator core's unused-regulator cleanup could pull it down.
- Argues it is missing: LN6001 enables `&sdhc_1` with `mmc-ddr-1_8v` / `mmc-hs200-1_8v` / `mmc-hs400-1_8v` and drives QCA8075/QCA8081 over MDIO, so there is real 1.8 V I/O on this board. If the LN6001 bootloader's RPM vote for `l5` is not sticky across the kernel handoff, eMMC and MDIO would be running off a rail nothing is holding up, and the failure mode (intermittent HS400 errors, PHY probe flakiness) is exactly the kind that survives a bench boot.

Only a scope or a rail-voltage readback on real hardware settles it. If you did verify the rail stays up without a Linux vote, a line in the commit message saying so would stop this coming back at un-draft time.

---

openwrt-ai (commented):
Reviewed 1 new commit (`b7b4144` → `4f48bd8`). Restricted to `target/linux/qualcommbe/`, the range contains exactly one non-rebase hunk: the new `&rpm_requests` block in `ipq9574-linksys-ln6001.dts`. Everything else is `main` moving under the branch — `askey_sbe1v1k` landing (which is why the LN6001 entries in `ipq95xx.mk` and `02_network` shifted anchors), and `source-only` dropping out of the target `FEATURES`.

That hunk is a required build fix, not cosmetics. [`ipq9574.dtsi:59`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/arm64/boot/dts/qcom/ipq9574.dtsi#L59) and its three sibling CPU nodes carry `cpu-supply = <&ipq9574_s1>`, and nothing under `patches-6.18/` touches those lines, so the label has to come from the board DTS — without this block dtc would reject the LN6001 DTS outright on an unresolved reference. The `qcom,rpm-mp5496-regulators` compatible, the `ipq9574_s1` label and the 725000/1075000 range all match upstream and both in-tree ipq9574 boards. I also re-resolved every label the DTS dereferences against the patched 6.18 dtsi (`pwm`, `qcom_ppe`, `pcs0_ch0..ch3`, `pcs2_ch0`, `nsscc`, `sdhc_1`, `mdio`, `mdio_pins`, `uart2_pins`, `sleep_clk`, `tlmm`, `blsp1_uart2`, `rpm_requests`) — all defined. Alphabetical placement of the new node between `&qcom_ppe` and `&sdhc_1` is consistent with the rest of the file, and the three board lists remain in the order agreed last round.

The commit message is byte-identical to the one on `b7b4144`, still matches the diff, and the `Signed-off-by:` trailer is well-formed. No CI statuses have been reported against `4f48bd8` yet, so there is nothing to ground against there.

Not re-raising the two open threads (public RAM-boot/stock-recovery procedure plus the remaining hardware-spec details, and the LAN/WAN MAC source) — both are still accurate and already tracked as Draft blockers in the PR body. Two parity nits inline on the new block; neither is blocking.

---

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
ikm-san: The omission is deliberate for this initial DTS. I do not have a direct L5 voltage measurement or public LN6001 rail mapping. The `mmc-*-1_8v` properties advertise signaling modes but do not identify the PMIC rail, and no enabled LN6001 node references L5. Adding an unreferenced always-on vote would therefore assert board wiring that has not been verified.

A real-hardware build using the same S1-only regulator description retained eMMC-backed configuration across repeated reboots and exercised Ethernet, including cable replug/link recovery, without observed storage or PHY instability. This is useful functional evidence, but not electrical proof of L5 or its consumers.

I will keep L5 omitted unless direct measurement or public board mapping identifies a required consumer. If that evidence becomes available, I will model the supply explicitly.

target/linux/qualcommbe/image/ipq95xx.mk
ikm-san: Addressed in c1bc35b7cca9. The single device commit now documents the Samsung 8GTF4R eMMC, RAM/UART basics, exact tftp/bootm RAM-boot procedure, temporary setenv/no-saveenv warning, and stock-return steps. The exact public-artifact retest remains listed in the PR body; the physical UART pin order and DRAM package marking are optional documentation gaps rather than source blockers.

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
ikm-san: Addressed in c1bc35b7cca9. Retail runtime evidence identifies devinfo key hw_mac_addr, and both stock wired LAN and WAN use the base value. The DTS now exposes only the named GPT partition through the read-only block NVMEM ASCII layout and assigns offset 0 to all five PPE ports. No unit address is recorded in the commit or PR.

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
ikm-san: Addressed in c1bc35b7cca9. The S1 node now includes the public voltage rationale explaining the bootloader voltage, regulator-registration behavior, and the 725 mV minimum used for 800 MHz operation.

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
```diff
@@ -0,0 +1,360 @@
...
+		#size-cells = <0>;
+
+		lan1: port@1 {
+			reg = <1>;
+			label = "lan1";
+			phy-mode = "psgmii";
+			managed = "in-band-status";
+			phy-handle = <&qca8075_0>;
+			pcs-handle = <&pcs0_ch0>;
+			nvmem-cells = <&hw_mac_addr 0>;
```
openwrt-ai, line 152: Every in-tree Linksys board that reads the same `hw_mac_addr` key out of `devinfo` via `ascii-eq-delim-env` splits the two: base+1 on the LAN ports, base+0 on WAN — [`ipq6000-mr7350.dts:378`](https://github.com/openwrt/openwrt/blob/c1bc35b7cca98107dd5529976d4e5fa0041b353b/target/linux/qualcommax/dts/ipq6000-mr7350.dts#L378) / [`:418`](https://github.com/openwrt/openwrt/blob/c1bc35b7cca98107dd5529976d4e5fa0041b353b/target/linux/qualcommax/dts/ipq6000-mr7350.dts#L418), [`ipq8072-mx8500.dts:436`](https://github.com/openwrt/openwrt/blob/c1bc35b7cca98107dd5529976d4e5fa0041b353b/target/linux/qualcommax/dts/ipq8072-mx8500.dts#L436) / [`:477`](https://github.com/openwrt/openwrt/blob/c1bc35b7cca98107dd5529976d4e5fa0041b353b/target/linux/qualcommax/dts/ipq8072-mx8500.dts#L477), and the qualcommbe sibling [`ipq9570-sbe1v1k.dts:400`](https://github.com/openwrt/openwrt/blob/c1bc35b7cca98107dd5529976d4e5fa0041b353b/target/linux/qualcommbe/dts/ipq9570-sbe1v1k.dts#L400) / [`:466`](https://github.com/openwrt/openwrt/blob/c1bc35b7cca98107dd5529976d4e5fa0041b353b/target/linux/qualcommbe/dts/ipq9570-sbe1v1k.dts#L466). The commit says retail uses the base value for both, which reads like the WAN address was checked against `hw_mac_addr` and the LAN side assumed rather than read off a running stock LAN port. Was `br-lan`'s address on stock firmware actually compared to `hw_mac_addr`, or only WAN's?

Only Linksys mesh nodes with no distinct WAN role ([`ipq8174-mx4300.dts:261-279`](https://github.com/openwrt/openwrt/blob/c1bc35b7cca98107dd5529976d4e5fa0041b353b/target/linux/qualcommax/dts/ipq8174-mx4300.dts#L261-L279)) use offset 0 across the board, so that precedent does not obviously carry to a router with a dedicated 2.5G WAN. Same question applies to lines 166, 185, 204 and 224.

---

openwrt-ai (commented):
Reviewed 1 new commit.

---

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
ikm-san: Yes, I compared the running stock `br-lan` address directly with `devinfo`'s `hw_mac_addr`; it matched the base value. The running stock WAN address also matched that base value.

That observation establishes the logical LAN bridge and WAN identities, not a factory-derived identity for each physical LAN member. I also inspected the stock member netdevs: they did not expose a stable base+1 LAN-port layout; most used random or locally administered addresses, with `eth0` being the exception. U-Boot's `ethaddr` is base+1, but that is the bootloader network identity and does not establish the stock Linux PPE-port mapping.

The offset-zero assignment on the four OpenWrt LAN port nodes is therefore deliberate: it replaces the stock member-port randomness with stable addresses while causing the resulting `br-lan` to retain the observed stock LAN identity. WAN also uses the observed base value. This does not claim that stock assigns the base address individually to every physical LAN port.

I will verify the resulting LAN-port, `br-lan`, and WAN addresses when RAM-booting the exact public initramfs. I do not plan to change LAN to base+1 solely from other-model precedent, because that would change the observed LN6001 `br-lan` identity.

ikm-san: Follow-up testing is now complete on the exact amended commit `77e77f5619433b87b9d9ef7c0893cbbae5191534`.

The named `devinfo` block NVMEM provider registered, and `lan1`-`lan4` plus `wan` each reported `addr_assign_type=0`, with no random-MAC warning. This confirms that the final public DTS supplies permanent rather than random addresses from `hw_mac_addr`; no raw address is included here.

I did not retain a raw cross-boot address comparison in the public evidence, so that explicit comparison remains on the Draft checklist. The offset-zero choice continues to be based on the separately observed stock `br-lan` and WAN identities described above, without claiming a stock identity for each physical LAN member.

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
ikm-san: Follow-up testing is now complete on the exact amended commit `77e77f5619433b87b9d9ef7c0893cbbae5191534`.

A public-source diagnostic DT that registered MP5496 L5 was rejected by the RPM interface with `-ENXIO` on the tested retail unit. The earlier eMMC, MDIO, and PWM failures were instead traced to missing board fixed-factor clock properties. With the 48 MHz XO and 24 MHz board clock described, and L5 omitted, the final initramfs image enumerated eMMC in HS400 Enhanced Strobe mode, attached all Ethernet PHYs, and registered all four PWM channels. No enabled LN6001 node references L5.

I have therefore kept L5 absent and documented this result in the amended commit. This establishes the Linux-visible RPM behavior on the tested unit; it does not claim a physical rail mapping or direct voltage measurement. No eMMC write was performed.

ikm-san:
Follow-up WAN validation is complete on amended commit
`f2fbe5c802ce625d76b393696b159a9b82feee8f`.

The prior generic Clause 45 compatible caused the PHY at
`90000.mdio-1:1c` to report ID zero and bind the generic driver. The amended
DTS uses the public QCA8081 ID compatible already used by other upstream
boards.

The exact initramfs tested was 10,848,128 bytes with SHA-256
`75d4af083ef976a7521f7f39a2cf25af5e4c6e5406d29aa1f29985d81a357b88`.
U-Boot received that exact byte count, verified the FIT, and started it from
RAM. At runtime the PHY bound as `Qualcomm QCA8081`; WAN established 1 Gbps
full duplex, reported a clean cable-down event, and renegotiated 1 Gbps full
duplex after replug without a PHY error, timeout or reset.

The initramfs root remained RAM-backed, no eMMC partition was mounted, and no
router storage was written. The PR body now records the updated artifact and
test evidence. The 2.5 Gbps WAN mode remains untested.

ikm-san:
Additional physical validation on the same exact RAM-booted commit is now
complete:

- a short WPS press produced a complete high-low-high transition on GPIO 37
- a short Reset press produced a complete high-low-high transition on GPIO 40;
  no long-press behavior was exercised
- red, green, blue and white PWM channels were selected one at a time and each
  color was visually confirmed; the original blue-on state was then restored

These checks changed only temporary RAM/runtime state. The tested source and
PR head remain `f2fbe5c802ce625d76b393696b159a9b82feee8f`, and no router storage
was written. The PR body has been updated accordingly.

ikm-san:
Moving this out of Draft for maintainer review.

All device-specific checks available on the retail unit are complete on the
exact PR head: RAM boot and stock return, partition/no-write safeguards, stable
wired identity, LAN wiring, QCA8081 WAN attach and 1 Gbps replug, both physical
buttons, and all four LED channels.

A 2.5 GbE-capable peer is not available here, so that link mode remains
explicitly disclosed as untested rather than blocking review. The hardware/PHY
is identified as QCA8081 and the WAN path is validated at 1 Gbps. Generic
PSGMII 100 Mbps/replug work remains separate in #24432 and is intentionally not
duplicated in this device commit. Profile/kernel CI still requires maintainer
workflow approval; all current FormalityCheck jobs are green.

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
```diff
@@ -0,0 +1,378 @@
...
+};
+
+&xo_board_clk {
+	clock-div = <2>;
+	clock-mult = <1>;
+};
+
+&xo_clk {
+	clock-frequency = <48000000>;
+};
```
openwrt-ai: nit: the `1`/`1` and `2`/`1` factors read as magic without the rationale that upstream [`ipq9574-rdp-common.dtsi:218-238`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/arm64/boot/dts/qcom/ipq9574-rdp-common.dtsi#L218-L238) and [`ipq9570-kiwi-dvk.dts:443-464`](https://github.com/openwrt/openwrt/blob/f2fbe5c802ce625d76b393696b159a9b82feee8f/target/linux/qualcommbe/dts/ipq9570-kiwi-dvk.dts#L443-L464) carry above these nodes — precedent is split, [`ipq9570-sbe1v1k.dts:541-553`](https://github.com/openwrt/openwrt/blob/f2fbe5c802ce625d76b393696b159a9b82feee8f/target/linux/qualcommbe/dts/ipq9570-sbe1v1k.dts#L541-L553) drops them. Copying them verbatim keeps future upstream syncs diff-clean. Placement is separately the one break in this file's otherwise alphabetical node order (`&pcs1` sits correctly between `&mdio` and `&pwm`); both siblings keep this group at the end of the file instead of between `&sleep_clk` and `&tlmm`.

```suggestion
/*
 * The bootstrap pins for the board select the XO clock frequency,
 * which automatically enables the right dividers to ensure the
 * reference clock output from WiFi is 48 MHZ.
 */
&ref_48mhz_clk {
	clock-div = <1>;
	clock-mult = <1>;
};

/*
 * The frequency of xo_board_clk is fixed to 24 MHZ, which is routed
 * from WiFi output clock 48 MHZ divided by 2.
 */
&xo_board_clk {
	clock-div = <2>;
	clock-mult = <1>;
};

&xo_clk {
	clock-frequency = <48000000>;
};
```

---

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
```diff
@@ -0,0 +1,378 @@
...
+				 <&nsscc NSS_CC_PORT4_RX_CLK>,
+				 <&nsscc NSS_CC_PORT4_TX_CLK>;
+			clock-names = "mac", "rx", "tx";
+			resets = <&nsscc PORT4_MAC_ARES>,
+				 <&nsscc PORT4_RX_ARES>,
+				 <&nsscc PORT4_TX_ARES>;
+			reset-names = "mac", "rx", "tx";
+		};
+
+		wan: port@6 {
```
openwrt-ai, line 221: nit: the commit message's validation paragraph and the PR body describe the same exact-commit run but disagree — the message says "LAN4 established a 1 Gbps link" and that "Physical key actuation, the remaining Ethernet ports and WAN speeds were not tested in this run", while the PR body reports LAN3 (not LAN4) at 1 Gbps, WAN at 1 Gbps full duplex including a cable-down/replug cycle, and short physical presses on both keys. Only the commit message survives into the merged history, so it is the copy worth reconciling.

---

openwrt-ai (commented):
Reviewed 1 new commit.

---

ikm-san:
Rebased onto current upstream `main` at `928bb37350d20a2a7737e3a3b5329261572ca533`.

New PR head: `4faa8e9a1c9dbd492efe18a91f953924ca4e639a`.

The four intervening upstream commits only touch Realtek DSA code. The LN6001 source diff remains identical to the hardware-tested `f2fbe5c802ce625d76b393696b159a9b82feee8f` patch (same tree and five changed paths); the only commit-message adjustment replaces the stale wording “built from this exact commit” with “built from this device patch.” I did not rebuild or reboot the router for this base-only rebase.

Re-run results on the rebased head: strict checkpatch 0 errors/0 warnings, `git diff --check` clean, both changed shell files parse cleanly, sanitation audit clean, and all three GitHub FormalityCheck jobs green.

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
ikm-san: Good catch, thank you. The commit message had retained an older validation snapshot. I amended the single commit at `7e38472870d04fb313287606405ebd8e0ad1b996` to distinguish the two test phases and record the later results: LAN4 reached 1 Gbps on the tested exact-ID artifact; LAN1-LAN3 were exercised on the preceding otherwise-identical image, with LAN3 at 1 Gbps and LAN1/LAN2 showing the separately tracked PSGMII reporting dependency. The message now also records WAN at 1 Gbps full duplex through cable-down/replug, all four visually checked LED channels, and short reset/WPS high-low-high transitions. The 2.5 Gbps WAN mode and PSGMII-dependent 100 Mbps/LAN-replug cases remain explicitly untested. The PR body now uses the same split. This is a commit-message-only correction; the source tree is unchanged.

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
ikm-san: Thanks, addressed in `cc6db21d22c42d13e67260a238d1c85674f719d9`. I copied the public upstream rationale comments verbatim and moved the unchanged `&ref_48mhz_clk`, `&xo_board_clk` and `&xo_clk` group to the end of the file, matching the sibling IPQ9574 layout. The values remain 1/1, 2/1 and 48 MHz; this changes only comments and DTS node ordering.

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
ikm-san: Follow-up cross-boot comparison is now complete. I compared a redacted ordered fingerprint of `lan1`-`lan4` and `wan` from the current RAM boot with the preceding cold RAM boot. The ordered fingerprints matched; all five interfaces reported `addr_assign_type=0`, and there was no random-MAC warning. No raw address or unit identity is being published.

This closes the promised cross-boot check. The offset-zero choice remains based on the separately observed stock `br-lan` and WAN base-value identities; it does not claim that stock assigns the base value to every physical LAN member.

openwrt-ai (commented):

---

ikm-san:
@robimarko, when you have time, could you please review this IPQ95xx device-support PR?

The current head has green Formality and qualcommbe/ipq95xx kernel/DTB CI, retail-hardware validation is documented, and all automated review feedback has been addressed. It is intentionally scoped to safe RAM-only support so the core board definition can be reviewed independently; persistent A/B installation, Wi-Fi and Bluetooth will be proposed separately after the base device support is accepted.

Please let me know if anything still blocks review or if you would prefer the branch rebased first.

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
```diff
@@ -0,0 +1,387 @@
...
+		led-upgrade = &led_blue;
+	};
+
+	chosen {
+		stdout-path = "serial0:115200n8";
+	};
+
+	memory@40000000 {
+		device_type = "memory";
+		reg = <0x0 0x40000000 0x0 0x40000000>;
```
robimarko, line 29: I doubt you need the memory node, U-Boot will patch that in

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
```diff
@@ -0,0 +1,387 @@
...
+	keys {
+		compatible = "gpio-keys";
+		pinctrl-0 = <&keys_pins>;
+		pinctrl-names = "default";
+
+		button-reset {
+			label = "reset";
+			linux,code = <KEY_RESTART>;
+			gpios = <&tlmm 40 GPIO_ACTIVE_LOW>;
+			debounce-interval = <60>;
```
robimarko, line 41: No debounce intervals, that is not needed

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
```diff
@@ -0,0 +1,387 @@
...
+		};
+
+		qca8075_3: ethernet-phy@13 {
+			reg = <0x13>;
+		};
+	};
+
+	qca8081: ethernet-phy@1c {
+		compatible = "ethernet-phy-id004d.d101";
+		reg = <0x1c>;
```
robimarko, line 123: Decimal please here

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
```diff
@@ -0,0 +1,387 @@
...
+		reset-assert-us = <40>;
```
robimarko, line 124: That's too short for sure

target/linux/qualcommbe/ipq95xx/base-files/lib/upgrade/platform.sh
```diff
@@ -5,7 +5,15 @@ RAMFS_COPY_BIN='fw_printenv fw_setenv head'
 RAMFS_COPY_DATA='/etc/fw_env.config /var/lock/fw_printenv.lock'
 
 platform_check_image() {
-	return 0;
+	case "$(board_name)" in
+	linksys,ln6001)
+		echo "Automatic sysupgrade is not supported on this device"
```
robimarko, line 10: Yeah this is not acceptable for a device accepted into OpenWrt

robimarko:
Until this is flashable and bootable on the device there is no point

ikm-san:
Thanks for the review, Robert. Updated in `9b8d01b30ef2`:

- Removed the fixed memory node; verified that U-Boot supplies 1 GiB RAM.
- Removed both debounce properties and retested short Reset/WPS presses.
- Changed the QCA8081 PHY address to decimal `28`.
- Increased reset assertion to 10,000 µs, following the public QCA8081 precedent in `904aa4386548`, while retaining the 150,000 µs deassertion delay. Cold-start and WAN cable-replug tests passed.
- Replaced the sysupgrade rejection with a flashable eMMC A/B installation path, including inactive-slot writes, readback verification, persistent overlay and boot confirmation.

Initial installation and settings-preserving upgrades were tested in both directions. The final revision also includes the SDCC controller-reset resource and passed B→A upgrade, repeated reboots and a physical cold start, with configuration retained.

Installation and recovery instructions are included in the commit message. Could you please take another look?

target/linux/qualcommbe/ipq95xx/base-files/etc/init.d/bootcount
```diff
@@ -0,0 +1,25 @@
...
+
+	magic="$(hexdump -v -n 4 -e '4/1 "%02x"' "$device")" || return 1
+	[ "$magic" = 11081120 ] || return 1
+	count="$(hexdump -v -s 4 -n 4 -e '4/1 "%02x"' "$device")" || return 1
+	[ "$count" != 00000000 ] || return 0
+
+	printf '\000\000\000\000' | dd of="$device" bs=1 seek=4 count=4 \
+		conv=notrunc 2>/dev/null || return 1
+	sync
+	[ "$(hexdump -v -s 4 -n 4 -e '4/1 "%02x"' "$device")" = 00000000 ]
```
openwrt-ai: `11081120` is the little-endian Linksys `BOOTCOUNT_MAGIC` (0x20110811), whose record is `{magic, count, checksum}` and is only accepted while [`checksum == magic + count`](https://github.com/openwrt/openwrt/blob/ec89d8d333817e85824a6a4f80fab25a269bd27f/package/system/mtd/src/linksys_bootcount.c#L132). Zeroing `count` without rewriting the checksum at offset 8 leaves a record that fails that check — [`mtd resetbc` sets both](https://github.com/openwrt/openwrt/blob/ec89d8d333817e85824a6a4f80fab25a269bd27f/package/system/mtd/src/linksys_bootcount.c#L173-L175).

```suggestion
	printf '\000\000\000\000' | dd of="$device" bs=1 seek=4 count=4 \
		conv=notrunc 2>/dev/null || return 1
	printf '\021\010\021\040' | dd of="$device" bs=1 seek=8 count=4 \
		conv=notrunc 2>/dev/null || return 1
	sync
	[ "$(hexdump -v -s 4 -n 4 -e '4/1 "%02x"' "$device")" = 00000000 ] &&
		[ "$(hexdump -v -s 8 -n 4 -e '4/1 "%02x"' "$device")" = 11081120 ]
```

---

target/linux/qualcommbe/ipq95xx/base-files/lib/upgrade/platform.sh
```diff
@@ -21,6 +32,9 @@ platform_do_upgrade() {
 		CI_DATAPART="rootfs_data"
 		emmc_do_upgrade "$1"
 		;;
+	linksys,ln6001)
+		ln6001_do_upgrade "$1" || exit 1
```
openwrt-ai: `exit 1` terminates `do_stage2` before its [`umount -a` / `reboot -f`](https://github.com/openwrt/openwrt/blob/ec89d8d333817e85824a6a4f80fab25a269bd27f/package/base-files/files/lib/upgrade/do_stage2#L18-L21), leaving the board wedged in the ramdisk with every service already killed. This path is reached on preflight/staging failure, where the selectors and the active slot are untouched, so rebooting onto the firmware it was running is the safe outcome; `ln6001_commit_failure` stays the one place that deliberately halts.

```suggestion
		ln6001_do_upgrade "$1" || return 1
```

---

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
```diff
@@ -0,0 +1,381 @@
...
+
+		qca8075_3: ethernet-phy@13 {
+			reg = <0x13>;
+		};
+	};
+
+	qca8081: ethernet-phy@1c {
+		compatible = "ethernet-phy-id004d.d101";
+		reg = <28>;
+		reset-assert-us = <10000>;
```
openwrt-ai: The commit message and PR body both report WAN cold-start/replug at 100 Mbps full duplex "with the revised QCA8081 reset delay", while [the 2026-09-02 report](https://github.com/openwrt/openwrt/pull/24736#issuecomment-5511499209) had the same port at 1 Gbps full duplex through a replug cycle with `reset-assert-us = <40>`. Is the drop attributable to the longer assertion, or just to a different link partner in the later run?

---

target/linux/qualcommbe/dts/ipq9554-linksys-ln6001.dts
```diff
@@ -0,0 +1,381 @@
...
+&sdhc_1 {
+	bus-width = <8>;
+	max-frequency = <384000000>;
+	mmc-ddr-1_8v;
+	mmc-hs200-1_8v;
+	mmc-hs400-1_8v;
+	mmc-hs400-enhanced-strobe;
+	pinctrl-0 = <&sdc_default_state>;
+	pinctrl-names = "default";
+	resets = <&gcc GCC_SDCC_BCR>;
```
openwrt-ai: nit: the SDCC block reset is a SoC-level resource, not a board one — `sdhc_1` lives in [`ipq9574.dtsi:452`](https://github.com/gregkh/linux/blob/8f3741e6feb045da5b406df0a80b42a1adfb289b/arch/arm64/boot/dts/qcom/ipq9574.dtsi#L452-L470) and `8devices,kiwi-dvk` / `askey,sbe1v1k` drive the same controller without it. Would this be better placed as a `patches-6.18` hunk on the dtsi so every ipq9574 board gets the reset path?

---

target/linux/qualcommbe/image/ipq95xx.mk
```diff
@@ -30,6 +30,19 @@ define Device/askey_sbe1v1k
...
+	$(call Device/FitImage)
+	DEVICE_VENDOR := Linksys
+	DEVICE_MODEL := LN6001
+	DEVICE_DTS := ipq9554-linksys-ln6001
+	SOC := ipq9554
+	DEVICE_PACKAGES += kmod-leds-pwm kmod-fs-f2fs mkf2fs f2fsck
+	IMAGES := sysupgrade.bin
+	IMAGE/sysupgrade.bin/squashfs := append-rootfs | pad-to 64k | \
+		check-size 128m | sysupgrade-tar rootfs=$$$$@ | append-metadata
```
openwrt-ai: nit: `IMAGES := sysupgrade.bin` is already set by [`Device/Default`](https://github.com/openwrt/openwrt/blob/ec89d8d333817e85824a6a4f80fab25a269bd27f/target/linux/qualcommbe/image/Makefile#L11), and the recipe below only differs from [`Device/EmmcImage`](https://github.com/openwrt/openwrt/blob/ec89d8d333817e85824a6a4f80fab25a269bd27f/target/linux/qualcommbe/image/Makefile#L30) by the `check-size`. Both siblings in this file pull the template in with `$(call Device/EmmcImage)`.

```suggestion
	DEVICE_PACKAGES += kmod-leds-pwm kmod-fs-f2fs mkf2fs f2fsck
	IMAGE/sysupgrade.bin/squashfs := append-rootfs | pad-to 64k | \
		check-size 128m | sysupgrade-tar rootfs=$$$$@ | append-metadata
```

---

package/boot/uboot-tools/uboot-envtools/files/qualcommbe_ipq95xx
```diff
@@ -12,6 +12,9 @@ case "$board" in
 askey,sbe1v1k)
 	ubootenv_add_mmc "0:APPSBLENV" "" "0x0" "0x40000" "0x40000"
 	;;
+linksys,ln6001)
+	ubootenv_add_mmc "u_env" "" "0x0" "0x40000" "0x40000"
+	;;
```
openwrt-ai: nit: envtools board entries plus the `PKG_RELEASE` bump are conventionally a separate commit scoped `uboot-envtools: ...`, e.g. [`04308ad`](https://github.com/openwrt/openwrt/commit/04308ad2845e1876a78e2f2bb0b1a0da1b9c8610) and [`c07b552`](https://github.com/openwrt/openwrt/commit/c07b55247090795c5f6d6cbd074ac26a331932a4). Splitting also keeps a rebuild that affects every target out of a `qualcommbe:`-scoped commit.

---

openwrt-ai (commented):
Reviewed 1 new commit.

---

PR #24554, opened 2026-08-03
realtek: add rtl960x DSA switch driver

zectaiga:
Adds a DSA driver for the RTL9607C switch core found in RTL960x SoCs.

The corresponding on-chip GMAC conduit and internal MDIO drivers have been merged through #24737 and #24840. This PR contains only the DSA switch driver.

Together with those merged drivers, this PR provides the driver support required by #20064, which adds the `RTL960X` platform symbol, DTS nodes, and target configuration. Until #20064 lands, `NET_DSA_RTL960X` remains unselectable.

This PR contains a single commit:

- `realtek: add rtl960x DSA switch driver`: switch bring-up, bridge/STP/VLAN offload, and FDB add/del/dump.

Thanks to @jameywine and @ProMix0 for their contributions and reviews in jameywine/openwrt#10.

Hardware-tested with the merged GMAC and MDIO drivers and the #20064 platform changes applied locally:

- clean GMAC/DSA probe, with the four physical internal GPHYs at MDIO addresses 0–3 and no PHY at address 4
- link at 1 Gbps/full duplex and bidirectional traffic on all three LAN-configured user ports, with no interface errors or drops
- VLAN-unaware and VLAN-aware bridges, including mixed tagged/untagged membership of one VID, live `vlan_filtering` changes, and host VLAN add/delete
- standalone ports and priority-tagged ingress
- dynamic FDB learning and static FDB add/dump/delete with hardware forwarding

PR #20064, opened 2025-09-16
realtek: add a new subtarget rtl9607c/rtl8198d with basic support

jameywine:
This patch/commit adds a new subtarget RTL9607C/RTL8198D to realtek platform with some initial support.

I am a simple hobbyist and since it is my first time contributing to this project, i would appreciate all feedback and advice from you to fix any kind of mistakes here and whatnot.
 
RTL9607C and RTL8198D both are identical dual core SoCs based on MIPS InterAptiv. RTL9607C is found in many ONTs with XPON capabilities while RTL8198D is found in many typical wireless routers.
The most recent GPL source code, which is based linux 5.10.x version, contains all the necessary drivers for these SoCs and the Realtek SDK for it. I have uploaded it to my github: https://github.com/jameywine/GPL-for-GP3000.

Surprisingly, it has a lot in common with RTL931X SoC, they both are interAptiv, use GIC and so most of the dtsi, config and makefile was copied from it. I've added a chip info for this RTL9607C/RTL8198D in prom.c and patch to include them as a config option.

**EDIT**: A few commits have been added since this post to introduce rtl960x support in different drivers:

1. I2C controller is supported.
2. Thermal sensor is supported.
3. Clock controller is supported.

Bootloader on these machines are based on Uboot 2020.01 with Bismarck Preloader and have tftp client capabilities which make it easy to experiment with. Rather interesting cause other Realtek SoCs have Realtek specific bootloader.

At this current moment, only initramfs booting on rtl9607c and rtl8198d based machines is supported. 

Flashing sysupgrade image should be possible thanks to existing `realtek,snand` and `realtek,ecc` drivers but the proper build recipe need to be crafted. Firmware needs to be tar archive with rootfs.squashfs and kernel files and a way to determine which ubi volume/mtd partition to flash to.

Instructions to boot initramfs from bootloader:
1. Connect to UART on the board and power on the machine.
2. Press any key in the first 2 seconds to interrupt the boot.
3. Enable tftp server on your device from which you want to transfer firmware from, with an ip set to ipaddr env variable.
4. Connect the ethernet cable to the rightmost LAN port (LAN1 or LAN4).
5. Run the command "tftp initramfs-kernel.bin".
6. After it is loaded, type "bootm" to boot into initramfs.

PR #24737, opened 2026-08-15
realtek: add rtl960x Ethernet (GMAC) driver

zectaiga:
Adds the RTL960x on-chip GMAC driver and moves the existing RTL838x Ethernet driver under the kernel Realtek vendor directory.

This driver was previously part of #24554 and is now split into its own series. Its review history is in that PR and in jameywine/openwrt#10.

This PR is a prerequisite for #20064, which adds the `RTL960X` platform symbol, DTS nodes, and target configuration. Until that lands, `NET_RTL960X_GMAC` remains unselectable. The corresponding DSA switch driver is tracked separately in #24554, which is also a prerequisite for #20064.

The series contains two commits:

- `realtek: move rtl838x Ethernet driver under realtek/`: moves the existing driver without source changes. The only non-mechanical hunk extends `NET_VENDOR_REALTEK` to `MACH_REALTEK_RTL`, so the driver stays visible on the SoC targets, which depend on neither PCI nor parport.
- `realtek: add rtl960x Ethernet (GMAC) driver`: adds the combined-NAPI RX/TX datapath, translates `rtl_otto` TX trailers to descriptor port steering, and passes the RX source port through `METADATA_HW_PORT_MUX`.

Thanks to @jameywine and @ProMix0 for their contributions and reviews in jameywine/openwrt#10.

Tested with #20064, the DSA driver from #24554, and local DTS/image support for the test device applied:

- `realtek/rtl838x` target compile after the RTL838x relocation
- RTL960x target compile/install and initramfs image generation
- clean GMAC/DSA probe and enumeration of all four internal PHYs
- link at 1 Gbps/full duplex and bidirectional traffic on all three LAN-configured user ports, with no interface errors or drops


target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
```diff
@@ -0,0 +1,818 @@
...
+			 * skb must stay pristine here -- it is retried as-is, so the DSA
+			 * trailer (pulled off below) must not be removed before we know
+			 * the frame is actually going on the ring.
+			 */
+			netif_stop_queue(dev);
+			return NETDEV_TX_BUSY;
+		}
+
+		/* Convert the rtl_otto trailer to descriptor port steering. */
+		if (netdev_uses_dsa(dev) && skb->len >= DSA_TRAILER_LEN) {
```
openwrt-ai, line 485: The trailer is matched on `netdev_uses_dsa()` alone, without checking that the conduit is actually running the `rtl_otto` tagger. A conduit's tagging protocol is runtime-switchable (`/sys/class/net/<conduit>/dsa/tagging`), so with any other tagger selected this still matches on the 3-byte `ab cd ef` magic and silently `skb_trim()`s 4 bytes off a frame that has no rtl_otto trailer.

The sibling in-tree driver guards exactly this: [`rteth_get_dsa_port()` at rtl838x_eth.c:1197](``https://github.com/openwrt/openwrt/blob/11332c7f63b8c5ce173e34c80dc704b08b783f78/target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl838x_eth.c#L1197-L1211)`` checks `dev->dsa_ptr->tag_ops->proto == DSA_TAG_PROTO_RTL_OTTO` before looking at the magic bytes.

```suggestion
		if (netdev_uses_dsa(dev) &&
		    dev->dsa_ptr->tag_ops->proto == DSA_TAG_PROTO_RTL_OTTO &&
		    skb->len >= DSA_TRAILER_LEN) {
```

The RX side has the same gap — `bool dsa = netdev_uses_dsa(dev);` on line 293 attaches a `METADATA_HW_PORT_MUX` dst regardless of which tagger is active, which would confuse a non-rtl_otto `rcv()`.

---

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.h
```diff
@@ -0,0 +1,237 @@
...
+#define CPUTAGCR_PROTOCOL_VAL	GENMASK(7, 0)
+#define CPUTAGCR_SIZE_8B	2
+#define CPUTAGCR_FMT_APOLLOPRO	8
+#define CPUTAGCR_PROTO_MASK_8370	0xff
+#define CPUTAGCR_PROTO_8370	0x04
+
+#define GMAC_CONFIG		0x4c
+#define CONFIG_RFIFO_SIZE	GENMASK(29, 28)
+#define CONFIG_EN_INTR_SPLIT	BIT(24)
+#define CONFIG_RX_SIDEBAND	GENMASK(23, 22)
```
openwrt-ai: These three bit definitions land in the `CONFIG_` namespace, which the kernel reserves for Kconfig symbols. Anything that scans sources for `CONFIG_*` (e.g. `scripts/checkkconfigsymbols.py`) will report them as undefined config options, and a future Kconfig symbol with one of these names would collide outright.

Renaming the register to `GMAC_CFG` keeps the file's own "register `GMAC_X` → bits `X_*`" convention while getting out of the reserved prefix:

```suggestion
#define GMAC_CFG		0x4c
#define CFG_RFIFO_SIZE		GENMASK(29, 28)
#define CFG_EN_INTR_SPLIT	BIT(24)
#define CFG_RX_SIDEBAND		GENMASK(23, 22)
```

None of the four are referenced from `rtl960x_gmac.c`, so dropping the block entirely is also fine.

---

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.h
```diff
@@ -0,0 +1,237 @@
+/* SPDX-License-Identifier: GPL-2.0-only */
+
+#ifndef __RTL960X_GMAC_H
+#define __RTL960X_GMAC_H
```
openwrt-ai, line 4: nit: the header is not self-contained — it uses `BIT()`/`GENMASK()`, `FIELD_PREP()` (in `TCR_CONFIG`) and `u32` (in the descriptor structs) but includes nothing. It only compiles because `rtl960x_gmac.c` happens to include `<linux/bitfield.h>` before it.

```suggestion
#ifndef __RTL960X_GMAC_H
#define __RTL960X_GMAC_H

#include <linux/bitfield.h>
#include <linux/bits.h>
#include <linux/types.h>
```

---

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.h
```diff
@@ -0,0 +1,237 @@
...
+/* Descriptor fields */
+#define DESC_OWN		BIT(31)
+#define DESC_EOR		BIT(30)
+#define DESC_FS			BIT(29)
+#define DESC_LS			BIT(28)
+#define DESC_LEN_MASK		0xfff
+#define TX_DESC_CRC		BIT(23)
+
+/* CPU-tag sideband fields in descriptor opts2/opts3 */
+#define RX_OPTS3_SRC_PORT(o3)	(((o3) >> 16) & 0xf)
```
openwrt-ai: nit: hand-rolled shift/mask here while every other multi-bit field in this header is a `GENMASK()` consumed with `FIELD_PREP()`/`FIELD_GET()` (`CPUTAGCR_*`, `IO_CMD_*`, `TCR_IFG`, …). Same for `TX_OPTS2_PORTMASK(pm)` on line 214 and `DESC_LEN_MASK 0xfff` on line 208.

Declaring them as `GENMASK(19, 16)` / `GENMASK(26, 16)` / `GENMASK(11, 0)` and using `FIELD_GET`/`FIELD_PREP` at the call sites in `rtl960x_gmac.c` would make the whole register description uniform (and gets the compile-time range checking that `FIELD_PREP` does).

---

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
```diff
@@ -0,0 +1,818 @@
...
+ */
+#define DSA_TRAILER_LEN		4
+
+/* Datapath sizing */
+#define RX_RING_SIZE		128	/* must be a power of two, 16..256 */
+#define TX_RING_SIZE		128
+#define RX_BUF_SIZE		1536	/* 0x600, per U-Boot RX_DESC_BUFFER_SIZE */
+#define RX_SHIFT		2	/* HW writes the frame at buf + 2 bytes */
+#define RX_MAX_FRAME_LEN	(RX_BUF_SIZE - RX_SHIFT)
+#define TX_MIN_LEN		60
```
openwrt-ai: nit: bare `60` where the file already uses the kernel constant for the same quantity — the RX path on line 325 spells the minimum as `ETH_ZLEN + ETH_FCS_LEN`.

```suggestion
#define TX_MIN_LEN		ETH_ZLEN
```

---

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
```diff
@@ -0,0 +1,818 @@
+// SPDX-License-Identifier: GPL-2.0-only
+/* Realtek RTL9607C ethernet driver */
```
openwrt-ai: nit: this file-header line names a single chip, while everything else describing the same driver is family-scoped: `MODULE_DESCRIPTION("Realtek RTL960x/LUNA GMAC driver")` (line 816), the Kconfig prompt `"Realtek RTL960X Ethernet MAC support"` in `717-net-ethernet-add-rtl960x-gmac-probe-driver.patch`, the probe message `"RTL960x/LUNA GMAC %pR irq %d"`, and the filename itself. (`RTL9607C` is still right for the compatible and the `RTL9607C_*_CONFIG` constants — those are genuinely chip-specific.)

```suggestion
/* Realtek RTL960x/LUNA GMAC driver */
```

---

openwrt-ai (commented):
Reviewed 2 commits (fresh review, full PR diff).

Both commit messages match their diffs. Checked the regenerated `716-*.patch` hunk offsets against `drivers/net/ethernet/realtek/{Kconfig,Makefile}` in v6.18.44 (`@@ -6,7`, `@@ -140,4`, `@@ -10,3`) and the `717-*.patch` offsets against the post-716 tree (`@@ -146,4`, `@@ -11,3`) — all line up. `CONFIG_NET_VENDOR_REALTEK=y` already comes from `target/linux/generic/config-6.18`, so moving `NET_RTL838X` inside `if NET_VENDOR_REALTEK` needs no subtarget config churn. The `RTL9607C_IO_CMD_CONFIG` field encoding does add up to the `0xd71f31f0` quoted in the commit message once `IO_CMD_RX_ENABLE|IO_CMD_TX_ENABLE` are OR'd in.

One substantive comment on the TX-side DSA trailer detection; the rest are marked `nit:`.

---

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
```diff
@@ -0,0 +1,818 @@
...
+
+	regmap_write(g->map32, GMAC_IDR0,
+		     (a[0] << 24) | (a[1] << 16) | (a[2] << 8) | a[3]);
+	regmap_write(g->map32, GMAC_IDR4, (a[4] << 24) | (a[5] << 16));
+}
+
+static void rtl960x_gmac_set_rx_mode(struct net_device *dev)
+{
+	u32 rcr = RCR_ACCEPT_BROADCAST | RCR_ACCEPT_MYPHYS;
+	struct rtl960x_gmac *g = netdev_priv(dev);
```
plappermaul: Please use something better than just "g" for the private driver structure. Realtek Target uses a mix of priv/ctrl in the drivers.

Kernel uses:

```
 $ grep -r -h -oP '\w+(?=\s*=\s*netdev_priv)' . | sort | uniq -c | sort -nr
   1896 priv
    931 adapter
    566 np
    500 lp
    454 bp
    316 port
    240 tp
    174 pi
    163 pdata
    153 fep
    122 edev
    120 nic_dev
    107 efx
```

plappermaul (changes requested):

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
ProMix0: Already discussed it here - https://github.com/openwrt/openwrt/pull/24554#discussion_r3717888488

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.h
zectaiga: Removed the unused register and bit definitions.

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.h
zectaiga: Applied. The source keeps its own `<linux/bitfield.h>` include because it also uses `FIELD_GET()` and `FIELD_PREP()` directly.

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.h
zectaiga: Applied. The three fields now use `GENMASK()`.

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
zectaiga: Changed TX_MIN_LEN to ETH_ZLEN.

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
zectaiga: Updated.

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
zectaiga: Renamed the private driver pointer to `priv` throughout the driver.

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
jameywine: @zectaiga please keep the file header as `/* Realtek RTL960X ethernet driver */` and instead remove mention of LUNA as it is realtek codename stuff and we don't care about that.

zectaiga: Updated. I restored the file header and removed the `LUNA` and `RTL8198D` references.

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
```diff
@@ -0,0 +1,819 @@
...
+			d->opts2 = TX_OPTS2_CPUTAG |
+				   FIELD_PREP(TX_OPTS2_PORTMASK, BIT(dest_port));
+			d->opts3 = TX_OPTS3_KEEP | TX_OPTS3_DISLRN | TX_OPTS3_L34_KEEP;
+		} else {
+			d->opts2 = 0;
+			d->opts3 = 0;
+		}
+		d->opts4 = 0;
+		opts1 = DESC_OWN | DESC_FS | DESC_LS | TX_DESC_CRC | len |
+			(last ? DESC_EOR : 0);
```
openwrt-ai: nit: `DESC_LEN_MASK` is now a `GENMASK(11, 0)` field read back with `FIELD_GET()` on the RX side (line 323), but all three descriptor *writes* still OR the raw length in. Functionally identical today because the field sits at bit 0, but it means only the read site follows the mask definition.

```suggestion
		opts1 = DESC_OWN | DESC_FS | DESC_LS | TX_DESC_CRC |
			FIELD_PREP(DESC_LEN_MASK, len) | (last ? DESC_EOR : 0);
```

Same applies to `d->opts1 = DESC_OWN | RX_BUF_SIZE;` on line 207 and `d->opts1 = DESC_OWN | RX_BUF_SIZE | (last ? DESC_EOR : 0);` on line 368.

---

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
```diff
@@ -0,0 +1,819 @@
+// SPDX-License-Identifier: GPL-2.0-only
+/* Realtek RTL960X ethernet driver */
```
openwrt-ai, line 2: nit: the naming cleanup in this round standardised everything the commit adds on `RTL960X` — this header, `MODULE_DESCRIPTION("Realtek RTL960X Ethernet driver")`, the probe message, the Kconfig prompt/help and the `717-*.patch` subject/body — but the OpenWrt commit message body still says `RTL960x`:

> Add a driver for the RTL960x on-chip Gigabit MAC […] When paired with the rtl960x DSA switch driver

`RTL960X` is also the spelling already in tree (`drivers/clk/realtek/Kconfig`: *"as well as chips from the RTL960X series"*). Worth folding into the same amend so the whole series reads consistently. (`realtek: add rtl960x …` in the subject is fine — that's the OpenWrt scope prefix.)

---

openwrt-ai (commented):
Re-reviewed the force-push (`11332c7` → `6b5067b`). `35a6df7` is unchanged; the second commit's message is byte-identical to the previously reviewed version, so no new commit-message issues.

Verified the mask translations are equivalent to what they replaced:

- `RX_OPTS3_SRC_PORT(o3) = ((o3) >> 16) & 0xf` → `GENMASK(19, 16)` ✔
- `TX_OPTS2_PORTMASK(pm) = ((pm) & 0x7ff) << 16` → `GENMASK(26, 16)` ✔ (11 bits at offset 16)
- `DESC_LEN_MASK 0xfff` → `GENMASK(11, 0)` ✔

`FIELD_PREP(TX_OPTS2_PORTMASK, BIT(dest_port))` stays in range — `dest_port < RTL960X_CPU_PORT` (9), so `BIT(8)` is the widest value and fits the 11-bit field. `TX_MIN_LEN = ETH_ZLEN` resolves via the existing `<linux/etherdevice.h>` include. The removed `GMAC_CONFIG` / `CONFIG_RFIFO_SIZE` / `CONFIG_EN_INTR_SPLIT` / `CONFIG_RX_SIDEBAND` block has no remaining references anywhere under `files-6.18/`, and no `LUNA` / `RTL8198D` strings are left in the driver or the patch. The `g` → `priv` rename is complete — every `netdev_priv()` / `container_of()` / callback site now uses `priv`, with no stragglers. It does push eight lines to 81–83 columns, which is well inside the kernel's 100-column limit and in line with the neighbouring `rtl838x_eth.c`.

Only two `nit:`-level items this round. The open thread on the TX DSA-trailer / tagger check is left as-is — @ProMix0 pointed at the prior discussion in #24554, so I'm not re-raising it.

---

plappermaul:
@jonasjelonek will merge this after final AI review comments are addressed. Fine from your side too?

target/linux/realtek/patches-6.18/716-net-ethernet-add-support-for-rtl838x-ethernet.patch
```diff
@@ -16,33 +16,36 @@ Submitted-by: Sander Vanheule <sander@svanheule.net>
...
- source "drivers/net/ethernet/rocker/Kconfig"
-+
+--- a/drivers/net/ethernet/realtek/Kconfig
++++ b/drivers/net/ethernet/realtek/Kconfig
+@@ -6,7 +6,7 @@
+ config NET_VENDOR_REALTEK
+ 	bool "Realtek devices"
+ 	default y
+-	depends on PCI || (PARPORT && X86)
++	depends on PCI || (PARPORT && X86) || MACH_REALTEK_RTL
```
jonasjelonek, line 30: Several PCI cards are gated by that `NET_VENDOR_REALTEK` symbol upstream. If you add `MACH_REALTEK_RTL` here, doesn't that inevitable pull the Realtek target stuff in for completely unrelated devices?

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
```diff
@@ -0,0 +1,819 @@
...
+
+#include <linux/bitfield.h>
+#include <linux/cleanup.h>
+#include <linux/delay.h>
+#include <linux/dma-mapping.h>
+#include <linux/etherdevice.h>
+#include <linux/ethtool.h>
+#include <linux/interrupt.h>
+#include <linux/mod_devicetable.h>
```
jonasjelonek: Can this be replaced or dropped? Upstream seems to work hard towards removing the direct inclusion of `linux/mod_devicetable.h`. See e.g. https://lore.kernel.org/all/ca270a534d0f230a939a3fb4a661808b35d6436d.1783329817.git.u.kleine-koenig@baylibre.com/

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
```diff
@@ -0,0 +1,819 @@
...
+	.probe = rtl960x_gmac_probe,
+	.driver = {
+		.name = "rtl960x-gmac",
+		.of_match_table = rtl960x_gmac_of_match,
+	},
+};
+module_platform_driver(rtl960x_gmac_driver);
+
+MODULE_DESCRIPTION("Realtek RTL960X Ethernet driver");
+MODULE_AUTHOR("Taiga Ogawa");
```
jonasjelonek: This should also carry an email address in <> behind your name.

jonasjelonek (commented):
Parts of the code would look less crowded if taking advantage of the ~100 chars limit instead of the legacy 80 limit. But that's just personal preference.

target/linux/realtek/patches-6.18/716-net-ethernet-add-support-for-rtl838x-ethernet.patch
ProMix0: Doesn't look like so, these drivers have individual config symbols

jonasjelonek: I think I mixed that with `select`. Seems ok then.

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
zectaiga: Applied FIELD_PREP(DESC_LEN_MASK, ...) to all three descriptor length writes.

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
zectaiga: Updated both occurrences in the commit body to `RTL960X`.

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
zectaiga: Replaced it with `<linux/of.h>`, since the driver defines an `of_device_id` match table.

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
zectaiga: Added.

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.h
jameywine: please don't drop these defines and instead just follow the suggested change to have them as CFG.

some are gonna be used when multi ring support is added like EN_INTR_SPLIT.

however, i don't have any insight on others. For example, the CFG_RFIFO_SIZE is set to 2 by re8686_rtl9607c nic driver which seems to set the gmac_rxfifo size to 2KB.
The CFG_RX_SIDEBAND set to 3 during hardware initialization, no clue what could mean. 
And there is also whatever config_tx_jumbo function which touches bit 16 of this register.

plappermaul: Just bring them back when needed.

jameywine: I am fine with bringing back EN_INTR_SPLIT part later but at least i very much would like to see @zectaiga add the missing CFG_RFIFO_SIZE part to rtl960x_gmac_init_hw as the original [re8686_rtl9607c wrote 2 to it during the same initialization](https://github.com/jameywine/GPL-for-GP3000/blob/5090fbcb6ba743dd7b1314811ef557bad0460147/linux-5.10.x/drivers/net/ethernet/realtek/rtl86900/nicDriver/re8686_rtl9607c.c#L5615).

openwrt-ai (commented):

Re-reviewed the amend (`6b5067b` → `92d9a60`). `35a6df7` is untouched; the second commit's message differs only by the two `RTL960x` → `RTL960X` body fixes, which matches what it says it does. The code delta is 7 insertions / 6 deletions and every hunk is a direct answer to an open thread:

- `FIELD_PREP(DESC_LEN_MASK, ...)` now applied at all three descriptor-length writes (lines 207, 368, 520). Behaviourally identical — `DESC_LEN_MASK` is `GENMASK(11, 0)`, so shift 0 — and both constant sites are in range for the `FIELD_PREP` build-time check (`RX_BUF_SIZE` = 1536 < 4096). The runtime TX site is bounded by `dev->max_mtu = RX_MAX_FRAME_LEN - ETH_HLEN - ETH_FCS_LEN` (1516), so `len` stays well inside 12 bits and the added masking never truncates a real frame.
- `<linux/mod_devicetable.h>` → `<linux/of.h>`: the only consumer is the `of_device_id` match table plus `MODULE_DEVICE_TABLE(of, ...)`, and `of.h` pulls in `mod_devicetable.h`, so nothing else needed adjusting. The new include also lands in the right alphabetical slot between `netdevice.h` and `platform_device.h`.
- `MODULE_AUTHOR("Taiga Ogawa <zectaiga@gmail.com>")` matches the `Signed-off-by:` and the surrounding upstream convention in `drivers/net/ethernet/realtek/` (`atp.c`, `8139cp.c`, `r8169_main.c` all carry an address).

CI on `92d9a60` is green across all six realtek subtarget kernel builds plus the patch and formality checks, so nothing to flag there.

Two open human threads remain outside this round's scope and I'm not re-raising either: the TX/RX tagger check (@ProMix0 pointed at the prior discussion in #24554) and @jameywine's request to write `CFG_RFIFO_SIZE` in `rtl960x_gmac_init_hw`.

---

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.h
zectaiga: Added. The driver now programs `CFG_RFIFO_SIZE` to `CFG_RFIFO_SIZE_2KB (2)` during hardware initialization, preserving the other fields in `GMAC_CFG`. The field read back as `2` on hardware, with a 1 Gbps link and bidirectional traffic working normally.

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
```diff
@@ -0,0 +1,822 @@
...
+	/*
+	 * Keep RXJUMBO set like the vendor driver always does; the accept
+	 * cutoff with the bit cleared is unvalidated on this hardware, and
+	 * oversize frames are dropped by the FS|LS + length check in the
+	 * RX path anyway.
+	 */
+	regmap_write(priv->map8, GMAC_CMD, CMD_RXCHKSUM | CMD_RXJUMBO);
+	regmap_write(priv->map32, GMAC_TCR, TCR_CONFIG);
+	regmap_update_bits(priv->map32, GMAC_CFG, CFG_RFIFO_SIZE,
+			   FIELD_PREP(CFG_RFIFO_SIZE, CFG_RFIFO_SIZE_2KB));
```
openwrt-ai, line 248: nit: the commit message wasn't amended along with this hunk, and it's the one paragraph that enumerates exactly this class of change:

> Align the DMA burst size, early TX, TX FIFO threshold, and RX interrupt trigger with the vendor driver. Keep the shorter RX packet timer because the vendor field value reproducibly added about 0.9 ms of ingress latency.

The new `CFG_RFIFO_SIZE` = 2KB programming is another vendor-driver alignment (per [re8686_rtl9607c.c:5615](``https://github.com/jameywine/GPL-for-GP3000/blob/5090fbcb6ba743dd7b1314811ef557bad0460147/linux-5.10.x/drivers/net/ethernet/realtek/rtl86900/nicDriver/re8686_rtl9607c.c#L5615)`` as @jameywine noted), so listing the RX FIFO size in that sentence would keep the message a complete description of the hardware setup the commit performs. Everything else in `rtl960x_gmac_init_hw` that isn't self-evident (the `CMD_RXJUMBO` retention, the CPU-tag parsing mode, the `IO_CMD` value) is accounted for there.

No functional concern with the hunk itself: `GMAC_CFG` at `0x4c` is 4-byte aligned for `map32`'s `reg_stride`, slots into the header in offset order between `GMAC_CPUTAGCR` (`0x48`) and `GMAC_CPUTAG1CR` (`0x50`), `FIELD_PREP(GENMASK(29, 28), 2)` is in range, and the read-modify-write preserves the `EN_INTR_SPLIT` / `RX_SIDEBAND` / TX-jumbo bits that were deliberately left undefined for now.

---

openwrt-ai (commented):
Re-reviewed the amend (`92d9a60` → `3625932`). `35a6df7` is untouched and the second commit's message is byte-identical to the previously reviewed version, so there are no new commit-message-vs-diff mismatches to report beyond the one nit inline.

The whole delta is 6 added lines answering @jameywine's [request](https://github.com/openwrt/openwrt/pull/24737#discussion_r3791904145) to program `CFG_RFIFO_SIZE` in `rtl960x_gmac_init_hw`:

- Header: `GMAC_CFG 0x4c` / `CFG_RFIFO_SIZE GENMASK(29, 28)` / `CFG_RFIFO_SIZE_2KB 2`. Offset ordering is correct (between `GMAC_CPUTAGCR` `0x48` and `GMAC_CPUTAG1CR` `0x50`), the `CFG_*` prefix follows the file's "register `GMAC_X` → bits `X_*`" convention and stays out of the reserved `CONFIG_` namespace, and `CFG_RFIFO_SIZE_2KB` matches the existing enum-value naming (`TCR_IFG_SHORT`, `CPUTAGCR_SIZE_8B`). No `GMAC_CFG` / `RFIFO` collision anywhere else under `target/linux/realtek/`.
- Source: `regmap_update_bits` rather than a full write, so `EN_INTR_SPLIT`, `RX_SIDEBAND` and the TX-jumbo bit at 16 keep their reset/bootloader values — consistent with @plappermaul's "just bring them back when needed". `0x4c` is 4-byte aligned for `map32`'s `reg_stride = 4`, `FIELD_PREP(GENMASK(29, 28), 2)` yields `0x20000000` and passes the compile-time range check, and the write lands after `stop_hw()` but before `IO_CMD` sets `IO_CMD_RX_ENABLE`, so the FIFO size is in place before RX starts. It's also within the `0x100` ethtool `get_regs` dump window, so the field is observable.

No checks have reported on `3625932` yet (`total_count: 0`), so nothing CI-grounded to flag.

The two open human threads are unchanged and out of scope for this round — the TX/RX tagger check (@ProMix0 pointed at the prior discussion in #24554) and the `CFG_EN_INTR_SPLIT` / `CFG_RX_SIDEBAND` defines, which @plappermaul deferred.

---

target/linux/realtek/files-6.18/drivers/net/ethernet/realtek/rtl960x_gmac.c
zectaiga: Updated the commit message to include the RX FIFO size.

plappermaul:

jameywine:
Awesome, thanks a lot!

Next order of business should probably be support for multiple rings and use of page pool.

And some things need be changed to be more in line with the linux re8686_rtl9607c nic driver as most of the things were copied over from uboot ethernet driver. Like for example, in uboot RX buf size is 0x600 but in re8686_rtl9607c driver, it is 1600 and I belive the latter to be more correct.

There is also a missing setting of netdev features and our GMAC should be quite capable similarly to r8169 chips.

PR #24739, opened 2026-08-15
ipq40xx: fix up UTR softpwm backlight

divinehawk:
Correct brightness table - these values come from the stock firmware. This was a transcription error in the original commit.

Rename the soft-pwm node to pwm so it meets pwm-gpio.yaml validation. 

Drop unneeded CONFIG_PWM_GPIO=m from kernel configuration. It gets pulled in through DEVICE_PACKAGES.

Fixes: e4de94870370 ("kernel: add softpwm/gpio-pwm backlight support for UTR")


robimarko:

PR #38, opened 2016-07-21
sunxi: add support for Orange Pi 2 (H3)

qinfengling:
- Add support Orange Pi 2
- Update 011-dt-sync-dts-files-with-kernel.patch to OpenWrts preferred format

Format links:https://wiki.openwrt.org/doc/devel/patches#prepare_quilt_configuration

Test report:Compile successful and boot with Orange Pi 2 successfully

Know issue: The network can't bring up, because the kernel driver is a bit older. It should work at least  on kernel version 4.5.1

@wigyori Could you help me to review and merge this little patch ? Thanks.

PR #24740, opened 2026-08-15
uhttpd: run in ujail as an unprivileged user

JuliusBairaktaris:
Splits out of #24558, which now carries pppd and odhcpd only. uhttpd needs root
only to bind 80/443 and read the TLS key, then keeps it for the process
lifetime. This runs it in a ujail with `CAP_NET_BIND_SERVICE` for the bind and
`CAP_SETUID` + `CAP_SETGID` for the drop, and passes `-Z` so uhttpd drops to uid
456 once its listeners are up:

    package/network/services/uhttpd/files/uhttpd.init, start_instance():

        [ -x /sbin/ujail -a -e /etc/capabilities/uhttpd.json ] && {
                procd_append_param command -Z uhttpd
                procd_add_jail uhttpd
                procd_set_param capabilities /etc/capabilities/uhttpd.json
                procd_set_param no_new_privs 1
        }

Dropping inside the daemon rather than with procd's `user` parameter is what
keeps the key at `root:root 0600`. The key path is a uci value, so dropping
before `execve()` would mean chowning an operator-supplied path to the
unprivileged user. `-Z` does not exist in the pinned uhttpd, so it rides along
as `100-main-add-Z-to-drop-privileges.patch`, openwrt/uhttpd#40 unchanged, and
goes away at the next `PKG_SOURCE_VERSION` bump that picks it up.

ubusd denies every non-root uid, so `/usr/share/acl.d/uhttpd.json` grants uid
456 what LuCI reaches. It ships with `uhttpd-mod-ubus` rather than the base
package, which links no ubus library and opens no connection to grant.
`session/create`, `grant` and `revoke` are withheld so a compromised uhttpd
cannot mint itself a privileged session.

The other objects rpcd owns get a method wildcard, and my understanding is that
rc, rpc-sys and iwinfo carry no session check of their own, so that wildcard is
safe only under openwrt/rpcd#38, which is not merged:

    package/network/services/uhttpd/files/uhttpd_acl.json:

        "rc": {
                "methods": [
                        "*"
                ]
        },
        "rpc-sys": {
                "methods": [
                        "*"
                ]
        },

On an ipq807x router LuCI login and the overview, flash, network, wireless and
system pages load at uid 456, with no ACL denial and no respawn; each is a
single run. The CGI script `/www/cgi-bin/luci` inherits the server uid, so which
pages read `/etc/config` there rather than through rpcd is unchecked. I used AI
to help me write this change.

Dependencies, in merge order:

1. openwrt/uhttpd#40 - https://github.com/openwrt/uhttpd/pull/40 - adds `-Z`, carried here as a package patch until it merges.
2. openwrt/rpcd#38 - https://github.com/openwrt/rpcd/pull/38 - hard prerequisite. `rpc_file_access()` returns true for a request with no session id, so with the `file` grant this PR adds and #38 out, a non-root uhttpd can call `file exec` untokened. Not before this one.
3. openwrt/rpcd#40 - https://github.com/openwrt/rpcd/pull/40 - `stream: true` on `file exec`, so the backup archive comes back as a descriptor instead of inline.
4. openwrt/cgi-io#6 - https://github.com/openwrt/cgi-io/pull/6 - makes `cgi-backup` and `cgi-exec` dispatch through rpcd instead of forking with uhttpd's uid; until it lands, a backup taken at uid 456 silently omits root-owned files.
5. openwrt/luci#8947 - https://github.com/openwrt/luci/pull/8947 - the session ACL for `sysupgrade --create-backup -`, enforced once #6 dispatches through rpcd.

Would you rather have `rc` and `rpc-sys` named method by method here instead of
wildcarded?

### DIFF
--- a/package/network/services/uhttpd/Makefile
+++ b/package/network/services/uhttpd/Makefile
@@ -5,7 +5,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=uhttpd
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL=$(PROJECT_GIT)/project/uhttpd.git
@@ -32,6 +32,7 @@ endef
 define Package/uhttpd
   $(Package/uhttpd/default)
   DEPENDS:= +USE_GLIBC:libcrypt-compat +libubox +libblobmsg-json +libjson-script +libjson-c
+  USERID:=uhttpd=456:uhttpd=456
 endef
 
 define Package/uhttpd/description
@@ -91,6 +92,8 @@ define Package/uhttpd/install
 	$(INSTALL_CONF) ./files/uhttpd.config $(1)/etc/config/uhttpd
 	$(VERSION_SED_SCRIPT) $(1)/etc/config/uhttpd
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/uhttpd $(1)/usr/sbin/uhttpd
+	$(INSTALL_DIR) $(1)/etc/capabilities
+	$(INSTALL_DATA) ./files/uhttpd.capabilities $(1)/etc/capabilities/uhttpd.json
 endef
 
 define Package/uhttpd-mod-lua/install
@@ -102,6 +105,8 @@ define Package/uhttpd-mod-ubus/install
 	$(INSTALL_DIR) $(1)/usr/lib $(1)/etc/uci-defaults
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/uhttpd_ubus.so $(1)/usr/lib/
 	$(INSTALL_DATA) ./files/ubus.default $(1)/etc/uci-defaults/00_uhttpd_ubus
+	$(INSTALL_DIR) $(1)/usr/share/acl.d
+	$(INSTALL_DATA) ./files/uhttpd_acl.json $(1)/usr/share/acl.d/uhttpd.json
 endef
 
 define Package/uhttpd-mod-ubus/postinst
--- a/package/network/services/uhttpd/files/uhttpd.init
+++ b/package/network/services/uhttpd/files/uhttpd.init
@@ -122,6 +122,13 @@ start_instance()
 	procd_set_param stderr 1
 	procd_set_param command "$UHTTPD_BIN" -f
 
+	[ -x /sbin/ujail -a -e /etc/capabilities/uhttpd.json ] && {
+		procd_append_param command -Z uhttpd
+		procd_add_jail uhttpd
+		procd_set_param capabilities /etc/capabilities/uhttpd.json
+		procd_set_param no_new_privs 1
+	}
+
 	config_get config "$cfg" config
 	if [ -z "$config" ]; then
 		mkdir -p /var/etc/uhttpd
--- a/package/network/services/uhttpd/files/uhttpd_acl.json
+++ b/package/network/services/uhttpd/files/uhttpd_acl.json
@@ -1,0 +1,111 @@
+{
+	"user": "uhttpd",
+	"access": {
+		"session": {
+			"methods": [
+				"access",
+				"destroy",
+				"get",
+				"login",
+				"set"
+			]
+		},
+		"file": {
+			"methods": [
+				"*"
+			]
+		},
+		"iwinfo": {
+			"methods": [
+				"*"
+			]
+		},
+		"luci": {
+			"methods": [
+				"*"
+			]
+		},
+		"luci-rpc": {
+			"methods": [
+				"*"
+			]
+		},
+		"luci.*": {
+			"methods": [
+				"*"
+			]
+		},
+		"rc": {
+			"methods": [
+				"*"
+			]
+		},
+		"rpc-sys": {
+			"methods": [
+				"*"
+			]
+		},
+		"system": {
+			"methods": [
+				"board",
+				"info",
+				"reboot",
+				"validate_firmware_image"
+			]
+		},
+		"uci": {
+			"methods": [
+				"*"
+			]
+		},
+		"dsl": {
+			"methods": [
+				"metrics",
+				"statistics"
+			]
+		},
+		"fingerprint": {
+			"methods": [
+				"fingerprint"
+			]
+		},
+		"hostapd.*": {
+			"methods": [
+				"del_client",
+				"wps_cancel",
+				"wps_start",
+				"wps_status"
+			]
+		},
+		"log": {
+			"methods": [
+				"read"
+			]
+		},
+		"network": {
+			"methods": [
+				"get_proto_handlers"
+			]
+		},
+		"network.device": {
+			"methods": [
+				"status"
+			]
+		},
+		"network.interface": {
+			"methods": [
+				"dump"
+			]
+		},
+		"network.rrdns": {
+			"methods": [
+				"lookup"
+			]
+		},
+		"service": {
+			"methods": [
+				"list"
+			]
+		}
+	}
+}
--- a/package/network/services/uhttpd/files/uhttpd.capabilities
+++ b/package/network/services/uhttpd/files/uhttpd.capabilities
@@ -1,0 +1,19 @@
+{
+	"bounding": [
+		"CAP_NET_BIND_SERVICE",
+		"CAP_SETGID",
+		"CAP_SETUID"
+	],
+	"effective": [
+		"CAP_NET_BIND_SERVICE",
+		"CAP_SETGID",
+		"CAP_SETUID"
+	],
+	"ambient": [],
+	"permitted": [
+		"CAP_NET_BIND_SERVICE",
+		"CAP_SETGID",
+		"CAP_SETUID"
+	],
+	"inheritable": []
+}
--- a/package/network/services/uhttpd/patches/100-main-add-Z-to-drop-privileges.patch
+++ b/package/network/services/uhttpd/patches/100-main-add-Z-to-drop-privileges.patch
@@ -1,0 +1,101 @@
+From 9751afbab3c4aca6da54f589c5d460010f953a48 Mon Sep 17 00:00:00 2001
+From: Julius Bairaktaris <julius@bairaktaris.de>
+Date: Thu, 6 Aug 2026 15:21:13 +0200
+Subject: [PATCH] main: add -Z to drop privileges after binding sockets
+
+uhttpd currently requires root for the whole lifetime of the process,
+even though all privileged work (binding the listeners, reading the
+TLS key) happens during startup. Add -Z to drop to an unprivileged
+user once that startup is done, using initgroups/setgid/setresuid so
+no saved-root or supplementary-group state survives. Service managers
+that cannot use a sandbox (plain procd, systemd) can then run the
+daemon unprivileged on privileged ports.
+
+The drop precedes the handler plugin initialization because the ubus
+plugin connects to ubusd there, and ubusd reads the peer credentials
+once, when the connection is accepted: a socket opened before the drop
+stays a uid 0 connection for the life of the process, and ubusd exempts
+uid 0 from every ACL it enforces. Connecting afterwards is what puts
+the daemon's ubus access under /usr/share/acl.d. The plugins need no
+privilege of their own - they dlopen a module and read a handler script
+- whereas the TLS key is read before the drop and keeps its root
+ownership.
+
+Assisted-by: Claude:claude-opus-5
+Signed-off-by: Julius Bairaktaris <julius@bairaktaris.de>
+---
+ main.c | 27 ++++++++++++++++++++++++++-
+ 1 file changed, 26 insertions(+), 1 deletion(-)
+
+--- a/main.c
++++ b/main.c
+@@ -30,7 +30,9 @@
+ 
+ #include <getopt.h>
+ #include <errno.h>
++#include <grp.h>
+ #include <netdb.h>
++#include <pwd.h>
+ #include <signal.h>
+ 
+ #include <libubox/usock.h>
+@@ -135,6 +137,7 @@ static int usage(const char *name)
+ 		"	-f              Do not fork to background\n"
+ 		"	-c file         Configuration file, default is '/etc/httpd.conf'\n"
+ 		"	-p [addr:]port  Bind to specified address and port, multiple allowed\n"
++		"	-Z user         Drop to the given user after binding sockets\n"
+ #ifdef HAVE_TLS
+ 		"	-s [addr:]port  Like -p but provide HTTPS on this port\n"
+ 		"	-C file         ASN.1 server certificate file\n"
+@@ -275,6 +278,7 @@ int main(int argc, char **argv)
+ 	struct alias *alias;
+ 	bool nofork = false;
+ 	char *port;
++	const char *priv_user = NULL;
+ 	int opt, ch;
+ 	int cur_fd;
+ 	int bound = 0;
+@@ -295,7 +299,7 @@ int main(int argc, char **argv)
+ 	init_defaults_pre();
+ 	signal(SIGPIPE, SIG_IGN);
+ 
+-	while ((ch = getopt(argc, argv, "A:ab:C:c:Dd:E:e:fh:H:I:i:K:k:L:l:m:N:n:O:o:P:p:qQ:Rr:Ss:T:t:U:u:Xx:y:")) != -1) {
++	while ((ch = getopt(argc, argv, "A:ab:C:c:Dd:E:e:fh:H:I:i:K:k:L:l:m:N:n:O:o:P:p:qQ:Rr:Ss:T:t:U:u:Xx:y:Z:")) != -1) {
+ 		switch(ch) {
+ #ifdef HAVE_TLS
+ 		case 'C':
+@@ -579,6 +583,10 @@ int main(int argc, char **argv)
+ 			                "ignoring -%c\n", ch);
+ 			break;
+ #endif
++		case 'Z':
++			priv_user = optarg;
++			break;
++
+ 		default:
+ 			return usage(argv[0]);
+ 		}
+@@ -614,6 +622,23 @@ int main(int argc, char **argv)
+ 	}
+ #endif
+ 
++	if (priv_user) {
++		struct passwd *pw = getpwnam(priv_user);
++
++		if (!pw) {
++			fprintf(stderr, "Error: Invalid user %s\n", priv_user);
++			return 1;
++		}
++
++		if (initgroups(priv_user, pw->pw_gid) ||
++		    setgid(pw->pw_gid) ||
++		    setresuid(pw->pw_uid, pw->pw_uid, pw->pw_uid)) {
++			fprintf(stderr, "Error: Failed to drop privileges to %s: %s\n",
++			        priv_user, strerror(errno));
++			return 1;
++		}
++	}
++
+ #ifdef HAVE_LUA
+ 	if (lua_handler || lua_prefix) {
+ 		fprintf(stderr, "Need handler and prefix to enable Lua support\n");


package/network/services/uhttpd/files/uhttpd.init
```diff
@@ -122,6 +122,13 @@ start_instance()
 	procd_set_param stderr 1
 	procd_set_param command "$UHTTPD_BIN" -f
 
+	[ -x /sbin/ujail -a -e /etc/capabilities/uhttpd.json ] && {
+		procd_append_param command -Z uhttpd
```
openwrt-ai, line 126: Everything under `/etc/config/` ships as `root:root 0600` — `$(INSTALL_CONF)` is [`install -m0600` (rules.mk:336)](https://github.com/openwrt/openwrt/blob/a5c4a3841cd633649d5bea406a8e7fa2c3249b57/rules.mk#L336), and this package installs its own config that way three lines above in the Makefile. Once `-Z uhttpd` lands, the in-process handlers and every CGI child run as uid 456 and lose all read access to `/etc/config`.

That is not hypothetical for the default install: the shipped config wires LuCI's Lua SGI in-process, [`list lua_prefix "/cgi-bin/luci=/usr/lib/lua/luci/sgi/uhttpd.lua"` (files/uhttpd.config:58)](https://github.com/openwrt/openwrt/blob/a5c4a3841cd633649d5bea406a8e7fa2c3249b57/package/network/services/uhttpd/files/uhttpd.config#L58). That handler reads UCI through libuci directly, not over ubus, so the `acl.d` grant does not cover it — a `uci.cursor()` read just comes back empty rather than erroring, which is the same silent-degradation shape as the 38-entry backup archive in the description.

The five dependencies listed in the PR body all cover the ubus/cgi-io path; none of them covers plain UCI reads from a request handler. Has this been measured, or does something have to widen `/etc/config` (group `uhttpd`, 0640) as part of the series?

---

package/network/services/uhttpd/files/uhttpd.init
```diff
@@ -122,6 +122,13 @@ start_instance()
 	procd_set_param stderr 1
 	procd_set_param command "$UHTTPD_BIN" -f
 
+	[ -x /sbin/ujail -a -e /etc/capabilities/uhttpd.json ] && {
```
openwrt-ai, line 125: The uid drop itself does not depend on the jail. uhttpd binds :80/:443 while still uid 0 and then `setuid()`s, so nothing has to be preserved across a capability drop — `-Z uhttpd` works identically on an image without procd-ujail. Gating it on this guard means that on any image built without ujail, uhttpd keeps running as root, while `/usr/share/acl.d/uhttpd.json` is still installed and still grants uid 456 access to rpcd's `file`, `rc` and `uci` objects — a grant that buys nothing there but is reachable by anything that happens to run as uid 456.

Is the coupling deliberate? Putting `procd_append_param command -Z uhttpd` outside the guard and leaving only `procd_add_jail` / `capabilities` / `no_new_privs` inside it would give the unprivileged uid on every image and keep the capability restriction opt-in, which seems closer to the stated goal.

---

package/network/services/uhttpd/Makefile
```diff
@@ -91,6 +92,10 @@ define Package/uhttpd/install
 	$(INSTALL_CONF) ./files/uhttpd.config $(1)/etc/config/uhttpd
 	$(VERSION_SED_SCRIPT) $(1)/etc/config/uhttpd
 	$(INSTALL_BIN) $(PKG_BUILD_DIR)/uhttpd $(1)/usr/sbin/uhttpd
+	$(INSTALL_DIR) $(1)/etc/capabilities
+	$(INSTALL_DATA) ./files/uhttpd.capabilities $(1)/etc/capabilities/uhttpd.json
+	$(INSTALL_DIR) $(1)/usr/share/acl.d
```
openwrt-ai: The base `uhttpd` package has no ubus dependency — [`DEPENDS:= +USE_GLIBC:libcrypt-compat +libubox +libblobmsg-json +libjson-script +libjson-c`](https://github.com/openwrt/openwrt/blob/a5c4a3841cd633649d5bea406a8e7fa2c3249b57/package/network/services/uhttpd/Makefile#L34); the proxy that needs these grants lives in `uhttpd-mod-ubus`, [`DEPENDS:=uhttpd +libubus +libblobmsg-json`](https://github.com/openwrt/openwrt/blob/a5c4a3841cd633649d5bea406a8e7fa2c3249b57/package/network/services/uhttpd/Makefile#L65). Installing the acl.d file from the base package means images built without the ubus plugin still hand uid 456 — and therefore every CGI process uhttpd spawns — call access to rpcd's `file`, `rc`, `uci` and `rpc-sys` objects that nothing in the image ever uses.

Moving these two lines into `Package/uhttpd-mod-ubus/install` scopes the grant to the images that need it. `/etc/capabilities/uhttpd.json` correctly stays in the base package, since the jail applies regardless of which plugins are present.

---

package/network/services/uhttpd/Makefile
```diff
@@ -91,6 +92,10 @@ define Package/uhttpd/install
...
+	$(INSTALL_DATA) ./files/uhttpd.acl $(1)/usr/share/acl.d/uhttpd.json
```
openwrt-ai: nit: every existing ACL source file in the tree is named `<pkg>_acl.json` — [`files/ntpd_acl.json`](https://github.com/openwrt/openwrt/blob/a5c4a3841cd633649d5bea406a8e7fa2c3249b57/package/utils/busybox/files/ntpd_acl.json), [`files/dnsmasq_acl.json`](https://github.com/openwrt/openwrt/blob/a5c4a3841cd633649d5bea406a8e7fa2c3249b57/package/network/services/dnsmasq/files/dnsmasq_acl.json), [`files/wpad_acl.json`](https://github.com/openwrt/openwrt/blob/a5c4a3841cd633649d5bea406a8e7fa2c3249b57/package/network/services/hostapd/files/wpad_acl.json). `files/uhttpd.acl` is the only one without a `.json` extension, and that is very likely what makes the formality check misread it as a raw file destined for `/etc/config/`. Renaming the source to `files/uhttpd_acl.json` (install path unchanged) matches the convention and should clear that error.

`files/uhttpd.capabilities` already matches [`files/ntpd.capabilities`](https://github.com/openwrt/openwrt/blob/a5c4a3841cd633649d5bea406a8e7fa2c3249b57/package/utils/busybox/files/ntpd.capabilities), so only the ACL file needs renaming. For what it's worth, the bot's other complaint is a false positive: busybox does not list `/etc/capabilities/ntpd.json` in its `conffiles` either.

---

package/network/services/uhttpd/Makefile
```diff
@@ -5,7 +5,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=uhttpd
-PKG_RELEASE:=1
+PKG_RELEASE:=2
```
openwrt-ai, line 8: nit: this branch is based on a `main` from before uhttpd was bumped. The parent commit still pins [`PKG_SOURCE_VERSION:=7b1bec45826bd78c8afc993435bdc0f1df2fe399` / `PKG_SOURCE_DATE:=2026-06-16`](https://github.com/openwrt/openwrt/blob/f4a7ae107e6ad41c37d0bfa152a35a6b2da79037/package/network/services/uhttpd/Makefile#L13-L14), while `main` is now at `60f64bec40c8113cf09815ec377761b1f4f95f22` / `2026-08-03`. That is what the formality bot is reporting as *"version updated from '60f64bec…' to '7b1bec45…'"* — from its point of view the PR downgrades the pin.

Rebasing onto current `main` makes `PKG_RELEASE:=2` correct relative to `60f64bec` and clears the warning. Note that when you later bump `PKG_SOURCE_VERSION` to the openwrt/uhttpd#40 merge commit, `PKG_RELEASE` has to go back to `1` in the same commit.

---

openwrt-ai (commented):
Reviewed 1 new commit; 5 inline comments. The commit message matches the diff.

Grounded the capability reasoning against procd's `jail_run()`/securebits handling and against the existing `sysntpd` / `wpad.init` jail blocks; the guard shape, the `.capabilities` sets and the `INSTALL_DATA` install paths all match in-tree practice, and uid 456 is free across the tree.

---

JuliusBairaktaris:
@brada4 ready for review

package/network/services/uhttpd/files/uhttpd_acl.json
```diff
@@ -0,0 +1,111 @@
...
+		"network.interface": {
+			"methods": [
+				"dump"
+			]
+		},
+		"network.rrdns": {
+			"methods": [
+				"*"
+			]
+		},
```
openwrt-ai, line 104: The wildcard here is justified in the commit message by grouping `network.rrdns` with "objects rpcd owns", where "rpcd re-checks the session ACL for non-root callers". That grouping does not look right for this object: rpcd in this tree builds only the `file`, `rpcsys`, `iwinfo` and `ucode` plugins ([rpcd/Makefile:104-107](https://github.com/openwrt/openwrt/blob/f82d657c79132eb4c32abbf5eeee65bc45de8674/package/system/rpcd/Makefile#L104-L107)), and `network.rrdns` does not appear anywhere in openwrt.git — it arrives from luci-base as a `/usr/libexec/rpcd/` exec plugin that rpcd dispatches by running the script, with no session ACL of its own.

The PR body already establishes that the re-check is not universal even for the one object that has one: `rpc_file_access()` returns true for a request carrying no session id, which is exactly why openwrt/rpcd#38 is listed as a hard prerequisite. If exec-plugin objects carry no re-check at all, then by this file's own stated policy `network.rrdns` belongs in the method-explicit group beside `log` and `service`, listing its single `lookup` method rather than `*`.

The same question applies more sharply to `rc` (line 38) and `rpc-sys` (line 43), where a sessionless wildcard reaches `init`/`upgrade_start`/`password_set` rather than a DNS lookup, and where rpcd#38 does not help since it only touches `rpc_file_access()`. Is there a session check on those paths that I am missing, or is the wildcard relying on uhttpd's ubus proxy being the only caller — which the CGI handlers' own connections would break?

---

package/network/services/uhttpd/files/uhttpd_acl.json
```diff
@@ -0,0 +1,111 @@
...
+		"dsl": {
+			"methods": [
+				"metrics",
+				"statistics"
+			]
+		},
```
openwrt-ai, line 66: nit: the method list is exact — it matches [`dsl_methods[]` in dsl_cpe_ubus.c:1124-1125](``https://github.com/openwrt/openwrt/blob/f82d657c79132eb4c32abbf5eeee65bc45de8674/package/network/config/ltq-vdsl-vr9-app/src/src/dsl_cpe_ubus.c#L1124-L1125``) exactly. But the commit message attributes the `dsl` object to `ltq-vdsl-vr9-app` alone, and `ltq-vdsl-vr11-app` (ipq40xx) compiles the very same file — [`$(CP) ../ltq-vdsl-vr9-app/src/src/dsl_cpe_ubus.c` at Makefile:81](https://github.com/openwrt/openwrt/blob/f82d657c79132eb4c32abbf5eeee65bc45de8674/package/network/config/ltq-vdsl-vr11-app/Makefile#L81) — so it registers the same object with the same two methods. Worth naming vr11 in the message too, so a later reader doesn't conclude the grant is lantiq-only.

Related, for the `fingerprint` block just below (lines 67-71): `ufp` and a `fingerprint` ubus object appear nowhere in openwrt.git, so unlike `dsl` this one can't be checked against the provider's `ubus_method` table. An inert grant for an absent object is harmless, but a typo in the object or method name fails silently and the page just comes back empty — worth confirming both strings against the feed package before merge.

---

openwrt-ai (commented):
Re-reviewed at `f82d657`. The branch was rebased onto current `main` and the single commit amended, so the new material relative to `a5c4a38` is: the `dsl`, `fingerprint` and `network.rrdns` grants in the ACL file, the carried `100-main-add-Z-to-drop-privileges.patch`, and the reworded commit message. The message matches the diff, so no commit-check findings.

The new patch holds up under the checks I could make without the uhttpd tree:

- `getpwnam("uhttpd")` resolving inside the jail is not an accident of the mount setup — in procd's [`jail_run()`](https://github.com/openwrt/procd/blob/2c1834abae171261f2d9161e5a28c36cbcc394a4/service/instance.c#L292) this instance produces only `-n`/`-C`/`-c`, and none of those touch `opts.namespace` in [jail.c:3510-3517](https://github.com/openwrt/procd/blob/2c1834abae171261f2d9161e5a28c36cbcc394a4/jail/jail.c#L3510-L3517), so there is no `CLONE_NEWNS` and `/etc/passwd` is the real one.
- The three capabilities do survive `execve()` despite `no_new_privs`: [`handle_privileged_root()`](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/security/commoncap.c#L789) sets pP' = bset | pI for a root exec, and the NNP downgrade at [commoncap.c:917-927](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/security/commoncap.c#L917-L927) only fires on `is_setid || __cap_gained(permitted, ...)`, neither of which holds once ujail has narrowed its own permitted set to the same three. That is consistent with the reported `CapBnd: 0x4c0` (SETGID | SETUID | NET_BIND_SERVICE) and `CapEff: 0` after the drop.
- `initgroups()` → `setgid()` → `setresuid()` is the correct order, the short-circuit `||` means a failed `initgroups()` cannot fall through to a partial drop, `-Z` collides with no existing option in the getopt string, and the patch's own diffstat (26 insertions, 1 deletion) matches the hunks.

CI is green on head, so nothing CI-grounded to report.

The earlier `PKG_SOURCE_VERSION` nit is resolved by the rebase. The `files/uhttpd.acl` → `files/uhttpd_acl.json` naming nit and the two open questions on `/etc/config` readability and the `-Z`-behind-the-ujail-guard coupling still stand from the previous round; the new patch's own commit message ("service managers that cannot use a sandbox (plain procd, systemd) can then run the daemon unprivileged") reads as further evidence for the latter.

---

package/network/services/uhttpd/files/uhttpd_acl.json
```diff
@@ -0,0 +1,111 @@
...
+		"network.interface": {
+			"methods": [
+				"dump"
+			]
+		},
+		"network.rrdns": {
+			"methods": [
+				"lookup"
```
openwrt-ai, line 102: nit: the PR description no longer matches this hunk. It still says *"Objects rpcd owns, including its `network.rrdns` plugin, get a method wildcard, since rpcd re-checks the session ACL for non-root callers"* — the exact grouping the commit message has since dropped in favour of *"network.rrdns is an exec plugin with no session ACL of its own, so it is named by its single method instead"*. As it stands the description argues for the `"*"` this push replaced.

Two smaller drifts in the same paragraph of the description: it attributes `dsl` to `ltq-vdsl-vr9-app` alone where the commit message now names `ltq-vdsl-vr11-app` as well, and it describes `/usr/share/acl.d/uhttpd.json` without mentioning that the file now ships with `uhttpd-mod-ubus` rather than the base package. Worth syncing, since the description is what a maintainer reads before the commit message.

---

openwrt-ai (commented):
Re-reviewed at `da5d97a`. The tree is otherwise identical to `f82d657` (no rebase this time), so the new material is exactly three changes, and all three answer earlier review threads:

- the acl.d install moved from `Package/uhttpd/install` to `Package/uhttpd-mod-ubus/install`;
- `files/uhttpd.acl` renamed to `files/uhttpd_acl.json` (install path unchanged);
- `network.rrdns` narrowed from `"*"` to `"lookup"`.

Checks I could make on the new state:

- `uhttpd-mod-ubus` carries [`DEPENDS:=uhttpd`](https://github.com/openwrt/openwrt/blob/da5d97ada4b0596fc5020eb89092f5bd7d6f910b/package/network/services/uhttpd/Makefile#L64), so the `uhttpd` user that the ACL's `"user"` key resolves against — created by `USERID` on the base package — exists wherever the file lands, and `PKG_RELEASE` is shared across the subpackages, so the single bump covers the moved file. `/etc/capabilities/uhttpd.json` correctly stayed in the base package, where the jail guard lives.
- The new source name matches the tree's convention for ACL sources ([`ntpd_acl.json`](https://github.com/openwrt/openwrt/blob/da5d97ada4b0596fc5020eb89092f5bd7d6f910b/package/utils/busybox/files/ntpd_acl.json), [`dnsmasq_acl.json`](https://github.com/openwrt/openwrt/blob/da5d97ada4b0596fc5020eb89092f5bd7d6f910b/package/network/services/dnsmasq/files/dnsmasq_acl.json), [`wpad_acl.json`](https://github.com/openwrt/openwrt/blob/da5d97ada4b0596fc5020eb89092f5bd7d6f910b/package/network/services/hostapd/files/wpad_acl.json)), and installing it under a different basename mirrors busybox's `ntpd_acl.json` → `/usr/share/acl.d/ntpd.json`. No in-tree package reloads ubusd from a postinst after dropping an acl.d file, so not doing so here matches practice too.
- The `network.rrdns` narrowing is strictly more restrictive than what the earlier `"*"` granted, so it cannot regress anything the description's measurements covered.

CI is green on head — nothing CI-grounded to report.

## Commit checks

- `da5d97a` "uhttpd: run in ujail as an unprivileged user" — the message matches the diff, including the `uhttpd-mod-ubus` placement and the rrdns rationale. But the amend left two body lines unwrapped, at 92 and 94 columns (`"ACL of its own, so it is named by its single method instead. The objects other daemons own -"` and `"and ltq-vdsl-vr11-app, and fingerprint from ufp where those are installed - are gated by ubusd"`); every other line in the message stays under 76.

Still open from earlier rounds and untouched by this push: the `/etc/config` readability question on `uhttpd.init`, the `-Z`-behind-the-ujail-guard coupling, and the `rc` / `rpc-sys` wildcards raised alongside the `network.rrdns` thread — rpcd#38 only touches `rpc_file_access()`, so it does not cover those two objects.

---

package/network/services/uhttpd/files/uhttpd.init
JuliusBairaktaris: > That is not hypothetical for the default install: the shipped config wires LuCI's Lua SGI in-process, `list lua_prefix "/cgi-bin/luci=/usr/lib/lua/luci/sgi/uhttpd.lua"`

That line ships, but it is inert on a booted snapshot image (r36061, AX3600 /
IPQ8074, LuCI from the same image): `uci show uhttpd.main` still carries the
`lua_prefix`, while `/usr/lib/uhttpd_lua.so` does not exist, and the init only
passes the prefix on when it does.

    package/network/services/uhttpd/files/uhttpd.init:148:
        [ -f /usr/lib/uhttpd_lua.so ] && {
                local len
                config_get len "$cfg" lua_prefix_LENGTH

What serves LuCI on that image is plain CGI. `/www/cgi-bin/luci` is a ucode
script that runs `/usr/share/ucode/luci/dispatcher.uc`, and a CGI child
inherits the server's uid, so under `-Z` the dispatcher reads UCI at uid 456:

    feeds/luci/modules/luci-base/htdocs/cgi-bin/luci:
        #!/usr/bin/env ucode
        ...
        import dispatch from 'luci.dispatcher';

A package install gets there by the other route rather than escaping it.
`uci show uhttpd.main` on this image carries no `ucode_prefix` at all;
installing luci-base as a package adds one, and the in-process handler it names
imports the same `luci.dispatcher`, still at the server's uid. So your point
holds either way.

`/etc/config` is not uniformly 0600 on that image, so whatever the exposure is,
it is per file. `ls -l /etc/config` gives dhcp, network, rpcd, system, uhttpd
and ppe-qos (a local package here) at 0600 root:root, and firewall, luci,
luci_plugins, wireless, openssl, radius, smp_affinity, sqm, ubihealthd and
ubootenv at 0644 root:root.

I have not measured the part that decides this: whether
`/cgi-bin/luci/admin/network/dhcp` degrades at uid 456.
`check_uci_depends()` in dispatcher.uc drops a menu node when `uci.load(config)`
yields nothing, and both `admin/network/dhcp` and `admin/network/dns` are gated
on `dhcp`, which is one of the 0600 files. Settling it needs an image built
from this branch with `-Z` active and a request for that page at uid 456; no
such image exists yet. If the page does 404, widening the affected files to
group `uhttpd` 0640 is the fix and it belongs in this series.

package/network/services/uhttpd/files/uhttpd.init
JuliusBairaktaris: > The uid drop itself does not depend on the jail. uhttpd binds :80/:443 while still uid 0 and then `setuid()`s, so nothing has to be preserved across a capability drop

I agree with that mechanically - `-Z` works on an image without procd-ujail.

The coupling is deliberate anyway, for two reasons. Decoupling produces a
configuration nothing in this PR measures: uid 456 with no
`PR_SET_NO_NEW_PRIVS` and no capability bound. Every number in the description
was taken at `CapEff: 0`, `CapBnd: 0x4c0`, `NoNewPrivs: 1`, and I'd rather not
ship a second, unmeasured arm as the default on the images that lack ujail.

The second is that the grant isn't present on those images. `acl.d/uhttpd.json`
moved to `uhttpd-mod-ubus` in the meantime:

    package/network/services/uhttpd/Makefile, Package/uhttpd-mod-ubus/install:
        $(INSTALL_DIR) $(1)/usr/share/acl.d
        $(INSTALL_DATA) ./files/uhttpd_acl.json $(1)/usr/share/acl.d/uhttpd.json

So the grant is not on every image built without ujail, only on the ones that
also install `uhttpd-mod-ubus`. On those, I agree it buys uhttpd nothing while
uhttpd is still root.

I'd rather keep it coupled until the unguarded arm has numbers of its own.

package/network/services/uhttpd/files/uhttpd_acl.json
JuliusBairaktaris: > `network.rrdns` does not appear anywhere in openwrt.git - it arrives from luci-base as a `/usr/libexec/rpcd/` exec plugin that rpcd dispatches by running the script, with no session ACL of its own.

That part is wrong on my side too, and the commit message said the same thing.
`network.rrdns` is a compiled rpcd plugin, `rpcd-mod-rrdns` in the luci feed,
registering one method:

    feeds/luci/libs/rpcd-mod-rrdns/src/rrdns.c, rpc_rrdns_api_init():
        static const struct ubus_method rrdns_methods[] = {
                UBUS_METHOD("lookup", rpc_rrdns_lookup, rpc_lookup_policy),
        };
        ...
        static struct ubus_object obj = {
                .name = "network.rrdns",

One method, so the grant names it rather than wildcarding; the commit message
is fixed to say that.

> The same question applies more sharply to `rc` (line 38) and `rpc-sys` (line 43) ... and where rpcd#38 does not help since it only touches `rpc_file_access()`.

You're right about rpcd as it stands, and the Dependencies section is wrong to
describe #38 in `rpc_file_access()` terms. `rc.c` carries no session reference
at all, `sys.c` includes `rpcd/session.h` without calling into it and
`iwinfo.c` has neither, so today those objects have no session check. I'll fix
that wording in the description.

#38 is not scoped to `rpc_file_access()`. It walks the objects rpcd has
registered, once, after the `api_init` calls, and swaps every method handler
for a guard:

    rpcd session.c, rpc_session_guard_objects():
        avl_for_each_element(&ctx->objects, obj, avl) {
                ...
                for (i = 0; i < obj->n_methods; i++)
                        methods[i].handler = rpc_session_guard;

The guard requires the session ACL for any caller ubusd reports as non-root,
with `RPC_DEFAULT_SESSION_ID` when the request carries no id. The exception is
the session object's own `access`, `destroy`, `get`, `login`, `set` and
`unset`, which stay reachable because they are how a caller authenticates:

    rpcd session.c, rpc_session_guard():
        if (req->acl.user && strcmp(req->acl.user, "root") &&
            !(!strcmp(obj->name, "session") && rpc_session_own_method(method))) {
                ...
                if (!rpc_session_access(sid, "ubus", obj->name, method))
                        return UBUS_STATUS_PERMISSION_DENIED;

package/network/services/uhttpd/files/uhttpd_acl.json
```diff
@@ -0,0 +1,111 @@
...
+				"get",
+				"login",
+				"set"
+			]
+		},
+		"file": {
+			"methods": [
+				"*"
+			]
+		},
```
openwrt-ai, line 17: nit: the reworded commit message now describes openwrt/rpcd#38 as putting the whole wildcard set behind the session ACL for non-root callers, but the PR body's dependency 2 still scopes #38 to `rpc_file_access()` and an untokened `file exec`. Those disagree about the same fact, and the body is what a maintainer reads first — worth syncing it to the commit message's wording.

---

openwrt-ai (commented):
Reviewed 1 new commit (message-only amend; tree identical to `da5d97a`).

---

PR #24742, opened 2026-08-15
mvebu: enable UAS for Turris Omnia and Turris MOX

BKPepe:
Add kmod-usb-storage-uas to the default package set for the Turris
Omnia and Turris MOX.

Both devices provide USB host ports that can be used with fast USB
storage devices such as SSDs. UAS provides a more efficient transport
than the legacy USB Mass Storage Bulk-Only Transport and allows such
devices to better utilize the available USB bandwidth.

Without the UAS driver installed, compatible devices silently fall
back to the legacy transport, which can limit storage performance.
**Users may not realize that kmod-usb-storage-uas needs to be installed
separately.**

The package is only about 10 KiB and remains a loadable kernel module,
so users can remove it independently if a particular USB storage
device has compatibility issues.

### DIFF
--- a/target/linux/mvebu/image/cortexa53.mk
+++ b/target/linux/mvebu/image/cortexa53.mk
@@ -6,7 +6,7 @@ define Device/cznic_turris-mox
     kmod-rtc-ds1307 kmod-i2c-pxa kmod-dsa kmod-dsa-mv88e6xxx kmod-sfp \
     kmod-phy-marvell kmod-phy-marvell-10g kmod-ath10k ath10k-board-qca988x \
     ath10k-firmware-qca988x kmod-mt7915e kmod-mt7915-firmware mwlwifi-firmware-88w8997 \
-    wpad-basic-mbedtls kmod-mwifiex-sdio kmod-btmrvl
+    wpad-basic-mbedtls kmod-mwifiex-sdio kmod-btmrvl kmod-usb-storage-uas
   SOC := armada-3720
   BOOT_SCRIPT := turris-mox
 endef
--- a/target/linux/mvebu/image/cortexa9.mk
+++ b/target/linux/mvebu/image/cortexa9.mk
@@ -108,7 +108,7 @@ define Device/cznic_turris-omnia
     wpad-basic-mbedtls kmod-ath9k kmod-ath10k-ct ath10k-firmware-qca988x-ct \
     kmod-mt7915-firmware partx-utils kmod-i2c-mux-pca954x kmod-leds-turris-omnia \
     kmod-turris-omnia-mcu kmod-gpio-button-hotplug omnia-eeprom omnia-mcu-firmware \
-    omnia-mcutool kmod-dsa-mv88e6xxx
+    omnia-mcutool kmod-dsa-mv88e6xxx kmod-usb-storage-uas
   IMAGES := sysupgrade.img.gz
   IMAGE/sysupgrade.img.gz := boot-scr | boot-img | sdcard-img | libdeflate-gzip | append-metadata
   SUPPORTED_DEVICES += armada-385-turris-omnia


champtar:
Why not apply it treewide if kmod-usbX is installed ?

openwrt-ai (commented):

---

neheb (approved):

geldot:
> Why not apply it treewide if kmod-usbX is installed ?

UAS isn't supported on the Linksys WRT3200ACM (mvebu-cortexa9) USB 3.0 controller, for example. There is no point in shipping the kmod in that default image. I don't recall if the kmod actively breaks USB, but at best it's confusing.

https://github.com/openwrt/packages/issues/12890

```
[    1.830001] usb 3-1: new SuperSpeed USB device number 2 using xhci-hcd
[    1.860632] usb 3-1: USB controller f10f8000.usb3 does not support streams, which are required by the UAS driver.
[    1.870949] usb 3-1: Please try an other USB controller if you wish to use UAS.
[    1.878291] usb-storage 3-1:1.0: USB Mass Storage device detected
```

That said, it might make sense to include alongside the usb mass storage kmods where it is known to work.

mrkiko:
Maybe we should reason about what we want to do in a generic fashion.
In other words, we are not adding USB storage-related modules to all devices offering USB support as far as I can tell.
Also, we load all modules at startup and not on-demand, so they may end up being just memory waste in some situations.
I can for sure agree the Turris is not a size-constrained device.

geldot:
> Maybe we should reason about what we want to do in a generic fashion. In other words, we are not adding USB storage-related modules to all devices offering USB support as far as I can tell. Also, we load all modules at startup and not on-demand, so they may end up being just memory waste in some situations. I can for sure agree the Turris is not a size-constrained device.

In practice there is no consistency. Many targets include `kmod-usb-storage` even though the device doesn't ship with a USB-attached drive internally or externally which would necessitate it.

A handful of targets include `kmod-usb-storage-uas`. Some targets even include a set of filesystem kmods in case the user decides to use them one day.

Several Gemini targets include ReiserFS:

```
target/linux/gemini/image/Makefile:             kmod-fs-nfsd kmod-fs-ntfs3 kmod-fs-reiserfs kmod-fs-vfat \
```

The Layerscape target builds filesystems directly into the kernel instead of kmods, including both of the NTFS implementations (at the same time).

There doesn't appear to be an overriding rule. Targets described at NAS devices by the manufacturer *tend* to have a wider range of storage-related kmods in the default image.

So, I think it's at the discretion of the target maintainer what they'd like to bundle in the image unless there's a better plan.

```
$ git grep kmod-usb-storage -- target/ | wc -l
35

$ git grep -w kmod-usb-storage-uas -- target/ | wc -l
3
```

BKPepe:
Is something blocking this to merge this? @robimarko ?

robimarko:
Honestly, I dont see the need to ship it by default, as its only used by mass storage devices.

geldot:
> The package is only about 10 KiB and remains a loadable kernel module, so users can remove it independently if a particular USB storage device has compatibility issues.

Does this PR pull in a chain of packages/kmods that wasn't included previously? I think that's worth clarifying in the commit message.

`kmod-usb-storage-uas` depends on `kmod-usb-storage`, which depends on `kmod-scsi-core` and so on. 

If USB storage enablement was the actual intention, it'd make more sense to bundle a proper selection of NAS packages as some other devices do. However, I'm with @robimarko on this one that it doesn't seem appropriate here.

> Without the UAS driver installed, compatible devices silently fall back to the legacy transport, which can limit storage performance. **Users may not realize that kmod-usb-storage-uas needs to be installed separately.**

That concern isn't specific to Turris. If it's a general concern, it needs a general solution, such as making `kmod-usb-storage` depend on `kmod-usb-storage-uas` or just combining the two packages (I'm not sure what the exact convention is with apk). But with the caveat I made earlier about devices like some mvebu that don't support UAS.

BKPepe:
> Does this PR pull in a chain of packages/kmods that wasn't included previously?

No, because the mvebu target builds the entire USB and SCSI stack into the kernel, not as
modules (`target/linux/mvebu/config-6.18`):
```
CONFIG_USB=y                    CONFIG_SCSI=y
CONFIG_USB_EHCI_HCD_ORION=y     CONFIG_SCSI_COMMON=y
CONFIG_USB_XHCI_MVEBU=y         CONFIG_BLK_DEV_SD=y
CONFIG_USB_STORAGE=y
```

`include/kernel.mk` detects built-in modules and emits an empty package
("NOTICE: module '...' is built-in."), so `kmod-usb-storage`, `kmod-scsi-core`
and `kmod-usb-core` ship no kernel modules at all here — just their `/etc/modules.d`
entries. `CONFIG_USB_UAS` is the only symbol not already in the target config, so
the PR would have added exactly one file: `uas.ko`.

robimarko:
> > Does this PR pull in a chain of packages/kmods that wasn't included previously?
> 
> No, because the mvebu target builds the entire USB and SCSI stack into the kernel, not as modules (`target/linux/mvebu/config-6.18`):
> 
> ```
> CONFIG_USB=y                    CONFIG_SCSI=y
> CONFIG_USB_EHCI_HCD_ORION=y     CONFIG_SCSI_COMMON=y
> CONFIG_USB_XHCI_MVEBU=y         CONFIG_BLK_DEV_SD=y
> CONFIG_USB_STORAGE=y
> ```
> 
> `include/kernel.mk` detects built-in modules and emits an empty package ("NOTICE: module '...' is built-in."), so `kmod-usb-storage`, `kmod-scsi-core` and `kmod-usb-core` ship no kernel modules at all here — just their `/etc/modules.d` entries. `CONFIG_USB_UAS` is the only symbol not already in the target config, so the PR would have added exactly one file: `uas.ko`.

I mean, in that case, why not simply enable that in the kernel config as well?

geldot:
> I mean, in that case, why not simply enable that in the kernel config as well?

OP tried to address that in the initial report:

> The package is only about 10 KiB and remains a loadable kernel module,
so users can remove it independently if a particular USB storage
device has compatibility issues.

But as I understand it, that isn't necessary, and there is a more correct way to opt out of UASP:

```
echo '174c:55aa:u' > /sys/module/usb_storage/parameters/quirks
```

Hence, it should be fine to compile into the kernel following the existing precedent. @BKPepe, would you like to revise the PR as such?

(The downside is ~20K dead weight on my WRT3200ACM whose USB 3 controller lacks hardware streams support, but I can live with that.)

PR #24745, opened 2026-08-15
airoha: add patch to move air_en8811h LED configuration to config_init

csharper2005:
Add a target-specific kernel patch for the airoha target that moves the LED GPIO pin configuration sequence from the probe phase to the config_init phase. This ensures that the pins are configured correctly when the PHY is initialized rather than just when it is probed.

Curently, LAN1 led usually doesn't work on Nokia XG-040G-MD and Nokia XG-040G-MF devices.

### DIFF
--- a/target/linux/airoha/patches-6.18/902-net-phy-air-en8811h-move-led-gpio-configuration-to-config-init.patch
+++ b/target/linux/airoha/patches-6.18/902-net-phy-air-en8811h-move-led-gpio-configuration-to-config-init.patch
@@ -1,0 +1,42 @@
+From 2001a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8 Mon Sep 17 00:00:00 2001
+From: Mikhail Zhilkin <csharper2005@gmail.com>
+Date: Sat, 15 Aug 2026 16:12:00 +0000
+Subject: [PATCH] net: phy: air_en8811h: move LED GPIO configuration to config_init
+
+Move the LED GPIO pin configuration sequence from the probe phase to the
+config_init phase. This ensures that the pins are configured correctly
+when the PHY is initialized rather than just when it is probed.
+
+Signed-off-by: Mikhail Zhilkin <csharper2005@gmail.com>
+---
+
+--- a/drivers/net/phy/air_en8811h.c
++++ b/drivers/net/phy/air_en8811h.c
+@@ -957,13 +957,6 @@ static int en8811h_probe(struct phy_devi
+ 	if (ret)
+ 		return ret;
+ 
+-	/* Configure led gpio pins as output */
+-	ret = air_buckpbus_reg_modify(phydev, EN8811H_GPIO_OUTPUT,
+-				      EN8811H_GPIO_OUTPUT_345,
+-				      EN8811H_GPIO_OUTPUT_345);
+-	if (ret < 0)
+-		return ret;
+-
+ 	return 0;
+ }
+ 
+@@ -1049,6 +1042,13 @@ static int en8811h_config_init(struct ph
+ 		return ret;
+ 	}
+ 
++	/* Configure led gpio pins as output */
++	ret = air_buckpbus_reg_modify(phydev, EN8811H_GPIO_OUTPUT,
++				      EN8811H_GPIO_OUTPUT_345,
++				      EN8811H_GPIO_OUTPUT_345);
++	if (ret < 0)
++		return ret;
++
+ 	return 0;
+ }
+ 


target/linux/airoha/patches-6.18/902-net-phy-air-en8811h-move-led-gpio-configuration-to-config-init.patch
openwrt-ai, line 1: `drivers/net/phy/air_en8811h.c` is a generic PHY driver, not an airoha-target file, and mediatek boards ship the same PHY: e.g. [GL.iNet MT2500 v2 (filogic.mk:1823)](https://github.com/openwrt/openwrt/blob/f90435123fd3d3f4476c5a5a9282fa41bebf9067/target/linux/mediatek/image/filogic.mk#L1823) and [Keenetic KN-3911 (filogic.mk:2317)](https://github.com/openwrt/openwrt/blob/f90435123fd3d3f4476c5a5a9282fa41bebf9067/target/linux/mediatek/image/filogic.mk#L2317) both pull `kmod-phy-airoha-en8811h`. Landing the fix under `target/linux/airoha/patches-6.18/` leaves every non-airoha board with this PHY unfixed. The existing in-tree patch to this same file is generic — [`generic/backport-6.18/785-v7.0-07-net-phy-air_en8811h-deprecate-airoha-pnswap-rx-and-a.patch`](``https://github.com/openwrt/openwrt/blob/f90435123fd3d3f4476c5a5a9282fa41bebf9067/target/linux/generic/backport-6.18/785-v7.0-07-net-phy-air_en8811h-deprecate-airoha-pnswap-rx-and-a.patch)`` — and a not-yet-upstream fix to it belongs in `target/linux/generic/pending-6.18/`.

I verified the patch applies with zero offset on top of the generic backports, so the move is mechanical.

nit: `902-` is the "Uncategorized" band in [PATCHES.md's target numbering table](https://github.com/openwrt/openwrt/blob/f90435123fd3d3f4476c5a5a9282fa41bebf9067/target/linux/generic/PATCHES.md#L73-L84); `7xx` is the "Network / PHY" band. `7xx` is free in `generic/pending-6.18/` as well (existing entries there stop at `732-*`).

---

target/linux/airoha/patches-6.18/902-net-phy-air-en8811h-move-led-gpio-configuration-to-config-init.patch
```diff
@@ -0,0 +1,42 @@
+From 2001a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8 Mon Sep 17 00:00:00 2001
+From: Mikhail Zhilkin <csharper2005@gmail.com>
+Date: Sat, 15 Aug 2026 16:12:00 +0000
+Subject: [PATCH] net: phy: air_en8811h: move LED GPIO configuration to config_init
+
+Move the LED GPIO pin configuration sequence from the probe phase to the
+config_init phase. This ensures that the pins are configured correctly
+when the PHY is initialized rather than just when it is probed.
```
openwrt-ai, line 8: The kernel-facing commit message only restates the move; it never says why the move is needed. [PATCHES.md](https://github.com/openwrt/openwrt/blob/f90435123fd3d3f4476c5a5a9282fa41bebf9067/target/linux/generic/PATCHES.md#L3-L7) requires each patch to carry "a proper commit message explaining what they change **and why**", and this is exactly what an upstream reviewer will ask for first.

The concrete facts are all in the OpenWrt commit message but not in the patch: LAN1's LED does not work on Nokia XG-040G-MD / XG-040G-MF. Worth adding the mechanism too — `en8811h_config_init()` calls `en8811h_restart_mcu()` on every invocation after the first, and the probe-time write to `EN8811H_GPIO_OUTPUT` does not survive that, so the pins fall back to inputs. If that is the mechanism you observed, state it; it turns the patch from "moved some code" into a reviewable bug fix.

nit: "Curently" in the OpenWrt commit message body (and the PR description) should be "Currently".

---

target/linux/airoha/patches-6.18/902-net-phy-air-en8811h-move-led-gpio-configuration-to-config-init.patch
```diff
@@ -0,0 +1,42 @@
...
+
+Signed-off-by: Mikhail Zhilkin <csharper2005@gmail.com>
+---
+
```
openwrt-ai, line 12: nit: `2001a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8` is not a real commit — the sequential nibbles give it away as a placeholder. Every other patch in this directory carries the SHA `git format-patch` actually produced (`921-net-airoha-npu-load-firmware-...` starts with `0509cfb2607a…`, `801-01-…` with `f2c6f8711c38…`), so a synthetic hash sends readers looking for a commit that does not exist. Dropping the line is cleaner than inventing one.

The `---` separator is also missing the diffstat that the sibling patches carry.

```suggestion
From: Mikhail Zhilkin <csharper2005@gmail.com>
Date: Sat, 15 Aug 2026 16:12:00 +0000
Subject: [PATCH] net: phy: air_en8811h: move LED GPIO configuration to config_init

Move the LED GPIO pin configuration sequence from the probe phase to the
config_init phase. This ensures that the pins are configured correctly
when the PHY is initialized rather than just when it is probed.

Signed-off-by: Mikhail Zhilkin <csharper2005@gmail.com>
---
 drivers/net/phy/air_en8811h.c | 14 +++++++-------
 1 file changed, 7 insertions(+), 7 deletions(-)

```

---

target/linux/airoha/patches-6.18/902-net-phy-air-en8811h-move-led-gpio-configuration-to-config-init.patch
```diff
@@ -0,0 +1,42 @@
...
+ 		return ret;
+ 
+-	/* Configure led gpio pins as output */
+-	ret = air_buckpbus_reg_modify(phydev, EN8811H_GPIO_OUTPUT,
+-				      EN8811H_GPIO_OUTPUT_345,
+-				      EN8811H_GPIO_OUTPUT_345);
+-	if (ret < 0)
+-		return ret;
+-
+ 	return 0;
```
openwrt-ai, line 26: nit: with the GPIO block removed, `en8811h_probe()` now ends on [`en8811h_clk_provider_setup()` followed by `if (ret) return ret;` / `return 0;` (air_en8811h.c:955-957)](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/air_en8811h.c#L955-L957) — a pattern upstream will ask you to collapse into `return en8811h_clk_provider_setup(&phydev->mdio.dev, &priv->hw);`. It only became redundant because of this removal, so it fits in this patch.

---

openwrt-ai (commented):
Reviewed 1 new commit. The code change itself looks right — I applied it on top of the generic `785-v7.0-0*` backports and it lands with zero offset, and moving the `EN8811H_GPIO_OUTPUT` write after the `en8811h_restart_mcu()` in `.config_init()` is consistent with the reported symptom. Main concern is patch placement (generic driver under a target directory); details inline.

---

PR #21896, opened 2026-02-06
airoha: an7581: Add support Nokia Bell XG-040G-MD

Cairongzeng:
airoha: an7581: Nokia Bell XG040GMD Router Mode

Hardware:

SoC: Airoha AN7581
RAM: 512MB DDR4 (Jinhua CA4S4G16V-F9HPC)
Flash: 256MB SPI NAND (SkyHigh S35ML02G300)
Ethernet:
1x 2.5GbE port (EcoNet EN8811H PHY) labeled as LAN1
3x 1GbE ports (internal AN7581 switch) labeled as LAN2, LAN3, LAN4
PON: EcoNet EN7572AN
USB: 1x USB 2.0, 1x USB 3.0
LEDs: 4x GPIO-controlled (power, PON, LOS, internet)
Buttons: 1x reset button

mtd: spinand: robust SkyHigh read workaround
add robust read workaround for SkyHigh chips

mtd: spinand: Add FudanMicro FM25G02B support
Add support for FudanMicro FM25G02B SPI NAND

Signed-off-by: Rongzeng Cai <cairongzeng@outlook.com>

PR #24746, opened 2026-08-15
airoha: add patch to handle premature ready bit for SkyHigh chips

csharper2005:
Add a new kernel patch for the airoha target to resolve issues where SkyHigh SPI NAND controllers drop the busy status bit prematurely before the internal operation settles. The patch adds a delay and validation loop triggered by a new SPINAND_HAS_PREMATURE_READY_BIT flag.

This pathch fixes bit-flips on Nokia XG-040G-MD devices with SkyHigh spi nand chips:
```
[2503209.483768] ubi0: schedule PEB 399 for scrubbing
[2503209.542280] ubi0: scrubbed PEB 399 (LEB 5:272), data moved to PEB 366
[2562754.294549] ubi0: fixable bit-flip detected at PEB 1316
[2562754.360134] ubi0: scrubbed PEB 1316 (LEB 5:195), data moved to PEB 399
[2589193.123216] ubi0: fixable bit-flip detected at PEB 511
[2589193.128571] ubi0: schedule PEB 511 for scrubbing
[2589193.197174] ubi0: scrubbed PEB 511 (LEB 5:116), data moved to PEB 364
[2689227.173721] ubi0: fixable bit-flip detected at PEB 1194
[2689227.179237] ubi0: schedule PEB 1194 for scrubbing
[2689227.221495] ubi0: fixable bit-flip detected at PEB 1194
[2689227.295181] ubi0: scrubbed PEB 1194 (LEB 5:76),

```

### DIFF
--- a/target/linux/airoha/patches-6.18/903-mtd-spinand-handle-premature-ready-bit-for-SkyHigh-chips.patch
+++ b/target/linux/airoha/patches-6.18/903-mtd-spinand-handle-premature-ready-bit-for-SkyHigh-chips.patch
@@ -1,0 +1,61 @@
+From 2001a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8 Mon Sep 17 00:00:00 2001
+From: Mikhail Zhilkin <csharper2005@gmail.com>
+Date: Sat, 15 Aug 2026 16:20:00 +0000
+Subject: [PATCH] mtd: spinand: handle premature ready bit for SkyHigh chips
+
+Some SPI NAND controllers drop the busy status bit prematurely before the
+internal operation has fully settled. Introduce a delay and validation
+loop when the SPINAND_HAS_PREMATURE_READY_BIT flag is set, and apply this
+flag to SkyHigh SPI NAND devices during initialization.
+
+Signed-off-by: Mikhail Zhilkin <csharper2005@gmail.com>
+---
+
+@@ -9,6 +9,7 @@
+ 
+ #define pr_fmt(fmt)	"spi-nand: " fmt
+ 
++#include <linux/delay.h>
+ #include <linux/device.h>
+ #include <linux/jiffies.h>
+ #include <linux/kernel.h>
+@@ -561,6 +562,19 @@ int spinand_wait(struct spinand_device *
+ 	if (ret)
+ 		return ret;
+ 
++	if (spinand->flags & SPINAND_HAS_PREMATURE_READY_BIT) {
++		ndelay(100);
++
++		ret = spinand_read_status(spinand, &status);
++		if (ret)
++			return ret;
++
++		if (status & STATUS_BUSY)
++			return -ETIMEDOUT;
++
++		*spinand->scratchbuf = status;
++	}
++
+ 	status = *spinand->scratchbuf;
+ 	if (!(status & STATUS_BUSY))
+ 		goto out;
+--- a/drivers/mtd/nand/spi/skyhigh.c
++++ b/drivers/mtd/nand/spi/skyhigh.c
+@@ -125,6 +125,7 @@ static const struct spinand_info skyhigh
+ 
+ static int skyhigh_spinand_init(struct spinand_device *spinand)
+ {
++	spinand->flags |= SPINAND_HAS_PREMATURE_READY_BIT;
+ 	/*
+ 	 * Config_Protect_En (bit 1 in Block Lock register) must be set to 1
+ 	 * before writing other bits. Do it here before core unlocks all blocks
+--- a/include/linux/mtd/spinand.h
++++ b/include/linux/mtd/spinand.h
+@@ -418,6 +418,7 @@ struct spinand_ecc_info {
+ #define SPINAND_HAS_PROG_PLANE_SELECT_BIT		BIT(2)
+ #define SPINAND_HAS_READ_PLANE_SELECT_BIT		BIT(3)
+ #define SPINAND_NO_RAW_ACCESS				BIT(4)
++#define SPINAND_HAS_PREMATURE_READY_BIT			BIT(5)
+ 
+ /**
+  * struct spinand_ondie_ecc_conf - private SPI-NAND on-die ECC engine structure


csharper2005:
cc @xiangtailiang @ikas-mc @Cairongzeng further to the discussion: https://github.com/openwrt/openwrt/pull/21896#issuecomment-4726208646  I rewrote the patch that the AI ​​was complaining about.

VitaliySochniy:
Initial boot tests on the Nokia XG-040G-MD look good. The device boots successfully, and the SkyHigh SPI NAND initializes as expected without immediate issues. Long-term testing for UBI bit-flips hasn't been conducted yet, but the primary test is OK.

target/linux/airoha/patches-6.18/903-mtd-spinand-handle-premature-ready-bit-for-SkyHigh-chips.patch
```diff
@@ -0,0 +1,61 @@
...
+
+Some SPI NAND controllers drop the busy status bit prematurely before the
+internal operation has fully settled. Introduce a delay and validation
+loop when the SPINAND_HAS_PREMATURE_READY_BIT flag is set, and apply this
+flag to SkyHigh SPI NAND devices during initialization.
+
+Signed-off-by: Mikhail Zhilkin <csharper2005@gmail.com>
+---
+
+@@ -9,6 +9,7 @@
```
openwrt-ai, line 14: The `--- a/drivers/mtd/nand/spi/core.c` / `+++ b/drivers/mtd/nand/spi/core.c` header is missing, so the two hunks below have no target file. `patch(1)` silently swallows them as part of the mail preamble and still exits 0 — it never even mentions core.c:

```
$ patch -p1 --dry-run --batch --forward < 903-....patch
checking file drivers/mtd/nand/spi/skyhigh.c
checking file include/linux/mtd/spinand.h
Hunk #1 succeeded at 415 (offset -3 lines).
EXIT=0
```

Net effect: the build succeeds, `SPINAND_HAS_PREMATURE_READY_BIT` gets defined and set in `skyhigh.c`, but the code in [`spinand_wait()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mtd/nand/spi/core.c#L548-L580) that would read the flag is never applied — the patch is a functional no-op. That also means the "boots fine" report on this PR doesn't yet exercise the new code path.

Adding the header makes both hunks apply (with small offsets, so please re-run `make target/linux/refresh` afterwards to fix the hunk line numbers):

```suggestion
--- a/drivers/mtd/nand/spi/core.c
+++ b/drivers/mtd/nand/spi/core.c
@@ -9,6 +9,7 @@
```

---

target/linux/airoha/patches-6.18/903-mtd-spinand-handle-premature-ready-bit-for-SkyHigh-chips.patch
openwrt-ai, line 1: Nothing in this patch is airoha-specific: it changes `drivers/mtd/nand/spi/core.c`, `drivers/mtd/nand/spi/skyhigh.c` and `include/linux/mtd/spinand.h`, and it arms the quirk for every SkyHigh chip on every board. Every other spinand core/vendor patch in the tree lives under `target/linux/generic/` — e.g. [`pending-6.18/487-mtd-spinand-Add-support-for-Etron-EM73D044VCx.patch`](``https://github.com/openwrt/openwrt/blob/ca85a3a23aeba479955162133270f5367e6aecbd/target/linux/generic/pending-6.18/487-mtd-spinand-Add-support-for-Etron-EM73D044VCx.patch)`` and [`pending-6.18/417-mtd-spi-nand-macronix-disable-continuous-read-for-MX.patch`](``https://github.com/openwrt/openwrt/blob/ca85a3a23aeba479955162133270f5367e6aecbd/target/linux/generic/pending-6.18/417-mtd-spi-nand-macronix-disable-continuous-read-for-MX.patch).`` The only spinand patch under `target/linux/airoha/patches-6.18/` is `901-snand-mtk-bmt-support.patch`, which really is a target-local BMT hack.

Keeping it here means other targets with SkyHigh SPI NAND don't get the fix, and it will silently conflict once someone adds the same flag generically.

Suggest moving it to `target/linux/generic/pending-6.18/` and renumbering: per [PATCHES.md's numbering table](https://github.com/openwrt/openwrt/blob/ca85a3a23aeba479955162133270f5367e6aecbd/target/linux/generic/PATCHES.md#L56) MTD patches use the `4xx` range, so something like `4NN-mtd-spinand-handle-premature-ready-bit-for-skyhigh-chips.patch` (lowercase shortname, per the [`pending` naming rule](https://github.com/openwrt/openwrt/blob/ca85a3a23aeba479955162133270f5367e6aecbd/target/linux/generic/PATCHES.md#L40-L44)).

---

target/linux/airoha/patches-6.18/903-mtd-spinand-handle-premature-ready-bit-for-SkyHigh-chips.patch
```diff
@@ -0,0 +1,61 @@
...
++#include <linux/delay.h>
+ #include <linux/device.h>
+ #include <linux/jiffies.h>
+ #include <linux/kernel.h>
+@@ -561,6 +562,19 @@ int spinand_wait(struct spinand_device *
+ 	if (ret)
+ 		return ret;
+ 
++	if (spinand->flags & SPINAND_HAS_PREMATURE_READY_BIT) {
++		ndelay(100);
```
openwrt-ai, line 27: `ndelay(100)` is 100 nanoseconds. The very next thing that happens is `spinand_read_status()`, which issues a full GET_FEATURE SPI transaction — that alone costs several microseconds of controller setup plus 2 bytes on the wire, i.e. one to two orders of magnitude more than the delay you just inserted. So the settle time this is supposed to provide is dominated by noise, and the only thing the block effectively adds is the extra status read.

Was `udelay(100)` intended? If 100 ns really is the number from the datasheet, a short comment naming the parameter (e.g. tRST/tPOST or whichever timing it is) would help, since a bare magic number here is impossible to review.

---

target/linux/airoha/patches-6.18/903-mtd-spinand-handle-premature-ready-bit-for-SkyHigh-chips.patch
```diff
@@ -0,0 +1,61 @@
...
++
++		ret = spinand_read_status(spinand, &status);
++		if (ret)
++			return ret;
++
++		if (status & STATUS_BUSY)
++			return -ETIMEDOUT;
```
openwrt-ai, line 34: Bailing out with `-ETIMEDOUT` after a single re-read throws away the rest of the wait budget. `spi_mem_poll_status()` above was given `SPINAND_WAITRDY_TIMEOUT_MS` and typically returns long before it is exhausted; here, if the chip happens to still be busy 100 ns later, the read/program/erase is failed outright instead of waiting the remaining time. That turns a transient "ready came up a hair early" into a hard I/O error — the opposite of what the patch is trying to fix.

Both the commit message and the patch's own description say this adds a "delay and validation loop", but there is no loop — it is one delay and one read. Either implement the retry loop that is described (re-poll until `!(status & STATUS_BUSY)` or the timeout expires), or drop the `-ETIMEDOUT` early return and let the existing tail of [`spinand_wait()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mtd/nand/spi/core.c#L563-L579) — which already does an "extra read, just in case the STATUS_READY bit has changed" and returns `-ETIMEDOUT` on busy — produce the result, and reword the message to match.

---

target/linux/airoha/patches-6.18/903-mtd-spinand-handle-premature-ready-bit-for-SkyHigh-chips.patch
```diff
@@ -0,0 +1,61 @@
...
++
++		*spinand->scratchbuf = status;
```
openwrt-ai, line 36: This write-back is dead code. `spinand_read_status()` is a thin wrapper around [`spinand_read_reg_op()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mtd/nand/spi/core.c#L23-L35), which performs the GET_FEATURE op straight into `spinand->scratchbuf` and only then does `*val = *spinand->scratchbuf;`. So by the time you get here `*spinand->scratchbuf` already holds exactly `status`, and the following `status = *spinand->scratchbuf;` re-reads the same byte.

Dropping this line (and the blank line above it) makes the intent clearer. Note that removing two lines invalidates the `@@ -561,6 +562,19 @@` hunk header, so regenerate the patch with `make target/linux/refresh` rather than hand-editing the counts.

---

target/linux/airoha/patches-6.18/903-mtd-spinand-handle-premature-ready-bit-for-SkyHigh-chips.patch
```diff
@@ -0,0 +1,61 @@
...
+ 	status = *spinand->scratchbuf;
+ 	if (!(status & STATUS_BUSY))
+ 		goto out;
+--- a/drivers/mtd/nand/spi/skyhigh.c
++++ b/drivers/mtd/nand/spi/skyhigh.c
+@@ -125,6 +125,7 @@ static const struct spinand_info skyhigh
+ 
+ static int skyhigh_spinand_init(struct spinand_device *spinand)
+ {
++	spinand->flags |= SPINAND_HAS_PREMATURE_READY_BIT;
```
openwrt-ai, line 48: There is a mismatch between where the quirk is applied and what the commit message blames. The message says "SkyHigh SPI NAND **controllers** drop the busy status bit prematurely" — but SkyHigh is the NAND die vendor; the SPI controller on the affected board is the Airoha SNFI (`drivers/spi/spi-airoha-snfi.c`). Setting the flag in [`skyhigh_spinand_init()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mtd/nand/spi/skyhigh.c#L126-L135) arms it unconditionally for every SkyHigh part behind every controller, including targets where nobody has reported the problem.

Which side actually misbehaves? If it is the chip, the message wording should say "chips" and it would be good to state which SkyHigh part IDs were reproduced on (the driver currently carries S35ML01G3/02G3/04G3 entries). If it is the Airoha controller, the flag belongs on the controller/board side rather than in the vendor chip driver, otherwise this regresses nothing today but silently costs an extra status read per operation on every other SkyHigh user.

Also, evidence-wise: the bit-flip log quoted in the commit message shows UBI *fixing* correctable bit-flips and scrubbing, which is normal NAND wear behaviour and is not obviously the same failure mode as a premature ready bit. Some explanation of how the ready-bit timing produces those particular UBI messages would make the fix much easier to accept.

---

target/linux/airoha/patches-6.18/903-mtd-spinand-handle-premature-ready-bit-for-SkyHigh-chips.patch
```diff
@@ -0,0 +1,61 @@
+From 2001a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8 Mon Sep 17 00:00:00 2001
```
openwrt-ai, line 1: nit: `2001a1b2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8` is a placeholder, not a real commit — it is just sequential hex bytes, and no such object exists in mainline. On a `From <sha> Mon Sep 17 00:00:00 2001` line it reads as a reference to an upstream commit that can be looked up, which will send the next person on a hunt. Since this is a new downstream patch (correctly not named `-vX.Y-`), either drop the line entirely or let `git format-patch` fill in the real SHA of your local commit.

---

openwrt-ai (commented):
## Commit checks

- `ca85a3a` "airoha: add patch to handle premature ready bit for SkyHigh chips" — the message says the patch "adds a delay and validation **loop**", but the added code does one `ndelay(100)` and one status read with no retry loop (see inline comment on lines 33-34).
- `ca85a3a` — the log excerpt in the body is delimited by bare `---` lines at column 0. That is the standard patch separator, so when this commit is exported with `git format-patch` and re-applied with `git am` (the route PRs take into the staging trees) everything from the first `---` onwards is liable to be dropped from the commit message. The PR description uses a fenced code block for the same text; indenting the log lines, or using a different delimiter, keeps it intact in the commit as well.
- `ca85a3a` — typo: "This **pathch** fixes bit-flips" (same typo in the PR description).


---

PR #24748, opened 2026-08-15
qualcommax: ipq50xx: fix CMCC PZ-L8 5 GHz UNII-3 channels 149-165

autumnshiu009-ship-it:
## Summary

CMCC PZ-L8 (IPQ5018 + QCN6122) cannot use 5 GHz UNII-3 channels 149-165. Two independent causes were found on real hardware:

1. The PZ-L8 device tree limits the 5 GHz radio to
   `ieee80211-freq-limit = <5150000 5730000>`, which filters out
   channels 149-165 before regulatory evaluation.
2. The QCN6122 firmware reports `high_5ghz_chan` without the UNII-3
   band, so ath11k marks channels above that value as
   `IEEE80211_CHAN_DISABLED`.

This PR fixes both:

- `qualcommax: ipq50xx: cmcc,pz-l8: fix 5 GHz UNII-3 channel limit`
  raises the DTS upper bound to 5850 MHz.
- `mac80211: ath11k: fix QCN6122 UNII-3 channels 149-165 disabled`
  carries the pending linux-wireless driver fix under
  `package/kernel/mac80211/patches/ath11k/`.

## Why this is not a regulatory workaround

- The board-2.bin BDF for PZ-L8 contains the UNII-3 frequencies.
- UNII-3 is legal in the target regulatory domains (e.g. CN and US).
- The fixes only stop incorrect hardware/firmware capability filtering;
  normal regulatory domain checks still apply afterwards.

## Tested on device

- Device: CMCC PZ-L8, OpenWrt SNAPSHOT r33948-a5c5ca0666,
  qualcommax/ipq50xx
- Kernel: `ath11k b00a040.wifi: qcn6122 hw1.0`
- Before: channel 149 cannot be selected/used by 5 GHz STA or AP
- After: 5 GHz STA and AP both work on channel 149 / HE80
  (`iwinfo`: ch149, center1 155, 866.7 Mbit/s STA link)

## Note on QCN6102/QCN6122 naming

Early teardown articles call this radio QCN6102, while the kernel and
firmware path use QCN6122. This PR does not resolve that naming
question; the driver change matches `ATH11K_HW_QCN6122_HW10` only.

## Attribution

Patch authored with assistance by V4Pro (DeepSeek Harness), tested and
submitted by Autumn Shiu <autumnshiu@foxmail.com>.

## Upstream driver submission

The ath11k driver fix was submitted to the linux-wireless mailing list
on 2026-08-15:

- To: linux-wireless@vger.kernel.org
- Cc: kvalo@kernel.org, ath11k@lists.infradead.org
- Message-ID: <tencent_CA929E144EEEC926E8CFF32B18ED3A26A907@qq.com>
- Archive: https://marc.info/?l=linux-wireless&m=178680667027761
- Lore: https://lore.kernel.org/linux-wireless/tencent_CA929E144EEEC926E8CFF32B18ED3A26A907@qq.com/


package/kernel/mac80211/patches/ath11k/930-wifi-ath11k-fix-qcn6122-unii3.patch
```diff
@@ -0,0 +1,19 @@
+diff --git a/drivers/net/wireless/ath/ath11k/mac.c b/drivers/net/wireless/ath/ath11k/mac.c
```
openwrt-ai: The provenance of this patch is unclear and self-contradictory. The commit message says "Carry a pending linux-wireless fix", which asserts the patch was posted upstream, but the PR body says "Patch authored by V4Pro (DeepSeek Harness), tested and submitted by autumnshiu009-ship-it" — i.e. written here, not taken from linux-wireless. There is no `Link:` to a lore.kernel.org posting anywhere in the commit or the patch.

Every other patch in this directory carries the original mail headers and authorship, e.g. `947-wifi-ath11k-fix-rssi-station-dump-for-IPQ5018-and-QC.patch` starts with `From <sha> Mon Sep 17 00:00:00 2001` / `From:` / `Date:` / `Subject:` and keeps the upstream `Signed-off-by:`. This file has none of that.

Please either (a) post the fix to linux-wireless first and add the `Link:` plus real mail headers, or (b) drop the "pending linux-wireless fix" wording from the commit message and mark it plainly as an OpenWrt-local workaround with a proper `From:`/`Subject:`/`Signed-off-by:` header block naming the actual author.

---

package/kernel/mac80211/patches/ath11k/930-wifi-ath11k-fix-QCN6122-UNII-3-channels-149-165-disa.patch
```diff
@@ -0,0 +1,19 @@
...
+ 	temp_reg_cap = reg_cap;
+ 
++	/* QCN6122 firmware omits the UNII-3 band (5725-5850 MHz, channels
++	 * 149-165) from high_5ghz_chan, which makes the driver mark those
++	 * channels IEEE80211_CHAN_DISABLED. Clamp it to the standard 5.8 GHz
++	 * upper bound so channels 149-165 remain usable.
++	 */
++	if (ar->ab->hw_rev == ATH11K_HW_QCN6122_HW10 &&
++	    reg_cap->high_5ghz_chan < 5825)
++		reg_cap->high_5ghz_chan = 5825;
```
openwrt-ai, line 47: This is keyed on `hw_rev` alone, so it changes behaviour for every QCN6122 board in the tree, not just the one tested. There are ten boards including `ipq5018-qcn6122.dtsi`, and only four of them constrain the radio in DT:

- capped at 5730 MHz in DT (unaffected, the DT limit is applied after this in `wiphy_read_of_freq_limits()`): `ipq5018-wrc-x3000gs2.dts`, `ipq5018-wrc-x3000gst2.dts`, `ipq5018-wn-dax3000gr.dts` — all JP-market devices where 5.8 GHz is not allocated.
- **no `ieee80211-freq-limit` at all**: `ipq5018-ax830.dts`, `ipq5018-gl-b3000.dts`, `ipq5018-mr3000d-ci.dts`, `ipq5018-mx2000.dts`, `ipq5018-mx6200.dts`, `ipq5018-scr50axe.dts`.

For those six, this silently raises `high_5ghz_chan` to 5825 regardless of what their own firmware/BDF reported, and none of them were tested here. If the truncation really is a universal QCN6122 firmware trait then that is arguably the right outcome, but it should be stated explicitly in the commit message. Do you have evidence (e.g. an `ath11k` reg-cap dump) that the reported `high_5ghz_chan` is identical across QCN6122 boards, rather than BDF-derived per board?

---

package/kernel/mac80211/patches/ath11k/930-wifi-ath11k-fix-QCN6122-UNII-3-channels-149-165-disa.patch
```diff
@@ -0,0 +1,19 @@
...
++++ b/drivers/net/wireless/ath/ath11k/mac.c
+@@ -10172,6 +10172,15 @@ static int ath11k_mac_setup_channels_rates(struct ath11k *ar,
+ 	reg_cap = &ar->ab->hal_reg_cap[ar->pdev_idx];
+ 	temp_reg_cap = reg_cap;
+ 
++	/* QCN6122 firmware omits the UNII-3 band (5725-5850 MHz, channels
++	 * 149-165) from high_5ghz_chan, which makes the driver mark those
++	 * channels IEEE80211_CHAN_DISABLED. Clamp it to the standard 5.8 GHz
++	 * upper bound so channels 149-165 remain usable.
++	 */
```
openwrt-ai, line 44: nit: the comment says the band is "5725-5850 MHz" but the clamp is 5825, which invites a later "cleanup" to 5850 that would be wrong. `high_5ghz_chan` is compared against channel *center* frequencies in [`ath11k_mac_update_ch_list()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/ath/ath11k/mac.c#L9947-L9961), not band edges, so 5825 is exactly ch165's center and 5850 would additionally enable ch169 ([`CHAN5G(169, 5845, 0)`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/ath/ath11k/mac.c#L98)), which is not allocated in CN or US. Worth spelling that out so the value is not "corrected" later:

```suggestion
+	/* QCN6122 firmware omits the UNII-3 band (5725-5850 MHz, channels
+	 * 149-165) from high_5ghz_chan, which makes the driver mark those
+	 * channels IEEE80211_CHAN_DISABLED. high_5ghz_chan is compared
+	 * against channel center frequencies, so clamp it to 5825 MHz, the
+	 * center of ch165 -- using the 5850 MHz band edge here would also
+	 * enable ch169, which is not allocated.
+	 */
```

---

openwrt-ai (commented):
## Commit checks

- `fbaaa19` "mac80211: ath11k: fix QCN6122 UNII-3 channels 149-165 disabled" — the message says "Carry a pending linux-wireless fix", but the PR body states the patch was authored for this PR ("Patch authored by V4Pro (DeepSeek Harness)"). No `Link:` to a linux-wireless posting is given and the patch file has no mail headers or original `Signed-off-by:`. Either substantiate the upstream claim with a lore link, or reword the message to describe it as an OpenWrt-local workaround. Details inline.

`9560655` looks fine: the DTS change matches its message, and 5850000 kHz is the correct upper bound — with the 20 MHz fit check in `wiphy_freq_limits_valid_chan()` it admits ch165 (5815-5835 kHz range) and still excludes ch169 (which would need 5855).


---

autumnshiu009-ship-it:
Thanks for the review. I've updated the branch:

- Commit author/committer/Signed-off-by now use a real full name and a real email (`Autumn Shiu <autumnshiu@foxmail.com>`).
- The embedded mac80211 patch now carries a proper `git format-patch` header block (`From <hash> ...`, `From:`, `Date:`, `Subject:`) plus the Signed-off-by, so it applies cleanly with `git am`.
- Bumped `package/kernel/mac80211` `PKG_RELEASE` to 4.
- Reworked the commit wording: the driver fix was posted to linux-wireless on 2026-08-15 (Message-ID included) and is carried here until it lands upstream, instead of the previous ambiguous "pending" phrasing.
- Adopted the suggestion to explain why the clamp is 5825 MHz (ch165 center) rather than the 5850 MHz band edge, which would enable ch169.

Regarding scope: I only have a CMCC PZ-L8 to test on, so I cannot yet prove from hardware logs that every QCN6122 board reports the same truncated `high_5ghz_chan`. The clamp only changes behaviour when firmware reports a value below 5825 MHz; boards whose device tree already limits the radio to 5730 MHz still have that DT limit applied afterwards. I'd welcome reports from other QCN6122 boards before/after this change.

package/kernel/mac80211/patches/ath11k/930-wifi-ath11k-fix-qcn6122-unii3.patch
```diff
@@ -0,0 +1,55 @@
...
+
+Signed-off-by: Autumn Shiu <autumnshiu@foxmail.com>
+---
+ drivers/net/wireless/ath/ath11k/mac.c | 9 +++++++++
+ 1 file changed, 9 insertions(+)
+
+diff --git a/drivers/net/wireless/ath/ath11k/mac.c b/drivers/net/wireless/ath/ath11k/mac.c
+index 2d55cdc..e23a295 100644
+--- a/drivers/net/wireless/ath/ath11k/mac.c
++++ b/drivers/net/wireless/ath/ath11k/mac.c
```
openwrt-ai: The new `git format-patch` framing added in this revision brings along the raw-git artefacts: the `diff --git` line, `index 2d55cdc..e23a295 100644`, and the `-- ` / `2.43.0` trailer at lines 53-54. None of the 164 patches under `package/kernel/mac80211/patches/` contain any of those — they are quilt-managed and start the diff directly at `--- a/...`, e.g. [`931-wifi-ath11k-Support-to-assign-m3-dump-memory.patch`](``https://github.com/openwrt/openwrt/blob/3cf57c966033c7a96ea989040d9129b27a03076c/package/kernel/mac80211/patches/ath11k/931-wifi-ath11k-Support-to-assign-m3-dump-memory.patch#L1-L11),`` which keeps the `From <sha> Mon Sep 17`/`From:`/`Date:`/`Subject:`/`---` header block but no `diff --git`, no `index`, no version trailer.

`make package/mac80211/refresh` regenerates the patch in that form (the header block above `---` is preserved). Dropping lines 33-34 here plus the `-- ` / `2.43.0` trailer at the end of the file gets you the same result:

```suggestion
--- a/drivers/net/wireless/ath/ath11k/mac.c
+++ b/drivers/net/wireless/ath/ath11k/mac.c
```

---

package/kernel/mac80211/patches/ath11k/930-wifi-ath11k-fix-QCN6122-UNII-3-channels-149-165-disa.patch
```diff
@@ -0,0 +1,55 @@
...
+band edge here would also enable channel 169, which is not allocated
+in CN or US.
+
+Tested on a CMCC PZ-L8 (IPQ5018 + QCN6122) with the 5 GHz radio as
+STA and AP on channel 149/HE80.
+
+Patch authored with assistance by V4Pro (DeepSeek Harness), tested
+and submitted by Autumn Shiu.
+
+Signed-off-by: Autumn Shiu <autumnshiu@foxmail.com>
```
openwrt-ai, line 29: The upstream-submission claim is still not resolvable from the tree. The commit message carries `Message-ID: <pz-l8-unii3-1786806458770268849-1000@autumn-NucBox-K6.local>` — a locally generated ID from a `.local` host, which nobody can look up — and the patch itself has no reference at all. The in-tree convention for a carried upstream-pending fix is a `Link:` line pointing at the archive, e.g. [`101-wifi-ath11k-fix-wrong-usage-of-resource_size-causing.patch:42`](``https://github.com/openwrt/openwrt/blob/3cf57c966033c7a96ea989040d9129b27a03076c/package/kernel/mac80211/patches/ath11k/101-wifi-ath11k-fix-wrong-usage-of-resource_size-causing.patch#L42)`` and [`ath12k/700-ath12k-read-calibration-variant-from-dt.patch:20`](``https://github.com/openwrt/openwrt/blob/3cf57c966033c7a96ea989040d9129b27a03076c/package/kernel/mac80211/patches/ath12k/700-ath12k-read-calibration-variant-from-dt.patch#L20).``

Please add `Link: https://lore.kernel.org/linux-wireless/<message-id>/` above the `Signed-off-by:` here once the posting appears in the archive. Without it a maintainer has no way to tell when this patch can be dropped, and no way to check that the version carried here matches what was posted.

---

package/kernel/mac80211/patches/ath11k/930-wifi-ath11k-fix-QCN6122-UNII-3-channels-149-165-disa.patch
```diff
@@ -0,0 +1,55 @@
+From 389f63fbe4deb21dadb21dbd48154fdfd84e812a Mon Sep 17 00:00:00 2001
+From: Autumn Shiu <autumnshiu@foxmail.com>
+Date: Sat, 15 Aug 2026 15:31:00 +0800
+Subject: [PATCH] wifi: ath11k: fix QCN6122 UNII-3 channels 149-165 disabled
```
openwrt-ai, line 4: nit: the filename `930-wifi-ath11k-fix-qcn6122-unii3.patch` doesn't track this `Subject:`. Sibling patches in this directory use the `git format-patch` derived name — subject verbatim including capitalisation, `:`/space collapsed to `-`, truncated at 52 chars — e.g. [`947-wifi-ath11k-fix-rssi-station-dump-for-IPQ5018-and-QC.patch`](``https://github.com/openwrt/openwrt/blob/3cf57c966033c7a96ea989040d9129b27a03076c/package/kernel/mac80211/patches/ath11k/947-wifi-ath11k-fix-rssi-station-dump-for-IPQ5018-and-QC.patch#L4)`` for `wifi: ath11k: fix rssi station dump for IPQ5018 and QCN6122`. `930-wifi-ath11k-fix-QCN6122-UNII-3-channels-149-165-disa.patch` would match this subject.

---

openwrt-ai (commented):
Re-reviewed after the force-push (`fbaaa19` → `3cf57c9`). Both new commit messages match their diffs, and the previous concern about the patch's provenance framing is partly addressed — the message no longer claims an ambiguous "pending" status and the `5825` vs `5850` rationale is now explained. Remaining points inline.

Verified against the ath11k source in `gregkh/linux` v6.18.44 (matching the `PKG_VERSION:=6.18.39` backports series):

- The patch's core premise holds: despite the name, `high_5ghz_chan` is a frequency in MHz, not a channel number — [`ath11k_mac_update_ch_list()` compares it against `band->channels[i].center_freq`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/ath/ath11k/mac.c#L9956-L9959), and it is elsewhere compared against [`ATH11K_MIN_6G_FREQ` (5925)](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/ath/ath11k/core.h#L577). Clamping to 5825 admits ch165 and leaves ch169 out, as the message states, and stays below the 6 GHz threshold so it cannot accidentally trip the `supports_6ghz` path.
- The clamp lands on the struct actually used: QCN6122 sets [`.single_pdev_only = false`](``https://github.com/openwrt/openwrt/blob/3cf57c966033c7a96ea989040d9129b27a03076c/package/kernel/mac80211/patches/ath11k/920-wifi-ath11k-add-hw-params-for-QCN6122.patch#L68),`` so `temp_reg_cap` is never re-pointed at a different `hal_reg_cap[]` entry and the mutation takes effect.
- `ATH11K_HW_QCN6122_HW10` is introduced by `920-wifi-ath11k-add-hw-params-for-QCN6122.patch`, which applies before this one — numbering is fine.
- The in-place write to `reg_cap` mutates the shared `ar->ab->hal_reg_cap[]`, which is also read by `ath11k_fill_band_to_mac_param()`; I checked the ordering and it is harmless here (`ath11k_wmi_cmd_init()` runs at core start, well before `ath11k_mac_register()`), so no change requested — just be aware upstream may ask for a local variable.

`PKG_RELEASE` 3 → 4 is correct for adding a patch.

---

autumnshiu009-ship-it:
Updated again:

- Removed the raw `git` artefacts (`diff --git`, `index`, and the `-- / 2.43.0` trailer) from the embedded patch; it now starts at `--- a/...` like the other quilt patches.
- Renamed the file to `930-wifi-ath11k-fix-QCN6122-UNII-3-channels-149-165-disa.patch` to match the Subject.

The `Link:` line will be added as soon as the posting is visible on lore.kernel.org/linux-wireless; the patch email was sent on 2026-08-15 with Message-ID `pz-l8-unii3-1786806458770268849-1000@autumn-NucBox-K6.local`. Until then the commit carries the same Message-ID so the two can be correlated.

target/linux/qualcommax/dts/ipq5018-pz-l8.dts
```diff
@@ -358,7 +358,7 @@
 	qcom,bdf-addr = <0x4d100000>;
 	qcom,m3-dump-addr = <0x4df00000>;
 
-	ieee80211-freq-limit = <5150000 5730000>;
+	ieee80211-freq-limit = <5150000 5850000>;
```
openwrt-ai, line 361: Both radio nodes on this board are disabled in-tree — [`&wifi { ... status = "disabled"; }` at ipq5018-pz-l8.dts:340-350](https://github.com/openwrt/openwrt/blob/ecc3f20f347aec55e5ef9768ddab959362c78eca/target/linux/qualcommax/dts/ipq5018-pz-l8.dts#L340-L350) and [`&wifi1` at :352-364](https://github.com/openwrt/openwrt/blob/ecc3f20f347aec55e5ef9768ddab959362c78eca/target/linux/qualcommax/dts/ipq5018-pz-l8.dts#L352-L364), with the comment block above `&q6v5_wcss` giving the OOM rationale. PZ-L8 is the only ipq50xx board in this state; every sibling (`ipq5018-ax830.dts`, `-mx2000.dts`, `-gl-b3000.dts`, `-scr50axe.dts`, …) sets `status = "okay"` on its `&wifi*` nodes.

So on a stock image this hunk changes nothing — the 5 GHz radio never probes, and the "tested as STA and AP on channel 149/HE80" result must come from a tree where the disable was dropped too. Is that intentional (land the correct limit now, re-enable the radios separately), or should this PR also flip `&wifi1` to `okay`? Either way it would help to say so in the commit message, since as written it reads like a user-visible fix.

---

openwrt-ai (commented):
Re-reviewed after the force-push (`3cf57c9` → `ecc3f20`). The only change is to the embedded ath11k patch: rename to `930-wifi-ath11k-fix-QCN6122-UNII-3-channels-149-165-disa.patch` and removal of the `diff --git`/`index` lines and the `-- / 2.43.0` trailer. Both points from the last round are addressed, and the result now matches the in-tree convention — e.g. [`931-wifi-ath11k-Support-to-assign-m3-dump-memory.patch`](``https://github.com/openwrt/openwrt/blob/ecc3f20f347aec55e5ef9768ddab959362c78eca/package/kernel/mac80211/patches/ath11k/931-wifi-ath11k-Support-to-assign-m3-dump-memory.patch)`` and [`947-wifi-ath11k-fix-rssi-station-dump-for-IPQ5018-and-QC.patch`](``https://github.com/openwrt/openwrt/blob/ecc3f20f347aec55e5ef9768ddab959362c78eca/package/kernel/mac80211/patches/ath11k/947-wifi-ath11k-fix-rssi-station-dump-for-IPQ5018-and-QC.patch)`` both keep the `From <sha> Mon Sep 17`/`From:`/`Date:`/`Subject:` block plus the `---` diffstat and start the payload at `--- a/...`. The 52-char filename truncation matches what `git format-patch` produces, and `930` slots correctly after the QCN6122 support patches (`920`-`925`).

Re-checked the hunk context against `gregkh/linux` v6.18.44 (matching the `PKG_VERSION:=6.18.39` backports series): [`reg_cap = &ar->ab->hal_reg_cap[ar->pdev_idx]; temp_reg_cap = reg_cap;` at mac.c:9992-9995](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/ath/ath11k/mac.c#L9992-L9995) still matches, so the stripped hunk applies with only an offset. Both CI runs on `ecc3f20` are green (Build Kernel, Build all core packages).

Still outstanding from the previous round, no action needed from me: the `Link:` to the lore posting is not yet in the commit message — the `Message-ID:` trailer is the only pointer, and its `@autumn-NucBox-K6.local` domain isn't something a reviewer can resolve. Worth adding the lore URL before merge.

One new question inline on the DTS side.

---

target/linux/qualcommax/dts/ipq5018-pz-l8.dts
autumnshiu009-ship-it: That's intentional: the in-tree `&wifi`/`&wifi1` nodes are disabled to avoid OOM on this memory-constrained board, so I left them disabled and kept this PR a metadata-only change for stock images. The tested PZ-L8 runs a local build that re-enables the radios before applying this fix; stock images are unaffected.

I've amended the DTS commit message to state this explicitly. Re-enabling Wi-Fi by default is a separate decision and out of scope for this PR.

autumnshiu009-ship-it:
Pushed a new revision addressing the two outstanding points:

1. **DTS inline comment**: the `&wifi`/`&wifi1` nodes stay disabled by design (OOM avoidance). The DTS commit message now states that this change is metadata-only for stock builds and that testing was done on a local build with `wifi1` re-enabled.
2. **Upstream `Link:`**: while checking the mailbox I found the SMTP provider had replaced my locally generated Message-ID with its own. The actual linux-wireless posting uses Message-ID `tencent_CA929E144EEEC926E8CFF32B18ED3A26A907@qq.com` and is archived at <https://marc.info/?l=linux-wireless&m=178680667027761>. The embedded patch and the commit now carry:

   `Link: https://lore.kernel.org/linux-wireless/tencent_CA929E144EEEC926E8CFF32B18ED3A26A907@qq.com/`

The PR description has been corrected as well.

openwrt-ai (commented):

Re-reviewed after the force-push (`ecc3f20` → `eaebf5f`). The only file change in this revision is inside `930-wifi-ath11k-fix-QCN6122-UNII-3-channels-149-165-disa.patch`: the `Date:` header and a new `Link: https://lore.kernel.org/linux-wireless/tencent_...@qq.com/` placed directly above `Signed-off-by:`, which is the right spot and matches the in-tree convention I pointed at last round. Both commit messages still match their diffs, and the DTS commit now states the metadata-only nature explicitly, which closes out the `&wifi1`/`status = "disabled"` question from the previous round.

Notes from re-verification, no action requested:

- Hunk context still matches `gregkh/linux` v6.18.44 (the series behind `PKG_VERSION:=6.18.39`): [`reg_cap = &ar->ab->hal_reg_cap[ar->pdev_idx]; temp_reg_cap = reg_cap;` in `ath11k_mac_setup_channels_rates()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/ath/ath11k/mac.c#L9992-L9995) is unchanged, so the patch applies with an offset only.
- The 5825 value still behaves as the message describes against the channel table — [`CHAN5G(165, 5825, 0)` / `CHAN5G(169, 5845, 0)`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/ath/ath11k/mac.c#L97-L98) — and stays below [`ATH11K_MIN_6G_FREQ`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/ath/ath11k/core.h#L577), so the `supports_6ghz` branch is not tripped.
- `PKG_RELEASE` is at 4, patch numbering `930` still slots after the QCN6122 support patches (`920`-`925`) with no collision.
- All jobs on `eaebf5f` are green (Build all affected Kernels, Build all core packages for selected target) — no failures to attribute.
- I could not verify the lore.kernel.org URL from this environment (outbound HTTPS is restricted to GitHub here), so I am taking it at face value rather than flagging it. A maintainer should confirm it resolves before merge, since it is the only handle for knowing when to drop this patch.

The two earlier inline threads are still open and unchanged, so I have not re-posted them: the `hw_rev`-keyed clamp applying to all ten QCN6122 boards, and the in-code comment wording ("standard 5.8 GHz upper bound") not matching the 5825 rationale now spelled out in the commit message.

---

autumnshiu009-ship-it:
Rebased the two commits onto current `openwrt/openwrt` main (`3d902d07a9d3`) because the PR branch had drifted 99 commits behind. No content changes in either commit; the tree is identical to the previous `eaebf5f` revision except for the new base.

PR #24749, opened 2026-08-15
realtek: rtl930x: add support for Hasivo S600WP-5GT-2S+

pktpls:
This commit adds support for the Hasivo S600WP-5GT-2S+ switch with SFP+ and PoE.
Based on board revision `RTL_5GT-2SX+ v1.02`.

This device is nearly identical to Hasivo S600WP-5GT-2SX-SE, except:
- More RAM and flash
- RJ45 console port
- Model name scheme, `2S+` vs `2SX`
- No case cut-out for the PoE LEDs
- No MAC address in nvmem

Hardware
--------

|          |                                                                |
|----------|----------------------------------------------------------------|
| SoC      | RTL9303 rev B                                                  |
| RAM      | 256 MB Samsung K4B2G1646F DDR3L                                |
| Flash    | 32 MB Macronix MX25L25645G SPI NOR, 29 MB usable by OpenWrt    |
| Ethernet | 5x RJ45 via 5x RTL8221B-VB-CG PHYs (2.5G/1G/100M/10M),         |
|          | 2x SFP+ via SoC (10G/2.5G/1G)                                  |
| PoE      | 1x 802.3bt 90 W (port 1)                                       |
|          | 3x 802.3at 30 W (ports 2, 3, 4)                                |
|          | via daughter board with Hasivo HS104PTI controller             |
| LEDs     | 5x orange/green 2.5G/1G link, 5x green 100M/10M link           |
|          | 4x orange internal PoE, 2x green 10G link, 2x green 1G link    |
|          | 1x green system, 1x green power                                |
| Button   | Reset                                                          |
| Console  | RJ45 38400 bps 8n1                                             |
| Power    | Barrel jack 48V, shipped PSU is 52V/1.25A                      |

Installing OpenWrt
------------------

1. Attach to RJ45 serial console port using a cisco cable.
2. Attach your computer to any RJ45 port.
3. Serve initramfs-kernel.bin on TFTP 192.168.1.111.
4. Power on the device.
5. Interrupt U-Boot by pressing `Ctrl+C`, then `Z`, then `H`, during 3 second countdown.
6. Run: `rtk network on`
7. Run: `tftpboot 0x84f00000 initramfs-kernel.bin ; bootm 0x84f00000`
8. Use `mtd dump` to make backups of all flash partitions.
9. Use SCP to copy `squashfs-sysupgrade.bin` to the device, then run `sysupgrade`.

Restoring factory firmware
--------------------------

OpenWrt uses the `RUNTIME` and `RUNTIME2` partitions as one combined partition.
To restore them from backups, boot from `initramfs-kernel.bin` just like during
the installation, then use `mtd write` to write your backups of the factory
`mtd5` and `mtd6` partitions.

Notes/Quirks
------------

- U-Boot interruption is obfuscated. Press `Ctrl+C`, then `Z`, then `H`,
  during the 3 second countdown.
- MAC address is stored on the `RUNTIME` or `RUNTIME2` partitions, which are used by OpenWrt.
  Instead, we generate one random MAC address and store it in the U-Boot environment.
- The PoE LEDs under each RJ45 port work correctly, but there's no cut-out in the case,
  so these LEDs are only barely visible.
- Under certain circumstances, the RJ45 ports can come up dead on boot.
  See https://github.com/openwrt/openwrt/issues/22140

### DIFF
--- a/target/linux/realtek/dts/rtl9303_hasivo_s600wp-5gt-2s-plus.dts
+++ b/target/linux/realtek/dts/rtl9303_hasivo_s600wp-5gt-2s-plus.dts
@@ -1,0 +1,377 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/dts-v1/;
+
+#include "rtl930x.dtsi"
+
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/leds/common.h>
+
+/ {
+	compatible = "hasivo,s600wp-5gt-2s-plus", "realtek,rtl9303-soc";
+	model = "Hasivo S600WP-5GT-2S+";
+
+	memory@0 {
+		device_type = "memory";
+		reg = <0x00000000 0x10000000>; /* 256 MiB */
+	};
+
+	aliases {
+		led-boot = &led_sys;
+		led-failsafe = &led_sys;
+		led-running = &led_sys;
+		led-upgrade = &led_sys;
+		i2c0 = &i2c_scl23_sda22;
+		i2c1 = &i2c_sfp0;
+		i2c2 = &i2c_sfp1;
+	};
+
+	chosen {
+		stdout-path = "serial0:38400n8";
+	};
+
+	keys {
+		compatible = "gpio-keys";
+
+		button-reset {
+			label = "reset";
+			gpios = <&gpio0 5 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		pinctrl-names = "default";
+		pinctrl-0 = <&pinmux_disable_sys_led>, <&pinmux_enable_led_sync>;
+
+		led_sys: led-0 {
+			gpios = <&gpio0 0 GPIO_ACTIVE_HIGH>;
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_STATUS;
+			linux,default-trigger = "heartbeat";
+		};
+	};
+
+	led_set {
+		compatible = "realtek,rtl9300-leds";
+
+		/* Port 25 is unused but present in the HC595 shift
+		 * register chain. Force it into the serial LED stream
+		 * to align all LED positions correctly.
+		 */
+		realtek,led-set0-force-port-mask = <(1 << 25)>;
+
+		/*
+		 * RJ45 LED0 (green): 1G/LINK/ACT
+		 * RJ45 LED1 (green): 10M/100M/LINK/ACT
+		 * RJ45 LED2 (orange): 2.5G/LINK/ACT
+		 */
+		led_set0 = <(RTL93XX_LED_SET_1G | RTL93XX_LED_SET_LINK | RTL93XX_LED_SET_ACT)
+			    (RTL93XX_LED_SET_10M | RTL93XX_LED_SET_100M |
+			     RTL93XX_LED_SET_LINK | RTL93XX_LED_SET_ACT)
+			    (RTL93XX_LED_SET_2P5G | RTL93XX_LED_SET_LINK | RTL93XX_LED_SET_ACT)>;
+
+		/*
+		 * SFP+ LED0: 10G/LINK/ACT
+		 * SFP+ LED1: 1G/LINK/ACT
+		 * SFP+ LED2: spacer to keep 3 slots per port
+		 */
+		led_set1 = <(RTL93XX_LED_SET_10G | RTL93XX_LED_SET_LINK | RTL93XX_LED_SET_ACT)
+			    (RTL93XX_LED_SET_1G | RTL93XX_LED_SET_LINK | RTL93XX_LED_SET_ACT)
+			    (RTL93XX_LED_SET_NONE)>;
+	};
+
+	/* I2C bus to the on-board HS104 PSE controller (PoE) and STC8 LED MCU. */
+	i2c_scl23_sda22: i2c-scl23-sda22 {
+		compatible = "i2c-gpio";
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		scl-gpios = <&gpio0 23 (GPIO_ACTIVE_HIGH | GPIO_OPEN_DRAIN)>;
+		sda-gpios = <&gpio0 22 (GPIO_ACTIVE_HIGH | GPIO_OPEN_DRAIN)>;
+		clock-frequency = <100000>;
+
+		pse0: ethernet-pse@d {
+			compatible = "hasivo,hs104";
+			reg = <0x0d>;
+			poll-interval-ms = <500>;
+			status = "okay";
+
+			pse-pis {
+				#address-cells = <1>;
+				#size-cells = <0>;
+
+				pse0_pi0: pse-pi@0 {
+					reg = <0>;
+					#pse-cells = <0>;
+				};
+				pse0_pi1: pse-pi@1 {
+					reg = <1>;
+					#pse-cells = <0>;
+				};
+				pse0_pi2: pse-pi@2 {
+					reg = <2>;
+					#pse-cells = <0>;
+				};
+				pse0_pi3: pse-pi@3 {
+					reg = <3>;
+					#pse-cells = <0>;
+				};
+			};
+		};
+
+		stc8: stc8@4d {
+			compatible = "hasivo,stc8-mfd", "syscon";
+			reg = <0x4d>;
+			#address-cells = <2>;
+			#size-cells = <0>;
+			hasivo,execute-bit = <0x40>;
+			hasivo,execute-bit-registers = <0x01 0x02>;
+			status = "okay";
+
+			led_lan1_poe: led@1,0 {
+				compatible = "register-bit-led";
+				reg = <0x01 0x00>;
+				mask = <0x08>;
+				color = <LED_COLOR_ID_ORANGE>;
+				function = "poe";
+				function-enumerator = <1>;
+				linux,default-trigger = "pse-0-000d:port0:delivering";
+				default-state = "off";
+			};
+
+			led_lan2_poe: led@1,1 {
+				compatible = "register-bit-led";
+				reg = <0x01 0x01>;
+				mask = <0x04>;
+				color = <LED_COLOR_ID_ORANGE>;
+				function = "poe";
+				function-enumerator = <2>;
+				linux,default-trigger = "pse-0-000d:port1:delivering";
+				default-state = "off";
+			};
+
+			led_lan3_poe: led@1,2 {
+				compatible = "register-bit-led";
+				reg = <0x01 0x02>;
+				mask = <0x02>;
+				color = <LED_COLOR_ID_ORANGE>;
+				function = "poe";
+				function-enumerator = <3>;
+				linux,default-trigger = "pse-0-000d:port2:delivering";
+				default-state = "off";
+			};
+
+			led_lan4_poe: led@1,3 {
+				compatible = "register-bit-led";
+				reg = <0x01 0x03>;
+				mask = <0x01>;
+				color = <LED_COLOR_ID_ORANGE>;
+				function = "poe";
+				function-enumerator = <4>;
+				linux,default-trigger = "pse-0-000d:port3:delivering";
+				default-state = "off";
+			};
+		};
+	};
+
+	i2c_sfp0: i2c-sfp0 {
+		compatible = "i2c-gpio";
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		scl-gpios = <&gpio0 7 (GPIO_ACTIVE_HIGH | GPIO_OPEN_DRAIN)>;
+		sda-gpios = <&gpio0 6 (GPIO_ACTIVE_HIGH | GPIO_OPEN_DRAIN)>;
+		i2c-gpio,delay-us = <50>;
+	};
+
+	i2c_sfp1: i2c-sfp1 {
+		compatible = "i2c-gpio";
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		scl-gpios = <&gpio0 11 (GPIO_ACTIVE_HIGH | GPIO_OPEN_DRAIN)>;
+		sda-gpios = <&gpio0 10 (GPIO_ACTIVE_HIGH | GPIO_OPEN_DRAIN)>;
+		i2c-gpio,delay-us = <50>;
+	};
+
+	sfp0: sfp-p26 {
+		compatible = "sff,sfp";
+		i2c-bus = <&i2c_sfp0>;
+		mod-def0-gpios = <&gpio0 9 GPIO_ACTIVE_LOW>;
+		tx-disable-gpios = <&gpio0 8 GPIO_ACTIVE_HIGH>;
+		maximum-power-milliwatt = <2500>;
+	};
+
+	sfp1: sfp-p27 {
+		compatible = "sff,sfp";
+		i2c-bus = <&i2c_sfp1>;
+		mod-def0-gpios = <&gpio0 13 GPIO_ACTIVE_LOW>;
+		maximum-power-milliwatt = <2500>;
+		/*
+		 * Stock firmware wires tx-disable to GPIO 12 (FB_TXDIS_1).
+		 * The pin is not claimed by the hardware I2C controller
+		 * (SDA pinmux is in GPIO mode) but still cannot be driven
+		 * HIGH for unknown reasons. Omitted for now.
+		 */
+	};
+};
+
+&spi0 {
+	status = "okay";
+
+	flash@0 {
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <10000000>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			/* stock is LOADER */
+			partition@0 {
+				label = "u-boot";
+				reg = <0x0000000 0x00e0000>;
+				read-only;
+			};
+
+			/* stock is BDINFO */
+			partition@e0000 {
+				label = "u-boot-env";
+				reg = <0x00e0000 0x0010000>;
+			};
+
+			/* stock is SYSINFO */
+			partition@f0000 {
+				label = "u-boot-env2";
+				reg = <0x00f0000 0x0010000>;
+			};
+
+			/* stock is CFG JFFS2 */
+			partition@100000 {
+				label = "jffs";
+				reg = <0x0100000 0x0100000>;
+			};
+
+			/* stock is LOG JFFS2 */
+			partition@200000 {
+				label = "jffs2";
+				reg = <0x0200000 0x0100000>;
+			};
+
+			/* stock is RUNTIME and RUNTIME2
+			 * RUNTIME is <0x0300000 0x0700000>
+			 * RUNTIME2 is <0x0a00000 0x1600000>
+			 */
+			partition@300000 {
+				compatible = "openwrt,uimage", "denx,uimage";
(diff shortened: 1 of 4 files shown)


pktpls:
Known problems:
- ~~Device sometimes hangs after `reboot: Restarting system`~~ probably a general rtl93xx issue
- ~~Port 1 sometimes stuck at 100M and not passing traffic~~ same issue as in #22140, likely out of scope for this PR

target/linux/realtek/image/rtl930x.mk
```diff
@@ -113,6 +113,17 @@ define Device/hasivo_s600wp-5gt-2sx-se
 endef
 TARGET_DEVICES += hasivo_s600wp-5gt-2sx-se
 
+define Device/hasivo_s600wp-5gt-2s-plus
+  SOC := rtl9303
+  DEVICE_VENDOR := Hasivo
+  DEVICE_MODEL := S600WP-5GT-2S+
+  DEVICE_PACKAGES := kmod-pse-hasivo-hs104 kmod-mfd-hasivo-stc8 \
+  gpiod-tools i2c-tools ethtool-full
```
pktpls: Development packages need to be removed

target/linux/realtek/image/rtl930x.mk
```diff
@@ -113,6 +113,17 @@ define Device/hasivo_s600wp-5gt-2sx-se
...
+  IMAGE_SIZE := 12288k
```
pktpls, line 121: Correct value is 29696k

jonasjelonek:
>     * Device sometimes hangs after `reboot: Restarting system`

probably not only this device is affected. I've noticed this also on some RTL931x devices, though not gave it more attention up to now.

pktpls:
> probably not only this device is affected. I've noticed this also on some RTL931x devices, though not gave it more attention up to now.

True, I remember the occasional reboot hang, too. Anyway now all reboots have been successful since I've posted this earlier. I'm not 100% sure anymore but initramfs vs. sysupgrade might make a difference?

---

About port 1 - if the device is powered on with something attached to port 1, the serdes fails calibration. I've tested with two devices, so it doesn't seem like a one-off hardware defect. It also fails with the PoE board removed. If the neighbor is attached after we've finished booting, the port works as expected.

The first device outright fails RX calibration, the port is unresponsive and only recovers with a powercycle:
```
[   10.532996] rtl83xx-switch 1b000000.ethernet-switch lan1: configuring for phy/usxgmii link mode
[   10.562021] rtl838x_eth 1b000000.ethernet-switch:ethernet eth0: Link is Up - 1Gbps/Full - flow control off
[   11.041330] realtek-otto-pcs 1b000000.ethernet-switch:pcs: SerDes 2: symbol error rate too high
[   11.491326] realtek-otto-pcs 1b000000.ethernet-switch:pcs: SerDes 2: symbol error rate too high
[   11.941368] realtek-otto-pcs 1b000000.ethernet-switch:pcs: SerDes 2: symbol error rate too high
[   11.970224] realtek-otto-pcs 1b000000.ethernet-switch:pcs: SerDes 2: RX calibration failed
[   11.997628] realtek-otto-pcs 1b000000.ethernet-switch:pcs: SerDes 2 configured for usxgmii mode
[   12.059493] 8021q: adding VLAN 0 to HW filter on device lan1
[   12.092909] rtl83xx-switch 1b000000.ethernet-switch lan1: Link is Up - 2.5Gbps/Full - flow control rx/tx
Press the [f] key and hit [enter] to enter failsafe mode
Press the [1], [2], [3] or [4] key and hit [enter] to select the debug level
[   16.712856] urandom-seed: Seeding with /etc/urandom.seed
[   16.930105] rtl83xx-switch 1b000000.ethernet-switch lan1: Link is Down
[   16.966544] procd: - early -
```

The second device seems to complete that calibration step, but never properly establishes a link. When it detects anything at all, it often ends up as 1G or 100M, but still doesn't pass any traffic. The port LED is lit but doesn't blink for incoming pings.

jonasjelonek:
> I'm not 100% sure anymore but initramfs vs. sysupgrade might make a difference?

Could be. I'm not 100% sure anymore but I stumbled upon this mostly on initramfs boot, never (or I just don't remember) on flash boot.

> About port 1 - if the device is powered on with something attached to port 1, the serdes fails calibration. I've tested with two devices, so it doesn't seem like a one-off hardware defect. It also fails with the PoE board removed. If the neighbor is attached after we've finished booting, the port works as expected.

Wait ... Given those are RTL8221B as you say, defining them with usxgmii is wrong. Those PHYs just support up to 2.5G. Basically the same PHYs as in XGS1210-12 B1. They should be defined with `phy-mode = "sgmii";` and `managed = "in-band-status";`

pktpls:
> Wait ... Given those are RTL8221B as you say, defining them with usxgmii is wrong. Those PHYs just support up to 2.5G. Basically the same PHYs as in XGS1210-12 B1. They should be defined with phy-mode = "sgmii"; and managed = "in-band-status";

Uuugh yes that's my own copy+paste fuckup, thanks for spotting this :)

pktpls:
Ok, with sgmii/in-band-status the error is now consistent, and affects all ports :) It looks like it's the same issue as https://github.com/openwrt/openwrt/issues/22140. I also noticed @carlosz1986 that you had the same issue with the other S600WP model.

```
[   13.669617] In rteth_mac_config, mode 1
[   13.683215] rtl83xx-switch 1b000000.ethernet-switch lan1: configuring for inband/sgmii link mode
[   13.712544] rtl838x_eth 1b000000.ethernet-switch:ethernet eth0: Link is Up - 1Gbps/Full - flow control off
[   14.181414] realtek-otto-pcs 1b000000.ethernet-switch:pcs: SerDes 2 configured for sgmii mode
[   14.312832] RTL8221B-VB-CG 2.5Gbps PHY realtek-mdio-0:01: rtl822x_serdes_write failed: -145
[   14.340599] rtl83xx-switch 1b000000.ethernet-switch lan1: phy_config_inband: -ETIMEDOUT
[   14.389568] 8021q: adding VLAN 0 to HW filter on device lan1
```

pktpls:
Yes, with @janh's patch the issue is gone:
```
> cat target/linux/generic/hack-6.18/726-net-phy-avoid-MMD-30-with-RTL8221B.patch
--- a/drivers/net/phy/phy_device.c
+++ b/drivers/net/phy/phy_device.c
@@ -926,6 +926,17 @@ static int get_phy_c45_devs_in_pkg(struc
        return 0;
 }

+static bool phy_c45_skip_mmd_read(int mmd, struct phy_c45_device_ids *c45_ids)
+{
+       if (c45_ids->device_ids[MDIO_MMD_PMAPMD] == 0x001cc849 ||
+                       c45_ids->device_ids[MDIO_MMD_PMAPMD] == 0x001cc84a) {
+               /* Reading MMD 30 on RTL8221B may cause the PHY to stop working. */
+               return mmd == MDIO_MMD_VEND1;
+       }
+
+       return false;
+}
+
 /**
  * get_phy_c45_ids - reads the specified addr for its 802.3-c45 IDs.
  * @bus: the target MII bus
@@ -993,6 +1004,10 @@ static int get_phy_c45_ids(struct mii_bu
                if (!(devs_in_pkg & (1 << i)))
                        continue;

+               /* Some PHYs react badly to reading from specific MMDs. */
+               if (phy_c45_skip_mmd_read(i, c45_ids))
+                       continue;
+
                if (i == MDIO_MMD_VEND1 || i == MDIO_MMD_VEND2) {
                        /* Probe the "Device Present" bits for the vendor MMDs
                         * to ignore these if they do not contain IEEE 802.3
```

(Although it would be nicer to do it via the MDIO driver and device tree, as also proposed in #22140.)

target/linux/realtek/dts/rtl9303_hasivo_s600wp-5gt-2s-plus.dts
```diff
@@ -0,0 +1,377 @@
...
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		port@0 {
+			reg = <0>;
+			label = "lan1";
+			pcs-handle = <&serdes2 0>;
+			phy-handle = <&phy0>;
+			phy-mode = "sgmii";
+			managed = "in-band-status";
```
openwrt-ai, line 303: nit: the DTS was corrected to `sgmii`/`in-band-status` (max 2.5G), but the hardware table in the commit message and PR body still says the RJ45 ports are `10G/5G/2.5G/1G/100M/10M`. RTL8221B-VB-CG is a 2.5G-only PHY (the kernel even names it that way — see the boot log in [this comment](https://github.com/openwrt/openwrt/pull/24749#issuecomment-5304657606)), so the Ethernet row should read something like `2.5G/1G/100M/10M`. Same applies to the four other port nodes below.

---

target/linux/realtek/dts/rtl9303_hasivo_s600wp-5gt-2s-plus.dts
```diff
@@ -0,0 +1,377 @@
...
+				default-state = "off";
+			};
+
+			led_lan4_poe: led@1,3 {
+				compatible = "register-bit-led";
+				reg = <0x01 0x03>;
+				mask = <0x01>;
+				color = <LED_COLOR_ID_ORANGE>;
+				function = "poe";
+				function-enumerator = <4>;
```
openwrt-ai, line 173: nit: the LED row of the hardware table doesn't match this file. It lists `5x orange internal PoE` twice (once at the end of the second line, once at the end of the third), while only four PoE LEDs are defined here (`led_lan1_poe` … `led_lan4_poe`), matching the four PoE ports. The same row's `4x green 10G/1G link` also doesn't line up with `led_set0`, which assigns LED0 (green, 1G) to all five RJ45 ports.

---

target/linux/realtek/dts/rtl9303_hasivo_s600wp-5gt-2s-plus.dts
```diff
@@ -0,0 +1,377 @@
...
+
+#include "rtl930x.dtsi"
+
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/leds/common.h>
+
+/ {
+	compatible = "hasivo,s600wp-5gt-2s-plus", "realtek,rtl9303-soc";
+	model = "Hasivo S600WP-5GT-2S+";
```
openwrt-ai, line 12: nit: the hardware specification in the commit message is missing the power supply line (input voltage + current), which the new-device checklist asks for. For a PoE switch with a 90 W bt port that's fairly relevant — worth adding a row next to `Console`.

---

target/linux/realtek/dts/rtl9303_hasivo_s600wp-5gt-2s-plus.dts
```diff
@@ -0,0 +1,377 @@
...
+			/* stock is BDINFO */
+			partition@e0000 {
+				label = "u-boot-env";
+				reg = <0x00e0000 0x0010000>;
+			};
+
+			/* stock is SYSINFO */
+			partition@f0000 {
+				label = "u-boot-env2";
+				reg = <0x00f0000 0x0010000>;
```
openwrt-ai, line 251: The near-identical sibling marks the SYSINFO partition read-only ([`rtl9303_hasivo_s600wp-5gt-2sx-se.dts:271-274`](``https://github.com/openwrt/openwrt/blob/5547fa2f5ee222cc693946dc1f47d8678d1c2eae/target/linux/realtek/dts/rtl9303_hasivo_s600wp-5gt-2sx-se.dts#L271-L274));`` here it is writable. That matters because the realtek uboot-envtools default arm registers `u-boot-env2` as the *sys* env ([`uboot-envtools/files/realtek:46-49`](https://github.com/openwrt/openwrt/blob/5547fa2f5ee222cc693946dc1f47d8678d1c2eae/package/boot/uboot-tools/uboot-envtools/files/realtek#L46-L49)), so `fw_setsys` would now be able to overwrite stock SYSINFO data on this board. Is dropping `read-only` intentional?

```suggestion
			partition@f0000 {
				label = "u-boot-env2";
				reg = <0x00f0000 0x0010000>;
				read-only;
```

---

target/linux/realtek/dts/rtl9303_hasivo_s600wp-5gt-2s-plus.dts
```diff
@@ -0,0 +1,377 @@
...
+			};
+
+			/* stock is RUNTIME and RUNTIME2
+			 * RUNTIME is <0x0300000 0x0700000>
+			 * RUNTIME2 is <0x0a00000 0x1600000>
+			 */
+			partition@300000 {
+				compatible = "openwrt,uimage", "denx,uimage";
+				label = "firmware";
+				reg = <0x0300000 0x1d00000>;
```
openwrt-ai, line 273: Because RUNTIME and RUNTIME2 are merged into a single `firmware` partition here, the "Restoring factory firmware" instruction ("use `mtd write` to write your backups of the factory `mtd5` and `mtd6` partitions") has no matching target on the OpenWrt side: OpenWrt's `mtd5` is the whole 29 MiB `firmware`, and `mtd6`/`mtd7` are the `kernel`/`rootfs` sub-partitions produced by the `openwrt,uimage` parser — writing the stock RUNTIME2 dump to OpenWrt's `mtd6` would land at the wrong offset. Worth spelling out the exact commands, e.g. write the RUNTIME dump to `firmware` and the RUNTIME2 dump at offset `0x700000` within it (or concatenate the two dumps first).

---

target/linux/realtek/image/rtl930x.mk
```diff
@@ -113,6 +113,16 @@ define Device/hasivo_s600wp-5gt-2sx-se
 endef
 TARGET_DEVICES += hasivo_s600wp-5gt-2sx-se
 
+define Device/hasivo_s600wp-5gt-2s-plus
```
openwrt-ai: nit: the Hasivo block in this file is alphabetically sorted (`f1100w-*` → `f1100wp-*` → `s1100w-*` → `s1100wp-8gt-se` → `s1100wp-8xgt-se` → `s600wp-*`). `hasivo_s600wp-5gt-2s-plus` sorts before `hasivo_s600wp-5gt-2sx-se` (`-` < `x`), so this block belongs above the `hasivo_s600wp-5gt-2sx-se` block at line 106 rather than below it. The `02_network` and `99_fwenv-store-ethaddr` additions are already in the right sort position.

---

openwrt-ai (commented):
Reviewed 1 new commit. The DTS is a clean derivative of the S600WP-5GT-2SX-SE, the partition layout adds up (0x300000 + 0x1d00000 = 32 MiB, matching `IMAGE_SIZE := 29696k`), and the `02_network` / `99_fwenv-store-ethaddr` additions are consistent with the random-MAC approach described in the commit message. CI on `5547fa2` is green. Findings are inline: mostly commit-message/hardware-table mismatches, plus one parity question about `u-boot-env2` no longer being `read-only`.

---

jonasjelonek:
I call that progress :)

I've seen it a couple of times now with different PHYs, they react badly to the PHYID read on certain MMDs, though this isn't always MMD_VEND1 and MMD_VEND2. Maybe we'll have at some point a solution but is upstreamable, I hardly believe this only affects Realtek switches/PHYs. I had that at least with a Broadcom one too.

target/linux/realtek/dts/rtl9303_hasivo_s600wp-5gt-2s-plus.dts
pktpls: No

target/linux/realtek/dts/rtl9303_hasivo_s600wp-5gt-2s-plus.dts
pktpls: No

pktpls:
In addition to addressing Claude's comments, I added `realtek,enable-pma-low-power` to the 2.5G ports, since RTL8221B-VB-CG allegedly supports it. These Hasivos run pretty hot, maybe it helps.

About the MMD issue, I could give the DTS route a try today?

target/linux/realtek/dts/rtl9303_hasivo_s600wp-5gt-2s-plus.dts
```diff
@@ -0,0 +1,383 @@
...
+&mdio_bus1 {
+	/* External RTL8221B-VB-CG PHY */
+	PHY_C45(24, 1)
+};
+
+&phy0 { realtek,enable-pma-low-power; };
+&phy8 { realtek,enable-pma-low-power; };
+&phy16 { realtek,enable-pma-low-power; };
+&phy20 { realtek,enable-pma-low-power; };
+&phy24 { realtek,enable-pma-low-power; };
```
openwrt-ai: `realtek,enable-pma-low-power` has no effect on these PHYs. The property is read exactly once in the tree, in `rtl826x_probe` ([742-net-phy-realtek-add-5G-and-10G-PHY-support.patch:249](``https://github.com/openwrt/openwrt/blob/64c878c4f129e91d0db3fbffa72454e33c815ec8/target/linux/generic/pending-6.18/742-net-phy-realtek-add-5G-and-10G-PHY-support.patch#L249)),`` and it only gates `rtl826x_suspend`. That probe is wired up solely for the 5G/10G parts (RTL8251L / RTL8254B / RTL8261BE / RTL8261N / RTL8264 / RTL8264B) — which is why the sibling that uses it, [`rtl9313_hasivo_s1300wp-8xgt-4s-plus.dts:377-419`](``https://github.com/openwrt/openwrt/blob/64c878c4f129e91d0db3fbffa72454e33c815ec8/target/linux/realtek/dts/rtl9313_hasivo_s1300wp-8xgt-4s-plus.dts#L377-L419),`` has an RTL8264 quad-PHY package.

`phy0`/`phy8`/`phy16`/`phy20`/`phy24` here are RTL8221B-VB-CG, which binds `rtl822x_probe` ([realtek_main.c:1913-1915](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/realtek/realtek_main.c#L1913-L1915), unchanged by [796-v7.0-net-phy-realtek-reunify-C22-and-C45-drivers.patch:87-91](``https://github.com/openwrt/openwrt/blob/64c878c4f129e91d0db3fbffa72454e33c815ec8/target/linux/generic/backport-6.18/796-v7.0-net-phy-realtek-reunify-C22-and-C45-drivers.patch#L87-L91))`` — that probe never looks at the property, so nothing changes for thermals here. Unless the intent is to land driver support for it in the same series, dropping the block keeps the DTS honest:

```suggestion
```

---

openwrt-ai (commented):
Reviewed the changes since `5547fa2`. The `.mk` block is now in the right alphabetical position, and the commit-message hardware table (Ethernet speeds, LED row, Power line) now matches the DTS — those earlier nits look resolved. CI on `64c878c` is fully green.

The one new thing in this push is the `realtek,enable-pma-low-power` block; see the inline comment — as far as I can tell it is inert for RTL8221B-VB-CG.

---

janh:
> About the MMD issue, I could give the DTS route a try today?

Otherwise, I could also create a PR with my patch. But of course an upstream(able) solution would be much better.

pktpls:
Clause was right about `realtek,enable-pma-low-power`.

If we defer the MMD thing, I think this PR might be complete.

> Otherwise, I could also create a PR with my patch. But of course an upstream(able) solution would be much better.

@janh let's continue this in #22140

openwrt-ai (commented):

The only change since `64c878c` is the removal of the five `realtek,enable-pma-low-power` overrides, which resolves the one open finding from the previous round — the removal is clean (no stray blank lines left behind) and the commit message needed no corresponding edit since the property was never described there. Commit message still matches the diff. CI on `31cb5ff` is fully green (all realtek subtarget kernel builds plus the three FormalityCheck jobs).

---

jonasjelonek (approved):

jonasjelonek:

PR #24757, opened 2026-08-16
ipq40xx: enable display feature on generic subtarget

divinehawk:
The Ubiquiti UniFi Travel Router lists kmod-drm-panel-mipi-dbi and kmod-backlight-pwm in DEVICE_PACKAGES. Both resolve through packages gated on @DISPLAY_SUPPORT, which is only selected when the target advertises the 'display' feature. Without it those packages are never built, and the image step fails with:

  ERROR: unable to select packages:
    kmod-backlight-pwm (no such package)
    kmod-drm-panel-mipi-dbi (no such package)

This has broken every ipq40xx/generic snapshot build since r35795.

### DIFF
--- a/target/linux/ipq40xx/generic/target.mk
+++ b/target/linux/ipq40xx/generic/target.mk
@@ -1,3 +1,3 @@
 BOARDNAME:=Generic
-FEATURES+=emmc nand
+FEATURES+=display emmc nand
 DEFAULT_PACKAGES += ath10k-board-qca4019


divinehawk:
@robimarko I believe this is the correct fix for the buildbot failures.

robimarko:

divinehawk:
Apparently this was only a partial fix. Any idea what we are missing?

robimarko:
Why do you mean partial?

Buildbots have not yet run for ipq40xx/generic

divinehawk:
https://github.com/openwrt/openwrt/actions/runs/31941979637/job/95162888159 Maybe a false negative?

robimarko:
Could be as it built for me locally

divinehawk:
Looks like it failed. I'm testing this which might fix it, but I am going to try and reproduce first before another PR. 

https://github.com/divinehawk/openwrt/commit/c4c5396425b792f101596fec3b8d7731639c9735

PR #24758, opened 2026-08-16
qualcommax: ipq807x/ipq60xx: add MTU change and jumbo frames

robimarko:
Currently, the EDMA and PPE drivers do not support changing the MTU nor using jumbo frames.

So, port the logic from SSDK and NSS-DP to allow changing hte MTU and using jumbo frames.
Frames up to 12288 are supported.


target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_main.c
```diff
@@ -508,6 +514,32 @@ static int qca_ppe_set_ageing_time(struct dsa_switch *ds, unsigned int msecs)
 	return 0;
 }
 
+static int qca_ppe_port_change_mtu(struct dsa_switch *ds, int port,
+				   int new_mtu)
+{
+	struct qca_ppe_priv *priv = ds_to_priv(ds);
+	u32 frame_size = new_mtu + ETH_HLEN;
```
openwrt-ai: `frame_size` budgets only the Ethernet header, but [`qca_ppe_port_max_mtu()`](``https://github.com/openwrt/openwrt/blob/016cee17f82ac619cb7014eb57649a58481834a9/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_main.c#L538-L541)`` advertises `PPE_MAX_FRAME_SIZE - ETH_HLEN - (2 * VLAN_HLEN)` (12266), i.e. it explicitly reserves 8 bytes for two VLAN tags, and [`EDMA_MAX_MTU`](``https://github.com/openwrt/openwrt/blob/016cee17f82ac619cb7014eb57649a58481834a9/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.h#L149``) does the same.

At the advertised maximum MTU the MRU/MTU programmed here is 12280, while a double-tagged frame is 12288 — exactly the size the RX buffer and the GMAC `MAXFR`/jumbo registers were sized for, but 8 bytes over the PPE threshold. The same holds at any MTU (a single-tagged 1500-byte frame is 1518 vs a 1514 threshold).

The two functions disagree about whether VLAN tags are in the budget. Either `frame_size` should add `2 * VLAN_HLEN` (and the setup default at line 455 follow suit), or `port_max_mtu()` should stop subtracting it — which did you intend?

---

target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_main.c
```diff
@@ -446,8 +449,11 @@ static int qca_ppe_setup(struct dsa_switch *ds)
 	for (i = 0; i < num_ports; i++) {
 		regmap_write(priv->regmap, PPE_CST_STATE(i), PPE_STP_FORWARDING);
 
-		regmap_write(priv->regmap, PPE_MRU_MTU_CTRL(i),
-			     PPE_DEFAULT_MTU | (PPE_DEFAULT_MTU << PPE_MTU_SHIFT));
+		regmap_write(priv->regmap,
+			     PPE_MRU_MTU_CTRL(i,
+					      priv->data->mru_mtu_ctrl_stride),
+			     FIELD_PREP(PPE_MRU_MTU_CTRL_MRU, PPE_DEFAULT_MTU) |
+			     FIELD_PREP(PPE_MRU_MTU_CTRL_MTU, PPE_DEFAULT_MTU));
```
openwrt-ai: `PPE_MC_MTU_CTRL_MTU` is not initialised anywhere — the only write to that register at probe time is the `TX_CNT_EN` bit in [`ppe_port_cnt_enable()`](``https://github.com/openwrt/openwrt/blob/016cee17f82ac619cb7014eb57649a58481834a9/target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_main.c#L161-L162).``

Now that `qca_ppe_port_change_mtu()` treats the multicast egress threshold as part of the MTU configuration, the initial state is inconsistent: MRU and unicast MTU are set to `PPE_DEFAULT_MTU` here, while the multicast MTU keeps its reset value until userspace changes the MTU at least once. Should this loop also program `FIELD_PREP(PPE_MC_MTU_CTRL_MTU, PPE_DEFAULT_MTU)`?

---

target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_ppe_main.c
```diff
@@ -151,7 +152,9 @@ static void ppe_xgmac_link_up(struct qca_ppe_priv *priv, int port,
 
 static void ppe_port_cnt_enable(struct qca_ppe_priv *priv, int port)
 {
-	regmap_update_bits(priv->regmap, PPE_MRU_MTU_CTRL(port) + 4,
+	regmap_update_bits(priv->regmap,
+			   PPE_MRU_MTU_CTRL(port,
+						priv->data->mru_mtu_ctrl_stride) + 4,
```
openwrt-ai: nit: the continuation is not aligned with the opening parenthesis of `PPE_MRU_MTU_CTRL(`; the two other call sites introduced by this same commit (lines 453-454 and 525-526) do align.

```suggestion
			   PPE_MRU_MTU_CTRL(port,
					    priv->data->mru_mtu_ctrl_stride) + 4,
```

---

target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
```diff
@@ -1098,6 +1098,8 @@ static int edma_irq_init(struct edma_priv *priv)
 static int edma_page_pool_create(struct edma_priv *priv)
 {
 	struct page_pool_params pp = {
+		.order     = get_order(NET_SKB_PAD + EDMA_RX_BUFFER_SIZE +
+				       SKB_DATA_ALIGN(sizeof(struct skb_shared_info))),
```
openwrt-ai: The order-2 buffers are allocated unconditionally, so the cost is paid on every qualcommax board even when nothing ever raises the MTU: the 2048-entry RX ring goes from 8 MiB to 32 MiB of permanently pinned memory, which is a large fraction of RAM on the smaller ipq60xx boards. The commit message states the number but not why it has to be unconditional.

A second, less visible effect: `napi_build_skb(page_address(page), page_size(page))` at line 375 now reports a ~16 KiB `truesize` for *every* received frame, including 64-byte ones. That feeds directly into `sk_rmem_alloc`/TCP window accounting and collapses the effective receive window for ordinary 1500-byte traffic, so this is not only a memory-footprint change.

Could the page-pool order be derived from the current conduit MTU and the pool re-created when the MTU actually grows, rather than sizing for 12288 up front?

---

openwrt-ai (commented):
## Commit checks

- `2235d492` "qualcommax: qca_ppe: use SoC-specific MRU/MTU stride" — the `Fixes:` tag references `142104bb90b7`, which does not exist. The commit it refers to is `142104bb90b3d7685b2a96f04f29b76cbd1e5ee9` ("qualcommax: add PPE driver"), and the last commit of this series (`016cee17`) already spells it `142104bb90b3` correctly.


---

target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
```diff
@@ -1095,20 +1117,95 @@ static int edma_irq_init(struct edma_priv *priv)
...
+}
+
+static u32 edma_rx_buffer_size(u8 order)
+{
+	return (PAGE_SIZE << order) - NET_SKB_PAD -
+	       SKB_DATA_ALIGN(sizeof(struct skb_shared_info));
+}
+
+static struct page_pool *edma_page_pool_create(struct edma_priv *priv,
+						       u8 order)
```
openwrt-ai: nit: the continuation line is indented 8 columns too far — it no longer lines up with `struct edma_priv *priv` under the opening paren (column 47). This looks like leftover alignment from the old, longer prototype.

```suggestion
					       u8 order)
```

---

target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
```diff
@@ -1095,20 +1117,95 @@ static int edma_irq_init(struct edma_priv *priv)
...
+		priv->rx_page_order = old_order;
+		priv->rx_buffer_size = edma_rx_buffer_size(old_order);
+		page_pool_destroy(new_pool);
+
+		restore_ret = edma_hw_init(priv);
+		if (restore_ret) {
+			netdev_err(netdev,
+				   "failed to restore receive buffers after MTU change\n");
+			return restore_ret;
+		}
```
openwrt-ai, line 1206: This early return leaves the NAPI/queue state unbalanced. When `running` was true, the function already did `netif_tx_disable()` + `edma_ndo_stop()` (which calls `napi_disable()` on both instances), and returning here skips the `if (running) edma_ndo_open(netdev)` at the bottom. The netdev is still `IFF_UP` from the core's point of view, so the next `ip link set ... down` calls `edma_ndo_stop()` → `napi_disable()` on an already-disabled NAPI, which spins forever in [`napi_disable_locked()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/net/core/dev.c#L7305) waiting for `NAPIF_STATE_SCHED` to clear.

Simply falling through to `edma_ndo_open()` is not obviously right either — after a failed `edma_hw_init()` the rings are freed (`ring->desc == NULL`) and re-enabling NAPI would poll against them. Some explicit handling (e.g. `netif_device_detach()` before returning, or keeping the NAPIs enabled but the rings disabled) is probably needed so the driver survives the subsequent `ndo_stop`.

---

target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
```diff
@@ -1095,20 +1117,95 @@ static int edma_irq_init(struct edma_priv *priv)
 	return 0;
 }
 
-static int edma_page_pool_create(struct edma_priv *priv)
+static u8 edma_rx_page_order(int mtu)
+{
+	size_t size = NET_SKB_PAD + EDMA_RX_PREHDR_SIZE + mtu + ETH_HLEN +
+		      2 * VLAN_HLEN +
+		      SKB_DATA_ALIGN(sizeof(struct skb_shared_info));
```
openwrt-ai, line 1125: `EDMA_MAX_MTU` in the same series budgets for `ETH_FCS_LEN` (`EDMA_MAX_FRAME_SIZE - ETH_HLEN - ETH_FCS_LEN - 2 * VLAN_HLEN`), but this order calculation does not. Since `edma_rx_buffer_size()` is the exact inverse of this expression, the advertised buffer is exactly `EDMA_RX_PREHDR_SIZE + mtu + ETH_HLEN + 2 * VLAN_HLEN` with no slack whenever `size` lands on a power-of-two boundary — e.g. `mtu = 3658` gives `size == 4096`, order 0, `rx_buffer_size == 3712`, while a double-tagged frame plus FCS needs 3716. Is the FCS stripped before EDMA writes into the buffer? If it can be present, the two expressions should agree.

---

target/linux/qualcommax/files/drivers/net/ethernet/qualcomm/qca_edma.c
```diff
@@ -358,6 +418,9 @@ static u32 edma_clean_rx(struct edma_priv *priv, int budget,
 
 		page_pool_dma_sync_for_cpu(priv->page_pool, page, 0,
 					   EDMA_RX_PREHDR_SIZE + pkt_len);
+		store_idx = le32_to_cpu(rxph->opaque);
+		if (unlikely(!edma_rx_page_take(priv, page, store_idx)))
+			goto next;
```
openwrt-ai, line 424: Every other bail-out in this loop (`SRC_INFO_TYPE` mismatch, `napi_build_skb()` failure) returns the page with `page_pool_put_full_page()`; this one drops it. The page stays counted as in-flight, so a later `page_pool_destroy()` — from `edma_ndo_change_mtu()` or `edma_remove()` — never reaches zero and the pool is only released by the retry workqueue with a "stalled pool shutdown" warning.

Is that deliberate, i.e. is the untracked case assumed to be a page already recycled (where a put would be a double-free)? If so, a comment saying so would help; otherwise the page should be put back here.

---

openwrt-ai (commented):
Reviewed the 4 rebased commits (previous review was against `016cee17`). The `Fixes:` SHA flagged last time is now correct (`142104bb90b3` exists in-tree). CI on `60a0e9b0` is green.

## Commit checks

- `b3046d90` "qualcommax: qca_ppe: add jumbo frame support" — the message only describes deriving the RX page order from the conduit MTU and recreating the pool/rings on an order change. The diff additionally rewrites the whole RX buffer lifecycle: it deletes `edma_rxfill_drain()` and `edma_rxdesc_drain()`, adds a `page_store` array to `struct edma_ring`, stamps an opaque cookie into the RX preheader in `edma_rx_fill()`, and adds `edma_rx_page_take()` plus `edma_rx_ring_alloc()`/`edma_rx_ring_free()`. That is the larger half of the diff and is a behavioural change on its own (drain no longer trusts the hardware indices). Worth either a paragraph in the message explaining why the hardware-index drain had to go, or splitting it into a preparatory patch.

---

robimarko:

lorand-horvath:
@robimarko Is this applicable to ipq50xx as well?

robimarko:
No, ipq50xx uses a completely different driver, it might already support it

PR #24247, opened 2026-07-16
tools: gnulib: update to branch stable-202507

mcprat:
Move to the July 2026 update of the last 2025 stable branch.

This branch includes many new modules and changes to modules
that were previously patched, so many patches are dropped.

Newer gnulib now also has a proper fts.h header
in the form of a template file for variable definitions
that are set during configuration,
so the patch 120-unmangle-darwin-fts-h.patch
in order to resolve differences between macOS and standard
fts.h headers is rewritten to patch the new header include guard.

Manually Adjusted Patch:
 - 120-unmangle-darwin-fts-h.patch

All other patches are automatically refreshed.

@robimarko @httpstorm for testing

hopefully this is the first update to gnulib that doesn't require another change for Macs

PR #24759, opened 2026-08-16
tools: gnulib: remove stale gl_-prefixed macros from staging

BKPepe:
Fixes macOS host builds broken by the gnulib stable-202507 update (#24247).

c820f097e0 temporarily installed gnulib autoconf macros as `gl_*.m4`. Although this was reverted in cda598a595, the previously installed files were never removed from `share/aclocal`.

On affected build trees, aclocal picks up both the stale `gl_*.m4` files and the current macros. The stale definitions can then win and cause elfutils' regenerated `ctype.h` to contain empty `#if` expressions on macOS.

This patch removes the stale `gl_*.m4` files from Host/Uninstall. Since Host/Install calls it first, existing build trees are automatically cleaned up on the next gnulib reinstall.

Verified on macOS (Apple Silicon): the stale files reproduce the elfutils build failure, while the cleanup allows a clean gnulib + elfutils rebuild to succeed.

### DIFF
--- a/tools/gnulib/Makefile
+++ b/tools/gnulib/Makefile
@@ -23,6 +23,7 @@ endef
 
 define Host/Uninstall
 	rm -rf $(STAGING_DIR_HOST)/bin/gnulib-tool $(STAGING_DIR_HOST)/share/gnulib
+	rm -f $(STAGING_DIR_HOST)/share/aclocal/gl_*.m4
 endef
 
 $(eval $(call HostBuild))


tools/gnulib/Makefile
```diff
@@ -23,6 +23,7 @@ endef
 
 define Host/Uninstall
 	rm -rf $(STAGING_DIR_HOST)/bin/gnulib-tool $(STAGING_DIR_HOST)/share/gnulib
+	rm -f $(STAGING_DIR_HOST)/share/aclocal/gl_*.m4
```
openwrt-ai, line 26: `gl_*.m4` is not the only class of leftover in `share/aclocal` that current trees carry. [`tools/gnulib/patches/310-modules-cond.patch`](https://github.com/openwrt/openwrt/blob/6c3a15e76727f094c08047cb8475947ecf456532/tools/gnulib/patches/310-modules-cond.patch#L1-L3) (added by a54eb7297f8) deletes `m4/cond.m4` and re-adds it as `m4/glthread_cond.m4`, so `cond.m4` is no longer installed — but any tree that installed gnulib *before* c820f097e0 still has `$(STAGING_DIR_HOST)/share/aclocal/cond.m4`, because `Host/Uninstall` has never removed it either.

That stale `cond.m4` re-creates precisely the automake `AM_CONDITIONAL` clash that the a54eb7297f8 rename was meant to fix, on the same set of pre-existing build trees this PR is trying to repair.

```suggestion
	rm -f $(STAGING_DIR_HOST)/share/aclocal/gl_*.m4
	rm -f $(STAGING_DIR_HOST)/share/aclocal/cond.m4
```

(Neither `tools/missing-macros` nor `tools/automake` installs `cond.m4` into `share/aclocal` — automake's copy lives in `share/aclocal-<ver>` — so removing it unconditionally is safe.)

More generally: `Host/Uninstall` still removes none of the unprefixed `*.m4` files that `Host/Install` writes into `share/aclocal` on line 19. Compare [`tools/libtool/Makefile:61`](https://github.com/openwrt/openwrt/blob/6c3a15e76727f094c08047cb8475947ecf456532/tools/libtool/Makefile#L61), which enumerates and deletes exactly the macros it installed. Without that, the next gnulib bump that renames or drops a macro file reproduces this same bug class and will need yet another one-off `rm` line here.

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message accurately describes the change, and the `Host/Install` -> `$(call Host/Uninstall)` ordering claim checks out. One inline note about a leftover the glob does not cover.

---

JuliusBairaktaris:
https://github.com/openwrt/openwrt/pull/24576/changes

robimarko:

PR #24761, opened 2026-08-16
ramips: add support for ASUS RT-AC52U B1

andy9a9:
RT-AC52U B1 Dual-Band Wi-Fi Router with two antennas and USB port

Specification:
- CPU: MediaTek MT7620A @ 580 MHz
- RAM: 64MB
- Flash: GigaDevice GD25Q128CSIG (16MB)
- WLAN: MT7620A(n) + MT7610E(ac) 2R/2T,
  - 802.11ac/n AC750 433Mbps (ac) + 300Mbps (n)
- Switch: Realtek RTL8367RB
- 4x 10/100/1000 Mbps Ethernet
- 1x 10/100/1000 Mbps WAN
- USB: USB 2.0 Controller
- 6x LED, 2x button

Installation (TFTP):
1. Power off the router.
2. Press and hold the Reset button.
3. Power on the router.
4. Wait 5 seconds, then release the Reset button.
5. Connect your PC to the router with Ethernet cable.
6. Configure a static IP address to `192.168.1.75/24`
7. Configure your TFTP client to use binary (octet) transfer mode.
8. Upload the firmware image (OpenWrt, stock, or Padavan) to `192.168.1.1`. The TFTP server accepts any filename.
9. Wait for the firmware flashing process to complete.

Known issues:
- Unable to control all LEDs while some of them are not wired to GPIO
  - I wasn't able to find the right one even if I loop over all available

### DIFF
--- a/target/linux/ramips/dts/mt7620a_asus_rt-ac52u-b1.dts
+++ b/target/linux/ramips/dts/mt7620a_asus_rt-ac52u-b1.dts
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+#include "mt7620a_asus_rt-ac5x.dtsi"
+
+/ {
+	compatible = "asus,rt-ac52u-b1", "ralink,mt7620a-soc";
+	model = "Asus RT-AC52U B1";
+
+	chosen {
+		bootargs = "console=ttyS0,115200";
+	};
+
+	rtl8367rb {
+		compatible = "realtek,rtl8367b";
+		realtek,extif = <6 1 0 1 1 1 1 1 1 2>;
+		mii-bus = <&mdio0>;
+	};
+};
+
+&state_default {
+	gpio {
+		groups = "i2c", "uartf", "wled", "ephy", "spi refclk";
+		function = "gpio";
+	};
+};
+
+&ethernet {
+	pinctrl-names = "default";
+	pinctrl-0 = <&rgmii1_pins &mdio_pins>;
+
+	port@5 {
+		status = "okay";
+		mediatek,fixed-link = <1000 1 1 1>;
+		phy-mode = "rgmii";
+	};
+
+	mdio0: mdio-bus {
+		status = "okay";
+	};
+};
+
+&gsw {
+	mediatek,ephy-base = /bits/ 8 <8>;
+};
+
+&pcie0 {
+	wifi@0,0 {
+		compatible = "mediatek,mt76";
+		reg = <0x0000 0 0 0 0>;
+		nvmem-cells = <&eeprom_factory_8000>;
+		nvmem-cell-names = "eeprom";
+	};
+};
--- a/target/linux/ramips/image/mt7620.mk
+++ b/target/linux/ramips/image/mt7620.mk
@@ -109,6 +109,17 @@ define Device/asus_rt-ac51u
 endef
 TARGET_DEVICES += asus_rt-ac51u
 
+define Device/asus_rt-ac52u-b1
+  SOC := mt7620a
+  IMAGE_SIZE := 16064k
+  DEVICE_VENDOR := ASUS
+  DEVICE_MODEL := RT-AC52U
+  DEVICE_VARIANT := B1
+  DEVICE_PACKAGES := kmod-mt76x0e kmod-usb2 kmod-usb-ohci \
+	kmod-usb-ledtrig-usbport kmod-switch-rtl8367b
+endef
+TARGET_DEVICES += asus_rt-ac52u-b1
+
 define Device/asus_rt-ac54u
   SOC := mt7620a
   IMAGE_SIZE := 16064k
--- a/target/linux/ramips/mt7620/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/mt7620/base-files/etc/board.d/02_network
@@ -153,6 +153,7 @@ ramips_setup_interfaces()
 		ucidef_add_switch "switch0" \
 			"1:lan:3" "2:lan:2" "3:lan:1" "4:wan" "6@eth0"
 		;;
+	asus,rt-ac52u-b1|\
 	edimax,br-6478ac-v2|\
 	lb-link,bl-w1200|\
 	tplink,archer-c2-v1)


target/linux/ramips/dts/mt7620a_asus_rt-ac52u-b1.dts
```diff
@@ -0,0 +1,45 @@
+#include "mt7620a_asus_rt-ac5x.dtsi"
```
openwrt-ai, line 3: nit: new DTS file is missing the SPDX license header. Nine of the ten most recently added `target/linux/ramips/dts/*.dts` files start with it, and the sibling [`mt7620a_asus_rt-ac54u.dts`](https://github.com/openwrt/openwrt/blob/c06daedd3e613684acf247c370ce126829b22229/target/linux/ramips/dts/mt7620a_asus_rt-ac54u.dts#L1) carries it too.

```suggestion
// SPDX-License-Identifier: GPL-2.0-or-later OR MIT

#include "mt7620a_asus_rt-ac5x.dtsi"
```

---

target/linux/ramips/dts/mt7620a_asus_rt-ac52u-b1.dts
```diff
@@ -0,0 +1,45 @@
...
+
+	mdio0: mdio-bus {
+		status = "okay";
+	};
+};
+
+&gsw {
+	mediatek,port4-gmac;
+	mediatek,ephy-base = /bits/ 8 <8>;
+};
```
openwrt-ai, line 44: `mediatek,port4-gmac` looks like it may not belong on this board. It clears `port4_ephy`, which switches GSW port 4 from the internal EPHY to an external GMAC and drops ephy init from 4 PHYs to 3 — see [`gsw_mt7620.c:43`](``https://github.com/openwrt/openwrt/blob/c06daedd3e613684acf247c370ce126829b22229/target/linux/ramips/files/drivers/net/ethernet/ralink/gsw_mt7620.c#L43``) and [`gsw_mt7620.c:245`](``https://github.com/openwrt/openwrt/blob/c06daedd3e613684acf247c370ce126829b22229/target/linux/ramips/files/drivers/net/ethernet/ralink/gsw_mt7620.c#L245).`` But here the RTL8367RB is wired to port 5, and `pinctrl-0` on line 20 only muxes `rgmii1_pins` — RGMII2, which port 4 would need as a GMAC, is never muxed.

Conflicting evidence in tree: [`mt7620a_tplink_archer-c2-v1.dts:70-83`](https://github.com/openwrt/openwrt/blob/c06daedd3e613684acf247c370ce126829b22229/target/linux/ramips/dts/mt7620a_tplink_archer-c2-v1.dts#L70-L83) uses the identical `rtl8367rb` node (same `realtek,extif = <6 1 0 1 1 1 1 1 1 2>`, same `mdio0` label, same `port@5` fixed-link) and the same 02_network switch case as this board, and has no `&gsw` override at all. Conversely [`mt7620a_zyxel_keenetic-viva.dts:160-180`](https://github.com/openwrt/openwrt/blob/c06daedd3e613684acf247c370ce126829b22229/target/linux/ramips/dts/mt7620a_zyxel_keenetic-viva.dts#L160-L180) does set `mediatek,port4-gmac` + `ephy-base = <8>`, but it enables `port@4` rather than `port@5`.

The `ephy-base = <8>` remap makes sense on its own (it keeps the internal EPHYs off MDIO addresses 0–4, which the RTL8367RB occupies). Is `mediatek,port4-gmac` intentional here, or copied over from the keenetic-viva topology?

---

target/linux/ramips/image/mt7620.mk
```diff
@@ -109,6 +109,18 @@ define Device/asus_rt-ac51u
...
 TARGET_DEVICES += asus_rt-ac51u
 
+define Device/asus_rt-ac52u-b1
+  SOC := mt7620a
+  IMAGE_SIZE := 16064k
+  DEVICE_VENDOR := ASUS
+  DEVICE_MODEL := RT-AC52U
+  DEVICE_VARIANT := B1
+  DEVICE_PACKAGES := kmod-mt76x0e kmod-usb2 kmod-usb-ohci \
+	kmod-usb-ledtrig-usbport kmod-switch-rtl8366-smi kmod-switch-rtl8367b
```
openwrt-ai: `kmod-switch-rtl8366-smi` is redundant here: `kmod-switch-rtl8367b` already pulls it in via `DEPENDS:=+kmod-switch-rtl8366-smi` — see [`netdevices.mk:1153-1156`](https://github.com/openwrt/openwrt/blob/c06daedd3e613684acf247c370ce126829b22229/package/kernel/linux/modules/netdevices.mk#L1153-L1156). None of the five existing ramips devices that use the RTL8367B list it explicitly (`mt7620.mk` lines 687, 1248, 1265, 1313, 1653).

```suggestion
  DEVICE_PACKAGES := kmod-mt76x0e kmod-usb2 kmod-usb-ohci \
	kmod-usb-ledtrig-usbport kmod-switch-rtl8367b
```

---

target/linux/ramips/image/mt7620.mk
```diff
@@ -109,6 +109,18 @@ define Device/asus_rt-ac51u
...
+  SUPPORTED_DEVICES += rt-ac52u-b1
```
openwrt-ai: nit: `SUPPORTED_DEVICES` exists to accept sysupgrade images whose metadata carries a legacy pre-DTS board name — that is why `asus_rt-ac51u` above has `SUPPORTED_DEVICES += rt-ac51u`. This device has never been supported by OpenWrt (`grep -ri rt-ac52u` over the tree returns nothing before this PR), so no image can ever carry `rt-ac52u-b1` as a board name and the line is dead weight. Unless it is meant to match some third-party (Padavan) metadata, please drop it.

---

openwrt-ai (commented):
## Commit checks

- c06daedd "ramips: add support for ASUS RT-AC52U B1" — this is new-device support (new `target/linux/ramips/dts/mt7620a_asus_rt-ac52u-b1.dts` + new `define Device/asus_rt-ac52u-b1` in `target/linux/ramips/image/mt7620.mk`), so per `.github/llm-review-rules.md` the commit message needs three sections. The **MAC address layout** section is missing entirely — please add one line per interface (LAN / WAN / 2.4G / 5G) saying where each MAC comes from, using `xx:xx:xx` for the unit-specific bytes. This matters here because the diff mixes two sources: the DTS inherits `nvmem-cells = <&macaddr_factory_28>` on `&ethernet` from `mt7620a_asus_rt-ac5x.dtsi`, while `02_network` overrides `lan_mac` from `factory 0x22`, and nothing documents where WAN/WLAN MACs come from.
- c06daedd — the **Hardware specification** section is missing the UART entry (header location, pinout, baud rate) and Power (voltage + current). The DTS sets `bootargs = "console=ttyS0,115200"`, overriding the `mt7620a.dtsi` default of 57600, so the console baud is known and worth stating.
- c06daedd — `- CPU: MediaTek MT7620A @ 580 MHz, WiSoC 5GHz` contradicts the WLAN line two rows below (`MT7620A(n) + MT7610E(ac)`): the MT7620A built-in radio is 2.4 GHz 802.11n, the 5 GHz AC radio is the separate MT7610E on PCIe. Suggest dropping the `WiSoC 5GHz` fragment.


---

target/linux/ramips/image/mt7620.mk
andy9a9: `SUPPORTED_DEVICES` removed

target/linux/ramips/image/mt7620.mk
andy9a9: `kmod-switch-rtl8366-smi` removed

target/linux/ramips/dts/mt7620a_asus_rt-ac52u-b1.dts
andy9a9: `mediatek,port4-gmac` removed

target/linux/ramips/dts/mt7620a_asus_rt-ac52u-b1.dts
andy9a9: `SPDX` section added

target/linux/ramips/dts/mt7620a_asus_rt-ac52u-b1.dts
```diff
@@ -0,0 +1,46 @@
...
+	};
+
+	mdio0: mdio-bus {
+		status = "okay";
+	};
+};
+
+&gsw {
+	mediatek,ephy-base = /bits/ 8 <8>;
+};
```
openwrt-ai, line 44: Possible lead on the "Unable to control all LEDs" known issue, now that `mediatek,port4-gmac` is gone and all five ports come off the RTL8367RB on port 5: the internal EPHY pins are still muxed as `ephy`. The kernel's mt7620 pinctrl defines that group as [`FUNC("ephy", 0, 40, 5)` at pinctrl-mt7620.c:64](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/pinctrl/mediatek/pinctrl-mt7620.c#L64) — 5 pins at GPIO 40–44 — and the inherited [`&state_default` in mt7620a_asus_rt-ac5x.dtsi:138-143](https://github.com/openwrt/openwrt/blob/b052a5c8fb517190eb268a3027294a6d065984f8/target/linux/ramips/dts/mt7620a_asus_rt-ac5x.dtsi#L138-L143) only frees `i2c`, `wled` and `uartf`, so GPIO 40–44 stay unavailable on this board. The identically wired [`mt7620a_tplink_archer-c2-v1.dts:63-68`](https://github.com/openwrt/openwrt/blob/b052a5c8fb517190eb268a3027294a6d065984f8/target/linux/ramips/dts/mt7620a_tplink_archer-c2-v1.dts#L63-L68) (same `rtl8367rb` node, same `realtek,extif`, same 02_network case) adds `"ephy"` and `"spi refclk"` to the gpio group for exactly this reason.

To probe those pins this board would need a `&state_default` override that also lists `"ephy"`, plus `&gpio2 { status = "okay"; };` — [`gpio2` is `status = "disabled"` in mt7620a.dtsi:164-182](https://github.com/openwrt/openwrt/blob/b052a5c8fb517190eb268a3027294a6d065984f8/target/linux/ramips/dts/mt7620a.dtsi#L164-L182) and it is the bank covering GPIO 40+. Worth re-checking before the "not wired to GPIO" note is finalised.

---

openwrt-ai (commented):
Re-reviewed at `b052a5c8` (amended since `c06daedd`). The three code changes in the new revision — SPDX header, `mediatek,port4-gmac` removal, redundant `kmod-switch-rtl8366-smi` / dead `SUPPORTED_DEVICES` removal — all look correct, and `mediatek,ephy-base = <8>` on its own is a valid combination (same as [`mt7620a_sitecom_wlr-4100-v1-002.dts:180`](https://github.com/openwrt/openwrt/blob/b052a5c8fb517190eb268a3027294a6d065984f8/target/linux/ramips/dts/mt7620a_sitecom_wlr-4100-v1-002.dts#L180)). No new code issues; one inline observation about the LED known-issue.

## Commit checks

- b052a5c8 "ramips: add support for ASUS RT-AC52U B1" — the **MAC address layout** section is still missing from the commit message. Per `.github/llm-review-rules.md` a new-device commit (new DTS + new `define Device/...`) needs one line per interface (LAN / WAN / 2.4G / 5G) saying where each MAC comes from, with `xx:xx:xx` for the unit-specific bytes. It matters here because the diff uses two different sources: `&ethernet` inherits `nvmem-cells = <&macaddr_factory_28>` from `mt7620a_asus_rt-ac5x.dtsi`, while the `02_network` hunk sets `lan_mac` from `factory 0x22`, and nothing states where the WAN and WLAN MACs come from.
- b052a5c8 — the **Hardware specification** section is still missing UART (header location, pinout, baud rate) and Power (voltage + current). The DTS pins `bootargs = "console=ttyS0,115200"`, overriding the `mt7620a.dtsi` default of 57600, so the console baud is known and worth writing down — it is also the only board in the `rt-ac5x` family that overrides it.

The `WiSoC 5GHz` contradiction from the previous round is fixed, thanks.

---

target/linux/ramips/dts/mt7620a_asus_rt-ac52u-b1.dts
andy9a9: I tried multiple experiments, but didn't find a way to make the other leds working

target/linux/ramips/dts/mt7620a_asus_rt-ac52u-b1.dts
```diff
@@ -0,0 +1,53 @@
...
+		mii-bus = <&mdio0>;
+	};
+};
+
+&state_default {
+	gpio {
+		groups = "i2c", "uartf", "wled", "ephy", "spi refclk";
+		function = "gpio";
+	};
+};
```
openwrt-ai, line 25: This override is currently inert, which is probably why the extra LEDs still don't respond. Freeing the pins is only half the job — the GPIO banks that contain them are still disabled:

- `"spi refclk"` is [`FUNC("spi refclk", 0, 37, 3)`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/pinctrl/mediatek/pinctrl-mt7620.c#L63) → GPIO 37–39, which live in `gpio1` (`ngpios = <16>`, base 24), and [`gpio1` is `status = "disabled"` in mt7620a.dtsi:144-161](https://github.com/openwrt/openwrt/blob/6239e9f4ff88736deecdee1a50770c1284ca64b7/target/linux/ramips/dts/mt7620a.dtsi#L144-L161).
- `"ephy"` is [`FUNC("ephy", 0, 40, 5)`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/pinctrl/mediatek/pinctrl-mt7620.c#L64) → GPIO 40–44, in `gpio2`, [also `status = "disabled"` in mt7620a.dtsi:164-181](https://github.com/openwrt/openwrt/blob/6239e9f4ff88736deecdee1a50770c1284ca64b7/target/linux/ramips/dts/mt7620a.dtsi#L164-L181).

`ralink_gpio_probe()` is a normal platform probe, so a disabled node never registers a `gpio_chip` and nothing can request lines 37–44. Every in-tree mt7620a board that frees these groups enables the banks explicitly — e.g. [`mt7620a_engenius_epg600.dts:84-90`](https://github.com/openwrt/openwrt/blob/6239e9f4ff88736deecdee1a50770c1284ca64b7/target/linux/ramips/dts/mt7620a_engenius_epg600.dts#L84-L90) (same `"ephy"` + `"spi refclk"` groups) and [`mt7620a_xiaomi_miwifi-mini.dts:72-78`](https://github.com/openwrt/openwrt/blob/6239e9f4ff88736deecdee1a50770c1284ca64b7/target/linux/ramips/dts/mt7620a_xiaomi_miwifi-mini.dts#L72-L78).

Worth retrying the LED hunt with the banks on (`&gpio1 15` = GPIO 39, `&gpio2 0` = GPIO 40, … — the same lines [`mt7620a_tplink_archer-c2-v1.dts:39-45`](https://github.com/openwrt/openwrt/blob/6239e9f4ff88736deecdee1a50770c1284ca64b7/target/linux/ramips/dts/mt7620a_tplink_archer-c2-v1.dts#L39-L45) uses on the identically wired board). If the remaining LEDs still don't show up after that, the pin mux override no longer buys anything and should be dropped again rather than left in.

```suggestion
&gpio1 {
	status = "okay";
};

&gpio2 {
	status = "okay";
};

&state_default {
	gpio {
		groups = "i2c", "uartf", "wled", "ephy", "spi refclk";
		function = "gpio";
	};
};
```

---

target/linux/ramips/mt7620/base-files/etc/board.d/02_network
```diff
@@ -153,6 +153,7 @@ ramips_setup_interfaces()
 		ucidef_add_switch "switch0" \
 			"1:lan:3" "2:lan:2" "3:lan:1" "4:wan" "6@eth0"
 		;;
+	asus,rt-ac52u-b1|\
```
openwrt-ai, line 156: nit: the previous revision also had this board in the `ramips_setup_macs` case at [02_network:344-347](``https://github.com/openwrt/openwrt/blob/6239e9f4ff88736deecdee1a50770c1284ca64b7/target/linux/ramips/mt7620/base-files/etc/board.d/02_network#L344-L347``) (`lan_mac=$(mtd_get_mac_binary factory 0x22)`), and that entry is gone now. With no `lan_mac`/`wan_mac` override, both `lan` and `wan` inherit eth0's address, i.e. `macaddr_factory_28` from [`mt7620a_asus_rt-ac5x.dtsi:146-147`](https://github.com/openwrt/openwrt/blob/6239e9f4ff88736deecdee1a50770c1284ca64b7/target/linux/ramips/dts/mt7620a_asus_rt-ac5x.dtsi#L146-L147) — the two interfaces end up with the identical MAC. Both siblings sharing that dtsi (`asus,rt-ac51u`, `asus,rt-ac54u`) keep the `factory 0x22` LAN MAC.

Was `factory 0x22` empty/invalid on your unit, or is the shared LAN/WAN MAC deliberate? A one-line note in the commit message would save the next reader the same question.

---

openwrt-ai (commented):
Re-reviewed at `6239e9f4` (amended since `b052a5c8`). Two changes in this revision: the new `&state_default` override freeing `"ephy"` / `"spi refclk"`, and the removal of this board from the `ramips_setup_macs` case in `02_network`. The commit message gained the UART and MAC-address sections. One inline finding on the pin mux (it can't take effect as written) and one question on the dropped LAN MAC.

## Commit checks

- 6239e9f4 "ramips: add support for ASUS RT-AC52U B1" — the new **MAC addresses** section is wrong for 2.4 GHz. It says `Wifi 2.4G, 5G = 'option eeprom' from factory_8000 partition`, but only the MT7610E on PCIe uses `factory` 0x8000 (`nvmem-cells = <&eeprom_factory_8000>` in the new DTS); the built-in MT7620A radio uses `eeprom_factory_0`, i.e. `factory` offset 0x0, via [`&wmac` in mt7620a_asus_rt-ac5x.dtsi:152-155](https://github.com/openwrt/openwrt/blob/6239e9f4ff88736deecdee1a50770c1284ca64b7/target/linux/ramips/dts/mt7620a_asus_rt-ac5x.dtsi#L152-L155). Please split that into two lines (2.4G → factory 0x0, 5G → factory 0x8000). Same section: `factory_28 partition` reads as a partition name — the partition is `factory` and 0x28 is the offset within it.
- 6239e9f4 — **Hardware specification** still has no Power line (input voltage + current); UART is now covered, thanks.

---

target/linux/ramips/mt7620/base-files/etc/board.d/02_network
andy9a9: I would say, the current setup is ok. Nobody will take a look on the previous commits, just the final one, so I guess don't have to explain it in commits.

target/linux/ramips/dts/mt7620a_asus_rt-ac52u-b1.dts
andy9a9: It didn't help. As I said, I tried to loop over all GPIOs with `gpiod-utils` => direct LED controlling -> no change

PR #24763, opened 2026-08-16
qualcommax: ipq50xx: correct default ECC settings

hzyitc:
Instand of using the ECC provided by the flash chip, QPIC handles ECC by itself. So the ECC capabilities of the flash chip is useless.

The max ECC capabilities of QPIC depends on the page size (include OOB). For ECC4, it needs 64B OOB space. For ECC8, it needs 80B OOB space. Use the max capabilities as the default value.


target/linux/qualcommax/patches-6.18/0421-spi-spi-qpic-snand-reject-when-no-enough-OOB-space.patch
```diff
@@ -0,0 +1,39 @@
...
+ 	bad_block_byte = mtd->writesize - ecc_cfg->cw_size * (cwperpage - 1) + 1;
+ 
++	if ((ecc_cfg->cw_size * cwperpage) > (mtd->writesize + mtd->oobsize)) {
++		dev_err(snandc->dev,
++			"Current ECC settings require %d bytes, but the flash only has %d+%d bytes.\n",
++			(ecc_cfg->cw_size * cwperpage), mtd->writesize,
++			mtd->oobsize);
++		ret = -EINVAL;
++		goto err_free_ecc_cfg;
++	}
```
openwrt-ai, line 30: The new `goto err_free_ecc_cfg` fires at a point where `snandc->qspi->oob_buf` has already been allocated ([spi-qpic-snand.c:305-312](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/spi/spi-qpic-snand.c#L305-L312)) and `nand->ecc.ctx.priv = ecc_cfg` has already been set ([:314](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/spi/spi-qpic-snand.c#L314)), but [`err_free_ecc_cfg`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/spi/spi-qpic-snand.c#L396-L398) only does `kfree(ecc_cfg)`. So every rejected probe leaks the `writesize + oobsize` `oob_buf` allocation and leaves `nand->ecc.ctx.priv` pointing at freed memory.

Everything the check needs (`ecc_bytes_hw`, `spare_bytes`, and the constants `bbm_size = 1` / `cw_data = 516`) is already determined right after the `switch (ecc_cfg->strength)`, so placing the check there — before the `oob_buf` allocation — makes the existing error label correct and avoids the leak.

---

target/linux/qualcommax/patches-6.18/0421-spi-spi-qpic-snand-reject-when-no-enough-OOB-space.patch
```diff
@@ -0,0 +1,39 @@
...
++
+ 	mtd_set_ooblayout(mtd, &qcom_spi_ooblayout);
+ 
+ 	/*
+@@ -1664,4 +1673,3 @@ module_platform_driver(qcom_spi_driver);
+ MODULE_DESCRIPTION("SPI driver for QPIC QSPI cores");
+ MODULE_AUTHOR("Md Sadre Alam <quic_mdalam@quicinc.com>");
+ MODULE_LICENSE("GPL");
+-
```
openwrt-ai: This second hunk is unrelated to the OOB check the patch is about: it deletes the blank line at the end of `drivers/spi/spi-qpic-snand.c`. Upstream v6.18.44 genuinely ends with `MODULE_LICENSE("GPL");` followed by one empty line ([spi-qpic-snand.c:1665-1667](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/spi/spi-qpic-snand.c#L1665-L1667)), so this looks like an editor/tooling artifact rather than an intended change. It also makes the header stats (`10 +++++++++-`, `9 insertions(+), 1 deletion(-)`) account for a deletion that is not part of the fix.

Please drop this hunk and regenerate the patch so the hunk headers and diffstat match — for `target/linux/qualcommax/patches-6.18/` that is `make target/linux/refresh` with the qualcommax target selected.

---

target/linux/qualcommax/patches-6.18/0421-spi-spi-qpic-snand-reject-when-no-enough-OOB-space.patch
```diff
@@ -0,0 +1,39 @@
+From 4f1aacff076719bd2d33bfd5f63e85e90957d30d Mon Sep 17 00:00:00 2001
+From: Ziyang Huang <hzyitc@outlook.com>
+Date: Sun, 16 Aug 2026 17:33:57 +0800
+Subject: [PATCH] spi: spi-qpic-snand: reject when no enough OOB space
+
+Print a clear error when the flash doesn't have enough space, instand of
+return -EBADMSG silently in read or write functions.
```
openwrt-ai, line 7: nit: two typos here — `no enough OOB space` in the subject should be `not enough OOB space`, and `instand of` on line 6 should be `instead of`. The same `Instand of` typo appears in `0422-spi-spi-qpic-snand-pick-default-ECC-settings-based-o.patch` line 7 and in the OpenWrt commit message body of 4a48b0a.

---

target/linux/qualcommax/patches-6.18/0422-spi-spi-qpic-snand-pick-default-ECC-settings-based-o.patch
```diff
@@ -0,0 +1,49 @@
...
+-	} else if (reqs->step_size && reqs->strength) {
+-		ecc_cfg->step_size = reqs->step_size;
+-		ecc_cfg->strength = reqs->strength;
+ 	} else {
+ 		/* use defaults */
+ 		ecc_cfg->step_size = NANDC_STEP_SIZE;
+-		ecc_cfg->strength = 4;
++		if (mtd->oobsize < 64) {
++			ret = -EOPNOTSUPP;
++			goto err_free_ecc_cfg;
```
openwrt-ai, line 42: This bails out with `-EOPNOTSUPP` and prints nothing, so a chip with too little OOB fails to probe with no clue why — exactly the silent-failure mode that `0421-spi-spi-qpic-snand-reject-when-no-enough-OOB-space.patch` in this same commit sets out to remove. Add a `dev_err(snandc->dev, ...)` naming `mtd->oobsize` before the `goto`, matching the message style of the other error paths in this function.

nit: the braces are also inconsistent — this branch is braced while the `else if`/`else` below are not. Linux coding style wants braces on every branch of the chain once one branch needs them. After editing, regenerate with `make target/linux/refresh` so the hunk header and diffstat stay in sync.

---

target/linux/qualcommax/patches-6.18/0422-spi-spi-qpic-snand-pick-default-ECC-settings-based-o.patch
```diff
@@ -0,0 +1,49 @@
...
++		} else if (mtd->oobsize < 80)
++			ecc_cfg->strength = 4;
++		else
++			ecc_cfg->strength = 8;
```
openwrt-ai, line 46: The hardcoded 64/80 thresholds are only valid for a 2048-byte page, which contradicts the commit message's own statement that "the max ECC capabilities of QPIC depends on the page size".

The real constraint is `cw_size * (writesize / 512) <= writesize + oobsize`. With `cw_size = 516 + ecc_bytes_hw + spare_bytes + bbm_size` that is 528 for 4-bit and 532 for 8-bit, i.e.

- 4-bit needs `oobsize >= writesize / 32` (64 B at 2 KiB, **128 B at 4 KiB**)
- 8-bit needs `oobsize >= writesize * 5 / 128` (80 B at 2 KiB, **160 B at 4 KiB**)

So on a 4 KiB-page SPI NAND with 128 B OOB this picks 8-bit, and the check added by `0421-…` then rejects the chip outright (532 * 8 = 4256 > 4096 + 128) — even though 4-bit fits exactly. Deriving both thresholds from `mtd->writesize` rather than hardcoding 64/80 would keep the selection correct for non-2 KiB pages.

---

target/linux/qualcommax/patches-6.18/0422-spi-spi-qpic-snand-pick-default-ECC-settings-based-o.patch
```diff
@@ -0,0 +1,49 @@
...
+ 	struct nand_ecc_props *user = &nand->ecc.user_conf;
+ 	struct nand_ecc_props *conf = &nand->ecc.ctx.conf;
+ 	struct mtd_info *mtd = nanddev_to_mtd(nand);
+@@ -265,13 +264,16 @@ static int qcom_spi_ecc_init_ctx_pipelin
+ 	if (user->step_size && user->strength) {
+ 		ecc_cfg->step_size = user->step_size;
+ 		ecc_cfg->strength = user->strength;
+-	} else if (reqs->step_size && reqs->strength) {
+-		ecc_cfg->step_size = reqs->step_size;
+-		ecc_cfg->strength = reqs->strength;
```
openwrt-ai, line 35: Dropping the `reqs` branch silently changes the effective ECC strength on every ipq50xx board that sets `nand-ecc-engine` but no explicit `nand-ecc-strength`, i.e. everything that currently takes the chip-reported requirements path:

- [ipq5018-gl-b3000.dts:197](https://github.com/openwrt/openwrt/blob/4a48b0aa1a1370fe566b885568d6a997cb10a27f/target/linux/qualcommax/dts/ipq5018-gl-b3000.dts#L197)
- [ipq5018-mx-base.dtsi:105](https://github.com/openwrt/openwrt/blob/4a48b0aa1a1370fe566b885568d6a997cb10a27f/target/linux/qualcommax/dts/ipq5018-mx-base.dtsi#L105)
- [ipq5018-ax850.dts:271](https://github.com/openwrt/openwrt/blob/4a48b0aa1a1370fe566b885568d6a997cb10a27f/target/linux/qualcommax/dts/ipq5018-ax850.dts#L271)
- [ipq5018-pz-l8.dts:105](https://github.com/openwrt/openwrt/blob/4a48b0aa1a1370fe566b885568d6a997cb10a27f/target/linux/qualcommax/dts/ipq5018-pz-l8.dts#L105)
- [ipq5018-ax6000.dts:125](https://github.com/openwrt/openwrt/blob/4a48b0aa1a1370fe566b885568d6a997cb10a27f/target/linux/qualcommax/dts/ipq5018-ax6000.dts#L125)
- plus ax830, scr50axe, wn-dax3000gr, wrc-x3000gst2

Any of those whose chip reports 4-bit/512 today gets 8-bit after this change if the part has >= 80 B OOB. That changes the on-flash codeword layout, so an already-deployed device would no longer be able to read its NAND after a sysupgrade to a build containing this patch.

Related evidence that 8-bit is not universally safe here: [ipq5018-wrc-x3000gs2.dts:151](https://github.com/openwrt/openwrt/blob/4a48b0aa1a1370fe566b885568d6a997cb10a27f/target/linux/qualcommax/dts/ipq5018-wrc-x3000gs2.dts#L151-L153) carries `/* strength=8 breaks NAND I/O, use 4 instead */`.

Has this been verified on the affected boards? If not, keeping the default at 4 and letting DT opt in to 8 would avoid the flag-day.

---

target/linux/qualcommax/dts/ipq5018-redmi-ax5400.dts
```diff
@@ -118,8 +118,6 @@
 		reg = <0>;
 
 		nand-ecc-engine = <&qpic_nand>;
-		nand-ecc-step-size = <512>;
-		nand-ecc-strength = <8>;
```
openwrt-ai, line 122: This board pinned 8-bit ECC explicitly; after the removal the strength is derived from `mtd->oobsize`, so it only stays at 8 if every flash variant fitted to this board reports >= 80 B OOB. If any variant ships with 64 B OOB it quietly drops to 4-bit, which changes the on-flash layout and makes existing installations unreadable after a sysupgrade.

Can you confirm the OOB size of the parts actually fitted? Same question for the identical removal in `target/linux/qualcommax/dts/ipq5018-mr3000d-ci.dts` lines 109-110.

---

openwrt-ai (commented):
## Commit checks

- `4a48b0a` "qualcommax: ipq50xx: correct default ECC settings" — the message describes only the new OOB-size-based default. It does not mention the two other user-visible changes in the same commit: the new probe-time hard failure added by `0421-spi-spi-qpic-snand-reject-when-no-enough-OOB-space.patch` (an unsupported OOB size now aborts `qcom_spi_ecc_init_ctx_pipelined()` instead of failing later), and the removal of `0401-spi-spi-qpic-snand-default-to-4-bit-ECC.patch`. Please either describe both in the body or split them into separate commits.


---

target/linux/qualcommax/patches-6.18/0422-spi-spi-qpic-snand-pick-default-ECC-settings-based-o.patch
openwrt-ai, line 46: The 64/80 constants are only correct for a 2048-byte page. The OOB requirement scales with the number of codewords per page, not with a fixed number:

- `bytes = ecc_bytes_hw + spare_bytes + bbm_size` → 4-bit: `7+4+1 = 12`, 8-bit: `13+2+1 = 16`
- `cw_size = cw_data(516) + bytes` → 528 / 532
- required OOB = `cw_size * cwperpage - writesize` = `16 * cwperpage` (4-bit) / `20 * cwperpage` (8-bit)

For `cwperpage == 4` (2048B page) that gives exactly 64 / 80, which is where the constants come from. For a 4096B-page chip (`cwperpage == 8`) the real requirement is 128 / 160. A 4096+128 chip would therefore be assigned 8-bit ECC here (128 ≥ 80), and the new check in `0421` would then reject it with `-EINVAL` at probe — the NAND fails to attach even though 4-bit ECC would have worked. See [`qcom_spi_ecc_init_ctx_pipelined()` cw_size computation](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/spi/spi-qpic-snand.c#L316-L322).

`cwperpage` is already computed at the top of the function, so the thresholds can be derived directly. This keeps the added-line count (and therefore the hunk header and diffstat) unchanged:

```suggestion
+		if (mtd->oobsize < 16 * cwperpage) {
+			ret = -EOPNOTSUPP;
+			goto err_free_ecc_cfg;
+		} else if (mtd->oobsize < 20 * cwperpage)
+			ecc_cfg->strength = 4;
+		else
+			ecc_cfg->strength = 8;
```

Two smaller things while respinning this hunk (both change the line count, so the diffstat/hunk header need regenerating afterwards — `make target/linux/refresh` with `qualcommax/ipq50xx` selected):

- Kernel coding style wants braces on every branch once one branch needs them; here only the first branch has them.
- The commit message repeats the same 2048-specific claim ("For ECC4, it needs 64B OOB space. For ECC8, it needs 80B OOB space") — worth rewording to say the requirement is per-codeword.

---

target/linux/qualcommax/patches-6.18/0422-spi-spi-qpic-snand-pick-default-ECC-settings-based-o.patch
```diff
@@ -0,0 +1,49 @@
...
+ 	} else {
+ 		/* use defaults */
+ 		ecc_cfg->step_size = NANDC_STEP_SIZE;
+-		ecc_cfg->strength = 4;
++		if (mtd->oobsize < 64) {
++			ret = -EOPNOTSUPP;
```
openwrt-ai, line 41: This bails out with no diagnostic at all, so a chip with too little OOB just makes the NAND disappear with a bare `-EOPNOTSUPP` in the probe chain. That directly contradicts the companion patch `0421`, whose whole stated purpose is "Print a clear error when the flash doesn't have enough space, instand of return -EBADMSG silently". Every other failure exit in [`qcom_spi_ecc_init_ctx_pipelined()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/spi/spi-qpic-snand.c#L276-L303) (`only %u bytes ECC step size is supported`, `only 4 or 8 bits ECC strength is supported`) emits a `dev_err()` first.

Please add a `dev_err()` naming the actual OOB size and the minimum required before the `goto`.

---

target/linux/qualcommax/patches-6.18/0421-spi-spi-qpic-snand-reject-when-no-enough-OOB-space.patch
```diff
@@ -0,0 +1,34 @@
...
+
+--- a/drivers/spi/spi-qpic-snand.c
++++ b/drivers/spi/spi-qpic-snand.c
+@@ -323,6 +323,15 @@ static int qcom_spi_ecc_init_ctx_pipelin
+ 	ecc_cfg->cw_size = ecc_cfg->cw_data + ecc_cfg->bytes;
+ 	bad_block_byte = mtd->writesize - ecc_cfg->cw_size * (cwperpage - 1) + 1;
+ 
++	if ((ecc_cfg->cw_size * cwperpage) > (mtd->writesize + mtd->oobsize)) {
++		dev_err(snandc->dev,
++			"Current ECC settings require %d bytes, but the flash only has %d+%d bytes.\n",
```
openwrt-ai, line 25: nit: `mtd->writesize` and `mtd->oobsize` are `unsigned int`, so they want `%u`, and none of the other `dev_err()` strings in this file end with a period (`"only 4 or 8 bits ECC strength is supported\n"`, `"failed to allocate BAM transaction\n"`).

```suggestion
+			"Current ECC settings require %d bytes, but the flash only has %u+%u bytes\n",
```

---

target/linux/qualcommax/patches-6.18/0421-spi-spi-qpic-snand-reject-when-no-enough-OOB-space.patch
```diff
@@ -0,0 +1,34 @@
...
++			(ecc_cfg->cw_size * cwperpage), mtd->writesize,
++			mtd->oobsize);
++		ret = -EINVAL;
++		goto err_free_ecc_cfg;
```
openwrt-ai, line 29: nit: this `goto err_free_ecc_cfg` sits after `snandc->qspi->oob_buf` has been `kmalloc()`'d, and [`err_free_ecc_cfg`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/spi/spi-qpic-snand.c#L397-L399) only does `kfree(ecc_cfg)`. `oob_buf` is never freed on any path — [`qcom_spi_ecc_cleanup_ctx_pipelined()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/spi/spi-qpic-snand.c#L402-L407) frees only `ecc_cfg` too — so the new error path leaks it.

The leak is pre-existing (the `qcom_alloc_bam_transaction()` failure path has the same problem), so it's fair to leave it out of this patch. But since both patches here look upstream-bound, a small preceding patch adding a `kfree(snandc->qspi->oob_buf)` unwind label would make the series cleaner and pre-empt the review comment upstream.

---

target/linux/qualcommax/dts/ipq5018-redmi-ax5400.dts
```diff
@@ -118,8 +118,6 @@
 		reg = <0>;
 
 		nand-ecc-engine = <&qpic_nand>;
```
openwrt-ai, line 120: nit: the commit subject scopes this as "ipq50xx", but only 2 of the 6 ipq5018 boards carrying explicit ECC properties are touched. Still setting `nand-ecc-strength`/`nand-ecc-step-size` after this PR:

- `target/linux/qualcommax/dts/ipq5018-scr50axe.dts:279-280` (strength 4)
- `target/linux/qualcommax/dts/ipq5018-wn-dax3000gr.dts:158-159` (strength 4)
- `target/linux/qualcommax/dts/ipq5018-wrc-x3000gs2.dts:152-153` (strength 4)
- `target/linux/qualcommax/dts/ipq5018-wrc-x3000gst2.dts:152-153` (strength 4)

I suspect the split is deliberate: the two boards you touch requested strength 8, which the new default reproduces on a ≥80B-OOB chip, whereas dropping the properties on the four 4-bit boards would silently promote them to 8-bit and change the on-flash layout. If that's the reasoning, a sentence in the commit message would help — otherwise the next person reading "correct default ECC settings for ipq50xx" will assume the remaining four were simply missed.

Also worth confirming for the two boards you *are* changing: if either flash is a 4096B-page part, see my comment on `0422` — the new default would pick 8-bit and then `0421` would refuse to attach the NAND.

---

openwrt-ai (commented):
Re-reviewed at `cab2797`. The force-push since `4a48b0a` only dropped the stray trailing-newline-removal hunk at the end of `0421` — that's fixed, thanks. Remaining comments are on the substance of the two new patches, which was not covered by the previous review.

## Commit checks

- `cab2797` "qualcommax: ipq50xx: correct default ECC settings" — unchanged from my previous review: the message body is a verbatim copy of the `0422` patch description and covers only the OOB-size-based default. It still doesn't mention the other two user-visible changes squashed into the same commit: the new probe-time hard failure added by `0421-spi-spi-qpic-snand-reject-when-no-enough-OOB-space.patch`, and the removal of `0401-spi-spi-qpic-snand-default-to-4-bit-ECC.patch` (which changes behaviour for chips advertising 1- or 2-bit ECC requirements). Please describe both in the body, or split them into separate commits.

---

robimarko:
@georgemoussalem Thoughts?

georgemoussalem:
> @georgemoussalem Thoughts?

Honest answer, don't know, but definitely worth testing. In addition, it would be good to get qcom to confirm this.
@hzyitc: can you send this upstream? The person at qcom who would know would be: 
Md Sadre Alam <md.alam@oss.qualcomm.com>

hzyitc:
> > @georgemoussalem Thoughts?
> 
> Honest answer, don't know, but definitely worth testing. In addition, it would be good to get qcom to confirm this. @hzyitc: can you send this upstream? The person at qcom who would know would be: Md Sadre Alam [md.alam@oss.qualcomm.com](mailto:md.alam@oss.qualcomm.com)

Have done it.

https://lore.kernel.org/all/SEYPR01MB58823B8C66E5044EADED1609C9D82@SEYPR01MB5882.apcprd01.prod.exchangelabs.com/

https://lore.kernel.org/all/SEYPR01MB5882463D860A9670617D5834C9A72@SEYPR01MB5882.apcprd01.prod.exchangelabs.com/

PR #24765, opened 2026-08-16
kernel: fix kmod-fb/kmod-backlight dependency direction on 6.18

divinehawk:
Since 6.18, fb.ko calls into the backlight class rather than the reverse, so it depends on backlight.ko when CONFIG_FB_BACKLIGHT is set, which in-tree only happens via FB_TFT's select. Building kmod-fb-tft on a 6.18 target with DISPLAY_SUPPORT therefore fails with:

```
 Package kmod-fb is missing dependencies for the following libraries:
 backlight.ko
```

Declare the dependency on kmod-fb, conditional on kmod-backlight being built. An unconditional +kmod-backlight would form a recursive dependency with KernelPackage/backlight's +kmod-fb, both as a kconfig select cycle and as a circular PACK_ prerequisite that make resolves by dropping an edge.

backlight.ko only references fb.ko on 6.12, so narrow that reverse dependency to +LINUX_6_12:kmod-fb, which leaves a single edge on 6.18.

### DIFF
--- a/package/kernel/linux/modules/video.mk
+++ b/package/kernel/linux/modules/video.mk
@@ -59,7 +59,7 @@ $(eval $(call KernelPackage,acpi-video))
 define KernelPackage/backlight
 	SUBMENU:=$(VIDEO_MENU)
 	TITLE:=Backlight support
-	DEPENDS:=@DISPLAY_SUPPORT +kmod-fb
+	DEPENDS:=@DISPLAY_SUPPORT +LINUX_6_12:kmod-fb
 	HIDDEN:=1
 	KCONFIG:=CONFIG_BACKLIGHT_CLASS_DEVICE \
 		CONFIG_BACKLIGHT_LCD_SUPPORT=y \
@@ -113,7 +113,7 @@ $(eval $(call KernelPackage,backlight-pwm))
 define KernelPackage/fb
   SUBMENU:=$(VIDEO_MENU)
   TITLE:=Framebuffer and framebuffer console support
-  DEPENDS:=@DISPLAY_SUPPORT
+  DEPENDS:=@DISPLAY_SUPPORT +PACKAGE_kmod-backlight:kmod-backlight
   KCONFIG:= \
 	CONFIG_FB \
 	CONFIG_FB_DEVICE=y \


package/kernel/linux/modules/video.mk
```diff
@@ -113,7 +113,7 @@ $(eval $(call KernelPackage,backlight-pwm))
 define KernelPackage/fb
   SUBMENU:=$(VIDEO_MENU)
   TITLE:=Framebuffer and framebuffer console support
-  DEPENDS:=@DISPLAY_SUPPORT
+  DEPENDS:=@DISPLAY_SUPPORT +kmod-backlight
```
openwrt-ai: This creates a circular package dependency: `KernelPackage/backlight` already carries [`DEPENDS:=@DISPLAY_SUPPORT +kmod-fb` at video.mk:62](https://github.com/openwrt/openwrt/blob/c4c5396425b792f101596fec3b8d7731639c9735/package/kernel/linux/modules/video.mk#L62), so after this change the two packages depend on each other.

`+pkg` is emitted by `scripts/package-metadata.pl` as `select PACKAGE_pkg`, so the generated `tmp/.config-package.in` ends up with a mutual select. Running the generator on just these two entries produces:

```
	config PACKAGE_kmod-backlight
		depends on DISPLAY_SUPPORT
		select PACKAGE_kmod-fb
	menuconfig PACKAGE_kmod-fb
		depends on DISPLAY_SUPPORT
		select PACKAGE_kmod-backlight
```

kconfig walks `sym->rev_dep.expr` in [`sym_check_sym_deps()`](https://github.com/openwrt/openwrt/blob/c4c5396425b792f101596fec3b8d7731639c9735/scripts/config/symbol.c#L1112-L1113) and will abort with `error: recursive dependency detected!`, which breaks `make menuconfig`/`defconfig` for every target with `DISPLAY_SUPPORT`.

The direction of the real symbol dependency also flipped between the two supported kernels, so an unconditional `+kmod-backlight` isn't right either way:

- 6.12: `backlight.ko` calls into `fb.ko` — [`fb_register_client()` at backlight.c:144](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/drivers/video/backlight/backlight.c#L144), plus `fb_bl_device()` at :102. The existing `kmod-backlight` -> `kmod-fb` edge is correct here, and `fb.ko` needs nothing from backlight.
- 6.18: the fbdev notifier block is gone from `backlight.c`, and `fb.ko` now calls backlight — [`fb_bl_notify_blank()` at fb_backlight.c:41-49](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/video/fbdev/core/fb_backlight.c#L41-L49).

On top of that, the whole of `fb_backlight.c` is inside `#if IS_ENABLED(CONFIG_FB_BACKLIGHT)`, and `CONFIG_FB_BACKLIGHT` is [`not set` in config-6.18](https://github.com/openwrt/openwrt/blob/c4c5396425b792f101596fec3b8d7731639c9735/target/linux/generic/config-6.18#L2116) — the only in-tree package turning it on is [`kmod-fb-tft` at video.mk:263](https://github.com/openwrt/openwrt/blob/c4c5396425b792f101596fec3b8d7731639c9735/package/kernel/linux/modules/video.mk#L263). So the `modinfo depends=backlight` that `scripts/gen-dependencies.sh` picks up only appears on 6.18 *with* `kmod-fb-tft` selected; in every other configuration this line just drags the backlight class device into all `kmod-fb` installs for no reason.

A fix that doesn't cycle probably has to touch both packages — e.g. make `kmod-backlight`'s `+kmod-fb` `+!LINUX_6_18:kmod-fb` and this one `+LINUX_6_18:kmod-backlight` (the `+LINUX_6_18:kmod-cec-core` form is already used at line 366) — so that only one edge exists per kernel version.

---

openwrt-ai (commented):
## Commit checks

- `c4c53964` "kernel: kmod-fb: add missing kmod-backlight dependency" — the message states flatly that "Modules packaged by kmod-fb reference symbols exported by backlight.ko", but that is only true on 6.18 *and* only when `CONFIG_FB_BACKLIGHT=y` (the entire body of `drivers/video/fbdev/core/fb_backlight.c` is guarded by `#if IS_ENABLED(CONFIG_FB_BACKLIGHT)`, and `CONFIG_FB_BACKLIGHT` is `not set` in both `target/linux/generic/config-6.12` and `config-6.18`). On 6.12 the reference goes the other way — `backlight.ko` calls `fb_register_client()`/`fb_bl_device()` from `fb.ko`. Please narrow the message to the configuration that actually reproduces the packaging error (which kernel, which package enabled `CONFIG_FB_BACKLIGHT=y`), since it determines what the correct fix looks like.


---

divinehawk:
Needs review and test

divinehawk:
Should fix https://github.com/openwrt/openwrt/pull/24757

package/kernel/linux/modules/video.mk
```diff
@@ -113,7 +113,7 @@ $(eval $(call KernelPackage,backlight-pwm))
 define KernelPackage/fb
   SUBMENU:=$(VIDEO_MENU)
   TITLE:=Framebuffer and framebuffer console support
-  DEPENDS:=@DISPLAY_SUPPORT
+  DEPENDS:=@DISPLAY_SUPPORT +PACKAGE_kmod-backlight:kmod-backlight
```
openwrt-ai, line 116: This makes `kmod-fb` and `kmod-backlight` depend on each other whenever `kmod-backlight` is selected, which is what distinguishes it from the `KernelPackage/drm` precedent cited in the commit message — `kmod-backlight` does not depend on `kmod-drm`, so that one is not a cycle. Here it is:

- [`filter_deps` at include/package-pack.mk:61](https://github.com/openwrt/openwrt/blob/e5eff3e1773a867612ff7d386aaaf4554344d87c/include/package-pack.mk#L61) resolves `PACKAGE_kmod-backlight:kmod-backlight` to plain `kmod-backlight` in `IDEPEND_kmod-fb` when `CONFIG_PACKAGE_kmod-backlight=y`.
- [`FixupDependencies` (package-pack.mk:73-78)](https://github.com/openwrt/openwrt/blob/e5eff3e1773a867612ff7d386aaaf4554344d87c/include/package-pack.mk#L73-L78) then emits `$(PACK_kmod-fb): $(PACK_kmod-backlight)`, while [`FixupReverseDependencies` (package-pack.mk:67-71)](https://github.com/openwrt/openwrt/blob/e5eff3e1773a867612ff7d386aaaf4554344d87c/include/package-pack.mk#L67-L71) emits `$(PACK_kmod-backlight): $(PACK_kmod-fb)` from [`KernelPackage/backlight`'s `+kmod-fb` (video.mk:62)](https://github.com/openwrt/openwrt/blob/e5eff3e1773a867612ff7d386aaaf4554344d87c/package/kernel/linux/modules/video.mk#L62).
- GNU make breaks such a cycle by silently dropping one edge. The package that then gets packed first writes its `.provides` without the other's `.ko`, because the collection loop skips a `.provides` file that does not exist yet, and `CheckDependencies` runs immediately after — [package-pack.mk:471-478](https://github.com/openwrt/openwrt/blob/e5eff3e1773a867612ff7d386aaaf4554344d87c/include/package-pack.mk#L471-L478).
- `compile` lists the `PACK_` targets in definition order, and `KernelPackage/backlight` is defined before `KernelPackage/fb` in this file, so on a serial clean build make should reach `PACK_kmod-backlight` first, drop its edge to `PACK_kmod-fb`, and pack `kmod-fb` first — i.e. reproduce the exact `Package kmod-fb is missing dependencies ... backlight.ko` error this change is meant to fix. With `-j` the order is simply nondeterministic.

Was this build-tested from a clean tree (empty `PKG_INFO_DIR`) with `kmod-fb-tft` selected on 6.18, i.e. with both packages actually in the build? If it was and it passed, please say so and ignore this — but the make-level cycle is real either way, and it also lands in the generated package metadata as `kmod-fb: Depends: kmod-backlight` plus `kmod-backlight: Depends: kmod-fb`.

A direction that avoids the cycle instead of relying on which edge make drops: on 6.18 `backlight.ko` no longer references `fb.ko` at all — [`drivers/video/backlight/backlight.c`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/video/backlight/backlight.c) in v6.18.44 has no `fb_*` call, whereas on v6.12.103 it does ([`fb_bl_device()` at backlight.c:102](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/drivers/video/backlight/backlight.c#L102), [`fb_register_client()` at backlight.c:144](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/drivers/video/backlight/backlight.c#L144)). So `KernelPackage/backlight`'s `+kmod-fb` is only needed on 6.12 and could be narrowed to `+LINUX_6_12:kmod-fb`, which removes the reverse edge on 6.18 without reintroducing the kconfig `select` recursion.

---

openwrt-ai (commented):
Reviewed 1 new commit (`e5eff3e1`). The rewritten commit message now matches what I can verify in the kernel trees — on v6.18.44 `fb.ko` calls `backlight_notify_blank()`/`backlight_notify_blank_all()` from `fb_backlight.c` under `CONFIG_FB_BACKLIGHT`, on v6.12.103 the reference goes the other way, `CONFIG_FB_BACKLIGHT=y` is set in-tree only by `KernelPackage/fb-tft`, and the `PACKAGE_x:x` form does suppress the kconfig `select` (`next if $depend eq $condition` in `scripts/package-metadata.pl`), so no recursive dependency there. One concern left inline about the make-level dependency cycle this introduces between `kmod-fb` and `kmod-backlight`.

---

package/kernel/linux/modules/video.mk
divinehawk: Implemented, needs a full build and retest

robimarko: Locally, it still fails:
```
Package kmod-fb is missing dependencies for the following libraries:
backlight.ko
```

robimarko:
```
diff --git a/package/kernel/linux/modules/video.mk b/package/kernel/linux/modules/video.mk
index 93725d33bb..85e20c2e0d 100644
--- a/package/kernel/linux/modules/video.mk
+++ b/package/kernel/linux/modules/video.mk
@@ -59,7 +59,7 @@ $(eval $(call KernelPackage,acpi-video))
 define KernelPackage/backlight
        SUBMENU:=$(VIDEO_MENU)
        TITLE:=Backlight support
-       DEPENDS:=@DISPLAY_SUPPORT +kmod-fb
+       DEPENDS:=@DISPLAY_SUPPORT +!LINUX_6_18:kmod-fb
        HIDDEN:=1
        KCONFIG:=CONFIG_BACKLIGHT_CLASS_DEVICE \
                CONFIG_BACKLIGHT_LCD_SUPPORT=y \
@@ -113,7 +113,7 @@ $(eval $(call KernelPackage,backlight-pwm))
 define KernelPackage/fb
   SUBMENU:=$(VIDEO_MENU)
   TITLE:=Framebuffer and framebuffer console support
-  DEPENDS:=@DISPLAY_SUPPORT
+  DEPENDS:=@DISPLAY_SUPPORT +LINUX_6_18:kmod-backlight
   KCONFIG:= \
        CONFIG_FB \
        CONFIG_FB_DEVICE=y \
```

robimarko:

PR #24467, opened 2026-07-29
ppp: fix precompiled filter option

LGA1150:
The upgrade to 2.5.0 accidentally dropped the CPP defines, so the precompiled-active-filter option has been missing and the demand dialing option in uci has been broken since then. Add the missing defines.

Fixes: 9cecf2b16e0e ("ppp: update to 2.5.0")
Closes: https://github.com/openwrt/openwrt/issues/24454

PR #24767, opened 2026-08-17
[25.12] ppp: fix precompiled filter option

LGA1150:
[ Upstream commit f59200bc56a4250028df060c893b64d82f561055 ]

The upgrade to 2.5.0 accidentally dropped the CPP defines, so the precompiled-active-filter option has been missing and the demand dialing option in uci has been broken since then. Add the missing defines.

Fixes: 9cecf2b16e0ea8560e50ef6719938bd80b963704 ("ppp: update to 2.5.0")
Closes: #24454

### DIFF
--- a/package/network/services/ppp/Makefile
+++ b/package/network/services/ppp/Makefile
@@ -10,7 +10,7 @@ include $(INCLUDE_DIR)/kernel.mk
 
 PKG_NAME:=ppp
 PKG_VERSION:=2.5.2
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/ppp-project/ppp
--- a/package/network/services/ppp/patches/310-precompile_filter.patch
+++ b/package/network/services/ppp/patches/310-precompile_filter.patch
@@ -13,17 +13,21 @@ Signed-off-by: Jo-Philipp Wich <jo@mein.io>
 
 --- a/configure.ac
 +++ b/configure.ac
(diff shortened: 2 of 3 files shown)


openwrt-ai (commented):
Reviewed 1 new commit (backport of `f59200bc56a4250028df060c893b64d82f561055` onto `openwrt-25.12`).

Diff content checks out: the two patch files end up identical to the upstream versions apart from the pre-existing ppp 2.5.2-vs-2.5.3 deltas on this branch (`configure.ac` line offsets one lower, and no `dhcpv6relay` plugin in `500-add-pptp-plugin.patch`). The `+4` shift applied to the hunk offsets in `310-precompile_filter.patch` and `500-add-pptp-plugin.patch` matches the 4 lines the new `AM_COND_IF` block adds, and 310/500 are the only in-tree ppp patches that touch `configure.ac`, so nothing else needed re-offsetting. `PKG_RELEASE` bump is there. CI is green.

## Commit checks

- f9b9d95 "ppp: fix precompiled filter option" — the `(cherry picked from commit f59200bc56a4250028df060c893b64d82f561055)` line is placed between the subject and the body rather than at the end of the message, so it does not sit in the trailer block (`git cherry-pick -x` appends it after `Signed-off-by:`). The message also diverges from upstream in a few other places: `Fixes:` carries the full 40-char SHA instead of the abbreviated `9cecf2b16e0e ("ppp: update to 2.5.0")` used on main, `Closes:` was shortened from the full issue URL to `#24454`, and upstream's `Link: https://github.com/openwrt/openwrt/pull/24467` plus `Signed-off-by: Hauke Mehrtens <hauke@hauke-m.de>` were dropped. Re-doing the pick with `git cherry-pick -x` preserves the original message verbatim and puts the trailer in the right spot.

---

PR #24768, opened 2026-08-17
[24.10] ppp: fix precompiled filter option

LGA1150:
[ Upstream commit f59200bc56a4250028df060c893b64d82f561055 ]

The upgrade to 2.5.0 accidentally dropped the CPP defines, so the precompiled-active-filter option has been missing and the demand dialing option in uci has been broken since then. Add the missing defines.

Fixes: 9cecf2b16e0ea8560e50ef6719938bd80b963704 ("ppp: update to 2.5.0")
Closes: #24454

### DIFF
--- a/package/network/services/ppp/Makefile
+++ b/package/network/services/ppp/Makefile
@@ -10,7 +10,7 @@ include $(INCLUDE_DIR)/kernel.mk
 
 PKG_NAME:=ppp
 PKG_VERSION:=2.5.1
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/ppp-project/ppp
--- a/package/network/services/ppp/patches/310-precompile_filter.patch
+++ b/package/network/services/ppp/patches/310-precompile_filter.patch
@@ -13,17 +13,21 @@ Signed-off-by: Jo-Philipp Wich <jo@mein.io>
 
 --- a/configure.ac
 +++ b/configure.ac
-@@ -306,6 +306,9 @@ AM_CONDITIONAL(PPP_WITH_PAM, test "x${wi
+@@ -306,6 +306,13 @@ AM_CONDITIONAL(PPP_WITH_PAM, test "x${wi
  # With libpcap support, activate pppd on network activity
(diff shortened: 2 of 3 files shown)


openwrt-ai (commented):
## Commit checks

- 1b3b598 "ppp: fix precompiled filter option" — the cherry-pick trailer names the correct upstream commit ([f59200bc](https://github.com/openwrt/openwrt/commit/f59200bc56a4250028df060c893b64d82f561055)), but the message body was rewritten rather than carried over as `git cherry-pick -x` produces it:
  - the `(cherry picked from commit ...)` line sits between the subject and the body; on this branch it goes at the *end* of the message (see e.g. [cd699811](https://github.com/openwrt/openwrt/commit/cd699811a8c02c63d2755bd6d76cf9db229cf0c8) and the surrounding 24.10 backports).
  - upstream's `Link: https://github.com/openwrt/openwrt/pull/24467` and `Signed-off-by: Hauke Mehrtens <hauke@hauke-m.de>` trailers were dropped; 24.10 backports normally keep the upstream message verbatim and append the new `Link:`/`Signed-off-by:` below the cherry-pick line.
  - `Fixes:` was expanded to the full 40-char SHA and `Closes:` shortened to `#24454`, where upstream has `9cecf2b16e0e ("ppp: update to 2.5.0")` and the full issue URL.

The referenced `Fixes:` commit 9cecf2b16e0e is present on `openwrt-24.10`, so the reference is valid on this branch.

Diff review: the backport is a faithful adaptation. Both patch files differ from their upstream (main) counterparts only in hunk-header line numbers, which is expected because 24.10 ships ppp 2.5.1 rather than 2.5.3; the added/removed lines are identical. The rebased offsets check out: `310-precompile_filter.patch` grows the first `configure.ac` hunk by 4 lines (`+306,9` → `+306,13`), so the second hunk's new-side offset correctly moves `362` → `366` with the old side unchanged at `359`, and `500-add-pptp-plugin.patch` correctly shifts `344` → `348` since it applies on top of 310 (those are the only two ppp patches on this branch touching `configure.ac`). `PKG_FIXUP:=autoreconf` is set, so the `configure.ac` change takes effect, and `PKG_RELEASE` is bumped. No inline issues found.

---

PR #24770, opened 2026-08-17
qualcomm: remove useless cpu erratum 

aiamadeus:
These targets currently uses CPU erratums for Cortex-A76, Cortex-A510,
Cortex-A710, and Neoverse-N2, which is useless and should be removed.


openwrt-ai (commented):
Reviewed 3 new commits. The config removals themselves check out: every prompted symbol dropped here (`ARM64_ERRATUM_1165522/1286807/2051678/2054223/2067961/2077057/2658417`, `AMPERE_ERRATUM_AC04_CPU_23`, `ROCKCHIP_ERRATUM_3568002`) already has a matching `# CONFIG_... is not set` line in `target/linux/generic/config-6.18`, so removing the `=y` from the target configs actually disables them rather than falling back to the `default y` in `arch/arm64/Kconfig`. The three `ARM64_WORKAROUND_*` symbols are prompt-less select-only booleans, so dropping them alongside the errata that selected them is correct. Neither target has a CPU affected by these errata (qualcommax is Cortex-A53, qualcommbe/IPQ9574 is Cortex-A73). Two commit-message nits below.

## Commit checks

- `e0ce0db` "qualcommax: move EC kconfig to generic" — the subject is scoped to `qualcommax`, but half the diff lands in `target/linux/generic/config-6.18`, which is shared by every target. In-tree precedent for moving symbols into the generic config uses a `kernel:` prefix, e.g. [860652f4b9da "kernel: 5.4: move some kconfig options to generic"](https://github.com/openwrt/openwrt/commit/860652f4b9da40c80ae96d55778a3cd887514e3d) and [1d03283eb240 "kernel: 5.4: remove some dup config options already set in generic"](https://github.com/openwrt/openwrt/commit/1d03283eb240759b58d7246ba72fabdb35fddb29). Consider `kernel: ...` (or `generic: ...`) so the subject scope matches the files touched.

- `e530ce9` "qualcommbe: remove useless cpu erratum" — the `Fixes:` trailer abbreviates the SHA to 7 characters. The commit referenced does exist ([0b25cef749314bafdf9e3b1bc6bf5aa48f57f44c](https://github.com/openwrt/openwrt/commit/0b25cef749314bafdf9e3b1bc6bf5aa48f57f44c)) and the claim is accurate — it did add both `CONFIG_AMPERE_ERRATUM_AC04_CPU_23=y` and `CONFIG_ROCKCHIP_ERRATUM_3568002=y` — but OpenWrt follows the kernel convention of at least 12 hex characters in `Fixes:`, cf. [8db23dc91a01](https://github.com/openwrt/openwrt/commit/8db23dc91a015bf843f1e3fbd0891574594e86f9) which uses `Fixes: 01fb4a6daadb ("qualcommbe: update pwm patches and add missing symbol")`. Please use `Fixes: 0b25cef74931 ("qualcommbe: enable 6.18 testing kernel")`; the same 7-char abbreviation appears in the body text ("The commit 0b25cef also added erratums ...").


---

robimarko:

PR #24771, opened 2026-08-17
lantiq: xrx200: fix fiber led on AVM FRITZ!Box 5490/5491

abajk:
The AVM 5490 and 5491 do not have Internet, LAN and DSL LEDs. Instead, they have a fiber and FON LED. This commit adds a trigger for the fiber LED and removes triggers for the LAN, DSL and Internet LEDs.

Fixes: 48b2df5a4164 ("lantiq: add support for AVM Fritzbox 5490/5491")

### DIFF
--- a/target/linux/lantiq/dts/vr9_avm_fritz5490.dtsi
+++ b/target/linux/lantiq/dts/vr9_avm_fritz5490.dtsi
@@ -10,14 +10,13 @@
 };
 
 &aliases {
-	led-dsl = &led_info_green;
-	led-internet = &led_internet;
 	led-wifi = &led_wifi;
 };
 
 &leds {
-	led_fiber: fiber {
-		label = "green:fiber";
+	fiber {
+		function = "fiber";
+		color = <LED_COLOR_ID_GREEN>;
 		gpios = <&gpio 47 GPIO_ACTIVE_LOW>;
 	};
 
@@ -27,8 +26,9 @@
 		gpios = <&gpio 36 GPIO_ACTIVE_LOW>;
 	};
 
-	led_internet: internet {
-		label = "green:internet";
+	fon {
+		function = "fon";
+		color = <LED_COLOR_ID_GREEN>;
 		gpios = <&gpio 35 GPIO_ACTIVE_LOW>;
 	};
 };
--- a/target/linux/lantiq/xrx200/base-files/etc/board.d/01_leds
+++ b/target/linux/lantiq/xrx200/base-files/etc/board.d/01_leds
@@ -48,11 +48,13 @@ avm,fritz3370-rev2-micron|\
 avm,fritz3390|\
 avm,fritz3490|\
 avm,fritz3490-micron|\
-avm,fritz5490|\
-avm,fritz5490-micron|\
 avm,fritz7490|\
 avm,fritz7490-micron)
 	ucidef_set_led_switch "lan" "LAN" "green:lan" "switch0" "0x17"
+	;;
+avm,fritz5490|\
+avm,fritz5490-micron)
+	ucidef_set_led_netdev "fiber" "fiber" "green:fiber" "wan" "link"
 	;;
 bt,homehub-v5a)
 	ucidef_set_led_default "dimmed" "dimmed" "dimmed" "0"


target/linux/lantiq/dts/vr9_avm_fritz5490.dtsi
```diff
@@ -27,8 +25,8 @@
 		gpios = <&gpio 36 GPIO_ACTIVE_LOW>;
 	};
 
-	led_internet: internet {
-		label = "green:internet";
+	fon {
+		label = "green:fon";
```
openwrt-ai: New LED node introduced with the deprecated `label = "<color>:<function>"` syntax; use the `color`/`function` bindings instead. There is no `LED_FUNCTION_FON` in [`dt-bindings/leds/common.h`](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/include/dt-bindings/leds/common.h), so the string fallback is the right form here — exactly what [`vr9_avm_fritz736x.dtsi:58-60`](https://github.com/openwrt/openwrt/blob/98b8e65a5ce1b16ed6eb454803e5c1d19c5de527/target/linux/lantiq/dts/vr9_avm_fritz736x.dtsi#L58-L60) already does for its `fon` LED, and what `led_wifi` a few lines above does here.

The resulting LED class device name stays `green:fon` (`led_compose_name()` builds `"<color>:<function>"` when no `label` is present), so nothing in `board.d` changes.

```suggestion
	fon {
		function = "fon";
		color = <LED_COLOR_ID_GREEN>;
```

Same applies to the `fiber` node at lines 17-18, whose opening line this commit already touches — `function = "fiber"; color = <LED_COLOR_ID_GREEN>;` also yields `green:fiber`, so the hardcoded string in `01_leds` keeps matching.

---

openwrt-ai (commented):
Reviewed 1 new commit; one inline comment.

Verified alongside the review: `Fixes: 48b2df5a4164` resolves to the correct commit and subject; the `wan` netdev named in the new `ucidef_set_led_netdev` call exists for `avm,fritz5490*` (`02_network` sets `ucidef_set_interfaces_lan_wan "lan1 lan2 lan3 lan4" "wan"`); the 5-argument form of `ucidef_set_led_netdev` (with an explicit `mode`) is supported; and dropping `led_fiber:`/`led_internet:` leaves no dangling phandle references in-tree. Removing `avm,fritz5490*` from the `green:lan` switch-LED group is consistent with the DTS, which never defined a LAN LED for this board. CI on `98b8e65` is green.

---

target/linux/lantiq/dts/vr9_avm_fritz5490.dtsi
abajk: Fixed

openwrt-ai (commented):
Reviewed 1 new commit (force-push `98b8e65` → `bb3a911`); no new issues found.

The only change since the last review is the DTS LED conversion, which resolves the earlier comment: both `fiber` and `fon` now use `function`/`color` instead of `label`, matching `led_wifi` in the same file. Re-verified alongside:

- `dt-bindings/leds/common.h` is already included by `vr9_avm_fritz5490.dtsi` (and by the parent `vr9_avm_fritzxx90.dtsi`), so `LED_COLOR_ID_GREEN` resolves.
- The resulting LED class device names are unchanged at `green:fiber` / `green:fon`: with no `label` and `devname_mandatory` false for `gpio-leds`, [`led_compose_name()` at led-core.c:552-570](https://github.com/gregkh/linux/blob/9d1a1a1a1d7a0e4e0d7d6f1b6c0a0f0e0d0c0b0a/drivers/leds/led-core.c#L552-L570) builds `"<color>:<function>"`. So the hardcoded `"green:fiber"` string in `01_leds` still matches.
- Dropping the `led_fiber:`/`led_internet:` phandle labels and the `led-dsl`/`led-internet` aliases leaves no dangling references in-tree; `led_info_green` remains defined in the parent dtsi for the other FRITZ!Box boards.
- Commit message still matches the diff, and `Fixes: 48b2df5a4164` resolves to the correct commit and subject.

No CI failures reported on `bb3a911`.

---

openwrt-ai:
Correction to my review above: the `led_compose_name()` permalink carries a bad commit SHA and will 404. The correct link (Linux v6.12.103, the `KERNEL_PATCHVER` for lantiq) is [`led_compose_name()` at led-core.c:552-570](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/drivers/leds/led-core.c#L552-L570). The point it supports is unchanged: with no `label` and `devname_mandatory` false, the LED name is composed as `"<color>:<function>"`, so `green:fiber` / `green:fon` are preserved.

---

PR #24007, opened 2026-06-30
kernel: fix FM25G01B/FM25G02B Quad I/O read dummy

Sanderrrs:
The FM25G01B/FM25G02B datasheets specify a single dummy byte for the 0xEB Quad I/O read-from-cache, but mainline's generic QUADIO template issues two. The extra dummy byte shifts the data phase and returns corrupted data with no ECC error, breaking boot on NC-1812/KN-1812. Use a dedicated read-from-cache variant set with ndummy=1 for 0xEB. FM25G01B datasheet: https://www.fmsh.com/nvm/FM25G01B_ds_eng.pdf FM25G02B datasheet: https://www.fmsh.com/nvm/FM25G02B_ds_eng.pdf

Tested on NC-1812

PR #24772, opened 2026-08-17
kernel: update FM25G{01,02}B SPI-NAND support to the accepted version

lexfrei:
The FM25G{01,02}B support carried in generic/pending-6.18 got accepted upstream: d5a5c9eb2ee9 in the MTD tree, heading for Linux 7.3. Upstream also took a follow-up fix right after (8211f2d74b35, Cc: stable), and it matters: as carried, the driver returns garbage on quad reads. These chips expect one dummy byte on the 0xEB Quad I/O read-from-cache, the generic read_cache_variants set sends two, so the data phase shifts by a byte and ECC still reports all clean. Silent corruption, no error anywhere. Any controller that selects the quad path hits this; FM25G02B ships in current Keenetic/Netcraze units.

This moves the pending patch to backport-6.18 with the upstream commit header (the accepted v5 is code-identical to the pending v3) and takes the fix on top as 438. Both drop out once the kernel moves past 7.3.


lexfrei:
Duplicate of #24007, which is hardware-tested and already approved. Closing in its favor.

PR #10592, opened 2022-09-03
add support for Midea morr-qr011

lonesafe:
Add support for Midea morr-qr011 device.

PR #24777, opened 2026-08-18
ramips: add support for Midea MORR-QR011

lonesafe:
## Summary

Add support for the Midea MORR-QR011, a wired MT7621A router with
256 MiB RAM, 32 MiB SPI NOR, one WAN port, four LAN ports and USB 3.0.

This supersedes #10592. The change is split into two signed-off commits:

1. A generic Linux 6.18 SPI-NOR fix for the MX25L25635E.
2. The ramips device DTS and image definition.

## Root cause of the SPI NOR failure

The MX25L25635E on the tested unit exposes a BFPT that Linux 6.18 cannot
parse. Without a legacy size, `spi_nor_needs_sfdp()` requires SFDP and the
probe returns `-EINVAL`. Ethernet then remains deferred because its MAC
address depends on NVMEM from the factory partition, and squashfs cannot
find its root filesystem.

The generic patch restores the name, 32 MiB size and legacy dual/quad read
capabilities used by Linux 6.12. This makes SFDP optional and permits the
known parameters to be used when BFPT parsing fails.

## Review changes

- Moved the SPI-NOR fix to `target/linux/generic/pending-6.18/` and split it
  from the board-support commit.
- Refreshed the patch against Linux 6.18.44; a second
  `make target/linux/refresh` leaves it unchanged.
- Removed the unused PCIe enablement; hardware logs report no device on any
  PCIe port.
- Removed the unverified `uart3` GPIO mux and retained only `wdt` for the
  reset button on GPIO 18.
- Expanded the device commit with literal Breed boot commands, recovery and
  stock-revert notes, MAC layout, and the known hardware inventory.

## Validation

- Rebased on current `openwrt/openwrt:main` (`7b7cf1454a`).
- `scripts/checkpatch.pl` passes both commits with 0 errors and 0 warnings.
- A clean ramips/mt7621 device-only build completes successfully.
- Both initramfs and squashfs-sysupgrade images are generated.
- The compiled profile reports the expected 32 MiB image limit and
  `midea,morr-qr011` supported-device string.
- Initramfs and squashfs-sysupgrade were boot-tested on hardware; SPI NOR,
  Ethernet links and DHCP work.

The board has one system/status LED, but its GPIO and polarity have not been
traced, so no LED node is guessed. The UART pad order and power-adapter rating
were likewise not recorded on the tested unit.

### DIFF
--- a/target/linux/generic/pending-6.18/478-mtd-spi-nor-macronix-restore-mx25l25635e-legacy.patch
+++ b/target/linux/generic/pending-6.18/478-mtd-spi-nor-macronix-restore-mx25l25635e-legacy.patch
@@ -1,0 +1,30 @@
+From: Zhenxiao Wang <85195433@qq.com>
+Date: Tue, 18 Aug 2026 16:00:00 +0800
+Subject: [PATCH] mtd: spi-nor: macronix: restore MX25L25635E legacy parameters
+
+Linux v6.12 described the MX25L25635E with a name, size and legacy read
+capabilities. These parameters were lost while retaining the JEDEC ID and
+chip fixups.
+
+Without a size, spi_nor_needs_sfdp() requires SFDP. Some MX25L25635E chips
+expose a BFPT which cannot be parsed, causing probing to fail. Restore the
+v6.12 parameters so SFDP is optional and the core can use the known 32 MiB
+geometry and read capabilities when BFPT parsing fails.
+
+Signed-off-by: Zhenxiao Wang <85195433@qq.com>
+---
+ drivers/mtd/spi-nor/macronix.c | 3 +++
+ 1 file changed, 3 insertions(+)
+
+--- a/drivers/mtd/spi-nor/macronix.c
++++ b/drivers/mtd/spi-nor/macronix.c
+@@ -136,6 +136,9 @@ static const struct flash_info macronix_
+ 	}, {
+ 		/* MX25L25635E, MX25L25645G */
+ 		.id = SNOR_ID(0xc2, 0x20, 0x19),
++		.name = "mx25l25635e",
++		.size = SZ_32M,
++		.no_sfdp_flags = SPI_NOR_DUAL_READ | SPI_NOR_QUAD_READ,
+ 		.fixups = &mx25l25635_fixups
+ 	}, {
+ 		/* MX66L51235F */
--- a/target/linux/ramips/dts/mt7621_midea_morr-qr011.dts
+++ b/target/linux/ramips/dts/mt7621_midea_morr-qr011.dts
@@ -1,0 +1,129 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+#include "mt7621.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+
+/ {
+	compatible = "midea,morr-qr011", "mediatek,mt7621-soc";
+	model = "Midea MORR-QR011";
+
+	aliases {
+		label-mac-device = &gmac0;
+	};
+
+	chosen {
+		bootargs = "console=ttyS0,115200";
+		bootargs-override = "console=ttyS0,115200";
+	};
+
+	keys {
+		compatible = "gpio-keys";
+
+		reset {
+			label = "reset";
+			gpios = <&gpio 18 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
+	};
+};
+
+&spi0 {
+	status = "okay";
+
+	flash@0 {
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <50000000>;
+		broken-flash-reset;
+		m25p,fast-read;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "u-boot";
+				reg = <0x0 0x30000>;
+				read-only;
+			};
+
+			partition@30000 {
+				label = "u-boot-env";
+				reg = <0x30000 0x10000>;
+				read-only;
+			};
+
+			partition@40000 {
+				label = "factory";
+				reg = <0x40000 0x10000>;
+				read-only;
+
+				nvmem-layout {
+					compatible = "fixed-layout";
+					#address-cells = <1>;
+					#size-cells = <1>;
+
+					macaddr_factory_e000: macaddr@e000 {
+						reg = <0xe000 0x6>;
+					};
+
+					macaddr_factory_e006: macaddr@e006 {
+						reg = <0xe006 0x6>;
+					};
+				};
+			};
+
+			partition@50000 {
+				compatible = "denx,uimage";
+				label = "firmware";
+				reg = <0x50000 0x1fb0000>;
+			};
+		};
+	};
+};
+
+&gmac0 {
+	nvmem-cells = <&macaddr_factory_e000>;
+	nvmem-cell-names = "mac-address";
+};
+
+&switch0 {
+	ports {
+		port@0 {
+			status = "okay";
+			label = "wan";
+			nvmem-cells = <&macaddr_factory_e006>;
+			nvmem-cell-names = "mac-address";
+		};
+
+		port@1 {
+			status = "okay";
+			label = "lan1";
+		};
+
+		port@2 {
+			status = "okay";
+			label = "lan2";
+		};
+
+		port@3 {
+			status = "okay";
+			label = "lan3";
+		};
+
+		port@4 {
+			status = "okay";
+			label = "lan4";
+		};
+	};
+};
+
+&state_default {
+	gpio {
+		groups = "wdt";
+		function = "gpio";
+	};
+};
--- a/target/linux/ramips/image/mt7621.mk
+++ b/target/linux/ramips/image/mt7621.mk
@@ -2268,6 +2268,15 @@ define Device/mercusys_mr70x-v1
 endef
 TARGET_DEVICES += mercusys_mr70x-v1
 
+define Device/midea_morr-qr011
+  $(Device/dsa-migration)
+  IMAGE_SIZE := 32448k
+  DEVICE_VENDOR := Midea
+  DEVICE_MODEL := MORR-QR011
+  DEVICE_PACKAGES := kmod-usb3 -uboot-envtools -wpad-basic-mbedtls
+endef
+TARGET_DEVICES += midea_morr-qr011
+
 define Device/MikroTik
   $(Device/dsa-migration)
   DEVICE_VENDOR := MikroTik


target/linux/ramips/patches-6.18/406-mtd-spi-nor-macronix-restore-mx25l25635e-legacy.patch
```diff
@@ -0,0 +1,26 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Zhenxiao Wang <85195433@qq.com>
```
openwrt-ai: The all-zero SHA in the `From ... Mon Sep 17 00:00:00 2001` line is a `git format-patch` placeholder for a commit that has no upstream equivalent, and it is misleading here. None of the 46 patches in `target/linux/ramips/patches-6.18/` carry a zero SHA; locally authored ones start straight at `From:`, e.g. [330-mt7620-avoid-direct-jal-in-random-helper.patch:1](``https://github.com/openwrt/openwrt/blob/69b1843321ccfe7ef478bc26852317593defc9f4/target/linux/ramips/patches-6.18/330-mt7620-avoid-direct-jal-in-random-helper.patch#L1).`` Drop the line.

```suggestion
From: Zhenxiao Wang <85195433@qq.com>
```

---

target/linux/ramips/patches-6.18/406-mtd-spi-nor-macronix-restore-mx25l25635e-legacy.patch
```diff
@@ -0,0 +1,26 @@
...
+known parameters when SFDP parsing fails.
+
+Signed-off-by: Zhenxiao Wang <85195433@qq.com>
+---
+ drivers/mtd/spi-nor/macronix.c | 3 +++
+ 1 file changed, 3 insertions(+)
+
+--- a/drivers/mtd/spi-nor/macronix.c
++++ b/drivers/mtd/spi-nor/macronix.c
+@@ -135,6 +135,9 @@ static const struct flash_info macronix_n
```
openwrt-ai: The hunk header is off by one line, so the patch applies with an offset instead of cleanly. `target/linux/generic/kernel-6.18` pins 6.18.44, and in that tree the first context line (`}, {`) is at [macronix.c:136](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mtd/spi-nor/macronix.c#L136-L141), not 135. No other in-tree patch touches `drivers/mtd/spi-nor/macronix.c`, so nothing shifts the line numbers. Applying it against a pristine 6.18.44 tree gives:

```
checking file drivers/mtd/spi-nor/macronix.c
Hunk #1 succeeded at 136 (offset 1 line).
```

Regenerate with `make target/linux/refresh` (ramips selected in `.config`), which produces the correct header:

```suggestion
@@ -136,6 +136,9 @@ static const struct flash_info macronix_n
```

---

target/linux/generic/pending-6.18/478-mtd-spi-nor-macronix-restore-mx25l25635e-legacy.patch
```diff
@@ -0,0 +1,26 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Zhenxiao Wang <85195433@qq.com>
+Date: Tue, 18 Aug 2026 16:00:00 +0800
+Subject: [PATCH] mtd: spi-nor: macronix: restore MX25L25635E legacy parameters
```
openwrt-ai, line 3: nit: this kernel patch is folded into the commit whose subject is `ramips: add support for Midea MORR-QR011`, but its effect is not board-scoped — it changes the `SNOR_ID(0xc2, 0x20, 0x19)` entry for every MX25L25635E/MX25L25645G user. Splitting it into its own commit (e.g. `ramips: fix MX25L25635E probing on kernel 6.18`) would keep the device-support commit reviewable on its own and make the fix easy to cherry-pick.

Two related points worth considering while splitting:

- The regression is in a generic driver and affects any target with this flash, not just ramips — `target/linux/generic/pending-6.18/` may be the better home, plus a submission upstream.
- The change is a faithful restore of the 6.12 entry (verified against `drivers/mtd/spi-nor/macronix.c` in v6.12.103: same `.name`, `.size = SZ_32M`, `.no_sfdp_flags = SPI_NOR_DUAL_READ | SPI_NOR_QUAD_READ`), and setting a non-zero `.size` is exactly what flips `spi_nor_needs_sfdp()` to false so a BFPT parse failure stops being fatal. Stating that mechanism in the patch body would help the next person who touches this entry.

---

target/linux/ramips/dts/mt7621_midea_morr-qr011.dts
```diff
@@ -0,0 +1,133 @@
...
+				label = "firmware";
+				reg = <0x50000 0x1fb0000>;
+			};
+		};
+	};
+};
+
+&pcie {
+	status = "okay";
+};
```
openwrt-ai: PCIe is enabled here but nothing appears to sit behind it: the commit message describes a wired router with no WiFi, the `mt7621.mk` entry drops wpad with `-wpad-basic-mbedtls` and pulls in no `kmod-mt76*` firmware, and the `factory` partition declares only the two MAC cells — no WiFi EEPROM cell. `&pcie` is `status = "disabled"` in [mt7621.dtsi:669](https://github.com/openwrt/openwrt/blob/69b1843321ccfe7ef478bc26852317593defc9f4/target/linux/ramips/dts/mt7621.dtsi#L669), and other wired-only MT7621 boards leave it that way — `mt7621_tplink_er605-v1.dts` and `-v2.dts` even mux the `pcie` pin group to GPIO instead.

Is this override intentional (some non-WiFi PCIe device on the board), or leftover? If nothing is attached, dropping the block avoids bringing up the controller for nothing.

---

target/linux/ramips/dts/mt7621_midea_morr-qr011.dts
```diff
@@ -0,0 +1,133 @@
...
+		};
+	};
+};
+
+&state_default {
+	gpio {
+		groups = "uart3", "wdt";
+		function = "gpio";
+	};
+};
```
openwrt-ai, line 129: nit: `"wdt"` is needed — it frees GPIO 18 for the reset key (`wdt_grp` is `FUNC("wdt rst", 0, 18, 1)` in [pinctrl-mt7621.c:50-53](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/pinctrl/mediatek/pinctrl-mt7621.c#L50-L53)) — but `"uart3"` frees GPIO 5-8 ([`uart3_grp`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/pinctrl/mediatek/pinctrl-mt7621.c#L39-L43)) and no node in this DTS consumes any of them. Combined with the absence of a `leds` node and of `led-boot`/`led-failsafe`/`led-running`/`led-upgrade` aliases, this looks like the remnant of an LED block that was dropped.

Do the board's LEDs sit on GPIO 5-8? If so they are worth describing (users otherwise get no boot/failsafe indication); if the board genuinely has no GPIO LEDs, `"uart3"` can come out of the group list.

---

openwrt-ai (commented):
## Commit checks

- `69b1843` "ramips: add support for Midea MORR-QR011" — the message carries all three required sections for a new device, but two of them are thin:
  - **Hardware specification**: no LED inventory at all (only `Button: reset` is listed), no UART header location or pinout (only the 3.3 V level and 115200 8N1), and no power input (voltage + current).
  - **Flash instructions**: the Breed procedure is paraphrased rather than given verbatim — there is no literal image URL or `wget` invocation, no literal `boot mem` example address, no statement of how Breed's autoboot is interrupted (which key, at which prompt) or which IP/port Breed serves on. "Recovery uses the same Breed initramfs procedure" covers re-flashing OpenWrt, but there is no reverting-to-stock note. As written a stranger with the same board would have to guess at several steps.

  The MAC address layout section is fine and matches the DTS (`macaddr_factory_e000` on `&gmac0`, `macaddr_factory_e006` on the WAN port), as do the switch port layout and the 32448k `IMAGE_SIZE` against the `0x50000 + 0x1fb0000` firmware partition.


---

target/linux/ramips/patches-6.18/406-mtd-spi-nor-macronix-restore-mx25l25635e-legacy.patch
lonesafe: Agreed. This is a locally authored OpenWrt patch, not a `git format-patch` import from an upstream commit. I will remove the all-zero placeholder so the patch starts directly with `From: Zhenxiao Wang <85195433@qq.com>`.

target/linux/ramips/patches-6.18/406-mtd-spi-nor-macronix-restore-mx25l25635e-legacy.patch
lonesafe: Confirmed. The failed kernel-patch CI reports the same issue: this hunk applies with an offset and `make target/linux/refresh` changes the start from line 135 to 136. I will use the refreshed form rather than editing the hunk header by hand.

target/linux/generic/pending-6.18/478-mtd-spi-nor-macronix-restore-mx25l25635e-legacy.patch
lonesafe: Agreed. The fallback changes the generic Macronix entry and is not board-scoped. I will split it into a preceding commit, move it to `target/linux/generic/pending-6.18/`, and explain that restoring a non-zero size makes `spi_nor_needs_sfdp()` false, allowing the known legacy parameters to be used when this device's BFPT cannot be parsed. The device-support commit will then contain only the board DTS/image definition. I will also prepare the driver fix for upstream Linux separately.

target/linux/ramips/dts/mt7621_midea_morr-qr011.dts
lonesafe: This override is not intentional. The tested boot log reports `pcie0 no card`, `pcie1 no card`, `pcie2 no card`, followed by `nothing connected in virtual bridges`. The block was inherited from the 2022 DTS submission, so I will remove the `&pcie` override and leave the controller disabled.

target/linux/ramips/dts/mt7621_midea_morr-qr011.dts
lonesafe: The older device report mentions one system/status LED, but the available test data does not establish its GPIO or polarity, and the DTS has never described an LED node. I therefore cannot confirm that GPIO 5-8 drive it. The `uart3` group is an unsubstantiated remnant inherited from the 2022 DTS; I will remove `"uart3"` and retain only `"wdt"` for the reset key. I will list the system LED as present but currently unsupported rather than inventing a GPIO mapping.

lonesafe:
Regarding the new Formality Check suggestion: the embedded patch already contains `From:`, `Date:` and `Subject:` headers. It intentionally does not contain a `From <hash> Mon Sep 17 00:00:00 2001` line because this is a locally authored OpenWrt pending patch with no upstream commit hash. The previous all-zero placeholder was removed in response to the inline review, since a fabricated hash would be misleading. Both current Formality Check jobs and `scripts/checkpatch.pl` pass. If the driver fix is accepted upstream later, the patch can be replaced by a proper upstream/backport patch carrying that real commit hash.

target/linux/ramips/dts/mt7621_midea_morr-qr011.dts
```diff
@@ -0,0 +1,129 @@
...
+
+	chosen {
+		bootargs = "console=ttyS0,115200";
+		bootargs-override = "console=ttyS0,115200";
+	};
+
+	keys {
+		compatible = "gpio-keys";
+
+		reset {
```
openwrt-ai, line 24: The child node name `reset` does not match the `gpio-keys` binding's node-name pattern, so this node is silently unvalidated by `dtbs_check`. The pattern is `^(button|event|key|switch|(button|event|key|switch)-[a-z0-9-]+|[a-z0-9-]+-(button|event|key|switch))$` ([gpio-keys.yaml:26](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/input/gpio-keys.yaml#L26)) — a bare `reset` matches none of the three alternatives.

110 older mt7621 DTS files still use the bare `reset` name, but the most recent device added to this target uses the conforming form: [`key-reset` in mt7621_zte_mf286c3.dtsi:98](https://github.com/openwrt/openwrt/blob/1069f30699476b661b153241719ee3a7511cd9bd/target/linux/ramips/dts/mt7621_zte_mf286c3.dtsi#L98). New files should follow the binding.

```suggestion
		key-reset {
```

---

target/linux/generic/pending-6.18/478-mtd-spi-nor-macronix-restore-mx25l25635e-legacy.patch
```diff
@@ -0,0 +1,30 @@
...
+
+Linux v6.12 described the MX25L25635E with a name, size and legacy read
+capabilities. These parameters were lost while retaining the JEDEC ID and
+chip fixups.
```
openwrt-ai, line 7: nit: "These parameters were lost" doesn't name the commit that lost them, and there is a concurrent PR fixing the neighbouring entry in the same table that does.

openwrt/openwrt#24783 adds `target/linux/generic/pending-6.18/498-…` and `499-mtd-spi-nor-macronix-Restore-fallback-parameters-for.patch` to `macronix_nor_parts[]` for the same class of regression on the adjacent `SNOR_ID(0xc2, 0x20, 0x18)` entry. Its 499 patch is the upstream-accepted fix (Reviewed-by Michael Walle and Miquel Raynal) and carries `Fixes: 947c86e481a0 ("mtd: spi-nor: macronix: Drop the redundant flash info fields")` plus `Cc: stable@vger.kernel.org` — i.e. the same cleanup that dropped `.name`/`.size`/`.no_sfdp_flags` from the `0xc2 0x20 0x19` entry you are restoring here. Adding the matching `Fixes:` tag (and `Cc: stable`) would make this patch's provenance clear in-tree and make it directly submittable when you send it upstream, as you said you would.

I could not verify that SHA myself in this session — it comes from #24783's patch header, so please confirm it before adding the tag.

No merge hazard either way: I applied 478, then 498, then 499 against a pristine v6.18.44 `drivers/mtd/spi-nor/macronix.c` and all three apply with no offsets, in either merge order. Worth a heads-up to the maintainer that the two PRs touch the same table.

---

openwrt-ai (commented):
Re-reviewed at `ecdf70f`, covering the two new commits since `69b1843`. All five findings from the previous round are addressed: the zero-SHA `From` line is gone, the hunk header is now `@@ -136,6 +136,9 @@` (verified against v6.18.44 — the `}, {` context line is at 136), the patch moved to `target/linux/generic/pending-6.18/478-…` and split into its own `kernel: mtd: …` commit, the `&pcie` override is removed, and `"uart3"` is dropped from the pinmux group list.

Spot-checks that came back clean:

- The restored fields are a verbatim match for the v6.12.103 entry (`.name = "mx25l25635e"`, `.size = SZ_32M`, `.no_sfdp_flags = SPI_NOR_DUAL_READ | SPI_NOR_QUAD_READ`).
- Patch number 478 is free and sits in the existing spi-nor block (476, 477, 479); no other in-tree patch touches `drivers/mtd/spi-nor/macronix.c` on this target.
- `$(Device/dsa-migration)` is correct here rather than redundant — `mt7621/base-files/etc/board.d/05_compat-version` defaults every board to `1.1`, so omitting it would leave `DEVICE_COMPAT_VERSION` at `1.0` and break sysupgrade against the running device.
- `kmod-usb3` needs no DTS work; `xhci@1e1c0000` is enabled by default in `mt7621.dtsi`.
- `IMAGE_SIZE := 32448k` matches the `0x50000 + 0x1fb0000` firmware partition, and the `midea_morr-qr011` block is correctly ordered between `mercusys_mr70x-v1` and `Device/MikroTik`.
- Both commit messages match their diffs, and the device commit now carries all three required sections with the unrecorded items (UART pad order, LED GPIO, PSU rating) explicitly called out rather than guessed.

Two minor items inline.

---

PR #24778, opened 2026-08-18
ath79: add support for TP-Link TL-WR942N v1

dormancygrace:
## Hardware

- Qualcomm Atheros QCA9561 at 775 MHz (DDR/AHB at 650/258 MHz)
- 128 MiB DDR2 RAM
- 16 MiB SPI NOR flash
- 2.4 GHz 3x3 802.11n Wi-Fi
- 4x 10/100 Mbps LAN and 1x 10/100 Mbps WAN
- 2x USB 2.0
- 11 LEDs, most controlled through a 74HC595 shift register
- Reset and Wi-Fi buttons
- 12 VDC, 1.5 A power supply

## Origin

This device was supported by the legacy ar71xx target before the OpenWrt 21.02 transition. This PR ports the board to ath79.

## MAC address layout

The base MAC is stored at offset `0xc` in the `product-info` partition as an ASCII record using hyphen-separated octets. The generic `mac-base` parser is extended for this format, so Ethernet and Wi-Fi receive their factory MAC addresses during driver probe without UCI overrides.

- LAN (`eth0`): base + 0
- WAN (`eth1`): base + 1
- Wi-Fi: base + 0

## Serial console

A UART header is present on the PCB. Production OEM U-Boot and firmware disable the serial console and use GPIOs 14 and 15 to control the two USB LEDs. The pinout and baud rate are not documented for production units; serial access therefore requires a UART-enabled U-Boot build.

### Vendor web interface

Upload the generated factory image through the normal firmware upgrade page.

### U-Boot TFTP recovery

1. Configure the host as `192.168.0.66/24` and start a TFTP server.
2. Rename the factory image to `WR942v1_recovery.bin` and place it in the TFTP root.
3. Hold Reset while powering on and release it when the WPS LED turns on.

The recovery mode is unavailable in beta OEM firmware.

### UART-enabled U-Boot

1. Interrupt autoboot with `tpl`.
2. Configure the router and TFTP server addresses.
3. Run:

```
tftp 0x81000000 openwrt-ath79-generic-tplink_tl-wr942n-v1-squashfs-sysupgrade.bin
erase 0x9f020000 +$filesize
cp.b 0x81000000 0x9f020000 $filesize
reset
```

## Tested

- Booted on TL-WR942N v1 hardware
- Factory MAC addresses on LAN, WAN and Wi-Fi
- LAN switch and WAN interface enumeration
- Both USB ports and their activity LEDs
- Open Wi-Fi AP on the LAN bridge
- Wi-Fi button generates rfkill events and toggles the radio
- WPS/Reset button reboots the device on a short press

### DIFF
--- a/target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
+++ b/target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
@@ -1,0 +1,284 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+#include "qca956x.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+
+/ {
+	compatible = "tplink,tl-wr942n-v1", "qca,qca9561";
+	model = "TP-Link TL-WR942N v1";
+
+	aliases {
+		led-boot = &status;
+		led-failsafe = &status;
+		led-running = &status;
+		led-upgrade = &status;
+		label-mac-device = &wmac;
+	};
+
+	led_spi {
+		compatible = "spi-gpio";
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		sck-gpios = <&gpio 18 GPIO_ACTIVE_HIGH>;	// 74HC595 SRCLK
+		mosi-gpios = <&gpio 17 GPIO_ACTIVE_HIGH>;	// 74HC595 SER
+		cs-gpios = <&gpio 20 GPIO_ACTIVE_LOW>;	// 74HC595 RCLK
+		num-chipselects = <1>;
+
+		led_gpio: led_gpio@0 {
+			compatible = "fairchild,74hc595";
+			reg = <0>;
+			gpio-controller;
+			#gpio-cells = <2>;
+			registers-number = <1>;
+			spi-max-frequency = <10000000>;
+		};
+	};
+
+	gpio-export {
+		compatible = "gpio-export";
+
+		gpio_shift_register_hub_reset {
+			gpio-export,name = "tp-link:power:usb";
+			gpio-export,output = <1>;
+			gpios = <&led_gpio 7 GPIO_ACTIVE_HIGH>;
+		};
+
+		gpio_shift_register_oe {
+			gpio-export,name = "tp-link:oe:sr";
+			gpio-export,output = <0>;
+			gpios = <&gpio 16 GPIO_ACTIVE_HIGH>;
+		};
+
+		gpio_shift_register_reset {
+			gpio-export,name = "tp-link:reset:sr";
+			gpio-export,output = <1>;
+			gpios = <&gpio 19 GPIO_ACTIVE_HIGH>;
+		};
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		status: status_led {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_STATUS;
+			gpios = <&gpio 22 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "heartbeat";
+		};
+
+		wifi {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WLAN;
+			gpios = <&led_gpio 6 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "phy0tpt";
+		};
+
+		lan1 {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_LAN;
+			function-enumerator = <1>;
+			gpios = <&led_gpio 3 GPIO_ACTIVE_LOW>;
+		};
+
+		lan2 {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_LAN;
+			function-enumerator = <2>;
+			gpios = <&led_gpio 2 GPIO_ACTIVE_LOW>;
+		};
+
+		lan3 {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_LAN;
+			function-enumerator = <3>;
+			gpios = <&led_gpio 1 GPIO_ACTIVE_LOW>;
+		};
+
+		lan4 {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_LAN;
+			function-enumerator = <4>;
+			gpios = <&led_gpio 0 GPIO_ACTIVE_LOW>;
+		};
+
+		wan_green {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WAN;
+			gpios = <&led_gpio 4 GPIO_ACTIVE_LOW>;
+		};
+
+		wan_amber {
+			color = <LED_COLOR_ID_AMBER>;
+			function = LED_FUNCTION_WAN;
+			gpios = <&led_gpio 5 GPIO_ACTIVE_LOW>;
+		};
+
+		wps {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_WPS;
+			gpios = <&gpio 21 GPIO_ACTIVE_LOW>;
+		};
+
+		usb1 {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_USB;
+			function-enumerator = <1>;
+			gpios = <&gpio 15 GPIO_ACTIVE_LOW>;
+		};
+
+		usb2 {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_USB;
+			function-enumerator = <2>;
+			gpios = <&gpio 14 GPIO_ACTIVE_LOW>;
+		};
+	};
+
+	keys {
+		compatible = "gpio-keys";
+
+		reset {
+			label = "Reset button";
+			linux,code = <KEY_RESTART>;
+			gpios = <&gpio 1 GPIO_ACTIVE_LOW>;
+			debounce-interval = <60>;
+		};
+
+		rfkill {
+			label = "RFKILL button";
+			linux,code = <KEY_RFKILL>;
+			gpios = <&gpio 2 GPIO_ACTIVE_LOW>;
+			debounce-interval = <60>;
+		};
+	};
+};
+
+&spi {
+	status = "okay";
+
+	flash@0 {
+		compatible = "jedec,spi-nor";
+		reg = <0>;
+		spi-max-frequency = <25000000>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "u-boot";
+				reg = <0x000000 0x020000>;
+				read-only;
+			};
+
+			partition@20000 {
+				compatible = "denx,uimage";
+				label = "firmware";
+				reg = <0x020000 0xe20000>;
+			};
+
+			product_info: partition@e40000 {
+				label = "product-info";
+				reg = <0xe40000 0x010000>;
+				read-only;
+
+				nvmem-layout {
+					compatible = "fixed-layout";
+					#address-cells = <1>;
+					#size-cells = <1>;
+
+					macaddr_product_info: macaddr@c {
+						compatible = "mac-base";
+						reg = <0xc 0x11>;
+						#nvmem-cell-cells = <1>;
+					};
+				};
+			};
+
+			partition@e50000 {
+				label = "partition-table";
+				reg = <0xe50000 0x010000>;
+				read-only;
+			};
+
+			partition@e60000 {
+				label = "oem-config";
+				reg = <0xe60000 0x040000>;
+				read-only;
+			};
+
+			partition@ea0000 {
+				label = "oem-vars";
+				reg = <0xea0000 0x150000>;
+				read-only;
+			};
+
+			art: partition@ff0000 {
+				label = "art";
+				reg = <0xff0000 0x010000>;
+				read-only;
+
+				nvmem-layout {
+					compatible = "fixed-layout";
+					#address-cells = <1>;
+					#size-cells = <1>;
+
+					cal_art_1000: calibration@1000 {
+						reg = <0x1000 0x440>;
+					};
+				};
+			};
+		};
+	};
+};
+
+&uart {
+	status = "okay";
+};
+
+&usb_phy0 {
+	status = "okay";
+};
+
+&usb0 {
+	status = "okay";
+};
+
+&usb_phy1 {
+	status = "okay";
+};
+
+&usb1 {
+	status = "okay";
+};
+
+&eth0 {
+	nvmem-cells = <&macaddr_product_info 1>;
+	nvmem-cell-names = "mac-address";
+	status = "okay";
+	phy-handle = <&swphy0>;
+
+	gmac-config {
+		device = <&gmac>;
+
+		switch-phy-addr-swap = <1>;
+		switch-phy-swap = <1>;
+	};
+};
+
+&eth1 {
+	nvmem-cells = <&macaddr_product_info 0>;
+	nvmem-cell-names = "mac-address";
+	status = "okay";
+};
+
+&wmac {
+	status = "okay";
+	nvmem-cells = <&macaddr_product_info 0>, <&cal_art_1000>;
+	nvmem-cell-names = "mac-address", "calibration";
+};
--- a/target/linux/ath79/generic/base-files/etc/board.d/02_network
+++ b/target/linux/ath79/generic/base-files/etc/board.d/02_network
@@ -554,7 +554,8 @@ ath79_setup_interfaces()
 		;;
 	tplink,tl-wr841hp-v2|\
 	tplink,tl-wr842n-v2|\
-	tplink,tl-wr941hp-v1)
+	tplink,tl-wr941hp-v1|\
+	tplink,tl-wr942n-v1)
 		ucidef_set_interface_wan "eth1"
 		ucidef_add_switch "switch0" \
 			"0@eth0" "1:lan:4" "2:lan:1" "3:lan:2" "4:lan:3"
--- a/target/linux/ath79/image/generic-tp-link.mk
+++ b/target/linux/ath79/image/generic-tp-link.mk
@@ -882,6 +882,18 @@ define Device/tplink_tl-wr941hp-v1
 endef
 TARGET_DEVICES += tplink_tl-wr941hp-v1
 
+define Device/tplink_tl-wr942n-v1
+  $(Device/tplink-safeloader-uimage)
+  SOC := qca9561
+  IMAGE_SIZE := 14464k
+  DEVICE_MODEL := TL-WR942N
+  DEVICE_VARIANT := v1
+  DEVICE_PACKAGES := kmod-usb2 kmod-usb-ledtrig-usbport
+  TPLINK_BOARD_ID := TLWR942NV1
+  SUPPORTED_DEVICES += tl-wr942n-v1
+endef
+TARGET_DEVICES += tplink_tl-wr942n-v1
+
 define Device/tplink_wbs210-v1
   $(Device/tplink-safeloader-okli)
   SOC := ar9344
--- a/target/linux/generic/pending-6.18/809-01-nvmem-core-generalize-mac-base-cells-handling.patch
+++ b/target/linux/generic/pending-6.18/809-01-nvmem-core-generalize-mac-base-cells-handling.patch
@@ -13,15 +13,19 @@ exposing a common helper.
 Such helper will change the nvmem_info_cell and apply the correct post
 process function to correctly parse the mac address.
 
+OpenWrt also normalizes hyphen-separated ASCII MAC addresses before
+parsing them, as used by some TP-Link devices.
+
 Since the API requires OF and is only related to layout, move the
 function in layouts.c to correctly handle non-OF NVMEM.
 
 Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
+Signed-off-by: Gleb Pesin <dormancygrace@gmail.com>
 ---
  drivers/nvmem/core.c           | 79 +--------------------------------
- drivers/nvmem/layouts.c        | 80 ++++++++++++++++++++++++++++++++++
+ drivers/nvmem/layouts.c        | 87 ++++++++++++++++++++++++++++++++++
  include/linux/nvmem-provider.h |  4 ++
- 3 files changed, 86 insertions(+), 77 deletions(-)
+ 3 files changed, 93 insertions(+), 77 deletions(-)
 
 --- a/drivers/nvmem/core.c
 +++ b/drivers/nvmem/core.c
@@ -142,7 +146,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  #include <linux/nvmem-consumer.h>
  #include <linux/nvmem-provider.h>
  #include <linux/of.h>
-@@ -21,6 +24,83 @@
+@@ -21,6 +24,90 @@
  #define to_nvmem_layout_device(_dev) \
  	container_of((_dev), struct nvmem_layout, dev)
  
@@ -162,11 +166,18 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
 +				     void *buf, size_t bytes)
 +{
 +	u8 mac[ETH_ALEN];
++	char macstr[3 * ETH_ALEN] = {};
++	int i;
 +
 +	if (WARN_ON(bytes != 3 * ETH_ALEN - 1))
 +		return -EINVAL;
 +
-+	if (!mac_pton(buf, mac))
++	memcpy(macstr, buf, sizeof(macstr) - 1);
++	for (i = 2; i < sizeof(macstr) - 1; i += 3)
++		if (macstr[i] == '-')
++			macstr[i] = ':';
++
++	if (!mac_pton(macstr, mac))
 +		return -EINVAL;
 +
 +	if (index)
--- a/target/linux/ath79/generic/base-files/etc/board.d/01_leds
+++ b/target/linux/ath79/generic/base-files/etc/board.d/01_leds
@@ -418,6 +418,15 @@ tplink,tl-wr1043n-v5)
 	ucidef_set_led_switch "lan3" "LAN3" "green:lan3" "switch0" "0x04"
 	ucidef_set_led_switch "lan4" "LAN4" "green:lan4" "switch0" "0x02"
 	;;
+tplink,tl-wr942n-v1)
+	ucidef_set_led_switch "lan1" "LAN1" "green:lan-1" "switch0" "0x04"
+	ucidef_set_led_switch "lan2" "LAN2" "green:lan-2" "switch0" "0x08"
+	ucidef_set_led_switch "lan3" "LAN3" "green:lan-3" "switch0" "0x10"
+	ucidef_set_led_switch "lan4" "LAN4" "green:lan-4" "switch0" "0x02"
+	ucidef_set_led_netdev "wan" "WAN" "green:wan" "eth1"
+	ucidef_set_led_usbport "usb_lower" "USB (lower)" "green:usb-1" "1-1-port2"
+	ucidef_set_led_usbport "usb_upper" "USB (upper)" "green:usb-2" "1-1-port1"
+	;;
 tplink,archer-c6-v2|\
 tplink,archer-c6-v2-us)
 	ucidef_set_led_switch "lan" "LAN" "green:lan" "switch0" "0x3c"


target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
```diff
@@ -0,0 +1,282 @@
...
+	status = "okay";
+
+	hub_port1: port@1 {
+		reg = <1>;
+		#trigger-source-cells = <0>;
+	};
+};
+
+&eth0 {
+	nvmem-cells = <&macaddr_product_info 0>;
```
openwrt-ai: The two MAC indices look swapped relative to the OEM/ar71xx assignment.

In ath79 `&eth0` is the standalone GMAC — here it carries `phy-handle = <&swphy0>` plus the `switch-phy-swap` gmac-config, i.e. it is the WAN port (it probes last and comes up as netdev `eth1`), while `&eth1` is the built-in-switch MAC (netdev `eth0`, LAN).

ar71xx assigned WAN = base + 1 and LAN = base + 0 for this exact board: [`ath79_init_mac(ath79_eth0_data.mac_addr, tmpmac, 1)` / `ath79_init_mac(ath79_eth1_data.mac_addr, tmpmac, 0)`](``https://github.com/openwrt/openwrt/blob/b7cc6616152a12b3f68c5f36e9ef052da5de889e/target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr942n-v1.c#L240-L252).`` The ath79 conversion of the sibling TL-WR941ND v5, which has the same topology, keeps that mapping: [`&eth0` index 1, `&eth1` index 0](https://github.com/openwrt/openwrt/blob/92474a7e743af30db5a6955fa3f47e056e2e4128/target/linux/ath79/dts/ar9341_tplink_tl-wr941nd-v5.dts#L99-L107).

As written, both Ethernet interfaces end up with the opposite address from stock.

```suggestion
	nvmem-cells = <&macaddr_product_info 1>;
```

The mirror change is needed on `&eth1` at line 273, which should use index `0`.


---

target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
```diff
@@ -0,0 +1,282 @@
...
+	};
+
+	led_spi {
+		compatible = "spi-gpio";
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		sck-gpios = <&gpio 18 GPIO_ACTIVE_HIGH>;	// 74HC595 SRCLK
+		mosi-gpios = <&gpio 17 GPIO_ACTIVE_HIGH>;	// 74HC595 SER
+		cs-gpios = <&gpio 20 GPIO_ACTIVE_HIGH>;	// 74HC595 RCLK
```
openwrt-ai: The 74HC595 latch line has to be active low. RCLK latches the shift register on its rising edge, which is the chip-select *deassert* edge; with `GPIO_ACTIVE_HIGH` the latch fires on assert instead, so the outputs end up one SPI transfer behind and the LED state lags.

Every 74HC595 `spi-gpio` node in ath79 uses active low here, e.g. [`cs-gpios = <&gpio 16 GPIO_ACTIVE_LOW>; // 74HC595 RCLK` in qca9563_tplink_archer-c7-v4.dts:27](https://github.com/openwrt/openwrt/blob/92474a7e743af30db5a6955fa3f47e056e2e4128/target/linux/ath79/dts/qca9563_tplink_archer-c7-v4.dts#L27), and the ar71xx board file left the CS polarity at the SPI default (active low) too.

```suggestion
		cs-gpios = <&gpio 20 GPIO_ACTIVE_LOW>;	// 74HC595 RCLK
```


---

target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
```diff
@@ -0,0 +1,282 @@
...
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		status: status_led {
+			label = "green:status";
+			gpios = <&gpio 22 GPIO_ACTIVE_LOW>;
+			linux,default-trigger = "heartbeat";
+		};
```
openwrt-ai, line 71: The `label = "<color>:<function>"` syntax is deprecated for new device support; use `color = <LED_COLOR_ID_*>` + `function = LED_FUNCTION_*` from `dt-bindings/leds/common.h`. All eleven LED nodes in this file (lines 67–122) need converting.

```suggestion
		status: status_led {
			color = <LED_COLOR_ID_GREEN>;
			function = LED_FUNCTION_STATUS;
			gpios = <&gpio 22 GPIO_ACTIVE_LOW>;
			linux,default-trigger = "heartbeat";
		};
```

Two follow-ups this pulls in:

* `#include <dt-bindings/leds/common.h>` has to be added next to the other includes at the top of the file (the sibling [qca9561_tplink_archer-c25-v1.dts](https://github.com/openwrt/openwrt/blob/92474a7e743af30db5a6955fa3f47e056e2e4128/target/linux/ath79/dts/qca9561_tplink_archer-c25-v1.dts#L7) does this).
* The enumerated LEDs (`lan1`..`lan4`, `usb1`, `usb2`) become `function = LED_FUNCTION_LAN; function-enumerator = <N>;`, which renames them to `green:lan-1` … in sysfs, so the names in `board.d/01_leds` (lines 422–428) have to be updated accordingly — see the existing [`ucidef_set_led_switch "lan1" "LAN1" "amber:lan-1" ...`](https://github.com/openwrt/openwrt/blob/92474a7e743af30db5a6955fa3f47e056e2e4128/target/linux/ath79/generic/base-files/etc/board.d/01_leds#L237) for the expected form.


---

target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
```diff
@@ -0,0 +1,282 @@
...
+
+		usb1 {
+			label = "green:usb1";
+			gpios = <&gpio 14 GPIO_ACTIVE_LOW>;
+		};
+
+		usb2 {
+			label = "green:usb2";
+			gpios = <&gpio 15 GPIO_ACTIVE_LOW>;
+		};
```
openwrt-ai, line 138: These two GPIOs are the other way round in the ar71xx board file: [`#define TL_WR942N_V1_GPIO_LED_USB2 14` / `#define TL_WR942N_V1_GPIO_LED_USB1 15`](``https://github.com/openwrt/openwrt/blob/b7cc6616152a12b3f68c5f36e9ef052da5de889e/target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr942n-v1.c#L41-L42),`` i.e. `green:usb1` was GPIO 15 and `green:usb2` was GPIO 14, while this DTS has usb1 on 14 and usb2 on 15.

The PR body says both USB ports and their activity LEDs were verified, so possibly the ar71xx naming was simply wrong relative to the case markings — but the `01_leds` hunk also maps `green:usb1` to hub port 2 and `green:usb2` to hub port 1, which is the same order ar71xx used. Was the swap intentional, or did the label/GPIO pairing get inverted here?


---

target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
```diff
@@ -0,0 +1,282 @@
...
+
+&usb0 {
+	#address-cells = <1>;
+	#size-cells = <0>;
+	status = "okay";
+
+	hub_port0: port@1 {
+		reg = <1>;
+		#trigger-source-cells = <0>;
+	};
```
openwrt-ai: `hub_port0` (and `hub_port1` at line 252) declare `#trigger-source-cells`, but nothing in this DTS references them — no LED node has a `trigger-sources = <&hub_port0>;` property. The USB LED triggers are instead configured from `board.d/01_leds` with the string form `1-1-port2` / `1-1-port1`, which addresses the ports of the internal hub on bus 1, not these root-hub ports.

So as it stands both `port@1` nodes (and the `#address-cells`/`#size-cells` they need) are dead DT. Either wire the LEDs to them via `trigger-sources` and drop the `ucidef_set_led_usbport` lines, or drop the port nodes and keep the `01_leds` entries.


---

target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
```diff
@@ -0,0 +1,282 @@
...
+			partition@0 {
+				label = "u-boot";
+				reg = <0x000000 0x020000>;
+				read-only;
+			};
+
+			partition@20000 {
+				compatible = "denx,uimage";
+				label = "firmware";
+				reg = <0x020000 0xE20000>;
```
openwrt-ai: nit: uppercase hex in `reg` values is inconsistent with the lowercase unit addresses in the node names right above/below (`partition@e40000`, `partition@ea0000`, `partition@ff0000`) and with the rest of `target/linux/ath79/dts/`.

```suggestion
				reg = <0x020000 0xe20000>;
```

Same for lines 171, 189, 195, 201 and 207.


---

target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
```diff
@@ -0,0 +1,282 @@
...
+			#gpio-cells = <2>;
+			registers-number = <1>;
+			lines-initial-states = /bits/ 8 <0xff>;
+			spi-max-frequency = <10000000>;
+		};
+	};
+
+	gpio-export {
+		compatible = "gpio-export";
+		#size-cells = <0>;
```
openwrt-ai: nit: `#size-cells` has no meaning on a `gpio-export` node — its children have no `reg`. None of the 54 `gpio-export` nodes currently in `target/linux/ath79/dts/` carry it (compare [qca9563_tplink_archer-c7-v4.dts:41-42](https://github.com/openwrt/openwrt/blob/92474a7e743af30db5a6955fa3f47e056e2e4128/target/linux/ath79/dts/qca9563_tplink_archer-c7-v4.dts#L41-L42), which this node is otherwise modelled on).

```suggestion
		compatible = "gpio-export";
```


---

target/linux/ath79/generic/base-files/etc/board.d/02_network
```diff
@@ -619,6 +619,11 @@ ath79_setup_interfaces()
 		ucidef_add_switch "switch0" \
 			"0@eth0" "3:lan:1" "4:lan:2"
 		;;
+	tplink,tl-wr942n-v1)
+		ucidef_set_interfaces_lan_wan "eth0.1" "eth1"
```
openwrt-ai: nit: the explicit `lan` device is redundant here — [`_ucidef_finish_switch_roles()`](https://github.com/openwrt/openwrt/blob/92474a7e743af30db5a6955fa3f47e056e2e4128/package/base-files/files/lib/functions/uci-defaults.sh#L205-L233) already derives `lan` = `eth0.1` from the `0@eth0` CPU port below. Every ath79 board with this exact switch layout sets only the WAN side, e.g. [`tplink,tl-wr842n-v2` / `tplink,tl-wr941hp-v1` at 02_network:556-561](``https://github.com/openwrt/openwrt/blob/92474a7e743af30db5a6955fa3f47e056e2e4128/target/linux/ath79/generic/base-files/etc/board.d/02_network#L556-L561).``

```suggestion
		ucidef_set_interface_wan "eth1"
```


---

target/linux/ath79/image/generic-tp-link.mk
```diff
@@ -748,6 +748,18 @@ define Device/tplink_tl-wr1043n-v5
...
 
+define Device/tplink_tl-wr942n-v1
+  $(Device/tplink-safeloader-uimage)
+  SOC := qca9561
+  IMAGE_SIZE := 14464k
+  DEVICE_MODEL := TL-WR942N
+  DEVICE_VARIANT := v1
+  DEVICE_PACKAGES := kmod-usb2 kmod-usb-ledtrig-usbport
+  TPLINK_BOARD_ID := TLWR942NV1
+  SUPPORTED_DEVICES += tl-wr942n tl-wr942n-v1
```
openwrt-ai: `tl-wr942n` was never a board name in ar71xx, so it can only weaken the sysupgrade compatibility check without ever matching anything. The ar71xx board-detect entry was [`*"TL-WR942N v1") name="tl-wr942n-v1"`](https://github.com/openwrt/openwrt/blob/b7cc6616152a12b3f68c5f36e9ef052da5de889e/target/linux/ar71xx/base-files/lib/ar71xx.sh#L1169-L1171) — one name only, which is also what every other converted device in this file does (`SUPPORTED_DEVICES += archer-c25-v1`, `+= tl-wr1043n-v5`, …).

```suggestion
  SUPPORTED_DEVICES += tl-wr942n-v1
```


---

target/linux/ath79/image/generic-tp-link.mk
```diff
@@ -748,6 +748,18 @@ define Device/tplink_tl-wr1043n-v5
 endef
 TARGET_DEVICES += tplink_tl-wr1043n-v5
 
+define Device/tplink_tl-wr942n-v1
```
openwrt-ai: nit: this file is sorted by device name and the block is inserted out of order — between `tplink_tl-wr1043n-v5` and `tplink_tl-wr1045nd-v2`. It belongs after [`tplink_tl-wr941hp-v1`](https://github.com/openwrt/openwrt/blob/92474a7e743af30db5a6955fa3f47e056e2e4128/target/linux/ath79/image/generic-tp-link.mk#L887-L895) and before `tplink_wbs210-v1`.


---

target/linux/generic/pending-6.18/809-04-nvmem-layouts-accept-hyphen-separated-mac-base.patch
openwrt-ai, line 1: The `810-` prefix is already taken in this directory by [`810-pci_disable_common_quirks.patch`](``https://github.com/openwrt/openwrt/blob/92474a7e743af30db5a6955fa3f47e056e2e4128/target/linux/generic/pending-6.18/810-pci_disable_common_quirks.patch),`` so quilt's apply order for the two becomes name-dependent and the numbering no longer conveys anything.

This patch also only makes sense on top of the existing nvmem `mac-base` series (`804-nvmem-core-support-mac-base-fixed-layout-cells.patch`, `809-01`…`809-03`) — it edits `nvmem_mac_base_ascii_read()` as introduced by `809-01`. Please renumber it into that group, e.g. `809-04-nvmem-layouts-accept-hyphen-separated-mac-base.patch`.


---

openwrt-ai (commented):
## Commit checks

- 92474a7e "ath79: add support for TP-Link TL-WR942N v1" — a new-device commit message needs three sections; two are missing and one is thin:
  - **Flash instructions** — absent entirely. The PR body mentions "the stock TP-Link web interface" and "the boot loader's standard TFTP recovery procedure", but neither is spelled out and neither is in the commit message. The [original ar71xx commit b7cc661](https://github.com/openwrt/openwrt/commit/b7cc6616152a12b3f68c5f36e9ef052da5de889e) has verbatim procedures for the OEM GUI, the U-Boot TFTP recovery (static IP `192.168.0.66/24`, image renamed to `WR942v1_recovery.bin`, hold reset until the WPS LED lights) and the serial/U-Boot path — those can largely be carried over.
  - **MAC address layout** — only "Factory MAC addresses are supplied through the product-info NVMEM layout". Please give one line per interface with the source, e.g. `LAN (eth0): product-info 0xc, base+N`, `WAN (eth1): …`, `WLAN: …`, using `xx:xx:xx` for the unit-specific bytes. Note the inline comment on the DTS: the base/offset assignment appears inverted relative to stock, so this section is worth getting right explicitly.
  - **Hardware specification** — covers SoC, RAM, flash, WiFi, Ethernet and USB, but not the LED/button count, the UART header (location, pinout, baud) or power input. The ar71xx commit's note that GPIO 14/15 are shared with UART TX/RX on some U-Boot versions is also worth preserving, since this DTS enables `&uart` *and* drives the USB LEDs from those pins.


---

target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
dormancygrace: Fixed. The DTS now follows the OEM/ar71xx layout: the standalone WAN GMAC uses base + 1 and the LAN switch MAC uses base + 0. The commit message and PR body now document the resulting netdev layout explicitly.

target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
dormancygrace: Fixed. RCLK is now declared GPIO_ACTIVE_LOW, matching the SPI chip-select semantics and the existing ath79 74HC595 users.

target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
dormancygrace: Fixed all eleven nodes. The DTS now includes leds/common.h and uses color/function/function-enumerator properties; the board.d names were updated to green:lan-1 through green:lan-4 and green:usb-1/2.

target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
dormancygrace: The swap was not intentional. Restored the ar71xx mapping: USB1 is GPIO 15 and USB2 is GPIO 14, while retaining the established internal-hub port associations in board.d.

target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
dormancygrace: Fixed using the second option: removed both unused root-hub port nodes and their address/size cells, while keeping the tested board.d USB-port trigger mappings.

target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
dormancygrace: Fixed all six values; the partition reg constants now use lowercase hexadecimal consistently.

target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
dormancygrace: Fixed. Removed the meaningless #size-cells property from gpio-export.

target/linux/ath79/generic/base-files/etc/board.d/02_network
dormancygrace: Fixed. The board now sets only the WAN interface explicitly; the LAN device is derived from the switch CPU-port role.

target/linux/ath79/image/generic-tp-link.mk
dormancygrace: Fixed. Removed the nonexistent tl-wr942n alias and retained only the legacy ar71xx board name tl-wr942n-v1.

target/linux/ath79/image/generic-tp-link.mk
dormancygrace: Fixed. Moved the profile after TL-WR941HP v1 and before WBS210 v1.

target/linux/generic/pending-6.18/809-04-nvmem-layouts-accept-hyphen-separated-mac-base.patch
dormancygrace: Fixed. Renamed the patch to 809-04-nvmem-layouts-accept-hyphen-separated-mac-base.patch so it follows the mac-base series and no longer collides with 810.

dormancygrace:
Addressed the commit-message review in the rewritten device commit. It now includes the complete hardware specification, explicit LAN/WAN/Wi-Fi MAC layout, vendor GUI installation, TFTP recovery, and UART-enabled U-Boot flashing instructions. It also documents that production OEM U-Boot disables serial and reuses GPIOs 14/15 for the USB LEDs; the production header pinout and baud rate are not documented.

target/linux/generic/pending-6.18/809-04-nvmem-layouts-accept-hyphen-separated-mac-base.patch
```diff
@@ -0,0 +1,36 @@
...
+ {
+ 	u8 mac[ETH_ALEN];
++	char macstr[3 * ETH_ALEN] = {};
++	int i;
+ 
+ 	if (WARN_ON(bytes != 3 * ETH_ALEN - 1))
+ 		return -EINVAL;
+ 
+-	if (!mac_pton(buf, mac))
++	memcpy(macstr, buf, sizeof(macstr) - 1);
```
neheb: this is interesting and I wonder if it makes more sense to edit the original target/linux/generic/pending-6.18/809-01-nvmem-core-generalize-mac-base-cells-handling.patch . I don't see any upstream activity regarding it.

dormancygrace: Good point. I folded the hyphen-separated ASCII MAC handling into the existing `809-01` patch and removed the standalone `809-04` patch. The rewritten series is now at `eef1dfbe`.

target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
```diff
@@ -0,0 +1,285 @@
...
+		cs-gpios = <&gpio 20 GPIO_ACTIVE_LOW>;	// 74HC595 RCLK
+		num-chipselects = <1>;
+
+		led_gpio: led_gpio@0 {
+			compatible = "fairchild,74hc595";
+			reg = <0>;
+			gpio-controller;
+			#gpio-cells = <2>;
+			registers-number = <1>;
+			lines-initial-states = /bits/ 8 <0xff>;
```
openwrt-ai: `lines-initial-states` is not part of the `fairchild,74hc595` binding — it belongs to `nxp,pcf857x` ([nxp,pcf8575.yaml:55](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/gpio/nxp,pcf8575.yaml#L55)), so this line is silently dropped.

[`gen_74x164_probe()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/gpio/gpio-74x164.c#L111-L131) reads only `registers-number` and allocates the output buffer with [`devm_kzalloc()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/gpio/gpio-74x164.c#L131), then writes it out — so the shift register is latched as all-zeros at probe, not `0xff`. With the LEDs wired `GPIO_ACTIVE_LOW`, that means every 74HC595-driven LED lights up and bit 7 (the USB hub reset exported as `tp-link:power:usb`) is driven low until `gpio-export` and `gpio-leds` probe and re-drive them.

The ar71xx board got `0xff` through `gen_74x164_chip_platform_data.init_data`, which has no DT equivalent; the closest in-tree analogue, [qca9563_tplink_archer-c7-v4.dts](https://github.com/openwrt/openwrt/blob/eef1dfbe7f3fc09cfa6520d98324293e96fd402b/target/linux/ath79/dts/qca9563_tplink_archer-c7-v4.dts#L31-L37), just omits it. No other DTS in `target/` uses this property. Also note `fairchild,74hc595.yaml` is `unevaluatedProperties: false`, so it would warn under `dtbs_check`.

```suggestion
			registers-number = <1>;
```


---

target/linux/ath79/generic/base-files/etc/board.d/02_network
```diff
@@ -619,6 +619,11 @@ ath79_setup_interfaces()
 		ucidef_add_switch "switch0" \
 			"0@eth0" "3:lan:1" "4:lan:2"
 		;;
+	tplink,tl-wr942n-v1)
+		ucidef_set_interface_wan "eth1"
+		ucidef_add_switch "switch0" \
+			"0@eth0" "1:lan:4" "2:lan:1" "3:lan:2" "4:lan:3"
+		;;
```
openwrt-ai: This new case is byte-identical to the existing [`tplink,tl-wr841hp-v2|tplink,tl-wr842n-v2|tplink,tl-wr941hp-v1)` group at 02_network:555-561](``https://github.com/openwrt/openwrt/blob/eef1dfbe7f3fc09cfa6520d98324293e96fd402b/target/linux/ath79/generic/base-files/etc/board.d/02_network#L555-L561``) — same `ucidef_set_interface_wan "eth1"`, same `"0@eth0" "1:lan:4" "2:lan:1" "3:lan:2" "4:lan:3"`. The file's convention is to add the compatible to the matching group rather than append a duplicate; ar71xx did exactly that for this board ([`tl-wr941nd-v5|\` + `tl-wr942n-v1)`](https://github.com/openwrt/openwrt/blob/b7cc6616152a12b3f68c5f36e9ef052da5de889e/target/linux/ar71xx/base-files/etc/board.d/02_network#L411-L418)).

Please add `tplink,tl-wr942n-v1|\` to the group at line 555 and drop these five lines — as written the block also lands at the very end of the `case`, right before the `*)` default, away from the other TP-Link entries.


---

target/linux/generic/pending-6.18/809-01-nvmem-core-generalize-mac-base-cells-handling.patch
```diff
@@ -162,11 +162,18 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
 +				     void *buf, size_t bytes)
 +{
 +	u8 mac[ETH_ALEN];
++	char macstr[3 * ETH_ALEN] = {};
++	int i;
```
openwrt-ai, line 170: nit: the code change itself is fine (17-byte cell, separators at 2/5/8/11/14, `macstr` NUL-terminated by the `= {}` initialiser, and `mac_pton()` only accepts `:` so the normalisation is required) — but the patch's own header was not updated along with it.

`Subject: [PATCH 1/2] nvmem: core: generalize "mac-base" cells handling` and the body still describe only the upstream refactor, and the diffstat at the top is now stale: `drivers/nvmem/layouts.c | 80 ++++...` and `3 files changed, 86 insertions(+), 77 deletions(-)` should be 87 and 93 after the 7 lines added here.

Since this is an OpenWrt-local addition folded into a patch carrying Christian Marangi's `Signed-off-by:` (per @neheb's suggestion), please note it in the patch description so whoever rebases this against the upstream submission can tell which hunks are ours.


---

openwrt-ai (commented):
Re-reviewed the 2 commits at `eef1dfbe` (previous review was against `92474a7e`). All eleven points from the last round are addressed; the commit message now carries hardware spec, flash instructions and MAC layout.

Verified against the ar71xx board file ([mach-tl-wr942n-v1.c](https://github.com/openwrt/openwrt/blob/b7cc6616152a12b3f68c5f36e9ef052da5de889e/target/linux/ar71xx/files/arch/mips/ath79/mach-tl-wr942n-v1.c)): the 74HC595 bit assignments (LAN4/3/2/1 = 0/1/2/3, WAN green/amber = 4/5, WLAN = 6, hub reset = 7), the USB LED GPIOs (usb1 = 15, usb2 = 14), the `eth0`/`eth1` MAC indices, the exported GPIO initial levels and the full flash layout (u-boot 128k / firmware 14464k @ 0x20000 / product-info / partition-table / oem-config / oem-vars / art = 16 MiB) all match. `macaddr@c` also lines up with the ar71xx NVRAM window (`0x1fe40008` + `strlen("MAC:")`). CI is green on both workflow runs for this head.

Three remaining items inline, one of which is a functional no-op in the DTS.

---

target/linux/ath79/dts/qca9561_tplink_tl-wr942n-v1.dts
dormancygrace: Removed the unsupported `lines-initial-states` property from the 74HC595 node.

target/linux/ath79/generic/base-files/etc/board.d/02_network
dormancygrace: Merged TL-WR942N v1 into the existing matching `02_network` case instead of keeping a duplicate block.

target/linux/generic/pending-6.18/809-01-nvmem-core-generalize-mac-base-cells-handling.patch
dormancygrace: Updated the downstream patch description and diffstat to document the OpenWrt hyphen-separated MAC extension.

openwrt-ai (commented):

Re-reviewed at `66e01638` (previous review was against `eef1dfbe`). The delta is three hunks, all addressing the last round; each checks out:

- `qca9561_tplink_tl-wr942n-v1.dts` — `lines-initial-states` dropped from the `fairchild,74hc595` node, leaving `registers-number` as the only vendor property, matching every other 74HC595 user in `target/linux/ath79/dts/`.
- `board.d/02_network` — `tplink,tl-wr942n-v1` folded into the existing `tl-wr841hp-v2|tl-wr842n-v2|tl-wr941hp-v1` group and the duplicate trailing block removed; the group's `ucidef_set_interface_wan "eth1"` / `"0@eth0" "1:lan:4" "2:lan:1" "3:lan:2" "4:lan:3"` is byte-identical to what was removed, and the new compatible sorts correctly within the group.
- `809-01-nvmem-core-generalize-mac-base-cells-handling.patch` — description and attribution now note the OpenWrt-local addition, and the metadata is consistent: the layouts.c hunk header went `@@ -21,6 +24,83 @@` → `@@ -21,6 +24,90 @@` and the diffstat 80 → 87 / 86 → 93, which matches the +8/-1 net of the hyphen-normalisation block. I re-validated every hunk header in the patch against its body; all line counts agree.

The normalisation itself is still correct against 6.18: [`mac_pton()` in lib/net_utils.c](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/lib/net_utils.c#L8-L26) rejects any separator other than `:`, the 17-byte `memcpy` into an 18-byte zeroed `macstr` keeps it NUL-terminated, and the `i = 2; i += 3` loop hits exactly the five separator positions.

Both workflow runs (Build Kernel, Build all core packages) are green on this head.

---

PR #24776, opened 2026-08-18
ramips: add Xiaomi Mi Router TR60x series

kokutoukiritsugu:
Add support for Xiaomi Mi Router TR60x series (TR606/TR608/TR609). The TR60x series shares the exact same hardware layout as the CR6606/CR6608/CR6609 series.

PR #24779, opened 2026-08-18
ramips: add unified Xiaomi Mi Router CR660x-TR60x profile

kokutoukiritsugu:
Add a unified "Xiaomi Mi Router CR660x-TR60x" device profile to provide a generic firmware image compatible across both CR660x and TR60x variants.

The CR660x (CR6606/CR6608/CR6609) and TR60x (TR606/TR608/TR609) series share identical hardware specs. Existing standalone profiles remain untouched.

### DIFF
--- a/package/boot/uboot-tools/uboot-envtools/files/ramips
+++ b/package/boot/uboot-tools/uboot-envtools/files/ramips
@@ -94,6 +94,7 @@ wifire,s1500-nbn)
 	ubootenv_add_mtd "u-boot-env" "0x0" "0x1000" "0x20000"
 	;;
 bolt,arion|\
+xiaomi,mi-router-cr660x-tr60x|\
 xiaomi,mi-router-cr6606|\
 xiaomi,mi-router-cr6608|\
 xiaomi,mi-router-cr6609)
--- a/target/linux/ramips/dts/mt7621_xiaomi_mi-router-cr660x-tr60x.dts
+++ b/target/linux/ramips/dts/mt7621_xiaomi_mi-router-cr660x-tr60x.dts
@@ -1,0 +1,8 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+#include "mt7621_xiaomi_mi-router-cr660x.dtsi"
+
+/ {
+	compatible = "xiaomi,mi-router-cr660x-tr60x", "mediatek,mt7621-soc";
+	model = "Xiaomi Mi Router CR660x-TR60x";
+};
--- a/target/linux/ramips/image/mt7621.mk
+++ b/target/linux/ramips/image/mt7621.mk
@@ -3622,6 +3622,12 @@ define Device/xiaomi_mi-router-cr660x
   DEVICE_PACKAGES += kmod-mt7915-firmware
 endef
 
+define Device/xiaomi_mi-router-cr660x-tr60x
+  $(Device/xiaomi_mi-router-cr660x)
+  DEVICE_MODEL := Mi Router CR660x-TR60x
+endef
+TARGET_DEVICES += xiaomi_mi-router-cr660x-tr60x
+
 define Device/xiaomi_mi-router-cr6606
   $(Device/xiaomi_mi-router-cr660x)
   DEVICE_MODEL := Mi Router CR6606
--- a/target/linux/ramips/mt7621/base-files/etc/board.d/01_leds
+++ b/target/linux/ramips/mt7621/base-files/etc/board.d/01_leds
@@ -255,6 +255,7 @@ xiaomi,mi-router-4a-gigabit-v2)
 xiaomi,mi-router-ac2100)
 	ucidef_set_led_netdev "wan-blue" "WAN (blue)" "blue:wan" "wan"
 	;;
+xiaomi,mi-router-cr660x-tr60x|\
 xiaomi,mi-router-cr6606|\
 xiaomi,mi-router-cr6608|\
 xiaomi,mi-router-cr6609)
--- a/target/linux/ramips/mt7621/base-files/etc/board.d/02_network
+++ b/target/linux/ramips/mt7621/base-files/etc/board.d/02_network
@@ -30,6 +30,7 @@ ramips_setup_interfaces()
 	tplink,mr600-v2-eu|\
 	xiaomi,mi-router-3-pro|\
 	xiaomi,mi-router-ac2100|\
+	xiaomi,mi-router-cr660x-tr60x|\
 	xiaomi,mi-router-cr6606|\
 	xiaomi,mi-router-cr6608|\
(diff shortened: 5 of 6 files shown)


kokutoukiritsugu:
build and flash test works

image

```
root@462359:~# ubus call system board
{
        "kernel": "6.18.44",
        "hostname": "462359",
        "system": "MediaTek MT7621 ver:1 eco:3",
        "model": "Xiaomi Mi Router CR660x-TR60x",
        "board_name": "xiaomi,mi-router-cr660x-tr60x",
        "rootfs_type": "squashfs",
        "release": {
                "distribution": "OpenWrt",
                "version": "SNAPSHOT",
                "firmware_url": "https://downloads.openwrt.org/",
                "revision": "r35846-db5ab5d0ec",
                "target": "ramips/mt7621",
                "description": "OpenWrt SNAPSHOT r35846-db5ab5d0ec",
                "builddate": "1787040747"
        }
}
```

target/linux/ramips/dts/mt7621_xiaomi_mi-router-cr660x-tr60x.dts
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+
+#include "mt7621_xiaomi_mi-router-cr660x.dtsi"
+
+/ {
+	compatible = "xiaomi,mi-router-cr660x-tr60x", "mediatek,mt7621-soc";
```
openwrt-ai, line 6: This introduces a compatible string that names no actual device — no board reports `xiaomi,mi-router-cr660x-tr60x`, it is a build-time umbrella. The in-tree way to publish one image under several marketing names is `DEVICE_ALT<N>_VENDOR` / `DEVICE_ALT<N>_MODEL` on the existing profile, plus `SUPPORTED_DEVICES` for the sysupgrade aliases — see [`Device/asus_rt-ax54` with three ALT names at mt7621.mk:504-513](https://github.com/openwrt/openwrt/blob/23a42990c1f8978b2201517c2ff6208ce78820bf/target/linux/ramips/image/mt7621.mk#L504-L513) and [`SUPPORTED_DEVICES += cudy,wr1300 R10` at mt7621.mk:827](https://github.com/openwrt/openwrt/blob/23a42990c1f8978b2201517c2ff6208ce78820bf/target/linux/ramips/image/mt7621.mk#L827).

As written the tree gains a fourth DTB and a fourth ~128 MB NAND image that are byte-for-byte identical to the CR6606/CR6608/CR6609 ones apart from two strings, while the three per-model profiles stay. If TR606/TR608/TR609 really are the same hardware, adding `DEVICE_ALT0_MODEL := TR606` (and ALT1/ALT2) to one of the existing CR660x profiles gets the same "one image, findable under every name" result on the firmware selector without a synthetic board name.

Also note the profile/file name `xiaomi_mi-router-cr660x-tr60x` is one hyphen away from the shared template `Device/xiaomi_mi-router-cr660x`, which makes the two easy to confuse when reading `mt7621.mk`.

---

target/linux/ramips/image/mt7621.mk
```diff
@@ -3622,6 +3622,12 @@ define Device/xiaomi_mi-router-cr660x
   DEVICE_PACKAGES += kmod-mt7915-firmware
 endef
 
+define Device/xiaomi_mi-router-cr660x-tr60x
+  $(Device/xiaomi_mi-router-cr660x)
+  DEVICE_MODEL := Mi Router CR660x-TR60x
+endef
+TARGET_DEVICES += xiaomi_mi-router-cr660x-tr60x
```
openwrt-ai, line 3629: This image is not actually interchangeable with the existing ones. `SUPPORTED_DEVICES` defaults to the profile name with `_` turned into `,` ([image.mk:551](https://github.com/openwrt/openwrt/blob/23a42990c1f8978b2201517c2ff6208ce78820bf/include/image.mk#L551)), so this image's metadata lists only `xiaomi,mi-router-cr660x-tr60x`. Running sysupgrade with it on a unit currently on the CR6606/CR6608/CR6609 image fails the compat check ("Device xiaomi,mi-router-cr6606 not supported by this image") and needs `-F`; the reverse direction breaks the same way once a unit is on this image, making it effectively a one-way trip.

If the unified profile stays, it should claim the model compatibles it is meant to replace:

```suggestion
define Device/xiaomi_mi-router-cr660x-tr60x
  $(Device/xiaomi_mi-router-cr660x)
  DEVICE_MODEL := Mi Router CR660x-TR60x
  SUPPORTED_DEVICES += xiaomi,mi-router-cr6606 xiaomi,mi-router-cr6608 \
	xiaomi,mi-router-cr6609
endef
TARGET_DEVICES += xiaomi_mi-router-cr660x-tr60x
```

---

package/boot/uboot-tools/uboot-envtools/files/ramips
```diff
@@ -94,6 +94,7 @@ wifire,s1500-nbn)
 	ubootenv_add_mtd "u-boot-env" "0x0" "0x1000" "0x20000"
 	;;
 bolt,arion|\
+xiaomi,mi-router-cr660x-tr60x|\
```
openwrt-ai, line 97: nit: these case lists are kept sorted, and `cr660x-tr60x` sorts after `cr6609` (`6` < `x`), so the new entry belongs on the line below `xiaomi,mi-router-cr6609` (which then loses its `)` to the new last entry) rather than above `xiaomi,mi-router-cr6606`.

Same ordering applies to the other four insertions in this PR: `target/linux/ramips/mt7621/base-files/etc/board.d/01_leds:258`, `.../02_network:33`, `.../lib/upgrade/platform.sh:156`, and the new `define Device/` block in `target/linux/ramips/image/mt7621.mk`, which sits before `xiaomi_mi-router-cr6606` instead of after `xiaomi_mi-router-cr6609`.

---

openwrt-ai (commented):
## Commit checks

- 23a42990 "ramips: add unified Xiaomi Mi Router CR660x-TR60x profile" — the commit adds a new `.dts` plus a new `define Device/` block, and TR606/TR608/TR609 are not supported by any existing in-tree profile, so this is the commit that introduces support for them. It carries none of the three sections the project asks for on new-device commits: hardware specification, flash instructions, and MAC address layout. At minimum, what is the install path for someone holding a TR608 (which image, how to get the device into a flashable state)? Related: "share identical hardware specs" is asserted without supporting evidence (SoC/RAM/flash chip, WiFi chip, partition layout) — the verification in the PR comment appears to be a build + flash on a CR660x unit, not on TR60x hardware. If the claim rests on an OEM source or a teardown, citing it in the commit message would make the unification reviewable.

---

kokutoukiritsugu:
#24776

PR #24786, opened 2026-08-18
octeon: add support for Ubiquiti UniFi Security Gateway Pro (USG-Pro-4)

N1kO23:
The Ubiquiti UniFi Security Gateway Pro 4 (USG-Pro-4) is a 1U rackmount
enterprise router powered by the Cavium Octeon CN6120 SoC. Previously
partially supported by the EdgeRouter Pro target.

## Specification:
- SoC: Cavium Octeon CN6120 (2 cores @ 1.0 GHz)
- RAM: 2 GB DDR3
- Flash: 8 MB NOR flash (U-Boot + eeprom) + 4 GB eMMC (/dev/mmcblk0)
- Ethernet: 4x 10/100/1000 Mbps
  - eth0 (LAN1): Broadcom PHY on SMI0
  - eth1 (LAN2): Broadcom PHY on SMI0
  - eth2 (WAN1): Broadcom PHY on SMI1 (RJ45 / SFP combo)
  - eth3 (WAN2): Broadcom PHY on SMI1 (RJ45 / SFP combo)
- Serial: RJ45 console, 115200 8N1 (standard RJ45-to-DB9 rollover cable)
- LEDs: 2x GPIO LEDs (white / blue logo)
- Buttons: 1x reset (GPIO 0)
- Power: 110-240 VAC input, internal AC/DC adapter, 60 W (24 V, 2.5 A)
  output, ~40 W max consumption, single (non-redundant) supply

### MAC address layout:
U-Boot loads a per-unit base MAC address into the Cavium boot descriptor
handed to the kernel. octeon_fill_mac_addresses() in
arch/mips/cavium-octeon/octeon-platform.c walks the pip interface's
ethernet@N child nodes in ascending order and, for each node whose
local-mac-address property is present, is exactly 6 bytes, and is not
already a valid address, patches in base+n and advances the counter; a
node with no such placeholder property is left untouched entirely.
octeon_3xxx.dtsi already carries a zero placeholder on ethernet@0-2, so
eth0/eth1/eth2 receive base+0/base+1/base+2 automatically. ethernet@3
(eth3/WAN2) is a new port this DTS adds and had no placeholder, so it
silently fell back to a random locally-administered address on every
boot; this DTS gives it the same zero placeholder so it receives base+3
from the same per-unit sequence as the other three ports.

## Installation via U-Boot:
1. Transfer openwrt-octeon-generic-ubnt_unifi-usg-pro-4-initramfs-kernel.bin
   to a USB drive
2. Intercept U-Boot on the serial console and load the image:
   - usb start
   - usb part 0
   - fatload usb 0:1 $loadaddr openwrt-octeon-generic-ubnt_unifi-usg-pro-4-initramfs-kernel.bin
3. Boot the initramfs image: bootoctlinux 0 numcores=2 endbootargs mem=0
4. Once booted into OpenWrt, perform a standard sysupgrade to install to
   the internal eMMC flash (/dev/mmcblk0)

Signed-off-by: Niko Huuskonen <niko.huuskonen.00@gmail.com>

### DIFF
--- a/target/linux/octeon/base-files/etc/board.d/01_network
+++ b/target/linux/octeon/base-files/etc/board.d/01_network
@@ -21,6 +21,9 @@ ubnt,edgerouter-6p)
 cisco,vedge1000)
 	ucidef_set_interfaces_lan_wan "mgmt0" "lan0"
 	;;
+ubnt,usg-pro-4)
+  ucidef_set_interfaces_lan_wan "eth0 eth1" "eth2 eth3"
+  ;;
 *)
 	ucidef_set_interfaces_lan_wan "eth0" "eth1"
 	;;
--- a/target/linux/octeon/base-files/etc/board.d/01_network
+++ b/target/linux/octeon/base-files/etc/board.d/01_network
@@ -22,8 +22,8 @@ cisco,vedge1000)
 	ucidef_set_interfaces_lan_wan "mgmt0" "lan0"
 	;;
 ubnt,usg-pro-4)
-  ucidef_set_interfaces_lan_wan "eth0 eth1" "eth2 eth3"
-  ;;
+	ucidef_set_interfaces_lan_wan "eth0 eth1" "eth2 eth3"
+	;;
 *)
 	ucidef_set_interfaces_lan_wan "eth0" "eth1"
 	;;
--- a/target/linux/octeon/base-files/etc/board.d/01_network
+++ b/target/linux/octeon/base-files/etc/board.d/01_network
@@ -21,6 +21,9 @@ ubnt,edgerouter-6p)
 cisco,vedge1000)
 	ucidef_set_interfaces_lan_wan "mgmt0" "lan0"
 	;;
+ubnt,usg-pro-4)
+	ucidef_set_interfaces_lan_wan "eth0 eth1" "eth2 eth3"
+	;;
 *)
 	ucidef_set_interfaces_lan_wan "eth0" "eth1"
 	;;
--- a/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
+++ b/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
@@ -186,8 +186,8 @@
 			#size-cells = <0>;
 			interrupts = <1 19 0 0x3f>;
 			
-			mmc_slot0:mmc-slot@0 {
-				compatible = "cavium,octeon-6130-mmc-slot", "mmc-spi-slot";
+			mmc_slot0: mmc-slot@0 {
+				compatible = "cavium,octeon-6130-mmc-slot";
 				reg = <0>;
 				spi-max-frequency = <26000000>;
 				voltage-ranges = <3300 3300>;
--- a/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
+++ b/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
@@ -125,22 +125,14 @@
 			#size-cells = <0>;
 			interrupts = <1 19 0 0x3f>;
 
-			mmc_slot0:mmc-slot@0 {
-				compatible = "cavium,octeon-6130-mmc-slot", "mmc-spi-slot";
+			mmc_slot0: mmc-slot@0 {
+				compatible = "cavium,octeon-6130-mmc-slot";
 				reg = <0>;
 				spi-max-frequency = <26000000>;
 				voltage-ranges = <3300 3300>;
 				cavium,bus-max-width = <8>;
 			};
 		};
-
-		/* Board-specific USB controller. Note this coexists with
-		 * the dtsi's "usbn" node (different name/unit-address, so
-		 * it isn't merged/replaced by this). If usbn is not present
-		 * on this SoC variant, consider deleting it explicitly with
-		 * /delete-node/ &usbn; to avoid a phantom, non-functional
-		 * node in the compiled tree.
-		 */
 
 		uctl: uctl@118006f000000 {
 			compatible = "cavium,octeon-6335-uctl";
@@ -176,14 +168,14 @@
 			 * trimmed to just the differing lines.
 			 */
 
-			ranges = <0 0  0x00 0x1f400000  0xc00000>,
-			<1 0  0x00 0x10000     0x30000000>,
-			<2 0  0x00 0x1d040000  0x10000>,
-			<3 0  0x00 0x1d050000  0x10000>,
-			<4 0  0x00 0x1d020000  0x10000>,
-			<5 0  0x00 0x10000     0x40000000>,
-			<6 0  0x00 0x10000     0x50000000>,
-			<7 0  0x00 0x10000     0x90000000>;
+			ranges = <0 0 0x00 0x1f400000 0xc00000>,
+            	 <1 0 0x00 0x10000 0x30000000>,
+            	 <2 0 0x00 0x1d040000 0x10000>,
+            	 <3 0 0x00 0x1d050000 0x10000>,
+            	 <4 0 0x00 0x1d020000 0x10000>,
+            	 <5 0 0x00 0x10000 0x40000000>,
+            	 <6 0 0x00 0x10000 0x50000000>,
+            	 <7 0 0x00 0x10000 0x90000000>;
 
 			cavium,cs-config@0 {
 				cavium,t-adr = <10>;
@@ -196,7 +188,6 @@
 				cavium,t-page = <25>;
 			};
 
-			/* cs2/cs3 do not exist in the dtsi - new chip selects */
 			cavium,cs-config@2 {
 				compatible = "cavium,octeon-3860-bootbus-config";
 				cavium,cs-index = <2>;
@@ -244,6 +235,10 @@
 				cavium,t-page = <300>;
 				cavium,t-rd-dly = <10>;
 			};
+
+			/delete-node/ cavium,cs-config@5;
+
+			/delete-node/ cavium,cs-config@6;
 		};
 	};
 
--- a/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
+++ b/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
@@ -1,104 +1,83 @@
 // SPDX-License-Identifier: GPL-2.0-only
 /*
-* Device tree source for Ubiquiti UniFi Security Gateway Pro (USG-Pro-4).
-*
-* Written by: Niko Huuskonen <niko.huuskonen.00@gmail.com>
-*/
+ * Device tree source for Ubiquiti UniFi Security Gateway Pro (USG-Pro-4).
+ *
+ * Written by: Niko Huuskonen <niko.huuskonen.00@gmail.com>
+ */
 
 /include/ "octeon_3xxx.dtsi"
 #include <dt-bindings/gpio/gpio.h>
 #include <dt-bindings/input/input.h>
 #include <dt-bindings/leds/common.h>
 
+/* octeon,6120 is an Octeon II (CN61xx-class) part, which exposes USB via
+ * EHCI/OHCI behind UCTL (defined below) - not the older 5750-generation
+ * USBN/USBC OTG block the dtsi provides for CN5xxx-class chips. That block
+ * doesn't correspond to real hardware on this SoC, so drop it rather than
+ * leaving a dead node in the compiled tree.
+ */
+
+/delete-node/ &usbn;
+
 / {
 	compatible = "ubnt,usg-pro-4", "cavium,octeon-6120";
 	model = "Ubiquiti UniFi Security Gateway Pro (USG-Pro-4)";
-	interrupt-parent = <&ciu>;
-	
-	
-	&soc@0 {
-		
+
+	soc@0 {
 		smi0: mdio@1180000001800 {
-			compatible = "cavium,octeon-3860-mdio";
-			reg = <0x11800 0x1800 0x00 0x40>;
-			#address-cells = <1>;
-			#size-cells = <0>;
-			
 			phy_eth0: ethernet-phy@4 {
 				compatible = "ethernet-phy-ieee802.3-c22";
 				reg = <4>;
 				interrupt-parent = <&gpio>;
 			};
-			
+
 			phy_eth1: ethernet-phy@5 {
 				compatible = "ethernet-phy-ieee802.3-c22";
 				reg = <5>;
 				interrupt-parent = <&gpio>;
 			};
 		};
-		
+
 		smi1: mdio@1180000001900 {
 			compatible = "cavium,octeon-3860-mdio";
 			reg = <0x11800 0x1900 0x00 0x40>;
 			#address-cells = <1>;
 			#size-cells = <0>;
-			
+
 			phy_eth2: ethernet-phy@0 {
 				compatible = "ethernet-phy-ieee802.3-c22";
 				reg = <0>;
 				interrupt-parent = <&gpio>;
 			};
-			
+
 			phy_eth3: ethernet-phy@1 {
 				compatible = "ethernet-phy-ieee802.3-c22";
 				reg = <1>;
 				interrupt-parent = <&gpio>;
 			};
 		};
-		
+
 		pip: pip@11800a0000000 {
-			compatible = "cavium,octeon-3860-pip";
-			reg = <0x11800 0xa0000000 0x00 0x2000>;
-			#address-cells = <1>;
-			#size-cells = <0>;
-			
-			interface@1 {
-				compatible = "cavium,octeon-3860-pip-interface";
-				reg = <1>;
-				#address-cells = <1>;
-				#size-cells = <0>;
-			};
-			
 			interface@0 {
-				compatible = "cavium,octeon-3860-pip-interface";
-				reg = <0>;
-				#address-cells = <1>;
-				#size-cells = <0>;
-				
 				ethernet@0 {
-					compatible = "cavium,octeon-3860-pip-port";
-					reg = <0>;
 					phy-handle = <&phy_eth0>;
 					rx-delay = <0>;
 					tx-delay = <0x10>;
 				};
-				
+
 				ethernet@1 {
-					compatible = "cavium,octeon-3860-pip-port";
-					reg = <1>;
 					phy-handle = <&phy_eth1>;
 					rx-delay = <0>;
 					tx-delay = <0x10>;
 				};
-				
+
 				ethernet@2 {
-					compatible = "cavium,octeon-3860-pip-port";
-					reg = <2>;
 					phy-handle = <&phy_eth2>;
 					rx-delay = <0>;
 					tx-delay = <0x10>;
 				};
-				
+
+				/* Port 3 does not exist in the dtsi - new port */
 				ethernet@3 {
 					compatible = "cavium,octeon-3860-pip-port";
 					reg = <3>;
@@ -108,11 +87,11 @@
 				};
 			};
 		};
-		
-		&uart0: serial@1180000000800 {
+
+		uart0: serial@1180000000800 {
 			clock-frequency = <600000000>;
 		};
-		
+
 		uart1: serial@1180000000c00 {
 			compatible = "cavium,octeon-3860-uart", "ns16550";
 			reg = <0x11800 0xc00 0x00 0x400>;
@@ -121,11 +100,7 @@
 			current-speed = <115200>;
 			interrupts = <0 35>;
 		};
-		
-		&twsi0 {
-			reg = <0x11800 0x1000 0x00 0x200>;
-		};
-		
+
 		twsi1: i2c@1180000001200 {
 			compatible = "cavium,octeon-3860-twsi";
 			reg = <0x11800 0x1200 0x00 0x200>;
@@ -134,7 +109,7 @@
 			clock-frequency = <100000>;
 			interrupts = <0 59>;
 		};
-		
+
 		spi: spi@1070000001000 {
 			compatible = "cavium,octeon-3010-spi";
 			reg = <0x10700 0x1000 0x00 0x100>;
@@ -142,23 +117,31 @@
 			#size-cells = <0>;
 			interrupts = <0 58>;
 		};
-		
+
 		mmc0: mmc@1180000002000 {
 			compatible = "cavium,octeon-6130-mmc";
 			reg = <0x11800 0x2000 0x00 0x100 0x11800 0x168 0x00 0x20>;
 			#address-cells = <1>;
 			#size-cells = <0>;
 			interrupts = <1 19 0 0x3f>;
-			
-			mmc_slot0: mmc-slot@0 {
-				compatible = "cavium,octeon-6130-mmc-slot";
+
+			mmc_slot0:mmc-slot@0 {
+				compatible = "cavium,octeon-6130-mmc-slot", "mmc-spi-slot";
 				reg = <0>;
 				spi-max-frequency = <26000000>;
 				voltage-ranges = <3300 3300>;
 				cavium,bus-max-width = <8>;
 			};
 		};
-		
+
+		/* Board-specific USB controller. Note this coexists with
+		 * the dtsi's "usbn" node (different name/unit-address, so
+		 * it isn't merged/replaced by this). If usbn is not present
+		 * on this SoC variant, consider deleting it explicitly with
+		 * /delete-node/ &usbn; to avoid a phantom, non-functional
+		 * node in the compiled tree.
+		 */
+
 		uctl: uctl@118006f000000 {
 			compatible = "cavium,octeon-6335-uctl";
 			reg = <0x11800 0x6f000000 0x00 0x100>;
@@ -167,14 +150,14 @@
 			ranges;
 			refclk-frequency = <12000000>;
 			refclk-type = "crystal";
-			
+
 			ehci0: ehci@16f0000000000 {
 				compatible = "cavium,octeon-6335-ehci", "usb-ehci";
 				reg = <0x16f00 0x00 0x00 0x100>;
 				interrupts = <0 56>;
 				big-endian-regs;
 			};
-			
+
 			ohci0: ohci@16f0000000400 {
 				compatible = "cavium,octeon-6335-ohci", "usb-ohci";
 				reg = <0x16f00 0x400 0x00 0x100>;
@@ -182,8 +165,17 @@
 				big-endian-regs;
 			};
 		};
-		
-		&bootbus {
+
+		bootbus: bootbus@1180000000000 {
+			/* Values sourced from a live devicetree dump on this
+			 * board's EdgeRouter Pro-based OpenWrt install, so
+			 * kept verbatim even though several entries differ
+			 * from the dtsi's generic ones. A DT property is
+			 * replaced as a whole (not merged entry-by-entry), so
+			 * the full list has to stay here rather than being
+			 * trimmed to just the differing lines.
+			 */
+
 			ranges = <0 0  0x00 0x1f400000  0xc00000>,
 			<1 0  0x00 0x10000     0x30000000>,
 			<2 0  0x00 0x1d040000  0x10000>,
@@ -192,24 +184,19 @@
 			<5 0  0x00 0x10000     0x40000000>,
 			<6 0  0x00 0x10000     0x50000000>,
 			<7 0  0x00 0x10000     0x90000000>;
-			
+
 			cavium,cs-config@0 {
-				compatible = "cavium,octeon-3860-bootbus-config";
-				cavium,cs-index = <0>;
-				cavium,bus-width = <8>;
 				cavium,t-adr = <10>;
 				cavium,t-ce = <50>;
 				cavium,t-oe = <50>;
 				cavium,t-we = <35>;
 				cavium,t-rd-hld = <25>;
 				cavium,t-wr-hld = <35>;
-				cavium,t-pause = <0>;
 				cavium,t-wait = <300>;
 				cavium,t-page = <25>;
-				cavium,pages = <0>;
-				cavium,t-rd-dly = <0>;
 			};
-			
+
+			/* cs2/cs3 do not exist in the dtsi - new chip selects */
 			cavium,cs-config@2 {
 				compatible = "cavium,octeon-3860-bootbus-config";
 				cavium,cs-index = <2>;
@@ -226,7 +213,7 @@
 				cavium,pages = <0>;
 				cavium,t-rd-dly = <0>;
 			};
-			
+
 			cavium,cs-config@3 {
 				compatible = "cavium,octeon-3860-bootbus-config";
 				cavium,cs-index = <3>;
@@ -244,11 +231,8 @@
 				cavium,pages = <0>;
 				cavium,t-rd-dly = <0>;
 			};
-			
+
 			cavium,cs-config@4 {
-				compatible = "cavium,octeon-3860-bootbus-config";
-				cavium,cs-index = <4>;
-				cavium,bus-width = <8>;
 				cavium,t-adr = <10>;
 				cavium,t-ce = <10>;
 				cavium,t-oe = <160>;
@@ -258,22 +242,21 @@
 				cavium,t-pause = <50>;
 				cavium,t-wait = <300>;
 				cavium,t-page = <300>;
-				cavium,pages = <0>;
 				cavium,t-rd-dly = <10>;
 			};
 		};
 	};
-	
+
 	leds {
 		compatible = "gpio-leds";
-		
+
 		led_blue: logo_blue {
 			color = <LED_COLOR_ID_BLUE>;
 			function = "logo";
 			gpios = <&gpio 14 GPIO_ACTIVE_HIGH>;
 			default-state = "off";
 		};
-		
+
 		led_white: logo_white {
 			color = <LED_COLOR_ID_WHITE>;
 			function = "logo";
@@ -281,11 +264,11 @@
 			default-state = "on";
 		};
 	};
-	
+
 	keys {
 		compatible = "gpio-keys-polled";
 		poll-interval = <20>;
-		
+
 		button-reset {
 			label = "reset";
 			linux,code = <KEY_RESTART>;
@@ -293,7 +276,7 @@
 			debounce-interval = <60>;
 		};
 	};
-	
+
 	aliases {
 		flash0 = &flash0;
 		gpio-controller = &gpio;
@@ -311,7 +294,7 @@
 		uart0 = &uart0;
 		uart1 = &uart1;
 	};
-	
+
 	memory@0 {
 		device_type = "memory";
 		reg = <0x00 0x00 0x00 0x10000000 0x00 0x20000000 0x00 0x70000000>;
--- a/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
+++ b/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
@@ -315,13 +315,15 @@
 		compatible = "gpio-leds";
 		
 		led_blue: led-14 {
-			label = "ubnt:blue:logo";
+			color = <LED_COLOR_ID_BLUE>;
+			function = "logo";
 			gpios = <&gpio 14 GPIO_ACTIVE_HIGH>;
 			default-state = "off";
 		};
-		
+
 		led_white: led-15 {
-			label = "ubnt:white:logo";
+			color = <LED_COLOR_ID_WHITE>;
+			function = "logo";
 			gpios = <&gpio 15 GPIO_ACTIVE_HIGH>;
 			default-state = "on";
 		};
--- a/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
+++ b/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
@@ -8,41 +8,15 @@
 /include/ "octeon_3xxx.dtsi"
 #include <dt-bindings/gpio/gpio.h>
 #include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
 
 / {
 	compatible = "ubnt,usg-pro-4", "cavium,octeon-6120";
 	model = "Ubiquiti UniFi Security Gateway Pro (USG-Pro-4)";
 	interrupt-parent = <&ciu>;
 	
-	soc@0 {
-		ciu: interrupt-controller@1070000000000 {
-			compatible = "cavium,octeon-3860-ciu";
-			interrupt-controller;
-			#interrupt-cells = <2>;
-			reg = <0x10700 0x00 0x00 0x7000>;
-		};
-		
-		gpio: gpio-controller@1070000000800 {
-			compatible = "cavium,octeon-3860-gpio";
-			reg = <0x10700 0x800 0x00 0x100>;
-			gpio-controller;
-			#gpio-cells = <2>;
-			interrupt-controller;
-			#interrupt-cells = <2>;
-			interrupts = <0 16>;
-		};
-		
-		dma0: dma-engine@1180000000100 {
-			compatible = "cavium,octeon-5750-bootbus-dma";
-			reg = <0x11800 0x100 0x00 0x08>;
-			interrupts = <0 63>;
-		};
-		
-		dma1: dma-engine@1180000000108 {
-			compatible = "cavium,octeon-5750-bootbus-dma";
-			reg = <0x11800 0x108 0x00 0x08>;
-			interrupts = <0 63>;
-		};
+	
+	&soc@0 {
 		
 		smi0: mdio@1180000001800 {
 			compatible = "cavium,octeon-3860-mdio";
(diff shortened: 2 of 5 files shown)


target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
```diff
@@ -0,0 +1,362 @@
...
+			#address-cells = <2>;
+			#size-cells = <1>;
+			ranges = <0 0  0x00 0x1f400000  0xc00000>,
+			<1 0  0x00 0x10000     0x30000000>,
+			<2 0  0x00 0x1d040000  0x10000>,
+			<3 0  0x00 0x1d050000  0x10000>,
+			<4 0  0x00 0x1d020000  0x10000>,
+			<5 0  0x00 0x10000     0x40000000>,
+			<6 0  0x00 0x10000     0x50000000>,
+			<7 0  0x00 0x10000     0x90000000>;
```
openwrt-ai: The parent-address and size cells are shifted by one in entries 1, 5, 6 and 7. `bootbus` declares `#address-cells = <2>` / `#size-cells = <1>` and its parent `soc@0` declares `#address-cells = <2>`, so each row is `<cs offset> <parent-hi parent-lo> <size>`. `<1 0  0x00 0x10000  0x30000000>` therefore maps CS1 at physical `0x0000_0000_0001_0000` with a 0x30000000-byte window, instead of at `0x00010000_30000000` with size 0. `0x30000000` / `0x40000000` / `0x50000000` / `0x90000000` are clearly addresses, not region sizes. Compare the values this node inherits from [octeon_3xxx.dtsi:119-126](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/boot/dts/cavium-octeon/octeon_3xxx.dtsi#L119-L126) and the sibling [cn6130_cisco_vedge1000.dts](``https://github.com/openwrt/openwrt/blob/521890cc1ce837ad8bda624b4aac45b30bc0931c/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6130_cisco_vedge1000.dts#L190-L197).``

Separately: entries 2 and 3 here take the addresses the base dtsi assigns to CS5 and CS6 (`0x1d040000` / `0x1d050000`). Is that a real board difference, or did the rows get shuffled along with the cell shift?

```suggestion
			ranges = <0 0  0x00 0x1f400000  0xc00000>,
				 <1 0  0x10000 0x30000000  0x00>,
				 <2 0  0x00 0x1d040000  0x10000>,
				 <3 0  0x00 0x1d050000  0x10000>,
				 <4 0  0x00 0x1d020000  0x10000>,
				 <5 0  0x10000 0x40000000  0x00>,
				 <6 0  0x10000 0x50000000  0x00>,
				 <7 0  0x10000 0x90000000  0x00>;
```


---

target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
```diff
@@ -0,0 +1,362 @@
...
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+
+/ {
+	compatible = "ubnt,usg-pro-4", "cavium,octeon-6120";
+	model = "Ubiquiti UniFi Security Gateway Pro (USG-Pro-4)";
+	interrupt-parent = <&ciu>;
+	
+	soc@0 {
+		ciu: interrupt-controller@1070000000000 {
```
openwrt-ai: This file re-declares almost everything the included `octeon_3xxx.dtsi` already provides — `ciu`, `gpio`, `dma0`, `dma1`, `smi0`, `pip`/`interface@0`/`interface@1`, `uart0`, `twsi0`, `bootbus`, `cavium,cs-config@0`, `cavium,cs-config@4` and `flash0` are all verbatim (or near-verbatim) copies of [octeon_3xxx.dtsi](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/boot/dts/cavium-octeon/octeon_3xxx.dtsi).

Both octeon siblings that include the same dtsi are written as overlays that only state the board deltas — see [cn5020_ubnt_usg.dts](``https://github.com/openwrt/openwrt/blob/521890cc1ce837ad8bda624b4aac45b30bc0931c/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn5020_ubnt_usg.dts)`` and [cn6130_cisco_vedge1000.dts](``https://github.com/openwrt/openwrt/blob/521890cc1ce837ad8bda624b4aac45b30bc0931c/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6130_cisco_vedge1000.dts),`` which is the closer match here since it is also a CN61xx. Please drop the copied nodes and keep only what differs (smi0/smi1 PHYs, the pip ports, mmc, uctl, the CS timings that actually differ, LEDs/keys/aliases/memory).

Beyond the churn this actively costs correctness: the copied `bootbus` `ranges` is what introduced the shifted-cells bug flagged below, and the copied `gpio` node replaces the dtsi's full 16-entry `interrupts` list with a single `<0 16>`.

---

target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
```diff
@@ -0,0 +1,362 @@
...
+		};
+		
+		smi0: mdio@1180000001800 {
+			compatible = "cavium,octeon-3860-mdio";
+			reg = <0x11800 0x1800 0x00 0x40>;
+			#address-cells = <1>;
+			#size-cells = <0>;
+			
+			phy_eth0: ethernet-phy@4 {
+				compatible = "broadcom,bcm", "ethernet-phy-ieee802.3-c22";
```
openwrt-ai: `"broadcom,bcm"` is not a compatible string that any driver or binding defines — there is no match for it anywhere in the kernel tree (Broadcom PHY bindings use concrete strings such as `broadcom,bcm8706`, see [broadcom-bcm87xx.txt](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/net/broadcom-bcm87xx.txt#L1-L20)). It is silently ignored, so drop it and keep only the generic c22 entry, matching [cn5020_ubnt_usg.dts:20-22](``https://github.com/openwrt/openwrt/blob/521890cc1ce837ad8bda624b4aac45b30bc0931c/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn5020_ubnt_usg.dts#L20-L22).``

```suggestion
				compatible = "ethernet-phy-ieee802.3-c22";
```

Same applies to lines 60, 73 and 79.

While here: the `interrupt-parent = <&gpio>;` on each of these four PHY nodes has no accompanying `interrupts` property, so it does nothing — either wire up the real PHY interrupt or drop the line.

---

target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
```diff
@@ -0,0 +1,362 @@
...
+			label = "ubnt:blue:logo";
+			gpios = <&gpio 14 GPIO_ACTIVE_HIGH>;
+			default-state = "off";
+		};
+		
+		led_white: led-15 {
+			label = "ubnt:white:logo";
+			gpios = <&gpio 15 GPIO_ACTIVE_HIGH>;
+			default-state = "on";
+		};
```
openwrt-ai: The `label = "<vendor>:<color>:<function>"` form is deprecated for new devices in OpenWrt — use `color` + `function`. There is no `LED_FUNCTION_LOGO` in `dt-bindings/leds/common.h`, so the string fallback `function = "logo";` is the right form here.

```suggestion
		led_blue: led-14 {
			color = <LED_COLOR_ID_BLUE>;
			function = "logo";
			gpios = <&gpio 14 GPIO_ACTIVE_HIGH>;
			default-state = "off";
		};

		led_white: led-15 {
			color = <LED_COLOR_ID_WHITE>;
			function = "logo";
			gpios = <&gpio 15 GPIO_ACTIVE_HIGH>;
			default-state = "on";
		};
```

This also needs `#include <dt-bindings/leds/common.h>` added next to the other includes at lines 9-10.

---

target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
```diff
@@ -0,0 +1,362 @@
...
+	};
+	
+	keys {
+		compatible = "gpio-keys";
+		
+		reset {
+			label = "reset";
+			gpios = <&gpio 0 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
```
openwrt-ai: The node name `reset` does not match the name pattern the `gpio-keys` binding requires — `^(button|event|key|switch|(button|event|key|switch)-[a-z0-9-]+|[a-z0-9-]+-(button|event|key|switch))$` ([gpio-keys.yaml:26](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/input/gpio-keys.yaml#L26)). Also add a debounce interval like the sibling [cn5020_ubnt_usg.dts:78-87](``https://github.com/openwrt/openwrt/blob/521890cc1ce837ad8bda624b4aac45b30bc0931c/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn5020_ubnt_usg.dts#L78-L87).``

```suggestion
		button-reset {
			label = "reset";
			linux,code = <KEY_RESTART>;
			gpios = <&gpio 0 GPIO_ACTIVE_LOW>;
			debounce-interval = <60>;
		};
```

---

target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
```diff
@@ -0,0 +1,362 @@
...
+		
+		mmc0: mmc@1180000002000 {
+			compatible = "cavium,octeon-6130-mmc";
+			reg = <0x11800 0x2000 0x00 0x100 0x11800 0x168 0x00 0x20>;
+			#address-cells = <1>;
+			#size-cells = <0>;
+			interrupts = <1 19 0 0x3f>;
+			
+			mmc_slot0:mmc-slot@0 {
+				compatible = "cavium,octeon-6130-mmc-slot", "mmc-spi-slot";
```
openwrt-ai: `"mmc-spi-slot"` describes an MMC card wired to a SPI master ([mmc-spi-slot.yaml](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/mmc/mmc-spi-slot.yaml#L21)); this slot is a child of the Octeon MMC controller, not of a SPI bus, so the string is wrong here. [cn6130_cisco_vedge1000.dts:307-309](``https://github.com/openwrt/openwrt/blob/521890cc1ce837ad8bda624b4aac45b30bc0931c/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6130_cisco_vedge1000.dts#L307-L309)`` uses the cavium compatible alone. The label also needs a space after the colon, unlike every other label in this file.

```suggestion
			mmc_slot0: mmc-slot@0 {
				compatible = "cavium,octeon-6130-mmc-slot";
```

---

target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
```diff
@@ -0,0 +1,362 @@
...
+
+/include/ "octeon_3xxx.dtsi"
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+
+/ {
+	compatible = "ubnt,usg-pro-4", "cavium,octeon-6120";
+	model = "Ubiquiti UniFi Security Gateway Pro (USG-Pro-4)";
+	interrupt-parent = <&ciu>;
+
```
openwrt-ai: nit: 40 of the added lines end in trailing whitespace — every "blank" separator line in this file is padded with the surrounding indentation tabs (this line, plus 24, 34, 40, 46, ...). Please strip them.

Two more formatting deviations from the neighbouring octeon DTS files:
- the header comment block (lines 3-6) puts `*` in column 0 instead of aligning as ` * `, unlike [cn5020_ubnt_usg.dts:1-7](``https://github.com/openwrt/openwrt/blob/521890cc1ce837ad8bda624b4aac45b30bc0931c/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn5020_ubnt_usg.dts#L1-L7);``
- the `ranges` continuation lines at 228-234 are not indented to line up under the first entry.

---

target/linux/octeon/base-files/etc/board.d/01_network
```diff
@@ -21,6 +21,9 @@ ubnt,edgerouter-6p)
 cisco,vedge1000)
 	ucidef_set_interfaces_lan_wan "mgmt0" "lan0"
 	;;
+ubnt,usg-pro-4)
+  ucidef_set_interfaces_lan_wan "eth0 eth1" "eth2 eth3"
+  ;;
```
openwrt-ai: This block is indented with two spaces; every other case arm in this file uses a single tab.

```suggestion
ubnt,usg-pro-4)
	ucidef_set_interfaces_lan_wan "eth0 eth1" "eth2 eth3"
	;;
```

---

target/linux/octeon/base-files/etc/board.d/01_network
```diff
@@ -21,6 +21,9 @@ ubnt,edgerouter-6p)
 cisco,vedge1000)
 	ucidef_set_interfaces_lan_wan "mgmt0" "lan0"
 	;;
+ubnt,usg-pro-4)
+  ucidef_set_interfaces_lan_wan "eth0 eth1" "eth2 eth3"
```
openwrt-ai: Passing two devices as the WAN argument does not do what the LAN argument does. `config_generate` force-bridges only `lan` when no `ports` array is present ([config_generate:107-130](https://github.com/openwrt/openwrt/blob/521890cc1ce837ad8bda624b4aac45b30bc0931c/package/base-files/files/bin/config_generate#L107-L130)); for any other interface it falls through to `set network.wan.device='...'` at [line 157](https://github.com/openwrt/openwrt/blob/521890cc1ce837ad8bda624b4aac45b30bc0931c/package/base-files/files/bin/config_generate#L157), so this ends up writing `option device 'eth2 eth3'`, which is not a resolvable device name.

Did you intend eth3 (WAN2) to be a second, independent WAN? If so it needs its own interface, e.g. `ucidef_set_interfaces_lan_wan "eth0 eth1" "eth2"` plus an explicit `ucidef_set_interface "wan2" device "eth3" protocol "dhcp"`. If you really do want them bridged, the `ports` form is needed. Note there is one in-tree precedent with a two-device WAN ([x86 02_network:87](https://github.com/openwrt/openwrt/blob/521890cc1ce837ad8bda624b4aac45b30bc0931c/target/linux/x86/base-files/etc/board.d/02_network#L87)), but it looks like it hits the same problem rather than being an endorsement of the pattern.

---

target/linux/octeon/image/Makefile
```diff
@@ -53,6 +53,19 @@ define Device/ubnt_edgerouter
...
+define Device/ubnt_usg-pro-4
+  DEVICE_VENDOR := Ubiquiti
+  DEVICE_MODEL := UniFi Security Gateway Pro (USG-Pro-4)
+  DEVICE_DTS := cn6120_ubnt_usg-pro-4
+  BOARD_NAME := ubnt,usg-pro-4
+  CMDLINE := $(ER_CMDLINE) ubnt_board=usg-pro-4
+  KERNEL := kernel-bin | patch-cmdline | append-dtb-to-elf
+  KERNEL_DEPENDS := $$(wildcard $(DTS_DIR)/$(DEVICE_DTS).dts)
+  DEVICE_PACKAGES := kmod-leds-gpio kmod-gpio-button-hotplug kmod-sfp kmod-phy-broadcom kmod-of-mdio
+  SUPPORTED_DEVICES += ubnt,usg-pro-4
```
openwrt-ai: nit: redundant. `Device/Default` already derives `SUPPORTED_DEVICES := $(subst _,$(comma),$(1))` ([include/image.mk:551](https://github.com/openwrt/openwrt/blob/521890cc1ce837ad8bda624b4aac45b30bc0931c/include/image.mk#L551)), which for the profile name `ubnt_usg-pro-4` is already `ubnt,usg-pro-4`, so this appends a duplicate. The neighbouring blocks only add entries when they differ from the derived value (`ubnt_unifi-usg` → `ubnt,usg`, `ubnt_edgerouter-lite` → `erlite`). Please drop the line.

---

target/linux/octeon/image/Makefile
```diff
@@ -53,6 +53,19 @@ define Device/ubnt_edgerouter
...
 
+define Device/ubnt_usg-pro-4
+  DEVICE_VENDOR := Ubiquiti
+  DEVICE_MODEL := UniFi Security Gateway Pro (USG-Pro-4)
+  DEVICE_DTS := cn6120_ubnt_usg-pro-4
+  BOARD_NAME := ubnt,usg-pro-4
+  CMDLINE := $(ER_CMDLINE) ubnt_board=usg-pro-4
+  KERNEL := kernel-bin | patch-cmdline | append-dtb-to-elf
+  KERNEL_DEPENDS := $$(wildcard $(DTS_DIR)/$(DEVICE_DTS).dts)
+  DEVICE_PACKAGES := kmod-leds-gpio kmod-gpio-button-hotplug kmod-sfp kmod-phy-broadcom kmod-of-mdio
```
openwrt-ai: nit: every other device block in this file uses `DEVICE_PACKAGES +=` (and keeps the list alphabetically sorted); `:=` would silently drop anything a shared block adds later.

```suggestion
  DEVICE_PACKAGES += kmod-gpio-button-hotplug kmod-leds-gpio kmod-of-mdio kmod-phy-broadcom kmod-sfp
```

---

target/linux/octeon/image/Makefile
```diff
@@ -53,6 +53,19 @@ define Device/ubnt_edgerouter
 endef
 TARGET_DEVICES += ubnt_edgerouter
 
+define Device/ubnt_usg-pro-4
+  DEVICE_VENDOR := Ubiquiti
+  DEVICE_MODEL := UniFi Security Gateway Pro (USG-Pro-4)
+  DEVICE_DTS := cn6120_ubnt_usg-pro-4
+  BOARD_NAME := ubnt,usg-pro-4
+  CMDLINE := $(ER_CMDLINE) ubnt_board=usg-pro-4
```
openwrt-ai: nit: `ubnt_board=usg-pro-4` is not consumed anywhere — grepping `target/` and `package/` for `ubnt_board` returns no hits, and nothing in this PR reads it either (board identification goes through the DT compatible via `01_sysinfo`). Dead kernel command-line token.

```suggestion
  CMDLINE := $(ER_CMDLINE)
```

Worth double-checking the rest of `ER_CMDLINE` too: it hardcodes `-mtdparts=phys_mapped_flash:640k(boot0)ro,640k(boot1)ro,64k(eeprom)ro`, which describes the EdgeRouter's NOR layout. If the USG-Pro-4's 8 MB NOR is partitioned differently, this should be its own `*_CMDLINE` rather than a reuse.

---

target/linux/octeon/image/Makefile
```diff
@@ -53,6 +53,19 @@ define Device/ubnt_edgerouter
 endef
 TARGET_DEVICES += ubnt_edgerouter
 
+define Device/ubnt_usg-pro-4
```
openwrt-ai: nit: two consistency points on this block.

1. It is inserted between `Device/ubnt_edgerouter` and `Device/ubnt_edgerouter-e300`, splitting the EdgeRouter group. Placing it next to `Device/ubnt_unifi-usg` (the other UniFi Security Gateway) would keep the file grouped by product family.
2. The existing USG uses the profile name `ubnt_unifi-usg`; this one uses `ubnt_usg-pro-4`, so the two members of the same family get different naming schemes and hence different image filenames (`...ubnt_unifi-usg-...` vs `...ubnt_usg-pro-4-...`). Was `ubnt_unifi-usg-pro-4` considered, or is the shorter form deliberate?

Also, the PR title says "UniFi Security Gateway Pro 4" while `DEVICE_MODEL` (line 58), the DTS `model`, and the commit subject all say "UniFi Security Gateway Pro (USG-Pro-4)" — worth aligning the title.

---

openwrt-ai (commented):
## Commit checks

- `521890cc` "octeon: add build config for Ubiquiti UniFi Security Gateway Pro (USG-Pro-4)" — the subject says "add build config", but the diff adds a complete new device: a 362-line DTS, a `define Device/ubnt_usg-pro-4` block, network/sysinfo/sysupgrade board handling. This is new device support, so the usual `octeon: add support for Ubiquiti UniFi Security Gateway Pro (USG-Pro-4)` subject fits the change better.

- `521890cc` — the commit message body is a changelog of the touched files ("Adds hardware references for network, sysinfo, platform" / "Adds make target to Makefile" / "Adds DTS file..."), not a device description. A commit introducing a new device needs the three standard sections in the message itself, none of which are present:
  1. **Hardware specification** — SoC, RAM, Flash, Ethernet, LEDs/Buttons, UART (header location, pinout, baud), Power. Most of this is already in the PR description; it just needs to move into the commit message.
  2. **Flash instructions** — the PR description has a U-Boot/USB procedure; please move it in and fix it while you do: the steps are numbered 1, 2, 4, 5 (no 3), and step 1 names `openwrt-octeon-generic-ubnt_usg-pro-4-initramfs-kernel.bin` while the `fatload` in step 2 uses `openwrt-octeon-ubnt_usg-pro-4-initramfs-kernel.bin`. A revert-to-stock note would be good too.
  3. **MAC address layout** — missing entirely, from both the commit message and the PR description. Where does each of eth0..eth3 get its MAC from? The DTS has no `nvmem-cells` for the pip ports, unlike the sibling [cn6130_cisco_vedge1000.dts](``https://github.com/openwrt/openwrt/blob/521890cc1ce837ad8bda624b4aac45b30bc0931c/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6130_cisco_vedge1000.dts#L26-L29),`` so presumably the Octeon bootloader patches `local-mac-address` into the appended DTB — please state that explicitly, one line per interface.

CI on `521890cc` is green (all three FormalityCheck runs passed), so nothing below is CI-derived.

---

target/linux/octeon/base-files/etc/board.d/01_network
N1kO23: The devices do in fact resolve correctly. They form a `br-wan` interface the same way as how multiple lan devices are under `br-lan`

target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
```diff
@@ -0,0 +1,319 @@
...
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+
+/ {
+	compatible = "ubnt,usg-pro-4", "cavium,octeon-6120";
+	model = "Ubiquiti UniFi Security Gateway Pro (USG-Pro-4)";
+	interrupt-parent = <&ciu>;
+	
+	
+	&soc@0 {
```
openwrt-ai: This `&soc@0` node — along with `&uart0: serial@1180000000800` (line 112), `&twsi0` (line 125) and `&bootbus` (line 186) — uses `&label` reference-override syntax nested inside the root `/ { }` node. That is not valid DTS. `&label { }` overrides are only allowed at the top level of the source (as siblings of `/ { }`), never as child nodes, and this file uses `/include/ "octeon_3xxx.dtsi"` (textual include), where a node is extended by matching its path/name — not by phandle reference. Two further problems compound it: a reference is `&soc`, never `&soc@0` (the `@unit` belongs to a node name, not a label), and `&uart0: serial@…` tries to attach a label to a reference, which is also invalid.

`octeon_3xxx.dtsi` already defines all four of these nodes, so extend them exactly as the sibling `cn6130_cisco_vedge1000.dts` does — with plain node names that merge by path:

- `&soc@0 {` -> `soc@0 {`
- `&uart0: serial@1180000000800 {` -> `serial@1180000000800 {`
- `&twsi0 {` -> `i2c@1180000001000 {`
- `&bootbus {` -> `bootbus@1180000000000 {`

As written, `dtc` rejects the file and the device image won't build. The FormalityCheck CI doesn't compile DTS, so it stays green despite this.

---

openwrt-ai (commented):
## Commit checks

- `c9ff91cc` "octeon: wip update cn6120_ubnt_usg-pro-4.dts" — the subject is still marked `wip`, and the body ("Remove duplicate entries, add import for leds, remove wrong compability definitions") understates what the commit does: it rewrites the `soc@0`, `uart0`, `twsi0` and `bootbus` nodes into `&`-reference overrides nested in the root node, which is invalid DTS and breaks the device build (see the inline comment on the DTS). Please fix the syntax and drop the `wip` marker before this lands.

- Series hygiene: this new-device PR now carries 8 commits, most of them `octeon: update …` fixups layered on the initial one. Consider squashing them into a single well-formed "add support" commit — folding in the hardware-spec / flash-instructions / MAC-layout sections noted in the earlier review — before merge.

The other new commits (`4b79fb41`, `27b9ab6e`, `0d0d8038`, `c13607c4`, `dd35f3ba`, `628a32c5`) match their diffs. FormalityCheck CI is green, but it does not compile the DTS, so nothing here is CI-derived.

---

target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
N1kO23: Yea this was a quick and dirty commit, realized after the fact. I'll revamp the file to clean it up properly in a bit after I verify the build compiles, boots and functions on real hardware

target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
```diff
@@ -0,0 +1,302 @@
...
+
+		mmc0: mmc@1180000002000 {
+			compatible = "cavium,octeon-6130-mmc";
+			reg = <0x11800 0x2000 0x00 0x100 0x11800 0x168 0x00 0x20>;
+			#address-cells = <1>;
+			#size-cells = <0>;
+			interrupts = <1 19 0 0x3f>;
+
+			mmc_slot0:mmc-slot@0 {
+				compatible = "cavium,octeon-6130-mmc-slot", "mmc-spi-slot";
```
openwrt-ai: This reverts [`0d0d8038`](https://github.com/openwrt/openwrt/commit/0d0d80389b757aa08723eee28adb25b341accc7f) from earlier in this same series: `"mmc-spi-slot"` is back in the compatible list and the space after the label colon is gone again. `mmc-spi-slot` describes an MMC card behind a SPI master; this slot is a child of the Octeon MMC controller, and the sibling [cn6130_cisco_vedge1000.dts:167-168](``https://github.com/openwrt/openwrt/blob/3c33550b068a6865bc225fc857aed7418359af28/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6130_cisco_vedge1000.dts#L167-L168)`` uses the cavium compatible alone.

```suggestion
			mmc_slot0: mmc-slot@0 {
				compatible = "cavium,octeon-6130-mmc-slot";
```

---

target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
```diff
@@ -0,0 +1,302 @@
...
+		};
+
+		/* Board-specific USB controller. Note this coexists with
+		 * the dtsi's "usbn" node (different name/unit-address, so
+		 * it isn't merged/replaced by this). If usbn is not present
+		 * on this SoC variant, consider deleting it explicitly with
+		 * /delete-node/ &usbn; to avoid a phantom, non-functional
+		 * node in the compiled tree.
+		 */
+
```
openwrt-ai: This comment is stale and now contradicts the file: it tells the reader to "consider deleting it explicitly with `/delete-node/ &usbn;`", which this same commit already does at line 20. It also still claims `usbn` "coexists with" `uctl`, which is no longer true after the deletion. The line-13 block already explains the reasoning, so this one can go.

```suggestion
```

---

target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
```diff
@@ -0,0 +1,302 @@
...
+		bootbus: bootbus@1180000000000 {
+			/* Values sourced from a live devicetree dump on this
+			 * board's EdgeRouter Pro-based OpenWrt install, so
+			 * kept verbatim even though several entries differ
+			 * from the dtsi's generic ones. A DT property is
+			 * replaced as a whole (not merged entry-by-entry), so
+			 * the full list has to stay here rather than being
+			 * trimmed to just the differing lines.
+			 */
```
openwrt-ai: This justification explains *why the list differs from the dtsi*, but it doesn't address the still-open cell-shift finding on the `ranges` values below — and "kept verbatim from a live dump" is hard to reconcile with what the numbers actually say.

`bootbus` is `#address-cells = <2>` / `#size-cells = <1>` under a `#address-cells = <2>` parent, so each row is `<cs offset> <parent-addr-hi parent-addr-lo> <size>`. Lining the rows up against [octeon_3xxx.dtsi:119-126](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/boot/dts/cavium-octeon/octeon_3xxx.dtsi#L119-L126) (6.18, the target's `KERNEL_PATCHVER`):

| this file | dtsi row it corresponds to |
|---|---|
| `<1 0 0x00 0x10000 0x30000000>` | `<1 0 0x10000 0x30000000 0>` |
| `<5 0 0x00 0x10000 0x40000000>` | `<2 0 0x10000 0x40000000 0>` |
| `<6 0 0x00 0x10000 0x50000000>` | `<3 0 0x10000 0x50000000 0>` |
| `<7 0 0x00 0x10000 0x90000000>` | `<7 0 0x10000 0x90000000 0>` |

Every one of those four is the dtsi row with its last three cells shifted right by one — the trailing `0` size dropped and a `0x00` prepended. So CS1 ends up mapped at physical `0x0000_0000_0001_0000` with a 0x30000000-byte window instead of at `0x00010000_30000000` with size 0. The CS→address reshuffle (rows 2/3 taking `0x1d040000`/`0x1d050000`) does look like a genuine board difference and matches the new `cs-config@2`/`@3` you added, so only the cell alignment is wrong. The sibling [cn6130_cisco_vedge1000.dts:183-190](``https://github.com/openwrt/openwrt/blob/3c33550b068a6865bc225fc857aed7418359af28/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6130_cisco_vedge1000.dts#L183-L190),`` which is also derived from a live tree, writes exactly the `<N 0 0x10000 0xADDR 0x00>` form.

While here: the continuation lines 180-186 are still not indented to line up under the first entry, unlike every other multi-line `ranges` in this directory.

---

target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
```diff
@@ -0,0 +1,302 @@
...
+				cavium,t-ce = <50>;
+				cavium,t-oe = <50>;
+				cavium,t-we = <35>;
+				cavium,t-rd-hld = <25>;
+				cavium,t-wr-hld = <35>;
+				cavium,t-wait = <300>;
+				cavium,t-page = <25>;
+			};
+
+			/* cs2/cs3 do not exist in the dtsi - new chip selects */
```
openwrt-ai: nit: this dedup pass overrides `cavium,cs-config@0`/`@4` and adds `@2`/`@3`, but leaves the dtsi's [`cavium,cs-config@5`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/boot/dts/cavium-octeon/octeon_3xxx.dtsi#L162) and [`@6`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/boot/dts/cavium-octeon/octeon_3xxx.dtsi#L179) in the compiled tree. Those carry the generic EdgeRouter-ish timings for chip selects this board's `ranges` now maps elsewhere, so they'd program bootbus timing registers for CS5/CS6 that nothing on this board uses. The sibling [cn6130_cisco_vedge1000.dts:210-214](``https://github.com/openwrt/openwrt/blob/3c33550b068a6865bc225fc857aed7418359af28/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6130_cisco_vedge1000.dts#L210-L214)`` explicitly `/delete-node/`s the inherited cs-config nodes it doesn't want. Was leaving `@5`/`@6` intentional, or should they be deleted the same way?

---

openwrt-ai (commented):
## Commit checks

- `3c33550b` "octeon: update cn6120_ubnt_usg-pro-4.dts" — the body lists three things (dedup, delete `&usbn`, formatting), but the diff also silently reverts [`0d0d8038`](https://github.com/openwrt/openwrt/commit/0d0d80389b757aa08723eee28adb25b341accc7f) from earlier in this same series: `"mmc-spi-slot"` is re-added to the `mmc-slot@0` compatible and the space after the label colon is removed again. See the inline comment. Introduce-then-fix-then-reintroduce within one series is worth avoiding — another reason to squash this branch into a single "add support" commit before merge.

The valuable part of this commit — converting the `&soc@0` / `&uart0:` / `&twsi0` / `&bootbus` reference-override syntax back to plain path-merged node names — does fix the DTS syntax problem flagged in the previous review, so the file should compile now (modulo the `ranges` semantics below). The `/delete-node/ &usbn;` is well-formed: `usbn` is labelled in [octeon_3xxx.dtsi:218](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/boot/dts/cavium-octeon/octeon_3xxx.dtsi#L218) and nothing else references it.

CI on `3c33550b` has no failures (FormalityCheck and "Check Kernel patches" green; the octeon/generic kernel build is still queued), so nothing here is CI-derived.

---

target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
```diff
@@ -0,0 +1,297 @@
...
+			 */
+
+			ranges = <0 0 0x00 0x1f400000 0xc00000>,
+            	 <1 0 0x00 0x10000 0x30000000>,
+            	 <2 0 0x00 0x1d040000 0x10000>,
+            	 <3 0 0x00 0x1d050000 0x10000>,
+            	 <4 0 0x00 0x1d020000 0x10000>,
+            	 <5 0 0x00 0x10000 0x40000000>,
+            	 <6 0 0x00 0x10000 0x50000000>,
+            	 <7 0 0x00 0x10000 0x90000000>;
```
openwrt-ai: The reformat in this commit made the indentation worse rather than fixing it. The seven continuation lines are now `12 spaces + TAB + space`, while line 171 and every other line in this file are tab-indented. Mixed leading spaces/tabs means the block only lines up at one specific tab width and lines up under nothing at the usual 8. The base [octeon_3xxx.dtsi:119-126](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/boot/dts/cavium-octeon/octeon_3xxx.dtsi#L119-L126) and the sibling [cn6130_cisco_vedge1000.dts:183-190](``https://github.com/openwrt/openwrt/blob/ef58cd1f510814cd1aa8439f432e2acc8de081c4/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6130_cisco_vedge1000.dts#L183-L190)`` both use `TAB TAB TAB TAB space` on the continuations, which puts them exactly under the first `<`.

The suggestion below also carries the cell-shift fix from the [still-open thread on this property](https://github.com/openwrt/openwrt/pull/24786#discussion_r3827564339) (rows 1/5/6/7 have their last three cells shifted right by one) so the whole block can be applied in one go — the CS→address reshuffle is left as you have it.

```suggestion
			ranges = <0 0 0x00 0x1f400000 0xc00000>,
				 <1 0 0x10000 0x30000000 0x00>,
				 <2 0 0x00 0x1d040000 0x10000>,
				 <3 0 0x00 0x1d050000 0x10000>,
				 <4 0 0x00 0x1d020000 0x10000>,
				 <5 0 0x10000 0x40000000 0x00>,
				 <6 0 0x10000 0x50000000 0x00>,
				 <7 0 0x10000 0x90000000 0x00>;
```

---

openwrt-ai (commented):
## Commit checks

- `ef58cd1f` "octeon: update cn6120_ubnt_usg-pro-4.dts" — bullet 2, "Fix bootbus ranges formatting", doesn't hold: the reformat swapped tab indentation on the seven `ranges` continuation lines for `12 spaces + TAB + space`, which is a regression against both the base dtsi and the sibling octeon DTS files. See the inline comment (its suggestion also folds in the still-unaddressed cell-shift fix so both can be applied at once).

- `ef58cd1f` — typo in bullet 1: `cavicum` -> `cavium`.

- Series hygiene, third mention: `"mmc-spi-slot"` has now been removed in [`0d0d8038`](https://github.com/openwrt/openwrt/commit/0d0d80389b757aa08723eee28adb25b341accc7f), reintroduced in [`3c33550b`](https://github.com/openwrt/openwrt/commit/3c33550b068a6865bc225fc857aed7418359af28) and removed again here, all inside one branch. Squashing this into a single "add support" commit — carrying the hardware-specification / flash-instructions / MAC-address-layout sections requested in the first review, none of which have landed yet — would erase all of that churn in one step.

The rest of `ef58cd1f` matches its message. The `/delete-node/ cavium,cs-config@5;` / `@6` additions do resolve the earlier question about inherited CS timings, and they match the pattern the sibling [cn6130_cisco_vedge1000.dts:210-214](``https://github.com/openwrt/openwrt/blob/ef58cd1f510814cd1aa8439f432e2acc8de081c4/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6130_cisco_vedge1000.dts#L210-L214)`` uses; the comma in the node name is fine for `dtc` (`PROPNODECHAR` includes `,`). Dropping the stale in-body USB comment is correct too.

CI on `ef58cd1f` is fully green — FormalityCheck, "Check Kernel patches (octeon, generic)" and the octeon/generic kernel build all passed — so nothing above is CI-derived.

---

target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6120_ubnt_usg-pro-4.dts
```diff
@@ -0,0 +1,297 @@
...
+
+		bootbus: bootbus@1180000000000 {
+			/* Values sourced from a live devicetree dump on this
+			 * board's EdgeRouter Pro-based OpenWrt install, so
+			 * kept verbatim even though several entries differ
+			 * from the dtsi's generic ones. A DT property is
+			 * replaced as a whole (not merged entry-by-entry), so
+			 * the full list has to stay here rather than being
+			 * trimmed to just the differing lines.
+			 */
```
openwrt-ai, line 174: nit: this comment block (lines 162-169) no longer describes the property below it. It says the values are "kept verbatim" from the live dump, but this commit re-derived rows 1, 5, 6 and 7 into the correct `<cs offset> <parent-addr-hi> <parent-addr-lo> <size>` cell layout — they are no longer the raw dump values. As written the comment tells the next reader not to touch numbers that were in fact already corrected.

Suggest rewording to something like "chip-select-to-address assignment taken from a live devicetree dump on this board's EdgeRouter Pro-based OpenWrt install; several entries differ from the dtsi's generic ones. A DT property is replaced as a whole (not merged entry-by-entry), so the full list has to stay here rather than being trimmed to just the differing lines." — i.e. keep the "why the full list is repeated" justification, drop the "verbatim" claim.

---

openwrt-ai (commented):
## Commit checks

- `58636aeb` "octeon: update cn6120_ubnt_usg-pro-4.dts" — the message says "Fix ranges formatting to match base and sibling dts files", but the diff does more than reformat: rows 1, 5, 6 and 7 also change value. `<1 0 0x00 0x10000 0x30000000>` becomes `<1 0 0x10000 0x30000000 0x00>` (and likewise for 5/6/7), which moves CS1 from a bogus mapping at physical `0x0000_0000_0001_0000` with a 0x30000000-byte window to `0x00010000_30000000` with size 0. That is the correct fix — it now matches the cell layout `bootbus` declares (`#address-cells = <2>` / `#size-cells = <1>` under a 2-cell parent) and the base [octeon_3xxx.dtsi:119-126](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/mips/boot/dts/cavium-octeon/octeon_3xxx.dtsi#L119-L126) — but it is a semantic change, not formatting, and the message should say so. Rows 2/3/4 really are whitespace-only, and the new `TAB TAB TAB TAB space` continuation indent matches the dtsi and [cn6130_cisco_vedge1000.dts:183-190](``https://github.com/openwrt/openwrt/blob/58636aeb9b5152dd5e0df23195382ac96489b2f8/target/linux/octeon/files/arch/mips/boot/dts/cavium-octeon/cn6130_cisco_vedge1000.dts#L183-L190)`` exactly.

- Still outstanding from earlier reviews, since this branch is now at 11 commits: the introducing commit `521890cc` carries none of the three sections a new-device commit needs — **hardware specification**, **flash instructions**, **MAC address layout**. The PR description has most of the first two (though its numbered steps still skip 3); the MAC layout is nowhere. Squashing the branch into one "octeon: add support for Ubiquiti UniFi Security Gateway Pro (USG-Pro-4)" commit carrying those three sections would resolve this and the accumulated `update …` churn in one step.

CI on `58636aeb` is green (all three FormalityCheck runs passed; no build job has been queued yet on this head), so nothing above is CI-derived.

---

openwrt-ai (commented):

---

PR #24787, opened 2026-08-18
microchipsw: use upstreamed QSPI and TAS patches

robimarko:
Use the upstreamed QSPI and TAS patches instead of the local patches.


robimarko:
Manually merged

PR #11113, opened 2022-10-30
Kernel: add packages for common USB DVB adapters

ghost:
DVB drivers are intended to be used with tvheadend package and easily available Astrometa USB DVB/analog tuners. More adapters can be added in the future.
This allows users to record and stream TV programs from DVB networks to tvheadend clients.

Major changes:
- Add new kernel module category: Multimedia
- Moved video cameras from Video to Multimedia. Video menu will contain only output devices.
- Add packages for these kernel modules:
  - Multimedia support
  - DVB USB (installs dvb-core and dvb-usb-v2)
  - Realtek RTL28xxU, RTL2830, RTL2832
  - Rafael Micro R820T
  - Panasonic MN88472 (no firmware)
  - Panasonic MN88473 (no firmware)
  - Sony CXD2841ER

Minor changes:
- USB audio needed a dependecy on kmod-multimedia, when kmod-multimedia is also selected.
- New required kernel config options were added for both kernel versions.

Tested on Linksys WRT-1900ACS v2 with OpenWrt git master, tvheadend package, Astrometa USB-DVB adapter and DVB-C TV source.

Signed-off-by: Marius Dinu <m95d+git@psihoexpert.ro>

PR #2620, opened 2019-12-16
kernel: add PCTV 461e DVB-S2 USB card support

psyborg55:
This changeset adds support for PCTV 461e DVB-S2 USB card by enabling related kernel modules and their dependencies.
The card has been tested on latest ath79 target source (Archer C7) with tvheadend as a streaming app. CPU load on router was in range 0.05-0.25 while streaming single h264 full HD channel.
Remote controller has not been tested.

Signed-off-by: Tomislav Požega <pozega.tomislav@gmail.com>

PR #1451, opened 2018-10-08
kernel: Initial DVB support

miska:
Add support for building some DVB modules. Not everything at the moment, plenty
is unsupported, but the list can grow over the time.

Signed-off-by: Michal Hrusecky <michal.hrusecky@nic.cz>

PR #24788, opened 2026-08-18
kernel: add initial DVB support

BKPepe:
This adds kernel packages for DVB adapters, frontends and tuners.

The packages come from the Turris OS build system, where they have been maintained since 2018. This is a follow-up of #1451 and of #11113, where I promised to revive #1451 and finish upstreaming it.

The review comments from #1451 are addressed here:

* The kernel symbols of the drivers which are not packaged are set in `target/linux/generic/config-6.18` instead of being passed through the `KCONFIG` of `kmod-dvb-core`, as @hauke asked for. Most of them are already in the generic config today, the remaining ones are added by this pull request.
* `kmod-v4l2` and `kmod-tveeprom` appeared outside of the menu structure. `kmod-v4l2` is dropped, because `kmod-video-core` already provides `videodev.ko`, and `kmod-tveeprom` is placed in the Video Support submenu.
* The recursive dependency of `kmod-dvb-tuner-dib0070` is gone. `make defconfig` selects all 78 packages and none of them is dropped for unmet dependencies.

Verification against kernel 6.18:

* All 26 module paths referenced by the packages exist in the kernel tree.
* Of the 128 kernel symbols which the packages disable, five do not exist anymore. `CONFIG_MEDIA_RC_SUPPORT`, `CONFIG_VIDEO_V4L2`, `CONFIG_DVB_HD29L2`, `CONFIG_DVB_USB_FRIIO` and `CONFIG_IR_SIR` are removed.
* There is no package name collision with the existing packages and the only file collision was the `videodev.ko` mentioned above.
* Enabling `DVB_CORE` and `RC_CORE`, which nothing in the tree has enabled so far, makes parts of the media driver menu visible that the generic config has no answer for, so `syncconfig` stops for input. 98 entries are added to cover that, including the symbols the new packages own themselves, so that a partial package selection configures as well.
* Build tested on mpc85xx with kernel 6.18.44: a full `CONFIG_ALL_KMODS` build of all 78 packages, and single package selections (`kmod-dvb-usb-ttusb2`, `kmod-video-em28xx`), each with no unanswered kconfig symbols left.

cc @adschm, who closed the previous pull requests including #2620


target/linux/generic/config-6.18
```diff
@@ -1955,7 +1955,32 @@ CONFIG_DVB_MAX_ADAPTERS=16
...
+# CONFIG_DVB_USB_MXL111SF is not set
+# CONFIG_DVB_USB_NOVA_T_USB2 is not set
+# CONFIG_DVB_USB_OPERA1 is not set
+# CONFIG_DVB_USB_PCTV452E is not set
+# CONFIG_DVB_USB_TECHNISAT_USB2 is not set
+# CONFIG_DVB_USB_UMT_010 is not set
 # CONFIG_DVB_USB_V2 is not set
+# CONFIG_DVB_USB_VP702X is not set
+# CONFIG_DVB_USB_VP7045 is not set
+# CONFIG_DVB_USB_ZD1301 is not set
```
openwrt-ai, line 2007: Kernel build fails on every target in this run (airoha/an7581, sifiveu/generic, armsr/armv7, octeon/generic, bcm27xx/bcm2708, bcm47xx/generic, malta/be) with an interactive `syncconfig` prompt:

```
  Cisco Cobalt support (VIDEO_COBALT) [N/m/?] n
  Conexant cx23418 MPEG encoder support (VIDEO_CX18) [N/m/?] (NEW) make[7]: *** [scripts/kconfig/Makefile:85: syncconfig] Error 1
make[6]: *** [Makefile:754: syncconfig] Error 2
make[5]: *** [.../linux-6.18.44/Makefile:849: include/config/auto.conf.cmd] Error 2
```

Turning on `CONFIG_MEDIA_SUPPORT=m` from the new packages makes the whole `drivers/media` menu visible, and this config file has no answer for some of the symbols that appear. `CONFIG_VIDEO_CX18` is the first one kconfig trips over; `CONFIG_VIDEO_CX23885` is missing too (its neighbours `VIDEO_BT848`, `VIDEO_COBALT`, `VIDEO_CX88`, `VIDEO_SAA7134` are all already present, so it looks like the list just went stale). Please add the missing `# CONFIG_... is not set` entries so a non-interactive `syncconfig` completes with `MEDIA_SUPPORT` enabled — best done by building one of these targets with `CONFIG_ALL_KMODS=y` and folding whatever kconfig still asks for back into this file.

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1086 @@
...
+define KernelPackage/media-tuner-tda827x/description
+ A DVB-T silicon tuner module. Say Y when you want to support this tuner.
+endef
+
+$(eval $(call KernelPackage,media-tuner-tda827x))
+
+define KernelPackage/media-tuner-tda8290
+  TITLE:=TDA 8290/8295 + 8275(a)/18271 tuner combo
+  DEPENDS:=+kmod-media-tuner-tda827x +kmod-media-tuner-tda18271
+  $(call MediaTuner,tda8290,CONFIG_MEDIA_TUNER_TDA8290)
```
openwrt-ai: `DEPENDS` is set here *before* `$(call MediaTuner,...)`, and `MediaTuner` uses a plain assignment (`DEPENDS:=+kmod-i2c-core`, line 665), so both tuner dependencies are silently discarded — `kmod-media-tuner-tda8290` ends up depending only on `kmod-i2c-core`. `tda8290.c` pulls in symbols from `tda827x` and `tda18271`, so the module will fail to load without them. Every other package in this file appends after the call; do the same here.

```suggestion
  $(call MediaTuner,tda8290,CONFIG_MEDIA_TUNER_TDA8290)
  DEPENDS+=+kmod-media-tuner-tda827x +kmod-media-tuner-tda18271
```

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1086 @@
...
+# ------------------------------- Media tuners --------------------------------
+# -----------------------------------------------------------------------------
+
+TUNER_MENU:=Media tuners
+
+define MediaTuner
+  SUBMENU:=$(TUNER_MENU)
+  KCONFIG:= \
+	CONFIG_MEDIA_SUPPORT=m \
+	CONFIG_MEDIA_DIGITAL_TV_SUPPORT=m \
```
openwrt-ai: `MEDIA_DIGITAL_TV_SUPPORT` is a `bool`, not a tristate — see [drivers/media/Kconfig:90-91](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/Kconfig#L90-L91). Setting it to `m` makes kconfig reject the value ("symbol value 'm' invalid"). Every other use in this file (lines 43 and 88) already uses `=y`.

```suggestion
	CONFIG_MEDIA_DIGITAL_TV_SUPPORT=y \
```

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1086 @@
...
+define KernelPackage/video-em28xx
+  TITLE:=Empia EM28xx USB devices support
+  KCONFIG:= \
+	CONFIG_VIDEO_EM28XX=m \
+	CONFIG_VIDEO_EM28XX_V4L2=m \
+	CONFIG_VIDEO_EM28XX_ALSA=m \
+	CONFIG_VIDEO_EM28XX_DVB=m \
+	CONFIG_VIDEO_EM28XX_RC=m
+  FILES:=$(LINUX_DIR)/drivers/media/usb/em28xx/em28xx.ko
+  AUTOLOAD:=$(call AutoProbe,video-em28xx)
```
openwrt-ai: `AutoProbe` takes module names, not package names. The file installed on line 1060 is `em28xx.ko`, so the generated `/etc/modules.d` entry `video-em28xx` will make `modprobe` fail at boot.

```suggestion
  AUTOLOAD:=$(call AutoProbe,em28xx)
```

Same problem on line 1076: `em28xx-dvb.ko` is autoloaded as `video-em28xx-dvb`.

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1086 @@
...
+define KernelPackage/video-em28xx/description
+  Empia EM28xx USB devices support
+endef
+
+$(eval $(call KernelPackage,video-em28xx))
+
+define KernelPackage/video-em28xx-dvb
+  TITLE:=DVB/ATSC support for em28xx based TV cards
+  FILES:=$(LINUX_DIR)/drivers/media/usb/em28xx/em28xx-dvb.ko
+  AUTOLOAD:=$(call AutoProbe,video-em28xx-dvb)
```
openwrt-ai:
```suggestion
  AUTOLOAD:=$(call AutoProbe,em28xx-dvb)
```

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1086 @@
...
+
+define KernelPackage/video-em28xx
+  TITLE:=Empia EM28xx USB devices support
+  KCONFIG:= \
+	CONFIG_VIDEO_EM28XX=m \
+	CONFIG_VIDEO_EM28XX_V4L2=m \
+	CONFIG_VIDEO_EM28XX_ALSA=m \
+	CONFIG_VIDEO_EM28XX_DVB=m \
+	CONFIG_VIDEO_EM28XX_RC=m
+  FILES:=$(LINUX_DIR)/drivers/media/usb/em28xx/em28xx.ko
```
openwrt-ai, line 1071: `VIDEO_EM28XX_V4L2`, `_ALSA` and `_RC` each build their own module — `em28xx-v4l.ko`, `em28xx-alsa.ko`, `em28xx-rc.ko` ([em28xx/Makefile:8-12](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/usb/em28xx/Makefile#L8-L12)) — but `FILES` only lists `em28xx.ko`, so those three end up built and never packaged. Either add them to `FILES` (or split them into their own packages, as `em28xx-dvb` already is), or drop the symbols from `KCONFIG` if they are not wanted.

Related: `CONFIG_VIDEO_EM28XX_DVB=m` is set here, but `em28xx-dvb.ko` is packaged by `kmod-video-em28xx-dvb`, which carries no `KCONFIG` of its own — selecting only `kmod-video-em28xx` therefore builds an unpackaged `em28xx-dvb.ko` too.

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1086 @@
...
+
+$(eval $(call KernelPackage,dvb-usb-dib0700))
+
+define KernelPackage/dvb-usb-dibusb-common
+  TITLE:=Common library for DiBcom USB drivers
+  FILES:=$(LINUX_DIR)/drivers/media/usb/dvb-usb/dvb-usb-dibusb-common.ko \
+         $(LINUX_DIR)/drivers/media/usb/dvb-usb/dvb-usb-dibusb-mc-common.ko
+  AUTOLOAD:=$(call AutoProbe,dvb-usb-dibusb-common dvb-usb-dibusb-mc-common)
+  $(call AddDepends/dvb-usb,+PACKAGE_kmod-dvb-dib3000mc:kmod-dvb-dib3000mc)
+endef
```
openwrt-ai, line 152: This package has no `KCONFIG` and is user-selectable, but neither of its two files is unconditionally built. Per [dvb-usb/Makefile:21,27](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/usb/dvb-usb/Makefile#L21-L30):

```
obj-$(CONFIG_DVB_USB_DIB3000MC) += dvb-usb-dibusb-common.o dvb-usb-dibusb-mc-common.o
obj-$(CONFIG_DVB_USB_DIBUSB_MB) += dvb-usb-dibusb-common.o dvb-usb-dibusb-mb.o
```

`dvb-usb-dibusb-mc-common.ko` is only produced by `CONFIG_DVB_USB_DIB3000MC` (the dvb-usb helper symbol, not the `CONFIG_DVB_DIB3000MC` frontend that `kmod-dvb-dib3000mc` sets). `kmod-dvb-usb-dibusb-mb` (line 176) depends on this package but only turns on `DVB_USB_DIBUSB_MB`, so selecting it without `kmod-dvb-usb-a800`/`-dibusb-mc` leaves `dvb-usb-dibusb-mc-common.ko` unbuilt and the package build fails on a missing file. The `make defconfig` run in the PR description selects everything, which hides this.

Also, since the package is a pure helper with no `KCONFIG`, it should probably carry `HIDDEN:=1` like other in-tree library kmods (e.g. [`kmod-regmap-core`](https://github.com/openwrt/openwrt/blob/e15cc65157efea28d8ec4af37bf18a55faf6adf3/package/kernel/linux/modules/other.mk#L633-L636)) so users cannot select it on its own.

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1086 @@
...
+
+$(eval $(call KernelPackage,dvb-dib3000mb))
+
+define KernelPackage/dvb-dibx000-common
+  SUBMENU:=$(DVB_MENU)
+  TITLE:=Common library for DiBX000 drivers
+  FILES:=$(LINUX_DIR)/drivers/media/dvb-frontends/dibx000_common.ko
+  AUTOLOAD:=$(call AutoProbe,dibx000_common)
+endef
```
openwrt-ai, line 591: Same shape as `kmod-dvb-usb-dibusb-common`: no `KCONFIG`, but user-selectable. `dibx000_common.ko` only exists when one of `DVB_DIB3000MC/7000M/7000P/8000/9000` is enabled ([dvb-frontends/Makefile:44-48](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/dvb-frontends/Makefile#L44-L48)), so selecting `kmod-dvb-dibx000-common` alone fails on a missing file. Adding `HIDDEN:=1` matches how other library-only kmods are declared in tree and makes it reachable only through the five packages that already pull it in.

```suggestion
define KernelPackage/dvb-dibx000-common
  SUBMENU:=$(DVB_MENU)
  TITLE:=Common library for DiBX000 drivers
  HIDDEN:=1
  FILES:=$(LINUX_DIR)/drivers/media/dvb-frontends/dibx000_common.ko
  AUTOLOAD:=$(call AutoProbe,dibx000_common)
endef
```

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1086 @@
...
+$(eval $(call KernelPackage,rc-core))
+
+define KernelPackage/dvb-core
+  SUBMENU:=$(DVB_MENU)
+  TITLE:=DVB core support
+  KCONFIG:= \
+	CONFIG_MEDIA_SUPPORT=m \
+	CONFIG_MEDIA_DIGITAL_TV_SUPPORT=y \
+	CONFIG_DVB_NET=y \
+	CONFIG_DVB_MAX_ADAPTERS=8 \
```
openwrt-ai: This silently lowers the in-tree value: [`target/linux/generic/config-6.18:1885`](https://github.com/openwrt/openwrt/blob/e15cc65157efea28d8ec4af37bf18a55faf6adf3/target/linux/generic/config-6.18#L1885) already carries `CONFIG_DVB_MAX_ADAPTERS=16`, which is also the kernel default ([dvb-core/Kconfig:32-36](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/dvb-core/Kconfig#L32-L36)). Selecting `kmod-dvb-core` would override it to 8. Is the reduction deliberate (memory saving on small devices), or should this line be dropped so the generic config value applies?

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1086 @@
...
+define KernelPackage/dvb-rtl2830/description
+ Realtek RTL2830 DVB-T
+endef
+
+$(eval $(call KernelPackage,dvb-rtl2830))
+
+define KernelPackage/dvb-rtl2832
+  TITLE:=Realtek RTL2832 DVB-T
+  $(call DvbFrontend,rtl2832,CONFIG_DVB_RTL2832)
+  DEPENDS+=+kmod-i2c-mux
```
openwrt-ai: nit: `DvbFrontend` already puts `+kmod-i2c-mux` in `DEPENDS` (line 364), so this line is a no-op. Same for lines 429, 441, 453, 465 and 477.

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1086 @@
+#
+# Copyright (C) 2018-2022 CZ.NIC, z. s. p. o. <https://www.nic.cz>
+#
+# This is free software, licensed under the GNU General Public License v2.
+# See /LICENSE for more information.
+#
+
+DVB_MENU:=DVB Support
+VIDEO_MENU:=Video Support
```
openwrt-ai, line 9: nit: `VIDEO_MENU` is already defined with the same value at [`video.mk:9`](https://github.com/openwrt/openwrt/blob/e15cc65157efea28d8ec4af37bf18a55faf6adf3/package/kernel/linux/modules/video.mk#L9), and all `modules/*.mk` are included into the same make scope, so this is a redundant redefinition that will drift if the menu is ever renamed.

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1086 @@
...
+ A driver for the silicon tuner XC5000 from Xceive.
+ This device is only used inside a SiP called together with a
+ demodulator for now.
+endef
+
+$(eval $(call KernelPackage,media-tuner-xc5000))
+
+define KernelPackage/tveeprom
+  SUBMENU:=$(VIDEO_MENU)
+  TITLE:=VIDEO_TVEEPROM
```
openwrt-ai: nit: the raw kconfig symbol leaks into the menuconfig entry. Every other package here uses a human-readable title, and this one also has no `/description` block.

```suggestion
  TITLE:=TV hardware capabilities EEPROM support
```

---

openwrt-ai (commented):
Reviewed 1 new commit. Verified the module paths and all 25 newly added kconfig symbols against the v6.18.44 kernel tree — those all check out, and there are no package-name collisions with the existing `modules/*.mk`. The commit message matches the diff.

Ten inline comments below: one on the kernel build failure CI is currently reporting, plus a few dependency/autoload bugs and some nits.

---

target/linux/generic/config-6.18
```diff
@@ -1955,9 +1966,35 @@ CONFIG_DVB_MAX_ADAPTERS=16
...
```
openwrt-ai, line 2007: Twelve `CONFIG_DVB_USB_*` symbols that the new packages own are still missing from this list, so they reintroduce the same interactive-`syncconfig` failure that the rest of this hunk fixes — just for a partial package selection instead of a full one.

`DVB_USB` is turned on by `kmod-dvb-usb` and `DVB_USB_V2` by `kmod-dvb-usb-v2`, and every driver symbol under them is a prompted tristate ([`DVB_USB_A800`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/usb/dvb-usb/Kconfig#L27-L30), [`DVB_USB_RTL28XXU`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/usb/dvb-usb-v2/Kconfig#L147-L150), …). So selecting e.g. only `kmod-dvb-usb-ttusb2` gives `CONFIG_DVB_USB=m` and leaves `DVB_USB_A800`, `DVB_USB_CXUSB`, `DVB_USB_DIB0700`, `DVB_USB_DIBUSB_MB`, `DVB_USB_DIBUSB_MC` unanswered → `(NEW)` prompt → build stops. The `CONFIG_ALL_KMODS` test in the commit message can't hit this because every symbol then gets an answer from its own package. Note that every other package-owned symbol in tree does carry an entry here (`# CONFIG_USB_NET_CDCETHER is not set`, `# CONFIG_VIDEO_PVRUSB2 is not set`, `# CONFIG_ATH9K is not set`, …).

Missing: `A800`, `AF9015`, `AF9035`, `ANYSEE`, `AZ6007`, `CXUSB`, `DIB0700`, `DIBUSB_MB`, `DIBUSB_MC`, `DVBSKY`, `RTL28XXU`, `TTUSB2`.

```suggestion
# CONFIG_DVB_USB_A800 is not set
# CONFIG_DVB_USB_AF9005 is not set
# CONFIG_DVB_USB_AF9015 is not set
# CONFIG_DVB_USB_AF9035 is not set
# CONFIG_DVB_USB_ANYSEE is not set
# CONFIG_DVB_USB_AU6610 is not set
# CONFIG_DVB_USB_AZ6007 is not set
# CONFIG_DVB_USB_AZ6027 is not set
# CONFIG_DVB_USB_CE6230 is not set
# CONFIG_DVB_USB_CINERGY_T2 is not set
# CONFIG_DVB_USB_CXUSB is not set
# CONFIG_DVB_USB_CXUSB_ANALOG is not set
# CONFIG_DVB_USB_DEBUG is not set
# CONFIG_DVB_USB_DIB0700 is not set
# CONFIG_DVB_USB_DIBUSB_MB is not set
# CONFIG_DVB_USB_DIBUSB_MC is not set
# CONFIG_DVB_USB_DIGITV is not set
# CONFIG_DVB_USB_DTT200U is not set
# CONFIG_DVB_USB_DTV5100 is not set
# CONFIG_DVB_USB_DVBSKY is not set
# CONFIG_DVB_USB_DW2102 is not set
# CONFIG_DVB_USB_EC168 is not set
# CONFIG_DVB_USB_GL861 is not set
# CONFIG_DVB_USB_GP8PSK is not set
# CONFIG_DVB_USB_LME2510 is not set
# CONFIG_DVB_USB_M920X is not set
# CONFIG_DVB_USB_MXL111SF is not set
# CONFIG_DVB_USB_NOVA_T_USB2 is not set
# CONFIG_DVB_USB_OPERA1 is not set
# CONFIG_DVB_USB_PCTV452E is not set
# CONFIG_DVB_USB_RTL28XXU is not set
# CONFIG_DVB_USB_TECHNISAT_USB2 is not set
# CONFIG_DVB_USB_TTUSB2 is not set
# CONFIG_DVB_USB_UMT_010 is not set
# CONFIG_DVB_USB_V2 is not set
# CONFIG_DVB_USB_VP702X is not set
# CONFIG_DVB_USB_VP7045 is not set
# CONFIG_DVB_USB_ZD1301 is not set
```

---

target/linux/generic/config-6.18
```diff
@@ -7859,6 +7907,9 @@ CONFIG_VHOST_MENU=y
 # CONFIG_VIDEO_DW9768 is not set
 # CONFIG_VIDEO_DW9807_VCM is not set
 # CONFIG_VIDEO_EM28XX is not set
+# CONFIG_VIDEO_EM28XX_ALSA is not set
```
openwrt-ai, line 7953: `CONFIG_VIDEO_EM28XX_DVB` is missing from this group. `_ALSA`, `_RC` and `_V4L2` were added here because they become visible once `VIDEO_EM28XX=m`, and [`VIDEO_EM28XX_DVB`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/usb/em28xx/Kconfig#L38-L40) has exactly the same trigger (`depends on VIDEO_EM28XX && DVB_CORE`, and `kmod-video-em28xx` pulls `kmod-dvb-usb` → `DVB_CORE=m`). Selecting `kmod-video-em28xx` without `kmod-video-em28xx-dvb` therefore leaves it unanswered.

```suggestion
# CONFIG_VIDEO_EM28XX_ALSA is not set
# CONFIG_VIDEO_EM28XX_DVB is not set
```

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1091 @@
...
+ This device is only used inside a SiP called together with a
+ demodulator for now.
+endef
+
+$(eval $(call KernelPackage,media-tuner-xc5000))
+
+define KernelPackage/tveeprom
+  SUBMENU:=$(VIDEO_MENU)
+  TITLE:=TV hardware capabilities EEPROM support
+  KCONFIG:=CONFIG_VIDEO_TVEEPROM
```
openwrt-ai, line 1052: `CONFIG_VIDEO_TVEEPROM` is a promptless, select-only symbol — [common/Kconfig:23-25](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/common/Kconfig#L23-L25) is just `tristate` + `depends on I2C`, with no title. kconfig discards a user value for such a symbol and recomputes it from the `select`s, so this `KCONFIG` line cannot turn it on: selecting `kmod-tveeprom` on its own produces no `tveeprom.ko` and the package build fails on the missing file.

It only gets built via `select VIDEO_TVEEPROM` from a driver, here `VIDEO_EM28XX` ([em28xx/Kconfig:2-6](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/usb/em28xx/Kconfig#L2-L6)) — and `kmod-video-em28xx` already depends on this package. Making it hidden, like `kmod-dvb-dibx000-common` in this same revision, keeps it reachable only that way:

```suggestion
  SUBMENU:=$(VIDEO_MENU)
  TITLE:=TV hardware capabilities EEPROM support
  HIDDEN:=1
  KCONFIG:=CONFIG_VIDEO_TVEEPROM
```

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1091 @@
...
+$(eval $(call KernelPackage,tveeprom))
+
+
+define KernelPackage/video-em28xx
+  TITLE:=Empia EM28xx USB devices support
+  KCONFIG:=CONFIG_VIDEO_EM28XX
+  FILES:=$(LINUX_DIR)/drivers/media/usb/em28xx/em28xx.ko
+  AUTOLOAD:=$(call AutoProbe,em28xx)
+  DEPENDS+=+kmod-tveeprom +kmod-video-core
+  $(call AddDepends/dvb-usb)
```
openwrt-ai: em28xx is not a dvb-usb driver, so `AddDepends/dvb-usb` pulls in a module it never uses: nothing under `drivers/media/usb/em28xx/` references `dvb_usb` (em28xx-dvb.c talks to dvb-core directly via `dvb_demux`/`dmxdev`/`dvb_net`). The effect is that `kmod-video-em28xx` hard-depends on `kmod-dvb-usb` and forces `CONFIG_DVB_USB=m`, and `dvb-usb.ko` gets installed and autoloaded on devices that only have an em28xx capture chip. The `SUBMENU:=$(DVB_MENU)` the macro sets is overridden by the next line anyway, so the call contributes nothing else.

```suggestion
  DEPENDS+=+kmod-usb-core +kmod-i2c-core
```

Same on line 1083 for `kmod-video-em28xx-dvb` — it already carries `+kmod-dvb-core +kmod-video-em28xx`, which is the complete set.

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1091 @@
...
+  TITLE:=Common library for DiBcom USB drivers
+  HIDDEN:=1
+  FILES:=$(LINUX_DIR)/drivers/media/usb/dvb-usb/dvb-usb-dibusb-common.ko
+  AUTOLOAD:=$(call AutoProbe,dvb-usb-dibusb-common)
+  $(call AddDepends/dvb-usb)
+endef
+
+$(eval $(call KernelPackage,dvb-usb-dibusb-common))
+
+define KernelPackage/dvb-usb-dibusb-mc-common
```
openwrt-ai, line 156: nit: this new package makes 78, and the config hunk now adds 54 symbols — which matches the commit message ("Add the 54 missing symbols", "all 78 packages build"). The PR description still says "the remaining 25 are added by this pull request" and "all 77 packages build", from before the force-push. Worth refreshing so the two don't disagree.

---

openwrt-ai (commented):
Re-reviewed the force-pushed commit (`e15cc65` → `f51ca5b`). The previous round's findings are addressed: `MEDIA_DIGITAL_TV_SUPPORT` is `=y` everywhere, `CONFIG_DVB_MAX_ADAPTERS=8` is gone, the `tda8290` `DEPENDS` now appends after `$(call MediaTuner,...)`, the redundant `+kmod-i2c-mux` lines are dropped, the em28xx `AUTOLOAD` names match the installed `.ko`s, `dibx000-common` is hidden, and `dvb-usb-dibusb-common` is split so each package's `FILES` are actually produced by the symbol it sets — that split matches [dvb-usb/Makefile:19-30](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/usb/dvb-usb/Makefile#L19-L30) (`DVB_USB_A800` and `DVB_USB_DIBUSB_MC` both `select DVB_USB_DIB3000MC`, which builds both helper objects). The commit message matches the diff (54 config lines, 78 packages).

Four things in the new revision, all verified against v6.18.44 (`gregkh/linux` @ `1efe5d0`):

- The kconfig gap that broke CI is only fixed for a full `CONFIG_ALL_KMODS` selection — twelve `DVB_USB_*` symbols and `VIDEO_EM28XX_DVB` still have no `# ... is not set` entry, so a partial selection hits the same interactive `syncconfig` stop.
- `kmod-tveeprom` sets a promptless select-only symbol, so it cannot build standalone.
- `kmod-video-em28xx{,-dvb}` depend on `kmod-dvb-usb`, which em28xx does not use.
- One nit about the PR description being stale relative to the amended commit.

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1093 @@
...
+ Kernel module to read the EEPROM found on TV tuner cards, which
+ describes the hardware capabilities of the card.
+endef
+
+$(eval $(call KernelPackage,tveeprom))
+
+
+define KernelPackage/video-em28xx
+  TITLE:=Empia EM28xx USB devices support
+  KCONFIG:=CONFIG_VIDEO_EM28XX
```
openwrt-ai: Dropping `$(call AddDepends/dvb-usb)` from this package also dropped the only thing that set `CONFIG_MEDIA_USB_SUPPORT=y` for it — `kmod-dvb-usb` carries that symbol in its own `KCONFIG` block.

`drivers/media/usb/em28xx/Kconfig` is only sourced from inside the `if MEDIA_USB_SUPPORT` region: [`if MEDIA_USB_SUPPORT` at drivers/media/usb/Kconfig:11](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/usb/Kconfig#L11) and [the `source` at drivers/media/usb/Kconfig:55-58](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/usb/Kconfig#L55-L58). [`kmod-video-core` (video.mk:811-815)](https://github.com/openwrt/openwrt/blob/7a1820d2369e7bcb16c425421d8733b052d69af7/package/kernel/linux/modules/video.mk#L811-L815) supplies `MEDIA_CAMERA_SUPPORT=y` and `VIDEO_DEV` but not `MEDIA_USB_SUPPORT`, and `config-6.18` keeps `# CONFIG_MEDIA_USB_SUPPORT is not set`.

So a selection of just `kmod-video-em28xx` (and `kmod-video-em28xx-dvb`, which reaches USB support only through this package) leaves `CONFIG_VIDEO_EM28XX` unreachable, `em28xx.ko` is never built, and the `FILES` entry fails. A `CONFIG_ALL_KMODS` build hides this because `kmod-dvb-usb` is selected anyway. If the standalone `kmod-video-em28xx` build mentioned in the PR description was run against the previous revision, it is worth re-running.

```suggestion
  KCONFIG:= \
	CONFIG_MEDIA_USB_SUPPORT=y \
	CONFIG_VIDEO_EM28XX
```

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1093 @@
...
+# --------------------------- DVB USB v2 drivers ------------------------------
+
+define KernelPackage/dvb-usb-v2
+  TITLE:=Support for various USB DVB devices v2
+  KCONFIG:= \
+	CONFIG_MEDIA_USB_SUPPORT=y \
+	CONFIG_DVB_USB_V2
+  FILES:=$(LINUX_DIR)/drivers/media/usb/dvb-usb-v2/dvb_usb_v2.ko
+  AUTOLOAD:=$(call AutoProbe,dvb_usb_v2)
+  DEPENDS:=+kmod-usb-core +kmod-i2c-core +kmod-dvb-core +PACKAGE_kmod-rc-core:kmod-rc-core
```
openwrt-ai: nit: `$(call AddDepends/dvb-core,+kmod-usb-core)` on the next line appends `+kmod-dvb-core +kmod-usb-core`, so both end up listed twice in the final `DEPENDS`. Only the i2c and the optional rc-core entries need spelling out here.

```suggestion
  DEPENDS:=+kmod-i2c-core +PACKAGE_kmod-rc-core:kmod-rc-core
```

---

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1093 @@
...
+  SUBMENU:=$(DVB_MENU)
+  DEPENDS+=+kmod-dvb-core $1
+endef
+
+# ----------------------------- utility drivers -------------------------------
+
+define KernelPackage/cypress-firmware
+  SUBMENU:=$(DVB_MENU)
+  TITLE:=Cypress firmware helper routines
+  KCONFIG:=CONFIG_CYPRESS_FIRMWARE
```
openwrt-ai, line 70: `CONFIG_CYPRESS_FIRMWARE` is a promptless, select-only symbol ([drivers/media/common/Kconfig:9-11](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/common/Kconfig#L9-L11)) — the same situation `kmod-tveeprom` had, which this revision fixed with `HIDDEN:=1`. Selecting `kmod-cypress-firmware` on its own leaves the symbol at `n`, so `cypress_firmware.ko` is never built and the `FILES` entry fails. Its only consumer in this file is `kmod-dvb-usb-af9015` (line 345), which already pulls it in via `AddDepends/dvb-usb-v2`, so hiding it costs nothing.

```suggestion
  HIDDEN:=1
  KCONFIG:=CONFIG_CYPRESS_FIRMWARE
```

---

openwrt-ai (commented):
Re-reviewed the amended commit (`f51ca5b` → `7a1820d`). The previous round's kconfig gap is closed: all 44 newly added entries in `config-6.18` exist as real symbols in v6.18.44 (`gregkh/linux` @ `1efe5d0`), the file stays free of duplicates and keeps its byte-order sorting, and sweeping every prompted symbol under `drivers/media/` against the new config leaves nothing reachable from `MEDIA_DIGITAL_TV_SUPPORT`/`RC_CORE`/`MEDIA_USB_SUPPORT` unanswered — the remaining unlisted ones all sit behind `MEDIA_PCI_SUPPORT`, `MEDIA_PLATFORM_DRIVERS`, `MEDIA_ANALOG_TV_SUPPORT`, CEC or radio menus that stay `n`. `kmod-tveeprom` is now `HIDDEN` and `VIDEO_EM28XX` does `select VIDEO_TVEEPROM`, so that one resolves correctly. The commit message tracks the diff (98 config lines, matching the actual `+` count).

Three things in this revision, all verified against v6.18.44:

- Removing `$(call AddDepends/dvb-usb)` from `kmod-video-em28xx` also removed the only source of `CONFIG_MEDIA_USB_SUPPORT=y`, which gates the em28xx Kconfig — the em28xx packages now cannot build on their own.
- `kmod-dvb-usb-v2` lists `+kmod-usb-core` and `+kmod-dvb-core` twice.
- `kmod-cypress-firmware` has the same promptless select-only symbol problem that `kmod-tveeprom` was just fixed for.

CI is green on `7a1820d`.

---

geldot:
That's quite a lot of new kmods. I think it's worth stating what userspace service can make use of DVB tuners, even if just one sentence. I imagine there's a daemon to stream the feed to network clients?

openwrt-ai (commented):

Re-reviewed the amended commit (`7a1820d` → `3634d69`, three hunks in `dvb.mk`). All three findings from the previous round are fixed, and the commit message is unchanged and still matches the diff. Verified against v6.18.44 (`gregkh/linux` @ `1efe5d0`):

- `kmod-cypress-firmware` is now `HIDDEN:=1`. Its only consumer in this file, `kmod-dvb-usb-az6007`, pulls it in explicitly, and `DVB_USB_AZ6007` is indeed the symbol carrying [`select CYPRESS_FIRMWARE` (dvb-usb-v2/Kconfig:75-78)](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/usb/dvb-usb-v2/Kconfig#L75-L78), so hiding it costs nothing.
- `kmod-dvb-usb-v2` no longer lists `+kmod-usb-core` / `+kmod-dvb-core` twice — `$(call AddDepends/dvb-core,+kmod-usb-core)` supplies both. Keeping `kmod-rc-core` optional is correct here: [`DVB_USB_V2` depends on `(RC_CORE || RC_CORE=n)`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/usb/dvb-usb-v2/Kconfig#L2-L4), unlike `DVB_USB`, which hard-depends on `RC_CORE` and correspondingly gets a hard `+kmod-rc-core`.
- `kmod-video-em28xx` now sets `CONFIG_MEDIA_USB_SUPPORT=y` itself, which is what makes `VIDEO_EM28XX` reachable — em28xx's Kconfig is sourced inside [`if MEDIA_USB_SUPPORT` … (usb/Kconfig:11, 55-58)](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/media/usb/Kconfig#L55-L58), and the surrounding `(MEDIA_CAMERA_SUPPORT || MEDIA_ANALOG_TV_SUPPORT || MEDIA_DIGITAL_TV_SUPPORT)` guard is satisfied by `MEDIA_CAMERA_SUPPORT=y` from `kmod-video-core`. `kmod-video-em28xx-dvb` inherits the symbol through its dependency on this package.

CI is green on `3634d69` — 179 jobs in the kernel workflow plus the core-package builds, no non-success conclusions.

---

One correction to my earlier round, informational only, **no change needed**: I asked for `# CONFIG_DVB_USB_* is not set` entries in `config-6.18` for symbols the new packages own themselves. That was over-cautious. [`gen_kconfig_overrides()` in scripts/package-metadata.pl:82-87](https://github.com/openwrt/openwrt/blob/3634d69d63d28d88710f8ab713511584a2ef68cb/scripts/package-metadata.pl#L82-L87) already emits `# CONFIG_X is not set` for every `Kernel-Config` symbol of an *unselected* package, as long as the entry carries no explicit `=value` — so a bare `KCONFIG:=CONFIG_DVB_USB_A800` is answered automatically on a partial selection. Only the `=y`-style entries (`MEDIA_USB_SUPPORT`, `MEDIA_DIGITAL_TV_SUPPORT`, …) and symbols no package owns genuinely need an answer in the generic config. The extra lines are harmless and match what plenty of in-tree package-owned symbols already do, so please leave them as they are rather than churning the file again.

---

BKPepe:
@adschm , @Hauke, @robimarko What do you think? I think I addressed all the feedback from @openwrt-ai, so this one should be ready.

geldot:
> @adschm , @hauke, @robimarko What do you think? I think I addressed all the feedback from @openwrt-ai, so this one should be ready.

There was also some non-AI feedback: https://github.com/openwrt/openwrt/pull/24788#issuecomment-5350790224

BKPepe:
> I think it's worth stating what userspace service can make use of DVB tuners, even if just one sentence. I imagine there's a daemon to stream the feed to network clients?

Sure, you can use them with e.g. [tvheadend](https://github.com/openwrt/packages/blob/aca7007595f4a5d468938cf8e1fea889ba7b5067/multimedia/tvheadend/Makefile) from the packages repository.

openwrt-ai (commented):

---

package/kernel/linux/modules/dvb.mk
openwrt-ai: fixed, thanks — solved by dropping `VIDEO_EM28XX_V4L2`/`_ALSA`/`_RC`/`_DVB` from this package's `KCONFIG` entirely.

---

package/kernel/linux/modules/dvb.mk
openwrt-ai: fixed, thanks — solved by splitting `dvb-usb-dibusb-mc-common` into its own package and marking both `HIDDEN:=1`.

---

package/kernel/linux/modules/dvb.mk
openwrt-ai: fixed, thanks

---

package/kernel/linux/modules/dvb.mk
openwrt-ai: fixed, thanks

---

package/kernel/linux/modules/dvb.mk
openwrt-ai: fixed, thanks

---

geldot:
> > I think it's worth stating what userspace service can make use of DVB tuners, even if just one sentence. I imagine there's a daemon to stream the feed to network clients?
> 
> Sure, you can use them with e.g. [tvheadend](https://github.com/openwrt/packages/blob/aca7007595f4a5d468938cf8e1fea889ba7b5067/multimedia/tvheadend/Makefile) from the packages repository.

Thanks! LGTM FWIW.

BKPepe:
@dangowrt Any thoughts? :)

package/kernel/linux/modules/dvb.mk
```diff
@@ -0,0 +1,1096 @@
...
+VIDEO_MENU:=Video Support
+
+
+# ------------------------------ core drivers ---------------------------------
+
+define KernelPackage/rc-core
+  SUBMENU:=$(DVB_MENU)
+  TITLE:=Remote Controller support
+  KCONFIG:= \
+	CONFIG_MEDIA_SUPPORT=m \
```
dangowrt, line 18: I don't think the explicit `=m` is needed here. Only the symbol name alone should be sufficient. I'd also kind of want to see Remote Controller support in `input.mk` or at least within the Input devices menu as this is not only useful in the context of DVB.

dangowrt (approved):
Minor nits regarding the use of `=m` in kmod packages as well as listing RC support in the DVB/Media menu instead of Input where it would belong imho. Otherwise looks good to me, and happy to see this new use-case landing. DVB is also useful beyond watching TV -- it can provide (not very accurate, but often good enough) time synchronization and also enable the use of data-over-satellite services (I doubt this is still relevant in 2026, but still)

package/kernel/linux/modules/dvb.mk
BKPepe: Okay, changed. :)

openwrt-ai (commented):

---

robimarko:

PR #24186, opened 2026-07-11
qualcommax: pcs: qca-uniphy: MAC mode and configure-once fixes

JuliusBairaktaris:
Three UNIPHY PCS fixes for the qualcommax target, sent together because they touch the same file and the same code path.

* **take a channel's clock pair as a bulk** — `pcs_enable()` leaves the RX clock enabled when the TX clock fails, and nothing rebalances the pair afterwards; `clk_bulk_enable()` unwinds what it enabled.
* **set MAC mode for PSGMII and QSGMII** — the channel mode field was left at its 1000BASE-X value.
* **configure the UNIPHY instance once per mode** — the instance is shared between the ports on it, so reconfiguring it on every link event disturbs the other ports that share it. Skipping a same-mode reconfiguration leaves a genuine mode change still doing the work. The cached mode is cleared before the sequence starts and written back only after its last step, so a bail-out anywhere in between leaves nothing cached that the hardware does not have; the channel's clock pair is restored on the way out, and a per-channel flag keeps a failed re-enable from making the next call disable an already-disabled clock. A per-instance mutex serialises the channels, which otherwise each arrive under their own phylink state lock.

Runtime tested on a Xiaomi AX3600 (IPQ8074, QCA8075 PSGMII, kernel 6.18): three boots, all four ports up with RX and TX counters moving, no clock warnings in `dmesg`, and one port swept 1000 → 100 → 10 Mbps with traffic at each step and the other channels of the instance untouched. Not reachable on that board and not exercised: the calibration bail-out, the XPCS reset paths, and the IPQ5018-only `pcs_enable()` / `pcs_disable()` sites.

Supersedes #24187, which is folded in here.

PR #24791, opened 2026-08-19
qualcommax: pcs: qca-uniphy: force the channel on every UNIPHY

JuliusBairaktaris:
A fixed link has no in-band word for the UNIPHY channel to take its speed from,
so the channel has to be forced. The driver does that on IPQ5018 only. It never
writes the speed field beside the force bit either, and that field is where a
forced channel reads its speed.

    target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c,
    qca_uniphy_pcs_config_mode(), before this series:

        /* TODO: Fix for IPQ6018 and IPQ8074 */
        if (uniphy->data->uniphy_type == UNIPHY_IPQ5018) {
                //set force mode for fixed link
                if (neg_mode == PHYLINK_PCS_NEG_OUTBAND && !phylink_expects_phy(pcs->phylink)) {
                        regmap_set_bits(uniphy->regmap,
                                        UNIPHY_CH_CTRL(upcs->channel),
                                        UNIPHY_CH_FORCE_MODE);
                }
        }

qca-ssdk writes that same UNIPHY_CH_CTRL bit on all three SoCs, so my
understanding is there is nothing SoC-specific here. The last commit drops the
SoC test, clears the bit for a port that is not a fixed link, and programs
UNIPHY_CH_SPEED_MODE from the speed `pcs_link_up()` is handed. Neither write is
made in USXGMII or 10GBASE-R, where the XPCS drives the channel.

That needs a PSGMII port to report an outband neg_mode, and it does not:
`phylink_get_inband_type()` omits PSGMII, so phylink reports
`PHYLINK_PCS_NEG_NONE` for it. The second commit backports 5ba017f9efef ("net:
phylink: treat PSGMII as an inband capable interface") into both kernel copies
the tree carries.

No board in the tree declares `managed = "in-band-status"` on a PSGMII port, so
every PSGMII port moves to `PHYLINK_PCS_NEG_OUTBAND`, PHY-managed and fixed
links alike. A PCS tells the two apart with `phylink_expects_phy()`:

    target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c:462

        static bool uniphy_ch_forced(struct phylink_pcs *pcs, unsigned int neg_mode)
        {
                return neg_mode == PHYLINK_PCS_NEG_OUTBAND &&
                       !phylink_expects_phy(pcs->phylink);
        }

Reading `pcs->phylink` there is what the first commit makes safe: the PCS list
is linked only after `phylink_mac_initial_config()` has run `.pcs_config()`, so
that first call sees NULL. DSA has no `.mac_select_pcs()`, so a DSA port with a
pcs-handle takes that route and a Zyxel NBG7815 panics on wan as netifd brings
it up. That patch moves the link above the initial config.

Fixes: 18cbd83a1443 ("generic: 6.18: import updated standalone PCS handling")
Report: https://github.com/JuliusBairaktaris/openwrt-nss-edma/issues/23
Rebased above #24186, which restructured `qca_uniphy_pcs_config_mode()`.

On a Xiaomi AX3600 (IPQ8074, kernel 6.18, four PSGMII channels on one UNIPHY)
every channel reads UNIPHY_CH_CTRL 0x844 after the clear: FORCE_MODE clear,
SPEED_MODE at its 1000 reset value, and a port linked at 100 Mbps holds that
value. All four ports pass traffic and dmesg carries no Oops or BUG. I have no
NBG7815, so the reported panic itself is not reproduced here.

No in-tree IPQ6018 or IPQ8074 board has a fixed link on a UNIPHY, so I exercised
the set side with a modified device tree on the AX3600: FORCE_MODE set,
SPEED_MODE following the declared speed. The eleven IPQ5018 boards already on
that path all declare 1000 Mbps, the value the field already holds.

I used AI to help me write the change. Can anyone with a fixed link on an
IPQ6018 or IPQ8074 UNIPHY read UNIPHY_CH_CTRL back with this applied?

### DIFF
--- a/target/linux/generic/backport-6.18/704-v7.3-net-phylink-treat-PSGMII-as-an-inband-capable-interface.patch
+++ b/target/linux/generic/backport-6.18/704-v7.3-net-phylink-treat-PSGMII-as-an-inband-capable-interface.patch
@@ -1,0 +1,71 @@
+From 5ba017f9efef3cf65cc60005aae4cbbf70b9b2b8 Mon Sep 17 00:00:00 2001
+From: Sandeep Sondagar <sandeepsondagar@gmail.com>
+Date: Sun, 9 Aug 2026 21:31:41 +0530
+Subject: [PATCH] net: phylink: treat PSGMII as an inband capable interface
+
+PSGMII (the Qualcomm 5-port SGMII) conveys the link negotiation result
+from the PHY back to the MAC through per-channel in-band SGMII words,
+exactly like SGMII and QSGMII.
+
+However, PHY_INTERFACE_MODE_PSGMII is missing from
+phylink_get_inband_type(), so phylink reports INBAND_NONE for it and
+phylink_pcs_neg_mode() falls back to PHYLINK_PCS_NEG_NONE. The PCS is
+then programmed in force mode and its control-register speed bits (which
+default to 1000base) are used, so a slower copper link - e.g. 100base-T
+- is reported as 1Gbps and cannot pass traffic.
+
+Classify PSGMII alongside SGMII and QSGMII as INBAND_CISCO_SGMII so the
+PCS negotiates in-band and the resolved link speed comes from the PHY
+in-band word.
+
+Also add PSGMII to the generic clause 22 PCS helper functions which
+handle the SGMII in-band word. Without this, a PCS using these helpers
+would still fall through to the default handling and force the link
+state to false in phylink_mii_c22_pcs_decode_state(), fail to encode
+the SGMII advertisement, and get rejected by phylink_get_link_timer_ns().
+
+Signed-off-by: Sandeep Sondagar <sandeepsondagar@gmail.com>
+Reviewed-by: Nicolai Buchwitz <nb@tipi-net.de>
+Link: https://patch.msgid.link/20260809-phylink-psgmii-v3-1-908dcd3a9e3d@gmail.com
+Signed-off-by: Jakub Kicinski <kuba@kernel.org>
+---
+ drivers/net/phy/phylink.c | 3 +++
+ include/linux/phylink.h   | 1 +
+ 2 files changed, 4 insertions(+)
+
+--- a/drivers/net/phy/phylink.c
++++ b/drivers/net/phy/phylink.c
+@@ -1036,6 +1036,7 @@ static enum inband_type phylink_get_inba
+ {
+ 	switch (interface) {
+ 	case PHY_INTERFACE_MODE_SGMII:
++	case PHY_INTERFACE_MODE_PSGMII:
+ 	case PHY_INTERFACE_MODE_QSGMII:
+ 	case PHY_INTERFACE_MODE_QUSGMII:
+ 	case PHY_INTERFACE_MODE_USXGMII:
+@@ -4093,6 +4094,7 @@ void phylink_mii_c22_pcs_decode_state(st
+ 		break;
+ 
+ 	case PHY_INTERFACE_MODE_SGMII:
++	case PHY_INTERFACE_MODE_PSGMII:
+ 	case PHY_INTERFACE_MODE_QSGMII:
+ 		if (neg_mode == PHYLINK_PCS_NEG_INBAND_ENABLED)
+ 			phylink_decode_sgmii_word(state, lpa);
+@@ -4173,6 +4175,7 @@ int phylink_mii_c22_pcs_encode_advertise
+ 			adv |= ADVERTISE_1000XPSE_ASYM;
+ 		return adv;
+ 	case PHY_INTERFACE_MODE_SGMII:
++	case PHY_INTERFACE_MODE_PSGMII:
+ 	case PHY_INTERFACE_MODE_QSGMII:
+ 		return 0x0001;
+ 	default:
+--- a/include/linux/phylink.h
++++ b/include/linux/phylink.h
+@@ -762,6 +762,7 @@ static inline int phylink_get_link_timer
+ {
+ 	switch (interface) {
+ 	case PHY_INTERFACE_MODE_SGMII:
++	case PHY_INTERFACE_MODE_PSGMII:
+ 	case PHY_INTERFACE_MODE_QSGMII:
+ 	case PHY_INTERFACE_MODE_USXGMII:
+ 	case PHY_INTERFACE_MODE_10G_QXGMII:
--- a/target/linux/generic/pending-6.12/770-02-net-phylink-introduce-internal-phylink-PCS-handling.patch
+++ b/target/linux/generic/pending-6.12/770-02-net-phylink-introduce-internal-phylink-PCS-handling.patch
@@ -189,7 +189,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  		return 0;
  
  	return phylink_pcs_inband_caps(pcs, interface);
-@@ -1397,10 +1444,36 @@ static void phylink_major_config(struct
+@@ -1398,10 +1445,36 @@ static void phylink_major_config(struct
  			pl->major_config_failed = true;
  			return;
  		}
@@ -227,7 +227,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  	phylink_pcs_neg_mode(pl, pcs, state->interface, state->advertising);
  
  	phylink_dbg(pl, "major config, active %s/%s/%s\n",
-@@ -1427,11 +1500,13 @@ static void phylink_major_config(struct
+@@ -1428,11 +1501,13 @@ static void phylink_major_config(struct
  	if (pcs_changed) {
  		phylink_pcs_disable(pl->pcs);
  
@@ -245,7 +245,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  
  		pl->pcs = pcs;
  	}
-@@ -1909,6 +1984,44 @@ int phylink_set_fixed_link(struct phylin
+@@ -1910,6 +1985,44 @@ int phylink_set_fixed_link(struct phylin
  }
  EXPORT_SYMBOL_GPL(phylink_set_fixed_link);
  
@@ -290,7 +290,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  /**
   * phylink_create() - create a phylink instance
   * @config: a pointer to the target &struct phylink_config
-@@ -1931,6 +2044,7 @@ struct phylink *phylink_create(struct ph
+@@ -1932,6 +2045,7 @@ struct phylink *phylink_create(struct ph
  			       const struct phylink_mac_ops *mac_ops)
  {
  	bool using_mac_select_pcs = false;
@@ -298,7 +298,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  	struct phylink *pl;
  	int ret;
  
-@@ -1952,9 +2066,21 @@ struct phylink *phylink_create(struct ph
+@@ -1953,9 +2067,21 @@ struct phylink *phylink_create(struct ph
  
  	mutex_init(&pl->state_mutex);
  	INIT_WORK(&pl->resolve, phylink_resolve);
@@ -320,7 +320,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  
  	pl->config = config;
  	if (config->type == PHYLINK_NETDEV) {
-@@ -2025,10 +2151,16 @@ EXPORT_SYMBOL_GPL(phylink_create);
+@@ -2026,10 +2152,16 @@ EXPORT_SYMBOL_GPL(phylink_create);
   */
  void phylink_destroy(struct phylink *pl)
  {
@@ -337,7 +337,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  	cancel_work_sync(&pl->resolve);
  	kfree(pl);
  }
-@@ -2473,6 +2605,7 @@ static irqreturn_t phylink_link_handler(
+@@ -2474,6 +2606,7 @@ static irqreturn_t phylink_link_handler(
   */
  void phylink_start(struct phylink *pl)
  {
@@ -345,7 +345,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  	bool poll = false;
  
  	ASSERT_RTNL();
-@@ -2499,6 +2632,10 @@ void phylink_start(struct phylink *pl)
+@@ -2500,6 +2633,10 @@ void phylink_start(struct phylink *pl)
  
  	pl->pcs_state = PCS_STATE_STARTED;
  
@@ -356,7 +356,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  	phylink_enable_and_run_resolve(pl, PHYLINK_DISABLE_STOPPED);
  
  	if (pl->cfg_link_an_mode == MLO_AN_FIXED && pl->link_gpio) {
-@@ -2543,6 +2680,8 @@ EXPORT_SYMBOL_GPL(phylink_start);
+@@ -2544,6 +2681,8 @@ EXPORT_SYMBOL_GPL(phylink_start);
   */
  void phylink_stop(struct phylink *pl)
  {
@@ -365,7 +365,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  	ASSERT_RTNL();
  
  	if (pl->sfp_bus)
-@@ -2560,6 +2699,14 @@ void phylink_stop(struct phylink *pl)
+@@ -2561,6 +2700,14 @@ void phylink_stop(struct phylink *pl)
  	pl->pcs_state = PCS_STATE_DOWN;
  
  	phylink_pcs_disable(pl->pcs);
--- a/target/linux/generic/pending-6.18/737-02-net-phylink-introduce-internal-phylink-PCS-handling.patch
+++ b/target/linux/generic/pending-6.18/737-02-net-phylink-introduce-internal-phylink-PCS-handling.patch
@@ -196,7 +196,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  		return 0;
  
  	return phylink_pcs_inband_caps(pcs, interface);
-@@ -1261,10 +1310,36 @@ static void phylink_major_config(struct
+@@ -1262,10 +1311,36 @@ static void phylink_major_config(struct
  			pl->major_config_failed = true;
  			return;
  		}
@@ -234,7 +234,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  	phylink_pcs_neg_mode(pl, pcs, state->interface, state->advertising);
  
  	phylink_dbg(pl, "major config, active %s/%s/%s\n",
-@@ -1291,11 +1366,13 @@ static void phylink_major_config(struct
+@@ -1292,11 +1367,13 @@ static void phylink_major_config(struct
  	if (pcs_changed) {
  		phylink_pcs_disable(pl->pcs);
  
@@ -252,7 +252,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  
  		pl->pcs = pcs;
  	}
-@@ -1823,6 +1900,44 @@ int phylink_set_fixed_link(struct phylin
+@@ -1824,6 +1901,44 @@ int phylink_set_fixed_link(struct phylin
  }
  EXPORT_SYMBOL_GPL(phylink_set_fixed_link);
  
@@ -297,7 +297,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  /**
   * phylink_create() - create a phylink instance
   * @config: a pointer to the target &struct phylink_config
-@@ -1844,6 +1959,7 @@ struct phylink *phylink_create(struct ph
+@@ -1845,6 +1960,7 @@ struct phylink *phylink_create(struct ph
  			       phy_interface_t iface,
  			       const struct phylink_mac_ops *mac_ops)
  {
@@ -305,7 +305,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  	struct phylink *pl;
  	int ret;
  
-@@ -1861,9 +1977,21 @@ struct phylink *phylink_create(struct ph
+@@ -1862,9 +1978,21 @@ struct phylink *phylink_create(struct ph
  	mutex_init(&pl->phydev_mutex);
  	mutex_init(&pl->state_mutex);
  	INIT_WORK(&pl->resolve, phylink_resolve);
@@ -327,7 +327,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  
  	pl->config = config;
  	if (config->type == PHYLINK_NETDEV) {
-@@ -1943,10 +2071,16 @@ EXPORT_SYMBOL_GPL(phylink_create);
+@@ -1944,10 +2072,16 @@ EXPORT_SYMBOL_GPL(phylink_create);
   */
  void phylink_destroy(struct phylink *pl)
  {
@@ -344,7 +344,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  	cancel_work_sync(&pl->resolve);
  	kfree(pl);
  }
-@@ -2448,6 +2582,7 @@ static irqreturn_t phylink_link_handler(
+@@ -2449,6 +2583,7 @@ static irqreturn_t phylink_link_handler(
   */
  void phylink_start(struct phylink *pl)
  {
@@ -352,7 +352,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  	bool poll = false;
  
  	ASSERT_RTNL();
-@@ -2474,6 +2609,10 @@ void phylink_start(struct phylink *pl)
+@@ -2475,6 +2610,10 @@ void phylink_start(struct phylink *pl)
  
  	pl->pcs_state = PCS_STATE_STARTED;
  
@@ -363,7 +363,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  	phylink_enable_and_run_resolve(pl, PHYLINK_DISABLE_STOPPED);
  
  	if (pl->cfg_link_an_mode == MLO_AN_FIXED && pl->link_gpio) {
-@@ -2518,6 +2657,8 @@ EXPORT_SYMBOL_GPL(phylink_start);
+@@ -2519,6 +2658,8 @@ EXPORT_SYMBOL_GPL(phylink_start);
   */
  void phylink_stop(struct phylink *pl)
  {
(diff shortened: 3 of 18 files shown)


robimarko:
```
### Finding

  - [P2] Force mode is enabled without programming the forced speed. The patch sets UNIPHY_CH_FORCE_MODE, which makes the channel consume UNIPHY_CH_SPEED_MODE, but that adjacent field is never updated (configuration
    (https://github.com/openwrt/openwrt/blob/7aacfaaa7dd9e10df0fa3a61a8384414cdf50756/target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c#L540-L547), link-up path (https://github.com/openwrt/openwrt/blob/7aacfaaa7dd9e10df0fa3a61a8384414cdf50756/target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c#L670-L720)).

    The register resets with speed encoding 2, meaning 1000 Mbps. Consequently, fixed SGMII links at 10 or 100 Mbps are forced to the stale 1000 Mbps encoding even though uniphy_link_up_sgmii() changes the RX/TX clock rates. They will not operate correctly.

    When force mode is active, pcs_link_up should also write:
      - 0 for 10 Mbps
      - 1 for 100 Mbps
      - 2 for 1000 Mbps

    into UNIPHY_CH_SPEED_MODE before toggling UNIPHY_CH_ADP_SW_RSTN.

  Non-blocking documentation issue: the statement that no in-tree board takes the fixed-link path is incorrect. For example, ipq5018-mr5500 configures gmac1 with a 1000 Mbps fixed link (https://github.com/openwrt/openwrt/blob/7aacfaaa7dd9e10df0fa3a61a8384414cdf50756/target/linux/qualcommax/dts/ipq5018-mr5500.dts#L59-L71), while gmac1 inherits its
  pcs-handle from ipq5018-ess.dtsi.

  I would request the forced-speed programming before merging head 7aacfaaa7d.
```

JuliusBairaktaris:
@robimarko Fixed in `646d1f9`, and thanks for the DTS correction — it was wrong: my scan only walked DSA `port@N` nodes, and the IPQ5018 fixed links hang off `&gmac1`, which inherits `pcs-handle` from `ipq5018-ess.dtsi`. Redone properly: 11 IPQ5018 boards declare a fixed link there, all at 1000 Mbps — the value the field already held, so the speed half was latent rather than visible — and no IPQ6018 or IPQ8074 board in the tree has a fixed link on a UNIPHY. Commit message and PR body corrected.

`pcs_link_up()` now programs `UNIPHY_CH_SPEED_MODE` (0/1/2 for 10/100/1000) before the `ADP_SW_RSTN` pulse, in the same encoding `qca_uniphy_pcs_get_state_sgmii()` already decodes out of the status register. 2500BASE-X has no encoding of its own in those two bits — SGMII+ carries the rate in the mode and the SSDK leaves the field at its 1000 reset value, which is what we write. The force-mode write itself is checked now as well.

robimarko:
This clearly needs to be runtime tested, I dont want any regressions

JuliusBairaktaris:
@robimarko `15befadd` — same code as before, now with the bench behind it.

On a Xiaomi AX3600 (IPQ8074, QCA8075 PSGMII, kernel 6.18) `FORCE_MODE` reads 0 on every channel after the clear, and `SPEED_MODE` follows the link: `UNIPHY_CH_CTRL` goes `0x844` / `0x842` / `0x840` at 1000 / 100 / 10 Mbps, the status field agrees, traffic passes at each step and the other channels of the instance are untouched. Before the change the same port sat at `0x844` (the 1000 encoding) while linked at 100 Mbps.

The set side stays untested: no IPQ6018 or IPQ8074 board in the tree has a fixed link on a UNIPHY, and the eleven IPQ5018 ones all declare 1000 Mbps.

(The branch briefly pointed at main by mistake, which auto-closed this PR — reopened, head is `15befadd`.)

target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c
```diff
@@ -536,15 +537,16 @@ static int qca_uniphy_pcs_config_mode(struct phylink_pcs *pcs,
...
-		}
-	}
+	/* A fixed link carries no in-band negotiation for the channel to take
+	 * its speed from, so the channel has to be forced.
+	 */
+	force = neg_mode == PHYLINK_PCS_NEG_OUTBAND &&
+		!phylink_expects_phy(pcs->phylink);
+	ret = regmap_update_bits(uniphy->regmap, UNIPHY_CH_CTRL(upcs->channel),
+				 UNIPHY_CH_FORCE_MODE,
+				 force ? UNIPHY_CH_FORCE_MODE : 0);
```
openwrt-ai: This write is no longer confined to the SGMII family: `qca_uniphy_pcs_config_mode()` is also the first thing [`qca_uniphy_pcs_config_usxgmii()` calls at pcs-qca-uniphy.c:615](``https://github.com/openwrt/openwrt/blob/15befaddb94dddfd58dc9eb06f8996f3e38e1a56/target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c#L615),`` so it now runs for `PHY_INTERFACE_MODE_USXGMII` and `PHY_INTERFACE_MODE_10GBASER` too, where the channel is driven by the XPCS and `UNIPHY_CH_CTRL` is not the register in play. The old code did not reach them on IPQ6018/IPQ8074 at all, and the SSDK writes this bit only from the SGMII mode-set paths your commit message cites.

No in-tree board hits the set side there today — the only USXGMII port, [nbg7815's `managed = "in-band-status"` at ipq8074-nbg7815.dts:377-379](https://github.com/openwrt/openwrt/blob/15befaddb94dddfd58dc9eb06f8996f3e38e1a56/target/linux/qualcommax/dts/ipq8074-nbg7815.dts#L377-L379), gives `PHYLINK_PCS_NEG_INBAND_*` so `force` is false and the clear only touches a bit XPCS mode ignores. But a 10GBASE-R or USXGMII fixed link would set `UNIPHY_CH_FORCE_MODE` on a channel that never consumes it. Was restricting this to the SGMII/QSGMII/PSGMII/1000BASE-X/2500BASE-X interfaces considered — either an interface check here or moving the block up into the SGMII branch of `qca_uniphy_pcs_config()`?

---

target/linux/qualcommax/files/include/linux/pcs/pcs-qca-uniphy.h
```diff
@@ -43,6 +43,9 @@
 
 #define UNIPHY_CH_CTRL(ch)		(UNIPHY_CH_BASE(ch) + 0x0)
 #define   UNIPHY_CH_SPEED_MODE		GENMASK(2, 1)
+#define     UNIPHY_CH_SPEED_10		0
+#define     UNIPHY_CH_SPEED_100		1
+#define     UNIPHY_CH_SPEED_1000	2
```
openwrt-ai, line 50: nit: the name and the nesting scope these to `UNIPHY_CH_CTRL`/`UNIPHY_CH_SPEED_MODE`, but they are also used to decode `UNIPHY_CH_STS_SPEED_MODE` — a different register and a different mask — in [`qca_uniphy_pcs_get_state_sgmii()` at pcs-qca-uniphy.c:303-311](``https://github.com/openwrt/openwrt/blob/15befaddb94dddfd58dc9eb06f8996f3e38e1a56/target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c#L303-L311).`` The two fields do share the encoding, so the code is right; a short comment here saying the encoding is common to the control and status fields would spare the next reader the cross-check.

---

openwrt-ai (commented):
Reviewed 1 commit; the commit message matches the diff, and the in-tree claims it makes hold up — 11 IPQ5018 boards declare a fixed link under `&gmac1`, all at `speed = <1000>`, and no IPQ6018/IPQ8074 board has a fixed link on a UNIPHY. The force condition (`PHYLINK_PCS_NEG_OUTBAND` plus `!phylink_expects_phy()`) does resolve to exactly `MLO_AN_FIXED` on 6.18: `PHYLINK_PCS_NEG_OUTBAND` is only produced for `MLO_AN_PHY`/`MLO_AN_FIXED`, and `phylink_expects_phy()` returns true for the former. Two inline notes, one of them a nit.

---

target/linux/generic/pending-6.18/737-02-net-phylink-introduce-internal-phylink-PCS-handling.patch
JuliusBairaktaris, line 1: @robimarko is it fine to edit a pending patch like so?

target/linux/generic/pending-6.18/737-02-net-phylink-introduce-internal-phylink-PCS-handling.patch
```diff
@@ -352,18 +352,20 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
...
  
- 	pl->pcs_state = PCS_STATE_STARTED;
+ 	pl->pcs_state = PCS_STATE_STARTING;
  
-+	/* link available PCS to phylink struct */
++	/* link available PCS to phylink struct, before the initial config
++	 * below runs .pcs_config() on the one it selects
++	 */
 +	list_for_each_entry(pcs, &pl->pcs_list, list)
 +		pcs->phylink = pl;
```
openwrt-ai: The same series is carried twice, and only the 6.18 copy is moved: [`pending-6.12/770-02-...patch:348-356`](``https://github.com/openwrt/openwrt/blob/708409469074e484aa6d78dd4d359f8180a686f3/target/linux/generic/pending-6.12/770-02-net-phylink-introduce-internal-phylink-PCS-handling.patch#L348-L356)`` still inserts the same loop after `pl->pcs_state = PCS_STATE_STARTED;`, i.e. after `phylink_mac_initial_config()` has already run `.pcs_config()` — exactly the window this commit closes. v6.12 `phylink_start()` has the identical shape ([`PCS_STATE_STARTING` at phylink.c:2234, `phylink_mac_initial_config()` at 2244, `PCS_STATE_STARTED` at 2246](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/drivers/net/phy/phylink.c#L2234-L2246)), so the same hunk applies there verbatim.

Nothing is broken on 6.12 today — no 6.12 target consumes the PCS list (realtek's `rtldsa_phylink_fill_available_pcs()` is under `files-6.18/`, and qualcommax/airoha/mediatek are all `KERNEL_PATCHVER:=6.18`) — but the two copies of the series now diverge, which will bite whoever refreshes or re-submits it. Was leaving `770-02` alone deliberate?

nit: relatedly, the commit subject is `generic:` with no kernel version while only `pending-6.18/` is touched; in-tree neighbours that touch a single series tend to say so (`generic: 6.18: disable CONFIG_SND_SOC_AN7581`, `kernel: 6.18: backport ksmbd multi-iov signing fix`).

---

openwrt-ai (commented):
Reviewed the 2 commits new since `15befadd`. Both messages match their diffs, so no commit checks.

The phylink change holds up against v6.18.44: `phylink_start()` sets `PCS_STATE_STARTING` before `phylink_mac_initial_config()` and `PCS_STATE_STARTED` after, so the hunk really does move the linking ahead of the first `.pcs_config()`; the refreshed hunk offsets are consistent (+30 against vanilla on all four hunks in that file); DSA's default `dsa_port_phylink_mac_ops` indeed has no `.mac_select_pcs`; and the other `phylink_mac_initial_config()` / `phylink_major_config()` callers are all gated on `!PHYLINK_DISABLE_STOPPED`, as the message claims. Both notes from my previous review are addressed — the XPCS modes are excluded from the `UNIPHY_CH_CTRL` write now, and the shared control/status encoding is documented in the header. CI is green on `70840946`.

One question left, inline, about the 6.12 copy of the same series.

---

target/linux/generic/pending-6.18/737-02-net-phylink-introduce-internal-phylink-PCS-handling.patch
robimarko: I would add a new one that documents your change.

Also, you gotta refresh all of the targets

target/linux/generic/pending-6.12/770-02-net-phylink-introduce-internal-phylink-PCS-handling.patch
```diff
@@ -345,17 +345,19 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
...
 +	list_for_each_entry(pcs, &pl->pcs_list, list)
 +		pcs->phylink = pl;
 +
- 	phylink_enable_and_run_resolve(pl, PHYLINK_DISABLE_STOPPED);
- 
- 	if (pl->cfg_link_an_mode == MLO_AN_FIXED && pl->link_gpio) {
+ 	/* Apply the link configuration to the MAC when starting. This allows
+ 	 * a fixed-link to start with the correct parameters, and also
+ 	 * ensures that we set the appropriate advertisement for Serdes links.
 @@ -2543,6 +2680,8 @@ EXPORT_SYMBOL_GPL(phylink_start);
```
openwrt-ai: The hunk above now inserts 6 lines instead of 4, but the new-side start lines of the two hunks after it were not shifted, so this file's metadata is internally inconsistent and the `Check Kernel patches` job fails on every 6.12 target:

```
target/linux/generic/pending-6.12/770-02-net-phylink-introduce-internal-phylink-PCS-handling.patch
Kernel patches for bcm53xx/generic require refresh. (run 'make target/linux/refresh' and force push this pr)
```

([bcm53xx/generic](https://github.com/openwrt/openwrt/actions/runs/32560544999/job/97005222987); same single-file diff on at91/sam9x, bcm27xx/bcm2708, bcm47xx/generic, bcm4908/generic, ipq806x/chromium and tegra/generic — the 6.18 variants of the same targets all pass.)

Arithmetic: `@@ -2487,6 +2620,12 @@` runs the new-side offset from 133 to 139, so `-2543` must map to `+2682`, and `-2560` on line 370 below to `+2701`. The [6.18 copy](``https://github.com/openwrt/openwrt/blob/ce68df60095f5248a07fc8bbba7b34020aeabd07/target/linux/generic/pending-6.18/737-02-net-phylink-introduce-internal-phylink-PCS-handling.patch#L355-L368)`` is consistent (`-2462,6 +2597,12` → `-2518,6 +2659,8`, +6 across the boundary), which is why only 6.12 fails.

```suggestion
@@ -2543,6 +2682,8 @@ EXPORT_SYMBOL_GPL(phylink_start);
```

Same +2 applies to line 370 (`@@ -2560,6 +2699,14 @@` → `+2701`). `make target/linux/refresh` (or the `*-refreshed` artifact CI attached) is the authoritative fix. If you go with @robimarko's suggestion of a separate patch on top instead of editing `770-02`/`737-02` in place, this goes away on its own — but the refresh across all targets is still needed either way.

---

openwrt-ai (commented):
Reviewed the range `70840946..ce68df60`. The only change is the one my last review asked about: the 6.12 copy of the series (`pending-6.12/770-02-...`) now gets the same move as the 6.18 one, and commit `8e329423`'s message grew a paragraph documenting exactly that. Both commit messages match their diffs, so no commit checks.

The moved hunk itself is correct against v6.12.103 — `PCS_STATE_STARTING` at [`phylink.c:2234`](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/drivers/net/phy/phylink.c#L2234-L2244) with `phylink_mac_initial_config()` at 2244, so `-2487` is precisely 12 lines above the old `-2499` anchor, and the inserted text is byte-identical to the 6.18 copy. One inline note: the following hunk headers in that file were not shifted along, which is what the failing `Check Kernel patches` jobs are reporting.

---

JuliusBairaktaris:
@robimarko i will submit the patches once you have confirmed the pr. Thanks as always

openwrt-ai (commented):
Reviewed the range `ce68df60..4f0afe2e`. The force-push reworks how the phylink change is packaged: `pending-6.18/737-02` and `pending-6.12/770-02` are byte-identical to their as-imported form on `main` again, and the move now rides in two new standalone patches (`pending-6.18/737-10`, `pending-6.12/770-08`) rather than an in-place edit of the imported series — which is what @robimarko asked for. Both commit messages match their diffs, so no commit checks.

The new patches are correct: each removes exactly the `pcs->phylink = pl` loop that `737-02`/`770-02` add after `pl->pcs_state = PCS_STATE_STARTED` (byte-identical, so the removal anchors cleanly) and re-inserts it after `PCS_STATE_STARTING`, the same move validated in my earlier reviews against v6.18.44 and v6.12.103. The qualcommax driver commit is unchanged from `ce68df60`.

Every earlier thread is addressed now: the imported series is no longer edited, both kernel copies carry the move, and all targets are refreshed — the `Check Kernel patches` jobs that were failing (qualcommax/ipq50xx, qualcommbe/ipq95xx, and the generic set) are green on `4f0afe2e`. No new issues found.

---

robimarko:
@Ansuel Can you look at the PCS patch?

target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c
```diff
@@ -709,6 +725,15 @@ static int uniphy_link_up_sgmii(struct phylink_pcs *pcs,
...
 
+	/* A forced channel takes its speed from here instead of from an
+	 * in-band word.
+	 */
+	ret = regmap_update_bits(uniphy->regmap, UNIPHY_CH_CTRL(upcs->channel),
+				 UNIPHY_CH_SPEED_MODE,
+				 FIELD_PREP(UNIPHY_CH_SPEED_MODE, speed_mode));
+	if (ret)
+		return ret;
+
```
Ansuel: Pass negmode to this function and guard this to enable only when inband is disabled. Makes it clear that the register is relevant only when force mode is used.

target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c
```diff
@@ -524,14 +533,17 @@ static int qca_uniphy_pcs_config_mode(struct phylink_pcs *pcs,
...
-		//set force mode for fixed link
-		if (neg_mode == PHYLINK_PCS_NEG_OUTBAND && !phylink_expects_phy(pcs->phylink)) {
-			regmap_set_bits(uniphy->regmap,
-					UNIPHY_CH_CTRL(upcs->channel),
-					UNIPHY_CH_FORCE_MODE);
-		}
+	/* The XPCS modes drive the channel from the XPCS and never read the
+	 * force bit.
+	 */
+	if (mode_ctrl != UNIPHY_XPCS_MODE) {
```
openwrt-ai, line 539: nit: `mode_ctrl` is a composite value — the SGMII-family branches OR `UNIPHY_CH0_MODE_CTRL_25M` and `UNIPHY_AUTONEG_MODE_ATH` on top of the mode bit, and only the two XPCS branches ([pcs-qca-uniphy.c:508-514](``https://github.com/openwrt/openwrt/blob/c338294d79896749ac987f1996c33f8f87485e45/target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c#L508-L514``)) leave it as the bare `UNIPHY_XPCS_MODE`. So the equality test is right only for as long as those two branches OR nothing in; the day one of them gains a field (a `UNIPHY_CH0_MODE_CTRL_25M` for the XPCS channel, say) this silently starts writing `UNIPHY_CH_CTRL` in XPCS mode again. A bit test is equivalent today and does not depend on that:

```suggestion
	if (!(mode_ctrl & UNIPHY_XPCS_MODE)) {
```

---

target/linux/generic/backport-6.18/704-v7.3-net-phylink-treat-PSGMII-as-an-inband-capable-interface.patch
```diff
@@ -0,0 +1,71 @@
...
+ include/linux/phylink.h   | 1 +
+ 2 files changed, 4 insertions(+)
+
+--- a/drivers/net/phy/phylink.c
++++ b/drivers/net/phy/phylink.c
+@@ -1036,6 +1036,7 @@ static enum inband_type phylink_get_inba
+ {
+ 	switch (interface) {
+ 	case PHY_INTERFACE_MODE_SGMII:
++	case PHY_INTERFACE_MODE_PSGMII:
```
openwrt-ai, line 42: nit: the commit message (and the PR body) says "every PSGMII port keeps the outband path it already takes", but a PSGMII port did not take the outband path before this hunk — `phylink_get_inband_type()` returned `INBAND_NONE` and `phylink_pcs_neg_mode()` bailed out with `PHYLINK_PCS_NEG_NONE` ([phylink.c:1092-1097](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/phylink.c#L1092-L1097)). After it, *every* PSGMII port that is not `managed = "in-band-status"` leaves the `!phylink_autoneg_inband()` branch as `PHYLINK_PCS_NEG_OUTBAND` ([phylink.c:1112-1150](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/phylink.c#L1112-L1150)) — PHY-managed ones too, not only the fixed links the next paragraph singles out.

It does look harmless in tree: of the three drivers that see a PSGMII `neg_mode`, qca-uniphy (this series), `ipq_pcs` (`qualcommbe/patches-6.18/0323-net-pcs-qcom-ipq9574-Update-IPQ9574-PCS-driver.patch`) and qca8k, none distinguishes `NEG_NONE` from `NEG_OUTBAND` — each only tests `neg_mode == PHYLINK_PCS_NEG_INBAND_ENABLED`. Given how precise the rest of the message is, it seems worth stating that explicitly rather than implying nothing changes for PHY-managed ports, since that is the neg_mode transition every PSGMII PCS now observes.

---

openwrt-ai (commented):
Reviewed the range `4f0afe2e..c338294d`: a rebase onto `3052766a`, one new commit, and the follow-up to @Ansuel's note. Both newly added commit messages match their diffs, so no commit checks.

`80d4180d` adds the PSGMII in-band classification to both kernel copies as `backport-{6.12,6.18}/704-v7.3-...` and refreshes the pending phylink patches that sort after it. The 6.18 copy's contexts line up with vanilla v6.18.44 — `phylink_get_inband_type()` at [phylink.c:1033-1046](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/phylink.c#L1033-L1046), `phylink_mii_c22_pcs_decode_state()`'s SGMII case with its `neg_mode` guard at [4092-4096](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/phylink.c#L4092-L4096) — and the 6.12 copy correctly drops that guard, which that kernel does not have. The "no board declares `managed = \"in-band-status\"` on a PSGMII port" claim holds: the three DTS files mentioning both (`ipq8072-haze`, `ipq8074-nbg7815`, `ipq8074-rax120v2`) use it on their USXGMII `swport6` only. Patch numbering is free at 704 in both directories, and `-v7.3-` matches the other net-next-era backports in the tree.

The driver commit answers @Ansuel: `uniphy_ch_forced()` now gates the `UNIPHY_CH_SPEED_MODE` write in the link-up path as well, so the field is only touched on a channel that consumes it; `speed_mode` is assigned on every path that can reach the write (each inner `default:` returns `-EINVAL`), and the write still precedes the `ADP_SW_RSTN` pulse @robimarko asked for. The force-mode write stays ahead of the cached-mode early return, so it runs on every `.pcs_config()`.

Two nits inline, nothing blocking. No failed jobs on `c338294d`.

---

target/linux/generic/backport-6.18/704-v7.3-net-phylink-treat-PSGMII-as-an-inband-capable-interface.patch
JuliusBairaktaris: > a PSGMII port did not take the outband path before this hunk - `phylink_get_inband_type()` returned `INBAND_NONE` and `phylink_pcs_neg_mode()` bailed out with `PHYLINK_PCS_NEG_NONE`

I agree with that, the sentence was wrong. The paragraph now reads:

    No board in the tree declares managed = "in-band-status" on a PSGMII
    port, so every PSGMII port moves from PHYLINK_PCS_NEG_NONE to
    PHYLINK_PCS_NEG_OUTBAND, PHY-managed and fixed links alike. A PCS
    driver can distinguish the two with phylink_expects_phy() and force
    its channel only on a fixed link.

> of the three drivers that see a PSGMII `neg_mode`, qca-uniphy (this series), `ipq_pcs` [...] and qca8k, none distinguishes `NEG_NONE` from `NEG_OUTBAND` - each only tests `neg_mode == PHYLINK_PCS_NEG_INBAND_ENABLED`

qca-uniphy distinguishes them after the third commit:

    target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c, uniphy_ch_forced():

    	return neg_mode == PHYLINK_PCS_NEG_OUTBAND &&
    	       !phylink_expects_phy(pcs->phylink);

On a PHY-managed PSGMII port that is false, so the new neg_mode writes 0 to
`UNIPHY_CH_FORCE_MODE`, the value the bit already held. That is what the
unmodified device tree arm on the AX3600 reads: every channel register keeps
its value.

target/linux/qualcommax/files/drivers/net/pcs/pcs-qca-uniphy.c
JuliusBairaktaris: > So the equality test is right only for as long as those two branches OR nothing in; the day one of them gains a field (a `UNIPHY_CH0_MODE_CTRL_25M` for the XPCS channel, say) this silently starts writing `UNIPHY_CH_CTRL` in XPCS mode again.

I agree the bit test is the more robust of the two.

The two pick the same branch for every value `mode_ctrl` can hold today, so
taking it means another bench run on the PCS commit for no change in
behaviour.

Would you like it taken anyway?

JuliusBairaktaris:
@robimarko @Ansuel - message-only amend of `generic: backport the phylink
PSGMII in-band classification`, the tree is byte-identical to the previous
push.

One correction to a number I posted earlier in this thread:

> `UNIPHY_CH_CTRL` goes `0x844` / `0x842` / `0x840` at 1000 / 100 / 10 Mbps

That was read on `15befadd`, where `pcs_link_up()` wrote the field on every
channel:

    ret = regmap_update_bits(uniphy->regmap, UNIPHY_CH_CTRL(upcs->channel),
                             UNIPHY_CH_SPEED_MODE,
                             FIELD_PREP(UNIPHY_CH_SPEED_MODE, speed_mode));

so what I measured there was the driver's own write, not the hardware. Since
`c338294d` that write sits behind `uniphy_ch_forced()` and a PHY-managed
channel is left alone.

I re-read the register on a build carrying no version of this series, where
nothing writes the field, which is the condition the current commit puts a
PHY-managed channel in. On the AX3600 with lan3 linked at 100 Mbps and lan1 at
1000 in the same read, three link-up cycles, all five channels of UNIPHY0 read
`0x844` every time - SPEED_MODE 2, the 1000 reset value. So the field does not
follow the in-band word, and a forced channel does need it programmed.

openwrt-ai (commented):

---

robimarko:

PR #24793, opened 2026-08-19
imx: backport dts fixes from linux-7.2 for Kontron OSM-S/BL i.MX8MP

sch-m:
This backports the latest dts fixes for the Kontron OSM-S/BL i.MX8MP.

### DIFF
--- a/target/linux/imx/patches-6.12/520-v7.2-Revert-arm64-dts-imx8mp-kontron-Add-support-for-reading-SD_VSEL-signal.patch
+++ b/target/linux/imx/patches-6.12/520-v7.2-Revert-arm64-dts-imx8mp-kontron-Add-support-for-reading-SD_VSEL-signal.patch
@@ -1,0 +1,60 @@
+From 22465a195af370ed6311be3adee479fb7c683685 Mon Sep 17 00:00:00 2001
+From: Peng Fan <peng.fan@nxp.com>
+Date: Fri, 3 Apr 2026 17:57:02 +0800
+Subject: Revert "arm64: dts: imx8mp-kontron: Add support for reading SD_VSEL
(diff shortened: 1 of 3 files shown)


openwrt-ai (commented):

Verification notes (for the record):

- All three upstream commits exist in mainline and the patch bodies match them verbatim: `22465a195af370ed6311be3adee479fb7c683685` (Revert SD_VSEL), `95f5bc1632ab8f243e517357b5da0dd28d1c6c92` (EERAM SPI clock), `5b19ca527471903920d713bc48518a036a95bf6b` (display power switch GPIO).
- Applying the full `target/linux/imx/patches-6.12/` stack (100 → 522) on top of v6.12 applies with no fuzz/offset, and the resulting `imx8mp-kontron-bl-osm-s.dts` / `imx8mp-kontron-osm-s.dtsi` hash byte-identical to their v7.2-rc7 blobs (`29ce863403…`, `ea69c639b3…`), so the two files are now fully in sync with mainline and nothing is missing.
- Numbering (520–522) and the `-v7.2-` prefix follow the existing convention in the directory (517–519 are `-v7.1-`). Keeping 512 plus the 520 revert rather than dropping 512 is the right call, since 512 is still needed for any kernel in [6.15, 7.2).
- `MX8MP_IOMUXC_SPDIF_TX__GPIO5_IO03` added by 522 does not collide with any other muxing in the enabled nodes — the BL carrier only enables `&pwm2` (SPDIF_RX), while SPDIF_TX/`&pwm3` stays unused.
- CI on `8e8bdce` is green (imx cortexa7/cortexa9/cortexa53 kernel builds + kernel patch check).

---

robimarko:

PR #24794, opened 2026-08-19
qualcommax: ipq50xx: add support for bluetooth

georgemoussalem:
This PR adds Bluetooth support for IPQ50xx chipsets.
It contains the patches, sent upstream for review for kernel support, and the firmware files.
In addition, it creates a new kernel module to be added to device recipes rather than building it in by default.
Lastly, enable Bluetooth on Linksys devices.


package/firmware/qca-bt-firmware/Makefile
```diff
@@ -0,0 +1,32 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=qca-bt-firmware
+PKG_RELEASE:=1
```
openwrt-ai, line 6: `PKG_LICENSE` / `PKG_LICENSE_FILES` are mandatory for a new package and are missing here (the formality bot flags this as a critical error). The sibling QCA blob package sets [`PKG_LICENSE_FILES:=LICENSE.qca_firmware` in `package/firmware/ath11k-firmware/Makefile`](https://github.com/openwrt/openwrt/blob/6110fa3f71830798c461c63edebe0c285321dac0/package/firmware/ath11k-firmware/Makefile#L19); since the blobs here are shipped from `files/` with no accompanying licence text, `PKG_LICENSE_FILES` has to stay empty (as [`package/firmware/mipi-dbi/Makefile`](https://github.com/openwrt/openwrt/blob/6110fa3f71830798c461c63edebe0c285321dac0/package/firmware/mipi-dbi/Makefile#L9) does) unless you also add the vendor licence file.

```suggestion
PKG_RELEASE:=1

PKG_LICENSE:=proprietary
PKG_LICENSE_FILES:=
```

Separately, worth stating in the commit message where these blobs were extracted from — every other QCA firmware package on this target (`ath11k-firmware`, `ipq-wifi`) pulls its blobs from an external git rather than committing them, so a maintainer will want the provenance recorded.

---

package/firmware/qca-bt-firmware/Makefile
```diff
@@ -0,0 +1,32 @@
...
+
+PKG_MAINTAINER:=George Moussalem <george.moussalem@outlook.com>
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/qca-bt-firmware-default
+  SECTION:=firmware
+  CATEGORY:=Firmware
+  DEPENDS:=@TARGET_qualcommax_ipq50xx
+  DEFAULTS:=n
```
openwrt-ai, line 16: `DEFAULTS` is not a recognised package variable, so this line does nothing. The variable the build system reads is `DEFAULT` ([`include/package-defaults.mk:44`](https://github.com/openwrt/openwrt/blob/6110fa3f71830798c461c63edebe0c285321dac0/include/package-defaults.mk#L44)), as used by the comparable blob package [`package/firmware/lantiq/lantiq-gphy-firmware/Makefile:16`](https://github.com/openwrt/openwrt/blob/6110fa3f71830798c461c63edebe0c285321dac0/package/firmware/lantiq/lantiq-gphy-firmware/Makefile#L16).

```suggestion
  DEFAULT:=n
```

---

package/firmware/qca-bt-firmware/Makefile
```diff
@@ -0,0 +1,32 @@
...
+PKG_RELEASE:=1
+
+PKG_MAINTAINER:=George Moussalem <george.moussalem@outlook.com>
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/qca-bt-firmware-default
+  SECTION:=firmware
+  CATEGORY:=Firmware
+  DEPENDS:=@TARGET_qualcommax_ipq50xx
```
openwrt-ai, line 15: nit: the two in-tree packages with the same shape — target-gated blobs shipped straight out of `files/` — both set `PKG_FLAGS:=nonshared`: [`package/firmware/lantiq/lantiq-gphy-firmware/Makefile:7`](https://github.com/openwrt/openwrt/blob/6110fa3f71830798c461c63edebe0c285321dac0/package/firmware/lantiq/lantiq-gphy-firmware/Makefile#L7) and [`package/firmware/ipq-wifi/Makefile:12`](https://github.com/openwrt/openwrt/blob/6110fa3f71830798c461c63edebe0c285321dac0/package/firmware/ipq-wifi/Makefile#L12). Was omitting it here deliberate, given the `@TARGET_qualcommax_ipq50xx` dependency?

---

target/linux/qualcommax/dts/ipq5018-mx-base.dtsi
```diff
@@ -74,6 +74,13 @@
 	pinctrl-names = "default";
 };
 
+&bluetooth {
+	status = "okay";
+
+	nvmem-cells = <&hw_mac_addr 0>;
```
openwrt-ai, line 80: Offset `0` of `hw_mac_addr` is already handed to an Ethernet MAC on every board that includes this dtsi, so the controller will come up with a BD address identical to an in-use interface MAC: [`ipq5018-mr5500.dts:65`](https://github.com/openwrt/openwrt/blob/6110fa3f71830798c461c63edebe0c285321dac0/target/linux/qualcommax/dts/ipq5018-mr5500.dts#L65) (`&gmac1`), [`ipq5018-mx2000.dts:46`](https://github.com/openwrt/openwrt/blob/6110fa3f71830798c461c63edebe0c285321dac0/target/linux/qualcommax/dts/ipq5018-mx2000.dts#L46), [`ipq5018-mx5500.dts:49`](https://github.com/openwrt/openwrt/blob/6110fa3f71830798c461c63edebe0c285321dac0/target/linux/qualcommax/dts/ipq5018-mx5500.dts#L49), [`ipq5018-mx6200.dts:62`](https://github.com/openwrt/openwrt/blob/6110fa3f71830798c461c63edebe0c285321dac0/target/linux/qualcommax/dts/ipq5018-mx6200.dts#L62) and [`ipq5018-spnmx56.dts:45`/`:58`](https://github.com/openwrt/openwrt/blob/6110fa3f71830798c461c63edebe0c285321dac0/target/linux/qualcommax/dts/ipq5018-spnmx56.dts#L45) all use `<&hw_mac_addr 0>` too. Is reusing offset 0 intentional (i.e. does the OEM firmware use the same address for BT), or should this get its own offset like the wifi radios do?

---

target/linux/qualcommax/patches-6.18/0190-net-of_net-Add-of_get_nvmem_eui48-helper-for-EUI-48-lookup.patch
```diff
@@ -0,0 +1,299 @@
+From patchwork Thu Jul 30 16:00:37 2026
```
openwrt-ai: This is a raw patchwork mbox dump — 112 lines of `Received:`/DKIM/List-* SMTP headers precede the real `From:`/`Date:`/`Subject:` at lines 113–115. It should be trimmed to the standard `From <hash> Mon Sep 17 00:00:00 2001` + `From:`/`Date:`/`Subject:` header so `git am` and the tree's patch checks work. Same applies to `0191-Bluetooth-hci_sync-Add-NVMEM-backed-BD-address-retrieval.patch` and `0192-Bluetooth-qca-Set-NVMEM-BD-address-quirks-when-address-is-invalid.patch`, which have the same problem.

Note the `Subject:` here reads `[PATCH v9 07/10]` — since only 3 of the 10 patches in that series are picked up, please add a note in the commit message (or the patch itself) saying which series revision they come from and why the rest is not needed.

---

target/linux/qualcommax/patches-6.18/0833-bluetooth-introduce-qualcomm-ipq5018-ipc-based-hci-driver.patch
```diff
@@ -0,0 +1,1296 @@
...
++
++		/*
++		 * The virtual instead of the physical address must be used as
++		 * the offset.
++		 */
++		if (phdr->p_vaddr > desc->mem_size ||
++		    phdr->p_filesz > desc->mem_size - phdr->p_vaddr) {
++			dev_err(desc->dev,
++				"Segment data exceeds the reserved memory area!\n");
++			goto release_fw;
```
openwrt-ai, line 910: `ret` is 0 here (it was cleared by the successful `qcom_scm_pas_init_image()` check just above), and this path never sets it before jumping to `release_fw`. So a segment that overflows the reserved region is logged but `btcomqipc_firmware_load()` returns success, and `btqcomipc_open()` goes on to `qcom_scm_pas_auth_and_reset()` on a partially-loaded image.

```suggestion
+			dev_err(desc->dev,
+				"Segment data exceeds the reserved memory area!\n");
+			ret = -EINVAL;
+			goto release_fw;
```

---

target/linux/qualcommax/patches-6.18/0833-bluetooth-introduce-qualcomm-ipq5018-ipc-based-hci-driver.patch
```diff
@@ -0,0 +1,1296 @@
...
++	ret = of_property_read_string(dev->of_node, "firmware-name",
++				      &desc->firmware);
++	if (ret < 0)
++		return ret;
++
++	desc->base = devm_platform_get_and_ioremap_resource(pdev, 0, &res);
++	if (IS_ERR(desc->base)) {
++		dev_err(dev, "unable to acquire memory-region resource\n");
++		return dev_err_probe(dev, PTR_ERR(desc->base),
++				     "unable to acquire memory-region resource\n");
```
openwrt-ai, line 1191: The bare `dev_err()` and the `dev_err_probe()` print the identical string, so this failure logs twice — and the `dev_err()` also defeats the `-EPROBE_DEFER` suppression that `dev_err_probe()` exists for. Every other error path in this probe uses `dev_err_probe()` alone.

```suggestion
+		return dev_err_probe(dev, PTR_ERR(desc->base),
+				     "unable to acquire memory-region resource\n");
```

---

target/linux/qualcommax/patches-6.18/0833-bluetooth-introduce-qualcomm-ipq5018-ipc-based-hci-driver.patch
```diff
@@ -0,0 +1,1296 @@
...
++
++	of_node_put(ipc_np);
++
++	lpo_clk = devm_clk_get_enabled(dev, "lpo");
++	if (IS_ERR(lpo_clk))
++		return dev_err_probe(dev, PTR_ERR(lpo_clk),
++				     "Failed to get lpo clock\n");
++
++	btss_reset = devm_reset_control_get_exclusive_deasserted(dev, NULL);
++	if (IS_ERR_OR_NULL(btss_reset))
```
openwrt-ai, line 1222: If `btss_reset` were ever `NULL`, `PTR_ERR(NULL)` is `0`, and `dev_err_probe(dev, 0, ...)` returns `0` — probe would report success while the reset was never deasserted. `devm_reset_control_get_exclusive_deasserted()` only ever returns an `ERR_PTR` on failure, so plain `IS_ERR()` is both correct and consistent with the `lpo_clk` check two lines above.

```suggestion
+	if (IS_ERR(btss_reset))
```

---

target/linux/qualcommax/patches-6.18/0833-bluetooth-introduce-qualcomm-ipq5018-ipc-based-hci-driver.patch
```diff
@@ -0,0 +1,1296 @@
...
++			goto spin_unlock;
++		}
++	} else {
++		btss_process_ack(desc);
++	}
++
++	rinfo = &(desc->rx_ctxt->sring_buf_info);
++	while (rinfo != NULL) {
++		ret = btss_process_rx(desc, rinfo, &ack,
++				      &desc->rx_ctxt->smsg_buf_cnt);
```
openwrt-ai, line 809: The `rx_count` argument of `btss_process_rx()` is only ever incremented (`(*rx_count)++` at the two `continue_rx` sites), never read — but the pointer passed in aims at `desc->rx_ctxt->smsg_buf_cnt`, a field of the `struct context_info` that lives in the shared BTSS memory region and describes how many short-message buffers the RX ring has. So every received frame bumps a firmware-owned descriptor field, and it wraps at 255 since it is a `u8`.

Is this deliberate (i.e. is the controller expected to consume a processed-frame count in that field), or should this be a driver-local counter? If the counter is genuinely unused, dropping the parameter entirely would be clearer.

---

target/linux/qualcommax/patches-6.18/0833-bluetooth-introduce-qualcomm-ipq5018-ipc-based-hci-driver.patch
```diff
@@ -0,0 +1,1296 @@
...
++static irqreturn_t btss_irq_handler(int irq, void *data)
++{
++	struct qcom_btss *desc = data;
++
++	queue_work(desc->wq, &desc->work);
++
++	return IRQ_HANDLED;
++}
++
++static int btcomqipc_firmware_load(struct qcom_btss *desc)
```
openwrt-ai, line 835: nit: `btcomqipc_` transposes the letters of the driver prefix used everywhere else in the file (`btqcomipc_open`, `btqcomipc_probe`, module name `btqcomipc`).

```suggestion
+static int btqcomipc_firmware_load(struct qcom_btss *desc)
```

The call site at line 981 (`ret = btcomqipc_firmware_load(desc);`) needs the same rename.

---

target/linux/qualcommax/patches-6.18/0833-bluetooth-introduce-qualcomm-ipq5018-ipc-based-hci-driver.patch
```diff
@@ -0,0 +1,1296 @@
...
++	u32 len;
++
++	/* Account for HCI packet type as it's not included in the skb payload */
++	len = skb ? skb->len + 1 : 0;
++
++	if (len > IPC_MSG_PLD_SZ) {
++		hdr |= IPC_HDR_LONG_MSG;
++
++		/* snapshot long message context in case we need to reverse */
++		lmsg_free_cnt = le16_to_cpu(desc->lmsg_ctxt.lmsg_free_cnt);
```
openwrt-ai, line 354: nit: `desc->lmsg_ctxt` is `struct long_msg_info`, a driver-private host-endian structure — its `lmsg_free_cnt` is a plain `u16` that the rest of the driver manipulates with plain arithmetic (`desc->lmsg_ctxt.lmsg_free_cnt -= blks;` in `btss_alloc_lmsg()`, `+= blks` in `btss_free_lmsg()`). Wrapping it in `le16_to_cpu()` here and `cpu_to_le16()` on the restore path at line 374 is a no-op on little-endian but is misleading and will trip sparse.

```suggestion
+		lmsg_free_cnt = desc->lmsg_ctxt.lmsg_free_cnt;
```

Line 374 should drop the matching `cpu_to_le16()`.

---

openwrt-ai (commented):
Reviewed 4 commits (full PR, first review). All four commit messages match the changes they carry, so no commit-message findings.

Inline comments cover the new firmware package's missing mandatory metadata, three concrete error-handling bugs in `btqcomipc.c` (one returning success on a failed firmware segment load), the raw patchwork headers on patches 0190–0192, and a question about the BD-address NVMEM offset colliding with the Ethernet MAC on all five Linksys boards.

---

target/linux/qualcommax/patches-6.18/0834-arm64-dts-qcom-ipq5018-add-node-required-for-bluetooth-support.patch
```diff
@@ -0,0 +1,42 @@
...
+
+Reviewed-by: Bartosz Golaszewski <bartosz.golaszewski@oss.qualcomm.com>
+Signed-off-by: George Moussalem <george.moussalem@outlook.com>
+---
+ arch/arm64/boot/dts/qcom/ipq5018.dtsi | 16 ++++++++++++++++
+ 1 file changed, 16 insertions(+)
+
+--- a/arch/arm64/boot/dts/qcom/ipq5018.dtsi
++++ b/arch/arm64/boot/dts/qcom/ipq5018.dtsi
+@@ -491,6 +491,22 @@
```
openwrt-ai, line 20: This hunk header is stale and is what fails [`Check Kernel patches (qualcommax, ipq50xx)`](https://github.com/openwrt/openwrt/actions/runs/32975269635/job/98199487446):

```
target/linux/qualcommax/patches-6.18/0834-arm64-dts-qcom-ipq5018-add-node-required-for-bluetooth-support.patch
Kernel patches for qualcommax/ipq50xx require refresh. (run 'make target/linux/refresh' and force push this pr)
```

0834 is the only patch the job reports. Replaying the `ipq5018.dtsi` hunks of the target series in order against v6.18.44 gives `Hunk #1 succeeded at 493 (offset 2 lines)` — the preceding USB patches (0823/0824/0826/0827) shifted the insertion point, so the recorded start line is two low. Please regenerate with `make target/linux/refresh` (qualcommax/ipq50xx selected in `.config`); the result is:

```suggestion
@@ -493,6 +493,22 @@
```


---

target/linux/qualcommax/patches-6.18/0190-net-of_net-Add-of_get_nvmem_eui48-helper-for-EUI-48-lookup.patch
```diff
@@ -0,0 +1,125 @@
+From 4278babf76f081a090e0f16c99738c98fdb041de Mon Sep 17 00:00:00 2001
+From: Loic Poulain <loic.poulain@oss.qualcomm.com>
+Date: Thu, 6 Aug 2026 16:38:40 +0200
+Subject: [PATCH v10 07/10] net: of_net: Add of_get_nvmem_eui48() helper for EUI-48 lookup
+ lookup
```
openwrt-ai, line 5: Thanks for stripping the patchwork SMTP headers. The unfolding of the `Subject:` header went wrong though: the original was a folded RFC 5322 header (`... helper for` / continuation ` EUI-48 lookup`), and the trim joined the two halves *and* kept the continuation line, so the last word is now duplicated — a consumer that unfolds this reads the subject as `... helper for EUI-48 lookup lookup`.

```suggestion
Subject: [PATCH v10 07/10] net: of_net: Add of_get_nvmem_eui48() helper for EUI-48 lookup
```

The same double-tail is on the other two patches from that series:

- `0191-Bluetooth-hci_sync-Add-NVMEM-backed-BD-address-retrieval.patch` lines 4–5 — `... BD address retrieval` + stray ` retrieval`
- `0192-Bluetooth-qca-Set-NVMEM-BD-address-quirks-when-address-is-invalid.patch` lines 4–5 — `... when address is invalid` + stray ` is invalid`


---

target/linux/qualcommax/patches-6.18/0192-Bluetooth-qca-Set-NVMEM-BD-address-quirks-when-address-is-invalid.patch
```diff
@@ -0,0 +1,33 @@
+From 4a3be4020ebb88d5c700718caf80de50e5d30327 Mon Sep 17 00:00:00 2001
+From: Loic Poulain <loic.poulain@oss.qualcomm.com>
+Date: Thu, 6 Aug 2026 16:38:42 +0200
+Subject: [v10,09/10] Bluetooth: qca: Set NVMEM BD address quirks when address is invalid
+ is invalid
```
openwrt-ai, line 5: nit: besides the duplicated tail, the prefix here is patchwork's comma form `[v10,09/10]` rather than the mail form `[PATCH v10 09/10]` that 0190 and the whole 0830–0834 block use. Worth normalising so the three patches of the same series read alike.

```suggestion
Subject: [PATCH v10 09/10] Bluetooth: qca: Set NVMEM BD address quirks when address is invalid
```

`0191-Bluetooth-hci_sync-Add-NVMEM-backed-BD-address-retrieval.patch:4` has the same `[v10,08/10]` form.


---

openwrt-ai (commented):
Re-reviewed at `332b6f36` against my last review at `6110fa3f`. Only the first commit changed (`868dda0b` → `60cd6a39`): the three NVMEM/BD-address patches were re-taken from the v10 posting with the patchwork SMTP headers stripped, and 0191's hunk offsets were refreshed. The `net/core/of_net.c` and `net/bluetooth/*` bodies are byte-identical to what was there before, and the 0190 hunks still line up with the tree after `generic/pending-6.18/683-of_net-add-mac-address-to-of-tree.patch`, so nothing regressed there. The other three commits are unchanged and their messages still match their diffs — no commit-message findings.

Three inline comments: the `Check Kernel patches (qualcommax, ipq50xx)` CI failure (0834's hunk header is two lines stale), and the mangled `Subject:` headers left behind by the mbox trim on 0190/0191/0192.

My earlier comments on `package/firmware/qca-bt-firmware/Makefile` (missing `PKG_LICENSE`/`PKG_LICENSE_FILES`, `DEFAULTS` → `DEFAULT`) and on `btqcomipc.c` still stand — the formality bot is still reporting the licence errors.

---

PR #24796, opened 2026-08-19
Add EdgeCORE ECS2100-52T support

luizluca:
~~This PR depends on https://github.com/openwrt/openwrt/pull/24803~~ (merged)


target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
```diff
@@ -0,0 +1,600 @@
...
+		reg = <0x0 0x10000000>;
+	};
+
+	leds {
+		pinctrl-names = "default";
+		pinctrl-0 = <&pinmux_disable_sys_led>;
+		compatible = "gpio-leds";
+
+		led_diag: diag {
+			label = "green:diag";
```
openwrt-ai: `label = "<color>:<function>";` is the deprecated LED binding form; new device trees should use `color` + `function`. There is no `LED_FUNCTION_*` constant for `diag`, so the plain string stays.

```suggestion
			color = <LED_COLOR_ID_GREEN>;
			function = "diag";
```

This also needs `#include <dt-bindings/leds/common.h>` added next to the other includes at line 6.

---

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
```diff
@@ -0,0 +1,600 @@
...
+		PHY_C22(22, 22)
+		PHY_C22(23, 23)
+	};
+};
+
+&mdio_bus1 {
+	/* External phy RTL8218B #4 */
+	ethernet-phy-package@24 {
+		compatible = "ethernet-phy-package";
+		reg = <24>;
```
openwrt-ai: The `reg` of an `ethernet-phy-package` is the MDIO address of the first PHY in the package on that bus, and the binding caps it at 31 ([`reg: maximum: 31` in ethernet-phy-package.yaml](``https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/net/ethernet-phy-package.yaml#L26-L28)).`` Here the switch port numbers are used instead: on `&mdio_bus1` the children sit at MDIO addresses 0-7, 8-15, 16-23 and 24-27 — [`PHY_C22(p, n)` in macros.dtsi](https://github.com/openwrt/openwrt/blob/fdac70471162e2ad475a42fd45ec23ac98f7ec08/target/linux/realtek/dts/macros.dtsi#L9-L13) puts `n` into `reg`, so `PHY_C22(24, 0)` is `ethernet-phy@0 { reg = <0>; }` — while the package nodes say 24/32/40/48. Three of those exceed the 31 maximum.

The identically laid out NETGEAR GS750E (also rtl8393, 6x RTL8218B) gets this right: [`rtl8393_netgear_gs750e.dts:153-193`](https://github.com/openwrt/openwrt/blob/fdac70471162e2ad475a42fd45ec23ac98f7ec08/target/linux/realtek/dts/rtl8393_netgear_gs750e.dts#L153-L193) uses `@0`/`@8`/`@16` on `mdio_bus1`.

```suggestion
	ethernet-phy-package@0 {
		compatible = "ethernet-phy-package";
		reg = <0>;
```

The same applies to lines 422-424 (`ethernet-phy-package@8` / `reg = <8>`), 439-441 (`@16` / `reg = <16>`) and 456-458 (`@24` / `reg = <24>`). `&mdio_bus0` above is already correct.

---

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
```diff
@@ -0,0 +1,600 @@
...
+};
+
+&mdio_bus0 {
+	reset-gpios = <&gpio0 23 GPIO_ACTIVE_LOW>;
+	reset-delay-us = <2000>;
+	reset-post-delay-us = <140000>;
+
+	/* External phy RTL8218B #1 */
+	ethernet-phy-package@0 {
+		compatible = "ethernet-phy-package";
```
openwrt-ai: nit: `compatible = "ethernet-phy-package";` is not part of the binding — the schema keys off `$nodename` and the kernel matches with `of_node_name_eq(child, "ethernet-phy-package")` in `of_mdio.c`, never on a compatible. None of the 56 in-tree DTs that use PHY packages set it. Suggest dropping the property from all seven package nodes (lines 353, 370, 387, 406, 423, 440, 457).

---

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
```diff
@@ -0,0 +1,600 @@
...
+
+		/* 
+		 * PWM signal generated by ADT7470 with a period of 714286ns (~1.4kHz).
+		 * HARDWARE NOTE: The fan uses a 3-pin adapter (DC voltage control).
+		 * Lacking native 4-pin PWM control, the power delivery circuit saturates 
+		 * at low duty cycle values. The real effective control range is between ~4% and ~12%.
+		 * 
+		 * cooling-levels scale calculation (0-255): 
+		 * Formula: ( duty_cycle_ns / 714286ns ) * 255
+		 *
```
openwrt-ai: nit: trailing whitespace on lines 154, 157, 159, 160 and 162.

```suggestion
		/*
		 * PWM signal generated by ADT7470 with a period of 714286ns (~1.4kHz).
		 * HARDWARE NOTE: The fan uses a 3-pin adapter (DC voltage control).
		 * Lacking native 4-pin PWM control, the power delivery circuit saturates
		 * at low duty cycle values. The real effective control range is between ~4% and ~12%.
		 *
		 * cooling-levels scale calculation (0-255):
		 * Formula: ( duty_cycle_ns / 714286ns ) * 255
		 *
```

---

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
```diff
@@ -0,0 +1,600 @@
...
+
+		PHY_C22_SFP(48, 24, 0)
+		PHY_C22_SFP(49, 25, 1)
+		PHY_C22_SFP(50, 26, 2)
+		PHY_C22_SFP(51, 27, 3)
+	};
+};
+
+
+&spi0 {
```
openwrt-ai, line 468: nit: double blank line; the rest of the file separates top-level overrides with a single one.

```suggestion

&spi0 {
```

There is also a stray blank line at line 510, right before the closing brace of `partition@180000`.

---

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
```diff
@@ -0,0 +1,600 @@
...
+				trip = <&switch_very_high>;
+				cooling-device = <&fan0 0 4>;
+			};
+		};
+	};
+};
+
+&ethernet0 {
+	nvmem-cells = <&macaddr_factory 0>;
+	nvmem-cell-names = "mac-address";
```
openwrt-ai, line 331: The commit message for a new device needs a **MAC address layout** section, and `fdac704` has none. This is where it matters: `ethernet0` takes the base MAC from `hwinfo@0x180000` offset 0, and [`02_network`](https://github.com/openwrt/openwrt/blob/fdac70471162e2ad475a42fd45ec23ac98f7ec08/target/linux/realtek/base-files/etc/board.d/02_network#L53-L76) then hands out `lan1`..`lan52` sequentially from the label MAC — none of which is documented anywhere. A few lines along the lines of:

```
MAC addresses:
  label/eth0 : hwinfo 0x00 (xx:xx:xx:xx:xx:00)
  lan1..lan52: label MAC + n, sequential
```

The **Hardware specification** section is also thinner than the rules ask for — it is missing LEDs/buttons, the UART header location/pinout/baud rate, and power input. The DT declares a diag LED, a reset key on `gpio1 33`, a GPIO watchdog and `serial0:115200n8`, so the information exists; the flash instructions in step 1 tell the user to "connect to serial port" without ever saying where it is or at what baud rate.

---

target/linux/realtek/image/Makefile
```diff
@@ -106,6 +106,13 @@ define Build/belkin-header
 	mv $@.new $@
 endef
 
+# Edgecore images use a 100-byte uImage header. The last two bytes contain
+# the board ID stored in the hwinfo partition.
+define Build/edgecore-header
+	uimage_padhdr -i $@ -o $@.new -l 36 -s 0:1.0.0.0 -x 34:$1
```
openwrt-ai: nit: every other host tool in this file is invoked through `$(STAGING_DIR_HOST)/bin/` (`rt-compress`, `7z`, `h3c-image`, `h3c-vfs`, `xikestor-nosimg`, `zynos-firmware`), and the tree spells macro arguments `$(1)`, not `$1`.

```suggestion
	$(STAGING_DIR_HOST)/bin/uimage_padhdr -i $@ -o $@.new -l 36 -s 0:1.0.0.0 -x 34:$(1)
```

---

target/linux/realtek/image/rtl839x.mk
```diff
@@ -9,6 +9,35 @@ define Device/d-link_dgs-1210-52
...
+	rt-loader | \
+	uImage none | \
+	edgecore-header 01d8
+  KERNEL_INITRAMFS := \
+	kernel-bin | \
+	append-dtb | \
+	rt-compress | \
+	rt-loader | \
+	uImage none | \
+	edgecore-header 01d8
```
openwrt-ai: nit: this open-codes the `kernel-bin | append-dtb | rt-compress | rt-loader` chain twice, while the sibling `Device/edgecore_ecs4100-12ph` right below (line 42) pulls it in via `$(Device/uimage-rt-loader)`, which exports `KERNEL/rt-loader` exactly so the header step can be appended.

```suggestion
  $(Device/uimage-rt-loader)
  KERNEL := $$(KERNEL/rt-loader) | uImage none | edgecore-header 01d8
  KERNEL_INITRAMFS := $$(KERNEL/rt-loader) | uImage none | edgecore-header 01d8
```

---

openwrt-ai (commented):
## Commit checks

- `01ee55c` "kernel/realtek: add PWM support" — the `kernel/realtek:` prefix does not exist in this tree. The diff only touches `target/linux/realtek/rtl839x/config-6.18`, so the conventional subject is `realtek: rtl839x: ...` (cf. `realtek: rtl930x: add support for Hasivo S600WP-5GT-2S+`, `realtek: add Realtek PSE MCU driver`). Same slash-prefix issue in `fdac704` (`realtek/rtl839x:`), which the formality bot already flagged for the capitalization.

The four `851-*` backport patches apply cleanly onto v6.18 hwmon sources, the `851` slot is free in `backport-6.18/`, and the `-vX.Y-` naming matches the tree's convention for not-yet-released versions. `uimage_padhdr`'s `-s`/`-x` options are present at the pinned firmware-utils revision (added by `c9da4861`, pulled in by `36946e2993`), so no tools bump is needed.


---

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
luizluca: OK. Changed

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
luizluca: dropped as requested.

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
luizluca: cleaned.

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
luizluca: removed.

target/linux/realtek/image/Makefile
luizluca: added the path as requested.

target/linux/realtek/image/rtl839x.mk
luizluca: OK

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
luizluca: I improved the vendor mac address generation comment. However, OpenWrt uses a different logic, with lan1 sharing the label mac and lan2.. using label mac + (n-1) with bit 2 set in the first octet.

luizluca: > The **Hardware specification** section is also thinner than the rules ask for — it is missing LEDs/buttons, the UART header location/pinout/baud rate, and power input. The DT declares a diag LED, a reset key on `gpio1 33`, a GPIO watchdog and `serial0:115200n8`, so the information exists; the flash instructions in step 1 tell the user to "connect to serial port" without ever saying where it is or at what baud rate.

I'm not sure what is missing.

target/linux/generic/backport-6.18/851-01-v7.3-dt-bindings-hwmon-add-adi-adt7470.patch
```diff
@@ -0,0 +1,84 @@
+From 9f446d82bdbcea89c2795a8a6ce7a7c2cac02129 Mon Sep 17 00:00:00 2001
+From: Luiz Angelo Daros de Luca <luizluca@gmail.com>
+Date: Wed, 5 Aug 2026 10:15:17 -0300
+Subject: [PATCH 1/4] dt-bindings: hwmon: add adi,adt7470
+
+The Analog Devices ADT7470 is a multichannel temperature sensor and
```
plappermaul, line 6: Make the backports a separate PR.

plappermaul (changes requested):

target/linux/generic/backport-6.18/851-01-v7.3-dt-bindings-hwmon-add-adi-adt7470.patch
luizluca: Sure: https://github.com/openwrt/openwrt/pull/24803.

luizluca: Sure. I put the backport commit as the first one and opened it as a new PR. I'll update this PR message.

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
```diff
@@ -0,0 +1,598 @@
...
+		#size-cells = <0>;
+
+		adt7470: fan-controller@2f {
+			compatible = "adi,adt7470";
+			reg = <0x2f>;
+			#pwm-cells = <3>;
+		};
+
+		lm75b: lm75b@48 {
+			compatible = "nxp,lm75a";
```
openwrt-ai: `nxp,lm75a` is not a documented compatible: [lm75.yaml lists `national,lm75` / `national,lm75a` / `national,lm75b`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/hwmon/lm75.yaml#L23-L25), and the driver's OF table has no `nxp,` LM75 entry at all ([lm75.c:901-906](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/hwmon/lm75.c#L901-L906) — the NXP prefixes it knows are `p3t*`/`pct2075`).

It binds today only via the I2C modalias fallback: [`of_alias_from_compatible()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/of/base.c#L1170-L1181) strips the vendor prefix, so [i2c-core-of.c:30](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/i2c/i2c-core-of.c#L30) registers the client as type `lm75a` and it matches `lm75_ids`. That picks the **lm75a** entry — [9-bit default resolution](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/hwmon/lm75.c#L201-L208) — while the node name and label say lm75b (11-bit). The node name should also be the generic class name rather than the part name.

```suggestion
		lm75b: temperature-sensor@48 {
			compatible = "national,lm75b";
```

Note this raises the sample resolution from 9 to 11 bits; if the fitted part is actually an LM75A, use `national,lm75a` and drop the `b` from the label/node instead. [`rtl8393_edgecore_ecs4100-12ph.dts:113-114`](https://github.com/openwrt/openwrt/blob/68400dcc5de11e475e42a31b7aee38cd4d21f7a0/target/linux/realtek/dts/rtl8393_edgecore_ecs4100-12ph.dts#L113-L114) carries the same `nxp,lm75a` string — that one is out of scope here, but worth a separate fix.

---

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
```diff
@@ -0,0 +1,598 @@
...
+		i2c-gpio,delay-us = <5>;
+		#address-cells = <1>;
+		#size-cells = <0>;
+	};
+
+	sfp0: sfp-p49 {
+		compatible = "sff,sfp";
+		i2c-bus = <&i2c0>;
+		los-gpio = <&gpio0 12 GPIO_ACTIVE_HIGH>;
+		mod-def0-gpio = <&gpio1 8 GPIO_ACTIVE_LOW>;
```
openwrt-ai: nit: the `sff,sfp` binding spells these plural — [`los-gpios` / `mod-def0-gpios` in sff,sfp.yaml](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/net/sff%2Csfp.yaml#L32-L41). The singular form works only because the gpiolib lookup falls back to the `-gpio` suffix, and it fails `dtbs_check`. Newer realtek DTs use the plural form, e.g. [`rtl9302_xikestor_sks8300-12e2t2x.dts:66-67`](``https://github.com/openwrt/openwrt/blob/68400dcc5de11e475e42a31b7aee38cd4d21f7a0/target/linux/realtek/dts/rtl9302_xikestor_sks8300-12e2t2x.dts#L66-L67).``

```suggestion
		los-gpios = <&gpio0 12 GPIO_ACTIVE_HIGH>;
		mod-def0-gpios = <&gpio1 8 GPIO_ACTIVE_LOW>;
```

Same applies to lines 86-87, 102-103 and 118-119.

---

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
```diff
@@ -0,0 +1,598 @@
...
+
+	keys: keys {
+		compatible = "gpio-keys-polled";
+		poll-interval = <20>;
+
+		reset {
+			label = "reset";
+			gpios = <&gpio1 33 GPIO_ACTIVE_LOW>;
+			linux,code = <KEY_RESTART>;
+		};
```
openwrt-ai, line 55: Answering "I'm not sure what is missing" from [the earlier thread](https://github.com/openwrt/openwrt/pull/24796#discussion_r3815504082) — concretely, against `68400dc`'s `Hardware:` list:

* **LEDs/Buttons** — no entry at all, yet the DT describes exactly one green diag LED on `gpio0 0` (driving boot/failsafe/running/upgrade) and one reset button on `gpio1 33` (`KEY_RESTART`), both in this hunk. One line each.
* **UART** — "RJ45 RS232 port on front panel" gives the location but not the settings; `chosen/stdout-path` says `115200n8`, and install step 1 tells the user to "connect to serial port" without ever naming a baud rate.
* **Power** — no input voltage/current line.
* **MAC address layout** — still no section. The `&ethernet0` comment added in this revision documents the *vendor* layout, but what an OpenWrt user sees is what [`realtek_setup_macs_lan()`](https://github.com/openwrt/openwrt/blob/68400dcc5de11e475e42a31b7aee38cd4d21f7a0/target/linux/realtek/base-files/etc/board.d/02_network#L29-L40) does for the group this board was added to: `lan1` = label MAC, `lan2`..`lan52` = label MAC with the locally-administered bit set, incrementing. Since that differs from the vendor scheme now in the DTS, spelling both out in the commit message (`label/eth0: hwinfo 0x00, xx:xx:xx:xx:xx:xx` + the per-port rule) is what makes the device reproducible from the message alone.

---

openwrt-ai (commented):
Re-reviewed at `68400dc` (branch was rewritten; all three commits carry new SHAs). Each commit message now matches its diff — including the split into `kernel:` / `realtek:` / device commits — so no commit-check findings this round.

Verified since the last pass:

- The `drivers/hwmon/Kconfig` hunk-header refreshes are complete and consistent: `851-03` adds one line inside `config SENSORS_ADT7470`, and all four in-tree patches that touch that file later (`bcm27xx/0310`, `microchipsw/0010-v7.1`, `mvebu/903`, `realtek/813`) are shifted by exactly +1. No other patch under `target/linux/*/patches-6.18/` or `target/linux/generic/*-6.18/` touches `drivers/hwmon/Kconfig`.
- `&mdio_bus1` PHY-package `reg`s are now the first MDIO address on that bus (0/8/16/24) and match the `PHY_C22*` children; the stray `compatible = "ethernet-phy-package"` is gone.
- `Device/edgecore_ecs2100-52t` now reuses `$(Device/uimage-rt-loader)`'s `KERNEL/rt-loader`, and `edgecore-header` calls `uimage_padhdr` through `$(STAGING_DIR_HOST)/bin/`.
- `kmod-thermal`, `kmod-hwmon-pwmfan`, `kmod-hwmon-adt7470`, `kmod-hwmon-lm75` and `kmod-eeprom-at24` all exist and `kmod-thermal` matches the sibling `edgecore_ecs4100-12ph` block; `CONFIG_SFP`, `CONFIG_I2C_GPIO` and `CONFIG_MDIO_I2C` are already set for rtl839x, so only the new `CONFIG_PWM*` symbols were needed.
- `pinmux_disable_sys_led`, `serdes12` and CPU `port@52` all exist in `rtl839x.dtsi`, and the `cooling-levels` values match the documented duty cycles for a 714286 ns period (`level * period / 255`).

Three inline comments, one of them answering your open question about the commit-message hardware section.

---

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
luizluca: Changed. It works better now.

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
luizluca: Done

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
luizluca: I'll add more HW info

openwrt-ai (commented):
Re-reviewed at `bf5a753`. Only the device commit changed since `68400dc` (the two backport/PWM commits carry identical SHAs), and its diff is limited to `rtl8393_edgecore_ecs2100-52t.dts`. Reviewed 1 new commit; no new issues found.

Verified this round:

- `los-gpio`/`mod-def0-gpio` → `los-gpios`/`mod-def0-gpios` on all four SFP cages, matching [`sff,sfp.yaml`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/Documentation/devicetree/bindings/net/sff%2Csfp.yaml#L32-L41). No other `-gpio` singular properties remain in the file.
- `national,lm75b` is in both the binding enum and the driver OF table ([`lm75.c:905-906`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/hwmon/lm75.c#L905-L906)), so it now binds through the OF match rather than the I2C modalias fallback and gets the 11-bit `lm75b` config. The `&lm75b` phandle referenced by the `thermal-sensors` property is unchanged.
- The commit message gained the LED, button and `115200n8` details that were missing before; they match the DT (single green `diag` LED on `gpio0 0`, `KEY_RESTART` on `gpio1 33`, `stdout-path = "serial0:115200n8"`).
- CI on this head is green: both build workflows and all three FormalityCheck contexts pass.

The two still-open points from the previous round (MAC address layout section, power input line in the hardware spec) are unchanged — the existing threads cover them, not repeating them here.

---

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
jonasjelonek: The bot wants to have that in the commit message.

jonasjelonek:
Please also add a small comment to the commit message that you're adding the `Build/edgecore-header` recipe.

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
```diff
@@ -0,0 +1,598 @@
...
+	};
+};
+
+&mdio_bus0 {
+	reset-gpios = <&gpio0 23 GPIO_ACTIVE_LOW>;
+	reset-delay-us = <2000>;
+	reset-post-delay-us = <140000>;
+
+	/* External phy RTL8218B #1 */
```
openwrt-ai, line 357: nit: the `Hardware:` list in the commit message still names no PHY or peripheral chips, although the DTS identifies all of them: 6x RTL8218B (this comment plus lines 373, 389, 407, 423, 439), one RTL8214FC for the four SFP cages (line 455), an RTL8231 GPIO expander on `&mdio_aux` (line 337), and — most relevant to this series — the ADT7470 fan controller (line 130) and the LM75B temperature sensor (line 136) that the PWM/thermal backport in `c39956e` exists for.

The sibling commits carry that level of detail: [`fa9f9259` "realtek/rtl839x: Edgecore ECS4100-12PH support"](https://github.com/openwrt/openwrt/commit/fa9f92595197e3e010cdc209089ff8ce5235876a) spells out "8x GbE RJ45 PoE (external RTL8218B)" and "2x SFP (external RTL8214FC)", and [`0c981506` "realtek: add support for Zyxel GS1920-24HPv2"](https://github.com/openwrt/openwrt/commit/0c98150650705ed5cf130c6258bca057a4fc36d4) has a dedicated "Fan controller: ADT7468" line. Two extra qualifiers under `Ethernet:` and one line for the fan controller / temperature sensor would close this out.

The rest of the message rework checks out: the MAC address layout section matches what [`realtek_setup_macs_lan()`](https://github.com/openwrt/openwrt/blob/a87960bdb8beb71258e0236cf396e76f1734bee1/target/linux/realtek/base-files/etc/board.d/02_network#L29-L40) does for the group this board joined (lan1 = label MAC, lan2.. = LA bit set, then incrementing), the vendor flash layout matches `partition@0`/`@100000`/`@180000`/`@200000` below, and the LED/button/UART/power lines match the DT.

---

openwrt-ai (commented):
Re-reviewed at `a87960b`. The branch was rewritten again — all three commits carry new SHAs — but `git diff bf5a753..a87960b` is empty, so nothing changed except the commit messages. Reviewed 3 new commits; each message matches its diff, so no commit-check findings.

Message-only deltas since `bf5a753`:

- `c39956e`: subject narrowed from "backport adt7470 fixes and PWM/thermal support" to "backport adt7470 PWM/thermal support". Accurate — the four `851-0*` patches are dt-bindings, `ADT7470_PWM_MAX`, the PWM-framework exposure and the thermal-zone sensor; none of them is a fix.
- `a8cc2fab`: unchanged text, still matches the two added symbols (`CONFIG_PWM`, `CONFIG_PWM_SYSFS`) in `rtl839x/config-6.18`; `edgecore_ecs4100-12ph`, the profile the footprint was measured on, is indeed an rtl839x device.
- `a87960b`: gained the power-input line, the vendor flash layout, the vendor + OpenWrt MAC address sections and the `Build/edgecore-header` note that @jonasjelonek asked for. All of these check out against the DTS partitions and against `02_network`. The `realtek/rtl839x:` prefix has in-tree precedent for this exact family (`fa9f9259`), so disregard my earlier remark about it.

CI on this head has nothing failing: the `Build Kernel` run is still in progress, with every completed job so far green or skipped.

One nit inline about the hardware spec; the still-open thread from the earlier round is otherwise resolved by the new message.

---

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
```diff
@@ -0,0 +1,598 @@
...
+				speed = <1000>;
+				full-duplex;
+			};
+		};
+	};
+};
+
+&uart1 {
+	status = "okay";
+};
```
plappermaul: Please explain in the commit message why you activate the second UART.

plappermaul (changes requested):

target/linux/realtek/dts/rtl8393_edgecore_ecs2100-52t.dts
luizluca: No need to explain. It is not used. It was a copy-paste leftover from the other ECS switch (which has PoE) and I missed the fact that uart1 a second uart. I'll remove the unused uart1 in the next revision.

openwrt-ai (commented):
Re-reviewed at `1506c22`. Reviewed 2 new commits; no new issues found.

The branch was rebased onto `54f7b24` and the backport commit is gone — the four `851-0*-v7.3-*` adt7470 patches are now in `target/linux/generic/backport-6.18/` on `main` via #24803, so this PR is correctly down to the PWM config commit plus the device commit.

Changes since `a87960b`:

- **`f1a6ac8`** "realtek: add PWM support" — message text is byte-identical to `a8cc2fab`, diff is still the same two symbols (`CONFIG_PWM`, `CONFIG_PWM_SYSFS`) in `rtl839x/config-6.18`. Matches.
- **`1506c22`** "realtek/rtl839x: add Edgecore ECS2100-52T support" — the only content change is the removal of the `&uart1 { status = "okay"; };` override that @plappermaul asked about. Correct to drop: `chosen/stdout-path` is `serial0`, which [`rtl839x.dtsi:95`](https://github.com/openwrt/openwrt/blob/1506c22791e52b8d5b6437919d6c8dc9f968f0c7/target/linux/realtek/dts/rtl839x.dtsi#L95) aliases to `&uart0`, and nothing else in the DTS references `uart1`.

The `Hardware:` list also gained the chip identifications from the last round's nit, and each one checks out against the DTS: "6x external RTL8218B" matches the six package comments at lines 357/373/389/407/423/439 (6 x 8 = 48 ports), "1x external RTL8214FC" matches line 455 (4 SFP cages), `RTL8231` matches the `realtek,rtl8231` expander at line 338, and the `adi,adt7470` (line 130) / `national,lm75b` (line 136) nodes back the fan-controller and temp-sensor lines. The "not used as temp sensor" qualifier is accurate — `switch-thermal` binds `thermal-sensors = <&lm75b>`, not the ADT7470.

Also re-verified: `IMAGE_SIZE := 30720k` matches `partition@200000`'s `0x1e00000`, and CI on this head is fully green (all seven realtek build jobs, the patch check, `build-pr-profile`, and all three FormalityCheck contexts).

---

target/linux/realtek/image/rtl839x.mk
```diff
@@ -9,6 +9,24 @@ define Device/d-link_dgs-1210-52
...
 TARGET_DEVICES += d-link_dgs-1210-52
 
+define Device/edgecore_ecs2100-52t
+  $(Device/uimage-rt-loader)
+  SOC := rtl8393
+  DEVICE_VENDOR := Edgecore
+  DEVICE_MODEL := ECS2100-52T
+  IMAGE_SIZE := 30720k
+  # 0x01d8 is the ECS2100-52T board ID
+  KERNEL := $$(KERNEL/rt-loader) | uImage none | edgecore-header 01d8
```
plappermaul: Final nit: Add a new variable like ECS_BOARD_ID to image/Makefile, Set it in the recipe. That is self explaining. Do not hand it over to the edcore-header function but directly reference it inside that.

plappermaul (changes requested):

target/linux/realtek/image/rtl839x.mk
luizluca: Sure. I'll also remove the comment here as the variable is already clear enough

target/linux/realtek/image/Makefile
```diff
@@ -106,6 +106,13 @@ define Build/belkin-header
 	mv $@.new $@
 endef
 
+# Edgecore images use a 100-byte uImage header. The last two bytes contain
+# the board ID stored in the hwinfo partition.
+define Build/edgecore-header
+	$(STAGING_DIR_HOST)/bin/uimage_padhdr -i $@ -o $@.new -l 36 -s 0:1.0.0.0 -x 34:$(ECS_BOARD_ID)
```
openwrt-ai, line 113: `ECS_BOARD_ID` is set inside `define Device/edgecore_ecs2100-52t` but never registered in `DEVICE_VARS`, so [`Device/Export`](https://github.com/openwrt/openwrt/blob/c2a66db42bf4bbcb58c460528b9763d858045d76/include/image.mk#L585-L588) never turns it into a target-specific variable. This recipe is expanded at rule-run time ([`$(call concat_cmd,$(KERNEL))` in image.mk:759](https://github.com/openwrt/openwrt/blob/c2a66db42bf4bbcb58c460528b9763d858045d76/include/image.mk#L759)), so `$(ECS_BOARD_ID)` resolves against the *global* value — whatever the last `define Device/...` evaluated in `rtl839x.mk` happened to leave behind — not against this device's value.

It produces `01d8` today only because exactly one device assigns it. Add a second Edgecore board with a different ID and both images silently get the same board ID. That is not hypothetical: `LINKSYS_MODEL` is unregistered in the same way, and [`rtl930x_nand.mk:28`](https://github.com/openwrt/openwrt/blob/c2a66db42bf4bbcb58c460528b9763d858045d76/target/linux/realtek/image/rtl930x_nand.mk#L28) / [`:37`](https://github.com/openwrt/openwrt/blob/c2a66db42bf4bbcb58c460528b9763d858045d76/target/linux/realtek/image/rtl930x_nand.mk#L37) set two different values in one subtarget, where one already wins for both devices.

Every other per-device variable consumed by a `Build/` recipe in this file is registered — `BELKIN_HEADER`, `CAMEO_*`, `FLASH_ADDR`, `H3C_*`, `ZYNFW_*`, `ZYXEL_VERS` in [the `DEVICE_VARS +=` list at lines 8-23](https://github.com/openwrt/openwrt/blob/c2a66db42bf4bbcb58c460528b9763d858045d76/target/linux/realtek/image/Makefile#L8-L23). `ECS_BOARD_ID` belongs there too (alphabetically between `CAMEO_ROOTFS_PART` and `FLASH_ADDR`).

---

openwrt-ai (commented):
Re-reviewed at `c2a66db`. Only the device commit changed since `1506c22` — `f1a6ac8` carries the same SHA, and the device commit's message text is byte-identical, so this is a code-only revision. Reviewed 1 new commit; the message still matches its diff, so no commit-check findings.

The whole delta is the `edgecore-header 01d8` → `ECS_BOARD_ID` refactor @plappermaul asked for in [this thread](https://github.com/openwrt/openwrt/pull/24796#discussion_r3835486671): the literal argument is gone from both `KERNEL` and `KERNEL_INITRAMFS`, and `Build/edgecore-header` now reads the variable directly. The recipe body is otherwise unchanged (`-l 36 -s 0:1.0.0.0 -x 34:`), so the emitted header is identical.

One inline issue: the new variable is missing from `DEVICE_VARS`, which is the second half of that request — without it the value reaches the recipe only by accident.

Nothing failing on CI for this head: the three FormalityCheck contexts and `build-pr-profile` are green, and the realtek kernel jobs are still queued.

---

plappermaul:

PR #24798, opened 2026-08-19
econet: en7528: add support for four JioFiber EN7528 ONTs

naseef:
This adds support for four ISP-provided GPON ONTs , all built on the EcoNet EN7528. They cover two board designs, each with a second product sharing its internals:

| Device              | Covered as                 | Notes                                 |
| ------------------- | -------------------------- | ------------------------------------- |
| JioFiber JCOW414    | primary                    |                                       |
| JioFiber JCO4032    | DEVICE_ALT0 of the JCOW414 | different form factor, same internals |
| JioFiber JCOW407    | primary                    |                                       |
| JioFiber JCOW407-IN | DEVICE_ALT0 of the JCOW407 | second-source SPI NAND                |

**Commits**

1. econet: make the TRX header and jump stub configurable per device

Preparatory, no behaviour change for existing devices:

- tclinux-trx.sh gains --loadaddr; the address the bootloader jumps to  was previously hardcoded to 0x80020000.
- TRX_MODEL and the new TRX_LOADADDR become per-device via DEVICE_VARS, and are reset in Device/Default. TRX_MODEL was a plain global, so the last device to set it won for every device in the target.
- The free-bootbase jump stub is emitted according to CONFIG_BIG_ENDIAN; it was unconditionally big-endian, which is wrong on the little-endian EN7528 subtarget.

2. econet: en7528: add support for JioFiber JCOW414/JCO4032

Introduces en7528_jiofiber.dtsi holding everything common to Jio's EN7528 boards  memory, flash layout, mtd-concat rootfs, PCIe, WiFi, switch and USB , leaving board files to describe only LEDs and buttons.

3. econet: en7528: add support for JioFiber JCOW407/JCOW407-IN

Adds only the second board file plus its device entry; reuses the dtsi unchanged.

**Hardware**

Common to all four:

- SoC EcoNet EN7528
- RAM 512 MiB DDR3
- Flash 256 MiB SPI NAND — Micron MT29F2G01, or Winbond W25M02G on the JCOW407
- WiFi MT7603E 2.4 GHz on PCIe port 0, MT7663 802.11ac 5 GHz on PCIe port 1
- Ethernet 4× 10/100/1000 behind the on-die MT7530 switch
- USB 2 ports
- Other 1× FXS, 1 GPON port

Where the two board designs differ:

|                 | JCOW414 / JCO4032 | JCOW407 / JCOW407-IN |
| --------------- | ----------------- | -------------------- |
| RGB status LED  | GPIO 27 / 25 / 23 | GPIO 2 / 3 / 4       |
| WPS button      | GPIO 4            | GPIO 1               |
| LAN LED colours | green on even pin | green on odd pin     |

**Installation**

Both boards are dual-image. OpenWrt goes in the first slot (tclinux); the bootloader picks the slot from one ASCII byte at flash offset 0x0DFC0000. Which slot the vendor firmware boots from varies between units, so the flag may need moving after flashing.

Full flash instructions : TFTP and XMODEM recovery, permanent install, slot switching from OpenWrt, from the vendor firmware, and from the bootloader , are in the individual commit messages, along with the MAC layout.

**Known limitation**

On the JCOW407 the green LAN LEDs are described in DT but stay dark: that bootloader leaves their pins muxed to the GPHY LED function, so the GPIO output has no effect. They start working once econet pinctrl support lands. The amber LEDs are unaffected, and all eight work on the JCOW414.

### DIFF
--- a/target/linux/econet/dts/en7528_jiofiber.dtsi
+++ b/target/linux/econet/dts/en7528_jiofiber.dtsi
@@ -1,0 +1,235 @@
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+
+#include "en7528.dtsi"
+
+#include <dt-bindings/gpio/gpio.h>
+#include <dt-bindings/input/input.h>
+#include <dt-bindings/leds/common.h>
+
+/ {
+	memory@0 {
+		device_type = "memory";
+		reg = <0x00000000 0x1c000000>;
+	};
+
+	chosen {
+		stdout-path = "/serial@1fbf0000:115200";
+		linux,usable-memory-range = <0x00020000 0x1bfe0000>;
+	};
+
+	reg_usb1_vbus: regulator-usb1-vbus {
+		compatible = "regulator-fixed";
+		regulator-name = "usb1-vbus";
+		regulator-min-microvolt = <5000000>;
+		regulator-max-microvolt = <5000000>;
+		gpios = <&gpio0 9 GPIO_ACTIVE_HIGH>;
+		enable-active-high;
+	};
+
+	/* Second USB port VBUS, no DT consumer, enabled unconditionally */
+	reg_usb2_vbus: regulator-usb2-vbus {
+		compatible = "regulator-fixed";
+		regulator-name = "usb2-vbus";
+		regulator-min-microvolt = <5000000>;
+		regulator-max-microvolt = <5000000>;
+		gpios = <&gpio0 11 GPIO_ACTIVE_HIGH>;
+		enable-active-high;
+		regulator-boot-on;
+		regulator-always-on;
+	};
+
+	virtual_flash {
+		compatible = "mtd-concat";
+		devices = <&rootA &root1>;
+
+		partitions {
+			compatible = "fixed-partitions";
+			#address-cells = <1>;
+			#size-cells = <1>;
+
+			partition@0 {
+				label = "ubi";
+				reg = <0x0 0x0>;
+			};
+		};
+	};
+};
+
+&usb {
+	status = "okay";
+	vbus-supply = <&reg_usb1_vbus>;
+};
+
+&pcie0 {
+	status = "okay";
+};
+
+&pcie1 {
+	status = "okay";
+};
+
+&slot0 {
+	status = "okay";
+
+	wifi@0,0 {
+		compatible = "mediatek,mt76";
+		reg = <0x0000 0 0 0 0>;
+		nvmem-cells = <&macaddr_bl 2>, <&eeprom_reservearea_1c0000>;
+		nvmem-cell-names = "mac-address", "eeprom";
+		ieee80211-freq-limit = <2400000 2500000>;
+	};
+};
+
+&slot1 {
+	status = "okay";
+
+	wifi@0,0 {
+		compatible = "mediatek,mt76";
+		reg = <0x0000 0 0 0 0>;
+		nvmem-cells = <&macaddr_bl 3>, <&eeprom_reservearea_4b000>;
+		nvmem-cell-names = "mac-address", "eeprom";
+		ieee80211-freq-limit = <5000000 6000000>;
+	};
+};
+
+&gmac0 {
+	status = "okay";
+	nvmem-cells = <&macaddr_bl 1>;
+	nvmem-cell-names = "mac-address";
+};
+
+&switch {
+	status = "okay";
+};
+
+&gsw_port1 {
+	label = "lan1";
+	status = "okay";
+};
+
+&gsw_port2 {
+	label = "lan2";
+	status = "okay";
+};
+
+&gsw_port3 {
+	label = "lan3";
+	status = "okay";
+};
+
+&gsw_port4 {
+	label = "lan4";
+	status = "okay";
+};
+
+&nand {
+	status = "okay";
+	econet,bmt;
+	econet,bbt-table-size = <250>;
+
+	partitions {
+		compatible = "fixed-partitions";
+		#address-cells = <1>;
+		#size-cells = <1>;
+
+		partition@0 {
+			label = "bootloader";
+			reg = <0x0 0x40000>;
+			read-only;
+
+			nvmem-layout {
+				compatible = "fixed-layout";
+				#address-cells = <1>;
+				#size-cells = <1>;
+
+				macaddr_bl: macaddr@3fe48 {
+					compatible = "mac-base";
+					reg = <0x3fe48 0x6>;
+					#nvmem-cell-cells = <1>;
+				};
+			};
+		};
+
+		partition@40000 {
+			label = "romfile";
+			reg = <0x40000 0x40000>;
+			read-only;
+		};
+
+		partition@80000 {
+			label = "tclinux";
+			reg = <0x80000 0x2800000>;
+			econet,enable-remap;
+		};
+
+		/* Nested inside of tclinux */
+		partition1@80000 {
+			label = "tclinux_kernel";
+			reg = <0x80000 0x400000>;
+			econet,enable-remap;
+		};
+
+		/* Concatenated with "unknown" to form the UBI volume */
+		rootA: partition1@480000 {
+			label = "tclinux_rootfs";
+			reg = <0x480000 0x2400000>;
+		};
+
+		partition@2880000 {
+			label = "tclinux_slave";
+			reg = <0x2880000 0x2800000>;
+		};
+
+		partition@5080000 {
+			label = "opt0";
+			reg = <0x5080000 0x800000>;
+		};
+
+		partition@5880000 {
+			label = "opt1";
+			reg = <0x5880000 0x800000>;
+		};
+
+		partition@6080000 {
+			label = "ubifs";
+			reg = <0x6080000 0x1a00000>;
+		};
+
+		partition@7a80000 {
+			label = "config";
+			reg = <0x7a80000 0x200000>;
+		};
+
+		partition@7c80000 {
+			label = "ubifs2";
+			reg = <0x7c80000 0x5a00000>;
+		};
+
+		/* Concatenated with tclinux_rootfs to form the UBI volume */
+		root1: partition@d680000 {
+			label = "unknown";
+			reg = <0xd680000 0x740000>;
+		};
+
+		partition@ddc0000 {
+			label = "reservearea";
+			reg = <0xddc0000 0x240000>;
+
+			nvmem-layout {
+				compatible = "fixed-layout";
+				#address-cells = <1>;
+				#size-cells = <1>;
+
+				/* 5GHz */
+				eeprom_reservearea_4b000: eeprom@4b000 {
+					reg = <0x4b000 0x4da8>;
+				};
+
+				/* 2.4GHz */
+				eeprom_reservearea_1c0000: eeprom@1c0000 {
+					reg = <0x1c0000 0x400>;
+				};
+			};
+		};
+	};
+};
--- a/target/linux/econet/dts/en7528_jiofiber_jcow414.dts
+++ b/target/linux/econet/dts/en7528_jiofiber_jcow414.dts
@@ -1,0 +1,111 @@
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+/dts-v1/;
+
+#include "en7528_jiofiber.dtsi"
+
+/ {
+	model = "JioFiber JCOW414";
+	compatible = "jiofiber,jcow414", "econet,en7528";
+
+	aliases {
+		led-boot = &led_status_red;
+		led-failsafe = &led_status_red;
+		led-running = &led_status_green;
+		led-upgrade = &led_status_blue;
+	};
+
+	leds {
+		compatible = "gpio-leds";
+
+		led_status_red: status-red {
+			color = <LED_COLOR_ID_RED>;
+			function = LED_FUNCTION_STATUS;
+			gpios = <&gpio0 27 GPIO_ACTIVE_LOW>;
+		};
+
+		led_status_green: status-green {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_STATUS;
+			gpios = <&gpio0 25 GPIO_ACTIVE_LOW>;
+		};
+
+		led_status_blue: status-blue {
+			color = <LED_COLOR_ID_BLUE>;
+			function = LED_FUNCTION_STATUS;
+			gpios = <&gpio0 23 GPIO_ACTIVE_LOW>;
+		};
+
+		led-lan1-green {
+			color = <LED_COLOR_ID_GREEN>;
+			function = LED_FUNCTION_LAN;
+			function-enumerator = <1>;
+			gpios = <&gpio0 30 GPIO_ACTIVE_LOW>;
+		};
+
+		led-lan1-amber {
+			color = <LED_COLOR_ID_AMBER>;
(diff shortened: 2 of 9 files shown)


naseef:
@cjdelisle Please review when you have time, especially the commit _econet: make the TRX header and jump stub configurable per device_

target/linux/econet/dts/en7528_jiofiber_jcow414.dts
```diff
@@ -0,0 +1,112 @@
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+/dts-v1/;
+
+#include "en7528_jiofiber.dtsi"
+
+/ {
+	model = "JioFiber JCOW414";
+	compatible = "jiofiber,jcow414", "jiofiber,jco4032",
+		     "econet,en7528";
```
openwrt-ai: `jiofiber,jco4032` is a sibling product, not a more generic form of `jiofiber,jcow414`, so it does not belong in this fallback chain. `compatible` is an ordered most-specific-to-most-generic list; every in-tree multi-entry example is a generalisation chain (e.g. [`"zbtlink,zbt-wg3526-32m", "zbtlink,zbt-wg3526", "mediatek,mt7621-soc"`](https://github.com/openwrt/openwrt/blob/25b08cdf8cb50dab4a40d7822981a1f2dc5e42f3/target/linux/ramips/dts/mt7621_zbtlink_zbt-wg3526-32m.dts#L4)), and other `DEVICE_ALT0_*` devices carry only the primary compatible (e.g. [`xiaomi,mi-router-4a-gigabit`](https://github.com/openwrt/openwrt/blob/25b08cdf8cb50dab4a40d7822981a1f2dc5e42f3/target/linux/ramips/dts/mt7621_xiaomi_mi-router-4a-gigabit.dts#L6)).

The extra string is also inert: [`02_sysinfo`](https://github.com/openwrt/openwrt/blob/25b08cdf8cb50dab4a40d7822981a1f2dc5e42f3/package/base-files/files/lib/preinit/02_sysinfo#L5) takes only the first entry for `board_name`, and `SUPPORTED_DEVICES` is derived from `DEVICE_NAME`, so nothing keys off it. The `DEVICE_ALT0_MODEL` in `en7528.mk` already documents the second product.

```suggestion
	compatible = "jiofiber,jcow414", "econet,en7528";
```

Same applies to `en7528_jiofiber_jcow407.dts:8-9`, where the ordering is additionally inverted: `jiofiber,jcow407-in` is the narrower variant (second-source NAND) but is listed as the *fallback* of `jiofiber,jcow407`.

---

target/linux/econet/image/Makefile
```diff
@@ -31,7 +35,7 @@ define Build/tclinux-free-bootbase-jump
 		$(call ERROR_MESSAGE,WARNING: KERNEL_LOADADDR changed); \
 		rm -f $@; \
 	}
-	printf '\010\000\200\000' > $@.new
+	printf '$(if $(filter le,$(TRX_ENDIAN)),\000\200\000\010,\010\000\200\000)' > $@.new
```
openwrt-ai: The stub is a MIPS `j 0x80020000` instruction inside the kernel image, so its byte order follows the *target CPU* endianness, not the TRX container format. `TRX_ENDIAN` only happens to agree here because each subtarget sets both consistently — and the one existing user of this macro, [`Device/huawei_hg2821t-u`](https://github.com/openwrt/openwrt/blob/25b08cdf8cb50dab4a40d7822981a1f2dc5e42f3/target/linux/econet/image/en751221.mk#L36), emits a raw `kernel.bin` with no TRX header at all, so `TRX_ENDIAN` is not even meaningful in that path.

OpenWrt's idiom for this is `$(CONFIG_BIG_ENDIAN)`, which tracks `ARCH:=mips` vs `mipsel` directly (see [`include/image.mk:61`](https://github.com/openwrt/openwrt/blob/25b08cdf8cb50dab4a40d7822981a1f2dc5e42f3/include/image.mk#L61) and [`include/image-commands.mk:514`](https://github.com/openwrt/openwrt/blob/25b08cdf8cb50dab4a40d7822981a1f2dc5e42f3/include/image-commands.mk#L514)):

```suggestion
	printf '$(if $(CONFIG_BIG_ENDIAN),\010\000\200\000,\000\200\000\010)' > $@.new
```

(The byte values themselves are correct — `08 00 80 00` BE / `00 80 00 08` LE both decode to `j 0x80020000`.)

---

target/linux/econet/dts/en7528_jiofiber.dtsi
```diff
@@ -0,0 +1,234 @@
...
+	reg_usb2_vbus: regulator-usb2-vbus {
+		compatible = "regulator-fixed";
+		regulator-name = "usb2-vbus";
+		regulator-min-microvolt = <5000000>;
+		regulator-max-microvolt = <5000000>;
+		gpios = <&gpio0 11 GPIO_ACTIVE_HIGH>;
+		enable-active-high;
+		regulator-boot-on;
+		regulator-always-on;
+	};
```
openwrt-ai, line 39: nit: `reg_usb2_vbus` has no consumer — the SoC has a single xHCI node and it takes `reg_usb1_vbus` via `vbus-supply` ([`xhci-mtk.c:444`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/usb/host/xhci-mtk.c#L444) claims the `"vbus"` supply). So this node only powers the second port as a side effect of `regulator-always-on` being enabled at registration, which is why it needs the `regulator-boot-on`/`regulator-always-on` pair that `reg_usb1_vbus` does not. That asymmetry is not obvious to the next reader; a one-line comment ("second USB port VBUS, no DT consumer — enabled unconditionally") would help.

---

target/linux/econet/image/en7528.mk
```diff
@@ -29,3 +29,33 @@ define Device/dasan_h660gm-a-generic
...
+  DEVICE_VENDOR := JioFiber
+  DEVICE_MODEL := JCOW407
+  DEVICE_ALT0_VENDOR := JioFiber
+  DEVICE_ALT0_MODEL := JCOW407-IN
+  DEVICE_DTS := en7528_jiofiber_jcow407
+  FACTORY_SIZE := 40m
+  TRX_LOADADDR := 0x80002000
+  KERNEL := kernel-bin | append-dtb | tclinux-free-bootbase-jump | lzma | \
+    kernel-trx
+  DEVICE_PACKAGES := kmod-usb3 kmod-mt7603 kmod-mt7615e kmod-mt7663-firmware-ap
```
openwrt-ai: nit: this block is identical to `Device/jiofiber_jcow414` above except for `DEVICE_MODEL`, `DEVICE_ALT0_MODEL` and `DEVICE_DTS` — `$(call Device/tclinux-ubi)`, `FACTORY_SIZE`, `TRX_LOADADDR`, the four-stage `KERNEL` chain and `DEVICE_PACKAGES` are all duplicated. This file's own convention is to factor that out, cf. [`define Device/dasan_h660gm-a`](https://github.com/openwrt/openwrt/blob/25b08cdf8cb50dab4a40d7822981a1f2dc5e42f3/target/linux/econet/image/en7528.mk#L10-L18), which the two H660GM-A variants pull in with `$(Device/dasan_h660gm-a)`. Since the series already factors the DTS side into `en7528_jiofiber.dtsi`, a matching `Device/jiofiber_en7528` base introduced in the JCOW414 commit and reused here would keep the two sides symmetric and stop the pair drifting apart later.

---

openwrt-ai (commented):
Reviewed 3 new commits (`c39d069`, `9529cab`, `25b08cd`). All three commit messages match their diffs, and both device commits carry the hardware-spec / flash-instruction / MAC-layout sections in full.

Verified with no findings: the free-bootbase jump encoding (`08 00 80 00` BE / `00 80 00 08` LE both decode to `j 0x80020000`, and the 122876-byte pad is exactly `0x80020000 - 0x80002000 - 4`); the `TRX_MODEL` global-leak fix (`DEVICE_VARS` + `Device/Default` reset is the right shape, and no existing en7528 device that sets it emits a TRX, so behaviour really is unchanged); the flash layout in `en7528_jiofiber.dtsi` (all partitions contiguous 0x0–0xe000000, `tclinux_kernel` + `tclinux_rootfs` sum to `tclinux`, `FACTORY_SIZE := 40m` matches `tclinux`, and the mtd-concat gives the 43.25 MiB the commit message claims); the `en75_chboot` boot-flag address (`reservearea` 0xddc0000 + 16 × 128k = 0xdfc0000, matching the documented offset); and the LED GPIO assignments against the even/odd claims in both commit messages.

Four comments inline, two of them nits.

---

target/linux/econet/dts/en7528_jiofiber_jcow414.dts
naseef: Agreed  and fixed

target/linux/econet/image/Makefile
naseef: Good catch , fixed

target/linux/econet/dts/en7528_jiofiber.dtsi
naseef: Fair, the asymmetry isn't obvious. Comment added.

target/linux/econet/image/en7528.mk
naseef: Agreed. Added Device/jiofiber_en7528 in the JCOW414 commit and both boards now pull it in

cjdelisle:
Tested on Smartfiber, also looks okay to my eyes.

target/linux/econet/image/Makefile
```diff
@@ -31,7 +35,7 @@ define Build/tclinux-free-bootbase-jump
 		$(call ERROR_MESSAGE,WARNING: KERNEL_LOADADDR changed); \
 		rm -f $@; \
 	}
-	printf '\010\000\200\000' > $@.new
+	printf '$(if $(CONFIG_BIG_ENDIAN),\010\000\200\000,\000\200\000\010)' > $@.new
```
openwrt-ai, line 38: nit: the selector is now `$(CONFIG_BIG_ENDIAN)`, but the prose describing it was not updated — `7df634a`'s message still says "pick the jump stub encoding from TRX_ENDIAN", and the PR description still says "The free-bootbase jump stub is emitted according to TRX_ENDIAN". Both should name `CONFIG_BIG_ENDIAN` so the commit message matches what the commit does.

(The change itself is right, and the "Behaviour is unchanged for the existing devices" claim still holds: the only existing user of this macro, [`Device/huawei_hg2821t-u`](https://github.com/openwrt/openwrt/blob/91ce3b999bb2e88666becb1e70663db3d10e050c/target/linux/econet/image/en751221.mk#L36), is on the en751221 subtarget, which is [`ARCH:=mips`](https://github.com/openwrt/openwrt/blob/91ce3b999bb2e88666becb1e70663db3d10e050c/target/linux/econet/en751221/target.mk#L1) — so `CONFIG_BIG_ENDIAN` is set there and the emitted bytes are byte-for-byte what they were before.)

---

openwrt-ai (commented):
Re-reviewed the force-push (`25b08cd` → `91ce3b9`, 3 commits). All four findings from the previous round are addressed: both `compatible` lists are now single-device chains, the jump stub selects on `$(CONFIG_BIG_ENDIAN)`, `reg_usb2_vbus` carries the explanatory comment, and `Device/jiofiber_en7528` factors the shared image recipe out in the JCOW414 commit with both boards pulling it in — matching the file's own `Device/dasan_h660gm-a` idiom.

Verified on the new delta with no findings: `CONFIG_BIG_ENDIAN` is the established selector in image makefiles ([`include/image.mk:61`](https://github.com/openwrt/openwrt/blob/91ce3b999bb2e88666becb1e70663db3d10e050c/include/image.mk#L61), [`include/image-commands.mk:514`](https://github.com/openwrt/openwrt/blob/91ce3b999bb2e88666becb1e70663db3d10e050c/include/image-commands.mk#L514)); the `Device/jiofiber_en7528` split is behaviour-preserving (every variable the two devices previously set individually is present in the base, in the same order relative to `$(call Device/tclinux-ubi)`); and the JCOW414/JCOW407 commit boundary is still clean, with the shared base and dtsi landing in the JCOW414 commit and the JCOW407 commit adding only its own board file plus five one-line additions.

CI on `91ce3b9` is green.

## Commit checks
- `7df634a` "econet: make the TRX header and jump stub configurable per device" — the body says the jump stub encoding is picked "from TRX_ENDIAN", but after the fix the diff picks it from `$(CONFIG_BIG_ENDIAN)`. Stale wording only; details inline.

---

bimalbiswal123:
any user-friendly installation guide for jco4032 ?

naseef:
> any user-friendly installation guide for jco4032 ?

Please check the commit message

target/linux/econet/image/Makefile
naseef: updated commit message and PR descirption

bimalbiswal123:
> > any user-friendly installation guide for jco4032 ?
> 
> Please check the commit message

Thanks, I saw that. Was hoping for a bit descriptive. Have a jco4032, can test it.

naseef:
> > > any user-friendly installation guide for jco4032 ?
> > 
> > 
> > Please check the commit message
> 
> Thanks, I saw that. Was hoping for a bit descriptive. Have a jco4032, can test it.

Please let me know what issue you facing? You need a USB to TTL cable though

openwrt-ai (commented):
Re-reviewed the force-push (`91ce3b9` → `9c7974e`, 3 commits). The tree is byte-for-byte unchanged (`git diff 91ce3b9..9c7974e` is empty); this push only rewrites the commit messages and PR description to address the previous round's commit-message finding.

That finding is now resolved: `0f9044b`'s message reads "pick the jump stub encoding from CONFIG_BIG_ENDIAN", matching the `$(if $(CONFIG_BIG_ENDIAN),...)` selector in the diff, and the PR description now says the same. No `TRX_ENDIAN` reference remains in the jump-stub prose (the surviving `--endian $(TRX_ENDIAN)` in the diff is the separate TRX-header field, correctly untouched).

No new issues found; all findings from prior rounds remain addressed.

---

naseef:
@jonasjelonek Please review this one

jonasjelonek (approved):
Looks good to me

jonasjelonek:

bimalbiswal123:
> > > > any user-friendly installation guide for jco4032 ?
> > > 
> > > 
> > > Please check the commit message
> > 
> > 
> > Thanks, I saw that. Was hoping for a bit descriptive. Have a jco4032, can test it.
> 
> Please let me know what issue you facing? You need a USB to TTL cable though

can't access u-boot, its locked

naseef:
> can't access u-boot, its locked 

These devices don't have uboot. Which model you have? Can you share bootlogs?

bimalbiswal123:
> > can't access u-boot, its locked
> 
> These devices don't have uboot. Which model you have? Can you share bootlogs?

Never mind, got it working. Thanks for your work :)

bimalbiswal123:
lan getting 100mbps - while building selected JCOW414. do I need to add any extra drivers?

naseef:
> lan getting 100mbps - while building selected JCOW414. do I need to add any extra drivers?

Linkspeed or speedtest?

bimalbiswal123:
linkspeed

naseef:
> linkspeed

check cable and also check with other link partners ( different PC, router etc)

Tapanhaz:
CI for JCOW414: https://github.com/Tapanhaz/jcow414-openwrt-build

PR #24799, opened 2026-08-19
libunwind: enable on musl/powerpc using libucontext

BKPepe:
On powerpc and powerpc64, libunwind implements `unw_tdep_getcontext()` using
libc's `getcontext()`:

```
#define unw_tdep_getcontext(uc)  (getcontext (uc), 0)
```

Unlike aarch64, arm, mips, riscv or x86_64, the powerpc ports do not ship
their own `getcontext.S`. musl declares `getcontext()` in `ucontext.h` but
does not implement it, so `libunwind.so` contains an undefined `getcontext`
symbol and linking a consumer against it fails:

```
ld: libunwind.so: undefined reference to `getcontext'
```

This restriction was originally added to work around that issue. Depend on
libucontext instead, which provides the SysV ucontext API for musl and
supports both ppc and ppc64.

The previous restriction only covered 32-bit powerpc, while powerpc64
resolves `unw_tdep_getcontext()` through libc in the same way, so the
libucontext dependency now covers both.

This also unblocks packages that already depend on libunwind
unconditionally, such as libjemalloc.

Verified on mpc85xx/p2020 (Turris 1.x, e500v2, musl):

* libucontext and libunwind both build, and `libunwind.so.8` gains
  `DT_NEEDED libucontext.so.1`
* linking a consumer against the packaged libunwind succeeds; without this
  change it fails with the undefined reference above
* `unw_backtrace()` and unwinding from a signal handler both return correct
  call chains on the device

powerpc64 was not build-tested; that part is based on `libunwind-ppc64.h`
resolving `unw_tdep_getcontext()` through libc in the same way.

cc: @blocktrron
Suggested-by: @neheb in https://github.com/openwrt/openwrt/commit/fca63d04b5075723d001eff788973edd88bcb636#commitcomment-167037367

### DIFF
--- a/package/libs/libunwind/Makefile
+++ b/package/libs/libunwind/Makefile
@@ -10,7 +10,7 @@ include $(TOPDIR)/rules.mk
 
 PKG_NAME:=libunwind
 PKG_VERSION:=1.8.3
-PKG_RELEASE:=3
+PKG_RELEASE:=4
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
 PKG_SOURCE_URL:=https://github.com/libunwind/libunwind/releases/download/v$(PKG_VERSION)/
@@ -32,7 +32,7 @@ define Package/libunwind
   CATEGORY:=Libraries
   TITLE:=The libunwind project
   URL:=https://www.nongnu.org/libunwind/
-  DEPENDS:=@!(USE_MUSL&&(powerpc)) +zlib
+  DEPENDS:=+zlib +(USE_MUSL&&powerpc||USE_MUSL&&powerpc64):libucontext
   ABI_VERSION:=8
 endef
 
@@ -46,6 +46,10 @@ CONFIGURE_ARGS += \
 	--disable-minidebuginfo
 
 TARGET_LDFLAGS += $(if $(CONFIG_USE_MUSL),-lssp_nonshared)
+# On powerpc/powerpc64 libunwind maps unw_getcontext() onto the libc
+# getcontext(), which musl does not implement.  Other architectures ship
+# their own getcontext.S, so they are unaffected.
+TARGET_LDFLAGS += $(if $(CONFIG_USE_MUSL),$(if $(filter powerpc powerpc64,$(ARCH)),-lucontext))
 
 define Package/libunwind/install
 	$(INSTALL_DIR) $(1)/usr/lib


neheb (approved):

package/libs/libunwind/Makefile
```diff
@@ -32,7 +32,7 @@ define Package/libunwind
   CATEGORY:=Libraries
   TITLE:=The libunwind project
   URL:=https://www.nongnu.org/libunwind/
-  DEPENDS:=@!(USE_MUSL&&(powerpc)) +zlib
+  DEPENDS:=+zlib +(USE_MUSL&&powerpc||USE_MUSL&&powerpc64):libucontext
```
openwrt-ai, line 35: `libucontext` is not in openwrt.git — it only exists in the packages feed (`libs/libucontext`). Without that feed installed the metadata scan emits [`WARNING: Makefile '%s' has a dependency on '%s', which does not exist` (package-metadata.pl:465)](https://github.com/openwrt/openwrt/blob/03b728b905d698931085e00d5925675caa50adb3/scripts/package-metadata.pl#L465), `PACKAGE_libucontext` is never defined, and libunwind becomes unbuildable on musl/powerpc from a plain core checkout — while `-lucontext` is still appended unconditionally below, so the link would fail rather than degrade gracefully.

There is precedent for core→feed conditional deps ([busybox:51](https://github.com/openwrt/openwrt/blob/03b728b905d698931085e00d5925675caa50adb3/package/utils/busybox/Makefile#L51) selects the feed-only `libpam`/`libtirpc`), but those are gated behind opt-in `BUSYBOX_CONFIG_*`/`USE_GLIBC` symbols a core-only user simply leaves off. Here the dependency is unavoidable for *every* musl powerpc/powerpc64 build of a core library. Is depending on the packages feed intentional here, or should `libucontext` be moved into openwrt.git alongside libunwind?

---

package/libs/libunwind/Makefile
```diff
@@ -46,6 +46,10 @@ CONFIGURE_ARGS += \
 	--disable-minidebuginfo
 
 TARGET_LDFLAGS += $(if $(CONFIG_USE_MUSL),-lssp_nonshared)
+# On powerpc/powerpc64 libunwind maps unw_getcontext() onto the libc
+# getcontext(), which musl does not implement.  Other architectures ship
+# their own getcontext.S, so they are unaffected.
+TARGET_LDFLAGS += $(if $(CONFIG_USE_MUSL),$(if $(filter powerpc powerpc64,$(ARCH)),-lucontext))
```
openwrt-ai, line 52: This fixes libunwind's own link, but probably not its consumers'. As quoted in the PR description, `unw_tdep_getcontext()` (which `libunwind.h` maps `unw_getcontext()` onto) is a *macro* — `(getcontext (uc), 0)` — so it expands inside the consumer's translation unit, and that object file gets its own undefined `getcontext` reference. Adding `-lucontext` here only gives `libunwind.so` a `DT_NEEDED libucontext.so.1`; with binutils' default `--no-copy-dt-needed-entries` a transitively-needed library does not satisfy the consumer's own undefined symbols.

That matches the verification described in the PR body: `unw_backtrace()` is a real out-of-line function in `libunwind.so`, so it links fine; a consumer calling `unw_getcontext()` would still hit `undefined reference to 'getcontext'`.

Should the installed `libunwind.pc` (copied by `Build/InstallDev` at [line 63](https://github.com/openwrt/openwrt/blob/03b728b905d698931085e00d5925675caa50adb3/package/libs/libunwind/Makefile#L63)) also gain `-lucontext` in `Libs`/`Libs.private` on musl/powerpc, so packages consuming libunwind via pkg-config pick it up?

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the diff; two questions inline about the `libucontext` dependency source and about consumers of the `unw_getcontext()` macro.

FWIW the `+(USE_MUSL&&powerpc||USE_MUSL&&powerpc64):libucontext` spelling is correct for OpenWrt's condition parser — `gen_condition()` strips parentheses and splits on `||` before `&&`, so the more natural-looking `USE_MUSL&&(powerpc||powerpc64)` would have been mis-parsed. Kconfig's `&&`-over-`||` precedence makes the generated `select ... if` correct too.

---

BKPepe:
I am putting this on hold. 

- I managed to fix it upstream, so we're now just waiting for the new libunwind 1.8.4 release. :)

PR #24801, opened 2026-08-19
build: add build targets to clean tmp / feeds

mcprat:
restoring a modern version of old clean targets that were removed in the past.

now everything in the `.SLIENT:` list is actually defined as a rule.

### DIFF
--- a/include/toplevel.mk
+++ b/include/toplevel.mk
@@ -271,12 +271,15 @@ package/symlinks-clean:
 help:
 	cat README.md
 
-distclean:
-	rm -rf bin build_dir .ccache .config* dl feeds key-build* logs package/feeds target/linux/feeds staging_dir tmp
+feedclean: tmpclean FORCE
+	rm -rf feeds package/feeds target/linux/feeds
+
+distclean: feedclean FORCE
+	rm -rf bin build_dir .ccache .config* dl key-build* logs staging_dir tmp
 	@$(_SINGLE)$(SUBMAKE) -C scripts/config clean
 
 ifeq ($(findstring v,$(DEBUG)),)
-  .SILENT: symlinkclean clean dirclean distclean config-clean download help tmpclean .config scripts/config/mconf scripts/config/conf menuconfig $(STAGING_DIR_HOST)/.prereq-build tmp/.prereq-package prepare-tmpinfo
+  .SILENT: feedclean clean dirclean distclean config-clean download help tmpclean .config scripts/config/mconf scripts/config/conf menuconfig $(STAGING_DIR_HOST)/.prereq-build tmp/.prereq-package prepare-tmpinfo
 endif
 .PHONY: help FORCE
 .NOTPARALLEL:
--- a/include/toplevel.mk
+++ b/include/toplevel.mk
@@ -91,6 +91,17 @@ prepare-tmpinfo: FORCE
 	./scripts/package-metadata.pl usergroup tmp/.packageinfo > tmp/.packageusergroup || { rm -f tmp/.packageusergroup; false; }
 	touch $(TOPDIR)/tmp/.build
 
+tmpclean: FORCE
+	rm -f $(PREP_MK)
+	rm -f $(STAGING_DIR_HOST)/.prereq-build
+	rm -rf $(TOPDIR)/tmp/info
+	rm -f $(TOPDIR)/tmp/.config-package.in
+	rm -f $(TOPDIR)/tmp/.config-target.in
+	rm -f $(TOPDIR)/tmp/.config-feeds.in
+	rm -f $(TOPDIR)/tmp/.package*
+	rm -f $(TOPDIR)/tmp/.target*
+	rm -f $(TOPDIR)/tmp/.build
+
 .config: ./scripts/config/conf $(if $(CONFIG_HAVE_DOT_CONFIG),,prepare-tmpinfo)
 	@+if [ \! -e .config ] || ! grep CONFIG_HAVE_DOT_CONFIG .config >/dev/null; then \
 		[ -e $(HOME)/.openwrt/defconfig ] && cp $(HOME)/.openwrt/defconfig .config; \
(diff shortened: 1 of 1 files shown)


include/toplevel.mk
```diff
@@ -91,6 +91,17 @@ prepare-tmpinfo: FORCE
 	./scripts/package-metadata.pl usergroup tmp/.packageinfo > tmp/.packageusergroup || { rm -f tmp/.packageusergroup; false; }
 	touch $(TOPDIR)/tmp/.build
 
+tmpclean: FORCE
+	rm -f $(PREP_MK)
+	rm -f $(STAGING_DIR_HOST)/.prereq-build
```
openwrt-ai, line 96: `PREP_MK` is not a file path — it is [`PREP_MK= OPENWRT_BUILD= QUIET=0` at toplevel.mk:5](https://github.com/openwrt/openwrt/blob/c24270d6bf856bb431a2eb8ade084624199e15af/include/toplevel.mk#L5), a pair of make variable assignments that only ever get passed on sub-make command lines ([lines 79](https://github.com/openwrt/openwrt/blob/c24270d6bf856bb431a2eb8ade084624199e15af/include/toplevel.mk#L79), [97](https://github.com/openwrt/openwrt/blob/c24270d6bf856bb431a2eb8ade084624199e15af/include/toplevel.mk#L97), [224](https://github.com/openwrt/openwrt/blob/c24270d6bf856bb431a2eb8ade084624199e15af/include/toplevel.mk#L224), [231](https://github.com/openwrt/openwrt/blob/c24270d6bf856bb431a2eb8ade084624199e15af/include/toplevel.mk#L231)).

So this recipe line expands to `rm -f OPENWRT_BUILD= QUIET=0`, i.e. it tries to remove two files literally named `OPENWRT_BUILD=` and `QUIET=0` from the top directory. It removes nothing the target means to remove (and `rm -f` hides that it did nothing).

```suggestion
	rm -f $(STAGING_DIR_HOST)/.prereq-build
```

---

include/toplevel.mk
```diff
@@ -114,7 +125,7 @@ config: scripts/config/conf prepare-tmpinfo FORCE
 	[ -L .config ] && export KCONFIG_OVERWRITECONFIG=1; \
 		$< $(KCONF_FLAGS) Config.in
 
-config-clean: FORCE
+config-clean: tmpclean FORCE
```
openwrt-ai, line 128: This is a behavior change to an existing target that neither the commit message ("Add a build target tmpclean") nor the PR body mentions. `config-clean` used to be scoped to the kconfig host binaries (`make -C scripts/config clean`); with this prerequisite it now also wipes `tmp/.packageinfo`, `tmp/.targetinfo`, `tmp/info/` and the generated `Config.in` fragments, so the next `make menuconfig` pays a full metadata rescan.

The two are otherwise unrelated — `tmp/.config-*.in` are produced by `scripts/package-metadata.pl` / `scripts/target-metadata.pl` in [`prepare-tmpinfo`](https://github.com/openwrt/openwrt/blob/c24270d6bf856bb431a2eb8ade084624199e15af/include/toplevel.mk#L78-L92), not by `scripts/config`. Is the coupling intentional? If so, it deserves its own commit (or at least a line in the commit message) explaining why; if not, `tmpclean` can stand alone and users who want both can run `make config-clean tmpclean`.

---

include/toplevel.mk
```diff
@@ -260,12 +271,15 @@ package/symlinks-clean:
...
+feedclean: tmpclean FORCE
+	rm -rf feeds package/feeds target/linux/feeds
+
+distclean: feedclean FORCE
+	rm -rf bin build_dir .ccache .config* dl key-build* logs staging_dir tmp
 	@$(_SINGLE)$(SUBMAKE) -C scripts/config clean
 
 ifeq ($(findstring v,$(DEBUG)),)
-  .SILENT: symlinkclean clean dirclean distclean config-clean download help tmpinfo-clean .config scripts/config/mconf scripts/config/conf menuconfig $(STAGING_DIR_HOST)/.prereq-build tmp/.prereq-package prepare-tmpinfo
+  .SILENT: feedclean clean dirclean distclean config-clean download help tmpclean .config scripts/config/mconf scripts/config/conf menuconfig $(STAGING_DIR_HOST)/.prereq-build tmp/.prereq-package prepare-tmpinfo
```
openwrt-ai, line 282: nit: the PR body says "now everything in the `.SILENT:` list is actually defined as a rule", but `tmp/.prereq-package` is still a dead entry — a tree-wide grep for `prereq-package` finds only this line, so nothing ever builds that target (the surviving stamp is `$(STAGING_DIR_HOST)/.prereq-build`, already listed). Since this PR is pruning the other stale entries (`symlinkclean`, `tmpinfo-clean`), it may as well go too, otherwise the claim in the description doesn't hold.

```suggestion
  .SILENT: feedclean clean dirclean distclean config-clean download help tmpclean .config scripts/config/mconf scripts/config/conf menuconfig $(STAGING_DIR_HOST)/.prereq-build prepare-tmpinfo
```

---

openwrt-ai (commented):
## Commit checks

- a9e8827 "build: add build target to clean tmp info" — the message describes only the addition of a new `tmpclean` target, but the diff also rewires the existing `config-clean` target to depend on it, which changes what `make config-clean` does. Either split that hunk out or document it in the commit body.

Reviewed 2 new commits; 3 inline comments.

---

include/toplevel.mk
```diff
@@ -260,12 +271,15 @@ package/symlinks-clean:
 help:
 	cat README.md
 
-distclean:
-	rm -rf bin build_dir .ccache .config* dl feeds key-build* logs package/feeds target/linux/feeds staging_dir tmp
+feedclean: tmpclean FORCE
+	rm -rf feeds package/feeds target/linux/feeds
+
+distclean: feedclean FORCE
+	rm -rf bin build_dir .ccache .config* dl key-build* logs staging_dir tmp
```
peterwillcn: rm -f scripts/config/conf

PR #24803, opened 2026-08-19
kernel: backport adt7470 PWM/thermal support from v7.3

luizluca:
Backport a series of upstream patches for the adt7470 hwmon driver, accepted for kernel v7.3. It exposes the fan control lines via the generic PWM framework and thermal zone sensors.

Instead of relying on userland scripts, the kernel can now control the fan speed directly through standard cooling maps and PWM fans configured in the DTS. Currently in the tree, only edgecore_ecs4100-12ph seems to use this chip.

### DIFF
--- a/target/linux/generic/backport-6.18/851-03-v7.3-hwmon-adt7470-Expose-fan-control-via-PWM-framework.patch
+++ b/target/linux/generic/backport-6.18/851-03-v7.3-hwmon-adt7470-Expose-fan-control-via-PWM-framework.patch
@@ -1,0 +1,299 @@
+From 8d1e8f8deefe7a5b00c5b3535b3eb63f23133d51 Mon Sep 17 00:00:00 2001
+From: Luiz Angelo Daros de Luca <luizluca@gmail.com>
+Date: Wed, 5 Aug 2026 10:15:19 -0300
+Subject: [PATCH 3/4] hwmon: (adt7470) Expose fan control via PWM framework
+
+The ADT7470 features four PWM outputs designed to control fans.
+Previously, these were only accessible through the legacy hwmon
+sysfs interface.
+
+Expose the ADT7470 fan control lines through the generic PWM framework.
+This allows generic PWM consumers described in Device Tree, such as
+"pwm-fan", to use the device through the "#pwm-cells" property.  This is
+particularly necessary for boards where the ADT7470 external temperature
+sensors are not populated and the chip is used strictly as a fan
+controller (e.g., Edgecore ECS-2100-52T switches). In such setups,
+delegating control to the generic PWM subsystem allows the kernel's
+thermal framework to manage the fans based on unrelated temperature
+zones, such as internal SoC sensors.
+
+When a PWM consumer applies a new PWM state, the driver automatically
+switches the corresponding PWM channel to manual mode so that the
+requested duty cycle takes effect. The duty cycle specified by the PWM
+framework is internally converted to the 0-255 scale expected by the
+hardware registers.
+
+To prevent conflicting access, the hwmon PWM sysfs attributes are hidden
+when the device is registered with PWM framework support (i.e. when
+"#pwm-cells" is present in DT and CONFIG_PWM is reachable).
+
+Signed-off-by: Luiz Angelo Daros de Luca <luizluca@gmail.com>
+Link: https://lore.kernel.org/r/20260805-adt7470_thermalzone-v6-3-5605a5f95467@gmail.com
+Signed-off-by: Guenter Roeck <linux@roeck-us.net>
+---
+ drivers/hwmon/Kconfig   |   1 +
+ drivers/hwmon/adt7470.c | 194 +++++++++++++++++++++++++++++++++++++++-
+ 2 files changed, 194 insertions(+), 1 deletion(-)
+
+--- a/drivers/hwmon/Kconfig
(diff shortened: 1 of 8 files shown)


luizluca:
The target tags added by bot might be misleading. It is just refreshing patches.

target/linux/generic/backport-6.18/851-03-v7.3-hwmon-adt7470-Expose-fan-control-via-PWM-framework.patch
```diff
@@ -0,0 +1,299 @@
...
+ 		case hwmon_pwm_input:
+ 		case hwmon_pwm_enable:
+@@ -1328,6 +1497,29 @@ static int adt7470_probe(struct i2c_clie
+ 
+ 	data->pwm_freq = (u32)freq_val;
+ 
++	data->use_pwm_framework = false;
++
++	if (device_property_present(dev, "#pwm-cells")) {
++		if (IS_REACHABLE(CONFIG_PWM)) {
```
openwrt-ai, line 277: This whole branch is dead in the tree today, so the PWM half of the series is a no-op for the only user: `CONFIG_PWM` is off tree-wide ([`target/linux/generic/config-6.18:5181`](https://github.com/openwrt/openwrt/blob/cc27ee2ca9330dbb421798dc8d677fa51097f795/target/linux/generic/config-6.18#L5181) — `# CONFIG_PWM is not set`) and no realtek subtarget config overrides it, so `IS_REACHABLE(CONFIG_PWM)` is 0 and only the `dev_warn()` path can ever run. On top of that, [`rtl8393_edgecore_ecs4100-12ph.dts`](https://github.com/openwrt/openwrt/blob/cc27ee2ca9330dbb421798dc8d677fa51097f795/target/linux/realtek/dts/rtl8393_edgecore_ecs4100-12ph.dts) has no `adi,adt7470` node at all (the chip comes up via the i2c address scan), so there is no `#pwm-cells` for `device_property_present()` to find either — `kmod-hwmon-adt7470` is just listed in [`rtl839x.mk`](https://github.com/openwrt/openwrt/blob/cc27ee2ca9330dbb421798dc8d677fa51097f795/target/linux/realtek/image/rtl839x.mk#L20).

Patch 04 (thermal zone) does take effect since `CONFIG_THERMAL_OF=y` on rtl839x, so the series is not entirely inert.

Is a follow-up planned that enables `CONFIG_PWM` (+ `PWM_FAN` / cooling maps) and adds the DT node with `#pwm-cells`/`#thermal-sensor-cells`? Worth mentioning in the commit message if so, otherwise this reads like it should land together with the consumer side.

(No build risk here — `depends on PWM || PWM=n` stays satisfiable with `PWM=n`, and `include/linux/pwm.h` in 6.18.44 provides the `!CONFIG_PWM` stubs for `devm_pwmchip_alloc()`/`devm_pwmchip_add()`, plus the waveform ops the series uses.)

---

openwrt-ai (commented):
## Commit checks

- cc27ee2c "kernel: backport adt7470 fixes and PWM/thermal support from v7.3" — the subject promises "fixes", but none of the four backported patches is a fix. `851-01` adds the dt-binding doc, `851-02` is a cleanup (magic `255` → `ADT7470_PWM_MAX`), and `851-03`/`851-04` add the PWM-framework and thermal-zone features. Something like "kernel: backport adt7470 PWM and thermal support from v7.3" would match the diff.

Verification notes (all clean, no action needed): the four new patches apply to v6.18.44 `drivers/hwmon/{adt7470.c,Kconfig}` with zero fuzz and zero offset, and each patch body's insertion/deletion counts match the `git format-patch` diffstat it carries verbatim (58/0, 4/3, 194/1, 2/0). The four refreshed target patches are the complete set that touches `drivers/hwmon/Kconfig` on 6.18 (bcm27xx 0310, microchipsw 0010, mvebu 903, realtek 813; the bcm27xx `patches-6.12` copy is correctly untouched), and all four re-apply at offset 0 against the Kconfig after `851-03`.

---

target/linux/generic/backport-6.18/851-03-v7.3-hwmon-adt7470-Expose-fan-control-via-PWM-framework.patch
jonasjelonek: Fine to keep, will be used by new EdgeCore device.

jonasjelonek:
Please adjust the commit title so it's not saying "fixes", since those really just to add support for something, not really fix anything. Then we're good to go.

luizluca:
> Please adjust the commit title so it's not saying "fixes", since those really just to add support for something, not really fix anything. Then we're good to go.

Thanks @jonasjelonek. It used to have some fixes but they were dropped when they reached kernel stable used by OpenWrt. I updated the commit message but I missed the subject. Fixed.

openwrt-ai (commented):

`git diff cc27ee2c..c39956ea` is empty — the tree is byte-identical to the previously reviewed revision, only the commit message changed. The subject is now "kernel: backport adt7470 PWM/thermal support from v7.3", which no longer claims "fixes" and matches what the four backported patches actually do (binding doc, `ADT7470_PWM_MAX` cleanup, PWM framework, thermal zone sensors). My earlier commit-check finding is resolved.

---

jonasjelonek:

Issue #24738, opened 2026-08-15
KSMBD bug in Kernels 6.18.x,  6.12.x and 6.6.x (the fix is ready to backport) 


GithPK:
### Describe the bug

Kernel module ksmbd crashes with list corruption in ksmbd_file_cache and then
deadlocks inside ksmbd_smb_check_shared_mode during smb2_open(), causing RCU
stall and complete hang of ksmbd worker thread.

This is a reproducible kernel-level bug.
- ksmbd-server  3.5.6-r1
- Filesystem: ext4
- Load: normal SMB file access (Windows client)

### OpenWrt version

r35781-71227664f2

### OpenWrt release

SNAPSHOT

### OpenWrt target/subtarget

 qualcommax/ipq807x

### Device

 Arcadyan AW1000

### Image kind

Official downloaded image

### Actual behaviour

[   48.641864] ksmbd: running
[  280.000009]  slab ksmbd_file_cache start ffffff80148c1080 pointer offset 192 size 376
[  280.000077] ------------[ cut here ]------------
[  280.006941] list_add corruption. next->prev should be prev (ffffff801462e358), but was ffffff80148c1140. (next=ffffff80148c1140).
[  280.011580] WARNING: CPU: 1 PID: 118 at lib/list_debug.c:29 __list_add_valid_or_report+0x94/0x108
[...]

PR #24657, opened 2026-08-10
kernel: 6.12-6.18: ksmbd: backport multi-iov signing fix

pesa1234:
Backport the ksmbd fix for signing responses whose SMB header and payload are stored in separate iovecs. The existing code only handles SMB2 READ specially, causing QUERY_INFO and CHANGE_NOTIFY responses to omit the header from the signature and be rejected by Windows clients.

The change is queued in the KSMBD maintainer code-review tree. Refresh the patch against 6.12 and 6.18

PR #24804, opened 2026-08-19
kernel: 6.12-6.18: ksmbd: backport session teardown fix

pesa1234:
Backport upstream "ksmbd: harden file lifetime during session teardown" for Linux 6.12 and 6.18.

The fix resolves multiple races in the ksmbd session and tree connection teardown paths. These races can cause file-list corruption, incorrect reference counting, use-after-free, leaked file references and sleeping while holding the file-table spinlock.

The commit first appeared in Linux v7.1, so add it to the generic backport directories and refresh it against Linux 6.12.103 and 6.18.44.

Upstream:
https://git.kernel.org/torvalds/c/a42896bebfcc

Validation:
- refreshed both patches with target/linux/refresh
- no fuzz or offset

please have a look also at this pr: https://github.com/openwrt/openwrt/pull/24657 not related but for me it is important.

Thanks

### DIFF
--- a/target/linux/generic/backport-6.12/502-v7.3-smb-server-fix-signing-multi-iov-response.patch
+++ b/target/linux/generic/backport-6.12/502-v7.3-smb-server-fix-signing-multi-iov-response.patch
@@ -1,4 +1,4 @@
-From a99deddaa603c56e824e0916d433e7f7d3015f53 Mon Sep 17 00:00:00 2001
+From a4c06ff2eb66fc704582a443b47798a52c4303e2 Mon Sep 17 00:00:00 2001
 From: ChenXiaoSong <chenxiaosong@kylinos.cn>
 Date: Thu, 23 Jul 2026 04:11:29 +0000
 Subject: [PATCH] smb/server: fix signing when a response uses more than one
@@ -25,7 +25,7 @@ Signed-off-by: Steve French <stfrench@microsoft.com>
 
 --- a/fs/smb/server/smb2pdu.c
 +++ b/fs/smb/server/smb2pdu.c
-@@ -9072,6 +9072,39 @@ int smb2_check_sign_req(struct ksmbd_wor
+@@ -9016,6 +9016,39 @@ int smb2_check_sign_req(struct ksmbd_wor
  }
  
  /**
@@ -41,7 +41,7 @@ Signed-off-by: Steve French <stfrench@microsoft.com>
 + * Return: The first iovec to sign.
 + */
 +static struct kvec *smb2_get_sign_rsp_iov(struct ksmbd_work *work,
-+					  struct smb2_hdr *hdr, int *n_vec)
++					   struct smb2_hdr *hdr, int *n_vec)
 +{
 +	int i;
 +
@@ -65,7 +65,7 @@ Signed-off-by: Steve French <stfrench@microsoft.com>
   * smb2_set_sign_rsp() - handler for rsp packet sign processing
   * @work:   smb work containing notify command buffer
   *
-@@ -9081,18 +9114,13 @@ void smb2_set_sign_rsp(struct ksmbd_work
+@@ -9025,18 +9058,13 @@ void smb2_set_sign_rsp(struct ksmbd_work
  	struct smb2_hdr *hdr;
  	char signature[SMB2_HMACSHA256_SIZE];
  	struct kvec *iov;
@@ -84,9 +84,9 @@ Signed-off-by: Steve French <stfrench@microsoft.com>
 -	}
 +	iov = smb2_get_sign_rsp_iov(work, hdr, &n_vec);
  
- 	ksmbd_sign_smb2_pdu(work->conn, work->sess->sess_key, iov, n_vec,
- 			    signature);
-@@ -9171,7 +9199,7 @@ void smb3_set_sign_rsp(struct ksmbd_work
+ 	if (!ksmbd_sign_smb2_pdu(work->conn, work->sess->sess_key, iov, n_vec,
+ 				 signature))
+@@ -9115,7 +9143,7 @@ void smb3_set_sign_rsp(struct ksmbd_work
  	struct channel *chann;
  	char signature[SMB2_CMACAES_SIZE];
  	struct kvec *iov;
@@ -95,7 +95,7 @@ Signed-off-by: Steve French <stfrench@microsoft.com>
  	char *signing_key;
  
  	hdr = ksmbd_resp_buf_curr(work);
-@@ -9193,12 +9221,7 @@ void smb3_set_sign_rsp(struct ksmbd_work
+@@ -9137,12 +9165,7 @@ void smb3_set_sign_rsp(struct ksmbd_work
  	hdr->Flags |= SMB2_FLAGS_SIGNED;
  	memset(hdr->Signature, 0, SMB2_SIGNATURE_SIZE);
  
--- a/target/linux/generic/backport-6.12/501-v7.1-ksmbd-harden-file-lifetime-during-session-teardown.patch
+++ b/target/linux/generic/backport-6.12/501-v7.1-ksmbd-harden-file-lifetime-during-session-teardown.patch
@@ -1,0 +1,386 @@
+From a42896bebfcc287ed1e61d820a888e33b1eb80ce Mon Sep 17 00:00:00 2001
+From: DaeMyung Kang <charsyam@gmail.com>
+Date: Tue, 28 Apr 2026 23:08:55 +0900
+Subject: [PATCH] ksmbd: harden file lifetime during session teardown
+
+__close_file_table_ids() is the per-session teardown that closes every
+fp belonging to a session (or to one tree connect on that session) by
+walking the session's volatile-id idr.  The current loop has three
+related problems on busy or racing workloads:
+
+  * Sleeping under ft->lock.  The session-teardown skip callback,
+    session_fd_check(), already sleeps in ksmbd_vfs_copy_durable_owner()
+    -> kstrdup(GFP_KERNEL) and down_write(&fp->f_ci->m_lock) (a
+    rw_semaphore).  Running the callback inside write_lock(&ft->lock)
+    trips CONFIG_DEBUG_ATOMIC_SLEEP / CONFIG_PROVE_LOCKING on a
+    durable-fd workload.
+
+  * Refcount accounting blind to f_state.  The unconditional
+    atomic_dec_and_test(&fp->refcount) does not distinguish
+    FP_INITED (idr-owned reference still intact) from FP_CLOSED (an
+    earlier ksmbd_close_fd() already consumed the idr-owned reference
+    while leaving fp in the idr because a holder kept refcount
+    non-zero).  When the latter races with teardown the same path
+    over-decrements into a holder reference and ksmbd_fd_put() later
+    UAFs that holder.
+
+  * FP_NEW window.  Between __open_id() publishing fp into the
+    session idr and ksmbd_update_fstate(..., FP_INITED) committing the
+    transition at the end of smb2_open(), an fp is in FP_NEW and an
+    intervening teardown that takes a transient reference and
+    unpublishes the volatile id leaves the original idr-owned
+    reference orphaned -- the opener is unaware that fp has been
+    unpublished, returns success to the client, and the fp leaks at
+    refcount = 1.
+
+Refactor __close_file_table_ids() to take a transient reference on fp
+and unpublish fp from the session idr *under ft->lock* before calling
+skip() outside the lock.  A transient ref protects lifetime but not
+concurrent field mutation, so the idr_remove() is what keeps
+__ksmbd_lookup_fd() through this session's idr from granting a new
+ksmbd_fp_get() reference to an fp whose fp->conn / fp->tcon /
+fp->volatile_id / op->conn / lock_list links are about to be rewritten
+by session_fd_check().  Durable reconnect is unaffected because it
+reaches fp through the global durable table (ksmbd_lookup_durable_fd
+-> global_ft).
+
+Decide n_to_drop together with any FP_INITED -> FP_CLOSED transition
+under ft->lock so teardown and ksmbd_close_fd() never both consume the
+idr-owned reference.  See ksmbd_mark_fp_closed() for the per-state
+accounting.  For the FP_NEW path to be safe, the opener has to learn
+that fp was unpublished: ksmbd_update_fstate() now returns -ENOENT
+when an FP_NEW -> FP_INITED transition finds f_state already advanced
+or the volatile id cleared (both committed by teardown under
+ft->lock); smb2_open() propagates that as STATUS_OBJECT_NAME_INVALID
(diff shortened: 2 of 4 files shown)


sppmasterspp:
@pesa1234 Thanks a lot for the fast reaction.
I have updated the package and I've added this and the other fix and now I'm testing it with heavy loads.
No issues so far.

openwrt-ai (commented):

Verification performed against the upstream tree and the two kernel versions this targets:

- `a42896bebfcc287ed1e61d820a888e33b1eb80ce` exists in mainline and first appears in **v7.1** (`fs/smb/server/vfs_cache.c` at `v7.0` has no `ksmbd_mark_fp_closed()`, at `v7.1` it does), so the `v7.1-` filename prefix is correct.
- Both patch files apply to `v6.12.103` and `v6.18.44` with zero fuzz and zero offset, matching the "no fuzz or offset" claim in the PR body.
- The post-apply result is byte-identical to upstream for every function the commit touches: `__ksmbd_remove_fd()`, `ksmbd_update_fstate()`, `ksmbd_mark_fp_closed()`, `__close_file_table_ids()` (all 132 lines), `ksmbd_close_tree_conn_fds()`, `ksmbd_close_session_fds()`, `ksmbd_destroy_file_table()`, the `session_fd_check()` `FP_INITED` guard, the `vfs_cache.h` prototype change, and the `err_out2:` hunk in `smb2pdu.c`. The embedded commit message is also verbatim upstream.
- No missing prerequisite: every helper the new code relies on already exists in both 6.12.103 and 6.18.44 — `ksmbd_fp_get()` already carries the `f_state != FP_INITED` guard, `__ksmbd_close_fd()` already tolerates a NULL `ft`, and `__put_fd_final()` already does its own `open_files_count` decrement.
- No missing follow-up: comparing against `v7.2`, the only later change in `__close_file_table_ids()` is `fp->durable_volatile_id = fp->volatile_id;`, and `durable_volatile_id` does not exist in 6.12/6.18 (it comes from a separate, unrelated commit), so nothing else needs to come along.
- Numbering/naming follow `target/linux/generic/PATCHES.md`: `5xx` is the filesystem range, `501` is free in both `backport-6.12` (existing `500-01..500-04`) and `backport-6.18`, and the name matches the `NNN-vX.Y-lowercase_shortname.patch` form.
- Both commit messages match their diffs exactly (one patch file each, in the directory named in the subject).
- CI on `cff6f49` reports no failed jobs.

---

jonasjelonek:
Please change the patch number from 501 to probably 502, since it's not from the same upstream series. And also do me a favor please, since I overlooked it: Add the version tag to the patches merged in #24657 the same as here. The patches aren't in Linus' tree yet but I'm pretty sure you can add v7.3

pesa1234:
> Please change the patch number from 501 to probably 502, since it's not from the same upstream series. And also do me a favor please, since I overlooked it: Add the version tag to the patches merged in #24657 the same as here. The patches aren't in Linus' tree yet but I'm pretty sure you can add v7.3

Done, thanks for spotting this. I renumbered the session teardown patches from 501 to 502 for both kernel versions and added the v7.3 tag to the multi-iov signing patches from #24657.

openwrt-ai (commented):
Reviewed the incremental changes since my previous review at `cff6f49` (one new commit, `3411efeccb7`, plus the rebased teardown patches); no new issues found.

The two follow-up requests from the review discussion are implemented correctly:

- **Renumber `501` -> `502` for the session-teardown backport.** `backport-6.12/` and `backport-6.18/` now carry `502-v7.1-ksmbd-harden-file-lifetime-during-session-teardown.patch`. The rename is content-identical (`similarity index 100%`), so the patch body already validated in my prior review is unchanged. `502` is free in both directories (`backport-6.12` has only `500-01..500-04`, `backport-6.18` has no other `5xx`), and keeping the teardown backport distinct from the `pending-*/501` multi-iov patch avoids a shared `501` across the combined series.

- **Add the `v7.3` tag to the multi-iov signing patches + refresh.** `pending-6.12/` and `pending-6.18/` patches are renamed to `501-v7.3-smb-server-fix-signing-multi-iov-response.patch`. The only content change is the hunk-header offsets shifting `+2` (e.g. `9014` -> `9016`, `9072` -> `9074`), which matches a `make target/linux/refresh` after inserting the `502` teardown backport: that patch's `err_out2:` hunk in `smb2pdu.c` (near line 3765) removes 2 lines and adds 4, a net `+2` ahead of every multi-iov hunk. The shift is uniform and consistent across all four hunks in both kernel versions.

Also checked:
- The `502` teardown patch still applies before the `pending` `501` multi-iov patch (backport dir precedes pending dir), so the refreshed offsets reflect the correct applied context.
- All three commit messages match their diffs: the two teardown commits each add one patch file for their kernel version; `3411efeccb7` performs exactly the four renames it describes.
- CI on head `3411efec` shows no failed jobs (all completed check runs are `success` or `skipped`).

---

jonasjelonek:
I probably got that wrong somehow, are the other patches not accepted yet officially? I just saw that they were placed in pending. Usually, once a patch is accepted upstream into a maintainer's tree, it'll very likely go with the same commit id into Linus' tree.

pesa1234:
> I probably got that wrong somehow, are the other patches not accepted yet officially? I just saw that they were placed in pending. Usually, once a patch is accepted upstream into a maintainer's tree, it'll very likely go with the same commit id into Linus' tree.

You were right. I was tracking a99ded... in the code-review branch and missed that the patch had already been picked into ksmbd-for-next as [a4c06ff...](https://github.com/smfrench/smb3-kernel/commit/a4c06ff2eb66fc704582a443b47798a52c4303e2). The content is unchanged; the commit ID changed when it was rebased onto the maintainer branch.
I’ve now moved both patches from pending-* to backport-*, updated the upstream commit ID, and refreshed them for the new ordering. They apply to Linux 6.12.103 and 6.18.44 without fuzz or offset. The v7.3 tag is retained.
Thanks!

Andrea

target/linux/generic/backport-6.12/502-v7.3-smb-server-fix-signing-multi-iov-response.patch
```diff
@@ -1,4 +1,4 @@
-From a99deddaa603c56e824e0916d433e7f7d3015f53 Mon Sep 17 00:00:00 2001
+From a4c06ff2eb66fc704582a443b47798a52c4303e2 Mon Sep 17 00:00:00 2001
```
openwrt-ai, line 1: nit: now that this claims to be a verbatim copy of `a4c06ff2eb66`, the body has one byte that isn't. The continuation line of the new helper is indented with 5 tabs + **2** spaces here, while both `a99deddaa603` and `a4c06ff2eb66` use 5 tabs + **3** spaces:

```
upstream:  \t\t\t\t\t   struct smb2_hdr *hdr, int *n_vec)
this file: \t\t\t\t\t  struct smb2_hdr *hdr, int *n_vec)
```

I diffed the post-apply `fs/smb/server/smb2pdu.c` against `a4c06ff2eb66:fs/smb/server/smb2pdu.c` for both 6.12.103 and 6.18.44; this is the only difference that isn't an expected 6.12/6.18-vs-7.x base adaptation (the `if (!ksmbd_sign_smb2_pdu(...))` / `if (!ksmbd_sign_smb3_pdu(...))` return-value checks and the `conn->binding` condition are pre-existing tree differences, not introduced by this patch).

The deviation is pre-existing — it came in with #24657 and is not introduced by this commit — but since this commit is specifically re-pointing the `From:` header at the maintainer-tree SHA, it seems like the right moment to make the body match. Same line in [backport-6.18/501-...patch#L44](``https://github.com/openwrt/openwrt/blob/c98f334cedf1339b76e01adc126803d8eedce9ee/target/linux/generic/backport-6.18/501-v7.3-smb-server-fix-signing-multi-iov-response.patch#L44)`` and in [backport-6.12/501-...patch#L44](``https://github.com/openwrt/openwrt/blob/c98f334cedf1339b76e01adc126803d8eedce9ee/target/linux/generic/backport-6.12/501-v7.3-smb-server-fix-signing-multi-iov-response.patch#L44).``

---

openwrt-ai (commented):
Reviewed the one new commit since my last review at `3411efec` (`c98f334` "kernel: 6.12-6.18: move ksmbd signing fix to backports"); one whitespace nit inline, nothing blocking.

Verification of the new commit:

- **Upstream SHA is real.** `a4c06ff2eb66fc704582a443b47798a52c4303e2` fetches from `smfrench/smb3-kernel` and carries exactly the subject/author/date/trailers embedded in the patch file (`ChenXiaoSong`, `Thu, 23 Jul 2026 04:11:29 +0000`, `Acked-by: Namjae Jeon`, `Signed-off-by: Steve French`). The relocated commit is content-identical to the old `a99deddaa603` in `fs/smb/server/smb2pdu.c`, which matches the PR-comment explanation that only the ID changed on rebase.
- **`pending-*` -> `backport-*` move is consistent with the tree.** Placing a not-yet-in-Linus patch under `backport-*` with a `v7.3` tag matches existing in-tree practice (`backport-6.18/436-v7.3-*`, `437-v7.3-*`, `755-v7.3-*`), and the filename follows the `NNN-vX.Y-lowercase_shortname.patch` form from `target/linux/generic/PATCHES.md`. `5xx` is the filesystem range and `501`/`502` are otherwise free in both directories (`backport-6.12` has `500-01..500-04`, `backport-6.18` has no other `5xx`).
- **Refresh is correct and complete.** The only content change in the two renamed files is the `From:` SHA plus four hunk headers shifting back `-2` (`9016`->`9014`, `9074`->`9072`, etc.), which is exactly what falls out of `501` now applying *before* `502` instead of after it. I applied `501` then `502` in order onto `fs/smb/server` from `v6.12.103` and `v6.18.44`: all hunks apply with zero fuzz and zero offset in both, so `502` correctly did not need re-refreshing (its `smb2pdu.c` hunk at `err_out2:` sits ~5000 lines above anything `501` touches).
- **No unrelated churn.** The PR's net change is 4 files; the two `target/linux/realtek/patches-6.18/*-v7.3-*` deletions visible in a raw `base..head` diff are merge-base artifacts of the branch point, not part of the PR.
- CI on `c98f334` is green: both workflow runs (`Build Kernel` 32419938458 and `Build all core packages` 32419938421) completed with conclusion `success`.

---

jonasjelonek:
Besides the AI-flagged issue, please do the following too:
- combine the two backports into one commit, there is no need to do it separately per kernel version
- keep the patches added by this PR as 501, renumber the previously added patches to 502 (to have those v7.1 patches before the v7.3 patches)
- do the move and renumber of the v7.3 patches also in a single commit

pesa1234:
> Besides the AI-flagged issue, please do the following too:
> 
>     * combine the two backports into one commit, there is no need to do it separately per kernel version
> 
>     * keep the patches added by this PR as 501, renumber the previously added patches to 502 (to have those v7.1 patches before the v7.3 patches)
> 
>     * do the move and renumber of the v7.3 patches also in a single commit

Done, thanks. I rebuilt the series as requested:
- combined the 6.12 and 6.18 session teardown backports into a single commit;
- kept the v7.1 patches as 501;
- moved and renumbered both v7.3 signing patches to 502 in a single commit;
- fixed the indentation issue flagged by the AI review in both patches.
I also refreshed and verified the 501 → 502 ordering against Linux 6.12.103 and 6.18.44.

openwrt-ai (commented):
Reviewed the rewritten series (2 commits, both new since my last review at `c98f334`); no new issues found. All four items from the maintainer's last comment are implemented, and the indentation nit I raised previously is resolved.

Verification of the reworked series:

- **Combined per-kernel commits.** `7fa5c36` now adds both `backport-6.12/501-v7.1-...` and `backport-6.18/501-v7.1-...` in one commit, and `4fc12f8` performs both `pending-* -> backport-*` moves plus both renumbers in one commit. Both commit messages match their diffs exactly (`--stat` shows 2 added files and 2 renames respectively).
- **Renumbering is the one requested.** v7.1 teardown patches keep `501`; the v7.3 signing patches move to `502`, so the v7.1 fix applies first. The teardown patch files are a 100%-similarity rename of the content I validated in my earlier reviews — unchanged.
- **Indentation now matches upstream byte-for-byte.** The continuation line is `5 tabs + 3 spaces` in `a4c06ff2eb66fc704582a443b47798a52c4303e2` (`fs/smb/server/smb2pdu.c`), which is what both `502-v7.3-*` patches now carry. I fetched the commit from `smfrench/smb3-kernel` and compared; the previous `5 tabs + 2 spaces` form was the deviation, so this change is correct for a verbatim backport even though 2 spaces would be the strictly-aligned form.
- **Refresh is correct.** The signing patches' hunk headers shift back to `+2` (`9014`->`9016`, `9072`->`9074`, etc.), which is exactly what falls out of `501` teardown now applying before `502`. The teardown patch itself needed no re-refresh: its only `smb2pdu.c` hunk sits at `err_out2:` (~line 3765/3777), ~5000 lines above anything the signing patch touches.
- **Apply test.** Copying `fs/smb/server` from `v6.12.103` (`25c09b42358e73e1476e517b296edb6344f2e4bd`) and `v6.18.44` (`1efe5d048a391de3ead2804b2e7f86376c356cc5`), then applying `501` followed by `502` with `patch -F0`: all hunks apply in both trees with zero fuzz and zero offset.
- **Post-apply result matches upstream.** For 6.18 the entire `smb2_get_sign_rsp_iov()` + `smb2_set_sign_rsp()` block is byte-identical to `a4c06ff`. The only diffs for 6.12 (`if (!ksmbd_sign_smb2_pdu(...))`) and for `smb3_set_sign_rsp()` in both (the `conn->binding == false` condition, `if (!ksmbd_sign_smb3_pdu(...))`) are pre-existing stable-branch context that predates this patch, not backport-introduced changes — the lines the patch actually alters (`int n_vec;`, the `iov = smb2_get_sign_rsp_iov(...)` call) are identical.
- **Upstream SHAs verified.** `a42896bebfcc287ed1e61d820a888e33b1eb80ce` ("ksmbd: harden file lifetime during session teardown", DaeMyung Kang, 2026-04-28) and `a4c06ff2eb66fc704582a443b47798a52c4303e2` ("smb/server: fix signing when a response uses more than one iov", ChenXiaoSong, 2026-07-23) both fetch and carry exactly the subject/author/date embedded in the patch files.
- **No numbering collision, no stale references.** `backport-6.12` holds `500-01..500-04` plus the new `501`/`502`; `backport-6.18` holds only `501`/`502`. `pending-6.12`/`pending-6.18` no longer carry a `501`. A tree-wide grep finds no other reference to either patch filename.
- Both commits carry a well-formed `Signed-off-by: Andrea Pesaresi <andreapesaresi82@gmail.com>`.
- CI on `4fc12f8` is green: `Build Kernel` (32462010318) and `Build all core packages` (32462010234) both completed with conclusion `success`.

---

jonasjelonek:

GithPK:
The KSMBD on OpenWrt SNAPSHOT r35906-3d1645ee26 with Kernel 6.18.44 and all the included fixes works fine now.

Thank you All for the commitment.

I close the issue #24738

PR #24811, opened 2026-08-20
ncurses: disable widec for host build

mgz0227:
The ncurses 6.6 update inherits the ABI 6 default of enabling
wide-character support. This silently changes host ncurses from narrow
to wide and conflicts with the packages feed's narrow-host assumption.

Python 3.13 then enables its wide-character curses path without matching
declarations and fails with:

```
error: implicit declaration of function 'wget_wch'
```

Explicitly disable widec for host builds. Target builds remain
explicitly wide-character.

The failure was reproduced on x86/64 in a clean GitHub-hosted build:
https://github.com/mgz0227/OpenWrt_x86/actions/runs/32355019702

Checked with `./scripts/checkpatch.pl --strict`.

### DIFF
--- a/package/libs/ncurses/Makefile
+++ b/package/libs/ncurses/Makefile
@@ -89,6 +89,7 @@ CONFIGURE_VARS += \
 HOST_CFLAGS += $(HOST_FPIC)
 
 HOST_CONFIGURE_ARGS += \
+	--disable-widec \
 	--enable-pc-files \
 	--without-cxx \
 	--without-cxx-binding \


openwrt-ai (commented):

The `Fixes:` trailer correctly references `18725c45a33a ("ncurses: bump to 6.6.20260801")`, and the one-line addition of `--disable-widec` to `HOST_CONFIGURE_ARGS` cleanly mirrors the explicit `--enable-widec` already used for the target build.

---

mgz0227:
Closing this because the proposed core change is incorrect.

The failing malta/be and x86/64 jobs show that --disable-widec also disables extended colors in the host tic. This reduces MAX_ENTRY_SIZE from 32768 to 4096, so tic skips xterm and xterm-256color and the target terminfo package cannot be assembled.

The original Python failure comes from combining the openwrt-25.12 packages feed narrow ncurses assumption with main wide host ncurses. It should be fixed at that integration boundary rather than changing the global core host tool.

Sorry for the noise.

dangowrt:
Thanks for the diagnosis and for catching the `MAX_ENTRY_SIZE`/tic regression before it landed — saved me from going down the same road.

Traced it further: host ncurses picks up wide-char by default now because ncurses 6.6+'s own `configure` gates `cf_dft_widec` on `cf_cv_abi_default` (== major version) landing in ABI 6/7/8/9 — that policy accumulated in Dickey's git tree sometime after the 6.4 tarball was cut, so jumping straight from 6.4 to a live snapshot pulled it in as a side effect, not a deliberate choice by anyone.

Agreed on fixing it at the integration boundary rather than the global host tool. Opened two PRs on that basis:
- #24817 — an independent ncurses-side hygiene fix (its own pkg-config files were landing in `staging_dir/host` instead of `staging_dir/hostpkg` — real bug, but not what breaks python3, since `PKG_CONFIG_PATH` searches both)
- openwrt/packages#30341 — the actual fix: python3's `ac_cv_header_ncursesw_*=no` overrides (from d883c02a4, written when host ncurses was narrow-only) now fight with its own pkg-config-based `ncursesw` detection instead of agreeing with it. Dropping the now-stale overrides lets both agree, verified end-to-end with a clean host python3 rebuild.

mgz0227:
Thanks for tracing this further and for opening both PRs.

The `cf_dft_widec` default change and the disagreement between Python's pkg-config detection and the forced `ac_cv_header_ncursesw_*=no` results make the root cause clear. I agree that openwrt/packages#30341 is the correct fix for the Python failure, while #24817 is a separate hostpkg/pkg-config hygiene fix.

My current downstream workaround is broader than necessary. I'll align it with openwrt/packages#30341: keep the non-namespaced header guards, drop only the stale `ncursesw/*` overrides, and rely on pkg-config for the staged include path. I'll verify that exact change against the openwrt-25.12/Python 3.13 + main ncurses combination and report back if it behaves differently.

No need to reopen #24811 from my side. Thanks again for following this through.

PR #1479, opened 2018-10-19
ipq40xx: add support for EnGenius EAP1300

gamism:
SOC:    IPQ4018 / QCA Dakota
CPU:    Quad-Core ARMv7 Processor rev 5 (v7l) Cortex-A7
DRAM:   256 MiB
NOR:    32 MiB
ETH:    Qualcomm Atheros QCA8072
WLAN1:  Qualcomm Atheros QCA4018 2.4GHz 802.11bgn 2:2x2
WLAN2:  Qualcomm Atheros QCA4018 5GHz 802.11a/n/ac 2:2x2
INPUT:  RESET Button
LEDS:   Power, LAN, MESH, WLAN 2.4GHz, WLAN 5GHz

1. Load Ramdisk via U-Boot

To set up the flash memory environment, do the following:
a. As a preliminary step, ensure that the board console port is connected to the PC using these RS232 parameters:
   * 115200bps
   * 8N1
b. Confirm that the PC is connected to the board using one of the Ethernet ports. Set a static ip 192.168.99.8 for Ethernet that connects to board. The PC must have a TFTP server launched and listening on the interface to which the board is connected. At this stage power up the board and, after a few seconds, press 4 and then any key during the countdown.

U-BOOT> set fdt_high 0x87000000 & set serverip 192.168.99.8 && set ipaddr 192.168.99.9 && tftpboot 0x84000000 openwrt-ipq40xx-engenius_eap1300-initramfs-fit-uImage.itb && bootm 0x84000000

2. Load image via GUI

a. Upgrade EAP1300 to FW v3.5.3.2
In the GUI, System Manager > Firmware > Firmware Upgrade, to do upgrade.
b. Transfer to OpenWrt from EnGenius.
In Firmware Upgrade page, to upgrade yours openwrt-ipq40xx-engenius_eap1300-squashfs-sysupgrade.bin.

3. Revert to EnGenius EAP1300
To flash openwrt-ipq40xx-engenius_eap1300-squashfs-factory.bin by using sysupgrade command and "DO NOT" keep configuration.
$ sysupgrade –n openwrt-ipq40xx-engenius_eap1300-squashfs-factory.bin

Signed-off-by: Steven Lin <steven.lin@senao.com>

PR #24812, opened 2026-08-20
ipq40xx: reduce kernel size on the generic subtarget

divinehawk:
Multiple devices in the ipq40xx target are limited on kernel size due to partition layout. Due to the addition of CONFIG_PWM=y in https://github.com/openwrt/openwrt/commit/e4de94870370e0948987a18f6409c353c891e28e, the kernel size exceeds the limit.

```
linux-ipq40xx_generic/netgear_rbr40-zImage.itb is too big: 3943708 > 3932160
linux-ipq40xx_generic/netgear_rbs40-zImage.itb is too big: 3943712 > 3932160
linux-ipq40xx_generic/netgear_rbr50-zImage.itb is too big: 3943716 > 3932160
linux-ipq40xx_generic/netgear_rbs50-zImage.itb is too big: 3943720 > 3932160
linux-ipq40xx_generic/netgear_srr60-zImage.itb is too big: 3943836 > 3932160
linux-ipq40xx_generic/netgear_srs60-zImage.itb is too big: 3943840 > 3932160
linux-ipq40xx_generic/engenius_eap1300-uImage.itb is too big: 5660548 > 5242880
```

Orbi (rbr40, rbs40, rbr50, rbs50, srr60, srs60) — over by ~11 KB against a real partition limit, from CONFIG_PWM=y (e4de948) and the display feature on the generic subtarget (20d94d5). The first commit modularizes drivers that don't need to be built in: new kmod-leds-lp5521, plus leds-lp5523, leds-lp5562, tlc591xx, gpio-nxp-74hc164, eeprom-at24, spi-gpio, and drops CONFIG_SPI_BITBANG=y.

EAP1300 — over by ~400 KB, unrelated to PWM, and against a limit that doesn't exist. The board has a single 30464k firmware partition with compatible = "denx,fit"; the parser reads the kernel length from the FIT header at runtime and the recipe never pads the kernel, so KERNEL_SIZE := 5120k marks no boundary on flash. IMAGE_SIZE := 25344k was the remainder of that notional split and was never enforced, since the sysupgrade recipe doesn't call check-size. This was flagged during the original device submission (#1479) and left in place. The second commit drops KERNEL_SIZE, sets IMAGE_SIZE to the full partition, and checks the combined image instead.

Only visible on the buildbot: ALL_KMODS selects the display kmods, which force CONFIG_VT, CONFIG_FRAMEBUFFER_CONSOLE and the input core (via select INPUT) into the kernel. The EAP1300 additionally has DEFAULT := n, so a local build never produces that image at all


target/linux/ipq40xx/image/generic.mk
```diff
@@ -847,7 +847,7 @@ define Device/meraki_common
 	BLOCKSIZE := 128k
 	PAGESIZE := 2048
 	DEVICE_DTS_LOADADDR := 0x89000000
-	DEVICE_PACKAGES := ath10k-firmware-qca9887-ct
+	DEVICE_PACKAGES := ath10k-firmware-qca9887-ct kmod-leds-lp5562
```
openwrt-ai: Three devices that inherit `meraki_common` reassign `DEVICE_PACKAGES` with `:=` after `$(call Device/meraki_common)`, which discards the `kmod-leds-lp5562` added here:

- [`meraki_z3`](https://github.com/openwrt/openwrt/blob/cb6d5deeddf69ac94c6a917e29fb6997367940ce/target/linux/ipq40xx/image/generic.mk#L890-L896)
- [`meraki_gx20`](https://github.com/openwrt/openwrt/blob/cb6d5deeddf69ac94c6a917e29fb6997367940ce/target/linux/ipq40xx/image/generic.mk#L898-L904)
- [`meraki_z3c`](https://github.com/openwrt/openwrt/blob/cb6d5deeddf69ac94c6a917e29fb6997367940ce/target/linux/ipq40xx/image/generic.mk#L906-L912)

All three include `qcom-ipq4029-meraki-insect.dtsi`, which instantiates an enabled `ti,lp5562` controller driving the boot/failsafe/running status LED. With `CONFIG_LEDS_LP5562=y` now dropped from `config-6.18`, these devices would ship with neither the built-in driver nor the module, losing their status LED.

Add `kmod-leds-lp5562` to each of those three device blocks' `DEVICE_PACKAGES` list. The `+=` device (`meraki_mr30h`) and the plain inheritors (`meraki_mr33`, `meraki_mr74`) already keep it.

---

openwrt-ai (commented):
Reviewed 2 new commits. Both commit messages match their diffs. One inline issue: a few meraki devices override `DEVICE_PACKAGES` with `:=` and would drop the newly-required `kmod-leds-lp5562`. Coverage for all other modularized drivers (gpio-74x164, lp5521, lp5523, tlc591xx) checks out across the affected devices and their inheriting variants.

---

target/linux/ipq40xx/image/generic.mk
```diff
@@ -862,13 +862,14 @@ define Device/meraki_mr30h
...
 	DEVICE_DTS_CONFIG := config@2
-	DEVICE_PACKAGES += ipq-wifi-meraki_mr30h
+	DEVICE_PACKAGES += kmod-leds-lp5562 ipq-wifi-meraki_mr30h
 endef
 TARGET_DEVICES += meraki_mr30h
 
 define Device/meraki_mr33
 	$(call Device/meraki_common)
 	DEVICE_MODEL := MR33
+	DEVICE_PACKAGES := kmod-leds-lp5562
```
openwrt-ai: Using `:=` here drops `ath10k-firmware-qca9887-ct`. `Device/meraki_common` now sets [`DEVICE_PACKAGES := ath10k-firmware-qca9887-ct`](https://github.com/openwrt/openwrt/blob/ee863d72a6d3b76c8f26314dce9af401932dc1f7/target/linux/ipq40xx/image/generic.mk#L847), and since `$(call Device/meraki_common)` is expanded inline before this line, a second `:=` replaces the value rather than extending it (`DEVICE_PACKAGES` starts empty from [`Device/Init`](https://github.com/openwrt/openwrt/blob/ee863d72a6d3b76c8f26314dce9af401932dc1f7/include/image.mk#L466)). Before this PR, MR33 and MR74 inherited both entries from `meraki_common`.

MR33 has an enabled PCIe `qcom,ath10k` node ([qcom-ipq4029-mr33.dts:21-27](https://github.com/openwrt/openwrt/blob/ee863d72a6d3b76c8f26314dce9af401932dc1f7/target/linux/ipq40xx/dts/qcom-ipq4029-mr33.dts#L21-L27)), so it would ship without the QCA9887 firmware and lose that radio.

Use `+=`, as `meraki_mr30h` does:

```suggestion
	DEVICE_PACKAGES += kmod-leds-lp5562
```

Same problem at line 888 for `meraki_mr74`, which has the same PCIe ath10k node ([qcom-ipq4029-mr74.dts:21-27](https://github.com/openwrt/openwrt/blob/ee863d72a6d3b76c8f26314dce9af401932dc1f7/target/linux/ipq40xx/dts/qcom-ipq4029-mr74.dts#L21-L27)).

Alternatively — all six devices that include `qcom-ipq4029-meraki-insect.dtsi` (mr30h, mr33, mr74, z3, z3c, gx20) need `kmod-leds-lp5562`, and they are exactly the set of `meraki_common` users apart from mr20/mr70 (underdog). Keeping `kmod-leds-lp5562` in `meraki_common` and switching the three `:=` overriders (`meraki_z3`, `meraki_gx20`, `meraki_z3c`) to `+=` would avoid repeating it six times and remove this class of override bug entirely.

---

openwrt-ai (commented):
Reviewed the amended `ipq40xx: switch to kernel modules to reduce kernel size` commit (`cb6d5de` → `ee863d7`). The `kmod-gpio-ncp-74hc164` → `kmod-gpio-nxp-74hc164` typo fix is correct (`KernelPackage/gpio-nxp-74hc164` in `package/kernel/linux/modules/gpio.mk`), and all five ipq40xx boards with a `fairchild,74hc595` node (rtl30vw, cm520-79f, ex61x0v2, wac510, rutx50) now carry it. lp5523, lp5521 and tlc591xx coverage also matches the DTS users.

One regression introduced by the meraki restructure: moving `kmod-leds-lp5562` out of `Device/meraki_common` and adding it back with `:=` on `meraki_mr33` and `meraki_mr74` overwrites the inherited `ath10k-firmware-qca9887-ct`. Details inline.

The `FormalityCheck` PKG_RELEASE warning on `package/kernel/linux` looks like a false positive — that Makefile has no `PKG_RELEASE`/`PKG_VERSION` to bump.

Commit message matches its diff; no commit-message issues.

---

divinehawk:
PWM code added 12kb, this PR reduced by 7kb, but we are still over by 5kb.

```
ipq40xx_generic/netgear_rbr40-zImage.itb is too big: 3937028 > 3932160
ipq40xx_generic/netgear_rbs40-zImage.itb is too big: 3937032 > 3932160
ipq40xx_generic/netgear_rbr50-zImage.itb is too big: 3937036 > 3932160
ipq40xx_generic/netgear_rbs50-zImage.itb is too big: 3937040 > 3932160
ipq40xx_generic/netgear_srr60-zImage.itb is too big: 3937156 > 3932160
ipq40xx_generic/netgear_srs60-zImage.itb is too big: 3937160 > 3932160
```
Modularize CONFIG_EEPROM_AT24 next

robimarko:
As a temporary thing, we could remove the display feature which compiles in stuff that cannot be built as a module and run your board without display support.

divinehawk:
Progress. 
```
ipq40xx_generic/netgear_rbr40-zImage.itb is too big: 3933764 > 3932160
ipq40xx_generic/netgear_rbs40-zImage.itb is too big: 3933768 > 3932160
ipq40xx_generic/netgear_rbr50-zImage.itb is too big: 3933772 > 3932160
ipq40xx_generic/netgear_rbs50-zImage.itb is too big: 3933776 > 3932160
ipq40xx_generic/netgear_srr60-zImage.itb is too big: 3933892 > 3932160
ipq40xx_generic/netgear_srs60-zImage.itb is too big: 3933896 > 3932160
```

Next is CONFIG_SPI_GPIO, should do it. My buildbot equivalent test takes a few hours, unfortunately.

target/linux/ipq40xx/image/generic.mk
```diff
@@ -847,14 +847,14 @@ define Device/meraki_common
...
-	DEVICE_PACKAGES := ath10k-firmware-qca9887-ct
+	DEVICE_PACKAGES := ath10k-firmware-qca9887-ct kmod-eeprom-at24 kmod-leds-lp5562
 endef
 
 define Device/meraki_mr20
 	$(call Device/meraki_common)
 	DEVICE_MODEL := MR20
 	DEVICE_DTS_CONFIG := config@4
-	DEVICE_PACKAGES := ipq-wifi-meraki_underdog
+	DEVICE_PACKAGES += ipq-wifi-meraki_underdog
```
openwrt-ai: Switching from `:=` to `+=` here also inherits `ath10k-firmware-qca9887-ct` and `kmod-leds-lp5562` from `Device/meraki_common`, and MR20 has neither. [`qcom-ipq4029-mr20.dts`](https://github.com/openwrt/openwrt/blob/e6ebdb3bcdb9fbece168374c71c6f4250bfdb796/target/linux/ipq40xx/dts/qcom-ipq4029-mr20.dts) is a 9-line file that only pulls in [`qcom-ipq4029-meraki-underdog.dtsi`](https://github.com/openwrt/openwrt/blob/e6ebdb3bcdb9fbece168374c71c6f4250bfdb796/target/linux/ipq40xx/dts/qcom-ipq4029-meraki-underdog.dtsi), which has no `&pcie0`/`qcom,ath10k` node at all (only the two built-in `&wifi0`/`&wifi1` radios) and drives its LEDs with [`gpio-leds`](https://github.com/openwrt/openwrt/blob/e6ebdb3bcdb9fbece168374c71c6f4250bfdb796/target/linux/ipq40xx/dts/qcom-ipq4029-meraki-underdog.dtsi#L26-L52), not an LP5562. The previous `:=` kept both out; this revision adds them back.

`kmod-eeprom-at24` genuinely is needed — underdog has the [at24c64 at `blsp1_i2c3`](https://github.com/openwrt/openwrt/blob/e6ebdb3bcdb9fbece168374c71c6f4250bfdb796/target/linux/ipq40xx/dts/qcom-ipq4029-meraki-underdog.dtsi#L56-L62) providing the MAC — so inheriting from `meraki_common` is the right shape; just mask the two that don't apply (`merge_packages` in [`include/image.mk`](https://github.com/openwrt/openwrt/blob/e6ebdb3bcdb9fbece168374c71c6f4250bfdb796/include/image.mk#L601) lets a later `-pkg` cancel an earlier `pkg`):

```suggestion
	DEVICE_PACKAGES += -ath10k-firmware-qca9887-ct -kmod-leds-lp5562 ipq-wifi-meraki_underdog
```

Same applies at line 879 for `meraki_mr70`, which uses the same underdog dtsi.

---

target/linux/ipq40xx/config-6.18
```diff
@@ -501,7 +494,6 @@ CONFIG_SOFTIRQ_ON_OWN_STACK=y
 CONFIG_SPARSE_IRQ=y
 CONFIG_SPI=y
 CONFIG_SPI_BITBANG=y
```
openwrt-ai: `SPI_GPIO` was the only symbol enabled in this target that selects `SPI_BITBANG` ([drivers/spi/Kconfig:483-486](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/spi/Kconfig#L483-L486); `SPI_QUP`, `SPI_MEM` and `MTD_SPI_NOR` don't). With `CONFIG_SPI_GPIO=y` dropped, this line keeps the bitbang core linked into vmlinux for every ipq40xx image — including the Orbi boards this PR is trying to shrink — with no built-in driver left to use it.

`kmod-spi-gpio` already depends on `kmod-spi-bitbang` ([spi.mk:49](https://github.com/openwrt/openwrt/blob/e6ebdb3bcdb9fbece168374c71c6f4250bfdb796/package/kernel/linux/modules/spi.mk#L46-L53)), so removing it here gets it built as a module alongside spi-gpio for exactly the five boards that need it. ramips and qualcommax, which also ship `kmod-spi-gpio` in `DEVICE_PACKAGES`, carry no `CONFIG_SPI_BITBANG` in their target configs.

```suggestion
```

---

openwrt-ai (commented):
Reviewed the changes since `ee863d7` (`kmod-eeprom-at24` + `kmod-spi-gpio` modularization and the meraki `+=` rework).

The earlier `:=`-override problem is fixed, and coverage checks out: the five boards given `kmod-spi-gpio` (rtl30vw, cm520-79f, ex61x0v2, wac510, rutx50) are exactly the five ipq40xx DTS files with a `compatible = "spi-gpio"` node, and `ex6100v2`/`ex6150v2` inherit it cleanly. For at24, all eight `meraki_common` users have an `atmel,24c64` (insect/underdog dtsi), and the only other ipq40xx at24 user, `luma_wrtq-329acn`, already carried `kmod-eeprom-at24`.

Two issues, both inline:
- `meraki_mr20` / `meraki_mr70` now inherit `ath10k-firmware-qca9887-ct` and `kmod-leds-lp5562`, neither of which exists on the underdog dtsi — the old `:=` masked them.
- `CONFIG_SPI_BITBANG=y` is left behind in `config-6.18` with no built-in consumer once `SPI_GPIO` is modular; it can go too and save a bit more vmlinux on the Orbi boards.

Commit messages match their diffs; CI on `e6ebdb3` is still running with no failures so far.

---

openwrt-ai (commented):
Reviewed the amended `ipq40xx: switch to kernel modules to reduce kernel size` commit (`e6ebdb3` → `f3a32c9`). Both findings from the previous round are addressed and the changes are correct:

- `meraki_mr20` / `meraki_mr70` now use `DEVICE_PACKAGES += -ath10k-firmware-qca9887-ct -kmod-leds-lp5562 ipq-wifi-meraki_underdog`. `merge_packages` in `include/image.mk` filters an earlier `pkg` when a later `-pkg` appears, and `$(call Device/meraki_common)` is expanded before the `+=`, so the ordering cancels both correctly while keeping the inherited `kmod-eeprom-at24` that the underdog dtsi's at24c64 needs. This matches the existing `-ath10k-firmware-qca9887-ct` idiom already used by `meraki_z3` / `meraki_z3c` / `meraki_gx20`.
- `CONFIG_SPI_BITBANG=y` is dropped from `config-6.18`; the only remaining reference in the tree is `KCONFIG:=CONFIG_SPI_BITBANG` in `package/kernel/linux/modules/spi.mk`, which `kmod-spi-gpio` depends on, so the five boards that need it pull it in as a module.

CI on `f3a32c9` is green — no failing check runs, including `Build Kernel with external toolchain (ipq40xx, chromium)` and the ipq40xx patch checks. The `FormalityCheck` PKG_RELEASE warning on `package/kernel/linux` is still the same false positive (that Makefile carries no `PKG_RELEASE`/`PKG_VERSION`).

Commit message matches its diff; no new issues found.

---

divinehawk:
I've eliminated all but this one..  it could be another issue separate than CONFIG_PWM:

`linux-ipq40xx_generic/engenius_eap1300-uImage.itb is too big: 5643876 > 5242880`

(still working on it)

target/linux/ipq40xx/image/generic.mk
```diff
@@ -458,9 +458,8 @@ define Device/engenius_eap1300
 	DEVICE_DTS_CONFIG := config@4
 	BOARD_NAME := eap1300
 	SOC := qcom-ipq4018
-	KERNEL_SIZE := 5120k
-	IMAGE_SIZE := 25344k
-	IMAGE/sysupgrade.bin := append-kernel | append-rootfs | pad-rootfs | append-metadata
+	IMAGE_SIZE := 30464k
+	IMAGE/sysupgrade.bin := append-kernel | append-rootfs | pad-rootfs | check-size | append-metadata
```
openwrt-ai, line 462: nit: the commit message says this matches "the other EnGenius devices in this target", but none of them actually use this shape — [`engenius_emd1`](https://github.com/openwrt/openwrt/blob/ef87cf2c5218d2ba4b8cc742d284077ab19a6e1b/target/linux/ipq40xx/image/generic.mk#L480-L490) has no `check-size` in its sysupgrade recipe at all, [`engenius_emr3500`](https://github.com/openwrt/openwrt/blob/ef87cf2c5218d2ba4b8cc742d284077ab19a6e1b/target/linux/ipq40xx/image/generic.mk#L500-L501) still carries `KERNEL_SIZE := 4096k`, and [`engenius_ens620ext`](https://github.com/openwrt/openwrt/blob/ef87cf2c5218d2ba4b8cc742d284077ab19a6e1b/target/linux/ipq40xx/image/generic.mk#L526-L530) keeps `KERNEL_SIZE := 5120k` plus a partition-remainder `IMAGE_SIZE := 21312k` (and it does need `KERNEL_SIZE`, its factory recipe pads with it). All three are also commented out of `TARGET_DEVICES`. The devices this change really lines up with are the extreme-networks ones ([ws-ap3915i](https://github.com/openwrt/openwrt/blob/ef87cf2c5218d2ba4b8cc742d284077ab19a6e1b/target/linux/ipq40xx/image/generic.mk#L538-L546)) — worth rewording so the justification doesn't point at devices that contradict it.

The change itself checks out: the `firmware` partition is `reg = <0x190000 0x1dc0000>` = 30464k in [qcom-ipq4018-eap1300.dts](https://github.com/openwrt/openwrt/blob/ef87cf2c5218d2ba4b8cc742d284077ab19a6e1b/target/linux/ipq40xx/dts/qcom-ipq4018-eap1300.dts#L190-L193), it ends exactly where `u-boot-env` begins, and `mtdsplit_fit` derives the kernel/rootfs split from the FIT header at runtime rather than from a fixed offset, so `KERNEL_SIZE` really was only a build-time constraint here.

---

openwrt-ai (commented):
Reviewed the one new commit since `f3a32c9` — `ef87cf2` "ipq40xx: fix image size limits for EnGenius EAP1300" (the earlier two commits are unchanged apart from the rebase).

The reasoning holds up:
- The `firmware` partition in `qcom-ipq4018-eap1300.dts` is `reg = <0x190000 0x1dc0000>`, i.e. exactly 30464k, and ends where `u-boot-env` (`0x1f50000`) starts — so the new `IMAGE_SIZE` is the full partition, not an approximation. The old `5120k + 25344k` split adds up to the same total, confirming it was a notional split.
- `mtdsplit_fit` (`target/linux/generic/files/drivers/mtd/mtdsplit/mtdsplit_fit.c`) reads `totalsize` from the FIT header and locates the rootfs with `mtd_find_rootfs_from(fit_offset + fit_size, ...)`, rounding the kernel partition up to an eraseblock at runtime. There is no fixed kernel boundary on flash, and `Device/FitImage` in this target does not pad, so dropping `KERNEL_SIZE` removes only the build-time `check-size` on the kernel — nothing on-flash changes.
- `IMAGE/sysupgrade.bin` gaining `check-size` before `append-metadata` matches the ordering used elsewhere in this file, and `KERNEL_SIZE` had no other consumer for this device (no `pad-to`, no factory recipe).

One nit inline on the commit message's "matching the other EnGenius devices in this target" justification, which the in-tree recipes don't support.

CI on `ef87cf2` is queued/running with no failures so far. The `FormalityCheck` `PKG_RELEASE` warning on `package/kernel/linux` remains the same false positive as in earlier rounds.

---

divinehawk:
Between the modules and the KERNEL_SIZE fix, this should resolve the buildbot failures. 

Several of the devices are still near the limit. 

Related PR https://github.com/openwrt/openwrt/pull/24858 to separate fb and fb-console so that devices with tiny screens don't have the fbconsole and VT code if it isn't needed.

robimarko:

PR #24814, opened 2026-08-20
arm-trusted-firmware-mediatek: update to latest upstream HEAD

dangowrt:
**Maintainer:** @dangowrt

Two real commits since the previous pin:

- mediatek: mt7622: move PMIC init before PLL/DRAM stage — fixes a hang in DRAM calibration on reboot when cpufreq had lowered Vcore before the reset, by moving PMIC init earlier so it no longer depends on the Vcore level left over from the previous boot.
- mediatek: mt7981, mt7986, mt7987, mt7988: add SMCCC TRNG support — implements the standard SMCCC TRNG interface (upstream U-Boot doesn't support the MTK-specific one; our `uboot-mediatek` carries a downstream patch for that), allowing U-Boot to provide a KASLR seed to Linux. Tested on MT7981 by upstream; not yet verified on the other three SoCs by upstream.

Mirror hash verified via an independent fresh clone of the pinned commit.


package/boot/arm-trusted-firmware-mediatek/Makefile
```diff
@@ -13,9 +13,9 @@ PKG_RELEASE:=1
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL=https://github.com/mtk-openwrt/arm-trusted-firmware.git
-PKG_SOURCE_DATE:=2026-01-23
-PKG_SOURCE_VERSION:=e06f258664198a901ff1c7c0c87802a115179451
-PKG_MIRROR_HASH:=e094f5c2838f366e9b2d87e85e0c324ac3b955dd02a493adced843031cf6b517
+PKG_SOURCE_DATE:=2026-05-21
+PKG_SOURCE_VERSION:=476cf6629f7675c9857b59458851dda64de88343
```
openwrt-ai, line 17: nit: the commit message and the PR body disagree on this point. The commit message says "U-Boot doesn't support the MTK-specific one", while the PR body qualifies it as "*upstream* U-Boot doesn't support the MTK-specific one; our `uboot-mediatek` carries a downstream patch for that". Without the qualifier the commit message reads as a claim about the U-Boot OpenWrt actually ships, which already drives the MTK TRNG directly via [181-rng-mtk-rng-v2-add-U-Boot-driver-for-Mediatek-TRNG-v.patch](``https://github.com/openwrt/openwrt/blob/b36e9e7de2dba42d8c3cb286c54627329cd0c461/package/boot/uboot-mediatek/patches/181-rng-mtk-rng-v2-add-U-Boot-driver-for-Mediatek-TRNG-v.patch)`` plus RNG DT nodes for exactly the four SoCs this ATF commit covers ([193-mediatek-mt7981-add-RNG-node.patch](``https://github.com/openwrt/openwrt/blob/b36e9e7de2dba42d8c3cb286c54627329cd0c461/package/boot/uboot-mediatek/patches/193-mediatek-mt7981-add-RNG-node.patch),`` [192](``https://github.com/openwrt/openwrt/blob/b36e9e7de2dba42d8c3cb286c54627329cd0c461/package/boot/uboot-mediatek/patches/192-mediatek-mt7986-add-RNG-node.patch),`` [194](``https://github.com/openwrt/openwrt/blob/b36e9e7de2dba42d8c3cb286c54627329cd0c461/package/boot/uboot-mediatek/patches/194-mediatek-mt7988-add-RNG-node.patch),`` [195](``https://github.com/openwrt/openwrt/blob/b36e9e7de2dba42d8c3cb286c54627329cd0c461/package/boot/uboot-mediatek/patches/195-mediatek-mt7987-add-RNG-node.patch)).`` Carrying the word "upstream" over from the PR body into the commit message would keep the two in sync.

---

openwrt-ai (commented):
Reviewed 1 new commit. Verified the following against the upstream repo:

- `476cf662` is the current tip of `refs/heads/mtksoc` in `mtk-openwrt/arm-trusted-firmware`, and `e06f2586..476cf662` is exactly the two commits named in the message, in that order — nothing skipped, no rebase.
- `PKG_SOURCE_DATE:=2026-05-21` matches the committer date of `476cf662` (2026-05-21 16:05:04 +0800).
- All seven patches under `package/boot/arm-trusted-firmware-mediatek/patches/` apply at the new pin with zero fuzz; neither new commit touches any file they patch (the SNFI/SPI-NAND drivers and `plat/mediatek/mt7987/bl2/`).
- The TRNG commit sets `TRNG_SUPPORT := 1` inside each `plat/mediatek/mt79{81,86,87,88}/platform.mk` and renames `apsoc_common/drivers/trng/v{1,2}/rng.c` to `trng.c`, updating every `bl31.mk` that referenced it. `mt7622` does not reference those files, so no dangling reference and no new `TFA_MAKE_FLAGS` entry is needed on the OpenWrt side.

One caveat on verification coverage: I could not reproduce `PKG_MIRROR_HASH` here (no `zstd` available in this environment), and the core-package build for this PR ran on `x86/64` and `malta/be`, neither of which builds this package — so the new mirror hash and the ATF build itself are resting entirely on your local verification.

---

peterwillcn (approved):

openwrt-ai (commented):

---

package/boot/arm-trusted-firmware-mediatek/Makefile
openwrt-ai: fixed, thanks

---

PR #24817, opened 2026-08-20
ncurses: fix host pkg-config file install location

dangowrt:
**Maintainer:** @dangowrt

ncurses' own `./configure` determines where to install its generated `*.pc` files by querying pkg-config at configure time, independently of `--prefix`/`HOST_BUILD_PREFIX`. On a typical build host that resolves to `staging_dir/host/lib/pkgconfig`, not `staging_dir/hostpkg/lib/pkgconfig`, which is where `HostBuild.mk` (and everything that depends on a host package via `PKG_BUILD_DEPENDS`/`HOST_BUILD_DEPENDS`) actually expects a package's own pkg-config files to live.

Explicitly pass `--with-pkg-config-libdir=$(STAGING_DIR_HOSTPKG)/lib/pkgconfig` for the host build, mirroring the target build's existing `--with-pkg-config-libdir=/usr/lib/pkgconfig`. Verified directly: after this change `ncurses.pc`/`ncursesw.pc` (and panel/menu/form counterparts) land in `staging_dir/hostpkg/lib/pkgconfig` with the correct embedded prefix, and nothing is installed into `staging_dir/host` any more.

This is independent of the widec/narrow default that started causing host-python3 build failures after 18725c45a3 ("ncurses: bump to 6.6.20260801") — see #24811 for that issue and `lang/python/python3`'s own fix in the packages feed (openwrt/packages, to follow). Even before this change, pkg-config still found ncurses's host `.pc` files via the `staging_dir/host` component of `PKG_CONFIG_PATH`, so this is a hygiene/correctness fix for staying within the intended hostpkg/host separation, not a fix for that failure by itself.

### DIFF
--- a/package/libs/ncurses/Makefile
+++ b/package/libs/ncurses/Makefile
@@ -12,7 +12,7 @@ PKG_CPE_ID:=cpe:/a:gnu:ncurses
 PKG_VERSION_MAJOR:=6.6
 PKG_SNAPSHOT_DATE:=20260801
 PKG_VERSION:=$(PKG_VERSION_MAJOR)$(if $(PKG_SNAPSHOT_DATE),.$(PKG_SNAPSHOT_DATE),)
-PKG_RELEASE:=1
+PKG_RELEASE:=2
 
 PKG_SOURCE_PROTO:=git
 PKG_SOURCE_URL:=https://github.com/ThomasDickey/ncurses-snapshots.git
@@ -90,6 +90,7 @@ HOST_CFLAGS += $(HOST_FPIC)
 
 HOST_CONFIGURE_ARGS += \
 	--enable-pc-files \
+	--with-pkg-config-libdir=$(STAGING_DIR_HOSTPKG)/lib/pkgconfig \
 	--without-cxx \
 	--without-cxx-binding \
 	--without-ada \
@@ -197,6 +198,27 @@ endef
 define Host/Compile
 	$(MAKE) -C $(HOST_BUILD_DIR) libs
 	$(MAKE) -C $(HOST_BUILD_DIR)/progs tic
+endef
+
+# Trees that built host ncurses before --with-pkg-config-libdir was added
+# have ncurses' *.pc files left over in $(STAGING_DIR_HOST)/lib/pkgconfig,
+# which still shadow the correct copies in $(HOST_BUILD_PREFIX)/lib/pkgconfig
+# since PKG_CONFIG_PATH searches STAGING_DIR_HOST first. Both spellings can
+# be present: the narrow names from 6.4 and older, and the widec names from
+# the 6.6 snapshots. Clean them up whenever host ncurses is (re)installed, so
+# a plain rebuild after pulling this change stops resolving pkg-config
+# ncurses/ncursesw to the stale prefix.
+define Host/CleanStaleHostPkgConfig
+	rm -f $(foreach l,ncurses panel menu form, \
+		$(STAGING_DIR_HOST)/lib/pkgconfig/$(l).pc \
+		$(STAGING_DIR_HOST)/lib/pkgconfig/$(l)w.pc)
+endef
+
+Hooks/HostInstall/Post := Host/CleanStaleHostPkgConfig $(Hooks/HostInstall/Post)
+
+# Also cover the explicit host-clean path.
+define Host/Uninstall
+	$(call Host/CleanStaleHostPkgConfig)
 endef
 
 $(eval $(call HostBuild))


package/libs/ncurses/Makefile
```diff
@@ -90,6 +90,7 @@ HOST_CFLAGS += $(HOST_FPIC)
 
 HOST_CONFIGURE_ARGS += \
 	--enable-pc-files \
+	--with-pkg-config-libdir=$(STAGING_DIR_HOSTPKG)/lib/pkgconfig \
```
openwrt-ai, line 93: Does this leave stale `.pc` files behind on existing build trees? [`include/host-build.mk:171-172`](https://github.com/openwrt/openwrt/blob/4deede8f04b3a1f5817fc48614fa74ca5403ceeb/include/host-build.mk#L171-L172) exports `PKG_CONFIG_PATH=$(STAGING_DIR_HOST)/lib/pkgconfig:$(HOST_BUILD_PREFIX)/lib/pkgconfig` — `staging_dir/host` comes *first*. On a tree that already built ncurses/host before this change, `staging_dir/host/lib/pkgconfig/ncurses*.pc` is still there (nothing in `HostBuild` removes it — no `Host/Uninstall` is defined for this package, and `host-clean` only drops the stamps and `HOST_BUILD_PREFIX` content), so those older files keep shadowing the newly correct ones in `staging_dir/hostpkg/lib/pkgconfig`.

That makes the "nothing is installed into `staging_dir/host` any more" claim true only for fresh trees, and incremental builders would silently keep resolving `pkg-config ncursesw` to the old prefix — which matters more than usual right now given the widec/narrow change in 18725c45a3 that #24811 is about. Worth adding a `Host/Uninstall` (or a `Hooks/HostInstall/Post` cleanup) that removes the strays from `$(STAGING_DIR_HOST)/lib/pkgconfig`?

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the diff, the referenced commit `18725c45a3` exists in `main`, and CI on `4deede8` is green. Verified against ncurses' `CF_WITH_PKG_CONFIG_LIBDIR` (aclocal.m4, version 25) that an absolute `--with-pkg-config-libdir` value is taken verbatim (`PKG_CONFIG_LIBDIR="$cf_search_path"`) without a directory-existence check, so passing a not-yet-created `staging_dir/hostpkg/lib/pkgconfig` is safe. One question inline about leftovers on incremental trees.

---

dangowrt:
Good catch, real gap. Confirmed the mechanism directly: `host-clean-build` (called by `host-clean`) already calls `$(call Host/Uninstall)` — nothing was defined for ncurses, so nothing removed the stale files. Added one that clears the four stale `*w.pc` files from `$(STAGING_DIR_HOST)/lib/pkgconfig`.

Verified end to end: recreated the stale-file scenario (touched all four in `staging_dir/host/lib/pkgconfig`, alongside an unrelated pre-existing `libdw.pc`), ran `make package/libs/ncurses/host/clean`, confirmed exactly the four target files are removed and `libdw.pc` is untouched.

Kept the `-I$(STAGING_DIR_HOSTPKG)/include/ncursesw` removal from `d883c02a4106` as-is per the PR description — not something this PR touches.

package/libs/ncurses/Makefile
```diff
@@ -199,6 +200,20 @@ define Host/Compile
...
+# copies in $(HOST_BUILD_PREFIX)/lib/pkgconfig since PKG_CONFIG_PATH
+# searches STAGING_DIR_HOST first. Clean those up on a host-clean so a
+# rebuild after this change stops resolving pkg-config ncursesw/panelw
+# to the stale prefix.
+define Host/Uninstall
+	rm -f $(STAGING_DIR_HOST)/lib/pkgconfig/ncursesw.pc \
+		$(STAGING_DIR_HOST)/lib/pkgconfig/panelw.pc \
+		$(STAGING_DIR_HOST)/lib/pkgconfig/menuw.pc \
+		$(STAGING_DIR_HOST)/lib/pkgconfig/formw.pc
+endef
```
openwrt-ai, line 222: Thanks for adding the cleanup — but as written it won't fire on the upgrade path it targets, for two independent reasons.

**1. `Host/Uninstall` is not reached by a normal rebuild.** Its only caller is [`host-clean-build` at include/host-build.mk:232-233](https://github.com/openwrt/openwrt/blob/e1eb83251b3efc0a9d4a1ea44e804e4fed9706a9/include/host-build.mk#L232-L233), which runs on an explicit `host-clean`/`host-clean-build`, or via `clean-build` — and `clean-build` is auto-invoked only from [`rebuild_check` in include/subdir.mk:52-58](https://github.com/openwrt/openwrt/blob/e1eb83251b3efc0a9d4a1ea44e804e4fed9706a9/include/subdir.mk#L52-L58), which sits inside `ifdef CONFIG_AUTOREMOVE`. `AUTOREMOVE` is an off-by-default `DEVEL` option ([config/Config-devel.in:52-53](https://github.com/openwrt/openwrt/blob/e1eb83251b3efc0a9d4a1ea44e804e4fed9706a9/config/Config-devel.in#L52-L53)), so for the ordinary `git pull && make` this recipe never runs. The verification in the commit message (`make package/libs/ncurses/host/clean`) is precisely the explicit-clean path, so it doesn't cover the incremental case.

What *does* happen on `git pull && make` is a full host rebuild: [`HOST_STAMP_PREPARED` at host-build.mk:38](https://github.com/openwrt/openwrt/blob/e1eb83251b3efc0a9d4a1ea44e804e4fed9706a9/include/host-build.mk#L38) embeds an md5 of `${CURDIR}`, so editing this Makefile changes the stamp name, wipes `HOST_BUILD_DIR`, and re-runs [`$(HOST_STAMP_INSTALLED)` at host-build.mk:209-211](https://github.com/openwrt/openwrt/blob/e1eb83251b3efc0a9d4a1ea44e804e4fed9706a9/include/host-build.mk#L209-L211) — including `Hooks/HostInstall/Post`. That hook fires exactly when you need it.

**2. The stale filenames are wrong for most affected trees.** `HOST_CONFIGURE_ARGS` never passed `--enable-widec`; the `*w.pc` names only appeared with 18725c45a3, which merely bumped 6.4 → 6.6.20260801 and inherited upstream's new widec default. A tree that last host-built 6.4 has the narrow `ncurses.pc`/`panel.pc`/`menu.pc`/`form.pc` in `$(STAGING_DIR_HOST)/lib/pkgconfig`, and since `HOST_BUILD_DIR` is version-suffixed that 6.4 install was never overwritten — those are the files `pkg-config ncurses` still resolves to a 6.4 prefix. The `*w.pc` strays only exist on trees host-built in the few days since 18725c45a3. The PR description also states both `ncurses.pc` *and* `ncursesw.pc` are produced, so both spellings need removing either way.

```suggestion
# Trees that built host ncurses before --with-pkg-config-libdir was added
# have ncurses' *.pc files left over in $(STAGING_DIR_HOST)/lib/pkgconfig,
# which still shadow the correct copies in $(HOST_BUILD_PREFIX)/lib/pkgconfig
# since PKG_CONFIG_PATH searches STAGING_DIR_HOST first. Both spellings can
# be present: the narrow names from 6.4 and older, and the widec names from
# the 6.6 snapshots. Clean them up whenever host ncurses is (re)installed, so
# a plain rebuild after pulling this change stops resolving pkg-config
# ncurses/ncursesw to the stale prefix.
define Host/CleanStaleHostPkgConfig
	rm -f $(foreach l,ncurses panel menu form, \
		$(STAGING_DIR_HOST)/lib/pkgconfig/$(l).pc \
		$(STAGING_DIR_HOST)/lib/pkgconfig/$(l)w.pc)
endef

Hooks/HostInstall/Post := Host/CleanStaleHostPkgConfig $(Hooks/HostInstall/Post)
```

(The hook assignment has to stay above `$(eval $(call HostBuild))`, which it is. Keeping the `Host/Uninstall` definition in addition is harmless if you want the explicit-clean path covered too.)

---

openwrt-ai (commented):
## Commit checks

- `e1eb832` "ncurses: fix host pkg-config file install location" — the new final paragraph claims the added `Host/Uninstall` "runs on host-clean-build (and therefore host-clean), so a normal rebuild after pulling this change stops resolving pkg-config ncursesw/panelw to the stale prefix". The first half is right, the conclusion isn't: `host-clean-build` is not part of a normal rebuild — it is reached from an explicit `host-clean`, or from `clean-build`, which is only auto-invoked under the off-by-default `CONFIG_AUTOREMOVE`. The described verification (`make package/libs/ncurses/host/clean`) exercises the explicit-clean path only, so it doesn't test the claim. The same paragraph also names only `ncursesw.pc`/`panelw`/`menuw`/`formw` as the strays, but trees predating 18725c45a3 have the narrow `ncurses.pc`/`panel.pc`/`menu.pc`/`form.pc` instead. Both points detailed inline.

The `--with-pkg-config-libdir` change itself and the `PKG_RELEASE` bump are unchanged from the previous head and still look correct. CI on `e1eb832` is green.

---

dangowrt:
Both real, applied exactly as suggested — thank you for the precise line-cited mechanism, saved me from guessing.

- Added `Host/CleanStaleHostPkgConfig`, hooked via `Hooks/HostInstall/Post` so it fires on a genuinely normal rebuild (verified: editing this Makefile changes `HOST_STAMP_PREPARED`'s checksum, forcing a full host reinstall on the next plain `make`/`host/compile` — no explicit clean needed). Kept `Host/Uninstall` calling the same helper for the explicit `host-clean` path.
- Now cleans both spellings (`ncurses.pc`/`panel.pc`/`menu.pc`/`form.pc` *and* the `w` variants), covering trees last host-built on 6.4-or-earlier as well as the recent 6.6 snapshots.

Verified both paths directly: recreated all eight stale files, ran a plain `make package/libs/ncurses/host/compile` with no clean — all eight removed via the new hook. Recreated a subset and ran the explicit `host/clean` — also removed, `libdw.pc` untouched in both cases.

Corrected the commit message's "normal rebuild" claim too, since as written it conflated the two cleanup paths.

PR #24819, opened 2026-08-20
generic: backport the phylink and phylib PHY fixes, carry the DSA one

lexfrei:
Two backports and one local carry, all touching the same path: a PHY whose driver is a module on a rootfs that is not mounted yet when a DSA switch probes.

The phylink and phylib patches are [the series submitted upstream](https://lore.kernel.org/netdev/20260906174643.4107607-1-f@lex.la/). One clears a stale pl->phydev when bringup fails after recording it, so a later disconnect cannot detach the same PHY twice; the other gives back the interrupt that phy_probe() replaces with PHY_POLL for a driver without interrupt support, when that driver is removed, when its own probe fails and when the attach is unwound. pending-6.18 carries them until a mainline version comes back through a stable bump.

The third patch keeps a DSA port whose PHY has no driver yet and connects it at ifup. It was [submitted](https://lore.kernel.org/netdev/20260822155259.87146-1-f@lex.la/) and turned down: binding the MAC to the PHY from the open path is the wrong layer, and a phylink-hotplug direction is under consideration instead. It stays here because without it this board loses its copper port entirely, and it is dropped as soon as an upstream mechanism exists.

No board in the tree hits the underlying problem today: every firmware-loading PHY currently sits on a MAC. The Keenetic KN-1012 PR puts an Airoha EN8811H on an MT7531 switch port and needs this to have its copper combo port at all.

Verified on that hardware: the port that was dropped at 1.9 s comes up on the real driver at ifup and forwards, the PHY takes its interrupt instead of polling, boot time unchanged, other ports unaffected.

The phylib patch needed a modified kernel to check here, because the phylink patch prevents the very cycle it fixes: the poller refuses to attach a PHY whose driver is generic. With that refusal removed and no EN8811H driver in the image, the poller binds the generic driver and detaches it once a second. The interrupt reads PHY_POLL while the generic driver is bound and 15 after the detach; on a build without the phylib patch it reads PHY_POLL both times. Forty-eight cycles per boot on each.


jonasjelonek:
I think we had the same issue in the Realtek target, but solved it for now with a change in the DSA driver, see https://github.com/openwrt/openwrt/commit/ea3d646a08524b1cf4560ae351845a125a7d22c6. But I favor a generic upstream approach anyway.

Could you send this to netdev as an RFC soon? I'd like to see some opinions from upstream on this.

target/linux/generic/pending-6.18/704-net-dsa-connect-a-late-arriving-phy-at-ifup.patch
```diff
@@ -0,0 +1,176 @@
...
++	phy_fwnode = dsa_user_phy_fwnode(dp);
++	if (!phy_fwnode)
++		return false;
++
++	phydev = fwnode_phy_find_device(phy_fwnode);
++	fwnode_handle_put(phy_fwnode);
++	if (!phydev)
++		return false;
++
++	lacks_driver = !READ_ONCE(phydev->mdio.dev.driver);
```
openwrt-ai, line 162: "no driver bound right now" cannot distinguish "the module is still in the unmounted rootfs" from "this PHY will never have a driver of its own", which is what forces the description to accept that a port permanently served by genphy is now registered and fails every ifup instead of being dropped once.

The driver core has a discriminator for exactly this: [`driver_deferred_probe_check_state()` at dd.c:289-302](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/base/dd.c#L289-L302) returns `-EPROBE_DEFER` while a module may still show up and `-ENODEV`/`-ETIMEDOUT` once initcalls are done and the deferred-probe window has closed. The in-tree Realtek workaround @jonasjelonek pointed at combines the two checks the same way — [`rtldsa_phy_load_deferred()` at common.c:214-221](``https://github.com/openwrt/openwrt/blob/ea3d646a08524b1cf4560ae351845a125a7d22c6/target/linux/realtek/files-6.18/drivers/net/dsa/rtl83xx/common.c#L214-L221).`` Folding it in here would narrow the predicate to ports that actually have a driver pending, and would drop the rest at setup as before. Is there a reason it does not apply on this path?

Related, and it would go away with the same change: the read is not serialized against a concurrent bind. `phy_detach()` releases genphy via [`device_release_driver()` at phy_device.c:1879-1882](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/phy_device.c#L1879-L1882) inside the failed connect; if the real driver binds between that and this read, `lacks_driver` is false and the port is dropped for the whole uptime — the outcome the patch exists to prevent. Narrow, but reachable when both the switch and the PHY driver are modules loading together.

---

target/linux/generic/pending-6.18/704-net-dsa-connect-a-late-arriving-phy-at-ifup.patch
```diff
@@ -0,0 +1,176 @@
...
++	struct dsa_switch *ds = dp->ds;
++	u32 phy_flags = 0;
++	int err;
++
++	if (dev->phydev)
++		return 0;
++
++	phy_fwnode = dsa_user_phy_fwnode(dp);
++	if (!phy_fwnode)
++		return 0;
```
openwrt-ai, line 111: nit: the fact that `dsa_user_phy_setup()` deliberately deferred this port is not recorded anywhere, so it has to be re-inferred here from `!dev->phydev` plus "a PHY node exists". That is also true for ports that legitimately have no `phydev` and never will: an `MLO_AN_FIXED` port, or `managed = "in-band-status"` with an 802.3z phy-mode, where [phylink.c:2285-2289](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/phylink.c#L2285-L2289) returns 0 without connecting and a `phy-handle` may still be present. Those ports now do a `fwnode_get_phy_node()` lookup and a full `phylink_of_phy_connect()` call on every single ifup — harmless, because phylink early-returns again, but it means the condition guarding this path is implicit rather than stated.

A `bool` on `struct dsa_port` set in `dsa_user_phy_setup()` next to the `netdev_info()` would make this run only for the ports it was written for, and would let the helper say what it is testing.

---

openwrt-ai (commented):
Fresh review of the single commit at `4ff5343`. Commit subject and body match the diff, and the patch applies cleanly and without offsets on 6.18.44 (`LINUX_VERSION-6.18 = .44`, and no other in-tree patch touches `net/dsa/user.c`); `704-` is free in `pending-6.18`. Refcounting in both new helpers looks right — `fwnode_phy_find_device()`'s reference is released with the matching `put_device()`, and the failure paths of `phylink_fwnode_phy_connect()` never leave `dev->phydev` set, which matches the note on the defensive `user_dev->phydev = NULL`.

Two comments inline, both on the keep-or-drop predicate rather than on the connect-at-open mechanism itself.

---

target/linux/generic/pending-6.18/704-net-dsa-connect-a-late-arriving-phy-at-ifup.patch
lexfrei: That's the limitation the header already spells out. The test matches every genphy-served PHY, including ones that will never get a driver of their own, so a permanently misconfigured port registers and fails each ifup with a warning instead of vanishing at setup. Nothing at that point separates 'module still in the unmounted rootfs' from 'never', so I keep the driverless test and take the wider set.

target/linux/generic/pending-6.18/704-net-dsa-connect-a-late-arriving-phy-at-ifup.patch
lexfrei: The deferral isn't recorded, so it's re-inferred from !dev->phydev plus a resolvable phy node. An explicit flag would be cleaner but means carrying state on dsa_port. The re-inference is self-contained and the header calls out the same-shape case.

target/linux/generic/pending-6.18/708-net-phy-restore-the-interrupt-after-a-generic-driver.patch
```diff
@@ -0,0 +1,82 @@
...
++ *                    generic bind cycle, and non-positive values are
++ *                    never restored
+  * @phylink: Pointer to phylink instance for this PHY
+  * @sfp_bus_attached: Flag indicating whether the SFP bus has been attached
+  * @sfp_bus: SFP bus attached to this PHY's fiber port
+@@ -730,6 +734,8 @@ struct phy_device {
+ 	 */
+ 	int irq;
+ 
++	int genphy_saved_irq;
```
openwrt-ai: The registration-time interrupt is already retained outside `phydev`, so this new member may be storing something the bus already holds. `phy_device_create()` seeds the PHY from the bus table — [`dev->irq = bus->irq[addr];` at phy_device.c:831](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/phy_device.c#L831) — and the fwnode path writes the resolved interrupt *back* into that table before it touches `phy->irq`, in both branches: [fwnode_mdio.c:85-90](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/mdio/fwnode_mdio.c#L85-L90). Nothing between registration and the genphy bind cycle writes `bus->irq[addr]` again, so `phydev->mdio.bus->irq[phydev->mdio.addr]` is exactly the value `phy_probe()` parks over, and `phy_detach()` could restore from it without adding a field to `struct phy_device`.

The one divergence I can construct is a specific driver that mutated `phydev->irq` during an earlier attach and then released the PHY to genphy — but the values that would set (`PHY_MAC_INTERRUPT`, `PHY_POLL`) are all non-positive and this patch already refuses to restore those. Is there a case where the two differ that the extra member is needed for? On a struct this widely embedded, upstream will likely ask the same.

---

target/linux/generic/pending-6.18/704-net-dsa-connect-a-late-arriving-phy-at-ifup.patch
```diff
@@ -0,0 +1,210 @@
...
++		/* A driver that was not there before this attempt is one that
++		 * bound while it ran: the failure came from the generic driver
++		 * and says nothing about this one, so try once more. A port
++		 * that had its driver all along keeps the single attempt.
++		 */
++		if (!had_driver && !retried) {
++			had_driver = true;
++			retried = true;
++			goto connect;
++		}
```
openwrt-ai, line 210: `had_driver = true;` is dead: `had_driver` is not read anywhere after this block, and on the second pass `!had_driver` short-circuits before `retried` is even evaluated, so `retried` alone already closes the one-shot. Keeping both assignments makes the two flags look like they gate different things when they gate the same one.

```suggestion
+		if (!had_driver && !retried) {
+			retried = true;
+			goto connect;
+		}
```

---

target/linux/generic/pending-6.18/704-net-dsa-connect-a-late-arriving-phy-at-ifup.patch
```diff
@@ -0,0 +1,210 @@
...
++
++	return lacks_driver;
++}
++
+ static int dsa_user_phy_setup(struct net_device *user_dev)
+ {
+ 	struct dsa_port *dp = dsa_user_to_port(user_dev);
+ 	struct device_node *port_dn = dp->dn;
+ 	struct dsa_switch *ds = dp->ds;
++	bool had_driver, retried = false;
```
openwrt-ai, line 173: nit: series hygiene — this hunk (and the retry block at lines 202-206, plus the new description paragraph and the `Submitted upstream:` line) rewrites a patch file that `4ff5343` added earlier in the same series, so `4ff5343` leaves a version of `704-...patch` in the tree that you have already superseded. Folding these changes back into `4ff5343` would keep each commit standing on its own and would leave `839af36` doing only what its subject says — adding `708`/`709` and refreshing the three patches below them. As it stands the subject advertises "the phylib and phylink halves", while the diff also adds a new retry-once mechanism to the DSA patch that the body covers only as "bring the DSA patch in line with what was sent".

---

openwrt-ai (commented):
Re-reviewed at `839af36`; the new material since `4ff5343` is the retry-once addition to `704`, the two new patches `708`/`709`, and the three refreshes.

Patch hygiene checks out. Applying `backport-6.18/*`, `pending-6.18/*` and `hack-6.18/*` in series order onto a pristine 6.18.44 tree (`LINUX_VERSION-6.18 = .44`) gives zero fuzz and zero offsets with `-F0`, and the refreshed hunk headers are arithmetically consistent with what the new patches insert: `709` adds 12 lines to `phylink.c` above every shifted hunk in `737-01`/`737-02`, and `708` adds 13 to `phy_device.c` and 6 to `phy.h` above the shifted hunks in `896-03`. `737-03`, `737-05` and `737-07` all sit below the `709` insertion point, so leaving them untouched is correct. `708` and `709` are free numbers in `pending-6.18`, and no target-specific `patches-6.18` touches these files.

`709` reads correct: the three cleared fields and the `phydev_mutex` → `phy->lock` → `state_mutex` order match `phylink_disconnect_phy()`; all three `phylink_bringup_phy()` callers already `phy_detach()` on failure, so `phy->phylink`/`phy_link_change` are unwound too; and `phylink_resolve()` holds `pl->phydev_mutex` across its whole body, so the `flush_work(&pl->resolve)` that `phylink_disconnect_phy()` does is not needed here to keep a concurrent resolve off the PHY the caller is about to detach. The early validation return is above the `pl->phydev = phy` assignment, so it correctly does not take the new path.

`708`'s premise holds — `fwnode_mdiobus_phy_device_register()` resolves the interrupt once at bus registration and nothing re-reads the node afterwards, so a genphy bind cycle is a one-way trip to `PHY_POLL`. `phy_remove()` does not touch `phydev->irq`, so restoring ahead of `device_release_driver()` is safe, and the `> 0` guard correctly skips `PHY_POLL`/`PHY_MAC_INTERRUPT`. One question inline on where the saved value is kept.

Three comments inline, one of them a nit on series hygiene.

---

lexfrei:
Folded the `704` rework back into the commit that adds the file, so each commit stands on its own again and the second one now does only what its subject says.

`had_driver = true;` is genuinely dead, thanks. It goes with the next push, together with the same removal upstream, since the sent version carries it too. That one needs a `make target/linux/refresh` run to regenerate the hunk header rather than a hand edit, and the machine that has the build tree is offline for a couple of days.

On `genphy_saved_irq` versus `bus->irq[addr]`: Andrew Lunn asked for exactly the same thing on the list, so that is two independent votes and the bus table wins. v2 upstream will restore from it and drop the new member; this backport follows once that is sent.

openwrt-ai (commented):

Both SHAs are new (`4ff5343`/`839af36` → `ac1277b`/`dc16cba`), but the tree at head differs from `839af36` by only two hunk headers — the `801-01` and `720` refreshes. The rest is the commit-boundary rework, which does what the earlier nit asked: the retry-once block and the description paragraphs now live in `ac1277b`, the commit that adds `704`, so that commit stands on its own, and `dc16cba` is left doing exactly what its subject says. Its body was corrected to name the refreshed set accurately ("phylink, PSE, airoha and ramips" rather than "the three phylink and PSE patches"). Both messages match their diffs.

The two new refreshes are correct, checked both ways. Arithmetically: `708` inserts 1+3+9 = 13 lines into `phy_device.c` above line 2030, so airoha `801-01`'s `@@ -2030` → `@@ -2043` is exact; `708`'s first `phy.h` hunk adds 4 lines at 566 and its second sits at 730, below the shifted site, so ramips `720`'s `@@ -659` → `@@ -663` is exact too. Empirically: `backport-6.18/*`, `pending-6.18/*`, `hack-6.18/*` followed by `airoha/patches-6.18/*` (and, on a second tree, `ramips/patches-6.18/*`) apply to a pristine 6.18.44 (`LINUX_VERSION-6.18 = .44`) with `-F0` and zero offsets. CI's "Check Kernel patches" agrees across every target it ran.

Nothing else in the tree needs refreshing. `layerscape/patches-6.18/702-...` is the only other target patch touching these three files, and its hunks sit at `phylink.c:274` and `phy.h:136`/`220` — all above `709`'s insertion at `phylink.c:2188` and `708`'s at `phy.h:566` — so leaving it alone is right.

The two findings still open from the last round are both acknowledged in [your comment](https://github.com/openwrt/openwrt/pull/24819#issuecomment-5386043278) — the dead `had_driver = true;` waiting on a `make target/linux/refresh` run, and `genphy_saved_irq` giving way to `bus->irq[addr]` after Andrew Lunn asked for the same upstream — so I am not restating either here.

One thing I checked while I was in the file and did not find a problem with: the late-connect path is symmetric across the netdev lifetime. `dsa_user_close()` does not disconnect, so a second `ifup` early-returns on `dev->phydev` rather than re-running the connect, and teardown is still covered because [`dsa_user_destroy()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/net/dsa/user.c#L2601-L2616) calls `phylink_disconnect_phy()` unconditionally, which is a no-op when `pl->phydev` is NULL and does the right thing when the PHY arrived at open instead of at setup. The `ds->user_mii_bus` fallback also cannot reach the new retry: it only runs when there is no `phy-handle` to resolve, which is exactly the case where `dsa_user_phy_fwnode()` returns NULL and `had_driver` is therefore already true.

---

target/linux/airoha/patches-6.18/801-01-net-phy-add-PHY_DETACH_NO_HW_RESET-PHY-flag.patch
```diff
@@ -107,7 +107,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  module_phy_driver(as21xxx_drivers);
 --- a/drivers/net/phy/phy_device.c
 +++ b/drivers/net/phy/phy_device.c
-@@ -2030,7 +2030,8 @@ void phy_detach(struct phy_device *phyde
+@@ -2043,7 +2043,8 @@ void phy_detach(struct phy_device *phyde
```
openwrt-ai: This refreshed hunk header is stale after the 708 rewrite in `6ce7ad7`. The new `708` inserts 14 lines into `phy_device.c` above `phy_detach()` (the 12-line `phy_restore_genphy_irq` helper at 1611, plus one line each in `phy_attach_direct`'s `error_module_put` path and in `phy_detach`), where the old `708` inserted 13. `6ce7ad7` re-normalized `896-03` for that `phy_device.c` shift (its `+1`), but this patch — which also sits below those insertions — was not refreshed, so it now applies with a one-line offset.

Applying the full generic 6.18 series onto a pristine 6.18.44 tree and then this patch:

```
checking file drivers/net/phy/phy_device.c
Hunk #1 succeeded at 2044 (offset 1 line).
```

CI's "Check Kernel patches" for the airoha target will show the same once it runs. Regenerate with `make target/linux/refresh` with an airoha target selected in `.config` (the project's patch-refresh flow, not a hand edit); the header should read `@@ -2044,7 +2044,8 @@ void phy_detach`.

---

target/linux/ramips/patches-6.18/720-NET-no-auto-carrier-off-support.patch
```diff
@@ -21,7 +21,7 @@ Signed-off-by: John Crispin <blogic@openwrt.org>
  	if (phydev->mii_ts && phydev->mii_ts->link_state)
 --- a/include/linux/phy.h
 +++ b/include/linux/phy.h
-@@ -659,6 +659,7 @@ struct phy_device {
+@@ -663,6 +663,7 @@ struct phy_device {
```
openwrt-ai: Same missed refresh as airoha `801-01`, in the other direction. The old `708` added a 4-line kernel-doc block to `include/linux/phy.h` (for the now-removed `genphy_saved_irq` member) above this `struct phy_device` hunk, and the previous round refreshed this header from `@@ -659` to `@@ -663` to account for it. The new `708` no longer touches `phy.h` at all, so those 4 lines are gone but this header was not moved back — it now applies with a negative offset.

Applying the full generic 6.18 series onto a pristine 6.18.44 tree and then this patch:

```
checking file include/linux/phy.h
Hunk #1 succeeded at 659 (offset -4 lines).
```

Regenerate with `make target/linux/refresh` with a ramips target selected; the header should read `@@ -659,6 +659,7 @@ struct phy_device {`. (The `phy_device.c` hunk at `@@ -268` is above every change and stays put.)

---

openwrt-ai (commented):
Re-reviewed at `6ce7ad7`. New since the last review (`dc16cba`, which the branch was rebased over) is the dead-`had_driver` removal in `704` and the phy-irq resync: `708`/`709` rewritten to the sent v2, with `737-01`, `737-02` and `896-03` refreshed for the new line counts.

Both findings open last round are resolved: the dead `had_driver = true;` is gone from the `704` retry, and `708` now restores the interrupt from `mdio.bus->irq[addr]` with the `genphy_saved_irq` struct member dropped — the form Andrew Lunn asked for on the list. The two new helpers read correctly: `phy_restore_genphy_irq()` reads back an always-present inline `bus->irq[]` slot and only undoes a `PHY_POLL` parking, and `phylink_clear_phydev()` is a behaviour-preserving dedup of the clear that `phylink_disconnect_phy()` already did.

One issue on the resync itself: `6ce7ad7` refreshed `737-01`, `737-02` and `896-03` for `708`'s new `phy_device.c`/`phy.h` line counts, but two target patches that also sit below `708`'s insertions were missed — airoha `801-01` (applies at +1) and ramips `720` (applies at -4). Both verified by applying the full generic 6.18 series onto a pristine 6.18.44 tree; details and the refresh command inline. layerscape `702` is unaffected — its `phylink.c` and `phy.h` hunks are all above the changed insertion points. CI's "Check Kernel patches" for airoha and ramips was still queued at review time and should surface the same two.

---

openwrt-ai (commented):
Re-reviewed at `0238282`. The branch was amended over `6ce7ad7`, and the only tree change is the two hunk headers flagged last round — no code moved.

Both are now correct, checked empirically. Applying `backport-6.18/*`, `pending-6.18/*` and `hack-6.18/*` in series order onto a pristine 6.18.44 tree (`target/linux/generic/kernel-6.18` gives `LINUX_VERSION-6.18 = .44`) is zero-fuzz and zero-offset with `-F0`, and airoha `801-01` (`@@ -2043` → `@@ -2044`) and ramips `720` (`@@ -663` → `@@ -659`) then both apply on top with no offset reported. CI's "Check Kernel patches" agrees: the airoha/an7581 and ramips/mt7620 jobs are green at this head, where they were still queued when I raised the two.

Nothing else in the tree needs the same treatment. `layerscape/patches-6.18/702-phy-Add-2.5G-SGMII-interface-mode.patch` is the only other target patch touching `phylink.c` or `phy.h`, and it still applies at zero offset after the full generic series — its hunks sit above both `708`'s and `709`'s insertion points.

## Commit checks
- `0238282` "generic: resync the phy-irq series with the sent v2" — the rewritten body documents only the airoha and ramips refreshes, but the commit also renumbers hunk headers in `737-01`, `737-02` and `896-03`. The body at `6ce7ad7` named exactly those three; the rewrite swapped which subset it covers rather than covering both, so three of the eight touched files now go unmentioned.

---

target/linux/generic/pending-6.18/708-net-phy-restore-the-interrupt-after-a-generic-driver.patch
```diff
@@ -0,0 +1,67 @@
...
+mdiobus_alloc() fills bus->irq[] with PHY_POLL for every address, and
+the bind cycle never writes to that table, so the entry still holds
+whatever the bus registered there. Restore phydev->irq from it on both
+exits, and only where the cycle left PHY_POLL. That guard preserves an
+interrupt mode a MAC installed on the attached PHY after connect, and
+it keeps a restored interrupt number out of the
+phy_connect_direct()/phy_disconnect() asymmetry, where such a MAC would
+have the interrupt requested and never freed.
+
+Submitted upstream: https://lore.kernel.org/netdev/20260827211638.63395-1-f@lex.la/
```
openwrt-ai: nit: both patches now point at the v3 thread (`20260827211638.63395-1`, same line in `709` at line 26), but the PR description still links the v2 posting — "The phylink and phylib patches are [the series submitted upstream](https://lore.kernel.org/netdev/20260824024029.41310-1-f@lex.la/)". Worth repointing the body at the same thread so a reader landing on the PR and a reader landing on the patch file end up in the same place.

(The DSA link in the body, `20260822155259.87146-1`, is a different series and correctly still matches `704`.)

---

openwrt-ai (commented):
Re-reviewed at `27261e0`. Everything new since `0238282` is inside the two pending patch files — no kernel hunk moved and no inserted line count changed — so nothing downstream needed a refresh this round, and the changed-file list agrees: `737-01`, `737-02`, `896-03` and ramips `720` are untouched here, and airoha `801-01`'s single header (`@@ -2044`) is carried over unchanged from the earlier commit.

Checked empirically on a pristine 6.18.44 tree (`target/linux/generic/kernel-6.18` gives `LINUX_VERSION-6.18 = .44`): `backport-6.18/*`, `pending-6.18/*` and `hack-6.18/*` in series order apply with `-F0` at zero fuzz and zero offset, and airoha, ramips and layerscape `patches-6.18/*` each then apply on top with no offset reported. `708`'s comment reword is exactly line-neutral — six lines before and after — which is why `@@ -1611,6 +1611,18 @@` still holds without a resync.

The rewritten `708` rationale is more accurate than the v2 wording, and both of its new load-bearing claims hold against 6.18.44. The parking really is inside probe, not in the attach path around it — [`phydev->irq = PHY_POLL` in `phy_probe()` at phy_device.c:3407](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/phy_device.c#L3407) — which is what makes the restore on the `error_module_put` exit meaningful rather than a no-op: `phy_attach_direct()` calls `d->driver->probe(d)` and only jumps there after it, and that label is reachable only from inside the `if (phydev->is_genphy_driven)` block, so the unguarded call there is still confined to the bind cycle. And [`bus->irq[i] = PHY_POLL` in `mdiobus_alloc_size()` at mdio_bus_provider.c:63](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/mdio_bus_provider.c#L63) confirms the new sentence about the table's default, so a bus that never filled it restores `PHY_POLL` onto `PHY_POLL` and the guard degrades to a no-op rather than to a bogus IRQ number.

`709` is metadata-only this round and reads the same as it did at `6ce7ad7`.

## Commit checks
- `27261e0` "generic: resync the phy-irq series with the sent v3" — "the phylib half reworded a code comment" understates what the commit does to `708`: besides the comment, its entire description is rewritten (the `mdiobus_alloc()` paragraph and the `phy_connect_direct()`/`phy_disconnect()` asymmetry rationale are both new) and it gains a `Fixes: 00db8189d984 (...)` tag it did not carry before. `709`'s half of the sentence is exact. The generic "Regenerate both backports from the new upstream commits" covers it loosely, but the sentence that spells out the per-half delta is the one a reader will trust, and a new `Fixes:` tag is the kind of thing worth naming.

---

lexfrei:
@jonasjelonek Sent it. Lunn dropped the DSA patch; phylink and phylib went to net as [v3](https://lore.kernel.org/netdev/20260827211638.63395-1-f@lex.la/). Late-PHY moved to an [architecture thread](https://lore.kernel.org/netdev/a230d199-5d4d-4637-aff3-e725a37e1da1@lunn.ch/), waiting on his answer there.

target/linux/generic/pending-6.18/704-net-dsa-connect-a-late-arriving-phy-at-ifup.patch
```diff
@@ -0,0 +1,214 @@
...
+driver present now, so the connect is retried once, and only for a PHY that
+had no driver when the attempt started. A port whose PHY was bound all along
+keeps its single attempt.
+
+Submitted upstream: https://lore.kernel.org/netdev/20260822155259.87146-1-f@lex.la/
+
+The maintainer turned the approach down: binding the MAC to the PHY
+from the open path is the wrong layer, and a phylink-hotplug
+direction is under consideration instead. Kept here as an OpenWrt
+carry until an upstream direction exists.
```
openwrt-ai, line 72: nit: the patch header and the commit message now disagree about the status of this patch. The header says the approach was turned down and that this is an OpenWrt carry until an upstream mechanism exists; `219fd4a`'s message says only `Submitted upstream: https://lore.kernel.org/netdev/20260822155259.87146-1-f@lex.la/`, which reads as "in flight". The PR body has the accurate version too, but the PR body is not what survives the merge — someone running `git log` on `pending-6.18/704-*` in six months gets the misleading half.

Since `219fd4a` is the commit that adds the file, the one-line "kept as a carry, dropped when a phylink-hotplug mechanism lands" fact belongs in its body as well.

---

openwrt-ai (commented):
Re-reviewed at `3519da7`. Since `27261e0` the branch was rebased onto `1ad2e78` and the four follow-up commits (`efbc075`, `0238282`, `27261e0`, plus the `ff7aac7` add) were folded back into the two they belonged to, so all four SHAs are new. The tree itself is not: `git diff 27261e0..3519da7` restricted to the seven files this PR touches is empty, so no patch text, hunk header or kernel hunk changed this round. The rebase is also inert for this series — the only 6.18-relevant commits between the two bases are the 6.12 bumps `bbddea7` and `a865bd9`.

Re-verified the series anyway, because the base moved. On a pristine 6.18.44 tree (`target/linux/generic/kernel-6.18` gives `LINUX_VERSION-6.18 = .44`), `backport-6.18/*`, `pending-6.18/*` and `hack-6.18/*` in series order apply with `-F0` at zero fuzz and zero offset, and airoha, ramips and layerscape `patches-6.18/*` each then apply on top with no offset reported. Those three are the complete set of target patches touching `phylink.c`, `phy_device.c` or `phy.h`, so nothing else needs a refresh. CI at this head agrees — both workflow runs (`Build Kernel`, `Build all core packages`) concluded green, "Check Kernel patches" included.

The one nit still open from last round is closed: the PR body now links the v3 thread (`20260827211638.63395-1`), matching what `708` and `709` carry. Everything else I flagged earlier is either resolved in the tree or answered in the thread, so I am not restating any of it. One new nit inline, on the `704` header.

## Commit checks
- `3519da7` "generic: add the phylib and phylink halves of the late-PHY fix" — "The phylink, PSE, airoha and ramips patches that apply later are refreshed" overstates the diff by one target: no ramips file is touched by this commit or anywhere in the PR. `ramips/patches-6.18/720-NET-no-auto-carrier-off-support.patch` was refreshed while `708` still added a kernel-doc block to `phy.h` for the old `genphy_saved_irq` member; the v2 rewrite dropped that member, the net shift went back to zero, and the refresh was correctly reverted — but the sentence was carried over from the pre-v2 wording when the follow-ups were squashed in. The commit refreshes phylink (`737-01`, `737-02`), PSE (`896-03`) and airoha (`801-01`); dropping "and ramips" makes it exact.

---

openwrt-ai (commented):
## Commit checks
- `1863aa8` "generic: add the phylib and phylink halves of the late-PHY fix" — still carries "The phylink, PSE, airoha and ramips patches that apply later are refreshed" while touching no ramips file; dropping "and ramips" makes it exact. (Same message as `3519da7`; re-flagging only because that SHA is gone from the branch.)

---

target/linux/generic/pending-6.18/704-net-dsa-connect-a-late-arriving-phy-at-ifup.patch
openwrt-ai: fixed, thanks

---

target/linux/generic/pending-6.18/708-net-phy-restore-the-interrupt-after-a-generic-driver.patch
openwrt-ai: fixed, thanks

---

target/linux/airoha/patches-6.18/801-01-net-phy-add-PHY_DETACH_NO_HW_RESET-PHY-flag.patch
openwrt-ai: fixed, thanks

---

target/linux/ramips/patches-6.18/720-NET-no-auto-carrier-off-support.patch
openwrt-ai: fixed, thanks

---

target/linux/generic/pending-6.18/704-net-dsa-connect-a-late-arriving-phy-at-ifup.patch
openwrt-ai: fixed, thanks

---

target/linux/generic/pending-6.18/704-net-dsa-connect-a-late-arriving-phy-at-ifup.patch
openwrt-ai: fixed, thanks

---

target/linux/generic/pending-6.18/708-net-phy-restore-the-interrupt-after-a-generic-driver.patch
openwrt-ai: solved by dropping the struct member and restoring from `mdio.bus->irq[addr]`, thanks

---

openwrt-ai (commented):

---

target/linux/generic/pending-6.18/896-03-net-phy-own-phydev-psec-via-PSE-notifier-and-remove-.patch
```diff
@@ -344,7 +344,7 @@ Tested-by: Jonas Jelonek <jelonek.jonas@gmail.com>
  	device_del(&phydev->mdio.dev);
  
  	/* Assert the reset signal */
-@@ -3701,8 +3846,14 @@ static int __init phy_init(void)
+@@ -3715,8 +3860,14 @@ static int __init phy_init(void)
```
openwrt-ai: Stale after the `708` comment reword in `a0792ed`: the helper block shrank from 12 lines to 11, so `708` now inserts 13 lines into `phy_device.c` instead of 14 and both of this patch's `phy_init`/`phy_exit` headers sit one line too low. Applying `backport-6.18/*`, `pending-6.18/*`, `hack-6.18/*` in series order onto a pristine 6.18.44 tree:

```
checking file drivers/net/phy/phy_device.c
Hunk #5 succeeded at 3859 (offset -1 lines).
Hunk #6 succeeded at 3880 (offset -1 lines).
```

Regenerate with `make target/linux/refresh`; line 362 needs the same `-1` (`@@ -3729,6 +3880,7 @@ err_ethtool_phy_ops:`).

```suggestion
@@ -3714,8 +3859,14 @@ static int __init phy_init(void)
```

---

target/linux/airoha/patches-6.18/801-01-net-phy-add-PHY_DETACH_NO_HW_RESET-PHY-flag.patch
```diff
@@ -107,7 +107,7 @@ Signed-off-by: Christian Marangi <ansuelsmth@gmail.com>
  module_phy_driver(as21xxx_drivers);
 --- a/drivers/net/phy/phy_device.c
 +++ b/drivers/net/phy/phy_device.c
-@@ -2030,7 +2030,8 @@ void phy_detach(struct phy_device *phyde
+@@ -2044,7 +2044,8 @@ void phy_detach(struct phy_device *phyde
```
openwrt-ai: Same one-line drift as `896-03`: `708`'s reworded comment gives back a line above `phy_detach()`, so this header is now one too high. Applying the full generic 6.18 series onto a pristine 6.18.44 tree and then this patch:

```
checking file drivers/net/phy/phy_device.c
Hunk #1 succeeded at 2043 (offset -1 lines).
```

Regenerate with `make target/linux/refresh` with an airoha target selected. ramips `720` and layerscape `702` are unaffected — I re-checked both at zero offset.

```suggestion
@@ -2043,7 +2043,8 @@ void phy_detach(struct phy_device *phyde
```

---

target/linux/airoha/patches-6.18/801-01-net-phy-add-PHY_DETACH_NO_HW_RESET-PHY-flag.patch
openwrt-ai: This file is gone from `main`: [`2700e9b277` "generic: 6.18: move AS21xxx patches from airoha"](https://github.com/openwrt/openwrt/commit/2700e9b277b18b85c946939dc2f3b33f2434cb1a) moved it to `target/linux/generic/pending-6.18/801-01-net-phy-add-PHY_DETACH_NO_HW_RESET-PHY-flag.patch`, so this edit is a modify/delete conflict — which is what GitHub is reporting as the PR's conflicting state.

After rebasing, the refresh belongs on the generic copy instead, and the number there is different because the patch now applies before the `896-*` series: its `phy_detach` header needs `@@ -1885` → `@@ -1898`. Verified by applying the rebased generic 6.18 series onto a pristine 6.18.44 tree — `Hunk #1 succeeded at 1898 (offset 13 lines)` — after which the whole series, plus ramips and layerscape `patches-6.18`, is zero-fuzz and zero-offset.

---

target/linux/generic/pending-6.18/896-03-net-phy-own-phydev-psec-via-PSE-notifier-and-remove-.patch
```diff
@@ -344,7 +344,7 @@ Tested-by: Jonas Jelonek <jelonek.jonas@gmail.com>
  	device_del(&phydev->mdio.dev);
  
  	/* Assert the reset signal */
-@@ -3701,8 +3846,14 @@ static int __init phy_init(void)
+@@ -3714,8 +3859,14 @@ static int __init phy_init(void)
```
openwrt-ai: Same stale base as airoha `801-01`: `2700e9b277` also refreshed this file, so on current `main` these two hunks sit at `@@ -3702`/`@@ -3717`, and adding `708`'s 13 lines puts them at `@@ -3715` and `@@ -3730` rather than the `-3714`/`-3729` computed here against the old base. Line 362 needs the matching `@@ -3730,6 +3881,7 @@ err_ethtool_phy_ops:`.

```suggestion
@@ -3715,8 +3860,14 @@ static int __init phy_init(void)
```

---

target/linux/generic/pending-6.18/896-03-net-phy-own-phydev-psec-via-PSE-notifier-and-remove-.patch
openwrt-ai: fixed, thanks

---

target/linux/airoha/patches-6.18/801-01-net-phy-add-PHY_DETACH_NO_HW_RESET-PHY-flag.patch
openwrt-ai: fixed, thanks

---

openwrt-ai (commented):
## Commit checks
- `f48534d` and `044e4a1` — the reworked trailer `Assisted-by: LLM` names no tool. Every `Assisted-by:` trailer on `main` identifies the model (`Assisted-by: Claude:claude-opus-5`, `Assisted-by: ChatGPT (OpenAI GPT-5.6 Sol)`); use a concrete name in that form.

---

target/linux/generic/pending-6.18/707-net-dsa-connect-a-late-arriving-phy-at-ifup.patch
openwrt-ai, line 1: nit: 707 is already taken in this directory by [`707-net-pcs-pcs-mtk-lynxi-fix-bpi-r3-serdes-configuratio.patch`](``https://github.com/openwrt/openwrt/blob/76fa06340471263fe24a82969d0f292fbbcc82d4/target/linux/generic/pending-6.18/707-net-pcs-pcs-mtk-lynxi-fix-bpi-r3-serdes-configuratio.patch),`` which predates this branch, so the rename off 704 landed on an occupied number and the two only order deterministically by the rest of the filename. 714 is free in `pending-6.18` and keeps the patch near 708/709.

---

openwrt-ai (commented):
## Commit checks
- `76fa063` "generic: add the phylib and phylink halves of the late-PHY fix" — "The phylink, PSE and airoha patches that apply later are refreshed" is wrong at both ends now. No airoha file is touched: [`2700e9b277`](https://github.com/openwrt/openwrt/commit/2700e9b277b18b85c946939dc2f3b33f2434cb1a) moved `801-01` to `generic/pending-6.18/`. And the commit also refreshes `generic/hack-6.18/790-net-phy-skip-MMD-30-for-RTL8221B-...` and `ramips/patches-6.18/720-NET-no-auto-carrier-off-support.patch`, neither of which the sentence covers.
- `991160c`, `76fa063` — `Assisted-by: LLM` still names no tool; every such trailer on `main` identifies the model (`Assisted-by: Claude:claude-opus-5`). Re-flagged only because the SHAs it was raised on are gone from the branch.

---

target/linux/airoha/patches-6.18/801-01-net-phy-add-PHY_DETACH_NO_HW_RESET-PHY-flag.patch
openwrt-ai: fixed, thanks — the refresh moved to the generic copy and the airoha file is untouched

---

target/linux/generic/pending-6.18/896-03-net-phy-own-phydev-psec-via-PSE-notifier-and-remove-.patch
openwrt-ai: fixed, thanks

---

target/linux/generic/pending-6.18/708-net-phy-take-the-interrupt-back-from-the-bus-on-detach.patch
```diff
@@ -0,0 +1,46 @@
...
+--- a/drivers/net/phy/phy_device.c
++++ b/drivers/net/phy/phy_device.c
+@@ -1679,6 +1679,11 @@ static void phy_detach_internal(struct p
+ 		phydev->is_genphy_driven = 0;
+ 	}
+ 
++	/* Whatever this attachment did to the interrupt, the bus that
++	 * described it still knows the number. Take it back from there.
++	 */
++	phydev->irq = phydev->mdio.bus->irq[phydev->mdio.addr];
```
openwrt-ai, line 42: For a bus whose driver never fills `bus->irq[]` this is not an inert restore but an overwrite with `PHY_POLL`, so "there is nothing to take back" holds for the table's value, not for `phydev->irq`. sxgbe is a concrete loser: it sets `phydev->irq` once at bus registration ([sxgbe_mdio.c:245-248](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/ethernet/samsung/sxgbe/sxgbe_mdio.c#L245-L248)) and connects/disconnects per ifup/ifdown ([`phy_connect()` at sxgbe_main.c:288](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/ethernet/samsung/sxgbe/sxgbe_main.c#L288), [`phy_disconnect()` at sxgbe_main.c:1219](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/ethernet/samsung/sxgbe/sxgbe_main.c#L1219)), so from the second ifup on it polls where it used to interrupt. Can the write be guarded on the table holding a valid number, or is the loss deliberate?

---

target/linux/generic/pending-6.18/708-net-phy-take-the-interrupt-back-from-the-bus-on-detach.patch
```diff
@@ -0,0 +1,46 @@
...
+whether to request an interrupt, and it holds no lock against the driver
+core, so a restore driven by an unbind would be racing that decision
+rather than ordered against it.
+
+A bus whose driver writes only phydev->irq and never the table is not
+covered, because the table then holds PHY_POLL and there is nothing to
+take back; lan78xx, smsc95xx and sxgbe are in that position today and
+registering the interrupt with the bus is theirs to do.
+
+Submitted upstream: https://lore.kernel.org/netdev/20260909204306.2374562-3-f@lex.la/
```
openwrt-ai, line 28: nit: the PR body still describes the superseded design — "when that driver is removed, when its own probe fails and when the attach is unwound" — and links `20260906174643.4107607-1`, while the patch now restores only in `phy_detach()` and cites `20260909204306.2374562`. Repoint the body at the current thread and drop the three restore sites.

---

openwrt-ai (commented):
## Commit checks
- `5673d14` "generic: add the phylib and phylink halves of the late-PHY fix" — "The phylink, PSE and airoha patches that apply later are refreshed": no airoha file is touched, [`2700e9b277`](https://github.com/openwrt/openwrt/commit/2700e9b277b18b85c946939dc2f3b33f2434cb1a) moved `801-01` to `generic/pending-6.18/`. Naming it generic (or by patch number) makes the sentence exact; the phylink and PSE halves are right.

---

openwrt-ai (commented):

---

target/linux/generic/pending-6.18/744-02-net-phylink-wait-for-PHYs-that-are-known-to-probe-late.patch
```diff
@@ -0,0 +1,379 @@
...
++	INIT_DELAYED_WORK(&pl->late_phy_poll, phylink_late_phy_poll);
+ 	INIT_LIST_HEAD(&pl->pcs_list);
+ 
+ 	pl->config = config;
+@@ -2276,6 +2299,8 @@ void phylink_destroy(struct phylink *pl)
+ 
+ 	mutex_lock(&pl->state_mutex);
+ 
++	phylink_late_phy_cancel(pl);
++
```
openwrt-ai: `phylink_late_phy_cancel()` does `cancel_delayed_work_sync()`, and it runs here while `pl->state_mutex` is held, but `phylink_late_phy_poll()` takes that same mutex — for the `link_interface == NA` fixup and again around `force_major_config`, plus inside `phylink_bringup_phy()`. A poll run already past `rtnl_trylock()` when `phylink_destroy()` is entered will block on `state_mutex` while the cancel waits for it. Moving the cancel above the `mutex_lock()` keeps the hunk line counts unchanged.

```suggestion
 
+	phylink_late_phy_cancel(pl);
+
 	mutex_lock(&pl->state_mutex);
 
```

---

target/linux/generic/pending-6.18/744-02-net-phylink-wait-for-PHYs-that-are-known-to-probe-late.patch
```diff
@@ -0,0 +1,379 @@
...
++			phy_device_free(phy_dev);
++
++		pl->late_phy_fwnode = phy_fwnode;
++		pl->late_phy_flags = flags;
++		pl->late_phy_poll_ms = PHYLINK_LATE_PHY_POLL_MS;
++		pl->late_phy_waited_ms = 0;
++		pl->late_phy_warned = false;
++		queue_delayed_work(system_freezable_power_efficient_wq,
++				   &pl->late_phy_poll, 0);
++		return 0;
```
openwrt-ai, line 309: This returns above the `host_interfaces` block that the in-tree [`706-net-phy-populate-host_interfaces-when-attaching-PHY.patch`](``https://github.com/openwrt/openwrt/blob/6886f6809e687dc1d8f5286c42f998105c3b78dd/target/linux/generic/pending-6.18/706-net-phy-populate-host_interfaces-when-attaching-PHY.patch#L36-L54)`` adds to this function, and `phylink_attach_phy()` does not carry it, so a PHY connected by the poller keeps an empty `host_interfaces` where the same PHY connected at setup would get the MAC's single-lane SerDes modes. That is exactly the 2.5GBASE-T case 706 was written for, which is also the EN8811H this series targets. Should the poller replay that block before attaching, or is the loss intended?

---

target/linux/generic/pending-6.18/744-02-net-phylink-wait-for-PHYs-that-are-known-to-probe-late.patch
```diff
@@ -0,0 +1,379 @@
+From: Aleksei Sviridkin <f@lex.la>
+Subject: [PATCH] net: phylink: wait for PHYs that are known to probe late
```
openwrt-ai, line 2: nit: the PR title and body still describe the dropped DSA design — "carry the DSA one", "keeps a DSA port whose PHY has no driver yet and connects it at ifup", "binding the MAC to the PHY from the open path is the wrong layer" — and link `20260822155259.87146-1`. The third patch is now this phylink poller plus the `needs-host-firmware` binding; repoint both at `20260908155729.4164814-1`.

---

target/linux/generic/pending-6.18/707-net-dsa-connect-a-late-arriving-phy-at-ifup.patch
openwrt-ai: solved by dropping the DSA patch entirely, thanks

---

openwrt-ai (commented):

---

target/linux/generic/pending-6.18/744-02-net-phylink-wait-for-PHYs-that-are-known-to-probe-late.patch
openwrt-ai: fixed, thanks — the cancel now sits above both `cancel_work_sync(&pl->resolve)` and the `state_mutex` section

---

openwrt-ai (commented):

---

PR #23727, opened 2026-06-09
mediatek: filogic: add Keenetic Buddy 6 / Buddy 6 SE and Netcraze rebrands

lexfrei:
Adds support for two Keenetic Buddy 6 variants on MT7981B and their Netcraze rebrands: the Buddy 6 (KN-3411 / NC-3411) and the Buddy 6 SE (KN-4410 / NC-4410). All four are dual-band Wi-Fi 6 mesh extenders with a single 1GbE port and 128 MiB SPI-NAND. All four are thin wrappers around one shared `.dtsi` and one image recipe, hence a single PR: the actual review surface is one board plus three compatibles and image ids.

A fresh take on the stalled #15551, rebased on current main.

Tested on hardware, Buddy 6 (KN-3411), on two units, one with each NAND chip this model ships:

- TFTP-booted the initramfs from the Keenetic U-Boot console.
- Installed via sysupgrade; boots from NAND.
- Verified NAND/partition layout, ethernet, both radios, all LEDs (toggled to confirm colours), the reset button and the factory MAC.

Netcraze NC-3411 is the same hardware sold under the Netcraze brand, so it shares the device tree and differs only in model, compatible and the factory image id. The device tree is split into a shared .dtsi with per-brand files, following the in-tree KN-1812/NC-1812 pattern. I don't own an NC-3411 unit, so that variant is untested on hardware.

The Buddy 6 SE (KN-4410 / NC-4410) is the same MT7981B design as the Buddy 6. Its vendor device tree is identical to the KN-3411 one apart from the model string, so both variants reuse the KN-3411 .dtsi and add only the per-model compatible and image id.

Tested on hardware, Buddy 6 SE (NC-4410), a unit with the HeYangTek HYF1GQ4UDACAE:

- Flashed the factory image through the bootloader TFTP recovery. The stock NDMS updater rejects it ("does not support dual image technology"): the image spans both firmware slots, which is the layout this family uses in-tree.
- Boots from NAND with the HeYangTek driver (#23742, merged); the partition layout matches stock, zero ECC errors.
- Verified ethernet with the factory MAC, both radios probing with firmware loaded, all three LEDs registered, and the reset button.

The KN-4410 variant differs only in the model string and image id and is untested on hardware, same as the NC-3411.

[...]

PR #24820, opened 2026-08-20
mediatek: filogic: add Keenetic KN-1012 and Netcraze NC-1012

lexfrei:
Adds the Keenetic KN-1012 and its Netcraze NC-1012 rebrand. MT7981B with a combo port: an SFP cage and an Airoha EN8811H 2.5G copper PHY share one SerDes behind a GPIO mux. The stock bootloader cannot be replaced, so U-Boot is chainloaded from it and applies the device tree overlay for whichever medium is plugged in at boot.

## Prerequisite

#24819 is a prerequisite for the copper side of the combo port. Without it the board boots, bridges, runs wifi and works over SFP, but lan4 on copper stays down: the switch attaches the PHY at probe, before air_en8811h can load its firmware from the still-unmounted rootfs, so phylink gets a genphy with no 2500baseT and DSA drops the port. The rest of this PR does not depend on it. The property that patch reads, `needs-host-firmware`, is not in this DTS: it goes in with #24819, next to the code that reads it.

## Tested on hardware

- copper combo port: works against gigabit and 2.5G peers, bridged and standalone, no interface errors
- SFP combo port with a DAC: module EEPROM read over I2C, link at 1 Gbit/s. That is the highest rate my SFP gear shares with this port: the far ends I have do 1, 5 or 10 Gbit/s, none 2.5. The SFP side should also run 2500BASE-X, which is the port's native mode, but I had no 2.5G partner to prove it
- media autoselect both ways: plug the DAC and it boots with the SFP overlay and moves the mux, unplug it and it comes back on copper, no manual steps
- both media at once, a DAC in the cage and a cable in the RJ45: SFP wins, the copper side stays down
- wifi as a working mesh node: AP on both bands plus 802.11s, real clients, moved traffic for hours
- sysupgrade and the vendor TFTP recovery path, buttons, LEDs, both USB controllers

The branch changed shape while these were collected. The current one, a base tree plus two overlays picked by the chainloader, was booted on the unit in both modes and covers the autoselect, both-media and sysupgrade items. The copper figures come from an earlier shape merged locally with #24819, which is how the board is meant to run; the SFP link figure, wifi, recovery, buttons, LEDs and USB were also measured on earlier shapes. The overlay form was checked offline against the two-tree form: same trees, only the way the choice is written changed.

## Notes for reviewers

- The two boards join the family entry in `11_fix_wifi_mac`, which keeps phy1 off rf-eeprom 0xa, the slot the DTS gives the WAN port. Adding the names is the whole of this PR's change to that file. phy1 comes up as 56:9c:53:26:4f:a1 and wan as d4:9c:53:26:4f:a2, measured on this unit.
- The combo port advertises 2500BASE-X only, which follows from phy-mode. On a SerDes the rate is not negotiated (Clause 37 settles duplex and pause, nothing else), so phylink picks the mode from the module. A 10G DAC against a 1G/10G switch gets no link; `ethtool -s lan4 advertise 0x20000000000` drops it to 1000BASE-X and the link comes up. A plain 1G module should be picked up automatically, I had none to check.
- The copper PHY node declares its interrupt on pio 0, edge falling like the BPI-R3 mini's EN8811H nodes; the vendor tree hangs the same line on the switch node as level low. On this tree the line goes unused: phylib parks the interrupt of a PHY attached while the generic driver is bound, which is what happens to this one at switch probe, so it polls. With #24819 merged, forced renegotiations delivered the link events over INTB. The line exists on the board either way, so the node describes it.
- FM25G02B ships its own ECC status decoder. Reusing the neighbouring FM25S01B one would misreport: six of the eight status codes map differently, including a healthy page with four bitflips reading as uncorrectable. What I could not verify is the FM25G02B mapping itself against the datasheet: the table is internally consistent (max matches the 8-bit requirement, buckets monotone), but if anyone has the datasheet, the status-register table is the one thing left uncheckable from here. The driver is [submitted to U-Boot upstream](https://lore.kernel.org/u-boot/20260822205540.1223-1-f@lex.la/) as a port of the Linux support.
- The two bootloader gaps noted here earlier are closed: the chainloaded U-Boot now reads its SPI-NAND in quad mode, matching the settings the stock loader's own FDT uses for this part, and `boot_tftp` tries both brands' image names. Verified booting on the unit.
- Chainloading a second bootloader to choose a device tree is a workaround. The mainline direction is the phy_port work representing SFP cages and combo ports properly; once that reaches this target I plan to rewrite the board on top of it.


package/boot/uboot-mediatek/patches/110-mtd-spinand-fmsh-add-support-for-FM25G02B.patch
```diff
@@ -0,0 +1,136 @@
...
++	}
++
++	return -EINVAL;
++}
++
+ static const struct spinand_info fmsh_spinand_table[] = {
++	SPINAND_INFO("FM25G02B",
++		     SPINAND_ID(SPINAND_READID_METHOD_OPCODE_DUMMY, 0xD2),
++		     NAND_MEMORG(1, 2048, 128, 64, 2048, 41, 1, 1, 1),
++		     NAND_ECCREQ(8, 528),
```
openwrt-ai, line 136: The ECC step size of 528 does not divide the 2048-byte page: `mtd->writesize / ecc.size` gives 3 steps and leaves ~464 bytes of every page outside the reported ECC coverage. No other SPI-NAND entry in either tree uses 528 — a survey of `drivers/mtd/nand/spi/*.c` in u-boot master and Linux 6.18 yields only `NAND_ECCREQ(8, 512)` (81x), `(4, 512)`, `(1, 512)` and `(6, 32)`. The neighbouring FudanMicro part with the same 2048+128 geometry and the same 8-bit requirement uses 512: [`NAND_ECCREQ(8, 512)` for FM25S01B](``https://github.com/openwrt/openwrt/blob/9198fdd3181785d24f7728fcf7fb178e0e5e4a2c/package/boot/uboot-mediatek/patches/100-20-mtd-spi-nand-add-support-for-FudanMicro-chips.patch#L123).``

```suggestion
+		     NAND_ECCREQ(8, 512),
```


---

package/boot/uboot-mediatek/patches/110-mtd-spinand-fmsh-add-support-for-FM25G02B.patch
```diff
@@ -0,0 +1,136 @@
...
++{
++	if (section)
++		return -ERANGE;
++
++	/* reserve 2 bytes for the BBM */
++	region->offset = 2;
++	region->length = 62;
++
++	return 0;
++}
```
openwrt-ai, line 90: This free-OOB layout is copied from FM25S01A (64-byte OOB, no exposed ECC region), but the ECC layout above is copied from FM25S01B (128-byte OOB, ECC at 64..127). The other FudanMicro part with exactly this geometry does *not* hand out bytes 2..63 in one block — it reserves 4 bytes at every 16-byte boundary: [`fm25s01b_ooblayout_free()` returns 4 sections of 12 bytes at `16 * section + 4`](``https://github.com/openwrt/openwrt/blob/9198fdd3181785d24f7728fcf7fb178e0e5e4a2c/package/boot/uboot-mediatek/patches/100-20-mtd-spi-nand-add-support-for-FudanMicro-chips.patch#L65-L74).``

If FM25G02B shares that per-sector spare-area structure, claiming 62 contiguous free bytes lets UBI/JFFS2 write over the chip's internal per-sector metadata. Is the single-section layout taken verbatim from the upstream `d5a5c9eb2ee9` entry, or is it the FM25S01A layout carried over by hand? Worth stating in the patch header either way, since the PR notes flag the FM25G02B tables as the part that could not be checked against a datasheet.

---

package/boot/uboot-mediatek/patches/461-add-keenetic_kn-1012.patch
```diff
@@ -0,0 +1,288 @@
...
++
+ endif
+--- a/arch/arm/Kconfig
++++ b/arch/arm/Kconfig
+@@ -104,6 +104,7 @@ config LINUX_KERNEL_IMAGE_HEADER
+ config LNX_KRNL_IMG_TEXT_OFFSET_BASE
+ 	depends on LINUX_KERNEL_IMAGE_HEADER
+ 	hex
++	default 0x41d80000 if MTK_KERNEL_IMAGE_HEADER
+ 	help
```
openwrt-ai: These last three hunks are not board support — they are new U-Boot infrastructure that happens to be needed by this board:

- `drivers/mtd/nmbm/Kconfig`: `ENABLE_NAND_NMBM`, `NMBM_MAX_RATIO`, `NMBM_MAX_BLOCKS`. Those symbols are referenced all over the tree today ([`board_nmbm_init()` in patch 129](``https://github.com/openwrt/openwrt/blob/9198fdd3181785d24f7728fcf7fb178e0e5e4a2c/package/boot/uboot-mediatek/patches/129-board-mediatek-wire-up-NMBM-support.patch#L127-L149),`` 100-09, 30x-*) but have never had a Kconfig entry, so `#ifdef CONFIG_ENABLE_NAND_NMBM` has been dead code for every mediatek board. This turns that code on tree-wide.
- `arch/arm/mach-mediatek/Kconfig`: `MTK_KERNEL_IMAGE_HEADER`.
- `arch/arm/Kconfig`: a `default 0x41d80000` for the generic `LNX_KRNL_IMG_TEXT_OFFSET_BASE`.

Bundled into a file named `461-add-keenetic_kn-1012.patch` with no `Subject:`/description, none of that is discoverable by the next person who bisects an NMBM regression. Please split the three Kconfig hunks into their own patch(es) with a header explaining what each enables, and keep 461 to the defconfig, DTS and defenv for this board.

---

package/boot/uboot-mediatek/patches/504-add-keenetic_kn-1012.patch
```diff
@@ -0,0 +1,288 @@
+--- /dev/null
```
openwrt-ai, line 15: nit: the `461-` prefix already has two occupants in this directory (`461-add-asus-zenwifi-bt8.patch`, `461-add-netis-nx30v2.patch`), and board patches there currently run up to `503-add-wavlink-wl-wnt100x3.patch`. A new board should take the next free number rather than adding a third file to a collided prefix, since quilt orders by filename and the intended position of this patch relative to the other two is then arbitrary.

---

package/boot/uboot-mediatek/patches/461-add-keenetic_kn-1012.patch
```diff
@@ -0,0 +1,288 @@
...
++CONFIG_LOG=y
++# CONFIG_BOARD_INIT is not set
++CONFIG_BOARD_LATE_INIT=y
++CONFIG_HUSH_PARSER=y
++CONFIG_SYS_PROMPT="MT7981> "
++CONFIG_CMD_CPU=y
++CONFIG_CMD_LICENSE=y
++CONFIG_CMD_BOOTMENU=y
++CONFIG_CMD_ASKENV=y
++CONFIG_CMD_ERASEENV=y
```
openwrt-ai: nit: this defconfig selects no environment backend at all, so the build lands on `ENV_IS_NOWHERE` and `eraseenv`/`saveenv` can only ever fail at runtime. Every other mt7981 board in this directory pairs `CONFIG_ENV_USE_DEFAULT_ENV_TEXT_FILE` with `CONFIG_ENV_IS_IN_UBI=y` plus `ENV_UBI_PART`/`ENV_UBI_VOLUME` (e.g. `475-add-teralink_tl3020-256mb.patch`, `501-add-netis-n6-v2.patch`) — though those boards also set `UBOOTENV_IN_UBI := 1` in their image recipe, which this one deliberately does not, so a read-only environment looks like the intended design for a chainloader.

If that is the intent, dropping `CONFIG_CMD_ERASEENV=y` (and `CONFIG_ENV_OVERWRITE=y` on line 165) would keep the defconfig honest about it.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -2282,6 +2282,155 @@ define Device/keenetic_kn-1812-common
...
+	@mv $@.new $@
+	@rm -rf $@.payload $@.uboot.lzma $@.uboot.its $@.its $@.fakefs $@.fakesquashfs
+endef
+
+define Device/keenetic_kn-1012-common
+  DEVICE_DTS_DIR := ../dts
+  # ethtool ships for combo-port diagnosis: -m reads the SFP module EEPROM and
+  # the link modes show which media is actually live.
+  DEVICE_PACKAGES := kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware \
+	kmod-usb3 kmod-phy-airoha-en8811h kmod-sfp ethtool
```
openwrt-ai: `ethtool` is a userspace diagnostic, not something the board needs to work — and no device anywhere in `target/linux/*/image/*.mk` currently puts it in `DEVICE_PACKAGES`. Shipping it by default grows the image for every user of these two profiles to serve a bring-up workflow; anyone diagnosing the combo port can `opkg install ethtool`. The continuation line also uses one tab where the neighbouring recipes use two ([kap-630](https://github.com/openwrt/openwrt/blob/9198fdd3181785d24f7728fcf7fb178e0e5e4a2c/target/linux/mediatek/image/filogic.mk#L2243-L2244), [kn-1812](https://github.com/openwrt/openwrt/blob/9198fdd3181785d24f7728fcf7fb178e0e5e4a2c/target/linux/mediatek/image/filogic.mk#L2269-L2271)).

```suggestion
  DEVICE_PACKAGES := kmod-mt7915e kmod-mt7981-firmware mt7981-wo-firmware \
		kmod-usb3 kmod-phy-airoha-en8811h kmod-sfp
```


---

target/linux/mediatek/image/filogic.mk
```diff
@@ -2282,6 +2282,155 @@ define Device/keenetic_kn-1812-common
...
+  UBINIZE_OPTS := -E 5
+  BLOCKSIZE := 128k
+  PAGESIZE := 2048
+  KERNEL_SIZE := 8192k
+  IMAGE_SIZE := 233984k
+  KERNEL := kernel-bin | lzma | kn1012-fit | kn1012-prepend-uboot
```
openwrt-ai, line 2508: Nothing bounds the size of the finished kernel blob. `kn1012-prepend-uboot` only checks that the chainloader half fits in its 1 MiB pad; the dual-media FIT concatenated behind it is unchecked, and the boot path has a hard 7 MiB ceiling on it — the defenv does [`mtd read nmbm0 $loadaddr $kernel_off $kernel_size` with `kernel_size=0x700000`](``https://github.com/openwrt/openwrt/blob/9198fdd3181785d24f7728fcf7fb178e0e5e4a2c/package/boot/uboot-mediatek/patches/461-add-keenetic_kn-1012.patch#L224-L234).`` Once the FIT (kernel + two DTBs) crosses that, the build still succeeds, the factory image still passes `check-size` against `IMAGE_SIZE`, and sysupgrade still writes it — but the flashed image overruns into `ubi` and `bootm` sees a truncated FIT.

`check-size $$(KERNEL_SIZE)` on the kernel pipeline catches it at build time (same 8 MiB budget: 1 MiB chainloader + 7 MiB readable window), matching the pattern used in `ath79/image/nand.mk` and `econet/image/en751221.mk`.

```suggestion
  KERNEL := kernel-bin | lzma | kn1012-fit | kn1012-prepend-uboot | check-size $$(KERNEL_SIZE)
```


---

target/linux/mediatek/image/filogic.mk
```diff
@@ -2282,6 +2282,155 @@ define Device/keenetic_kn-1812-common
...
+		echo "chainloader does not fit in the 1 MiB it is padded to"; \
+		false; \
+	}
+	$(call Build/pad-to,1024k)
+	cat $@ $@.payload > $@.new
+	@mv $@.new $@
+	@rm -rf $@.payload $@.uboot.lzma $@.uboot.its $@.its $@.fakefs $@.fakesquashfs
+endef
+
+define Device/keenetic_kn-1012-common
```
openwrt-ai, line 2499: nit: this block lands between `Device/keenetic_kn-1812-common` and `Device/keenetic_kn-1812`, splitting that pair and breaking the file's alphabetical device ordering (kap-630 → kn-1812 → kn-3711 → kn-3811). `keenetic_kn-1012-common` / `keenetic_kn-1012` belong just above `Device/keenetic_kn-1812-common` at line 2267. The Netcraze side is already placed correctly.

---

target/linux/mediatek/dts/mt7981b-keenetic-kn-1012.dtsi
```diff
@@ -0,0 +1,452 @@
...
+			};
+		};
+	};
+};
+
+&wifi {
+	nvmem-cell-names = "eeprom";
+	nvmem-cells = <&eeprom_factory_0>;
+	status = "okay";
+};
```
openwrt-ai, line 450: nit: the commit message documents the MAC layout as `5g rf-eeprom 0xa, shared with WAN`, but the PR body calls that same value a collision that #24623 exists to fix. Whichever way that resolves, the commit message table will describe the wrong thing once both land — either it states a deliberate WAN/5 GHz collision that is about to be corrected, or it needs updating to the post-#24623 source.

Related: the sibling Keenetic boards get their phy1 address from [the `keenetic,kap-630|kn-3711|kn-3811|kn-3911|netcraze,nap-630` case in `11_fix_wifi_mac`](``https://github.com/openwrt/openwrt/blob/9198fdd3181785d24f7728fcf7fb178e0e5e4a2c/target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac#L179-L186)`` and this board is absent from it — which the PR body says #24623 supplies, so noting it only so the dependency stays visible at merge time.

---

openwrt-ai (commented):
Reviewed 3 commits (first review of this PR).

## Commit checks

- `6266e54` "uboot-mediatek: add Keenetic KN-1012 chainload build" — the subject scopes this to one board's chainload build, but `461-add-keenetic_kn-1012.patch` also adds three tree-wide U-Boot Kconfig changes: the `ENABLE_NAND_NMBM` / `NMBM_MAX_RATIO` / `NMBM_MAX_BLOCKS` symbols in `drivers/mtd/nmbm/Kconfig` (which have been referenced by dead `#ifdef`s across the whole mediatek tree until now), `MTK_KERNEL_IMAGE_HEADER` in `arch/arm/mach-mediatek/Kconfig`, and a `LNX_KRNL_IMG_TEXT_OFFSET_BASE` default in the generic `arch/arm/Kconfig`. Neither the commit message nor the (absent) patch header mentions any of it. Details inline.

Everything else checks out against the tree: `IMAGE_SIZE`/partition arithmetic matches the concat layout and its KN-3811 sibling exactly, the ZYIMAGE_ID pattern follows the Keenetic/Netcraze convention, `0x600000`/`0x700000` in the defenv line up with the 1 MiB chainloader pad at the head of `firmware_1`, the U-Boot `gpio input 23` / `gpio set|clear 27` polarity is consistent with the `GPIO_ACTIVE_LOW` mux and mod-def0 in the Linux DTS, `02_network`'s default case already yields `lan1 lan2 lan3 lan4` + `wan`, and the commit message carries all three sections required for a new device. CI is green on the head commit.


---

package/boot/uboot-mediatek/patches/110-mtd-spinand-fmsh-add-support-for-FM25G02B.patch
lexfrei: Verbatim from d5a5c9eb2ee9, the mainline entry that added this chip: it uses NAND_ECCREQ(8, 528). I'd keep the U-Boot port identical to the accepted Linux entry rather than diverge.

package/boot/uboot-mediatek/patches/110-mtd-spinand-fmsh-add-support-for-FM25G02B.patch
lexfrei: Also verbatim from d5a5c9eb2ee9. Its fm25g01b_ooblayout_free returns one region at offset 2 length 62, and both parts share it. The header names the two upstream commits.

package/boot/uboot-mediatek/patches/461-add-keenetic_kn-1012.patch
lexfrei: Done. The three Kconfig hunks are their own patch now, numbered 409 so it lands after 408 which touches the same Kconfig, with a header explaining each symbol.

package/boot/uboot-mediatek/patches/504-add-keenetic_kn-1012.patch
lexfrei: Fixed, it's 504 now, the next free number.

package/boot/uboot-mediatek/patches/461-add-keenetic_kn-1012.patch
lexfrei: Dropped CONFIG_CMD_ERASEENV and CONFIG_ENV_OVERWRITE. The read-only environment is the intent for a chainloader.

target/linux/mediatek/image/filogic.mk
lexfrei: Removed ethtool, and fixed the continuation indent to two tabs.

target/linux/mediatek/image/filogic.mk
lexfrei: include/image.mk already appends check-size $(KERNEL_SIZE) to the KERNEL chain after concat_cmd, so the 8 MiB budget is checked at build time. An explicit one would run twice on the same bytes.

target/linux/mediatek/image/filogic.mk
lexfrei: Moved above keenetic_kn-1812-common, alphabetical now.

target/linux/mediatek/dts/mt7981b-keenetic-kn-1012.dtsi
lexfrei: Reworded to '5g rf-eeprom 0xa, the second MAC slot of the WiFi EEPROM'. Neutral fact, stays correct whichever way #24623 lands.

package/boot/uboot-mediatek/patches/409-mediatek-Kconfig-entries-for-NMBM-and-kernel-image-header.patch
```diff
@@ -0,0 +1,66 @@
...
+
+- ENABLE_NAND_NMBM, NMBM_MAX_RATIO and NMBM_MAX_BLOCKS: the NMBM code
+  and its call sites have been guarded by ifdef CONFIG_ENABLE_NAND_NMBM
+  without any Kconfig entry declaring the symbol, so that code was
+  unreachable from a defconfig. Boards whose stock firmware manages bad
+  blocks through NMBM need to enable it to see the same flash layout.
+- MTK_KERNEL_IMAGE_HEADER: lets a MediaTek platform select the ARM64
+  Linux image header on the U-Boot binary, which a vendor loader that
+  only boots kernel-shaped images requires for chainloading. The symbol
+  is promptless on purpose: a board selects it, a defconfig cannot.
```
openwrt-ai: The header claims the symbol is promptless and can only be reached via `select`, but the hunk below gives `MTK_KERNEL_IMAGE_HEADER` a prompt string (`bool "Build U-Boot with an arm64 Linux kernel image header"`, line 48) and nothing in the tree selects it — the board turns it on straight from its defconfig: [`CONFIG_MTK_KERNEL_IMAGE_HEADER=y` in 504-add-keenetic_kn-1012.patch](``https://github.com/openwrt/openwrt/blob/0a77fc16649456fdddafd65bce2fbec0db8ecd40/package/boot/uboot-mediatek/patches/504-add-keenetic_kn-1012.patch#L135).``

A promptless symbol could not be set from a defconfig at all, so the code is the correct half here and the description is what needs adjusting.

```suggestion
  only boots kernel-shaped images requires for chainloading. A board
  defconfig enables it directly.
```

---

package/boot/uboot-mediatek/patches/409-mediatek-Kconfig-entries-for-NMBM-and-kernel-image-header.patch
```diff
@@ -0,0 +1,66 @@
...
++	  to chainload U-Boot as its payload.
++
+ endif
+--- a/arch/arm/Kconfig
++++ b/arch/arm/Kconfig
+@@ -104,6 +104,7 @@ config LINUX_KERNEL_IMAGE_HEADER
+ config LNX_KRNL_IMG_TEXT_OFFSET_BASE
+ 	depends on LINUX_KERNEL_IMAGE_HEADER
+ 	hex
++	default 0x41d80000 if MTK_KERNEL_IMAGE_HEADER
```
openwrt-ai, line 62: Patching a `default ... if <platform symbol>` into the generic `arch/arm/Kconfig` is not how U-Boot expresses this. Every platform that needs a `LNX_KRNL_IMG_TEXT_OFFSET_BASE` value redeclares the symbol in its *own* Kconfig with a bare `default` instead — [`arch/arm/mach-snapdragon/Kconfig:26`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/arch/arm/mach-snapdragon/Kconfig#L26-L27), [`arch/arm/mach-rockchip/Kconfig:761`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/arch/arm/mach-rockchip/Kconfig#L761-L762), [`arch/arm/mach-apple/Kconfig:28`](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/arch/arm/mach-apple/Kconfig#L28-L29), plus mach-mmp and mach-tegra. Following that pattern drops `arch/arm/Kconfig` from this patch entirely, leaving nothing outside `mach-mediatek` to carry and rebase across U-Boot bumps:

```
config LNX_KRNL_IMG_TEXT_OFFSET_BASE
	default 0x41d80000 if MTK_KERNEL_IMAGE_HEADER
```

placed alongside `MTK_KERNEL_IMAGE_HEADER` in the `arch/arm/mach-mediatek/Kconfig` hunk above.

Separately, `0x41d80000` is exactly `CONFIG_TEXT_BASE - 0x80000` for this one board (504 sets `CONFIG_TEXT_BASE=0x41e00000`), yet the condition is the platform-wide `MTK_KERNEL_IMAGE_HEADER`, which any `TARGET_MT7981` board can enable. A second board with a different `TEXT_BASE` would silently get a wrong TEXT_OFFSET in its image header. A one-line comment tying the constant to KN-1012's `TEXT_BASE` would make that trap visible.

---

target/linux/mediatek/image/filogic.mk
```diff
@@ -2282,6 +2419,16 @@ define Device/keenetic_kn-1812-common
...
 
+# FIT with one kernel and both media DTBs (copper on port 5 vs SFP on
+# port 5); the chainloaded U-Boot picks config-copper or config-sfp per
+# boot from the SFP mod-def0 state.
+# mediatek sets separate_ramdisk, so an initramfs kernel is built with an
+# empty CONFIG_INITRAMFS_SOURCE and carries no rootfs of its own: the FIT
+# has to supply the cpio. The guard keeps it out of the flash kernel,
+# which takes its rootfs from UBI.
+kn1012_initrd = $(and $(findstring with-initrd,$(1)),$(CONFIG_TARGET_ROOTFS_INITRAMFS_SEPARATE))
+kn1012_initrd_data = $(KERNEL_BUILD_DIR)/initrd$(if $(TARGET_PER_DEVICE_ROOTFS),.$(ROOTFS_ID/$(DEVICE_NAME))).cpio$(strip $(call Build/initrd_compression))
```
openwrt-ai: nit: the kn-1012 device blocks moved up, but these two helpers and the comment describing the dual-media FIT stayed behind — they now sit between `Device/keenetic_kn-1812-common` and `Device/keenetic_kn-1812`, splitting that pair, and ~150 lines below the [`define Build/kn1012-fit`](https://github.com/openwrt/openwrt/blob/0a77fc16649456fdddafd65bce2fbec0db8ecd40/target/linux/mediatek/image/filogic.mk#L2267) that is their only consumer. Nothing breaks (recursively-expanded variables are resolved at recipe time), but the comment no longer sits next to the recipe it explains.

Moving both lines and the comment to just above `define Build/kn1012-fit` puts them back with the code they belong to and leaves the kn-1812 pair contiguous again.

---

openwrt-ai (commented):
Re-reviewed the force-pushed range (3 rewritten commits, `9198fdd..0a77fc1`). All eight points from the previous round are addressed: the Kconfig hunks are split into their own `409-` patch with a header, the board patch moved to the next free number `504-`, `CMD_ERASEENV`/`ENV_OVERWRITE` are gone from the read-only defconfig, `ethtool` is dropped and the continuation indent is two tabs, the kn-1012 device blocks are back in alphabetical order, and the MAC table wording is now neutral with respect to #24623. All three commit messages match their diffs — `6da9af7` now states the Kconfig split explicitly.

Two corrections to my own earlier comments, for the record:

- The `check-size $$(KERNEL_SIZE)` request was wrong — [`include/image.mk:759-760`](https://github.com/openwrt/openwrt/blob/0a77fc16649456fdddafd65bce2fbec0db8ecd40/include/image.mk#L759-L760) already appends it after the KERNEL chain whenever `KERNEL_SIZE` is set, so the 8 MiB budget is enforced at build time as you said.
- The `From 0000000...` placeholder line added to the patch headers matches existing practice in this directory (`111-mtd-nand-spi-otp-*.patch`), no objection.

Three new findings inline, on the newly-added `409-` patch and the filogic.mk reshuffle. The `ECCREQ(8, 528)` and single-region `ooblayout_free` threads stay open only because `d5a5c9eb2ee9` is not reachable from any tree I can read here (it postdates 6.18.44) — I could not confirm the "verbatim from upstream" claim either way, and take you at your word.

CI is green on `0a77fc1` (22 checks, mediatek/filogic kernel + package builds included).

---

package/boot/uboot-mediatek/patches/409-mediatek-Kconfig-entries-for-NMBM-and-kernel-image-header.patch
lexfrei: The description was stale, from a revision where the symbol really was promptless. Fixed the header; a defconfig enabling it directly is the intended shape.

package/boot/uboot-mediatek/patches/409-mediatek-Kconfig-entries-for-NMBM-and-kernel-image-header.patch
lexfrei: Moved the default into mach-mediatek next to MTK_KERNEL_IMAGE_HEADER, following the snapdragon/rockchip pattern, so the generic arch/arm/Kconfig is out of the patch. Also added a comment tying 0x41d80000 to this board's TEXT_BASE.

target/linux/mediatek/image/filogic.mk
lexfrei: Moved the helpers and the comment right above Build/kn1012-fit; the kn-1812 pair is contiguous again.

target/linux/mediatek/dts/mt7981b-keenetic-kn-1012.dtsi
```diff
@@ -0,0 +1,453 @@
...
+
+&mdio_bus {
+	/*
+	 * Airoha EN8811H, left polling: pio 36 is SMI_MDC and is muxed for
+	 * MDIO by this board, and pio 38 already carries the switch interrupt.
+	 */
+	phy13: ethernet-phy@d {
+		reg = <0xd>;
+		airoha,pnswap-rx;
```
openwrt-ai, line 230: nit: this one-liner belongs squashed into `f7ede66` ("mediatek: filogic: add Keenetic KN-1012 and Netcraze NC-1012"), the immediately preceding commit that adds this very file. The board does not exist in the tree before that commit, so there is no released state being fixed here — as a separate commit it only creates a bisect point where the copper combo port links at any speed and transmits nothing. The explanation in `963dc7f`'s body is worth keeping; it just belongs in the introducing commit, or as a comment on the node, the way the `usbreset` and `sfp1mux` quirks are already documented in this file.

Squashing would also settle the author identity, which currently differs across the series: `f7e4af3`, `4ad530e` and `f7ede66` are authored and signed off as `Aleksei Sviridkin <3811295@gmail.com>`, `963dc7f` as `Aleksei Sviridkin <f@lex.la>`.

The property itself checks out — it is consumed by [`device_property_read_bool(dev, "airoha,pnswap-rx")` at `drivers/net/phy/air_en8811h.c:1008`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/air_en8811h.c#L1008) in 6.18.44, which is this target's `KERNEL_PATCHVER`, and the same property and placement are used by `mt7981b-openwrt-one.dts`, `mt7986a-bananapi-bpi-r3-mini.dts` and `mt7987a-rfb-eth0-mt7531.dtso`.

---

package/boot/uboot-mediatek/patches/409-mediatek-Kconfig-entries-for-NMBM-and-kernel-image-header.patch
```diff
@@ -0,0 +1,64 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Aleksei Sviridkin <3811295@gmail.com>
+Date: Thu, 21 Aug 2026 12:00:00 +0300
+Subject: [PATCH] mediatek: add Kconfig entries for NMBM and the kernel image header
+
+Three Kconfig additions that board patches can select:
```
openwrt-ai: nit: this lead-in is now out of step with the bullets it introduces. After the rework nothing in this patch is reached by a `select` from a board patch: `ENABLE_NAND_NMBM` and `MTK_KERNEL_IMAGE_HEADER` are both switched on directly from the board defconfig (`CONFIG_ENABLE_NAND_NMBM=y`, `CONFIG_MTK_KERNEL_IMAGE_HEADER=y` in `504-add-keenetic_kn-1012.patch`), and `LNX_KRNL_IMG_TEXT_OFFSET_BASE` only gains a `default`. Bullet two already says "A board defconfig enables it", which contradicts the header line.

```suggestion
Three Kconfig additions the board defconfigs rely on:
```

---

openwrt-ai (commented):
Re-reviewed the range `0a77fc1..963dc7f` — three rewritten commits plus one new commit (`963dc7f`). Net change is small: the `409-` patch reshaped, the filogic.mk helpers moved, and one DTS property added.

Both points from the previous round are addressed, and I verified the fixes rather than taking them on trust:

- `LNX_KRNL_IMG_TEXT_OFFSET_BASE` is now redeclared in `arch/arm/mach-mediatek/Kconfig` with a bare `default`, exactly the shape used by [mach-snapdragon](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/arch/arm/mach-snapdragon/Kconfig#L26-L27), [mach-rockchip](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/arch/arm/mach-rockchip/Kconfig#L761-L762) and [mach-apple](https://github.com/u-boot/u-boot/blob/ece349ade2973e220f524ce59e59711cc919263f/arch/arm/mach-apple/Kconfig#L28-L29); `arch/arm/Kconfig` is out of the patch. The type is inherited from the `hex` declaration in `arch/arm/Kconfig`, the new block sits inside `if ARCH_MEDIATEK`, and `TARGET_MT7981` selects `ARM64` so the `select LINUX_KERNEL_IMAGE_HEADER` does not select past an unmet dependency. `0x41d80000` is `0x41e00000 - 0x80000`, matching `CONFIG_TEXT_BASE` in the board defconfig, and the new comment now says so.
- Hunk arithmetic on the reshaped patch is right: `@@ -222,4 +222,20 @@` matches the 16 added lines, and line 222 in the post-`301-` tree really is `config RESET_BUTTON_SETTLE_DELAY`.
- The `kn1012_initrd` / `kn1012_initrd_data` helpers and their comment now sit directly above `Build/kn1012-fit`, their only consumer, and the `keenetic_kn-1812-common` / `keenetic_kn-1812` pair is contiguous again.
- `airoha,pnswap-rx` is a real binding on this kernel, and putting it on `phy13` in the shared dtsi is harmless for the SFP tree, where nothing attaches that PHY node.

The `NAND_ECCREQ(8, 528)` and single-region `ooblayout_free` threads stay open on the same footing as last round — `d5a5c9eb2ee9` is still not reachable from anything I can read here, so I can neither confirm nor refute the "verbatim from upstream" claim.

Two nits inline, both cosmetic. CI is green on `963dc7f` (22 checks, mediatek/filogic kernel and package builds included).

---

target/linux/mediatek/dts/mt7981b-keenetic-kn-1012.dtsi
lexfrei: Squashed into the introducing commit; the series is one author again.

package/boot/uboot-mediatek/patches/409-mediatek-Kconfig-entries-for-NMBM-and-kernel-image-header.patch
lexfrei: Taken as suggested.

openwrt-ai (commented):

Re-reviewed `963dc7f..f8fb414`. The series went from four commits to three; the only tree change in the whole range is one line of prose in the `409-` patch header, so this round was mostly verifying that the squash did what it says.

- `git diff 963dc7f f8fb414` touches exactly one file — `409-mediatek-Kconfig-entries-for-NMBM-and-kernel-image-header.patch`, the lead-in now reading "Three Kconfig additions the board defconfigs rely on". `target/` and the other two U-Boot patches are byte-identical to the previously reviewed state, so nothing was lost or silently reshuffled in the rewrite.
- The squash of `963dc7f` into `f8fb414` is clean: `airoha,pnswap-rx` is still on `phy13` in the shared dtsi, and the P/N-swap explanation moved into the introducing commit's body rather than being dropped. There is no longer a bisect point where the copper combo port links but does not transmit.
- Author identity is now uniform across all three commits — `Aleksei Sviridkin <3811295@gmail.com>` as author and in `Signed-off-by` on each, which is what the FormalityCheck "Git & Commits" job checks.
- All three commit messages still match their diffs, including the new pnswap paragraph in `f8fb414`.

The `NAND_ECCREQ(8, 528)` and single-region `ooblayout_free` threads remain open on the same footing as the last two rounds: `d5a5c9eb2ee9` is not reachable from any tree available here, so the "verbatim from upstream" claim is neither confirmed nor refuted from my side — noting it only so it stays visible at merge time, not as an objection.

CI is green on `f8fb414` (22 checks, mediatek/filogic kernel build, package build and all three FormalityCheck jobs included).

---

target/linux/mediatek/dts/mt7981b-keenetic-kn-1012.dtsi
```diff
@@ -0,0 +1,456 @@
...
+	/*
+	 * Airoha EN8811H. The vendor tree declares this PHY's INTB on the
+	 * switch node instead, but pio 0 is the PHY's own line and the switch
+	 * has its own on pio 38.
+	 */
+	phy13: ethernet-phy@d {
+		reg = <0xd>;
+		airoha,pnswap-rx;
+		interrupt-parent = <&pio>;
+		interrupts = <0 IRQ_TYPE_LEVEL_LOW>;
```
openwrt-ai: Every other EN8811H node I can reach declares INTB edge-triggered, not level: [`mt7981b-openwrt-one.dts:162`](https://github.com/openwrt/openwrt/blob/a9b9cb9d70e971036efbe0ddfd82d8a36c054a95/target/linux/mediatek/dts/mt7981b-openwrt-one.dts#L162), [`mt7986a-bananapi-bpi-r3-mini.dts:179`](https://github.com/openwrt/openwrt/blob/a9b9cb9d70e971036efbe0ddfd82d8a36c054a95/target/linux/mediatek/dts/mt7986a-bananapi-bpi-r3-mini.dts#L179) and [`:215`](https://github.com/openwrt/openwrt/blob/a9b9cb9d70e971036efbe0ddfd82d8a36c054a95/target/linux/mediatek/dts/mt7986a-bananapi-bpi-r3-mini.dts#L215) in this tree, plus [the same PHY in mainline 6.18.44](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/arch/arm64/boot/dts/mediatek/mt7986a-bananapi-bpi-r3-mini.dts#L201) — four instances, all `IRQ_TYPE_EDGE_FALLING`.

The distinction matters more than usual for this driver. [`en8811h_clear_intr()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/air_en8811h.c#L1141-L1155) is wired as both [`.config_intr` and the body of `.handle_interrupt`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/air_en8811h.c#L1199-L1200), and all it does is issue the MCU INTCLR command — it never masks the source. So `phy_disable_interrupts()` cannot actually stop the PHY from pulling INTB low. Under a level trigger, an INTB asserted while phylib has the IRQ parked — precisely the state your commit message describes this PHY going through at switch probe — keeps re-entering genirq until it gives up with `nobody cared` and disables the line permanently, which would also take out anything sharing it later. Under edge-falling a stuck-low line is simply never re-latched.

Was `IRQ_TYPE_LEVEL_LOW` measured on this board, or carried over from the switch node's `IRQ_TYPE_LEVEL_HIGH` two nodes down? If the former, a word in the comment above would keep the next reader from "fixing" it to match the siblings.

```suggestion
		interrupts = <0 IRQ_TYPE_EDGE_FALLING>;
```

---

openwrt-ai (commented):
Re-reviewed `f8fb414..a9b9cb9`. All three commits were rewritten; the tree delta is small — hunk-header context regenerated on `110-` and `409-`, a `Date:` weekday corrected, quad-I/O widths added to the U-Boot DTS, a second TFTP image name in the defenv, and the EN8811H INTB now described in the Linux dtsi.

Verified rather than taken on trust:

- Hunk arithmetic on the regenerated patches holds. `110-`'s `@@ -120,7 +130,90 @@` is exactly 7 context + 83 added lines, and the `+130` start follows the 10 lines the first hunk adds. The first hunk correctly gained *no* function-context suffix (a `#define` does not match diff's default function regex) while the second did — that asymmetry is what a real refresh produces, not a hand edit.
- `504-`: `@@ -0,0 +1,108 @@`, `+1,102` and `+1,23` each land exactly on the next `---` / EOF, so the two new `spi-*-bus-width` lines and the new `altbootfile` line are accounted for.
- `spi-tx-bus-width = <4>` selects something that exists: `write_cache_variants` in `100-20-mtd-spi-nand-add-support-for-FudanMicro-chips.patch` really does carry `SPINAND_PROG_LOAD_1S_1S_4S_OP`, and the FM25G02B entry sets `SPINAND_HAS_QE_BIT`. The pair also now matches what the Linux dtsi already declared for the same chip.
- `boot_tftp`'s `&&` / `;` chaining is right — a failed download skips its own `bootm` and falls through to the second name, and a successful one boots without reaching it.
- `Fri, 21 Aug 2026` is the correct weekday.
- The new `interrupts` property resolves and is wired to something real: `IRQ_TYPE_LEVEL_LOW` reaches `mt7981b.dtsi` via `arm-gic.h` → `irq.h`, `&pio` is an `interrupt-controller` with `#interrupt-cells = <2>`, pio 0 is claimed by no other node in this board's DTS and by none of the pin groups this board selects (`mdio_pins`, `spi0_flash_pins`, `i2c0_1` = pins 30/31), and `air_en8811h` does implement `.handle_interrupt`. The trigger *type* is the one thing I'd question; inline.

The `NAND_ECCREQ(8, 528)` and single-region `ooblayout_free` threads stay open on the same footing as the previous three rounds — `d5a5c9eb2ee9` is still not reachable from any tree available here, so the "verbatim from upstream" claim is neither confirmed nor refuted from my side. Noting it only so it stays visible at merge time. The new "Submitted upstream" paragraph in the `110-` header is a useful addition for exactly that reason.

## Commit checks

- `a9b9cb9` "mediatek: filogic: add Keenetic KN-1012 and Netcraze NC-1012" — carries two `Signed-off-by` lines for the same person under different addresses (`<3811295@gmail.com>` and `<f@lex.la>`); the other two commits in the series carry only the first, and the author identity on all three is `<3811295@gmail.com>`. Looks like a `git commit -s` landing on an already-signed message during the squash. FormalityCheck passes either way, so purely cosmetic.

Otherwise all three commit messages match their diffs, including the new quad-mode and dual-name TFTP paragraphs in `b7e173d` and the INTB paragraph in `a9b9cb9`.

CI on `a9b9cb9` has nothing failed: 10 jobs green (all three FormalityCheck jobs included), 4 still queued or running (the mediatek/filogic kernel build and patch check, and the two package builds), plus one cancelled job belonging to a superseded run.

---

lexfrei:
Switched to `IRQ_TYPE_EDGE_FALLING` and measured it on the board rather than taking either side on trust.

With edge falling, three forced autonegotiation restarts each delivered both link events (interrupt counter 1 to 3 to 5 to 7), the link came back at 1 Gbit/s every time, the counter stayed put over 30 s of idle, and nothing was left unhandled. Traffic through the port after that: 40/40, no errors, no flaps.

The vendor DTS really does say level low, on the switch node together with the gsw interrupt, which is where the value came from. It works in normal operation, but your point about the driver clearing the source without masking it stands, and matching the other four EN8811H nodes is worth more than matching a vendor tree that describes the line on a different node. The comment now records both, so nobody flips it back to the vendor value later.

openwrt-ai (commented):

Re-reviewed `a9b9cb9..4decdd9`. Only the board-support commit was rewritten (`3e3b6fd` and `b7e173d` are byte-identical), and the whole tree delta is four lines in `mt7981b-keenetic-kn-1012.dtsi` — the INTB trigger on `phy13` plus the comment above it.

Verified rather than taken on trust:

- The trigger is now `IRQ_TYPE_EDGE_FALLING`, and the claim that this "matches every other EN8811H in the tree" holds: those are the only three other EN8811H nodes that declare an interrupt at all, and all three are edge falling — [`mt7981b-openwrt-one.dts:162`](https://github.com/openwrt/openwrt/blob/4decdd9166340b4080516dadb78521746e5aa205/target/linux/mediatek/dts/mt7981b-openwrt-one.dts#L162), [`mt7986a-bananapi-bpi-r3-mini.dts:179`](https://github.com/openwrt/openwrt/blob/4decdd9166340b4080516dadb78521746e5aa205/target/linux/mediatek/dts/mt7986a-bananapi-bpi-r3-mini.dts#L179) and [`:215`](https://github.com/openwrt/openwrt/blob/4decdd9166340b4080516dadb78521746e5aa205/target/linux/mediatek/dts/mt7986a-bananapi-bpi-r3-mini.dts#L215). `mt7981b-mercusys-mr85x.dts` and the airoha-target boards declare the PHY without an interrupt, so they are not counter-examples.
- The driver rationale in the new comment is accurate against this target's `KERNEL_PATCHVER` (6.18): [`en8811h_handle_interrupt()`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/air_en8811h.c#L1158-L1171) does nothing but `en8811h_clear_intr()` + `phy_trigger_machine()`, and the same `en8811h_clear_intr` is also wired as [`.config_intr`](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/air_en8811h.c#L1199-L1200) — nothing masks the source, which is exactly the situation edge triggering tolerates and level triggering does not.
- `IRQ_TYPE_EDGE_FALLING` comes from the same `dt-bindings/interrupt-controller/irq.h` that already reached this file via `arm-gic.h`, so the include set is unchanged, and the mt7531 node two nodes down keeps its own separate `IRQ_TYPE_LEVEL_HIGH` on pio 38.
- The commit message's new paragraph matches the diff, including the measurement note, and the `airoha,pnswap-rx` paragraph from the earlier squash is still in place.

The `NAND_ECCREQ(8, 528)` and single-region `ooblayout_free` threads remain open on the same footing as the previous four rounds: `d5a5c9eb2ee9` is not reachable from any tree available here, so the "verbatim from upstream" claim is neither confirmed nor refuted from my side. Noting it only so it stays visible at merge time, not as an objection.

## Commit checks

- `4decdd9` "mediatek: filogic: add Keenetic KN-1012 and Netcraze NC-1012" — unchanged from last round, repeated only because the commit was rewritten: it still carries two `Signed-off-by` lines for the same person under different addresses (`<3811295@gmail.com>`, `<f@lex.la>`), while the other two commits carry only the first. All three are authored as `<3811295@gmail.com>` and committed as `<f@lex.la>`. FormalityCheck passes either way, so purely cosmetic.

CI is green on `4decdd9` (22 checks, mediatek/filogic kernel build and patch check, both package builds and all three FormalityCheck jobs included; nothing failed).

---

openwrt-ai (commented):

The head moved from `4decdd9` to `587610f`, but this is a rebase onto current `main` (`f0204d78`), not a content change. The PR's own diff is unchanged: base..head is the same 11 files / 1179 insertions as the previously reviewed state, and an interdiff of the two PR-only diffs comes back empty apart from hunk-header line numbers shifting and a couple of neighbouring context lines (`mt7981_mercusys_mr85x` in the `uboot-mediatek/Makefile` `UBOOT_TARGETS` list, a moved `netcraze_nap-630` anchor in `filogic.mk`) that main introduced around the unchanged additions. All three commit messages still match their diffs.

The two long-standing open threads carry over on the same footing as the previous five rounds — the U-Boot FM25G02B `NAND_ECCREQ(8, 528)` and single-region `ooblayout_free`, whose "verbatim from upstream" provenance (`d5a5c9eb2ee9`) I still cannot confirm or refute from the trees available here (it postdates v6.18.44). Noting it only so it stays visible at merge time, not as an objection.

CI on `587610f` has nothing failed: all three FormalityCheck jobs and the changed-files/changed-packages steps are green, the kernel and package builds are still queued.

---

openwrt-ai (commented):

The head moved from `587610f` to `7bf0b47`, but this is a rebase from `f0204d78` onto current `main` (`1ad2e78e`), not a content change. Verified rather than assumed:

- The PR-only diffs are byte-identical apart from blob hashes and hunk-header line numbers. `git diff f0204d78..587610f` and `git diff 1ad2e78e..7bf0b47` differ in exactly four places, all of them positional: the `uboot-mediatek/Makefile` `UBOOT_TARGETS` hunk moved from `@@ -1361` to `@@ -1397`, `platform.sh` from `@@ -149` to `@@ -152`, and the two `filogic.mk` hunks from `@@ -2264`/`@@ -2753` to `@@ -2319`/`@@ -2808`. No added, removed or reworded lines anywhere in the 1274-line PR diff.
- All three commit messages, author identities and committer identities are byte-identical to the previously reviewed state — the only difference in `git log --format='%H%n%an <%ae>%n%cn <%ce>%n%B'` output across the two ranges is the three SHAs themselves.
- The rebase pulled in the two BPi-R4 Pro 8X commits (`3aa068ac71f`, `554b5dcc519`), which are the only base-branch changes touching `package/boot/uboot-mediatek` or `target/linux/mediatek` in the range. Neither collides with this PR: the new `472-add-bpi-r4-pro-8x.patch` leaves `504-add-keenetic_kn-1012.patch` and `409-` unaffected (504 is still the highest number in the directory), the new `mt7988_bananapi_bpi-r4-pro-8x-*` entries sort well below `mt7981_keenetic_kn-1012` in `UBOOT_TARGETS`, and `package/boot/uboot-mediatek/Makefile` carries no `PKG_VERSION`/`PKG_HASH` change, so no patch refresh is owed.

The two long-standing open threads carry over unchanged, on the same footing as the previous six rounds: the U-Boot FM25G02B `NAND_ECCREQ(8, 528)` and the single-region `ooblayout_free`. I checked again against the reference tree available here rather than just repeating the previous note — [`drivers/mtd/nand/spi/fmsh.c` at v6.18.44](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mtd/nand/spi/fmsh.c#L53-L63) still contains only the `FM25S01A` entry with its `NAND_ECCREQ(1, 512)`, and `d5a5c9eb2ee9` is not a valid object there, so the "ported from upstream" claim remains neither confirmed nor refuted from my side. Noting it only so it stays visible at merge time, not as an objection.

CI is green on `7bf0b47`: 22 check runs, all completed, none failed — mediatek/filogic kernel build and kernel-patch check, both external-toolchain package builds (x86/64, malta/be) and all three FormalityCheck jobs included; the only non-success conclusions are two `skipped` ccache-upload steps.

---

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
```diff
@@ -179,8 +179,10 @@ case "$board" in
 	keenetic,kap-630|\
 	keenetic,kn-3711|\
 	keenetic,kn-3811|\
+	keenetic,kn-1012|\
```
openwrt-ai: nit: `kn-1012` sorts before `kn-3711`, not between `kn-3811` and `kn-3911`; every other case list in this file is in board-name order. The `netcraze,nc-1012` placement is already correct.

```suggestion
	keenetic,kn-1012|\
	keenetic,kn-3711|\
	keenetic,kn-3811|\
```

---

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
```diff
@@ -179,8 +179,10 @@ case "$board" in
...
 	keenetic,kn-3911|\
-	netcraze,nap-630)
+	netcraze,nap-630|\
+	netcraze,nc-1012)
```
openwrt-ai, line 203: nit: this belongs squashed into `7bf0b47`, the immediately preceding commit that introduces both boards — as a separate commit it leaves a bisect point where phy1 falls back to `MT_EE_MAC_ADDR2` (rf-eeprom `0xa`) and collides with the WAN address the DTS takes from [`macaddr_factory_a`](https://github.com/openwrt/openwrt/blob/80626bf6d5d22636285c7332e5eb759963ab41b9/target/linux/mediatek/dts/mt7981b-keenetic-kn-1012.dtsi#L381-L384). That commit's MAC table also still reads `5g rf-eeprom 0xa, the second MAC slot of the WiFi EEPROM`, which stops being true once this override lands; squashing lets both be corrected in one place.

---

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
```diff
@@ -179,8 +179,10 @@ case "$board" in
 	keenetic,kap-630|\
 	keenetic,kn-3711|\
 	keenetic,kn-3811|\
+	keenetic,kn-1012|\
 	keenetic,kn-3911|\
-	netcraze,nap-630)
+	netcraze,nap-630|\
```
openwrt-ai, line 202: nit: the PR body still states "#24623 already carries both board names", but that PR's head (`284fee8`) only rewrites the derivation — its hunk keeps `keenetic,kn-3811|keenetic,kn-3911|netcraze,nap-630` as untouched context, so this commit is now the only place the two names are added. #24623's own commit message still claims "KN-1012 and NC-1012 join the same list", which is worth fixing on that side so the pair isn't read as duplicating each other.

---

openwrt-ai (commented):
## Commit checks

- `80626bf` "mediatek: filogic: derive the KN-1012 phy1 MAC like its siblings" — the message's premise, that the family entry "keeps phy1 out of both the VIF-allocator and MBSSID address schemes", is only true after #24623. In this tree the entry is still `macaddr_setbit_la "$(mtd_get_mac_binary rf-eeprom 0x4)"`, which per #24623's own analysis is exactly the address the VIF allocator hands the second 2.4 GHz VIF — so standing alone this commit puts the pair *on* the collision, not out of it. Reword to say it joins the entry where that fix will land, and name the #24623 dependency.

---

target/linux/mediatek/dts/mt7981b-keenetic-kn-1012.dtsi
openwrt-ai: fixed, thanks

---

openwrt-ai (commented):

---

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
openwrt-ai: fixed, thanks

---

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
openwrt-ai: fixed, thanks — squashed, and the MAC table now reads `5g rf-eeprom 0x4 with the locally administered bit set`.

---

target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac
openwrt-ai: fixed, thanks — solved on this side by rewording the PR body to say the boards join the entry here; the #24623 commit message is still the other half.

---

openwrt-ai (commented):
## Commit checks

- `b526324`, `68c2660`, `4b5ed71` — the new `Assisted-by: LLM` trailer names no tool or model; all 44 `Assisted-by:` trailers on `main` are specific (`Assisted-by: Claude:claude-opus-5`, `Assisted-by: ChatGPT (OpenAI GPT-5.6 Sol)`). Use the model-specific form; it carries no email either, so `check_noreply_email` still passes.

---

openwrt-ai (commented):
## Commit checks

- `c8ba844`, `2d81c37`, `5982f28` — `Assisted-by: LLM` still names no tool or model; every `Assisted-by:` trailer on `main` is specific (`Assisted-by: Claude:claude-opus-5`, `Assisted-by: Claude Code (Anthropic Claude Sonnet 5)`). Use the model-specific form.

---

target/linux/mediatek/dts/mt7981b-keenetic-kn-1012.dtsi
```diff
@@ -0,0 +1,462 @@
...
+	 * switch node instead, but pio 0 is the PHY's own line and the switch
+	 * has its own on pio 38. The vendor also declares it level low; edge
+	 * falling matches every other EN8811H in the tree and survives the
+	 * driver clearing the source without masking it.
+	 */
+	phy13: ethernet-phy@d {
+		reg = <0xd>;
+		airoha,pnswap-rx;
+		/* firmware from the rootfs, mounted after the switch probes */
+		needs-host-firmware;
```
openwrt-ai, line 230: Nothing reads `needs-host-firmware` — not this tree, and not [`air_en8811h.c` / `phy_device.c` in 6.18.44](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/phy/air_en8811h.c), this target's `KERNEL_PATCHVER` — so on `main` it is an inert line, and the commit message documents the pnswap and INTB quirks on this same node but not this one. Name the #24819 dependency in the commit body so a later reader isn't left guessing whether it is dead or broken. It is also an unprefixed property with no binding; a vendor prefix would keep an OpenWrt-local flag out of the generic DT namespace.

---

target/linux/mediatek/dts/mt7981b-keenetic-kn-1012.dtsi
```diff
@@ -0,0 +1,462 @@
...
+			gpio-export,output = <0>;
+			gpios = <&pio 25 GPIO_ACTIVE_LOW>;
+		};
+
+		/*
+		 * Selects which media reaches MT7531 port 5: a low pin routes it
+		 * to the SFP cage, a high pin to the EN8811H copper PHY. The SFP
+		 * device tree flips this default. gpio-export writes the value
+		 * raw, so it is the pin level and not the logical one. Exported
+		 * so the active path can be verified and forced from userspace.
```
openwrt-ai, line 128: nit: only the *initial* value is raw. sysfs goes through [`gpiod_get_value_cansleep()` / `gpiod_set_value_cansleep()` in gpiolib-sysfs.c:172](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/gpio/gpiolib-sysfs.c#L172), which does apply `GPIO_ACTIVE_LOW` — so `sfp1mux/value` reads 0 for the copper default this line sets to 1, and forcing SFP from userspace means writing 1, not 0. Worth one clause here, since the sentence right after invites exactly that.

```suggestion
		 * raw, so it is the pin level and not the logical one. Exported
		 * so the active path can be verified and forced from userspace,
		 * where the sysfs value is logical and thus inverted: 1 is SFP.
```

---

openwrt-ai (commented):
Reviewed 3 new commits.

---

lexfrei:
Rebased on current main.

To be explicit about the related PRs in the description: none of them blocks this one. Without them the board still boots, bridges and runs wifi, the pieces they fix are just off. The only real interaction is #24623, which edits the same case block in the wifi MAC fixup, so whichever of the two lands second needs a rebase.

openwrt-ai (commented):

---

lexfrei:
@jonasjelonek dropped the redundant `&watchdog` override here too, same as on #23727. Rebased on current main.

None of the related PRs in the description blocks this one. Without them the board still boots, bridges and runs wifi, the pieces they fix are just off.

openwrt-ai (commented):

---

openwrt-ai (commented):
## Commit checks

- `6d7351d` "mediatek: filogic: add Keenetic KN-1012 and Netcraze NC-1012" — the rebase brought in `cc96ed1e6dc`, so the entry these boards join now computes `(octet0 | 0x2) ^ 0x80`; the table's `5g  rf-eeprom 0x4 with the locally administered bit set` no longer describes it (LA-only would give `d6:9c:…`, the PR body measures `56:9c:…`). Update that line, and drop "The derivation that entry uses is being reworked for the whole family separately" — that rework is in the tree this now sits on.

---

openwrt-ai (commented):

---

target/linux/mediatek/dts/mt7981b-keenetic-kn-1012.dtsi
lexfrei: Done, the comment now says the sysfs value is logical and 1 selects the SFP cage.

target/linux/mediatek/dts/mt7981b-keenetic-kn-1012.dtsi
lexfrei: Dropped from this PR. The property only means something once the phylink change that reads it lands, so it moves there together with its consumer.

target/linux/mediatek/dts/mt7981b-keenetic-kn-1012.dtsi
```diff
@@ -0,0 +1,457 @@
...
+	/*
+	 * Airoha EN8811H. The vendor tree declares this PHY's INTB on the
+	 * switch node instead, but pio 0 is the PHY's own line and the switch
+	 * has its own on pio 38. The vendor also declares it level low; edge
+	 * falling matches every other EN8811H in the tree and survives the
+	 * driver clearing the source without masking it.
+	 */
+	phy13: ethernet-phy@d {
+		reg = <0xd>;
+		airoha,pnswap-rx;
```
openwrt-ai, line 229: nit: the PR body's #24819 bullet still says "The copper PHY node here carries `needs-host-firmware`, which is the property that patch reads; a kernel without it ignores the line" — that property is gone from this node now. Reword that bullet to match.

---

openwrt-ai (commented):
Reviewed 3 new commits.

---

target/linux/mediatek/dts/mt7981b-keenetic-kn-1012.dtsi
lexfrei: Done, the bullet now says the property arrives with #24819.

lexfrei:
@jonasjelonek rebased on current main, the redundant watchdog override is gone and the two remaining bot points are closed. Anything else needed here before it can go in?

target/linux/mediatek/dts/mt7981b-keenetic-kn-1012.dtsi
```diff
@@ -0,0 +1,457 @@
...
+		interrupts = <38 IRQ_TYPE_LEVEL_HIGH>;
+	};
+};
+
+&switch {
+	ports {
+		#address-cells = <1>;
+		#size-cells = <0>;
+
+		port@0 {
```
jonasjelonek: Possible to use `ethernet-ports` and `ethernet-port` here? Upstream allows both and doesn't seem to prefer one explicitly in the bindings, however I guess the more specific variant is preferred implicitly.

Applies to all other DTS additions here.

target/linux/mediatek/image/filogic.mk
```diff
@@ -2396,6 +2396,153 @@ define Device/keenetic_kap-630
...
+
+define Build/kn1012-fit
+	( \
+		printf '/dts-v1/;\n/ {\n'; \
+		printf '\tdescription = "OpenWrt FIT (dual media)";\n'; \
+		printf '\t#address-cells = <1>;\n\timages {\n'; \
```
jonasjelonek: Can you explain why this is needed? Because I don't get it yet and it seems quite strange to build kind of a DT structure here in the recipe with prints.

target/linux/mediatek/dts/mt7981b-keenetic-kn-1012.dtsi
lexfrei: Switched to `ethernet-ports` with `ethernet-port@N` children, here and in the other DTS files this PR adds.

target/linux/mediatek/image/filogic.mk
lexfrei: The printf ITS is gone. I wrote it by hand without checking what the tree already had, and the recipe now uses the standard `fit` step with `DEVICE_DTS_OVERLAY`, like rax3000m does, so mkits.sh builds the kernel FIT with the two overlays and the chainloader is just prepended.

One hand-built FIT is left, the wrapper that carries the chainloaded U-Boot. The stock bootloader only starts a payload shaped like the vendor image, node names `kernel@1` / `fdt@1` / `config@1` and the description "Linux 4.9" included, and mkits.sh hardcodes its own names and descriptions. It is a fixed 3-node blob, nothing per-device.

lexfrei:
@jonasjelonek pushed the rework. The hand-written ITS is gone, the recipe uses the standard `fit` step with `DEVICE_DTS_OVERLAY`, and the switch node uses `ethernet-ports`. Please take another look.

package/boot/uboot-mediatek/patches/409-arm-mediatek-default-the-kernel-image-header-text-offset-base.patch
```diff
@@ -0,0 +1,30 @@
...
+
+Submitted upstream:
+https://lore.kernel.org/u-boot/20260920151512.401899-2-f@lex.la/
+
+Assisted-by: LLM
+Signed-off-by: Aleksei Sviridkin <f@lex.la>
+---
+--- a/arch/arm/mach-mediatek/Kconfig
++++ b/arch/arm/mach-mediatek/Kconfig
+@@ -215,4 +215,7 @@ config MTK_TZ_MOVABLE
```
openwrt-ai: This hunk is anchored against pristine U-Boot, not the quilt-patched tree: [`301-mt7622-generic-reset-button-ignore-env.patch`](``https://github.com/openwrt/openwrt/blob/8ec12cd963fea1d64393730f50b9756abdd56aea/package/boot/uboot-mediatek/patches/301-mt7622-generic-reset-button-ignore-env.patch#L46-L58)`` already inserts 7 lines between `bool` and `endif`, so `select OF_SYSTEM_SETUP / bool / <blank> / endif` no longer exists anywhere in the file. Applying 301 then this one, `patch` reports `Hunk #1 succeeded at 173 with fuzz 2 (offset -42 lines)` and drops the symbol inside the nested `if ARM64` block rather than at the end of `if ARCH_MEDIATEK`; with `-F0` it fails outright. Regenerate with `make package/uboot-mediatek/refresh` — 110-, 112-, 113- and 504- all apply at fuzz 0, so this is the only one affected.

---

openwrt-ai (commented):
Reviewed 4 new commits.

---

openwrt-ai (commented):

---

package/boot/uboot-mediatek/patches/409-arm-mediatek-default-the-kernel-image-header-text-offset-base.patch
openwrt-ai: fixed, thanks

---

openwrt-ai (commented):

---

lexfrei:
Rebased on current main. The U-Boot kernel image header patch is no longer headed upstream, so 409 now says it is kept local, and the chainload commit message says the same. I also asked on netdev whether the phy_port work is meant to cover a combo port like this one. The description has both.

target/linux/mediatek/image/filogic.mk
```diff
@@ -2396,6 +2396,91 @@ define Device/keenetic_kap-630
...
+		printf '\t\tconfig@1 {\n'; \
+		printf '\t\t\tdescription = "NDMS";\n'; \
+		printf '\t\t\tkernel = "kernel@1";\n'; \
+		printf '\t\t\tfdt = "fdt@1";\n\t\t};\n'; \
+		printf '\t};\n};\n'; \
+	) > $@.uboot.its
+	PATH=$(LINUX_DIR)/scripts/dtc:$(PATH) mkimage -f $@.uboot.its $@.uboot.itb
+	mv $@ $@.payload
+	mv $@.uboot.itb $@
+	$(call Build/append-squashfs4-fakeroot)
```
openwrt-ai, line 2440: nit: the rework shortened the comment block above this recipe to cover only the hand-written ITS; the part explaining why a fake squashfs is glued to the chainloader FIT before the 1 MiB pad and the real FIT is gone. Add one line here so the step stays readable — unlike the kn-3811 recipe, this squashfs does not end up at the end of the kernel partition.

---

openwrt-ai (commented):
Reviewed 4 new commits.

---

PR #24821, opened 2026-08-20
realtek: pse: replace pending with backport patches

jonasjelonek:
The Realtek PSE MCU driver patches have been accepted upstream, being available in v7.3. Replace the current pending patches with the backported variants. This requires some additional DTS changes since the final version differs from what was merged as pending.

### DIFF
--- a/target/linux/realtek/dts/rtl9301_zyxel_xgs1930-28hp.dts
+++ b/target/linux/realtek/dts/rtl9301_zyxel_xgs1930-28hp.dts
@@ -176,13 +176,6 @@
 		tx-fault-gpio = <&gpio1 28 GPIO_ACTIVE_HIGH>;
 		los-gpio = <&gpio1 24 GPIO_ACTIVE_HIGH>;
 	};
-
-	poe_power: regulator-poe {
-		compatible = "regulator-fixed";
-		regulator-name = "poe-power";
-		gpio = <&gpio1 12 GPIO_ACTIVE_HIGH>;
-		enable-active-high;
-	};
 };
 
 &gpio0 {
@@ -204,8 +197,8 @@
 			compatible = "zyxel,xgs1930-28hp-pse", "realtek,pse-mcu-gen2-smbus";
 			reg = <0x20>;
 
-			enable-gpios = <&gpio0 19 GPIO_ACTIVE_HIGH>;
-			power-supply = <&poe_power>;
+			disable-ports-gpios = <&gpio1 12 GPIO_ACTIVE_LOW>;
+			reset-gpios = <&gpio0 19 GPIO_ACTIVE_LOW>;
 
 			pse-pis {
 				#address-cells = <1>;
--- a/target/linux/realtek/patches-6.18/817-net-pse-pd-realtek-pse-mcu-add-per-por-legacy-detec.patch
+++ b/target/linux/realtek/patches-6.18/817-net-pse-pd-realtek-pse-mcu-add-per-por-legacy-detec.patch
@@ -1,4 +1,4 @@
-From b69e39ba5853c587a594ed0475f873519301f326 Mon Sep 17 00:00:00 2001
+From bbb5e80bbaf0d73b73c4bcd85c7941439436ef5c Mon Sep 17 00:00:00 2001
 From: Jonas Jelonek <jelonek.jonas@gmail.com>
 Date: Tue, 30 Jun 2026 20:46:30 +0000
 Subject: [PATCH] net: pse-pd: realtek-pse-mcu: add per-port legacy-detection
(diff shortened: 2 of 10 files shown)


target/linux/realtek/dts/rtl9301_zyxel_xgs1930-28hp.dts
```diff
@@ -176,13 +176,6 @@
 		tx-fault-gpio = <&gpio1 28 GPIO_ACTIVE_HIGH>;
 		los-gpio = <&gpio1 24 GPIO_ACTIVE_HIGH>;
 	};
-
-	poe_power: regulator-poe {
```
openwrt-ai, line 180: Dropping this node (and the one in `rtl9313_zyxel_xs1930-12hp.dts`) removes the last `regulator-fixed` user in the whole realtek target, and the upstream driver no longer includes `<linux/regulator/consumer.h>` — after this PR `git grep -l 'regulator-fixed\|devm_regulator_get\|regulator_enable' target/linux/realtek/` comes back empty.

`CONFIG_REGULATOR=y` / `CONFIG_REGULATOR_FIXED_VOLTAGE=y` were added to all six subtarget configs by exactly the commit that introduced the regulator-based version of this driver ([301d98f81fa "realtek: add Realtek PSE MCU driver"](https://github.com/openwrt/openwrt/commit/301d98f81fa7baedcb172f7d47c05b12770302a5)), so they now look dead:

```
target/linux/realtek/rtl838x/config-6.18:225-226
target/linux/realtek/rtl839x/config-6.18:235-236
target/linux/realtek/rtl930x/config-6.18:226-227
target/linux/realtek/rtl930x_nand/config-6.18:235-236
target/linux/realtek/rtl931x/config-6.18:231-232
target/linux/realtek/rtl931x_nand/config-6.18:239-240
```

`REGULATOR` is a user-selectable menuconfig symbol, so `kernel_oldconfig` will not drop it by itself. Worth removing those two lines from the six subtarget configs in this commit so the revert of the regulator dependency is complete.

---

target/linux/realtek/patches-6.18/817-net-pse-pd-realtek-pse-mcu-add-per-por-legacy-detec.patch
```diff
@@ -19,11 +19,11 @@ default. Each port gets a sysfs directory with a detection_legacy file:
...
 
-The detection-type command and its value encoding are dialect-specific, so
-both are taken from the dialect: the command opcode (0x09 on Realtek, 0x10
-on Broadcom) goes through the opcode table, and the standard/legacy byte
+The detection-type command and its value encoding are dialect-specific,
+so both are taken from the dialect: the command opcode (0x10 on Gen1,
+0x09 on Gen2) goes through the opcode table, and the standard/legacy byte
 values come from the dialect too. These differ - standard detection is 0x0
-on Realtek but 0x2 on Broadcom, where 0x0 instead disables detection - so a
+on Gen2 but 0x2 on Gen1, where 0x0 instead disables detection - so a
```
openwrt-ai, line 26: nit: the commit body here was retermed from Realtek/Broadcom to Gen2/Gen1, but the kerneldoc comment this same patch adds still uses the old wording — lines 73-74 of this file read:

```
	 * The encoding is dialect-specific: standard detection is 0x0 on RTL
	 * but 0x2 on BCM (where 0x0 instead disables detection entirely), so
```

That is now actively misleading, because the gen1 dialect covers RTL8238B as well as BCM59111/BCM59121 (see the `rtl8238b_info`/`bcm5911x_info` chip table in 816-01), so the split is per-dialect, not per-silicon-vendor. Suggest changing "on RTL"/"on BCM" to "on Gen2"/"on Gen1" to match the reworded body.

---

openwrt-ai (commented):
Reviewed 1 new commit. Commit message matches the diff, and CI on `06702b0` is green (including "Check Kernel patches").

Verified locally against `linux-6.18.44`: the whole 816-01 → 819 series applies with zero fuzz in order, the `PSE_REALTEK_MCU*` Kconfig symbols are unchanged from the pending version (so `package/kernel/linux/modules/pse-pd.mk` needs no update), and `kzalloc_obj(*range)` without an explicit GFP is valid on 6.18 (`include/linux/slab.h`, defaults to `GFP_KERNEL`).

The GPIO polarity conversions all preserve the previous electrical behaviour on the four affected boards — old `enable-gpios` (`GPIO_ACTIVE_HIGH` + `GPIOD_OUT_HIGH`) and new `reset-gpios` (`GPIO_ACTIVE_LOW` + `GPIOD_OUT_LOW`) both leave the line physically high, and the same holds for the `poe_power` regulator → `disable-ports-gpios` swap on `xgs1930-28hp` (`&gpio1 12`) and `xs1930-12hp` (`&gpio0 30`). No sister board was missed: the other `realtek,pse-mcu-*` DTS files carry no `enable-gpios` at all.

Two inline notes.

---

target/linux/realtek/dts/rtl9301_zyxel_xgs1930-28hp.dts
jonasjelonek: Will be kept for future changes, especially wiring up `vpwr-supply` on `pse-pis` nodes. Also for #24376, which does exactly this.

target/linux/realtek/patches-6.18/817-net-pse-pd-realtek-pse-mcu-add-per-por-legacy-detec.patch
jonasjelonek: > because the gen1 dialect covers RTL8238B as well as BCM59111/BCM59121 (see the `rtl8238b_info`/`bcm5911x_info` chip table in 816-01)

That is wrong. Gen1 covers setups with BCM PSE, Gen2 covers setups with RTL PSE.

Anyway, fixed the wording.

openwrt-ai (commented):
Re-reviewed after the force-push (`06702b0` → `e194a34`). The only delta is the kerneldoc wording fix in `817-...-per-por-legacy-detec.patch` ("on RTL"/"on BCM" → "on Gen2"/"on Gen1"), which now matches the reworded commit body and the rest of the series — `git grep -E 'on (RTL|BCM)\b'` over the PSE patches comes back empty, and 816-01 consistently frames Gen1 as the Broadcom-PSE dialect and Gen2 as the Realtek-PSE one. Thanks for the correction on my earlier misreading of the chip table; the dialect split is per-generation as you described.

The two replaced lines keep the hunk line counts identical, so the patch metadata is unaffected. CI on `e194a34` is fully green (Check Kernel patches, all six subtarget kernel builds, FormalityCheck).

---

PR #24825, opened 2026-08-21
econet: en751221: add support for Nokia G-140W-F

AKoo7:
## Description

Add support for the **Nokia G-140W-F**, an ISP-provided GPON ONT / home
router based on the EcoNet EN751221 (EN7526G).

It is the same OEM platform as the already-supported **Nokia G-240G-E** —
the SPI-NAND partition layout is identical (verified byte-for-byte against
a full per-partition dump of a unit) — so the board support closely mirrors
that device, with WiFi wired up the same way as the SmartFiber XP8421-B
(which carries the same two radios).

### What works
- Boot from NAND, `tclinux` slot
- Ethernet LAN — the four RJ45 jacks are aggregated behind the on-die
  switch and come up as a single bridged `eth0`, the same way the other
  EN751221 boards present their ports (this subtarget has no DSA switch
  driver yet)
- 2.4 GHz WiFi (MT7603E) and 5 GHz WiFi (MT7662), calibrated from on-chip
  eFuse; per-radio MAC from the `ri` MAC base
- USB (xHCI)

### Not included
- **GPON WAN** — the EN7528/EN751221 xPON driver is not in tree yet, so the
  device is contributed as a LAN/WiFi board. WAN can be enabled later once
  that driver lands, without changing this board file.

## Hardware
| | |
|---|---|
| SoC | EcoNet EN751221 (EN7526G) |
| RAM | 256 MiB DDR3 |
| Flash | 256 MiB SPI NAND (Micron MT29F2G01) |
| WiFi | MT7603E (2.4 GHz) + MT7662 (5 GHz) |
| Ethernet | 4x 10/100/1000 |
| USB | 1x USB port |
| Telephony | 1x FXS (RJ11) phone jack |

## Installation
Serial console (115200 8n1); interrupt the bootloader and use its built-in
TFTP server to load an OpenWrt initramfs image into RAM (any file name
other than the vendor's `tcboot.bin` / `tclinux.bin`), boot it, then write
the sysupgrade image to the `tclinux` partition and reboot.

## Testing
- Full image build from a clean tree against `main` (`make -j12`,
  kernel 6.18.44) — builds cleanly, produces
  `openwrt-econet-en751221-nokia_g140w-f-squashfs-tclinux.trx`.
- DTB compiles with no new warnings.
- Board brings up LAN + both WiFi radios on real hardware.

### DIFF
--- a/target/linux/econet/dts/en751221_nokia_g140w-f.dts
+++ b/target/linux/econet/dts/en751221_nokia_g140w-f.dts
@@ -1,0 +1,163 @@
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+/dts-v1/;
+
+#include "en751221.dtsi"
+
+/ {
+	model = "Nokia G-140W-F";
+	compatible = "nokia,g140w-f", "econet,en751221";
+
+	memory@0 {
+		device_type = "memory";
+		reg = <0x00000000 0x10000000>;
+	};
+
+	chosen {
+		stdout-path = "/serial@1fbf0000:115200";
+		linux,usable-memory-range = <0x00020000 0x0ffe0000>;
+	};
+};
+
+&nand {
+	status = "okay";
+	econet,bmt;
+
+	partitions {
+		compatible = "fixed-partitions";
+		#address-cells = <1>;
+		#size-cells = <1>;
+
+		partition@0 {
+			label = "bootloader";
+			reg = <0x0 0x40000>;
+			read-only;
+		};
+
+		partition@40000 {
+			label = "romfile";
+			reg = <0x40000 0x40000>;
+		};
+
+		partition@80000 {
+			label = "tclinux";
+			reg = <0x80000 0x2200000>;
+			econet,enable-remap;
+		};
+
+		partition@480000 {
+			label = "rootfs";
+			reg = <0x480000 0x1e00000>;
+			linux,rootfs;
+		};
+
+		partition@2280000 {
+			label = "kernel_slave";
+			reg = <0x2280000 0x300000>;
+		};
+
+		partition@2580000 {
+			label = "rootfs_slave";
+			reg = <0x2580000 0x1f00000>;
+		};
+
+		partition@4480000 {
+			label = "kernel_oflt";
+			reg = <0x4480000 0x300000>;
+		};
+
+		partition@4780000 {
+			label = "rootfs_oflt";
+			reg = <0x4780000 0xc00000>;
+		};
+
+		partition@5380000 {
+			label = "config";
+			reg = <0x5380000 0x800000>;
+		};
+
+		partition@5b80000 {
+			label = "log";
+			reg = <0x5b80000 0xc00000>;
+		};
+
+		partition@6780000 {
+			label = "extfs";
+			reg = <0x6780000 0x600000>;
+		};
+
+		partition@6d80000 {
+			label = "bosa";
+			reg = <0x6d80000 0x40000>;
+		};
+
+		partition@6dc0000 {
+			label = "flag";
+			reg = <0x6dc0000 0x40000>;
+		};
+
+		partition@6e00000 {
+			label = "flagback";
+			reg = <0x6e00000 0x40000>;
+		};
+
+		partition@6e40000 {
+			label = "ri";
+			reg = <0x6e40000 0x40000>;
+
+			nvmem-layout {
+				compatible = "fixed-layout";
+				#address-cells = <1>;
+				#size-cells = <1>;
+
+				macaddr_ri_3e: macaddr@3e {
+					compatible = "mac-base";
+					reg = <0x3e 0x6>;
+					#nvmem-cell-cells = <1>;
+				};
+			};
+		};
+
+		partition@6e80000 {
+			label = "riback";
+			reg = <0x6e80000 0x40000>;
+		};
+	};
+};
+
+&gmac0 {
+	status = "okay";
+	nvmem-cells = <&macaddr_ri_3e 0>;
+	nvmem-cell-names = "mac-address";
+};
+
+/*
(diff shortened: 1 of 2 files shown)


target/linux/econet/image/en751221.mk
```diff
@@ -54,6 +54,16 @@ define Device/nokia_g240g-e
 endef
 TARGET_DEVICES += nokia_g240g-e
 
+define Device/nokia_g140w-f
```
openwrt-ai, line 57: nit: the device blocks in this file are in strict alphabetical order (`chinamobile_gs3101`, `en751221_generic`, `huawei_hg2821t-u`, `nokia_g240g-e`, `smartfiber_xp8421-b`, `tplink_archer-vr1200v-v2`, `zyxel_pmg5617ga`). `nokia_g140w-f` sorts before `nokia_g240g-e`, so this block belongs above the `define Device/nokia_g240g-e` block at line 47 rather than after it.

---

target/linux/econet/dts/en751221_nokia_g140w-f.dts
```diff
@@ -0,0 +1,163 @@
...
+
+		partition@6d80000 {
+			label = "bosa";
+			reg = <0x6d80000 0x40000>;
+		};
+
+		partition@6dc0000 {
+			label = "flag";
+			reg = <0x6dc0000 0x40000>;
+		};
```
openwrt-ai, line 96: nit: this board declares the same `flag` / `flagback` pair and the same dual-slot layout (`tclinux` + `kernel_slave`/`rootfs_slave`) as the G-240G-E, but is not added to the boot-flag switcher. `en75_chboot` has a per-board case list and the [`nokia,g240g-e)` case at en75_chboot:122-129](https://github.com/openwrt/openwrt/blob/cb4561496513e8586f0914691e2f35a9da58d518/target/linux/econet/base-files/sbin/en75_chboot#L122-L129) keys off exactly this `flag` partition. Since the commit message states the SPI-NAND layout is byte-for-byte identical to the G-240G-E, is the omission intentional, or should `nokia,g140w-f` be folded into that case (e.g. `nokia,g140w-f|\` above `nokia,g240g-e)`)? Without it there is no in-tree way to flip back to the vendor OS slot.

---

target/linux/econet/dts/en751221_nokia_g140w-f.dts
```diff
@@ -0,0 +1,163 @@
...
+	memory@0 {
+		device_type = "memory";
+		reg = <0x00000000 0x10000000>;
+	};
+
+	chosen {
+		stdout-path = "/serial@1fbf0000:115200";
+		linux,usable-memory-range = <0x00020000 0x0ffe0000>;
+	};
+};
```
openwrt-ai, line 19: nit: no `gpio-leds` / `gpio-keys-polled` nodes are described, so the front-panel LEDs and the reset button on this ONT are unusable. Other econet boards do wire these up — see [`en7526f_chinamobile_gs3101.dts:24`](https://github.com/openwrt/openwrt/blob/cb4561496513e8586f0914691e2f35a9da58d518/target/linux/econet/dts/en7526f_chinamobile_gs3101.dts#L24-L60) and [`en7528_dasan_h660gm-a.dtsi:28`](https://github.com/openwrt/openwrt/blob/cb4561496513e8586f0914691e2f35a9da58d518/target/linux/econet/dts/en7528_dasan_h660gm-a.dtsi#L28). The G-240G-E this file is derived from also omits them, and the commit message says nothing about LEDs/buttons either way — is the GPIO map simply not known yet, or was this left out deliberately?

---

openwrt-ai (commented):
## Commit checks

- `cb456149` "econet: en751221: add support for Nokia G-140W-F" — this is a new-device commit (new `.dts` under `target/linux/econet/dts/` plus a new `define Device/...` in `target/linux/econet/image/`), so the message needs the three standard sections. Two are there (Specifications, Installation), but:
  - **MAC address layout is missing entirely.** The Notes only say "a per-radio MAC is supplied via nvmem, derived from the 'ri' MAC base". The DTS actually encodes a concrete layout — `ri` partition at offset `0x3e`, `mac-base` with index 0 → `gmac0`, 1 → 2.4 GHz radio, 2 → 5 GHz radio. Please spell that out, one line per interface, using `xx:xx:xx` for the unit-specific bytes.
  - **Specifications is missing LEDs/Buttons (count + type), the UART header location/pinout, and power (voltage + current).** The baud rate is covered in Installation, but the header itself isn't located.
  - **Installation is light on the verbatim details.** It says to TFTP "an OpenWrt initramfs image" and then "write the sysupgrade image to the 'tclinux' partition", but doesn't name the actual artifacts this device builds (only `tclinux.trx` is produced — `IMAGES := tclinux.trx`, and `platform_check_image`/`platform_do_upgrade` in `target/linux/econet/base-files/lib/upgrade/platform.sh` do not cover this board), nor the bootloader TFTP commands / IP setup, nor the command used to write the partition. The bar in the project rules is "could a stranger flash this device from the commit message alone?" — as written they'd have to guess. A reverting-to-stock note would also help, given the vendor OS stays intact in the `kernel_slave`/`rootfs_slave` slots.

Everything else lines up: the DTS is a faithful copy of `en751221_nokia_g240g-e.dts` plus the two PCIe radios wired the way `en751221_smartfiber_xp8421-b.dts` does, `DEVICE_MODEL`/`model`/`compatible`/`DEVICE_DTS` are consistent, and `DEVICE_PACKAGES := kmod-usb3 kmod-mt7603 kmod-mt76x2` matches MT7603E + MT7662. The "no EEPROM partition, calibrate from eFuse" claim checks out against the driver — `mt7603_eeprom_init()` falls through to `mt7603_efuse_init()` when `mt76_eeprom_init()` finds nothing, and mt76x2 has the equivalent path (Linux 6.18.44, the target's `KERNEL_PATCHVER`).

---

PR #24827, opened 2026-08-21
qualcommbe: fix 1000BASE-X SFP support on IPQ9574

BalsamCHIHI:
Add a kernel patch to fix an oversight in the out-of-tree qcom_ppe driver where PHY_INTERFACE_MODE_1000BASEX was missing from the MAC clock rate configuration switch statement.

Without this patch, inserting a 1G fiber SFP module into the 10G MAC throws an -EOPNOTSUPP (-95) error during phylink negotiation, preventing the link from coming up. This patch routes 1000BASE-X to the GMII clock helper, restoring native 1G SFP module hot-swapping on IPQ9574 boards like the 8devices Kiwi-DVK.

### DIFF
--- a/target/linux/qualcommbe/patches-6.18/0342-net-qualcomm-Update-IPQ9574-PPE-driver.patch
+++ b/target/linux/qualcommbe/patches-6.18/0342-net-qualcomm-Update-IPQ9574-PPE-driver.patch
@@ -10,6 +10,8 @@ issue for the performance improvement.
 Signed-off-by: Luo Jie <quic_luoj@quicinc.com>
 Alex G: print error code when failing to connect phylink
 Signed-off-by: Alexandru Gagniuc <mr.nuke.me@gmail.com>
+Balsam CHIHI: route 1000BASE-X to GMII clock helper for 1G SFP support
+Signed-off-by: Balsam CHIHI <balsam.chihi@moment.tech>
 ---
  drivers/net/ethernet/qualcomm/ppe/Makefile    |   2 +-
  drivers/net/ethernet/qualcomm/ppe/edma.c      | 158 +---
@@ -2263,8 +2265,7 @@ Signed-off-by: Alexandru Gagniuc <mr.nuke.me@gmail.com>
  {
 -	struct ppe_device *ppe_dev = ppe_port->ppe_dev;
 -	int ret;
-+	int ret, i;
- 
+-
 -	ret = reset_control_assert(ppe_port->rstcs[PPE_PORT_CLK_RST_MAC]);
 -	if (ret)
 -		goto error;
@@ -2283,7 +2284,8 @@ Signed-off-by: Alexandru Gagniuc <mr.nuke.me@gmail.com>
 -	ret = reset_control_deassert(ppe_port->rstcs[PPE_PORT_CLK_RST_MAC]);
 -	if (ret)
 -		goto error;
--
++	int ret, i;
+ 
 -	ret = reset_control_deassert(ppe_port->rstcs[PPE_PORT_CLK_RST_RX]);
 -	if (ret)
 -		goto error;
@@ -2353,7 +2355,7 @@ Signed-off-by: Alexandru Gagniuc <mr.nuke.me@gmail.com>
  	if (ret)
  		goto err_mac_config;
  
-@@ -676,8 +702,111 @@ static void ppe_port_mac_config(struct p
+@@ -676,8 +702,112 @@ static void ppe_port_mac_config(struct p
  	return;
  
  err_mac_config:
@@ -2432,6 +2434,7 @@ Signed-off-by: Alexandru Gagniuc <mr.nuke.me@gmail.com>
 +	case PHY_INTERFACE_MODE_SGMII:
 +	case PHY_INTERFACE_MODE_QSGMII:
 +	case PHY_INTERFACE_MODE_PSGMII:
++	case PHY_INTERFACE_MODE_1000BASEX:
 +	case PHY_INTERFACE_MODE_2500BASEX:
 +		rate = ppe_port_mac_clock_rate_get_gmii(speed);
 +		break;
@@ -2467,7 +2470,7 @@ Signed-off-by: Alexandru Gagniuc <mr.nuke.me@gmail.com>
  }
  
  /* PPE port GMAC link up configuration */
-@@ -688,6 +817,11 @@ static int ppe_port_gmac_link_up(struct
+@@ -688,6 +818,11 @@ static int ppe_port_gmac_link_up(struct
  	int ret, port = ppe_port->port_id;
  	u32 reg, val;
  
@@ -2479,7 +2482,7 @@ Signed-off-by: Alexandru Gagniuc <mr.nuke.me@gmail.com>
  	/* Set GMAC speed */
  	switch (speed) {
  	case SPEED_1000:
-@@ -700,8 +834,8 @@ static int ppe_port_gmac_link_up(struct
+@@ -700,8 +835,8 @@ static int ppe_port_gmac_link_up(struct
  		val = GMAC_SPEED_10;
  		break;
  	default:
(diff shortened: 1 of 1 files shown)


robimarko:
Instead of a new patch, please update the existing patch adding it in the first place:
https://github.com/openwrt/openwrt/blob/main/target/linux/qualcommbe/patches-6.18/0342-net-qualcomm-Update-IPQ9574-PPE-driver.patch

target/linux/qualcommbe/patches-6.18/0363-net-ethernet-qualcomm-ppe-add-1000basex-clock.patch
```diff
@@ -0,0 +1,38 @@
...
+(ppe_port_mac_clock_rate_get_gmii), as it shares the same 1.25 Gbps baud rate
+requirements as SGMII. This restores 1G SFP hot-swapping capability on IPQ9574.
+
+Signed-off-by: Balsam CHIHI <balsam.chihi@moment.tech>
+---
+ drivers/net/ethernet/qualcomm/ppe/ppe_port.c | 1 +
+ 1 file changed, 1 insertion(+)
+
+diff --git a/drivers/net/ethernet/qualcomm/ppe/ppe_port.c b/drivers/net/ethernet/qualcomm/ppe/ppe_port.c
+index e37dd06ea..bb872df20 100644
```
openwrt-ai: This patch still carries raw `git format-patch` metadata that OpenWrt's quilt-managed patches don't have. Concretely, in `target/linux/qualcommbe/patches-6.18/`:

- **`index e37dd06ea..bb872df20 100644`** — zero of the 65 existing patches in that directory contain an `index <old>..<new>` line.
- **The `-- ` / `2.43.0` footer on lines 36-37** — zero of the 65 existing patches carry a git version footer either.
- **The hunk header on line 28** is untruncated (`@@ -775,6 +775,7 @@ static int ppe_port_mac_speed_clock_rate_set(struct ppe_port *ppe_port,`). Refreshed patches truncate the context function name at quilt's width — compare [`0357-...-complete-USXGMII-in-band-AN.patch:45`](``https://github.com/openwrt/openwrt/blob/7385558352213d8ddfbb088f7f87f38f085aada4/target/linux/qualcommbe/patches-6.18/0357-net-pcs-qcom-ipq9574-complete-USXGMII-in-band-AN.patch#L45),`` which reads `@@ -472,12 +475,16 @@ static int ipq_pcs_config_usxgmii(struct`.

Please regenerate this as a quilt patch with `make target/linux/refresh` (qualcommbe selected in `.config`) rather than dropping a `git format-patch` output in, so the metadata and hunk context match the rest of the directory. Note this also becomes moot if you follow the maintainer's request to fold the hunk into `0342-net-qualcomm-Update-IPQ9574-PPE-driver.patch` instead.

---

target/linux/qualcommbe/patches-6.18/0363-net-ethernet-qualcomm-ppe-add-1000basex-clock.patch
```diff
@@ -0,0 +1,38 @@
+From b19658f3ae93795683b0917ab7f2f128883ffcaa Mon Sep 17 00:00:00 2001
+From: Balsam CHIHI <balsam.chihi@moment.tech>
+Date: Fri, 21 Aug 2026 12:18:55 +0200
+Subject: [PATCH] net: ethernet: qualcomm: ppe: fix 1000BASE-X MAC clocking
```
openwrt-ai: nit: two filename issues, if this ends up staying a standalone patch.

1. The filename slug (`net-ethernet-qualcomm-ppe-add-1000basex-clock`) doesn't derive from this `Subject:` line (`net: ethernet: qualcomm: ppe: fix 1000BASE-X MAC clocking`) — "add ... clock" vs "fix ... MAC clocking". Every other patch in this directory uses the `format-patch`-derived slug of its own subject.
2. The number `0363` skips `0362`, which is free — the directory currently runs `0336`…`0361`, then jumps to the `0400`+ pwm block. A new patch here would normally be `0362-...`.

---

openwrt-ai (commented):
Reviewed 1 new commit. The functional change itself checks out: [`ppe_port_mac_speed_clock_rate_set()` in 0342](``https://github.com/openwrt/openwrt/blob/7385558352213d8ddfbb088f7f87f38f085aada4/target/linux/qualcommbe/patches-6.18/0342-net-qualcomm-Update-IPQ9574-PPE-driver.patch#L2431-L2447)`` really does omit `PHY_INTERFACE_MODE_1000BASEX` while [`ppe_port_mac_config()`](``https://github.com/openwrt/openwrt/blob/7385558352213d8ddfbb088f7f87f38f085aada4/target/linux/qualcommbe/patches-6.18/0342-net-qualcomm-Update-IPQ9574-PPE-driver.patch#L2318-L2327)`` already maps it to `PPE_MAC_TYPE_GMAC`, and `ppe_port_mac_clock_rate_get_gmii(SPEED_1000)` returns 125 MHz, which is the right rate for a 1.25 Gbaud 1000BASE-X link. `ppe_port.c` is not present in the v6.18.44 tree, so this is indeed OpenWrt-local code with no upstream counterpart to sync against.

Two inline comments, both about patch packaging rather than the fix.

---

BalsamCHIHI:
Hi @robimarko ,
patch "0342-net-qualcomm-Update-IPQ9574-PPE-driver.patch" updated.

openwrt-ai (commented):

Both earlier comments are addressed: the standalone `0363-*.patch` is gone and the hunk is now folded into `0342-net-qualcomm-Update-IPQ9574-PPE-driver.patch` as @robimarko asked, so the `git format-patch` metadata (`index ..`, `-- `/`2.43.0` footer, untruncated hunk context) is no longer present.

Verification done on this revision:

- Reconstructed the pre- and post-image of `drivers/net/ethernet/qualcomm/ppe/ppe_port.c` from the patch at base (`8df271d`) and at head. The old side is byte-identical and the new side differs by exactly one line — `case PHY_INTERFACE_MODE_1000BASEX:` in [`ppe_port_mac_speed_clock_rate_set()`](``https://github.com/openwrt/openwrt/blob/cd2c1da5277bd9112ac29678362810d844b88925/target/linux/qualcommbe/patches-6.18/0342-net-qualcomm-Update-IPQ9574-PPE-driver.patch#L2434-L2439).`` The reshuffling of `+	int ret, i;` relative to the surrounding `-` lines around [line 2265](``https://github.com/openwrt/openwrt/blob/cd2c1da5277bd9112ac29678362810d844b88925/target/linux/qualcommbe/patches-6.18/0342-net-qualcomm-Update-IPQ9574-PPE-driver.patch#L2265-L2290)`` is a refresh artifact with no effect on either image.
- Hunk headers are consistent: `@@ -676,8 +702,111 @@` → `112`, and every following new-side offset in `ppe_port.c` is bumped by one, so the patch looks properly regenerated rather than hand-edited. `Check Kernel patches` is green on this head.
- The routing is consistent with the rest of the driver: [`ppe_port_mac_config()`](``https://github.com/openwrt/openwrt/blob/cd2c1da5277bd9112ac29678362810d844b88925/target/linux/qualcommbe/patches-6.18/0342-net-qualcomm-Update-IPQ9574-PPE-driver.patch#L2326-L2327)`` already maps `PHY_INTERFACE_MODE_1000BASEX` to `PPE_MAC_TYPE_GMAC`, the mode is already in the `mac_interfaces[]` table fed into `supported_interfaces`, and `ppe_port_mac_clock_rate_get_gmii(SPEED_1000)` returns 125 MHz — the right rate for a 1.25 Gbaud 1000BASE-X link.
- `ppe_port.c` still does not exist in `gregkh/linux` v6.18.44 (`drivers/net/ethernet/qualcomm/ppe/` there has only `ppe.c`, `ppe_config.c`, `ppe_debugfs.c` and headers), so there is no upstream counterpart this could diverge from.
- The `Balsam CHIHI: ...` + `Signed-off-by:` attribution added to the patch header matches the existing `Alex G: ...` precedent in the same file.

---

robimarko:
Manually merged

PR #24829, opened 2026-08-21
mediatek: filogic: fix Teltonika RUTC50 pcie reset GPIO export

simsasss:
DTS GPIOs export properties handling changed in commit 0ca78951739a1e6e2f569438764381bd5fdb5190. This prevents USB hub startup by holding pcie reset. Update the device-tree accordingly. This commit has been tested on the device.

### DIFF
--- a/target/linux/mediatek/dts/mt7981a-teltonika-rutc50.dts
+++ b/target/linux/mediatek/dts/mt7981a-teltonika-rutc50.dts
@@ -45,7 +45,7 @@
 
 		gpio_pcie_reset {
 			gpio-export,name = "pcie_reset";
-			gpio-export,output = <1>;
+			gpio-export,output = <0>;
 			gpios = <&pio 3 GPIO_ACTIVE_LOW>;
 		};
 


target/linux/mediatek/dts/mt7981a-teltonika-rutc50.dts
```diff
@@ -45,7 +45,7 @@
 
 		gpio_pcie_reset {
 			gpio-export,name = "pcie_reset";
-			gpio-export,output = <1>;
+			gpio-export,output = <0>;
```
openwrt-ai, line 48: Flipping the DTS value papers over a semantic change that affects every active-low `gpio-export` node in the tree, so the fix may belong in the hack patch rather than here.

Before 0ca78951739a1e6e2f569438764381bd5fdb5190 the exporter went through `devm_gpio_request_one()`, which ends in [`gpiod_direction_output_raw()` at gpiolib-legacy.c:52](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/gpio/gpiolib-legacy.c#L52) — `gpio-export,output` was a **raw** pin level, unaffected by `GPIO_ACTIVE_LOW`. After that commit, [`dflags = val ? GPIOD_OUT_HIGH : GPIOD_OUT_LOW`](``https://github.com/openwrt/openwrt/blob/e3e423a6a63279bf511129f867c2386a1b1ca000/target/linux/generic/hack-6.18/800-GPIO-add-named-gpio-exports.patch#L43-L44)`` is handed to `devm_fwnode_gpiod_get_index()`, which lands in [`gpiod_direction_output_nonotify()` at gpiolib.c:4866-4867](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/gpio/gpiolib.c#L4866-L4867) after `GPIOD_FLAG_ACTIVE_LOW` has already been set — i.e. a **logical** level that gets inverted for active-low pins. That is exactly the PCIe reset assertion you observed.

The same inversion applies to ~70 other `gpio-export` children across ath79, ipq40xx, lantiq, mediatek, qualcommax and ramips that combine `gpio-export,output` with `GPIO_ACTIVE_LOW`, e.g. [`modem_power` in mt7981b-openfi-6c.dts:78-80](https://github.com/openwrt/openwrt/blob/e3e423a6a63279bf511129f867c2386a1b1ca000/target/linux/mediatek/dts/mt7981b-openfi-6c.dts#L78-L80) and [`sim` in mt7981b-zbtlink-zbt-z8106ax.dtsi:116-118](``https://github.com/openwrt/openwrt/blob/e3e423a6a63279bf511129f867c2386a1b1ca000/target/linux/mediatek/dts/mt7981b-zbtlink-zbt-z8106ax.dtsi#L116-L118).`` No DTS in main has been flipped for this so far, so after this patch RUTC50's `gpio-export,output` would mean the opposite of every other board's, and it would regress again if the hack patch is ever restored to raw semantics.

Would it be preferable to restore the previous behaviour inside `target/linux/generic/hack-6.18/800-GPIO-add-named-gpio-exports.patch` (and its `hack-6.12` copy) — e.g. acquire the descriptor with `GPIOD_ASIS` and then apply the requested level with `gpiod_direction_output_raw()` — so all affected boards keep working with their existing device trees? If the project has instead decided the new logical semantics are the correct interpretation, then this DTS change is right, but it would be worth saying so in the commit message and converting the other boards too.

---

openwrt-ai, line 48: nit: the two facts that make this one-line change reviewable live only in the PR description and are lost on merge — the commit message says just "DTS GPIOs export properties handling changed" with no SHA, and does not mention that the USB hub failed to start or that this was tested on hardware. Please move them into the commit message and add the standard trailer, which is used elsewhere in the tree (e.g. `f65530c1ca3`, `7b7cf1454ab`):

```
Fixes: 0ca78951739a ("kernel: use fwnode for GPIO export driver")
```

---

openwrt-ai (commented):
Reviewed 1 new commit. The diagnosis is correct — 0ca78951739a switched `gpio-export,output` from a raw pin level to a logical one — but the same inversion hits ~70 other active-low `gpio-export` nodes tree-wide, so I left an inline question about whether the fix belongs in the hack patch instead. Plus one commit-message nit.

The failing `FormalityCheck / Git & Commits` run (missing `Signed-off-by` matching the author) is already covered by the bot comment, so not repeated here.

---

openwrt-ai (commented):

The tree diff is byte-identical to what I reviewed at `e3e423a` — the only change is in the commit message, which now names `0ca78951739a1e6e2f569438764381bd5fdb5190` (verified: `kernel: use fwnode for GPIO export driver`). The message still matches what the diff does.

Retracting my earlier inline question about fixing this in `800-GPIO-add-named-gpio-exports.patch` instead: since that review, [`f1bf2d18392e` ("mediatek: filogic: fix GatoNetworks GDSP GPIOs exports")](https://github.com/openwrt/openwrt/commit/f1bf2d18392ef7b9f5f7bc6773055cfbddde371b) landed on `main`, flipping `gpio-export,output` on two `GPIO_ACTIVE_LOW` nodes for exactly the same reason. So the project has settled on adjusting the DTS, and this PR follows that precedent. `pcie_reset` is also the only `gpio-export,output` node in this file with an active-low flag — `modem_power`, `modem_reset` and `digital_output` are `GPIO_ACTIVE_HIGH` and correctly left alone — so there is no parity gap within the board.

CI on `bd65a2f` is fully green, including `FormalityCheck / Git & Commits`, which failed at the previous head.

My earlier `nit:` about folding the PR-description details (USB hub not starting, tested on hardware) and a `Fixes:` trailer into the commit message still stands on its existing thread; not repeating it here.

---

jonasjelonek:
I think this isn't needed anymore with #24832. Can you confirm?

simsasss:
> I think this isn't needed anymore with #24832. Can you confirm?

Yes, its ok with OpenWrt SNAPSHOT r36072-f1a3f07139.

PR #24830, opened 2026-08-21
mt76: install default MT7915 EEPROM bins

KNIGAZV:
The MT7915 default EEPROM bins are part of the mt76 sources and are installed for the MT7996 family, but not for MT7915.

Devices that carry no calibration data in flash, and whose efuse does not hold a complete EEPROM either, depend on them: mt7915_eeprom_load() fails with -EINVAL and the driver falls back to request_firmware() for mediatek/mt7915_eeprom_dbdc.bin. Without the file the probe never finishes and no PHY is registered at all - the radio is simply missing.

Installing the two bins adds 7 KiB to kmod-mt7915-firmware.

Tested on a TP-Link EX220 v1 (NAND), whose support I am submitting separately: with this change both bands come up and a client associates at HE rates over 80 MHz, a local transfer peaking at 541 Mbit/s.

### DIFF
--- a/package/kernel/mt76/Makefile
+++ b/package/kernel/mt76/Makefile
@@ -1,7 +1,7 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=mt76
-PKG_RELEASE=1
+PKG_RELEASE:=2
 
 PKG_LICENSE:=BSD-3-Clause-Clear
 PKG_LICENSE_FILES:=
@@ -612,6 +612,8 @@ define KernelPackage/mt7915-firmware/install
 		$(PKG_BUILD_DIR)/firmware/mt7915_wa.bin \
 		$(PKG_BUILD_DIR)/firmware/mt7915_wm.bin \
 		$(PKG_BUILD_DIR)/firmware/mt7915_rom_patch.bin \
+		$(PKG_BUILD_DIR)/firmware/mt7915_eeprom.bin \
+		$(PKG_BUILD_DIR)/firmware/mt7915_eeprom_dbdc.bin \
 		$(1)/lib/firmware/mediatek
 endef
 


EPinci:
Why you filed this PR in here instead of https://github.com/openwrt/mt76?

untilscour:
> Why you filed this PR in here instead of https://github.com/openwrt/mt76?

Look at the diff - it is obviously a packaging change for the `mt7915-firmware` kmod, not code changes of the driver

package/kernel/mt76/Makefile
```diff
@@ -612,6 +612,8 @@ define KernelPackage/mt7915-firmware/install
 		$(PKG_BUILD_DIR)/firmware/mt7915_wa.bin \
 		$(PKG_BUILD_DIR)/firmware/mt7915_wm.bin \
 		$(PKG_BUILD_DIR)/firmware/mt7915_rom_patch.bin \
+		$(PKG_BUILD_DIR)/firmware/mt7915_eeprom.bin \
+		$(PKG_BUILD_DIR)/firmware/mt7915_eeprom_dbdc.bin \
```
openwrt-ai, line 616: The default-EEPROM fallback this fixes is not specific to chip 0x7915 — the same `mt7915_eeprom_name()` serves the whole family and returns `mediatek/mt7916_eeprom.bin` for MT7916, `mediatek/mt7981_eeprom_mt7976_dbdc.bin` for MT7981, and one of five `mediatek/mt7986_eeprom_mt79{75,76}*.bin` variants for MT7986 ([`mt7915_eeprom_name()`, mt7915/eeprom.c:67-95](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/net/wireless/mediatek/mt76/mt7915/eeprom.c#L67-L95), v6.18.44 — matches the mt76 package sources).

The sibling install blocks just below still ship only `wa`/`wm`/`rom_patch` ([Makefile:620-646](https://github.com/openwrt/openwrt/blob/58593b5024af9e8a5f0585154d3aeccaeb089e27/package/kernel/mt76/Makefile#L620-L646)), so a board on those SoCs with no flash calibration and an incomplete efuse hits exactly the probe failure described in the commit message. Is leaving MT7916/MT7981/MT7986 out intentional (e.g. no such boards in tree today), or worth covering in the same commit?

Worth noting for MT7981 specifically: `mt7981_eeprom_mt7976_dbdc.bin` is already extracted by [`Build/InstallDev`, Makefile:742-745](https://github.com/openwrt/openwrt/blob/58593b5024af9e8a5f0585154d3aeccaeb089e27/package/kernel/mt76/Makefile#L742-L745), but only into `STAGING_DIR_IMAGE` for image generation — it never lands in `/lib/firmware/mediatek` on the device, so the runtime fallback is still unavailable there.

---

openwrt-ai (commented):
Reviewed 1 new commit; the commit message matches the diff (Signed-off-by present and matching the author). One inline question about coverage of the sibling MT7916/MT7981/MT7986 firmware packages, which share the same default-EEPROM fallback path.

---

castiel652:
Why? Does your board not have it on the NAND?

package/kernel/mt76/Makefile
KNIGAZV: Leaving them out was not deliberate - I limited the change to the family I can actually test. The same fallback path exists for MT7916, MT7981 and MT7986, and the MT7981 case you describe (extracted only into STAGING_DIR_IMAGE, never into /lib/firmware/mediatek) looks like the same gap.
I have no hardware for any of those, so I did not want to ship files I cannot verify boot. Happy to extend the commit to all four families if maintainers prefer coverage over tested-only changes - say the word and I will push it.

KNIGAZV:
> Why? Does your board not have it on the NAND?

No, it does not. I dumped every service partition of this board - boot, boot_env, misc_ro, misc_rw, misc_isp, bflag and the second firmware slot
and there is no calibration blob in any of them: no 0x7915 magic followed by a plausible MAC anywhere, and the two partitions that would be the obvious place (misc_isp, bflag) are erased. The vendor firmware does not carry one either.
The efuse does not hold a complete EEPROM, so mt7915_eeprom_load() ends in the -EINVAL branch and the driver goes to request_firmware() for mediatek/mt7915_eeprom_dbdc.bin - which nothing installs today. Without this change the probe stops there and no PHY is registered at all; with it both bands come up and a client pulls 541 Mbit/s over 80 MHz.

castiel652:
I remember TP-Link devices usually store it as file.

Can you try mounting `misc_ro` and see if there's anything on there?

KNIGAZV:
> I remember TP-Link devices usually store it as file.
> 
> Can you try mounting `misc_ro` and see if there's anything on there?

You were right, thank you - that hint is what found it.

misc_ro turned out to hold a JFFS2 filesystem with five files named after their original flash offsets: 0x0014F100 (6 bytes, the label MAC), 0x0014F200, 0x0014F300, 0x0014F500 and 0x00180000 (608 KB). The MT7915 EEPROM sits inside the last one at offset 0x90000. My raw scan for the 0x7915 magic across the partition found nothing because the vendor compressed those JFFS2 nodes with zlib.

Extracting it is measurably better than the default EEPROM I was proposing here: -44 dBm instead of -82 at the same half metre, HE-MCS 11 instead of HE-MCS 3, 692 instead of 541 Mbit/s on a local transfer. So this pull request is going away and #24831 now does the extraction.

KNIGAZV:
Closing this: my board does not need it after all.

castiel652 was right that TP-Link keeps the calibration as a file. The misc_ro partition holds a JFFS2 filesystem with five files named after their original flash offsets, and the largest one carries the MT7915 EEPROM at offset 0x90000. Extracting it from there is the correct fix, and it is strictly better than the generic default: with the factory data the client reports -44 dBm at half a metre and negotiates HE-MCS 11, while the same link on the default EEPROM reports -82 dBm and settles at HE-MCS 3 - a local transfer peaks at 692 instead of 541 Mbit/s.

Worse, the two changes are mutually exclusive: 11-mt76-caldata starts with "[ -e /lib/firmware/$FIRMWARE ] && exit 0", so shipping the default bins would stop the extraction from ever running - on this board and on any other that grows a caldata entry later.

The device support pull request (#24831) now carries the extraction plus CONFIG_JFFS2_ZLIB, which the vendor's compression makes necessary.

The gap the bot pointed out for MT7916/MT7981/MT7986 is real and still unaddressed, but it should be argued by someone who owns one of those boards and can show the failure, not by me on inference.

Djfe:
if anyone ever stumbles across this:
a fallback does not make sense since it includes the files for many devices that don't need it. if your board really does not come with an eeprom file you can supply the eeprom-data via its dts instead
see https://github.com/openwrt/openwrt/blob/f0204d78e1b42d9e7fdf57a59b0aaa09aea0abcc/target/linux/mediatek/dts/mt7986a-bananapi-bpi-r3-mini.dts#L589

PR #24831, opened 2026-08-21
ramips: add support for TP-Link EX220 v1 (NAND)

KNIGAZV:
Adds support for the parallel-NAND revision of the TP-Link EX220,
distributed by the Russian ISP MTS and labelled "EX220(RU)MTS) Ver:1.30".
It shares the model name with the already supported tplink,ex220-v1, but
the board is different: 128 MB of parallel NAND behind NMBM instead of
16 MB of SPI-NOR, and a different flash layout. The bootloader identifies
it - U-Boot prints "NAND:  128 MiB" followed by "NMBM has been
successfully attached".

Tested on the device, on current main (6.18):

- installation exactly as written in the commit message: serial console,
  initramfs over YMODEM, then sysupgrade -n
- sysupgrade afterwards, both keeping and wiping the configuration
- failsafe mode
- all five ethernet ports verified with a cable, both WiFi bands
  (HE20 on 2.4 GHz, HE80 on 5 GHz), all seven LEDs, both buttons
- MAC addresses: label MAC on the ethernet side, label+2 and label+3 on
  the two radios

Changes since the first revision, all from review:

- the flash is parallel NAND, not SPI-NAND - corrected here, in the
  device tree and in the commit message
- added the &state_default pinmux block, which also silences the
  "there is not valid maps for state default" line printed on every boot
- carried over mediatek,disable-radar-background from both siblings
- dropped the duplicate nvmem cell: a single mac-base cell referenced
  with the argument works, the CPU port gets the label MAC and
  rx_packets increments normally
- the radio calibration is now read from the vendor rom partition
  instead of relying on a default EEPROM, which is what #24830 was for;
  that pull request is closed

On the calibration, since it drove most of the review: castiel652 was
right that TP-Link stores it as a file. misc_ro holds a JFFS2 filesystem
with five files named after their original flash offsets, and the
largest one carries the MT7915 EEPROM at 0x90000. A raw scan for the
0x7915 magic finds nothing because those nodes are zlib-compressed,
which is why CONFIG_JFFS2_ZLIB comes with this patch.

The difference is not cosmetic. Same client, same half metre, default
EEPROM first and factory calibration second:

- signal: -82 dBm at SNR 3, versus -44 dBm at SNR 42
- negotiated rate: HE-MCS 3, versus HE-MCS 11
- local transfer: 541 Mbit/s, versus 692 Mbit/s


target/linux/ramips/dts/mt7621_tplink_ex220-v1-nand.dts
```diff
@@ -0,0 +1,268 @@
...
+			status = "okay";
+			label = "lan4";
+		};
+
+		port@4 {
+			status = "okay";
+			label = "wan";
+		};
+	};
+};
```
openwrt-ai, line 257: The `&state_default` pinmux override is missing, so every GPIO this board uses for LEDs and buttons is left on its non-GPIO function.

The MT7621 pin groups ([`pinctrl-mt7621.c:37-51`](https://github.com/gregkh/linux/blob/1eb4dd6f04cec9eeee9a2fb52b0e30f0ffdb5aec/drivers/pinctrl/mediatek/pinctrl-mt7621.c#L37-L51)) map to the pins used here as: `i2c` = 3-4, `uart3` = 5-8, `jtag` = 13-17, `wdt` = 18. This DTS drives GPIO 3, 4 (LED WPS/LAN), 7, 8 (both buttons), 13, 14, 15, 16 (WAN amber/green, WLAN 5G/2G) and 18 (power LED) — i.e. exactly those four groups, and none of them are muxed to `gpio`.

Both siblings do it: [`mt7621_tplink_ex220-v1.dts:265-270`](https://github.com/openwrt/openwrt/blob/0acc59c25feb23a2ed45fcdb99bcabaaf061e704/target/linux/ramips/dts/mt7621_tplink_ex220-v1.dts#L265-L270) and [`mt7621_tplink_ex220-v2.dts:225-230`](https://github.com/openwrt/openwrt/blob/0acc59c25feb23a2ed45fcdb99bcabaaf061e704/target/linux/ramips/dts/mt7621_tplink_ex220-v2.dts#L225-L230) carry the identical block. It presumably appears to work because the vendor bootloader already leaves the mux in a usable state, but that is not something to rely on.

```suggestion
};

&state_default {
	gpio {
		groups = "i2c", "uart3", "jtag", "wdt";
		function = "gpio";
	};
};
```

---

target/linux/ramips/dts/mt7621_tplink_ex220-v1-nand.dts
```diff
@@ -0,0 +1,268 @@
...
+			nvmem-cells = <&macaddr_base_5c50bc 3>;
+			nvmem-cell-names = "mac-address";
+		};
+	};
+};
+
+&gmac0 {
+	nvmem-cells = <&macaddr_misc_ro_5c50bc>;
+	nvmem-cell-names = "mac-address";
+};
```
openwrt-ai, line 227: When you tried the `mac-base` cell here, did you pass the cell argument — i.e. `nvmem-cells = <&macaddr_base_5c50bc 0>;`, not `<&macaddr_base_5c50bc>`?

`macaddr_base_5c50bc` declares `#nvmem-cell-cells = <1>` (line 153), so a consumer must supply the offset. Referencing it without the argument makes the phandle-with-args parse fail, the interface never gets its MAC from nvmem, and you end up with a random address on the CPU port — which would fit the "link up, TX works, `rx_packets` stays 0" symptom you describe rather better than a kernel regression does.

The in-tree sibling does exactly that and works: [`mt7621_tplink_ex220-v1.dts:223-226`](https://github.com/openwrt/openwrt/blob/0acc59c25feb23a2ed45fcdb99bcabaaf061e704/target/linux/ramips/dts/mt7621_tplink_ex220-v1.dts#L223-L226) has `nvmem-cells = <&macaddr_rom_file_f100 0>;` on `&gmac0` against a `mac-base` cell.

If that turns out to be it, `macaddr_misc_ro_5c50bc` (lines 146-148) and the explanatory comment above it can both go, leaving one cell for all three consumers.

---

target/linux/ramips/dts/mt7621_tplink_ex220-v1-nand.dts
```diff
@@ -0,0 +1,268 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
+/*
+ * TP-Link EX220 v1, labelled "EX220(RU)MTS) Ver:1.30".
+ *
+ * Not to be confused with tplink,ex220-v1: that one is built around
+ * 16 MB of SPI-NOR and uses a completely different flash layout. This
+ * revision carries 128 MB of SPI-NAND behind NMBM and is recognisable
```
openwrt-ai: nit: this is parallel NAND, not SPI-NAND. The DTS enables `&nand`, which is the MT7621 NAND flash interface ([`mt7621.dtsi:484-493`](https://github.com/openwrt/openwrt/blob/0acc59c25feb23a2ed45fcdb99bcabaaf061e704/target/linux/ramips/dts/mt7621.dtsi#L484-L493), `compatible = "mediatek,mt7621-nfc"`), driven by [`mt7621_nand.c`](https://github.com/openwrt/openwrt/blob/0acc59c25feb23a2ed45fcdb99bcabaaf061e704/target/linux/ramips/files/drivers/mtd/nand/raw/mt7621_nand.c) under `drivers/mtd/nand/raw/`. Real SPI-NAND boards in this target hang the chip off `&spi0` with `compatible = "spi-nand"` instead — e.g. [`mt7621_plasmacloud_pax1800-lite.dts:145`](https://github.com/openwrt/openwrt/blob/0acc59c25feb23a2ed45fcdb99bcabaaf061e704/target/linux/ramips/dts/mt7621_plasmacloud_pax1800-lite.dts#L145). (The `spi-nand` node name in the dtsi's `nand_pins` refers to routing the parallel NAND signals onto the SPI pin group via `function = "nand1"`, not to the SPI-NAND protocol.)

Same wording appears in the commit message ("128 MB of SPI-NAND managed by NMBM" and the `Flash:` line) and in the PR body — worth fixing in all three so the next person comparing the two EX220 revisions isn't misled.

```suggestion
 * revision carries 128 MB of parallel NAND behind NMBM and is recognisable
```

---

target/linux/ramips/dts/mt7621_tplink_ex220-v1-nand.dts
```diff
@@ -0,0 +1,268 @@
...
+};
+
+&pcie {
+	status = "okay";
+};
+
+&pcie1 {
+	wifi@0,0 {
+		compatible = "mediatek,mt76";
+		reg = <0x0000 0 0 0 0>;
```
openwrt-ai, line 195: nit: both sibling EX220 revisions set `mediatek,disable-radar-background;` on their MT7915 node — [`mt7621_tplink_ex220-v1.dts:204`](https://github.com/openwrt/openwrt/blob/0acc59c25feb23a2ed45fcdb99bcabaaf061e704/target/linux/ramips/dts/mt7621_tplink_ex220-v1.dts#L204) and the same in `mt7621_tplink_ex220-v2.dts`. This board is the same SoC and the same DBDC MT7915E, so the omission looks like a parity gap rather than a deliberate difference. Is background radar detection actually usable here, or should the property be carried over?

---

openwrt-ai (commented):
Reviewed 1 new commit (0acc59c). Comments inline.

The commit message carries the hardware specification, flash instructions and MAC address layout, and matches what the diff does. CI is green on the head commit.

---

target/linux/ramips/dts/mt7621_tplink_ex220-v1-nand.dts
KNIGAZV: Added, thanks - this was a real gap.
It also silences the "mt7621-pinctrl pinctrl: there is not valid maps for state default" line the board printed on every boot, which I had written off as harmless. LEDs and buttons keep working with the groups muxed to gpio.

target/linux/ramips/dts/mt7621_tplink_ex220-v1-nand.dts
KNIGAZV: You are right, and my note was wrong. I retested with a single mac-base cell and the argument on &gmac0, exactly as the sibling does: the CPU port gets the label MAC and eth0 rx_packets increments normally (301 -> 333 over a few pings). The duplicate plain cell and the comment above it are gone.
For the record on what I originally saw: the interface did get the correct label MAC, not a random one, while rx_packets stayed at 0 - so a failed phandle-with-args parse does not explain it. I could not reproduce it after adding the &state_default block you point out below, on current main. I am not claiming causation without a bisect; the simpler device tree works, so it stands.

target/linux/ramips/dts/mt7621_tplink_ex220-v1-nand.dts
KNIGAZV: Correct - it hangs off the MT7621 NAND controller (mediatek,mt7621-nfc), not the SPI bus. Fixed in the device tree header, the commit message and the PR body.

target/linux/ramips/dts/mt7621_tplink_ex220-v1-nand.dts
KNIGAZV: Carried over - same SoC and the same DBDC MT7915E as both siblings, so the omission was an oversight rather than a deliberate difference.

target/linux/ramips/dts/mt7621_tplink_ex220-v1-nand.dts
```diff
@@ -0,0 +1,263 @@
...
+		#size-cells = <0>;
+		mediatek,disable-radar-background;
+
+		/*
+		 * The board holds no calibration data, neither in flash nor
+		 * in the efuse, so mt76 falls back to the default EEPROM and
+		 * takes the addresses from here. Both bands need their own
+		 * address, otherwise the two PHYs end up bridged together
+		 * with an identical MAC.
+		 */
```
openwrt-ai, line 207: This comment is now false. It says the board holds no calibration data "neither in flash nor in the efuse, so mt76 falls back to the default EEPROM" — but this revision adds an [`11-mt76-caldata`](``https://github.com/openwrt/openwrt/blob/4d8f4abd381bb9af39b085ec5c59da99dca3784a/target/linux/ramips/mt7621/base-files/etc/hotplug.d/firmware/11-mt76-caldata#L12-L18)`` entry that mounts `misc_ro` and copies the vendor EEPROM out of it, and the commit message spends two paragraphs on exactly that. Leaving the old wording in the DTS points the next reader at the default EEPROM and hides the `misc_ro` dependency, which is the one thing that must not be broken when the flash layout is touched.

```suggestion
		/*
		 * The calibration data is not reachable over nvmem: it sits
		 * inside a JFFS2 file in the misc_ro partition, so it is
		 * copied over the default EEPROM at boot by the target's
		 * 11-mt76-caldata hotplug script. Both bands still need
		 * their own address from here, otherwise the two PHYs end
		 * up bridged together with an identical MAC.
		 */
```

---

target/linux/ramips/mt7621/base-files/etc/hotplug.d/firmware/11-mt76-caldata
```diff
@@ -9,6 +9,13 @@ board=$(board_name)
 case "$FIRMWARE" in
 "mediatek/mt7915_eeprom_dbdc.bin")
 	case "$board" in
+	tplink,ex220-v1-nand)
+		EEPROM=/lib/firmware/$FIRMWARE
+		mkdir -p /tmp/rom_file
+		mount -t jffs2 -o ro "/dev/mtdblock$(find_mtd_index misc_ro)" /tmp/rom_file
+		dd if=/tmp/rom_file/0x00180000 of=$EEPROM bs=1 skip=$((0x90000)) count=$((0xe00)) 2>/dev/null
+		umount /tmp/rom_file
+		;;
```
openwrt-ai, line 22: Nothing here is checked, so every failure mode is silent and lands the board back on the generic default EEPROM — the -82 dBm / HE-MCS 3 case the commit message argues against — with no trace in the log. If the JFFS2 mount fails (bad block in `misc_ro`, `CONFIG_JFFS2_ZLIB` dropped again, `find_mtd_index` returning nothing so the device string becomes `/dev/mtdblock`), `dd` fails into `2>/dev/null` and the only visible effect is a bare `umount: can't unmount /tmp/rom_file` on the console.

The tree has helpers for this in [`caldata.sh`](https://github.com/openwrt/openwrt/blob/4d8f4abd381bb9af39b085ec5c59da99dca3784a/package/base-files/files/lib/functions/caldata.sh#L6-L19), already sourced at the top of this script: `caldata_dd <src> <dst> <count> <offset>` (one `read()` instead of 3584 `bs=1` iterations, and it returns a status) plus `caldata_die` for the message. The `mount ... || { ... }` shape has a precedent at [`bcm4908 platform.sh:315`](https://github.com/openwrt/openwrt/blob/4d8f4abd381bb9af39b085ec5c59da99dca3784a/target/linux/bcm4908/base-files/lib/upgrade/platform.sh#L315).

```suggestion
	tplink,ex220-v1-nand)
		EEPROM=/lib/firmware/$FIRMWARE
		mkdir -p /tmp/rom_file
		mount -t jffs2 -o ro "/dev/mtdblock$(find_mtd_index misc_ro)" /tmp/rom_file || \
			caldata_die "failed to mount misc_ro"
		caldata_dd /tmp/rom_file/0x00180000 $EEPROM 0xe00 0x90000
		ret=$?
		umount /tmp/rom_file
		[ "$ret" = 0 ] || caldata_die "failed to extract calibration data from misc_ro"
		;;
```

Worth noting this still does not catch a short read: if `/tmp/rom_file/0x00180000` is smaller than `0x90000 + 0xe00` on some unit, `dd` writes a truncated EEPROM, returns 0, and mt76 consumes it. A `[ "$(wc -c < $EEPROM)" = 3584 ]` guard after the copy would close that too, if you think it is worth the line.

---

target/linux/ramips/dts/mt7621_tplink_ex220-v1-nand.dts
```diff
@@ -0,0 +1,263 @@
+// SPDX-License-Identifier: GPL-2.0-or-later OR MIT
```
openwrt-ai, line 1: nit: this file is committed with mode `100755`. It is the only executable one of the 822 files in `target/linux/ramips/dts/` — both siblings, `mt7621_tplink_ex220-v1.dts` and `mt7621_tplink_ex220-v2.dts`, are `100644`. Probably picked up from a Windows/exFAT working copy. `chmod 644` on the file and amend.

---

openwrt-ai (commented):
Reviewed the changes since `0acc59c` (force-push to `4d8f4ab`): the `&state_default` block, the single `mac-base` cell with the cell argument, `mediatek,disable-radar-background`, the `misc_ro` calibration extraction and `CONFIG_JFFS2_ZLIB`. The four earlier points are all addressed correctly; the mac-base node name `macaddr@5c50bc` matches the dominant convention in this target (130 of 132 mac-base cells in `target/linux/ramips/dts/`), and `CONFIG_ZLIB_DEFLATE`/`CONFIG_ZLIB_INFLATE` are already `=y` in `mt7621/config-6.18`, so the new symbol needs no companion lines. Three new comments inline, on the calibration path the revision introduces.

## Commit checks

- `4d8f4ab` "ramips: add support for TP-Link EX220 v1 (NAND)" — the flash type fix did not land everywhere in the message. The second paragraph still reads "128 MB of **SPI-NAND** managed by NMBM instead of 16 MB of SPI-NOR", while the `Flash:` line below it says "128 MB parallel NAND, ESMT F59L1G81MB", the DTS header now says "128 MB of parallel NAND behind NMBM", and the PR body says the same. The chip hangs off `&nand` (`mediatek,mt7621-nfc`), so the `Flash:` line is the correct one and the paragraph should follow it.

CI on the head commit has no failures — the FormalityCheck runs passed and the ramips kernel builds are queued or still running.

---

target/linux/ramips/dts/mt7621_tplink_ex220-v1-nand.dts
KNIGAZV: You are right, the comment described the previous revision. Replaced with
your wording - it names the misc_ro dependency explicitly, which is the
part that must not be broken if the flash layout is ever touched.

target/linux/ramips/mt7621/base-files/etc/hotplug.d/firmware/11-mt76-caldata
KNIGAZV: Fixed, thanks - that was the worst of the four, because the failure mode
it describes is silent and lands the board back on the -82 dBm / HE-MCS 3
case with nothing in the log to explain it.

Now uses caldata_dd and caldata_die, with the mount guarded and the
return value checked after the umount. I also kept the short-read guard
you called optional: a truncated EEPROM is exactly the kind of thing
that would be blamed on the radio for weeks.

Verified on hardware after the change: the extracted file still matches
the vendor blob (md5 9a19c873...), and iwinfo reports link quality 70/70
where the default EEPROM gave 25/70.

target/linux/ramips/dts/mt7621_tplink_ex220-v1-nand.dts
KNIGAZV: Fixed - chmod 644 and amended. It came from a working copy on a Windows
filesystem, which has no permission bits at all.

KNIGAZV:
> the flash type fix did not land everywhere in the message

Fixed - the second paragraph now says parallel NAND as well. The chip
hangs off &nand (mediatek,mt7621-nfc), so the Flash: line was the correct
one and the paragraph had been left behind.

target/linux/ramips/mt7621/base-files/etc/hotplug.d/firmware/11-mt76-caldata
```diff
@@ -9,6 +9,17 @@ board=$(board_name)
...
+	tplink,ex220-v1-nand)
+		EEPROM=/lib/firmware/$FIRMWARE
+		mkdir -p /tmp/rom_file
+		mount -t jffs2 -o ro "/dev/mtdblock$(find_mtd_index misc_ro)" /tmp/rom_file || \
+			caldata_die "failed to mount misc_ro"
+		caldata_dd /tmp/rom_file/0x00180000 $EEPROM 0xe00 0x90000
+		ret=$?
+		umount /tmp/rom_file
+		[ "$ret" = 0 ] || caldata_die "failed to extract calibration data from misc_ro"
+		[ "$(wc -c < $EEPROM)" = 3584 ] || caldata_die "short calibration data in misc_ro"
```
openwrt-ai: Both guards fire exactly once and then hand the bad file to mt76 anyway. `caldata_dd` has already written `$EEPROM` under `/lib/firmware/`, which is on the writable overlay, and `caldata_die` only prints and [`exit 1`s](https://github.com/openwrt/openwrt/blob/73a6c642583eec8b6eb84de7e5fb16cb0195d611/package/base-files/files/lib/functions/caldata.sh#L16-L19) — it does not remove the partial file. On the next firmware request (next boot, or the second band's request in the same boot) [line 3](``https://github.com/openwrt/openwrt/blob/73a6c642583eec8b6eb84de7e5fb16cb0195d611/target/linux/ramips/mt7621/base-files/etc/hotplug.d/firmware/11-mt76-caldata#L3)`` sees the file exists, `exit 0`s before any check runs, and the truncated EEPROM is loaded silently — the exact outcome the new short-read guard was added to prevent, now made permanent instead of one-shot.

Removing the file on both failure paths is enough; `$EEPROM` is guaranteed not to have existed before this script ran, so `rm -f` cannot delete anything else.

```suggestion
		[ "$ret" = 0 ] || { rm -f "$EEPROM"; caldata_die "failed to extract calibration data from misc_ro"; }
		[ "$(wc -c < $EEPROM)" = "$((0xe00))" ] || { rm -f "$EEPROM"; caldata_die "short calibration data in misc_ro"; }
```

The suggestion also swaps the literal `3584` for `$((0xe00))` so the expected size and the `caldata_dd` count argument on line 17 stay the same token — the `netis,n6` branch right below uses `$((0xe00))` for the same constant.

---

openwrt-ai (commented):
Reviewed the changes since `4d8f4ab` (force-push to `73a6c64`): the DTS calibration comment, the hardened `misc_ro` extraction in `11-mt76-caldata`, the `100755` -> `100644` mode fix, and the "SPI-NAND" -> "parallel NAND" correction in the second paragraph of the commit message. All four earlier points are addressed; the commit message now agrees with the `Flash:` line, the DTS header and the PR body, and it still carries the hardware specification, flash instructions and MAC address layout matching what the diff does.

One new comment inline, on the failure handling the revision introduces — the guards themselves are the right ones, but they leave the bad artefact behind for the next boot to pick up.

For the record on things checked and found fine: `caldata_dd`'s `bs=0xe00 skip=0x90000 count=1` with `iflag=skip_bytes,fullblock` reads exactly the range the commit message describes; `CONFIG_JFFS2_FS=y` comes from `target/linux/generic/config-6.18`, so the new `CONFIG_JFFS2_ZLIB=y` in `mt7621/config-6.18` has its dependency met; and `/lib/firmware/mediatek/` exists on this device because `DEVICE_PACKAGES` pulls in `kmod-mt7915-firmware`.

CI on the head commit has no failures — the three FormalityCheck runs passed.

---

target/linux/ramips/mt7621/base-files/etc/hotplug.d/firmware/11-mt76-caldata
KNIGAZV: Right, and the failure this creates is worse than the one I was guarding
against: caldata_die leaves the partial file behind, so line 3 short-
circuits on every later request and the truncated EEPROM becomes
permanent rather than one-shot. Applied your suggestion, including the
$((0xe00)) token so the size and the caldata_dd count argument stay in
sync.

Checked the logic against a deliberately short source: the guard fires
and the file is gone afterwards, so the next firmware request retries
instead of consuming the leftover. Normal path re-verified on the device
after the change - the extracted blob still matches the vendor one.

KNIGAZV:
It's been quiet for a week, so a quick status: all three rounds of comments are addressed, every thread is resolved, CI is green on 7e5917c and the branch merges cleanly. Nothing is waiting on me.

If anyone is opening this fresh: it's a TP-Link EX220 v1 with 128 MB parallel NAND behind NMBM. The EX220 already in the tree is the SPI one, so this is a separate device rather than a variant patch. The one part worth a look is the vendor radio calibration - it sits as a file inside a JFFS2 filesystem in misc_ro, which is why the patch needs CONFIG_JFFS2_ZLIB=y and mounts the partition instead of reading a raw offset.

Verified on hardware: the extracted blob matches the vendor one (md5 9a19c873...) both in /lib/firmware and inside the driver, via the debugfs eeprom file of both PHYs. Same client, fallback EEPROM: −82 dBm, HE-MCS 3. Vendor blob: −44 dBm, HE-MCS 11. The device has been running as an AP since, without OOM or UBI errors.

I still have it on the bench, so anything that needs a measurement rather than an argument, ask.

openwrt-ai (commented):

---

KNIGAZV:
@namiltd, could you please take a look at this ramips/mt7621 device-support PR, or point me to the right reviewer?  
It has been idle for two weeks after all review comments were addressed. The branch is clean, CI is green, and the parallel-NAND board has been tested on hardware, including installation, sysupgrade, failsafe, Ethernet, both Wi-Fi bands, LEDs/buttons, MAC addresses, and factory radio calibration. The device is still available for any additional tests.

namiltd:
Ping @robimarko, @jonasjelonek 

However, it would be useful to have someone else run this software on their router as well.
Look for other users at https://forum.openwrt.org

PR #24832, opened 2026-08-21
kernel: fix gpio-export initial output for active-low GPIOs

neheb:
The gpio-export driver was converted from devm_gpio_request_one (raw GPIO API) to devm_fwnode_gpiod_get_index (gpiod API). The gpiod API sets FLAG_ACTIVE_LOW on the descriptor before calling gpiod_direction_output, which inverts the value. This means gpio-export,output = <0> on an active-low GPIO now results in physical HIGH instead of the intended physical LOW.

Fix all 27 gpio-export child nodes that combine GPIO_ACTIVE_LOW with gpio-export,output = <0> by flipping the output value to <1>.

Fixes: 0ca78951739a ("kernel: use fwnode for GPIO export driver")

ping @lorand-horvath @jonasjelonek

### DIFF
--- a/target/linux/generic/hack-6.12/800-GPIO-add-named-gpio-exports.patch
+++ b/target/linux/generic/hack-6.12/800-GPIO-add-named-gpio-exports.patch
@@ -15,7 +15,7 @@ Signed-off-by: John Crispin <blogic@openwrt.org>
  
  #include "gpiolib.h"
  #include "gpiolib-of.h"
-@@ -1205,3 +1207,61 @@ void of_gpiochip_remove(struct gpio_chip
+@@ -1205,3 +1207,65 @@ void of_gpiochip_remove(struct gpio_chip
  
  	of_node_put(np);
  }
@@ -37,22 +37,26 @@ Signed-off-by: John Crispin <blogic@openwrt.org>
 +		struct gpio_desc *desc;
 +		const char *name = NULL;
 +		bool dmc;
-+		enum gpiod_flags dflags;
++		int err;
 +		int i = 0;
 +
-+		if (!fwnode_property_read_u32(child, "gpio-export,output", &val))
-+			dflags = val ? GPIOD_OUT_HIGH : GPIOD_OUT_LOW;
-+		else
-+			dflags = GPIOD_IN;
-+
 +		fwnode_property_read_string(child, "gpio-export,name", &name);
++
 +		while (true) {
-+			desc = devm_fwnode_gpiod_get_index(dev, child, NULL, i, dflags,
++			desc = devm_fwnode_gpiod_get_index(dev, child, NULL, i,
++					GPIOD_ASIS,
 +					name ? name : fwnode_get_name(child));
 +			if (PTR_ERR(desc) == -ENOENT)
 +				break;
 +			if (IS_ERR(desc))
 +				return PTR_ERR(desc);
++
++			if (!fwnode_property_read_u32(child, "gpio-export,output", &val))
++				err = gpiod_direction_output_raw(desc, !!val);
++			else
++				err = gpiod_direction_input(desc);
++			if (err)
++				return err;
 +
 +			dmc = fwnode_property_present(child, "gpio-export,direction_may_change");
 +			gpio_export_with_name(desc, dmc, name);
--- a/target/linux/mediatek/dts/mt7981b-gatonetworks-gdsp.dts
+++ b/target/linux/mediatek/dts/mt7981b-gatonetworks-gdsp.dts
@@ -60,13 +60,13 @@
 
 		modem1 {
 			gpio-export,name = "modem1";
-			gpio-export,output = <1>;
+			gpio-export,output = <0>;
 			gpios = <&pio 2 GPIO_ACTIVE_LOW>;
 		};
 
 		modem2 {
 			gpio-export,name = "modem2";
-			gpio-export,output = <1>;
+			gpio-export,output = <0>;
 			gpios = <&pio 14 GPIO_ACTIVE_LOW>;
 		};
 	};
--- a/target/linux/generic/hack-6.18/800-GPIO-add-named-gpio-exports.patch
+++ b/target/linux/generic/hack-6.18/800-GPIO-add-named-gpio-exports.patch
@@ -15,7 +15,7 @@ Signed-off-by: John Crispin <blogic@openwrt.org>
  
  #include "gpiolib.h"
  #include "gpiolib-of.h"
-@@ -1302,3 +1304,61 @@ bool of_gpiochip_instance_match(struct g
+@@ -1302,3 +1304,65 @@ bool of_gpiochip_instance_match(struct g
  
  	return false;
  }
@@ -37,22 +37,26 @@ Signed-off-by: John Crispin <blogic@openwrt.org>
 +		struct gpio_desc *desc;
 +		const char *name = NULL;
 +		bool dmc;
-+		enum gpiod_flags dflags;
++		int err;
 +		int i = 0;
 +
-+		if (!fwnode_property_read_u32(child, "gpio-export,output", &val))
-+			dflags = val ? GPIOD_OUT_HIGH : GPIOD_OUT_LOW;
-+		else
-+			dflags = GPIOD_IN;
-+
 +		fwnode_property_read_string(child, "gpio-export,name", &name);
++
 +		while (true) {
-+			desc = devm_fwnode_gpiod_get_index(dev, child, NULL, i, dflags,
++			desc = devm_fwnode_gpiod_get_index(dev, child, NULL, i,
++					GPIOD_ASIS,
 +					name ? name : fwnode_get_name(child));
 +			if (PTR_ERR(desc) == -ENOENT)
 +				break;
 +			if (IS_ERR(desc))
 +				return PTR_ERR(desc);
++
++			if (!fwnode_property_read_u32(child, "gpio-export,output", &val))
++				err = gpiod_direction_output_raw(desc, !!val);
++			else
++				err = gpiod_direction_input(desc);
++			if (err)
++				return err;
 +
 +			dmc = fwnode_property_present(child, "gpio-export,direction_may_change");
 +			gpio_export_with_name(desc, dmc, name);


target/linux/ath79/dts/qca9563_tplink_archer-c7-v4.dts
```diff
@@ -43,7 +43,7 @@
 
 		gpio_shift_register_oe {
 			gpio-export,name = "tp-link:oe:sr";
-			gpio-export,output = <0>;
+			gpio-export,output = <1>;
```
openwrt-ai: The fix only covers `output = <0>`, but the regression from 0ca78951739a is symmetric, so this leaves half the affected nodes broken — and in this very file it leaves the two halves of one shift-register inconsistent.

The old path passed the value to [`gpiod_direction_output_raw()` in gpiolib-legacy.c:55](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/drivers/gpio/gpiolib-legacy.c#L55) (identical at [line 52 on 6.18](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/gpio/gpiolib-legacy.c#L52)), i.e. `FLAG_ACTIVE_LOW` was set on the descriptor but deliberately *not* applied to the initial value. `devm_fwnode_gpiod_get_index(..., GPIOD_OUT_HIGH/LOW)` applies it. So on an active-low GPIO **both** values changed meaning: `<0>` went raw LOW → raw HIGH (what this PR fixes) and `<1>` went raw HIGH → raw LOW (not fixed).

Six lines below, [`gpio_shift_register_reset`](https://github.com/openwrt/openwrt/blob/28991f67dcfb75a6010c904a15da35d9c56e0d5f/target/linux/ath79/dts/qca9563_tplink_archer-c7-v4.dts#L51-L55) still has `gpio-export,output = <1>;` with `<&gpio 21 GPIO_ACTIVE_LOW>` — that is the 74HC595 `/SRCLR`. It used to be driven raw HIGH (out of reset); after the fwnode conversion it is driven raw LOW, holding the shift register in reset, while this hunk re-enables `/OE`. Same pattern in [`qca9531_alfa-network_n2q.dts`](https://github.com/openwrt/openwrt/blob/28991f67dcfb75a6010c904a15da35d9c56e0d5f/target/linux/ath79/dts/qca9531_alfa-network_n2q.dts#L20-L31), where `poe-passthrough` (`<1>`, active-low) sits directly above the `usb-power` node this PR flips.

In total there are 36 `GPIO_ACTIVE_LOW` + `gpio-export,output = <1>` child nodes in 21 files that this PR does not touch — ath79 (`qca9563_tplink_archer-c7-v4.dts`, `qca9531_tplink_tl-mr3420-v3.dts`, `qca9531_alfa-network_n2q.dts`, `qca9533_mikrotik_routerboard-map-2nd.dts`, `qca9533_mikrotik_routerboard-95x.dtsi`, `qca9558_mikrotik_routerboard-962uigs-5hact2hnt.dts`), ramips (13 nodes in `mt7621_sercomm_na502s.dts` alone, plus `mt7621_sercomm_na502.dts`, `mt7621_unielec_u7621-01.dtsi`, `mt7621_unielec_u7621-06.dtsi`, `mt7621_asus_4g-ax56.dts`, `mt7621_zyxel_lte3301-plus.dts`, `mt7621_mikrotik_ltap-2hnd.dts`, `mt7628an_hak5_wifi-pineapple-mk7.dts`, `rt3050_arcwireless_freestation5.dts`), mediatek (`mt7981a-teltonika-rutc50.dts`, `mt7981b-openfi-6c.dts`, `mt7981b-gatonetworks-gdsp.dts`, `mt7988a-smartrg-mt-stuart-phase1.dtsi`) and **ipq40xx** (`qcom-ipq4029-mr30h.dts`, `qcom-ipq4029-z3c.dts`) — a target that isn't in this PR's diff or labels at all. Most of those are `*_reset` / `*_power` / modem-enable lines, so the inversion is not cosmetic.

Have you considered restoring the old semantics in the hack patch instead? Requesting with `GPIOD_ASIS` and then calling `gpiod_direction_output_raw()` for the `gpio-export,output` case would repair all 71 nodes in one place and keep out-of-tree DTS files working, rather than redefining the meaning of `gpio-export,output` for every consumer. If you prefer the DTS sweep, the `<1>` → `<0>` half needs to land in the same commit, otherwise `git bisect` lands on a tree where some boards are broken either way.

---

openwrt-ai (commented):
## Commit checks

- 28991f67 "kernel: fix gpio-export initial output for active-low GPIOs" — two message/diff mismatches:
  - The message (and the PR body) says "Fix all 27 gpio-export child nodes that combine GPIO_ACTIVE_LOW with gpio-export,output = <0>", but the diff flips **35** lines across 31 files. 35 is also the correct in-tree count of live `GPIO_ACTIVE_LOW` + `output = <0>` nodes (a 36th in `target/linux/ramips/dts/rt3662_edimax_br-6475nd.dts` is inside a `/* */` block and rightly skipped), so the diff is right and the number in the message is wrong.
  - Subject scope `kernel:` doesn't match the files touched — there is no kernel or patch change in this commit, only `target/linux/{ath79,lantiq,mediatek,qualcommax,ramips}/dts/*`. The in-tree convention for a cross-target DTS sweep is `treewide:` (cf. `treewide: dts: reg: use new syntax for arrays`, `treewide: convert ranges to new syntax`). If the direction changes to fixing the hack patch instead (see inline comment), `kernel:` becomes correct.


---

lorand-horvath:
@neheb Thanks for fixing these. I'm just wondering whether this is the complete story, since these particular units are only those that needed flipping from <0> to <1>. However, it seems there are other situations, such as the `mt7981a-teltonika-rutc50`, which was recently changed from <1> to <0> in https://github.com/openwrt/openwrt/pull/24829/changes
Could there be more devices like it?

neheb:
I think the patch should be fixed to restore prior behavior.

target/linux/generic/hack-6.12/800-GPIO-add-named-gpio-exports.patch
```diff
@@ -37,23 +37,27 @@ Signed-off-by: John Crispin <blogic@openwrt.org>
...
++			desc = devm_fwnode_gpiod_get_index(dev, child, NULL, i,
++					GPIOD_ASIS,
 +					name ? name : fwnode_get_name(child));
 +			if (PTR_ERR(desc) == -ENOENT)
 +				break;
 +			if (IS_ERR(desc))
 +				return PTR_ERR(desc);
 +
++			if (!fwnode_property_read_u32(child, "gpio-export,output", &val))
++				err = gpiod_direction_output_raw(desc, !!val);
```
openwrt-ai, line 55: This restores the old semantics correctly — the pre-0ca78951739a path went through `devm_gpio_request_one()`, which sets `FLAG_ACTIVE_LOW` and then calls [`gpiod_direction_output_raw(desc, !!(flags & GPIOF_OUT_INIT_HIGH))` at gpiolib-legacy.c:55](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/drivers/gpio/gpiolib-legacy.c#L55) (same on 6.18), i.e. active-low deliberately not applied to the initial value. `GPIOD_ASIS` + `gpiod_direction_output_raw()` reproduces that exactly, since [`gpiod_configure_flags()`](https://github.com/gregkh/linux/blob/25c09b42358e73e1476e517b296edb6344f2e4bd/drivers/gpio/gpiolib.c#L4528-L4570) still sets `FLAG_ACTIVE_LOW` before the early return on `GPIOD_ASIS`.

But one in-tree DTS workaround for the regression has already been merged and is not reverted here: [`f1bf2d18` ("mediatek: filogic: fix GatoNetworks GDSP GPIOs exports", merged 2026-08-21)](https://github.com/openwrt/openwrt/commit/f1bf2d18392ef7b9f5f7bc6773055cfbddde371b) flipped `modem1`/`modem2` from `<0>` to `<1>` on `GPIO_ACTIVE_LOW` pins for exactly this reason, and those [two nodes are still `<1>` at this PR's head](https://github.com/openwrt/openwrt/blob/e4f31a21f159bb11628f85a475527df7a6f46766/target/linux/mediatek/dts/mt7981b-gatonetworks-gdsp.dts#L60-L71). Once the driver goes back to raw output values, `<1>` on an active-low pin means raw HIGH again — i.e. the GDSP modems get re-broken by this commit, on a device the commit message of f1bf2d18 says was tested on both single- and multi-modem variants. Please add a revert of f1bf2d18 to this series so the tree is consistent at every commit.

Also worth a note in the PR/on the other thread: #24829 (still open, same workaround for `mt7981a-teltonika-rutc50` `pcie_reset`) becomes wrong once this lands and should be closed rather than merged.

Everything else checks out — I applied both patches to v6.12.103 and v6.18.44 with `patch -p1 --dry-run` and they land with no fuzz or offset, and the two copies of the probe function are byte-identical.

---

target/linux/generic/hack-6.12/800-GPIO-add-named-gpio-exports.patch
```diff
@@ -37,23 +37,27 @@ Signed-off-by: John Crispin <blogic@openwrt.org>
...
++			else
++				err = gpiod_direction_input(desc);
++			if (err)
++				return err;
```
openwrt-ai, line 59: nit: this is the one spot where the commit does not restore prior behaviour. Before 0ca78951739a the direction setup lived inside `devm_gpio_request_one()`, and its failure path was `continue;` — a single unconfigurable GPIO was skipped and the remaining exports on the board still came up. Returning here aborts `of_gpio_export_probe()`, so devm unwinds and *no* GPIO on the board gets exported. Intentional (loud failure on a broken DTS), or should it keep the old skip-and-continue? Same in `hack-6.18`.

---

target/linux/generic/hack-6.12/800-GPIO-add-named-gpio-exports.patch
```diff
@@ -37,23 +37,27 @@ Signed-off-by: John Crispin <blogic@openwrt.org>
...
-+			dflags = val ? GPIOD_OUT_HIGH : GPIOD_OUT_LOW;
-+		else
-+			dflags = GPIOD_IN;
-+
 +		fwnode_property_read_string(child, "gpio-export,name", &name);
++
 +		while (true) {
-+			desc = devm_fwnode_gpiod_get_index(dev, child, NULL, i, dflags,
++			desc = devm_fwnode_gpiod_get_index(dev, child, NULL, i,
++					GPIOD_ASIS,
```
openwrt-ai, line 47: nit: the PR title and body still describe the abandoned approach — "kernel: fix gpio-export initial output for active-low GPIOs" / "Fix all 27 gpio-export child nodes that combine GPIO_ACTIVE_LOW with gpio-export,output = &lt;0&gt; ... by flipping the output value to &lt;1&gt;". The branch now contains no DTS changes at all. Worth updating both so the merge context matches what is actually being done.

Same for the commit message: it explains *what* changed (request-time flags → `GPIOD_ASIS` + explicit direction call) but never states the user-visible regression it fixes — that `gpio-export,output` stopped being a raw value, so every `GPIO_ACTIVE_LOW` export came up inverted. A sentence on that would make the `Fixes:` tag self-explanatory and give people hitting the bug something to search for.

---

openwrt-ai (commented):
Re-reviewed after the force-push. The branch was rewritten from the 31-file DTS sweep into a single driver-side fix, so the commit my previous review was anchored on (`28991f67`) is no longer in the branch and this is effectively a fresh review of `e4f31a21`.

The new direction is the right one, and the commit message matches its diff. Both patch files apply to v6.12.103 and v6.18.44 with no fuzz or offset, and the two copies of the probe function are byte-identical apart from the pre-existing 6.12/6.18 difference in the `gpiolib-sysfs.c` hunks.

One thing that needs to land in the same series (inline): the DTS workaround `f1bf2d18` for GatoNetworks GDSP is already on `main` and is not reverted here, so that board gets re-inverted by this commit.

---

target/linux/generic/hack-6.12/800-GPIO-add-named-gpio-exports.patch
neheb: LLM says this is fine:

Verified the claim against all three versions. My take: keep return err; and rebut the nit.
Reasoning:
- True parity was already broken by 0ca7895 itself, not by this commit. Pre-0ca7895, a plain request failure (GPIO busy) also hit continue;; 0ca7895 turned that into return PTR_ERR(desc). Making only the direction call continue would leave it inconsistent with the IS_ERR(desc) → return two lines above it.
- Skip-and-continue is the worse failure mode here: the DTS asked for output-high/low or input and the chip refused; exporting anyway silently publishes a sysfs entry with an unapplied direction — bad for the reset/enable lines gpio-export typically drives. A broken DTS failing loudly at probe beats a half-working board.
- The path is near-unreachable in practice — the descriptor was just validated by the core, so gpiod_direction_output_raw()/input() failing implies a controller-level problem affecting every other GPIO on that chip anyway.

lorand-horvath:
@neheb @jonasjelonek Could you please merge this?

lorand-horvath:
@neheb Is there an issue with this PR? Why isn't it merged into main?

neheb:
someone else needs to.

lorand-horvath:
@robimarko Would you please merge this?

target/linux/mediatek/dts/mt7981b-gatonetworks-gdsp.dts
```diff
@@ -60,13 +60,13 @@
 
 		modem1 {
 			gpio-export,name = "modem1";
-			gpio-export,output = <1>;
+			gpio-export,output = <0>;
```
openwrt-ai, line 63: nit: the revert body is bare — just `This reverts commit f1bf2d18392ef7b9f5f7bc6773055cfbddde371b.` with no reason. [`f1bf2d18`](https://github.com/openwrt/openwrt/commit/f1bf2d18392ef7b9f5f7bc6773055cfbddde371b) was a correct workaround when it was written (it repaired a real regression from `0ca7895`, and its message says it was tested on single- and multi-modem GDSP units), so on its own this reads as undoing a tested bug fix — `git log` on this file will show "fix GatoNetworks GDSP GPIOs exports" immediately followed by an unexplained revert. One sentence naming the preceding commit in the series would make it self-contained, e.g. *"The preceding commit restores raw output values in the gpio-export driver, so the DTS workaround is no longer needed and would otherwise re-invert both modem GPIOs."* In-tree reverts usually carry that rationale ([`32e4796045a`](https://github.com/openwrt/openwrt/commit/32e4796045a), [`412526f0fee`](https://github.com/openwrt/openwrt/commit/412526f0fee), [`b5ed5604cf9`](https://github.com/openwrt/openwrt/commit/b5ed5604cf9)).

Worth stating there too: the bisect window is unavoidable, so a reader should not try to "fix" it by reordering. At [`bcf26dd7`](https://github.com/openwrt/openwrt/commit/bcf26dd79a4f3b4fefa5f6519ef413cfd6833f05) alone GDSP is inverted (`<1>` + `GPIO_ACTIVE_LOW` → `gpiod_direction_output_raw(desc, 1)` → raw HIGH), and swapping the two commits does not help either (`<0>` with the still-unfixed driver → `GPIOD_OUT_LOW` + active-low → raw HIGH as well). Only squashing would close it, and keeping the revert separate is the better trade for attribution — saying so in the message pre-empts the question.

The revert itself is exact: the resulting blob (`3689aacdc2`) is byte-identical to the pre-`f1bf2d18` file, and `<0>` on these two `GPIO_ACTIVE_LOW` pins is raw LOW again under the restored `gpiod_direction_output_raw()` path, matching pre-`0ca7895` behaviour.

---

openwrt-ai (commented):
Re-reviewed against `2596e2a6`. Since `e4f31a21` the tree changed only by the added revert commit — the two hack patches are byte-identical to what I reviewed last round — so this pass covers the two commit messages and the new DTS hunk.

- `bcf26dd7` "kernel: gpio-export: restore prior direction setup" — same diff as `e4f31a21`; the message now documents both deliberate deviations from pre-`0ca7895` behaviour (no skip-and-continue on direction-setup failure, raw output value rather than gpiod flags). Message matches the diff.
- `2596e2a6` "Revert \"mediatek: filogic: fix GatoNetworks GDSP GPIOs exports\"" — exact revert; the resulting blob matches the pre-`f1bf2d18` file byte for byte. This resolves the point raised on the previous review.

I also re-checked the surrounding tree at `2596e2a6`:

- `f1bf2d18` is the only DTS commit merged since `0ca7895` that worked around the regression, so nothing else in-tree needs reverting alongside it.
- No board added since `0ca7895` depends on the new (broken) semantics. The only post-`0ca7895` DTS file combining `GPIO_ACTIVE_LOW` with `gpio-export,output = <1>` is `mt7988a-smartrg-mt-stuart-phase1.dtsi` (`gps_enable`), but that node dates back to `c71b68acdd6` (2024) and only moved files in the 2026-08-10 restructure — it is correct under the restored semantics.
- #24829 (`mt7981a-teltonika-rutc50` `pcie_reset`) is still unmerged and the node is still `<1>` on `main`, which is the right state once this lands. That PR should be closed rather than merged, as noted previously.

CI on `2596e2a6` is green — Build Kernel, Build all core packages and Build PR Profile all completed `success`.

One nit inline on the revert's commit message.

---

target/linux/mediatek/dts/mt7981b-gatonetworks-gdsp.dts
jonasjelonek: Please give at least a small explanation. Just having "This reverts ... (full stop)" forces everyone to look around the other commits and figure out why this happened.

neheb: OK.

openwrt-ai (commented):
## Commit checks

- `5cb86b42` "Revert \"mediatek: filogic: fix GatoNetworks GDSP GPIOs exports\"" — nit: grammar in the new paragraph, "this need to be reverted" should be "this needs to be reverted".

---

target/linux/ath79/dts/qca9563_tplink_archer-c7-v4.dts
openwrt-ai: solved by fixing the driver instead of the DTS, thanks

---

target/linux/generic/hack-6.12/800-GPIO-add-named-gpio-exports.patch
openwrt-ai: fixed, thanks

---

FredApps:
Hardware before/after on an ASUS 4G-AX56 (ramips/mt7621), same physical unit, same DTS, only the branch changed.

The node under test is unchanged on both branches:

```dts
modem_reset {
	gpio-export,name = "reset:modem";
	gpio-export,output = <1>;
	gpios = <&gpio 18 GPIO_ACTIVE_LOW>;
};
```

| Build | gpio-export uses | `reset:modem` |
|---|---|---|
| 25.12 `r0-b6a820e` | `devm_gpio_request_one(&pdev->dev, gpio, flags, ...)` | **`out hi`** |
| main `r0-347ea96` | `devm_fwnode_gpiod_get_index(...)`, `dflags = val ? GPIOD_OUT_HIGH : GPIOD_OUT_LOW` | **`out lo`** |

```
# 25.12
gpio-530 ( |reset:modem  ) out hi ACTIVE LOW
gpio-524 ( |reboot:modem ) out hi

# main
gpio-18  ( |reset:modem  ) out lo ACTIVE LOW
gpio-12  ( |reboot:modem ) out hi
```

So the active-low output inverts, HIGH to LOW, exactly as described. The active-high export on the same board (`reboot:modem`, also `output = <1>`) is unchanged, which isolates it to polarity handling. The differing debugfs numbers are only the gpiochip base (512), same pins.

**Two caveats for anyone else measuring this**, both of which cost me a wrong reading first time:

1. **`gpio_switch` masks it completely.** If the board has a `ucidef_add_gpio_switch` entry, `/etc/init.d/gpio_switch` writes the logical value through `/sys/class/gpio/<name>/value` at S94 on every boot. sysfs honours `active_low`, so it normalises the pin on *both* branches and the regression disappears. My first 25.12 reading was `out lo` for this reason and was meaningless. `/etc/init.d/gpio_switch disable` and reboot before reading.
2. `git merge-base --is-ancestor 0ca78951739a HEAD` answers **NO** on a shallow clone even when the commit is in. Check which API the `800-GPIO-add-named-gpio-exports.patch` in the tree uses instead.

**On the fix itself — I think it is incomplete.** This PR flips `output = <0>` to `<1>` for the 27 nodes combining `GPIO_ACTIVE_LOW` with `output = <0>`. The node above is `GPIO_ACTIVE_LOW` with `output = <1>`, so it is not in that set, yet it is demonstrably affected. Unless I have misread the intent, the mirror case needs flipping to `<0>` and appears to have been missed. A quick grep for active-low exports with a non-zero output should turn up the rest.

On this particular board the inversion is latent rather than fatal — the modem stayed up in both states, ttyUSB0-3 present and LTE connected — but it would bite on a device where that line actually holds the modem in reset.

Happy to re-test on this hardware if you want a specific patch verified.

neheb:
That's an old version of this PR?

jonasjelonek:

lorand-horvath:
@neheb I confirm this PR fixed the main LED on my AX59U. Many thanks!

FredApps:
Post-merge verification on the same unit, same DTS node, same procedure as my earlier before/after.

| Build | gpio-export path | `reset:modem` |
| --- | --- | --- |
| 25.12 `r0-b6a820e`, kernel 6.12.103 | `devm_gpio_request_one` (raw) | `out hi` |
| main `r0-347ea96`, before this PR | `devm_fwnode_gpiod_get_index` + `GPIOD_OUT_HIGH/LOW` | `out lo` |
| main `r0-d8f9a15a` incl. 8f3285db, kernel 6.18.44 | `gpiod_direction_output_raw` | **`out hi`** |

```
# main r0-d8f9a15a, kernel 6.18.44
gpio-12  (                    |reboot:modem        ) out hi
gpio-18  (                    |reset:modem         ) out hi ACTIVE LOW
```

The node is unchanged (`GPIO_ACTIVE_LOW` with `output = <1>`), and `gpio_switch` was disabled and the board rebooted before both readings, so neither is the normalised value. The active-low export is back to the level it had on 25.12, and the active-high export on the same board never moved. The inversion is resolved here. Modem still enumerates normally afterwards (2cb7:0a05, ttyUSB0-3 present).

One correction to my earlier comment: the "I think it is incomplete" note was written against the revision that flipped `output` in the 27 active-low DTS nodes — you had already replaced that with the kernel-side fix, which is what @neheb was pointing out. The `_raw` call covers the mirror case I was worried about (`GPIO_ACTIVE_LOW` with a non-zero `output`) without touching any DTS at all, so the concern does not apply to what was merged.

carlosz1986:
FYI this also caught boards added during the window between 0ca7895 and this commit, where output = <1> on an ACTIVE_LOW line was tested as "asserted". Saw it on my Cudy LT700 (modem power enable, fixed in #24689 with output = <0>).

PR #18819, opened 2025-05-16
adb: Switch to mbedtls

pesintta:
Currently adb uses libopenssl for certain authentication tasks between the host and the target device such as certificate generation, hashing, base64 encoding and pki signatures.

Add a patch to use functionalities available in mbedtls instead.

Also switch package makefile and dependency to libmbedtls and drop patches and references to lib{crypto,openssl} as they are no longer required.

This conserves considerable amount of space on the device as openwrt ships with libmbedtls by default.

PR #24835, opened 2026-08-21
adb and adb-enablemodem: remove packages

sanderpqr:
The `adb` package is built using a 2014-era snapshot and is no longer maintained by Google. It has since changed to the `android-sdk-platform-tools` package, which is not currently supported by OpenWrt.

This old `adb` package is a blocker to the update of Mbed TLS 4.1: `010-mbedtls.patch` uses mbedtls_pk_rsa() and raw MPI arithmetic, both of which moved out of the public API with no PSA equivalent.

`adb-enablemodem` is the only package with a dependency on `adb`. This package sends a ADB command to switch certain 3G/4G dongles into modem mode. The only devices using this package are the [TP-Link TL-MR6400 v1.0](https://openwrt.org/toh/tp-link/tl-mr6400_v1) and [TP-Link Archer MR200 v1](https://openwrt.org/toh/tp-link/archer_mr200). Both are 8/64 devices.

If future support for `adb-enablemodem`, or `adb` in general, should remain available, the `android-sdk-platform-tools` package, needs to be added.


sanderpqr:
See previous discussion in https://github.com/openwrt/openwrt/pull/18819

package/network/utils/adb-enablemodem/Makefile
```diff
@@ -1,26 +0,0 @@
...
-define Build/Compile
-        true
-endef
-
-define Package/adb-enablemodem/install
-	$(INSTALL_DIR)	$(1)/etc/init.d
-	$(INSTALL_BIN)	./files/adb-enablemodem	$(1)/etc/init.d/adb-enablemodem
-endef
-
-$(eval $(call BuildPackage,adb-enablemodem))
```
openwrt-ai, line 26: The package is removed but two in-tree `DEVICE_PACKAGES` entries still reference it, leaving dangling deps:

- [`target/linux/ath79/image/generic-tp-link.mk:570-571`](https://github.com/openwrt/openwrt/blob/640917c5e86735cd79f23f356eea608f0dbf4384/target/linux/ath79/image/generic-tp-link.mk#L570-L571) — `Device/tplink_tl-mr6400-v1`
- [`target/linux/ramips/image/mt7620.mk:1294-1295`](https://github.com/openwrt/openwrt/blob/640917c5e86735cd79f23f356eea608f0dbf4384/target/linux/ramips/image/mt7620.mk#L1294-L1295) — `Device/tplink_archer-mr200`

These are exactly the two devices named in the commit message, so the removal commit should drop `adb-enablemodem` from both `DEVICE_PACKAGES` lines (leaving `kmod-usb-serial-option` as the last entry and folding the continuation back if it fits). CI here only built malta/be and x86/64 core packages, so neither image build was exercised — the unresolved package name won't show up until someone builds ath79/generic or ramips/mt7620.

---

package/network/utils/adb-enablemodem/Makefile
```diff
@@ -1,26 +0,0 @@
-include $(TOPDIR)/rules.mk
-
-PKG_NAME:=adb-enablemodem
-PKG_VERSION:=2017.03.05
```
openwrt-ai, line 4: nit: the two commit messages give different reasons for the same fact. `0a401395` ("adb-enablemodem: remove package") closes with *"If support for 8/64 devices persists in future releases, the adb-enablemodem needs to be updated to use the android-sdk-platform-tools package instead"*, while `640917c5` and the PR body say *"If future support for adb-enablemodem, or adb in general, should remain available, the android-sdk-platform-tools package needs to be added"*. The 8/64 flash-size framing is unrelated to the adb/Mbed TLS rationale and reads as a separate (unstated) argument for dropping these devices; consider aligning the first commit's closing paragraph with the wording used in the second.

---

openwrt-ai (commented):
Reviewed 2 new commits. Removal itself is complete (no leftover files under `package/utils/adb/` or `package/network/utils/adb-enablemodem/`), commit ordering is correct (dependent package first), and both commits carry a valid `Signed-off-by`. One blocking issue and one nit inline.

---

namiltd:
So by removing the odb package you disable support for TP-Link TL-MR6400 v1.0 and TP-Link Archer MR200 v1 until someone adds android-sdk-platform-tools? Their users will not be satisfied.

sanderpqr:
> So by removing the odb package you disable support for TP-Link TL-MR6400 v1.0 and TP-Link Archer MR200 v1 until someone adds android-sdk-platform-tools? Their users will not be satisfied.

No, it doesn't disable the devices. It disables a package that sets their 3G/4G modules into modem mode. Both devices only have 8MB of flash and 64MB of RAM, so support beyond 25.12 is already doubtful.

Mbed TLS is the default SSL library in OpenWrt. Main and 25.12 are currently on the 3.6 LTS branch, but support ends by March 2027. So, OpenWrt needs to update to the 4.1 LTS branch in the coming months. There is a lot of work ahead due to significant changes in the Mbed TLS APIs. The ancient 2014-era snapshot of `adb` is not compatible with these new APIs.

I agree that a PR that disables `adb`, adds `android-sdk-platform-tools` and rewrites `adb-enablemodem` is the most optimal solution. But given the amount of work to support the limited usage and age of these devices, I'm proposing to drop `adb` (and consequently, `adb-enablemodem`) and focus the limited developer time/attention to adopt the core packages to the new Mbed TLS APIs instead.

package/network/utils/adb-enablemodem/Makefile
sanderpqr: The 8/64 flash statement is relevant in the sense that support for these devices in future releases is doubtful due to their limited flash and RAM sizes, thus the work required to keep the adb-enablemodem working isn't currently worth the effort. The devices will continue working without issue on the 25.12 branch.

namiltd:
A router with a modem without the ability to run the modem limits its use.
So in my opinion adb should not be removed until an alternative is implemented.

target/linux/ramips/image/mt7620.mk
```diff
@@ -1292,7 +1292,7 @@ define Device/tplink_archer-mr200
   TPLINK_HWREV := 0x4a
   IMAGES := sysupgrade.bin
   DEVICE_PACKAGES := kmod-mt76x0e kmod-usb2 kmod-usb-net-rndis \
-	kmod-usb-serial-option adb-enablemodem
+	kmod-usb-serial-option
```
openwrt-ai, line 1295: nit: series ordering — the two `DEVICE_PACKAGES` cleanups land *after* the package removals, so at `0a40139` and `640917c` the tree references a package that no longer exists and `ath79/generic` / `ramips/mt7620` image builds break for those two commits (git-bisect hazard). CI didn't catch it because only `malta/be` and `x86/64` core-package builds plus kernel-only target builds run here — no per-device image assembly.

Reordering so the two target `.mk` commits come first (`ath79:` and `ramips: mt7620:` dropping the package from `DEVICE_PACKAGES`, then `adb-enablemodem: remove package`, then `adb: remove package`) keeps every commit buildable. Squashing the target changes into `0a40139` would work too.

---

target/linux/ath79/image/generic-tp-link.mk
```diff
@@ -568,7 +568,7 @@ define Device/tplink_tl-mr6400-v1
   DEVICE_VARIANT := v1
   TPLINK_HWID := 0x64000001
   DEVICE_PACKAGES := kmod-usb2 kmod-usb-net-rndis \
-	kmod-usb-serial-option adb-enablemodem
+	kmod-usb-serial-option
```
openwrt-ai, line 571: nit: with `adb-enablemodem` gone the line continuation is no longer needed — the remaining three packages fit on one line (72 chars). Every other wrapped `DEVICE_PACKAGES` in this file wraps because it genuinely exceeds 80 columns, so this is now the odd one out.

```suggestion
  DEVICE_PACKAGES := kmod-usb2 kmod-usb-net-rndis kmod-usb-serial-option
```

Does not apply to `tplink_archer-mr200` in [`target/linux/ramips/image/mt7620.mk:1294-1295`](https://github.com/openwrt/openwrt/blob/93d8a4a5991bcd88f9c60c397d862c86c3e7d6c0/target/linux/ramips/image/mt7620.mk#L1294-L1295) — the extra `kmod-mt76x0e` there pushes it to 85 chars, so keeping the wrap is correct.

---

openwrt-ai (commented):
Reviewed 2 new commits (`9cb3f24`, `93d8a4a`). Both fix the dangling `DEVICE_PACKAGES` references flagged in the previous review, and `git grep` confirms no `adb` / `adb-enablemodem` references remain anywhere in the tree at head. Both messages match their diffs and carry `Signed-off-by`. Two nits inline (series ordering, leftover line continuation); nothing blocking. CI on `93d8a4a` is green/pending with no failures.

---

sppmasterspp:
I want to share my personal usage experience of ADB and of course I welcome adding support for android-sdk-platform-tools but removing ADB package before that seems premature to me as it will limit my usage case.
Here is my real life usage - https://github.com/openwrt/openwrt/pull/18819#issuecomment-5439211041

> It has since changed to the `android-sdk-platform-tools` package, which is not currently supported by OpenWrt.

I don't know the reason why this isn't supported by OpenWrt.

sanderpqr:
> Here is my real life usage - [#18819 (comment)](https://github.com/openwrt/openwrt/pull/18819#issuecomment-5439211041)

Do I understand correctly from your comment that you've basically reverted that commit so that `adb` works with OpenSSL instead of Mbed TLS? That could be an alternative route if no one wants to pick up adding `android-sdk-platform-tools`, but the switch from OpenSSL to Mbed TLS was done for a reason as highlighted in the discussion. (And I still believe we shouldn't create band-aids to keep a 12 year old package alive...)

> > It has since changed to the `android-sdk-platform-tools` package, which is not currently supported by OpenWrt.
> 
> I don't know the reason why this isn't supported by OpenWrt.

I think the answer if fairly easy, no-one has made the time to add the new package...

sppmasterspp:
> Do I understand correctly from your comment that you've basically reverted that commit so that `adb` works with OpenSSL instead of Mbed TLS?

Yes. Not exactly a revert but a local new commit to restore back OpenSSL backend.

sanderpqr:
> > Do I understand correctly from your comment that you've basically reverted that commit so that `adb` works with OpenSSL instead of Mbed TLS?
> 
> Yes. Not exactly a revert but a local new commit to restore back OpenSSL backend.

Feel free to open a PR... ;)

PR #24840, opened 2026-08-22
realtek: mdio: add support for RTL9607C internal mdio bus

jameywine:
This commit adds support for internal mdio bus on RTL9607C SoC. It is used to connect to 4 internal phys on the embedded switch.

Since there is an ongoing upstream work on porting mdio-realtek-otto to mainline linux, make a separate rtl960x mdio driver for now. Once it lands there should be no reason not to merge these two into one.

cc: @plappermaul @ProMix0 from https://github.com/openwrt/openwrt/pull/24554#discussion_r3835456960

### DIFF
--- a/target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
+++ b/target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
@@ -1,0 +1,246 @@
+// SPDX-License-Identifier: GPL-2.0-only
+
+#include <linux/mfd/syscon.h>
+#include <linux/module.h>
+#include <linux/of.h>
+#include <linux/of_mdio.h>
+#include <linux/phy.h>
+#include <linux/platform_device.h>
+#include <linux/regmap.h>
+#include <linux/types.h>
+
+#define RTMD_MAX_PORTS				4
+#define RTMD_PAGE_SELECT			0x1f
+#define RTMD_RAW_PAGE(p)			((p) - 1)
+#define RTMD_960X_NUM_INT_PAGES			4096
+
+/* MDIO bus registers/fields */
+#define RTMD_960X_GPHY_IND_WD			0x0
+#define RTMD_960X_GPHY_IND_CMD			0x4
+#define   RTMD_960X_CMD_READ_C22		0
+#define   RTMD_960X_CMD_WRITE_C22		BIT(22)
+#define   RTMD_960X_CMD_RUN			BIT(21)
+#define   RTMD_960X_C22_DATA(pn, addr)		((pn) << 16 | (addr))
+#define RTMD_960X_GPHY_IND_RD			0x8
+#define   RTMD_960X_GPHY_BUSY			BIT(16)
+#define   RTMD_960X_GPHY_READ_MASK		GENMASK(15, 0)
+#define RTMD_960X_WRAP_GPHY_MISC		0x114
+#define   RTMD_960X_PHY_PATCH_DONE		BIT(0)
+#define RTMD_960X_OCP_PHY_BASE			0xa400
+
+struct rtmd_port {
+	int page;
+	bool raw;
+};
+
+struct rtmd_ctrl {
+	struct mutex lock;
+	struct regmap *map;
+	struct rtmd_port port[RTMD_MAX_PORTS];
+};
+
+static int rtmd_960x_to_ocp_addr(u32 page, u32 reg, u32 *addr)
+{
+	/* Translate phy reg to a PHY OCP address
+	 *
+	 * The phy reg 31 is reserved for page select
+	 * The phy regs 16~23 are for paged access to OCP addresses
+	 * The phy regs 0~15 and 24~30 are on base 0xa400 ocp address
+	 *
+	 * With phy registers 16~23, a user can realistically reach any
+	 * PHY OCP address, including the ones on base 0xa400.
+	 *
+	 * Masking page to 12 bits is presumably because it is the max page
+	 * of the mdio bus. Since PHY OCP Address is only 15 bits long, the
+	 * registers over 7 (exlucing 16-23 because they are substacted by 16)
+	 * overlap with page bits because they essentially shifted by 1 and
+	 * that is very strange. But that doesn't matter because all 0~15/24~30
+	 * phy registers are fixed to a standard phy ocp base of 0xa400 which
+	 * seems to be typical to realtek chips, like rtl8365mb, so there is
+	 * precedent in upstream linux for that.
+	 */
+	if (reg > 30)
+		return -EINVAL;
+	if (reg > 15 && reg < 24)
+		*addr = ((page & 0xfff) << 4) + (reg - 16) * 2;
+	else
+		*addr = RTMD_960X_OCP_PHY_BASE + reg * 2;
+
+	return 0;
+}
+
+static int rtmd_960x_internal_run_cmd(struct mii_bus *bus, u32 pn, u32 page, u32 reg, u32 cmd)
+{
+	struct rtmd_ctrl *ctrl = bus->priv;
+	u32 ocp_addr, c22_adr, val;
+	int ret;
+
+	ret = rtmd_960x_to_ocp_addr(page, reg, &ocp_addr);
+	if (ret)
+		return ret;
+
+	c22_adr = RTMD_960X_C22_DATA(pn, ocp_addr);
+
+	ret = regmap_write(ctrl->map, RTMD_960X_GPHY_IND_CMD, cmd | RTMD_960X_CMD_RUN | c22_adr);
+	if (ret)
+		return ret;
+
+	ret = regmap_read_poll_timeout(ctrl->map, RTMD_960X_GPHY_IND_RD,
+				       val, !(val & RTMD_960X_GPHY_BUSY), 20, 500000);
+	if (ret)
+		dev_warn_once(&bus->dev, "access timed out\n");
+
+	return ret;
+}
+
+static int rtmd_960x_write_int_c22(struct mii_bus *bus, u32 pn, u32 page, u32 reg, u32 val)
+{
+	struct rtmd_ctrl *ctrl = bus->priv;
+	int ret;
+
+	ret = regmap_write(ctrl->map, RTMD_960X_GPHY_IND_WD, val);
+	if (ret)
+		return ret;
+
+	return rtmd_960x_internal_run_cmd(bus, pn, page, reg, RTMD_960X_CMD_WRITE_C22);
+}
+
+static int rtmd_960x_read_int_c22(struct mii_bus *bus, u32 pn, u32 page, u32 reg, u32 *val)
+{
+	struct rtmd_ctrl *ctrl = bus->priv;
+	int ret;
+
+	ret = rtmd_960x_internal_run_cmd(bus, pn, page, reg, RTMD_960X_CMD_READ_C22);
+	if (ret)
+		return ret;
+
+	ret = regmap_read(ctrl->map, RTMD_960X_GPHY_IND_RD, val);
+	if (ret)
+		return ret;
+
+	*val &= RTMD_960X_GPHY_READ_MASK;
+
+	return ret;
+}
+
+static int rtmd_read_c22(struct mii_bus *bus, int phy, int regnum)
+{
+	struct rtmd_ctrl *ctrl = bus->priv;
+	int ret, page, val = 0;
+
+	if (phy >= RTMD_MAX_PORTS)
+		return 0xffff;
+
+	scoped_guard(mutex, &ctrl->lock) {
+		page = ctrl->port[phy].page;
+		if (regnum == RTMD_PAGE_SELECT &&
+		    page != RTMD_RAW_PAGE(RTMD_960X_NUM_INT_PAGES))
+			return page;
+
+		ctrl->port[phy].raw = (page == RTMD_RAW_PAGE(RTMD_960X_NUM_INT_PAGES));
+		ret = rtmd_960x_read_int_c22(bus, phy, page, regnum, &val);
+	}
+
+	dev_dbg(&bus->dev, "rd_PHY(phy=0x%02x, pag=0x%04x, reg=0x%04x) = 0x%04x, ret = %d\n",
+		phy, page, regnum, val, ret);
+
+	return ret ? ret : val;
+}
+
+static int rtmd_write_c22(struct mii_bus *bus, int phy, int regnum, u16 val)
+{
+	struct rtmd_ctrl *ctrl = bus->priv;
+	bool do_write = false;
+	int ret = 0, page;
+
+	if (phy >= RTMD_MAX_PORTS)
+		return -ENODEV;
+
+	scoped_guard(mutex, &ctrl->lock) {
+		page = ctrl->port[phy].page;
+		if (regnum == RTMD_PAGE_SELECT)
+			ctrl->port[phy].page = val;
+
+		do_write = !ctrl->port[phy].raw &&
+			   (regnum != RTMD_PAGE_SELECT ||
+			    page == RTMD_RAW_PAGE(RTMD_960X_NUM_INT_PAGES));
+
+		if (do_write) {
+			ctrl->port[phy].raw = (page == RTMD_RAW_PAGE(RTMD_960X_NUM_INT_PAGES));
+			ret = rtmd_960x_write_int_c22(bus, phy, page, regnum, val);
+		} else {
+			ctrl->port[phy].raw = false;
+		}
+	}
+
+	if (do_write)
+		dev_dbg(&bus->dev,
+			"wr_PHY(phy=0x%02x, pag=0x%04x, reg=0x%04x, val=0x%04x), ret = %d\n",
+			phy, page, regnum, val, ret);
+
+	return ret;
+}
+
+static int rtmd_probe(struct platform_device *pdev)
+{
+	struct device *dev = &pdev->dev;
+	struct rtmd_ctrl *ctrl;
+	struct mii_bus *bus;
+	int ret;
+
+	bus = devm_mdiobus_alloc_size(dev, sizeof(*ctrl));
+	if (!bus)
+		return -ENOMEM;
+
+	ctrl = bus->priv;
+
+	ret = devm_mutex_init(dev, &ctrl->lock);
+	if (ret)
+		return ret;
+
+	ctrl->map = syscon_node_to_regmap(pdev->dev.of_node->parent);
+	if (IS_ERR(ctrl->map))
+		return PTR_ERR(ctrl->map);
+
+	/*
+	 * PHY_PATCH_DONE enables phy control via SoC. This is required for phy access,
+	 * including patching. Must always be set before the phys are probed.
+	 */
+	ret = regmap_update_bits(ctrl->map, RTMD_960X_WRAP_GPHY_MISC,
+				 RTMD_960X_PHY_PATCH_DONE, RTMD_960X_PHY_PATCH_DONE);
+	if (ret)
+		return ret;
+
+	bus->name = "Realtek MDIO bus";
+	bus->read = rtmd_read_c22;
+	bus->write = rtmd_write_c22;
+	bus->parent = dev;
+	snprintf(bus->id, MII_BUS_ID_SIZE, "realtek-mdio-int");
+
+	ret = devm_of_mdiobus_register(dev, bus, dev->of_node);
+	if (ret)
+		return dev_err_probe(dev, ret, "cannot register MDIO bus\n");
+
+	platform_set_drvdata(pdev, bus);
+
+	return 0;
+}
+
+static const struct of_device_id rtmd_ids[] = {
+	{ .compatible = "realtek,rtl9607-int-mdio" },
+	{ /* sentinel */ }
+};
+MODULE_DEVICE_TABLE(of, rtmd_ids);
+
+static struct platform_driver rtmd_driver = {
+	.probe = rtmd_probe,
+	.driver = {
+		.name = "mdio-rtl960x-otto",
+		.of_match_table = rtmd_ids,
+	},
+};
+
+module_platform_driver(rtmd_driver);
+
+MODULE_DESCRIPTION("Realtek Otto RTL960X MDIO driver");
+MODULE_LICENSE("GPL");
--- a/target/linux/realtek/patches-6.18/820-add-rtl960x-mdio-driver.patch
+++ b/target/linux/realtek/patches-6.18/820-add-rtl960x-mdio-driver.patch
@@ -1,0 +1,26 @@
+--- a/drivers/net/mdio/Kconfig
++++ b/drivers/net/mdio/Kconfig
+@@ -216,6 +216,13 @@ config MDIO_REALTEK_OTTO_SERDES
+         help
+           This driver provides MDIO access to the RTL83xx/RTL93xx SerDes.
+ 
++config MDIO_REALTEK_OTTO_RTL960X
++	tristate "Realtek Otto RTL960X MDIO interface support"
++	default MACH_REALTEK_RTL && RTL960X
++	depends on (MACH_REALTEK_RTL && RTL960X) || COMPILE_TEST
++	help
++	  This driver supports the internal MDIO bus on RTL960x SoCs.
++
+ config MDIO_THUNDER
+ 	tristate "ThunderX SOCs MDIO buses"
+ 	depends on 64BIT
+--- a/drivers/net/mdio/Makefile
++++ b/drivers/net/mdio/Makefile
+@@ -25,6 +25,7 @@ obj-$(CONFIG_MDIO_REGMAP)		+= mdio-regma
+ obj-$(CONFIG_MDIO_REALTEK_OTTO)	+= mdio-realtek-otto.o
+ obj-$(CONFIG_MDIO_REALTEK_OTTO_AUX)	+= mdio-realtek-otto-aux.o
+ obj-$(CONFIG_MDIO_REALTEK_OTTO_SERDES)	+= mdio-realtek-otto-serdes.o
++obj-$(CONFIG_MDIO_REALTEK_OTTO_RTL960X)	+= mdio-realtek-otto-rtl960x.o
+ obj-$(CONFIG_MDIO_SMBUS)		+= mdio-smbus.o
+ obj-$(CONFIG_MDIO_SUN4I)		+= mdio-sun4i.o
+ obj-$(CONFIG_MDIO_THUNDER)		+= mdio-thunder.o


jameywine:
oh fun, accidentally put the comma in between u32 and reg in the `rtmd_960x_internal_run_cmd` function while i was cleaning the patch in prep for pull request.
That is fixed now.

plappermaul:
From looking at the code this is nothing that will survive the upstreaming of the MDIO driver. That alone is hard enough  - see https://lore.kernel.org/netdev/20260817162632.1208891-1-markus.stockhausen@gmx.de/

Please make a separate MDIO driver from that.

ProMix0:
@jameywine you could take separated driver [here](https://github.com/ProMix0/openwrt/commit/e4f0c4e0ef48f82cbd36d2b32479e64a944adf2d#diff-67f6978a7e2528eb9191c706a80a62e3f680b8a19d67525b3891a95721c93a4c)

But, again, I don't see any benefits in moving it out from DSA

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto.c
```diff
@@ -199,6 +202,19 @@
 #define RTMD_931X_SMI_10GPHY_POLLING_SEL3	0x0cfc
 #define RTMD_931X_SMI_10GPHY_POLLING_SEL4	0x0d00
 
+#define RTMD_960X_GPHY_IND_WD			(0x0)
+#define RTMD_960X_GPHY_IND_CMD			(0x4)
+#define   RTMD_960X_CMD_READ_C22		0
+#define   RTMD_960X_CMD_WRITE_C22		BIT(22)
+#define   RTMD_960X_CMD_RUN			BIT(21)
+#define   RTMD_960X_C22_DATA(addr, pn)		((pn) << 16 | (addr))
```
openwrt-ai: The parameters are declared `(addr, pn)` but the only call site passes `RTMD_960X_C22_DATA(pn, ocp_addr)` (line 563), so the macro expands to `ocp_addr << 16 | pn` — the arguments are effectively swapped.

That is not just cosmetic: `rtmd_960x_to_ocp_addr()` produces addresses up to `0xffff` (`((page & 0xfff) << 4) + 14` and `0xa400 + 60`), so `ocp_addr << 16` overwrites bits 21 and 22 of `RTMD_960X_GPHY_IND_CMD` — exactly `RTMD_960X_CMD_RUN` and `RTMD_960X_CMD_WRITE_C22` — while the port number lands in the address field. The only layout that fits the register is address in bits 15:0 and port in bits 20:16, i.e. `(pn) << 16 | (addr)`, which is what the macro body already does.

The other `*_C22_DATA` macros in this file are all declared in call order (`(page, reg)` / called `(page, reg)`), so renaming the parameters here keeps that convention and fixes the call site:

```suggestion
#define   RTMD_960X_C22_DATA(pn, addr)		((pn) << 16 | (addr))
```

---

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto.c
```diff
@@ -522,6 +538,69 @@ static int rtmd_931x_write_c45(struct mii_bus *bus, u32 pn, u32 devnum, u32 regn
...
+	return ret;
+}
+
+static int rtmd_960x_write_int_c22(struct mii_bus *bus, u32 pn, u32 page, u32 reg, u32 val)
+{
+	struct rtmd_ctrl *ctrl = rtmd_bus_to_ctrl(bus);
+
+	regmap_write(ctrl->map, RTMD_960X_GPHY_IND_WD, val);
+
+	return rtmd_960x_internal_run_cmd(bus, pn, page, reg, RTMD_960X_CMD_WRITE_C22);
```
openwrt-ai: The return value of this `regmap_write()` is dropped. If loading the write-data register fails, the command is still issued and the function reports whatever `rtmd_960x_internal_run_cmd()` returns, so a failed MDIO write can be reported as success with stale data written to the PHY. Every other regmap access in this driver checks its return.

```suggestion
	struct rtmd_ctrl *ctrl = rtmd_bus_to_ctrl(bus);
	int ret;

	ret = regmap_write(ctrl->map, RTMD_960X_GPHY_IND_WD, val);
	if (ret)
		return ret;

	return rtmd_960x_internal_run_cmd(bus, pn, page, reg, RTMD_960X_CMD_WRITE_C22);
```

---

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto.c
```diff
@@ -78,6 +78,9 @@
 #define RTMD_931X_NUM_BUSES			4
 #define RTMD_931X_NUM_PAGES			8192
 #define RTMD_931X_NUM_PORTS			56
+#define RTMD_960X_NUM_INT_BUSES			1
+#define RTMD_960X_NUM_INT_PAGES			4096
+#define RTMD_960X_NUM_INT_PORTS			32
```
openwrt-ai: `num_ports` is not a port count in this driver — it is the CPU port index. [`rtmd_map_ports()`](``https://github.com/openwrt/openwrt/blob/48764b051412d07c8da9b156d15a1c30d7e551d4/target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto.c#L1085-L1089)`` skips `pn == num_ports` as the CPU port and rejects `pn > num_ports`. All four existing configs match the SoC CPU port exactly (`RTMD_838X_NUM_PORTS` 28 / `RTL838X_CPU_PORT` 28, 839x 52/52, 930x 28/28, 931x 56/56 — see [mach-rtl-otto.h](``https://github.com/openwrt/openwrt/blob/48764b051412d07c8da9b156d15a1c30d7e551d4/target/linux/realtek/files-6.18/arch/mips/include/asm/mach-rtl-otto/mach-rtl-otto.h#L46-L50)),`` and `RTL9607_CPU_PORT` is 9.

With 32 here, a `port@9` CPU node in the DT is not recognised as the CPU port (it falls through into `sds_ports` instead), and the bounds check accepts port numbers 10..32 that do not exist on this SoC.

```suggestion
#define RTMD_960X_NUM_INT_PORTS			9
```

---

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto.c
```diff
@@ -522,6 +538,69 @@ static int rtmd_931x_write_c45(struct mii_bus *bus, u32 pn, u32 devnum, u32 regn
...
+		*addr = RTMD_960X_OCP_PHY_BASE + reg * 2;
+
+	return 0;
+}
+
+static int rtmd_960x_internal_run_cmd(struct mii_bus *bus, u32 pn, u32 page, u32 reg, u32 cmd)
+{
+	struct rtmd_ctrl *ctrl = rtmd_bus_to_ctrl(bus);
+	u32 ocp_addr, c22_adr;
+	int ret, val;
```
openwrt-ai: nit: `val` is handed to `regmap_read_poll_timeout()`, which expands to `regmap_read(map, addr, &val)` and therefore expects `unsigned int *`. Declaring it `int` is a signedness mismatch with the regmap API; the equivalent variable in [`rtmd_run_cmd()`](``https://github.com/openwrt/openwrt/blob/48764b051412d07c8da9b156d15a1c30d7e551d4/target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto.c#L306``) is `u32 cmdstate;`.

```suggestion
	u32 ocp_addr, c22_adr, val;
	int ret;
```

---

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto.c
```diff
@@ -522,6 +538,69 @@ static int rtmd_931x_write_c45(struct mii_bus *bus, u32 pn, u32 devnum, u32 regn
 	return rtmd_run_cmd(bus, RTMD_931X_CMD_WRITE_C45, &cmd_data, NULL);
 }
 
+static int rtmd_960x_to_ocp_addr(u32 page, u32 reg, u32 *addr)
+{
+	if (reg > 30)
+		return -EINVAL;
+	if (reg > 15  && reg < 24)
```
openwrt-ai: nit: double space before `&&`.

```suggestion
	if (reg > 15 && reg < 24)
```

---

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto.c
```diff
@@ -199,6 +202,19 @@
 #define RTMD_931X_SMI_10GPHY_POLLING_SEL3	0x0cfc
 #define RTMD_931X_SMI_10GPHY_POLLING_SEL4	0x0d00
 
+#define RTMD_960X_GPHY_IND_WD			(0x0)
```
openwrt-ai: nit: plain register offsets are written without parentheses everywhere else in this header block (`0xa100`, `0x03dc`, `0xca00`, `0x0c00`, ...; the sole exception is `RTMD_839X_PHYREG_CTRL`). Same applies to lines 206, 211, 214 and 216 below.

```suggestion
#define RTMD_960X_GPHY_IND_WD			0x0
```

---

openwrt-ai (commented):
Reviewed 1 new commit. The commit message matches the diff, and CI on `48764b0` is green, so no build-grounded findings.

Main things found: an argument-order mismatch in `RTMD_960X_C22_DATA` that puts the OCP address on top of the RUN/WRITE command bits, an unchecked `regmap_write()` in the C22 write path, and `RTMD_960X_NUM_INT_PORTS` not matching the CPU-port convention the other four configs follow. Details inline.

Note this review only covers the code as written against this driver; it does not weigh in on @plappermaul's request to split the internal bus into a separate MDIO driver, which would supersede most of the placement questions here.

---

jameywine:
> From looking at the code this is nothing that will survive the upstreaming of the MDIO driver. That alone is hard enough - see https://lore.kernel.org/netdev/20260817162632.1208891-1-markus.stockhausen@gmx.de/

Interesting, it should be possible to add that to the upstream version later on when it happens. I just don't think there is any hard differences that would sharing a single mdio driver a non option. But that is for upstream maintainers to decide when it happens.

> Please make a separate MDIO driver from that.

But until then, i am cool with a separate mdio driver until your mdio driver lands.

> @jameywine you could take separated driver [here](https://github.com/ProMix0/openwrt/commit/e4f0c4e0ef48f82cbd36d2b32479e64a944adf2d#diff-67f6978a7e2528eb9191c706a80a62e3f680b8a19d67525b3891a95721c93a4c)

Thanks, it would definitely require some cleanup that i can take care of.

> But, again, I don't see any benefits in moving it out from DSA

I think it can go both ways really. There are certainly cases of some being separated and some being together with dsa stuff... but for now, it is just to get things going forward.

jameywine:
As requested, moved the rtl960x to a new mdio driver.

@ProMix0 Since your version was based on the older mdio-realtek-otto, i had change quite a bit to match newer one and so can you check if it is still functional?

jameywine:
Changed the patch prefix to 820 as i did not notice the patch 809 coming after 807 that was also changing mdio Kconfig and Makefile files.

jameywine:
changed the new config to depend both MACH_REALTEK_RTL and RTL960X to fix the build issue on switch chips

target/linux/realtek/patches-6.18/820-add-rtl960x-mdio-driver.patch
```diff
@@ -0,0 +1,26 @@
...
+@@ -216,6 +216,13 @@ config MDIO_REALTEK_OTTO_SERDES
+         help
+           This driver provides MDIO access to the RTL83xx/RTL93xx SerDes.
+ 
++config MDIO_REALTEK_OTTO_RTL960X
++	tristate "Realtek Otto RTL960X MDIO interface support"
++	default MACH_REALTEK_RTL && RTL960X
++	depends on (MACH_REALTEK_RTL && RTL960X) || COMPILE_TEST
++	help
++	  This driver supports the internal MDIO bus on RTL960x SoCs.
```
ProMix0, line 12: Looks like DSA should depend on it?

ProMix0:
> @ProMix0 Since your version was based on the older mdio-realtek-otto, i had change quite a bit to match newer one and so can you check if it is still functional?

Except for a few obvious mistakes, it is

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
```diff
@@ -0,0 +1,276 @@
...
+struct rtmd_ctrl {
+	struct mutex lock;
+	struct regmap *map;
+	const struct rtmd_config *cfg;
+	struct rtmd_port port[RTMD_MAX_PHY];
+};
+
+struct rtmd_chan {
+	struct rtmd_ctrl *ctrl;
+	s8 port[PHY_MAX_ADDR];
```
ProMix0: Not initialized, and the purpose of this port mapping is questionable

target/linux/realtek/patches-6.18/820-add-rtl960x-mdio-driver.patch
jameywine: that would belong with dsa config would it not?

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
jameywine: i see...

i guess some parts of port mapping from `rtmd_map_ports` should come over here then? Unless i am overcomplicating things

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
```diff
@@ -0,0 +1,276 @@
...
+struct rtmd_port {
+	int page;
+	bool raw;
+};
+
+struct rtmd_ctrl {
+	struct mutex lock;
+	struct regmap *map;
+	const struct rtmd_config *cfg;
+	struct rtmd_port port[RTMD_MAX_PHY];
```
openwrt-ai: `RTMD_MAX_PHY` is not defined anywhere — not in this file, not in the parent [`mdio-realtek-otto.c`](``https://github.com/openwrt/openwrt/blob/f6618038305ec65676e64bd7ec9ad5930e1a61c7/target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto.c),`` and not in any kernel header pulled in here. This is a hard compile error (`error: 'RTMD_MAX_PHY' undeclared`). Conversely [`RTMD_MAX_PORTS`](``https://github.com/openwrt/openwrt/blob/f6618038305ec65676e64bd7ec9ad5930e1a61c7/target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c#L13)`` is defined but never used, and the parent driver sizes the same array with it ([line 224](``https://github.com/openwrt/openwrt/blob/f6618038305ec65676e64bd7ec9ad5930e1a61c7/target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto.c#L224)).``

```suggestion
	struct rtmd_port port[RTMD_MAX_PORTS];
```

---

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
```diff
@@ -0,0 +1,276 @@
...
+		return ret;
+
+	*val &= RTMD_960X_GPHY_READ_MASK;
+
+	return ret;
+}
+
+static int rtmd_read_c22(struct mii_bus *bus, int phy, int regnum)
+{
+	struct rtmd_ctrl *ctrl = rtmd_ctrl_from_bus(bus);
```
openwrt-ai: `rtmd_ctrl_from_bus()` does not exist. The helper defined in this file (and in the parent driver) is [`rtmd_bus_to_ctrl()`](``https://github.com/openwrt/openwrt/blob/f6618038305ec65676e64bd7ec9ad5930e1a61c7/target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c#L55-L58),`` which is what every other function here uses — a second compile error in the same file.

```suggestion
	struct rtmd_ctrl *ctrl = rtmd_bus_to_ctrl(bus);
```

---

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
```diff
@@ -0,0 +1,276 @@
...
+static struct rtmd_ctrl *rtmd_bus_to_ctrl(struct mii_bus *bus)
+{
+	return ((struct rtmd_chan *)bus->priv)->ctrl;
+}
+
+static int rtmd_phy_to_port(struct mii_bus *bus, int phy)
+{
+	struct rtmd_chan *chan = bus->priv;
+
+	return chan->port[phy];
```
openwrt-ai: Beyond the missing initialisation @ProMix0 already noted on `chan->port[]`: the returned `pn` is used unchecked as an index into `ctrl->port[]`, which has only `RTMD_MAX_PORTS` (4) entries, while `chan->port[]` is `s8` and covers all `PHY_MAX_ADDR` (32) MDIO addresses. Only `pn < 0` is rejected in `rtmd_read_c22()`/`rtmd_write_c22()`, so any port number 4..127 that ends up in the table becomes an out-of-bounds read *and write* of `ctrl->port[pn].page` / `.raw`.

The parent driver gets this bound from `rtmd_map_ports()`, which rejects [`pn > ctrl->cfg->num_ports`](``https://github.com/openwrt/openwrt/blob/f6618038305ec65676e64bd7ec9ad5930e1a61c7/target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto.c#L996-L999)`` before ever storing into the reverse-lookup table. Whatever form the port setup takes here, please validate against `RTMD_MAX_PORTS` at the point the table is filled (or add an upper-bound check here).

---

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
```diff
@@ -0,0 +1,276 @@
...
+	ctrl->map = syscon_node_to_regmap(pdev->dev.of_node->parent);
+	if (IS_ERR(ctrl->map))
+		return PTR_ERR(ctrl->map);
+
+	/*
+	 * PHY_PATCH_DONE enables phy control via SoC. This is required for phy access,
+	 * including patching. Must always be set before the phys are probed.
+	 */
+	regmap_update_bits(ctrl->map, RTMD_960X_WRAP_GPHY_MISC,
+			   RTMD_960X_PHY_PATCH_DONE, RTMD_960X_PHY_PATCH_DONE);
```
openwrt-ai: The return value is dropped, but the comment right above says this bit is *required* for PHY access. If the write fails, probe continues and registers the MDIO bus anyway, and every subsequent PHY access silently misbehaves instead of the driver failing to probe. Every other regmap call in this file checks its return.

```suggestion
	ret = regmap_update_bits(ctrl->map, RTMD_960X_WRAP_GPHY_MISC,
				 RTMD_960X_PHY_PATCH_DONE, RTMD_960X_PHY_PATCH_DONE);
	if (ret)
		return ret;
```

---

target/linux/realtek/patches-6.18/820-add-rtl960x-mdio-driver.patch
```diff
@@ -0,0 +1,26 @@
+--- a/drivers/net/mdio/Kconfig
++++ b/drivers/net/mdio/Kconfig
+@@ -216,6 +216,13 @@ config MDIO_REALTEK_OTTO_SERDES
+         help
+           This driver provides MDIO access to the RTL83xx/RTL93xx SerDes.
+ 
++config MDIO_REALTEK_OTTO_RTL960X
++	tristate "Realtek Otto RTL960X MDIO interface support"
++	default MACH_REALTEK_RTL && RTL960X
++	depends on (MACH_REALTEK_RTL && RTL960X) || COMPILE_TEST
```
openwrt-ai, line 10: There is no `config RTL960X` symbol in this tree — [300-01-enhance-realtek-platform-config.patch](``https://github.com/openwrt/openwrt/blob/f6618038305ec65676e64bd7ec9ad5930e1a61c7/target/linux/realtek/patches-6.18/300-01-enhance-realtek-platform-config.patch#L43-L71)`` only adds `RTL838X`, `RTL839X`, `RTL930X` and `RTL931X`, and no subtarget under `target/linux/realtek/` sets `CONFIG_RTL960X`. Kconfig treats the undefined symbol as `n`, so `default` never evaluates to `y` and the option is only reachable via `COMPILE_TEST`. No `config-6.18` in this PR adds `CONFIG_MDIO_REALTEK_OTTO_RTL960X=y` either (the other four subtargets list the three existing `MDIO_REALTEK_OTTO*` symbols explicitly).

Net effect: nothing in the tree compiles `mdio-realtek-otto-rtl960x.c`, which is why CI is green on `f661803` despite the two undeclared identifiers in it.

In-tree [717-net-ethernet-add-rtl960x-gmac-probe-driver.patch](``https://github.com/openwrt/openwrt/blob/f6618038305ec65676e64bd7ec9ad5930e1a61c7/target/linux/realtek/patches-6.18/717-net-ethernet-add-rtl960x-gmac-probe-driver.patch#L29)`` already carries the same `depends on RTL960X`, so I assume this is deliberately waiting on a separate RTL960X platform/subtarget series — can you confirm, and is that series a prerequisite for this one?

---

openwrt-ai (commented):
Re-reviewed after the force-push from `48764b0` to `f661803`, which restructured the change from an addition to `mdio-realtek-otto.c` into a standalone `mdio-realtek-otto-rtl960x.c` driver plus patch 820. Everything is new relative to my last review; all six of my earlier inline comments are now outdated and the three code findings among them (`RTMD_960X_C22_DATA` argument order, unchecked `regmap_write()` in the C22 write path, `u32 val` for `regmap_read_poll_timeout()`) are addressed in the new file.

The main new findings: the file does not compile — `RTMD_MAX_PHY` and `rtmd_ctrl_from_bus()` are both undeclared — and nothing in the tree currently builds it, since `CONFIG_RTL960X` does not exist and no subtarget config enables `MDIO_REALTEK_OTTO_RTL960X`. That combination is why CI passes. Also flagged an unchecked `regmap_update_bits()` in probe and the missing upper bound on the port index. Details inline.

## Commit checks

- `f661803` "realtek: mdio: add support for RTL9607C internal mdio bus" — the body says port mapping was among the things *not* carried over from the parent driver, but the diff keeps the `rtmd_chan.port[]` reverse-lookup table and `rtmd_phy_to_port()` while dropping only the code that fills them, which is the uninitialised-table problem @ProMix0 raised. Either the table should go too, or the message should say the mapping is still to be added. Two typos in the same paragraph: "seperate" → "separate", "expect for" → "except for".

---

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
ProMix0: I can't see any port mapping useful for this driver

I'm quite surprised that DSA API handles ports as plain numbers - I've expected some device structure, potentially associated with correct MDIO bus and PHY bus id. So yes, looks like mapping should be done in DSA driver

target/linux/realtek/patches-6.18/820-add-rtl960x-mdio-driver.patch
ProMix0: Not sure. Still skeptical about separation thing

jameywine:
at least small steps at a time, removed the reminiscence of port mapping and some other fixes noticed by the claude review bot.

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
jameywine: that is gone now, and hopefully is still okay.

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
```diff
@@ -0,0 +1,268 @@
...
+	return ret;
+}
+
+static int rtmd_read_c22(struct mii_bus *bus, int phy, int regnum)
+{
+	struct rtmd_ctrl *ctrl = rtmd_bus_to_ctrl(bus);
+	int ret, page, val = 0;
+
+	if (phy >= RTMD_MAX_PORTS)
+		return -ENODEV;
```
ProMix0: `0xffff`?

ProMix0:
Works now :+1:

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
jameywine: what exactly do you mean?

ProMix0: I think we discussed it somewhere. If there is no device at specified address, drivers should return 0xffff. There was some examples in tree, and `mdio` utility interprets negative values as errors during bus scan

jameywine: sure i guess. I did see some dsa drivers do that but not the rtl8365mb dsa driver which i find more surprising as it is very related to us.

jameywine:
changed it to return 0xffff and fixed some typos

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
```diff
@@ -0,0 +1,268 @@
...
+
+struct rtmd_chan {
+	struct rtmd_ctrl *ctrl;
+};
+
+struct rtmd_config {
+	u16 num_pages;
+	int (*read_c22)(struct mii_bus *bus, u32 pn, u32 page, u32 reg, u32 *val);
+	int (*write_c22)(struct mii_bus *bus, u32 pn, u32 page, u32 reg, u32 val);
+};
```
plappermaul: Now that this driver is self contained you can remove that structure and call the functions directly.

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
```diff
@@ -0,0 +1,268 @@
...
+{
+	return ((struct rtmd_chan *)bus->priv)->ctrl;
+}
+
+static int rtmd_960x_to_ocp_addr(u32 page, u32 reg, u32 *addr)
+{
+	if (reg > 30)
+		return -EINVAL;
+	if (reg > 15 && reg < 24)
+		*addr = ((page & 0xfff) << 4) + (reg - 16) * 2;
```
plappermaul, line 65: This is not self explanatory. Why just and the page with that constant?  Additionally Add a meaningful comment here.

plappermaul (changes requested):

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
```diff
@@ -0,0 +1,268 @@
...
+struct rtmd_ctrl {
+	struct mutex lock;
+	struct regmap *map;
+	const struct rtmd_config *cfg;
+	struct rtmd_port port[RTMD_MAX_PORTS];
+};
+
+struct rtmd_chan {
+	struct rtmd_ctrl *ctrl;
+};
```
openwrt-ai: Now that `s8 port[PHY_MAX_ADDR]` is gone, `struct rtmd_chan` wraps a single pointer and buys nothing: `rtmd_probe_bus()` allocates 8 bytes of `bus->priv` just to store `ctrl`, and every accessor pays a double dereference through [`rtmd_bus_to_ctrl()`](``https://github.com/openwrt/openwrt/blob/17fbb2518ada5112ac04cd893b8a371766d1251e/target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c#L54-L57).`` In the parent driver the struct exists because it carries per-bus state (the SMI bus index and the phy→port table); here there is no per-bus state left.

Dropping it means: `bus = devm_mdiobus_alloc(dev);` + `bus->priv = ctrl;` in `rtmd_probe_bus()`, and `rtmd_bus_to_ctrl()` becoming `return bus->priv;`. Not suggesting a block since the change spans three hunks.

Same flavour of simplification as @plappermaul's note on `struct rtmd_config` above.

---

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
```diff
@@ -0,0 +1,268 @@
...
+	ctrl->map = syscon_node_to_regmap(pdev->dev.of_node->parent);
+	if (IS_ERR(ctrl->map))
+		return PTR_ERR(ctrl->map);
+
+	/*
+	 * PHY_PATCH_DONE enables phy control via SoC. This is required for phy access,
+	 * including patching. Must always be set before the phys are probed.
+	 */
+	ret = regmap_update_bits(ctrl->map, RTMD_960X_WRAP_GPHY_MISC,
+			   RTMD_960X_PHY_PATCH_DONE, RTMD_960X_PHY_PATCH_DONE);
```
openwrt-ai: nit: adding `ret = ` shifted the call six columns right but the continuation line kept its old indentation, so the second line no longer lines up with the opening paren. The parent driver aligns continuations to the paren throughout, e.g. [`rtmd_run_cmd()` at mdio-realtek-otto.c:313-314](``https://github.com/openwrt/openwrt/blob/17fbb2518ada5112ac04cd893b8a371766d1251e/target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto.c#L313-L314),`` as does `regmap_read_poll_timeout()` earlier in this file.

```suggestion
	ret = regmap_update_bits(ctrl->map, RTMD_960X_WRAP_GPHY_MISC,
				 RTMD_960X_PHY_PATCH_DONE, RTMD_960X_PHY_PATCH_DONE);
```

---

openwrt-ai (commented):
Re-reviewed the delta from `f661803` to `17fbb25` (removal of `rtmd_phy_to_port()`/`rtmd_chan.port[]`, direct `phy` indexing with a `RTMD_MAX_PORTS` bound, `0xffff` for absent addresses, and the `regmap_update_bits()` return check).

All four of my previous code findings are addressed: `RTMD_MAX_PHY` and `rtmd_ctrl_from_bus()` are gone, the port-index upper bound is now enforced in both `rtmd_read_c22()` and `rtmd_write_c22()`, and probe fails if `PHY_PATCH_DONE` cannot be set. The commit message also now matches the diff (the port-mapping leftovers it claimed were dropped really are dropped) and the two typos are fixed, so no commit checks this round. CI on `17fbb25` is green.

Only two small things left from the new delta, both inline: `struct rtmd_chan` is now a one-pointer wrapper that can go away, and the continuation line of the `regmap_update_bits()` call lost its paren alignment.

Still open from my last review and not addressed here: [`CONFIG_RTL960X` does not exist in this tree](https://github.com/openwrt/openwrt/pull/24840#discussion_r3838931013), so nothing builds this file — that thread is unanswered, and it is why the green CI above says little about the driver itself.

---

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
jameywine: removed all of it along with channel stuff

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
jameywine: i tried to explain to my best ability but i just follow what realtek does in their code. I dunno why it is the way it is

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
jameywine: removed it like suggested

target/linux/realtek/patches-6.18/820-add-rtl960x-mdio-driver.patch
jameywine: Yes, there is no RTL960X config and its tied to another pull request that adds rtl960x subtarget to realtek.

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
ProMix0: Registers 0-15 is a standard ones and defined in C22. The memory with corresponding values have `0xa400` as a base value. Other registers are vendor-specific. For realtek it traditionally means that registers 16-23 is for paged access (emulated for MDIO, actually just use page address as a base)

That looks like description of logic behind the code. It could be partially moved to code (i.e. `IS_C22_STANDARD_REGISTER(reg)`)

target/linux/realtek/patches-6.18/820-add-rtl960x-mdio-driver.patch
```diff
@@ -0,0 +1,26 @@
+--- a/drivers/net/mdio/Kconfig
++++ b/drivers/net/mdio/Kconfig
+@@ -216,6 +216,13 @@ config MDIO_REALTEK_OTTO_SERDES
+         help
+           This driver provides MDIO access to the RTL83xx/RTL93xx SerDes.
+ 
++config MDIO_REALTEK_OTTO_RTL960X
```
plappermaul, line 7: Final nit, While you share some code with the rest of the target this is no Otto platform. Please use a better naming for the driver file and the CONFIG.

plappermaul (changes requested):

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
plappermaul: That is the same logic as for the Otto MDIO bus. Nevertheless SDK and OpenWrt does not nail the lower registers to a fixed page.

ProMix0: > Nevertheless SDK and OpenWrt does not nail the lower registers to a fixed page

Well, [it does](https://github.com/jameywine/GPL-for-GP3000/blob/5090fbcb6ba743dd7b1314811ef557bad0460147/linux-5.10.x/drivers/net/ethernet/realtek/rtl86900/sdk/src/hal/mac/drv.c#L11291)

plappermaul: I talked about Otto.

ProMix0: Just like you said, it isn't Otto

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
```diff
@@ -0,0 +1,239 @@
...
+
+#include <linux/fwnode.h>
+#include <linux/fwnode_mdio.h>
+#include <linux/mfd/syscon.h>
+#include <linux/of.h>
+#include <linux/of_mdio.h>
+#include <linux/phy.h>
+#include <linux/platform_device.h>
+#include <linux/regmap.h>
+#include <linux/types.h>
```
openwrt-ai: `<linux/fwnode.h>` and `<linux/fwnode_mdio.h>` are leftovers from the parent driver — no `fwnode_*` identifier appears anywhere in this file. The parent needs them because it walks the DT via fwnode ([`rtmd_get_bus_node()` at mdio-realtek-otto.c:756](``https://github.com/openwrt/openwrt/blob/aba96e2fcb8b0a17fb660ec20a6aa707c48d0792/target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto.c#L756));`` this one only uses `devm_of_mdiobus_register()`.

Conversely `<linux/module.h>` is missing even though the file uses `module_platform_driver()`, `MODULE_DEVICE_TABLE()`, `MODULE_DESCRIPTION()` and `MODULE_LICENSE()`. It builds today only because `<linux/phy.h>` happens to pull it in transitively ([phy.h:23](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/include/linux/phy.h#L23), v6.18.44). The sibling [mdio-realtek-otto-serdes.c](``https://github.com/openwrt/openwrt/blob/aba96e2fcb8b0a17fb660ec20a6aa707c48d0792/target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-serdes.c#L6)`` includes it explicitly.

```suggestion
#include <linux/mfd/syscon.h>
#include <linux/module.h>
#include <linux/of.h>
#include <linux/of_mdio.h>
#include <linux/phy.h>
#include <linux/platform_device.h>
#include <linux/regmap.h>
#include <linux/types.h>
```

---

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
```diff
@@ -0,0 +1,239 @@
...
+
+struct rtmd_ctrl {
+	struct mutex lock;
+	struct regmap *map;
+	struct rtmd_port port[RTMD_MAX_PORTS];
+};
+
+static int rtmd_960x_to_ocp_addr(u32 page, u32 reg, u32 *addr)
+{
+	/* Translate phy reg to an PHY OCP address
```
openwrt-ai: nit: "an PHY" → "a PHY".

```suggestion
	/* Translate phy reg to a PHY OCP address
```

Same paragraph, line 51: "nobody stops a user to use" is not grammatical — "nothing stops a user from using" reads as intended.

---

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
```diff
@@ -0,0 +1,239 @@
...
+	 * including patching. Must always be set before the phys are probed.
+	 */
+	ret = regmap_update_bits(ctrl->map, RTMD_960X_WRAP_GPHY_MISC,
+				 RTMD_960X_PHY_PATCH_DONE, RTMD_960X_PHY_PATCH_DONE);
+	if (ret)
+		return ret;
+
+	bus->name = "Realtek MDIO bus";
+	bus->read = rtmd_read_c22;
+	bus->write = rtmd_write_c22;
```
openwrt-ai, line 216: nit: the PR description still describes the first version of this change (the in-place edit of `mdio-realtek-otto.c`): *"the enable and disable polling functions are now behind the if checks for poll_ctrl from driver data. The same goes for c45 reads and writes."*

None of that is in the PR any more — this driver registers only `read`/`write` (C22), has no `read_c45`/`write_c45`, and has no `poll_ctrl` or driver-data config struct at all. The commit message is accurate and matches the diff; only the PR body is stale, so it is worth updating before this gets picked up.

---

openwrt-ai (commented):
Re-reviewed the delta from `17fbb25` to `aba96e2` (dropping `struct rtmd_chan`, `struct rtmd_config` and `rtmd_bus_to_ctrl()`, folding `rtmd_probe_bus()` into `rtmd_probe()` with `devm_mdiobus_alloc_size(dev, sizeof(*ctrl))`, and adding the OCP-address comment).

Both findings from my last review are addressed: the one-pointer wrapper is gone and `bus->priv` now holds `rtmd_ctrl` directly, and the `regmap_update_bits()` continuation line is aligned to the paren. The devm teardown order is still correct after the restructure — the bus is allocated first, so it is unregistered and the mutex destroyed before the allocation is released. The commit message matches the diff, so no commit checks this round, and CI on `aba96e2` is green.

Only nits left, all inline: two unused `fwnode` includes plus a missing `<linux/module.h>`, a typo in the new comment, and a stale PR description.

Still open from earlier rounds and untouched here: [`CONFIG_RTL960X` does not exist in this tree](https://github.com/openwrt/openwrt/pull/24840#discussion_r3838931013) — now confirmed by the author as depending on a separate RTL960X subtarget PR, which means nothing builds this file yet and the green CI above says nothing about the driver itself. Not flagging it again; noting it so the dependency stays visible.

---

target/linux/realtek/patches-6.18/820-add-rtl960x-mdio-driver.patch
jameywine: > this is no Otto platform

But [it is](https://github.com/jameywine/GPL-for-GP3000/tree/main/bootloader/u-boot-2020.01/arch/otto) (Phoebus is a codename (of many) for RTL9607C).
If you have any other reason apart from that, then sure but until then i'll nack this request.

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
jameywine: > Registers 0-15 is a standard ones and defined in C22

Not quite but close enough. There are certainly instances where it is not the case like register 9 which is for "test modes". There are also registers beyond 23 which are not paged

> For realtek it traditionally means that registers 16-23 is for paged access

Given the subtraction of 16 (which would mean the reg 0-7 and 0-e with *2), they are essentially covering *all* the ocp addresses if you think about it. For example registers 0-7 with page 0xa40 would give you range of 0xa400 to 0xa40e ocp addresses. You can pretty access any of the standard registers and much more with that.

And because of that, maybe nailing to a fixed page is as necessary as it may be? It just there so you don't have to select pages for registers 0-15 and 24-30 every time.

But i am also fine with keeping it as it is. At least there is some precedent for it in upstream kernel [1](https://elixir.bootlin.com/linux/v7.2/source/drivers/net/dsa/realtek/rtl8365mb_main.c#L882) and [2](https://elixir.bootlin.com/linux/v7.2/source/drivers/net/phy/realtek/realtek_main.c#L166)

jameywine:
Something else worth mentioning, do we need c45 stuff? The internal PHYs have EEE functionality and i don't know it is handled without c45 functions, if at all really.

jameywine:
addressed some stuff from the bot and cleaned up description of rtmd_960x_to_ocp_addr() a bit.

target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-rtl960x.c
```diff
@@ -0,0 +1,237 @@
...
+static int rtmd_960x_to_ocp_addr(u32 page, u32 reg, u32 *addr)
+{
+	/* Translate phy reg to a PHY OCP address
+	 *
+	 * The phy reg 31 is reserved for page select
+	 * The phy regs 16~23 are for paged access to OCP addresses
+	 * The phy regs 0~15 and 24~30 are on base 0xa400 ocp address
+	 *
+	 * With phy registers 16~23, user can realistically reach any
+	 * PHY OCP address which can include things on base 0xa400.
```
openwrt-ai: nit: the rewrite fixed "an PHY" on line 44 but the replacement sentence drops an article — "user can realistically reach" → "a user can realistically reach". "which can include things on base 0xa400" is also vaguer than it needs to be given the preceding line already names that base.

```suggestion
	 * With phy registers 16~23, a user can realistically reach any
	 * PHY OCP address, including the ones on base 0xa400.
```

---

openwrt-ai (commented):
Re-reviewed the delta from `aba96e2` to `8e73835`. The only changes to this PR's two files are the include-list fix and the reworded OCP-translation comment; the rest of the range is the rebase onto current `main`.

All three nits from my last review are addressed: `<linux/fwnode.h>`/`<linux/fwnode_mdio.h>` are gone (no `fwnode_*` identifier remains in the file), `<linux/module.h>` is now included explicitly, and the PR description no longer describes the superseded in-place edit of `mdio-realtek-otto.c`. The remaining implicit includes (`bits.h`, `mutex.h`, `cleanup.h`) match what [mdio-realtek-otto-serdes.c](``https://github.com/openwrt/openwrt/blob/8e73835e97bb5dc78e31f2f39c20f7fdaa93e48a/target/linux/realtek/files-6.18/drivers/net/mdio/mdio-realtek-otto-serdes.c#L3-L12)`` relies on, so no change needed there.

The commit message is unchanged from `aba96e2` and still matches the diff, so no commit checks this round. CI on `8e73835` is green across all six realtek subtargets plus the FormalityCheck jobs.

Only one grammar nit left, inline.

Still open from earlier rounds and not something I'm re-flagging, just keeping visible:
- [`CONFIG_RTL960X` does not exist in this tree](https://github.com/openwrt/openwrt/pull/24840#discussion_r3838931013) — confirmed by the author as depending on a separate RTL960X subtarget PR, so nothing builds this file yet and the green CI above says nothing about the driver itself.
- [@plappermaul's request to explain the `page & 0xfff` mask](https://github.com/openwrt/openwrt/pull/24840#discussion_r3844679036) — the reworded comment clarifies *what* regs 16~23 reach but still doesn't say why the page is masked to 12 bits. That thread is an active discussion between you, @plappermaul and @ProMix0, so I'll leave it to you rather than pile on.
- [@plappermaul's naming request](https://github.com/openwrt/openwrt/pull/24840#discussion_r3865845059), which you've nacked — a maintainer call, not mine.

---

jameywine:
fixed a wording and added more stuff to the comment in rtmd_960x_to_ocp_addr(), for better or worse who knows.

plappermaul:

PR #24841, opened 2026-08-22
bcm53xx: abort unsafe NAND fallback

dhrm1k:
When the NAND-aware upgrade path runs into a preparation error, its helpers return to `platform_other_do_upgrade()`. The caller then continues through the legacy whole-image upgrade path.

This can bypass checks such as the kernel partition size check and write an image after its NAND-specific preparation has failed.

Stop immediately on genuine preparation errors. Keep the intentional fallback for non-UBI images, which continues through `default_do_upgrade()` and preserves configuration using `mtd -j`.

### DIFF
--- a/target/linux/bcm53xx/base-files/lib/upgrade/platform.sh
+++ b/target/linux/bcm53xx/base-files/lib/upgrade/platform.sh
@@ -233,7 +233,7 @@ platform_do_upgrade_nand_trx() {
 		-2 $dir/root
 	[ $? -ne 0 ] && {
 		echo "Failed to extract TRX partitions."
-		return
+		exit 1
 	}
 
 	# Firmwares without UBI image should be flashed "normally"
@@ -253,7 +253,7 @@ platform_do_upgrade_nand_trx() {
 	local kernel_length=$(wc -c $dir/kernel | cut -d ' ' -f 1)
 	[ $kernel_length -gt $linux_length ] && {
 		echo "New kernel doesn't fit \"linux\" partition."
-		return
+		exit 1
 	}
 	rm -f /tmp/null.bin
 	rm -f /tmp/kernel.trx
@@ -263,7 +263,7 @@ platform_do_upgrade_nand_trx() {
 		-f /tmp/null.bin
 	[ $? -ne 0 ] && {
 		echo "Failed to create simple TRX with new kernel."
-		return
+		exit 1
 	}
 
 	# Prepare UBI image (drop unwanted extra blocks)
@@ -274,7 +274,7 @@ platform_do_upgrade_nand_trx() {
 	truncate -s $ubi_length $dir/root
 	[ $? -ne 0 ] && {
 		echo "Failed to prepare new UBI image."
-		return
+		exit 1
 	}
 
 	# Flash
@@ -295,7 +295,7 @@ platform_do_upgrade_nand_seamaseal() {
 		-o $dir/seama.entity
 	[ $? -ne 0 ] && {
 		echo "Failed to extract Seama entity."
-		return
+		exit 1
 	}
 	local entity_size=$(wc -c $dir/seama.entity | cut -d ' ' -f 1)
 
@@ -311,7 +311,7 @@ platform_do_upgrade_nand_seamaseal() {
 	done
 	[ $ubi_offset -eq 0 ] && {
 		echo "Failed to find UBI in Seama entity."
-		return
+		exit 1
 	}
 
 	local ubi_length=0
@@ -371,8 +371,8 @@ platform_other_do_upgrade() {
 				;;
 		esac
 
-		# Above calls exit on success.
-		# If we got here something went wrong.
+		# NAND-aware helpers exit after successful UBI upgrades.
+		# Returned images continue through the legacy whole-image path.
 		echo "Writing whole image to NAND flash. All erase counters will be lost."
 	}
 


neheb (approved):

target/linux/bcm53xx/base-files/lib/upgrade/platform.sh
```diff
@@ -373,6 +373,10 @@ platform_other_do_upgrade() {
 
 		# Above calls exit on success.
 		# If we got here something went wrong.
+		[ -n "$UPGRADE_BACKUP" ] && {
+			echo "Refusing raw NAND fallback: configuration cannot be preserved."
+			exit 1
+		}
```
openwrt-ai: This guard cannot fire on the case the commit is about, and only fires on the two fallbacks the commit says it keeps.

After the first five hunks, every genuine preparation error in `platform_do_upgrade_nand_trx()` / `platform_do_upgrade_nand_seamaseal()` calls `exit 1` and never returns to this point. That leaves exactly two ways for control to reach line 376:

1. the deliberate non-UBI fallback — [`echo "Provided firmware doesn't use UBI for rootfs."; return` at platform.sh:241-244](https://github.com/openwrt/openwrt/blob/5d79fed231b8e3937e74b2336cee095a51c0db80/target/linux/bcm53xx/base-files/lib/upgrade/platform.sh#L241-L244);
2. the `NAND aware sysupgrade is unsupported for $BCM53XX_FW_FORMAT format` branch just above.

Since [`SAVE_CONFIG` defaults to 1](https://github.com/openwrt/openwrt/blob/5d79fed231b8e3937e74b2336cee095a51c0db80/package/base-files/files/sbin/sysupgrade#L34) and stage2 exports `UPGRADE_BACKUP` accordingly, a plain `sysupgrade image.bin` on a NAND bcm53xx board now hard-aborts on both of those instead of flashing.

The stated justification also doesn't hold for case 1: [`default_do_upgrade()` at common.sh:311-313](https://github.com/openwrt/openwrt/blob/5d79fed231b8e3937e74b2336cee095a51c0db80/package/base-files/files/lib/upgrade/common.sh#L311-L313) already passes `-j "$UPGRADE_BACKUP"` to `mtd`, which does append the config archive to the jffs2 tail of a squashfs image — precisely the non-UBI image kind this fallback exists for. Config is only unrecoverable when the image *is* UBI-based, and that situation can no longer reach here anyway.

Suggest either dropping this hunk (the `exit 1` conversions on their own already fix the "bypasses the kernel size check / silently discards config" problem), or narrowing it so it triggers only for a UBI rootfs — e.g. have the non-UBI branch signal "deliberate fallback" distinctly from "unsupported format" and let only the latter refuse.

---

target/linux/bcm53xx/base-files/lib/upgrade/platform.sh
```diff
@@ -373,6 +373,10 @@ platform_other_do_upgrade() {
 
 		# Above calls exit on success.
 		# If we got here something went wrong.
+		[ -n "$UPGRADE_BACKUP" ] && {
+			echo "Refusing raw NAND fallback: configuration cannot be preserved."
```
openwrt-ai: nit: if this guard stays, the message leaves the user stuck — it names the refusal but not the way forward. Every other abort in this file is a genuine "your image is broken" condition, whereas this one is recoverable by re-running without config preservation.

```suggestion
			echo "Refusing raw NAND fallback: configuration cannot be preserved. Use \"sysupgrade -n\" to flash without keeping settings."
```

---

target/linux/bcm53xx/base-files/lib/upgrade/platform.sh
```diff
@@ -233,7 +233,7 @@ platform_do_upgrade_nand_trx() {
 		-2 $dir/root
 	[ $? -ne 0 ] && {
 		echo "Failed to extract TRX partitions."
-		return
+		exit 1
```
openwrt-ai, line 236: The five `return` -> `exit 1` conversions in this file look right and match the pre-existing convention two blocks down ([`Unable to find "linux" partition size` -> `exit 1` at platform.sh:248-251](https://github.com/openwrt/openwrt/blob/5d79fed231b8e3937e74b2336cee095a51c0db80/target/linux/bcm53xx/base-files/lib/upgrade/platform.sh#L248-L251)), which was already the odd one out. Same applies to lines 256, 266, 277, 298 and 314. No change requested here — flagging only that these conversions alone already close the "raw fallback bypasses the kernel size check" hole described in the commit message; see my comment on the `UPGRADE_BACKUP` hunk.

---

openwrt-ai (commented):
## Commit checks

- `5d79fed2` "bcm53xx: abort unsafe NAND fallback" — the body says "Keep the intentional raw fallback for non-UBI images, but refuse it when configuration preservation was requested." Those two clauses cancel each other in practice: `SAVE_CONFIG` defaults to 1, so `UPGRADE_BACKUP` is set for an ordinary `sysupgrade image.bin`, and the non-UBI fallback the sentence promises to keep is the main path the new guard kills. The body also states the raw fallback "cannot restore [the saved configuration] into the new rootfs_data volume" — true for a UBI image, but `default_do_upgrade()` does restore it via `mtd -j` for the non-UBI images this fallback is for. Worth either reworking the guard or restating the message to match what the diff actually does. Details inline.


---

target/linux/bcm53xx/base-files/lib/upgrade/platform.sh
dhrm1k: fixed by dropping the `UPGRADE_BACKUP` guard.

dhrm1k:
@neheb gentle ping when you have a moment. I updated the commit after your review and removed the `UPGRADE_BACKUP` guard because it also blocked the intentional non-UBI fallback, where `default_do_upgrade()` can preserve the configuration using `mtd -j`.

The remaining change only turns genuine NAND preparation errors into hard failures. Could you please take another look at the latest revision?

openwrt-ai (commented):

---

target/linux/bcm53xx/base-files/lib/upgrade/platform.sh
openwrt-ai: solved by dropping the guard entirely, thanks

---

PR #24843, opened 2026-08-22
wifi-scripts: mac80211: fix 6 GHz 40 MHz configuration for non-AP modes

astrouski-google:
In `iw_htmode()`, 40 MHz modes (HE40/EHT40) on the 6 GHz band fell through to the 5 GHz HT40+/HT40- calculation. Since 802.11n HT modes do not exist in 6 GHz, `iw` commands like `iw dev <ifname> set freq <freq> HT40+` fail in cfg80211 with `-EINVAL` (`invalid channel definition`). Fix this by adding 6 GHz band handling in `iw_htmode()` to calculate and return the 40 MHz center frequency (`40 <center_freq>`).

### DIFF
--- a/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh
+++ b/package/network/config/wifi-scripts/files-ucode/lib/netifd/wireless/mac80211.sh
@@ -125,6 +125,20 @@ function iw_htmode(config) {
 	case "HE20":
 	case "EHT20":
 		return "HT20";
+	case "HT40":
+	case "VHT40":
+	case "HE40":
+	case "EHT40":
+		switch (config.band) {
+		case "2g":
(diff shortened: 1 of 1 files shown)


astrouski-google:
@PolynomialDivision Hi, can you pls take a look here also? I've hit that issue when creating monitor mode for HE40 and EHT40 on 6G. Because of incorrect iw_htmode return value it tried to use HT40 mode instead which is not appropriate for 6G band. I've validate that after fix monitor interface start working.

robimarko:
This should probably be fixed in the legacy shell scripts as well

robimarko:

astrouski-google:
> This should probably be fixed in the legacy shell scripts as well

@robimarko 
I can create a separate PR then with changes in old one too.

robimarko:
Yes, just make a new PR for that

PR #24844, opened 2026-08-22
mediatek: mt7622: remove bogus dsa-migration from Netgear WAX206

namiltd:
The Netgear WAX206 was added to OpenWrt long after the mt7622 target had already been converted to DSA. Consequently, this device never shipped with a swconfig-based network setup and there is no DSA migration to perform.

Remove the spurious $(Device/dsa-migration) reference. Device/dsa-migration is not defined for the mediatek target, so the reference expands to nothing and no DEVICE_COMPAT_VERSION was ever set for this device; removing it has no effect on sysupgrade compatibility.

### DIFF
--- a/target/linux/mediatek/image/mt7622.mk
+++ b/target/linux/mediatek/image/mt7622.mk
@@ -333,7 +333,6 @@ endef
 TARGET_DEVICES += mediatek_mt7622-rfb1-ubi
 
 define Device/netgear_wax206
-  $(Device/dsa-migration)
   DEVICE_VENDOR := NETGEAR
   DEVICE_MODEL := WAX206
   DEVICE_DTS := mt7622-netgear-wax206


target/linux/mediatek/image/mt7622.mk
```diff
@@ -333,7 +333,6 @@ endef
 TARGET_DEVICES += mediatek_mt7622-rfb1-ubi
 
 define Device/netgear_wax206
-  $(Device/dsa-migration)
```
openwrt-ai, line 336: nit: the commit message hedges ("which is either a no-op or misleadingly suggests..."), but this can be stated definitively: `Device/dsa-migration` is never defined under `target/linux/mediatek/` — the only definitions in-tree are in [ipq806x/image/generic.mk:12](https://github.com/openwrt/openwrt/blob/9981578d3e7b1a2f0a0638a94e0c982994c411de/target/linux/ipq806x/image/generic.mk#L12), [lantiq/image/vr9.mk:3](https://github.com/openwrt/openwrt/blob/9981578d3e7b1a2f0a0638a94e0c982994c411de/target/linux/lantiq/image/vr9.mk#L3), [mvebu/image/cortexa9.mk:38](https://github.com/openwrt/openwrt/blob/9981578d3e7b1a2f0a0638a94e0c982994c411de/target/linux/mvebu/image/cortexa9.mk#L38) and [ramips/image/mt7621.mk:240](https://github.com/openwrt/openwrt/blob/9981578d3e7b1a2f0a0638a94e0c982994c411de/target/linux/ramips/image/mt7621.mk#L240). So `$(Device/dsa-migration)` expands to the empty string here and `DEVICE_COMPAT_VERSION` / `DEVICE_COMPAT_MESSAGE` were never set for this device.

Worth saying so explicitly in the commit message, since it is the thing a reviewer needs to check before merging: because no compat version was ever set, removing the line cannot change sysupgrade compatibility for existing WAX206 installations. Suggested rewording of the last paragraph:

> Remove the spurious $(Device/dsa-migration) reference. Device/dsa-migration is not defined for the mediatek target, so the reference expands to nothing and no DEVICE_COMPAT_VERSION was ever set for this device; removing it has no effect on sysupgrade compatibility.

This is also the last remaining `dsa-migration` reference in the mediatek target, so no sibling devices need the same treatment.

---

openwrt-ai (commented):
Reviewed 1 new commit. The change itself is correct: `Device/dsa-migration` has no definition in the mediatek target, so the removed line was an unexpanded no-op and dropping it does not alter `DEVICE_COMPAT_VERSION` for the WAX206. One commit-message nit inline.

---

robimarko:

namiltd:
@robimarko The patch is ​​also correct with version 25.12, please merge it with the openwrt-25.12 branch as well.

robimarko:
Backported to 25.12

PR #24845, opened 2026-08-22
modemsignal-leds: drive cellular signal LEDs from the modem

carlosz1986:
Please consider this as a draft for now :-) It works on my device ...but still!

OpenWrt has rssileds for wireless RSSI but nothing equivalent for cellular. More than 20 in-tree boards declare cellular signal LEDs and not one of them has anything driving those LEDs.

This is the cellular counterpart of rssileds: a small daemon that polls a modem and maps the sigal onto the board's *:signalN, *:rssiN or *:mobile-N LEDs, with a shell backend for AT (+CSQ, +QCSQ), QMI (uqmi) and MBIM (umbim).

No per-board configuration is needed in the common case. The init script auto-discovers a same-colour set of those LEDs and the first ncm/qmi/mbim interface: several LEDs become a bar graph, a single one a connection indicator. LEDs that already have a job are never taken, so a board running rssileds for wireless keeps its display. A mixed-colour set is ambiguous and is left to an explicit uci config.

Enabled on one board: Cudy LT700 Outdoor
Technically ten boards that have an unclaimed, uniform-colour signal LED group and already ship uqmi or comgt-ncm should work: Teltonika RUT955 (+ the inherited H7V3C0 variant) and RUTC50, TOZED ZLT S12 PRO, TOTOLINK LR1200, TP-Link MR600 v2, Archer MR200 v5/v6 and TL-MR6400 v4/v5/v7. Those are untested by me. For the MR200 I am still puzzled what works what not. Development was done on a Cudy LT700 Outdoor with a Quectel EG060W, which is [now merged](https://github.com/openwrt/openwrt/pull/24689).

I would love to get feedback here.


carlosz1986:
@rbqvq do you know how the signal led on the Hongdian H8850 (RF (Modem Signal): GPIO/2 active-low) is driven? cheers.

carlosz1986:
@dangowrt, hey Daniel ...This package is heavily inspired (basically a simple copy) by the rssileds package. As you are  involved there, would you mind to share some feedback? Best.

rbqvq:
> @rbqvq do you know how the signal led on the Hongdian H8850 (RF (Modem Signal): GPIO/2 active-low) is driven? cheers.

On the original vendor firmware, it appears to be OEM software driven

carlosz1986:
> On the original vendor firmware, it appears to be OEM software driven

Exactly, that's the same for the cudy lt700 - so I created this PR. Would you mind testing this package for the mt7628an_hongdian_h8850-v20.dts and help me? I don't have your deivce on my workbench. I assume the RF is the signal one?
You would need to change something like this (fully untested of course):

```
diff --git a/target/linux/ramips/dts/mt7628an_hongdian_h8850-v20.dts b/target/linux/ramips/dts/mt7628an_hongdian_h8850-v20.dts
index 8983203e2c..af32eb604d 100644
--- a/target/linux/ramips/dts/mt7628an_hongdian_h8850-v20.dts
+++ b/target/linux/ramips/dts/mt7628an_hongdian_h8850-v20.dts
@@ -24,10 +24,10 @@
 	leds {
 		compatible = "gpio-leds";
 
-		led-rf {
+		led-signal {
 			gpios = <&gpio 2 GPIO_ACTIVE_LOW>;
 			color = <LED_COLOR_ID_GREEN>;
-			function = "rf";
+			function = "signal";
 		};
 
 		led_net: led-net {
diff --git a/target/linux/ramips/image/mt76x8.mk b/target/linux/ramips/image/mt76x8.mk
index 173beba69d..715b7e7c67 100644
--- a/target/linux/ramips/image/mt76x8.mk
+++ b/target/linux/ramips/image/mt76x8.mk
@@ -421,7 +421,8 @@ define Device/hongdian_h8850-v20
   DEVICE_VENDOR := Hongdian
   DEVICE_MODEL := H8850
   DEVICE_VARIANT := v20
-  DEVICE_PACKAGES := kmod-usb2 kmod-usb-ohci kmod-usb-net-qmi-wwan kmod-usb-serial-option uqmi
+  DEVICE_PACKAGES := kmod-usb2 kmod-usb-ohci kmod-usb-net-qmi-wwan kmod-usb-serial-option uqmi \
+	modemsignal-leds
 endef
 TARGET_DEVICES += hongdian_h8850-v20
 
diff --git a/target/linux/ramips/mt76x8/base-files/etc/board.d/04_led_migration b/target/linux/ramips/mt76x8/base-files/etc/board.d/04_led_migration
index c845fadd73..da91d825d9 100644
--- a/target/linux/ramips/mt76x8/base-files/etc/board.d/04_led_migration
+++ b/target/linux/ramips/mt76x8/base-files/etc/board.d/04_led_migration
@@ -4,6 +4,9 @@
 board=$(board_name)
 
 case "$board" in
+hongdian,h8850-v20)
+	migrate_leds 'green:rf=green:signal'
+	;;
 keenetic,kn-1613)
 	migrate_leds 'green:internet=green:wan'
 	;;
```

package/network/utils/modemsignal-leds/files/modemsignal-leds.init
```diff
@@ -0,0 +1,200 @@
...
+		esac
+	done
+}
+
+# LEDs that already have a job in /etc/config/system - a netdev, switch,
+# usbport or rssi trigger, or a user assignment. rssileds drives its LEDs
+# this way, so this is what keeps us off a wireless RSSI display on a board
+# that has both a modem and rssileds.
+claimed_leds() {
+	uci -q show system | sed -n "s/^system\.[^.]*\.sysfs='\(.*\)'$/\1/p"
```
openwrt-ai, line 105: `claimed_leds()` only inspects `/etc/config/system`, but the boot/failsafe/running/upgrade LED is never configured there — it comes from the DT `led-boot`/`led-failsafe`/`led-running`/`led-upgrade` aliases and is written straight to sysfs by [`diag.sh`](https://github.com/openwrt/openwrt/blob/a40cbde146d50f3cf7c14b1781a90518eba100a6/package/base-files/files/etc/diag.sh#L6-L9).

That collides on one of the boards this series enables: on the Telco T1 all four aliases point at `blue:rssi0` ([`qca9531_telco_t1.dts:15-18`](https://github.com/openwrt/openwrt/blob/a40cbde146d50f3cf7c14b1781a90518eba100a6/target/linux/ath79/dts/qca9531_telco_t1.dts#L15-L18)), and `list_signal_leds()` matches it, so autodiscovery hands `blue:rssi0` to the bar graph. `led_release_trigger()` then clears its trigger and the daemon keeps overwriting brightness — the running indication is gone and the failsafe/upgrade blink fights the poll loop.

Consider excluding the four DT status LEDs as well, e.g. resolving them with `get_dt_led boot|failsafe|running|upgrade` from [`/lib/functions/leds.sh:44`](https://github.com/openwrt/openwrt/blob/a40cbde146d50f3cf7c14b1781a90518eba100a6/package/base-files/files/lib/functions/leds.sh#L44) and filtering them out here.

---

package/network/utils/modemsignal-leds/files/modemsignal-leds.init
```diff
@@ -0,0 +1,200 @@
...
+# ... -> "mobile" / "signal" / "rssi"
+led_base()  { led_group "$1" | sed 's/.*://'; }
+
+# zero-config: drive an unclaimed, same-colour set of cellular signal LEDs.
+# A lone LED becomes a connection indicator (status); a set becomes a bar
+# graph. signal<N>/rssi<N> are unambiguous levels; mobile-<N> is not - the
+# MOFI5500 uses mobile-1/-2 for two modem modules - so it is only treated as
+# a bar graph when there is exactly one modem.
+autodiscover() {
+	local leds n iface ifaces nif mode=bars oIFS="$IFS" name groups base i=0
```
openwrt-ai: `level` is assigned on line 143 but missing from this `local` list, so it leaks into the calling shell (rc.common's) environment.

```suggestion
	local leds n iface ifaces nif mode=bars oIFS="$IFS" name groups base level i=0
```

---

package/network/utils/modemsignal-leds/src/modemsignal-leds.c
```diff
@@ -0,0 +1,278 @@
...
+	close(fds[1]);
+	f = fdopen(fds[0], "r");
+	if (f) {
+		if (fscanf(f, "%d", &q) != 1)
+			q = -1;
+		fclose(f);
+	} else {
+		close(fds[0]);
+	}
+	waitpid(pid, &status, 0);
```
openwrt-ai: The backend child has no timeout. `b_at()` drives `gcom` on the modem's control tty, which can block indefinitely when the modem stops answering (and `flock -n` only protects against *this* daemon's other instances, not against a wedged port), so `fscanf()` and then `waitpid()` block with it and the LEDs freeze at their last value.

It also makes shutdown unreliable: the handlers installed with `signal()` on lines 251-252 get `SA_RESTART` semantics on Linux/glibc/musl, so SIGTERM will not break out of the blocked `fscanf()` — `/etc/init.d/modemsignal-leds stop` hangs until procd escalates to SIGKILL.

Consider bounding the child, e.g. `alarm(interval)` with a no-op `SIGALRM` handler installed via `sigaction()` without `SA_RESTART`, and `kill(pid, SIGKILL)` before `waitpid()` on expiry.

---

package/network/utils/modemsignal-leds/files/backend
```diff
@@ -0,0 +1,136 @@
...
+	mbim)		b_mbim "$dev" ;;
+	ncm|3g|wwan)	b_at   "$dev" ;;
+	# Anything else has no control channel we can query - on a plain
+	# cdc-ether modem on proto dhcp, "device" is a netdev, not a tty.
+	*)		echo -1 ;;
+	esac
+}
+
+case "$1" in
+at)	b_at   "$2" ;;
```
openwrt-ai: The netifd-collision guard documented on lines 106-108 only applies to `auto`: `b_auto()` calls `iface_up()` first, but this direct `at` entry point does not. A user following the documented explicit config (`option backend 'at'` + `option device '/dev/ttyUSB2'`, [`modemsignal-leds.config:21-22`](``https://github.com/openwrt/openwrt/blob/a40cbde146d50f3cf7c14b1781a90518eba100a6/package/network/utils/modemsignal-leds/files/modemsignal-leds.config#L21-L22))`` therefore gets exactly the interleaving with netifd's AT traffic that the comment warns about — and that config is the one recommended for multi-modem boards, where the AT port is busiest.

Could the gate be made unconditional, e.g. by having the init script always pass the interface name along so `b_at()` can call `iface_up()` too?

---

target/linux/ramips/image/mt76x8.mk
```diff
@@ -845,7 +845,7 @@ define Device/totolink_lr1200
   IMAGE_SIZE := 7872k
   DEVICE_VENDOR := TOTOLINK
   DEVICE_MODEL := LR1200
-  DEVICE_PACKAGES := kmod-mt76x2 kmod-usb2 uqmi
+  DEVICE_PACKAGES := kmod-mt76x2 kmod-usb2 uqmi modemsignal-leds
```
openwrt-ai, line 848: nit: The PR description says "Enabled on the three Hongdian boards that have both uniform-colour mobile LEDs and uqmi already in DEVICE_PACKAGES", but the eleven board commits in this series enable it on Telco T1, Teltonika RUT955 (+ the inherited H7V3C0 variant), Teltonika RUTC50, TOZED ZLT S12 PRO, TOTOLINK LR1200 and six TP-Link models — no Hongdian board is touched. Worth refreshing the description so reviewers know which boards to look at.

---

package/network/utils/modemsignal-leds/Makefile
```diff
@@ -0,0 +1,58 @@
...
+	$(INSTALL_DIR) $(1)/usr/sbin
+	$(INSTALL_BIN) $(PKG_BUILD_DIR)/modemsignal-leds $(1)/usr/sbin/modemsignal-leds
+	$(INSTALL_DIR) $(1)/etc/init.d
+	$(INSTALL_BIN) ./files/modemsignal-leds.init $(1)/etc/init.d/modemsignal-leds
+	$(INSTALL_DIR) $(1)/etc/config
+	$(INSTALL_CONF) ./files/modemsignal-leds.config $(1)/etc/config/modemsignal-leds
+	$(INSTALL_DIR) $(1)/etc/uci-defaults
+	$(INSTALL_BIN) ./files/modemsignal-leds.defaults $(1)/etc/uci-defaults/91-modemsignal-leds
+	$(INSTALL_DIR) $(1)/usr/share/modemsignal-leds
+	$(INSTALL_BIN) ./files/backend $(1)/usr/share/modemsignal-leds/backend
```
openwrt-ai: nit: `backend` is an executable being installed into `/usr/share` with `$(INSTALL_BIN)`. rssileds — the package this one is modelled on — puts its equivalent shell helper in [`/usr/libexec/led-trigger/`](https://github.com/openwrt/openwrt/blob/a40cbde146d50f3cf7c14b1781a90518eba100a6/package/network/utils/rssileds/Makefile#L43-L44) and keeps `/usr/share` for data. `/usr/libexec/modemsignal-leds/backend` would match; the `.gcom` scripts on lines 52-53 are genuine data and can stay where they are, but note `backend` currently locates them via `dirname "$0"` (files/backend line 15), so that would need an explicit path.

---

openwrt-ai (commented):
Reviewed all 12 commits (fresh review, head `a40cbde`). Commit messages match their diffs; findings are inline. CI on this head is green/queued, so nothing CI-grounded.

---

rbqvq:
> > On the original vendor firmware, it appears to be OEM software driven
> 
> Exactly, that's the same for the cudy lt700 - so I created this PR. Would you mind testing this package for the mt7628an_hongdian_h8850-v20.dts and help me? I don't have your deivce on my workbench. I assume the RF is the signal one?

Yes, it's RF leds. And same with H8922 device.

NOTE: On my H8922 device, only first modem is installed (it seems PCB only reserve one aerial for modem, so I think only one modem for this HW revision, so ignore the second RF led)

rbqvq:
Any plan for ModemManager?

carlosz1986:
> Any plan for ModemManager?

I guess that's rather easy to add here... But I would say I keep it for later.
The PR has already a lot of functionality and my worry is that this is never accepted if we add more...
So I'll come back to you.
best.

package/network/utils/modemsignal-leds/src/modemsignal-leds.c
```diff
@@ -0,0 +1,303 @@
...
+		 * killed with it rather than left holding the modem's tty */
+		setpgid(0, 0);
+		dup2(fds[1], STDOUT_FILENO);
+		close(fds[0]);
+		close(fds[1]);
+		execvp(argv[0], argv);
+		_exit(127);
+	}
+
+	close(fds[1]);
```
openwrt-ai, line 178: `kill(-pid, SIGKILL)` on line 190 signals process group `pid`, but that group only comes into existence when the child runs `setpgid(0, 0)` on line 169. If the parent gets there first, no process group with that id exists yet, `kill()` fails with `ESRCH`, nothing is signalled, and `waitpid()` on line 191 then blocks unbounded on a backend that may be stuck on the modem tty — precisely the hang this commit set out to remove.

The window is narrow but it is reachable in exactly the case that matters: SIGTERM no longer restarts `fscanf()` (that is the point of the `sigaction()` change), so a SIGTERM arriving right after `fork()` returns the read immediately, `alarm(0)` cancels the timeout, and the kill races the child's `setpgid()`.

The usual fix is to set the group from both sides — whichever call runs first wins and the other is a harmless no-op (`EACCES` once the child has exec'd):

```suggestion
	setpgid(pid, pid);	/* races the child's own setpgid(); either may win */
	close(fds[1]);
```

---

package/network/utils/modemsignal-leds/files/modemsignal-leds.init
```diff
@@ -0,0 +1,213 @@
...
+
+	case "$backend" in
+		auto)	target="$interface" ;;
+		*)	target="$device" ;;
+	esac
+
+	procd_open_instance "$s"
+	procd_set_param command "$BIN" -i "$refresh" -m "$mode"
+	append_leds "$mode" "$n"
+	procd_append_param command "$BACKEND" "$backend" "$target" "$interface"
```
openwrt-ai, line 83: `interface` falls back to the *section name* ([`config_get interface "$s" interface "$s"` at modemsignal-leds.init:57](``https://github.com/openwrt/openwrt/blob/1583f6d32c5eb9e6bf3ae7734bec72fd4571eeb7/package/network/utils/modemsignal-leds/files/modemsignal-leds.init#L57)),`` and `b_at()` now drops every sample when the gate fails ([`files/backend:66`](https://github.com/openwrt/openwrt/blob/1583f6d32c5eb9e6bf3ae7734bec72fd4571eeb7/package/network/utils/modemsignal-leds/files/backend#L66)). So an explicit AT config whose section name is not also a live `network.<name>` never samples at all: an anonymous `config modem`, or the `modem0`/`modem1` shape shipped in [`files/profiles/example:14-23`](``https://github.com/openwrt/openwrt/blob/1583f6d32c5eb9e6bf3ae7734bec72fd4571eeb7/package/network/utils/modemsignal-leds/files/profiles/example#L14-L23``) with `option interface` omitted. `iface_up` is then false forever, the backend returns -1 on every poll and the LEDs stay dark with nothing in the log explaining why — a regression against the previous head for that config.

Passing the name on only when it really is an interface keeps the netifd gate where it helps and restores the old ungated behaviour otherwise (`target` is computed above, so the `auto` path is unaffected):

```suggestion
	# gate on the interface only when it really is one; otherwise an explicit
	# at/device config would never get past iface_up() in the backend
	[ "$(uci -q get "network.$interface")" = interface ] || interface=""
	procd_append_param command "$BACKEND" "$backend" "$target" "$interface"
```

---

package/network/utils/modemsignal-leds/files/backend
```diff
@@ -0,0 +1,140 @@
...
+	# Anything else has no control channel we can query - on a plain
+	# cdc-ether modem on proto dhcp, "device" is a netdev, not a tty.
+	*)		echo -1 ;;
+	esac
+}
+
+case "$1" in
+at)	b_at   "$2" "$3" ;;
+qmi)	b_qmi  "$2" ;;
+mbim)	b_mbim "$2" ;;
```
openwrt-ai: The netifd gate now reaches `b_at` through `$3`, but the explicit `qmi` and `mbim` entry points still receive no interface, so they poll the cdc-wdm control node regardless of what netifd is doing to it.

That leaves the two paths disagreeing with each other inside this same file: `b_auto()` gates *every* protocol — [`iface_up "$iface" || { echo -1; return; }` at backend:120](https://github.com/openwrt/openwrt/blob/1583f6d32c5eb9e6bf3ae7734bec72fd4571eeb7/package/network/utils/modemsignal-leds/files/backend#L120) runs before it dispatches to `b_qmi`/`b_mbim` — while the explicit path gates only AT.

Is the gate deliberately AT-only (because uqmi/umbim allocate their own client ID and tolerate the overlap), or should `$3` be plumbed into `b_qmi`/`b_mbim` too? If it is deliberate, a note next to `iface_up()` saying so would stop the next reader from "fixing" it.

---

package/network/utils/modemsignal-leds/src/modemsignal-leds.c
```diff
@@ -0,0 +1,303 @@
...
+	if (f) {
+		if (fscanf(f, "%d", &q) != 1)
+			q = -1;
+		fclose(f);
+	} else {
+		close(fds[0]);
+	}
+	alarm(0);
+
+	/* the read returned early or was interrupted: do not wait on the child */
```
openwrt-ai: nit: this comment says the opposite of what the code does — the very next statement is a blocking `waitpid()`, and the `kill()` is unconditional, so it also runs on the normal path where the backend already printed its number and exited.

```suggestion
	/* the backend answered or the alarm expired: kill the whole group so a
	 * stuck gcom cannot keep holding the modem's tty, then reap the child */
```

---

openwrt-ai (commented):
Re-reviewed at head `1583f6d`. The branch was rebased/amended since `a40cbde`, so all 11 commits carry new SHAs; the actual change is confined to five files (`Makefile`, `files/backend`, `files/modemsignal-leds.init`, `src/modemsignal-leds.c`, `ath79/image/generic.mk`).

Three of the four earlier findings are addressed: `backend` moved to `/usr/libexec`, `level` added to the `local` list in `autodiscover()`, and the AT poll is now gated on `iface_up`. The boot/failsafe/running/upgrade collision is handled twice over — the new `status_leds()` filter plus dropping the `ath79: telco-t1` commit from the series. Follow-up findings on the new `run_backend()` timeout and on the interface plumbing are inline.

All commit messages match their diffs; CI on this head is green.

---

package/network/utils/modemsignal-leds/files/modemsignal-leds.init
carlosz1986: fixed... -> status_leds() now excludes the DT led-boot/failsafe/running/upgrade aliases

FredApps:
Tested on hardware — ASUS 4G-AX56 (ramips/mt7621, Fibocom FG621-EA). **It works with zero configuration.**

Built your branch's package onto main alongside my pending board patches, and flashed with `sysupgrade -n` so autodiscovery ran against a wiped config rather than anything I had set up by hand.

```
modemsignal-leds: autodiscover: 3 rssi LED(s) mode=bars, iface 'wwan'

/usr/sbin/modemsignal-leds -i 10 -m bars \
  -l white:rssi-1,0,33 -l white:rssi-2,33,66 -l white:rssi-3,66,100 \
  /usr/libexec/modemsignal-leds/backend auto wwan
```

No uci config was created or needed. It found the uniform `white:rssi-1..3` set and the `wwan` ncm interface on its own.

Correlation check at `+CSQ: 20,99` — 20/31 is 64.5%, which lands in the middle band:

| LED | band | brightness |
|---|---|---|
| `white:rssi-1` | 0–33 | 1 |
| `white:rssi-2` | 33–66 | 1 |
| `white:rssi-3` | 66–100 | 0 |

Two of three bars lit. Correct. (Worth noting for anyone reading the panel: on this chassis the bars fill right-to-left, `rssi-1` being the rightmost lens — that is a silkscreen matter, not something the daemon needs to know.)

This board is **not** in your enabled list, and it could be. Its two blockers are removed by openwrt/openwrt#24974 and openwrt/openwrt#24975, which give the 4G-AX56 a `proto ncm` interface named `wwan` and pull in `comgt-ncm`; the `white:rssi-1..3` group already matches your `:(signal|rssi|mobile)(-?[0-9]+)?$` discovery unmodified. For this test I just added `modemsignal-leds` to the device's `DEVICE_PACKAGES` by hand.

Once those land I am happy to send the one-line addition to `mt7621.mk` myself, or you can fold `asus_4g-ax56` into this PR and I will re-test whatever you push.

Context worth having: the 4G-AX56's three signal LEDs are currently listed as an "Unsupported Function" on its wiki page, because nothing in tree drives cellular signal LEDs. Your daemon is the first thing that does, and it needed no board-specific code to do it.

package/network/utils/modemsignal-leds/files/modemsignal-leds.init
```diff
@@ -0,0 +1,217 @@
...
+	esac
+
+	procd_open_instance "$s"
+	procd_set_param command "$BIN" -i "$refresh" -m "$mode"
+	append_leds "$mode" "$n"
+	# gate on the interface only when it really is one: "interface" falls back
+	# to the section name above, and an explicit at/qmi config named e.g.
+	# modem0 would otherwise never get past iface_up() in the backend
+	[ "$(uci -q get "network.$interface")" = interface ] || interface=""
+	procd_append_param command "$BACKEND" "$backend" "$target" "$interface"
```
openwrt-ai, line 83: The new guard reasons about the section-name fallback for `at`/`qmi`, but `auto` still consumes it unchecked on line 67: `config modem 'modem0'` with `option backend 'auto'` and no `option interface` yields `target=modem0`, [`iface_up "$iface"` at backend:122](https://github.com/openwrt/openwrt/blob/86f0ab9f45b9a8b4397ef7a06fdf726ab6ba77ca/package/network/utils/modemsignal-leds/files/backend#L122) is false forever and the LEDs stay dark with nothing logged. Resolving the interface before the `case` and rejecting an empty `target` covers that and the `backend at` / missing `device` case too.

```suggestion
	# "interface" falls back to the section name above, so only use it when it
	# really names a network interface: gate the backend on it then, and drop
	# it otherwise so an explicit at/qmi config named e.g. modem0 still gets
	# past iface_up() in the backend
	[ "$(uci -q get "network.$interface")" = interface ] || interface=""

	case "$backend" in
		auto)	target="$interface" ;;
		*)	target="$device" ;;
	esac
	[ -n "$target" ] || { logger -t modemsignal-leds "modem '$s': no usable interface/device, skipping"; return 0; }

	procd_open_instance "$s"
	procd_set_param command "$BIN" -i "$refresh" -m "$mode"
	append_leds "$mode" "$n"
	procd_append_param command "$BACKEND" "$backend" "$target" "$interface"
```

---

package/network/utils/modemsignal-leds/files/backend
```diff
@@ -0,0 +1,142 @@
+#!/bin/sh
+# SPDX-License-Identifier: GPL-2.0-or-later
+# modemsignal-leds backend: print a normalized signal quality 0..100
+# (or a negative value = no service / error) for one modem.
+#
+#   backend at   <ttydev>            # AT+CSQ via gcom
+#   backend qmi  <cdc-wdm>           # uqmi --get-signal-info
+#   backend mbim <cdc-wdm>           # umbim signal
```
openwrt-ai: nit: all three entry points now take an optional third `<iface>` argument that this synopsis does not mention.

```suggestion
#   backend at   <ttydev> [<iface>]   # AT+CSQ via gcom
#   backend qmi  <cdc-wdm> [<iface>]  # uqmi --get-signal-info
#   backend mbim <cdc-wdm> [<iface>]  # umbim signal
```

Same staleness in the `iface_up()` comment on lines 53-55, which still explains the gate purely in terms of the AT control port although `b_qmi`/`b_mbim` now use it too.

---

package/network/utils/modemsignal-leds/files/modemsignal-leds.config
```diff
@@ -0,0 +1,30 @@
...
+# running rssileds for wireless keeps its display.
+#
+# Nothing is started when no modem interface exists. A mixed-colour set is
+# ambiguous, and so is <color>:mobile-<N> on a multi-modem device, where the
+# index is the module and not a level; both need an explicit config below.
+#
+# Configure explicitly to override, or for multi-modem devices:
+#
+# config modem 'wwan'
+#     option interface 'wwan'       # netif to derive backend + device from
```
openwrt-ai: nit: `interface` is no longer only an `auto` input — with `backend` set to `at`/`qmi`/`mbim` it is what enables the netifd gate, and leaving it out silently means polling right through interface bring-up.

```suggestion
#     option interface 'wwan'       # netif: derives backend + device for
#                                   # 'auto', and pauses polling while netifd
#                                   # brings it up/down for all backends
```

---

openwrt-ai (commented):
Reviewed 11 new commits; commit messages match their diffs.

---

package/network/utils/modemsignal-leds/files/backend
openwrt-ai: solved by having the init script pass the interface through to `b_at` whenever it names a real netif, thanks

---

package/network/utils/modemsignal-leds/Makefile
openwrt-ai: fixed, thanks

---

package/network/utils/modemsignal-leds/files/modemsignal-leds.init
openwrt-ai: fixed, thanks

---

package/network/utils/modemsignal-leds/files/backend
openwrt-ai: solved by plumbing `$3` into `b_qmi`/`b_mbim` as well, thanks

---

carlosz1986:
@FredApps thx for testing, I will put your tag.... I can add a commit later for your device as well :-) 

I still hope for some feedback from others :-) 

pinging again some folks that worked heavily on that area: @hauke @blocktrron @Leo-PL @feckert @jonasjelonek @blogic @nbd168  ... 

I mean the main question that is still open: Does a standalone polling daemon belong here at all, or should this hook into netifd/ubus instead? Based what I see in the rssileds package I would say yes...

cheers.

FredApps:
On the daemon-vs-netifd/ubus question, a data point from `proto ncm` on this board (ASUS 4G-AX56, Fibocom FG621-EA):

netifd publishes no signal data whatsoever for this protocol. The complete key set of `network.interface.wwan` status, which is `proto ncm`, is:

```
autostart, available, data, delegation, dns_metric, dynamic, inactive,
l3_device, manufacturer, metric, neighbors, pending, proto, route, up,
updated, uptime
```

Nothing for rssi/rsrp/rsrq/csq/band, at any nesting level.

So for ncm there is nothing to hook into today. The signal still has to be polled out of the modem over AT, and routing it through netifd/ubus would first mean teaching each proto to publish it — a much larger change that does not remove the polling, only moves it. Standalone looks right to me, at least until some proto does publish signal quality; the daemon could then prefer ubus and fall back to AT polling where it does not.

ModemManager is the obvious exception, since it already exposes signal quality on D-Bus — which is probably what makes the question feel open.

On the board commit: glad to take it, though it depends on #24975, which adds this board's LED nodes. Once that lands the `mt7621.mk` side is a one-liner.

One thing this board could contribute here rather than in my own PR: it has a colour-coded RAT trio (`red`/`yellow`/`blue:mobile` for 2G/3G/4G) which I deliberately left out of #24975. A RAT-mode extension belongs in something like this, where it would serve every cellular board instead of one. Happy to work it up as a follow-up if you want it.

openwrt-ai (commented):
Reviewed 11 new commits; no new issues found.

---

package/network/utils/modemsignal-leds/files/modemsignal-leds.init
openwrt-ai: fixed, thanks

---

package/network/utils/modemsignal-leds/files/backend
openwrt-ai: fixed, thanks

---

package/network/utils/modemsignal-leds/files/modemsignal-leds.config
openwrt-ai: fixed, thanks

---

carlosz1986:
> A RAT-mode extension belongs in something like this, where it would serve every cellular board instead of one. Happy to work it up as a follow-up if you want it.

Yeah :-) We can do this later.
Let's first get some approval on the basic design of this PR :-)

feckert:
Can we make this optional? And do not add them to the DEVICE_PACKAGES by default?

carlosz1986:
> Can we make this optional? And do not add them to the DEVICE_PACKAGES by default?

Sure...But let me shortly clarify: You mean that everyone has to install the package manually or to remove the commits for the devices that are still untested?
For my lt700 it works pretty nice.  I mean if we keep it activated by default every user has directly a full working experience.
What is the reason to not activate it?

cheers,
Carlo

FredApps:
On the default question: `rssileds`, which this package is modelled on, is already in `DEVICE_PACKAGES` itself â€” 20 occurrences in `ath79/image/generic.mk` and 5 more across `ramips/image/mt7620.mk` and `mt7621.mk`. So per-device default-on has direct precedent for this kind of LED helper.

That does not dispose of @feckert's point. `DEVICE_PACKAGES` can only be dropped by building your own image, and unlike `rssileds` this ships a daemon that polls the modem's AT port rather than a passive LED trigger â€” which is a fair thing to want off by default. But opt-in only would make this stricter than the package it supersedes, so it seems worth deciding deliberately rather than by default.

One data point for the trade-off: autodiscovery is why this worked on my board with no configuration at all, and that only reaches users who actually get the package installed.

On your first question — manual install for everyone, versus dropping it only from the devices that are still untested — the same precedent points at the second. `rssileds` is not enabled across the board either; it is added device by device, in each case where someone had the hardware and knew the LEDs were there. Keeping `modemsignal-leds` on the boards where someone has actually run it, and leaving it off the rest until they are, matches how the existing package is handled and is a narrower change than opt-in everywhere.

If that is the split, the ASUS 4G-AX56 belongs on the tested side — the run above was autodiscovery against a wiped config after `sysupgrade -n`, not a hand-tuned setup.

FredApps:
> the main question that is still open: Does a standalone polling daemon belong here at all, or should this hook into netifd/ubus instead?

I went through the tree rather than argue it from taste. There is nothing to hook into.

Every cellular proto handler in core, checked for *any* mention of signal quality, not just for publishing it:

| proto handler | `rssi\|rsrp\|rsrq\|csq\|signal\|sinr` mentions | published via `proto_add_data` |
| --- | --- | --- |
| `ncm.sh` | 0 | 0 |
| `directip.sh` | 0 | 0 |
| `3g.sh` | 0 | 0 |
| `qmi.sh` | 0 | 0 |
| `mbim.sh` | 0 | 0 |

For context on what they *do* publish: `ncm.sh` adds `manufacturer` and `model`, `qmi.sh` adds `cid_4`, `pdh_4`, `cid_6` and a timezone. Signal quality appears nowhere, at any nesting level, in any of them.

ModemManager is the case worth looking at closely, since it is the one stack that genuinely has the data. Its proto script does not publish it either — no `proto_add_data` in it at all — and the way it exposes periodic updates is `mmcli --modem=... --signal-setup=<seconds>`, driven by a `signalrate` UCI option. That is a polling interval. Even the D-Bus stack polls; it just polls somewhere else.

So "hook into netifd/ubus instead" is not an alternative that exists today. Taking that route means first teaching five proto handlers to poll the modem and publish the result, which does not remove the polling — it multiplies it into five implementations instead of one, and still leaves any board not running one of those protos unserved.

Standalone looks right to me on that evidence, and the forward-compatible shape is the one you already have: if a proto ever does start publishing signal quality, the backend can prefer ubus and fall back to AT polling, with nothing changing on the LED side.

FredApps:
Built this branch's package against the matching `ramips/mt7621` 25.12-SNAPSHOT SDK and ran the real thing on the other vendor's modem — ASUS 4G-AX56, Fibocom FG621-EA.

**Autodiscovery, no configuration, after a plain `apk add`:**

```
modemsignal-leds: autodiscover: 3 rssi LED(s) mode=bars, iface 'wwan'
/usr/sbin/modemsignal-leds -i 10 -m bars \
  -l white:rssi-1,0,33 -l white:rssi-2,33,66 -l white:rssi-3,66,100 ...
```

**Eight ifdown/ifup cycles with the daemon running at the default `-i 10`:** all eight reconnected and got a lease, and the modem was at `+CFUN: 4` after every single teardown — so `proto ncm` reached its disconnect command every time. No `Unsupported modem`, nothing in the log.

**The `iface_up()` gate and the `flock` are load-bearing, and I can put a number on what they are worth.** I first measured with a deliberately naive stand-in — a bare `gcom` poll loop with neither guard — to see what happens without them. Same hardware, same 10 s interval:

| | connects OK | teardown reached disconnect |
| --- | --- | --- |
| no poller at all | 5/5 | 5/5 (`+CFUN: 4`) |
| unguarded poll loop | 10/13 | at least one left `+CFUN: 1` |
| **this package, `-i 10`** | **8/8** | **8/8** |

The unguarded failures look exactly like the one you fixed in 5ae89060f4, reached by a different route — an empty manufacturer from a corrupted read, then:

```
netifd: wwan (27331): WARNING: Variable '' does not exist or is not an array/object
netifd: wwan (27331): Unsupported modem
```

teardown returning before the disconnect is sent, leaving the data call up on the modem.

One correction to my own numbers, in your favour: that stand-in polled with comgt's `getstrength.gcom`, so part of the damage was self-inflicted. As the comment in your `backend` says, that script opens with `flash 0.1`, a DTR pulse that hangs up the data call — and diffing the two files on the device, that line is the *only* difference from your `getcsq.gcom`. So I reproduced a hazard you had already documented rather than finding a new one. The separate point stands on its own: polling a control node netifd owns, without gating on interface state, is destructive.

For @feckert's question this cuts the opposite way to what I expected when I started looking: the shared-port risk is real, it is already handled here, and on this evidence it is not a reason to keep the package out of `DEVICE_PACKAGES`.

carlosz1986:
thx @FredApps for detailed testing :-)

carlosz1986:
I removed all devices commits that are not fully tested yet. The LT700 works proper, so we get this PR merged faster...
@FredApps you can add your device on top once your pr lands please ...
 
I kindly ping @hauke and @jonasjelonek for feedback or approval to get this merged :-)
cheers...

target/linux/ramips/image/mt7621.mk
```diff
@@ -792,7 +792,7 @@ define Device/cudy_lt700-outdoor-v1
   UIMAGE_NAME := R43
   DEVICE_PACKAGES := kmod-mt7603 kmod-mt7615e kmod-mt7663-firmware-ap \
 	kmod-usb3 kmod-usb-net-cdc-ncm kmod-usb-serial-option comgt-ncm \
-	-uboot-envtools
+	modemsignal-leds -uboot-envtools
```
openwrt-ai, line 795: nit: the PR body still says the series enables this on ten boards (RUT955 + H7V3C0, RUTC50, ZLT S12 PRO, LR1200, MR600 v2, Archer MR200 v5/v6, TL-MR6400 v4/v5/v7) and that the Cudy LT700 Outdoor is "not yet in tree" — the series now touches only this device, and it is in tree ([`mt7621.mk:786`](https://github.com/openwrt/openwrt/blob/e4a645339956fbac7e77509b7d7ef80b68ef9336/target/linux/ramips/image/mt7621.mk#L786)). Worth refreshing so reviewers arriving now see the actual scope.

---

PR #24847, opened 2026-08-22
kernel: workaround premature ready bit on SkyHigh SPI-NAND chips

VitaliySochniy:
Nokia XG-040G-MD units built with SkyHigh SPI-NAND occasionally log UBI "fixable bit-flip" events in production, here after about a month of uptime:

```
[2503209.483768] ubi0: schedule PEB 399 for scrubbing
[2503209.542280] ubi0: scrubbed PEB 399 (LEB 5:272), data moved to PEB 366
[2562754.294549] ubi0: fixable bit-flip detected at PEB 1316
[2562754.360134] ubi0: scrubbed PEB 1316 (LEB 5:195), data moved to PEB 399
[2589193.123216] ubi0: fixable bit-flip detected at PEB 511
[2589193.128571] ubi0: schedule PEB 511 for scrubbing
[2589193.197174] ubi0: scrubbed PEB 511 (LEB 5:116), data moved to PEB 364
[2689227.173721] ubi0: fixable bit-flip detected at PEB 1194
[2689227.179237] ubi0: schedule PEB 1194 for scrubbing
[2689227.221495] ubi0: fixable bit-flip detected at PEB 1194
[2689227.295181] ubi0: scrubbed PEB 1194 (LEB 5:76),
```

`spinand_wait()` never re-verifies that the chip is still ready once it has first observed a non-busy sample: it polls `REG_STATUS` via `spi_mem_poll_status()` until `STATUS_READY` clears and trusts that single result. The extra `spinand_read_status()` further down is only a fallback for controllers whose hardware `->poll_status()` offload didn't leave a non-busy sample in `spinand->scratchbuf`, not a second confirmation. On SkyHigh chips the ready bit has been observed to flip back to busy shortly after first going ready, so the cache can be read before page data has settled, or the next command issued before a program/erase has actually completed.

This adds a `SPINAND_PREMATURE_READY_QUIRK` flag, set in the chip table for the SkyHigh S35ML01G301, S35ML01G300, S35ML02G300 and S35ML04G300 entries. The reset, page load (regular and continuous read), program and erase waits go through a new `spinand_wait_ready_confirmed()` helper: for flagged parts it only trusts a ready result from `spinand_wait()` once a follow-up status read still shows the device ready, and calls `spinand_wait()` again otherwise, so the normal delay/poll and any hardware `->poll_status()` offload are kept. Each `spinand_wait()` gets its own `SPINAND_WAITRDY_TIMEOUT_MS` budget and no new one is started once that span has elapsed, so a flagged part can spend up to roughly twice that time waiting. The reset issued from `spinand_detect()` runs before `spinand->flags` is assigned and so keeps the unflagged behavior; the resume path is covered. Parts without the flag get plain `spinand_wait()`, unchanged.

The patch lives in `target/linux/generic/pending-6.18/` since it changes generic `drivers/mtd/nand/spi/core.c`, not airoha-specific code. The other changed files are hunk offset refreshes of target patches that touch the same `core.c` (airoha 901, econet 902, mediatek 010/121/330/435/436/960, qualcommax 0412).

**Not yet confirmed to eliminate the bit-flips over a long soak on hardware** — this is a workaround for the observed behavior, not a proven fix.

Testing: `make target/linux/refresh` is clean on airoha/an7581, mediatek/filogic, qualcommax/ipq50xx and econet/en751221; `checkpatch.pl --strict` reports no issues; the airoha kernel builds with `CONFIG_KERNEL_WERROR=y` and all kmods without a single diagnostic. This revision was flashed with sysupgrade to a Nokia XG-040G-MD with a SkyHigh chip: the upgrade (real erase and program cycles through the new path) completed and the board booted cleanly, with no UBI or SPI-NAND errors. The reset on the resume path isn't exercised by a plain boot, so that one call site is compile-tested only.

### DIFF
--- a/target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
+++ b/target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
@@ -1,0 +1,168 @@
+From f5e0d7642e7239aa5e014f821756405cf29b8293 Mon Sep 17 00:00:00 2001
+From: Vitaliy Sochnev <sochnev.v.74@gmail.com>
+Date: Tue, 18 Aug 2026 23:04:36 +0100
+Subject: [PATCH] mtd: spinand: workaround premature ready bit on SkyHigh chips
+
+Nokia XG-040G-MD units built with SkyHigh SPI-NAND occasionally log UBI
+"fixable bit-flip" events in production, weeks apart:
+
+  ubi0: fixable bit-flip detected at PEB 1316
+  ubi0: scrubbed PEB 1316 (LEB 5:195), data moved to PEB 399
+  ubi0: fixable bit-flip detected at PEB 511
+  ubi0: schedule PEB 511 for scrubbing
+
+spinand_wait() never re-verifies that the chip is still ready once it
+has first observed a non-busy sample: it polls REG_STATUS via
+spi_mem_poll_status() until STATUS_READY clears (or times out) and
+trusts that single result. The extra spinand_read_status() call further
+down only runs as a fallback for controllers whose hardware
+->poll_status() offload didn't already leave a non-busy sample in
+spinand->scratchbuf - it is not a second confirmation of readiness.
+
+On SkyHigh chips the ready bit has been observed to flip back to busy
+shortly after first going ready, so a caller can act on stale/settling
+state - reading the cache register before page data has settled, or
+issuing the next command before a program/erase has actually completed.
+
+Add a SPINAND_HAS_PREMATURE_READY_BIT quirk flag and a SkyHigh-specific
+wait that wraps spinand_wait() and only trusts readiness once a
+follow-up status read still shows non-busy, retrying otherwise (bounded
+by the existing SPINAND_WAITRDY_TIMEOUT_MS). SkyHigh's ->init() sets the
+flag; the read, write and erase paths use the wait for any manufacturer
+that sets it, everyone else keeps using spinand_wait() unchanged.
+
+Not yet confirmed to eliminate the bit-flips over a long soak on
+hardware - posted as a workaround for the observed behavior.
+
+Suggested-by: Mikhail Zhilkin <csharper2005@gmail.com>
+Signed-off-by: Vitaliy Sochnev <sochnev.v.74@gmail.com>
+---
+ drivers/mtd/nand/spi/core.c    | 49 ++++++++++++++++++++++++++++++++++-------
+ drivers/mtd/nand/spi/skyhigh.c |  2 ++
+ include/linux/mtd/spinand.h    |  1 +
+ 3 files changed, 45 insertions(+), 7 deletions(-)
+
+--- a/drivers/mtd/nand/spi/core.c
++++ b/drivers/mtd/nand/spi/core.c
+@@ -613,6 +613,37 @@ static int spinand_lock_block(struct spi
+ 	return spinand_write_reg_op(spinand, REG_BLOCK_LOCK, lock);
+ }
+ 
++/*
++ * SkyHigh SPI-NAND chips have been observed to flip the ready bit back
++ * to busy shortly after first reporting ready. Wrap spinand_wait() -
++ * which still does the normal delay/poll and any hardware
++ * ->poll_status() offload - and only trust its result once a follow-up
++ * status read confirms the chip is still non-busy, retrying otherwise.
++ */
++static int spinand_skyhigh_wait_ready(struct spinand_device *spinand,
++				      unsigned long initial_delay_us,
++				      unsigned long poll_delay_us, u8 *s)
++{
++	unsigned long timeo = jiffies + msecs_to_jiffies(SPINAND_WAITRDY_TIMEOUT_MS);
++	u8 status;
++	int ret;
++
++	do {
++		ret = spinand_wait(spinand, initial_delay_us, poll_delay_us,
++				   &status);
++		if (ret < 0)
++			return ret;
++
++		ret = spinand_read_status(spinand, &status);
++		if (ret)
++			return ret;
++	} while ((status & STATUS_BUSY) && time_before(jiffies, timeo));
++
++	*s = status;
++
++	return status & STATUS_BUSY ? -ETIMEDOUT : 0;
++}
++
+ /**
+  * spinand_read_page() - Read a page
+  * @spinand: the spinand device
+@@ -636,10 +667,16 @@ int spinand_read_page(struct spinand_dev
+ 	if (ret)
+ 		return ret;
+ 
+-	ret = spinand_wait(spinand,
+-			   SPINAND_READ_INITIAL_DELAY_US,
+-			   SPINAND_READ_POLL_DELAY_US,
+-			   &status);
++	if (spinand->flags & SPINAND_HAS_PREMATURE_READY_BIT)
++		ret = spinand_skyhigh_wait_ready(spinand,
++						  SPINAND_READ_INITIAL_DELAY_US,
++						  SPINAND_READ_POLL_DELAY_US,
++						  &status);
++	else
++		ret = spinand_wait(spinand,
++				   SPINAND_READ_INITIAL_DELAY_US,
++				   SPINAND_READ_POLL_DELAY_US,
++				   &status);
+ 	if (ret < 0)
+ 		return ret;
+ 
+@@ -683,10 +720,16 @@ int spinand_write_page(struct spinand_de
+ 	if (ret)
+ 		return ret;
+ 
+-	ret = spinand_wait(spinand,
+-			   SPINAND_WRITE_INITIAL_DELAY_US,
+-			   SPINAND_WRITE_POLL_DELAY_US,
+-			   &status);
++	if (spinand->flags & SPINAND_HAS_PREMATURE_READY_BIT)
++		ret = spinand_skyhigh_wait_ready(spinand,
++						  SPINAND_WRITE_INITIAL_DELAY_US,
++						  SPINAND_WRITE_POLL_DELAY_US,
++						  &status);
++	else
++		ret = spinand_wait(spinand,
++				   SPINAND_WRITE_INITIAL_DELAY_US,
++				   SPINAND_WRITE_POLL_DELAY_US,
++				   &status);
+ 	if (ret)
+ 		return ret;
+ 
+@@ -1070,10 +1113,16 @@ static int spinand_erase(struct nand_dev
+ 	if (ret)
+ 		return ret;
+ 
+-	ret = spinand_wait(spinand,
+-			   SPINAND_ERASE_INITIAL_DELAY_US,
+-			   SPINAND_ERASE_POLL_DELAY_US,
+-			   &status);
++	if (spinand->flags & SPINAND_HAS_PREMATURE_READY_BIT)
++		ret = spinand_skyhigh_wait_ready(spinand,
++						  SPINAND_ERASE_INITIAL_DELAY_US,
++						  SPINAND_ERASE_POLL_DELAY_US,
++						  &status);
++	else
++		ret = spinand_wait(spinand,
++				   SPINAND_ERASE_INITIAL_DELAY_US,
++				   SPINAND_ERASE_POLL_DELAY_US,
++				   &status);
+ 
+ 	if (!ret && (status & STATUS_ERASE_FAILED))
+ 		ret = -EIO;
+--- a/drivers/mtd/nand/spi/skyhigh.c
++++ b/drivers/mtd/nand/spi/skyhigh.c
+@@ -125,6 +125,8 @@ static const struct spinand_info skyhigh
+ 
+ static int skyhigh_spinand_init(struct spinand_device *spinand)
+ {
++	spinand->flags |= SPINAND_HAS_PREMATURE_READY_BIT;
++
+ 	/*
+ 	 * Config_Protect_En (bit 1 in Block Lock register) must be set to 1
+ 	 * before writing other bits. Do it here before core unlocks all blocks
+--- a/include/linux/mtd/spinand.h
++++ b/include/linux/mtd/spinand.h
+@@ -418,6 +418,7 @@ struct spinand_ecc_info {
+ #define SPINAND_HAS_PROG_PLANE_SELECT_BIT		BIT(2)
+ #define SPINAND_HAS_READ_PLANE_SELECT_BIT		BIT(3)
+ #define SPINAND_NO_RAW_ACCESS				BIT(4)
++#define SPINAND_HAS_PREMATURE_READY_BIT		BIT(5)
+ 
+ /**
+  * struct spinand_ondie_ecc_conf - private SPI-NAND on-die ECC engine structure
--- a/target/linux/airoha/patches-6.18/901-snand-mtk-bmt-support.patch
+++ b/target/linux/airoha/patches-6.18/901-snand-mtk-bmt-support.patch
@@ -8,7 +8,7 @@
  
  int spinand_read_reg_op(struct spinand_device *spinand, u8 reg, u8 *val)
  {
-@@ -1719,6 +1720,7 @@ static int spinand_probe(struct spi_mem
+@@ -1768,6 +1769,7 @@ static int spinand_probe(struct spi_mem
  	if (ret)
  		return ret;
  
@@ -16,7 +16,7 @@
  	ret = mtd_device_register(mtd, NULL, 0);
  	if (ret)
  		goto err_spinand_cleanup;
-@@ -1726,6 +1728,7 @@ static int spinand_probe(struct spi_mem
+@@ -1775,6 +1777,7 @@ static int spinand_probe(struct spi_mem
  	return 0;
  
  err_spinand_cleanup:
@@ -24,7 +24,7 @@
  	spinand_cleanup(spinand);
  
  	return ret;
-@@ -1744,6 +1747,7 @@ static int spinand_remove(struct spi_mem
+@@ -1793,6 +1796,7 @@ static int spinand_remove(struct spi_mem
  	if (ret)
  		return ret;
  
--- a/target/linux/econet/patches-6.18/902-snand-mtk-bmt-support.patch
+++ b/target/linux/econet/patches-6.18/902-snand-mtk-bmt-support.patch
@@ -8,7 +8,7 @@
  
  int spinand_read_reg_op(struct spinand_device *spinand, u8 reg, u8 *val)
  {
-@@ -1719,6 +1720,7 @@ static int spinand_probe(struct spi_mem
+@@ -1768,6 +1769,7 @@ static int spinand_probe(struct spi_mem
  	if (ret)
  		return ret;
  
@@ -16,7 +16,7 @@
  	ret = mtd_device_register(mtd, NULL, 0);
  	if (ret)
  		goto err_spinand_cleanup;
-@@ -1726,6 +1728,7 @@ static int spinand_probe(struct spi_mem
+@@ -1775,6 +1777,7 @@ static int spinand_probe(struct spi_mem
  	return 0;
  
  err_spinand_cleanup:
@@ -24,7 +24,7 @@
  	spinand_cleanup(spinand);
  
  	return ret;
-@@ -1744,6 +1747,7 @@ static int spinand_remove(struct spi_mem
+@@ -1793,6 +1796,7 @@ static int spinand_remove(struct spi_mem
  	if (ret)
  		return ret;
  
--- a/target/linux/mediatek/patches-6.18/010-v7.0-mtd-spinand-add-support-for-Dosilicon-DS35Q1GA-DS35M.patch
+++ b/target/linux/mediatek/patches-6.18/010-v7.0-mtd-spinand-add-support-for-Dosilicon-DS35Q1GA-DS35M.patch
@@ -42,7 +42,7 @@ Signed-off-by: Miquel Raynal <miquel.raynal@bootlin.com>
  obj-$(CONFIG_MTD_SPI_NAND) += spinand.o
 --- a/drivers/mtd/nand/spi/core.c
 +++ b/drivers/mtd/nand/spi/core.c
-@@ -1226,6 +1226,7 @@ static const struct nand_ops spinand_ops
+@@ -1275,6 +1275,7 @@ static const struct nand_ops spinand_ops
  static const struct spinand_manufacturer *spinand_manufacturers[] = {
  	&alliancememory_spinand_manufacturer,
  	&ato_spinand_manufacturer,
--- a/target/linux/mediatek/patches-6.18/121-hack-spi-nand-1b-bbm.patch
+++ b/target/linux/mediatek/patches-6.18/121-hack-spi-nand-1b-bbm.patch
@@ -11,7 +11,7 @@ Signed-off-by: Chuanhong Guo <gch981213@gmail.com>
 ---
 --- a/drivers/mtd/nand/spi/core.c
 +++ b/drivers/mtd/nand/spi/core.c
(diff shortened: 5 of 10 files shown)


target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
```diff
@@ -0,0 +1,102 @@
...
++		if (ret)
++			return ret;
++
++		if (status & STATUS_BUSY)
++			continue;
++
++		ret = spinand_read_status(spinand, &status);
++		if (ret)
++			return ret;
++	} while ((status & STATUS_BUSY) && time_before(jiffies, timeo));
```
openwrt-ai, line 75: This loop fires back-to-back `GET_FEATURE` ops with no delay at all, for up to 400 ms. The code path it replaces sleeps: the software fallback in `spi_mem_poll_status()` does `udelay(initial_delay_us)` and then polls via `read_poll_timeout()` with `polling_delay_us` between samples — [spi-mem.c:1004-1013](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/spi/spi-mem.c#L1004-L1013). Two consequences:

- A page load takes tens of µs, so instead of ~one status read every 5 µs this issues as many status transactions as the SPI bus can carry, on *every* page read of *every* SkyHigh chip. That is a measurable read-throughput and CPU cost, not just a latency change.
- Nothing in the loop sleeps or calls `cond_resched()`, and the timeout is `jiffies`-based. With a synchronous SPI controller driver, a chip that is genuinely stuck busy will spin a CPU for the full `SPINAND_WAITRDY_TIMEOUT_MS` window rather than yielding, which `spinand_wait()` does not do today.

Please add the initial delay before the first sample and `SPINAND_READ_POLL_DELAY_US` between iterations. An alternative that keeps the existing fast path entirely is to call `spinand_wait()` as-is and then only accept the result once a follow-up `spinand_read_status()` still reads non-busy, retrying `spinand_wait()` otherwise — that reuses the hardware `->poll_status()` offload where the controller has one, which this open-coded loop bypasses.

Note that changing the line count here invalidates the patch's `@@ -613,6 +613,36 @@` hunk header and the `43 insertions(+), 6 deletions(-)` diffstat, so the patch needs regenerating afterwards (`make target/linux/refresh` with a 6.18 target selected).

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
```diff
@@ -0,0 +1,102 @@
...
+ 		return ret;
+ 
+-	ret = spinand_wait(spinand,
+-			   SPINAND_READ_INITIAL_DELAY_US,
+-			   SPINAND_READ_POLL_DELAY_US,
+-			   &status);
+-	if (ret < 0)
+-		return ret;
++	if (spinand->id.data[0] == 0x01) {
++		/* SPINAND_MFR_SKYHIGH, see drivers/mtd/nand/spi/skyhigh.c */
```
openwrt-ai: Open-coding the raw manufacturer byte here is fragile — it duplicates a constant that only exists inside `skyhigh.c`, and it relies on `spinand->id.data[0]` still holding the matched ID (true today, but `spinand_id_detect()` overwrites that buffer on each RDID method attempt, and qualcommax's `0412-mtd-spinand-qpic-only-support-max-4-bytes-ID.patch` in this same PR touches that matching path).

`spinand->manufacturer` is set by `spinand_manufacturer_match()` during `spinand_detect()`, well before any page read, and `skyhigh_spinand_manufacturer` is already declared at [spinand.h:370](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/include/linux/mtd/spinand.h#L370), so a direct pointer compare is both safer and self-documenting:

```suggestion
+	if (spinand->manufacturer == &skyhigh_spinand_manufacturer) {
+		/* SkyHigh, see drivers/mtd/nand/spi/skyhigh.c */
```

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
```diff
@@ -0,0 +1,102 @@
...
++		ret = spinand_skyhigh_wait_ready(spinand, &status);
```
openwrt-ai: Is confining the workaround to the read path deliberate? If the SkyHigh ready bit really drops back to busy after first reporting ready, the same hazard exists on the program and erase paths — [`spinand_write_page()` core.c:686](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mtd/nand/spi/core.c#L686) and [`spinand_erase()` core.c:1073](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mtd/nand/spi/core.c#L1073) both use the unmodified `spinand_wait()`.

That matters for the diagnosis, not just for coverage: a premature ready on the *read* path makes the cache read return wrong data on that read only, so it should look like frequent, transient, non-reproducible ECC events. The log in the commit message shows UBI scrubbing PEBs and the flips not recurring on the destination PEB, which reads more like marginal cells left behind by a program that returned before the write settled. Worth saying in the commit message why read-only coverage is the right scope, or extending it.

(FWIW the continuous-read path at [core.c:805](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mtd/nand/spi/core.c#L805) has the same load-then-read-cache sequence, but `skyhigh.c` sets no `->set_cont_read`, so it is unreachable for these chips — no change needed there.)

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
```diff
@@ -0,0 +1,102 @@
...
+
+  ubi0: fixable bit-flip detected at PEB 1316
+  ubi0: scrubbed PEB 1316 (LEB 5:195), data moved to PEB 399
+  ubi0: fixable bit-flip detected at PEB 511
+  ubi0: schedule PEB 511 for scrubbing
+
+spinand_wait() already re-reads the status register once "just in case
+the STATUS_READY bit has changed since our last check", but that extra
+read only fires off the very first not-busy sample reported by
+spi_mem_poll_status() - it never re-verifies that the chip stays ready.
```
openwrt-ai: nit: this describes `spinand_wait()`'s extra read backwards. Per [core.c:563-571](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mtd/nand/spi/core.c#L563-L571), the extra `spinand_read_status()` only runs when the *buffered* status still reads busy — it is the fallback for controllers whose hardware `->poll_status()` never wrote into `spinand->scratchbuf`. When `spi_mem_poll_status()` reports non-busy through the scratchbuf, `spinand_wait()` takes the `goto out` and does no extra read at all.

The claim you actually need is simpler and correct as written in the next sentence: `spinand_wait()` never re-verifies that the chip is *still* ready after it first observes non-busy. Since this patch is destined for upstream (it sits in `generic/pending-6.18/`), it's worth getting that paragraph right before it goes to linux-mtd.

---

openwrt-ai (commented):
Reviewed 1 new commit against v6.18.44 (`KERNEL_PATCHVER:=6.18` for airoha/econet/mediatek/qualcommax).

The patch-refresh side of this looks correct: the new patch nets +37 lines in `drivers/mtd/nand/spi/core.c`, and every hunk header in the nine refreshed target patches shifts by exactly +37. I also checked that no other target carries a `patches-6.18` patch touching `core.c` — the two `generic/backport-6.18/` and one `generic/pending-6.18/487-*` patches that touch it all apply before `489-*` and are correctly left alone. Slot 489 is free.

Four inline comments on the new patch: one on the unthrottled polling loop, one on the manufacturer check, one question on why the workaround covers only the read path, and one on the commit message's description of `spinand_wait()`.

---

geokvant:
5 days uptime with this patch
Zero warnings in my case

[    2.332309] ubi0: PEB size: 131072 bytes (128 KiB), LEB size: 126976 bytes
[    2.352950] ubi0: good PEBs: 2047, bad PEBs: 0, corrupted PEBs: 0
[    2.375399] ubi0: available PEBs: 0, total reserved PEBs: 2047, PEBs reserved for bad PEB handling: 40

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
```diff
@@ -0,0 +1,168 @@
...
+ 
+ 	if (!ret && (status & STATUS_ERASE_FAILED))
+ 		ret = -EIO;
+--- a/drivers/mtd/nand/spi/skyhigh.c
++++ b/drivers/mtd/nand/spi/skyhigh.c
+@@ -125,6 +125,8 @@ static const struct spinand_info skyhigh
+ 
+ static int skyhigh_spinand_init(struct spinand_device *spinand)
+ {
++	spinand->flags |= SPINAND_HAS_PREMATURE_READY_BIT;
```
openwrt-ai: `spinand->flags` is written in exactly one place in the tree — [core.c:1415](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mtd/nand/spi/core.c#L1415), from the chip table — and no manufacturer `->init()` ORs into it, so this introduces a new pattern that linux-mtd will likely push back on. Put the quirk in the four `SPINAND_INFO()` entries instead, turning [`SPINAND_NO_RAW_ACCESS` at skyhigh.c:91](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/drivers/mtd/nand/spi/skyhigh.c#L91) and its three siblings into `SPINAND_NO_RAW_ACCESS | SPINAND_HAS_PREMATURE_READY_BIT`; that also scopes it per part if only some S35ML* dies are affected.

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
```diff
@@ -0,0 +1,168 @@
...
++/*
++ * SkyHigh SPI-NAND chips have been observed to flip the ready bit back
++ * to busy shortly after first reporting ready. Wrap spinand_wait() -
++ * which still does the normal delay/poll and any hardware
++ * ->poll_status() offload - and only trust its result once a follow-up
++ * status read confirms the chip is still non-busy, retrying otherwise.
++ */
++static int spinand_skyhigh_wait_ready(struct spinand_device *spinand,
++				      unsigned long initial_delay_us,
++				      unsigned long poll_delay_us, u8 *s)
```
openwrt-ai: `spinand_skyhigh_wait_ready()` lives in generic `core.c` and runs for any manufacturer that sets `SPINAND_HAS_PREMATURE_READY_BIT`, so both the manufacturer-specific name and the SkyHigh-only comment above it are wrong for what the code does. Rename it neutrally (e.g. `spinand_wait_ready_confirmed()`) and move the `spinand->flags &` test inside it, which also collapses the three triplicated `if`/`else` call sites back to a single call each.

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
```diff
@@ -0,0 +1,168 @@
...
+ 	/*
+ 	 * Config_Protect_En (bit 1 in Block Lock register) must be set to 1
+ 	 * before writing other bits. Do it here before core unlocks all blocks
+--- a/include/linux/mtd/spinand.h
++++ b/include/linux/mtd/spinand.h
+@@ -418,6 +418,7 @@ struct spinand_ecc_info {
+ #define SPINAND_HAS_PROG_PLANE_SELECT_BIT		BIT(2)
+ #define SPINAND_HAS_READ_PLANE_SELECT_BIT		BIT(3)
+ #define SPINAND_NO_RAW_ACCESS				BIT(4)
++#define SPINAND_HAS_PREMATURE_READY_BIT		BIT(5)
```
openwrt-ai: nit: two tabs puts `BIT(5)` at column 48, while the four flags above it all land at column 56 — [spinand.h:413-417](https://github.com/gregkh/linux/blob/1efe5d048a391de3ead2804b2e7f86376c356cc5/include/linux/mtd/spinand.h#L413-L417). Needs three tabs.

```suggestion
+#define SPINAND_HAS_PREMATURE_READY_BIT			BIT(5)
```

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
```diff
@@ -0,0 +1,168 @@
...
+Not yet confirmed to eliminate the bit-flips over a long soak on
+hardware - posted as a workaround for the observed behavior.
+
+Suggested-by: Mikhail Zhilkin <csharper2005@gmail.com>
+Signed-off-by: Vitaliy Sochnev <sochnev.v.74@gmail.com>
+---
+ drivers/mtd/nand/spi/core.c    | 49 ++++++++++++++++++++++++++++++++++-------
+ drivers/mtd/nand/spi/skyhigh.c |  2 ++
+ include/linux/mtd/spinand.h    |  1 +
+ 3 files changed, 45 insertions(+), 7 deletions(-)
```
openwrt-ai: nit: the diffstat header is stale — applying this patch to v6.18.44 gives 64 insertions / 12 deletions (73 changed lines in `core.c`), not 45/7. `make target/linux/refresh` regenerates the hunks but not this block, so it needs fixing by hand before the patch goes to linux-mtd.

```suggestion
 drivers/mtd/nand/spi/core.c    | 73 +++++++++++++++++++++++++++++++++++-------
 drivers/mtd/nand/spi/skyhigh.c |  2 ++
 include/linux/mtd/spinand.h    |  1 +
 3 files changed, 64 insertions(+), 12 deletions(-)
```

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
```diff
@@ -0,0 +1,168 @@
...
+  * @spinand: the spinand device
+@@ -636,10 +667,16 @@ int spinand_read_page(struct spinand_dev
+ 	if (ret)
+ 		return ret;
+ 
+-	ret = spinand_wait(spinand,
+-			   SPINAND_READ_INITIAL_DELAY_US,
+-			   SPINAND_READ_POLL_DELAY_US,
+-			   &status);
++	if (spinand->flags & SPINAND_HAS_PREMATURE_READY_BIT)
```
openwrt-ai: nit: the PR description still describes this as keyed on "manufacturer ID `0x01`" and used "from `spinand_read_page()` for this manufacturer only" — both are false at this head. Worth syncing it with the commit message before the PR leaves draft.

---

openwrt-ai (commented):
Reviewed 1 new commit against v6.18.44.

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
openwrt-ai: fixed, thanks — solved by the alternative, wrapping `spinand_wait()` instead of open-coding the poll.

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
openwrt-ai: fixed, thanks — solved by dropping the ID check for a `SPINAND_HAS_PREMATURE_READY_BIT` flag.

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
openwrt-ai: fixed, thanks — the write and erase paths are covered now too.

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
openwrt-ai: fixed, thanks.

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
```diff
@@ -0,0 +1,209 @@
...
+spinand->scratchbuf - it is not a second confirmation of readiness.
+
+On SkyHigh chips the ready bit has been observed to flip back to busy
+shortly after first going ready, so a caller can act on stale/settling
+state - reading the cache register before page data has settled, or
+issuing the next command before a program/erase has actually completed.
+
+Add a SPINAND_HAS_PREMATURE_READY_BIT flag and set it on the SkyHigh
+S35ML* entries. Route the page load (regular and continuous read),
+program and erase waits through a new spinand_wait_ready_confirmed()
```
openwrt-ai: [`spinand_reset_op()` at core.c:605](https://github.com/gregkh/linux/blob/8f3741e6feb045da5b406df0a80b42a1adfb289b/drivers/mtd/nand/spi/core.c#L605) still uses plain `spinand_wait()`, so the resume path ([core.c:1560](https://github.com/gregkh/linux/blob/8f3741e6feb045da5b406df0a80b42a1adfb289b/drivers/mtd/nand/spi/core.c#L1560)) acts on an unconfirmed ready after RESET. Is leaving it out deliberate? (The probe-time reset at [core.c:1458](https://github.com/gregkh/linux/blob/8f3741e6feb045da5b406df0a80b42a1adfb289b/drivers/mtd/nand/spi/core.c#L1458) runs inside `spinand_detect()` before `spinand->flags` is assigned, so it cannot be covered either way.)

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
```diff
@@ -0,0 +1,209 @@
...
+ 		     SPINAND_ECCINFO(&skyhigh_spinand_ooblayout,
+ 				     skyhigh_spinand_ecc_get_status)),
+ };
+--- a/include/linux/mtd/spinand.h
++++ b/include/linux/mtd/spinand.h
+@@ -418,6 +418,7 @@ struct spinand_ecc_info {
+ #define SPINAND_HAS_PROG_PLANE_SELECT_BIT		BIT(2)
+ #define SPINAND_HAS_READ_PLANE_SELECT_BIT		BIT(3)
+ #define SPINAND_NO_RAW_ACCESS				BIT(4)
++#define SPINAND_HAS_PREMATURE_READY_BIT			BIT(5)
```
openwrt-ai: nit: the `SPINAND_HAS_*_BIT` names in this header all mean "the chip's registers contain this bit" ([spinand.h:413-417](https://github.com/gregkh/linux/blob/8f3741e6feb045da5b406df0a80b42a1adfb289b/include/linux/mtd/spinand.h#L413-L417)), while this one means "the chip's ready bit misbehaves" — a quirk, which `SPINAND_NO_RAW_ACCESS` shows is named differently. Something like `SPINAND_PREMATURE_READY_QUIRK` would not read as a feature-bit flag; the rename also touches the helper, the four `skyhigh.c` entries and the patch's commit message.

---

openwrt-ai (commented):
## Commit checks

- `ecadad61` "kernel: workaround premature ready bit on SkyHigh SPI-NAND chips" — "bounded by the existing SPINAND_WAITRDY_TIMEOUT_MS" overstates the bound: `timeo` only gates loop re-entry, so a `spinand_wait()` started just inside the deadline can add another full 400 ms, making the worst case ~2x. The patch header carries the same claim and needs the same correction.

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
openwrt-ai: fixed, thanks.

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
openwrt-ai: fixed, thanks.

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
openwrt-ai: fixed, thanks.

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
openwrt-ai: fixed, thanks.

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
openwrt-ai: fixed, thanks.

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
```diff
@@ -0,0 +1,234 @@
...
+
+Add a SPINAND_PREMATURE_READY_QUIRK flag and set it on the SkyHigh
+S35ML* entries. Route the reset, page load (regular and continuous
+read), program and erase waits through a new
+spinand_wait_ready_confirmed() helper: for flagged parts it only trusts
+a ready result from spinand_wait() once a follow-up status read still
+shows the device ready, and calls spinand_wait() again otherwise. Each
+spinand_wait() gets its own SPINAND_WAITRDY_TIMEOUT_MS budget and no new
+one is started once that span has elapsed, so a flagged part can spend
+up to roughly twice that time waiting. For everything else the helper is
```
openwrt-ai, line 36: nit: the PR description has drifted from this paragraph again on three points — it still names the flag `SPINAND_HAS_PREMATURE_READY_BIT`, it lists only page load/program/erase (not reset), and it still says the retry is "bounded by `SPINAND_WAITRDY_TIMEOUT_MS`" rather than roughly twice it. Worth syncing before the PR leaves draft.

---

openwrt-ai (commented):
Reviewed 1 new commit against v6.18.52.

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
openwrt-ai: fixed, thanks.

---

target/linux/generic/pending-6.18/489-mtd-spinand-skyhigh-workaround-premature-ready-bit.patch
openwrt-ai: fixed, thanks.

---

csharper2005 (approved):

PR #24852, opened 2026-08-22
mediatek: drop unused kmod-mt7916-firmware for JioRouter AX6000 JIDU6101

SahilB69:
Drop the unused kmod-mt7916-firmware package from DEVICE_PACKAGES.

The JioRouter AX6000 (JIDU6101) does not have a PCIe radio and relies
solely on the MT7986 SoC integrated Wi-Fi, which is covered by
kmod-mt7986-firmware.

### DIFF
--- a/target/linux/mediatek/image/filogic.mk
+++ b/target/linux/mediatek/image/filogic.mk
@@ -2216,7 +2216,7 @@ define Device/jiorouter_ax6000-jidu6101
   DEVICE_VARIANT := JIDU6101
   DEVICE_DTS := mt7986a-jiorouter-ax6000-jidu6101
   DEVICE_DTS_DIR := ../dts
-  DEVICE_PACKAGES := kmod-usb3 kmod-mt7915e kmod-mt7916-firmware kmod-mt7986-firmware mt7986-wo-firmware
+  DEVICE_PACKAGES := kmod-usb3 kmod-mt7915e kmod-mt7986-firmware mt7986-wo-firmware
   UBINIZE_OPTS := -E 5
   UBOOTENV_IN_UBI := 1
   BLOCKSIZE := 128k


neo6699new (approved):

ctpt14378 (approved):

Baibhav253 (approved):

target/linux/mediatek/image/filogic.mk
```diff
@@ -2216,7 +2216,7 @@ define Device/jiorouter_ax6000-jidu6101
   DEVICE_VARIANT := JIDU6101
   DEVICE_DTS := mt7986a-jiorouter-ax6000-jidu6101
   DEVICE_DTS_DIR := ../dts
-  DEVICE_PACKAGES := kmod-usb3 kmod-mt7915e kmod-mt7916-firmware kmod-mt7986-firmware mt7986-wo-firmware
+  DEVICE_PACKAGES := kmod-usb3 kmod-mt7915e kmod-mt7986-firmware mt7986-wo-firmware
```
openwrt-ai, line 2219: nit: the removal itself looks right, but the justification in the commit message is backwards. `kmod-mt7916-firmware` explicitly [`DEPENDS+=+kmod-mt7915e`](https://github.com/openwrt/openwrt/blob/a6356ceb18d4ff1d5087e926222e14d8e022e4a4/package/kernel/mt76/Makefile#L243-L247), so "the device uses the MT7915E driver" is not a reason to drop the MT7916 firmware — MT7916 cards use that same driver.

The actual reason the package is dead weight here is that the board has no PCIe radio at all: `mt7986a-jiorouter-common.dtsi` has no `&pcie`/`&pcie_phy` node and only enables the SoC's built-in [`&wifi`](https://github.com/openwrt/openwrt/blob/a6356ceb18d4ff1d5087e926222e14d8e022e4a4/target/linux/mediatek/dts/mt7986a-jiorouter-common.dtsi#L276-L282) (DBDC, PHY0/PHY1 only, matching the two `PHYNBR` cases in [11_fix_wifi_mac](``https://github.com/openwrt/openwrt/blob/a6356ceb18d4ff1d5087e926222e14d8e022e4a4/target/linux/mediatek/filogic/base-files/etc/hotplug.d/ieee80211/11_fix_wifi_mac#L174-L178)),`` which is served by `kmod-mt7986-firmware`. The resulting package list also matches what the other PCIe-less MT7986 boards in this file already use (e.g. lines 3340 and 3476). Consider rewording the second paragraph along those lines so the rationale holds up.

---

openwrt-ai (commented):
Reviewed 1 new commit. The change itself checks out: the board has no PCIe radio, so `kmod-mt7916-firmware` installs blobs nothing on this device loads, and dropping it does not break any dependency (`kmod-mt7986-firmware` / `mt7986-wo-firmware` are independent packages). One nit inline about the wording of the commit-message rationale.

---

Aditya-750:
Is this extra driver actually causing any noticeable performance degradation, higher RAM usage, or unnecessary CPU load? If it isn’t causing any real issue, what’s the point of removing it? If it were actually needed, the PR author would have raised it themselves.

